Issue #5157, opened 2019-02-09
Strange Streams behavior


scheibo:
Something fishy seems to be occurring with [lib/streams.js](https://github.com/Zarel/Pokemon-Showdown/blob/master/lib/streams.js) where code after `while ((chunk = await stream.read()) {}` does not get executed. I noticed this when trying to run [sim/battle-stream-example.js](https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/battle-stream-example.js), but a more minimal example of the bug can be found below:

```javascript
'use strict';
const Streams = require('../lib/streams');

class FooStream extends Streams.ObjectReadWriteStream {
	_write(message) {}
	_end() { this.push(null); }
}

function getStreams(stream) {
	let omni = new Streams.ObjectReadWriteStream({
		write(data) {
			stream.write(data);
		}
	});

	(async() => {
		console.log('before logged');
		let chunk;
		while ((chunk = await stream.read())) {
			if (chunk !== 'end') {
				omni.push(chunk);
			}
		}
		console.log('never logged');
	})();

	return {omni};
}

let fooStream = new FooStream();
const streams = getStreams(fooStream);

(async() => {
	console.log('also logged');
	let chunk;
	while ((chunk = await streams.omni.read())) {
		console.log(chunk);
	}
	console.log('also never logged');
})();

console.log('start');
for (let i = 0; i < 10; i++) {
	fooStream.push(`data ${i}`);
}
fooStream.push('end');

console.log('stop');
```

When run, this code outputs:

```
before logged
also logged
start
stop
data 0
data 1
data 2
data 3
data 4
data 5
data 6
data 7
data 8
data 9
```

With `never logged` and `also never logged` not appearing. If I replace the `await streams.read()` method with a different async method the code after the `while` will execute, so something Streams is doing seems to be to blame here. I tried single stepping through streams with a debugger but couldn't garner any additional insight.

Zarel:
So, a stream can be thought of as a string (or, in the case of an object string, an array) that you go through asynchronously.

But the array or string doesn't end until you tell it to end. With the current API, you tell the stream to end by sending `.push(null)`. This says "we've reached the end of the string/array".

So say we start an object stream. We do `stream.push("data 0"); stream.push("data 1");`

This is kind of like the array `["data 0", "data 1", ...]`

[...]

PR #5181, opened 2019-02-20
Optimize Dex#getEffect by caching the result.

scheibo:
I've been taking a look at optimizing the simulator (see [pkmn.cc/optimize](https://pkmn.cc/optimize) - comments welcome), and `getEffect` seems ripe for improvements. I can't confirm this caching is 100% safe - I don't know if `Effect`s are supposed to be immutable or not, but the 'tests pass', for whatever that's worth. There's already caching for `Move`/`Ability`/`Item` etc, but `getEffect` only very rarely hits those code paths, so it's actually missing out on most of the benefit. 

This change should result in a **~30% increase in overall battle performance** over repeated runs, but if this caching would break things theres still some other options for restructuring the method for performance, as outlined in the doc. I think there's more performance gains to be had by refactoring `getRelevantEffects`, but it would require a substantially larger change architecturally, so I'd hope to get some buy in before tackling that.

PR #5278, opened 2019-03-09
Add benchmarking/profiling support to dev-tools/harness.js.

scheibo:
Not sure if you'd want this committed, but I'm trying to update my testing harness so I can get back to optimizing the simulator and I'm encountering issues, specifically what appears to be #5157 again - [`BattleStream#_end`](https://github.com/Zarel/Pokemon-Showdown/blob/fc95f6e7d96db09fa2896bbd595b15327c2d112e/sim/battle-stream.ts#L151) never gets called if I add logging to it, and my harness only runs a single battle before exiting. Maybe I'm still 'holding it wrong', but my old code works against https://github.com/pkmn-cc/Pokemon-Showdown/tree/profile or 94b2d65 give or take, so maybe I broke myself with all my refactoring?


Zarel:
This pull request **introduces 2 alerts** when merging c831027a9fdb7e6d38da57be1953b7a7bfcfef08 into fc95f6e7d96db09fa2896bbd595b15327c2d112e - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-56ca808d2a5ecf92598fb8d4ee60a9c54d7a4f8e)

**new alerts:**

---

scheibo:
Actually yeah, the stream is not being closed, its not my harness code thats bad. See https://github.com/Zarel/Pokemon-Showdown/issues/5157#issuecomment-471137961

Zarel:
I hope assert failures would report the random seed that failed?

scheibo:
> I hope assert failures would report the random seed that failed?

*quoi?*

I've changed random approach away from `--random-seed=<SEED>` to `PRNG` anyway for simplicity, but I still don't understand your question.

---

Anyway, I've got this working now without timing. I'm going to add coarse timing statistics here (around the whole battle, not in the individual methods. Without an optimizing compiler like Closure to strip it, I can't leave instrumentation code around as it would incur function call overhead). **Do you think you'd be willing to let me check something like this in?** I think with some documentation it will make it easier for other devs to benchmark their changes + it can also be used for some sort of smoke test if we wanted (run a bunch of battles in see if anything crashes).

I've beefed up the `RandomPlayerAI` which might make it more intimidating as an example, but now it can at least actually exercise a lot more functionality and create more realistic playouts.

Zarel:
This pull request **introduces 2 alerts** when merging 97d00ebcfc9fa18f98e8cfafc5595a03ef1f74b2 into 9e7ba68fc4caedf503638bd7edda6b781df18d65 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-2cb15c2baa60afb0dfcd3b1a28d00f6912ec92b9)

**new alerts:**

---

Zarel:
> _quoi?_

Like, if a test fails, I should hope that the failing test reports the random seed causing it to fail, so that the failure is reproducible.

scheibo:
>  so that the failure is reproducible.

Where's the fun in that?

But yeah, the code is currently deterministic, so you if you get a bug with one run, a subsequent run with the same seed will give you the same bug. However, it can be improved to be more useful (instead of having to potentially run M/N battles before reencountering the error...).

However, there's no 'test' or 'failure' really as it currently stands, the code either runs without errors (and will optionally dump timing statistics) or throws and exits unsuccessfully. However, if we wanted to use it for smoke testing purposes (if it were say `require`d instead of run from the command line), then yeah, having it track which seeds caused issues would be useful. I'll work on that.

Zarel:
This pull request **introduces 2 alerts** when merging 8c427a052538b786c7d2b651649cd885509681e4 into 1b2aeb696b37f79da61aa74b484f15923157a9cc - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-2c8adba99c9048e7abbbeb9f45c312ed8980d2f2)

**new alerts:**

---

Zarel:
This pull request **introduces 2 alerts** when merging d8b4ae6a96b624154f204900fe868833e89dc8fe into 8fec7543ed34634a47b753fafbbfa0db524c822f - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-64fbbd6559fd15aec0fb9c337c6f2c85210652d9)

**new alerts:**

---

dev-tools/harness.js
```diff
@@ -0,0 +1,156 @@
...
+
+'use strict';
+
+const BattleStreams = require('../.sim-dist/battle-stream');
+const Dex = require('../.sim-dist/dex');
+const PRNG = require('../.sim-dist/prng').PRNG;
+const RandomPlayerAI = require('../.sim-dist/examples/random-player-ai').RandomPlayerAI;
+const Streams = require('../.lib-dist/streams');
+
+const DEFAULT_SEED = [0x09917, 0x06924, 0x0e1c8, 0x06af0];
```
Slayer95: I am afraid this is a very boring seed.

IIRC, calling ``.sample(array)`` will first result in picking the first element at least 12 consecutive times for ``.length < 8``. Therefore, the tested formats will be highly biased, as well as the chosen moves.

Though maybe I am worrying over something of no actual relevance.

dev-tools/harness.js
```diff
@@ -0,0 +1,156 @@
...
+				games = [];
+			}
+
+			let seed = this.prng.seed;
+			try {
+				const game = this.runGame(format);
+				if (!this.async) await game;
+				games.push(game);
+			} catch (e) {
+				console.error(`Error running game with seed: [${seed}]\n`, e);
```
Slayer95: We've got to report the format as well in order to reproduce the failure.

dev-tools/harness.js
scheibo: Oh, thanks for letting me know! For my original benchmarking experiments I was using `node --random-seed`, but decided to switch to the `PRNG` for this more 'polished' version. I don't know that the seed matters a ton given I run at default 100 games and usually over 1000, but no reason to choose a known 'bad' seed. Any suggestions? usually something like `0,0,0,0` is bad for a lot of generators.

Slayer95: We could get the upper 64 bits of a crypto hash of the date. That would allow for both some consistency and variability in the results.

dev-tools/harness.js
scheibo: > We've got to report the format as well in order to reproduce the failure.

Very true, good catch!

> Also, it could be useful if the error would output instructions to run the harness with the required parameters?

My main motivation for this harness is two-fold: 

1. primarily for selfish reasons (ie. what motivates most of my work :P), so I dont have to work as hard to maintain a separate `profiling` branch up to date with `HEAD` to profile PS
2. because I think it could be helpful to others interested in performance. I want to add a `'BENCHMARKING.md'` or `'PERFORMANCE.md'` which would explain how to run the harness and profile the app using various tools, because thats something I have a lot of experience with, and IME the easier I make it the more likely other people will care about measuring the impact of their changes etc going forward.

Using this harness as a way to smoke out the simulator is really an indirect goal, and I'm not fully sure about exactly what that use case would entail which is why any 'support' for it at this point probably seems half baked (**suggestions very welcome**). I think my stretch goal for smoke tests would be to test every move + template + gen (though obviously not all possible combinations and interactions) to ensure none of them `throw` to give us some assurance the simulator is solid and is less likely to crash (maybe some kind of custom random team generator would be required + running the battles in parallel + it would still be slow and we'd want to have a separate test suite outside of unit tests which maybe only gets run on demand or by Travis or something?)

Really, I don't know. How could something like this be useful to you? This PR is very raw (I only really opened it at this early stage so I could provide context to Zarel about a bug that ended up being due to my own refactoring), and I'm very open to making it more broadly useful than my benchmarking focus (which admittedly isn't even apparent at this stage in the development because I havent even included any timing logic yet...)

Slayer95: > How could something like this be useful to you?

A use case that comes to my mind is testing experimental OMs, which are likely to be frail. For example, running this would give us assurance that Digimon Showdown works properly :)

>  I want to add a 'BENCHMARKING.md' or 'PERFORMANCE.md' which would explain how to run the harness and profile the app using various tools

Awesome. That would be enough to satisfy my concern raised.

Slayer95: OFC, I shouldn't minimize the benchmarking benefits -running benchmarks in PS has always been such a PITA-, but since it's still a WIP... I wait and see :)

dev-tools/harness.js
Zarel: Did you steal the default seed from the tests? The test seed is I think designed to minimize critical hits.

Zarel: We commonly use `1,2,3,4` as a seed. I don't know if that's bad for the generator.

dev-tools/harness.js
Zarel: You could report the entire inputlog, honestly.

Zarel:
This pull request **introduces 2 alerts** when merging 07d53cbfed704f895250ef2223969e030865ba0d into 29f9c978169425c3333929dacec68f10c56253d3 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-a279145edacfe0d4cdc5620c1567b4323edbb1d4)

**new alerts:**

---

dev-tools/harness.js
```diff
@@ -0,0 +1,148 @@
...
+		const streams = BattleStreams.getPlayerStreams(new BattleStreams.BattleStream());
+		const spec = {formatid: format, seed: this.prng.seed};
+		const p1spec = {name: "Bot 1", team: this.generateTeam(format)};
+		const p2spec = {name: "Bot 2", team: this.generateTeam(format)};
+
+		/* eslint-disable no-unused-vars */
+		const p1 = new RandomPlayerAI(streams.p1,
+			Object.assign({seed: this.nextSeed()}, this.p1options));
+		const p2 = new RandomPlayerAI(streams.p2,
+			Object.assign({seed: this.nextSeed()}, this.p2options));
```
Slayer95: The usage of ``.nextSeed()`` here implies that ``p1`` and ``p2`` issue the same commands off by one turn, right? That makes it seem so unsound...

However, just like my seed comment, I wonder if we should care at all.

scheibo:
Yay! Making progress on timing! (I'm working on a different branch because I don't want to this PR to notify chat/reviewers until its more backing)

table

(disregard the actually numbers, kind of bogus since i'm only testing with one run currently, and Dex loading skews things dramatically)

Zarel:
This pull request **introduces 2 alerts** when merging b9b27306987c1da939c33c3ba07f808af43c5611 into 2a9442eb50a340b58278a5f50f332d6b5e5e1b42 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-88a9855f2c62ba92745d30f89df2bc87b78e6f59)

**new alerts:**

---

scheibo:
Two more `TODO`s left, but its getting very close, such that I figured I'd push to the branch because its not just 'code i threw together for an example' at this point. 

- The 'profiling' library I've written is called [trakr](https://github.com/scheibo/trakr/blob/master/src/index.ts), which is a general purpose, barebones, zero-deps lib. However, I found I was putting too much logic into formatting in `dev-tools/harness.js`, so I've created a wrapper library with deps ([trakkr](https://github.com/scheibo/trakkr/blob/master/src/index.ts)) which is still plenty minimal and generic, but the 'minimal deps base library' + 'optional deps extension library which wraps and reexposes the same API, only enhanced' is a pattern I'm into. Basically, these libraries should be useful for other projects, and Pokemon-Showdown doesn't seem like it needs to be in the business of maintaining a generic profiling library. It's an optional dep in the harness so that if we're running in test mode we don't even need it.
  - trakkr still needs support for aggregating stats and outputting them, this shouldn't be very hard, I just need to make it take a robust aggregation function and format things properly. (I also need to share the `Timer` formatting code with `Tracker`, but thats not really something PS! is concerned with).
- None of the profiling code is checked in, nor do I have any plans for doing so. See 2892070 for an example of what I've instrumented during my testing. In an ideal world, I'd be able to check in my profiling and the code would be removed during an optimizing build step, but thats not really an option unless `tsickle` + Closure Compiler could be made to work, and both are those are finicky and painful enough that I don't think its an avenue worth pursuing for this project. The harness will make it easy to run benchmarks once you instrument the code with timer calls, but it will still require some manual effort.
- Hopefully the PRNG/seeding is closer to being '@Slayer95 -approved' (more critiques welcome!). `--benchmark` mode wants a fixed seed for comparable results over time, but for testing we want it to be good and random.
- Exceptions still don't work correctly. If I manually go and `throw` an error in `Battle`, it doesn't get caught by my `try..catch` even with @Zarel's 18dd3da, for reasons I was trying to outline on #5297 (it ends up as a rejected promise). However, if we fix #5297 (which is now blocking this again, so I'll go after it again), catching the errors should work, at which point this should be useful as a testing harness. And I've already proved I can fix it, its just a matter of working on what the correct approach and API is.

dev-tools/harness.js
```diff
@@ -0,0 +1,209 @@
...
+module.exports = Runner;
+
+// Kick off the Runner if we're being called from the command line.
+if (require.main === module) {
+	// 'move' 70% of the time (ie. 'switch' 30%) and ' mega' 60% of the time that its an option.
+	const AI_OPTIONS = {move: 0.7, mega: 0.6};
+
+	const options = {totalGames: 100, p1options: AI_OPTIONS, p2options: AI_OPTIONS, timer: () => NOOP_TIMER};
+	// If we have more than one arg, or the arg looks like a flag, we need minimist to understand it.
+	if (process.argv.length > 2 || process.argv.length === 2 && process.argv[2].startsWith('-')) {
```
Slayer95: Surely you meant ``process.argv[1]``

dev-tools/harness.js
```diff
@@ -0,0 +1,209 @@
...
+		// 16 frames off the previous.
+		const spec = {formatid: format, seed: this.prng.seed};
+		const p1spec = {name: "Bot 1", seed: this.newSeed()};
+		const p2spec = {name: "Bot 2", seed: this.newSeed()};
+
+		const p1 = new RandomPlayerAI( // eslint-disable-line no-unused-vars
+			streams.p1, Object.assign({seed: this.newSeed()}, this.p1options));
+		const p2 = new RandomPlayerAI( // eslint-disable-line no-unused-vars
+			streams.p2, Object.assign({seed: this.newSeed()}, this.p2options));
+
```
Slayer95: You forgot to ``start()`` the AIs?

dev-tools/harness.js
```diff
@@ -0,0 +1,209 @@
...
+		  if (missing('minimist')) shell('npm install trakkr');
+			const trakkr = require('trakkr');
+			options.timer = () => new trakkr.Timer();
+			// Choose which formatter to use - we don't need to tweak the sort or write a custom
+			// formatter because its almost as though the defaults were written for our usecase...
+			options.formatter = new trakkr.Formatter(!!argv.full, trakkr.SORT,
+				argv.csv ? trakkr.CSV : (argv.tsv ? trakkr.TSV : trakkr.TABLE));
+		}
+		if (argv.seed) options.prng = argv.prng.split(',').map(s => Number(s));
```
Slayer95: ``argv.seed.split``

dev-tools/harness.js
scheibo: I didn't, this is before `start()` as added (they used to auto-start)

scheibo:
Still not ready, but I'm making progress.

scheibo:
*A further update on "pre's interminable quest to build benchmarking/performance infrastructure that no one asked for..."*

 Yesterday I added BenchmarkJS support to the script, and made a distinction between **tracing** (generating trace events from manually instrumented sections to then view in `chrome://tracing`), **profiling** (using manually instrumented sections to compute breakdowns of where the application spends its time) and now **benchmarking** (running the same scenario multiple times and then reasoning about the multiple runs in aggregate). 

The motivation is to be able to determine if a performance change we make is actually statistically significant, which is slightly challenging because running the suite takes a while (~12s, and we want multiple runs) and theres actually high variance between the runs. Today I spent a while porting the mishmash of R and Python I used to use to validate whether a performance improvement was statistically significant or not to JS. I 'validated' it by reverting the `getEffect` caching from #5181, as well as a 'null hypothesis' verification run for two consecutive runs at `HEAD`.

comparison

In the first case, the `getEffect` optimization is so large we hardly needed this new logic to verify it was an improvement, but you can see from the second example why such statistical verification is required for smaller changes - between identical runs there's enough variation that we would have no way of actually knowing we had an improvement. This large variance unfortunately mean that optimizations will need to be fairly large and obvious to be picked out from the noise (and I'll continue looking to reduce the variance), but I'm happy enough with the benchmarking and comparison logic that I'm going to try to actually validate some smaller optimization ideas to see if they move the needle at all.

This branch is still very much WIP as I'm still organically adding features and experimenting with output as I figure out what I need. **I wonder though if I should just get the testing logic committed (without any of the benchmark/profile/tracing/formatting options I keep bikeshedding) so that the harness can be used by others for looking for errors/debugging etc, and followup with my performance logic later?**

scheibo:
*(This isn't updated, I was just doing branch cleanup and accidentally deleted this because its sometimes hard to remember which of the 50 branches i had were still necessary)*

PR #5279, opened 2019-03-09
Implement /cmd roominfo

Kaiepi:
At the moment, bots can't differentiate between hidden and private rooms, and while they can get the modjoin status and list of auth in rooms, it's very difficult to parse (I had to write a grammar to parse /roomauth's popup, but even then couldn't work out a way to get the room title while accounting for other servers that may have translated the command's output).

This allows bots to be able to get the room title, type, visibility, modjoin status, auth list, and userlist of a room easily.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4262,7 +4262,7 @@ const commands = {
 				roominfo.users.push(userinfo);
 			}
 
-			connection.send(`|queryresponse|roomroominfo|${JSON.stringify(roominfo)}`);
+			connection.send(`|queryresponse|roominfo|${JSON.stringify(roominfo)}`);
 		} else {
 			// default to sending null
 			connection.send(`|queryresponse|${cmd}|null`);
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4223,6 +4223,46 @@ const commands = {
 			Ladders(toId(target)).getTop().then(result => {
 				connection.send('|queryresponse|laddertop|' + JSON.stringify(result));
 			});
+		} else if (cmd === 'roominfo') {
+			if (!trustable) return false;
+
+			let targetRoom = Rooms.get(target);
+			if (!targetRoom) return false;
+			if (targetRoom.isPrivate && !user.inRooms.has(room.id) && !user.games.has(room.id)) {
+				return false;
+			}
+
+			let visibility;
+			if (targetRoom.isPrivate) {
+				visibility = (targetRoom.isPrivate === 'hidden') ? 'hidden' : 'private';
+			} else {
+				visibility = 'public';
+			}
+
+			let roominfo = {
+				title: targetRoom.title,
+				type: targetRoom.type,
+				visibility: visibility,
+				modjoin: targetRoom.modjoin || targetRoom.modchat,
+				auth: {},
+				users: [],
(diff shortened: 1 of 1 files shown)


Zarel:
Heh, this was previously planned to be a `|metadata|` message sent from the room.

I still think that's probably the right approach? The client could use that metadata too. Although we never really agreed on what to call it.

Zarel:
`|roominfo|` is probably not a name, in hindsight.

server/chat-commands.js
```diff
@@ -4223,6 +4223,47 @@ const commands = {
...
+			}
+
+			for (let userid in targetRoom.users) {
+				let user = targetRoom.users[userid];
+				let rank = room.auth ? (room.auth[userid] || user.group) : user.group;
+				let userinfo = `${rank}${user.name}`;
+				roominfo.users.push(userinfo);
+			}
+
+			connection.send(`|queryresponse|roomroominfo|${JSON.stringify(roominfo)}`);
```
```suggestion
			connection.send(`|queryresponse|roominfo|${JSON.stringify(roominfo)}`);
```

Zarel:
This is also fine, though.

Kaiepi:
Sending a message like this when joining the room sounds like a better idea. `|title|` and `|users|` could be merged with this, along with the room type included in `|init|` (leaving it empty)

Kaiepi:
Wait does the client need the room type in `|init|` to figure out where to put the room?

server/chat-commands.js
Kaiepi: Damn `:s` in vim

Kaiepi:
Never mind, just saw your post. If you think this is fine I'm ok with it

server/chat-commands.js
```diff
@@ -4223,6 +4223,47 @@ const commands = {
...
+				for (let userid in targetRoom.auth) {
+					let rank = targetRoom.auth[userid];
+					if (!roominfo.auth[rank]) roominfo.auth[rank] = [];
+					roominfo.auth[rank].push(userid);
+				}
+			}
+
+			for (let userid in targetRoom.users) {
+				let user = targetRoom.users[userid];
+				let rank = room.auth ? (room.auth[userid] || user.group) : user.group;
```
whalemer: is ``User#getIdentity`` fast enough to use here?

Kaiepi: I wasn't aware that it existed. It'd probably be a better idea to use that here so bots can know who's muted or locked

Zarel:
I actually like this better than sending the message when joining the room, mostly because most of this information will be out of date, and it's not like the server is sending live updates.

That, and, there's a lot of information here that I wouldn't want to send to every client, for perf reasons too.

server/chat-commands.js
Zarel: You accidentally clicked "Resolve" instead of committing this.

Kaiepi: I force pushed without pulling the commit it made, my bad

Zarel: And/or you accidentally overwrote your fix with the `getIdentity` change.

Zarel:
@Kaiepi 
> I force pushed without pulling the commit it made, my bad

```
git config --global alias.forcepush push --force-with-lease
```

PR #5280, opened 2019-03-09
Fix race with replace in pokemon-showdown launcher

scheibo:


scheibo:
Trying to go to bed but I keep thinking about this - i think it's wrong, its too risky. Instead we should make it so we always build except in the `simulate-battle` case, at which point we rely on the heuristic. No reason to cause other users headaches.

scheibo:
Alright, I think this now has the correct tradeoff between speed and safety.

Zarel:
My concerns do apply to all of the command-line functionality besides starting the server, not just `simulate-battle`.

PR #5284, opened 2019-03-09
Make move choice parsing flexible regarding targeting/flag order

scheibo:


Zarel:
"Fix" is an interesting way to describe changing the order of the expected elements.

Zarel:
It might be ideal to accept targeting/mega in any order, though.

Zarel:
Also we should probably make choice failures throw in a test context.

scheibo:
> "Fix" is an interesting way to describe changing the order of the expected elements.

`SIM-PROTOCOL.md` seems to imply `mega` comes last?

> `move MOVESPEC mega`, to mega-evolve and make a move
> `MOVESPEC` is:
> `MOVESLOTSPEC` or `MOVESLOTSPEC TARGETSPEC`

Zarel:
It's way more likely that the protocol doc is wrong, than that the code is wrong.

Zarel:
I think the ideal solution here is probably to accept targeting/mega in any order. Then we don't have to fix the doc _or_ the client!

scheibo:
> I think the ideal solution here is probably to accept targeting/mega in any order. Then we don't have to fix the doc or the client!

OK, ill look into this. 

> Also we should probably make choice failures throw in a test context.

And this

scheibo:
Alright, I've made it so its now agnostic to ordering, but `throw`ing breaks a lot of tests and is kind of unrelated to this PR so I've moved those changes to #5286 (my approach there might also be wrong, but we can talk about that over there).

Zarel:
This pull request **introduces 1 alert** when merging b813fa6e220169c83c840effcd4fd431fe674c5c into 9e7ba68fc4caedf503638bd7edda6b781df18d65 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-c1f391939b8a933ed0608f0ab4a1b1b3da7983f3)

**new alerts:**

* 1 for Useless assignment to local variable

---

sim/side.ts
```diff
@@ -629,17 +629,37 @@ export class Side {
...
+				// tslint:disable-next-line:one-variable-per-declaration
+				let targetLoc, willMega, willUltra, willZ;
+				// We're flexible with the ordering of target vs. flags - if `data `ends
+				// with a number then we assume target comes last, otherwise flags.
+				if (/[1-3]$/.test(data)) {
+					[targetLoc, data] = parseLoc(data);
+					[willMega, willUltra, willZ, data] = parseFlags(data);
+				} else {
+					[willMega, willUltra, willZ, data] = parseFlags(data);
+					[targetLoc, data] = parseLoc(data);
```
Zarel: This seems kind of inefficient.

```js
let targetLoc = 0;
let willMegaEtc = false;
while (true) {
	if (/\s-?[1-3]$/.test(data)) {
		targetLoc = parseInt(data.slice(-2), 10);
		data = data.slice(0, data.lastIndexOf(' '));
	} else if (data.endsWith(' mega')) {
		willMegaEtc = true;
		data = data.slice(0, -5);
	} else if (data.endsWith(' ultra')) {
		willMegaEtc = true;
		data = data.slice(0, -6);
	} else if (data.endsWith(' zmove')) {
		willMegaEtc = true;
		data = data.slice(0, -6);
	} else {
		break;
	}
}
```

scheibo: Done.

sim/side.ts
```diff
@@ -629,17 +629,24 @@ export class Side {
 
 			switch (choiceType) {
 			case 'move':
-				let targetLoc = 0;
-				if (/\s-?[1-3]$/.test(data)) {
-					targetLoc = parseInt(data.slice(-2), 10);
-					data = data.slice(0, data.lastIndexOf(' '));
+				let [targetLoc, willMega, willUltra, willZ] = [0, '', '', ''];
```
Zarel: I still don't like this initialization.

```suggestion
				let targetLoc = 0;
				let megaOrZ = '';
```

sim/side.ts
```diff
@@ -629,17 +629,24 @@ export class Side {
...
+				while (true) {
+					if (/\s-?[1-3]$/.test(data)) {
+						targetLoc = parseInt(data.slice(-2), 10);
+						data = data.slice(0, data.lastIndexOf(' '));
```
```suggestion
						data = data.slice(0, -2).trim();
```

In hindsight, this is probably easier to read.

sim/side.ts
```diff
@@ -629,17 +629,24 @@ export class Side {
...
+						break;
+					}
 				}
-				const willMega = data.endsWith(' mega') ? 'mega' : '';
-				if (willMega) data = data.slice(0, -5);
-				const willUltra = data.endsWith(' ultra') ? 'ultra' : '';
-				if (willUltra) data = data.slice(0, -6);
-				const willZ = data.endsWith(' zmove') ? 'zmove' : '';
-				if (willZ) data = data.slice(0, -6);
 				this.chooseMove(data, targetLoc, willMega || willUltra || willZ);
```
```suggestion
				this.chooseMove(data, targetLoc, megaOrZ);
```

sim/side.ts
Zarel: I actually explicitly think `willMega, willUltra, willZ` should be combined into one variable – it's done when `chooseMove` is called, anyway.

Slayer95:
This doesn't reject invalid commands such as ``/move 1 1 mega ultra``. Unintended, right?

Zarel:
The old code didn't, either. We could certainly throw if `megaOrZ` was truthy when we went to set it. Feel free to pullreq that change.

test/simulator/misc/choice-parser.js
```diff
@@ -80,6 +80,21 @@ describe('Choice parser', function () {
...
+				]);
+				battle.join('p2', 'Guest 2', 1, [
+					{species: "Blaziken", ability: 'speedboost', item: 'blazikenite', moves: ['blazekick']},
+					{species: "Aggron", ability: 'sturdy', moves: ['irondefense']},
+				]);
+
+				assert.ok(battle.choose('p1', `move 1 1 mega, move smog 1`));
+				assert.ok(battle.choose('p2', `move 1 mega 2, move irondefense`));
+			});
+
```
Slayer95: This test was placed under ``Switch requests > Doubles/Triples``. It should be under ``Move requests > Generic``

Slayer95:
Oh you are right. The loop just made it more obvious.

Zarel:
Where by "throw" I mean emit a choice error.

scheibo:
> This test was placed under ``Switch requests > Doubles/Triples``. It should be under ``Move requests > Generic``

_Originally posted by @Slayer95 in https://github.com/Zarel/Pokemon-Showdown/pull/5284_


> Where by "throw" I mean emit a choice error.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5284#issuecomment-471244025_

Will follow up, gimme 5

PR #5285, opened 2019-03-09
Random Battle: Update Leafeon

ascriptmaster:
Smogon Analysis uses Double Edge with Normalium-Z instead of X-Scissor,
which shares too many resistances with Leaf Blade. (Use Return for Gen
6 because there is no Z Crystal there.)

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -2174,7 +2174,7 @@ let BattleFormatsData = {
 		doublesTier: "(DUU)",
 	},
 	leafeon: {
-		randomBattleMoves: ["swordsdance", "leafblade", "healbell", "xscissor", "synthesis", "knockoff"],
+		randomBattleMoves: ["swordsdance", "leafblade", "healbell", "doubleedge", "synthesis", "knockoff"],
 		randomDoubleBattleMoves: ["swordsdance", "leafblade", "substitute", "xscissor", "protect", "helpinghand", "knockoff"],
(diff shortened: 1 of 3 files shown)


TheImmortal:
Is it worth it tho? On something like Sawsbuck it would be good. But it's non-STAB and Leafeon in general is not sweeping any teams.

ascriptmaster:
It sure is better than X-Scissor, which shares like 4 resistances to grass versus normal sharing 1, meanwhile a +2 Z Double Edge can do a lot of damage in a pinch against some fire or flying mon off its base 110 attack, hence its appearance in the analysis. At worst I'd be willing to drop return from gen 6, which is not as effective but still is better coverage than X-Scissor

TheImmortal:
Okay I’ll see what Randoms people have to say because I’m 50/50.

ascriptmaster:
So, lose the return or no?

TheImmortal:
The consensus is that Double Edge without Z is not good, especially combined with Life Orb. Second, it's a waste of a Z-Move that a another mon could have, and non-STAB as well. And third, people like being able to hit Grass-types with X-Scissor. I'm gonna have to agree unfortunately.

ascriptmaster:
Huh. I thought I had the opposite impression when I asked. The X-Scissor thing seems odd considering many Grass mons are Poison anyways. I'll have to do more research and come back later with improvements.

PR #5266, opened 2019-03-07
Support for 4 player battles

MacChaeger:

PR #5286, opened 2019-03-09
Make Battle throw on choice errors in tests

scheibo:
> Also we should probably make choice failures throw in a test context.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5284#issuecomment-471216992_


scheibo:
A lot of the tests fail because everywhere that is doing `assert.false(battle.choose(...))` will now throw before returning false.

sim/battle.ts
```diff
@@ -33,6 +33,7 @@ interface BattleOptions {
 	p1?: PlayerOptions; // Player 1 data
 	p2?: PlayerOptions; // Player 2 data
 	debug?: boolean; // show debug mode option
+	test?: boolean; // whether we are in a test environment
```
Zarel: Maybe `strictChoices`?

Zarel:
I like this strategy. The tests can be revised to `assert.throws`.

scheibo:
Man, that was a PITA! Clearly `strictChoices` is valuable given it finds so many typos in tests.

I gave up on some of them and just turned `strictChoices` on. I will file a bug to followup and see if we can turn on `strictChoices` everywhere (though maybe `decisions.js` will still want it off), but I've kind of lost my patience on this for now and maybe someone else can take up the long tail of the tests.

scheibo:
> So, uh, the button to make this PR not a draft any more isn't there.

_Originally posted by @MacChaeger in https://github.com/Zarel/Pokemon-Showdown/pull/5266#issuecomment-471245327_

This is broken for me as well (did GitHub push a bad build before the weekend?). @Zarel, can you please mark this as ready for review?

test/assert.js
```diff
@@ -74,6 +74,26 @@ assert.holdsItem = function (pokemon, message) {
...
+	if (pokemon && move) {
+		assert.throws(() => fn(), Error, new RegExp(`[Invalid choice] Can't move:.*${pokemon}.*${move}`, 'i'));
+	} else {
+		assert.throws(() => fn(), Error, new RegExp(`[Invalid choice] Can't move:`));
+	}
+};
+
+assert.cantUndo = function (fn) {
+	assert.throws(() => fn(), Error, new RegExp("[Invalid choice] Can't undo:"));
+};
```
Slayer95: Gotta escape the brackets

scheibo:
```
Pokemon-Showdown/test/assert.js
  83:48  error  Unnecessary escape character: \[  no-useless-escape
  83:64  error  Unnecessary escape character: \]  no-useless-escape
  85:48  error  Unnecessary escape character: \[  no-useless-escape
  85:64  error  Unnecessary escape character: \]  no-useless-escape
  90:47  error  Unnecessary escape character: \[  no-useless-escape
  90:63  error  Unnecessary escape character: \]  no-useless-escape
```

Zarel: No, `eslint` is correct. Your escapes are useless. `"\["` means the same thing as `"["` – it's a string passed to a regex construct, not a real regex. You need to do `"\\[`.

test/assert.js
```diff
@@ -74,6 +74,26 @@ assert.holdsItem = function (pokemon, message) {
...
+assert.cantMove = function (fn, pokemon, move) {
+	if (pokemon && move) {
+		assert.throws(() => fn(), Error, new RegExp(`\\[Invalid choice\\] Can't move:.*${pokemon}.*${move}`, 'i'));
+	} else {
+		assert.throws(() => fn(), Error, new RegExp(`\\[Invalid choice\\] Can't move:`));
+	}
+};
+
+assert.cantUndo = function (fn) {
+	assert.throws(() => fn(), Error, new RegExp("\\[Invalid choice\\] Can't undo:"));
```
Zarel: In hindsight I'm confused why this isn't just a regex literal.

```suggestion
	assert.throws(() => fn(), Error, /\[Invalid choice\] Can't undo:/);
```

scheibo: It should be. I did the `cantUndo` case with interpolation that actual required the constructor first, and then just copied and pasted the rest. Fixed this and above.

test/assert.js
```diff
@@ -74,6 +74,26 @@ assert.holdsItem = function (pokemon, message) {
...
+	if (pokemon && move) {
+		assert.throws(() => fn(), Error, new RegExp(`\\[Invalid choice\\] Can't move:.*${pokemon}.*${move}`, 'i'));
+	} else {
+		assert.throws(() => fn(), Error, /\[Invalid choice\] Can't move:/);
+	}
+};
+
+assert.cantUndo = function (fn) {
+	assert.throws(() => fn(), Error, /\[Invalid choice\] Can't undo:/);
+};
```
Slayer95: Uhm, actually... I have just checked the ``assert`` API at https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message:

> assert.throws(fn[, error][, message])
fn <Function>
error <RegExp> | <Function> | <Object> | <Error>
message <string>

[...]

If specified, error can be a Class, RegExp [...]

If specified, message will be appended to the message provided by the AssertionError if the fn call fails to throw or in case the error validation fails.

So, from above, ``error`` should be a ``RegExp`` matching the choice error, and ``message`` should be something like ``'Choice unexpectedly accepted``

Slayer95: Per above, ``error`` should be a ``RegExp`` matching the choice error, and ``message`` should be something like ``'Choice unexpectedly accepted``

scheibo: Ah, so right now its just matching the `Error` constructor, so its basically just a vanilla `assert.throws` with a weird regexp error message? Thanks for being so 'flexible', JS.

scheibo: Switched things around, but `message` still doesn't seem to be working correctly. Will dig in further.

scheibo: OK, I'm dumb. `message` is working, it just doesn't work the way I thought it would. I thought it would always be used if the test fails, but its only used if the *assertion* fails (ie/ *no* error is thrown). If the test fails because an error is thrown, you see the *exception*'s message. This makes sense and is fine, I just didn't understand what was going on and couldn't verify my `message`s were working correctly until I figured it out.

I think I'm done with this change. I'd be happy to submit this as is and let you take over the remaining cases if you're still up for it. :)

Slayer95: Great! Oh, oops! Sorry for the merge conflict :[

scheibo: No worries.

test/simulator/moves/curse.js
```diff
@@ -140,11 +142,12 @@ describe('XY/ORAS Curse targetting when becoming Ghost the same turn', function
 		let p1active = battle.p1.active;
 		let p2active = battle.p2.active;
 
-		// p1: Kecleon uses Curse last in the turn.
-		// p2: Electric attack on Kecleon, then Ghost.
 		battle.makeChoices(
-			`move ${battle.p1.active[0].set.moves[0]}, move ${battle.p1.active[1].set.moves[0]}, move ${battle.p1.active[2].set.moves[0]}`,
-			`move aurasphere ${curseUser.position + 1}, move lick ${curseUser.position + 1}, move harden`);
+			// p1: Kecleon uses Curse last in the turn.
```
scheibo: well, duh. I don't know why i decided to overcomplicate things

test/simulator/moves/curse.js
```diff
@@ -140,11 +142,12 @@ describe('XY/ORAS Curse targetting when becoming Ghost the same turn', function
...
+			// p2: Electric attack on Kecleon, then Ghost.
+			`move 1, move 1, move 1`,
```
scheibo: well, duh. I don't know why i decided to overcomplicate things

Zarel:
I unfortunately can't commit this because of the number of bugged tests, but here are some things that could be integrated with this PR, representing the future of PS APIs:

First: `makeChoices()` should auto-choose the defaults for everyone:

```js
	/**
	 * Convenience method for easily making choices.
	 */
	makeChoices(...inputs: string[]) {
		if (inputs.length) {
			for (const [i, input] of inputs.entries()) {
				this.sides[i].choose(input);
			}
		} else {
				side.autoChoose();
			}
		}
		this.commitDecisions();
	}
```

Second, `commitDecisions` should throw if the battle is still waiting on choices:

```js
	commitDecisions() {
		this.updateSpeed();

		let oldQueue = this.queue;
		this.queue = [];
		if (!this.allChoicesDone()) throw new Error("Not all choices done");
```

Third, `allChoicesDone()` replaces `checkActions()`: it's just the same thing, but with `commitDecisions` removed. The relevant call in `battle.choose` becomes:

```js
		if (this.allChoicesDone()) this.commitDecisions();
```

scheibo:
> I unfortunately can't commit this because of the number of bugged tests

Sorry, what? I'm confused because this change *fixes* the bugged tests? I can understand if you mean "I don't want to commit the bit where `emitChoiceError` throws right now", but surely you want the bit that fixes dumb typos (eg. https://github.com/Zarel/Pokemon-Showdown/pull/5286/files#diff-bfd13248eb2395e8e715f095140dd4b5) and poorly written tests?

Zarel:
I mean I don't want to commit the code I pasted. Not the PR. I'm sorry, in hindsight, that could have been misinterpreted.

The code I pasted can't be pushed directly into master because master has so many bugged tests, so I instead want it committed to this PR's branch, which fixes those bugged tests and is related as a strictness improvement.

scheibo:
Ah OK. I was worried my Sunday spent correctly other people's typos in tests was going to be all for naught and got defensive ;)

test/simulator/abilities/pickup.js
```diff
@@ -53,9 +53,9 @@ describe('Pickup', function () {
 			{species: 'Clefable', ability: 'unaware', item: 'ejectbutton', moves: ['followme']},
 			{species: 'Magikarp', ability: 'rattled', moves: ['splash']},
 		]);
-		battle.makeChoices('move shadowsneak, move rest', 'move uturn, move followme');
-		battle.makeChoices('move shadowsneak, move rest', 'switch 3'); // Eject Button activated
-		battle.makeChoices('move shadowsneak, move rest', 'switch 3');
+		battle.makeChoices('move shadowsneak 1, move rest', 'move uturn 1, move followme');
+		battle.makeChoices('', 'switch 3'); // Eject Button activated
+		battle.makeChoices('', 'switch 3');
```
Slayer95: As per latest @Zarel's proposal, this line should change to explicity proc the decisions (add the missing command).

scheibo:
@Slayer95 - I know you were doing some work on this? Did you make any progress? What do you think of [Zarel's suggestion above](https://github.com/Zarel/Pokemon-Showdown/pull/5286#issuecomment-472350403)?

Slayer95:
Uh? Sorry for not being clear. I only intended to handle the tests...
As for @Zarel 's proposal, it seems fair :+1:

scheibo:
Oh, I blanked on aa32787 that you already committed.

scheibo:
@Zarel https://github.com/Zarel/Pokemon-Showdown/pull/5286#issuecomment-472350403: :heavy_check_mark:

PR #5288, opened 2019-03-10
Make the `switch` command more flexible

scheibo:
Make it take a case-insensitive nickname or species id,
something several of our test cases already rely upon.

### DIFF
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -456,10 +456,10 @@ export class Side {
 			slot = parseInt(slotText!, 10) - 1;
 		}
 		if (isNaN(slot) || slot < 0) {
-			// maybe it's a name!
+			// maybe it's a name/species id!
 			slot = -1;
 			for (const [i, mon] of this.pokemon.entries()) {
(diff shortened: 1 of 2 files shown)


sim/side.ts
```diff
@@ -456,10 +456,10 @@ export class Side {
 			slot = parseInt(slotText!, 10) - 1;
 		}
 		if (isNaN(slot) || slot < 0) {
-			// maybe it's a name!
+			// maybe it's a name/species id!
 			slot = -1;
 			for (const [i, mon] of this.pokemon.entries()) {
-				if (slotText === mon.name) {
+				if (slotText!.toLowerCase() === mon.name.toLowerCase() || toId(slotText) === mon.speciesid) {
```
Slayer95: This is redundant: ``toId(slotText) === mon.speciesid`` implies ``slotText!.toLowerCase() === mon.name.toLowerCase()``. Furthermore, ``toId(slotText)`` could be called outside of the loop.

PR #5289, opened 2019-03-10
Rename most of room-battle.js's exports

MacChaeger:


server/room-battle.js
```diff
@@ -1017,7 +1017,7 @@ class Battle {
 
 exports.RoomBattlePlayer = BattlePlayer;
 exports.RoomBattleTimer = BattleTimer;
-exports.RoomBattle = Battle;
+exports.RoomBattle = BattleGame;
```
Zarel: Wait, it's exported as `RoomBattle`? On second thought, why not just call it `RoomBattle` everywhere? Ditto `BattlePlayer` and `BattleTimer`.

PR #5290, opened 2019-03-10
Reject `move` commands with multiple flags

scheibo:
> This test was placed under ``Switch requests > Doubles/Triples``. It should be under ``Move requests > Generic``

_Originally posted by @Slayer95 in https://github.com/Zarel/Pokemon-Showdown/pull/5284_

> Where by "throw" I mean emit a choice error.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5284#issuecomment-471244025_

---

I also snuck in `'auto'` as an alias for `'default'`, as I've discovered many tests are using it during the cleanup I'm working on in #5286

### DIFF
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -629,19 +629,25 @@ export class Side {
 
 			switch (choiceType) {
 			case 'move':
-				let targetLoc = 0;
+				const original = data;
+				const error = () => this.emitChoiceError(`Conflicting arguments for "move": ${original}`);
+				let targetLoc: number | undefined;
 				let megaOrZ = '';
 				while (true) {
 					if (/\s-?[1-3]$/.test(data)) {
+						if (targetLoc !== undefined) error();
 						targetLoc = parseInt(data.slice(-2), 10);
 						data = data.slice(0, -2).trim();
 					} else if (data.endsWith(' mega')) {
+						if (megaOrZ) error();
 						megaOrZ = 'mega';
 						data = data.slice(0, -5);
 					} else if (data.endsWith(' zmove')) {
+						if (megaOrZ) error();
 						megaOrZ = 'zmove';
 						data = data.slice(0, -6);
 					} else if (data.endsWith(' ultra')) {
+						if (megaOrZ) error();
 						megaOrZ = 'ultra';
 						data = data.slice(0, -6);
 					} else {
@@ -665,6 +671,7 @@ export class Side {
 				if (data) return this.emitChoiceError(`Unrecognized data after "pass": ${data}`);
 				this.choosePass();
 				break;
+			case 'auto':
 			case 'default':
 				this.autoChoose();
 				break;
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -635,19 +635,19 @@ export class Side {
 				let megaOrZ = '';
 				while (true) {
 					if (/\s-?[1-3]$/.test(data)) {
-						if (targetLoc !== undefined) error();
+						if (targetLoc !== undefined) return error();
 						targetLoc = parseInt(data.slice(-2), 10);
 						data = data.slice(0, -2).trim();
 					} else if (data.endsWith(' mega')) {
-						if (megaOrZ) error();
+						if (megaOrZ) return error();
 						megaOrZ = 'mega';
 						data = data.slice(0, -5);
 					} else if (data.endsWith(' zmove')) {
-						if (megaOrZ) error();
+						if (megaOrZ) return error();
 						megaOrZ = 'zmove';
 						data = data.slice(0, -6);
 					} else if (data.endsWith(' ultra')) {
-						if (megaOrZ) error();
+						if (megaOrZ) return error();
 						megaOrZ = 'ultra';
 						data = data.slice(0, -6);
 					} else {
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -638,7 +638,7 @@ export class Side {
 					// We need to special case 'Conversion 2' so it doesn't get
 					// confused with 'Conversion' erroneously sent with the target
 					// '2' (since Conversion targets 'self', targetLoc can't be 2).
-					if (/\s-?[1-3]$/.test(data) && toId(data) !== 'conversion2') {
+					if (/\s(?:-|\+)?[1-3]$/.test(data) && toId(data) !== 'conversion2') {
 						if (targetLoc !== undefined) return error();
 						targetLoc = parseInt(data.slice(-2), 10);
 						data = data.slice(0, -2).trim();
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -634,7 +634,11 @@ export class Side {
 				let targetLoc: number | undefined;
 				let megaOrZ = '';
 				while (true) {
-					if (/\s-?[1-3]$/.test(data)) {
+					// If data ends with a number, treat it as a target location.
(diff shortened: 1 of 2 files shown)


sim/side.ts
```diff
@@ -629,19 +629,25 @@ export class Side {
...
 			switch (choiceType) {
 			case 'move':
-				let targetLoc = 0;
+				const original = data;
+				const error = () => this.emitChoiceError(`Conflicting arguments for "move": ${original}`);
+				let targetLoc: number | undefined;
 				let megaOrZ = '';
 				while (true) {
 					if (/\s-?[1-3]$/.test(data)) {
+						if (targetLoc !== undefined) error();
```
Slayer95: I might be missing something, but shouldn't these be early ``return error()``; ?

Zarel: This is going to be a mess when in conjunction with `Conversion 2`

Zarel: How did we actually do this in the past, anyway?

Zarel: It looks like neither Conversion nor Conversion 2 are targeted, but various consumers might want to send a target anyway.

Zarel: If a Pokémon in move slot two sends `move Conversion 2`, and has both `Conversion` and `Conversion 2` in its moveset, what should happen?

scheibo: IIRC, its position would be `-2`, not `2`. `2` would be targetting the opponent, which would be an invalid target anyway (regardless of whether its `Conversion` or `Conversion 2`).

I think in the `Conversion[ 2]` case we should just special case it to ignore targetting information, because theyre both `target: self`

scheibo: these should all be `return`s, yeah.

Zarel: Oh, good point. That sounds good to me.

sim/side.ts
```diff
@@ -629,19 +629,25 @@ export class Side {
...
 			case 'move':
-				let targetLoc = 0;
+				const original = data;
+				const error = () => this.emitChoiceError(`Conflicting arguments for "move": ${original}`);
+				let targetLoc: number | undefined;
 				let megaOrZ = '';
 				while (true) {
 					if (/\s-?[1-3]$/.test(data)) {
+						if (targetLoc !== undefined) return error();
 						targetLoc = parseInt(data.slice(-2), 10);
```
Zarel: So can you add a check for Conversion 2?

scheibo: Yeah, I'm doing that now. :) I assumed you'd notice it wasn't present and wouldn't merge without it, and I wanted to give you a chance to respond.

Slayer95:
Conversion 2 is an annoying case indeed. It used to be buggy but it now works by forcing the client to send the move ID.

Zarel:
We could also require `+` for targeting the enemy.

sim/side.ts
```diff
@@ -629,19 +629,25 @@ export class Side {
 
 			switch (choiceType) {
 			case 'move':
-				let targetLoc = 0;
+				const original = data;
+				const error = () => this.emitChoiceError(`Conflicting arguments for "move": ${original}`);
+				let targetLoc: number | undefined;
 				let megaOrZ = '';
 				while (true) {
 					if (/\s-?[1-3]$/.test(data)) {
```
```suggestion
					if (/\s(?:-|\+)?[1-3]$/.test(data)) {
```

Zarel: No reason not to support it as an option. Will be useful if we get another move ending in a number.

test/simulator/misc/choice-parser.js
```diff
@@ -155,6 +140,27 @@ describe('Choice parser', function () {
 			});
 		});
 
+		it('should allow mega evolving and targeting in the same move in either order', function () {
```
Slayer95: Hm, you have placed it outside the inner ``describe`` blocks :|

scheibo: Oh man, I'm doing really bad at placing this in the correct place.

sim/side.ts
scheibo: Done.

sim/side.ts
scheibo: Done.

test/simulator/misc/choice-parser.js
Slayer95: Also, this could well be tested in a Singles game. I am not sure why you create a Doubles one...

scheibo: Hopefully the latest commit finally found the correct spot for this.

scheibo:
> We could also require + for targeting the enemy.

*Requiring* it seems like a breaking change, but allowing clients to use it (like we're currently doing in the PR) seems reasonable.

> Conversion 2 is an annoying case indeed. It used to be buggy but it now works by forcing the client to send the move ID.

Yeah... I added a bunch of `console.log`s when testing the latest commits to see that it was actually doing the right thing with `Conversion 2`. I'm not confident enough to claim this implementaiton is perfect, but it doesn't seem to break in any obvious way :)

test/simulator/misc/choice-parser.js
```diff
@@ -153,6 +138,27 @@ describe('Choice parser', function () {
...
+					{species: "Aggron", ability: 'sturdy', moves: ['irondefense']},
+				]);
+
+				const badChoices = [`move 1 1 2`, `move 1 1 mega ultra`, `move 1 mega zmove 2`];
+				for (const badChoice of badChoices) {
+					const choice = `${badChoice}, move Conversion 2`;
+					assert.false(battle.choose('p1', choice), `Choice '${choice}' should be rejected`);
+				}
+
+				assert.ok(battle.choose('p1', `move 1 1 mega, move Conversion 2 1`));
```
Zarel: Shouldn't this error out for invalid target?

test/simulator/misc/choice-parser.js
```diff
@@ -153,6 +138,27 @@ describe('Choice parser', function () {
...
+				assert.ok(battle.choose('p2', `move Conversion 2 1 zmove, move irondefense`));
+			});
 		});
```
Slayer95: At last it's well-placed. But isn't this test now too overloaded? Conversion 2 is an outstanding issue and its behaviour should be explicitly tested independently IMO.

test/simulator/misc/choice-parser.js
Slayer95: It only errors in Gen 2-4. Conversion 2 targets adjacent since BW.

test/simulator/misc/choice-parser.js
scheibo: Yeah... felt kind of like I was piling on - thanks for keeping me honest. I agree actually thoroughly testing Conversion + Conversion 2 deserves its own test, but I'm a little burnt out tonight (been in bed sick and coding all day). I'll separate it in the morning.

test/simulator/misc/choice-parser.js
Zarel: I was lied to!

Zarel: Oh wait I got Conversion and Conversion 2 confused. Nevermind.

Zarel:
These changes so far are good; any more changes can go in a new pull request.

test/simulator/misc/choice-parser.js
scheibo: ....or i'll followup tomorrow because its already merged :)

PR #5292, opened 2019-03-10
Enable prefer-const and max 120 character lines for .ts files

scheibo:


Zarel:
I'm not sure how important `prefer-const` is to me.

scheibo:
>*If a variable is never modified, using the `const` declaration is better. Const declaration tells readers, “this variable is never modified,” reducing cognitive load and improving maintainability.*

from the documentation and ["Why `prefer-const`"](https://medium.com/tanda-developers/why-do-we-prefer-const-d805b9dc03be) might offer some small explanation of the potential benefits.

It's not a hill I'm going to die on, and I'm always OK with PRs being rejected. I noticed the TODOs in the `tslint.json` config (that I had copied from the client) and I personally use `const` in my own code, so I figured I'd try to get rid of them. It doesn't seem like its particularly onerous on people, and helps with consistency. But once again, if you don't want it I'm fine to drop this (though maybe I should remove the TODO?) :).

Zarel:
So, I like `const`, especially if it's opt-in. Your link makes a good case that globals should be `const` if they're supposed to be immutable, but that's already true.

But I don't think it's so useful as to be proactively making everything at all a `const` the moment it's not edited, and then back the moment it is. That seems like a lot of work for comparatively little benefit.

Especially since JavaScript isn't _amazing_ at functional code style. It can do it, but without stuff like `switch` expressions, a lot of things that _could_ be const aren't. And so switching between `let` and `const` a lot is going to need to happen.

scheibo:
> That seems like a lot of work for comparatively little benefit.

I think that you're overestimating the number of times we'd actually need to switch back and the work involved. I'd guess it happens in less than 10-20% of cases here. That doesnt seem too high, and its a trivial change for the developer to make when needed.

---

What about the 120 character line length? Is that also something you no longer care about? I can separate that into a different PR if you want that and not the `const`?


Anyway, both of these are less important - I only went into the `tslint.json` to begin with to turn on [`no-floating-promises`](https://palantir.github.io/tslint/rules/no-floating-promises/), which *is* in fact a 'hill I'm willing to die on' or at least argue more passionately for ;). Expect the PR for that imminently.

scheibo:
I've thought about it, and I've decided is that why I like `const` is that it makes code easier to read: if I'm reading a function and I see `const` variables declared, I can mentally make note that I don't need to watch them for changes etc. Each `let` adds another thing to track to figure out what's going on, and with PS!'s large and dense methods, this often becomes cumbersome.

What `prefer-const` provides is the guarantee that every time you see a `let` you know its state *is* going to change somewhere (as opposed to today, where it means it *might*, which just adds cognitive overhead for no benefit).

Zarel:
To be fair, I was the one who wrote "TODO: enable prefer-const". I certainly prefer it on... I'm just a bit nervous about enabling it for a codebase worked on with a lot of other devs who might not like an even pickier linter. I feel the same way about the line limit, which is why the line limit is a warning.

Zarel:
Especially since, while developing, it might be common to write to a variable, and then comment out the code that writes to it, and then not be able to run tests.

I might compromise on making it a warning, like the line limit?

scheibo:
> Especially since, while developing, it might be common to write to a variable, and then comment out the code that writes to it, and then not be able to run tests.

Ah, good point. I'm completely OK with making it a warning.

Zarel:
Also, can tslint auto-fix a `prefer-const` warning?

sim/battle.ts
```diff
@@ -353,7 +355,11 @@ export class Battle extends Dex.ModdedDex {
 		return super.getFormat(format, true);
 	}
 
-	addPseudoWeather(status: string | PureEffect, source: Pokemon | 'debug' | null = null, sourceEffect: Effect | null = null): boolean {
+	addPseudoWeather(
+		status: string | PureEffect,
+		source: Pokemon | 'debug' | null = null,
+		sourceEffect: Effect | null = null): boolean {
```
```suggestion
		sourceEffect: Effect | null = null
	): boolean {
```

sim/battle.ts
```diff
@@ -353,7 +355,11 @@ export class Battle extends Dex.ModdedDex {
...
+
```
```suggestion
```

sim/battle.ts
```diff
@@ -1835,10 +1858,12 @@ export class Battle extends Dex.ModdedDex {
 
 		if (damage) {
 			if (this.gen <= 1 && effect.recoil && source) {
-				this.damage(this.clampIntRange(Math.floor(damage * effect.recoil[0] / effect.recoil[1]), 1), source, target, 'recoil');
+				this.damage(
+					this.clampIntRange(Math.floor(damage * effect.recoil[0] / effect.recoil[1]), 1), source, target, 'recoil');
```
Zarel: Generally I prefer to reduce line lengths by introducing temporary variables. Temporary variables are also good as documentation.

```js
				const amount = this.clampIntRange(Math.floor(damage * effect.recoil[0] / effect.recoil[1]), 1);
				this.damage(amount, source, target, 'recoil');
```

sim/battle.ts
```diff
@@ -1835,10 +1858,12 @@ export class Battle extends Dex.ModdedDex {
...
 			}
 			if (this.gen <= 4 && effect.drain && source) {
-				this.heal(this.clampIntRange(Math.floor(damage * effect.drain[0] / effect.drain[1]), 1), source, target, 'drain');
+				this.heal(
+					this.clampIntRange(Math.floor(damage * effect.drain[0] / effect.drain[1]), 1), source, target, 'drain');
```
Zarel: Ditto (`const amount`)

sim/battle.ts
```diff
@@ -2147,22 +2172,24 @@ export class Battle extends Dex.ModdedDex {
...
-		let baseDamage = tr(tr(tr(tr(2 * level / 5 + 2) * basePower * attack) / defense) / 50);
+		const baseDamage = tr(tr(tr(tr(2 * level / 5 + 2) * basePower * attack) / defense) / 50);
 
 		// Calculate damage modifiers separately (order differs between generations)
 		return this.modifyDamage(baseDamage, pokemon, target, move, suppressMessages);
 	}
 
-	modifyDamage(baseDamage: number, pokemon: Pokemon, target: Pokemon, move: ActiveMove, suppressMessages: boolean = false) {
+	modifyDamage(
+		baseDamage: number, pokemon: Pokemon, target: Pokemon, move: ActiveMove, suppressMessages: boolean = false) {
```
```suggestion
		baseDamage: number, pokemon: Pokemon, target: Pokemon, move: ActiveMove, suppressMessages: boolean = false
	) {
```

sim/battle.ts
```diff
@@ -2147,22 +2172,24 @@ export class Battle extends Dex.ModdedDex {
...
+
```
```suggestion
```

sim/battle.ts
```diff
@@ -2241,13 +2268,14 @@ export class Battle extends Dex.ModdedDex {
...
+		const numSlots = source.side.active.length;
 		if (!Math.abs(targetLoc) && Math.abs(targetLoc) > numSlots) return false;
 
-		let sourceLoc = -(source.position + 1);
-		let isFoe = (targetLoc > 0);
-		let isAdjacent = (isFoe ? Math.abs(-(numSlots + 1 - targetLoc) - sourceLoc) <= 1 : Math.abs(targetLoc - sourceLoc) === 1);
-		let isSelf = (sourceLoc === targetLoc);
+		const sourceLoc = -(source.position + 1);
+		const isFoe = (targetLoc > 0);
+		const isAdjacent =
```
```suggestion
		const acrossFromTargetLoc = -(numSlots + 1 - targetLoc);
		const isAdjacent = (isFoe ?
			Math.abs(acrossFromTargetLoc - sourceLoc) <= 1 :
			Math.abs(targetLoc - sourceLoc) === 1);
```

sim/battle.ts
```diff
@@ -2241,13 +2268,14 @@ export class Battle extends Dex.ModdedDex {
...
+			(isFoe ? Math.abs(-(numSlots + 1 - targetLoc) - sourceLoc) <= 1 : Math.abs(targetLoc - sourceLoc) === 1);
```
```suggestion
```

sim/battle.ts
```diff
@@ -2830,7 +2863,8 @@ export class Battle extends Dex.ModdedDex {
 		} else if (action.choice === 'megaEvo' && this.gen >= 7) {
 			this.eachEvent('Update');
 			// In Gen 7, the action order is recalculated for a Pokémon that mega evolves.
-			const moveIndex = this.queue.findIndex(queuedAction => queuedAction.pokemon === action.pokemon && queuedAction.choice === 'move');
+			const moveIndex = this.queue.findIndex(
```
```suggestion
			const moveIndex = this.queue.findIndex(queuedAction =>
				queuedAction.pokemon === action.pokemon && queuedAction.choice === 'move'
			);
```

sim/battle.ts
```diff
@@ -2830,7 +2863,8 @@ export class Battle extends Dex.ModdedDex {
...
+				queuedAction => queuedAction.pokemon === action.pokemon && queuedAction.choice === 'move');
```
```suggestion
```

sim/battle.ts
```diff
@@ -3176,19 +3210,25 @@ export class Battle extends Dex.ModdedDex {
 		}
 	}
 
-	runMove(move: string | Move, target: Pokemon, targetLoc?: number, sourceEffect?: Effect | null, zMove?: string, externalMove?: boolean) {
+	runMove(
+		move: string | Move, target: Pokemon, targetLoc?: number,
+		sourceEffect?: Effect | null, zMove?: string, externalMove?: boolean) {
```
```suggestion
		sourceEffect?: Effect | null, zMove?: string, externalMove?: boolean
	) {
```

sim/battle.ts
```diff
@@ -3176,19 +3210,25 @@ export class Battle extends Dex.ModdedDex {
...
 		throw new Error(`The runMove function needs to be implemented in scripts.js or the battle format.`);
 	}
 
-	useMove(move: string | Move, pokemon: Pokemon, target?: Pokemon | null, sourceEffect?: Effect | null, zMove?: string): boolean {
+	useMove(
+		move: string | Move, pokemon: Pokemon, target?: Pokemon | null,
+		sourceEffect?: Effect | null, zMove?: string): boolean {
```
```suggestion
		sourceEffect?: Effect | null, zMove?: string
	): boolean {
```

sim/battle.ts
```diff
@@ -3176,19 +3210,25 @@ export class Battle extends Dex.ModdedDex {
...
 	}
 
 	/**
 	 * target = undefined: automatically resolve target
 	 * target = null: no target (move will fail)
 	 */
-	useMoveInner(move: string | Move, pokemon: Pokemon, target?: Pokemon | null, sourceEffect?: Effect | null, zMove?: string): boolean {
+	useMoveInner(
+		move: string | Move, pokemon: Pokemon, target?: Pokemon | null,
+		sourceEffect?: Effect | null, zMove?: string): boolean {
```
```suggestion
		sourceEffect?: Effect | null, zMove?: string
	): boolean {
```

sim/battle.ts
```diff
@@ -3198,7 +3238,8 @@ export class Battle extends Dex.ModdedDex {
 
 	moveHit(
 		target: Pokemon | null, pokemon: Pokemon, move: string | Move,
-		moveData?: ActiveMove | SelfEffect | SecondaryEffect, isSecondary?: boolean, isSelf?: boolean): number | undefined | false {
+		moveData?: ActiveMove | SelfEffect | SecondaryEffect,
+		isSecondary?: boolean, isSelf?: boolean): number | undefined | false {
```
```suggestion
		isSecondary?: boolean, isSelf?: boolean
	): number | undefined | false {
```

sim/dex.ts
```diff
@@ -260,7 +260,10 @@ class ModdedDex {
 	 * Returns false if the target is immune; true otherwise.
 	 * Also checks immunity to some statuses.
 	 */
-	getImmunity(source: {type: string} | string, target: {getTypes: () => string[]} | {types: string[]} | string[] | string): boolean {
+	getImmunity(
+		source: {type: string} | string,
+		target: {getTypes: () => string[]} | {types: string[]} | string[] | string): boolean {
```
```suggestion
		target: {getTypes: () => string[]} | {types: string[]} | string[] | string
	): boolean {
```

sim/dex.ts
```diff
@@ -275,7 +278,10 @@ class ModdedDex {
 		return true;
 	}
 
-	getEffectiveness(source: {type: string} | string, target: {getTypes: () => string[]} | {types: string[]} | string[] | string): number {
+	getEffectiveness(
+		source: {type: string} | string,
+		target: {getTypes: () => string[]} | {types: string[]} | string[] | string): number {
```
```suggestion
		target: {getTypes: () => string[]} | {types: string[]} | string[] | string
	): number {
```

sim/dex.ts
```diff
@@ -260,7 +260,10 @@ class ModdedDex {
 	 * Returns false if the target is immune; true otherwise.
 	 * Also checks immunity to some statuses.
 	 */
-	getImmunity(source: {type: string} | string, target: {getTypes: () => string[]} | {types: string[]} | string[] | string): boolean {
+	getImmunity(
+		source: {type: string} | string,
+		target: {getTypes: () => string[]} | {types: string[]} | string[] | string): boolean {
+
```
```suggestion
```

sim/dex.ts
```diff
@@ -275,7 +278,10 @@ class ModdedDex {
 		return true;
 	}
 
-	getEffectiveness(source: {type: string} | string, target: {getTypes: () => string[]} | {types: string[]} | string[] | string): number {
+	getEffectiveness(
+		source: {type: string} | string,
+		target: {getTypes: () => string[]} | {types: string[]} | string[] | string): number {
+
```
```suggestion
```

sim/dex.ts
```diff
@@ -653,11 +659,13 @@ class ModdedDex {
 		let statName: keyof StatsTable;
 		for (statName in modStats) {
 			const stat = baseStats[statName];
-			modStats[statName] = Math.floor(Math.floor(2 * stat + set.ivs[statName] + Math.floor(set.evs[statName] / 4)) * set.level / 100 + 5);
+			modStats[statName] = Math.floor(Math.floor(2 * stat + set.ivs[statName] +
```
```suggestion
			const tr = this.trunc;
			modStats[statName] = tr(tr(2 * stat + set.ivs[statName] + tr(set.evs[statName] / 4)) * set.level / 100 + 5);
```

You'll need to also pull `trunc` out of `battle.ts` and into `dex.ts`.

sim/dex.ts
```diff
@@ -653,11 +659,13 @@ class ModdedDex {
...
+				Math.floor(set.evs[statName] / 4)) * set.level / 100 + 5);
 		}
 		if ('hp' in baseStats) {
 			const stat = baseStats['hp'];
-			modStats['hp'] = Math.floor(Math.floor(2 * stat + set.ivs['hp'] + Math.floor(set.evs['hp'] / 4) + 100) * set.level / 100 + 10);
+			modStats['hp'] = Math.floor(Math.floor(2 * stat + set.ivs['hp'] +
+				Math.floor(set.evs['hp'] / 4) + 100) * set.level / 100 + 10);
```
Zarel: Ditto. (`tr`)

sim/dex.ts
```diff
@@ -1265,7 +1278,8 @@ class ModdedDex {
 				return new TypeError(`${filePath}, if it exists, must export a non-null object`);
 			}
 			if (!dataObject[key] || typeof dataObject[key] !== 'object') {
-				return new TypeError(`${filePath}, if it exists, must export an object whose '${key}' property is a non-null object`);
+				return new TypeError(
+					`${filePath}, if it exists, must export an object whose '${key}' property is a non-null object`);
```
Zarel: I would whitelist `return new TypeError` in the long line regex.

sim/dex.ts
Zarel: (n.b. this might require some changes to how `trunc` is implemented; I remember there's some magic with how it works in custom games)

sim/dex.ts
scheibo: Its kind of a dense regex, so I'll maybe let you edit that one...

FWIW, I'm also not sure I'd want line length exceptions to be whitelisted - I think there should be so few places that breaking the line limit should be necessary that manual overrides should be sufficient (remember, I'm from a world where the automatic formatter decides how long my lines are and where they wrap!).

*(In news somewhat related to this change, I've come around and agree with you that I don't like what the publicly availably automatic code formatting options do to the PS! codebase, see https://github.com/Zarel/Pokemon-Showdown/compare/master...pkmn-cc:prettier if you want to see how much **less** prettier things become...)*

sim/dex.ts
scheibo: Done. It seems we use `Math.floor` in a lot of places we should be using `trunc`... (out of scope for this PR, but maybe worth fixing?)

tslint.json
```diff
@@ -12,13 +12,13 @@
...
-        // I do want to bring this down to 120 at some point but maybe later
+        "prefer-const": {
+            "severity": "warning"
+        },
         "max-line-length": {
             "severity": "warning",
             "options": {
-                "limit": 140,
+                "limit": 120,
                 "ignore-pattern": "\\.replace\\(('\\[|/)|^\\s*(// \\s*)?(return |(let |const )?[a-zA-Z0-9$.]+ \\+?= (\\$\\()?|[a-zA-Z0-9$.]+\\()?['\"`]"
```
Zarel:
```suggestion
                "ignore-pattern": "\\.replace\\(('\\[|/)|^\\s*(// \\s*)?(return |throw (new )?|(let |const )?[a-zA-Z0-9$.]+ \\+?= (\\$\\()?|[a-zA-Z0-9$.]+\\()?['\"`]"
```

Zarel:
```suggestion
                "ignore-pattern": "^\\s*(// \\s*)?((let |const )?[a-zA-Z0-9$.]+ \\+?= (\\$\\()?|(return |throw )?(new )?([a-zA-Z0-9$.]+\\()?)?['\"`/]"
```

Zarel: Sorry for delaying until a conflict appeared; I didn't have much free time yesterday.

scheibo: Why remove `replace`? It means the zalgo regex at sim/dex.ts:253 now is too long?

sim/dex.ts
```diff
@@ -880,7 +888,8 @@ class ModdedDex {
 			}
 		}
 		if (matches.length > 1) {
-			throw new Error(`More than one thing matches "${rule}"; please use something like "-item:metronome" to disambiguate`);
+			throw new Error(
+				`More than one thing matches "${rule}"; please use something like "-item:metronome" to disambiguate`);
```
Zarel: this can be put back

sim/dex.ts
```diff
@@ -690,7 +697,8 @@ class ModdedDex {
 			const spcDV = Math.floor(ivs.spa / 2);
 			return {
 				type: hpTypes[4 * (atkDV % 4) + (defDV % 4)],
-				power: Math.floor((5 * ((spcDV >> 3) + (2 * (speDV >> 3)) + (4 * (defDV >> 3)) + (8 * (atkDV >> 3))) + (spcDV % 4)) / 2 + 31),
+				power: Math.floor(
+					(5 * ((spcDV >> 3) + (2 * (speDV >> 3)) + (4 * (defDV >> 3)) + (8 * (atkDV >> 3))) + (spcDV % 4)) / 2 + 31),
```
Zarel: All the `Math.floor`s here can be `~~`, which is hopefully enough to fit.

tslint.json
Zarel: The zalgo regex should be:

```js
		name = name.replace(
			/[\u0300-\u036f\u0483-\u0489\u0610-\u0615\u064B-\u065F\u0670\u06D6-\u06DC\u06DF-\u06ED\u0E31\u0E34-\u0E3A\u0E47-\u0E4E]{3,}/g,
			''
		);
```

sim/dex.ts
scheibo: There's one `Math.floor(...)` on the too long line, and replacing it with `(~~...)` its still too long and also becomes obsfuscated (the parens are required here, but even without them its still too long).

Zarel: They should still be changed to `trunc`.

sim/battle-stream.ts
```diff
@@ -81,6 +81,7 @@ export class BattleStream extends Streams.ObjectReadWriteStream {
 	}
 
 	_writeLine(line: string) {
+		// tslint:disable-next-line:prefer-const
```
Zarel:
```
"prefer-const": [true, {"destructuring": "all"}]
```

https://palantir.github.io/tslint/rules/prefer-const/

Zarel: In general, I _hate_ using `tslint:disable` and equivalents, and will prefer basically anything else.

lib/crashlogger.ts
```diff
@@ -28,14 +28,14 @@ export = function crashlogger(error: Error | string, description: string, data:
 	let stack = typeof error === 'string' ? error : error.stack;
 	if (data) {
 		stack += `\n\nAdditional information:\n`;
-		for (let k in data) {
+		for (const k in data) {
 			// @ts-ignore
```
Zarel: I think I'd rather just define this as `data: AnyObject | null`

sim/dex.ts
```diff
@@ -1160,8 +1182,9 @@ class ModdedDex {
 			if (j < 0) return null;
 			const ability = buf.substring(i, j);
 			const template = dexes['base'].getTemplate(set.species);
-			// @ts-ignore
-			set.ability = (template.abilities && ['', '0', '1', 'H'].includes(ability) ? template.abilities[ability || '0'] : ability);
+			set.ability = (template.abilities && ['', '0', '1', 'H'].includes(ability)
+				// @ts-ignore
+				? template.abilities[ability || '0'] : ability);
```
```js
			set.ability = ['', '0', '1', 'H', 'S'].includes(ability) ?
				template.abilities[ability as '0' || '0'] || (ability === '' ? '' : '!!!ERROR!!!') :
				ability;
```

is the corresponding client code, incidentally.

The server code can probably be updated to match.

sim/dex.ts
```diff
@@ -1265,7 +1288,8 @@ class ModdedDex {
 				return new TypeError(`${filePath}, if it exists, must export a non-null object`);
 			}
 			if (!dataObject[key] || typeof dataObject[key] !== 'object') {
-				return new TypeError(`${filePath}, if it exists, must export an object whose '${key}' property is a non-null object`);
+				return new TypeError(
+					`${filePath}, if it exists, must export an object whose '${key}' property is a non-null object`);
```
Zarel: (This is not exactly resolved in the sense that I would still like it reverted.)

sim/pokemon.ts
```diff
@@ -1498,7 +1511,8 @@ export class Pokemon {
 		linkedStatus = linkedStatus.toString();
 		for (const linkedPoke of linkedPokemon) {
 			if (linkedPoke.volatiles[linkedStatus]) {
-				linkedPoke.volatiles[linkedStatus].linkedPokemon.splice(linkedPoke.volatiles[linkedStatus].linkedPokemon.indexOf(this), 1);
+				linkedPoke.volatiles[linkedStatus].linkedPokemon.splice(
+					linkedPoke.volatiles[linkedStatus].linkedPokemon.indexOf(this), 1);
```
Zarel:
```suggestion
					linkedPoke.volatiles[linkedStatus].linkedPokemon.indexOf(this), 1
				);
```

Zarel:
```js
	removeLinkedVolatiles(linkedStatus: string | Effect, linkedPokemon: Pokemon[]) {
		linkedStatus = linkedStatus.toString();
		for (const linkedPoke of linkedPokemon) {
			const volatileData = linkedPoke.volatiles[linkedStatus];
			if (!volatileData) continue;
			volatileData.linkedPokemon.splice(volatileData.linkedPokemon.indexOf(this), 1);
			if (volatileData.linkedPokemon.length === 0) {
				linkedPoke.removeVolatile(linkedStatus);
			}
		}
	}
```

alternate `removeLinkedVolatiles`

sim/side.ts
```diff
@@ -190,7 +190,8 @@ export class Side {
 		return this.battle.sample(actives);
 	}
 
-	addSideCondition(status: string | Effect, source: Pokemon | 'debug' | null = null, sourceEffect: Effect | null = null): boolean {
+	addSideCondition(
+		status: string | Effect, source: Pokemon | 'debug' | null = null, sourceEffect: Effect | null = null): boolean {
```
```suggestion
		status: string | Effect, source: Pokemon | 'debug' | null = null, sourceEffect: Effect | null = null
	): boolean {
```

sim/side.ts
```diff
@@ -538,7 +540,8 @@ export class Side {
 			}
 			if (this.choice.switchIns.has(pos)) {
 				if (autoFill) continue;
-				return this.emitChoiceError(`Can't choose for Team Preview: The Pokémon in slot ${pos + 1} can only switch in once`);
+				return this.emitChoiceError(
+					`Can't choose for Team Preview: The Pokémon in slot ${pos + 1} can only switch in once`);
```
Zarel: New regex allows this form; this can be reverted.

sim/side.ts
```diff
@@ -600,7 +603,8 @@ export class Side {
 
 	choose(input: string) {
 		if (!this.currentRequest) {
-			return this.emitChoiceError(this.battle.ended ? `Can't do anything: The game is over` : `Can't do anything: It's not your turn`);
+			return this.emitChoiceError(
+				this.battle.ended ? `Can't do anything: The game is over` : `Can't do anything: It's not your turn`);
```
Zarel: New regex allows this form; this can be reverted.

Zarel:
```suggestion
				this.battle.ended ? `Can't do anything: The game is over` : `Can't do anything: It's not your turn`
			);
```

sim/team-validator.ts
```diff
@@ -87,7 +87,7 @@ export class Validator {
 		}
 
 		for (const [rule] of ruleTable) {
```
Zarel:
```suggestion
		for (const rule of ruleTable.keys()) {
```

sim/team-validator.ts
```diff
@@ -487,19 +493,20 @@ export class Validator {
 			(format.requirePlus || format.requirePentagon)) {
 			problems.push(`${name} has Secret Sword, which is only compatible with Keldeo-Ordinary obtained from Gen 5.`);
 		}
-		if (!lsetData.sources && lsetData.sourcesBefore <= 3 && dex.getAbility(set.ability).gen === 4 && !template.prevo && dex.gen <= 5) {
+		if (!lsetData.sources && lsetData.sourcesBefore <= 3 &&
+			dex.getAbility(set.ability).gen === 4 && !template.prevo && dex.gen <= 5) {
```
Zarel: I tend to strongly prefer intermediate variables, when code is too long to fit on one line. It has the added benefit of the intermediate variable's name being self-documenting.

```suggestion
		const hasGen3moves = !lsetData.sources && lsetData.sourcesBefore <= 3;
		if (hasGen3moves && dex.getAbility(set.ability).gen === 4 && !template.prevo && dex.gen <= 5) {
```

sim/team-validator.ts
```diff
@@ -439,12 +444,13 @@ export class Validator {
 				if (lsetData.sources.length > 1) {
 					problems.push(`${name} has an event-exclusive move that it doesn't qualify for (only one of several ways to get the move will be listed):`);
 				}
-				let eventProblems = this.validateSource(set, lsetData.sources[0], template, ` because it has a move only available`);
+				const eventProblems = this.validateSource(
+					set, lsetData.sources[0], template, ` because it has a move only available`);
```
Zarel:
```suggestion
					set, lsetData.sources[0], template, ` because it has a move only available`
				);
```

scheibo:
Thanks for solidifying why I appreciate automatic formatters in this review - formatting choices become completely automatic instead of the subjective 'whatever the project owner thinks in his head looks better' (I'm in no way disputing any of your edits - they all are improvements, it just doesn't scale - there's no way collaborators will always be able to tell what your preference is).

Zarel:
None of the "break things out into their own variable" improvements can be done automatically.

The "ending line should be same indentation as starting line" changes probably can be linted if we cared enough to figure out the relevant linter rule. I consider them relatively nitpicky.

(eslint does have `function-paren-newline: "consistent"` which is some of what I want)

Zarel:
So, yeah, I don't actually think there are subjective things improvable with an autoformatter, in any of my suggested changes, here. I only suggest two kinds of changes:

1. AST changes that an autoformatter wouldn't be able to handle
2. Changes according to rules simple enough to make a linter enforce, if we cared enough, and in no way subjective

Slayer95:
Shouldn't we enable ``prefer-const`` in ESLint then? It's autofixable.

scheibo:
Its autofixable in both. I'd be happy to see it enabled in ESLint as a warning as well.

Zarel:
I'd rather not. It's easy enough to do it during the transition to `.ts` files.

PR #5296, opened 2019-03-10
Randbats: Ensure that Genesect's forme matches its drive if it has one

MacChaeger:


Zarel:
Randbats should be able to generate any forme of any Pokémon... this'll need to be fixed at some point.

TheImmortal:
> 
> 
> Randbats should be able to generate any forme of any Pokémon... this'll need to be fixed at some point.

You do know that Genesect form is literally a sprite difference and being stuck with a coverage move it doesn't need at the loss of an item? People complained about these shitty Genesect sets and so they were removed. Formes with actual differences I can agree with.

Zarel:
Yes, I also know that Unown is strictly outclassed by like any other Pokémon.

Zarel:
I don't know, it seems wrong to just not use a Pokémon at all just because it's bad.

TheImmortal:
I don’t think you understand how seriously Random Battle is taken. People play this competitively. We’re already forced to use f’n joke sets.

It’s not even a Pokémon. It’s literally an item. Every Pokémon is in Randoms (except I think Castform-Snow but I can fix that).

Zarel:
Maybe the default should be Battle Factory. :|

You can play competitively with suboptimal Pokémon. That always seemed like the point of randbat to me.

TheImmortal:
We do have suboptimal sets, or sets that you don’t see competitively you could say. Douse Drive isn’t optimal, for example. But it isn’t a literal handicap on Genesect as the other Drives are.

If you want the default format to emphasize the official Smogon tiers more, then yes, Battle Factory should be default. But I think Random Battle appeals to the casual audience more. While we try to make it competitive, it’s still fun at the same time. We’re not Battle Factory level of strictness, and not Challenge Cup level of randomness.

Zarel:
https://dex.pokemonshowdown.com/moves/technoblast

Techno Blast was buffed to 120 power, so it's not _strictly_ outclassed by Thunderbolt and Flamethrower and Ice Beam anymore. It might be weaker than other sets but it still seems like a Pokémon to me.

I hope we're not also excluding Darmanitan-Z and Meloetta-P on the grounds that they're weaker than other sets.

MacChaeger:
I've played over 2000 games of Gen 7 Randbats and I've never seen a Darm-Z.

TheImmortal:
Meloetta-P is actually a good set, it's always been in.

Zarel:
It seems the best way to deal is just to raise the level on handicap sets if they're that much of a handicap.

Zarel:
Incidentally, I've heard Belly Drum Darm-Z can be interesting. Not a great set but has some surprise value, and sometimes that's the best you can do in randbats.

TheImmortal:
Absolutely not. That's akin to complex bans, they'll be no end. In addition, levels would be arbitrarily picked.

I wouldn't be completely opposed to Darmantian-Zen. I'll see if I can come up with a decent set.

Zarel:
I mean, we're talking about tiering formes separately, a time-honored tradition.

Zarel:
And imo way better than ability bans and move bans.

PR #5297, opened 2019-03-10
Fix 'floating' promises and enable lint errors going forward

scheibo:
If no one uses `catch(...)` on a `Promise` (or `try..catch` around an `await` in an `async` function), the exception only gets handled by a global `process.on('unhandledRejection', ...)` handler after a cycle of the event loop which gives the caller no way of handling the error. ie/ **exceptions thrown in promises do not propagate the same way as you might expect based on how exceptions in synchronous code propagate.**

The ways to fix this and avoid leaving 'orphaned'/'dangling'/'floating' promises are:
1) `return` the promise so it gets chained correctly. **THIS CHANGES THE BEHAVIOR** - downstream listeners now **wait** for the returned promise to complete instead (ie, execute *after* instead of *concurrently*). This may or may not be desirable.
2) provide a mechanism to propagate the failure through a separate 'error channel'.
3) log (`console.error`) or ignore the warning with `// tslint-disable`

---

- I used (1) for `lib/streams.ts`, but this may be incorrect, it may want to use (2) instead. 
- (1) isn't an option for `sim/battle-streams.ts` because the promises are in the `constructor` or methods hidden from the caller, so the (2) works well here, and solves my problems in #5278. 
- Logging the `sim/battle-stream-example.ts` seems sufficient because its just a CLI example.
- `lib/process-manager.ts` gave me problems so I punted with (3). (1) is not really an option. **Given the nature of the code, what do we want to do when an error happens?** swallow it (ie/ current behavior today, in which case the comments can stay and ill remove `FIXME`), log (`console.error`), ???.

Having `tslint` cover this for us going forward is important as its very easy to have errors slip through the cracks without it. Its very easy to turn off with a comment in places where we really do intend to orphan the promise.


sim/battle-stream.ts
```diff
@@ -225,11 +225,12 @@ export class BattlePlayer {
...
 	readonly debug: boolean;
 
-	constructor(playerStream: Streams.ObjectReadWriteStream, debug: boolean = false) {
+	constructor(
+		playerStream: Streams.ObjectReadWriteStream, debug: boolean = false, errorHandler?: (e: any) => void) {
 		this.stream = playerStream;
 		this.log = [];
 		this.debug = debug;
-		this.listen();
+		this.listen().catch(errorHandler || ((e: any) => { throw e; }));
```
Zarel: I'm not sure what this catch actually adds here? It just turns an unhandled promise rejection into an unhandled error.

Zarel: Oh, error handler. Like I said on Discord, I think the correct call here is to make the stream itself forward the error.

Zarel:
"Propagate the error through the stream" seems like the only correct way to deal with it.

scheibo:
> "Propagate the error through **the** stream" seems like the only correct way to deal with it.

How do you mean, exactly. 'The' stream (`BattleStream`) doesn't have an error propagation mechanism, which is what this PR adds (via `errorHandler`). Maybe you mean something like using an _additional_ stream like I'm doing in d8b4ae6a96b624154f204900fe868833e89dc8fe? Should clients pass it in like (an `errorStream` as opposed to `errorHandler`?) or should it be an inherent field within `BattleStream` (and another stream returned by `getPlayerStreams` etc)? Thats `battle-stream.ts` though, what about `process-manager.ts` etc?

FWIW, my original logic behind the 'simpler' `errorHandler` vs an error stream was then 'you don't need a dependency on streams, clients can choose whether or not to use streams for errors' (which is what I chose to do in my harness code), but thats kind of stupid logic because all the things taking `errorHandler` are dealing with streams anyway, so adding another concept just complicates things.

scheibo:
Or like maybe all streams in `lib/streams` have an `error` stream to them as well that you can subscribe to? So a `ReadWriteStream` is really a read, write AND error stream?

Zarel:
I'm saying `ReadStream` and `ObjectReadStream` itself should have an error propagation mechanism added to it. The errors would be stored in an error queue, and passed as promise rejections of `stream.read()`.

scheibo:
> I'm saying `ReadStream` and `ObjectReadStream` itself should have an error propagation mechanism added to it. The errors would be stored in an error queue, and passed out as promise rejections of `stream.read()`.

Given your knowledge of how `lib/streams.ts` works it seems like you might be able to do it much quicker than I could? IIUC though, while it does seem like the correct approach for the streams library to take, many of the changes in this commit would still be required or need alternative fixes, even if `streams` is fixed, unless I'm mistaken?

For example, its not clear to me how this would fix the case of `this.listen()`/`this._listen()` in a `constructor`: constructors can't be `async`, so without `await` or `.catch(...)` this still leaks errors? **The caller wrapping a `try..catch` around the constructor call doesn't catch errors thrown by `listen`.** 

Similarly, if the async block inside [getPlayerStreams](https://github.com/Zarel/Pokemon-Showdown/pull/5297/files#diff-6c85ba131ed11dc878ef70d51e43dac4L190) throws, it turns into a unhanded rejection regardless of any tweaks we make to `lib/streams.ts`?

Enhancing `lib/streams.ts` to throw also would do nothing about the dangling promises in `lib/process-manager.ts`?

Zarel:
The `listen` function itself could catch and forward the errors – you can put the `try...catch` inside the listen function. You'll have to lint-ignore the floating promise, though.

And `process-manager` is implemented via a `Stream` protocol, so as far as I'm aware it would solve that problem, too.

Zarel:
I guess I can try to find some time today to implement this into Streams.

scheibo:
> The `listen` function itself could catch and forward the errors –

Is that not exactly what I did, using `.catch(...)` in the constructor? You're just moving the `errorHandler` down a level? Or what mechanism would it use for forwarding (something built into `stream` somewhere, presumably?)

> And `process-manager` is implemented via a Stream protocol, so as far as I'm aware it would solve that problem, too.

Well, in many of the cases its the stream functions it uses that result in dangling promises (but not all). I think there's more going on with promise futures and `process-manager`, but I'm OK to punt on it because its not immediately relevant to my interests ;)

>  guess I can try to find some time today to implement this into Streams.

Awesome, I was about to ask you if you could at least try throwing up an example solution for one of the cases. I think we are talking past each other here, and maybe if you start implementing it you will see the concerns I have (or I'll see how you address them!).

**In particular, try having some move/ability/etc in `Battle` throw an error and try to have something using `BattleStream` be able to catch it and log.**

Zarel:
My main idea is that I don't think there's a need for a separate `errorHandler` callback when Promises already have a mechanism for propagating errors: rejecting the Promise. I'll try to show you what I mean today.

scheibo:
> when Promises already have a mechanism for propagating errors: rejecting the Promise

I agree and we should definitely be utilizing Promise rejection more in `streams`. This will solve like 80-90% of the problem if we do it right. The remaining bit (and which is probably the part I've been harping on, resulting in the source of us not being in sync here) comes from how rejected promises get treated by the interpreter at runtime once there's nothing chained to the promise/you leave an `async` function. The 'last mile' is where my issue lies.

Zarel:
I haven't had the time to test this yet, but this is what I meant:

https://github.com/Zarel/Pokemon-Showdown/commit/18dd3da86243576b1c41043966f2bc21c5612981

scheibo:
OK, returning to this again. I've patched in your change, removed my 'error handler' logic from `battle-stream.ts` and synced to `HEAD`. The new code certainly removes the unhandled promise rejection, but now the program termninates on error with a success status code and no message or stack trace, which if anything is worse. I'll start digging

scheibo:
`pushError` pushes the error throw to the `BattleSteam`s `errorBuf`, but the streams set up by `getPlayerStreams` are different stream objects so don't ever receive the error.

Zarel:
Oh, yeah, the individual streams need to handle the error. I haven't really done that part yet.

I'm surprised not handling them leads to errors being swallowed, though. It should still have raised an unhandled promise rejection, just in a different place.

scheibo:
I fixed it

Zarel:
How does the code detect that no further code will execute?

OH!

Zarel:
`readError` needs to happen after `loadIntoBuffer`, and `loadIntoBuffer` should exit when the error queue is non-empty.

scheibo:
85e5f75fafce665839fb35a9a6bbc6755a12eaf1 makes it so that my harness catches errors that occur when they're thrown in battle.

Zarel:
Maybe. I'm going to push my changes directly to master, though, so `git blame` in the future can find the correct person responsible for the streams library.

scheibo:
With `getPlayerStreams` it makes sense where we forward the error, but I'm having trouble reasoning about the `listen` in `BattlePlayer` and `BattleTextStream` (ie. https://github.com/Zarel/Pokemon-Showdown/pull/5297/files#diff-6c85ba131ed11dc878ef70d51e43dac4R241). The only stream we have to `pushError` to if we caught it would be the very same stream that we got the error from in the first place, which isn't very useful. Either we need a different place to send the error, or we need to explain in the place where we disable the `no-floating-promises` why doing so is safe (ie/ who is going to handle the error)

Zarel:
On second thought, can you rebase on top of `stream-error-propagate`? It'd be nice for you to do your own testing before it gets pushed to master.

scheibo:
> On second thought, can you rebase on top of stream-error-propagate? It'd be nice for you to do your own testing before it gets pushed to master.

Merged in `Zarel/master` to get the Streams typing changes and then cherry-picked `Zarel/streams-error-propagate` (rebasing was a bitch), so I think I should have all the stream fixes. I reverted my change reordering around `Monitor` in `battle-stream.ts` on this branch that was previously necessary as well. I then merged this branch with my `pre/harness` branch for testing, and it worked! I think if you land `streams-error-propagate` and I land the changes in this PR, we should be good.

Zarel:
> Merged in `Zarel/master` to get the Streams typing changes and then cherry-picked `Zarel/streams-error-propagate` (rebasing was a bitch), so I think I should have all the stream fixes.

It's confusing to me that this would be easier than rebasing. In theory, both approaches should have the exact same merge conflicts. I guess if you rebase, you have to handle them one at a time instead of all at once, but in practice it shouldn't be too much of a difference.

scheibo:
Rebase makes me go through multiple rounds (which I hate), as well as handle merges that dont exist if you do it at all once. It *shouldn't* make a difference, but it was way easier in this case, and almost always is IME

Zarel:
Ah, yes, I think it's because your Git history tends to be less clean than mine – in PR branches, I'll `reset --hard` instead of reverting, and `commit --fixup` instead of making fix commits. I guess different philosophies – "preserve all the history" vs "make the history easy to read/review".

scheibo:
My philosophy on a working branch is "commit all my changes ever, even if I just immediately `revert` it or `reset hard`" to then get rid of them, so if I screw up, I can always `git reflog` to save my ass. I 100% believe in the clean history on `master`, but thats already taken care of these days with how Github lets your squash rebase to merge a PR.

Case in point - i tried rebasing onto master just now since you've committed your branch and it gave me conflicts. I aborted and merged, no conflicts.

---

Anyway, I believe committing this branch as it currently stands will improve the state of the world and unblock my harness, but I'm still unhappy with my all the `tslint:disable FIXME` I had to do (especially in `battle-stream.ts` with `listen`). Do you have a solution for that (or an explanation I can put in a comment)?

sim/battle-stream-example.ts
```diff
@@ -98,7 +98,7 @@ console.log("p2 is " + p2.constructor.name);
 	while ((chunk = await streams.omniscient.read())) {
 		console.log(chunk);
 	}
-})();
+})().catch(err => console.error(err));
```
Zarel: This seems unnecessary. Uncaught Promise rejections are already printed to console, right?

Zarel:
I would just not enable `no-floating-promises` in TSLint.

We don't have try-blocks for all our thrown errors (and possible runtime errors are unbounded); nor do I think we need catch handlers for all our Promises that might reject. In both cases, we catch the ones we want to catch; the ones we don't catch will go to our global error handlers (logging to console, among other things) and then we'll fix the bug. I don't think anything needs to be changed about that status quo.

Zarel:
I do see the value in making exceptions in Promises _easier_ to catch as the need arises, but I think "log the exception to console" isn't bad as a default, so the linter rule doesn't seem necessary.

sim/battle-stream-example.ts
Zarel: In general, if a linter forces code that obviously isn't best practice, I prefer to use a linter-ignore comment, rather than do the non-best-practice thing.

Zarel:
> My philosophy on a working branch is "commit all my changes ever, even if I just immediately `revert` it or `reset hard`" to then get rid of them, so if I screw up, I can always `git reflog` to save my ass. I 100% believe in the clean history on `master`, but thats already taken care of these days with how Github lets your squash rebase to merge a PR.

My philosophy on a working branch is also "commit + reset hard" so it's in my reflog, but not "commit + revert". It's mostly about fewer commits being easier to code review, easier to see at a glance which commits are and aren't intended to end up in master, and easier to rebase (rebasing is useful to me, because it replays the individual steps that caused the conflict, instead of just mushing all the changes together).

sim/battle-stream-example.ts
scheibo: How uncaught promises are treated depends on the Node version, and at the moment an error message is logged to the console but in a different way than an uncaught exception (and this wasn't the behavior in the past). In this particular case, I would argue that the `catch` is in fact best practice, and adding a `// tslint:disable` would accomplish a similar thing in this case but would be less explicit.

That being said - I could care less about whether we `// tslint:disable` or `catch` here. This is not an important thing to get bogged down about in this PR.

scheibo:
> I would just not enable `no-floating-promises` in TSLint.

I am **strongly** opposed to this. This is about correctness, not simply a style issue like `prefer-const` or `line-length`. I don't want to make an [*argumentum ab auctoritate*](https://en.wikipedia.org/wiki/Argument_from_authority), but in my experience at work when we switched our APIs over to return `ListenableFutures` in Java (analogous to Promises), countless bugs were the result of orphaned futures (ie. dangling promises). In this very project we already know dangling promises resulted in bugs with `sim/battle-stream.ts` and `lib/streams.ts` error propagation that blocked #5278. Perhaps more importantly than error propagation is that forcing the developer to avoid dangling promises prevents race-related bugs which are among the most difficult and pernicious to track down. Chaining Promises properly *matters*, and `no-floating-promises` makes this significantly easier.

> We don't have try-blocks for all our thrown errors (and possible runtime errors are unbounded);

This is completely unrelated. You don't *need* `try` blocks for every error - given how error propagation works, code can still decide to still catch the error at a higher level if it wanted to. Promises don't work like that - if you don't catch and propagate the error manually, you *can't* catch it at a higher level, period.

> In both cases, we catch the ones we want to catch; the ones we don't catch will go to our global error handlers (logging to console, among other things) and then we'll fix the bug. I don't think anything needs to be changed about that status quo.

See above - Promise error propagation is different, thus error handling needs to be different. But I want to reiterate - `no-floating-promises` is *not* just about ensuring error propagation - avoiding races is by far the more important benefit.

> I do see the value in making exceptions in Promises easier to catch as the need arises, but I think "log the exception to console" isn't bad as a default, so the linter rule doesn't seem necessary.

You're getting distracted by the one case in the PR where I'm logging to the `console` because thats how the error should be handled at that level. That's by far the least interesting and important part of this PR.

Zarel:
> I don't want to make an _argumentum ab auctoritate_

I wouldn't call it authority, necessarily. You're speaking from experience, and that's probably the most reliable thing to base a technical decision on.

And, like, for the record, I trust you. You seem kind of... worried, I guess? I'm not going to force you to turn off `no-floating-promises` if you want it on, I just want to discuss your reasoning.

> In this very project we already know dangling promises resulted in bugs with `sim/battle-stream.ts` and `lib/streams.ts` error propagation that blocked #5278.

I wouldn't call them bugs; they seemed more like missing features to me, and they were missing features that seemed pretty easy to add when they were needed.

> Perhaps more importantly than error propagation is that forcing the developer to avoid dangling promises prevents race-related bugs which are among the most difficult and pernicious to track down. Chaining Promises properly _matters_, and `no-floating-promises` makes this significantly easier.

Interestingly enough, I don't think I've ever had Promise-race-related bugs, and we've been using Promises for a while. Intuitively, I feel like we either need the return value or we don't. If we need the return value, we'll wire everything up and it'll work fine (un-dangling Promises at this point doesn't feel "too late" to me, it feels like "the right time"), and if we don't, it's never been a big deal.

I don't think PS has actually had many race conditions in general – I think we're relatively good at, like, making sure separate threads/processes aren't sharing resources, and communicating exclusively through message passing.

> This is completely unrelated. You don't _need_ `try` blocks for every error - given how error propagation works, code can still decide to still catch the error at a higher level if it wanted to. Promises don't work like that - if you don't catch and propagate the error manually, you _can't_ catch it at a higher level, period.

I mean, you _can_, you just have to un-dangle the Promises you need.

That only works for synchronous code, anyway. We get crashes in stuff like SockJS all the time that we have no way to catch except with a global listener.

> You're getting distracted by the one case in the PR where I'm logging to the `console` because thats how the error should be handled at that level. That's by far the least interesting and important part of this PR.

I mean, in your list of solutions, #3 is "log the error".

You're telling me that #1 should be the default choice and #3 should be so rare that we should `tslint-disable` everywhere it comes up. Which, I mean, I do understand. My ideal solution here is a `spawn` keyword, an opt-in to "dangling this Promise is intentional", because the corresponding `tslint-disable` sigil is really hard to memorize.

It's also too bad dangling Promises is often necessary at the top level, because top-level `await` has opponents whose arguments never made _any_ sense to me (the arguments are something like "top-level `await` is bad in modules", which is like zero reason not to do it in a hashbang script).

But otherwise, I don't really have an objection.

lib/fs.ts
```diff
@@ -164,7 +164,7 @@ class FSPath {
 			return;
 		}
 
-		this.writeUpdateNow(dataFetcher, options);
+		return this.writeUpdateNow(dataFetcher, options);
```
Zarel: This one shouldn't be returned, though.

The old implementation of this function was actually async, and the doc comment said "don't use the returned Promise for anything; it's meaningless".

After this function was refactored into `writeUpdateNow`/`checkNextUpdate`/`finishUpdate`, it intentionally no longer returns anything. This return value would in fact still be meaningless.

(`writeUpdateNow`'s return value is irrelevant, and I don't have a good sense for whether or not `checkNextUpdate` should return a `Promise`.)

Zarel:
```suggestion
		// tslint:disable-next-line:no-floating-promises
		this.writeUpdateNow(dataFetcher, options);
```

Zarel: As a recap: `writeUpdate` will write an update the next time it's safe to do so.

If a previous call to `writeUpdate` hasn't finished writing, it will wait for the previous write to finish.

If `writeUpdate` is called multiple times while a previous call hasn't finish, it will only write the most recent update.

`writeUpdateNow` is only the case where there is no previous call. Returning it would only sometimes mean "this update has finished writing".

You _could_ rewrite this function to return a Promise accurately tracking when the update has finished writing, but it would take a lot more code and work, and I don't see a pressing need to support that feature, relative to the amount of work it would require.

lib/fs.ts
```diff
@@ -192,7 +192,7 @@ class FSPath {
...
 	}
 	finishUpdate() {
```
Zarel: There's a dangling Promise in this function, by the way. Apparently TSLint doesn't recognize `Promise<void> | void` as a Promise.

lib/fs.ts
```diff
@@ -178,7 +178,7 @@ class FSPath {
 			throttleTimer: null,
 		};
 		pendingUpdates.set(this.path, update);
-		this.safeWrite(dataFetcher(), options).then(() => this.finishUpdate());
+		return this.safeWrite(dataFetcher(), options).then(() => this.finishUpdate());
```
Zarel: Now here's the question – do you actually want to return this Promise? It's also meaningless – under normal conditions, this Promise will never resolve; it's an infinite loop repeatedly writing the latest update.

Zarel: But sometimes it does resolve, if nothing needs to be updated for a while. Nothing good would happen if someone tried to await this Promise.

Zarel: Similar concerns apply to the next two functions, by the way. The Promises you're newly returning are meaningless – all the meaningful Promises already being used.

sim/battle-stream.ts
```diff
@@ -220,6 +220,7 @@ export function getPlayerStreams(stream: BattleStream) {
 				streams.p1.push(p1Update);
 				const p2Update = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n([^\n]*)\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
 				streams.p2.push(p2Update);
+				// TODO p3/p4?
```
scheibo: @MacChaeger shouldn't we be doing something with the `p3` and `p4` streams here?

sim/battle-stream.ts
```diff
@@ -250,9 +255,9 @@ export class BattlePlayer {
 		this.stream = playerStream;
 		this.log = [];
 		this.debug = debug;
-		this.listen();
```
scheibo: If we want to maintain the 'start on creation' approach instead of an explicit `start` call (which I've renamed from `listen`, but I'd be happy to change back to `listen`), I would instead suggest we make the constructor private and add a static factory method which can then return a started instance and the promise in case someone cares:

```javascript
const {p1, started} = BattlePlayer.create(...);
// started.catch(err => ...)  if they want
```

sim/battle-stream.ts
MacChaeger: Per Zarel’s request, we’re still using a 4 channel system. `p3` and ``p4`` use channels 2 and 3.

MacChaeger: 2 and 3 of 1-4 that is

scheibo: I'm not sure how that works? Your teammate gets to see your exact HP etc in a 2v2? and what about free for all?

Regardless of how the `'|split'` is supposed to work - we set up a `ObjectReadWriteStream` for `p3` and `p4` here and don't push to them in the `update` case (only `sideupdate`/`end` and errors). Is this intentional? From your explanation it seems like we should be pushing the `p1update` to `p3` and the `p2update` to `p4`? Or does `update` never get used in multi-battles, in which case we should at least add a comment

MacChaeger: Yes, your teammate gets to see exact HP in 4 player battles, even when your teammate isn't really your teammate, because Zarel didn't think that mattered enough to use 6 channels. That said, I guess we probably should be pushing the updates like you suggested, though I'm not entirely sure what this function even gets used for.

scheibo: Hmm, not sure I agree with that? I'm going to open a different PR for this because its unrelated to `floating-promises` and I'd like to reserve this PR for Zarel and I to keep writing technical essays back and forth at each other. ;)

lib/fs.ts
Zarel: I return `Promise<T> | T` in a few places in the Streams API for perf reasons; in theory it's not hard to `await` or Promise.resolve` the result if you want a guaranteed Promise.

Zarel:
Incidentally, I really like Swift's approach to this: not handling the return value of a function is an error, unless either the caller or function actively marks the return value as ignorable. Much nicer than a linter comment, too.

sim/battle-stream.ts
Zarel: I wonder if we just make `BattlePlayer` extend a `Stream` and just foward errors to itself?

lib/fs.ts
Zarel: I mean, none of the problem is API design – the ideal API seems like common sense to me. If it's as easy as you say, feel free to pull request it – I can't think of any reason I'd reject it.

scheibo:
> Incidentally, I really like Swift's approach to this: not handling the return value of a function is an error, unless either the caller or function actively marks the return value as ignorable. Much nicer than a linter comment, too.

In Java we have [`@CheckReturnValue`/`@CanIgnoreReturnValue`](http://errorprone.info/bugpattern/CheckReturnValue), which is preferable to just special casing Promises for sure (though also requires a plugin). Does such a thing exist for NodeJS? [TSLint](https://palantir.github.io/tslint/rules/) doesn't appear to have anything like it, and searching for other static analysis tools which could provide such a thing did not bear any fruit.

Also, yeah, Swift is a better designed language. No arguments from me there. :)

sim/battle-stream.ts
scheibo: Hmm. What would it stream other than errors? And how would using it look? `new BattlePlayer().catch(err => ...)` instead of `new BattlePlayer.start().catch(err => ...)`? I'm not sure it buys a whole lot, and is somewhat confusing logically (a player seems like something which would act on the stream, not be a stream itself?) but could work.

Zarel: It could also be a native Node EventEmitter with error events.

scheibo: I'd actually rather stay away from that route if possible. I'd like to be able to run the entirety of `sim` purely in browser for my '[PocketMon](https://docs.google.com/document/d/1xKy1b1SExXqlb6KXfmfbuifVCjAgPUbnusekIZuKQ8s/edit#heading=h.jb61j01iy1m4)' implementation)

Zarel: That's currently supported? I'm pretty surprised, but that makes me happy.

Zarel: Making `BattlePlayer` extend `Promise` might also be an option, but that might be excessive.

scheibo: I had it working at one point in my prototype, yeah (with some shimming etc). I mean, I bet a lot more could work given the low amount of dependencies (I could probably just change some of the socket stuff slightly and get it to play well as well?), but I did have two random bots playing each other on an android web view on my phone.

Zarel:
`no-unused-expression` is the kind of like that, by which I mean the exact opposite of that. :p

Zarel:
I don't like the `.start()` approach, but since I can't think of a better approach, I'm reluctantly accepting.

Issue #2978, opened 2016-12-02
Formally addressing the way PS goes about pull requests


panpawn:
**Pull Requests**
These are a major part of the PS development culture.  However, in my opinion, there's a huge issue with the way that some pull requests are handled in that they sometimes are merged into the master branch before they are necessarily ready for the production server or to otherwise go live.

Some people might think that there's nothing wrong with prematurely merged pull requests, and that we typically do a lot to prevent that, and generally I would agree.   However, this is _not_ always the case; sometimes, we don't seem to do the best that we could to prevent this from happening, based on my observations.

When a pull request is prematurely merged due to a lack of review, this can leave bugs or even crashes in the current version of PS, which otherwise could of been prevented.  This has happened several times before, but we should try to do the best to prevent this to our ability.

When we don't try to do the best we can to prevent this, it can cause either other developers or the developer who worked on the pull request to go back and try to find how to work on a fix, which often times takes that developer away from working on whatever it was they were previously.

In comparison, several other projects on GitHub will make it so that a pull request much get X number of approvals before being merged, or alternatively, must be open for a minimum period of time.  I believe that something to this effect would help the current issue of some pull requests just not being open long enough.

So, all of this comes down to several differing meanings of what a "pull request" is among the development team.

To me, a pull request are changes that someone is proposing to the project, but not necessarily the best way of going about those changes.  For this reason, I am a firm believer that in general, a pull request should be reviewed, preferably by several people if several people are currently working on that part of PS that is relating to said pull request.

A big part of being open-source is collaboration, and if a pull request isn't open long enough for people to actually help the collaboration of that change, that is straying from what I think a pull request is, which is probably why I feel this issue is important.
 
It has come to my attention that owners of repositories can actually set a minimum number of approvals before a PR can actually be merged, but I think we could just do a "thumb up" kind of a thing or something on the actual PR to simplify things or something, though.  

Also, somewhat unrelated to this but worthy to mention: I think we should be holding collaborators to the same contributing standards as everyone else.  Namely, this has to do with commit name standards specifically, which I have seen several instances where collaborators don't follow what is outlined in CONTRIBUTING.md.

The meaning of this issue is to think about the things that I've talked about, and to see if there is actually room for improvement in these areas, and if so, to give helpful feedback and your thoughts on the topics that I've outlined above.

With that being said, what does everyone else think?

panpawn:
[...]

PR #5299, opened 2019-03-11
Allow any user to ignore help tickets

scheibo:
Really, users below '%' should be ignoring help tickets completely by default - right now they show up in chat and more importantly, the preview hovertext leaks ticket information to unprivileged users who happen to be in the room. I would have thought https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-plugins/helptickets.js#L375 would cover this, but apparently not? (maybe it should be `'lock'`? Why do we check auth for `'lock`' in some cases and `'mute'` in others?)

### DIFF
--- a/server/users.js
+++ b/server/users.js
@@ -663,21 +663,18 @@ class User extends Chat.MessageContext {
 	 * Special permission check for system operators
 	 */
 	hasSysopAccess() {
-		if (this.isSysop && Config.backdoor) {
-			// This is the Pokemon Showdown system operator backdoor.
+		// This is the Pokemon Showdown system operator backdoor.
 
-			// Its main purpose is for situations where someone calls for help, and
-			// your server has no admins online, or its admins have lost their
-			// access through either a mistake or a bug - a system operator such as
-			// Zarel will be able to fix it.
+		// Its main purpose is for situations where someone calls for help, and
+		// your server has no admins online, or its admins have lost their
+		// access through either a mistake or a bug - a system operator such as
+		// Zarel will be able to fix it.
 
-			// This relies on trusting Pokemon Showdown. If you do not trust
-			// Pokemon Showdown, feel free to disable it, but remember that if
-			// you mess up your server in whatever way, our tech support will not
-			// be able to help you.
-			return true;
-		}
-		return false;
+		// This relies on trusting Pokemon Showdown. If you do not trust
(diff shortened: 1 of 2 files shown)


server/users.js
```diff
@@ -663,21 +663,18 @@ class User extends Chat.MessageContext {
...
-			// Zarel will be able to fix it.
-
-			// This relies on trusting Pokemon Showdown. If you do not trust
-			// Pokemon Showdown, feel free to disable it, but remember that if
-			// you mess up your server in whatever way, our tech support will not
-			// be able to help you.
-			return true;
-		}
-		return false;
+		// This is the Pokemon Showdown system operator backdoor.
```
scheibo: Completely unrelated, but noticed this when I was trying to grok how auth worked.

Zarel:
I would rather figure out why the original check doesn't work.

scheibo:
> I would rather figure out why the original check doesn't work.

Yeah, lets do that as well, but I also don't see why it should require privileges to ignore something which you shouldn't be seeing in the first place.

Zarel:
I think that's a good reason to require privileges.

Zarel:
For various reasons, but a big one is as a signal of "this isn't a feature that affects you".

scheibo:
> To me, a pull request are changes that someone is proposing to the project, but not necessarily the best way of going about those changes

_Originally posted by @panpawn in https://github.com/Zarel/Pokemon-Showdown/pull/2978_

I mainly just wanted to get the ball rolling to fix the leak I noticed, maybe @bumbadadabum can help me figure out why the original check doesn't work?

Zarel:
It's not a leak in that we trust anyone in the Staff room with that information.

Zarel:
```
>> Users('pre').can('mute', null, room)
<< false
```

I'm really confused, because the only time `ignoreTickets` is read is in

```js
if (user.can('mute', null, room) && !user.ignoreTickets)
```

Zarel:
WAIT

This isn't about you getting notified by them. This is about them appearing at all to you. Ignoring help tickets won't change that; it only ignores the notification.

Zarel:
Which is unfortunately intentional: Help tickets are a feature of the Staff room, they can't be easily removed from some screens but not others.

AustinXII:
A lot of staff go on alts in the staff room too and it would be disadvantageous to not display them to those users

PR #5274, opened 2019-03-08
Cleanup stat naming and types (storedStats, StatNameExceptHP)

scheibo:
> I would probably rename these to `storedStats` and `baseStoredStats`; to make it clearer that they aren't the Pokémon's "actual" stats.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5267_

PR #5301, opened 2019-03-12
Fix Stadium mod's stat modification logic

scheibo:
> Could this have affected the attack and speed loss for burn and paralysis in Gen 1 Stadium?
>
> Here is a replay https://replay.pokemonshowdown.com/gen1stadiumou-874684443
See turn 3 for speed being unaffected by paralysis and turn 31 for attack being unaffected by burn.

_Originally posted by @SamanthaLola in https://github.com/Zarel/Pokemon-Showdown/pull/5274#issuecomment-471769625_

I purposefully deleted this (https://github.com/Zarel/Pokemon-Showdown/pull/5274/files#diff-19490bbe318ddb0e323a7bd29127f453L12) because its identical to `getStat` in `gen1` which this mod `inherits` from. What did I miss about inheritance, why do I need to add this back?

### DIFF
--- a/data/mods/stadium/scripts.js
+++ b/data/mods/stadium/scripts.js
@@ -9,6 +9,14 @@ let BattleScripts = {
 	gen: 1,
 	// BattlePokemon scripts. Stadium shares gen 1 code but it fixes some problems with it.
 	pokemon: {
+		getStat(statName, unmodified) {
+			statName = /** @type {StatNameExceptHP} */(toId(statName));
+			// @ts-ignore - type checking prevents 'hp' from being passed, but we're paranoid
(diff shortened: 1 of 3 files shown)


xfix:
The issue is that it wasn't inherited, being a part of `pokemon` object.

scheibo:
> The issue is that it wasn't inherited

Yes, I did figure out as much :)

> being a part of `pokemon` object.

> `pokemon` needs `inherit: true`, just like old-gen moves, abilities, etc.

I think the correct answer to what my misunderstanding was is 'inheritance is not *deep*' (ie/ the fact that `BattleScripts` inherits `gen1` does not recursively seem to bring everything over)

PR #5302, opened 2019-03-12
Fix regressions in team generators

Slayer95:
- Reimplements 23573646d.
- The `side` parameter is required for Seasonal School Schemes and Seasonal Strikes Back, as well as #5303.

### DIFF
--- a/data/mods/gen6/random-teams.js
+++ b/data/mods/gen6/random-teams.js
@@ -1029,10 +1029,11 @@ class RandomGen6Teams extends RandomTeams {
 	}
 
 	/**
+	 * @param {PlayerOptions} [side]
 	 * @param {number} [depth]
 	 * @return {RandomTeamsTypes["RandomFactorySet"][]}
 	 */
-	randomFactoryTeam(depth = 0) {
+	randomFactoryTeam(side, depth = 0) {
 		let forceResult = (depth >= 4);
 
 		// The teams generated depend on the tier choice in such a way that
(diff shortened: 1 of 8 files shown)


Zarel:
Can you show me what a test failure looks like in a random team generation?

Slayer95:
I am afraid not. The issue is that it's possible to use the information from your own team: pokémon, moves, items, abilities, etc., and from that deduct what the team of your foe is. However, I don't have the calculations at hand anymore.

Zarel:
Oh, sorry, I got this mixed up with a different commit.

data/mods/gen6/random-teams.js
```diff
@@ -1029,10 +1029,11 @@ class RandomGen6Teams extends RandomTeams {
 	}
 
 	/**
+	 * @param {PlayerOptions} [side]
 	 * @param {number} [depth]
 	 * @return {RandomTeamsTypes["RandomFactorySet"][]}
 	 */
-	randomFactoryTeam(depth = 0) {
+	randomFactoryTeam(side, depth = 0) {
```
Zarel: I don't like having to pass a `side` param everywhere. Can you just stick the seed in the generator object? It should already have one, right?

Slayer95: > Can you just stick the seed/PRNG property in the generator object? It should already have one, right?

This change belongs to 25edaa3, which is unrelated to the PRNG, and that's what already happens.

``side`` is only used for the player name. I figured I'd restore the old team generator signature... However, an alternative is a ``setMeta()`` function that would set some ``.meta.playerName``.

Zarel: What's the player name used for?

Slayer95: As a seed to create a random team specific to the player

Old Seasonals:
- https://github.com/Zarel/Pokemon-Showdown/blob/old-seasonals/2013/scripts.js#L339
- https://github.com/Zarel/Pokemon-Showdown/blob/old-seasonals/2013/scripts.js#L424
- https://github.com/Zarel/Pokemon-Showdown/blob/old-seasonals/2014/scripts.js#L50

SSB Proposal:
- https://github.com/Zarel/Pokemon-Showdown/blob/5d5b1afde80840f9ea3c2b37488c81fda45f200e/data/mods/ssb/random-teams.js#L650-L653

Zarel:
```
> battle.inputLog
[ '>start {"seed":[3064,22480,24090,47930]}',
  '>player p1 {"name":"p1","seed":[44477,54230,45879,47132]}',
  '>player p2 {"name":"p2","seed":[3893,43553,65434,45120]}' ]
```

Oh! This is how you're doing the inputlog. Interesting.

Zarel:
Okay, I think I like this.

PR #5303, opened 2019-03-12
SSB: Add a superior chance of rolling the player's staffmon

Slayer95:
Should be fun :)

(Gotta rebase after #5302)

### DIFF
--- a/data/mods/ssb/random-teams.js
+++ b/data/mods/ssb/random-teams.js
@@ -19,7 +19,10 @@
 const RandomTeams = require('../../random-teams');
 
 class RandomStaffBrosTeams extends RandomTeams {
-	randomStaffBrosTeam() {
+	/**
+	 * @param {PlayerOptions} [side]
+	 */
+	randomStaffBrosTeam(side) {
 		/** @type {PokemonSet[]} */
 		let team = [];
 		/** @type {SSBSets} */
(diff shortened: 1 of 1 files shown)


xfix:
I don't know how I feel about it. One issue I see with that is that SSB is a ladder format, and this change is advantageous to staff members with good mons.

Slayer95:
Indeed. Please discuss!

AustinXII:
Wasn’t this a thing on one of the older versions? I don’t think it was an issue then and it actually guaranteed your mon

xfix:
I don't remember it being a case ever, I know I didn't have Xatu in every single of my battles in previous Super Staff Bros. formats.

Slayer95:
FWIW, I had this implemented in my own server version just like @AustinXII 's.

I'd argue that SSB is eminently a for-the-lols format. However, if the OM staff values the leaderboard results, they might have misgivings.

HoeenCoder:
SSB is used in competetive formats, namely the random battles team tour and other major randbats tours. I'd rather not skew team generation to possible help a specific player win a game.

force:
RBTT is over. dunno if that will play a factor in changing your mind but whatever

HoeenCoder:
I know its over now, but I don't want to remove a sense of balance, alot of work was put into quality control, and while were not perfect (that would be really hard to get) I think were in a good spot overall.

As a result I'm closing this. TD;LR I appreciate the thought and think its a nice idea, but I don't want to sacrifice balance/randomness of the format to get it (I dont want to skew games so some staff with high tier mons win more often because of it).

PR #5305, opened 2019-03-12
Add mafia themes

Snaquaza:


Zarel:
Mafia is Hoeen I think? @HoeenCoder?

server/chat-plugins/mafia-data.js
```diff
@@ -1705,6 +1719,27 @@ const themes = {
 		6: `Solo Assassin, Guard, Guard, Guard, Guard, King`,
 		7: `Solo Assassin, Guard, Guard, Guard, Guard, Guard, King`,
 	},
+	tms: `baby_too_much_scum`,
+	btms: `baby_too_much_scum`,
+	too_much_scum: `baby_too_much_scum`,
+	baby_too_much_scum: {
+		name: `Too Much Scum`,
```
HoeenCoder: Any reason the name dosent match the id?

Lucas-Meijer: Oh, that was honestly just because at first I made the name "Baby Too Much Scum" and having too much scum as an alias. Then I discovered Too Much Scum was a theme too so just edited it to say that and hold both setups

HoeenCoder: If its the same theme but two setups:
swap one or the other (id should match name, the other option can be an alias).

If its seperate themes:
It would probably be best to split this into two themes to avoid confusion.

PR #5306, opened 2019-03-12
Add autoconfirmed user metadata to /cmd userdetails

Kaiepi:


Zarel:
Wait... do we actually want this info being made public??? On second thought, we really don't, do we?

Slayer95:
It's sent as a boolean. Why not? In any case, it would make sense to make it available in the ``/whois`` for room auth.

Zarel:
I'm fine with showing roomauth who's autoconfirmed, but I don't think we want regular users with that information.

Zarel:
I guess it _is_ public, though. Based on reg date and battle wins, both of which are public.

PR #5307, opened 2019-03-12
Alias /forfeit as /ff

Slayer95:
Used in League of Legends.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -3787,6 +3787,7 @@ const commands = {
 	 * Battle commands
 	 *********************************************************/
 
+	ff: 'forfeit',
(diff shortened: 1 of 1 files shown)


Zarel:
We've removed all the synonyms of "forfeit" because people keep trying to trick other people into using them. I think that's a likely problem with this one, too.

Slayer95:
I figured there was a chance of that issue rising, but I didn't realize there was precedent of alias deletion :/

Slayer95:
I see. It's da5c916e065ce4f2394bc3396c795d52e677dab2. I am not sure if the meaning ``/ff`` is more or less intuitive than ``/concede`` ...

PR #5308, opened 2019-03-13
Random Battle: Fix Malamar

ascriptmaster:
Currently, Superpower is being rejected with RestTalk, but it shouldn't
do that if it is a Malamar, because of Contrary.


ascriptmaster:
(no need for a change to the gen 6 file because Malamar doesn't use RestTalk in that gen for randbats)

PR #5309, opened 2019-03-13
Improve /sp

TheJetOU:
This commit fixes/changes: 

- Fixes alts always having `undefined`

- Sorts expireTime to ascending order

- Moves some code to server/punishments.js

- Uses table instead of `<p>`

- Handles errors more correctly

(I had a local git branch that I commited when I did each fix/change with a proper description for ease of reviewning but my laptop froze and it got corrupted)

### DIFF
--- a/server/chat-plugins/info.js
+++ b/server/chat-plugins/info.js
@@ -2397,55 +2397,30 @@ const pages = {
 		let buf = "";
 		this.extractRoom();
 		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
-		buf += `<div class="pad"><h2>List of active punishments:</h2>`;
+		if (!this.room.chatRoomData) return;
 		if (!this.can('mute', null, this.room)) return;
-		if (!this.room.chatRoomData) {
-			return buf + `<div class="notice message-error">This page is unavailable in temporary rooms / non-existent rooms.</div>`;
-		}
-		const store = new Map();
-		const possessive = (word) => {
-			const suffix = word.endsWith('s') ? `'` : `'s`;
-			return `${word}${suffix}`;
-		};
 
-		if (Punishments.roomUserids.get(this.room.id)) {
-			for (let [key, value] of Punishments.roomUserids.get(this.room.id)) {
-				if (!store.has(value)) store.set(value, [new Set([value.id]), new Set()]);
-				store.get(value)[0].add(key);
+		if (Punishments.getPunishmentsOfRoom(this.room).length) {
+			buf += `<div class="pad"><h2>List of active punishments:</h2>`;
+			buf += `<table style="border: 1px solid black; border-collapse:collapse; width:100%;"><tr><th style="border: 1px solid black;">Username</th><th style="border: 1px solid black;">Punishment type</th><th style="border: 1px solid black;">Expire time</th><th style="border: 1px solid black;">Reason</th><th style="border: 1px solid black;">Alts</th>`;
+			if (user.can('ban')) buf += `<th style="border: 1px solid black;">IP</th>`;
+			buf += `</tr>`;
+			for (const entry of Punishments.getPunishmentsOfRoom(this.room)) {
+				Array.prototype.sort.call(entry, function(a, b) {return a.expiresIn - b.expiresIn;});
+				buf += `<tr>`;
+				if (!entry.alts) entry.alts = [];
+				if (!entry.ip) entry.ip = [];
+				let expireString = Chat.toDurationString(entry.expiresIn, {precision: 1});
+				let punishDesc = "";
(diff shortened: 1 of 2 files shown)


server/punishments.js
```diff
@@ -1559,6 +1559,46 @@ Punishments.getRoomPunishments = function (user, options) {
...
+	if (room.muteQueue) {
+		for (const entry of room.muteQueue) {
+			let expiresIn = new Date(entry.time).getTime() - Date.now();
+			if (expiresIn < 0) continue;
+			output.push({"punishType": 'MUTE', "id": entry.userid, "expiresIn": expiresIn});
+		}
+	}
+	return output.sort((a, b) => {
+		return a.expiresIn - b.expiresIn;
+	});
```
TheJetOU: I think I should probably move this code to the page

TheJetOU:
@bumbadadabum

server/punishments.js
bumbadadabum: Yeah

server/chat-plugins/info.js
```diff
@@ -2397,55 +2397,29 @@ const pages = {
...
-			buf += `<p>- ${possessive(id)} ${punishType.toLowerCase()} expires in ${expireString}.${punishDesc}</p>`;
-		}
 
-		if (this.room.muteQueue) {
-			for (const entry of this.room.muteQueue) {
-				let expiresIn = new Date(entry.time).getTime() - Date.now();
-				if (expiresIn < 0) continue;
-				let expireString = Chat.toDurationString(expiresIn, {precision: 1});
-				buf += `<p>- ${possessive(entry.userid)} mute expires in ${expireString}.</p>`;
+		if (Punishments.getPunishmentsOfRoom(this.room).length) {
```
bumbadadabum: Put this in a variable instead of generating the list twice.

bumbadadabum:
Oh shit I never saw the tag @TheJetOU my apologies

Looks good outside of the small comments I left. Can you maybe show a screenshot?

TheJetOU:
@bumbadadabum, Did what you told me to and fixed other mistakes I noticed (will send you a screenshot of the output whenever I see you in PS).

server/punishments.js
```diff
@@ -1559,6 +1559,45 @@ Punishments.getRoomPunishments = function (user, options) {
...
+	}
+
+	if (store) {
+		for (const [punishment, data] of store) {
+			let [punishType, id, expireTime, reason] = punishment;
+			let expiresIn = new Date(expireTime).getTime() - Date.now();
+			if (expiresIn < 1000) continue;
+			let alts = [...data[0]].filter(user => user !== id);
+			let ip = [...data[1]];
+			output.push({"punishType": punishType, "id": id, "expiresIn": expiresIn, "reason": reason, "alts": alts, "ip": ip});
```
bumbadadabum: Ips are plural here

TheJetOU:
Should be good now, @bumbadadabum?

TheJetOU:
bumping this @bumbadadabum

bumbadadabum:
oh my god I did not get emailed for this for some reason

server/chat-plugins/info.js
```diff
@@ -2397,55 +2397,39 @@ const pages = {
 		let buf = "";
 		this.extractRoom();
 		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
-		buf += `<div class="pad"><h2>List of active punishments:</h2>`;
+		if (!this.room.chatRoomData) return;
```
bumbadadabum: So the command refuses a room with a hyphen but this line doesn't. Also this should still return the html you removed earlier shouldn't it.

server/punishments.js
```diff
@@ -1559,6 +1559,45 @@ Punishments.getRoomPunishments = function (user, options) {
...
+	}
+
+	if (Punishments.roomIps.get(room.id)) {
+		for (let [key, value] of Punishments.roomIps.get(room.id)) {
+			if (!store.has(value)) store.set(value, [new Set([]), new Set()]);
+			store.get(value)[1].add(key);
+		}
+	}
+
+	if (store) {
```
bumbadadabum: This is always true btw, you literally initialize the variable a few lines before. You can just iterate over it without this check.

bumbadadabum:
Sorry for making you wait so long

server/chat-plugins/info.js
TheJetOU: > the command refuses a room with a hyphen but this line doesn't.

I added that line because for some reason in the page it catches battle rooms while in the commands it still continued and I didn't care enough to investigate the reasoning 
> this should still return the html you removed earlier shouldn't it.

I don't see a reason to how 'List of active punishments:' if it's an invalid room.

bumbadadabum:
Not the list thing but the error message. Better than just showing nothing

TheJetOU:
@bumbadadabum, it shows 'Invalid Room'.

bumbadadabum:
I guess that's good enough.

PR #5310, opened 2019-03-14
Run effectiveness events even for string type

urkerab:
Reported [here](https://www.smogon.com/forums/posts/8066734/).

This is a regression from commit 13614b9 which among other things stopped `onEffectiveness` events from being sent when running type effectiveness for a string rather than an `ActiveMove`, so to handle these I picked an innocuous move and manually set its type.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1598,15 +1598,17 @@ export class Pokemon {
 
 	runEffectiveness(moveOrType: ActiveMove | string) {
 		let totalTypeMod = 0;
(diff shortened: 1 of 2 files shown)


Zarel:
I don't really like the idea of running it on a fake Judgment, but I guess I have no better ideas.

Zarel:
Can you add a test?

urkerab:
I just took the existing type effectiveness test from Stealth Rock, copied it to the Inverse Battle test, and switched the order of the Pokémon so that they are in descending damage order again.

Zarel:
Looks good, thanks!

PR #5311, opened 2019-03-14
Random Doubles Updates

ascriptmaster:
- Fix Guts Dynamic Punch Machamp.
- Mega Heracross with Protect doesn't need Guts in its base form.
- Give Kingler Liquidation for its Sheer Force.

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -1497,7 +1497,7 @@ let BattleFormatsData = {
 	},
 	kingler: {
 		randomBattleMoves: ["liquidation", "xscissor", "rockslide", "swordsdance", "agility", "superpower", "knockoff"],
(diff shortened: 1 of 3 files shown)


TheImmortal:
I'd argue Guts Heracross is useful. Also no need for the gen 6 change because the gen 6 file doesn't have any doubles stuff anyway.

ascriptmaster:
The Guts check line doesn't prevent it from being Guts, it just prevents it from ALWAYS being Guts. Based on its ability ratings this will give it a 2/3 chance of being Moxie and 1/3 Guts, as per usual. I guess adding the exception isn't particularly necessary though? It just seemed silly to me

TheImmortal:
Can we just change it to ``hasMove['protect']  && !isDoubles``

ascriptmaster:
You know what? Yes that's a good idea

PR #5317, opened 2019-03-18
Push 'update' to p3 & p4 in `BattleStream`.

scheibo:
> Yes, your teammate gets to see exact HP in 4 player battles, even when your teammate isn't really your teammate, because Zarel didn't think that mattered enough to use 6 channels. That said, I guess we probably should be pushing the updates like you suggested, though I'm not entirely sure what this function even gets used for.

_Originally posted by @MacChaeger in https://github.com/Zarel/Pokemon-Showdown/pull/5297_

---

I think this fix is required given the current design, but I'd also like to start a discussion around the design decisions regarding channels/streams and 4 player battles. 

- From a correctness POV, revealing information unnecessarily seems very undesirable. What is the difficultly with adding additional channels? I think in 2 player battles we should have 2 channels and in 4 player battles we should have 4?
- From an efficiency POV - I'm not sure we should be creating and updating `p3` and `p4` streams in every battle in this class, given I expect single and doubles to be substantially more popular and we should be optimizing for the common case. Similar to above - in multi battles we can create and maintain all 4 streams, but keep it with 2 streams in the common case?

### DIFF
--- a/sim/battle-stream.ts
+++ b/sim/battle-stream.ts
@@ -172,7 +172,7 @@ export class BattleStream extends Streams.ObjectReadWriteStream<string> {
 }
 
 /**
- * Splits a BattleStream into omniscient, spectator, p1, and p2
+ * Splits a BattleStream into omniscient, spectator, p1, p2, p3 and p4
  * streams, for ease of consumption.
  */
 export function getPlayerStreams(stream: BattleStream) {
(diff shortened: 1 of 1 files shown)


scheibo:
>  I'm not sure we should be creating and updating `p3` and `p4` streams in every battle in this class, given I expect single and doubles to be substantially more popular and we should be optimizing for the common case. Similar to above - in multi battles we can create and maintain all 4 streams, but keep it with 2 streams in the common case?

Hmm, seems like me might have our hands tied here a bit - when we `getPlayerStreams(battleStream)` theres no guarantee `battleStream` has been started to be able to query `battleStream.battle.gameType` to determine how many streams to set up.

Zarel:
Maybe instead of having just one `|split` message, we can have multiple.

`|split|p1`, etc. That'd probably be easier to read/etc than literally six repetitions of the same message.

Zarel:
so my new API doesn't deal with the possibility of privileged information to multiple players in the same line, but I'm 90% sure that case doesn't come up.

scheibo:
`'|split|p1|...'` sounds like an improvement because it would prevent my concern of `p3` and `p4` building up buffers in 2 player battles after this change. How exactly does the API cut down on duplication? Would you just append more players (`'|split|p1,p3|...'` to indicate that they get the same message), or do you mean you still have 6 messages, they're just no longer duplicated because they all have a prefix now?

Zarel:
The idea is that

```
|split|p1
SECRET
PUBLIC
```

would mean the same thing as

```
|split
PUBLIC
SECRET
PUBLIC
SECRET
```

and

```
|split|p2
SECRET
PUBLIC
```

would mean the same thing as

```
|split
PUBLIC
PUBLIC
SECRET
SECRET
```

scheibo:
Forked this conversation into a new bug because commenting on a closed PR is easy to forget about. I like the new protocol and I'd be willing to give the implementation a shot.

PR #5320, opened 2019-03-19
Life Orb/Shell Bell should take effect before Pickpocket

urkerab:
Thanks to @MacChaeger we can now move AfterMoveSecondarySelf to happen before AfterMoveSecondary and still have it only fire once, although I'm not sure I'm using the right target.

The other effects that use AfterMoveSecondarySelf are Fell Stinger, which doesn't care as its target has been KO'd anyway, and Meloetta's forme change, which seems fairly innocuous to me...

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -297,6 +297,7 @@ let BattleAbilities = {
 	"berserk": {
 		desc: "When this Pokemon has more than 1/2 its maximum HP and takes damage from an attack bringing it to 1/2 or less of its maximum HP, its Special Attack is raised by 1 stage. This effect applies after all hits from a multi-hit move; Sheer Force prevents it from activating if the move has a secondary effect.",
 		shortDesc: "This Pokemon's Sp. Atk is raised by 1 when it reaches 1/2 or less of its max HP.",
+		onAfterMoveSecondaryPriority: 4,
 		onAfterMoveSecondary(target, source, move) {
 			if (!source || source === target || !target.hp || !move.totalDamage) return;
 			const lastAttackedBy = target.getLastAttackedBy();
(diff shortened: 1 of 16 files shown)


Marty-D:
Would it make sense to just toss the whole thing in the new `afterMoveSecondaryEvent` before the AfterMoveSecondary part to send the appropriate target? I haven't had a chance to test this yet.

urkerab:
Ah, so what we want is to change Life Orb/Shell Bell to use onSourceAfterMoveSecondary, and put their priority between Red Card and Pickpocket... assuming that will do the right thing, @MacChaeger ?

urkerab:
No, because `onSourceAfterMoveSecondary` still runs once per target, so that's no good...

urkerab:
Ugh, not quite right, this breaks Lock-On...

urkerab:
> No, because onSourceAfterMoveSecondary still runs once per target, so that's no good...

I made it so that `onSource` events only run once when called as a spread event. This affects `onSourceAfterMoveSecondary` (good), `onSourceTryHit` (not used) and `onSourceInvulnerability`. Lock-On uses the latter, so I had to tweak it to use `onAnyInvulnerability` instead (fortunately it already checks both source and target).

This now makes the `AfterMoveSecondarySelf` event redundant, so I switched the users to `onSourceAfterMoveSecondary`, except for Fell Stinger and Relic Song, which simply use `onAfterMoveSecondary`.

At this point all I had to do was set the correct priority on the handlers. Thanks to @SadisticMystic for listing the effects in priority order for me.

Zarel:
Okay, what's the status of this?

AnnikaCodes:
Is this still in progress?

urkerab:
Well, it looks like the status of this is that at the very least it needs rebasing...

mia-pi-git:
Going to close this since it's been inactive for several years. Please feel free to reopen it if it's still relevant.

PR #5323, opened 2019-03-19
Typescript tours

whalemer:


server/tournaments/index.js
```diff
@@ -9,42 +9,51 @@ const TOURBAN_DURATION = 14 * 24 * 60 * 60 * 1000;
...
 		this.id = room.id;
 		this.room = room;
-		this.title = Dex.getFormat(format).name + ' tournament';
+		this.title = format.name + ' tournament';
 		this.isTournament = true;
 		this.allowRenames = false;
 		this.players = Object.create(null);
 		this.playerCount = 0;
-		this.playerCap = parseInt(playerCap) || Config.tourdefaultplayercap || 0;
+		this.playerCap = Number(playerCap) || Config.tourdefaultplayercap || 0;
```
Zarel: ?

server/tournaments/index.js
```diff
@@ -9,42 +9,51 @@ const TOURBAN_DURATION = 14 * 24 * 60 * 60 * 1000;
...
 		this.id = room.id;
 		this.room = room;
-		this.title = Dex.getFormat(format).name + ' tournament';
+		this.title = format.name + ' tournament';
 		this.isTournament = true;
 		this.allowRenames = false;
 		this.players = Object.create(null);
 		this.playerCount = 0;
-		this.playerCap = parseInt(playerCap) || Config.tourdefaultplayercap || 0;
+		this.playerCap = playerCap ? parseInt(playerCap) : Config.tourdefaultplayercap || 0;
```
Zarel: This doesn't change my reaction.

?

What's going on here?

server/tournaments/index.js
```diff
@@ -832,8 +951,12 @@ class Tournament {
 			}
 		}
 
-		this.updateFor(user);
+		this.updateFor(user, null);
```
Zarel: Shouldn't `updateFor` be changed to default the second argument?

server/tournaments/index.js
```diff
@@ -9,42 +9,51 @@ const TOURBAN_DURATION = 14 * 24 * 60 * 60 * 1000;
...
+	/**
+	 * @param {ChatRoom} room
+	 * @param {Format} format
+	 * @param {Object} generator TODO: type this properly
+	 * @param {string | undefined} playerCap
+	 * @param {boolean} isRated
+	 */
 	constructor(room, format, generator, playerCap, isRated) {
-		if (room.tourNumber) {
-			room.tourNumber++;
```
Zarel: You seem to have removed the `tourNumber` tracker without adding it back in elsewhere? Is it useless or something?

server/tournaments/index.js
whalemer: the conditional is because typescript doesn't like ``parse[Int | Float](undefined)``

server/tournaments/index.js
whalemer: according to gentle observation and ``rg``, yes, it used to be appended to `teambuilderFormat`

server/tournaments/index.js
```diff
@@ -115,14 +128,21 @@ class Tournament {
...
 		try {
 			this.teambuilderFormat = Dex.validateFormat(`${this.originalFormat}@@@${rules}`);
 		} catch (e) {
 			output.errorReply(`Custom rule error: ${e.message}`);
 			return false;
 		}
-		this.customRules = Dex.getFormat(this.teambuilderFormat, true).customRules;
+
+		const customRules = Dex.getFormat(this.teambuilderFormat, true).customRules;
+		if (!customRules || !Array.isArray(customRules)) throw new Error(`Setting invalid custom rules in tour: ${customRules}`);
```
Zarel: `!customRules` should probably be an `errorReply`, if it's even possible.

server/tournaments/index.js
```diff
@@ -320,25 +362,36 @@ class Tournament {
...
 	}
-	removeUser(user, output) {
-		if (!(user.userid in this.players)) {
-			output.sendReply('|tournament|error|UserNotAdded');
+
+	/**
+	 * @param {string} userid
+	 * @param {CommandContext} [output]
+	 */
+	removeUser(userid, output) {
```
Zarel: Why change this function to take a `userid` instead of a `User`?

whalemer: Banning an offline user with no object that's in the tour will try to remove them without a user object.

server/tournaments/index.js
```diff
@@ -154,6 +177,7 @@ class Tournament {
 	forceEnd() {
 		if (this.isTournamentStarted) {
 			if (this.autoDisqualifyTimer) clearTimeout(this.autoDisqualifyTimer);
+			if (!this.inProgressMatches) throw new Error(`Unexpected null in tours`);
```
Zarel: I don't know if I like all the new `Unexpected null` throws. Especially if they do the same thing as `// @ts-ignore`

Zarel: I've thought about this, and I think the best way is to declare these as non-null (and cover them up with `ts-ignore` or an `any`-cast.

Zarel:
This pull request **introduces 1 alert** when merging 7c6b52714a565e77fcad95bc8330bcadbee2e8af into 101e973ab9cbd8d12e60ba306e08d0f38b52e86f - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-bd535bb96761f5dc6501195de7163120817b4505)

**new alerts:**

* 1 for Useless assignment to property

---

server/tournaments/index.js
Zarel:
```suggestion
		this.playerCap = (playerCap ? parseInt(playerCap) : Config.tourdefaultplayercap) || 0;
```

Both need to fall back, though.

server/tournaments/index.js
```diff
@@ -56,19 +65,25 @@ class Tournament {
...
+		this.isTournamentStarted = false;
+
+		this.availableMatches = new Map();
+		this.inProgressMatches = new Map();
+		this.pendingChallenges = new Map();
+		this.disqualifiedUsers = new Map();
+		this.autoDisqualifyWarnings = new Map();
+		this.lastActionTimes = new Map();
+
+		this.isBracketInvalidated = true;
```
Zarel:
```suggestion
```

This is duplicate code of two lines down

server/tournaments/index.js
```diff
@@ -56,19 +65,25 @@ class Tournament {
...
+
```
Zarel:
```suggestion
```

server/tournaments/index.js
whalemer: my bad, forgot how precedence worked

PR #5324, opened 2019-03-19
Don't show boost failure when stockpile wears off after it didn't boost

urkerab:
Just a hunch, but if you're at +6 before you stockpile and then the stockpiled effect wears off it probably shouldn't say that the stat couldn't go any higher.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -16458,7 +16458,13 @@ let BattleMovedex = {
 				if (curSpD !== target.boosts.spd) this.effectData.spd--;
 			},
 			onEnd(target) {
-				this.boost({def: this.effectData.def, spd: this.effectData.spd}, target, target);
+				if (this.effectData.def || this.effectData.spd) {
+					/** @type {SparseBoostsTable} */
+					let boosts = {};
+					if (this.effectData.def) boosts.def = this.effectData.def;
+					if (this.effectData.spd) boosts.spd = this.effectData.spd;
+					this.boost(boosts, target, target);
+				}
 				this.add('-end', target, 'Stockpile');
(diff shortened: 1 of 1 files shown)


MacChaeger:
Is there a boost failure message if the attempted boost is all zeroes to begin with? That seems like a bug in and of itself.

Zarel:
Can we get someone to actually test this? @DaWoblefet?

DaWoblefet:
This is sort of ambiguous wording to me, so I'm not sure what's being changed, but I'm sure one of these is what the PR is after:
- Using Stockpile when at +6 Def/+6 Sp. Def and user has 0-2 Stockpiles: animation plays, Stockpile increases counter, doesn't raise Def/Sp. Def.
- Using Stockpile when at +6 Def/+6 Sp. Def and user has 3 Stockpiles: no animation plays, "But it failed!" message.

urkerab:
@MacChaeger I'm happy to rewrite this PR into that if that's what people would prefer. @DaWoblefet This is for the scenario when a) you are already at +6 b) you stockpile c) you spit up or swallow.

DaWoblefet:
In that particular case, Spit Up/Swallow don't fail - Stockpile increases the Stockpile counter at b), but doesn't boost Def/Sp. Def like mentioned earlier. Using Stockpile/Spit Up will then execute their effects as a 1-Stockpile attack, remove Stockpile, and not touch the Def/Sp. Def (because Stockpile didn't increase Def/Sp. Def).

Zarel:
@CTCC1 I'm guessing this is a coincidence, because this is an RNG update. The reason Rough Skin happens twice is because Ice Fang freezes. This is definitely a bug caused by @MacChaeger's spreadMoveHit refactor.

PR #5325, opened 2019-03-19
Cleanup sim/battle.ts

scheibo:
Final part of my 'post Typescript-ification cleanup' (yay!). This is a big class, I got tired and probably missed some more cleanup opportunities, but I was only willing to make so many passes...

Trying to make things easier on you as a reviewer, so instead of squashing all the commits together I tried to keep each operation I did separate so that you could review each consecutive commit and see exactly what is happening at each step. Let me know if approach helps at all.

I know we discussed merging `scripts.js` into `battle.ts` instead of all the throwing methods, but I ran into trouble when I tried that due to the mismatched types between `globals` and `battle` (fixed in this CL), and also I wanted to take a step back and think more about what the end goal should look like (no point in merging `scripts.js` and causing churn and making `git blame` harder if we're just going to create a `mechanics.ts` file or something soon anyway).


sim/battle.ts
```diff
@@ -279,19 +279,16 @@ export class Battle extends Dex.ModdedDex {
...
 		this.weather = '';
 		this.weatherData = {id: ''};
 		return true;
 	}
 
 	effectiveWeather() {
-		if (this.suppressingWeather()) return '';
-		return this.weather;
+		return this.suppressingWeather() ? '' : this.weather;
 	}
```
Slayer95: It may be just me, but I think ``return default`` is better than a hard to read ternary?

sim/battle.ts
```diff
@@ -3467,46 +3340,104 @@ export class Battle extends Dex.ModdedDex {
...
 		}
-		// @ts-ignore - prevent type | null
-		this.sides[0] = null;
-		// @ts-ignore - prevent type | null
-		this.sides[1] = null;
 		for (const action of this.queue) {
 			delete action.pokemon;
 		}
+		// @ts-ignore - readonly
 		this.queue = [];
```
Slayer95: ``queue`` doesn't seem to be marked as ``readonly`` (anymore)?

sim/battle.ts
```diff
@@ -175,18 +193,10 @@ export class Battle extends Dex.ModdedDex {
...
-		if (options.p3) {
-			this.setPlayer('p3', options.p3);
-		}
-		if (options.p4) {
-			this.setPlayer('p4', options.p4);
-		}
+		if (options.p1) this.setPlayer('p1', options.p1);
+		if (options.p2) this.setPlayer('p2', options.p2);
+		if (options.p3) this.setPlayer('p3', options.p3);
+		if (options.p4) this.setPlayer('p4', options.p4);
```
Slayer95: Maybe this can be switched to a loop over a ``SIDE_IDS: SideID[]`` constant?

sim/battle.ts
scheibo: Entirely fair, but unfortunately completely subjective. We don't really have any rules around when we return a ternary and when we do the double `return`, and my JS background is in writing for Web circa 2010 where code size is king, which is probably why I prefer the former (also because I write C/C++/Java for my day job).

In a similar vein - when do we omit the braces from an `if` (or even `for` loops sometimes :S)? The answer is once again murky.

I've tried to use the rule that if the ternary would be long or complex then I would use the double `return` - in this specific case the ternary is pretty short and I don't think its necessarily hard to read if you're familiar with the ternary operator? But I could be very mistaken.

Regarding `if` - if the statement is short and an early `return`/`continue`/`break` then no braces seems like a no brainer. If its a variable assignment (`if (!foo) foo = ...`) or a method call (`if (foo) bar()`) then I'm less sure its always the right thing to do (and I'm really not a big fan for braceless loops except in rare cases), but theres no standard in the codebase. Do you and @Zarel have guidelines/rules of thumb you follow? I don't really have a huge preference here, I'm used to automatic formatters which insert all the braces everywhere regardless.

data/items.js
```diff
@@ -152,7 +152,6 @@ let BattleItems = {
 			if (effect.effectType === 'Move' && effect.id !== 'confused') {
 				this.add('-enditem', target, 'Air Balloon');
 				target.item = '';
-				this.itemData = {id: '', target: this};
```
Slayer95: This should've been ``target.itemData = {id: '', target: this};``

data/items.js
```diff
@@ -161,7 +160,6 @@ let BattleItems = {
...
```
Slayer95: Ditto

sim/battle.ts
Slayer95: As far as I can tell, the codebase follows the pattern that if one of the code paths is falsey or an error, it should early return, thus making the semantic split more explicit.

sim/battle.ts
```diff
@@ -2358,11 +2317,8 @@ export class Battle extends Dex.ModdedDex {
 	}
 
 	getTargetLoc(target: Pokemon, source: Pokemon) {
-		if (target.side === source.side) {
-			return -(target.position + 1);
-		} else {
-			return target.position + 1;
-		}
+		const position = target.position + 1;
+		return (target.side === source.side) ? -position : position;
```
Slayer95: This case looks great for a ternary IMO :)

sim/battle.ts
scheibo: For sure, but that's not quite what I'm talking about here. When does the early return/throw use braces and when doesn't it? And there are oodles of places where we're not returning/throwing and still don't use braces.

FWIW - I reverted back to the `return default` approach here.

Slayer95: Unfortunately, that's... quite arbitrary... But you may notice that with/without braces is most often defined in a per-file basis, according to who the author was...

Zarel: I'm fine with either approach here. This is one of those style issues I've never bothered nitpicking since I don't have a strong opinion.

sim/battle.ts
```diff
@@ -175,17 +193,10 @@ export class Battle extends Dex.ModdedDex {
...
-		}
-		if (options.p2) {
-			this.setPlayer('p2', options.p2);
-		}
-		if (options.p3) {
-			this.setPlayer('p3', options.p3);
-		}
-		if (options.p4) {
-			this.setPlayer('p4', options.p4);
+		for (const side of SIDE_IDS) {
```
Zarel: I don't really like the `SIDE_IDS`, especially not imported from `sim/side`, and especially not unnamespaced.

Zarel: Considering it's only used here, it might be better as a literal.

sim/battle.ts
```diff
@@ -573,8 +564,7 @@ export class Battle extends Dex.ModdedDex {
 			hasRelayVar = false;
 		}
 
-		// @ts-ignore
-		if (effect.effectType === 'Status' && target.status !== effect.id) {
+		if (effect.effectType === 'Status' && (target instanceof Pokemon) && target.status !== effect.id) {
```
```suggestion
		if (effect.effectType === 'Status' && (target as Pokemon).status !== effect.id) {
```

We try to use duck typing rather than `instanceof`.

sim/battle.ts
```diff
@@ -2834,7 +2779,7 @@ export class Battle extends Dex.ModdedDex {
 					break;
 				}
 				// in gen 5+, the switch is cancelled
-				this.debug('A Pokemon can\'t switch between when it runs out of HP and when it faints');
+				this.debug("A Pokemon can't switch between when it runs out of HP and when it faints");
```
Zarel: I wonder if this should be a hint.

sim/battle.ts
```diff
@@ -3306,7 +3235,7 @@ export class Battle extends Dex.ModdedDex {
 		} else if (left && !right && right !== 0) {
 			return left;
 		} else if (typeof left === 'number' && typeof right === 'number') {
-			// @ts-ignore
+			// @ts-ignore - Typescript doesn't realize T and U both are numbers
 			return left + right;
```
Zarel: I wonder if we could do something like `return (left + right) as T;`

sim/battle.ts
```diff
@@ -3317,148 +3246,94 @@ export class Battle extends Dex.ModdedDex {
...
-	 * target = undefined: automatically resolve target
-	 * target = null: no target (move will fail)
-	 */
-	useMoveInner(
-		move: string | Move, pokemon: Pokemon, target?: Pokemon | null,
-		sourceEffect?: Effect | null, zMove?: string
-	): boolean {
-		throw new Error(`The useMoveInner function needs to be implemented in scripts.js or the battle format.`);
+	calcRecoilDamage(damageDealt: number, move: Move): number {
+		throw newUnimplementedError('calcRecoilDamage');
```
Zarel: Why `newUnimplementedError` instead of `new UnimplementedError`?

sim/side.ts
```diff
@@ -69,7 +71,7 @@ export class Side {
 		if (sideScripts) Object.assign(this, sideScripts);
 
 		this.battle = battle;
-		this.id = ['p1', 'p2', 'p3', 'p4'][sideNum] as SideID;
+		this.id = SIDE_IDS[sideNum];
```
Zarel: I guess it's used in two places, but I'm still not sure how I like this.

sim/battle.ts
scheibo: is `Side.IDS` any better?

sim/battle.ts
scheibo: `singleEvent` does this (I've already tried to shard the code between the suppression checks in `runEvent` and `singleEvent` but ran into problems)

sim/battle.ts
scheibo: Because from my research properly implementing an Error class seemed kind of obnoxious in JS, but I guess I'm wrong.

sim/battle.ts
Zarel: It is, but I'm still partial to the array literal.

sim/battle.ts
Zarel: I'm tentatively fine with either resolution.

sim/battle.ts
```diff
@@ -3442,6 +3441,9 @@ export class Battle extends Dex.ModdedDex {
 	}
 }
 
-function newUnimplementedError(name: string) {
-	return new Error(`The ${name} function needs to be implemented in scripts.js or the battle format.`);
+class UnimplementedError extends Error {
+	constructor(name: string) {
```
Zarel: I vaguely remember there being some trick you have to do to make Error subclasses work correctly.

sim/battle.ts
scheibo: Going to stick with the `instanceof` like with `singleEvent`.

sim/battle.ts
scheibo: *d' accord. je m'en fiche*

sim/battle.ts
scheibo: > We can inherit from Error and other built-in error classes normally, just need to take care of name property and don’t forget to call super.

https://javascript.info/custom-errors

[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types) talks about `Error.captureStackTrace` on Node, but then also links to [Stack Overflow](https://stackoverflow.com/questions/1382107/whats-a-good-way-to-extend-error-in-javascript), which seems to agree that in ES6 the way the first site I quoted above explained things is correct. (but given how its not immediately obvious what the 'correct' way is, its perhaps not surprising I originally punted by just making this a `function`...).

Anyway, went ahead and threw in the `Error.captureStackTrace` check.

sim/battle.ts
Slayer95: See e.g. https://github.com/Zarel/Pokemon-Showdown/blob/master/server/loginserver.js#L26-L27

(A few years ago it was really complicated, yes. It even required usage of non-standard JS provided by the V8 engine).

sim/battle.ts
Slayer95: See my comment above: ``Error.captureStackTrace()`` is no longer needed.

sim/battle.ts
scheibo: Cool. The pattern of setting `this.name` in the constructor also seems to work just the same.

PR #5326, opened 2019-03-19
Extract and enhance RandomPlayerAI from `battle-stream-example.ts`

scheibo:
From #5278 - I'm still busy [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) how exactly I want to do profiling, but I'd like to land this part of it which is kind of separate anyway.

- I made an `examples` sub directory to hopefully indicate that `random-player-ai.ts` isn't a 'blessed' AI by any means, but also make it so that other tools (like my forthcoming harness) can still require it. I'm not sure I like the stutter in `examples/battle-stream-example.ts`, but calling it `examples/battle-stream.ts` seemed a little too close to `battle-stream.ts` one level up? Maybe `examples/play-random-game.ts` would be a better name? Another alternative is to leave everything in `sim`, but call this `example-random-ai.ts`/`random-ai-example.ts` or something?
- The main difference with the AI is that it can now actually execute zMoves/Megas/switching instead of just staying in and attacking until it faints. I also had to add fixes for double battles and targetting that the previous AI was lacking. I tried to get `git` to see this as a branch from `battle-stream-example.ts`, but no matter what I tried it seems like the changes I made are substantial enough that it won't let me track the history :(


sim/examples/random-player-ai.ts
```diff
@@ -0,0 +1,157 @@
...
+	) {
+		super(playerStream, debug);
+		this.move = options.move || 1.0;
+		this.mega = options.mega || 0;
+		this.prng = options.seed && !Array.isArray(options.seed) ? options.seed : new PRNG(options.seed);
+	}
+
+	receiveError(error: Error) {
+		if (!error.message.startsWith('[Invalid choice]')) throw error;
+		// Somehow we had a choice error - just let the simulator autochoose for us.
```
Slayer95: I think this piece of documentation would be better if it specified that this case might happen if a Pokémon is afflicted by Imprison, Arena Trap, Magnet Pull or Shadow Tag.

scheibo: I think it also maybe happens with Struggle? It also conveniently papers over bugs (as we discovered in #5286...). But yeah, explaining why its sometimes needed is a good call.

Slayer95: Not really. If the Pokémon must Struggle, the request object lists only ``Struggle`` in the possible moves.

scheibo: Ack.

PR #5329, opened 2019-03-20
Random Battle: Update Heliolisk, Blacephalon, Seviper

ascriptmaster:
- Rain Dance does not work well with Heliolisk; it prefers wallbreaker
roles instead.
- Mind Blown is not good if the Z-Crystal is already taken.
- Seviper learns Swords Dance in Gen 7.

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -4143,7 +4143,7 @@ let BattleFormatsData = {
 		doublesTier: "(DUU)",
 	},
 	seviper: {
-		randomBattleMoves: ["flamethrower", "sludgewave", "gigadrain", "darkpulse", "switcheroo", "coil", "earthquake", "poisonjab", "suckerpunch"],
+		randomBattleMoves: ["flamethrower", "sludgewave", "gigadrain", "darkpulse", "switcheroo", "swordsdance", "earthquake", "poisonjab", "suckerpunch"],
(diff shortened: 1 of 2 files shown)


ascriptmaster:
That's an interesting Scarfed set. I guess I can just not reject Mind Blown with no Z-Move, but I still want to keep the check that allows CM Blacephalon to not reject Fire Blast if there is no Z-Move.

Actually, maybe Mind Blown can be rejected if it has setup and can't use a Z-Move, and Fire Blast be rejected otherwise? The Sitrus Berry set is a cute idea but I don't think it's a particularly good idea in Randoms.

Speaking of Rain Dance Heliolisk getting Life Orb, maybe we should reconsider all the other manual weather setters and whether they should keep their weather stones. Sunny Day Pyroar was another complaint in the thread, but after chatting with the Random Battles room on Showdown I found people tended to agree that it was ok as is. It might be better with Life Orb, though. That said, the other Rain Dancers are mostly Swift Swimmers (or Rain Dish Ludicolo) so maybe they appreciate the Damp Rock anyways.

ascriptmaster:
Reverted Heliolisk for now, and updated the Mind Blown condition.

TheImmortal:
That's fair. I have the Helio change in my next update. The thing about Pyroar is that it uses SolarBeam. Helio also has Rain Dish to cancel out Life Orb.

TheImmortal:
Randoms people say Heat Rock is good on stuff. But Life Orb is better on most setters. I'll do that.

Issue #5243, opened 2019-03-04
Add a BasicEffect.isAvailable(format) method


scheibo:
Follow up after the 8289374 bug fix.

```
>> Dex.mod('gen1').getTemplate('hydreigon').isNonstandard
<< false
>> Dex.mod('gen1').getTemplate('hydreigon').tier
<< "OU"
```
Is unintuitive. @HoeenCoder suggests it could be useful for OM's, but @TheImmortal claims it isn't. I think its worth fixing.

Somewhat related: shouldn't CAP Pokemon have a generation associated with them? 
I think for CAP (depending on how you treat the prevos) it would be:
```
- Gen IV: [-1, -2, -3, -4, -5, -6, -7,  -8, -9, -10, -11] + [-101, -102, -103, -104, -116, -117, -118, -119]
- Gen V: [-12, -13, -14, -15, -16, -17] + [-106, -107, -108, -109, -110]
- Gen VI: [-17, -18, -19, -20, -21, -22] + [-111, -112, -113, -114, -115]
- Gen VI: [-23, -24, -25, -26, -27] + [-120]
```

Neither of these changes would have prevented the bug 8289374 fixed, but the way `Dex` works was revealed to be less obvious/intuitive due to the bug.

Zarel:
I'd say to set `isNonstandard` to `true` (it was probably originally set to `true` but lost at some point), but otherwise keep it around.

Zarel:
`tier` as well can be set to `Illegal`.

Slayer95:
IIRC, Crystal had a server with a Gen 1 OM with Gen 2 Pokémon, so evidence shows it's useful.

I'd suggest changing ``isNonstandard`` to a string: ``'CAP' | 'Glitch' | 'Hack' | 'Future'``. That would also simplify work on rulesets. See e.g. this unintuitive line https://github.com/Zarel/Pokemon-Showdown/blob/cfb1cf1731eb5273ecc3796ff2a5b7bd92896a4a/data/rulesets.js#L162

Slayer95:
Rather than forcing every generation to get its own deep clone of each data entry, it would be better to have a ``BasicEffect#isAvailable()`` method that reads the ``isNonstandard`` flag as well as the ``gen`` in order to find out the relevant data.

Sort of like how old gens use ``getCategory(move)`` rather than checking ``move.category``

Zarel:
I disagree with `.isAvailable`. We already make a deep clone for every gen.

PR #5330, opened 2019-03-20
Make isNonstandard specify reason for non-standardness

scheibo:
> I'd suggest changing ``isNonstandard`` to a string: ``'CAP' | 'Glitch' | 'Hack' | 'Future'``. That would also simplify work on rulesets.

_Originally posted by @Slayer95 in https://github.com/Zarel/Pokemon-Showdown/issues/5243#issuecomment-471334629_

This is purely meant to be a data-related change, and does not encompass all possible work on #5243 (eg. potentially cleaning up existing usages of `isNonstandard`, setting `isNonstandard: 'Future'` and `tier: 'Illegal'` for future gen mons, etc)


Slayer95:
Semantically, ``Hack`` should be something usable in Hackmons.

Pet mods may use ``Custom``, or preferably the mod name as an identifier. That would allow checking for ``isNonstandard === modName`` on data initialization scripts.

scheibo:
> Pet mods, such as SSB, may use Custom, or preferably the mod name as an identifier. That would allow checking for .isNonstandard === modName on data initialization scripts.

I'm fine with 'SSB' instad of 'Custom', but this requires we are able to enumerate all the types in the `type` declaraton.

> Would it be possible to tell TS to accept all strings according to a regex? Say: `/Mod: [a-z0-9]+`/?

[No](https://github.com/Microsoft/TypeScript/issues/6579). The best we could do would be to use `'| string'` which is less desirable from a type safety perspective.

Zarel:
I like this, but what does `Hack` mean?

Zarel:
I don't see `Hack` used anywhere, either.

Zarel:
Oh, I see, it was renamed `Custom`.

Slayer95:
> I like this, but what does Hack mean?

> Semantically, Hack should be something usable in Hackmons.

FWIW, I originally intended that keyword for Pokestar mons.

Zarel:
Strictly speaking, Pokestar mons should be `Past` but this is probably better in practice. There's some clarity gain from `Past Glitch` and `Past Pokestar` but I think in practice that's not worth the uglier comparisons you'd need.

Zarel:
Actually... it might.

sim/dex-data.ts
```diff
@@ -103,11 +103,11 @@ export class BasicEffect implements EffectData {
...
 	desc: string;
 	/**
-	 * Is this item/move/ability/pokemon nonstandard? True for effects
+	 * Is this item/move/ability/pokemon nonstandard? Specified for effects
 	 * that have no use in standard formats: made-up pokemon (CAP),
-	 * glitches (Missingno etc), and Pokestar pokemon.
+	 * glitches (Missingno etc), Pokestar pokemon, etc.
 	 */
-	isNonstandard: boolean;
+	isNonstandard: Nonstandard | false;
```
Zarel:
```suggestion
	isNonstandard: Nonstandard | null;
```

(and elsewhere)

I've been trying to move away from the `| false` pattern.

dev-tools/globals.ts
```diff
@@ -35,6 +35,7 @@ type SparseStatsTable = Partial<StatsTable>;
 type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
 type BoostsTable = {[boost in BoostName]: number };
 type SparseBoostsTable = Partial<BoostsTable>;
+type Nonstandard = 'Glitch' | 'Past' | 'Future' | 'CAP' | 'LGPE' | 'Pokestar' | 'Custom';
```
Zarel:
```suggestion
type Nonstandard = 'Glitch' | 'Past' | 'Future' | 'CAP' | 'LGPE' | 'Pokestar' | 'Custom' | null;
```

Or maybe the `| null` should be here? Up to you.

sim/dex-data.ts
scheibo: Becaus Typescript has had issues with it (thinking its `boolean`)?[ I don't really like `null` anywhere TBH](https://www.youtube.com/watch?v=PSGEjv3Tqo0&feature=youtu.be&t=9m21s), *maybe* its OK as a missing `Object` reference, and `false` seems to be more explicit and read better (`isNonstandard: false` is more clear then `null`).

I'm fine to switch this, I'm just curious about your reasoning?

Zarel: Because `false` here is a sentinel value, rather than an actual boolean (it's impossible for this to be `true`).

And nulls are the usual sentinel value for "there's nothing here". Between `null` and `undefined`, I went with `null` mostly because it was easier to type, especially since `T?` in JSDoc meant `T | null`, making it much easier to type than `T | undefined`.

`null` also allows `null` to be a non-default value – i.e. for `foo(null)` to be different from `foo()`. I guess this might be an anti-pattern though; I don't think we actually use this anywhere.

I don't think the difference between `null` and `undefined` is big enough to be worth redoing the entire codebase towards `undefined`.

scheibo: > Because `false` here is a sentinel value, rather than an actual boolean (it's impossible for this to be `true`). And nulls are the usual sentinel value for "there's nothing here"

Fair. This argument seems the most valid. My complaint is that https://github.com/Zarel/Pokemon-Showdown/pull/5330/commits/909eb214285b95ee9b2eaa760047bd95bffd9a1b#diff-a9f5618c5913fe6f011f25e44b509967 now doesn't read as well because there's an asymmetry between `isUnreleased` and `isNonstandard`, but I can look past that.

> I went with null mostly because it was easier to type, especially since `T?` in JSDoc meant `T | null`, making it much easier to type than `T | undefined`.

Yeah, unfortunately that is now the opposite in native Typescript. Instead of `T?` we now write `T | null = null` which is very unfortunate. 

> `null` also allows `null` to be a non-default value – i.e. for `foo(null)` to be different from `foo()`. I guess this might be an anti-pattern though; I don't think we actually use this anywhere.

I agree, this is definitely an anti pattern

> I don't think the difference between `null` and `undefined` is big enough to be worth redoing the entire codebase towards `undefined`.

I don't know, I'd consider it to clean up the `.ts` method signatures

Zarel: `null` is still significantly fewer letters than `undefined` overall, and there are many places you can't just use `?`, such as variable declarations. For the most part, I don't have a problem with `?` in method signatures; there are relatively few where the difference matters (and for those, we already have a defined difference between `null` and `undefined`).

Issue #5331, opened 2019-03-20
Instead of having just one `|split` message, we should have multiple.


scheibo:
> Maybe instead of having just one `|split` message, we can have multiple.
>
> `|split|p1`, etc. That'd probably be easier to read/etc than literally six repetitions of the same message.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5317#issuecomment-474154108_

> The idea is that
> 
> ```
> |split|p1
> SECRET
> PUBLIC
> ```
>
> would mean the same thing as
>
> ```
> |split
> PUBLIC
> SECRET
> PUBLIC
> SECRET
> ```
>
> and
>
> ```
> |split|p2
> SECRET
> PUBLIC
> ```
>
> would mean the same thing as
>
> ```
> |split
> PUBLIC
> PUBLIC
> SECRET
> SECRET
> ```

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5317#issuecomment-474678889_

I think this protocol change is important in a world with @MacChaeger's 4 player battles to allow for not sharing information between teammates (or an opponent in a free-for-all!) Because this changes protocol, I believe the change needs to first be made on the client (to handle the new protocol as well as the old one), after which point the server can be updated

Zarel:
This change doesn't affect client, which only receives sanitized messages. It only affects the sim protocol – i.e. the protocol the sim process and the main process use to communicate with each other.

scheibo:
Ah! Quite right, I forgot that the server handles this. I was thinking maybe replays have `splits` in them, but thats clearly wrong as theyre from the spectators POV. It seems its mainly `battle.ts`/`battle-stream.ts`/`server/sockets.js` and maybe a few other places that need updating.

PR #5332, opened 2019-03-20
Change protocol for '|split' message type (#5331)

scheibo:
Okay, so this passes our test suite and `sim/battle-stream-example.ts` runs, but I have **absolutely no confidence** in it and ran into problems so wanted to sent it out to get some preliminary feedback.

- [x] not confident with the regexes, especially regarding when they should be adding or stripping newlines
- [x] the `Battle#add` API. *erghmagerd*, nice and ugly. I think moving the `reportExactHP` check into `Pokemon#getHealth` is the right direction (it already checks `this.battle.reportPercentages`, so it makes sense to have everything in one place?), but `Battle#add` + `getDetails`/`getHealth` is ugly AF. Currently as written, **(a)** its only a partial function because it can't take multiple functions if the `side` doesn't agree and **(b)** its taking a `function` in the first place, as opposed to the `{side, secret, shared}` object, because it conflicts with `AnyObject` in the signature. One solution might be to turn it into a `[SideId, string, string]` array to be able to detect it in `parts` with an `Array.isArray` check instead of `typeof`, but its still awkward.

I'm looking for feedback on this approach, in particular what the most elegant way to handle `add` + `getHealth`/`getDetails` might be, and how best to test this and be confident I'm not breaking everything (roomlog? sockets? etc).


sim/battle-stream.ts
```diff
@@ -209,33 +209,32 @@ export function getPlayerStreams(stream: BattleStream) {
...
-				const specUpdate = data.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-				streams.spectator.push(specUpdate);
-				const omniUpdate = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n[^\n]*/g, '');
-				streams.omniscient.push(omniUpdate);
+			case 'update': {
+				const split = SPLIT_REGEX.exec(data);
+				for (const name in streams) {
+					// @ts-ignore - index signature
+					const s = streams[name];
+					s.push(split ? (name === split[0] || name === 'omniscient' ? split[1] : split[2]) : data);
```
MacChaeger: Can this be broken up? It’s pretty hard to read a nested ternary.

sim/battle-stream.ts
```diff
@@ -209,33 +209,32 @@ export function getPlayerStreams(stream: BattleStream) {
...
-				const p2Update = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n([^\n]*)\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-				streams.p2.push(p2Update);
-				// p3 and p4 share update information with p1 and p2 respectively.
-				streams.p3.push(p1Update);
-				streams.p4.push(p2Update);
-				const specUpdate = data.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-				streams.spectator.push(specUpdate);
-				const omniUpdate = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n[^\n]*/g, '');
-				streams.omniscient.push(omniUpdate);
+			case 'update': {
```
MacChaeger: What’s with the braces inside a switch statement? Isn’t that inconsistent with the style of the repository?

server/sockets.js
```diff
@@ -567,31 +569,15 @@ if (cluster.isMaster) {
...
-					socket.write(messages[2]);
-					break;
-				default:
-					if (!messages[0]) {
-						messages[0] = message.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-					}
-					socket.write(messages[0]);
-					break;
-				}
+				socket.write((roomChannel && roomChannel.get(socketid) === channelid) ? secret : shared);
```
MacChaeger: Isn’t `channelid` a `SideID` and not a number like what the switch statement was looking for?

MacChaeger:
Yeah, gotta edit `RoomBattlePlayer#updateChannel` and take out the `% 2`

Zarel:
I would definitely have retained the regexes.

You actually only need two regexes: A regex for "this is privileged", and a regex for "this is not privileged".

Your first step is to grab the privileged messages you do have access to:

```js
.replace(/\n\|split|p1\n([^\n]*)\n(?:[^\n]*)/g, '\n$1')
```

(Replace `p1` with `p2` etc for other player channels, or `p1` with `p[1234]` for the omniscient channel. Don't do any replacements at this phase for the spectator channel.)

Your next step is to discard all the privileged messages you don't have access to:

```js
.replace(/\n\|split|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '\n')
```

This should probably all be consolidated in a single function.

```ts
/** -1:omniscient, 0:spectator */
function extractChannel(message: string, channelid: -1 | 0 | 1 | 2 | 3 | 4 = 0) {
  if (channelid === -1) {
    // Grab all privileged messages
    return message.replace(/\n\|split|p[1234]\n([^\n]*)\n(?:[^\n]*)/g, '\n$1');
  }

  // Grab privileged messages channel has access to
  switch (channelid) {
  case 1: message = message.replace(/\n\|split|p1\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
  case 2: message = message.replace(/\n\|split|p2\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
  case 3: message = message.replace(/\n\|split|p3\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
  case 4: message = message.replace(/\n\|split|p4\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
  }

  // Discard remaining privileged messages
  // Note: the last \n? is for privileged messages that are empty when non-privileged
  return message.replace(/\n\|split|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '\n');
}
```

server/roomlogs.js
```diff
@@ -90,10 +90,11 @@ class Roomlog {
 		log = [];
 		for (let i = 0; i < this.log.length; ++i) {
 			const line = this.log[i];
-			if (line === '|split') {
-				const ownLine = this.log[i + channel + 1];
+			const split = /\|split\|p(\d)/g.exec(line);
+			if (split) {
+				const ownLine = this.log[i + (Number(split[0]) === channel ? 1 : 2)];
```
Zarel: What does `channel` mean here? The previous mapping was `0:spectator, 1:p1, 2:p2, 3:omniscient`, I think, but this refactor should probably change that mapping.

sim/README.md
```diff
@@ -122,7 +122,15 @@ An update which should be sent to all players and spectators.
...
-One message type that only appears here is `|split`. This splits the next six lines into `spectator`, `p1`, `p2`, `p3`, `p4`, and `omniscient` messages. The `p1`, `p2`, `p3`, and `p4` logs will have exact HP values only for the corresponding player, while the `spectator` log will not have exact HP values for either player, and the `omniscient` logs will have exact HP values for both.
+One message type that only appears here is `|split|PLAYERID`:
+
+    |split|PLAYERID
+    SECRET
+    PUBLIC
+
+- `PLAYERID` - one of `p1`, `p2`, `p3`, or `p4`.
+- `SECRET` - messages for the specific player or an omniscient observer (details which may contain information about exact details of the player's set, like exact HP)
+- 'PUBLIC` - message with public details suitable for display to opponents / teammates / spectators.
```
Zarel:
```suggestion
- 'PUBLIC` - message with public details suitable for display to opponents / teammates / spectators. Note that this may be empty.
```

sim/battle.ts
```diff
@@ -3105,41 +3107,47 @@ export class Battle extends Dex.ModdedDex {
...
-				typeof part !== 'function' ? part : part(side)
-			).join('|');
-			this.log.push(sideUpdate);
+		let side: SideID | null = null;
+		const secret = [];
+		const shared = [];
+		for (const part of parts) {
+			if (typeof part === 'function') {
+				const split = part();
+				// FIXME: Can only handle a single side
```
Zarel: I don't think this needs to be fixed, I think this is an intentional design.

sim/battle-stream.ts
scheibo: Its required in certain in ES6 to create a block scope to get around the issue declaring variables with `let` or `const` which could conflict with other `case` statements. We do it elsewhere in the repo as well.

In this case though, I think the collision no longer exists, so I can remove them.

server/sockets.js
scheibo: Going of the switch, the `channelid` is the string representation of a number, not a SideID (ie/ no 'p' prefix).

server/roomlogs.js
scheibo: true, I'm neglecting `omniscient`. I'm assuming its now 0 - 5.

sim/battle.ts
scheibo: Unfortunately it doesn't work: data/moves.js:11599 (Pain Split) does two `getHealth` calls, which means we need multiple `split`s I think?

scheibo:
> I would definitely have retained the regexes.

The regexes have a lot of duplication - the only different in `case 1` and `case 2` in your `switch` is `p1` vs `p2`, where we should be able to do `p\d`?

Also, 'grab all' then 'discard all' is inefficient - why not just grab the particular part of the `split` that is relevant to us?

Zarel:
I assumed writing it that way lets V8 precompile them – with how often these are used, some optimization would be nice.

I don't understand what you mean by "grab the particular part of the `split` that is relevant to us" – isn't that what my sample code does?

server/room-battle.js
```diff
@@ -71,11 +71,11 @@ class RoomBattlePlayer {
 	updateChannel(/** @type {User | Connection} */ user) {
 		if (user instanceof Users.Connection) {
 			// "user" is actually a connection
-			Sockets.channelMove(user.worker, this.game.id, this.slotNum % 2 + 1, user.socketid);
+			Sockets.channelMove(user.worker, this.game.id, this.slotNum + 1, user.socketid);
```
Zarel: In theory it should still be `% 2` in multi battles.

server/roomlogs.js
```diff
@@ -12,6 +12,9 @@
 /** @type {typeof import('../lib/fs').FS} */
 const FS = require(/** @type {any} */('../.lib-dist/fs')).FS;
 
+const SPECTATOR_CHANNEL = 5;
+const OMNISCIENT_CHANNEL = 5;
```
Zarel: ummm these should probably not both be `5`.

Also under my proposed change, spectator would be `0` and omniscient would be `-1`.

server/roomlogs.js
```diff
@@ -90,16 +93,18 @@ class Roomlog {
...
-				i += 4;
+				i += 2;
 			} else {
 				log.push(line);
 			}
 		}
 		let textLog = log.join('\n') + '\n';
-		if (channel === 0) {
+		if (channel === SPECTATOR_CHANNEL) {
 			return textLog.replace(/\n\|choice\|\|\n/g, '\n').replace(/\n\|seed\|\n/g, '\n');
```
Zarel: `|choice|` and `|seed|` are no longer used; this entire line can be removed.

server/roomlogs.js
```diff
@@ -90,16 +93,18 @@ class Roomlog {
 		log = [];
 		for (let i = 0; i < this.log.length; ++i) {
 			const line = this.log[i];
-			if (line === '|split') {
-				const ownLine = this.log[i + channel + 1];
+			const split = /\|split\|p(\d)/g.exec(line);
+			if (split) {
+				const offset = (channel === OMNISCIENT_CHANNEL || Number(split[0]) === channel) ? 1 : 2;
```
Zarel: I would probably not draw the line there.

```suggestion
				const canSeePrivileged = (channel === Number(split[0]) || channel === OMNISCIENT_CHANNEL);
```

server/roomlogs.js
```diff
@@ -90,16 +93,18 @@ class Roomlog {
...
+				const ownLine = this.log[i + offset];
```
Zarel:
```suggestion
				const ownLine = this.log[i + (canSeePrivileged ? 1 : 2)];
```

server/sockets.js
Zarel: I'm nearly 100% confident this isn't doing the right thing. You're only sending the split part of the message, not the entire message.

server/sockets.js
```diff
@@ -567,31 +569,15 @@ if (cluster.isMaster) {
 			room = rooms.get(roomid);
 			if (!room) return;
 
-			/** @type {[string?, string?, string?]} */
-			let messages = [null, null, null];
-			message = data.substr(nlLoc + 1);
+			const split = SPLIT_REGEX.exec(data.substr(nlLoc + 1));
```
Zarel: At this point, `data` is something like

```
>battle-foobar
|start|battle
|split|p1
|switch|p1: Pikachu|350/350
|switch|p1: Pikachu|100/100
```

You want to send to player 1:

```
>battle-foobar
|start|battle
|switch|p1: Pikachu|350/350
```

and to player 2:

```
>battle-foobar
|start|battle
|switch|p1: Pikachu|100/100
```

Zarel: As far as I can tell, your code instead sends to player 1:

```

|switch|p1: Pikachu|350/350
```

and to player 2:

```

|switch|p1: Pikachu|100/100
```

The function I pasted in comments will be useful for your sending the correct message.

sim/battle-stream.ts
```diff
@@ -209,24 +209,24 @@ export function getPlayerStreams(stream: BattleStream) {
...
-				const p2Update = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n([^\n]*)\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-				streams.p2.push(p2Update);
-				// p3 and p4 share update information with p1 and p2 respectively.
-				streams.p3.push(p1Update);
-				streams.p4.push(p2Update);
-				const specUpdate = data.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-				streams.spectator.push(specUpdate);
-				const omniUpdate = data.replace(/\n\|split\n[^\n]*\n[^\n]*\n[^\n]*/g, '');
-				streams.omniscient.push(omniUpdate);
+				const split = SPLIT_REGEX.exec(data);
```
Zarel: This has the same flaw as the `:` handler in `sockets`. This is again a situation you need the function I pasted in PR comments.

sim/battle.ts
```diff
@@ -3123,7 +3129,7 @@ export class Battle extends Dex.ModdedDex {
 	}
 
 	getDebugLog() {
-		return this.log.join('\n').replace(/\|split\n.*\n.*\n.*\n/g, '');
+		return this.log.join('\n').replace(/\|split|.*\n.*\n.*\n/g, '');
```
Zarel: This is again incorrect, and is again something best handled with the function I pasted in comments.

sim/battle.ts
scheibo: So I came up with something which I think works. Before:

```
  '|split',
  '|-sethp|p1a: Gourgeist|44/48|p2a: Dusknoir|48/48|[from] move: Pain Split',
  '|-sethp|p1a: Gourgeist|251/271|p2a: Dusknoir|48/48|[from] move: Pain Split',
  '|-sethp|p1a: Gourgeist|44/48|p2a: Dusknoir|231/231|[from] move: Pain Split',
  '|-sethp|p1a: Gourgeist|251/271|p2a: Dusknoir|231/231|[from] move: Pain Split',
```

after:

```
  '|split|p1',
  '|-sethp|p1a: Gourgeist|251/271|p2a: Dusknoir|[from] move: Pain Split',
  '|-sethp|p1a: Gourgeist|44/48|p2a: Dusknoir|48/48|[from] move: Pain Split',
  '|split|p2',
  '|-sethp|p1a: Gourgeist|44/48|p2a: Dusknoir|231/231|[from] move: Pain Split',
  '|-sethp|p1a: Gourgeist|44/48|p2a: Dusknoir|48/48|[from] move: Pain Split',
```

Its the only rare case where the new protocol is more verbose, but handling it also removes my qualms about `add` being a [partial function](https://wiki.haskell.org/Partial_functions).

server/sockets.js
scheibo: I think the fundamental misunderstanding I had is that message would be one logical command, not multiple.

server/sockets.js
scheibo: Yep, didn't realize the message would contain multiple commands. Thanks.

server/room-battle.js
scheibo: Who's "theory"? Do you really share information in 2v2? Also, do we know the gameType here?

sim/battle.ts
Zarel: I'n confused.

Did you mean:

```
|split|p1
|-sethp|p1a: Gourgeist|251/271|[from] move: Pain Split
|-sethp|p1a: Gourgeist|44/48|[from] move: Pain Split
|split|p2
|-sethp|p2a: Dusknoir|231/231|[from] move: Pain Split
|-sethp|p2a: Dusknoir|48/48|[from] move: Pain Split
```

Because that's what I would have done.

Zarel: I did think of this as a possible problem, but when I checked Leech Seed and saw that it didn't do this, I didn't think of Pain Split.

server/sockets.js
```diff
@@ -446,6 +446,29 @@ if (cluster.isMaster) {
 		}
 	}, 10 * MINUTES);
 
+	/**
+	 * @param {string} message
+	 * @param {number} channelid
+	 */
+	const extractChannel = (message, channelid) => {
```
Zarel: Apparently you can't use `function extractChannel` if you want to make it `const`. I still think the `function` notation looks better.

Zarel: It's at least more consistent with the rest of the codebase. If we want to break with established style, we should probably do it all at once.

server/sockets.js
```diff
@@ -567,31 +590,14 @@ if (cluster.isMaster) {
...
-					break;
-				default:
-					if (!messages[0]) {
-						messages[0] = message.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-					}
-					socket.write(messages[0]);
-					break;
-				}
+				const channelid = roomChannel ? Number(roomChannel.get(socketid)) : 0;
+				if (!messages[channelid]) messages[channelid] = extractChannel(message, channelid);
```
Zarel: May need to be `messages[channelid + 1]` under the new definition of `channelid`.

sim/battle.ts
```diff
@@ -2954,8 +2966,28 @@ export class Battle extends Dex.ModdedDex {
 		}
 	}
 
+	// omniscient = -1, spectator = 0
+	static updateForSide(data: string, side: -1 | 0 | 1 | 2 | 3 | 4 = 0) {
```
Zarel: `extractChannel` seems to be the better name here.

This name is kind of misleading since I'd expect functions starting with `update` to be impure, and return a success state.

sim/battle.ts
Zarel: Also: `add` is _still_ a partial function. :p

Just throw if you get privileged data from more than one side in a line.

server/sockets.js
scheibo: lol i had gone with `const` over `function` because literrally everything else in this block is a `const`, not a` function`. also, IIRC functions names and definitions gets hoisted in JS, so declaring them in a block is sort of misleading.

sim/battle.ts
scheibo: AIUI, there aren't "channels" in the sim (unlike in the server), the closet analog is side. Would you be OK with `extractUpdateForSide`?

sim/battle.ts
scheibo: > Because that's what I would have done.

`|-sethp|p1a: Gourgeist|251/271|[from] move: Pain Split` alters the definition of `-sethp` and requires client changes, no? You changed the number of args. Also, just noticed the p1 split in my `example` above is wrong, I copied it before I fixed the code.

> Also: `add` is *still* a partial function. :p

Is it? Am I missing something obvious? What valid inputs result in us throwing?

> Just throw if you get privileged data from more than one side in a line.

And modify the Pain Split calls + client?

server/sockets.js
Zarel: Functions are apparently block-scoped in ES2015 (previously, functions in if-blocks were undefined behavior).

But you're right, this is fine and should be left alone. Matching the other code in this file is better. Although probably some sort of restructure so we're not defining functions in if-blocks is probably the _best_ approach here.

sim/battle.ts
Zarel: Considering you're passing a channel ID to this function, this function introduces channels to the sim. The function should probably take a side ID if you didn't want to do that.

server/sockets.js
scheibo: Hmm, The old code was `'1'`, `'2'`, and `default`, with no `roomChannel` becoming `0`, where `0` was actually grabbing the first of the 4 splits (which corresponds to spectator, as channelid = 0 still does). What makes you think it should be +1?

scheibo:
https://github.com/Zarel/Pokemon-Showdown/pull/5332#discussion_r271574936, https://github.com/Zarel/Pokemon-Showdown/pull/5332#discussion_r271886621, and https://github.com/Zarel/Pokemon-Showdown/pull/5332#discussion_r271529554 are still pending resolution.

server/sockets.js
Zarel: `channelid` is currently an int ranging from `-1` to `4`, right? `+1` makes that `0` to `5`, allowing it to be an array index. `messages` is a length-6 array, after all.

sim/battle.ts
Zarel: The client is definitely designed to handle single-argument `|-sethp|`. I intentionally designed it for that. The client might need minor updates for animation stuff, and maybe the first one should be tagged `|[silent]`, but that's about it.

By "`add` is a partial function", I mean that it's not valid to pass it two separate splitter functions for different sides, and it's not possible to prevent that with the type system, so it should be prevented by throwing a runtime error.

server/room-battle.js
Zarel: I'm relatively sure they share max-HP information, which is the main reason why have the channel system in the first place.

server/sockets.js
scheibo: Are we reading the same code? `messages` is 5 elements (was previously 3 under the 4 stream system). I also don't think `channelid` is an int from -1 to 4? `roomChannel.get(socketid)` seems to return a string `'1'` or `'2'` under the old code?

server/room-battle.js
scheibo: max-HP is one thing, but theres also a bunch of other privileged information. do we need research to figure out exactly what is shared? I guess it doesn't matter because with the code on HEAD it gets shared anyway, so this wouldnt be introducing leaks (though they might already exist). 

This only applies to 2v2 though (GameType `'multi'`), right? Can we detect that here?

sim/battle.ts
scheibo: > The client is definitely designed to handle single-argument |-sethp|. I intentionally designed it for that. The client might need minor updates for animation stuff, and maybe the first one should be tagged |[silent], but that's about it.

OK. I didn't document `|-sethp|` (which I will in a followup CL, as well as `|callback|` which is also absent), so I didn't know this. Looking at `src/battle.ts` in the client, youre correct, it looks like it would handle missing a pokemon.

> By "`add` is a partial function", I mean that it's not valid to pass it two separate splitter functions for different sides, and it's not possible to prevent that with the type system, so it should be prevented by throwing a runtime error.

I think my confusion comes from that fact that `add` on HEAD is not a partial function, and my remodified version here is also not partial, but you're saying it *should* be partial, not that it *is* partial. I'm OK with that, I just think we were talking past each other.

Zarel: > (which I will in a followup CL, as well as `|callback|` which is also absent)

In theory `|callback|` should no longer exist...?

> I think my confusion comes from that fact that `add` on HEAD is not a partial function, and my remodified version here is also not partial, but you're saying it _should_ be partial, not that it _is_ partial. I'm OK with that, I just think we were talking past each other.

What's not partial about your remodified version? What happens if you pass it two splitter functions for different sides?

server/room-battle.js
Zarel: As far as I'm aware, the only privileged information is max-HP, gen-3 Pressure activation, and a bunch of hints. I'm fine with all of those being shared; that seems a much better solution than trying to separate them out into shared and unshared data.

Yes, this only applies to `'multi'`. In theory that should be `this.game.gameType` or something along those lines.

Zarel: Even more ideally, the player should track its own channel, so it doesn't need to be calculated on-the-fly here.

server/sockets.js
Zarel: You're right, `channelid` is a string, which makes it even less suited for being stuffed into a key of `messages` – I'm surprised TypeScript isn't complaining.

scheibo: Its not a string anymore - I `Number` it on purpose now.

Zarel: Oh wait no it's being cast to `Number`, no wonder it works here.

server/sockets.js
```diff
@@ -567,31 +590,14 @@ if (cluster.isMaster) {
...
-					socket.write(messages[2]);
-					break;
-				default:
-					if (!messages[0]) {
-						messages[0] = message.replace(/\n\|split\n([^\n]*)\n[^\n]*\n[^\n]*\n[^\n]*/g, '\n$1').replace(/\n\n/g, '\n');
-					}
-					socket.write(messages[0]);
-					break;
-				}
+				const channelid = roomChannel ? Number(roomChannel.get(socketid)) : 0;
```
Zarel: So `channelid` is guaranteed to be an integer from 0 to 4? That should probably be documented somewhere.

Zarel: In fact, `type ChannelID = 0 | 1 | 2 | 3 | 4` would probably help.

sim/battle.ts
scheibo: > In theory `|callback|` should no longer exist...?

#5414

> What's not partial about your remodified version? What happens if you pass it two splitter functions for different sides?

That's literally what Pain Split is currently doing (passing in two different splitters) and my code now handles it (though, with a different output than you want, as covered above). I'm OK to make `add` partial and change the logic in Pain Split to make two calls though.

server/sockets.js
scheibo:
```javascript
/**
 * roomid:socketid:channelIndex
 * @type {Map<string, Map<string, string>>}
 */
const roomChannels = new Map();
```

No earthly close why `channelIndex` is a `string`, but the fact that its an `index` means I'm now more confident its not going to be negative.

scheibo: Ah, its a `string` because it comes from `'.roomid, .channelid, .socketid'` which are all strings.

Zarel: Yep, I've investigated at it's in fact `0 | 1 | 2 | 3 | 4`.

Now all that remains is documenting it so people don't accidentally pass the wrong thing in.

server/room-battle.js
scheibo: Done, I think.

server/sockets.js
scheibo: Added a `ChannelID` type. Converting `roomid`/`socketid` to be numbers is I think out of scope.

server/room-battle.js
```diff
@@ -47,13 +47,16 @@ class RoomBattlePlayer {
 		user.updateSearch();
 
 		this.slot = slot;
-		this.slotNum = Number(slot.charAt(1)) - 1;
+		const channelIndex = Number(slot.charAt(1));
+		this.slotNum = channelIndex - 1;
+		this.channelIndex = /** @type {0 | 1 | 2 | 3 | 4} */
+			(game.gameType === 'multi' ? channelIndex % 2 + 1 : channelIndex);
```
scheibo: I think I still need the + 1 here (even though I've removed it from `channelIndex` because its not longer based on `slotNum`) because anything % 2 is going to be 0  | 1?

server/room-battle.js
```diff
@@ -47,13 +47,16 @@ class RoomBattlePlayer {
...
 		this.active = true;
 		this.eliminated = false;
 
 		for (const connection of user.connections) {
 			if (connection.inRooms.has(game.id)) {
-				Sockets.channelMove(connection.worker, this.game.id, this.slotNum + 1, connection.socketid);
+				Sockets.channelMove(connection.worker, this.game.id, this.channelIndex, connection.socketid);
```
scheibo: I don't know why this wasn't `this.slotNum % 2 + 1` before?

sim/README.md
```diff
@@ -122,7 +122,15 @@ An update which should be sent to all players and spectators.
 
   [sim-protocol]: https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/SIM-PROTOCOL.md
 
-One message type that only appears here is `|split`. This splits the next six lines into `spectator`, `p1`, `p2`, `p3`, `p4`, and `omniscient` messages. The `p1`, `p2`, `p3`, and `p4` logs will have exact HP values only for the corresponding player, while the `spectator` log will not have exact HP values for either player, and the `omniscient` logs will have exact HP values for both.
```
scheibo: Incidentally, this information was incorrect, there are still only 4 streams on HEAD. ;)

data/moves.js
```diff
@@ -11564,8 +11564,9 @@ let BattleMovedex = {
 		onHit(target, pokemon) {
 			let averagehp = Math.floor((target.hp + pokemon.hp) / 2) || 1;
 			target.sethp(averagehp);
+			this.add('-sethp', target, target.getHealth, '[from] move: Pain Split');
```
```suggestion
			this.add('-sethp', target, target.getHealth, '[from] move: Pain Split', '[simult]');
```

Zarel: I wonder if it should be necessary to specify which is the attacker and which is the defender.

Probably one needs to be silent.

server/room-battle.js
```diff
@@ -47,13 +47,16 @@ class RoomBattlePlayer {
 		user.updateSearch();
 
 		this.slot = slot;
-		this.slotNum = Number(slot.charAt(1)) - 1;
+		const channelIndex = Number(slot.charAt(1));
```
Zarel: Calling this `channelIndex` is pretty confusing since it's before the modulo.

server/room-battle.js
Zarel: `%2+1` will reverse the channels. I'd honestly rather figure this out from the `slotNum`, even if it's a bit silly to subtract 1 and then add 1.

scheibo: Done.

data/moves.js
scheibo: Played it out locally, it seems like the animation happens once, but I get messages twice:
```
Gengar used Pain Split!
The battlers shared their pain!
The battlers shared their pain!
```

Zarel: Yeah, I would just add `[silent]` to the first `-sethp`.

scheibo: Yeah, that seems to work.

Zarel:
I'll merge timed with a restart.

scheibo:
> I'll merge timed with a restart.

Do you have an ETA for this (does the server ever have scheduled restated, or is it only restarted in emergencies or when its been 'too long')? I only ask because its already missed one restart and you seemed to imply you wanted to done urgently. I worry this change will become forgotten and rot.

PR #5333, opened 2019-03-21
Pull Field out of Battle

scheibo:
@MacChaeger wants to move `sideConditions` to be positional `fieldConditions` instead, and `Battle` already does way too much. TBD whether we eventually move more into here or how `battle.ts`/`scripts.js` gets merged or split up further, but this step seems logical regardless of how things shake out.

Typescript has no complaints with this commit as is, but 31 failing tests has me believe I've broken everything which touches weather/terrain/pseudoWeather, which is not unexpected. Opening as a draft in this very unfinished state as a heads up and to get early high level feedback before I dive into actually fixing the tests.


Zarel:
This pull request **introduces 1 alert** when merging cac552f3eb08c3658888d75a028af10c90cce9df into 1cade1340a080b8094593085439204b6305dae22 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-7ffb6e75c352c33f5923388b45c4996b560ff2c8)

**new alerts:**

* 1 for Unused variable, import, function or class

---

Zarel:
I don't know how I feel about this, since it seems to just be tracking two state variables.

I kind of wonder if a lot of this code could be less-duplicated in some way.

Zarel:
Some sort of `EffectTracker` or something.

Zarel:
I was more open to this idea when I thought all mechanics could be consolidated here. If it's just these two, I'm iffier.

MacChaeger:
I'm planning on moving side conditions here later in the form of positional field effects, which could also involve grouping weathers, terrains, and pseudoweathers into one category, perhaps `fieldConditions`

scheibo:
> I don't know how I feel about this, since it seems to just be tracking two state variables.

3, but who's counting? It will eventually house side conditions as well, which would probably bump the class up to 300-400 lines, which is fairly substantial IMO.

I'm completely OK to drop this this. I'm not wedded to this design, its proven useful to get the juices flowing a bit regarding organizational issues.

> I kind of wonder if a lot of this code could be less-duplicated in some way.
> Some sort of `EffectTracker` or something.

Yeah, when I was [working on coming up with stricter types](https://github.com/pkmn-cc/state/blob/master/src/field.ts) I had these sharing a common `PersistentEffect` base type.

Really though, I'm not sure about `Side` either. It has a lot of things that seem fairly arbitrarily pulled out of `Battle` as well, and has to fall back and call `this.battle` a lot of the time which I think points to too tight of a coupling. I still would like Battle/Field/Side/Pokemon to be mostly stateful + objects, and pull out both the event system (runEvent/singleEvent/findEventHandlers) and the choice system (present in both Battle and Side) somehow.

scheibo:
> I was more open to this idea when I thought all mechanics could be consolidated here. If it's just these two, I'm iffier.

I think this is a case of "beggars can't be choosers" - `Battle` is doing a ton, and while this doesn't pull a huge chunk out of it, it is a logical split. Mac well make it bigger with his side conditions work, and maybe at that point we can also look into refactoring all 4 logical things `Field` would be responsible to share more logic. I don't think it should be done as part of this change or Mac's change - let's keep the reorganizing (this), features (Mac's work) and refactoring to share logic separate instead of trying to do everything at once.

dev-tools/globals.ts
```diff
@@ -219,7 +220,7 @@ interface EventMethods {
 	onEat?: ((this: Battle, pokemon: Pokemon) => void) | false
 	onEatItem?: (this: Battle, item: Item, pokemon: Pokemon) => void
 	onEffectiveness?: (this: Battle, typeMod: number, target: Pokemon | null, type: string, move: ActiveMove) => void
-	onEnd?: (this: Battle, pokemon: Pokemon & Side) => void
+	onEnd?: (this: Battle, target: Pokemon & Side & Field) => void
```
scheibo: Using `&` here is really sketchy - this should really be `|` because handlers are really expecting one or the other, not a combination of both.

Zarel: I assume the `&` is explicitly to shut up a TypeScript complaint that's really hard to deal with normally.

Zarel:
This pull request **introduces 1 alert** when merging c7d675cc8b01e13e2ad653e61b9dd38909f91887 into e2beda76a8bd4949336094a253e25d6e778bd823 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-37b7ff737af1e00cb193d17faba38c6e44440182)

**new alerts:**

* 1 for Unused variable, import, function or class

---

Zarel:
All the things that call `this.battle` should probably be merged into the new `Mechanics` system; however we decide to do that.

Also, I don't see any value in attaching side conditions to `Field`; I think they're better off in their current position.

dev-tools/globals.ts
scheibo: I don't think so, unless by "TypeScript complaint" you mean the compiler is telling us "you're not using types properly at all". Try changing these to `|` and run `npm run tsc` - handlers just want to not bother checking what type of object theyve been given and want to go about making assumptions about what fields they have.

So yeah, `&` is definitely for convenience, but makes a farce of the typesystem because its not true.

Zarel:
I'm not opposed to `Field`, btw – I'm overall indecisive. I'll probably end up merging this.

scheibo:
> All the things that call `this.battle` from `Pokemon`/`Side` should probably be merged into the new `Mechanics` system; however we decide to do that.

Still thinking on that.  I promise to send out a design doc soon, @MacChaeger just seemed to need `Field` first.

> Also, I don't see any value in attaching side conditions to `Field`; I think they're better off in their current position.

We can cover in @MacChaeger's upcoming PR when/if he moves them.

> I'm not opposed to `Field`, btw – I'm overall indecisive. I'll probably end up merging this.

Followed by immediately merging, lol.

PR #5334, opened 2019-03-22
Allow multi-line htmlbox

TheJetOU:


server/chat-plugins/info.js
```diff
@@ -2292,7 +2292,7 @@ const commands = {
 		if (!target) return this.parse('/help htmlbox');
 		target = this.canHTML(target);
 		if (!target) return;
-
+		target = target.replace(/\n/g, '');
```
Zarel: This should at least be `' '` – presence/absence of whitespace is significant in HTML.

This also affects `<pre>` blocks, which is harder to deal with.

Zarel: It looks like `&#10;` is the safest replacement.

Zarel: Although you have to do something smarter than a straight replacement – replace with a space inside tags and `&#10;` outside of tags.

Zarel:
This pull request **introduces 3 alerts** when merging fc52b4a6c5643015a2649d128f2dcfb401f42247 into 9f5a727ad1dd89bea529d9aaf9843040774407a4 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-8ccbbd42dc0666ddd2971e258441bb26524dbf24)

**new alerts:**

* 3 for Useless assignment to local variable

---

server/chat-plugins/info.js
Slayer95: Is there a need for all that? We might as well "escape" ``\n`` as ``\f`` and have the client convert it back.

Zarel: The HTML entity removes the need to unescape.

Zarel:
The 3 useless assignments are because you should be using `html` instead of `target` in three of your changes.

server/chat-plugins/info.js
```diff
@@ -2292,7 +2292,19 @@ const commands = {
...
+		if (target.includes('\n')) {
+			let OpenTags = 0;
+			let CloseTags = 0;
+			for (let line of target.split('\n')) {
+				OpenTags = OpenTags + (line.match(/<[^/].*>/g) || []).length;
+				CloseTags = CloseTags + (line.match(/<\/.*>/g) || []).length;
+				if (OpenTags > CloseTags) {
+					target = target.replace('\n', ' ')
+				} else if (OpenTags === CloseTags) {
+					target = target.replace('\n', '&#10;')
```
TheJetOU: Should I use ternaries here? I feel that nested ternaries are a bit unreadable

Zarel: You wouldn't need to nest them.

Also this isn't very DRY; you might want to write a function for this.

Zarel: I'd implement this as a few regexes.

Zarel:
```ts
target = target.replace(/<[^>]*>/g, tag => tag.replace(/\n/g, ' '));
target = target.replace('\n', '&#10;');
```

This probably belongs as a help function in `Chat`, rather than duplicated five places here.

TheJetOU: Care to explain why you don't replace `\n` globally in the last line (or was it mistake in your end), because some tags will end up on a linebreak, if the right criteria is met.

Zarel: Oh that was a typo. Should be:

```js
target = target.replace(/<[^>]*>/g, tag => tag.replace(/\n/g, ' '));
target = target.replace(/\n/g, '&#10;');
```

server/chat-plugins/info.js
```diff
@@ -2292,7 +2292,7 @@ const commands = {
 		if (!target) return this.parse('/help htmlbox');
 		target = this.canHTML(target);
 		if (!target) return;
-
+		if (target.includes('\n')) target = Chat.removeLineBreaksInHTML(target);
```
Zarel: The `if` part is unnecessary.

Zarel: `removeLineBreaksInHTML` seems kind of verbose, but I can't think of a better name.

`collapseNewlinesInHTML`?

Zarel: `collapseLineBreaksHTML`?

TheJetOU: I mean, those are just as verbose, maybe `collpaseHTML`?

Zarel: Yeah, but they're easier to understand.

Zarel: I currently like `collapseLineBreaksHTML` the most.

TheJetOU: I have no strong opinion on it, so sure.

PR #5340, opened 2019-03-23
Combine pseudo weathers and side conditions into field conditions

MacChaeger:
The goal of this PR is to make side conditions into positional field conditions, removing the need for any hacks in multi battles with respect to things such as hazards. It should also greatly simplify things like Wish. Furthermore, merging terrains and weathers into the general field conditions should simplify the interaction between Leftovers and Grassy Terrain.

### DIFF
--- a/data/items.js
+++ b/data/items.js
@@ -142,7 +142,7 @@ let BattleItems = {
 			basePower: 10,
 		},
 		onStart(target) {
-			if (!target.ignoringItem() && !this.field.getPseudoWeather('gravity')) {
+			if (!target.ignoringItem() && !this.field.getFieldCondition('gravity')) {
 				this.add('-item', target, 'Air Balloon');
 			}
 		},
@@ -2609,7 +2609,7 @@ let BattleItems = {
 		},
 		onEffectiveness(typeMod, target, type, move) {
 			if (!target) return;
-			if (target.volatiles['ingrain'] || target.volatiles['smackdown'] || this.field.getPseudoWeather('gravity')) return;
+			if (target.volatiles['ingrain'] || target.volatiles['smackdown'] || this.field.getFieldCondition('gravity')) return;
 			if (move.type === 'Ground' && target.hasType('Flying')) return 0;
 		},
 		// airborneness negation implemented in sim/pokemon.js:Pokemon#isGrounded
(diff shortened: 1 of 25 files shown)


MacChaeger:
Obviously, I have yet to actually finish changing side conditions, but I'm working on it.

Zarel:
> removing the need for any hacks in multi battles with respect to things such as hazards

The "hack" is a check for the existence of allied sides, and it's approximately one line:

https://github.com/Zarel/Pokemon-Showdown/blob/6d563f555ab317834e942b7ea1ec0274adbc0fd1/sim/battle.ts#L745

Your proposed code seems much more hacky, and doesn't seem like it _removes_ the need for a check for the existence of allied sides, so much as that it spreads it out over more code.

> It should also greatly simplify things like Wish

Wish and Future-Sight-likes appear to be the only slot-positional conditions. Supporting them natively in the future would certainly be nice, but that seems completely irrelevant to moving side conditions to the field. If anything, they would make the most sense as a new `slotConditions` array-of-tables on a `Side`.

> Furthermore, merging terrains and weathers into the general field conditions should simplify the interaction between Leftovers and Grassy Terrain.

You mean in terms of residual order? I don't see how this this adds or removes the need for complex sorting of `Residual` handlers.

Zarel:
In conclusion, I've _still_ yet to see any advantage to this change. Putting side conditions in `side.ts` makes intuitive sense to me so I'm not willing to do it unless I see any advantage at all to this approach.

Like, I can see the advantage to treating "an entire group of allies" as a position, but I think the most natural place for storing that position-condition data is `side.sideConditions`, in `p1`/`p2` in multi-battles.

You are free to continue working on this if you think there is some _other_ advantage I'll be able to understand once I see the code, just, if I don't end up accepting it, you've been warned.

MacChaeger:
That's what I had when I went to bed (before you commented). It's not done by any means; I still have to sort out terrains and weathers only activating once per Pokemon, duration only decreasing once per turn, and other things; there are 41 failing tests as of now. What I'm going to do now instead of fixing those tests is try another approach to getting side conditions to make sense in multi battles. Specifically, I'm going to have all side conditions be stored on sides `p1` and `p2` in multi battles, which will necessitate implementing `slotConditions`.

Zarel:
Fair. That might be pretty close to what I'd want, then.

PR #5345, opened 2019-03-25
Restore Cancel Button

### LINKED ISSUE #5344: Cancel button broken after #5266
> the latest git code on my server seems to have cancel button not working, and I believe it's likely the choice related work that was done recently that caused it...pre do you mind looking into it?

_[18:49:29]  **Freedom SK**_

I bisected and found 6ac82666966135445b4f7a0732a9f8da270f4ea7 is where the cancel button is no longer presented as an option in battles. @MacChaeger, can you PTAL?

MacChaeger:
Fixes #5344

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1022,15 +1022,12 @@ export class Battle extends Dex.ModdedDex {
 		}
 		}
 
-		let allRequestsMade = true;
-		for (const request of requests) {
-			if (!request) allRequestsMade = false;
-		}
+		let allRequestsMade = !requests.some(request => !request);
 		for (let i = 0; i < this.sides.length; i++) {
 			const side = this.sides[i];
 			const request = requests[i];
 			if (request) {
-				if (!this.supportCancel || allRequestsMade) request.noCancel = true;
+				if (!this.supportCancel || !allRequestsMade) request.noCancel = true;
 				side.emitRequest(request);
 			} else {
 				side.emitRequest({wait: true, side: side.getRequestData()});
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1022,7 +1022,7 @@ export class Battle extends Dex.ModdedDex {
 		}
 		}
 
-		let allRequestsMade = !requests.some(request => !request);
+		let allRequestsMade = requests.every(request => request);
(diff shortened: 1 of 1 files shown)


sim/battle.ts
```diff
@@ -1022,15 +1022,12 @@ export class Battle extends Dex.ModdedDex {
 		}
 		}
 
-		let allRequestsMade = true;
-		for (const request of requests) {
-			if (!request) allRequestsMade = false;
-		}
+		let allRequestsMade = !requests.some(request => !request);
```
scheibo:
```

```suggestion
		let allRequestsMade = requests.every(request => request);
```

sim/battle.ts
```diff
@@ -1022,15 +1022,12 @@ export class Battle extends Dex.ModdedDex {
...
 		for (let i = 0; i < this.sides.length; i++) {
 			const side = this.sides[i];
 			const request = requests[i];
 			if (request) {
-				if (!this.supportCancel || allRequestsMade) request.noCancel = true;
+				if (!this.supportCancel || !allRequestsMade) request.noCancel = true;
```
scheibo: Can we add a regression test for this (or fix any cancel tests we already have that should have caught this)?

MacChaeger: For the record, this only affected whether or not the cancel button actually appeared client-side. It didn't actually stop users from sending `/choose undo` and cancelling their choice.

scheibo: True. So less critical to worry about, but if we could cover it with a test without too much effort itd still be worth doing.

Zarel: Also, basically zero users know about `/choose undo`, so this is approximately as critical; it just means the test needs to be designed to account for this case.

scheibo: I think the naive way of adding the test would be to add a new method to ``test/assert.js`` which allows for asserting on the latest ``request`` object on the battle log for a side (unless theres an easier way to access the request in a test than searching through the battle log)

MacChaeger: Yeah I was just saying that's why the tests passed.

Zarel: We should probably rewrite the relevant code to make it easier to extract requests; that sounds less effort than screwing around with `test/assert.js`.

PR #5348, opened 2019-03-25
Test 'noCancel' behavior to prevent #5344

scheibo:
> We should probably rewrite the relevant code to make it easier to extract requests; that sounds less effort than screwing around with `test/assert.js`.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5345_

- saves `lastRequest` emitted by `Side`
- renames `currentRequest` to `requestState` (open to naming suggestions) to avoid confusion with `lastRequest`, given they are different shapes
- check `noCancel` state in tests - some tests seem to have uninuitive `noCancel` behavior - added comments ~and left as `FIXME` for now. **NOTE:** Not sure if the implementation or tests are actually incorrect or if I don't understand the intended behavior, but I figured I'd highlight.~


test/simulator/misc/decisions.js
```diff
@@ -726,15 +726,17 @@ describe('Choice extensions', function () {
...
+			it.skip(`should disallow to ${mode} decisions after every player has sent an unrevoked action`, function () {
 				battle = common.createBattle({cancel: true});
 				battle.setPlayer('p1', {team: [{species: "Bulbasaur", ability: 'overgrow', moves: ['tackle', 'growl']}]});
 				battle.setPlayer('p2', {team: [{species: "Charmander", ability: 'blaze', moves: ['tackle', 'growl']}]});
 
 				battle.choose('p1', 'move tackle');
 				battle.choose('p2', 'move growl');
-				if (mode === 'revoke') battle.undoChoice('p1');
-				battle.choose('p1', 'move growl');
+				assert(battle.p1.lastRequest.noCancel);
```
Slayer95: The next turn has already started, so ``noCancel`` is falsey, and ``battle.undoChoice()`` is noop.

Slayer95:
RE: maybe-cases

Being unable to cancel only applies if the player chooses to move/switch when *that Pokémon* is maybe-disabled/maybe-trapped, *respectively*, so the global ``noCancel`` shouldn't be ``true``.

However, it could be tested if ``noCancel`` were a per-Pokémon flag of the form ``true | 'move' | 'switch'``. (Is it worth it?)

test/simulator/misc/decisions.js
scheibo: Ah! Of course. I've added a comment.

Slayer95: ``battle.undoChoice()`` could be changed to error if ``side.actions.length === 0``.

FWIW, in normal play, that code path would be taken if a player sends a bad decision and tries to undo it immediately afterwards -before the room battle is aware that the decision was faulty.

scheibo:
I'm OK just commenting about why the maybe-cases are they way they are for now. Extending it seems useful, but a little more work than my immediate goal of regression testing #5344.

Zarel:
Maybe activeRequest is a better name than lastRequest? Also requestState is tracked inside it, right? No need to track it separately.

Slayer95:
``requestType`` would be best imo. The names ``activeRequest`` and ``lastRequest`` suggest that it's the actual object returned by ``getRequestData()``.

Zarel:
`activeRequest`/`lastRequest` _is_ the actual data returned by `getRequestData()`, isn't it?

Slayer95:
Oh, mb

scheibo:
> Maybe `activeRequest` is a better name than `lastRequest`?

I may be confused about the ordering here - `requestState` is the request we're currently building, and `lastRequest` is the request which was sent. We refer to the current `requestState` to be able to build `lastRequest`. I don't think `active` is accurate?

> Also `requestState` is tracked inside it, right? No need to track it separately.

No? The emitted request shape differs depending on the `requestState` (`'switch'` requests have `'forceSwitch': true'`, `'teampreview'` requests have '`teamPreview': true`, `'move'` request have '`active'`), but its not explicit.

Zarel:
`activeRequest` is the request that we're waiting for a response for. In fact, when `commitDecisions` is called, `activeRequest` should be cleared.

scheibo:
Done. I was wrong, I didn't think that would work.

I know you don't like `get`ters, but it really seems logical in this case... (and also means I don't need to modify the request object and/or change a zillion callsites to be more verbose)

Zarel:
I would have solved this by actually adding a `type` property on the `activeRequest`.

PR #5352, opened 2019-03-25
Update dependencies

scheibo:
Noticed the yellow badge in the README. Not sure about the `sockjs` dependency, was it pinned at `'0.3.18'` for a reason, or did #2795 just forget `'^'`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "0.11.2",
   "dependencies": {
     "probe-image-size": "^4.0.0",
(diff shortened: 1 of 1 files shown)


Slayer95:
``^0.3.18`` doesn't make sense. See the [SemVer spec](https://semver.org/#spec-item-4)

scheibo:
sockjs is milking that 'initial development' state even harder than Pokemon Showdown milks 'Beta'.

PR #5353, opened 2019-03-26
Prevent a tournament exploit

MacChaeger:


server/tournaments/index.js
```diff
@@ -781,7 +781,11 @@ class Tournament {
 
 		// Prevent battles between offline users from starting
 		const from = Users.get(challenge.from.userid);
-		if (!from || !from.connected || !user.connected) return;
+		if (!from || !from.connected || !user.connected) {
+			// Prevent an exploit
+			if (from && !from.connected) this.cancelChallenge(from);
```
scheibo: I'd move this out of the first `if` and just make it 2 single line `if`s - it doesn't need to be nested, and is kind of odd given the similarity between the conditions. I'd also add some more explanation about what the 'exploit' is exactly.

MacChaeger: Thing is, we always want to `return` after cancelling the challenge, and I think it would be odd to make them two separate `if` statements when the second condition will always be true if the first one is.

MacChaeger: I also deliberately avoided describing the exploit in order to not publicize it.

scheibo: Fair enough, maybe come back and add an explanation once this is landed and the servers have been patched.

What about:

```javascript
if (!from || !user.connected) return;
if (!from.connected) {
  // Prevent an exploit
  if (from) this.cancelChallenge(from);
  return;
}
```

MacChaeger: That works.

MacChaeger:
OK I give up on the co-author thing

scheibo:
Aha I appreciate the effort, but don't worry about it in the future. :)

server/tournaments/index.js
```diff
@@ -781,7 +781,12 @@ class Tournament {
 
 		// Prevent battles between offline users from starting
 		const from = Users.get(challenge.from.userid);
-		if (!from || !from.connected || !user.connected) return;
+		if (!from || !user.connected) return;
+		if (!from.connected) {
+			// Prevent an exploit
+			this.cancelChallenge(from);
```
Slayer95: Seems like this should be handled in ``User#onDisconnect()``

MacChaeger: `User` objects don't store any reference to room games they're in, so we'd have to iterate through every single room on the simulator looking for tours the user is in and cancel their challenges. After all, there's nothing stopping a user from leaving a room while they're entered in its tournament.

Slayer95: It's ``user.games``

Zarel: Also like you should just cancel all challenges on disconnect; roomgames in theory shouldn't come into it.

server/tournaments/index.js
```diff
@@ -736,12 +736,12 @@ class Tournament {
 		this.update();
 	}
 	cancelChallenge(user, output) {
-		if (!this.isTournamentStarted) {
+		if (!this.isTournamentStarted && output) {
```
Zarel: Err, this should still return if no output, right?

PR #5354, opened 2019-03-26
Reject Magnet Pull on non-Earth Power Probopass

ascriptmaster:
Without Earth Power, there isn't anything Probopass can actually do to hurt the Steels that it traps, so it might as well get Sturdy in those situations.

### DIFF
--- a/data/mods/gen6/random-teams.js
+++ b/data/mods/gen6/random-teams.js
@@ -629,6 +629,8 @@ class RandomGen6Teams extends RandomTeams {
 					rejectAbility = template.types.includes('Ground');
 				} else if (ability === 'Limber') {
 					rejectAbility = template.types.includes('Electric');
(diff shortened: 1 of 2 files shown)


TheImmortal:
Hmm. Should we look at Golem-Alola as well?

ascriptmaster:
I personally feel like Alolan Golem has the Attack to hit hard with Thunder Punch, and Alolan Golem should always have it based on the code, as long as it doesn't have Return (in which case it has Galvanize). This is why I was focusing on Probopass, because it has weak Special Attack and both STABs are resisted by Steel. I'll leave it to your judgment, though.

Also, if Magnet Pull is rejected on Golem, it will get Galvanize, which is not good for it without Return.

TheImmortal:
Fair. Btw, forgot to change gen 6.

ascriptmaster:
ok

PR #5355, opened 2019-03-26
Random Battle: Fix odds of various forme differences

ascriptmaster:
Castform is moved to 1/3 odds because normal Castform isn't rolled.


TheImmortal:
We forgot about Alola formes.

ascriptmaster:
ah.

PR #5358, opened 2019-03-26
Fix doubled recoil

MacChaeger:
Fixes both Mind Blown and standard recoil moves.

### DIFF
--- a/data/scripts.js
+++ b/data/scripts.js
@@ -652,16 +652,17 @@ let BattleScripts = {
 			}
 			nullDamage = false;
 
-			for (let i = 0; i < damage.length; i++) {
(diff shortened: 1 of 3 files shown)


data/scripts.js
```diff
@@ -652,16 +652,17 @@ let BattleScripts = {
 			}
 			nullDamage = false;
 
-			for (let i = 0; i < damage.length; i++) {
+			for (let j = 0; j < damage.length; j++) {
```
Zarel: Is there a reason this isn't a for...of?

MacChaeger: Because at the time I originally wrote this loop I didn't know about the `.entries` trick.

PR #5359, opened 2019-03-26
Sim: Improve querying Pokémon

Slayer95:
- Implements ``Pokemon#nearbyAllies()``, ``Pokemon#nearbyFoes()``
- Implements ``Battle#getAllPokemon()``, ``Battle#getAllActive()``
- Refactors accordingly

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -965,6 +965,26 @@ export class Battle extends Dex.ModdedDex {
 		return null;
 	}
 
+	getAllPokemon() {
+		const pokemonList: Pokemon[] = [];
+		for (const side of this.sides) {
+			pokemonList.push(...side.pokemon);
+		}
+		return pokemonList;
+	}
+
+	getAllActive() {
+		const pokemonList: Pokemon[] = [];
+		for (const side of this.sides) {
+			for (const pokemon of side.active) {
+				if (pokemon && !pokemon.fainted) {
+					pokemonList.push(pokemon);
+				}
+			}
+		}
+		return pokemonList;
+	}
+
 	makeRequest(type?: string) {
 		if (type) {
 			this.currentRequest = type;
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -247,10 +247,8 @@ export class Battle extends Dex.ModdedDex {
 	}
 
 	updateSpeed() {
-		for (const side of this.sides) {
-			for (const pokemon of side.active) {
-				if (pokemon) pokemon.updateSpeed();
-			}
+		for (const pokemon of this.getAllActive()) {
+			pokemon.updateSpeed();
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


sim/battle.ts
```diff
@@ -965,6 +958,26 @@ export class Battle extends Dex.ModdedDex {
 		return null;
 	}
 
+	getAllPokemon() {
+		const qPokemon: Pokemon[] = [];
```
scheibo: why `qPokemon` and not just `pokemon`?

Zarel:
This pull request **introduces 1 alert** when merging 9ec996e7c592d2a2ec92d030342667885a875dc1 into d499a019f0cdf5fa2f0eaa52d5c1498fcfb4457f - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-5a0bd173c1bfd15d1f2987f72fde1ebfb946dd80)

**new alerts:**

* 1 for Useless conditional

---

sim/battle.ts
Slayer95: Because ``pokemon`` should only be used for the type ``Pokemon``, AFAIAC.

Slayer95:
LGTM alert seems to be a false positive, for code that isn't introduced in this PR.

sim/battle.ts
scheibo: I'm not sure I agree, but if you feel that way, I still think `qPokemon` is an abomination that looks like wannabe Hungarian notation or something. How about `all` or `allPokemon` here and `all` or `allActive` below?

Slayer95: ``pokemonList`` probably?

scheibo: better than `qPokemon`, but I'm generally not a fan of encoding the type/container in the variable name. My preference is usually to just add an `'s'`, but I think we can both agree `'pokemons'` is right out :P. `allPokemon` and `allActive` seems like it would be fairly explicit given they are the return values for methods named `getAllPokemon` and `getAllActive`, no?

Zarel: I weakly prefer `pokemonList`

sim/pokemon.ts
```diff
@@ -537,11 +537,14 @@ export class Pokemon {
 				side.n % 2 === team ? side.active : []
 			);
 		}
-		if (adjacentOnly) allies = allies.filter(ally => this.battle.isAdjacent(this, ally));
 		return allies.filter(ally => ally && !ally.fainted);
 	}
 
-	foes(adjacentOnly?: boolean): Pokemon[] {
+	nearbyAllies(): Pokemon[] {
```
Zarel: This is going to conflict with the same change I commented on MacChaeger's PR...

Slayer95: What's it? I can only find this comment, which seems to agree with my approach...

> I often wonder if these should be different functions, rather than the same function with different booleans passed to them. How much code do they even share?

See also: [https://ariya.io/2011/08/hall-of-api-shame-boolean-trap](Hall of API shame)

sim/battle.ts
scheibo: I don't 'strongly' prefer `all*`, I only strongly dislike `qPokemon`. `pokemonList` works for me if you two prefer it.

sim/pokemon.ts
Zarel: Oh yes I agree. I'm just saying MacChaeger will have fun merging this.

Zarel: I did suggest this to MacChaeger before you even PRed this, after all. :p

I'm not sure how much I like "nearby" vs "adjacent", though.

Zarel: `adjacentAllies` etc seems like it would be better; to use the same word throughout the codebase.

Zarel:
Hey the `adjacentAllies` naming thing was also important to me!

Slayer95:
Uh... I don't know how I missed that...
I only saw
> Zarel approved these changes 15 hours ago

PR #5360, opened 2019-03-27
Fix multihit moves continuing after a KO

MacChaeger:


test/simulator/abilities/aftermath.js
```diff
@@ -0,0 +1,21 @@
...
+let battle;
+
+describe('Atermath', function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it("should hurt attackers by 1/4 their max HP when this Pokemon is KOed by a contact move", function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Voltorb', level: 1, item: 'rockyhelmet', ability: 'aftermath', moves: ['sleeptalk']}]});
```
MacChaeger: The reason Rocky Helmet isn't in its own test is that I'm actually testing if they can work on the same hit, because it's possible there may be a bug related to this.

Zarel:
```suggestion
		battle.setPlayer('p1', {team: [
			{species: 'Voltorb', level: 1, item: 'rockyhelmet', ability: 'aftermath', moves: ['sleeptalk']},
		]});
```

test/simulator/abilities/aftermath.js
```diff
@@ -0,0 +1,21 @@
...
+		battle.setPlayer('p2', {team: [{species: 'Nidoqueen', ability: 'sheerforce', moves: ['doublekick']}]});
```
Zarel:
```suggestion
		battle.setPlayer('p2', {team: [
			{species: 'Nidoqueen', ability: 'sheerforce', moves: ['doublekick']},
		]});
```

data/scripts.js
```diff
@@ -604,8 +604,10 @@ let BattleScripts = {
 		/** @type {(Pokemon | false | null)[]} */
 		let targetsCopy = targets.slice(0);
 		let i;
-		for (i = 0; i < hits && !damage.includes(false); i++) {
+		for (i = 0; i < hits; i++) {
```
urkerab: If you put the `pokemon.hp` check here then this will happen after the `i++` automatically.

Zarel: I would rename `i` to `hit` and `hits` to `hitCount`.

data/scripts.js
```diff
@@ -647,9 +649,7 @@ let BattleScripts = {
 			// Modifies targetsCopy (which is why it's a copy)
 			[moveDamage, targetsCopy] = this.spreadMoveHit(targetsCopy, pokemon, move, moveData);
 
-			if (!moveDamage.some(val => val !== false)) {
-				break;
-			}
+			if (!moveDamage.some(val => val !== false)) break;
 			nullDamage = false;
 
 			for (let i = 0; i < damage.length; i++) {
```
Zarel: Look, shadowing!

data/scripts.js
Zarel: I would also probably do `hit = 1; hit <= hitCount; hit++`

So `hit` matches `move.hit`.

MacChaeger: The hp check is at the end so that explosion still does damage, but I hadn’t thought of just further checking whether or not we’re on the first hit. On that note, @DaWoblefet how does Parental Bond Final Gambit work?

data/scripts.js
MacChaeger: Didn’t I make this `j` in another commit?

MacChaeger: No, that was another loop.

MacChaeger: Guess there’s no need if I’m renaming the first `i`

data/scripts.js
Zarel: I very much don't want the `pokemon.hp` check here.

data/scripts.js
MacChaeger: Actually it was this loop.

data/scripts.js
DaWoblefet: Parental Bond Final Gambit only hits once.

urkerab: Ugh, Explosion. Thanks for explaining. I still think the extra `++` is awkward though, but maybe not as awkward as switching to e.g.

    let hit = 1;
    do {
      // ...
    } while (++hit <= targetHits && pokemon.hp && !damage.includes(false));

data/scripts.js
```diff
@@ -584,20 +584,20 @@ let BattleScripts = {
 		for (let i = 0; i < targets.length; i++) damage[i] = 0;
 		move.totalDamage = 0;
 		pokemon.lastDamage = 0;
-		let hits = move.multihit || 1;
-		if (Array.isArray(hits)) {
+		let targetHits = move.multihit || 1;
```
Zarel: Why is this `targetHits` and not `hitCount`?

PR #5363, opened 2019-03-27
Fixes for target resolution

### LINKED ISSUE #5361: Imprison and Snatch targetting is broken
A. Steps
1. Use Imprison
2. Try to use any move
3. See that you actually Struggle

B. Steps
1. Doubles: Use Snatch and an snatchable move
2. See that a slow foe Pokémon snatches the move

### LINKED ISSUE #5362: Redirection should work if the player targeted an empty ally slot
From https://github.com/Zarel/Pokemon-Showdown/issues/2367#issuecomment-435642464

Slayer95:
- Storm Drain and Lightningrod now redirect attacks originally targetted against an empty allied slot of the attacker.
- Imprison and Snatch no longer affect both ``self`` and ``foeSide``.

Fixes #5361 and fixes #5362.

### DIFF
--- a/data/mods/gen3/scripts.js
+++ b/data/mods/gen3/scripts.js
@@ -71,11 +71,11 @@ let BattleScripts = {
 			return false;
 		}
 
-		const targets = pokemon.getMoveTargets(move, target);
+		const {targets, pressureTargets} = pokemon.getMoveTargets(move, target);
 
 		if (!sourceEffect || sourceEffect.id === 'pursuit') {
 			let extraPP = 0;
-			for (const source of targets) {
+			for (const source of pressureTargets) {
 				const ppDrop = this.runEvent('DeductPP', source, pokemon, move);
 				if (ppDrop !== true) {
 					extraPP += ppDrop || 0;
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2186,7 +2186,8 @@ export class Battle extends Dex.ModdedDex {
 				target = pokemon.side.active[-targetLoc - 1];
 			}
 			if (target && !(target.fainted && target.side !== pokemon.side)) {
-				// We aren't targetting a fainted foe.
+				// Target is unfainted: no need to retarget
+				// Or target is a fainted ally: attack shouldn't retarget
 				return target;
 			}
 
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2185,16 +2185,13 @@ export class Battle extends Dex.ModdedDex {
 			} else {
 				target = pokemon.side.active[-targetLoc - 1];
 			}
-			if (target) {
-				if (!target.fainted) {
-					// target exists and is not fainted
-					return target;
-				} else if (target.side === pokemon.side) {
-					// fainted allied targets don't retarget
-					return null;
-				}
+			if (target && !(target.fainted && target.side !== pokemon.side)) {
+				// We aren't targetting a fainted foe.
+				return target;
 			}
-			// chosen target not valid, retarget randomly with resolveTarget
+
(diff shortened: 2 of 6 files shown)


sim/battle.ts
```diff
@@ -2185,16 +2185,13 @@ export class Battle extends Dex.ModdedDex {
 			} else {
 				target = pokemon.side.active[-targetLoc - 1];
 			}
-			if (target) {
-				if (!target.fainted) {
-					// target exists and is not fainted
-					return target;
-				} else if (target.side === pokemon.side) {
-					// fainted allied targets don't retarget
```
Zarel: Is this wrong?

sim/pokemon.ts
```diff
@@ -560,8 +560,10 @@ export class Pokemon {
 		return this.foes().filter(foe => this.battle.isAdjacent(this, foe));
 	}
 
-	getMoveTargets(move: Move, target: Pokemon): Pokemon[] {
+	getMoveTargets(move: Move, target: Pokemon): {hit: Pokemon[], pp: Pokemon[]} {
```
Zarel: I feel like `pressureTargets` is a better name than `pp`.

`targets`, `pressureTargets`.

sim/battle.ts
Slayer95: They don't retarget by default, but still can be redirected by Storm Drain / Lightningrod.

Slayer95:
Updated~

sim/pokemon.ts
```diff
@@ -610,11 +615,12 @@ export class Pokemon {
 			if (move.pressureTarget) {
 				// At the moment, this is the only supported target.
 				if (move.pressureTarget === 'foeSide') {
-					targets.push(...this.foes());
```
Slayer95: I must say that it's baffling that this used to work...

sim/battle.ts
Slayer95: BTW, it might be unclear, but this diff is equivalent to just changing the ``return null`` for ``return target``.

Zarel: Is the prevent-retargeting code done elsewhere or something?

data/mods/gen3/scripts.js
```diff
@@ -71,11 +71,11 @@ let BattleScripts = {
 			return false;
 		}
 
-		const targets = pokemon.getMoveTargets(move, target);
+		const {targets, ppTargets} = pokemon.getMoveTargets(move, target);
```
Zarel: I said `pressureTargets` because I think "pp" is unnecessarily vague. How am I supposed to figure it out means `targetsForThePurposesOfCalculatingAdditionalPPDeductionFromEffects`? Pressure is _waaaayyyyy_ clearer.

sim/battle.ts
Zarel: Oh, I understand now.

sim/battle.ts
```diff
@@ -2185,16 +2185,13 @@ export class Battle extends Dex.ModdedDex {
...
-					return null;
-				}
+			if (target && !(target.fainted && target.side !== pokemon.side)) {
+				// We aren't targetting a fainted foe.
```
Zarel: Then this comment is wrong.

Zarel: We could even do the checks separately!

Slayer95: Wrong how?

Zarel:
```suggestion
				// Target is unfainted: no need to retarget
				// Or target is a fainted ally: attack shouldn't retarget
```

Slayer95:
Updated ~

Zarel:
Thanks!

PR #5365, opened 2019-03-27
Send move effectiveness data to the client

Slayer95:
Also fixes `disabledSource`

### DIFF
--- a/data/mods/ssb/scripts.js
+++ b/data/mods/ssb/scripts.js
@@ -187,7 +187,7 @@ let BattleScripts = {
 		if (!item.zMove) return;
 		if (item.zMoveUser && !item.zMoveUser.includes(pokemon.template.species)) return;
 		let atLeastOne = false;
-		/**@type {AnyObject?[]} */
+		/**@type {RequestSubMoveSlot?[]} */
 		let zMoves = [];
 		for (const moveSlot of pokemon.moveSlots) {
 			if (moveSlot.pp <= 0) {
--- a/data/mods/ssb/scripts.js
+++ b/data/mods/ssb/scripts.js
@@ -199,7 +199,11 @@ let BattleScripts = {
 			if (zMoveName) {
 				let zMove = this.getMove(zMoveName);
 				if (!zMove.isZ && zMove.category === 'Status') zMoveName = "Z-" + zMoveName;
-				zMoves.push({move: zMoveName, target: zMove.target});
+				zMoves.push({
+					move: zMoveName,
+					target: zMove.target,
+					effectiveness: this.getEffectivenessHints(pokemon, zMove, zMove.target),
+				});
 			} else {
 				zMoves.push(null);
 			}
--- a/dev-tools/globals.ts
+++ b/dev-tools/globals.ts
@@ -1,9 +1,12 @@
 type Battle = import('./../sim/battle').Battle
 type Field = import('./../sim/field').Field
 type ModdedDex = typeof import('./../sim/dex')
-type Pokemon = import('./../sim/pokemon').Pokemon
 type Side = import('./../sim/side').Side
 type Validator = import('./../sim/team-validator').Validator
+
+type Pokemon = import('./../sim/pokemon').Pokemon
+type RequestPokemonData = import('./../sim/pokemon').RequestPokemonData
+type RequestSubMoveSlot = import('./../sim/pokemon').RequestSubMoveSlot
 
 type PageTable = import('./../server/chat').PageTable
 type ChatCommands = import('./../server/chat').ChatCommands
@@ -750,7 +753,7 @@ interface BattleScriptsData {
 	calcRecoilDamage?: (this: Battle, damageDealt: number, move: Move) => number
 	canMegaEvo?: (this: Battle, pokemon: Pokemon) => string | undefined | null
 	canUltraBurst?: (this: Battle, pokemon: Pokemon) => string | null
-	canZMove?: (this: Battle, pokemon: Pokemon) => (AnyObject | null)[] | void
+	canZMove?: (this: Battle, pokemon: Pokemon) => (RequestSubMoveSlot | null)[] | void
 	forceSwitch?: (this: Battle, damage: SpreadMoveDamage, targets: SpreadMoveTargets, source: Pokemon, move: ActiveMove, moveData: ActiveMove, isSecondary?: boolean, isSelf?: boolean) => SpreadMoveDamage
 	getActiveZMove?: (this: Battle, move: Move, pokemon: Pokemon) => ActiveMove
 	getSpreadDamage?: (this: Battle, damage: SpreadMoveDamage, targets: SpreadMoveTargets, source: Pokemon, move: ActiveMove, moveData: ActiveMove, isSecondary?: boolean, isSelf?: boolean) => SpreadMoveDamage
@@ -792,7 +795,7 @@ interface ModdedBattlePokemon {
 	boostBy?: (this: Pokemon, boost: SparseBoostsTable) => boolean | number
 	calculateStat?: (this: Pokemon, statName: StatNameExceptHP, boost: number, modifier?: number) => number
 	getActionSpeed?: (this: Pokemon) => number
-	getRequestData?: (this: Pokemon) => {moves: {move: string, id: string, target?: string, disabled?: boolean}[], maybeDisabled?: boolean, trapped?: boolean, maybeTrapped?: boolean, canMegaEvo?: boolean, canUltraBurst?: boolean, canZMove?: AnyObject | null}
+	getRequestData?: (this: Pokemon) => RequestPokemonData
 	getStat?: (this: Pokemon, statName: StatNameExceptHP, unboosted?: boolean, unmodified?: boolean, fastReturn?: boolean) => number
 	getWeight?: (this: Pokemon) => number
 	hasAbility?: (this: Pokemon, ability: string | string[]) => boolean
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -6,6 +6,7 @@
  */
 import Dex = require('./dex');
 global.toId = Dex.getId;
+import {MoveEffectiveness} from './constants';
 import * as Data from './dex-data';
 import {Field} from './field';
 import {Pokemon} from './pokemon';
@@ -2897,6 +2898,23 @@ export class Battle extends Dex.ModdedDex {
 		}
 
 		if (once) this.hints.add(hint);
+	}
+
+	getEffectivenessHints(source: Pokemon, move: Move, targetType: string) {
+		const effectiveness: MoveEffectiveness[] = [];
+		for (const target of source.allies(true).concat(source.foes(true))) { // Order matters
+			if (!target || target === source || target.fainted) {
+				effectiveness.push(MoveEffectiveness.INVALID);
+				continue;
+			}
+
+			if (!this.validTarget(target, source, targetType)) {
+				effectiveness.push(MoveEffectiveness.INVALID);
+			} else {
+				effectiveness.push(target.getEffectivenessCode(move));
+			}
+		}
+		return effectiveness;
 	}
 
 	add(...parts: (string | number | boolean | ((side: 0 | 1 | boolean) => string) | AnyObject | null | undefined)[]) {
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1448,7 +1448,7 @@ export class Battle extends Dex.ModdedDex {
 				pokemon.isStaleHP = pokemon.hp;
 			}
 		}
-		if (pokemon.getMoves().length === 0) {
+		if (pokemon.getMoveRequest().length === 0) {
 			pokemon.isStaleCon++;
 			pokemon.isStaleSource = 'struggle';
 		}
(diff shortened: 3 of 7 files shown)


sim/pokemon.ts
```diff
@@ -528,7 +542,7 @@ export class Pokemon {
 		return null;
 	}
 
-	allies(): Pokemon[] {
+	allies(faintedToo?: boolean): Pokemon[] {
```
Slayer95: I plead guilty !

sim/pokemon.ts
```diff
@@ -18,6 +18,20 @@ interface MoveSlot {
...
+	id: string;
+	move: string;
+	pp?: number;
+	maxpp?: number;
+	target?: string;
+	targetZ?: string;
+	disabled?: string | boolean;
+	disabledSource?: string;
+	effectiveness?: number[];
+	effectivenessZ?: number[];
```
Slayer95: Needed because of Flying Press and Freeze Dry

sim/pokemon.ts
```diff
@@ -18,6 +18,20 @@ interface MoveSlot {
...
 }
 
+ /** A Pokemon's move slot as sent to the client. */
+interface RequestMoveSlot {
+	id: string;
+	move: string;
+	pp?: number;
+	maxpp?: number;
+	target?: string;
+	targetZ?: string;
```
Slayer95: Overdue

sim/pokemon.ts
scheibo: `includeFainted`

sim/pokemon.ts
```diff
@@ -537,14 +551,15 @@ export class Pokemon {
...
+		if (faintedToo) return allies;
 		return allies.filter(ally => ally && !ally.fainted);
 	}
 
 	nearbyAllies(): Pokemon[] {
 		return this.allies().filter(ally => this.battle.isAdjacent(this, ally));
 	}
 
-	foes(): Pokemon[] {
+	foes(faintedToo?: boolean): Pokemon[] {
```
scheibo: ditto

sim/dex.ts
```diff
@@ -314,6 +314,23 @@ class ModdedDex {
...
+	getEffectivenessCode(
+		move: Move,
+		target: Pokemon
+	): number {
+		if (!this.getImmunity(move, target)) return 3;
+		if (move.ohko || move.damageCallback || move.damage) {
+			return 0;
+		}
+		const baseMod = this.getEffectiveness(move, target);
+		// @ts-ignore
```
scheibo: Can you make this `// @ts-ignore - dynamic lookup` or something similar? I'm trying to add explanations to when we use `// @ts-ignore` (like is [generally recommended](https://source.android.com/setup/contribute/code-style#use-standard-java-annotations) for `@SuppressWarnings` in Java), which helps make sure people are not ignoring things for the wrong reasons and makes it so we know which we can remove in the future when Typescript gets smarter (index signature crap) or we refactor.

Slayer95: Sure!

sim/pokemon.ts
Zarel: MacChaeger called it "forTargeting", but I like `includeFainted` more than either.

sim/dex.ts
```diff
@@ -314,6 +314,24 @@ class ModdedDex {
...
+			}
+		}
+		if (move.category === 'Status' || move.ohko || move.damageCallback || move.damage) {
+			return 0; // Effective
+		}
+		const baseMod = this.getEffectiveness(move, target);
+		// @ts-ignore - ModdedDex stubs Battle
+		const moveMod = move.onEffectiveness && move.onEffectiveness.call(this, baseMod, null, move.type, move);
+		const finalTypeMod = typeof moveMod === 'number' ? moveMod : baseMod;
+		if (finalTypeMod > 0) return 1; // Super effective
```
Zarel: Sending numeric enums over the protocol seems like a bad idea. At least send a string enum.

sim/pokemon.ts
```diff
@@ -18,6 +18,20 @@ interface MoveSlot {
 	virtual?: boolean;
 }
 
+ /** A Pokemon's move slot as sent to the client. */
+interface RequestMoveSlot {
+	id: string;
+	move: string;
```
Zarel: Do we send `id`? It's a bit redundant with `move`, isn't it?

sim/pokemon.ts
Zarel: I think the order is usually `zTarget` and `zEffectiveness`.

sim/pokemon.ts
```diff
@@ -684,9 +700,7 @@ export class Pokemon {
 		return (lockedMove === true) ? null : lockedMove;
 	}
 
-	getMoves(lockedMove?: string | null, restrictData?: boolean):
-	{move: string, id: string, disabled?: string | boolean,
-		disabledSource?: string, target?: string, pp?: number, maxpp?: number}[] {
+	getMoves(lockedMove?: string | null, restrictData?: boolean): RequestMoveSlot[] {
```
Zarel: `getMoveRequest` is probably the best name here. Request functions have always been a mess.

sim/dex.ts
```diff
@@ -314,6 +314,24 @@ class ModdedDex {
 		}
 	}
 
+	getEffectivenessCode(move: Move, target: Pokemon): number {
```
Zarel: I'm not sure if a function taking a `Pokemon` belongs in `dex.ts`...

sim/pokemon.ts
Slayer95: Yes. It is redundant, but it's still used by the client, so we'd need a fixed client deployed before changing this.

Slayer95:
Like this? Updated ~

sim/dex.ts
Slayer95: On hindsight, what I meant is ``Template``...

sim/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @license MIT
+ */
+
+const k = {
+	EFFECTIVENESS: {
+		INVALID: 'na',
+		NEUTRAL: 'e',
+		NO_EFFECT: 'ne',
+		NOT_VERY: 'nve',
+		SUPER: 'se',
```
Zarel: This is not what I mean by string-enum.

I mean like,

```
type MoveEffectiveness = 'invalid' | 'neutral' | 'immune' | 'not-very-effective' | 'super-effective';
```

sim/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+	},
+	MV_RESULT: {
```
Zarel: These three should be the only constants on `Battle`, and they should be directly on `Battle` imo.

Slayer95:
Now literally using ``enum``s !

sim/constants.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Simulator constants
+ * Pokemon Showdown - http://pokemonshowdown.com/
+ *
+ * @license MIT
+ */
+
+export const enum MoveEffectiveness {
```
scheibo: This doesn't work with sucrase

scheibo: > Compiles TypeScript code to JavaScript, removing type annotations and handling features like enums. Does not check types. Sucrase transforms each file independently, so you should enable the `isolatedModules` TypeScript flag so that the typechecker will disallow the few features like const enums that need cross-file compilation.

Which means we should probably turn on `isolatedModules` .

sim/constants.ts
scheibo: I think string-enum is the way to go here.

sim/constants.ts
Slayer95: ! The emitted code seems valid though?

Slayer95: Oh, I see. It's downgraded to a normal enum...

sim/battle.ts
```diff
@@ -3103,7 +3103,7 @@ export class Battle extends Dex.ModdedDex {
 		throw new UnimplementedError('canUltraBurst');
 	}
 
-	canZMove(pokemon: Pokemon): (AnyObject | null)[] | void {
+	canZMove(pokemon: Pokemon): ({move: string, target: string} | null)[] | void {
```
scheibo: please modify in `globals.ts` as well

sim/constants.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Simulator constants
+ * Pokemon Showdown - http://pokemonshowdown.com/
+ *
+ * @license MIT
+ */
+
+export enum MoveEffectiveness {
```
scheibo: Why a `constants.ts`? This can just be in `pokemon.ts`, no?

Slayer95: To facilitate modding, and with future refactors of magic numbers in mind.

sim/battle.ts
Slayer95: Is it necessary? Wouldn't it be enough to define an ``Interface`` in the same file?

Slayer95: Or, I guess, ``import`` it to the ``globals.ts``?

scheibo: No, I mean just make sure the signatures of `canZMove` here and `globals.ts` match to avoid misleading developers. 

I'm going to work to eliminate `globals.ts`, but in the mean time I think its helpful if you can look at either here or `globals.ts` to figure out the signature, as opposed to having to look at both and merge in your head (which is what Typescript does)

sim/constants.ts
scheibo: How does it help with those (sorry, maybe you could elaborate with an example? I'm better with examples :P)?

I'm not saying having a `constants.ts` file is wrong *a priori*, it just seems out of place currently.

Slayer95: Actually, there is a better reason: it's now going to be imported by ``battle.ts`` too (WIP) :p

Slayer95:
I took a step back and I think this is now looking much better

data/mods/ssb/scripts.js
```diff
@@ -187,7 +187,7 @@ let BattleScripts = {
 		if (!item.zMove) return;
 		if (item.zMoveUser && !item.zMoveUser.includes(pokemon.template.species)) return;
 		let atLeastOne = false;
-		/**@type {AnyObject?[]} */
+		/**@type {(RequestSubMoveSlot | null)[]} */
```
scheibo: This can `RequestMoveSlot?[]` because `?` means `| null` in `@type`

dev-tools/globals.ts
```diff
@@ -1,10 +1,13 @@
 type Battle = import('./../sim/battle').Battle
 type Field = import('./../sim/field').Field
 type ModdedDex = typeof import('./../sim/dex')
-type Pokemon = import('./../sim/pokemon').Pokemon
 type Side = import('./../sim/side').Side
 type Validator = import('./../sim/team-validator').Validator
 
+type Pokemon = import('./../sim/pokemon').Pokemon
+type RequestPokemonData = import('./../sim/pokemon').RequestPokemonData
```
scheibo: do `RequestPokemonData` and `RequestSubMoveSlot` work as `declare let Foo: import(...).Foo`?

Slayer95: idk ;-;

data/mods/ssb/scripts.js
Slayer95: Fixed

Zarel:
This pull request **introduces 1 alert** when merging 54dde847a24cb4dfceeedebe7ec9a7cb116a9a55 into 1f8eeae108009779f2f119a42ac1cb52e8e0cc4c - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-dbe175f0e00052ba2e8be2f54deaf23cbb0604cb)

**new alerts:**

* 1 for Unused variable, import, function or class

---

Zarel:
This pull request **introduces 1 alert** when merging d003c103928cef54972f23632a077127ef7feb98 into 1f8eeae108009779f2f119a42ac1cb52e8e0cc4c - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-acb8ca0cec6fe966c49095e958fd1ad88b1d5e64)

**new alerts:**

* 1 for Unused variable, import, function or class

---

sim/battle.ts
```diff
@@ -6,6 +6,7 @@
  */
 import Dex = require('./dex');
 global.toId = Dex.getId;
+import {MoveEffectiveness} from './constants';
```
Zarel: I still don't like having an external constants file. We've always just declared constants at the place they're used. Especially with string enums, you can just use the strings themselves instead of needing to import types.

Slayer95: > type MoveEffectiveness = 'invalid' | 'neutral' | 'immune' | 'resist' | 'weak';

Didn't you want type-checking? ``enum``s are the ideal way for that :|

scheibo: Type literals in Typescript accomplish the same thing, and are the pattern we've been using throughout the codebase

scheibo:
Is this blocked on Zarel and my opposition to `MoveEffectiveness` in its `constants.ts` file? I think making a `MoveEffectiveness`  type literal (`type MoveEffectiveness = 'invalid' | 'neutral' | 'immune' | 'resist' | 'weak'`) and putting it in `Battle` (where `pokemon.ts` can still import it - just with `import {Battle, MoveEffectiveness} from './battle'` would be preferable (or it could be in `Pokemon` and imported by `Battle`, though this seems sort of backwards)

Slayer95:
More importantly, it's still missing client-side support. I'd have wanted this for AFD, but we are past the deadline !

AnnikaCodes:
@KrisXV are you working on this?

mia-pi-git:
Going to close this since it's been inactive for several years. Please feel free to reopen it if it's still relevant.

PR #5370, opened 2019-03-28
Introduce `dev-tools/harness.js` for testing/benchmarking

scheibo:
> I wonder though if I should just get the testing logic committed (without any of the benchmark/profile/tracing/formatting options I keep bikeshedding) so that the harness can be used by others for looking for errors/debugging etc, and followup with my performance logic later?

_Originally posted by @scheibo in https://github.com/Zarel/Pokemon-Showdown/pull/5278#issuecomment-477388726_

I keep finding new yaks to shave with benchmarking/profiling, but this works as is for testing.


scheibo:
This could have caught the crashes from the server restarts today (provided it ran long enough and the PRNG was in its favor to give a mon 'Leech Seed' etc). I think to be more generally useful for smoke tests we should create a random team generator which doesn't aim to build 'good' mons, but instead aims for coverage - just making sure we have the best chance of playing battles with the most Pokemon/Moves/Abilities etc. Obviously we wouldn't be able to test each combination and interplay of factors, but just making sure we've used all 1000+ Pokemon and 1000+ moves would only take < 100 custom games (1000/12 with no team validation) which takes about 15s to run and would give us some confidence that we can at least avoid crashes. We could do this for each format (not just the random formats), including OMs and it would still take < 10 minutes. Definitely not something to run on each commit, but something that could be used to know if restarting the server will crash on battles on not. 

Creating such a generator is out of scope for this PR,  but where I think would be useful to take this on the testing front.

Slayer95:
> I think to be more generally useful for smoke tests we should create a random team generator which doesn't aim to build 'good' mons, but instead aims for coverage

That's the Hackmons Cup generator.

scheibo:
Made a strawman [prototype team generator](https://gist.github.com/scheibo/7089d7416d03a8dfbc6561c25b385504) inspired by Hackmons Cup. We want to be sure to exhaust all possible Items/Moves/Pokemon/Abilities in the pools before repeating. For each generation, we can then pre-generate a bunch of random teams and create custom games (without species clause, EV checks, etc) to run through the harness. I'll need to add logic to the runner to allow it to play in non-random metas, but thats not very difficult and I'll do it in the followup CL with the team generation logic.

scheibo:
@Zarel - PTAL when you're feeling better. I'm worried you'll confuse this for the perma-WIP #5278 which used to have this title.

---

My 'smoke test project' (which will follow up on this PR) is going well - I've got it running on all the singles formats pretty well ([logs](https://gist.github.com/scheibo/293f953c4d5124fd2880786521d2d858)). Doubles is proving to be troublesome (possibly a bug in the AI, given the logs), the pool sampling could maybe be improved (eg. 1516-1590, where it seems stuck not using one move for a while), and I need to fix logging/repeatability (in particular, when the bug occurs at the wrong level, like its doing now. Error logging/reproduction works well if the problem is in `sim`, less well if its in my generator/AI code :P). I reverted some of the fixes from yesterday's crashes and it catches them, which seems to validate my mission.

dev-tools/HARNESS.md
```diff
@@ -0,0 +1,51 @@
...
+    the next format.
+
+### Concurrency
+
+The harness runs games sequentially by default, but can be configured to run
+games asynchronously.
+
+-   **`--async`**: runs each game concurrently instead of waiting for the
+    previous game to complete before starting the next. Note that since battles
+    should not be blocking to begin with, this mode is not expected to have
```
Zarel: "since battles should not be blocking"?

sim/battle-stream.ts
```diff
@@ -41,12 +41,14 @@ function splitFirst(str: string, delimiter: string, limit: number = 1) {
 export class BattleStream extends Streams.ObjectReadWriteStream<string> {
 	readonly debug: boolean;
 	readonly keepAlive: boolean;
+	readonly retainBattle: boolean;
```
Zarel: We can just always retain the battle, instead of needing a setting for it.

sim/battle-stream.ts
```diff
@@ -41,12 +41,14 @@ function splitFirst(str: string, delimiter: string, limit: number = 1) {
 export class BattleStream extends Streams.ObjectReadWriteStream<string> {
 	readonly debug: boolean;
```
Zarel: Also why are these readonly? It makes sense that a consumer might want to change these.

sim/battle-stream.ts
Zarel: Like, the only reason we don't is because a past V8 had GC issues which we thought preemptively deleting references might help fix.

Slayer95:
> [slow battle] 2204ms - >start {"formatid":"digimondigimonshowdown","seed":[27939,54995,12592,54036]}
>player p1 {"name":"Bot 1","seed":[62304,1829,61481,48193]}
>player p2 {"name":"Bot 2","seed":[22370,56598,38891,14058]}
Run `node dev-tools/harness 1 --format=digimondigimonshowdown --seed=49587,60385,24801,60692` to debug (optionally with `--output` and/or `--input` for more info):
 Error: Push after end of read stream
    at BattleStream.resolvePush (/home/ubuntu/workspace/data/Pokemon-Showdown/.lib-dist/streams.js:544:37)
    at BattleStream.pushError (/home/ubuntu/workspace/data/Pokemon-Showdown/.lib-dist/streams.js:526:8)
    at BattleStream._write (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/battle-stream.js:63:10)
    at BattleStream.write (/home/ubuntu/workspace/data/Pokemon-Showdown/.lib-dist/streams.js:726:15)
    at ObjectReadWriteStream.write [as _write] (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/battle-stream.js:195:12)
    at ObjectReadWriteStream.write (/home/ubuntu/workspace/data/Pokemon-Showdown/.lib-dist/streams.js:726:15)
    at RandomPlayerAI.choose (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/battle-stream.js:300:15)
    at RandomPlayerAI.receiveError (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/examples/random-player-ai.js:36:8)
    at RandomPlayerAI.receiveLine (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/battle-stream.js:286:16)
    at RandomPlayerAI.receive (/home/ubuntu/workspace/data/Pokemon-Showdown/.sim-dist/battle-stream.js:274:9)

scheibo:
@Slayer95 - thats the same error I'm seeing on https://github.com/Zarel/Pokemon-Showdown/pull/5370#issuecomment-478179648, and I though I fixed it in the past on #5278. Maybe I didn't port everything relevant over from that PR.

sim/battle-stream.ts
scheibo: Gotcha.

dev-tools/HARNESS.md
scheibo: You're right, thats misleading. Added more explanation.

sim/battle-stream.ts
scheibo: Because when I originally ported it I did a manual `prefer-readonly` because I don't like things to be mutable :).

Slayer95:
Oh, right! That's a Doubles format. Just posted to make sure you didn't miss it.

sim/battle-stream.ts
Zarel: I personally think of `readonly` and `const` as "it would be a mistake to mutate these" and/or "it would break something if these were mutated", rather than just "nothing is currently mutating these".

scheibo: Yes, but I'm of the opinion that it's almost a mistake to ever mutate anything. I prefer to design all my code to be immutable, which is why for me 'it would break something if these were mutated' is my default.

Zarel:
Should I wait until the Doubles issue is fixed...?

scheibo:
> Should I wait until the Doubles issue is fixed...?

Its not like anything breaks if this gets committed as is, but its OK if you want to hold off, I'm working to fix the bug now.

scheibo:
https://github.com/Zarel/Pokemon-Showdown/commit/134fb8b852f897025327cd296cb58ccf35d69678 is what fixed the push after end of stream last time. Commenting out the `battle.destroy()` fixes things as well (but obviously isn't the right answer)

Slayer95:
And this seems to still have memory issues (now running without ``--async`` after figuring out that ``--async`` *wasn't* actually sequential).

History

Notes
1. First peak is after running about half an hour. By then, the output has plenty of ``[slow battle]`` warnings.
2. Second peak is a short run
3. Screenshot taken during a 1000 battles run

scheibo:
https://gist.github.com/scheibo/84e0331f38f724646824a29776cb0f89 is a repro of the push after stream that can be fed into `pokemon-showdown simulate-battle` (though it materializes as a different error). The `battle.inputLog` annoyingly doesn't capture the issue at all, because a) the input log != what the exact inputs were and b) the battle is over when additional pushes are made.

I've also made `RandomPlayerAI#receiveError` log for debugging, because it seems like the AI is papering over issues with `'default'` like I was worried it would (and like our tests were wont to do). I think maybe the `request.active.length > 1` check which I copied from the original `RandomPlayerAI` to determine whether we're not in singles and thus targeting information is required or not is incorrect - doesn't it fail if we're the only Pokemon left for a player? Anyway, I have plenty of fun things to debug when I wake up tomorrow.

> First peak is after running about **half an hour.**

> **512 MB**

OK, so these are definitely conditions I didn't account for! I've never run it for more than 5 minutes (2000 battles), and my computers have 16GB and 64GB respectively. I will look into this.

Slayer95:
> the request.active.length > 1 check which I copied from the original RandomPlayerAI to determine whether we're not in singles and thus targeting information is required or not is incorrect - doesn't it fail if we're the only Pokemon left for a playe

No, that check is fine.

scheibo:
Ah, because it should just be `[{}, null]` when theres only one poke?

Slayer95:
No, active Pokémon cannot contain ``null`` anymore. You might see some ``null`` checks in the code though. That enables support for bringing only 1 Pokémon in Doubles battles, which is currently forbidden in cartridge.

Slayer95:
Heap snapshot reveals that tons of ``update | [...]`` strings are retained.

scheibo:
I really need to go to bed, but #5381 was a pretty simple fix. Unfortunately, my 'push after stream' error no longer repros with that (the battles play out successfully), but I need to fix the 'push after stream' error as well. For my own reference, for this PR I want to fix:

- [ ] remove `receiveError()` `default` fallback. I originally punted, but at the minimum the AI should more strictly check the error its getting to detect if its 'trapped' or 'disabled' case, not just an invalid choice. I think it should also do the extra legwork to save the previous request and re-choose taking into account the error, but maybe I'll continue to punt on that.
- [ ] fix push after stream error
- [ ] come up with a better way of saving the true input log
- [ ] memory issues @Slayer95 has identified

The AI doesn't look like it will handle triple/rotation/multi battles very well either, but that's out of scope for now.

In my `smoke` branch, I need

- [ ] a way to repro runs better. Because of how the team generator + AI works, you have to rerun at the 'format' level instead of the 'battle' level, because its no longer the case the battle state and AI is fully determined from its starting seed - the state of the pools matters as well. 
- [ ] I need to special case options that only work with certain pokemon (z crystals with `zMoveUsers`/mega or ultra stones with `megaEvolves`/ are there any other moves/abilities/items that only work with certain pokemon?) because otherwise its very unlikely the smoke tests will be able to actually test mega evolution etc
- [x] fix team preview (need to choose `'team 2'` instead of just `'2'`, obviously)

dev-tools/harness.js
```diff
@@ -0,0 +1,229 @@
...
+	async run() {
+		let games = [];
+		let format, lastFormat;
+		let failures = 0;
+		while ((format = this.getNextFormat())) {
+			if (this.all && lastFormat && format !== lastFormat) {
+				// If we ran async, we need to now wait for each game and determine its status.
+				// NOTE: Promise.all doesn't help us here because it will resolve when the first
+				// rejection occurs, we need to wait for *all* the rejections.
+				if (this.async) for (const game of games) await game;
```
Slayer95: Actually, ``await`` has the same issue, since a failure would raise an exception.

Slayer95: However, as it is implemented, ``game`` never rejects.

scheibo: Yeah, it never rejecting is why this pattern is used.

scheibo: Right, I forgot I changed my approach around catching and tracking failures such that this pattern is wrong (in case you were curious, at one point these loops had a `try..catch` around the `await` and I didn't use the `catch(...) method`, so they did actually work). When I changed to use `catch` I got rid of the unnecessary `try..catch`, not realizing that these could now just use `Promise.all`.

dev-tools/harness.js
```diff
@@ -0,0 +1,229 @@
...
+				failures++;
+				console.error(
+					`Run \`node dev-tools/harness 1 --format=${format} --seed=${seed.join()}\` ` +
+					`to debug (optionally with \`--output\` and/or \`--input\` for more info):\n`, err);
+			});
+
+			if (!this.async) await game;
+			games.push(game);
+			lastFormat = format;
+		}
```
Slayer95: After removing the implementations of ``--async`` and ``--all``, I tried to rewrite this loop in all sorts of ways, but I couldn't convince V8 to GC the intermediate ``Promise`` and ``TickObject`` instances this creates...

An implementation that uses good old callbacks doesn't leak though :)
https://hastebin.com/odoguzecuw.js

Slayer95:
>  I need to special case options that only work with certain pokemon (z crystals with zMoveUsers/mega or ultra stones with megaEvolves/ are there any other moves/abilities/items that only work with certain pokemon?) because otherwise its very unlikely the smoke tests will be able to actually test mega evolution etc

You don't need to worry about that. If there is a mismatch between item and Pokémon, the request has the ``canMegaEvo``, etc. flags off.

scheibo:
> We need to listen to the `|callback|cant`| and `|callback|trapped|` messages.

Ah, was not aware of those. I don't think theyre documented in `SIM-PROTOCOL.md` either. :S

> What you do have to check, however, is that no more than 1 Pokémon should try to mega-evolve/ultra-burst/z-move in the same turn.

We already handle this, I think? `canMegaEvo`, `canZMove`, `canUltraBurst` guard against this - though maybe there's an edge case where we could try to Ultra + Mega in the same turn - is that allowed?

> You don't need to worry about that.

I'm not worried about the AI not handling mega evolution properly (it does it correctly already), I'm worried that in my `smoke` branch it doesnt e\ver get the opportunity to mega evolve because the teams it generates will rarely if ever have the correct combinations.

Zarel:
I keep on forgetting `|callback|cant` and `|callback|trapped` still exist. Shouldn't they be choice errors by now? Are they still used by anything? They should probably be removed entirely.

Zarel:
> We already handle this, I think? `canMegaEvo`, `canZMove`, `canUltraBurst` guard against this - though maybe there's an edge case where we could try to Ultra + Mega in the same turn - is that allowed?

It's allowed. But it's more that, it's possible for `canMegaEvo` to be `true` for multiple active Pokémon, but it's still only legal for one of them to mega evolve.

scheibo:
> But it's more that, it's possible for `canMegaEvo` to be `true` for multiple active Pokémon, but it's still only legal for one of them to mega evolve.

Yeah, I handle this, with `canMegaEvo`, `canZMove`, `canUltraBurst` variables (thats not what they're called in the request object), sorry that wasn't clear.

scheibo:
> I keep on forgetting `|callback|cant` and `|callback|trapped` still exist. Shouldn't they be choice errors by now? Are they still used by anything? They should probably be removed entirely.

Do they exist? I can't find them. I only see choice errors?

Zarel:
Okay, I panicked for nothing. Yes, I remember replacing `|callback|cant` and `|callback|trapped` with the current choice error system; I was afraid I missed something. Slayer95 probably just remembers a time where they still existed.

Zarel:
I've nearly entirely removed the `|callback|` system; currently the only vestige is:

https://github.com/Zarel/Pokemon-Showdown-Client/blob/5e52773beef7e67cd46944447d60d2b507ee71a4/src/battle.ts#L975-L984

scheibo:
4965b5996706c3933f27fab37cd55b7e35b8e8fb is where I'm testing out my changes to the AI so that it throws on issues instead of using `default` to paper over them. Handling being trapped properly was fairly easy, still working on the other errors though.

scheibo:
Actually, I lied:

```
sideupdate
p2
|callback|cant|p2a: Uxie||imprison

sideupdate
p2
|error|[Invalid choice] Can't move: Uxie's Imprison is disabled
```

It does in fact still exist!

```
sim/side.ts:315:    emitCallback(...args: (string | number | AnyObject)[]) {
sim/side.ts:408:                    this.emitCallback('cantz', pokemon);
sim/side.ts:462:                            this.emitCallback('cant', pokemon, disabledSource, moveid);
sim/side.ts:475:                    this.emitCallback('cantmega', pokemon);
sim/side.ts:483:                    this.emitCallback('cantmega', pokemon);
sim/side.ts:559:                            this.emitCallback('trapped', pokemon.position);
```

TBH though, I think we still want it, at least for trapping - see my comments in 697b700ea2b0ea6c745445e6a4d16206002aec1d - either we need to indicate the Pokemon which are trapped in the choice error, or just use `callback`. I think for maybe disabled/trapped a formal and specified `callback` is much better than parsing a arbitrary string error, and I think we should remove the choice errors in those two cases and use the callback instead.

dev-tools/harness.js
scheibo: FYI: When I get the code functionally correct I'm going to dig into the memory issues, but I really appreciate that you spent the time to pinpointed the reason being promises. However, I dont think a callback based implementation is the way to go unless absolutely necessary, because `await` and `Promise` are idiomatic and if using them with the PS APIs causes issues then we probably need to fix the APIs

scheibo:
```
[20:29:07] @Slayer95: pre, fixed the memory leak by doing await new Promise(resolve => {setImmediate(resolve)}) every N games run
[20:30:14] @Slayer95: as it turns out, there is not a single setImmediate in the BattleStream implementation, so it's in fact a sync stream
[20:30:50] @Slayer95: now, exactly how that impacts the GC logic isn't very clear
[20:30:56] @pre: i think thats a `BattleStream` or `Steams` issue then?
[20:31:01] @pre: but thanks for figuring that out
[20:31:15] @Slayer95: there are very few articles discussing memory leaks from looping async
[20:31:19] @pre: i was literally just commenting on your callback implementation from yesterday
[20:31:42] @Slayer95: yes, a fix in BattleStream / Streams should be feasible
```

Zarel:
Err, tell me whenever you think this is ready.

scheibo:
> Err, tell me whenever you think this is ready.

I've been tracking that with https://github.com/Zarel/Pokemon-Showdown/pull/5370#issuecomment-478211581. We need to fix #5391 for memory issues, and I'll pull 17159e7f29f3386f9c400f35c277f2d768171926 (`RandomPlayerAI` changes) out to land in a separate PR as well, at which point I believe we can land this (though once again, I think we could always land this at any point given it doesnt break anything in its current state, but making sure there are no known issues first is probably a better call)

dev-tools/harness.js
```diff
@@ -0,0 +1,237 @@
...
+			streams.p1, Object.assign({seed: this.newSeed()}, this.p1options)).start();
+		const p2 = new RandomPlayerAI(
+			streams.p2, Object.assign({seed: this.newSeed()}, this.p2options)).start();
+
+		streams.omniscient.write(`>start ${JSON.stringify(spec)}\n` +
+			`>player p1 ${JSON.stringify(p1spec)}\n` +
+			`>player p2 ${JSON.stringify(p2spec)}`);
+
+		let chunk;
+		while ((chunk = await Promise.race([streams.omniscient.read(), p1, p2]))) {
```
Slayer95: Why isn't this ``await streams.omniscient.read()`` ?

scheibo: Because then we'd be orphaning the `p1`/`p2` `start` promises.

Zarel:
This pull request **introduces 2 alerts** when merging f7675264bb0428e7d6057cded67c03969eb5a6b4 into ac5c0a5760cff588b9c2569d317c9afb7f92543c - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-77f47b18f5567e7f8d0b15982130ecf4bb7d32a4)

**new alerts:**

---

dev-tools/harness.js
scheibo: FWIW, The `race` really does appear to be required to catch all errors properly, see https://gist.github.com/scheibo/e044fe2e99b08518ad30b8ab992f24d7 - in both cases we discover the error, but only in the case with `race` can we actually log properly (and continue running afterwards, to potentially find more bugs!). I'm OK with leaving the `race` off to get this submitted, because it will still be useful/identify issues, but having actionable output in the cases where its an AI bug is pretty useful.

Slayer95: Oh? Good to know. So far, the bug seems to have been accepted, but without further comments.

Slayer95:
> why does LGTM not respect that?

Because LGTM is not ESLint ;)
You'd need to add ``// lgtm`` at the end of the line

scheibo:
> Because LGTM is not ESLint ;)

I should have rephrased as "I think LGTM should understand ESLint directives given its widely used and having to ignore the same violation two different ways is obnoxious"

dev-tools/harness.js
```diff
@@ -0,0 +1,247 @@
...
+		}
+		process.exit(code + i);
+	}
+
+	register() {
+		process.on('unhandledRejection', (r, p) => RejectionTracker.onUnhandledRejection(r, p));
+		process.on('rejectionHandled', p => RejectionTracker.onRejectionHandled(p));
+		process.on('exit', c => RejectionTracker.onExit(c));
+	}
+}();
```
Slayer95: Given that the ``register()`` method refers to a unique object, it makes more sense for ``RejectionTracker`` to be a singleton:

```js
const RejectionTracker = {
  unhandled: [],
  ...
};

Comment (scheibo) [reply dev-tools/harness.js]: Good point.

**Comment by Zarel** [comment]
This pull request **introduces 2 alerts** when merging f393bc7fb36a8b71b863c934071a76f13d0a8719 into 2978546982fcad839afe3de4c7c7a12da4924198 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-bfa4d0be3b5f215259ff500c7461ab1a7b0359a0)

**new alerts:**

---

**Comment by Slayer95** [comment]
It no longer reports the error locations, but it still counts them. LGTM

**Comment by scheibo** [comment]
What does LGTM buy us that Travis running out linter and tests doesn't?

**Comment by Zarel** [comment]
This pull request **introduces 2 alerts** when merging 3669dc19ef88eb42ee93838ca8f50c83cfb138e8 into a53d6a8af762e6bc98382ac8f090c92ef2a57d82 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-8c819cd87cdf555aab88aae4a2f568f6e4287b8b)

**new alerts:**

---

**Comment by Slayer95** [comment]
Who knows !

**Comment by scheibo** [comment]
OMFG ITS STILL COMPLAINING. (ノಠ益ಠ)ノ彡┻━┻

**Comment by Zarel** [comment]
This pull request **introduces 2 alerts** when merging f242844b4fb1abc8000d00dcc2cf2b18ac966eb1 into 0f8241045b5829f8b1e2532d369d9cc614e16ab5 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-558bf0473761bcbe0b8d9b43f74c4b9572b03af1)

**new alerts:**

---

File: dev-tools/harness.js [hunk 8]
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return false;
+	}
+}
+
+// Tracks whether some promises threw errors that weren't caught so we can log
+// and exit with a non-zero status to fail any tests. This "shouldn't happen"
+// because we're "great at propagating promises (TM)", but better safe than sorry.
+const RejectionTracker = {
```
Zarel: Hm, you recently switched this from `new class {` to `{`. Want to talk about your reasoning?

I'm still undecided which is better. I'm leaning towards `new class {` because it feels more natural in TypeScript.

Zarel: There's also `class` with literally everything `static`. Which is also an option I'm undecided about.

scheibo: https://github.com/Zarel/Pokemon-Showdown/pull/5370#discussion_r271503441 is my reasoning.

scheibo: I'm also happy to follow up and bikeshed this to look however you want, but I have several PRs blocked on this landing so it would be awesome if we could merge first and argue over the insignificant bits later :)

Zarel:
@scheibo, LGTM is a static code analyzer, it finds flaws that our linter and tests can't. I've had it dig up a few bugs that even TSLint isn't capable of seeing, bugs involving variables being guaranteed to be a specific value at a specific point.

After repeated complaints I thought I got them to support ESLint syntax for unused variables, though?

Zarel:
Also, the "introduces 2 alerts" message is counting suppressed alerts. They won't show up in the main LGTM page; it's just here so you can't slip bad code past me, I think.

PR #5372, opened 2019-03-29
Fix duration of slot conditions

MacChaeger:
- Includes bonus Perish Song test

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -803,9 +803,9 @@ export class Battle extends Dex.ModdedDex {
 			this.resolveLastPriority(handlers, callbackName);
 		}
 		const side = pokemon.side;
-		for (const j in side.slotConditions[pokemon.position]) {
-			const slotConditionData = side.slotConditions[pokemon.position][j];
-			const slotCondition = side.getSlotCondition(pokemon, j);
+		for (const conditionid in side.slotConditions[pokemon.position]) {
+			const slotConditionData = side.slotConditions[pokemon.position][conditionid];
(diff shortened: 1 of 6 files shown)


MacChaeger:
This is actually not correct...

MacChaeger:
...well it's good enough for 2 player formats, but I'll need to rework it for 4 player formats.

MacChaeger:
Ok now it should work with 4 players as well.

sim/battle.ts
```diff
@@ -802,6 +802,24 @@ export class Battle extends Dex.ModdedDex {
 			});
 			this.resolveLastPriority(handlers, callbackName);
 		}
+		const side = pokemon.side;
+		for (const j in side.slotConditions[pokemon.position]) {
```
scheibo: `j` doesn't seem like a good name for this.

sim/side.ts
```diff
@@ -272,6 +272,7 @@ export class Side {
 		const slotConditionData = this.slotConditions[target][status.id] = {
 			id: status.id,
 			target: this,
+			targetPosition: target,
```
scheibo: when is this used?

MacChaeger: Never, but I thought it might be nice to have? I'll remove it.

scheibo: You can keep it if you'd like, I just wanted to make sure I wasn't missing something :)

MacChaeger: Someone can re-add it in the future if for some reason it becomes necessary, but if it does then something is probably wrong with how slot conditions are implemented.

PR #5379, opened 2019-03-29
Turn on more tslint checks

scheibo:
These ones we already mostly follow already, as evidenced by the fact that there are no code changes to go along with this config change.

Others I find interesting, but which cause code changes:

- `prefer-method-signature`: has a fixer, makes it so theres only one way to do things and doesn't involve too many code changes
- `only-arrow-functions`: more succint, better binding semantics, why not?
- `prefer-template`: this is something our string guidelines dictate, but there are a whole bunch of fixes required and no automatic fixer, so out of scope for now.
- `return-undefined`: i like this for the explicitness it provides, but it doesn't play perfectly with eg. the `void | Promise<void>` pattern
- ~`no-unbound-method`: currently there are a number of places where we fail (eg. https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/battle.ts#L773), but since our code works I feel they could be false positives? It also seems to complain about static methods, so I'm not sure how useful it is.~


tslint.json
```diff
@@ -11,7 +11,18 @@
         "member-access": [true, "no-public"],
         "no-empty": [true, "allow-empty-catch", "allow-empty-functions"],
         "no-string-literal": false,
-        "interface-over-type-literal": false,
+        "interface-over-type-literal": true,
+        "arrow-return-shorthand": true,
+        "no-angle-bracket-type-assertion": true,
+        "no-redundant-jsdoc": true,
+        "no-unnecessary-initializer": true,
```
Zarel: I really don't like this one (and also, using the ESLint version has caused bugs in the past). I think an initializer should be used to mean "this might be read before it's written to".

tslint.json
```diff
@@ -11,7 +11,18 @@
...
+        "no-unsafe-finally": true,
+        "object-literal-shorthand": true,
```
Zarel: I'm surprised we don't currently use this.

tslint.json
```diff
@@ -11,7 +11,18 @@
...
+        "prefer-for-of": true,
+        "prefer-readonly": true,
```
Zarel: In practice this won't come up much because we don't explicitly mark things as private, but I still don't know how much I agree with this. Like `prefer-const`, I think there's a difference between "this isn't currently modified by anything" and "this shouldn't be modified".

tslint.json
Zarel: Incidentally, LGTM will already catch a truly unnecessary initializer, based on static analysis.

tslint.json
Zarel: I, um, blame what I typed there on being sick. What I actually meant to say was:

I'm surprised we don't currently disable this check. I remember being frustrated with it (in either ESLint or TSLint) somewhat recently.

Zarel: I'm fine with enabling this, I guess.

PR #5382, opened 2019-03-30
Make sure BattleStream only calls _destroy once

scheibo:
Fixes `Error: Push after end of read stream` from https://github.com/Zarel/Pokemon-Showdown/pull/5370#issuecomment-478203045.

`this.push(null)` triggers `this._end()` which calls `this._destroy()` already.

This was fine before (though redundant), but no longer `null`-ing out `this.battle` (either by default or through `{retainBattle: true}`) causes the double `_destroy` to become a problem because the `if (this.battle) this.battle.destroy()` check no longer works, and `Battle#destroy()` is not meant to be called more than once.

### DIFF
--- a/sim/battle-stream.ts
+++ b/sim/battle-stream.ts
@@ -39,8 +39,8 @@ function splitFirst(str: string, delimiter: string, limit: number = 1) {
 }
 
 export class BattleStream extends Streams.ObjectReadWriteStream<string> {
-	readonly debug: boolean;
-	readonly keepAlive: boolean;
+	debug: boolean;
+	keepAlive: boolean;
 	battle: Battle | null;
 
 	constructor(options: {debug?: boolean, keepAlive?: boolean} = {}) {
@@ -92,10 +92,7 @@ export class BattleStream extends Streams.ObjectReadWriteStream<string> {
 			options.send = (t: string, data: any) => {
 				if (Array.isArray(data)) data = data.join("\n");
 				this.push(`${t}\n${data}`);
-				if (t === 'end' && !this.keepAlive) {
-					this.push(null);
-					this._destroy();
-				}
+				if (t === 'end' && !this.keepAlive) this.push(null);
 			};
 			if (this.debug) options.debug = true;
 			this.battle = new Battle(options);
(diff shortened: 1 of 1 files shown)


sim/battle-stream.ts
```diff
@@ -39,8 +39,8 @@ function splitFirst(str: string, delimiter: string, limit: number = 1) {
 }
 
 export class BattleStream extends Streams.ObjectReadWriteStream<string> {
-	readonly debug: boolean;
-	readonly keepAlive: boolean;
+	debug: boolean;
+	keepAlive: boolean;
```
Zarel: errr

scheibo: You had wanted this in the previous change?

Zarel:
I guess this turned out to be more complicated than you expected?

scheibo:
> I guess this turned out to be more complicated than you expected?

OK, so whats confusing me is `this._end` also does a `this.push(null)`, which scared me and made me think my reasoning about `this.push(null)` triggering `this._end()` was wrong, but I looked it over and that still appears to be the case.

sim/battle-stream.ts
Zarel: Yeah but I thought I merged it in the previous change.

scheibo: https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/battle-stream.ts#L42 still there on HEAD

sim/battle-stream.ts
```diff
@@ -164,10 +161,7 @@ export class BattleStream extends Streams.ObjectReadWriteStream<string> {
 		this._destroy();
 	}
 	_destroy() {
-		if (this.battle) {
-			this.battle.destroy();
-		}
-		this.battle = null;
```
Zarel: Wasn't this line also removed? Am I thinking of a different stream or did this never get merged or what?

scheibo: You're thinking of c45ff0e4bcd87d0101ea97b046551767d41e1c84 on #5370 which has not been merged into HEAD yet. I just redid those changes on this PR because its not like they were specific to #5370, they just were discovered there, and landing them earlier seems useful given all my PRs involving the harness seem to get bogged down :)

PR #5385, opened 2019-03-30
Helpticket: Use for-of instead of for-in in arrays

TheJetOU:
It's more readable. Although it's a bit slower (https://jsperf.com/for-in-vs-for-of-in-arrays), I'd take 3% of less speed for readability in a place where speed isn't _that_ important

### DIFF
--- a/server/chat-plugins/helptickets.js
+++ b/server/chat-plugins/helptickets.js
@@ -1033,8 +1033,8 @@ let commands = {
 				this.privateModAction(displayMessage);
 			}
 
-			for (let i in affected) {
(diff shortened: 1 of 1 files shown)


Slayer95:
``console.log()`` is too slow itself, so it renders your benchmark ineffective. Please check https://jsperf.com/foreach-vs-for-loop-vs-for-in-vs-for-of-vs-babel-for-of instead.

Zarel:
Yeah, `CONTRIBUTING.md` already talks about how the best way to iterate arrays is.

PR #5387, opened 2019-03-30
Sim: Z Moves may only be disabled when all base moves are

Slayer95:


Marty-D:
This prevents the battle from progressing if all your moves are disabled (from Taunt or whatever combination of effects) and you could have otherwise used Z-Power. Clicking Struggle at that point (whether or not you clicked Z-Power first) causes this:

```
[Invalid choice] Can't move: Your Chansey doesn't have a move 0
[Invalid choice] Can't make choices: You sent choices for 2 Pokémon, but this is a singles game!
[Invalid choice] Can't make choices: You sent choices for 3 Pokémon, but this is a singles game!
[Invalid choice] Can't make choices: You sent choices for 4 Pokémon, but this is a singles game!
```
etc.

What the game really does is if all your moves are disabled, clicking Fight skips straight to Struggle because you wouldn't have been allowed to click any of your moves, and can't access the Z-Power menu from there.

Slayer95:
! So that's how it is

Slayer95:
Updated accordingly. Please take a look, @Marty-D

Marty-D:
That part looks good now, but there seems to be a regression somewhere.

**Turn 5**
Chansey used Disable!
The opposing Chansey's Sleep Talk was disabled!
The opposing Chansey used Disable!
Chansey's Disable was disabled!
-
**Turn 6**
The opposing Chansey can't use Sleep Talk!
Chansey can't use Disable!
-
Turn 6 was successfully clicking Z-Power -> Z-Sleep Talk and Z-Power -> Z-Disable, respectively.

Slayer95:
Oh? Interesting...

So I am checking the instances of disabling, and assuming the following:
- Attract *can* prevent Z Moves.
- Dazzling / Queenly Majesty and Truant *can* prevent Z Moves.
- Disable, Gravity, Heal Block, Imprison. Taunt, Throat Chop *cannot* prevent Z Moves.

Marty-D:
That's a mix of move selection and move execution prevention though.
Attract and Truant happen before execution, while every other case you mentioned allows the Z-Move to be selected and executed, except Gravity only allows selection. Once executed, it may still be blocked by priority-blocking like Dazzling.

Slayer95:
> except Gravity only allows selection

That should be the case both for Z and regular moves, right? Then we should also remove the ``onDisableMove`` handler from Gravity.

Marty-D:
> 
> 
> > except Gravity only allows selection
> 
> That should be the case both for Z and regular moves, right? Then we should also remove the `onDisableMove` handler from Gravity.

There are currently no Z-Move "moves" that have the Gravity flag. But even if there were and they were in your actual moveset they would be disabled, like Clangorous Soulblaze is disabled during Throat Chop or Extreme Evoboost is disabled during Taunt.

Slayer95:
> while every other case you mentioned allows the Z-Move to be selected and executed

> like Clangorous Soulblaze is disabled during Throat Chop or Extreme Evoboost is disabled during Taunt.

Hm?

> regular moves, right?

I am not sure you really answered my question. I mean, Fly should be selectable during Gravity, but fail on usage, right?

Anyway, I updated this to the best of my current understanding...

Slayer95:
> Taunt
       should not prevent the target from using Z-Powered Status moves:

Is that test wrong?

Marty-D:
>while every other case you mentioned allows the Z-Move to be selected and executed

This was re: Z-Moves used from the Z-Power menu, not the regular move selection menu.

I believe this PR was good to go before your last change :(

Marty-D:
Fly is not selectable during Gravity, no.

Slayer95:
Reverted. So confusing!

Marty-D:
Perfect, thanks! Catch me on PS if you want to clarify anything.

scheibo:
Thanks @Slayer95 and @Marty-D for fixing my mistake.

Zarel:
> `Struggle` shouldn't have been shown. Looks like a client issue?

PS always displays the forced move button for forced moves, instead of automatically progressing the battle like the games do. Mostly for historical reasons, although I feel like giving users a button to press makes them feel better.

PR #5389, opened 2019-03-31
Fix Weakness Policy and Clangorous Soulblaze in doubles

MacChaeger:
Such a simple title for such a massive change...

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -131,7 +131,7 @@ let BattleAbilities = {
 		shortDesc: "If this Pokemon (not its substitute) takes a critical hit, its Attack is raised 12 stages.",
 		onHit(target, source, move) {
 			if (!target.hp) return;
-			if (move && move.effectType === 'Move' && this.getHitData(target).crit) {
+			if (move && move.effectType === 'Move' && move.getHitData(target).crit) {
 				target.setBoost({atk: 6});
 				this.add('-setboost', target, 'atk', 12, '[from] ability: Anger Point');
 			}
@@ -891,8 +891,8 @@ let BattleAbilities = {
 	},
 	"filter": {
 		shortDesc: "This Pokemon receives 3/4 damage from supereffective attacks.",
-		onSourceModifyDamage(damage, source, target) {
-			if (this.getHitData(target).typeMod > 0) {
+		onSourceModifyDamage(damage, source, target, move) {
+			if (move.getHitData(target).typeMod > 0) {
 				this.debug('Filter neutralize');
 				return this.chainModify(0.75);
 			}
@@ -2141,7 +2141,7 @@ let BattleAbilities = {
 	"neuroforce": {
 		shortDesc: "This Pokemon's attacks that are super effective against the target do 1.25x damage.",
 		onModifyDamage(damage, source, target, move) {
-			if (move && this.getHitData(target).typeMod > 0) {
+			if (move && move.getHitData(target).typeMod > 0) {
 				return this.chainModify([0x1400, 0x1000]);
 			}
 		},
@@ -2522,8 +2522,8 @@ let BattleAbilities = {
 	"prismarmor": {
 		desc: "This Pokemon receives 3/4 damage from supereffective attacks. Moongeist Beam, Sunsteel Strike, and the Mold Breaker, Teravolt, and Turboblaze Abilities cannot ignore this Ability.",
 		shortDesc: "This Pokemon receives 3/4 damage from supereffective attacks.",
-		onSourceModifyDamage(damage, source, target) {
-			if (this.getHitData(target).typeMod > 0) {
+		onSourceModifyDamage(damage, source, target, move) {
+			if (move.getHitData(target).typeMod > 0) {
 				this.debug('Prism Armor neutralize');
 				return this.chainModify(0.75);
 			}
@@ -3110,8 +3110,8 @@ let BattleAbilities = {
 	},
 	"sniper": {
 		shortDesc: "If this Pokemon strikes with a critical hit, the damage is multiplied by 1.5.",
-		onModifyDamage(damage, source, target) {
-			if (this.getHitData(target).crit) {
+		onModifyDamage(damage, source, target, move) {
+			if (move.getHitData(target).crit) {
 				this.debug('Sniper boost');
 				return this.chainModify(1.5);
 			}
@@ -3170,8 +3170,8 @@ let BattleAbilities = {
 	},
 	"solidrock": {
 		shortDesc: "This Pokemon receives 3/4 damage from supereffective attacks.",
-		onSourceModifyDamage(damage, source, target) {
-			if (this.getHitData(target).typeMod > 0) {
+		onSourceModifyDamage(damage, source, target, move) {
+			if (move.getHitData(target).typeMod > 0) {
 				this.debug('Solid Rock neutralize');
 				return this.chainModify(0.75);
 			}
@@ -3643,8 +3643,8 @@ let BattleAbilities = {
 	},
 	"tintedlens": {
 		shortDesc: "This Pokemon's attacks that are not very effective on a target deal double damage.",
-		onModifyDamage(damage, source, target) {
-			if (this.getHitData(target).typeMod < 0) {
+		onModifyDamage(damage, source, target, move) {
+			if (move.getHitData(target).typeMod < 0) {
 				this.debug('Tinted Lens boost');
 				return this.chainModify(2);
 			}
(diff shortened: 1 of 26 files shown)


sim/battle.ts
```diff
@@ -19,6 +19,11 @@ interface FaintedPokemon {
 	effect: Effect | null;
 }
 
+interface MoveHitData {
+	crit: {[key: string]: boolean};
```
Zarel: `key` is really vague, and appears to be `": "` plus the Pokémon nickname, which is probably unintentional.

data/abilities.js
```diff
@@ -131,7 +131,7 @@ let BattleAbilities = {
 		shortDesc: "If this Pokemon (not its substitute) takes a critical hit, its Attack is raised 12 stages.",
 		onHit(target, source, move) {
 			if (!target.hp) return;
-			if (move && move.effectType === 'Move' && move.crit) {
+			if (move && move.effectType === 'Move' && this.getHitData(target).crit) {
```
Zarel: This always seemed really awkward to me.

Then it hit me: Why _isn't_ this a property of `move`? That seems like a better place for hit data than even more temporary data in `Battle`.

Zarel: For that matter, I wonder if it makes more sense to have a different `moveCopy` object for each target.

Most move state isn't shared in spread moves, right?

Zarel: Actually, "different ActiveMove for each target" sounds like a significantly nicer solution.

sim/battle.ts
MacChaeger: Oh yeah it should be `.slice(0, 3`

MacChaeger:
When @Slayer95 and I were thinking of how to go about doing this, he suggested that, but @scheibo thought it would be bad for performance. The real problem with that idea, however, lies in the fact that we have to run the `'AfterDamage'` event simultaneously on multiple Pokémon at once, which means that this solution would require passing an array of moves to numerous functions, most notably `runEvent`.

data/abilities.js
MacChaeger: Oops, I meant to reply to this.

MacChaeger: Also, most move-state _is_ shared between targets. Literally the only things not shared are `crit` and `typeMod`.

Zarel: Accuracy, base power, etc isn't shared either, right?

Zarel: What move state _is_ shared?

sim/battle.ts
MacChaeger: What would you call “p1a”, “p2b”, etc.? “targetingid”?

data/abilities.js
MacChaeger: We don't store either of those. We calculate them once per target, then throw them away because they aren't needed any more. What we do store is the following, and all of these are shared per target (this excludes things only relevant to single target moves, such as `ActiveMove#allies` which only matters for Beat Up):
- `ability`
- `aerilateBoosted`
- `auraBooster`
- `galvanizeBoosted`
- `hasAuraBreak`
- `hasBounced`
- `hasSheerForce`
- `isExternal`
- `magnitude`
- `normalizeBoosted`
- `pixilateBoosted`
- `pranksterBoosted`
- `refrigerteBoosted`
- `selfDropped`
- `spreadHit` (obviously)
- `stab`

sim/battle.ts
Zarel: `slotid`?

Zarel: Also I wouldn't do this at all. :[

Zarel:
Okay, that's fair, but your `moveHitData` table should still be on the move, and not as global battle state.

MacChaeger:
What about the functions that are used to access and set the data easily? Should those be on the `ActiveMove` object as well?

data/abilities.js
MacChaeger: Oh, `zBrokeProtect` is another thing that's not shared between targets, so I'll add that to this PR.

Zarel:
`Pokemon.prototype` might not be a bad place for the accessors tbh.

Zarel:
I guess `ActiveMove` isn't a _bad_ place for it

sim/dex-data.ts
```diff
@@ -859,6 +854,73 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 	}
 }
 
+export class ActiveMove extends Move {
```
Zarel: Hm I don't think I like the idea of `ActiveMove` being its own class. Especially since it doesn't actually extend `Move`, it actually makes move properties non-readonly.

Zarel:
Blah, none of this is bad enough to justify delaying a major bugfix.

sim/dex-data.ts
```diff
@@ -859,6 +854,73 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
+	crit(target: Pokemon) {
+		this.moveHitData.crit[target.toString().slice(0, 3)] = true;
+	}
+
+	setTypeModFor(target: Pokemon, typeMod: number) {
+		this.moveHitData.typeMod[target.toString().slice(0, 3)] = typeMod;
+	}
+
+	zBreakProtect(target: Pokemon) {
+
```
urkerab: Something missing?

data/mods/gen2/scripts.js
```diff
@@ -588,7 +588,7 @@ let BattleScripts = {
 		let noburndrop = false;
 
 		// The move is a critical hit. Several things happen here.
-		if (move.crit) {
+		if (isCrit) {
```
Slayer95: ``this.runEvent('CriticalHit', target, null, move)`` should modify ``isCrit`` so that this check works correctly.

data/statuses.js
```diff
@@ -398,7 +398,7 @@ let BattleStatuses = {
 			}
 			const hitMove = new this.Data.Move(data.moveData);
 
-			this.tryMoveHit(target, data.source, /** @type {ActiveMove} */(hitMove));
+			this.tryMoveHit(target, data.source, /** @type {ActiveMove} */(/** @type {unknown} */(hitMove)));
```
scheibo: This is wrong, or at the very least misleading. Because you've made `ActiveMove` a class you need to correct the callsites to actually have an `ActiveMove` object, not a cast `Move` (or arbitrary object). 

```
$ git grep "@type {ActiveMove}"
data/mods/gen1/moves.js:                                  this.moveHit(target, pokemon, move, /** @type {ActiveMove} */ ({damage: this.effectData.totalDamage * 2}));
data/mods/gen1/scripts.js:                        move = /** @type {ActiveMove} */ ({
data/mods/gen2/moves.js:                                  /** @type {ActiveMove} */
data/mods/gen2/scripts.js:                        move = /** @type {ActiveMove} */ ({
data/mods/gen2/statuses.js:                       move = /** @type {ActiveMove} */ ({
data/mods/gen3/moves.js:                                  /** @type {ActiveMove} */
data/mods/gen3/moves.js:                  let moveData = /** @type {ActiveMove} */ ({
data/mods/gen4/moves.js:                                  let moveData = /** @type {ActiveMove} */ ({
data/mods/gen4/moves.js:                  let moveData = /** @type {ActiveMove} */ ({
data/mods/gen4/moves.js:                  let moveData = /** @type {ActiveMove} */ ({
data/mods/gen6/statuses.js:                       this.damage(damage, pokemon, pokemon, /** @type {ActiveMove} */ ({
data/mods/gennext/moves.js:                                       let moveData = /** @type {ActiveMove} */ ({
data/mods/stadium/scripts.js:                     move = /** @type {ActiveMove} */ ({
data/statuses.js:                 this.damage(damage, pokemon, pokemon, /** @type {ActiveMove} */ ({
data/statuses.js:                 this.tryMoveHit(target, data.source, /** @type {ActiveMove} */(/** @type {unknown} */(hitMove)));
```

Are all places where we are casting which may or may not be OK, but in certain places it definitely isn't: 

```
A battle crashed with this stack trace:
TypeError: move.getHitData is not a function
    at Battle.modifyDamage (/home/ps/showdown/data/mods/gen4/scripts.js:44:23)
    at Battle.getDamage (/home/ps/showdown/.sim-dist/battle.js:2047:15)
    at Battle.onTry (/home/ps/showdown/data/mods/gen4/moves.js:700:22)
    at Battle.singleEvent (/home/ps/showdown/.sim-dist/battle.js:416:25)
    at Battle.tryMoveHit (/home/ps/showdown/data/mods/gen3/scripts.js:182:13)
    at Battle.useMoveInner (/home/ps/showdown/data/mods/gen3/scripts.js:147:18)
    at Battle.useMove (/home/ps/showdown/data/scripts.js:138:25)
    at Battle.runMove (/home/ps/showdown/data/scripts.js:93:31)
    at Battle.runAction (/home/ps/showdown/.sim-dist/battle.js:2557:9)
    at Battle.go (/home/ps/showdown/.sim-dist/battle.js:2777:9)
```

PR #5397, opened 2019-04-02
Battle Factory: Add tests for EVs, learnsets, and natures

KrisXV:


KrisXV:
@zarel these are the same timeout errors ^

test/simulator/misc/random-teams.js
```diff
@@ -128,13 +138,27 @@ describe(`Factory sets`, function () {
 							assert(ability.name === abilityName, `miscapitalized ability "${abilityName}" of ${species}`);
 						}
 
+						for (const natureName of [].concat(set.nature)) {
```
scheibo: what is going on here? it would be better as

```javascript
for (const natureName of [set.nature]) { ... }
```

But you don't even need the loop at all if theres only one element! (and if I'm missing something and `set.nature` is already an array, just loop over that instead of the `concat`).

test/simulator/misc/random-teams.js
```diff
@@ -26,6 +27,14 @@ function isValidSet(gen, set) {
 	return true;
 }
 
+function validLearnset(move, set, tier) {
+	const validator = TeamValidator(`gen7${tier}`);
+	const template = validator.dex.getTemplate(set.species || set.name);
+	let lsetProblem = validator.checkLearnset(move, template);
+	if (lsetProblem) return false;
```
scheibo: Returning `false` and `true` right beside each other usually points to an issue. Just 
```javascript
return !validator.checkLearnset(move, template);
```
on line 33 and cut out everything after.

test/simulator/misc/random-teams.js
```diff
@@ -128,13 +138,27 @@ describe(`Factory sets`, function () {
...
 						}
 
+						for (const natureName of [].concat(set.nature)) {
+							const nature = Dex.getNature(natureName);
+							assert(nature.exists, `invalid nature "${natureName}" of ${species}`);
+							assert(nature.name === natureName, `miscapitalized nature "${natureName}" of ${species}`);
+						}
+
 						for (const moveSpec of set.moves) {
 							for (const moveName of [].concat(moveSpec)) {
```
scheibo: same odd `[].concat` pattern - what am I missing?

test/simulator/misc/random-teams.js
KrisXV: i have zero clue why i didnt think of that

test/simulator/misc/random-teams.js
KrisXV: set.nature can be an array or a string, and i just did this because it was consistent with the other loops

scheibo: I see. I think that they can be written more cleanly as `[...set.nature]` then :)

test/simulator/misc/random-teams.js
KrisXV: I wish I had an answer for this, but I didn't write it, sorry

test/simulator/misc/random-teams.js
```diff
@@ -26,6 +27,12 @@ function isValidSet(gen, set) {
...
 
+function validLearnset(move, set, tier) {
+	const validator = TeamValidator(`gen7${tier}`);
+	const template = validator.dex.getTemplate(set.species || set.name);
+	return !validator.checkLearnset(move, template);
+}
+
 describe(`Random Team generator`, function () {
 	for (const gen of ALL_GENS) {
 		it(`should successfully create valid Gen ${gen} teams`, function () {
```
Slayer95: This makes for ludicrously long tests. Add ``this.timeout(0)`` below.

test/simulator/misc/random-teams.js
Slayer95: That would result in Nature Jolly becoming ``['J', 'o', 'l', 'l', 'y']``.
``Array.prototype.concat()`` is special in that it accepts as arguments single elements to be pushed, but it also flattens arrays with ``Symbol.isConcatSpreadable``.

scheibo: TIL, thanks for the explanation.

test/simulator/misc/random-teams.js
scheibo: I mean.... we don't really want to have a test suite with ludicrously long tests either though....

Slayer95: Then the test can be conditionally run depending on the arguments passed to ``mocha``.

Personally, I like to unlock long tests by passing ``-t NUM``, which also sets the timeout for tests. We could have that as ``npm run fulltest``

scheibo: having additional options for long tests seems like it would be used (selfishly thinking of my smoke test use case)

test/simulator/misc/random-teams.js
Zarel: Considering how obscure this is, it should probably be commented somewhere.

test/simulator/misc/random-teams.js
KrisXV: In doing this, other tests cause to fail, such as Intimidate in gen 4, and endless battle clause.

Zarel:
I think the main thing this needs is some sort of flag to opt into running the full test.

Slayer95:
> I think the main thing this needs is some sort of flag to opt into running the full test.

Did. Run with ``npx mocha -t [timeout]``

scheibo:
> Did. Run with `npx mocha -t [timeout]`. Faster if ran also with `-g Factory`

Can we update Travis to then run with `-t` to trigger these tests then? Our CI should be testing these IMO, even if we dont want every developer to be forced to?

Slayer95:
New ``mocha`` release and this doesn't seem to work anymore? Could I be that unlucky?

Slayer95:
Not sure what happened :s

Anyway, reworked to a declarative format, and implemented ``npm run fulltest``.

PR #5400, opened 2019-04-03
Ban Trapinch from LC

Merritty:
https://www.smogon.com/forums/threads/usum-lc-trapinch-suspect-voting.3648825/post-8086522


KrisXV:
You also have to ban it in formats.js in the off chance it moves out of LC Uber

PR #5402, opened 2019-04-03
Random Battle Gen 3 and 4: Fix Togetic

ascriptmaster:
It does not get Super Luck in this generation, so it should always get
Serene Grace, but it keeps getting rejected.

### DIFF
--- a/data/mods/gen3/random-teams.js
+++ b/data/mods/gen3/random-teams.js
@@ -435,68 +435,77 @@ class RandomGen3Teams extends RandomGen4Teams {
 			ivs = {hp: 31, atk: 31, def: 31, spa: 31, spd: 31, spe: 31};
 		}
 
-		let abilities = Object.values(baseTemplate.abilities);
+		let abilities = Object.values(baseTemplate.abilities).filter(a => this.getAbility(a).gen === 3);
 		abilities.sort((a, b) => this.getAbility(b).rating - this.getAbility(a).rating);
 		let ability0 = this.getAbility(abilities[0]);
 		let ability1 = this.getAbility(abilities[1]);
-		ability = ability0.name;
-		if (ability0.gen !== 3) ability = ability1.name;
-		if (ability0.gen === 3 && ability1.gen === 3) {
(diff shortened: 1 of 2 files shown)


TheImmortal:
Uh, instead of hardcoding, shouldn't Hustle be rejected as well? Then it selects the highest rated ability gets picked (doesn't look like I copied that over from future gens but I should) which is Serene Grace.

ascriptmaster:
Ability selection is mostly the same as the current gen, where it selects the highest rated ability (33% chance of choosing the second highest if it's within a 0.5 point difference), then takes the second choice if the first choice is rejected. 

Serene Grace is rated 3, and Hustle is 2, so it will always choose Serene Grace first, and thus the check needs to be in Serene Grace, and not Hustle.

TheImmortal:
No it used to be like this. Look at future gen ability selection. I updated it some time ago. If all the abilities are rejected, it defaults to the highest rated.

TheImmortal:
bfa0394e938190e756c9111e6df38ff73d847e74

So you can just reject Hustle now, in gen 4 at least.

ascriptmaster:
What about gen 3? That's where the bug was originally reported.

TheImmortal:
Gen 3 had these weird lines: https://github.com/Zarel/Pokemon-Showdown/blob/master/data/mods/gen3/random-teams.js#L443

I wanted to look into it and see if it's necessary.

ascriptmaster:
I can fix the gen 3 issues. Give me a second.

ascriptmaster:
Done. Hustle would've actually been rejected on Togetic with the current system anyways, since Seismic Toss doesn't increment the Hustle counter, so the only thing that needed to be done was the ability update.

ascriptmaster:
never mind i broke something with that change, somehow

ascriptmaster:
ah I forgot the else statement at the end of the ability selection code

ascriptmaster:
Apparently removing the gen 3 ability check breaks the set generation by giving mons abilities they didn't have in that gen, so whatever.

Issue #5399, opened 2019-04-02
Invalid choices should abort Side#choose()


scheibo:
In this doubles battle scenario, the `p2` has 1 mon remaining, so sends `'>p2 move 3 2, pass'`, only to get back 2 choice errors: 

```
sideupdate
p2
[...]

Issue #5393, opened 2019-04-01
[Gen 4] Custap Berry + Trump Card  interaction causes a crash


scheibo:
```
 TypeError: Cannot read property 'pp' of null
    at Battle.basePowerCallback (/Pokemon-Showdown/data/moves.js:18263:21)
    at Battle.getDamage (/Pokemon-Showdown/.sim-dist/battle.js:1953:39)
    at Battle.getSpreadDamage (/Pokemon-Showdown/data/scripts.js:806:25)
    at Battle.spreadMoveHit (/Pokemon-Showdown/data/scripts.js:755:17)
    at Battle.hitStepMoveHitLoop (/Pokemon-Showdown/data/scripts.js:653:37)
    at Battle.trySpreadMoveHit (/Pokemon-Showdown/data/scripts.js:338:26)
    at Battle.useMoveInner (/Pokemon-Showdown/data/scripts.js:248:22)
    at Battle.useMove (/Pokemon-Showdown/data/scripts.js:138:25)
    at Battle.runMove (/Pokemon-Showdown/data/scripts.js:93:31)
    at Battle.runAction (/Pokemon-Showdown/.sim-dist/battle.js:2552:9)
```

See [input log](https://gist.githubusercontent.com/scheibo/84a7cc5ab30ac1e3723d434282abd7e7/raw/c38859c56c213d65467432e50b60cd5dcb628874/error.log) to repro.

*Yay! My smoke test works!*

Slayer95:
Looks like Custap Berry should be removed from the possible ``sourceEffect`` values of moves, just like Instruct is in ``useMoveInner()``

Issue #5270, opened 2019-03-07
Serializable simulator state


yuzeh:
Now that simulator code is pretty well separated from server code, the idea of serializing and reloading a game in the middle seems way more achievable.

My vision for what this looks like is adding two more operations to the simulator protocol (in addition to `start`, `player`, `p1`, `p2`):
* `dumpstate` -- the process would write a JSON blob to stdout with the current state of the game
* `loadstate <JSON Blob>` -- the process would overwrite its current state with the JSON blob.

Two main use cases come to mind:
* May be useful for debugging and unit testing
* Allows for efficient forward modeling, making methods like Monte Carlo Tree Search (or some partially observed variant of it) possible

Let me know if you'd accept a PR that implements this.

scheibo:
OMG! I am on a mission to do the exact same thing (for MCTS). I wasn't going to add `dumpstate` or `loadstate` though, because the server's true state is != the client's observed/inferred state. I don't know about debugging, but I think it would be relatively useful for recreating games in tournament settings that got screwed up do the DCing etc.

Zarel:
We already have a solution for recreating games – inputlogs are now tracked, and tournament directors know about the `/importinputlog` and `/exportinputlog` commands.

The only advantage of dumping state over inputlogs is performance, since state can be deterministically recreated from inputlogs.

I don't have a problem with supporting serializing simulator state if you want to implement it, though.

scheibo:
@yuzeh - I'm probably still a month or so away from being ready to implement it, but it seems like the PR would be accepted if you wrote it ahead of that. I'd be happy to review or offer support if you wanted to take it on.

yuzeh:
`loadstate` would be excellent for loading a particular game state without having to read the whole inputlog up until that point. `loadstate` would also be good for loading hypothetical states (i.e. ones not necessarily generated by `savestate`) and potentially invalid states as well.

@scheibo great to hear that you're looking into MTCS for Pokemon battling! I'm using RL (without any forward search) right now and seeing promising results in gen7randombattle. That said, I won't have very much time to implement this in the next two months...

scheibo:
Alright, well then I might get to it first, but Ill loop you in to see that my solution works for you if I do. :)

[...]

PR #5405, opened 2019-04-03
Add a smoke test runner to dev-tools

scheibo:
This was fun to write, but given how its significantly more complex than I'd wish I wouldn't be surprised if there were bugs (but it doesn't need to be perfect to be useful, it's stochastic by design).

I've been running this in 'infinite' mode on one of my cores for 24hrs+ (and found #5393 and #5399). The idea is that before restarting the server, admins should run `node dev-tools/smoke 5` or whatever and wait maybe 5 minutes as insurance against obvious crashes that would force a restart again anyway.


Zarel:
Unfortunately, I don't think this will catch most bugs. Very few bugs are actually crashes; they're usually just incorrect mechanics.

Actual integration tests would probably also be nice. Maybe a `./pokemon-showdown --no-fs-writing` mode, which also listens on a nonstandard port, so it can't screw with an actively-running server.

dev-tools/smoke.js
```diff
@@ -0,0 +1,433 @@
...
+ *
+ * @license MIT
+ */
+
+'use strict';
+
+const child_process = require('child_process');
+const shell = cmd => child_process.execSync(cmd, {stdio: 'inherit', cwd: __dirname});
+
+// Run the build script if we're being called from the command line.
```
Zarel: This would probably benefit from documentation for calling it from the command line.

test/dev-tools/smoke.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const assert = require('assert');
+
+const {SmokeRunner} = require('./../../dev-tools/smoke');
+
+describe('SmokeRunner', async () => {
+	it('should run successfully', async () => {
+		const opts = {format: 'gen7doublescustomgame', maxGames: 1, prng: [1, 2, 3, 4]};
+		assert.strictEqual(await (new SmokeRunner(opts).run()), 0);
```
Zarel: How much time does this take?

scheibo: Running a single battle? 100ms.

Zarel: On _your_ computer. Poor Kris, on the other hand...

scheibo:
> Unfortunately, I don't think this will catch most bugs. Very few bugs are actually crashes; they're usually just incorrect mechanics.

Correct, this will 'only' catch crashes and protocol issues, like the 2 its already caught. That's more than nothing.

> Actual integration tests would probably also be nice. Maybe a `./pokemon-showdown --no-fs-writing` mode, which also listens on a nonstandard port, so it can't screw with an actively-running server.

This is also on my TODO list, these smoke tests do not preclude more integration tests.

Zarel:
Oh, yeah, I wasn't saying otherwise. That was just a somewhat-related thought.

test/dev-tools/smoke.js
scheibo: Kris is busy adding even slower tests with his team validation :) https://github.com/Zarel/Pokemon-Showdown/pull/5397#discussion_r271467476

WDYT about @Slayer95's idea of having a separate option for slow tests (which Travis or people with servers under their desk to run)? I'd want to run a whole cycle of the smoke tests (~1 minute) instead of just a single battle if we had that.

dev-tools/smoke.js
scheibo: Unlike the harness (which will get even more args once I add benchmarking support...), this one is designed to pretty much just run without args. 5f539a9 adds an example. The negative cycles option is kind of hidden on purpose, and passing in format + seed is for debugging purposes (and the error message will tell you about them when relevant).

test/dev-tools/smoke.js
Zarel: Yeah, I like the idea of having a separate option for slow tests. But we're already stretching Travis CI to its limit, so it'd definitely just be something we'd have to use a different server for.

Also I hear Travis CI is imploding. We should probably look into switching. I hear Azure Pipelines is a thing? Also Circle CI? @Slayer95, can you investigate?

dev-tools/smoke.js
Zarel: I think it'd benefit from a short summary of what a smoke test is (I obviously know, but you know, for those ignorant people who certainly aren't myself). And what a cycle is. And why are negative cycles hidden on purpose? Is that you're "run infinitely" setting? I don't understand why you'd want to hide that.

test/dev-tools/smoke.js
Slayer95: > Also I hear Travis CI is imploding.

That seems to point to last week's outage, which is the only major issue in the last 4 months as per https://www.traviscistatus.com/

They are yet to publish a postmortem about it though.

> But we're already stretching Travis CI to its limit

I have noticed some delay in Travis yea. But is that some actual limit as opposed to the direct effect of our longer build time?

> Azure Pipelines is a thing? Also Circle CI? @Slayer95, can you investigate?

I guess I'll have a look.

dev-tools/smoke.js
```diff
@@ -0,0 +1,435 @@
...
+		'gen2customgame',
+		'gen1customgame'];
+
+	// Because of our handwavy accounting of 'usage', as well as to account for ordering effects and
+	// the interplay of various combinations, we should actually run multiple cycles and exhaust our
+	// pools multiple times to provide more confidence that we're actually smoking out issues.
+	//
+	// If cycles is negative the smoke test will instead run indefinitely in case you wish to
+	// dedicate some computer resources to finding PS! edge case crashes instead of mining crypto,
+	// folding proteins, or searching for extraterrestrials/large primes.
```
Zarel: Does folding@home even matter ever since AlphaFold annihilated everyone else?

dev-tools/smoke.js
```diff
@@ -0,0 +1,435 @@
...
+		let failures = 0;
+		do {
+			for (let format of formats) {
+				failures += await new SmokeRunner({
+					format, cycles, prng, maxFailures, log: true,
+				}).run();
+				process.stdout.write('\n');
+				if (failures >= maxFailures) break;
+			}
+		} while (cycles < 0); // eslint-disable-line no-unmodified-loop-condition
```
Zarel: `repeatForever = cycles < 0; do ... while (repeatForever)`

is probably enough to shut up the linter, and is probably also more readable.

Zarel: Wait, no it's not.

Zarel: Still, it might be the best way to write this code.

```js
if (cycles < 0) {
  repeatForever = true;
  cycles = -cycles;
}
```

dev-tools/smoke.js
scheibo: Ahah reviewing my jokes, I love it!

test/dev-tools/smoke.js
Zarel: > I have noticed some delay in Travis yea. But is that some actual limit as opposed to the direct effect of our longer build time?

Kris pointed out he got a timeout in Travis when he added the extra BF validator.

Slayer95: That should be the default test timeout from ``mocha`` itself...,

dev-tools/smoke.js
scheibo: I tried:

```js
const forever = cycles < 0;
do {
 ...
} while (forever);
```

And I still get:

```
/dev-tools/smoke.js
  429:12  error  'forever' is not modified in this loop  no-unmodified-loop-condition
```

Maybe the `forever` variable makes things more readable though? Iunno, I'm not convinced, because then it looks like the loop might always be an infinite loop at first glance.

dev-tools/smoke.js
scheibo: Done. I had hidden the negative cycles because I didn't know anyone but me would be interested in running infinitely (though if someone looked at the code theyd see it called out in a comment).

test/dev-tools/smoke.js
scheibo: Are we blocking on this? I can disable both `test/dev-tools/{smoke,harness}.js` tests if you're concerned (I don't know if you saw it in the previous PR, because while a single smoke test is slower, theyre both around the same O(100ms) run time?)

scheibo:
>  I don't think this will catch most bugs.

Incidentally, if this were landed and @Marty-D had run this before `/hotpatch formats`, the latest `TypeError: move.getHitData is not a function` crash would have been caught. (caught within 30s of running on my machine) :)

dev-tools/smoke.js
Zarel: I think the code I suggest makes things more readable, yes, even after realizing it wouldn't shut up the linter, which I realized when I said "Wait, no it's not".

dev-tools/smoke.js
```diff
@@ -0,0 +1,452 @@
...
+
+	// Because of our handwavy accounting of 'usage', as well as to account for ordering effects and
+	// the interplay of various combinations, we should actually run multiple cycles and exhaust our
+	// pools multiple times to provide more confidence that we're actually smoking out issues.
+	//
+	// If cycles is negative the smoke test will instead run indefinitely in case you wish to
+	// dedicate some computer resources to finding PS! edge case crashes instead of mining crypto,
+	// folding proteins, or searching for extraterrestrials/large primes.
+	const cycles = Number(process.argv[2]) || DEFAULT_CYCLES;
+	const forever = cycles < 0;
```
Zarel: It would be nice to then make `cycles` positive here, instead of needing that `Math.abs` in a weird spot.

scheibo: Yeah, but I prefer the code this way so I can leave `cycles` `const` as well as be less verbose. :)

Let's argue about this some more instead of merging!

Zarel: See, this is why I don't like `const`, it forces you to write worse code because you start optimizing for this sort of thing. :p

test/dev-tools/smoke.js
Zarel: We're not blocking on anything. I'm not opposed to merging and figuring out the rest later.

dev-tools/smoke.js
Zarel:
```js
	const cyclesSetting = Number(process.argv[2]) || DEFAULT_CYCLES;
	const forever = cyclesSetting < 0;
	const cycles = Math.abs(cyclesSetting);
```

Zarel:
At some point I'd like documentation on what the difference is between the harness and this, because they seem to do the same thing (run battles randomly and report the errors that result).

But you sound antsy. Nothing's blocking this, if you want to merge and figure out the rest later.

dev-tools/smoke.js
scheibo: I mean, obviously I thought of that, I just didn't like it as because `cyclesSetting` seems contrived. You could also pass `cycle: Math.abs(cycle)` to the `SmokeRunner`, but that also kind of makes the code uglier as well IMO because the shortcut notation looks better

I don't know why I'm being stubborn about this because its so trivial, but I like it the way it is. *shrugs*. I wonder if this is just a ['just remove the  duck'](https://rachelbythebay.com/w/2013/06/05/duck/) scenario given how I'm sure the `Pool` logic etc is more complicated and more worth this sort of attention.

scheibo:
> At some point I'd like documentation on what the difference is between the harness and this, because they seem to do the same thing (run battles randomly and report the errors that result).

Well, the smoke test runner builds on top of the harness, as does my forthcoming benchmarking and profiling logic (which will probably be added to harness instead of its own file). I imagine I'll probably build more things on top of the harness as well in the future. Logically it's all the same shit, they could all be in one monolithic file, but I don't think thats as maintainable, and the `Pool`/`TeamGenerator`/`CoordinatedPlayerAI` logic was distinct enough from the stuff in `harness.js` that it seemed like it could be a separate thing. PS seems to like monolithic files.
 
> But you sound antsy. Nothing's blocking this, if you want to merge and figure out the rest later.

I'm only antsy because I find it really useful (I've been running it against all my recent changes and its saved me several times when the `npm run test` suite passed and missed bugs) and I'd like others to be able to use it. :)

dev-tools/smoke.js
Zarel: You wouldn't be resisting this much if this were actually a duck. I just don't like that `cycles` isn't actually a cycle count, only its absolute value is. Hence why I think `cyclesSetting` is a better name for it.

Zarel: I guess it's not a huge deal to me in that like most things, if you insisted, I'd let it go. I'm just not seeing the value here. It seems like the variable names are bad for seemingly no reason.

Zarel: In my experience, good variable names tends to be one of the more important things for readability.

Zarel:
> Well, the smoke test runner builds on top of the harness, as does my forthcoming benchmarking and profiling logic (which will probably be added to harness instead of its own file). I imagine I'll probably build more things on top of the harness as well in the future. Logically it's all the same shit, they could all be in one monolithic file, but I don't think thats as maintainable, and the `Pool`/`TeamGenerator`/`CoordinatedPlayerAI` logic was distinct enough from the stuff in `harness.js` that it seemed like it could be a separate thing. PS seems to like monolithic files.

I do prefer files being separate. It's just that I'd like the line separating them to be sensible, which is important because I want developers to be able to know what file the thing they want to change is in. PS has monolithic files because it often isn't clear to me what the best way to separate them is.

In this case, I don't understand what the line is between a "harness" and a "smoke test", especially since `HARNESS.md` says "run 100 random battles and report any errors that occurred" and "the number of battles run can be configured" – which seems identical to your documentation of the smoke test runner?

> I'm only antsy because I find it really useful (I've been running it against all my recent changes and its saved me several times when the `npm run test` suite passed and missed bugs) and I'd like others to be able to use it. :)

I'm not opposed to you squash-merging this whenever you personally feel it's ready.

dev-tools/smoke.js
scheibo: Only now that I've removed the `Math.abs` did I remember why I needed it - the error message on line 340 doesn't match the command run. Doesnt really matter, because you should be able to use negative or positive to repro, but now I at least remember *why* I wrote it the way I did.

scheibo:
> PS has monolithic files because I procrastinate figuring out the best way to separate them.

[Perfect is the enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good). Nothing is set in stone, you can always reorganize at a later time if the original separation wasn't correct.

> In this case, I don't understand what the line is between a "harness" and a "smoke test"

`harness.js` is poorly named, the relevant classes are `Runner`, `MultiRunner` and `SmokeRunner`.
*All* are [test harnesses](https://en.wikipedia.org/wiki/Test_harness). `Runner` runs a single battle. `MultiRunner` runs many. `SmokeRunner` also runs many but with coordination instead of purely randomly, but there still should be room to somehow better merge these two. I think the main problem though is the `harness.js` vs. `smoke.js` file names - running `node harness.js 100` runs 100 random battles which could *also* smoke out crashes/be described as a smoke test, its just less thorough/exhaustive than `smoke.js`. 

Improving the organization/naming/documentation here is a goal of mine, but I don't know the answer just yet, I figure these things out of time. I keep adding crap to these (like for testing #5270), so I'm going to be forced to rethink the design to make things more maintainable and intuitive in the long run.

Zarel:
> [Perfect is the enemy of the good](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good). Nothing is set in stone, you can always reorganize at a later time if the original separation wasn't correct.

On the other hand, nothing is set in stone, you can always split the monolithic files into smaller files later.

In the meantime, I like filenames that make sense. ;)

PR #5409, opened 2019-04-04
Allow Room Owners to use addhtml

TheMezStrikes:
While writing this I noticed addhtml is not included in the comment decribing all permissions. I'm bad at writing descriptions so I didn't fix it but I thought I'd mention it

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -493,6 +493,7 @@ exports.grouplist = [
 		inherit: '@',
 		jurisdiction: 'u',
 		declare: true,
+		addhtml: true,
 		modchat: true,
 		roomonly: true,
 		gamemanagement: true,
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -482,6 +482,7 @@ exports.grouplist = [
 		roomdriver: true,
 		editroom: true,
 		declare: true,
+		addhtml: true,
 		modchatall: true,
 		roomonly: true,
 		gamemanagement: true,


xfix:
Should it `inherit: '*'`, I wonder, what do you think, @Zarel?

config/config-example.js
```diff
@@ -493,6 +493,7 @@ exports.grouplist = [
 		inherit: '@',
```
xfix: I wonder whether instead of doing this, shouldn't this be changed to inherit from `*` instead of `@`. @Zarel?

bumbadadabum:
Tiny nitpick but you should edit server/rooms.js#L604 because it would now list ROs as normal users. Super niche but needs to be said. Also what xfix said might be a better way to implement this as ROs can promote bots, there's no reason for there to ever be a thing a bot can do that a RO can't.

config/config-example.js
Zarel: I don't want bots part of the regular inheritance chain, even if ROs are supposed to be a strict superset. I've always modeled them as a fork.

xfix:
For a pull request called "Allow Room Owners to use addhtml", it suspiciously doesn't change the room owners permissions, is it intentional?

Zarel:
I mean, I do think this is intentional. Bots can make groupchats and be their host, and being able to control those rooms is probably useful. So it's just the PR title that's wrong.

TheMezStrikes:
Nah that's not intentional, I'm actually stupid and forgot to add the permission to Room Owners

Zarel:
I hope this doesn't backfire...

PR #5411, opened 2019-04-05
Allow Global Voices and higher to pm /code

AustinXII:


AustinXII:
@Zarel

Zarel:
This code does not check for global voice or higher...?

server/chat-plugins/info.js
```diff
@@ -2329,6 +2330,7 @@ const commands = {
 	},
 	codehelp: [
 		`!code [code] - Broadcasts code to a room. Accepts multi-line arguments. Requires: + % @ & # ~`,
+		`In order to use !code in pms you must be a global voice or higher`,
```
Zarel:
```suggestion
		`In order to use !code in private messages you must be a global voice or higher`,
```

AustinXII:
@Zarel there is a room check, since pms are a  global room you need to be global voice or higher. I did test this :P

Zarel:
Oh! So this was already implemented, we just accidentally omitted the `"!code": true` part to make it work.

AustinXII:
Well you can broadcast /code in normal chat rooms so I don’t know if that was the case but it definitely made for an quick and easy PR

Issue #5415, opened 2019-04-05
ActiveMove is problematic


scheibo:
#5389 made `ActiveMove` a class, which brought some concerns:

> Hm I don't think I like the idea of `ActiveMove` being its own class. Especially since it doesn't actually extend `Move`, it actually makes move properties non-readonly.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/diffs_

and importantly also [caused some crashes](https://github.com/Zarel/Pokemon-Showdown/pull/5389/files#r272753761). It also presents some challenges for serialization in #5270, as I need to include a `_class_` tag (WIP: 'feeling cute, might delete later'):

```json
  "lastMove": {
    "_class_": "ActiveMove",
    "move": "[Move:willowisp]",
    "isViable": true,
    "zMoveBoost": {
      "atk": 1
    },
    "contestType": "Beautiful",
    "hit": 1,
    "moveHitData": {
      "crit": {},
      "typeMod": {},
      "zBrokeProtect": {}
    },
    "totalDamage": 0
  }
```

However, I'm not sure the previous approach (before it was made it a `class`) was any better, as we'd often simply cast arbitrary objects to `/** @type {ActiveMove} */`, and simply cloning + adding crap to a `Data.Move` is [not ideal for performance, as thanks to the monomorphism optimizations](https://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html) we would ideally be assigning everything on creation in the literal/`constructor`. From a serialization point of view, detecting an `ActiveMove` without the constructor becomes more difficult (looking for a `hit` field or similar is somewhat error prone, so I'd probably end up not being able to collapse the attributes cloned from `Move`).

- Serialization concerns would have us create an object that looks like:

```js
interface ActiveMove {
  move: Data.Move;
  activeCrap: ...
}
```

- Performance concerns would lead us to not modifying fields of the existing `Move` object, which the `ActiveMove`-as-a-`class` and the strawman interface above would support.
- From conversations with @Zarel it seems like he'd prefer the simplicity of reverting back to the old `ActiveMove` format and simply added a `getHitData` to `Pokemon`.

[...]

PR #5416, opened 2019-04-06
Move Data.ActiveMove methods to Pokemon

scheibo:
See #5415 for context - having these methods is unsafe and leads
to crashes because not all `ActiveMove`s are created through the
`Data.ActiveMove` constructor. Instead of `Pokemon`, these could
alternatively be static methods on` ActiveMove` (or the `ActiveMove`
`class` could be completely abolished and reverted back to an
`interface`), but #5415 will deal with `ActiveMove` long term, this just
fixes the crashes.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -525,6 +525,34 @@ export class Pokemon {
 		return null;
 	}
 
+	getMoveHitData(move: ActiveMove) {
+		move = Pokemon.ensureActiveMove(move);
+		return {
+			crit: move.moveHitData.crit[this.toString().slice(0, 3)] || false,
+			typeMod: move.moveHitData.typeMod[this.toString().slice(0, 3)] || 0,
+			zBrokeProtect: move.moveHitData.zBrokeProtect[this.toString().slice(0, 3)] || false,
+		};
+	}
+
+	setMoveCrit(move: ActiveMove) {
+		Pokemon.ensureActiveMove(move).moveHitData.crit[this.toString().slice(0, 3)] = true;
+	}
+
+	setMoveTypeModFor(move: ActiveMove, typeMod: number) {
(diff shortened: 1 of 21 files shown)


sim/pokemon.ts
```diff
@@ -525,6 +525,35 @@ export class Pokemon {
...
 	}
 
+	getMoveHitData(move: ActiveMove) {
+		move = Pokemon.ensureActiveMove(move);
+		const slotid = this.toString().slice(0, 3);
+		return {
+			crit: move.moveHitData.crit[slotid] || false,
+			typeMod: move.moveHitData.typeMod[slotid] || 0,
+			zBrokeProtect: move.moveHitData.zBrokeProtect[slotid] || false,
+		};
```
urkerab: I'm wondering whether the move hit data should be stored the other way around, i.e. `move.moveHitData[slotid][datatype]` instead of `move.moveHitData[datatype][slotid]`. I guess it would look something like this:

    getMoveHitData(move: ActiveMove) {
        const slotid = this.toString().slice(0, 3);
        if (!move.moveHitData) move.moveHitData = {};
        if (!move.moveHitData[slotid]) move.moveHitData[slotid] = { crit: false, typeMode: 0, zBrokeProtect: false };
        return move.moveHitData[slotid];
    }

    setMoveCrit(move: ActiveMove) {
        this.getMoveHitData(move).crit = true;
    }

    setMoveTypeModFor(move: ActiveMove, typeMod: number) {
        this.getMoveHitData(move).typeMod = number;
    }

    setMoveZBreakProtect(move: ActiveMove) {
        this.getMoveHitData(move).zBrokeProtect = true;
    }

Zarel: I actually did bring this up earlier. I agree, that would be better.

Zarel: I'll work on this myself.

urkerab: Ah, sorry, I failed to make the connection.

Issue #5412, opened 2019-04-05
Replays have commit SHAs that do not reflect anything on the public server


yuzeh:
https://github.com/Zarel/Pokemon-Showdown/blob/6c1ab309774e3ae1111abea114dc63a1cd5bc49f/server/room-battle.js#L1080-L1087

Is this what is being written into replay logs on the first line?

Zarel:
Yeah, there was an unpushed commit on the server. I've pushed it now: https://github.com/Zarel/Pokemon-Showdown/commit/12f49b20e16c0fab0cd1a1ca9ac1a5a1cc8d7e82

yuzeh:
I think it might be more than that. Here's a few examples:

* https://replay.pokemonshowdown.com/gen7randombattle-887606429.json references `664768c238d88325eea8f848b9726dcf8abd958e` which doesn't exist in the public repo
* https://replay.pokemonshowdown.com/gen7randombattle-879384978.json references `dab6083d84cf10c82638ffd4b09ad28aa8458e44` which also doesn't exist in the public repo

Is the master branch being overwritten locally on the deploy machine(s)?

Zarel:
664768c238d8 is actually ac5c0a5760cff588b9c25 in the public repo

dab6083d84cf10c is actually 9315b8d4d83df3 in the public repo

They're named differently because 12f49b2 was being constantly rebased on top of them.

yuzeh:
Oof, okay. I'm going to go back to using guess and check, then.

scheibo:
Out of curiosity, what are you using it for?

yuzeh:
Testing two ideas:
1. Imitation learning
2. Computing a measure of how "human-like" my current agent is

Slayer95:
Looks like it'd be better to run ``git merge-base origin/master HEAD`` ?

yuzeh:
@Slayer95 agreed. I'll submit a PR.

yuzeh:
@Slayer95 I've thought about this a bit more. If rebasing on the deploy-local master branch is going to be a common occurrence, I'd rather the system generate a confusing SHA than provide a SHA that is potentially wrong (and won't reconstruct properly).

@Zarel thoughts?

Slayer95:
Local edits to the battle engine shouldn't really happen -most should be to the server infrastructure AFAICT.

[...]

PR #5417, opened 2019-04-06
use origin/master as basis for replay version tag

yuzeh:
See #5412

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -1080,7 +1080,7 @@ if (!PM.isParentProcess) {
 	global.__version = '';
 	try {
 		const execSync = require('child_process').execSync;
-		const out = execSync('git merge-base master HEAD', {
+		const out = execSync('git merge-base origin/master HEAD', {
 			stdio: ['ignore', 'pipe', 'ignore'],
 		});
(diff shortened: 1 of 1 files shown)


Zarel:
In general, Main should never be running with unpushed commits.

I think using local head is better, because in the exceptions:

1. Sometimes Main runs on a detached HEAD on an older version, if there's a critical bug in origin/master HEAD.

2. If there _is_ an unpushed commit, it'll probably be pushed soon.

Slayer95:
Would you be fine with reporting some sort of compound version from the latest ``HEAD`` commit and the ancestor in ``origin/master`` ?

Zarel:
Sure, if they differ.

yuzeh:
Not sure I completely follow. What should this look like?

This is my idea:
```
>version-HEAD `git merge-base HEAD HEAD`
>version-master `git merge-base master HEAD`
>version-origin-master `git merge-base origin/master HEAD`
```

Zarel:
I would say `>version` as before, and `>version-origin` only if they differ.

Slayer95:
Fixed in #5542.

PR #5418, opened 2019-04-07
Remove |callback| in favor of |error|

### LINKED ISSUE #5414: Remove |callback| in favor of |error| + updated request
There was discussion on #5370 about this, but https://github.com/Zarel/Pokemon-Showdown/pull/5370#issuecomment-478308587 discovered the existing error message is insufficent, and either making `|error|` structured or resending the updated `request` along with an `|error|` would be required to replace `|callback|`. If we resent the request with approriate information updated (ie. `maybeTrapped` -> `trapped`, etc), it would remove the need for `RandomPlayerAI` to have any state, at the cost of additional bandwidth usage in the error cases (expected to be negligible in terms of oveall bandwidth impact).

scheibo:
Fixes #5414.

---

Currently this won't work because the `RandomPlayerAI` will break as receiving the choice `|error|` will cause it to throw. The primary problem is that we don't know which choice error corresponds to the request - with this protocol the AI receives an updated request + an error, but it could issue an updated choice before it receives the error and then cant tell if the error corresponded to the first or second request. 

We could not issue a choice error if we reissue the request, but that seems wrong to me as the original choice was in fact erroneous. I think the best solution is going to be to that instead of emitting the request + error, we emit the request as *part* of the `|error|` (ie/ `|error|<MESSAGE>(|<REQUEST>)?`) provided an update occurred.

I'm also not sure of scenarios where we would ever need to update the request for zmove/mega/ultra burst. Isn't it always known whether you can do one of those? Why did we have `|callback|`'s for those?

### DIFF
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -680,7 +680,12 @@ Mean Look or something), you will receive a message starting with:
 
 `|error|[Invalid choice]`
 
-This will tell you to send a different decision.
+This will tell you to send a different decision. If your previous choice
+revealed additional information, the error will contain an updated `REQUEST`
+object (in the same format as `|request|`) to base your decisions off of:
+
+`|error|[Invalid choice] MESSAGE|REQUEST`
+
 
 ### Choice requests
 
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -680,7 +680,7 @@ revealed additional information (For example: a move disabled by Imprison
 or a trapping effect), the error will be followed with a `|request|` command
 to base your decision off of:
 
-`|error|[Unavailable choice] MESSAGE`
+`|error|[Unavailable choice] MESSAGE`  
 `|request|REQUEST`
 
 ### Choice requests
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -678,14 +678,15 @@ against race conditions involving `/undo` (the cancel button).
 If an invalid decision is sent (trying to switch when you're trapped by
 Mean Look or something), you will receive a message starting with:
 
-`|error|[Invalid choice]`
+`|error|[Invalid choice] MESSAGE`
 
 This will tell you to send a different decision. If your previous choice
-revealed additional information, the error will contain an updated `REQUEST`
-object (in the same format as `|request|`) to base your decisions off of:
+revealed additional information (For examplse: a move disabled by Imprison
+or a trapping effect), the error will be followed with a `|request|` command
+to base your decision off of:
 
-`|error|[Invalid choice] MESSAGE|REQUEST`
-
+`|error|[Unavailable choice] MESSAGE`
+`|request|REQUEST`
 
 ### Choice requests
 
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -676,7 +676,7 @@ Mean Look or something), you will receive a message starting with:
 `|error|[Invalid choice] MESSAGE`
 
 This will tell you to send a different decision. If your previous choice
-revealed additional information (For examplse: a move disabled by Imprison
+revealed additional information (For example: a move disabled by Imprison
 or a trapping effect), the error will be followed with a `|request|` command
 to base your decision off of:
 
--- a/sim/examples/random-player-ai.ts
+++ b/sim/examples/random-player-ai.ts
@@ -15,11 +15,6 @@ export class RandomPlayerAI extends BattlePlayer {
 	readonly mega: number;
 	readonly prng: PRNG;
 
-	trapped: Set<number>;
-	disabled: Map<number, Set<string>>;
-	lastRequest?: AnyObject;
-	retry: boolean;
-
 	constructor(
 		playerStream: ObjectReadWriteStream<string>,
 		options: {move?: number, mega?: number, seed?: PRNG | PRNGSeed | null } = {},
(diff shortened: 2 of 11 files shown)


Slayer95:
> we don't know which choice error corresponds to the request 

We should leverage the ``rqid`` feature to link the error to the original request.

> I'm also not sure of scenarios where we would ever need to update the request for zmove/mega/ultra burst. Isn't it always known whether you can do one of those? Why did we have |callback|'s for those?

Yea, we don't need to preserve these cases.

scheibo:
> We should leverage the `rqid` feature to link the error to the original request.

Ah, good idea. So we can add `rqid` to the `|error|` instead of combining error + request together? However, this does mean the AI needs to become stateful again (to track the `rqid` of the choice it made). Though I guess it doesnt need to be one or the other - `|error|` can both specify the `rqid` its relevant for as well as forward along the updated request information (if any).

> Yea, we don't need to preserve these cases.

Thanks, will remove.

scheibo:
I'm happy with this as of 2bcff68 (ie. the resulting AI code is pretty clean). I'll modify the 2 tests which depend on `|callback|` if we're OK with the protocol. I think adding a `rqid` to choice errors would be nice to have but is not my ideal end state (because its stateful), so I don't think i want to block on that.

sim/SIM-PROTOCOL.md
```diff
@@ -680,7 +680,12 @@ Mean Look or something), you will receive a message starting with:
 
 `|error|[Invalid choice]`
 
-This will tell you to send a different decision.
+This will tell you to send a different decision. If your previous choice
+revealed additional information, the error will contain an updated `REQUEST`
+object (in the same format as `|request|`) to base your decisions off of:
+
+`|error|[Invalid choice] MESSAGE|REQUEST`
```
Zarel: Hm.

You clearly need something in a choice error, to tell the random player AI "don't re-make the same choice, wait for a new request". But I don't like bundling the request here.

I think I'd prefer something like

```
|error|[Unavailable choice] MESSAGE
|request|NEWREQUEST
```

Zarel: Which I think would be backward-compatible across all existing sim uses?

scheibo: Having the two messages complicates `BattlePlayer#receiveLine`, no?

Zarel: Not any more than having a possible `REQUEST` term in `|error|`, in my view.

Zarel: Oh, I see what you mean. I'm surprised `receiveLine` isn't already designed to be multiline. But it would need to be sooner or later, anyway.

scheibo: Is that so? We expect clients to support lookahead?

Zarel: I'm not sure what you mean by "lookahead"? We expect clients to process messages line-by-line.

scheibo: `BattlePlayer` receive becomes:

```js
receive(chunk: string) {
	const lines = chunk.split('\n');
	if (lines.length >= 2 &&
		lines[0].startsWith('|error|[Invalid Choice]') &&
		lines[1].startsWith('|request|')) {
		const err = lines[0].slice(7);
		const request = lines[1].slice(9);
		this.receiveError(new Error(err), JSON.parse(request));
		const removed = lines.splice(0, 2);
		if (this.debug) console.log(removed);
	}

	for (const line of lines) {
		this.receiveLine(line);
	}
}
```

It needs to 'peek' (ie/ lookahead, a very common term in [parsing](https://en.wikipedia.org/wiki/Parsing#Lookahead)).

scheibo: Actually, I seem to recall a very angry comment from you in the client code when someone broke the 'line-by-line' requirement, so I don't think you want `BattlePlayer` to have to do this. If I'm mistaken and lookahead is not required, please show me the code `BattlePlayer`/`RandomPlayer` should be using with the separate `|error|` + `|request|` protocol.

Zarel: Oh. You're right, I don't want lookahead at all. That's why I suggested different names `[Unavailable choice]` vs `[Invalid choice]`, to make it clear whether or not a corrected request was forthcoming.

Zarel: This is also why I said I'd expect it to be backward compatible. Because `[Unavailable choice]` is already not considered a choice error by the client, so no additional code would be necessary.

Zarel: Something like `[Disabled choice]` might also work.

scheibo: I didn't notice the `[Unavailable choice]` distinction in your original comment or appreciate the significance.

sim/SIM-PROTOCOL.md
```diff
@@ -673,9 +673,15 @@ battle will continue.
...
+`|error|[Invalid choice] MESSAGE`
 
-This will tell you to send a different decision.
+This will tell you to send a different decision. If your previous choice
+revealed additional information (For example: a move disabled by Imprison
+or a trapping effect), the error will be followed with a `|request|` command
+to base your decision off of:
+
+`|error|[Unavailable choice] MESSAGE`
+`|request|REQUEST`
```
Zarel: This should either be multiline code, or the first line needs to have two spaces after it to force a linebreak.

sim/examples/random-player-ai.ts
```diff
@@ -29,53 +24,15 @@ export class RandomPlayerAI extends BattlePlayer {
...
-			// If we get a choice error, we retry the choice using the last request provided
-			// we've got a '|callback' updating our state regarding what Pokemon are trapped
-			// or disabled.
-			this.makeChoice(this.lastRequest!);
-		} else {
-			throw error;
-		}
+		// If we made an unavailable choice we will receive a followup request to
+		// allow us the opportunity to correct our decision.
+		if (!error.message.startsWith('[Unavailable choice]')) throw error;
```
Zarel:
```suggestion
		if (error.message.startsWith('[Unavailable choice]')) return;
		throw error;
```

Really minor nitpick, but I think this makes the code more readable.

sim/SIM-PROTOCOL.md
scheibo: Done

sim/side.ts
```diff
@@ -314,19 +314,16 @@ export class Side {
...
 		this.battle.send('sideupdate', `${this.id}\n|request|${JSON.stringify(update)}`);
 		this.activeRequest = update;
 	}
 
-	emitChoiceError(message: string) {
+	emitChoiceError(message: string, unavailable?: boolean) {
 		this.choice.error = message;
-		this.battle.send('sideupdate', `${this.id}\n|error|[Invalid choice] ${message}`);
-		if (this.battle.strictChoices) throw new Error(`[Invalid choice] ${message}`);
+		const type = `[${unavailable ? 'Unavailable' : 'Invalid'} choice]`;
```
Zarel: Bikeshedding time! What should be the new name for "Unavailable choice"?

I'm currently leaning towards "Invalid choice".

scheibo: Oh god, I should have merged it myself after you approved ;)

We can't make the name "Invalid choice", the whole point is that its *different* from "Invalid choice" so that it can be distinguished. Do you mean "Disabled choice"? I'm indifferent

Zarel: Err, I mean "Disabled choice". Typo on my part.

Zarel: "Unavailable" mainly strikes me as kind of complicated for English. It might not play as well with non-native speakers, even if it is more precise/accurate than "Disabled".

Zarel: Redundancy is maybe a concern. "[Unavailable] Move x is disabled" is less redundant than "[Disabled] Move x is disabled".

Other question: "[Unavailable]" vs "[Unavailable choice]"?

Zarel: (None of this actually matters; feel free to merge.)

scheibo: I went with "Unavailable choice" as opposed to "Disabled choice" for the redundancy/stutter it introduces, as you identified. I also think "choice" is useful in the description for symmetry with "Invalid choice" and because "Unavailable" makes it seem like simulator itself is unavailable (like a 500 error on a web server) instead of the choice that was made being unavailable.

I'm also not in love with "Unavailable choice" FWIW, it just seems like the least bad option.

Zarel: Fair. I have nothing to add there.

PR #5422, opened 2019-04-07
Send blockpms and blockchallenges status in |updateuser|

scheibo:
Required for Zarel/Pokemon-Showdown-Client#1263.

---

The client (+ Preact client) already seem resilient to having additional parts to `|updateuser|`. I think though it would be better to make a JSON field `SETTINGS` instead of tacking on additional user settings as needed?

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -227,11 +227,12 @@ represented by a space), and the rest of the string being their username.
 > Finish logging in (or renaming) by sending: `/trn USERNAME,0,ASSERTION`
 > where `USERNAME` is your desired username and `ASSERTION` is `data.assertion`.
 
-`|updateuser|USERNAME|NAMED|AVATAR`
+`|updateuser|USERNAME|NAMED|AVATAR|BLOCKPMS|BLOCKCHALLENGES`
 
-> Your name or avatar was successfully changed. Your username is now `USERNAME`.
-> `NAMED` will be `0` if you are a guest or `1` otherwise. And your avatar is
-> now `AVATAR`.
+> Your name, avatar or blocking status was successfully changed. Your username is
+> now `USERNAME`. `NAMED` will be `0` if you are a guest or `1` otherwise. Your
+> avatar is now `AVATAR`. `BLOCKPMS` and `BLOCKCHALLENGES` will respectively be
+> `0` if true globally and `1` otherwise.
 
 `|formats|FORMATSLIST`
 
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -227,12 +227,12 @@ represented by a space), and the rest of the string being their username.
 > Finish logging in (or renaming) by sending: `/trn USERNAME,0,ASSERTION`
 > where `USERNAME` is your desired username and `ASSERTION` is `data.assertion`.
 
-`|updateuser|USERNAME|NAMED|AVATAR|BLOCKPMS|BLOCKCHALLENGES`
+`|updateuser|USERNAME|NAMED|AVATAR|BLOCKPMS|SETTINGS`
 
 > Your name, avatar or blocking status was successfully changed. Your username is
 > now `USERNAME`. `NAMED` will be `0` if you are a guest or `1` otherwise. Your
-> avatar is now `AVATAR`. `BLOCKPMS` and `BLOCKCHALLENGES` will respectively be
-> `0` if true globally and `1` otherwise.
+> avatar is now `AVATAR`. `SETTINGS` is a JSON object representing the current
+> state of various user settings.
 
 `|formats|FORMATSLIST`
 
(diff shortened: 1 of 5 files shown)


PROTOCOL.md
```diff
@@ -227,11 +227,12 @@ represented by a space), and the rest of the string being their username.
 > Finish logging in (or renaming) by sending: `/trn USERNAME,0,ASSERTION`
 > where `USERNAME` is your desired username and `ASSERTION` is `data.assertion`.
 
-`|updateuser|USERNAME|NAMED|AVATAR`
+`|updateuser|USERNAME|NAMED|AVATAR|BLOCKPMS|BLOCKCHALLENGES`
```
Zarel: Hm, I would probably want to send the rest via JSON, rather than ossifying them?

server/chat-commands.js
```diff
@@ -593,13 +593,14 @@ const commands = {
...
 	ignorepm: 'ignorepms',
-	ignorepms(target, room, user) {
+	ignorepms(target, room, user, connection) {
 		if (user.ignorePMs === (target || true)) return this.errorReply("You are already blocking private messages! To unblock, use /unblockpms");
 		user.ignorePMs = true;
 		if (target in Config.groups) {
 			user.ignorePMs = target;
 			return this.sendReply(`You are now blocking private messages, except from staff and ${target}.`);
 		}
+		connection.send(user.getUpdateuserText());
```
Zarel: Why not `user.send`?

Heck, why not `user.update()`?

Zarel: The purpose of this is to get all the user's connections in sync, right?

PROTOCOL.md
scheibo: > I think though it would be better to make a JSON field `SETTINGS` instead of tacking on additional user settings as needed?

Agreed.

server/chat-commands.js
scheibo: > The purpose of this is to get all the user's connections in sync, right?

 Yes, sorry, I dont know much (anything?) about non-sim code. What's the preferred way to do this?

Zarel: `connection.send` will only send a message to the connection that sent the connection.

`user.send` will send a message to all connections associated with a user. Which is the one you want here, because you want all connections (and therefore all clients) to know that the user's settings have been updated.

I also think `user.send(user.getUpdateuserText());` should be broken out into its own function `user.update();`

PROTOCOL.md
```diff
@@ -227,11 +227,12 @@ represented by a space), and the rest of the string being their username.
 > Finish logging in (or renaming) by sending: `/trn USERNAME,0,ASSERTION`
 > where `USERNAME` is your desired username and `ASSERTION` is `data.assertion`.
 
-`|updateuser|USERNAME|NAMED|AVATAR`
+`|updateuser|USERNAME|NAMED|AVATAR|BLOCKPMS|SETTINGS`
```
Zarel:
```suggestion
`|updateuser|USERNAME|NAMED|AVATAR|SETTINGS`
```

I assume this was a typo.

server/users.js
```diff
@@ -987,6 +986,14 @@ class User extends Chat.MessageContext {
 		if (isForceRenamed) this.trackRename = oldname;
 		return true;
 	}
+	getUpdateuserText() {
+		const named = this.named ? 1 : 0;
+		const settings = {blockPMs: this.ignorePMs, blockChallenges: this.blockChallenges};
```
Zarel: We might want to rename `ignorePMs` to `blockPMs` while we're making this change.

Zarel: I also wonder if we should start bundling user settings together, instead of having them on the top level of `User`. We have a bunch more (although most of them probably shouldn't be sent here).

scheibo: > We might want to rename `ignorePMs` to `blockPMs` while we're making this change.

Done

> I also wonder if we should start bundling user settings together, instead of having them on the top level of `User`. We have a bunch more (although most of them probably shouldn't be sent here).

I think that's maybe out of scope and can be done in a followup by someone?

Zarel: Oh yeah the latter one was definitely not a suggestion for this PR.

PROTOCOL.md
```diff
@@ -227,11 +227,12 @@ represented by a space), and the rest of the string being their username.
 > Finish logging in (or renaming) by sending: `/trn USERNAME,0,ASSERTION`
 > where `USERNAME` is your desired username and `ASSERTION` is `data.assertion`.
 
-`|updateuser|USERNAME|NAMED|AVATAR`
+`|updateuser|USERNAME|NAMED|AVATAR|SETTINGS`
 
-> Your name or avatar was successfully changed. Your username is now `USERNAME`.
-> `NAMED` will be `0` if you are a guest or `1` otherwise. And your avatar is
-> now `AVATAR`.
+> Your name, avatar or blocking status was successfully changed. Your username is
```
Zarel:
```suggestion
> Your name, avatar or settings were successfully changed. Your username is
```

PR #5423, opened 2019-04-07
Update SIM-PROTOCOL regarding |request|

### LINKED ISSUE #5420: |request|'s REQUEST object should include a rqid
SIM-PROTOCOL.md documents a `rqid` field, but the `rqid` is actually implemented in `server/roombattle.js` instead of in the `sim`, so its not really part of the protocol. The logic should be moved into `sim` so that clients/middleware doesn't need to track it and so that it could be used in the simulator logic (eg. https://github.com/Zarel/Pokemon-Showdown/pull/5418#issuecomment-480553928)

scheibo:
Closes #5420.

### DIFF
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -124,8 +124,10 @@ The beginning of a battle will look something like this:
 > Gives a JSON object containing a request for a choice (to move or
 > switch). To assist in your decision, `REQUEST.active` has information
 > about your active Pokémon, and `REQUEST.side` has information about your
-> your team as a whole. `REQUEST.rqid` is an optional request ID (see
-> "Sending decisions" for details).
+> your team as a whole. If a switch is required, `REQUEST.forcedSwitch` will
+> contain your options. If a Team Preview decision is required,
+> `REQUEST.teamPreview` will be true. If no decision is expected,
+> `REQUEST.wait` will be set.
 
 `|inactive|MESSAGE` or `|inactiveoff|MESSAGE`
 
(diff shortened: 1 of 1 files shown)


sim/SIM-PROTOCOL.md
```diff
@@ -670,10 +672,11 @@ To be exact, `CHOICE` is one of:
 Once a choice has been set for all players who need to make a choice, the
 battle will continue.
 
-All decisions except `/undo` can be sent with `|RQID` at the end. `RQID` is
-`REQUEST.rqid` from `|request|`. Each `RQID` is a unique number used to
-identify which action the request was intended for and is used to protect
-against race conditions involving `/undo` (the cancel button).
+All decisions except `/undo` can be sent with `|RQID` at the end, where `RQID`
+can be computed from the number of `|request|`'s received (eg. the fifth
+`|request|` is `RQID` 5). The `RQID` is used to identify which action the
```
Zarel: This part still mentions RQID.

I think it might make sense to document RQID in SIM-PROTOCOL, and just note that it doesn't appear if you're using the sim directly rather than connecting through the server.

scheibo: It deliberately mentions rqid, but it explains that its computable not that its part of request.

TBH though, im beyond caring about this, I was just trying to do some due diligence followup. If you or anyone else wants to make the documentation more clear I'd be happy, but this is not worth it to me at this point.

Zarel: Sure.

PR #5425, opened 2019-04-08
Camouflage and Soak should fail if they don't change the type

urkerab:
... in much the same way that Protean does nothing if the user is already the correct type.

Bulbapedia says that Soak doesn't fail for pure Water types in Gen 5.

### DIFF
--- a/data/mods/gen5/moves.js
+++ b/data/mods/gen5/moves.js
@@ -934,6 +934,16 @@ let BattleMovedex = {
 		basePower: 40,
 		flags: {protect: 1, mirror: 1, sound: 1},
 	},
+	soak: {
+		inherit: true,
+		onHit(target) {
+			if (!target.setType('Water')) {
+				this.add('-fail', target);
+				return null;
+			}
+			this.add('-start', target, 'typechange', 'Water');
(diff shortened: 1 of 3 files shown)


data/mods/gen5/moves.js
```diff
@@ -934,6 +934,16 @@ let BattleMovedex = {
 		basePower: 40,
 		flags: {protect: 1, mirror: 1, sound: 1},
 	},
+	soak: {
+		inherit: true,
+		onHit(target) {
+			if (!target.setType('Water')) {
+				this.add('-fail', target);
```
Zarel: Why not `return false`?

DaWoblefet:
I should note that the fail for Soak at least reveals the target with its animation, unlike pretty much all other cases for "But it failed!". I'll test Camouflage today as well.

data/mods/gen5/moves.js
Zarel: Oh, because the animation. I would suggest adding a comment to explain that.

urkerab: I've added some comments; I hope you like them.

PR #5426, opened 2019-04-08
Support multiline commands in PM

Zarel:
I'm not pushing this directly because it's not hotpatchable. So... PR it is.

### DIFF
--- a/server/chat.js
+++ b/server/chat.js
@@ -106,13 +106,30 @@ class PatternTester {
 	/**
 	 * @param {string} text
 	 */
-	test(text) {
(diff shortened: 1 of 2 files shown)


Zarel:
Hmm, in hindsight, it wouldn't _break_ anything to hotpatch, so it might not be bad to just immediately merge. Still, though, with how hacky this code is, it could probably use code review.

Zarel:
The long-term plan is still for every command to be multiline, up until the next line that starts with `/`.

Slayer95:
LGTM

Issue #5391, opened 2019-03-31
Idiomatic Streams usage with Promises causes memory leaks


scheibo:
From https://github.com/Zarel/Pokemon-Showdown/pull/5370#discussion_r270618747

```
[20:29:07] @Slayer95: pre, fixed the memory leak by doing await new Promise(resolve => {setImmediate(resolve)}) every N games run
[20:30:14] @Slayer95: as it turns out, there is not a single setImmediate in the BattleStream implementation, so it's in fact a sync stream
[20:30:50] @Slayer95: now, exactly how that impacts the GC logic isn't very clear
[20:30:56] @pre: i think thats a `BattleStream` or `Steams` issue then?
[20:31:01] @pre: but thanks for figuring that out
[20:31:15] @Slayer95: there are very few articles discussing memory leaks from looping async
[20:31:19] @pre: i was literally just commenting on your callback implementation from yesterday
[20:31:42] @Slayer95: yes, a fix in BattleStream / Streams should be feasible
```

Zarel:
Streams are supposed to be either sync or async depending on which you prefer.

Also, in theory the `async` keyword compiles to the equivalent of a `setImmediate`.

Zarel:
Also: BattleStreams are designed to be sync.

scheibo:
So this should be closed as WAI? And users should just use `setImmediate` as needed? Perhaps we should add that guidance to `STREAMS.md`?

Zarel:
Well, if it leads to a memory leak under normal usage, clearly _something_ needs to be fixed. We should just figure out why.

Zarel:
It's unclear to me why using a sync stream would cause a memory leak. Do you have details?

Slayer95:
> Also, in theory the async keyword compiles to the equivalent of a setImmediate.

Not really.

```js
await x();
y();
```

is mostly equivalent to

```js
x(() => process.nextTick(y));
```

But ofc, ``nextTick()`` doesn't schedule ``y`` for the *next* tick, but for the *current* tick.

Slayer95:
When running this excerpt from #5370 

```js
		while ((format = this.getNextFormat())) {
			const seed = this.prng.seed;
			lastFormat = format;
			await new Runner(Object.assign({format}, this.options)).run().catch(err => {
				failures++;
				console.error(
					`Run \`node dev-tools/harness 1 --format=${format} --seed=${seed.join()}\` ` +
[...]

PR #5427, opened 2019-04-08
Battle toJSON/fromJSON

### LINKED ISSUE #5270: Serializable simulator state

scheibo:
Allow for serializing and deserializing simulator state.
Fixes #5270.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -11,6 +11,7 @@ import {Field} from './field';
 import {Pokemon} from './pokemon';
 import {PRNG, PRNGSeed} from './prng';
 import {Side} from './side';
+import {State} from './state';
 
 /** A Pokemon that has fainted. */
 interface FaintedPokemon {
@@ -30,6 +31,7 @@ interface BattleOptions {
 	p3?: PlayerOptions; // Player 3 data
 	p4?: PlayerOptions; // Player 4 data
 	debug?: boolean; // show debug mode option
+	deserialized?: boolean;
 	strictChoices?: boolean; // whether invalid choices should throw
 }
 
@@ -48,6 +50,7 @@ export type RequestState = 'teampreview' | 'move' | 'switch' | '';
 export class Battle extends Dex.ModdedDex {
 	readonly id: '';
 	readonly debugMode: boolean;
+	readonly deserialized: boolean;
 	readonly strictChoices: boolean;
 	readonly format: string;
 	readonly formatData: AnyObject;
@@ -118,6 +121,7 @@ export class Battle extends Dex.ModdedDex {
 
 		this.id = '';
 		this.debugMode = format.debug || !!options.debug;
+		this.deserialized = !!options.deserialized;
 		this.strictChoices = !!options.strictChoices;
 		this.format = format.id;
 		this.formatData = {id: format.id};
@@ -205,6 +209,14 @@ export class Battle extends Dex.ModdedDex {
 				this.setPlayer(side, options[side]!);
 			}
 		}
+	}
+
+	toJSON(): AnyObject {
+		return State.serializeBattle(this);
+	}
+
+	static fromJSON(serialized: string | AnyObject): Battle {
+		return State.deserializeBattle(serialized);
 	}
 
 	get p1() {
@@ -1544,6 +1556,8 @@ export class Battle extends Dex.ModdedDex {
 	}
 
 	start() {
+		// deserialized should use restart instead
+		if (this.deserialized) return;
 		// need all players to start
 		if (!this.sides.every(side => !!side)) return;
 
@@ -1587,6 +1601,17 @@ export class Battle extends Dex.ModdedDex {
 
 		this.addToQueue({choice: 'start'});
 		this.midTurn = true;
+		if (!this.requestState) this.go();
+	}
+
+	restart(send?: (type: string, data: string | string[]) => void) {
+		if (!this.deserialized) throw new Error('Attempt to restart a battle which has not been deserialized');
+		// @ts-ignore - readonly
+		this.send = send;
+		if (!this.started) {
+			this.start();
+			return;
+		}
 		if (!this.requestState) this.go();
 	}
 
@@ -2667,7 +2692,7 @@ export class Battle extends Dex.ModdedDex {
 					foeActive.removeVolatile('substitutebroken');
 				}
 			}
-			delete action.pokemon.draggedIn;
+			action.pokemon.draggedIn = null;
 			break;
 		case 'runPrimal':
 			if (!action.pokemon.transformed) {
@@ -3004,7 +3029,7 @@ export class Battle extends Dex.ModdedDex {
 		const format = this.getFormat();
 		let team = options.team;
 		if (typeof team === 'string') team = Dex.fastUnpackTeam(team);
-		if (!format.team && team) return team;
+		if ((!format.team || this.deserialized) && team) return team;
 
 		if (!options.seed) {
 			options.seed = PRNG.generateSeed();
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1621,7 +1621,6 @@ export class Battle extends Dex.ModdedDex {
 		if (!this.deserialized) throw new Error('Attempt to restart a battle which has not been deserialized');
 		// @ts-ignore - readonly
 		this.send = send;
-		if (!this.started) this.start();
 	}
 
 	boost(
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1595,9 +1595,8 @@ export class Battle extends Dex.ModdedDex {
 			this.add('rated', typeof this.rated === 'string' ? this.rated : '');
 		}
 
-		if (format.onBegin) {
-			format.onBegin.call(this);
-		}
+		if (format.onBegin) format.onBegin.call(this);
+		if (format.trunc) this.trunc = format.trunc;
 		for (const rule of this.getRuleTable(format).keys()) {
 			if (rule.startsWith('+') || rule.startsWith('-') || rule.startsWith('!')) continue;
 			const subFormat = this.getFormat(rule);
@@ -1619,6 +1618,10 @@ export class Battle extends Dex.ModdedDex {
 
 	restart(send?: (type: string, data: string | string[]) => void) {
 		if (!this.deserialized) throw new Error('Attempt to restart a battle which has not been deserialized');
+
+		const format = this.getFormat();
+		if (format.trunc) this.trunc = format.trunc;
+
 		// @ts-ignore - readonly
 		this.send = send;
 	}
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1032,11 +1032,33 @@ export class Battle extends Dex.ModdedDex {
 			type = this.requestState;
 		}
 
-		// default to no request
-		const requests: any[] = Array(this.sides.length).fill(null);
 		for (const side of this.sides) {
 			side.activeRequest = null;
 		}
+
+		const maxTeamSize = this.getMaxTeamSize();
+		if (type === 'teampreview') {
+			this.add('teampreview' + (maxTeamSize !== 6 ? '|' + maxTeamSize : ''));
+		}
+
+		const requests = this.getRequests(type, maxTeamSize);
+		for (let i = 0; i < this.sides.length; i++) {
+			this.sides[i].emitRequest(requests[i]);
+		}
+
+		if (this.sides.every(side => side.isChoiceDone())) {
+			throw new Error(`Choices are done immediately after a request`);
+		}
+	}
+
+	getMaxTeamSize() {
+		const teamLengthData = this.getFormat().teamLength;
+		return (teamLengthData && teamLengthData.battle) || 6;
+	}
+
+	getRequests(type: RequestState, maxTeamSize: number) {
+		// default to no request
+		const requests: any[] = Array(this.sides.length).fill(null);
 
 		switch (type) {
 		case 'switch': {
@@ -1054,10 +1076,6 @@ export class Battle extends Dex.ModdedDex {
 		}
 
 		case 'teampreview':
-			const teamLengthData = this.getFormat().teamLength;
-			let maxTeamSize = teamLengthData && teamLengthData.battle;
-			this.add('teampreview' + (maxTeamSize ? '|' + maxTeamSize : ''));
-			if (!maxTeamSize) maxTeamSize = 6;
 			for (let i = 0; i < this.sides.length; i++) {
 				const side = this.sides[i];
 				side.maxTeamSize = maxTeamSize;
@@ -1077,19 +1095,14 @@ export class Battle extends Dex.ModdedDex {
 
 		const allRequestsMade = requests.every(request => request);
 		for (let i = 0; i < this.sides.length; i++) {
-			const side = this.sides[i];
-			const request = requests[i];
-			if (request) {
-				if (!this.supportCancel || !allRequestsMade) request.noCancel = true;
-				side.emitRequest(request);
+			if (requests[i]) {
+				if (!this.supportCancel || !allRequestsMade) requests[i].noCancel = true;
 			} else {
-				side.emitRequest({wait: true, side: side.getRequestData()});
+				requests[i] = {wait: true, side: this.sides[i].getRequestData()};
 			}
 		}
 
-		if (this.sides.every(side => side.isChoiceDone())) {
-			throw new Error(`Choices are done immediately after a request`);
-		}
+		return requests;
 	}
 
 	tiebreak() {
@@ -1608,11 +1621,7 @@ export class Battle extends Dex.ModdedDex {
 		if (!this.deserialized) throw new Error('Attempt to restart a battle which has not been deserialized');
 		// @ts-ignore - readonly
 		this.send = send;
-		if (!this.started) {
-			this.start();
-			return;
-		}
-		if (!this.requestState) this.go();
+		if (!this.started) this.start();
 	}
 
 	boost(
--- a/sim/state.ts
+++ b/sim/state.ts
@@ -1,0 +1,356 @@
+/**
+ * Simulator State
+ * Pokemon Showdown - http://pokemonshowdown.com/
+ *
+ * @license MIT
+ */
+
+import {Battle} from './battle';
+import Dex = require('./dex');
+import * as Data from './dex-data';
+import {Field} from './field';
+import {Pokemon} from './pokemon';
+import {PRNG} from './prng';
+import {Choice, Side} from './side';
+
+// The simulator supports up to 24 different Pokemon on a team. Serialization
+// uses letters instead of numbers to indicate indices/positions, but where
+// the simulator only gives a position to active Pokemon, serialization
+// uses letters for every Pokemon on a team. Active pokemon will still
+// have the same letter as their position would indicate, but non-active
+// team members are filled in with subsequent letters.
+const POSITIONS = 'abcdefghijklmnopqrstuvwx';
+
+// Several types we serialize as 'references' in the form '[Type]' because
+// they are either circular or they are (or at least, should be) immutable
+// and thus can simply be reconsituted as needed.
+// NOTE: Template is not strictly immutable as some OM formats rely on an
+// onModifyTemplate event - deserialization is not possible for such formats.
+type Referable = Battle | Field | Side | Pokemon | PureEffect | Ability | Item | Move | Template;
+
+// Certain fields are either redundant (transient caches, constants, duplicate
+// information) or require special treatment. These sets contain the specific
+// keys which we skip during default (de)serialization (and the keys which)
+// need special treatment from these sets are then handled manually.
+
+// Battle inherits from Dex, but all of Dex's fields are redundant - we can
+// just recreate the Dex from the format.
+const BATTLE = new Set([
+	...Object.keys(Dex), 'inherit', 'cachedFormat', 'zMoveTable', 'teamGenerator',
+	'NOT_FAIL', 'FAIL', 'SILENT_FAIL', 'field', 'sides', 'prng', 'hints', 'deserialized',
+]);
+const FIELD = new Set(['id', 'battle']);
+const SIDE = new Set(['battle', 'team', 'pokemon', 'choice']);
+const POKEMON = new Set([
+	'side', 'battle', 'set', 'name', 'fullname', 'id', 'species', 'speciesid', 'happiness',
+	'level', 'pokeball', 'baseTemplate', 'baseHpType', 'baseHpPower', 'baseMoveSlots',
+]);
+const CHOICE = new Set(['switchIns']);
+// ActiveMove (theoretically) includes all of the fields of Move, but we need
+// to dynamically create that skip set as Moves have different fields depending
+// on their data.
+const ACTIVE_MOVE = new Set(['move']);
+
+export const State = new class {
+	// REFERABLE is used to determine which objects are of the Referable type by
+	// comparing their constructors. Unfortunately, due to circular module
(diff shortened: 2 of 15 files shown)


sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	private fromRef(ref: string, battle: Battle): Referable | undefined {
+		// References are sort of fragile - we're mostly just counting on there
+		// being a low chance that some string field in a simulator object will not
+		// 'look' like one. However, it also needs to match one of the Referable
+		// class types to be decode, so we're probably OK. We could make the reference
+		// markers more esoteric with additional sigils etc to avoid collisions, but
+		// we're making a conscious decision to favor readability over robustness.
+		if (ref[0] !== '[' && ref[ref.length - 1] !== ']') return undefined;
```
Slayer95: We prefer ``charAt()`` over bracket notation for strings.

sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+		const [type, id] = ref.split(':');
+		switch (type) {
+			case 'Side': return battle.sides[Number(id[1]) - 1];
+			case 'Pokemon': return battle.sides[Number(id[1]) - 1].pokemon[POSITIONS.indexOf(id[2])];
+			case 'Ability': return battle.getAbility(id);
+			case 'Item': return battle.getItem(id);
+			case 'Move': return battle.getMove(id);
+			case 'PureEffect': return battle.getEffect(id);
+			case 'Template': return battle.getTemplate(id);
+			default: return undefined; // maybe we actually got unlucky and its a string
```
Slayer95: ``switch`` and ``case`` should have the same indentation level

sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+		this.deserialize(state, field, FIELD, field.battle);
+	}
+
+	serializeSide(side: Side): /* Side */ AnyObject {
+		const state: /* Side */ AnyObject = this.serialize(side, SIDE, side.battle);
+		state.pokemon = new Array(side.pokemon.length);
+		const team = new Array(side.pokemon.length);
+		for (const [i, pokemon] of side.pokemon.entries()) {
+			state.pokemon[i] = this.serializePokemon(pokemon);
+			team[side.team.indexOf(pokemon.set)] = i + 1;
```
Slayer95: Any reason to use one-based indices here? ``side.team`` isn't used for anything after the battle is started, so it probably doesn't deserve this treatment.

sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+		}
+		state.team = team.join(team.length > 9 ? ',' : '');
```
Slayer95: This seems to be overcomplicating this field too.

I figure you are concerned about size of the serialized string? In that case we might as well omit ``.team`` from the output.

sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	private deserializeSide(state: /* Side */ AnyObject, side: Side) {
+		this.deserialize(state, side, SIDE, side.battle);
+		const ordered = new Array(side.pokemon.length);
+		const team = state.team.split(state.team.length > 9 ? ',' : '');
+		for (const [i, pos] of team.entries()) {
+			ordered[Number(pos) - 1] = side.pokemon[i];
+		}
+		side.pokemon = ordered;
```
Slayer95: I am not sure what you are doing here.

You seem to be restoring the Pokémon to its order at the start of the battle? Why?

sim/state.ts
```diff
@@ -0,0 +1,340 @@
...
+	private deserializeChoice(state: /* Choice */ AnyObject, choice: Choice, battle: Battle) {
+		this.deserialize(state, choice, CHOICE, battle);
+		choice.switchIns = new Set(state.switchIns);
+	}
+
+	// Simply looking for a 'hit' field to determine if an object is an ActiveMove or not seems
+	// pretty fragile, but its no different than what the simulator is doing. We go further and
+	// also check if the object has an 'id', as that's what we will intrepret as the Move.
+	private isActiveMove(obj: AnyObject): obj is ActiveMove {
+		return obj.hasOwnProperty('hit') && obj.hasOwnProperty('id');
```
Slayer95: Why don't you check for ``isCopy``, which is *exactly* what the simulator is doing?

sim/state.ts
scheibo: Kind of dumb but w/e. Seems like something the linter should enforce?

sim/state.ts
scheibo: Because I find it more readable and it looks more like the `team` command you send the simulator.

Slayer95:
This commit spends a lot of effort in (de)serialization of ``Effect``s, which AFAICT are only used for the ``sourceEffect`` slots of volatiles and field conditions. However, for the sim, their only relevant piece of data is their ``id``s (In fact, nowadays, ``move.sourceEffect`` is no longer an ``Effect``, but an ID).

Thoughts, @Zarel?

sim/state.ts
scheibo: No. I want to support 24 pokemon, so I was just encoding this as characters originally, but that led to confusion because index `a` (0) and `pokemon` `a` didn't line up. I chose 1-based number like the `team` command, and I don't like `1,2,3,4,5,6` because its too verbose. 

Basically, "i chose this format because it makes for the most readable output"

sim/state.ts
Slayer95: But why would you want to read ``state.team`` ? According to this implementation, it wouldn't match the actual ``/team`` command sent, wouldn't it?

sim/state.ts
scheibo: I'm doing the opposite (could use a comment). We create a `new Battle` and pass in `team`, but that creates the `pokemon` array in the wrong order. This code reshuffles the Pokemon from the starting order to the `pokemon` order.

sim/state.ts
Slayer95: We already enforce that for ``.js`` files. Idk if tslint has that setting.

sim/state.ts
scheibo: Um, what?

https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/dex.ts#L458

sim/state.ts
Slayer95: Again, why would you "read" this output? To start with, JSON isn't meant to be human-readable. If we want readibility we should be using YAML.

Still, this is to be used to be parsed by a computer, so the simplest code is best.

scheibo:
> This commit spends a lot of effort in (de)serialization of `Effect`s, which AFAICT are only used for the` sourceEffect` slots of volatiles and field conditions. However, for the sim, their only relevant piece of data is their `id`s (In fact, nowadays, `move.sourceEffect` is no longer an `Effect`, but an ID).

Not sure what you're talking about here? `toRef`/`fromRef`? Pretty sure only `Ability` and `ActiveMove` are `Effect`s, and we're effectively collapsing everything into an `id` anyway, which seems to be what you're arguing for?

sim/state.ts
Slayer95: Uh? Fine then, I wonder why would @Zarel change this...

Slayer95:
I was not arguing "for" anything. I was describing the situation: only their IDs are relevant.

Which implies the question: should the simulator keep using ``Effect``s are values for ``sourceEffect`` or should it switch to ``ID``s?

sim/state.ts
Slayer95: Oh, I see!

Still, re: my other comments, note that ``team`` doesn't match the actual ``/team`` command, after Pokémon are switched out.

sim/state.ts
scheibo: Maybe the new 'eslint-is-tslint' will fix this?

scheibo:
> I was not arguing "for" anything. I was describing the situation: only their IDs are relevant.

Sorry, I didn't mean to make you seem confrontational! (And I really appreciate your reviews!)

> Which implies the question: should the simulator keep using `Effect`s are values for `sourceEffect` or should it switch to IDs?

My original argument was going to be that IDs are slower because they require reconstitution (and creating `Effect` is kind of slow because it has a lot of fields), but with the caching we have it should just be a lookup to get back the `Effect` object so I'd be down for switching to using `IDs`.

sim/state.ts
Slayer95: I would suggest to switch these to use ``map()``. I find this code pretty confusing.

It's equivalent to ``const team = side.team.map(set => pokemon.set.indexOf(set) + 1)``... ... Is it???? :[

sim/state.ts
Slayer95: Yup!

sim/state.ts
Zarel: Also we prefer `ref.slice(-1)` to `ref.charAt(ref.length - 1)`

sim/state.ts
scheibo: JSON is 100% meant to human readable. It's text! This is without a doubt one of the main reasons it is so popular and used in non-JS languages for protocol even though its clearly inferior bandwidth and parsing wise than a binary protocol. Binary is for computers, text is for humans. Albeit, in Javascript writing a binary protocol is a PITA, but you could still base64 encode the whole thing if it wasnt meant for humans.

I dont want to support YAML, YAML is a dependency and slower on JS, where JSON is native.

This is meant to be parsed by a computer, but I need to a) read it in order to debug it while I'm implementing it b) plan to continue reading it going forward. As I am working towards a MCTS based AI, I assume I'm going to be the primary user of this serialization, and having it readable is important to me.

---

`side.state.team` = the command *I* would need to issue at `/team` to transform `side.state.pokemon` into the `side.state.team` order. That is intuitively how I think about the meaning of the field, and after trying several different encoding schemes for team, it proves to be the simplest. The code to parse it is still trivial for the computer, and even if it werent, the size of a team is bounded so we could sacrifice performance for readability if we needed to.

Zarel:
Although IDs are nonzero ambiguous for Effects.

Zarel:
Actually, in hindsight, no, they aren't.

Zarel:
Specifically, `effect.fullname` isn't ambiguous, as far as I'm aware.

sim/state.ts
Slayer95: I believe that meaning is not valuable enough for humans... for us to care for its readability... But I don't think this is worth arguing over either.

scheibo: If you think *this* isn't worth it, wait until you see what I've been working on this morning! While I'm debugging this, I constantly need to read over a `control.json` and `test.json`, and `activeRequest` on `sides` is a big block of annoying redundant info to scroll over, so now I'm working to collapse *that*. That's significantly more work than team encoding, but is of similar impact on readability. Readability matters, because I literally can't make progress on this if I can't debug WTF is going on, and I can only debug it if the output is readable.


(I'm actually not sure about how much work it will be to collapse `activeRequest` now that we mutate it thanks to removing `|callback|`. I think it would be ideal if the respective `getRequestData` will return the updated request automatically, but if not my plan is to elide `activeRequest` and instead encode an `activeRequestDiff` which will usually be `{}`, but in the cases where `activeRequest` was updated it will contain information about what things were revealed to be `disabled`/`trapped` etc.).

Slayer95: > Readability matters, because I literally can't make progress on this if I can't debug WTF is going on

Since you say it like so, this is literally like adding ``console.log()`` statements all over the place. The final protocol should be clean and needn't have the debug features.

Slayer95:
Regarding the issue of strings in the battle conflicting with the serialization format, we only need to add `battle.log`` to the skipped keys -it can be edited by admins with ``/a``. As for other user-editable content, it's already sanitized by ``Dex#getName()``, so it can't contain brackets.

Also, this is missing tests:
- Validate the serialization format spec.
- Ensure that a deserialized battle can be properly resumed and completed.

scheibo:
> Also, this is missing tests

I think you've missed https://github.com/Zarel/Pokemon-Showdown/issues/5270#issuecomment-480937346 (there's a lot of context in the other updates on that as well), and the fact that this is a draft PR :).

Slayer95:
Yes, I have missed all that thread!

scheibo:
Oof. Circled back to this while on the plane yesterday and made some progress, but its still as difficult as ever. I now have a passing test, for whatever that is worth.

- I'm collapsing `activeRequest` now, which is mostly for debuggability/size (its a large redundant object we can compute)
- `moveSlots` and `baseMoveSlots` want to be sharing the exact same object (identity) in certain circumstances, which I've handled elsewhere with 'refs' but here I'm just using `slice`. I have a `TODO` for now, but I think more work is needed here. **I'm also very worried more places depend on shared identity.**
- I had to write `State.equal` for testing (and its kind of ugly because we do an equal comparison twice so as to get a nice error message). I'm trying to avoid adding any dependencies here.
- `DualStream` and `tools/simulate --dual` is WIP...

Probably going to park this for another couple of weeks until I forget how painful it is :)

scheibo:
Realized I can just elide `undefined` values before the `stringify` step, which makes `assert.deepStrictEqual` work = much cleaner and simpler. `--dual` now fails as well, which actually makes me happy because I was deeply skeptical things actually worked before.

scheibo:
Yay! I think I finally trust my testing setup to a larger extent. Currently passing 38/100 random battles.

Manually picked some seeds to figure out what needs to be fixed with `node tools/simulate multi 1 --dual --seed=SEED`

- [ ] how to handle megas/formechanges (the `pokemon` template isn't going to match the `team`) = `1,2,3,0`, `1,2,3,1`, `1,2,3,6`, `1,2,3,7` (and more)
- [ ] 'mustrecharge' is getting send 4 instead of 3(!) times, which surely seems buggy in the first place? =
 `1,2,3,7`
- [ ] showCure shows up for some reason? =  `1,2,6,4`, `1,9,3,4`

There are surely more, but those are the 3 error patterns I've noticed so far. Fixing megas/forme changes will do the most - maybe I need to encode team if theres a difference?

scheibo:
12/100 failing now - this PR isn't as painful when I can actually make progress! Turns out I need to add the `Promise.race([...])` logic back to the harness or else it will abort after the first failure if I'm throwing in `compare`...

I know why `mustrecharge` fails, but I'm still debating how I want to solve it.

- [ ] `trapped` doesn't match, I think this has to do with my request collapsing logic = `--format=gen5randombattle --seed=32978,33296,27302,30395`, `--format=gen5randombattle --seed=38358,51500,25461,1053`, `--format=gen7randombattle --seed=5764,7213,6686,12539`
- [ ] some discrepancies with PP usage/moveSlots, maybe related to the identity issue =  `--format=gen1randombattle --seed=65339,57240,60857,62385`, `--format=gen2randombattle --seed=53388,44488,65417,35372`
- [ ] hpType mismatch (maybe need to retain `baseHpType`?) = `--format=gen2randombattle --seed=58320,1548,19233,21864`, `--format=gen2randombattle --seed=53079,50189,20513,41681`

scheibo:
> I know why `mustrecharge` fails, but I'm still debating how I want to solve it.

Decided to 'solve' it a simpler way during serialization, but #5462 will track a proper fix.

Moved the outstanding issues here:

- [ ] showCure shows up for some reason? =  `1,2,6,4`, `1,9,3,4`
- [ ] `trapped` doesn't match, ~I think~ confirmed this has to do with my request collapsing logic = `--format=gen5randombattle --seed=32978,33296,27302,30395`, `--format=gen5randombattle --seed=38358,51500,25461,1053`, `--format=gen7randombattle --seed=5764,7213,6686,12539`
- [ ] some discrepancies with PP usage/moveSlots, maybe related to the identity issue =  `--format=gen1randombattle --seed=65339,57240,60857,62385`, `--format=gen2randombattle --seed=53388,44488,65417,35372`

Zarel:
This pull request **introduces 2 alerts** when merging a6033290f8537b46d1a0254a3d373965bc73b29b into 82f0f9c71b7dbe1c1a4d689f0b8f9750efdd9154 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-8ab5a00ead012091702d374842c3c5e474891349)

**new alerts:**

---

Zarel:
This pull request **introduces 2 alerts** when merging f542f93baeec64ebd5a8497bf569dcadcf1623a1 into 82f0f9c71b7dbe1c1a4d689f0b8f9750efdd9154 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-a8339e246aba93077bd3185e271ec8ec242bc64b)

**new alerts:**

---

scheibo:
Getting so close! I'm 100/100 with random seeds so I'm now running it with the 'smoke' runner (`node tools/simulate exhaustive`) and seem to only be running into Trick Room issues - we're almost there!

As an aside - I have no clue how someone would be able to do this with any confidence without the `tootls/simulate` scripts...

Zarel:
This pull request **introduces 2 alerts** when merging e5cb25fc3179f4a10023994260bbaf49ff741557 into 82f0f9c71b7dbe1c1a4d689f0b8f9750efdd9154 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-1362710fff5b8a27ee871654258e929d03c7d2c2)

**new alerts:**

---

scheibo:
Trick Room is off because of f7094909c2289a7e33c834ef52fe2846608c988e. Go figure.

scheibo:
2 errors remaining after several thousands of rounds of battles! Once fixed, I'll probably also need to revert the `Promise.race` stuff until #5391 is fixed :|

```
Run `node tools/simulate exhaustive --cycles=5 --format=gen2customgame --seed=33971,18589,31546,54979`:
 TypeError: Cannot read property 'counter' of undefined
```
This could even be a real error on master TBH. I'll check in a few hours when I'm back

```
Run `node tools/simulate exhaustive --cycles=5 --format=gen1customgame --seed=1504,37375,12061,30501`:
 Error: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

  {
    abilityOrder: 59,
...
                move: 'Mist',
+               pp: 5,
-               pp: 47,
                target: 'allySide',
                target: 'allySide',
+               used: false,
+               virtual: true
-               used: true
              },
...
    turn: 71
  }
```
More `moveSlot` fun...

scheibo:
Oh yeah, if @Zarel is not down with 15d9abe (I don't think custom games should arbitrarily have different behavior, but its clearly intentional) I will just not support serialization for custom games and have my harness code explicitly hardcode the `trunc` swapping logic in `--dual` mode so that my comparisons still work. In theory I could run `format.onBegin` in `Battle.restart` and just do my lame detach + reattach `battle.log` trick, but doing a bunch of work to then throw it away for most formats is super inefficient, and this needs to be efficient.

Zarel:
This pull request **introduces 2 alerts** when merging 8cd55461734e305955b0223eddfb6c61dad1e492 into 82f0f9c71b7dbe1c1a4d689f0b8f9750efdd9154 - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-d0f5003895e77c08d261f10cd1f0fbdc15782544)

**new alerts:**

---

scheibo:
> More `moveSlot` fun...

This one is great! Pokemon A Mimic's Pokemon B and then Pokemon B Transforms into Pokemon A! *ay caramba*

scheibo:
Ran this through about 20k random battles at this point (and I'm actually trust my testing set up again, unlike earlier on when errors where being hidden) so I'm finally ready for this to be reviewed (thanks @yuzeh for being patient on this one :P).

- 15d9abe might be contentious as mentioned in https://github.com/Zarel/Pokemon-Showdown/pull/5427#issuecomment-486880430, but I can workaround if need be
- I'm surprised Typescript doesn't complain about `delete` changing the domain of a type or impossible checks to `undefined` (eg. `showCure: boolean`)
- collapsing `activeRequest` is the most work (recomputing it on deserialization), but its actually a huge object on each `side` that like doubles the serialized output, so I think its worth it. But who knows, maybe I'll do some benchmarking later and want to eat the extra size for performance. Its fairly simple to back out, so I'd like to land it as is for readability
- @Slayer95: any more unit tests that you'd like to see? Before you said "Validate the serialization format spec." - do you mean like a [golden files test](https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files) or something? It seems like itd need constant updates (but I'm probably misinterpreting you)? The `--dual` mode is what I care most about, but I'm hesitant to turn it on, even for `npm run fulltest`, because it basically triples the time it takes to run the already long tests (playing 2 battles at once is obviously 2x, and then `toJSON` + `fromJSON` each turn adds up to about one battle's worth of latency)

Zarel:
I'm not down with `15d9abe`

Zarel:
You could replace `onBegin` with `trunc` and just manually overwrite it.

scheibo:
> I'm not down with 15d9abe

I've reverted it. I don't really agree or understand, but I also don't want this PR to be the place to argue about that.

sim/tools/runner.ts
```diff
@@ -129,3 +139,72 @@ class RawBattleStream extends BattleStreams.BattleStream {
...
+				fs.writeFileSync('logs/test.json', JSON.stringify(test, null, 2));
+			}
+			throw new Error(err.message);
+		}
+
+		if (end) return;
+		const send = this.test.battle.send;
+		// NOTE: retaining trunc as part of the round trip is only required for certain custom
+		// game formats which override it in the format onBegin logic to have non-standard behavior.
+		const trunc = this.test.battle.trunc;
```
Zarel: `trunc` is a function and can't be serialized into JSON.

I'd recommend just defining `trunc` directly into the format so it can be manually imported.

scheibo: Obviously. `trunc` isn't being serialized here - we're saving it *before* we round trip and reinstalling it afterwards.

> I'd recommend just defining `trunc` directly into the format so it can be manually imported.

???

Zarel: > ???

```diff
-onBegin() { this.trunc = n => Math.trunc(n); },
+trunc(n) { return Math.trunc(n); },
```

Add some glue code for replacing `trunc` with the version in the format, at the start of a battle, and away we go.

scheibo: Done.

Zarel:
Neat, looks good. Do you think this is ready to merge?

scheibo:
> Do you think this is ready to merge?

Yes. Very confident at this point (>1M battles simulated...). I don't know if @Slayer95 wanted to weigh in, but I think I could always fix things in a follow up/add additional testing.

sim/battle.ts
```diff
@@ -1020,12 +1032,34 @@ export class Battle extends Dex.ModdedDex {
...
 
-		// default to no request
-		const requests: any[] = Array(this.sides.length).fill(null);
 		for (const side of this.sides) {
 			side.activeRequest = null;
 		}
 
+		const maxTeamSize = this.getMaxTeamSize();
+		if (type === 'teampreview') {
+			this.add('teampreview' + (maxTeamSize !== 6 ? '|' + maxTeamSize : ''));
```
urkerab: ... I had specifically changed this in PR #4281, and you went and changed it back again...

Zarel: Probably you should just change it again, but with a comment this time. :p

PR #5428, opened 2019-04-08
Pull out non-sim code into RoomBattleStream

scheibo:
`BattleStream` is the only part of `sim` which depends on `Monitor` and `Chat`. In order to get `sim` to function as a standalone package (design forthcoming, *'OP will surely deliver'*), we need to either pull out `Chat.stringify` and the `crashlogger` (*a la* [this gist](https://gist.github.com/scheibo/f399c7c3d02a857c36c0f8b6d5263a5a) that has probably rotted pretty hard by this point), or do what I've done in this PR which is to pull out the functionality which requires those dependencies into the `server` logic instead of `sim`. I think the approach taken here is preferable as general `sim` code isn't interested in slow battle logging etc.

The main motivation for doing this now as opposed to after I've sent out a more thorough design doc regarding modularization is because I'm looking to get (optional, for development) source map support working (I'm tired of clicking through to the `.sim-dist` files in VS Code and editing the compiled file by accident...). This PR, plus a change which splits the `sim` parts out of `globals.ts` and `globals.d.ts` (ie. into additional `sim.globals.ts` and `sim.d.ts`) means we can add a `tsconfig` which can optionally be used to actually emit code with a source map instead of only having the option of using the combination of `sucrase` for compilation and `tsc` for typechecking. I'm not proposing we actually *use* just `tsc` for production compilation (its slow), but I'd like to be able to do something like `node build --dev` or `node build --sourcemap` so that I can then run `node -r source-map-support/register foo.js` during testing to get a better debugging experience (and maybe prove that source maps are awesome and that we should be looking to use them in the client to allow for proper minification etc).

I've shared the motivation of this change as context, but **I think this change is desirable for its own merits** and hopefully we can not get sidetracked on talking about the motivation (modularization, possible source map support) at this point in time.

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -13,6 +13,8 @@
 
 'use strict';
 
+/** @type {typeof import('../sim/battle-stream').BattleStream} */
+const BattleStream = require(/** @type {any} */ ('../.sim-dist/battle-stream')).BattleStream;
 /** @type {typeof import('../lib/fs').FS} */
 const FS = require(/** @type {any} */('../.lib-dist/fs')).FS;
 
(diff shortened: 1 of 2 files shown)


scheibo:
Slightly less sure about `eval` because you might theoretically want it for `pokemon-showdown simulate-battle` debugging, but that could just also depend on `RoomBattleStream` as well.

Slayer95:
> Slightly less sure about eval because you might theoretically want it for pokemon-showdown simulate-battle debugging

A better approach would probably be to trigger a ``debugger`` breakpoint after the input log is processed, so this shouldn't be a concern.

Zarel:
Doesn't TypeScript support incremental compilation now? It might not even be slow!

PR #5429, opened 2019-04-09
Use Proxy wherever applicable

Kaiepi:
`server/rooms.js`, `server/users.js`, and `server/loginserver.js` are the modules to my knowledge that mutate functions and objects used for `module.exports` in ways I don't particularly like. This aims to remove those mutations by exporting a `Proxy` instead. A side effect of this is properties of `Rooms`, `Users`, and `LoginServer` will be immutable (unless that's unwanted).

This currently fails tslint because `RoomGame` and `RoomGamePlayer` don't get recognized as superclasses of various chat plugin classes. I have no clue why. Some help regarding that would be appreciated so I hopefully don't have to put `// @ts-ignore` before 234 lines of code.

This is just a draft PR since I'm not sure what @Zarel's thoughts are about this.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -85,7 +85,7 @@
 		"init-declarations": "off",
 		"no-shadow": "off",
 		"no-label-var": "error",
-		"no-restricted-globals": ["error", "Proxy", "Reflect", "Symbol", "WeakSet"],
+		"no-restricted-globals": ["error", "Reflect", "Symbol", "WeakSet"],
 		"no-shadow-restricted-names": "error",
 		"no-undef": ["error", {"typeof": true}],
 		"no-undefined": "off",
--- a/dev-tools/global.d.ts
+++ b/dev-tools/global.d.ts
@@ -11,7 +11,8 @@ import {PRNG as PRNGType, PRNGSeed as PRNGSeedType} from './../sim/prng';
 import {Side as SideType} from './../sim/side';
 import {TeamValidator as TeamValidatorType} from'./../sim/team-validator';
 
-import RoomsType = require('./../server/rooms');
+import __RoomsType__ = require('./../server/rooms');
+type RoomsType = Proxy<__RoomsType__>
 import RoomlogsType = require('./../server/roomlogs');
 import LadderStoreType = require('./../server/ladders-remote');
 import LaddersType = require('./../server/ladders');
@@ -70,10 +71,12 @@ declare global {
 	const BasicRoom: typeof RoomsType.BasicRoom
 	const BasicChatRoom: typeof RoomsType.ChatRoom
 	const RoomGame: typeof RoomsType.RoomGame
+	const RoomGamePlayer: typeof RoomsType.RoomGamePlayer
 	const RoomBattle: typeof RoomsType.RoomBattle
 	const Rooms: typeof RoomsType
 	const Roomlogs: typeof RoomlogsType
 	const Roomlog: typeof RoomlogsType.Roomlog
+	type Room = GlobalRoom | ChatRoom | GameRoom
 
 	// users
 	const Users: typeof UsersType
--- a/dev-tools/globals.ts
+++ b/dev-tools/globals.ts
@@ -19,7 +19,20 @@ declare let Monitor: typeof import("../server/monitor");
 
 declare let LoginServer: typeof import('../server/loginserver');
 
-// type RoomBattle = AnyObject;
+let __RoomsType__: typeof import("../server/rooms")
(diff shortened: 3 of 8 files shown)


scheibo:
> `server/rooms.js`, `server/users.js`, and `server/loginserver.js` are the modules to my knowledge that mutate functions and objects used for `module.exports` in ways I don't particularly like.

`sim/team-validator.ts` as well I think?

---

I'd like to understand this pattern and why its used at all in the first place. I'm willing to bet there's a more straightforward way of accomplishing something similar without needing to resort to `Object.assign` on a `function` or reaching into the tool chest of esoteric language power tools like `Proxy`. I've written a heck of a lot more non-JS code than JS, so maybe I'm just unenlightened as to why this `Object.assign`-on-`exports` business is required, but it strikes me as odd that there would be a need for it in Javascript when I've never reached for it in my own code.

Kaiepi:
> ...or reaching into the tool chest of esoteric language power tools like `Proxy`. I've written a heck of a lot more non-JS code than JS, so maybe I'm just unenlightened as to why this Object.assign-on-exports business is required, but it strikes me as odd that there would be a need for it in Javascript when I've never reached for it in my own code.

`Proxy` is used a lot in Perl 6, though in that language it only affects getting/setting values, rather than the slew of properties they have in JavaScript

Slayer95:
... Have you at least measured the performance impact of this change?

xfix:
???  This is JavaScript, not Perl 6.

Have you considered using getters?

```javascript
module.exports = {
    get rooms() {
        return rooms;
    },
};
```

.eslintrc
```diff
@@ -85,7 +85,7 @@
 		"init-declarations": "off",
 		"no-shadow": "off",
 		"no-label-var": "error",
-		"no-restricted-globals": ["error", "Proxy", "Reflect", "Symbol", "WeakSet"],
+		"no-restricted-globals": ["error", "Reflect", "Symbol", "WeakSet"],
```
Slayer95: FWIW, banning ``Reflect`` but not ``Proxy`` makes zero sense.

Kaiepi: Oh, right

Kaiepi:
> ... Have you at least measured the performance impact of this change?

...ouch, calling `Rooms.get('lobby')` is about 10x slower with `Proxy`.

> Have you considered using getters?

I have, but as PS' code stands `Rooms` and `Users` need to be callable.

Maybe `Rooms` and `Users` should just not be callable and calls to them should be replaced with `Rooms.get`/`Users.get`. I'll test that out to see if that performs better tomorrow

Zarel:
I don't think your taste issues are good reasons to use Proxy here.

Once we migrate to TypeScript, TypeScript will be able to enforce their immutability. Until then, all this is overkill and results in worse readability _and_ worse performance.

xfix:
@Kaiepi 

Uhm, a getter CAN return a function.

```javascript
> ({get a() { return x => x + this.one }, one: 1}).a(2)
3
```

scheibo:
> I'd like to understand this pattern and why its used at all in the first place

No one has bothered to answer this yet, from what I can tell. Is it just sugar? Is it really worth using the `Object.assign` pattern at all?

Slayer95:
It's just sugar because @Zarel believes that ``Users(id)`` and ``Rooms(id)`` is a nicer API than ``Users.get(id)`` / ``Rooms.get(id)``.

Thumbs up if you *don't* think so.

scheibo:
Not only do I not think so, I think its actively *worse* because I don't know what its doing. `Users.get` is more explicit.

Kaiepi:
> I don't think your taste issues are good reasons to use Proxy here.
> Once we migrate to TypeScript, TypeScript will be able to enforce their immutability. Until then, all this is overkill and results in worse readability and worse performance.

...which is why I'm looking into using `Users.get`/`Rooms.get` instead now. There is a reason I didn't finish writing this before getting other people's thoughts first.

> Uhm, a getter CAN return a function.

I know. The issue is `Users`/`Rooms` themselves being functions with properties added on to them. I don't see how getters would completely fix having to use `Object.assign` with them.

Zarel:
> It's just sugar because @Zarel believes that `Users(id)` and `Rooms(id)` is a nicer API than `Users.get(id)` / `Rooms.get(id)`.
> 
> Thumbs up if you _don't_ think so.

:(

It's definitely really nice when I'm using the main REPL.

Slayer95:
Closing. Further discussion can be held in #5432

PR #5430, opened 2019-04-09
Update dependencies

scheibo:
Motivated by `--incremental` support on TS 3.4, only to try running it and notice no improvement. :( Turns out that since we use `noEmit` it doesnt work:  https://github.com/Microsoft/TypeScript/issues/30661. There are some [other features of 3.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html) which should prove useful though, but it looks like we'll need to continue to wait for faster typechecking.

Changes to `lib/streams.ts` are due to https://github.com/DefinitelyTyped/DefinitelyTyped/commit/ff1b97f5229632971c4f2927686fde76457bd5a5

### DIFF
--- a/lib/streams.ts
+++ b/lib/streams.ts
@@ -12,6 +12,8 @@
 
 const BUF_SIZE = 65536 * 4;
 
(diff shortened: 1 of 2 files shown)


Zarel:
Weren't we going to use `tsc` as our main compiler?

scheibo:
> Weren't we going to use `tsc` as our main compiler?

One day

Issue #2444, opened 2016-03-13
Ideas for new developers


Zarel:
This is an issue of suggestions for things people who want to start contributing to Pokémon Showdown can work on.

First, the Development room is the right place to discuss working on new features: http://psim.us/dev

- **Fix mechanics bugs** – #2367 tracks the mechanics bugs we're aware of.

- **Fixes to text hiding** – Text hiding in `/roomban`, `/hidetext`, etc currently work by checking for user objects, which expire after 40 minutes of being offline. There are several possible improvements here:
  1. Actually clear the text from the roomlog, so users joining after the text is hidden don't see it at all. This is useful if the text being hidden is a secret, and also saves bandwidth and room clutter.
  2. To support very inactive rooms, scan the roomlog for the passed username, as well as checking for user objects.
  3. To support users who change usernames multiple times, scan the roomlog for the user's entire `prevNames` list.

- **Move timer code from rooms.js to simulator.js** – A pretty straightforward refactor.

- **Rewrite sockets.js in Go** – You'll probably need https://github.com/igm/sockjs-go – this is a pretty hard task but doesn't need anyone particularly familiar with the codebase. Node.js's support for WebSocket seems to have multiple bugs which are very hard to fix (see nodejs/node#3072 and nodejs/node#5469) so changing languages may be easier. Fortunately, PS's architecture is multi-process which makes it possible to change language for one process without having to rewrite the entire thing from scratch.

- **Refactor tournaments to use RoomGamePlayer** – See #2331 – A self-contained refactor.

robotmayo:
Whats the current status on the Golang rewrite? I'm a Javascript dev looking to get into Golang. If possible id like to assist whoever is leading that.

Zarel:
@robotmayo No one's started on that yet. Feel free to try it yourself. In addition to golang, you'll need to be able to do IPC between Node and Go, which shouldn't be too difficult. You may be interested in https://github.com/jgranstrom/gonode

Joimer:
@robotmayo Maybe @xfix Would be interested in that (sorry if you aren't, xfix :)). If you start the project, I'd like to follow it, as I'd like to get more into Golang as well.

Slayer95:
@Zarel, I see the list randomly growing with no notification.
Each item should probably be its own issue.

Zarel:
I think that's unnecessary, but I guess I can comment every time I add something.

acz13:
[...]

PR #5431, opened 2019-04-10
Migrate tests to Jest

scheibo:
> * **Migrate tests to Jest** – Jest (https://jestjs.io/) supports parallel test-running, which would be nice to have over Mocha. Most of its API surface is similar, so this shouldn't take too long.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/issues/2444_

Did some investigation here. Have some failing tests still, but I figured I'd dump what I've figured out so far:

1) Jest has a pretty similar API, but since it doesn't use `this` we could switch `function () { ... }` to `() => { ... }` which is the preferred style. I didn't bother with this.
2) A lot of our tests do weird things with timeouts, possibly due to them getting hit by `Dex` loading? Why do half of these fiddle with timeouts? I can see why team generator might, but basic tests for abilities or moves? Smells like a bug to me.
3) Jest's test runner is much noisier by default, and we'd probably need to write our own/depend on a new one. The biggest difference is that its logging even on success (which I'm OK with), but this means we get tons of:

   ```
   PASS  test/simulator/abilities/roughskin.js
   ● Console
 
     console.log server/monitor.js:135
       NEW GLOBAL: global
     console.log server/monitor.js:135
       NEW CHATROOM: lobby
     console.log server/monitor.js:135
       NEW CHATROOM staff
   ```

    for each test. Maybe we could stub out `Monitor.notice` etc in the `beforeAll`?

4) Hot take: Jest will be slower. jest Is what my computer looks like running Jest, and I'm pretty sure most PS! developers don't have that kind of hardware sitting under their desk. To make matters worse - even on my box, its slower (13-14s instead of 10-12s)! I'm guess this is possibly due to the verbosity of the logging (writing that much to the console is slow), but I'd bet even on my computer its not going to be much faster than Mocha due to `Dex` loading. `Dex` loading is super slow, and if we're naively running tests in parallel processes the way Jest does, we're paying the price multiple times. The optimal way of running PS is to spin up a number of processes roughly equal to the number of cores (maybe NUM_CORES-1, to allow for the GC thread), and then run `Dex` loading once in each one, and then run multiple battles in a loop (incidently, the PS server does this :P)

5) Jest doesn't like `Dex` loading, it seems to cause failures when a test is `skip`ped:

```
ReferenceError: You are trying to `import` a file after the Jest environment has been torn down.

      at ModdedDex.includeFormats (.sim-dist/dex.js:1643:17)
      at ModdedDex.get formats [as formats] (.sim-dist/dex.js:246:10)
      at ModdedDex.loadData (.sim-dist/dex.js:1584:80)
      at ModdedDex.includeData (.sim-dist/dex.js:1550:10)
      at Object.<anonymous>.process.nextTick (server/chat-plugins/info.js:2549:7)
.sim-dist/dex.js:1651
      throw new TypeError(`Exported property 'Formats' from "./config/formats.js" must be an array`);
      ^

TypeError: Exported property 'Formats' from "./config/formats.js" must be an array
    at ModdedDex.includeFormats (.sim-dist/dex.js:1651:13)
    at ModdedDex.get formats [as formats] (.sim-dist/dex.js:246:10)
    at ModdedDex.loadData (.sim-dist/dex.js:1584:80)
    at ModdedDex.includeData (.sim-dist/dex.js:1550:10)
    at Object.<anonymous>.process.nextTick (server/chat-plugins/info.js:2549:7)
    at processTicksAndRejections (internal/process/task_queues.js:79:9)
```

All in all, useful exercise, but not super sure about how much we want to invest in Jest?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,7 +15,7 @@
 		"Config": false, "Monitor": false, "toId": false, "Dex": false, "LoginServer": false,
 		"Users": false, "Punishments": false, "Rooms": false, "Verifier": false, "Chat": false,
 		"Tournaments": false, "Dnsbl": false, "Sockets": false, "TeamValidator": false,
-		"TeamValidatorAsync": false, "Ladders": false
+		"TeamValidatorAsync": false, "Ladders": false, "beforeAll": false, "afterAll": false
 	},
 	"extends": "eslint:recommended",
 	"rules": {
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,7 @@
+'use strict';
+
+module.exports = {
+	testEnvironment: 'node',
+	setupFilesAfterEnv: ['<rootDir>/test/main.js'],
+	testMatch: ['<rootDir>/test/simulator/**/*.js', '<rootDir>/test/application/*.js', '<rootDir>/test/chat-plugins/*.js', '<rootDir>/test/dev-tools/*.js'],
+};
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "The server for the Pokémon Showdown battle simulator",
   "version": "0.11.2",
   "dependencies": {
+    "jest": "^24.7.1",
     "probe-image-size": "^4.0.0",
     "sockjs": "0.3.19",
     "sucrase": "^3.10.1"
@@ -21,7 +22,7 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "test": "eslint --cache . && tslint --project . && node build && mocha && tsc",
+    "test": "eslint --cache . && tslint --project . && node build && jest && tsc",
     "tsc": "tsc",
     "lint": "eslint --cache ."
   },
@@ -63,7 +64,6 @@
     "@types/sockjs": "^0.3.31",
     "eslint": "^5.16.0",
     "husky": "^1.1.2",
-    "mocha": "^6.1.2",
     "tslint": "^5.15.0",
     "typescript": "^3.4.2"
   }
--- a/test/application/ladders-matchmaker.js
+++ b/test/application/ladders-matchmaker.js
@@ -5,6 +5,9 @@ const assert = require('assert');
 global.Ladders = require('../../server/ladders');
 const {Connection, User} = require('../../dev-tools/users-utils');
 
+let p1, p2;
+let s1, s2;
+
 describe('Matchmaker', function () {
 	const FORMATID = 'gen7ou';
 	const addSearch = (player, rating = 1000, formatid = FORMATID) => {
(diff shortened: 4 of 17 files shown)


xfix:
If I had to guess Jest is slower because it runs multiple processes, which prevents the JIT from optimizing too much, and loading takes a while. 48 cores is kinda a lot.

scheibo:
> *That is because Chrome’s build system is smart enough to spawn more processes if you have more cores, which means that there are more terminating processes fighting over the global lock.*

'smart' is used generously here, because clearly its resulting in worse results. Having more cores can definitely slow things down if they're not used properly though.

> If I had to guess Jest is slower because it runs multiple processes, which prevents the JIT from optimizing too much, and loading takes a while

Yup. I'm curious though, when I get home I'll try it on my laptop which I think only has maybe 4 cores. I think its still going to be slower because the processes Jest spawns wont reuse `Dex` loading or benefit from caching etc.

That being said: https://github.com/facebook/jest/issues/6694

Slayer95:
In my machine, with a [barebones initialization](https://gist.github.com/Slayer95/b78c77c15112c8da0c4ebd5a42236c1d), and running only the ``simulator/`` tests, ``mocha`` uses ``~23 seconds`` in my machine (2 cores [1]).

For a similar setup, ``jest`` uses ``~140 seconds``. This duration doesn't change noticeably with ``--maxWorkers=2``, but it does increase a lot with ``maxWorkers=4`` (despite Jest's estimates that each of those would result in halving the run time).

On the other hand, switching to a very simple runner which condenses all the simulator test suites into [4 suites]() reduces the total time to ``~40 seconds``. If it's condensed into [2 suites](), the total time is ~30 seconds.... It seems to be just slightly faster than the case of 1 suite (~33 seconds), but I don't think that's a conclusive result....

Overall, it's worrying that this results in a performance degradation of anywhere from 50% to 500%.

-------------------

[1] Initially I thought I had 4 cores, which is why I also tested for the cases o 4 workers / 4 suites.....

scheibo:
TL;DR: "Jest seems slow". Do we want to even bother to try and make it work? I'm happy to close this (once @Zarel weighs in) and remove the idea from #2444.

Slayer95:
> Jest has a pretty similar API, but since it doesn't use this we could switch function () { ... } to () => { ... } which is the preferred style. I didn't bother with this.

ESLint's rule ``prefer-arrow-callback`` has an autofixer. Just add it to ``test/.eslintrc`` and run it with ``--fix``.

> A lot of our tests do weird things with timeouts, possibly due to them getting hit by Dex loading? Why do half of these fiddle with timeouts? I can see why team generator might, but basic tests for abilities or moves? Smells like a bug to me.

Those sometimes to often timed out in my machine. The only explanation I could fathom was that they ran heavy events, but taking several seconds was always odd. That said, since the sim is at least 50% more performant nowadays, those custom timeouts might not be needed anymore.

> we get tons of:

PASS  test/simulator/abilities/roughskin.js
● Console

  console.log server/monitor.js:135
    NEW GLOBAL: global
  console.log server/monitor.js:135
    NEW CHATROOM: lobby
  console.log server/monitor.js:135
    NEW CHATROOM staff
for each test. Maybe we could stub out Monitor.notice etc in the beforeAll?

We only need to tweak ``Config.loglevel``. Furthermore, we don't actually need to start up the whole server for each test. It's possible to start just the sim itself nowadays. Of course, application tests still need the server, so that's the scope of the fix.

> Jest doesn't like Dex loading, it seems to cause failures when a test is skipped:

It doesn't like us loading Dex in ``process.nextTick()`` from chat plugins. We can probably wait for a ``setImmediate()`` to resolve in the ``beforeAll()`` for application/plugins tests.

scheibo:
> ESLint's rule `prefer-arrow-callback` has an autofixer. Just add it to `test/.eslintrc` and run it with `--fix`.

Yeah, `jest-codemods` does this as well, but I didn't want to cause so much churn.

> Those sometimes to often timed out in my machine. The only explanation I could fathom was that they ran heavy events, but taking several seconds was always odd. That said, since the sim is at least 50% more performant nowadays, those custom timeouts might not be needed anymore.

I think this is worth followup. I don't know that I'm the best to try it though, because I think the tests run the fastest on my machines. Maybe @KrisXV can help us here?

> We only need to tweak `Config.loglevel`. 

Ahh, TIL.

> Furthermore, we don't actually need to start up the whole server for each test. It's possible to start just the sim itself nowadays. Of course, application tests still need the server, so that's the scope of the fix.

I was going to say we should followup on that (not starting the server), but I guess it currently doesnt matter if we're sticking with mocha because the application tests run in the same process as the sim tests anyway.

> It doesn't like us loading Dex in `process.nextTick()` from chat plugins. We can probably wait for a `setImmediate()` to resolve in the `beforeAll()` for application/plugins tests.

Good call. Though once again, not necessarily worth changing if we're not going to use Jest.

Thanks @Slayer95 for diving into this! You definitely did a better job than my naive attempt!

Zarel:
I'd agree with closing and removing the idea. I only ever wanted Jest because it seemed like parallel tests would be a speed boost, but I did give up on using Jest on client after benchmarking. It looks like it doesn't offer anything to be worth it.

scheibo:
I've been using it for all my personal libraries because I thought it would be better as well, but after now benchmarking on PS im reevaluating. Anyway, will close and remove the suggestion, thanks!

PR #5432, opened 2019-04-10
Remove Object.assign(function) pattern from TeamValidator{Async}

scheibo:
> It's just sugar because @Zarel believes that ``Users(id)`` and ``Rooms(id)`` is a nicer API than ``Users.get(id)`` / ``Rooms.get(id)``.
>
> Thumbs up if you *don't* think so.

_Originally posted by @Slayer95 in https://github.com/Zarel/Pokemon-Showdown/pull/5429#issuecomment-481463245_

This is my counter argument. I even added some sugar back in 111f26b because in that case `new TeamValidatorAsync.TeamValidatorAsync(...)` looks worse than `TeamValidatorAsync.create(...)` IMO. I don't think its necessary for `TeamValidator` and `new TeamValidator(...)` is shorter than `TeamValidator.create(...)` anyway, but I can maybe see adding a factory method there as well for consistency. Either way, I believe this pattern is more straightforward than the previous one, and having to type `.create` or `new ` doesn't seem like such a large imposition as to merit adding the confusing `Object.assign(function)` pattern.

I'm happy to migrate `Rooms`/`Users` in a separate PR, I just wanted to throw something concrete out there. (I only really planned to change `TeamValidator` as a strawman, but `TeamValidatorAsync` is such a close analog that it seems wrong to have them both have different APIs. Which is why I'd also be OK with a `create` method on `TeamValidator`, FWIW)

### DIFF
--- a/dev-tools/global.d.ts
+++ b/dev-tools/global.d.ts
@@ -47,7 +47,6 @@ declare global {
 	const Side: SideType
 	const Sim: typeof SimType
 	const TeamValidator: typeof TeamValidatorType
-	const Validator: typeof TeamValidatorType.Validator
 	const BattleStream: BattleStreamType.BattleStream
 
 	// dex data
--- a/dev-tools/globals.ts
+++ b/dev-tools/globals.ts
@@ -3,7 +3,7 @@ type Field = import('./../sim/field').Field
 type ModdedDex = typeof import('./../sim/dex')
 type Pokemon = import('./../sim/pokemon').Pokemon
 type Side = import('./../sim/side').Side
-type Validator = import('./../sim/team-validator').Validator
+type Validator = import('./../sim/team-validator').TeamValidator
 
 type PageTable = import('./../server/chat').PageTable
 type ChatCommands = import('./../server/chat').ChatCommands
--- a/dev-tools/globals.ts
+++ b/dev-tools/globals.ts
@@ -3,7 +3,7 @@ type Field = import('./../sim/field').Field
 type ModdedDex = typeof import('./../sim/dex')
 type Pokemon = import('./../sim/pokemon').Pokemon
 type Side = import('./../sim/side').Side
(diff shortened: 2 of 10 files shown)


Zarel:
I originally designed the API as `TeamValidator(formatid)` instead of `new TeamValidator(formatid)` because it abstracts out any caching that may or may not be happening.

In other words, I think of it as "get a team validator for this format" rather than "create a team validator for this format".

Zarel:
I think of `FS` the same way, incidentally – the fact that it creates a new object in the background is an implementation detail, not something I want as part of the API itself. And `Users` and `Rooms` are obviously getters.

scheibo:
>  I think of it as "get a team validator for this format" rather than "create a team validator for this format".

OK, cool. Let's call the factory method `get` then.

Zarel:
If you fix these conflicts, I'll totally merge this change.

Zarel:
After a few more conversations, I've come to accept that my fancy APIs aren't good enough for the rest of you. >:(

scheibo:
_That moment when your desire to 'have < 25 issues/PRs' trumps your design ethos_

Zarel:
> _That moment when your desire to 'have < 25 issues/PRs' trumps your design ethos_

Hah, I'd be focusing on way more than this one PR if that was the case. I was probably going to agree to this eventually anyway, I just needed to think about it some more. I eventually gave in while biking with Bumba. But I was biking so I couldn't do anything about it immediately, and then a bunch of other stuff happened and I forgot about this.

I remembered again after trying to trim my to-do list today, and finding this on it.

scheibo:
It was less work to just redo this change from scratch than it was the figure out how to deal with all the merge conflicts to salvage this PR.

**Does your approval of this PR indicate you'd also be willing to accept `User` and `Room` migrating away from this pattern as well?**

Zarel:
Yes.

PR #5433, opened 2019-04-10
Make Pokemon.moves a read only array

urkerab:
We're still trying to `push` uselessly to it in a couple of places.

I don't know if you want the Typescript annotation or whether it's the best way of annotating the requirement.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -382,7 +382,7 @@ export class Pokemon {
 		this.m = {};
 	}
 
-	get moves() {
+	get moves(): readonly string[] {
 		return this.moveSlots.map(moveSlot => moveSlot.id);
 	}
 
@@ -923,7 +923,6 @@ export class Pokemon {
 				used: false,
 				virtual: true,
 			});
-			this.moves.push(toId(moveName));
 		}
 		let boostName: BoostName;
 		for (boostName in pokemon.boosts) {
--- a/data/mods/ssb/moves.js
+++ b/data/mods/ssb/moves.js
@@ -1405,7 +1405,6 @@ let BattleMovedex = {
 						used: false,
 						virtual: true,
 					});
-					pokemon.moves.push(move.id);
 				}
 			},
 			onModifySpAPriority: 1,
@@ -1439,7 +1438,6 @@ let BattleMovedex = {
 						used: false,
 						virtual: true,
 					});
-					pokemon.moves.push(move.id);
 				}
 			},
 		},
@@ -1660,7 +1658,6 @@ let BattleMovedex = {
 							used: false,
 							virtual: true,
 						});
-						pokemon.moves.push(move.id);
 					}
 				}
 			},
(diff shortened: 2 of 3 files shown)


Slayer95:
I am not sure about this...
In theory, it's *okay* to do the following:

```js
const extendedMoves = pokemon.moves;
extendedMoves.push(id);
doSomething(extendedMoves);
```

But we would also be banning that operation...

scheibo:
I think given the small number of `moves` a Pokemon has, having them perform another copy in that scenario would not be excessive?

Slayer95:
> perform another copy

So does ``.slice()`` called on a ``ReadonlyArray<T>`` return a mutable ``Array<T>`` ?

I'd be more inclined to use an ``as`` cast.

scheibo:
I'm not sure, `ReadonlyArray` seems pretty aspirational with Typescript tbh. In my experience, its not always perfect at checking these things -its not like an `ImmutableList` in Guava, or even something that went ended up calling `Object.freeze` on (which I'm not suggesting).

Zarel:
I mean, it makes more sense to use `concat` than `slice`.

Zarel:
We could also just deprecate `pokemon.moves` entirely.

sim/pokemon.ts
```diff
@@ -382,7 +382,7 @@ export class Pokemon {
 		this.m = {};
 	}
 
-	get moves() {
+	get moves(): ReadonlyArray<string> {
```
Zarel:
```suggestion
	get moves(): readonly string[] {
```

urkerab: With this change I get tons of errors, starting with:

ERROR: sim/pokemon.ts:385:23 - block is empty

Am I doing something wrong?

scheibo: I don't know that Zarel's suggestion is valid pre Typescript 3.4 (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html). You probably need to update your Typescript version.

Zarel: I thought we already updated our TypeScript version?

Zarel:
Overall, though, I think this is the correct direction. It's not like we're using the mutability of copies anywhere here.

sim/pokemon.ts
scheibo: We did, I dont know if @urkerab did (is it automatic? if the repo updates `package.json`, will running `npm test` cause it to download the new deps?)

Zarel: Oh, it's not automatic, yeah, you'll all need to manually rerun `npm i`.

urkerab: (I had to downgrade `node` so that I could upgrade `npm` and then re-upgrade `node` so I could `npm install` and `npm test` but it all works now, phew!)

Zarel:
This pull request **introduces 206 alerts** when merging b13438d59dbe322cba6cd6d2a48c0ee063d72ba4 into 3e735aa6ce89d88d4c1288cfe95aa055aaf24e0a - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-e91ade3429bc6ca6bb75da1f42522fe89296e064)

**new alerts:**

* 206 for Syntax error

---

PR #5436, opened 2019-04-11
Update test folder structure to match the source code

Slayer95:
Folders renamed:
``simulator`` → ``sim``
``application`` → ``server``
``chat-plugins`` → ``server/chat-plugins``

Files renamed:
``server/ladders-matchmaker.js`` → ``server/ladders.js``
``server/simulator.js`` → ``server/room-battle.js``
``server/team-validator.js`` → ``sim/team-validator.js``
``sim/misc/choice-parser.js`` → ``sim/choice-parser.js``
``sim/misc/decisions.js`` → ``sim/decisions.js``
``sim/misc/dex.js`` → ``sim/dex.js``
``sim/misc/eventemitter.js`` → ``sim/events.js``

(OP updated in real time)


Slayer95:
This is gonna conflict with #5397. I guess we should finish & merge that first.

test/mocha.opts
```diff
@@ -1,4 +1,4 @@
-test/main.js test/simulator/**/*.js test/application/*.js test/chat-plugins/*.js test/dev-tools/*.js
+test/main.js test/dev-tools/**/*.js test/server/**/*.js test/sim/**/*.js
```
scheibo: I placed dev-tools last tactically so it wouldn't be the first test run and have to pay the `Dex` loading penalty on the test which is already on the slow side. The real answer is to probably disable timeouts for those tests/set them to be long? They're currently `skip`ped though, so it doesn't really matter.

scheibo:
> If you'd prefer `application` → `server`, say the word.

I think this would be nice? If we're going to match, we should actually match!

test/mocha.opts
Slayer95: Oh, I see. We should preload ``Dex`` in ``main.js``.

Slayer95:
Right. Took the chance to update some files with outdated names / paths.

Slayer95:
I wonder if it would be better to rename the current ``test/simulator`` folder to `test/sim-integration``, so that we reserve ``test/sim`` for testing the actual ``sim/*.ts`` modules.

Slayer95:
Actually, it'd be better to rename the current test/simulator folder to test/sim-integration, so that we reserve test/sim for testing the actual sim/*.ts modules.

Updated ~

scheibo:
Hmm, those names don't make as much sense to me anymore? An individual move or ability seems like it would logically belong under `sim` (as theyre as close to a unit test as we have), but `sim-integration` seems to imply something higher level.

I think we should have the `sim/choice.js`, `sim/decisions.js`  (even though we they're not testing a `decisions.ts` file) etc still, but `sim/abilities/damp.js` etc seems more intuitive to me, even though there's no `sim/abilities` directory in the non-test tree. You could place the tests under a `data/abilities` path etc instead of `sim`, but I don't think that simplifies things. 

<i>As an aside: I think two of the issues at play here are that:

1) `battle.ts` is too monolithic (I believe we've proposed pulling out the choice or decision logic from the event system logic from the battle state logic from the field logic etc, at which point it probably *would* make sense to have each file have its own matching test)
2) We have non-data in the `data` sub directory, and the `data` subdirectory is not under `sim`.

While I'm in the process of thinking of better ways to solve those two issues, I don't think this PR is the place to do so. Thus my recommendation of trying to go for what is the most 'logical' place for the tests (ie. I don't think its necessarily logical to look for unit tests under a path that contains 'integration') as opposed to whatever maps best the current directory structure (which somewhat ironically may seem to be inconsistent with https://github.com/Zarel/Pokemon-Showdown/pull/5436#issuecomment-482352614)</i>

**TL;DR:** I'd remove `sim-integration` and just put everything under `sim`, because the distinction seems more confusing, even if it is more 'correct'.

Slayer95:
Great analysis.

While a more obvious structure would be ``test/data/(abilities|moves|items)/*.js``, I am not happy with that idea because I am not convinced of their being proper unit tests for the *data* (one proper such test would be the BF sets test in #5397).

Therefore, I thought ``sim-integration`` would be a good name. However, the tests are indeed much closer to the concept of unit tests than to actual integration tests.

I'll roll-back to ``test/sim/unit.js`` + ``test/sim/(abilities|items|moves|moves).js`` for now.

Suggestions welcome!

Zarel:
I like all the updates thus far. Good luck with the merge conflict. ;)

Slayer95:
> I like all the updates thus far. Good luck with the merge conflict. ;)

It turned out to be a seamless 3-way merge.

PR #5437, opened 2019-04-12
Test: Implement `npm run fulltest` and flagging tests as slow

Slayer95:
- Tests flagged with (slow) will no longer be run in `npm test`.
- Travis CI will now run `fulltest`.
- Unskip slow dev-tools tests and flag as such.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "test": "eslint --cache . && tslint --project . && node build && mocha && tsc",
+    "test": "eslint --cache . && tslint --project . && node build && mocha -i -g \"\\(slow\\)\" && tsc",
+    "fulltest": "eslint --cache . && tslint --project . && node build && mocha && tsc",
     "tsc": "tsc",
     "lint": "eslint --cache ."
   },
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ cache:
 notifications:
   webhooks: https://www.travisbuddy.com/
   on_success: change
+script: npm run fulltest
 travisBuddy:
   insertMode: update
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -21,7 +21,8 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "test": "eslint --cache . && tslint --project . && node build && mocha && tsc",
+    "test": "eslint --cache . && tslint --project . && node build && mocha -i -g \"\\(slow\\)\" && tsc",
+    "fulltest": "eslint --cache . && tslint --project . && node build && mocha && tsc",
     "tsc": "tsc",
     "lint": "eslint --cache ."
```
scheibo: What about changing `"lint"` to `"eslint ---cache .&& tslint --project ."`? I would then do:

```json
"pretest": "npm run lint",
"test": "mocha -i -g \"\\(slow\\)\"",
"posttest": "npm run tsc",
"fulltest": "npm run pretest && mocha && npm run posttest",
```

`pretest` and `posttest` are special cased by `npm` to automatically run before `npm test`, but we'd need to specify them manually for `posttest`. The main motivation behind this change would be to remove duplication between the `"test"` and `"posttest"` scripts.

Zarel:
Thanks! Um, what order should I be merging this and #5397 in?

package.json
```diff
@@ -21,7 +21,8 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "test": "eslint --cache . && tslint --project . && node build && mocha && tsc",
+    "test": "eslint --cache . && tslint --project . && node build && mocha -i -g \"\\(slow\\)\" && tsc",
```
Zarel: I like everything about this except that `mocha` by itself runs slow tests.

Zarel: But I guess there's no easy solution here.

Zarel:
(After code reviewing it became obvious which order I should be merging them)

PR #5438, opened 2019-04-12
Include diffs of all User settings in |updateuser|

### LINKED ISSUE #5424: Rework User settings + |updateuser| protocol
> I also wonder if we should start bundling user settings together, instead of having them on the top level of `User`. We have a bunch more (although most of them probably shouldn't be sent here).

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5422_

> I'm not entirely sure I want settings in `|updateuser|`
> there's an argument to be made that we don't want to resend all user information whenever a setting is updated
> it's fine right now but we might regret this one day when we have like fifty settings
> although I guess if that happens we can revisit then
> and probably just send the JSON object as a delta

scheibo:
Fixes #5424. The client code is already able to handle diffs.

- in certain scenarios I think we want to continue to send all the values instead of just the diffs
- `User.lastConnected` gets updated in `User.markInactive()`, but we cant send an `|updateuser|` at that point
- `User.isStaff` and `User.isSysop` usually get updated right before we'd be sending all the values anyway, so I'm trying to avoid doing a double `update`
- WTF is going on with `specialNextBattle`? I only see it being read, not being written?

    ```
    server/rooms.js:1615:               /** @type {(User & {specialNextBattle: boolean})[]} */
    server/rooms.js:1638:               if (players.some(user => user.specialNextBattle)) {
    server/rooms.js:1639:                       const p1Special = players[0].specialNextBattle;
    server/rooms.js:1642:                               if (user.specialNextBattle !== p1Special) {
    server/rooms.js:1643:                                       mismatch += ` vs. "${user.specialNextBattle}"`;
    ```

### DIFF
--- a/server/users.js
+++ b/server/users.js
@@ -434,6 +434,8 @@ class Connection {
 
 /** @typedef {[string, string, Connection]} ChatQueueEntry */
 
+const SETTINGS = ['isSysop', 'isStaff', 'blockChallenges', 'blockPMs', 'ignoreTickets', 'lastConnected', 'inviteOnlyNextBattle'];
+
 // User
 class User extends Chat.MessageContext {
 	/**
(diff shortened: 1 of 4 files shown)


Zarel:
`specialNextBattle` is only written to in private chat plugins. It's intended to support special battle types.

Zarel:
It's intended to work similarly to `inviteOnlyNextBattle`, but instead of making the next battle invite-only, it gives the next battle a special property up to the developer to script.

server/users.js
```diff
@@ -1198,6 +1213,7 @@ class User extends Chat.MessageContext {
 			// We're not resetting .trusted/.autoconfirmed so those accounts
 			// can still be locked after logout.
 		}
+		// NOTE: can't do a this.update(...) at this point because we're no longer connected.
```
Zarel: This function should maybe be renamed `markDisconnected` for clarity.

scheibo:
> `specialNextBattle` is only written to in private chat plugins. It's intended to support special battle types.

I can't call `user.update('specialNextBattle')` if I can't see where it's written, so I think it might be out of scope for this change?

server/users.js
scheibo: Done

Zarel:
Looks good.

PR #5002, opened 2018-11-29
Add an AFK timer

bumbadadabum:

PR #5440, opened 2019-04-12
Ban '@' from usernames

scheibo:
> Step 1: ban `@` from usernames, everywhere.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown-Client/pull/1265#issuecomment-482610450_

From #5002.


Zarel:
Man, let me code review first. :|

First of all, this is inside the `Config.disablebasicnamefilter` block. This is a protocol change, not a basic namefilter. It belongs with the ban on starting symbols: always-on, non-optional.

bumbadadabum:
I... totally missed that line. Mostly because I really didn't think of anyone ever disabling this basic filter...

Issue #5337, opened 2019-03-22
Illegal moves in RBY/GSC randbats data


Slayer95:
- [x] [gen2] PERSIAN: Move 'icebeam' cannot be learnt
- [x] [gen2] TYPHLOSION: Move 'counter' cannot be learnt
- [x] [gen2] NOCTOWL: Move 'doubleedge' cannot be learnt
- [x] [gen2] LEDIAN: Move 'substitute' cannot be learnt
- [x] [gen2] JUMPLUFF: Move 'substitute' cannot be learnt
- [x] [gen2] AIPOM: Move 'substitute' cannot be learnt
- [x] [gen2] GIRAFARIG: Move 'substitute' cannot be learnt
- [x] [gen2] HOUNDOOM: Move 'substitute' cannot be learnt
- [x] [gen1] ZUBAT: Move 'screech' cannot be learnt
- [x] [gen1] OMANYTE: Move 'seismictoss' cannot be learnt

TheImmortal:
@xCrystal

Slayer95:
They are inactive in Github since September...

TheImmortal:
Uh, @ascriptmaster then.

Slayer95:
Gen 1 issues introduced in 892719ddbe2. There was no set information before.
Gen 2 issues introduced from 07bbdd0ab9 onwards. Set information before [here](https://github.com/Zarel/Pokemon-Showdown/blob/85c83be66685c436995bbc1a4df472401f5a8e74/mods/gen2/formats-data.js).

TheImmortal:
Gen 1 fixed in 99caab2

scheibo:
Was about to close, but unless I'm missing something, Gen 2 Houndoom still needs fixes? @mark7328?

PR #5442, opened 2019-04-13
Rework mocha fulltest configuration

scheibo:
_#justbikesheddingthings_

Optional follow up to #5437, maybe you guys are interested in some or none of the following:

- leverage [mocha's merging order](https://mochajs.org/#merging) with an impossible regex and `test/mocha.opts` to make `npx mocha` be the default which skips slow tests
- put `eslint` and `tslint` together under the `lint` script
- use standard [`posttest` and `pretest` hooks](https://docs.npmjs.com/misc/scripts)

In other news, @KrisXV's test is damn slow. How long do we think is OK for Travis? Can I run a full cycle of smokes tests instead of just the one canary one?

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,12 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "test": "eslint --cache . && tslint --project . && node build && mocha -i -g \"\\(slow\\)\" && tsc",
-    "fulltest": "eslint --cache . && tslint --project . && node build && mocha && tsc",
     "tsc": "tsc",
-    "lint": "eslint --cache ."
+    "lint": "eslint --cache . && tslint --project .",
+    "pretest": "npm run lint && npm run build",
+    "test": "mocha",
+    "posttest": "npm run tsc",
+    "fulltest": "npm run pretest && mocha -g '$^' && npm run posttest"
   },
   "husky": {
     "hooks": {
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,4 +1,5 @@
 test/main.js test/dev-tools/**/*.js test/server/**/*.js test/sim/**/*.js
+-i -g "\(slow\)"
 -R dot
 -u bdd
 --exit


Slayer95:
> How long do we think is OK for Travis? Can I run a full cycle of smokes tests instead of just the one canary one?

I'd like to push the test which revealed #5337 as well.

scheibo:
I think there's two things we need to take into account, 1) Travis timeouts and 2) how long we'd want to wait on commits.

> - When a job produces no log output for 10 minutes.
> - When a job on a public repository takes longer than 50 minutes.

_https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts_

Looks like a test case can be ~10 minutes long (before it adds a `.` for having finished). That being said, people will probably get upset if Travis takes > 10 minutes before they get results (LGTM latency already draws complaints).

WDYT? I think (1) is  less of a limiting factor than (2). With my 2 tests runnings more iterations plus the existing ~50s long test, I think we'd be around ~3-5 minutes? How long is your test? And do you agree that all the slow test combined should probably try to stay under 10 minutes?

Slayer95:
> That being said, people will probably get upset if Travis takes > 10 minutes before they get results (LGTM latency already draws complaints).

CI time should be a very small fraction of the overall turnover time of most PRs imo, granted the author runs `npm test`.

> How long is your test?
* cough cough * I seem to have overestimated its duration since I last ran it: just 3 seconds in my machine.......

scheibo:
I've decided to [run more cycles of my tests](https://github.com/Zarel/Pokemon-Showdown/blob/37310be828268354da2db0a88eaa69ecc8ef475b/test/dev-tools/runners.js) - doing a single exhaustion cycle of each supported metagame + 100 random battles only bumps the time up to '3m' (presumably somewhere between 3 and 4 minutes) on my laptop which still seems pretty palatable for `fulltest`.

> _cough cough_ I seem to have overestimated its duration since I last ran it: just 3 seconds in my machine.......

I feel we have _regular_ tests slower than that\*! Check it in!

_\*Maybe we should audit our existing tests and move some of the slow ones to `fulltest`?_

Zarel:
If it's 3 seconds, why is Kris's so much slower? I guess because Kris's is a full validation and yours is just a "does learnset contain move"

Slayer95:
> If it's 3 seconds, why is Kris's so much slower? I guess because Kris's is a full validation and yours is just a "does learnset contain move"

Yeah, we don't care whether the final sets are legal in randbats..

PR #5445, opened 2019-04-13
Fix misspecified event handler priorities

### LINKED ISSUE #5444: Misspecified event handler priorities
> [base] battery.onBasePowerPriority
[base] dryskin.onBasePowerPriority
[base] flowergift.onModifyAtkPriority
[base] heatproof.onBasePowerPriority
[base] thickfat.onModifyAtkPriority
[gen1] flowergift.onModifyAtkPriority
[gen1] thickfat.onBasePowerPriority
[gen1] mudsport.onBasePowerPriority
[gen1] watersport.onBasePowerPriority
[gen5] mudsport.onBasePowerPriority
[gen5] watersport.onBasePowerPriority
[gennext] thickfat.onModifyAtkPriority
[gennext] flowergift.onModifyAtkPriority
[gennext] heatproof.onBasePowerPriority
[ssb] thiccerfat.onModifyAtkPriority
[ssb] murkyambush.onBasePowerPriority

No corresponding event handler, or it should be ``on(Source|Foe|Ally|Any)EventPriority``

Slayer95:
Fixes #5444

Stuff I noticed:
- Battery is assigned a priority of 8. However, from 9635ea7d5643 I'd expect it to have a priority of 3.
- Dry Skin currently doesn't boost Lava Plume's friendly fire. Bug?

cc @Marty-D

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -235,7 +235,7 @@ let BattleAbilities = {
 	},
 	"battery": {
 		shortDesc: "This Pokemon's allies have the power of their special attacks multiplied by 1.3.",
-		onBasePowerPriority: 8,
+		onAllyBasePowerPriority: 8,
 		onAllyBasePower(basePower, attacker, defender, move) {
 			if (attacker !== this.effectData.target && move.category === 'Special') {
 				this.debug('Battery boost');
@@ -790,7 +790,7 @@ let BattleAbilities = {
 				return null;
 			}
 		},
-		onBasePowerPriority: 7,
+		onFoeBasePowerPriority: 7,
 		onFoeBasePower(basePower, attacker, defender, move) {
 			if (this.effectData.target !== defender) return;
 			if (move.type === 'Fire') {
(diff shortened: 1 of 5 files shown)


Marty-D:
V4's grouping of modifiers is more or less accurate to the actual chaining order, but since Battery's an Ability it's grouped with the other "user" Abilities. It also appears Helping Hand lands between move modifiers; we might want to overhaul this entire grouping concept. http://bbs10.aimix-z.com/mtpt.cgi?room=sonota&mode=view2&f=140&no=27 for reference.

Re: Dry Skin, looks like a holdover from before Doubles was even implemented, good catch!

Slayer95:
Hm is that list sorted? Otherwise, I can only find information regarding how the game performs rounding.

> Helping Hand lands between move modifiers; we might want to overhaul this entire grouping

Overhaul? Sounds like just splitting move modifiers in two stages?

Marty-D:
The list is sorted, yeah. By overhaul I mean even within any given "group", the modifiers should be chained in the listed order without race conditions.

Marty-D:
Oh heh, I hadn't even looked at what this PR changed because I was so distracted by V4's linked commit I'd never seen before. This is good! Other issues can be tackled another time.

PR #5449, opened 2019-04-15
Random Battle: Update Mandibuzz

ascriptmaster:
Removing Brave Bird prevents Banded sets.

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -6552,7 +6552,7 @@ let BattleFormatsData = {
 		tier: "LC",
 	},
 	mandibuzz: {
-		randomBattleMoves: ["foulplay", "knockoff", "roost", "taunt", "whirlwind", "toxic", "uturn", "bravebird", "defog"],
+		randomBattleMoves: ["foulplay", "knockoff", "roost", "taunt", "whirlwind", "toxic", "uturn", "defog"],
 		randomDoubleBattleMoves: ["knockoff", "roost", "taunt", "tailwind", "snarl", "uturn", "bravebird", "protect"],
 		eventPokemon: [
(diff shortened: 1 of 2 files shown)


TheImmortal:
Eh, I think removing Knock Off is the play. It's pretty much secondary to Foul Play on every analysis. Brave Bird is good so you aren't just setup bait for Fighting mons.

TheImmortal:
I mean ideally we'd reject Knock Off with Foul play anyway but that isn't bad on offensive mons like Yveltal

ascriptmaster:
I chose Brave Bird because while Knock Off is secondary to Foul Play on every analysis, Brave Bird isn't on any analysis outright. That said, 1760 usage has both at around the same usage, although the 1500 Knock Off usage is double Brave Bird's. 

I do wonder about which is better though, since I don't know how much Brave Bird can do off that base 65 attack and no offensive item....

TheImmortal:
Foul Play is twice as much as Knock Off on usage. While Brave Bird is low, Flying STAB isn't guaranteed in Randoms so it should be fine.

Also, sample damage: 85 Atk Mandibuzz Brave Bird vs. 85 HP / 85 Def Infernape: 206-246 (79.5 - 94.9%)

TheImmortal:
U-turn is even less usage than Brave Bird if we took stats into account

ascriptmaster:
Hmmm. How useful is Mandibuzz having a slow U-Turn in randoms?

TheImmortal:
Not sure. It's useful is OMs, I know that. I don't have much experience with it in Randoms but worth keeping for now.

TheImmortal:
Whirlwind is something we could consider for removal if we wanted it to roll more useful moves.

PR #5450, opened 2019-04-15
Split dev-tools/globals* into sim and server

scheibo:
All this PR accomplishes is moving things out of `dev-tools/` and into more 'logical' locations (which will have the side effect of letting us eventually remove the `dev-tools` directory). It is important to note that this is very much '[rearranging the deck chairs on the Titanic](https://en.wiktionary.org/wiki/rearrange_the_deck_chairs_on_the_Titanic)': **`sim`/`data` (and `lib`) still implicitly depend on some declarations in `server/globals*`**, `tsc` treats this change exactly as they were all in the same file anyway. Some concrete examples: code in `data/` still depends on `Monitor` declared in `server/globals.ts`, `lib/process-manager.ts` needs `ChildProcess` declared in `server/globals.d.ts` etc.

This change does almost nothing to help us properly modularize `sim` (design doc forthcoming) or eliminate global usage in our application (also a very worthwhile goal that will be pursued in the near future). **This is purely meant to be an incremental improvement on our current state.**


tslint.json
```diff
@@ -68,6 +68,6 @@
     "rulesDirectory": [],
     "linterOptions": {
       // TODO: why is tslint looking at .js files anyway?
-      "exclude": ["node_modules/", "**/*.js", "dev-tools/global*"]
+      "exclude": ["node_modules/", "**/*.js", "**/globals*.ts"]
```
Zarel:
```suggestion
      "exclude": ["node_modules/", "**/*.js", "*/globals.ts"]
```

seems equivalent?

scheibo: `globals.d.ts`?

scheibo: Can probably remove the `**` though

Zarel: I'm confused that `globals*.ts` would match `global.d.ts` – the latter is missing the first `s`.

Zarel: Also: is `**/*.js` necessary? Is `*.js` enough?

scheibo: `**/*.js` is necessary, and you're right, I had meant `global*.ts` and typoed but clearly it didn't matter. Maybe it skips `.d.ts` files by default (and yet I can't get it to exclude `.js` files? the irony...)?

Updated to be `*/globals.ts`

Zarel:
Man, your merges do horrible things to the "Files changed" panel. :(

PR #5452, opened 2019-04-17
Allow staff to hidetext themselves

whalemer:


AustinXII:
Jw, why?

Zarel:
Should probably allow any target, if we're going to do it that way.

server/chat-commands.js
```diff
@@ -2598,7 +2598,7 @@ const commands = {
 		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${target} not found or has no roomlogs.`);
 		if (!targetUser && !user.can('lock')) return this.errorReply(`User ${name} not found.`);
 		let userid = toId(this.inputUsername);
-		if (!this.can('mute', targetUser, room)) return;
+		if (!(user === targetUser && user.can('mute', null, room)) && !this.can('mute', targetUser, room)) return;
```
```suggestion
```

whalemer: Isn't this just completely removing the permission check?
do you mean ``if (!this.can('mute', null, room)) return;``?

whalemer:
>Jw, why?

I've seen/experienced a number of situations where staff have accidentally said sensitive information (priroom names, potential security issues, even just things like trivia answers). It's a lot easier when you can just hidetext yourself instead of needing to find a RO or leader.

server/chat-commands.js
Zarel: Oh yes, I did mean that.

```suggestion
		if (!this.can('mute', null, room)) return;
```

PR #5454, opened 2019-04-17
Support requesting ties

TheJetOU:


Zarel:
I believe the agreed-upon name was `/offerdraw`

Zarel:
There's also no real reason for `/offerdraw` and `/acceptdraw` to be different commands. They should be aliases; the implementation should be exactly the same whether you're first or second to offer the draw, though.

Zarel:
Also, the draw state should actually be on the battle object. Don't take advantage of the lack of TypeScript in commands to do this... :/

Actually @scheibo we should probably migrate chat commands directly to TypeScript...

Zarel:
Honestly, I wonder if players and users should be more separated. Currently a `RoomGamePlayer` will only exist if a corresponding user exists, but maybe they should be disassociated so player state can be kept around in a consistent location, instead of strewn around the `RoomGame` as is currently done.

(None of this should probably be done in your PR, though – I'm just musing.)

TheJetOU:
> There's also no real reason for `/offerdraw` and `/acceptdraw` to be different commands. They should be aliases; the implementation should be exactly the same whether you're first or second to offer the draw, though.

I mean, it doesn't really makes sense to accept a draw by using `/offerdraw`.

TheJetOU:
Not sure why my VS Code isn't reporting these errors like it used to :(

server/chat-commands.js
```diff
@@ -3748,6 +3748,30 @@ const commands = {
 		this.parse(`/help importinputlog`);
 	},
 
+	acceptdraw: 'offerdraw',
+	offerdraw(target, room, user) {
+		const battle = room.battle;
+		if (!battle) return this.errorReply("This command can only be used in battle rooms.");
+		if (!this.can('roomvoice', null, room)) return;
+		if (!target) {
```
Zarel: There shouldn't need to be a `target`. The offer to a draw should be sent to all other players in the room.

Zarel: I also think `X is offering a draw.` should be public. Only the "Accept draw", "Reject" buttons should be private.

Slayer95:
Draw offers should reset every turn.

server/chat-commands.js
TheJetOU: Only reason I used target was to be able to differentiate when it's `acceptdraw` or `offerdraw` but I guess I can use `cmd`?

TheJetOU: Oh wait, I can use `otherPlayer`!

server/chat-commands.js
```diff
@@ -3748,6 +3748,26 @@ const commands = {
 		this.parse(`/help importinputlog`);
 	},
 
+	acceptdraw: 'offerdraw',
+	offerdraw(target, room, user) {
+		const battle = room.battle;
+		if (!battle) return this.errorReply("This command can only be used in battle rooms.");
+		if (!this.can('roomvoice', null, room)) return;
```
TheJetOU: Should I just make it so players can use it? If a room owner or above wants to draw said room owner can just use `/forcetie`.

Zarel: Yes, players should be able to use it. Is that not the case?

TheJetOU: > only players

TheJetOU: Eh, I guess it's useful sometimes, supported it.

server/room-battle.js
```diff
@@ -559,6 +559,7 @@ class RoomBattle {
 		request.choice = choice;
 
 		this.stream.write(`>${player.slot} ${choice}`);
+		this.allowTie = ['', ''];
```
Slayer95: This resets ``allowTie`` too often. It needs to be reset when we get a new request from the sim.

Zarel:
We should probably initialize draw offers to `null` rather than `['', '']`

server/chat-commands.js
Zarel: It doesn't make sense for ROs/leaders to offer a draw – you'd still need both players to agree to it, right?

TheJetOU: Yeah, I know the code does that. Leader offers a deal, if both players accept the battle is tied.

TheJetOU:
Weird, VS Code didn't complain.

xfix:
> I mean, it doesn't really makes sense to accept a draw by using /offerdraw.

I would say it's reasonable. If you want to offer a draw, then chances are you want a draw. If an opponent offered a draw already, it's reasonable to assume you accept it.

TheJetOU:
Yeah, that makes sense now @xfix.

@Zarel, should be ready for review now.

Slayer95:
@xfix, if you use the right premises in [Temporal logic](https://en.wikipedia.org/wiki/Temporal_logic), the equivalence of those aliases won't hold.

In simpler terms, we need to implement a mechanism to avoid race conditions in which accepting a draw in turn ``n`` translates in offering a draw in turn ``n+1`` after the opponent changes their mind and decides to make a choice.

TheJetOU:
> In simpler terms, we need to implement a mechanism to avoid race conditions in which accepting a draw in turn n translates in offering a draw in turn n+1 after the opponent changes their mind and decides to make a choice.

@Slayer95, Isn't that what already happens because of [this commit](https://github.com/Zarel/Pokemon-Showdown/pull/5454/commits/989525de3d57aa2cadcb0e8e910fb61c09754e7e)?
If Player X and Player Y are playing, In turn 2 Player X offers a draw, Player Y attempts to accept the draw in turn 3, It wouldn't tie the battle, It would just show Player Y offering a draw once again.

Slayer95:
> @Slayer95, Isn't that what already happens because of this commit?
If Player X and Player Y are playing, In turn 2 Player X offers a draw, Player Y attempts to accept the draw in turn 3, It wouldn't tie the battle, It would just show Player Y offering a draw once again.

Indeed, but that's not desirable! The scenario I am considering is as follows:

08:00:01.00 Player Y sends ``/choose move closecombat``
08:00:08.00 Player X sends ``/offerdraw``
08:00:10.00 Player X sends ``/choose switch chansey`` (Player X changed his mind)
08:00:10.05 Player Y sends ``/acceptdraw`` (Player Y believes the turn is the same)

Assuming Player X and Player Y have the same latency, the turn proceeded, and Chansey fainted. And Player Y is now allowing a tie! So Player X lost their dear Chansey and arguably might have lost any chances to win the game, but then he could just do ``/acceptdraw`` and end the battle as a tie!

server/chat-commands.js
Zarel: Oh, that's not a bad idea.

Zarel:
Hm, that's a good point. I'm not sure that requires a workaround, though.

Slayer95:
Well, with a proper GUI, offering and accepting draws would send the ``rqid``.

However, if that's not planned for the near future, the simplest solution would be removing the equivalence between ``/acceptdraw`` and ``/offerdraw``. Then the scenario above would just result in an error: ``"No draw offer available. It might have been automatically withdraw as the current turn started."``

Slayer95:
Interestingly enough, we could also make ``/offerdraw`` *accept* draws, but make it so that ``/acceptdraw`` doesn't *offer* draws, and that would be best imo.

TheJetOU:
I'm more leaning towards the first solution which is removing the equivalency between `/offerdraw` and `/acceptdraw`.
What do you think, @Zarel?

Zarel:
If the code isn't closely related, that means you're writing it wrong.

TheJetOU:
Okay, the code is much better now, I think. But the race condition is still present :|

TheJetOU:
Race condition is fixed!

TheJetOU:
Bumping @Zarel.

TheJetOU:
@Zarel, the Smogon Tournament Directors have stated they don't want `/offerdraw` and `/acceptdraw` in their server, so set `allowRequestingTies` to `false` when updating Smogtours.

config/config-example.js
```diff
@@ -171,6 +171,12 @@ exports.crashguardemail = null;
  */
 exports.disablebasicnamefilter = false;
 
+/**
+ * allow requesting ties - allows the use of commands `/offerdraw` and `/acceptdraw`
+ * If this setting  is set to `false` it will disallow users to request or offer draws.
+ */
+exports.allowRequestingTies = true;
```
Zarel: PS config is currently all-lowercase by convention.

Zarel:
(This is pending my refactor of `BattlePlayer`)

TheJetOU:
This Pull Request is in the same situation as #5456, because since players don't have the player rank, they can't support ties.

TheJetOU:
(Tagging @scheibo to add the 'Do Not Merge' label)

server/chat-commands.js
```diff
@@ -3740,6 +3740,39 @@ const commands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ & ~`],
 
+	acceptdraw: 'acceptdraw',
+	offerdraw(target, room, user, connection, cmd) {
+		const battle = room.battle;
```
scheibo: Move these two variables down 5 lines to where theyre actually used

server/chat-commands.js
```diff
@@ -3740,6 +3740,39 @@ const commands = {
...
+		if (!battle.allowTie) {
+			battle.allowTie = [];
+			for (const player of playerIDS) {
+				if (player === user.userid) {
+					battle.allowTie.push(player);
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} is offering a draw, <button class="button">do you accept</button>`
```
scheibo: Not sure about the wording here, but probably only `'accept'` should be in the `button`

server/chat-commands.js
```diff
@@ -3740,6 +3740,39 @@ const commands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ & ~`],
 
+	acceptdraw: 'acceptdraw',
+	offerdraw(target, room, user, connection, cmd) {
+		const battle = room.battle;
+		const playerIDS = Object.keys(battle.playerTable);
```
scheibo: despite renaming `toID`, I think `playerIds` would still be more typical for the codebase.

server/chat-commands.js
TheJetOU: `battle` is used exactly one line down, so just moved `playerIDs` (now `playerIds`).

server/chat-commands.js
TheJetOU: Sure, I also forgot to make the button send the command, lol.

server/chat-commands.js
TheJetOU: I like camelCase so went with `playerIds` :p

Zarel: I'd imagine `playerIDs` works fine.

server/chat-commands.js
```diff
@@ -3740,6 +3740,39 @@ const commands = {
...
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} is offering a draw, do you <button class="button" name="send" value="/acceptdraw">accept</button>?`
+				);
+			}
+			this.add(`${user.name} is offering a draw`);
+		} else {
+			if (!playerIds.includes(user.userid)) return this.errorReply("Must be a player to accept draws");
+			battle.allowTie.push(user.userid);
```
Zarel: So you can force a battle to tie by just using `/offerdraw` twice? :p

TheJetOU: Oh right, good catch.

TheJetOU:
Bumping this.

server/room-battle.js
```diff
@@ -631,6 +634,8 @@ class RoomBattle {
 		}
 	}
 	receive(/** @type {string[]} */ lines) {
+		this.allowTie = null;
```
Zarel: It makes more sense to call this `wantsTie`, and also, it makes more sense to make this a property of `BattlePlayer` than `Battle`

config/config-example.js
```diff
@@ -171,6 +171,12 @@ exports.crashguardemail = null;
  */
 exports.disablebasicnamefilter = false;
 
+
```
Zarel:
```suggestion
```

server/chat-commands.js
Zarel: So you can force a battle to tie just by using `/offerdraw`, changing your username, and using `/offerdraw` again? :p

server/room-battle.js
Zarel: Putting it on `BattlePlayer` also fixes all the vulns I mentioned earlier.

Zarel: In which case this would be:

```suggestion
		for (const player of this.players) player.wantsTie = false;
```

server/chat-commands.js
```diff
@@ -3740,6 +3740,40 @@ const commands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ & ~`],
 
+	acceptdraw: 'acceptdraw',
+	offerdraw(target, room, user, connection, cmd) {
+		const battle = room.battle;
+		if (!battle) return this.errorReply("Must be in a battle room.");
+		if (!Config.allowrequestingties) return this.errorReply("This server does not allow the offering/accpeting of draws");
```
Zarel: Typo in `accpeting`, but I'm not a huge fan of this phrasing, anyway.

```suggestion
		if (!Config.allowrequestingties) return this.errorReply("This server does not allow offering draws.");
```

server/chat-commands.js
```diff
@@ -3740,6 +3740,40 @@ const commands = {
...
+		if (!this.can('roomvoice', null, room)) return;
+		if (cmd === 'acceptdraw' && !battle.allowTie) return this.errorReply("No draw offer available. It might have been automatically withdrawn as the current turn started.");
```
Zarel: Better phrasing, also the property on `Player`

```suggestion
		if (cmd === 'acceptdraw' && !battle.players.some(player => player.wantsTie)) {
			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
		}
```

server/chat-commands.js
```diff
@@ -3740,6 +3740,40 @@ const commands = {
...
+		if (!battle.allowTie) {
+			battle.allowTie = [];
+			for (const player of playerIds) {
+				if (player === user.userid) {
+					battle.allowTie.push(player);
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} is offering a draw, do you <button class="button" name="send" value="/acceptdraw">accept</button>?`
```
```suggestion
					Chat.html`|html|${user.name} wants this game to be a tie, do you <button class="button" name="send" value="/acceptdraw">accept tie</button>?`
```

Ideally there's also a reject button.

config/config-example.js
```diff
@@ -171,6 +171,11 @@ exports.crashguardemail = null;
  */
 exports.disablebasicnamefilter = false;
 
+/**
+ * allowrequestingties - disables the use of `/offerdraw` and `/acceptdraw`
```
panpawn: Is this comment right? Its description sounds like it contradicts the variable name to me?

server/chat-commands.js
```diff
@@ -3796,6 +3796,42 @@ const commands = {
...
+			if (battle.allowTie.includes(user.userid)) return this.errorReply("You have already accepted to draw this battle.");
+			battle.allowTie.push(user.userid);
+			this.add(`${user.userid} accepts to draw this battle`);
+			if (playerIds.length === battle.allowTie.length) {
+				this.add(`All the players in this battle (${Chat.toListString(playerIds)}) accept to draw this battle.`);
+				room.battle.tie();
+			}
+		}
+	},
+	offerdrawhelp: [`/offerdraw - Offers a draw to all others players in a battle, if all accept it is then tied.`],
```
panpawn: This could say that it requires being a player in the battle.

server/chat-commands.js
```diff
@@ -3796,6 +3796,42 @@ const commands = {
...
+		if (!battle.allowTie) {
+			battle.allowTie = [];
+			for (const player of playerIds) {
+				if (player === user.userid) {
+					battle.allowTie.push(player);
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} wants this game to be a tie; <button class="button" name="send" value="/acceptdraw">accept tie</button>?`
```
panpawn: I think this could be phrased better. What about something like:
"Your opponent wants this battle to end in a tie. If you want, you can ``<button>``accept this battle to end in a tie``</button>``"

panpawn:
You're using both ``tie`` and ``draw`` in the UI. I would suggest picking one that remains constant thru the UI (not necessarily the commands, but the prompts the users get etc)

server/chat-commands.js
Zarel: @panpawn I think your phrasing is pretty bad English, and also I don't like messages that are unnecessarily verbose. The shorter the message, the easier it is to read, and the less likely someone will skim over something important.

panpawn: At the least, I think this line should be clear that this will end the battle, and should use a word like "end"

Zarel:
I'd prefer "tie", since I think it's the more common word. `/requesttie` and `/accepttie`. I'm amenable to using "draw" in the command aliases.

server/chat-commands.js
Zarel: Fair, I wouldn't mind replacing "wants this game to be a tie" with "wants this game to end in a tie". I just think the "If you want" part is unnecessarily long.

server/chat-commands.js
TheJetOU: Moderators and above can also. I guess I can add a `Requires: x`.

config/config-example.js
TheJetOU: Yeah, I noticed it when testing.

TheJetOU:
> I'd prefer "tie", since I think it's the more common word. /requesttie and /accepttie. I'm amenable to using "draw" in the command aliases.

It was called `/accepttie`` in the beginning but you said..

> I believe the agreed-upon name was /offerdraw

Zarel:
Yeah, that's my bad. "Offer draw" is way more common than "request tie", but "tie" by itself is more common than "draw". We should definitely support both `/offerdraw` and `/requesttie` as aliases, but the text should probably consistently use "tie".

server/room-battle.js
```diff
@@ -53,6 +53,8 @@ class RoomBattlePlayer extends RoomGames.RoomGamePlayer {
 
 		/** @type {BattleRequestTracker} */
 		this.request = {rqid: 0, request: '', isWait: 'cantUndo', choice: ''};
+		/** @type {boolean} */
```
```suggestion
```

type inference exists

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+	accepttie: 'offertie',
+	offertie(target, room, user, connection, cmd) {
+		const battle = room.battle;
+		if (!battle) return this.errorReply("Must be in a battle room.");
+		if (!Config.allowrequestingties) {
+			return this.errorReply("This server does not allow offering ties.");
+		}
+		if (!this.can('roomvoice', null, room)) return;
+		if (cmd === 'accepttie' && !battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No tie offer available. It might have been automatically withdrawn as the current turn started.");
```
Zarel: "automatically withdrawn" is unnecessarily complicated phrasing.

Zarel: And also not really the core of the problem which is that the opponent took back the offer.

```suggestion
			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+		}
+		const playerIds = Object.keys(battle.playerTable);
```
Zarel: This is unnecessary; `players` exists. Unrelatedly, caching the current player is probably useful here.

```suggestion
		const player = battle.playerTable[user.userid];
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+		if (!battle.players.some(player => player.wantsTie)) {
+			for (const player of playerIds) {
```
Zarel: Moving the `wantsTie` setting outside of the loop makes sense.

```suggestion
			if (player) player.wantsTie = true;
			for (const otherPlayer of battle.players) {
				if (otherPlayer !== player) {
					otherPlayer.send(Chat.html`|html|${user.name} wants this game to end in a tie; <button class="button" name="send" value="/accepttie">accept tie</button>?`);
				}
			}
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+				if (player === user.userid) {
+					battle.playerTable[player].wantsTie = true;
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} wants this game to end in a tie; <button class="button" name="send" value="/accepttie">accept tie</button>?`
+				);
+			}
+			this.add(`${user.name} is offering a tie`);
```
Zarel: Maybe move this line above the others, so it appears above the "accept tie" button.

Zarel: Actually, once this is above the others, you don't even need to repeat it in "accept tie" button.

Remember to `room.update()` if you need to.

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+		const playerIds = Object.keys(battle.playerTable);
+		if (!battle.players.some(player => player.wantsTie)) {
+			for (const player of playerIds) {
+				if (player === user.userid) {
+					battle.playerTable[player].wantsTie = true;
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} wants this game to end in a tie; <button class="button" name="send" value="/accepttie">accept tie</button>?`
```
```suggestion
					Chat.html`|uhtml|offertie|<button class="button" name="send" value="/accepttie"><strong>accept tie</strong></button> <button class="button" name="send" value="/rejecttie">reject</button>`
```

And then make it so `/rejecttie` hides the `offertie` uhtml.

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+			if (!playerIds.includes(user.userid)) {
+				return this.errorReply("Must be a player to accept ties");
+			}
+			if (battle.playerTable[user.userid].wantsTie) {
+				return this.errorReply("You have already accepted to tie this battle.");
+			}
+			battle.playerTable[user.userid].wantsTie = true;
+			this.add(`${user.userid} accepts to tie this battle.`);
+			if (battle.players.every(player => player.wantsTie)) {
+				this.add(`All the players in this battle (${Chat.toListString(playerIds)}) accept to tie this battle.`);
```
```suggestion
				this.add(`All the players in this battle (${Chat.toListString(battle.players.map(player => player.name))}) agree to tie this battle.`);
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+			this.add(`${user.name} is offering a tie`);
+		} else {
+			if (!playerIds.includes(user.userid)) {
+				return this.errorReply("Must be a player to accept ties");
+			}
+			if (battle.playerTable[user.userid].wantsTie) {
+				return this.errorReply("You have already accepted to tie this battle.");
+			}
+			battle.playerTable[user.userid].wantsTie = true;
+			this.add(`${user.userid} accepts to tie this battle.`);
```
```suggestion
			this.add(`${user.userid} agrees to tie this battle.`);
```

"accepts to" doesn't really work in English.

server/chat-commands.js
Zarel: Actually, I still don't believe this needs to exist at all. Maybe the correct solution is that draw offers should cancel and prevent moves from being made.

Zarel: No, the timer still exists. Rather, I think draw offers should last across turns, instead of needing reconfirming every turn.

Zarel: Or maybe something weird, like lasting exactly two turns.

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+			}
+			this.add(`${user.name} is offering a tie`);
+		} else {
+			if (!playerIds.includes(user.userid)) {
+				return this.errorReply("Must be a player to accept ties");
+			}
+			if (battle.playerTable[user.userid].wantsTie) {
+				return this.errorReply("You have already accepted to tie this battle.");
+			}
+			battle.playerTable[user.userid].wantsTie = true;
```
Zarel: These four lines can be deleted now

server/chat-commands.js
```diff
@@ -3796,6 +3796,49 @@ const commands = {
...
+					continue;
+				}
+				Users(player).sendTo(
+					room,
+					Chat.html`|html|${user.name} wants this game to end in a tie; <button class="button" name="send" value="/accepttie">accept tie</button>?`
+				);
+			}
+			this.add(`${user.name} is offering a tie`);
+		} else {
+			if (!playerIds.includes(user.userid)) {
```
```suggestion
			if (!player) {
```

Zarel:
This pull request **introduces 1 alert** and **fixes 3** when merging ab9243e15d84bccc3afba3f3df90823d1fd7984f into 05693188483b62a88154dc08b0e5e59865e34cbd - [view on LGTM.com](https://lgtm.com/projects/g/Zarel/Pokemon-Showdown/rev/pr-88c3453c12a8aae8269d46d0f5bbaf89ec8e4abf)

**new alerts:**

* 1 for Syntax error

**fixed alerts:**

* 2 for Useless assignment to local variable
* 1 for Incomplete URL substring sanitization

---

server/chat-commands.js
TheJetOU: I feel like that's a discussion for Policy Review, so we should go with the safe option for now which is resetting after every turn.

server/chat-commands.js
TheJetOU: I added this but removed it later. It just adds some form of code complexity because it gets harder to check if it's an offer or an acceptance. I don't feel like (the current code) which is just an ` } else {` is all that bad and probably simpler

server/chat-plugins/mafia.js
```diff
@@ -51,8 +51,8 @@
 
 /** @type {typeof import('../../lib/fs').FS} */
 const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
-const LOGS_FILE = 'config/chat-plugins/mafia-logs.json';
-const BANS_FILE = 'config/chat-plugins/mafia-bans.json';
+const LOGS_FILE = '../../config/chat-plugins/mafia-logs.json';
+const BANS_FILE = '../../config/chat-plugins/mafia-bans.json';
```
TheJetOU: I accidentally added this but I couldn't run a PS server without it and even then it still crashed when ran. I'll go ahead and remove it and investigate what's actually happening later.

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+		if (!battle) return this.errorReply("Must be in a battle room.");
+		if (!Config.allowrequestingties) {
+			return this.errorReply("This server does not allow offering ties.");
+		}
+		if (!this.can('roomvoice', null, room)) return;
+		if ((cmd === 'accepttie' || cmd === 'rejecttie') && !battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		const player = battle.playerTable[user.userid];
+		if (cmd === 'rejecttie' && battle.players.some(player => player.wantsTie)) {
```
Zarel: I would make `/rejecttie` a separate command.

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+			if (!player) {
+				return this.errorReply("Must be a player to reject ties.");
+			}
+			if (player.wantsTie) player.wantsTie = false;
+			for (const player of battle.players) {
+				player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie offer rejected)`);
+			}
+			return this.add(`${user.userid} rejects to tie this battle`);
```
```suggestion
			return this.add(`${user.userid} rejects tying this battle`);
```

`declines to tie this battle` also works, but might sound too formal?

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		const player = battle.playerTable[user.userid];
+		if (cmd === 'rejecttie' && battle.players.some(player => player.wantsTie)) {
+			if (!player) {
+				return this.errorReply("Must be a player to reject ties.");
+			}
+			if (player.wantsTie) player.wantsTie = false;
+			for (const player of battle.players) {
+				player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie offer rejected)`);
```
```suggestion
				player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie rejected)`);
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+				return this.errorReply("Must be a player to accept ties");
+			}
+			if (!player.wantsTie) {
+				player.wantsTie = true;
+			} else {
+				return this.errorReply("You have already agreed to tie this battle.");
+			}
+			this.add(`${user.userid} agrees to tie this battle.`);
+			if (battle.players.every(player => player.wantsTie)) {
+				this.add(`All the players in this battle (${Chat.toListString(battle.players.map(player => player.name))}) agree to tie this battle.`);
```
```suggestion
				if (battle.players.length > 2) {
					this.add(`All players have agreed to tie.`);
				}
```

- no need to report in two-player battles, "[user] agrees to tie" is enough
- no need to list users; "[user] agrees to tie" is enough

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+		} else {
+			if (!player) {
+				return this.errorReply("Must be a player to accept ties");
+			}
+			if (!player.wantsTie) {
+				player.wantsTie = true;
+			} else {
+				return this.errorReply("You have already agreed to tie this battle.");
+			}
+			this.add(`${user.userid} agrees to tie this battle.`);
```
```suggestion
			this.add(`${user.name} agrees to a tie.`);
```

Zarel:
This is looking good; I approve of everything except the minor things I commented.

server/chat-commands.js
```diff
@@ -3796,6 +3796,60 @@ const commands = {
...
+				player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie offer rejected)`);
+			}
+			return this.add(`${user.userid} rejects to tie this battle`);
+		}
+		if (!battle.players.some(player => player.wantsTie)) {
+			this.add(`${user.name} is offering a tie.`);
+			room.update();
+			for (const otherPlayer of battle.players) {
+				if (otherPlayer !== player) {
+					otherPlayer.sendRoom(Chat.html`|uhtml|offertie|<button class="button" name="send" value="/accepttie"><strong>accept tie</strong></button> <button class="button" name="send" value="/rejecttie">reject</button>`);
```
```suggestion
					otherPlayer.sendRoom(Chat.html`|uhtml|offertie|<button class="button" name="send" value="/accepttie"><strong>Accept tie</strong></button> <button class="button" name="send" value="/rejecttie">Reject</button>`);
```

server/chat-commands.js
TheJetOU: I feel like using 'rejects' is better because the command name is `/rejecttie` and not `/declinetie`

server/chat-commands.js
```diff
@@ -3796,6 +3796,69 @@ const commands = {
...
+			return this.errorReply("Must be a player to reject ties.");
+		}
+		if (!battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		if (player.wantsTie) player.wantsTie = false;
+		for (const player of battle.players) {
+			player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie rejected)`);
+		}
+		return this.add(`${user.name} rejects to tie this battle`);
```
```suggestion
		return this.add(`${user.name} rejects tying this battle`);
```

Zarel: I have a better idea:

```suggestion
		return this.add(`${user.name} rejected the tie.`);
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,69 @@ const commands = {
...
+		const player = battle.playerTable[user.userid];
+		if (!player) {
+			return this.errorReply("Must be a player to reject ties.");
+		}
+		if (!battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		if (player.wantsTie) player.wantsTie = false;
+		for (const player of battle.players) {
+			player.sendRoom(Chat.html`|uhtmlchange|offertie|(Tie rejected)`);
```
```suggestion
			player.sendRoom(Chat.html`|uhtmlchange|offertie|`);
```

The "[user] rejected the tie" message is enough.

server/chat-commands.js
```diff
@@ -3796,6 +3796,69 @@ const commands = {
...
+		if (!battle) return this.errorReply("Must be in a battle room.");
+		const player = battle.playerTable[user.userid];
+		if (!player) {
+			return this.errorReply("Must be a player to reject ties.");
+		}
+		if (!battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		if (player.wantsTie) player.wantsTie = false;
+		for (const player of battle.players) {
```
Zarel: Shadowing `player`...

```suggestion
		for (const otherPlayer of battle.players) {
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,69 @@ const commands = {
...
+			}
+		} else {
+			if (!player) {
+				return this.errorReply("Must be a player to accept ties.");
+			}
+			if (!player.wantsTie) {
+				player.wantsTie = true;
+			} else {
+				return this.errorReply("You have already agreed to tie this battle.");
+			}
```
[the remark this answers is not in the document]
```suggestion
			}
			player.sendRoom(Chat.html`|uhtmlchange|offertie|`);
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,70 @@ const commands = {
...
+				}
+			}
+		} else {
+			if (!player) {
+				return this.errorReply("Must be a player to accept ties.");
+			}
+			if (!player.wantsTie) {
+				player.wantsTie = true;
+			} else {
+				return this.errorReply("You have already agreed to tie this battle.");
```
```suggestion
				return this.errorReply("You have already agreed to a tie.");
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,70 @@ const commands = {
...
+			}
+			player.sendRoom(Chat.html`|uhtmlchange|offertie|`);
+			this.add(`${user.name} agrees to tie this battle.`);
```
```suggestion
			this.add(`${user.name} agrees to a tie.`);
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,70 @@ const commands = {
...
+			if (battle.players.every(player => player.wantsTie)) {
+				if (battle.players.length > 2) {
+					this.add(`All players have agreed to a tie.`);
+				}
+				room.battle.tie();
+			}
+		}
+	},
+	offertiehelp: [`/offertie - Offers a tie to all others players in a battle, if all accept it is then tied. Requires: \u2606 @ # & ~`],
```
```suggestion
	offertiehelp: [`/offertie - Offers a tie to all other players in a battle; if all accept, it ends in a tie. Requires: \u2606 @ # & ~`],
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,70 @@ const commands = {
...
+		if (!battle.players.some(player => player.wantsTie)) {
+			return this.errorReply("No other player is requesting a tie right now. It was probably canceled.");
+		}
+		if (player.wantsTie) player.wantsTie = false;
+		for (const otherPlayer of battle.players) {
+			otherPlayer.sendRoom(Chat.html`|uhtmlchange|offertie|`);
+		}
+		return this.add(`${user.name} rejected the tie.`);
+	},
+	rejecttiehelp: [`/rejecttie - If a tie exists in a battle, the tie rejected.`],
```
```suggestion
	rejecttiehelp: [`/rejecttie - Rejects a tie offered by another player in a battle.`],
```

server/chat-commands.js
```diff
@@ -3796,6 +3796,70 @@ const commands = {
...
+			if (!player.wantsTie) {
+				player.wantsTie = true;
+			} else {
+				return this.errorReply("You have already agreed to tie this battle.");
+			}
+			player.sendRoom(Chat.html`|uhtmlchange|offertie|`);
+			this.add(`${user.name} agrees to tie this battle.`);
+			if (battle.players.every(player => player.wantsTie)) {
+				if (battle.players.length > 2) {
+					this.add(`All players have agreed to a tie.`);
```
```suggestion
					this.add(`All players have accepted the tie.`);
```

Zarel:
Okay, just some grammar fixes and we should be done!

PR #5456, opened 2019-04-20
Refactor /(allow)exportinput log

TheJetOU:
This refactor changes:
- Allows multiple people to extract input logs

- Scales beyond 2 players

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -3904,7 +3904,7 @@ const commands = {
 		if (!battle.allowExtraction[user.userid]) {
 			battle.allowExtraction[user.userid] = new Set();
 			for (const player of battle.players) {
-				const playerUser = Users(player.name);
+				const playerUser = player.getUser();
 				if (!playerUser) continue;
 				if (playerUser.userid === user.userid) {
 					battle.allowExtraction[user.userid].add(user.userid);
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -3864,59 +3864,57 @@ const commands = {
 
 	allowexportinputlog(/** @type {string} */ target, /** @type {Room?} */ room, /** @type {User} */ user) {
 		const battle = room.battle;
-		if (!battle) return this.errorReply(`Must be in a battle.`);
-		if (!battle.allowExtraction) return this.errorReply(`Someone must have requested extraction.`);
-		const targetUser = Users.getExact(target);
-		if (!targetUser) return this.errorReply(`User ${target} not found.`);
-
-		if (toID(battle.p1.name) === user.userid) {
-			battle.allowExtraction[0] = targetUser.userid;
-		} else if (toID(battle.p2.name) === user.userid) {
-			battle.allowExtraction[1] = targetUser.userid;
-		} else {
-			return this.errorReply(`Must be a player in the battle.`);
+		if (!battle) {
+			return this.errorReply(`Must be in a battle.`);
 		}
-		this.addModAction(`${user.userid} consents to sharing battle team and choices with ${targetUser.userid}.`);
-		if (battle.allowExtraction.join(',') !== `${targetUser.userid},${targetUser.userid}`) return;
-
-		this.addModAction(`${targetUser.name} has extracted the battle input log.`);
-		const inputLog = battle.inputLog.map(Chat.escapeHTML).join(`<br />`);
-		targetUser.sendTo(room, `|html|<div class="chat"><code style="white-space: pre-wrap; overflow-wrap: break-word; display: block">${inputLog}</code></div>`);
+		const targetUser = Users.getExact(target);
+		if (!targetUser) {
+			return this.errorReply(`User ${target} not found.`);
+		}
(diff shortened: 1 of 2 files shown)


server/chat-commands.js
```diff
@@ -3689,25 +3684,20 @@ const commands = {
 		}
 		if (!this.can('exportinputlog', null, room)) return;
```
TheJetOU: ...Why? Why can't players export their own input log? It shouldn't be anyone's business if Player X and Player Y both consent to extracting their own input logs.

TheJetOU:
BTW, this code structure is how i envision for #5454 to be.

server/chat-commands.js
Zarel: They can...

TheJetOU: No? Only Leaders have `exportinputlog` set to true, Administrators have `exportinputlog` set to true by default, because they have root access,

Zarel: Try it on Main. I already changed it, I just haven't updated the example config yet.

TheJetOU: Oh, that makes sense. Created #5457 to fix that.

TheJetOU:
Bumping this

Zarel:
This is pending my refactor of `BattlePlayer`

TheJetOU:
Why doesn't GitHub supported converting a regular Pull Request to a Draft Pull Request.. :(

There are some bug that will need to be fix before this can be updated (if someone uses `/exportinputlog`, it will crash assuming the server has been restarted)

Some of the bugs I've noticed are:

- Players don't have the rank player, so they won't have player permissions

- Things like `battle.playerTable`, `battle.players`, `battle.p1`, are empty/null when the battle, which doesn't work with `/exportinputlog` as it must  only be used when the battle end, thus it will crash.

I'll see if I have time to tackle these, but I'm not that familiar with `server/room-battle`'s code

scheibo:
> Why doesn't GitHub supported converting a regular Pull Request to a Draft Pull Request..

Yeah, this is dumb. Added `do not merge` for now to signal this, please comment when this PR is ready again and I'll remove and review :)

Zarel:
I pushed 2a31bc308cd90e5f135 which should fix the things being `null` at the end of a battle.

server/chat-commands.js
```diff
@@ -3863,59 +3863,56 @@ const commands = {
...
-			if (p2) p2.sendTo(room, Chat.html`|html|${user.name} wants to extract the battle input log. <button name="send" value="/allowexportinputlog ${user.userid}">Share your team and choices with "${user.name}"</button>`);
-		}
-
-		if (battle.allowExtraction.join(',') !== `${user.userid},${user.userid}`) {
-			this.addModAction(`${user.name} wants to extract the battle input log.`);
-			return;
+		if (!battle.allowExtraction[user.userid]) {
+			battle.allowExtraction[user.userid] = new Set();
+			for (const player of battle.players) {
+				const playerUser = Users(player.name);
```
TheJetOU: I would just do `player.getUser()` but it seems like `player.userid` is cleared when the battle ends, I'm not really sure why that's the case.

Zarel: Oh, it's the case because the user isn't considered a player when the battle ends (the user is allowed to change name without forfeiting, etc).

Zarel: It does make this code awkward, though. I wonder if there's a solution here.

Zarel: (This in general shouldn't be done because `player.name` isn't guaranteed to be a username; that's why `player.userid` exists in the first place.)

TheJetOU: Is this a better solution?
```javascript
const players = [battle.p1, battle.p2, battle.p3, battle.p4].filter(player => player);
for (const player of players) { ... }```

TheJetOU: > (This in general shouldn't be done because player.name isn't guaranteed to be a username; that's why player.userid exists in the first place.)

How so, is ``RoomBattlePlayer.name`` also not guaranteed to be a real name?

Zarel: No, none of those really deal with the core problem, which is relatively obscure.

One of them is `/importinputlog` in which case I think the player names can be whatever you want them to be?

This is probably fine as-is.

TheJetOU: I still don't understand why we clear `player.userid`.

Players can still change their name because the user rename methods check `this.games` which wouldn't contain `RoomBattle` because `unlinkUser` deletes it.

Zarel: Um, I want `player.userid` to stay in sync with `user.games`.

I guess that's not _strictly_ necessary though? I suppose I can accept a desync at the end of game, or, like, a flag or something.

Zarel: Anyway, yes, overall, I think that's a good idea. Let's just keep `player.userid` uncleared.

TheJetOU: Okay, I'll just make another PR for this, since it's a bit unrelated to this. I already have merge conflicts anyway..

TheJetOU: #5695.

PR #5458, opened 2019-04-20
Improve roomevents

TheJetOU:
This was on the "More common suggestions" in the [suggestions thread](https://www.smogon.com/forums/threads/3534365/).

### DIFF
--- a/server/chat-plugins/room-events.js
+++ b/server/chat-plugins/room-events.js
@@ -74,7 +74,7 @@ exports.commands = {
 				return this.errorReply("There are currently no planned upcoming events for this room to start.");
 			}
 			if (!target) return this.errorReply("Usage: /roomevents start [event name]");
-			target = toId(target);
+			target = toID(target);
 			if (!room.events[target]) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
 			if (room.events[target].started) return this.errorReply(`The event ${room.events[target].eventName} has already started.`);
 			for (const u in room.users) {
--- a/server/chat-plugins/room-events.js
+++ b/server/chat-plugins/room-events.js
@@ -22,7 +22,11 @@ exports.commands = {
 			let buff = '<table border="1" cellspacing="0" cellpadding="3">';
 			buff += '<th>Event Name:</th><th>Event Description:</th><th>Event Date:</th>';
 			for (let i in room.events) {
-				buff += `<tr><td>${Chat.escapeHTML(room.events[i].eventName)}</td><td>${Chat.formatText(room.events[i].desc, true)}</td><td><time>${Chat.escapeHTML(room.events[i].date)}</time></td></tr>`;
+				let timeRemaining = new Date(room.events[i].date).getTime() - new Date().getTime();
+				if (!timeRemaining) timeRemaining = "The time remaining for this event is not available";
+				if (timeRemaining < 0) timeRemaining = "This event will start soon";
+				if (room.events[i].started) timeRemaining = "This event has started";
+				buff += `<tr title="${isNaN(timeRemaining) ? timeRemaining : `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`}"><td>${Chat.escapeHTML(room.events[i].eventName)}</td><td>${Chat.formatText(room.events[i].desc, true)}</td><td><time>${Chat.escapeHTML(room.events[i].date)}</time></td></tr>`;
 			}
 			buff += '</table>';
 			return this.sendReply(`|raw|<div class="infobox-limited">${buff}</div>`);
@@ -55,11 +59,34 @@ exports.commands = {
 				eventName: eventName,
 				date: date,
 				desc: desc,
+				started: false,
 			};
 
 			room.chatRoomData.events = room.events;
 			Rooms.global.writeChatRoomData();
 		},
+
+		begin: 'start',
+		start(target, room, user) {
+			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+			if (!this.can('ban', null, room)) return false;
+			if (!room.events || Object.keys(room.events).length === 0) {
+				return this.errorReply("There are currently no planned upcoming events for this room to start.");
+			}
+			if (!target) return this.errorReply("Usage: /roomevents start [event name]");
+			target = toId(target);
+			if (!room.events[target]) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
+			if (room.events[target].started) return this.errorReply(`The event ${room.events[target].eventName} has already started.`);
+			for (const u in room.users) {
+				if (Users(u).connected) Users(u).sendTo(room, Chat.html`|notify|A new roomevent in ${room.title} has started!|The "${room.events[target].eventName}" roomevent has started!`);
+			}
+			this.add(Chat.html`|raw|<div class="broadcast-blue"><b>The "${room.events[target].eventName}" roomevent has started!</b></div>`);
+			this.modlog('ROOMEVENT', null, `started "${target}"`);
+			room.events[target].started = true;
+			room.chatRoomData.events = room.events;
+			Rooms.global.writeChatRoomData();
+		},
+
 		delete: 'remove',
 		remove(target, room, user) {
 			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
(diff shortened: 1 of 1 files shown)


server/chat-plugins/room-events.js
```diff
@@ -55,6 +55,7 @@ exports.commands = {
 				eventName: eventName,
 				date: date,
 				desc: desc,
+				timeRemaining: new Date(date).getTime() - new Date().getTime(),
```
Slayer95: Gotta make sure that ``date`` is a valid parseable JS date.

TheJetOU: It's checked if `timeRemaining` is a false-y value in L25, should I move it to there?

Slayer95: Uh........ Yeah..... ``timeRemaning`` shouldn't be stored, but validated and calculated on-demand.

TheJetOU: Done.

Slayer95:
This feature needs to be engineered carefully. The scheduling of many events may be incompatible with a naive countdown implementation, most notably events with a continuous duration, or those held in rooms with hand-wavy timetables.

TheJetOU:
@Slayer95, I mean there are two ways to solve this really

- Explicitly mention that it's a 'Predicted' time.
- Turn it into an input provided by the user.

Or, build an AI which can predict human inaccuracy.

bumbadadabum:
How about a mouseover instead? Maybe put it in the same column as the date? Regardless, it shouldn't be shown as N/A in its own column for events that don't match the exact time element we expect. I agree with Slayer95 that this feature is too naive to be an "always visible" thing. I don't think the issue with hand-wavy timetables is new with this addition though. The time already doesn't match with the current features available.

Slayer95:
We must start from the analysis of the use cases.
1. Date is scheduled as ``yyyy/MM/dd``. Are we concerned about this? I don't think so. So long as users are aware of the current date, they can figure out that the event is "tomorrow", "in two days", "next week", or whatever. (*)

2. Date is scheduled as ``yyyy/MM/dd hh``. This is where the main use case lies. If the French room timetable claims the event will be held at ``3:00pm``, is that Paris' time? UTC? Or perhaps it's automagically the local time of my overseas country?

So, if we focus on use case number 2:

Does the server currently have the required information to provide a reliable answer? I don't think so. To fix that, we need to first provide a reliable and friendly way for room staff to schedule their events. How?

a) We could force everyone to input times as UTC, and then tell them: ``"this event will be scheduled within 8 hours, do blah blah to confirm"``.
b) We could implement some client-side helper command using ``<input type="datetime-local" />`` and then convert it to universal time with ``new Date(input.value).toISOString()``.

Once we have that, the server can then reliably say ``"the room staff intends this event to start in 3 hours"``. But what happens once the time is due? I think we could have the column be ``"Status"``, and then it would change to ``"Starting soon"`` and notify the room staff. The room staff would then do their thing and manually change the event to ``"Started"`` (we could even have a button in the event list command for that). This proposal would also apply if we decide to have a tooltip as opposed to an independent column.

--------------

(*) Of course, calculating times this is isn't quite trivial for large timezone offsets, but that issue reduces to use case number 2.

TheJetOU:
@bumbadadabum, that seems reasonable enough, and I agree with it.

@Slayer95, 

>If the French room timetable claims the event will be held at 3:00pm, is that Paris' time? UTC? Or perhaps it's automagically the local time of my overseas country?

The time is sent through the HTML5 tag `<time>` which I think assumes the time is in UTC+0 and converts it to your local timezone, so roomstaff (should) input the time in UTC+0, I don't think that's document though in `/help roomevents` though.

> a) We could force everyone to input times manually as UTC, and then tell them: "this event will be scheduled within 8 hours, do blah blah to confirm", or "this event will not have countdown information available, please input date as UTC, or do blah blah to confirm"

Yeah, that makes sense. We could also make it an input form (Client-side) which would have the fields: Name, Description, Date (explicitly mention to input in UTC and a specific format) and clicking "Add" would send the data to the server.

> Once we have that, the server can then reliably say "the room staff intends this event to start in 3 hours". But what happens once the time is due? I think we could have the column be "Status", and then it would change to "Starting soon" and notify the room staff. The room staff would then do their thing and manually change the event to "Started" (we could even have a button in the event list command for that). This proposal would also apply if we decide to have a tooltip as opposed to an independent column.

Yeah, a `/roomevent start [event name]` makes sense, it would declare (highlight) to all the users in the room that [event name] started.

TheJetOU:
Okay, implemented the above features.

I think it's now reliable enough to warrant it's own column. What do you think @Slayer95 and @bumbadadabum?

I should probably also change the title, I don't think I can change branch names though.

TheJetOU:
Bumping this.

bumbadadabum:
I forgot to respond to Slayer's comments but I personally think adding a start command and confirm command very much overcomplicate the in essence simplistic feature for very little gain. We don't have a confirmation for any other HTML displayed function that can be edited (roomfaqs, roomintros, etc) and I really don't see why roomevents need to be the exception. It doesn't actually need to be entered as UTC either, since you can add timezones in already. If we want to make putting in time more user-friendly, there are better ways to do it than to have a confirmation command. I think that is very much overkill.

TheJetOU:
@bumbadadabum, while a confirmation message might be a bit overkill, I think a start command is quite nice, notifying all the users in the room at the actual start date is quite nice IMO.

bumbadadabum:
Fair enough. That can stay I suppose.

TheJetOU:
@bumbadadabum

server/chat-plugins/room-events.js
```diff
@@ -22,15 +22,19 @@ exports.commands = {
...
+				buff += `<tr title="${isNaN(timeRemaining) ? timeRemaining : `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`}"><td>${Chat.escapeHTML(room.events[i].eventName)}</td><td>${Chat.formatText(room.events[i].desc, true)}</td><td><time>${Chat.escapeHTML(room.events[i].date)}</time></td></tr>`;
 			}
 			buff += '</table>';
 			return this.sendReply(`|raw|<div class="infobox-limited">${buff}</div>`);
 		},
 		new: 'add',
 		create: 'add',
 		edit: 'add',
-		add(target, room, user) {
+		add(target, room, user, connection, cmd) {
```
bumbadadabum: This is no longer needed

```suggestion
		add(target, room, user) {
```

server/chat-plugins/room-events.js
```diff
@@ -157,6 +184,7 @@ exports.commands = {
 	roomeventshelp: [
 		`/roomevents - Displays a list of upcoming room-specific events.`,
 		`/roomevents add [event name] | [event date/time] | [event description] - Adds a room event. A timestamp in event date/time field like YYYY-MM-DD HH:MM±hh:mm will be displayed in user's timezone. Requires: @ # & ~`,
+		`/roomevents start [event] - Declares to the room that [event] has started.`,
```
bumbadadabum:
```suggestion
		`/roomevents start [event name] - Declares to the room that the event has started.`,
```

server/chat-plugins/room-events.js
```diff
@@ -55,11 +59,34 @@ exports.commands = {
...
+			if (!this.can('ban', null, room)) return false;
+			if (!room.events || Object.keys(room.events).length === 0) {
+				return this.errorReply("There are currently no planned upcoming events for this room to start.");
+			}
+			if (!target) return this.errorReply("Usage: /roomevents start [event name]");
+			target = toId(target);
+			if (!room.events[target]) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
+			if (room.events[target].started) return this.errorReply(`The event ${target} has already started.`);
+			for (const u in room.users) {
+				if (Users(u).connected) Users(u).sendTo(room, `|notify|A new roomevent in ${room.title} has started!|The "${target}" roomevent has started!`);
```
bumbadadabum: I would use the `.eventName` and not `target` for displaying to users.

bumbadadabum: HTML escaped of course

server/chat-plugins/room-events.js
```diff
@@ -78,7 +78,7 @@ exports.commands = {
 			if (!room.events[target]) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
 			if (room.events[target].started) return this.errorReply(`The event ${target} has already started.`);
 			for (const u in room.users) {
-				if (Users(u).connected) Users(u).sendTo(room, `|notify|A new roomevent in ${room.title} has started!|The "${target}" roomevent has started!`);
+				if (Users(u).connected) Users(u).sendTo(room, Chat.html`|notify|A new roomevent in ${room.title} has started!|The "${room.events[target].eventName}" roomevent has started!`);
 			}
 			this.add(Chat.html`|raw|<div class="broadcast-blue"><b>The "${target}" roomevent has started!</b></div>`);
```
bumbadadabum: You forgot it here

TheJetOU:
I reverted 'toId -> toID' because I could not get tests to pass, so I'll just create another PR after this, I guess? Or maybe someone has a better solution.

TheJetOU:
Rebased, should be good now I think. @bumbadadabum?

PR #5460, opened 2019-04-21
Past Gens: Fix confusion interaction with self-destructing moves in gen 2

jmclemo6:
This should fix https://github.com/Zarel/Pokemon-Showdown/projects/3#card-18664063


Zarel:
Looks great! @Marty-D?

Marty-D:
Neat, thanks!

PR #5465, opened 2019-04-26
Fixes crash in gen3doubles due to missing target

### LINKED ISSUE #5464: TypeError: Cannot read property 'fainted' of undefined [gen3doubles]
```
Run `node tools/simulate exhaustive --cycles=1 --format=gen3doublescustomgame --seed=14286,2514,46811,43200`:
 TypeError: Cannot read property 'fainted' of undefined
    at Battle.useMoveInner (data/mods/gen3/scripts.js:135:29)
    at Battle.useMove (data/scripts.js:138:25)
    at Battle.runMove (data/scripts.js:93:31)
    at Battle.runAction (.sim-dist/battle.js:2564:9)
    at Battle.go (.sim-dist/battle.js:2784:9)
    at Battle.commitDecisions (.sim-dist/battle.js:2862:8)
    at Battle.choose (.sim-dist/battle.js:2819:35)
    at RawBattleStream._writeLine (.sim-dist/battle-stream.js:95:17)
    at RawBattleStream._writeLines (.sim-dist/battle-stream.js:67:10)
    at RawBattleStream._write (.sim-dist/battle-stream.js:55:9)
```

[Input Logs](https://gist.github.com/scheibo/83da7f801f08f2b2c339132058e7e12c)

---

@MacChaeger related to #5216?

scheibo:
Fixes #5464

Simple fix, but I didn't just commit it directly because I want someone confirm that `targets` being empty is actually expected and that there's not a deeper issue at play.

### DIFF
--- a/data/mods/gen3/scripts.js
+++ b/data/mods/gen3/scripts.js
@@ -132,7 +132,7 @@ let BattleScripts = {
 			if (move.spreadHit) this.attrLastMove('[spread] ' + hitTargets.join(','));
 		} else {
 			target = targets[0];
-			let lacksTarget = target.fainted;
+			let lacksTarget = !target || target.fainted;
 			if (!lacksTarget) {
 				if (move.target === 'adjacentFoe' || move.target === 'adjacentAlly' || move.target === 'normal' || move.target === 'randomNormal') {
 					lacksTarget = !this.isAdjacent(target, pokemon);


Zarel:
Looks intentional to me. I'll wait in case @Marty-D knows more, but if you don't, Marty, just merge it.

Marty-D:
The deeper issue in this case would be that it's not actually possible to intentionally target empty slots prior to Gen 5. The games will just not let you click a slot without a Pokemon in it. If you happen to be attacking an empty slot in Gen 3 or 4 because the intended target fainted before the user's action, single-target moves are redirected into an opponent instead. Helping Hand specifically will not even request a target at all in Gen 3, and in Gen 4 will only do so if your ally slot is currently occupied; if it isn't, Helping Hand just acts like it does if you used it in Singles and fails.

But until something happens with all that, this is useful to prevent these crashes. Good catch, thanks!

Zarel:
The answer there would be to generate a choice error if someone tries to target an empty slot?

PR #5466, opened 2019-04-28
Fix bug in /msg

TheJetOU:
Reported here: https://www.smogon.com/forums/posts/8114587

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -473,14 +473,15 @@ const commands = {
 	w: 'msg',
 	msg(target, room, user, connection) {
 		if (!target) return this.parse('/help msg');
+		if (!target.includes(',')) {
+			this.errorReply("You forgot the comma.");
+			return this.parse('/help msg');
+		}
 		target = this.splitTarget(target);
 		let targetUser = this.targetUser;
 		if (this.targetUsername === '~') {
 			this.room = Rooms.global;
 			this.pmTarget = null;
-		} else if (!target) {
-			this.errorReply("You forgot the comma.");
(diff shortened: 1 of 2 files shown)


server/chat.js
```diff
@@ -386,7 +386,7 @@ class MessageContext {
 		if (commaIndex < 0) {
 			return [target.trim(), ''];
 		}
-		return [target.substr(0, commaIndex).trim(), target.substr(commaIndex + 1).trim()];
+		return [target.slice(0, commaIndex).trim(), target.slice(commaIndex + 1).trim()];
```
TheJetOU: This is unnecessary for the fix, but I just thought it was clearer for future people reading the code.

server/chat-commands.js
```diff
@@ -473,14 +473,15 @@ const commands = {
...
+		if (!target.includes(',')) {
+			this.errorReply("You forgot the comma.");
+			return this.parse('/help msg');
+		}
 		target = this.splitTarget(target);
 		let targetUser = this.targetUser;
 		if (this.targetUsername === '~') {
 			this.room = Rooms.global;
 			this.pmTarget = null;
-		} else if (!target) {
```
Zarel: This check still does need to be done, just with a different message, like "You can't leave your message empty."

TheJetOU: Isn't that handled here? https://github.com/Zarel/Pokemon-Showdown/blob/d6b4705870849d54543038321d3b5c898beea0b6/server/chat.js#L1162-L1165

Zarel: Good point. Yeah, that should be good.

PR #5467, opened 2019-04-30
Merge global type declarations into dex-data

scheibo:
Currently the PR title is purely aspirational, as `tsc` seems to not want this to happen. I've attempted to migrate `Ability` as a strawman, and I run into the [following errors](https://gist.github.com/scheibo/f7b32f535b289643ca9b025928153f53):

```
...
data/abilities.js:321:17 - error TS2339: Property 'secondaries' does not exist on type 'Effect'.
  Property 'secondaries' does not exist on type 'Ability'.

321  			if (!effect.secondaries) this.add("-fail", target, "unboost", "Defense", "[from] ability: Big Pecks", "[of] " + target);
     			            ~~~~~~~~~~~
...
```

I can't figure out why it can't follow `Ability -> AbilityData -> EffectData` or `Ability -> BasicEffect -> EffectData`, but unless we can figure this out, unifying these two files won't happen :(. Any suggestions for what to try to resolve this would be useful, as I'm fairly confident that if we can unify `Ability` we can unify all of the rest.

### DIFF
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -451,6 +451,21 @@ export class Item extends BasicEffect implements Readonly<BasicEffect & ItemData
 	}
 }
 
+interface AbilityEventMethods {
+	onCheckShow?: (this: Battle, pokemon: Pokemon) => void
+	onEnd?: (this: Battle, target: Pokemon & Side & Field) => void
+	onPreStart?: (this: Battle, pokemon: Pokemon) => void
+	onStart?: (this: Battle, target: Pokemon) => void
+}
+
+export interface AbilityData extends EffectData, AbilityEventMethods, EventMethods {
+	rating: number
+	isUnbreakable?: boolean
+	suppressWeather?: boolean
+}
+
+export interface ModdedAbilityData extends Partial<AbilityData>, ModdedEffectData {}
+
 export class Ability extends BasicEffect implements Readonly<BasicEffect & AbilityData> {
 	readonly effectType: 'Ability';
 	/** Represents how useful or detrimental this ability is. */
--- a/sim/globals.ts
+++ b/sim/globals.ts
@@ -6,6 +6,10 @@ type PRNGSeed = import('./prng').PRNGSeed;
 type Side = import('./side').Side
 type Validator = import('./team-validator').Validator
 
+type Ability = import('./dex-data').Ability
+type AbilityData = import('./dex-data').AbilityData
+type ModdedAbilityData = import('./dex-data').ModdedAbilityData
+
 interface AnyObject {[k: string]: any}
 type DexTable<T> = {[key: string]: T}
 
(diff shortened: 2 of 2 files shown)


Zarel:
I don't think `implements` does anything but check compatibility. That's probably what's going on here. Honestly, the whole `.secondaries` hack itself should probably be fixed.

Zarel:
Background: `.secondaries` is only supposed to exist on `Move`s. We added it to `BasicEffect` because some code was reading it on arbitrary effects and we were too lazy to change it.

Zarel:
`if (!effect.secondaries)` can probably be replaced with `if (effect.effectType !== 'Move')`.

scheibo:
`secondaries` is not really the issue here, it just happens to be the first thing `tsc` complains about. If you look at the full logs linked in the description (or patch locally with '`git apply <(curl -L https://github.com/Zarel/Pokemon-Showdown/pull/5467.patch)`' and try it yourself) you'll see it complains about other fields as well.

scheibo:
I'm wondering if there's some circular require issue between `globals` and `dex-data` that is just hidden by a bad error message? I ran into subtle circular dependencies with `state.ts` that I had to resolve, and the error message which similarly confusing.

scheibo:
https://gist.github.com/scheibo/4ca5cd2678212ddd9eb82ce2b4e22dc3 is my minimal repro case, I get

```
src/index.ts:34:23 - error TS2339: Property 'ignoreKlutz' does not exist on type 'Effect'.
  Property 'ignoreKlutz' does not exist on type 'Ability'.

34    console.log(effect.ignoreKlutz ? 'item' : 'ability')
                         ~~~~~~~~~~~


Found 2 errors.
```

Which I think is the same problem as we have here?

scheibo:
OK, pretty sure this is just the manifestion of a **circular dependency issue** with a terrible error message. I ran into circular deps before with the `state.ts` `toJSON` PR and had to figure out a workaround, but the errors are incredibly cryptic (though different than these. Typescript seems to want to pretend that circular dependency issues aren't a thing, so just fails in mysterious ways when the problem crops up instead of, you know, telling you that theres a fucking circular dep issue).

https://gist.github.com/b15a7f959436390c6626e761b6f23a2a is a minimal repro. Run that, and you get:

```
src/index.ts:34:23 - error TS2339: Property 'secondaries' does not exist on type 'Effect'.
  Property 'secondaries' does not exist on type 'Ability'.

34    console.log(effect.secondaries ? 'item' : 'ability')
                         ~~~~~~~~~~~

src/index.ts:39:15 - error TS2339: Property 'secondaries' does not exist on type 'BasicEffect'.

39 console.log(b.secondaries);
                 ~~~~~~~~~~~


Found 4 errors.
````

*(4 errors? I count two, you stupid compiler, but thats beside the point)*

If you collapse everything into one file everything compiles.

scheibo:
I don't think I can do this piecemeal. I think I'm going to need to pretty much untangle all of `sim/globals.ts` completely such that type definitions live in more approriate files and are imported where needed instead of being globals definitions floating around. That was my end goal anyway, I was just hoping it could be done incrementally.

Zarel:
Yeah, that massively sucks. I guess you can omit the `implements` part if that makes the incremental part simpler; you lose some safety but it's only temporary, right?

PR #5468, opened 2019-04-30
Introduce an ID type to sim

scheibo:
Similar to `ID` in Pokemon-Showdown-Client, but also including `''` to avoid a bunch of casts. More places can likely to converted to `ID`, and eventually certain `toId` calls can be removed, but this commit just lays the groundwork.

### DIFF
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -37,14 +37,14 @@ export class Tools {
 	 * Dex.getId is generally assigned to the global toId, because of how
 	 * commonly it's used.
 	 */
-	static getId(text: any): string {
+	static getId(text: any): ID {
 		if (text && text.id) {
 			text = text.id;
 		} else if (text && text.userid) {
 			text = text.userid;
 		}
 		if (typeof text !== 'string' && typeof text !== 'number') return '';
-		return ('' + text).toLowerCase().replace(/[^a-z0-9]+/g, '');
+		return ('' + text).toLowerCase().replace(/[^a-z0-9]+/g, '') as ID;
 	}
 }
 const toId = Tools.getId;
@@ -56,7 +56,7 @@ export class BasicEffect implements EffectData {
 	 * becomes "mrmime", and "Basculin-Blue-Striped" becomes
 	 * "basculinbluestriped".
 	 */
-	id: string;
+	id: ID;
 	/**
 	 * Name. Currently does not support Unicode letters, so "Flabébé"
 	 * is "Flabebe" and "Nidoran♀" is "Nidoran-F".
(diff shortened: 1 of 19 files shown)


Zarel:
I'd want to avoid `/** @type {ID} */ (` basically everywhere... I was going to delay this until `as ID` was good enough.

sim/dex-data.ts
```diff
@@ -141,8 +141,8 @@ export class BasicEffect implements EffectData {
 		this.duration = data.duration;
 		this.noCopy = !!data.noCopy;
 		this.affectsFainted = !!data.affectsFainted;
-		this.status = data.status || undefined;
-		this.weather = data.weather || undefined;
+		this.status = toId(data.status) || undefined;
```
Zarel: I'm not a huge fan of unnecessary `toId` calls; Dex constructors are an inner loop.

sim/pokemon.ts
```diff
@@ -50,10 +50,10 @@ export class Pokemon {
 
 	baseTemplate: Template;
 	template: Template;
-	speciesData: AnyObject;
+	speciesData: AnyObjectWithID;
```
Zarel: I was thinking about removing the ID, actually.

Zarel: (I don't think we've ever needed to read it in the past eight years of supporting it.)

scheibo:
> I'd want to avoid `/** @type {ID} */ (` basically everywhere... I was going to delay this until `as ID` was good enough.

Excluding `gennext`, its pretty minimal, all things considered?

sim/dex-data.ts
scheibo: Yeah, my problem was `EffectData` can't be given the `ID` type (without `/** @type {ID} */(...)` casts EVERYWHERE, or `as ID` when Typescript-ed).

However, maybe we can just cast here instead of `toId` and just assume (or add a test to validate) that the data files are in fact correct IDs?

Zarel:
I guess I'm okay with this.

sim/dex-data.ts
scheibo: Changes this to a cast

sim/pokemon.ts
scheibo: Removed `AnyObjectWithID` typing. Actually removing the `id: ''` field is better saved for a future change (maybe one which also renames them to `fooState` like I believe you requested in the past?)

scheibo:
- removed `AnyObjectWithID`
- now casting in `BasicEffect` instead of calling `toId`

> I guess I'm okay with this.

Should be ready to go now, please merge or approve :)

scheibo:
Sneak peak: I'll be sending out https://github.com/pkmn-cc/Pokemon-Showdown/compare/sim-id2...pkmn-cc:getEffect immediately after this PR goes in to start minimizing `toId` usage where currently possible (though we can surely eliminate a lot more with more work)

sim/pokemon.ts
Zarel: I approve of that.

Zarel:
me:

> I'd want to avoid `/** @type {ID} */ (` basically everywhere... I was going to delay this until `as ID` was good enough.

you:

> Yeah, my problem was `EffectData` can't be given the `ID` type (without `/** @type {ID} */(...)` casts EVERYWHERE, or `as ID` when Typescript-ed).

Oh, so _that's_ how you avoided "`/** @type {ID} */ (` basically everywhere"

PR #5469, opened 2019-04-30
Notify users the first time a PM/challenge is blocked

scheibo:
> One solution would be to show a warning the first time a PM or challenge is blocked, every session.

_Originally posted by @Zarel in https://github.com/Zarel/Pokemon-Showdown-Client/issues/1271#issuecomment-488094948_

I'm not entirely sure this code does exactly what we want because it isn't completely clear to me what is meant by 'session' in this case (should the notifications be once-per-connection instead? Is there another way to handle sessions?). Also, not sure if the messages should just be plain or if we want `|error|` or something.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -552,10 +552,7 @@ const commands = {
 		if (!targetUser || !targetUser.connected) return this.errorReply(`User ${this.targetUsername} is not currently online.`);
 		if (!(targetUser in room.users) && !user.can('addhtml')) return this.errorReply("You do not have permission to use this command to users who are not in this room.");
 		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) {
-			if (!targetUser.blockPMsNotified) {
-				targetUser.send(`The user '${user.name}' attempted to PM you but was blocked. To enable PMs, use /unblockpms.`);
-				targetUser.blockPMsNotified = true;
-			}
+			Chat.maybeNotifyBlocked('pm', targetUser, user);
 			return this.errorReply("This user is currently blocking PMs.");
 		}
 		if (targetUser.locked && !user.can('lock')) return this.errorReply("This user is currently locked, so you cannot send them a pminfobox.");
@@ -584,10 +581,7 @@ const commands = {
 		if (!targetUser || !targetUser.connected) return this.errorReply(`User ${this.targetUsername} is not currently online.`);
 		if (!(targetUser in room.users) && !user.can('addhtml')) return this.errorReply("You do not have permission to use this command to users who are not in this room.");
 		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) {
-			if (!targetUser.blockPMsNotified) {
-				targetUser.send(`The user '${user.name}' attempted to PM you but was blocked. To enable PMs, use /unblockpms.`);
-				targetUser.blockPMsNotified = true;
-			}
+			Chat.maybeNotifyBlocked('pm', targetUser, user);
 			return this.errorReply("This user is currently blocking PMs.");
 		}
 		if (targetUser.locked && !user.can('lock')) return this.errorReply("This user is currently locked, so you cannot send them UHTML.");
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -551,7 +551,13 @@ const commands = {
 
 		if (!targetUser || !targetUser.connected) return this.errorReply(`User ${this.targetUsername} is not currently online.`);
 		if (!(targetUser in room.users) && !user.can('addhtml')) return this.errorReply("You do not have permission to use this command to users who are not in this room.");
-		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) return this.errorReply("This user is currently blocking PMs.");
+		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) {
+			if (!targetUser.blockPMsNotified) {
+				targetUser.send(`The user '${user.name}' attempted to PM you but was blocked. To enable PMs, use /unblockpms.`);
+				targetUser.blockPMsNotified = true;
+			}
+			return this.errorReply("This user is currently blocking PMs.");
+		}
 		if (targetUser.locked && !user.can('lock')) return this.errorReply("This user is currently locked, so you cannot send them a pminfobox.");
 
 		// Apply the infobox to the message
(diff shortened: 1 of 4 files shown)


server/chat-commands.js
```diff
@@ -551,7 +551,13 @@ const commands = {
 
 		if (!targetUser || !targetUser.connected) return this.errorReply(`User ${this.targetUsername} is not currently online.`);
 		if (!(targetUser in room.users) && !user.can('addhtml')) return this.errorReply("You do not have permission to use this command to users who are not in this room.");
-		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) return this.errorReply("This user is currently blocking PMs.");
+		if (targetUser.blockPMs && targetUser.blockPMs !== user.group && !user.can('lock')) {
+			if (!targetUser.blockPMsNotified) {
+				targetUser.send(`The user '${user.name}' attempted to PM you but was blocked. To enable PMs, use /unblockpms.`);
```
Zarel: I would open the Settings menu.

scheibo: Done.

server/chat.js
```diff
@@ -1933,6 +1934,27 @@ Chat.fitImage = async function (url, maxHeight = 300, maxWidth = 300) {
...
+ * Notifies a targetUser that a user was blocked from reaching them due to a setting they have enabled.
+ * @param {'pm'|'challenge'} blocked
+ * @param {User} targetUser
+ * @param {User} user
+ */
+Chat.maybeNotifyBlocked = function (blocked, targetUser, user) {
+	const options = 'or change it in the <button name="openOptions" class="subtle">Options</button> menu in the upper right.';
+	if (blocked === 'pm') {
+		if (!targetUser.blockPMsNotified) {
+			targetUser.send(`|html|The user '${user.name}' attempted to PM you but was blocked. To enable PMs, use /unblockpms ${options}`);
```
Zarel: Err, this sends to Lobby and will do nothing if Lobby isn't open.

Zarel: You probably want a PM from `~`, but that'll definitely notify. Maybe add some hardcoding to make it not notify?

sparkychildcharlie:
want to bring up that this applies for the first user that pms/challenges you, and any further pms/challenges even from another user wont notify.

this should also probably remove those properties used for checking if the user was notified if the user _unblocks_ pms/challenges

Zarel:
> this should also probably remove those properties used for checking if the user was notified if the user unblocks pms/challenges

Nah, it's fine to retain. If a user explicitly turns on blocking, they probably know why.

scheibo:
> want to bring up that this applies for the first user that pms/challenges you, and any further pms/challenges even from another user wont notify.

That is the intention? (_"One solution would be to show a warning the **first** time a PM or challenge is blocked"_)

> this should also probably remove those properties used for checking if the user was notified if the user _unblocks_ pms/challenges

> Nah, it's fine to retain. If a user explicitly turns on blocking, they probably know why.

"explicitly turns on blocking" -> there's no other way to turn on blocking, unless youre referring to the persistent block challenge toggle, which someone would have to explicitly toggle in the past in order to block challenges...? I think I agree with @sparkychildcharlie that we should be clearing these after unblock for consistency.

server/chat.js
scheibo: getting a PM to tell you that youre blocking PMs is kind of annoying. is there a way I can just make this show up in the chat of all the rooms the user is in?

Zarel:
> That is the intention? (_"One solution would be to show a warning the **first** time a PM or challenge is blocked"_)

+1. To confirm, that's my intention too.

> "explicitly turns on blocking" -> there's no other way to turn on blocking, unless youre referring to the persistent block challenge toggle, which someone would have to explicitly toggle in the past in order to block challenges...? I think I agree with @sparkychildcharlie that we should be clearing these after unblock for consistency.

I'm referring to the persistent blockchallenge toggle, yes.

There are several problems with "someone would have to explicitly toggle in the past in order to block challenges" – one is that misclicks become a much longer-term problem; another is that sometimes friends will maliciously change settings on computers they don't own as a prank.

server/chat.js
scheibo: @Zarel

Zarel: All the rooms? That feels even worse to me. Some sort of "notification" that doesn't notify would be nice...

scheibo: I use the single panel layout, and a PM just causes the 'Home' tab to light up and its annoying to have to switch tabs to get a warning about something I don't care about (I don't have 'friends maliciously changing settings on my computer'). The only logical place in the single panel UI to show a message is in the chat stream, which is why I suggested that. Ideally the message would be sent to only the room that you're "in" (or PM if the user is in no rooms/battles), but I don't think the server knows which room a particular user is in at any given time (which wouldn't even work, because users can switch rooms and miss the message).

Failing any better UI for notification, a specially cased PM from `~` is I guess one solution, but I'm not super chuffed about it

Zarel: Hm, a popup would save you the need to switch rooms.

Zarel: I really don't like popups for non-user-initiated actions, though.

Zarel: Specially cased PM might be the best solution here.

scheibo: I've got `|pm|~|` working, but I need a way to distinguish the 'notification' PM in some way to know not to notify for it (string matching on the message would work, but seems super kludge)

Zarel: I guess a new message type `/nonotify`?

scheibo: Done (depends on Zarel/Pokemon-Showdown-Client#1280 which would need to land first).

Zarel:
Client part merged; server part will take longer.

scheibo:
> Client part merged; server part will take longer.

To review? Or is there something concrete youve mentioned already that I'm missing?

Zarel:
There'll be a lot of users still using the old client. I usually either time with a restart or wait a few days for everyone to update.

scheibo:
Gotcha. However, this is just part 1 of solving Zarel/Pokemon-Showdown-Client#1271, the second part was to add client support to ignore the defaults the server sends and always override initially with `Dex.prefs`. If the client and server are going to be pushed at the same time, should I also open a PR on the client for that, so the next time you restart you could merge in both the client and server pending PRs and deliver persistent settings all at once on the synchronized push, or would you rather wait for two client push cycles (for the non-notifying PM, and a second one for persistent notifications)?

Zarel:
Sure, you can open the other PR. I can either sync them if we restart, or wait for two cycles if we don't.

scheibo:
Remind me why this is blocked on a restart? The client changes have landed and been deployed to the best of my knowledge, so this should be fine to land now, no?

PR #5470, opened 2019-05-01
Other Metagames of the Month May 2019

urkerab:
It passes tsc, but I've not run this version live, so you should triple-check it...


TheImmortal:
I've tested and they both seem to work as intended.

sim/globals.ts
```diff
@@ -1049,10 +1050,11 @@ interface FormatsData extends EventMethods {
 	onBegin?: (this: Battle) => void
 	onChangeSet?: (this: ModdedDex, set: PokemonSet, format: Format, setHas?: AnyObject, teamHas?: AnyObject) => string[] | void
 	onModifyTemplate?: (this: Battle, template: Template, target: Pokemon, source: Pokemon, effect: Effect) => Template | void
+	onStart?: (this: Battle) => void
```
Slayer95: This doesn't seem to be needed. Am I missing something?

Issue #5453, opened 2019-04-17
Improve Endless Battle Clause


scheibo:
https://gist.github.com/scheibo/534d9bb858a0824082a3b9dce1fee61c (discovered while I was implementing the `RandomPlayerAI` for my harness) and https://replay.pokemonshowdown.com/gen7uu-894258335 are two separate examples of Endless Battle Clause triggering prematurely due to a Pokemon which was frozen erroneously getting penalized.

```
13:03:52 |c| pre (away)|the Kingdra would eventually thaw out (its not Gen 1), so it seems like the battle should have been allowed to continue
13:06:36 |c|+Volco|imo it should exclude status aliments with the exception of sleep by rest because rest with leppa would be endless, but that's just me
...
13:25:27 |c|@MacChaeger|Also, I believe I once triggered EBC using two paralyzed Slakings
13:29:15 |c|@DaWoblefet|I've triggered Endless Battle Clause from Heal Pulsing an opponent in doubles
13:29:27 |c|@DaWoblefet|no Leppas involved
```

I think statuses (paralyzed/sleeping/frozen/confusion/attraction) should not trigger EBC.

scheibo:
I'm curious about the `ppstall` `staleSource` in general (what this gets classified as) - what are some examples of what a user could do maliciously to trigger it?

scheibo:
https://github.com/Zarel/Pokemon-Showdown/blob/master/sim/battle.ts#L1450 also doesn't show up in the logs - even if we don't change the implementation to ignore status, we should at least be warning so the user can change their action if possible

Slayer95:
Re: specification

https://www.smogon.com/forums/threads/endless-battle-clause.3495354/
https://www.smogon.com/forums/threads/endless-stall-clause.3610486/
https://www.smogon.com/forums/threads/endless-battle-issues.3596079/
https://www.smogon.com/forums/threads/endless-battle-clause-for-ps-ladder.3624183/

Have fun!

Slayer95:
Note that once we have a concrete proposal to update the Endless Battle Clause, it would need to go through the Policy Review forum, and that the Smogon guys not necessarily care about edge cases triggered by an AI doing strictly random choices.

xfix:
In my opinion, this is a bugfix to Endless Battle Clause rather than a specification change.

scheibo:
[...]

PR #5472, opened 2019-05-02
Redo Endless Battle Clause with a new specification

scheibo:
Based on discussion in #5453, this change implements the core of the new EBC:

- Any Pokémon hit by a foe’s Heal Pulse or Floral Healing becomes stale.
- Any Pokémon _using_ (consuming) an _acquired Leppa Berry_ (defined as a Leppa Berry which the Pokemon did not start the battle with, for example, one obtained Fling, Trick, Harvest, Recycle, Bug Bite, Pluck, Pickup etc) becomes stale.
- If all Pokémon on the field are stale and at least one team does not have the option of switching to a non-stale Pokémon, with at least one Pokémon on a side's staleness having been inflicted by a Pokémon of the opposing side, the game ends. If only one side started the battle with the _rudimentary means to perform Leppa Berry-cycling_ (defined to be a team containing a Pokémon holding a Leppa Berry and a Pokémon with Recycle/Harvest/Pickup, it loses, otherwise the battle ends in a tie.

---

The remaining conditions still need to be implemented

- Moves selected by Metronome do not affect staleness in any way (ie. acquiring a Leppa Berry through Metronome will not count as having acquired a Leppa Berry for the purposes of staleness tracking when it is later consumed, nor will Heal Pulse via Metronome confer the stale state to a foe Pokémon).
- If all Pokémon on the field are forced to use _Struggle ineffectively_ (where the damage from Struggle does not result in their opponent losing HP by the end of the turn) and at least one team does not have the option of switching to Pokémon that could do something other than Struggle ineffectively, the battle ends in a tie.

https://github.com/Zarel/Pokemon-Showdown/issues/5453#issuecomment-488847446 brings up the possibility for several changes to conditions (I've already taken the liberty of deciding that Pickup should be considered in the same class as Harvest).

When the spec is finalized, I also intend to have us host it on https://pokemonshowdown.com/pages/ebc (or similar), instead of having it simply being 'implementation-defined'.

### DIFF
--- a/data/items.js
+++ b/data/items.js
@@ -2944,10 +2944,6 @@ let BattleItems = {
 			moveSlot.pp += 10;
 			if (moveSlot.pp > moveSlot.maxpp) moveSlot.pp = moveSlot.maxpp;
 			this.add('-activate', pokemon, 'item: Leppa Berry', moveSlot.move, '[consumed]');
-			if (pokemon.item !== 'leppaberry') {
-				pokemon.isStale = 2;
-				pokemon.isStaleSource = 'useleppa';
-			}
 		},
 		num: 154,
 		gen: 3,
@@ -6374,10 +6370,6 @@ let BattleItems = {
 			moveSlot.pp += 5;
 			if (moveSlot.pp > moveSlot.maxpp) moveSlot.pp = moveSlot.maxpp;
 			this.add('-activate', pokemon, 'item: Mystery Berry', moveSlot.move);
-			if (pokemon.item !== 'leppaberry') {
-				pokemon.isStale = 2;
-				pokemon.isStaleSource = 'useleppa';
-			}
 		},
 		num: 154,
 		gen: 2,
--- a/data/mods/gen1/scripts.js
+++ b/data/mods/gen1/scripts.js
@@ -127,8 +127,6 @@ let BattleScripts = {
 			const moveSlot = pokemon.moveSlots.find(moveSlot => moveSlot.id === move.id);
 			if (moveSlot && moveSlot.pp < 0) {
 				moveSlot.pp = 63;
-				pokemon.isStale = 2;
-				pokemon.isStaleSource = 'ppoverflow';
 				this.hint("In Gen 1, if a player is forced to use a move with 0 PP, the move will underflow to have 63 PP.");
 			}
 		}
--- a/data/mods/gen3/scripts.js
+++ b/data/mods/gen3/scripts.js
@@ -67,7 +67,6 @@ let BattleScripts = {
 		if (!target) {
 			this.attrLastMove('[notarget]');
 			this.add('-notarget', pokemon);
-			if (move.target === 'normal') pokemon.isStaleCon = 0;
 			return false;
 		}
 
@@ -141,7 +140,6 @@ let BattleScripts = {
 			if (lacksTarget && !move.isFutureMove) {
 				this.attrLastMove('[notarget]');
 				this.add('-notarget', pokemon);
-				if (move.target === 'normal') pokemon.isStaleCon = 0;
 				return false;
 			}
 			damage = this.tryMoveHit(target, pokemon, move);
--- a/data/moves.js
+++ b/data/moves.js
@@ -5598,12 +5598,15 @@ let BattleMovedex = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1, reflectable: 1, heal: 1, mystery: 1},
-		onHit(target) {
+		onHit(target, source) {
 			let success = false;
 			if (this.field.isTerrain('grassyterrain')) {
 				success = !!this.heal(this.modify(target.maxhp, 0.667)); // TODO: find out the real value
 			} else {
 				success = !!this.heal(Math.ceil(target.maxhp * 0.5));
+			}
+			if (success && target.side.id !== source.side.id) {
+				target.staleness = 'external';
 			}
 			return success;
 		},
@@ -7355,6 +7358,9 @@ let BattleMovedex = {
 				success = !!this.heal(this.modify(target.maxhp, 0.75));
 			} else {
 				success = !!this.heal(Math.ceil(target.maxhp * 0.5));
+			}
+			if (success && target.side.id !== source.side.id) {
+				target.staleness = 'external';
 			}
 			return success;
 		},
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -565,30 +565,7 @@ let BattleFormats = {
 		effectType: 'Rule',
 		name: 'Endless Battle Clause',
 		desc: "Prevents players from forcing a battle which their opponent cannot end except by forfeit",
-		// implemented in sim/battle.js
-
-		// A Pokémon has a confinement counter, which starts at 0:
-		// +1 confinement whenever:
-		// - it has no available moves other than Struggle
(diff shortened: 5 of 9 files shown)


scheibo:
Please see https://github.com/Zarel/Pokemon-Showdown/issues/5453#issuecomment-489184022 regarding the conditions I left out - I don't feel they (or the other cases the new spec misses) justify the code complexity required. If the community doesn't want to wait 1000 turns for their battles to end in ties, `/offerdraw` from #5454 seems like a good shortcut (or revisiting a battle time limit).

Zarel:
How complicated is the spec? I feel like the best place to host it is a doc comment in `rulesets.js`. I'd accept it as a `.md` file in `data/` if you wanted someplace better.

Zarel:
I explicitly rejected `/offerdraw` as a solution for endless battles for one reason: it's only a viable option if both players are _capable_ of playing a game to a close. If one player needs to leave in a shorter timeframe than three hours (previous obligation, etc), the other player is incentivized to reject the draw, and win on grounds other than game skill.

scheibo:
> How complicated is the spec? I feel like the best place to host it is a doc comment in `rulesets.js`. I'd accept it as a `.md` file in `data/` if you wanted someplace better.

The spec is currently just the 3 conditions outlined in the description of this PR. The length/complexity of the spec is not why I don't want it to exclusively in the codebase, the reasons are to (a) accurately list the rules that govern the game that is being played in a human readable way and (b) distinguish between implementation and specification (where an edge case involving endless battles may be the result of a poor specification or a poor implementation of said specification). 

TBH, I'd actually like *all* our rules to be better defined (eg. outline in detail what is meant by 'putting more than one of their opponent's Pokemon to sleep at a time' for Sleep Clause - what does this mean for Effect Spore etc?), EBC is just the the one I happened to focus on now (and which has the potential to be the most hairy, especial with its implementation prior to the commit). Basically, I think its important to have a specification for all rules in English that players can read/understand/shape and then code which attempts to implementation said specification, not the opaque implementation-is-the-specification we have today.

> `/offerdraw`

You're right, its the wrong solution.

Taking a step back about why I even started this quest to change EBC in the first place: I view false positives (prematurely ending battles and declaring a winner) to be considerably worse than allowing for the possibility of endless battles. Especially so if the players getting dinged can't reason about/plan around the circumstances because they don't know how it works (or can't know how it works, because its too complex).

The revised specification aims to be straightforward to understand and implement (especially given I'm deliberately cutting out the Metronome/ineffective Struggle logi), and seems to cut out a lot of false positives (its still too complicated for me to claim that there will never be a false positive, but I imagine any that exist would be fairly niche), however, has the undesirable effect of not covering a lot of the edge cases the original implementation does (https://github.com/Zarel/Pokemon-Showdown/issues/5453#issuecomment-488847446 covers some of them), potentially leaving players to play out 1000 turns fairly mindlessly to end in a tie (against an opponent who might be utilizing his/her timer to the fullest extent possible, potentially maliciously).

**Can we do better than 1000 turns?** 
- Why 1000 turns? Can we instead use data from the logs to determine what an appropriate cut off would be to ensure 99% (or whatever arbitrary N you want) will terminate before then? (My gut feeling is that 1000 turns is currently very long tail). The turn limit could even be per generation, to provide more of a buffer for the comparitively longer games in the earlier gens.
- I feel it might be worth distinguishing between the limit for ladder vs. tournament play vs. custom play. The metagames in ladder vs. tournament play are already different, I don't see why we couldn't add a clock time deadline on ladder games vs. tournament games (a malicious Leppa abuser might win battles in ladder play simply by players forfeitting after turn 200 because they dont want to waste their day playing out 800 more turns for a tie just to avoid losing 20 rating points, but a player is a tournament is more likely to be willing to stick it out to the end because a single loss could mean elimination).
- Actually [hammer out a 50 turn rule](https://www.smogon.com/forums/threads/hammering-out-a-fifty-turn-rule-proposal.3610169/page-2) equivalent, accepting that it will disallow valid battles (which the current EBC before the PR also definitely does, as evidenced by the replays in #5453). Maybe there can be a gating condition involved, maybe not (eg. a Leppa Berry is involved).

This all might benefit from Policy Review.

Zarel:
I do want to merge this and _then_ refine via Policy Review. The impression I get is that the current EBC implementation has a number of bugs introduced by recent refactors.

Zarel:
The main reason I haven't insta-merged is that I get the impression the new clause also has a number of false positives, mainly to do with someone bringing a Leppacycler and accidentally getting the Leppa relocated.

Zarel:
I might be tentatively fine with merging and reacting quickly to bug reports, though. Leppacycling itself is pretty rare.

scheibo:
> The impression I get is that the current EBC implementation has a number of bugs introduced by recent refactors.

Almost certainly

> The main reason I haven't insta-merged is that I get the impression the new clause also has a number of false positives, mainly to do with someone bringing a Leppacycler and accidentally getting the Leppa relocated.

Can you thing of a concrete scenario I can add as a test case?

> I might be tentatively fine with merging and reacting quickly to bug reports, though. Leppacycling itself is pretty rare.

I think this is mergeable and will have strictly less false positives than the current implementation at the cost of making it much easier for malicious players to force their opponents into a decision of 'sit here for 7 hours and mindlessly play out this game to get a tie or forfeit'.

Zarel:
> Can you thing of a concrete scenario I can add as a test case?

Not easily. Let's just wait until someone reports an EBC bug. :p

> I think this is mergeable and will have strictly less false positives than the current implementation at the cost of making it much easier for malicious players to force their opponents into a decision of 'sit here for 7 hours and mindlessly play out this game to get a tie or forfeit'.

_Two and a half_ hours, actually. I calculated this out when we settled on 1000 turns.

Zarel:
> The spec is currently just the 3 conditions outlined in the description of this PR. The length/complexity of the spec is not why I don't want it to exclusively in the codebase, the reasons are to (a) accurately list the rules that govern the game that is being played in a human readable way and (b) distinguish between implementation and specification (where an edge case involving endless battles may be the result of a poor specification or a poor implementation of said specification).
> 
> TBH, I'd actually like _all_ our rules to be better defined (eg. outline in detail what is meant by 'putting more than one of their opponent's Pokemon to sleep at a time' for Sleep Clause - what does this mean for Effect Spore etc?), EBC is just the the one I happened to focus on now (and which has the potential to be the most hairy, especial with its implementation prior to the commit). Basically, I think its important to have a specification for all rules in English that players can read/understand/shape and then code which attempts to implementation said specification, not the opaque implementation-is-the-specification we have today.

I still don't understand what this has to do with hosting the spec on the PS CMS instead of in a doc comment or in GitHub?

scheibo:
> I don't disagree with any of this, but I still don't understand what this has to do with hosting the spec on the PS CMS instead of in a doc comment or in GitHub?

I suggested PS because it seems the most user friendly place (besides maybe a page on the Smogon dex) for specifications of rules, the same way it explains various mechanics which are also critical for understanding the game. Basically, I don't see why these rules would be treated any different than item/move descriptions (we don't link people to `data/moves.js` to explain about how particular moves work). A markdown doc in the repo would be 'OK', but it still feels a little too 'behind the scenes' ?

Zarel:
Hm, PSdex is probably the right place, then.

https://github.com/Zarel/Pokemon-Showdown-Dex/tree/master/articles

Which can be accessed in the Dex.

PR #5474, opened 2019-05-03
Fix Knock Off And Rapid Spin

TheJetOU:
Knock Off and Rapid Spin should still take effect even if the user faints in Generation 3 and Generation 4,

Ignore the branch name.


Marty-D:
Thanks!

Issue #697, opened 2014-02-27
Bringing one Pokémon to doubles causes side.active[1] to be null


Zarel:
So, the current way PS works is, if you bring one Pokémon to doubles, `side.pokemon` is an array of length 1, and `side.active` is an array of length 2, with `side.active[1] === null`.

Approximately 90% of our last 20 crashes have been because one of our coders (cough hi @Slayer95 @TheImmortal @Zarel) forgot that `side.active[1]` can be `null`, and write something that ends up being `side.active[1].fainted`, which crashes.

This is actually why the Tools getter functions like `Tools.getAbility` always return an object, even for abilities that don't exist. You can still do `this.ability === Tools.getAbility(input).id` or whatever even if it's not a real ability.

The question is, do we want to do the same here? Instead of `null`, we could just have a placeholder object, like just `{fainted:true}`, which would allow us to test for `pokemon.fainted` and `!pokemon.hp` without having to test for existence first.

PR #5475, opened 2019-05-04
Avoid calling toId where possible with Dex.getEffectByID

scheibo:
**This changes reduces `toID` calls by ~15.7% (from `10238733` -> `8628355` over 100 random battles).**

After #5468, we can now start to avoid unnecessary `toId` calls in critical places. This commit avoids `toId` in certain obvious places by introducing a `Dex.getEffectByID` method which can be used internally to skip both `toId` and unnecessary lookups. This commit also performs a minor optimization by passing `id` to `Data` constructors (which saves a call to `toId` in the `BasicEffect` constructor, and which we could have done before #5468 introduced an `ID` type, I only just noticed it now).

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -407,7 +407,7 @@ class ModdedDex {
 			if (!template.tier) template.tier = 'Illegal';
 			if (!template.doublesTier) template.doublesTier = template.tier;
 		} else {
-			template = new Data.Template({name, exists: false});
+			template = new Data.Template({id, name, exists: false});
 		}
 		if (template.exists) this.templateCache.set(id, template);
 		return template;
@@ -439,7 +439,7 @@ class ModdedDex {
 		if (id && this.data.Movedex.hasOwnProperty(id)) {
 			move = new Data.Move({name}, this.data.Movedex[id]);
 		} else {
-			move = new Data.Move({name, exists: false});
+			move = new Data.Move({id, name, exists: false});
 		}
 		if (move.exists) this.moveCache.set(id, move);
 		return move;
@@ -486,22 +486,30 @@ class ModdedDex {
 			// @ts-ignore
 			return effect;
 		}
+		return this.getEffectByID(id, effect);
+	}
+
+	getEffectByID(id: ID, effect?: Effect | Move): PureEffect {
+		if (!id) return nullEffect;
+
+		if (!effect) effect = this.effectCache.get(id);
+		if (effect) return effect as PureEffect;
 
 		let found;
 		if (this.data.Formats.hasOwnProperty(id)) {
-			effect = new Data.Format({name}, this.data.Formats[id]);
+			effect = new Data.Format({name: id}, this.data.Formats[id]);
 		} else if (this.data.Statuses.hasOwnProperty(id)) {
-			effect = new Data.PureEffect({name}, this.data.Statuses[id]);
+			effect = new Data.PureEffect({name: id}, this.data.Statuses[id]);
 		} else if ((this.data.Movedex.hasOwnProperty(id) && (found = this.data.Movedex[id]).effect) ||
 							 (this.data.Abilities.hasOwnProperty(id) && (found = this.data.Abilities[id]).effect) ||
 							 (this.data.Items.hasOwnProperty(id) && (found = this.data.Items[id]).effect)) {
-			effect = new Data.PureEffect({name: found.name || name}, found.effect!);
+			effect = new Data.PureEffect({name: found.name || id}, found.effect!);
 		} else if (id === 'recoil') {
-			effect = new Data.PureEffect({name: 'Recoil', effectType: 'Recoil'});
+			effect = new Data.PureEffect({id, name: 'Recoil', effectType: 'Recoil'});
 		} else if (id === 'drain') {
-			effect = new Data.PureEffect({name: 'Drain', effectType: 'Drain'});
+			effect = new Data.PureEffect({id, name: 'Drain', effectType: 'Drain'});
 		} else {
-			effect = new Data.PureEffect({name, exists: false});
+			effect = new Data.PureEffect({id, name: id, exists: false});
 		}
 
 		this.effectCache.set(id, effect);
@@ -563,7 +571,7 @@ class ModdedDex {
 		if (this.data.Formats.hasOwnProperty(id)) {
 			effect = new Data.Format({name}, this.data.Formats[id], supplementaryAttributes);
 		} else {
-			effect = new Data.Format({name, exists: false});
+			effect = new Data.Format({id, name, exists: false});
 		}
 		return effect;
 	}
@@ -590,7 +598,7 @@ class ModdedDex {
 		if (id && this.data.Items.hasOwnProperty(id)) {
 			item = new Data.Item({name}, this.data.Items[id]);
 		} else {
-			item = new Data.Item({name, exists: false});
+			item = new Data.Item({id, name, exists: false});
 		}
 
 		if (item.exists) this.itemCache.set(id, item);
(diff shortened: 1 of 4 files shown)


scheibo:
`Dex.getEffectByID` not `Dex.getEffectById` because we're talking about an `ID` type and not an `Id` type (and really `toId` should be `toID` as well by the same logic, IMO). I don't feel super strongly about this though and would be fine with `Dex.getEffectById`, I just think `byID` is more correct.

I wish there were package-private modifiers in Typescript, as `Dex.getEffect` should be the public facing API and `Dex.getEffectByID` is just sort of meant of be a package internal optimization thing.

Zarel:
I named `toId` based on precedents set by DOM APIs with stuff like `getElementById`. But PS conventions elsewhere seem to think `toID` is better. I'm fine with `ID` instead of `Id` here.

I've also considered `ID()`, by analogy to how e.g. `String()` is a casting function. TypeScript _does_ support that, and it would be consistent with our precedent that globals are allcaps. Judging by votes on `Rooms()` vs `Rooms.get()`, you might all prefer `ID.get()`, which, ew.

sim/dex.ts
```diff
@@ -486,22 +486,30 @@ class ModdedDex {
...
+			effect = new Data.PureEffect({name: found.name || id}, found.effect!);
 		} else if (id === 'recoil') {
-			effect = new Data.PureEffect({name: 'Recoil', effectType: 'Recoil'});
+			effect = new Data.PureEffect({id, name: 'Recoil', effectType: 'Recoil'});
 		} else if (id === 'drain') {
-			effect = new Data.PureEffect({name: 'Drain', effectType: 'Drain'});
+			effect = new Data.PureEffect({id, name: 'Drain', effectType: 'Drain'});
 		} else {
-			effect = new Data.PureEffect({name, exists: false});
+			effect = new Data.PureEffect({id, name: id, exists: false});
```
Zarel: This should probably throw instead of creating a placeholder.

Zarel: I don't know how I feel about how PS creates placeholders. I don't _think_ there are any situations we currently use it that aren't bugs? But in any case, if the placeholder doesn't even have a name, that should definitely throw.

Zarel: Well, `drain` etc are currently placeholders. We'd have to make sure to make actual entries in `statuses.js` for them.

scheibo: > I named `toId` based on precedents set by DOM APIs with stuff like `getElementById`. But PS conventions elsewhere seem to think `toID` is better. I'm fine with `ID` instead of `Id` here.

I'm fine with either `Id` or `ID`, I just don't like the inconsistency of having a type called `ID` and methods like `fooId`. To that end, I'd be happy to rename `toId` -> `toID` in the codebase(s)...

> I've also considered `ID()`, by analogy to how e.g. `String()` is a casting function. TypeScript _does_ support that, and it would be consistent with our precedent that globals are allcaps. 

Since when are our globals all caps??? We have a ton of globals, none of which are ALLCAPS from what I can tell?

> Judging by votes on `Rooms()` vs `Rooms.get()`, you might all prefer `ID.get()`, which, ew.

*Oh, you did see that PR then... ;)*

I prefer `toID`. If `ID` were a class, then yeah, probably `new ID()`/`ID.create`/`ID.get`. Though this might be the one single place where I'd be OK with `ID()` given its ubiquity and how its more analogous to a primitive than something like `Rooms` or `Users` (or, `TeamValidator`...)

scheibo:
> I named `toId` based on precedents set by DOM APIs with stuff like `getElementById`. But PS conventions elsewhere seem to think `toID` is better. I'm fine with `ID` instead of `Id` here.

I'm fine with either `Id` or `ID`, I just don't like the inconsistency of having a type called `ID` and methods like `fooId`. To that end, I'd be happy to rename `toId` -> `toID` in the codebase(s)...

> I've also considered `ID()`, by analogy to how e.g. `String()` is a casting function. TypeScript _does_ support that, and it would be consistent with our precedent that globals are allcaps. 

Since when are our globals all caps??? We have a ton of globals, none of which are ALLCAPS from what I can tell?

> Judging by votes on `Rooms()` vs `Rooms.get()`, you might all prefer `ID.get()`, which, ew.

*Oh, you did see that PR then... ;)*

I prefer `toID`. If `ID` were a class, then yeah, probably `new ID()`/`ID.create`/`ID.get`. Though this might be the one single place where I'd be OK with `ID()` given its ubiquity and how its more analogous to a primitive than something like `Rooms` or `Users` (or, `TeamValidator`...)

sim/dex.ts
scheibo: I have a note to myself to talk to you about the `{exists: false}` pattern - why don't we just return `undefined`? Obviously, our code doesnt want to need to check for `undefined`, but creating bulky objects (`PureEffect` has a good 20+ fields) because we don't want to check `undefined` (which the compiler can help with statically) seems really inefficient.

Zarel:
> I'm fine with either `Id` or `ID`, I just don't like the inconsistency of having a type called `ID` and methods like `fooId`. To that end, I'd be happy to rename `toId` -> `toID` in the codebase(s)...

I'm fine with that.

> Since when are our globals all caps??? We have a ton of globals, none of which are ALLCAPS from what I can tell?

Sorry, I just meant capitalized, like UpperCamelCase. Which all our globals except `toId` are.

> _Oh, you did see that PR then... ;)_

I see every PR. If I leave one open, it means I'm undecided. I don't ever intentionally pocket veto anything.

With that one specifically... blah if it's me against literally everyone else, I guess I have to capitulate, just give me some more time.

> I prefer `toID`. If `ID` were a class, then yeah, probably `new ID()`/`ID.create`/`ID.get`. Though this might be the one single place where I'd be OK with `ID()` given its ubiquity and how its more analogous to a primitive than something like `Rooms` or `Users` (or, `TeamValidator`...)

I've been leaning `ID()` for a while now. It's definitely something we can push on the server. On the client, I'm a bit more nervous about making `ID` a global, because it's a generic name and can conflict with poorly-written extensions and malware (client has had crashes in the past due to malware taking the variable name `app` in the global namespace).

sim/dex.ts
Zarel: The biggest reason we don't just return `undefined` is because https://github.com/tc39/proposal-optional-chaining is _still_ Stage 1 after approximately eight years of literally everyone asking for it.

Zarel: Honestly, we even considered making placeholder `Pokemon` objects, for similar reasons in #697.

The only reason we didn't was because a researcher pointed out that pokemon slots can never actually be completely empty.

scheibo:
I think changing `toId(...)` -> `toID(...)` or `ID(...)` is out of scope for this change? I think we'd also want to get some feedback about it from others before doing it? Similarly, what we do with the placeholder objects also seems like a candidate for a different PR? I'm happy to follow up on both of these.

Zarel:
Yeah, sure, that's out of scope of this PR.

scheibo:
I've filed bugs to follow up on the issues you've noted so far. Do you have more thoughts, are should we merge this?

sim/dex.ts
Zarel: This doesn't seem resolved – you've neither added a throw nor said that you think it shouldn't.

Zarel: Ohhh, I see, you consider that a separate change. I originally thought this was a regression, but I guess in practice it isn't.

Zarel: I still consider this a distinct change, though! I'm recommending throwing, instead of returning `null`, because there are significantly fewer `getEffectByID` calls, and it's significantly easier to audit every call and note that they'd all be bugs if not found.

Zarel: I guess that can be a separate change, though

PR #5476, opened 2019-05-04
[Gen 5] 1v1: Ban Perish Song

AustinXII:
[15:18:44] dom: can some1 ban perish song from gen5 1v1 i lead that tier thanks


AustinXII:
@TheImmortal

PR #5481, opened 2019-05-06
Past Custom Game "canUseRandomTeam", Level 9999/24 Pokemon

Tmi489:
Using future stuff in past game custom game still doesn't work so copied the exact same ruleset (gen6's) for all CGs other than Gen7

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -1854,7 +1854,11 @@ let Formats = [
 
 		mod: 'gen3',
 		searchShow: false,
+		canUseRandomTeam: true,
 		debug: true,
+		maxLevel: 9999,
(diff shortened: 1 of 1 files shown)


Zarel:
None of those control using future stuff in past gens.

Zarel:
I can take a look at what's going on here, but I'm swamped and probably won't get to this for a long time...

scheibo:
Probably an `exists` check?

Zarel:
Future Pokémon in past gens should still have `exists = true`

Tmi489:
i still want level 9999 and 24 mons though

Zarel:
That's fair.

PR #5482, opened 2019-05-07
Typescript server/verifier.js

scheibo:
Still going to futz around to see if I can eliminate the `no-var-requires` ignores, but this should work.


build
```diff
@@ -57,9 +57,13 @@ try {
 
 sucrase('./sim', './.sim-dist');
 sucrase('./lib', './.lib-dist');
+sucrase('./server', './.server-dist');
 // NOTE: replace is asynchronous - add additional replacements for the same path in one call instead of making multiple calls.
 replace(path.join(__dirname, '.sim-dist'), [
-	{regex: new RegExp(`(require\\\(.*?)(lib)(.*?\\\))`), replace: `$1.lib-dist$3`},
+	{regex: new RegExp(`(require\\\(.*?)(lib)(.*?\\\))`, 'g'), replace: `$1.lib-dist$3`},
```
scheibo: Not necessarily needed for now, but seems like a foot gun if any imports are added.

PR #5484, opened 2019-05-07
Chat-Plugins: Fix regression in weakness command

jmclemo6:
This fixes the weakness regression that was discussed in the dev room on PS.


Zarel:
Thanks!

PR #5485, opened 2019-05-07
TypeScript server/loginserver.js

TheJetOU:


server/loginserver.ts
```diff
@@ -136,8 +127,7 @@ class LoginServerInstance {
 
 		if (!requests.length) return;
 
-		/** @type {((val: LoginServerResponse) => void)[]} */
-		let resolvers = [];
+		let resolvers = <((val: LoginServerResponse) => void)[]> [];
```
scheibo:
```suggestion
		let resolvers: ((val: LoginServerResponse) => void)[] = [];
```

server/loginserver.ts
```diff
@@ -149,18 +139,16 @@ class LoginServerInstance {
...
 		requestOptions.headers = {
 			'Content-Type': 'application/x-www-form-urlencoded',
 			'Content-Length': postData.length,
 		};
 
-		/** @type {any} */
-		let response = null;
-		// @ts-ignore
-		let req = http.request(requestOptions, res => {
+		let response = <null | string | object> null;
```
scheibo:
```suggestion
		let response: string | object | null = null;
```

But `AnyObject` might be better

server/loginserver.ts
```diff
@@ -12,24 +12,22 @@
 const LOGIN_SERVER_TIMEOUT = 30000;
 const LOGIN_SERVER_BATCH_TIME = 1000;
 
+// @ts-ignore
```
scheibo: Please add a comment about what is being `@ts-ignore` / why its necessary. Are you sure you want a `@ts-ignore` here and not `// tslint:disable-next-line no-var-requires`?

scheibo: Or is this to avoid some having to `ts-ignore` on usage instead? Ignoring at such a high level here seems dangerous.

TheJetOU: Yeah, that's probably better.

server/loginserver.ts
```diff
@@ -223,10 +223,14 @@ const LoginServer = Object.assign(new LoginServerInstance(), {
 });
 
 FS('./config/custom.css').onModify(() => {
-	LoginServer.request('invalidatecss');
+	LoginServer.request('invalidatecss').catch(err => {
```
scheibo: catching and rethrowing does not useful other than trick the stupid linter. You want to `return` here and either `console.err` or ``// disable`` below

scheibo: To be clear, rethrowing does nothing. The whole point of the linter warning is to tell you that if `LoginServer.request('invalidatecss')` fails, nothing will noticed because the error is swallowed up. I

server/loginserver.ts
```diff
@@ -38,69 +36,67 @@ function parseJSON(/** @type {string} */ json) {
...
+		const urlObject = url.parse(`${this.uri}action.php?act=action&serverid='${Config.serverid}&servertoken=` +
+		`'${encodeURIComponent(Config.servertoken)}&nocache=${new Date().getTime() + dataString}`);
 		return new Promise((resolve, reject) => {
-			// @ts-ignore TypeScript bug: http.get signature
-			let req = http.get(urlObject, res => {
-				Streams.readAll(res).then(buffer => {
-					let data = parseJSON(buffer).json || null;
-					resolve([data, res.statusCode || 0, null]);
+
+			const req = http.get(urlObject, (res: any) => {
```
TheJetOU: I'm not sure what type I should give `res` to.

scheibo: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/https.d.ts and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts should help answer that (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L177 appears to be the answer).

However, `Streams.readAll` just needs `NodeJS.ReadableStream`, which `stream.Readable` is (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts#L20), so if you want to avoid `any`, `NodeJS.ReadableStream` seems correct

server/loginserver.ts
```diff
@@ -131,40 +127,40 @@ class LoginServerInstance {
...
+			},
 		};
 
-		/** @type {any} */
-		let response = null;
-		// @ts-ignore
-		let req = http.request(requestOptions, res => {
+		let response: AnyObject | null =  null;
+
+		const req = http.request(requestOptions, (res: any) => {
```
scheibo: See above regarding typing and propagating the errors

server/loginserver.ts
```diff
@@ -216,22 +215,29 @@ class LoginServerInstance {
...
+const LoginServer = Object.assign(new LoginServerInstance(), {
 	TimeoutError,
 
 	ladderupdateServer: new LoginServerInstance(),
 	prepreplayServer: new LoginServerInstance(),
 });
 
 FS('./config/custom.css').onModify(() => {
-	LoginServer.request('invalidatecss');
+	LoginServer.request('invalidatecss').catch((err: string) => {
```
scheibo: id move both these calls into a method where you then then handle the error in one place.

server/loginserver.ts
TheJetOU: Hm, that gives an error because `NodeJS.ReadableStream` does not have `statusCode`, I guess I could write a one-liner type for those?

scheibo: Maybe like:

```typescript
interface IncomingMessage extends NodeJS.ReadableStream {
  statusCode: number;
}
```

server/loginserver.ts
scheibo: Though I'm also guessing the 'handling' will just be to `// tslint:disable`` and ignore the possibility for errors - I dont know where `console.err` logs go on the server or whether anyone looks at them, and I dont know how much we care if the 'invalidatecss' require fails

server/loginserver.ts
TheJetOU: Oh right yeah, I forgot about interfaces, lol. Thank you.

server/loginserver.ts
TheJetOU: @Zarel

server/loginserver.ts
```diff
@@ -37,6 +37,7 @@ function parseJSON(json: string) {
 }
 
 type LoginServerResponse = [AnyObject | null, number, Error | null];
+type response = NodeJS.ReadableStream | {statusCode: string};
```
scheibo: We typically use Capitalized types. Also, this should be `&`, not `|`.

TheJetOU: That code doesn't exist anymore. It's from an old commit.

TheJetOU: It's now an interface

server/loginserver.ts
Zarel: We very much don't want to `console.err`. This should definitely just be `tslint:disable`d. Uncaught promises will go directly to Crashlogger which is exactly where we want them.

TheJetOU: did

server/loginserver.ts
```diff
@@ -216,22 +215,25 @@ class LoginServerInstance {
 		this.requestTimerPoke();
 	}
 	getLog() {
-		return this.requestLog + (this.lastRequest ? ' (' + Chat.toDurationString(Date.now() - this.lastRequest) + ' since last request)' : '');
+		return `${this.requestLog}` +
+		`${this.lastRequest ? ` (${Chat.toDurationString(Date.now() - this.lastRequest)} since last request)` : ''}`;
```
scheibo: It seems like the ternary could be moved out here.

server/loginserver.ts
```diff
@@ -38,69 +36,69 @@ function parseJSON(/** @type {string} */ json) {
...
 		this.openRequests++;
 		let dataString = '';
 		if (data) {
-			for (let i in data) {
+			for (const i in data) {
 				dataString += '&' + i + '=' + encodeURIComponent('' + data[i]);
 			}
 		}
-		const urlObject = url.parse(this.uri + 'action.php?act=' + action + '&serverid=' + Config.serverid + '&servertoken=' + encodeURIComponent(Config.servertoken) + '&nocache=' + new Date().getTime() + dataString);
+		const urlObject = url.parse(`${this.uri}action.php?act=action&serverid='${Config.serverid}&servertoken=` +
```
scheibo: I think this was more readable at `HEAD`?

```
		let postData = 'serverid=' + Config.serverid +
			'&servertoken=' + encodeURIComponent(Config.servertoken) +
			'&nocache=' + new Date().getTime() +
			'&json=' + encodeURIComponent(JSON.stringify(dataList)) + '\n';
```

server/loginserver.ts
```diff
@@ -38,69 +36,74 @@ function parseJSON(/** @type {string} */ json) {
...
-	/**
-	 * @param {string} action
-	 * @param {AnyObject?} data
-	 * @return {Promise<LoginServerResponse>}
-	 */
-	instantRequest(action, data = null) {
+	instantRequest(action: string, data: AnyObject | null = null): Promise<LoginServerResponse> {
 		if (this.openRequests > 5) {
-			return Promise.resolve(/** @type {LoginServerResponse} */ (
+			return Promise.resolve((
```
scheibo: `((`? Seems you can lose some parens?

server/loginserver.ts
```diff
@@ -38,69 +36,74 @@ function parseJSON(/** @type {string} */ json) {
...
 		if (data) {
-			for (let i in data) {
+			for (const i in data) {
 				dataString += '&' + i + '=' + encodeURIComponent('' + data[i]);
 			}
 		}
-		const urlObject = url.parse(this.uri + 'action.php?act=' + action + '&serverid=' + Config.serverid + '&servertoken=' + encodeURIComponent(Config.servertoken) + '&nocache=' + new Date().getTime() + dataString);
+
+		const urlString = this.uri + 'action.php?' +
+		'?act=action&serverid=' + Config.serverid +
```
scheibo: I think this was indented before?

server/loginserver.ts
```diff
@@ -38,69 +36,74 @@ function parseJSON(/** @type {string} */ json) {
...
-	 * @param {string} action
-	 * @param {AnyObject?} data
-	 * @return {Promise<LoginServerResponse>}
-	 */
-	request(action, data = null) {
+
+	request(action: string, data: AnyObject | null = null): Promise<LoginServerResponse> {
 		if (this.disabled) {
-			return Promise.resolve(/** @type {LoginServerResponse} */ (
+			return Promise.resolve((
```
scheibo: ditto

server/loginserver.ts
```diff
@@ -216,22 +220,26 @@ class LoginServerInstance {
 		this.requestTimerPoke();
 	}
 	getLog() {
-		return this.requestLog + (this.lastRequest ? ' (' + Chat.toDurationString(Date.now() - this.lastRequest) + ' since last request)' : '');
+		const lastRequestString = this.lastRequest ?
```
scheibo:
```typescript
if (!this.lastRequest) return this.requestLog;
return `${this.requestLog}${Chat.toDurationString} ...
```

server/loginserver.ts
TheJetOU: url is reserved for the module, so used `urlObject`.

scheibo: Ah, right. didn't catch that. Maybe `actionUrl`, because its for the action.php endpoint?

server/loginserver.ts
TheJetOU: Huh, that's weird, why did the original have two parens? That's definitely not the syntax of Promise.resolve

server/loginserver.ts
TheJetOU: Don't have any strong opinions on it, so sure.

server/loginserver.ts
```diff
@@ -12,24 +12,22 @@
 const LOGIN_SERVER_TIMEOUT = 30000;
 const LOGIN_SERVER_BATCH_TIME = 1000;
 
+// tslint:disable-next-line no-var-requires
 const http = Config.loginserver.startsWith('http:') ? require("http") : require("https");
```
scheibo: `"http"` and `"https"` should be single quotes

(which wasn't introduced by this change, but lets fix it while we're here :P)

server/loginserver.ts
```diff
@@ -38,71 +36,75 @@ function parseJSON(/** @type {string} */ json) {
...
-			for (let i in data) {
+			for (const i in data) {
 				dataString += '&' + i + '=' + encodeURIComponent('' + data[i]);
 			}
 		}
-		const urlObject = url.parse(this.uri + 'action.php?act=' + action + '&serverid=' + Config.serverid + '&servertoken=' + encodeURIComponent(Config.servertoken) + '&nocache=' + new Date().getTime() + dataString);
+
+		const actionUrl = url.parse(this.uri + 'action.php' +
+			'?act=action&serverid=' + Config.serverid +
+			'&servertoken=' + `'${encodeURIComponent(Config.servertoken)}'` +
```
scheibo: `encodeURIComponent` does not need to be interpolated into a string

server/loginserver.ts
TheJetOU: Sure, lol.

server/loginserver.ts
```diff
@@ -131,40 +133,41 @@ class LoginServerInstance {
...
 			'&json=' + encodeURIComponent(JSON.stringify(dataList)) + '\n';
-		/** @type {any} */
-		let requestOptions = url.parse(this.uri + 'action.php');
-		requestOptions.method = 'post';
-		requestOptions.headers = {
-			'Content-Type': 'application/x-www-form-urlencoded',
-			'Content-Length': postData.length,
+
+		const requestOptions = {
+			url: url.parse(this.uri + 'action.php'),
```
scheibo:
```suggestion
			url: url.parse(`${this.uri}action.php`),
```

server/loginserver.ts
```diff
@@ -216,22 +219,25 @@ class LoginServerInstance {
 		this.requestTimerPoke();
 	}
 	getLog() {
-		return this.requestLog + (this.lastRequest ? ' (' + Chat.toDurationString(Date.now() - this.lastRequest) + ' since last request)' : '');
+		if (!this.lastRequest) return this.requestLog;
+		return `${this.requestLog}  (${Chat.toDurationString(Date.now() - this.lastRequest)} since last request)`;
```
scheibo: I
```suggestion
		return `${this.requestLog} (${Chat.toDurationString(Date.now() - this.lastRequest)} since last request)`;
```

server/loginserver.ts
TheJetOU: I accidentally implemented this but, there needs to be a `'` and I think that looks better than using `"` and then escaping `'`.

TheJetOU: Oh wait, that's not true.

PR #5486, opened 2019-05-07
Only add the '-status slp' message in statuses

### LINKED ISSUE #5023: Two sleep messages from Rest
When a Pokemon uses Rest, the server sends the following messages:

|move|p2a: Giratina|Rest|p2a: Giratina
|-status|p2a: Giratina|slp
|-heal|p2a: Giratina|100/100 slp|[silent]
|-status|p2a: Giratina|slp|[from] move: Rest

It seems that the first message is redundant.

scheibo:
Fixes #5023


Marty-D:
👍

PR #5489, opened 2019-05-07
Protocol: replace |customgroups| with an all purpose |groups| message

Kaiepi:
This sends a list of the server's configured groups to the user on
connect to propagate the client's groups state.

- |customgroups| was dependent on a years-old version of group
configuration, making it useless for newer servers.
- Decoupling the state of groups between clients and servers like this
is more friendly to bots. Hardcoding the groups list is no longer
necessary, so if the bot dynamically sets the group list on connect, it
will be able to connect to any server regardless of custom groups and
still function.
- The official client no longer needs to be updated whenever main's
groups list changes.

https://github.com/Zarel/Pokemon-Showdown-Client/pull/1272 should be merged before this

### DIFF
--- a/server/rooms.js
+++ b/server/rooms.js
@@ -591,37 +591,62 @@ class GlobalRoom extends BasicRoom {
 		}
 		return this.formatList;
 	}
-	get configRankList() {
-		if (Config.nocustomgrouplist) return '';
-
-		// putting the resultant object in Config would enable this to be run again should config.js be reloaded.
-		if (Config.rankList) {
-			return Config.rankList;
+	/**
+	 * @param {AnyObject} group
+	 */
+	getGroupWithType(group) {
+		let {symbol, id, name} = group;
+		let type;
+		if (group.editroom || group.makeroom || group.root) {
+			type = 'leadership';
+		} else if (!group.declare && (group.mute || group.ban)) {
+			type = 'staff';
+		} else if (group.punishgroup) {
+			type = 'punishment';
+		} else if (!group.broadcast) {
+			type = 'default';
+		} else {
+			type = 'normal';
 		}
-		let rankList = [];
+		return {symbol, id, name, type};
+	}
+	get configGroupList() {
+		// Putting the resultant object in Config would enable this to be run again should config.js be reloaded.
+		if (Config.rankList) return Config.rankList;
 
-		for (let rank in Config.groups) {
-			if (!Config.groups[rank] || !rank) continue;
+		const typeOrder = ['punishment', 'default', 'normal', 'staff', 'leadership'];
+		let rankList;
 
-			let tarGroup = Config.groups[rank];
-			let groupType = tarGroup.addhtml || (!tarGroup.mute && !tarGroup.root) ? 'normal' : (tarGroup.root || tarGroup.declare) ? 'leadership' : 'staff';
+		if (Config.grouplist) {
+			// a and b are implicitly any
+			// @ts-ignore
+			rankList = Config.grouplist.map(this.getGroupWithType).sort((a, b) => {
+				let ret = typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type);
+				if (ret) return ret;
 
-			rankList.push({symbol: rank, name: (Config.groups[rank].name || null), type: groupType}); // send the first character in the rank, incase they put a string several characters long
+				// g is implicitly any
+				// @ts-ignore
+				ret = Config.grouplist.findIndex(g => g.id === a.id) - Config.grouplist.findIndex(g => g.id === b.id);
+				return ret;
+			});
+		} else {
+			const groups = Object.values(Config.groups);
+			const punishGroups = Object.values(Config.punishgroups);
+			rankList = [...groups, ...punishGroups]
+				.map(this.getGroupWithType)
+				.sort((a, b) => {
+					let ret = typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type);
+					if (ret) return ret;
+
+					const groupList = (b.type === 'punishment') ? punishGroups : groups;
+					ret = groupList.findIndex(g => g.id === a.id) - groupList.findIndex(g => g.id === b.id);
+					return ret;
+				});
 		}
 
-		const typeOrder = ['punishment', 'normal', 'staff', 'leadership'];
-
-		rankList = rankList.sort((a, b) => typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type));
-
-		// add the punishment types at the very end.
-		for (let rank in Config.punishgroups) {
-			rankList.push({symbol: Config.punishgroups[rank].symbol, name: Config.punishgroups[rank].name, type: 'punishment'});
-		}
-
-		Config.rankList = '|customgroups|' + JSON.stringify(rankList) + '\n';
+		Config.rankList = `|groups|${JSON.stringify(rankList)}`;
 		return Config.rankList;
 	}
-
 	/**
 	 * @param {string} filter "formatfilter, elofilter"
 	 */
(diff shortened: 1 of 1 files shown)


server/rooms.js
```diff
@@ -591,37 +591,55 @@ class GlobalRoom extends BasicRoom {
...
-			if (!Config.groups[rank] || !rank) continue;
-
-			let tarGroup = Config.groups[rank];
-			let groupType = tarGroup.addhtml || (!tarGroup.mute && !tarGroup.root) ? 'normal' : (tarGroup.root || tarGroup.declare) ? 'leadership' : 'staff';
-
-			rankList.push({symbol: rank, name: (Config.groups[rank].name || null), type: groupType}); // send the first character in the rank, incase they put a string several characters long
+	/**
+	 * @param {AnyObject} group
+	 */
+	configGroupListMapper(group) {
```
scheibo: This seems to be hiding a `getGroupType` method inside of it

server/rooms.js
```diff
@@ -591,37 +591,55 @@ class GlobalRoom extends BasicRoom {
...
 
-		rankList = rankList.sort((a, b) => typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type));
-
-		// add the punishment types at the very end.
-		for (let rank in Config.punishgroups) {
-			rankList.push({symbol: Config.punishgroups[rank].symbol, name: Config.punishgroups[rank].name, type: 'punishment'});
+		if (Config.grouplist) {
+			// @ts-ignore
+			rankList = Config.grouplist.map(this.configGroupListMapper).sort((a, b) => (
+				typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type) ||
```
scheibo: This is really dense. `map` -> `sort` with a relatively complex comparator. Perhaps could benefit being broken up to make it more digestable.

server/rooms.js
```diff
@@ -591,37 +591,55 @@ class GlobalRoom extends BasicRoom {
...
 		const typeOrder = ['punishment', 'normal', 'staff', 'leadership'];
+		let rankList;
 
-		rankList = rankList.sort((a, b) => typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type));
-
-		// add the punishment types at the very end.
-		for (let rank in Config.punishgroups) {
-			rankList.push({symbol: Config.punishgroups[rank].symbol, name: Config.punishgroups[rank].name, type: 'punishment'});
+		if (Config.grouplist) {
+			// @ts-ignore
```
scheibo: Its helpful to include a comment explaining why you are using `// @ts-ignore` (or in any language when you suppress warnings) such that other developers can more easily reason about whether the usage is justified.

server/rooms.js
```diff
@@ -591,37 +591,55 @@ class GlobalRoom extends BasicRoom {
...
+			rankList = Config.grouplist.map(this.configGroupListMapper).sort((a, b) => (
+				typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type) ||
+					// @ts-ignore
+					Config.grouplist.findIndex(g => g.id === a.id) - Config.grouplist.findIndex(g => g.id === b.id)
+			));
+		} else {
+			const groups = Object.values(Config.groups);
+			const punishGroups = Object.values(Config.punishgroups);
+			rankList = [
+				...groups.map(this.configGroupListMapper),
```
scheibo: Seems like you could do the map on `rankList` instead of on both `groups` and `punishGroups` individually?

server/rooms.js
```diff
@@ -591,37 +591,55 @@ class GlobalRoom extends BasicRoom {
...
-
-		for (let rank in Config.groups) {
-			if (!Config.groups[rank] || !rank) continue;
-
-			let tarGroup = Config.groups[rank];
-			let groupType = tarGroup.addhtml || (!tarGroup.mute && !tarGroup.root) ? 'normal' : (tarGroup.root || tarGroup.declare) ? 'leadership' : 'staff';
-
-			rankList.push({symbol: rank, name: (Config.groups[rank].name || null), type: groupType}); // send the first character in the rank, incase they put a string several characters long
+	/**
+	 * @param {AnyObject} group
```
scheibo: Is this really `AnyObject`? You seem to depend on `symbol`, `id` and `name` keys  - that should be reflected here

server/rooms.js
Kaiepi: Already discussed this in the dev room, the problem was more a matter of naming

Kaiepi:
Updated

server/rooms.js
Kaiepi: `group` here is the group as it's defined in config, not the group object with `symbol`, `id`, `name`, and `type` keys

server/rooms.js
Kaiepi: Fixed now

server/rooms.js
Kaiepi: Done

server/rooms.js
Kaiepi: Fixed now

server/rooms.js
Kaiepi: Wait, I misread. How would I type it so it includes those three keys but ignores any others the group may have?

server/rooms.js
```diff
@@ -591,37 +591,62 @@ class GlobalRoom extends BasicRoom {
...
-			let groupType = tarGroup.addhtml || (!tarGroup.mute && !tarGroup.root) ? 'normal' : (tarGroup.root || tarGroup.declare) ? 'leadership' : 'staff';
+		const typeOrder = ['punishment', 'default', 'normal', 'staff', 'leadership'];
+		let rankList;
 
-			rankList.push({symbol: rank, name: (Config.groups[rank].name || null), type: groupType}); // send the first character in the rank, incase they put a string several characters long
-		}
-
-		const typeOrder = ['punishment', 'normal', 'staff', 'leadership'];
+		if (Config.grouplist) {
+			// a and b are implicitly any
```
scheibo: And this is why I ask for people to add comments for these! :) The 'implicitly any' error doesn't mean you should be ignoring, it means you need to specify the types explicitly (it can't infer them so its implicitly giving them the `any` type)

server/rooms.js
scheibo: Thats how interfaces work already with structural typing? 

```ts
interface Foo {
  foo: number;
}

function inc(foo: Foo) {
  return foo.foo + 1;
}

const bar = {foo: 5, bar: 10}; // bar implicitly implements Foo
console.log(inc(bar)); // => 6
```

scheibo: Though actually, looking at it again, you care about more than just `symbol`, `id`, and `name` (you want `editroom`, `makeroom`, `root` etc).

server/rooms.js
scheibo: This is where the type is just:

```ts
interface TypedGroupInfo {
  type: 'punishment' | 'default' | 'normal' | 'staff' | 'leadership';
  symbol: '~' | '#' | '&' | '@' | '%' | '*' | '+' | ...; // or just string
  id: ID;
  name: string;
}

File: server/rooms.js [hunk 7]
```diff
@@ -591,37 +591,62 @@ class GlobalRoom extends BasicRoom {
...
+			rankList = [...groups, ...punishGroups]
+				.map(this.getGroupWithType)
+				.sort((a, b) => {
+					let ret = typeOrder.indexOf(b.type) - typeOrder.indexOf(a.type);
+					if (ret) return ret;
 
-		// add the punishment types at the very end.
-		for (let rank in Config.punishgroups) {
-			rankList.push({symbol: Config.punishgroups[rank].symbol, name: Config.punishgroups[rank].name, type: 'punishment'});
+					const groupList = (b.type === 'punishment') ? punishGroups : groups;
```
scheibo: I believe you can actually combine both branches and sort functions into one simply by using this comparator function and setting `groups = Config.groupList` and `punishGroups = []` for the first branch.

AnnikaCodes:

PR #5490, opened 2019-05-08
TypeScript ladders-remote and ladders-local

TheJetOU:


server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
-		/** @type {Promise<LadderRow[]>?} */
 		this.ladderPromise = null;
+		this.saving = false;
 	}
 
-	/**
-	 * @return {Promise<LadderRow[]>}
-	 */
-	getLadder() {
+	getLadder(): Promise<LadderRow[]> {
```
Zarel: PS by convention doesn't specify return type except when it can't be inferred. I'd prefer to keep that consistent.

Zarel: (You specify all return types in this file, which is mostly unnecessary.)

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
 
 // ladderCaches = {formatid: ladder OR Promise(ladder)}
 // Use Ladders(formatid).ladder to guarantee a Promise(ladder).
 // ladder is basically a 2D array representing the corresponding ladder.tsv
 //   with userid in front
-/** @typedef {[string, number, string, number, number, number, string]} LadderRow [userid, elo, username, w, l, t, lastUpdate] */
-/** @type {Map<string, LadderRow[] | Promise<LadderRow[]>>} formatid: ladder */
-let ladderCaches = new Map();
+type LadderRow = /** [userid, elo, username, w, l, t, lastUpdate */
+[string, number, string, number, number, number, string];
```
Zarel:
```suggestion
type LadderRow = [string, number, string, number, number, number, string];
```

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
+import {FS} from '../lib/fs';
 
 // ladderCaches = {formatid: ladder OR Promise(ladder)}
 // Use Ladders(formatid).ladder to guarantee a Promise(ladder).
 // ladder is basically a 2D array representing the corresponding ladder.tsv
 //   with userid in front
-/** @typedef {[string, number, string, number, number, number, string]} LadderRow [userid, elo, username, w, l, t, lastUpdate] */
-/** @type {Map<string, LadderRow[] | Promise<LadderRow[]>>} formatid: ladder */
-let ladderCaches = new Map();
+type LadderRow = /** [userid, elo, username, w, l, t, lastUpdate */
```
Zarel:
```suggestion
/** [userid, elo, username, w, l, t, lastUpdate */
```

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
+[string, number, string, number, number, number, string];
+type Cache = Map<string, LadderRow[] | Promise<LadderRow[]>>;
```
Zarel:
```suggestion
```

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
+
+const ladderCaches: Cache = new Map();
```
Zarel:
```suggestion
/** formatid: ladder */
const ladderCaches = new Map<string, LadderRow[] | Promise<LadderRow[]>>();
```

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
+				const ladder = await cachedLadder;
 				return (this.ladder = ladder);
 			}
 			// @ts-ignore
 			return (this.ladder = cachedLadder);
 		}
 		try {
 			const data = await FS('config/ladders/' + this.formatid + '.tsv').readIfExists();
-			let ladder = /** @type {LadderRow[]} */ ([]);
+			const ladder: LadderRow[] = ([]);
```
Zarel:
```suggestion
			const ladder: LadderRow[] = [];
```

server/ladders-local.ts
```diff
@@ -292,11 +288,13 @@ class LadderStore {
 
 			let reasons = '' + (Math.round(p1newElo) - Math.round(p1elo)) + ' for ' + (p1score > 0.9 ? 'winning' : (p1score < 0.1 ? 'losing' : 'tying'));
 			if (reasons.charAt(0) !== '-') reasons = '+' + reasons;
-			room.addRaw(Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; <strong>${Math.round(p1newElo)}</strong><br />(${reasons})`);
+			room.addRaw(Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; ` +
+			`<strong>${Math.round(p1newElo)}</strong><br />(${reasons})`);
```
Zarel: This isn't being affected by `Chat.html` even though it should be!

server/ladders-local.ts
TheJetOU: We use the type `Cache` twice, isn't it better to just make it a type since it's a bit verbose?

server/ladders-local.ts
Zarel:
```suggestion
```

server/ladders-local.ts
```diff
@@ -292,11 +288,13 @@ class LadderStore {
 
 			let reasons = '' + (Math.round(p1newElo) - Math.round(p1elo)) + ' for ' + (p1score > 0.9 ? 'winning' : (p1score < 0.1 ? 'losing' : 'tying'));
 			if (reasons.charAt(0) !== '-') reasons = '+' + reasons;
-			room.addRaw(Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; <strong>${Math.round(p1newElo)}</strong><br />(${reasons})`);
+			room.addRaw(Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; ` +
```
Zarel:
```suggestion
			room.addRaw(
				Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; <strong>${Math.round(p1newElo)}</strong><br />(${reasons})`
			);
```

server/ladders-local.ts
```diff
@@ -292,11 +288,13 @@ class LadderStore {
...
+			`<strong>${Math.round(p1newElo)}</strong><br />(${reasons})`);
 
 			reasons = '' + (Math.round(p2newElo) - Math.round(p2elo)) + ' for ' + (p2score > 0.9 ? 'winning' : (p2score < 0.1 ? 'losing' : 'tying'));
 			if (reasons.charAt(0) !== '-') reasons = '+' + reasons;
-			room.addRaw(Chat.html`${p2name}'s rating: ${Math.round(p2elo)} &rarr; <strong>${Math.round(p2newElo)}</strong><br />(${reasons})`);
+			room.addRaw(Chat.html`${p2name}'s rating: ${Math.round(p2elo)} &rarr; ` +
```
Zarel:
```suggestion
			room.addRaw(
				Chat.html`${p2name}'s rating: ${Math.round(p2elo)} &rarr; <strong>${Math.round(p2newElo)}</strong><br />(${reasons})`
			);
```

Same issue. Your second line isn't being sanitized.

server/ladders-local.ts
```diff
@@ -292,11 +288,13 @@ class LadderStore {
...
+			`<strong>${Math.round(p2newElo)}</strong><br />(${reasons})`);
```
Zarel:
```suggestion
```

server/ladders-local.ts
Zarel: Wait, we do? I didn't know. That's fine, then.

Zarel: Just don't call it `Cache`. Namespacing is important. Call it `LadderCache`.

Zarel: Oh, wait, we don't actually use it twice. We only do because you wrote the code wrong.

TheJetOU: Sure.

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
 class LadderStore {
-	/**
-	 * @param {string} formatid
-	 */
-	constructor(formatid) {
+	formatid: string;
+	ladder: LadderRow[] | null;
+	ladderPromise: Promise<LadderRow[]> | null;
+	saving: boolean;
+	static formatsListPrefix: string;
```
Zarel:
```suggestion
	static formatsListPrefix = '|,LL';
```

server/ladders-local.ts
```diff
@@ -15,59 +15,57 @@
...
+	static ladderCaches: Cache;
```
Zarel:
```suggestion
	static ladderCaches = ladderCaches;
```

Zarel:
Yeah, `formatsListPrefix` is an LGTM bug. Good thing the correct code doesn't trigger the LGTM bug. :p

server/ladders-local.ts
TheJetOU: I didn't see the last comment lol, @Zarel, with your review shouldn't we still give it a type though?

Zarel: No? Like I mentioned in other places, you shouldn't be using types where they can be inferred.

server/ladders-local.ts
TheJetOU: That's >120 chars though :|

server/ladders-local.ts
TheJetOU: Hopefully https://github.com/Zarel/Pokemon-Showdown/pull/5490/commits/365868d02ab1c6addd50bf147362a63f21a6f7a5 fixed what you mean?

server/ladders-local.ts
```diff
@@ -292,11 +288,17 @@ class LadderStore {
 
 			let reasons = '' + (Math.round(p1newElo) - Math.round(p1elo)) + ' for ' + (p1score > 0.9 ? 'winning' : (p1score < 0.1 ? 'losing' : 'tying'));
 			if (reasons.charAt(0) !== '-') reasons = '+' + reasons;
-			room.addRaw(Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; <strong>${Math.round(p1newElo)}</strong><br />(${reasons})`);
+			room.addRaw(
+				Chat.html`${p1name}'s rating: ${Math.round(p1elo)} &rarr; ` +
+				Chat.html`<strong>${Math.round(p1newElo)}</strong><br />(${reasons})`
```
Zarel: What's your problem with doing it all on one line? It fits in 120 cols (I'm pretty sure) and I thought we carved out an exception in the line length limit for string literals anyway.

TheJetOU: It doesn't fit in 120 cols. https://i.imgur.com/M94ROln.png

TheJetOU: I guess I can disable it for those two lines?

server/ladders-local.ts
Zarel: It's not though... tslint counts tabs as one char each. (Also as mentioned elsewhere, we carve out a tslint exception for strings so they are always allowed on one line.)

server/ladders-local.ts
Zarel: Yep, that's what I wanted; thanks!

server/ladders-local.ts
Zarel: Oh, no, that's a flaw in my regex, then. You can leave this as-is.

TheJetOU: Reverted.

TheJetOU:
I'm confused. I'm not writing to `formatsListPrefix`, I'm just giving it a type, not a value.

server/ladders-local.ts
Zarel: Hey, wait, `Sanitize html in one line` passed linting! Your VS Code seems to not be picking up on my filter regex?

server/ladders-local.ts
```diff
@@ -15,59 +15,58 @@
...
 
-	/**
-	 * @return {Promise<LadderRow[]>}
-	 */
 	getLadder() {
 		if (!this.ladderPromise) this.ladderPromise = this.load();
 		return this.ladderPromise;
 	}
-	/**
-	 * Internal function, returns a Promise for a ladder
```
Zarel: Hey, why'd you delete this? The doc comment part is still important.

server/ladders-local.ts
```diff
@@ -90,15 +89,17 @@ class LadderStore {
 		if (this.saving) return;
 		this.saving = true;
 		const ladder = await this.getLadder();
+		if (!ladder) return;
```
Zarel: What's with all of this stuff? `getLadder` should never return a falsy value.

Zarel:
At some point I'm going to want classes to be ordered:

- instance properties
- instance methods
- static properties
- static methods

But that can wait until later.

server/ladders-local.ts
TheJetOU: Oh, that's,,, weird. I'll revert my reverted commit I guess.

server/ladders-local.ts
TheJetOU: Oh, that's my bad.

server/ladders-local.ts
TheJetOU: Oh, that _was_ because when I was adding return values for every function, I think I added `null` for `getLadder()`, so TypeScript was complaining that it could be null, but now that I don't add return values unless they're _not_ deducible, those checks are useless.

Zarel:
Looks good to me now. @scheibo?

server/ladders-local.ts
scheibo: @Zarel this is still generating linter warnings.

PR #5492, opened 2019-05-09
Update RoomGame API

Zarel:
So in our previous iteration, `RoomGamePlayer` was too closely linked to `User`, making the idea of substituting players unnecessarily difficult.

This new API supports `RoomGamePlayer`s that aren't connected to any user. Such `RoomGamePlayer`s will have `.userid === ''`, and won't show up in `game.players`. A new array, `game.playerList`, will hold all players, with and without users. `game.players` has been renamed `game.playerTable` and will only be a table of players with users.

(For diff readability, `game.players` has not been renamed in this PR, but will be renamed at merge time.)

Games that don't need to support players without users won't need to any major updates, except for the `.players` to `.playerTable` rename, which is done here. Other than that, the main thing is to make sure `addPlayer` and `removePlayer` manipulate `playerList`, so that `destroy` still works like before.

In addition, this contains some major refactors of `RoomBattle` and `Tournament` to take advantage of the updates.


Zarel:
I'll merge sometime after 24 hours of putting this up.

scheibo:
> A new array, `game.playerList`, will hold all players, with and without users. `game.players` has been renamed `game.playerTable` and will only be a table of players with users.

Hungarian notation is pretty garbage. These should probably be renamed `game.allPlayers` and `game.playersWithUsers` or something similar which actually reflects what they are meant to hold, not what type they happen to be.

Zarel:
Sure, you can do `players`/`playersByUserid` instead of the current `players`/`playerTable` if you'd like.

scheibo:
> (For diff readability, `game.players` has not been renamed in this PR, but will be renamed at merge time.)

Have you given up on `playerList`? That's the main one I didn't like.

Zarel:
> Have you given up on `playerList`? That's the main one I didn't like.

I thought it _was_ renamed at merge time? Was my message unclear?

Zarel:
Oh, my message _was_ unclear.

There were two planned renames:

- `game.players` -> `game.playerTable`
- `game.playerList` -> `game.players`

Both were done at merge-time.

Zarel:
I thought about doing them in two steps, so people would have more sensible error messages while updating any RoomGame not in Zarel:master. But then I didn't.

PR #5493, opened 2019-05-09
Fix Fusion Bolt and Fusion Flare's interaction with each other

jmclemo6:
This should fix https://github.com/Zarel/Pokemon-Showdown/projects/3#card-18662639

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -6187,10 +6187,12 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
 		onBasePower(basePower, pokemon) {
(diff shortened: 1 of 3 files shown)


Marty-D:
`moveThisTurn` isn't necessarily the last move used by any Pokemon that turn. In Triples for example, up to 4 Pokemon could have used a different move in between, in which case the next Fusion move shouldn't be doubled.

jmclemo6:
RIP. I'll take a more intensive crack at this later.

jmclemo6:
@Marty-D does this look alright now?

Marty-D:
Looks good to me, thanks! 👍

PR #5497, opened 2019-05-10
Stop hidetext message from getting hidden

whalemer:
Moving the log message to after the unlink messages means that the hide/show prompt has nothing to attach to, making it look weird.


Zarel:
So what's the fix for hiding your own messages? Hardcode in client?

whalemer:
The "user's messages were cleared" shouldn't hide at all now, since it's not attributed to a user

Zarel:
I'm not sure how much I like that. But I guess the downsides are pretty minor.

whalemer:
The other thing I thought of would be to change the unlink message to include all users (instead of a separate message for each user), and then make the client put in the "user's messages were cleared" message, but that seems a little dodgy.

PR #5498, opened 2019-05-14
Scavengers: Viewhunt/Queue improvements/fixes

sparkychildcharlie:
- Add "unrated" hunts to queue system
- fix /forcestarthunt, an add an unrated option for this.
- allow /viewhunt to show the current question while leaving the unknown answer blank

I plan to soon do typescripting with scav stuff so it's like the other chat-plugins

<sub>I swear I am a good neko!!!</sub>

### DIFF
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -1024,7 +1024,7 @@ let commands = {
 	 * Hunt queuing
 	 */
 	queueunrated: 'queue',
-	queueregular: 'queue',
+	queuerated: 'queue',
 	queue(target, room, user) {
 		if (room.id !== 'scavengers' && !(room.parent && room.parent.id === 'scavengers')) return this.errorReply("This command can only be used in the scavengers room.");
 		if (!target) {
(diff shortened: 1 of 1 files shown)


Zarel:
As far as I know, we have no devs on Scavengers? Someone will need to run these changes past Scavs auth.

sparkychildcharlie:
If i held back on my salty shitposting I would be lol. Anyways regarding this, it has full approval from Meicoo (screenshots on Discord)

server/chat-plugins/scavengers.js
```diff
@@ -1543,7 +1556,7 @@ exports.commands = {
 			"- /scav resetladder - resets the monthly scavenger leaderboard. (Requires: # & ~)",
 			"- /scav setpoints [1st place], [2nd place], [3rd place], [4th place], [5th place], ... - sets the point values for the wins. Use `/scav setpoints` to view what the current point values are. (Requires: # & ~)",
 			"- /scav setblitz [value] ... - sets the blitz award to `value`. Use `/scav setblitz` to view what the current blitz value is. (Requires: # & ~)",
-			"- /scav queue <em>[host] | [hint] | [answer] | [hint] | [answer] | [hint] | [answer] | ...</em> - queues a scavenger hunt to be started after the current hunt is finished. (Requires: % @ * # & ~)",
+			"- /scav queue(unrated) <em>[host] | [hint] | [answer] | [hint] | [answer] | [hint] | [answer] | ...</em> - queues a scavenger hunt to be started after the current hunt is finished. (Requires: % @ * # & ~)",
```
Zarel: I don't think we use the `(unrated)` syntax, it's pretty confusing here. I'd make them different commands in the list.

sparkychildcharlie: What would you prefer instead? I'm just using the same syntax as the /start(mini/official/unratrd/practice)hunt help entry above)

Zarel: > What would you prefer instead?

> I'd make them different commands in the list.

Zarel: Oh, I didn't know they did that for the other ones.

Zarel: I'd at least suggest `queue(rated/unrated)` to make it clearer what's going on there.

sparkychildcharlie:
@Zarel I've updated it :3

xfix:
I'm not a fan of this huge line 183, but the change seems fine.

PR #5500, opened 2019-05-14
Track forcerenamed names

whalemer:


Zarel:
How does this differ from the old tracker?

whalemer:
This is designed to stop
user gets forcerenamed -> user goes to a different name -> user immediately alts back
The current tracker only notices the first rename.

server/chat.js
```diff
@@ -1948,6 +1948,8 @@ Chat.filterWords = {};
 Chat.monitors = {};
 /** @type {Map<string, string>} */
 Chat.namefilterwhitelist = new Map();
+/** @type {Map<string, string>} */
```
Zarel: This should probably document what the key and value are.

Zarel:
```suggestion
/**
 * Inappropriate userid : forcerenaming staff member's userid
 * @type {Map<string, string>}
 */
```

Zarel: Also I never thought we should be blacklisting userids. Sometimes we want to forcerename people for e.g. having swastika emoji in their name, but there's nothing wrong with the userid.

whalemer: Hence why the only thing this does is notify staff

Zarel: Point of order: Don't resolve conversations after replying to them, you make it so I can't see it. Just reply; I'll resolve it once I've seen it.

server/chat-commands.js
```diff
@@ -2535,12 +2535,23 @@ const commands = {
...
 		targetUser.send(`|nametaken||${user.name} considers your name inappropriate${(reason ? `: ${reason}` : ".")}`);
 		return true;
 	},
-	forcerenamehelp: [`/forcerename OR /fr [username], [reason] - Forcibly change a user's name and shows them the [reason]. Requires: % @ & ~`],
+	forcerenamehelp: [
+		`/forcerename OR /fr [username], [reason] - Forcibly change a user's name and shows them the [reason]. Requires: % @ & ~`,
+		`/nameok [username] - Unmarks a forcerenamed username, stopping staff from being notified when it is used. Requires % @ & ~`,
+	],
+
+	nameok(target) {
```
Zarel: Don't we have an `/allowname` or something?

```js
	allowname(target, room, user) {
		if (" +".includes(user.group)) return fakeUnrecognized.call(this);
		target = toId(target);
		Chat.namefilterwhitelist.set(target, user.name);
		Rooms('staff').add("'" + target + "' was allowed as a username by " + user.name + ".");
		Rooms('staff').update();
		Rooms('upperstaff').add("'" + target + "' was allowed as a username by " + user.name + ".");
		Rooms('upperstaff').update();
	},
```

in our private chat plugin.

Feel free to merge those?

bumbadadabum:
There is an /allowname in server/chat-plugins/chat-monitor.js

whalemer:
>Don't resolve conversations after replying to them

Didn't realise I was doing that, my bad.

server/chat-commands.js
```diff
@@ -2535,12 +2535,16 @@ const commands = {
...
+		Chat.forceRenames.set(targetUser.userid, user.userid);
 		Ladders.cancelSearches(targetUser);
 		targetUser.resetName(true);
 		targetUser.send(`|nametaken||${user.name} considers your name inappropriate${(reason ? `: ${reason}` : ".")}`);
 		return true;
 	},
-	forcerenamehelp: [`/forcerename OR /fr [username], [reason] - Forcibly change a user's name and shows them the [reason]. Requires: % @ & ~`],
+	forcerenamehelp: [
+		`/forcerename OR /fr [username], [reason] - Forcibly change a user's name and shows them the [reason]. Requires: % @ & ~`,
+		`/allowname [username] - Unmarks a forcerenamed username, stopping staff from being notified when it is used. Requires % @ & ~`,
```
Zarel: As far as I can tell, you didn't implement `/allowname`.

whalemer:
https://github.com/Zarel/Pokemon-Showdown/blob/a394da45cb71a8992565e04d05c163ecf8489229/server/chat-plugins/chat-monitor.js#L228-L230

It's handled here

Zarel:
@whalemer that code doesn't "Unmarks a forcerenamed username" like the added doc suggests, though.

whalemer:
It functionally does the same thing, though, and I feel like the command is descriptive enough in that way. You're saying that the name is ok, both to the automated namefilter and to someone that previously forcerenamed it.

Zarel:
Hm, okay, I see it now.

There's still some value in clearing it from the `forceRenames` map, but I suppose it isn't urgent.

PR #5501, opened 2019-05-15
LoginServer: Fix makeRequests

TheJetOU:


server/loginserver.ts
```diff
@@ -151,13 +151,11 @@ class LoginServerInstance {
 			'&nocache=' + new Date().getTime() +
 			'&json=' + encodeURIComponent(JSON.stringify(dataList)) + '\n';
 
-		const requestOptions = {
-			url: url.parse(`${this.uri}action.php`),
-			method: 'post',
-			headers: {
+		const requestOptions = url.parse(this.uri + 'action.php');
```
scheibo:
```suggestion
		const requestOptions: any = url.parse(`${this.uri}action.php`);
```

server/loginserver.ts
```diff
@@ -151,13 +151,11 @@ class LoginServerInstance {
...
 
-		const requestOptions = {
-			url: url.parse(`${this.uri}action.php`),
-			method: 'post',
-			headers: {
+		const requestOptions: AnyObject = url.parse(`${this.uri}action.php`);
+		requestOptions.method = 'post';
+		requestOptions.headers = {
 				'Content-Type': 'application/x-www-form-urlencoded',
 				'Content-Length': postData.length,
```
Zarel: Weird that linter doesn't consider this an indentation error.

scheibo: True. 032096b53b36566b1f66d5e1a81ea781a3b97411 fixes, I was mainly concerned with unbreaking the ladder and not about style

PR #5502, opened 2019-05-15
Chat-commands: Fix hotpatching loginserver

TheJetOU:


server/chat-commands.js
```diff
@@ -3029,8 +3029,8 @@ const commands = {
 				this.sendReply("Formats have been hot-patched.");
 			} else if (target === 'loginserver') {
 				FS('config/custom.css').unwatch();
-				Chat.uncache('./server/loginserver');
-				global.LoginServer = require('./loginserver').LoginServer;
+				Chat.uncacheDir('./.server-dist/loginserver');
```
scheibo: `uncacheDir` seems odd here because `loginserver` is not a directory?

```suggestion
				Chat.uncache('./.server-dist/loginserver');
```

TheJetOU: Yeah, that was a mistake on my end.

PR #5503, opened 2019-05-16
Remove additional toID calls

scheibo:
The other `getEffect` calls all are going to require `data` to be Typescripted because we need to perform *beaucoup* casting.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1627,7 +1627,7 @@ export class Battle extends Dex.ModdedDex {
 	) {
 		if (!targetArray) return [0];
 		let retVals: (number | false | undefined)[] = [];
-		if (typeof effect === 'string' || !effect) effect = this.getEffect(effect);
+		if (typeof effect === 'string' || !effect) effect = this.getEffectByID((effect || '') as ID);
 		for (const [i, curDamage] of damage.entries()) {
 			const target = targetArray[i];
 			let targetDamage = curDamage;
@@ -1751,7 +1751,7 @@ export class Battle extends Dex.ModdedDex {
 		if (!damage) return 0;
 		damage = this.clampIntRange(damage, 1);
 
-		if (typeof effect === 'string' || !effect) effect = this.getEffect(effect);
+		if (typeof effect === 'string' || !effect) effect = this.getEffectByID((effect || '') as ID);
 
 		// In Gen 1 BUT NOT STADIUM, Substitute also takes confusion and HJK recoil damage
 		if (this.gen <= 1 && this.currentMod !== 'stadium' &&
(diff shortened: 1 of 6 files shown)


sim/battle.ts
```diff
@@ -2342,7 +2342,7 @@ export class Battle extends Dex.ModdedDex {
 				}
 			} else if (action.choice === 'switch' || action.choice === 'instaswitch') {
 				if (typeof action.pokemon.switchFlag === 'string') {
-					action.sourceEffect = this.getEffect(action.pokemon.switchFlag);
+					action.sourceEffect = (this.getMove(action.pokemon.switchFlag as ID) as unknown) as PureEffect;
```
scheibo: We get around this is `Dex.getEffect` with `// @ts-ignore`, but apparently thats really bad and `Foo -> unknown -> Bar` is the 'correct' way to abuse the type system.

Zarel: I wouldn't use parens for the double cast. What does "apparently" imply here?

sim/pokemon.ts
```diff
@@ -113,10 +113,10 @@ export class Pokemon {
 
 	/**
 	 * If the switch is called by an effect with a special switch
-	 * message, like U-turn or Baton Pass, this will be the fullname of
+	 * message, like U-turn or Baton Pass, this will be the ID of
 	 * the calling effect.
 	 */
-	switchFlag: boolean | string;
+	switchFlag: ID | boolean;
```
Zarel: Is `switchFlag` ever `true`? If not, this should be `ID | null`, which renders your cast above a significantly easier `!` rather than `as ID`

scheibo: https://github.com/Zarel/Pokemon-Showdown/blob/41d16984527562dc72bbcc46cb6ce85bd1ee2da4/data/abilities.js#L857-L858

sim/battle.ts
Zarel: Ideally, with the right other changes, this can be:

```suggestion
					action.sourceEffect = this.getMove(action.pokemon.switchFlag!) as any;
```

sim/pokemon.ts
Zarel: Very sad. The comment should probably be updated.

Zarel: Hm, is there any harm in setting `switchFlag` to `'emergencyexit' as ID`?

scheibo: There are multiple places where `switchFlag` is set to `true`, not just `Emergency Exit` (`Battle#checkFainted`, `Wimp Out`, `Eject Button` and Gen 4 `Healing Wish` / `Lunar Dance`). Refactoring these to use `ID` instead of `true` is out of scope for this change, IMO

sim/battle.ts
scheibo: Not sure why you'd want to cast to `any`.

sim/pokemon.ts
Zarel: Fair.

sim/battle.ts
Zarel: It's an easier cast and approximately as safe.

scheibo: Done.

PR #5504, opened 2019-05-19
Don't notify using a forcerenamed name for namelocked users

whalemer:


Zarel:
On second thought, this seems like the incorrect solution.

Zarel:
It should still warn if other, non-namelocked users use that name, right?

Zarel:
Hm, actually, other users will still get namelocked for trying to use that name, so I guess the warning isn't necessary in that case, either.

PR #5508, opened 2019-05-20
Implement TypeScript interface for config groups

Kaiepi:
Needed for https://github.com/Zarel/Pokemon-Showdown/pull/5489

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
  *                       lower in rank than the current group.
  *     - roomonly: forces the group to be a per-room moderation rank only.
  *     - globalonly: forces the group to be a global rank only.
+ *
+ *   Permissions are a string key such as `mute` followed by a boolean. `true` means group
+ *   members have that permission, `false` means they don't, and `undefined` means permission
+ *   is inherited from the `inherit` group.
+ *
+ *   Some permissions, such as mute/ban, have a target (user you're doing the action to).
+ *   These, when set to `true`, will use the jurisdiction list in the `jurisdiction` key to
+ *   determine whether or not you have permission to do it to that user. You can also set them
+ *   to a string to use it as a custom jurisdiction list, replacing the one in the
+ *   `jurisdiction` key.
+ *
  *   All the possible permissions are as follows:
  *     - console: Developer console (>>).
  *     - lockdown: /lockdown and /endlockdown commands.
  *     - hotpatch: /hotpatch, /crashfixed and /savelearnsets commands.
  *     - ignorelimits: Ignore limits such as chat message length.
  *     - promote: Promoting and demoting. Will only work if the target user's current
- *                  group and target group are both in jurisdiction.
+ *                  group and target group are both in jurisdiction. Has a target.
  *     - room<rank>: /roompromote to <rank> (eg. roomvoice)
  *     - makeroom: Create/delete chatrooms, and set modjoin/roomdesc/privacy
  *     - editroom: Editing properties of rooms
  *     - editprivacy: Set modjoin/privacy only for battles
- *     - ban: Banning and unbanning.
- *     - mute: Muting and unmuting.
- *     - lock: locking (ipmute) and unlocking.
+ *     - ban: Banning and unbanning. Target is subject to jurisdiction.
+ *     - mute: Muting and unmuting. Target is subject to jurisdiction.
+ *     - lock: locking (ipmute) and unlocking. Target is subject to jurisdiction.
  *     - receivemutedpms: Receive PMs from muted users.
  *     - forcerename: /fr command.
- *     - ip: IP checking.
- *     - alts: Alt checking.
+ *     - ip: IP checking. Target is subject to jurisdiction.
+ *     - alts: Alt checking. Target is subject to jurisdiction.
  *     - modlog: view the moderator logs.
  *     - broadcast: Broadcast informational commands.
  *     - declare: /declare command.
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -429,7 +429,7 @@ exports.disablehotpatchall = false;
  *     - makeroom: Create/delete chatrooms, and set modjoin/roomdesc/privacy
  *     - editroom: Editing properties of rooms
  *     - editprivacy: Set modjoin/privacy only for battles
- *     - ban: Banning and unbanning. Target is subject to jurisdiction.
+ *     - ban: Banning and unbanning. Has a target.
  *     - mute: Muting and unmuting. Target is subject to jurisdiction.
  *     - lock: locking (ipmute) and unlocking. Target is subject to jurisdiction.
  *     - receivemutedpms: Receive PMs from muted users.
--- a/server/globals.ts
+++ b/server/globals.ts
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
 declare let Sockets: typeof import('./sockets');
 // let TeamValidator: typeof import('../sim/team-validator');
 declare let TeamValidatorAsync: typeof import('./team-validator-async');
+
+interface Group {
+	symbol: string,
+	id: string,
+	name: string,
+
+	root?: boolean,
+	inherit?: string,
+	jurisdiction?: string,
+	roomonly?: boolean,
+	globalonly?: boolean,
+	console?: boolean,
+	lockdown?: boolean,
+	hotpatch?: boolean,
+	ignorelimits?: boolean,
+	promote?: string | boolean,
+	makeroom?: boolean,
+	editroom?: boolean,
+	editprivacy?: boolean,
+	ban?: string | boolean,
+	mute?: string | boolean,
+	lock?: string | boolean,
+	receivemutedpms?: boolean,
+	forcerename?: boolean,
+	ip?: string | boolean,
+	alts?: string | boolean,
+	modlog?: boolean,
+	broadcast?: boolean,
+	declare?: boolean,
+	announce?: boolean,
+	modchat?: boolean,
+	modchatall?: boolean,
+	potd?: boolean,
+	forcewin?: boolean,
+	battlemessage?: boolean,
+	tournaments?: boolean,
+	gamemoderation?: boolean,
+	gamemanagement?: boolean,
+	minigame?: boolean,
+	game?: boolean,
+
+	// Needed for room<rank> settings and to allow for custom settings on other
+	// servers.
+	[key: string]: string | boolean | undefined
+}
--- a/server/globals.ts
+++ b/server/globals.ts
@@ -22,7 +22,9 @@ interface Group {
 	id: string,
 	name: string,
 
+	/** Permission bypass. `true` will give the group every permission, bypassing all other settings. */
 	root?: boolean,
+
 	inherit?: string,
 	jurisdiction?: string,
 	roomonly?: boolean,
--- a/server/globals.ts
+++ b/server/globals.ts
@@ -23,12 +23,17 @@ interface Group {
 	name: string,
 
 	/** Permission bypass. `true` will give the group every permission, bypassing all other settings. */
+	/** Permission bypass. `true` will give the group every permission, bypassing all other settings. */
 	root?: boolean,
 
+	/** Group symbol to inherit undefined permissions from. */
 	inherit?: string,
+	/** List of default allowed targets for targeted permissions like `mute`, `ban`, etc. */
 	jurisdiction?: string,
+	/** `true` to disallow globally promoting users to this group. */
 	roomonly?: boolean,
 	globalonly?: boolean,
+
 	console?: boolean,
 	lockdown?: boolean,
 	hotpatch?: boolean,


sim/globals.ts
```diff
@@ -1370,3 +1370,48 @@ interface PokemonModData {
...
+	battlemessage?: boolean,
+	tournaments?: boolean,
+	gamemoderation?: boolean,
+	gamemanagement?: boolean,
+	minigame?: boolean,
+	game?: boolean,
+
+	// Needed for room<rank> settings and to allow for custom settings on other
+	// servers.
+	[key: string]: any
```
Zarel: `[key: string]: unknown`?

`[key: string]: boolean | string`?

Kaiepi: I didn't want to make it `boolean | string` in case other servers use settings that aren't those types.

I'm way out of date with TypeScript, I didn't know `unknown` existed. That'd probably fit better

Kaiepi:
Updated

sim/globals.ts
```diff
@@ -1370,3 +1370,48 @@ interface PokemonModData {
 	originalSpecies?: string; // Mix and Mega
 	[key: string]: any;
 }
+
+interface Group {
```
scheibo: I believe this belongs in `server/globals.ts`, not `sim/globals.ts`.

Kaiepi:
Updated

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
 declare let Sockets: typeof import('./sockets');
 // let TeamValidator: typeof import('../sim/team-validator');
 declare let TeamValidatorAsync: typeof import('./team-validator-async');
+
+interface Group {
+	symbol: string,
+	id: string,
```
scheibo: Should `id` be an `ID` type? (is it always all lowercase etc?)

scheibo:
LG pending the question about `ID` :)

server/globals.ts
Zarel: Yes, I think group IDs are guaranteed to be IDs. However, having to type `/** @type {ID} */ ('moderator')` might be too annoying in config.js, if that's the plan.

Zarel:
Overall I'd prefer this to be in the same PR as the thing using it.

server/globals.ts
scheibo: True. Is `config.js` even included in typechecking though? Its not in `tsconfig.json`'s `include`, but maybe it still gets pulled in?

Zarel: I'm not sure. It _was_, but at some point this was changed, I think.

Kaiepi:
>Overall I'd prefer this to be in the same PR as the thing using it.
I split this into two different pullreqs since `Group` may eventually be used by other parts of PS. I can put this commit in the `|groups|` pullreq if you want, though

Zarel:
I guess most of the problem here is that it's hard to review when I don't understand the use-case.

You don't need to merge them into one PR; just answer the question about IDs.

Kaiepi:
I don't think `ID` should be the type of `id` since server owners may not make the id a proper ID. Maybe if `config.js` were TypeScript it'd make more sense, but for the time being I don't think so.

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	battlemessage?: boolean,
+	tournaments?: boolean,
+	gamemoderation?: boolean,
+	gamemanagement?: boolean,
+	minigame?: boolean,
+	game?: boolean,
+
+	// Needed for room<rank> settings and to allow for custom settings on other
+	// servers.
+	[key: string]: unknown
```
Zarel:
```suggestion
	[key: string]: string | boolean
```

`unknown` is definitely excessive. We know for a fact that every permission is either `string` (custom jurisdiction), `true` (use the `jurisdiction` key), or `false | undefined` (access denied)

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	root?: boolean,
+	inherit?: string,
+	jurisdiction?: string,
+	roomonly?: boolean,
+	globalonly?: boolean,
+	console?: boolean,
+	lockdown?: boolean,
+	hotpatch?: boolean,
+	ignorelimits?: boolean,
+	promote?: string,
```
Zarel:
```suggestion
	promote?: string | boolean,
```

Any permission with a target can take either a custom jurisdiction, or `true` meaning to fall back to the `jurisdiction` key.

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	makeroom?: boolean,
+	editroom?: boolean,
+	editprivacy?: boolean,
+	ban?: boolean,
```
Zarel:
```suggestion
	ban?: string | boolean,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	mute?: boolean,
```
Zarel:
```suggestion
	mute?: string | boolean,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	lock?: boolean,
```
Zarel:
```suggestion
	lock?: string | boolean,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	receivemutedpms?: boolean,
+	forcerename?: boolean,
+	ip?: boolean,
```
Zarel:
```suggestion
	ip?: string | boolean,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	alts?: boolean,
```
Zarel:
```suggestion
	alts?: string | boolean,
```

server/globals.ts
Kaiepi: I think that's missing from the documentation in `config.js`. I'll add that if that's the case

Zarel: Yeah, it could very well be missing. I'm kind of bad at documentation sometimes.

config/config-example.js
```diff
@@ -413,18 +413,24 @@ exports.disablehotpatchall = false;
...
+ *                  group and target group are both in jurisdiction. Setting this to
+ *                  true will fallback to the previous jurisdiction's permission level.
  *     - room<rank>: /roompromote to <rank> (eg. roomvoice)
  *     - makeroom: Create/delete chatrooms, and set modjoin/roomdesc/privacy
  *     - editroom: Editing properties of rooms
  *     - editprivacy: Set modjoin/privacy only for battles
- *     - ban: Banning and unbanning.
- *     - mute: Muting and unmuting.
- *     - lock: locking (ipmute) and unlocking.
+ *     - ban: Banning and unbanning. Setting to true will fallback to the
```
Zarel: I wouldn't phrase it this way. It doesn't fall back to the previous jurisdiction, it falls back to the current group's default jurisdiction.

I'd also just phrase all of the individual permissions as "Target subject to jurisdiction", and describe how jurisdiction works at the top of the comment.

config/config-example.js
```diff
@@ -413,18 +413,24 @@ exports.disablehotpatchall = false;
  *     - hotpatch: /hotpatch, /crashfixed and /savelearnsets commands.
```
Zarel: So where it currently says:

```
 *   All the possible permissions are as follows:
```

```
 *
 * Permissions are a string key such as `mute` followed by a boolean. `true` means group
 * members have that permission, `false` means they don't, and `undefined` means permission
 * is inherited from the `inherit` group.
 * 
 * Some permissions, such as mute/ban, have a target (user you're doing the action to). These,
 * when set to `true`, will use the jurisdiction list in the `jurisdiction` key to determine whether
 * or not you have permission to do it to that user. You can also set them to a string to use that
 * string as a custom jurisdiction list, replacing the one in the `jurisdiction` key.
```

At which point, on the individual permissions, you can just say "has a target".

config/config-example.js
Kaiepi: Something like this?
```
 *     - promote: Promoting and demoting. Will only work if the target user's current
 *                  group and target group are both in jurisdiction. Target is subject to
 *                  jurisdiction. Setting this to true will fallback to the current group's
 *                  default jurisdiction.
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+interface Group {
+	symbol: string,
+	id: string,
+	name: string,
+
+	root?: boolean,
+	inherit?: string,
+	jurisdiction?: string,
+	roomonly?: boolean,
+	globalonly?: boolean,
```
Zarel:
```suggestion
	globalonly?: boolean,

```

(Separate the permission metadata above from the permissions list below.)

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
 declare let Sockets: typeof import('./sockets');
 // let TeamValidator: typeof import('../sim/team-validator');
 declare let TeamValidatorAsync: typeof import('./team-validator-async');
+
+interface Group {
+	symbol: string,
+	id: string,
+	name: string,
+
+	root?: boolean,
```
Zarel:
```suggestion
	/** Permission bypass. `true` will give the group every permission, bypassing all other settings. */
	root?: boolean,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	inherit?: string,
```
Zarel:
```suggestion
	/** Group symbol to inherit undefined permissions from. */
	inherit?: string,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	jurisdiction?: string,
```
Zarel:
```suggestion
	/** List of default allowed targets for targeted permissions like `mute`, `ban`, etc. */
	jurisdiction?: string,
```

server/globals.ts
```diff
@@ -16,3 +16,48 @@ declare let IPTools: typeof import('./ip-tools').IPTools;
...
+	roomonly?: boolean,
```
Zarel:
```suggestion
	/** `true` to disallow globally promoting users to this group. */
	roomonly?: boolean,
```

server/globals.ts
Zarel:
```suggestion
	/** `true` to disallow room-promoting users to this group. */
	globalonly?: boolean,

```

config/config-example.js
Zarel: More like

```
 *     - promote: Promoting and demoting. Will only work if the target user's current
 *                  group and target group are both in jurisdiction. Has a target.
```

Kaiepi:
I'm not sure how I feel about documenting what the individual settings do outside of `config.js` since they could potentially change in the future, possibly making these comments misleading if someone forgets to update them.

config/config-example.js
```diff
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
...
- *                  group and target group are both in jurisdiction.
+ *                  group and target group are both in jurisdiction. Has a target.
  *     - room<rank>: /roompromote to <rank> (eg. roomvoice)
  *     - makeroom: Create/delete chatrooms, and set modjoin/roomdesc/privacy
  *     - editroom: Editing properties of rooms
  *     - editprivacy: Set modjoin/privacy only for battles
- *     - ban: Banning and unbanning.
- *     - mute: Muting and unmuting.
- *     - lock: locking (ipmute) and unlocking.
+ *     - ban: Banning and unbanning. Target is subject to jurisdiction.
```
Zarel:
```suggestion
 *     - ban: Banning and unbanning. Has a target.
```

I meant to use "Has a target" rather than "Target is subject to jurisdiction" for all of these, in case that was unclear.

Kaiepi:
Updated

config/config-example.js
```diff
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
...
+ *                  group and target group are both in jurisdiction. Has a target.
  *     - room<rank>: /roompromote to <rank> (eg. roomvoice)
  *     - makeroom: Create/delete chatrooms, and set modjoin/roomdesc/privacy
  *     - editroom: Editing properties of rooms
  *     - editprivacy: Set modjoin/privacy only for battles
- *     - ban: Banning and unbanning.
- *     - mute: Muting and unmuting.
- *     - lock: locking (ipmute) and unlocking.
+ *     - ban: Banning and unbanning. Has a target.
+ *     - mute: Muting and unmuting. Target is subject to jurisdiction.
```
Zarel:
```suggestion
 *     - mute: Muting and unmuting. Has a target.
```

config/config-example.js
```diff
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
...
+ *     - lock: locking (ipmute) and unlocking. Target is subject to jurisdiction.
```
Zarel:
```suggestion
 *     - lock: locking (ipmute) and unlocking. Has a target.
```

config/config-example.js
```diff
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
...
  *     - receivemutedpms: Receive PMs from muted users.
  *     - forcerename: /fr command.
- *     - ip: IP checking.
- *     - alts: Alt checking.
+ *     - ip: IP checking. Target is subject to jurisdiction.
```
Zarel:
```suggestion
 *     - ip: IP checking. Has a target.
```

config/config-example.js
```diff
@@ -407,24 +407,35 @@ exports.disablehotpatchall = false;
...
+ *     - alts: Alt checking. Target is subject to jurisdiction.
```
Zarel:
```suggestion
 *     - alts: Alt checking. Has a target.
```

AnnikaCodes:

PR #5510, opened 2019-05-21
Clear forcerename monitor names on namelock more completely

whalemer:


Zarel:
Okay, now this one I don't get. Is the entire list of usernames namelocked? Is that why you're doing this?

whalemer:
Yes, (i think) that that list of names will namelock on login, so reporting isn't necessary

Zarel:
I _still_ think it makes more sense not to report any names that are namelocked on login, rather than manually clearing them all here.

whalemer:
I don't think that's possible without changing when the namefilter is checked, they're not namelocked at that point

Zarel:
The simple solution is to move the reporting to outside of the namefilter check.

Zarel:
And/or check namelocks before namefilter.

whalemer:
Can I move it to a loginfilter?

Zarel:
That's a possibility.

whalemer:
bump?

PR #5512, opened 2019-05-23
Fix quotation marks when a username is already in use

SuddenlyQ:
Fixes a bug which was mentioned in the Smogon bug thread. I'm Annika0 on PS btw.


scheibo:
Thanks, Annika0 :)

PR #5513, opened 2019-05-26
Provide a hint regarding '/offertie' in EBC situations

scheibo:
Not confident about the wording in the slightest, and obviously `/offertie` is a lame solution to EBC shortcomings, but I think it's going to be difficult for user's to figure out `/offertie` is even an option without this or something similar.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1456,7 +1456,7 @@ export class Battle extends Dex.ModdedDex {
 			}
 			const turnsLeftText = (turnsLeft === 1 ? `1 turn` : `${turnsLeft} turns`);
 			this.add('bigerror', `You will auto-tie if the battle doesn't end in ${turnsLeftText} turns (on turn 1000).`);
-			this.hint("Consider using ``/offertie`` to see if your opponent would be willing to accept a tie earlier if you feel this battle is a stalemate.");
+			this.hint("If you want to tie earlier, consider using `/offertie`.");
 		}
 
 		// Are all Pokemon on every side stale, with at least one side containing an externally stale Pokemon?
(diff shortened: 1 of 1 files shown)


sim/battle.ts
```diff
@@ -1455,7 +1455,7 @@ export class Battle extends Dex.ModdedDex {
 				return true;
 			}
 			const turnsLeftText = (turnsLeft === 1 ? `1 turn` : `${turnsLeft} turns`);
-			this.add('bigerror', `You will auto-tie if the battle doesn't end in ${turnsLeftText} turns (on turn 1000).`);
+			this.add('bigerror', `You will auto-tie if the battle doesn't end in ${turnsLeftText} turns (on turn 1000). Consider using '/offertie' to see if your opponent would be willing to accept a tie earlier if you feel this battle is a stalemate.`);
```
Zarel: Maybe not this entire message as a `bigerror`?

scheibo: Made this a `hint`. Does ` `` `   (I still cant escape it properly on GH, and I dont care) work the way I hope?

Zarel: Man, don't resolve a conversation immediately after asking a question.

Anyway, I don't actually know.

Zarel: I just tested and it turns out the answer is "no".

sim/battle.ts
```diff
@@ -1456,6 +1456,7 @@ export class Battle extends Dex.ModdedDex {
 			}
 			const turnsLeftText = (turnsLeft === 1 ? `1 turn` : `${turnsLeft} turns`);
 			this.add('bigerror', `You will auto-tie if the battle doesn't end in ${turnsLeftText} turns (on turn 1000).`);
+			this.hint("Consider using ``/offertie`` to see if your opponent would be willing to accept a tie earlier if you feel this battle is a stalemate.");
```
```suggestion
			this.hint("If you want to tie earlier, consider `/offertie`.");
```

PR #5515, opened 2019-05-31
Add commands for upper staff to add custom avatars

maxalexandderpi:
I was told basing off of other's code was fine if i credited as so, please excuse me if i'm wrong otherwise.

### DIFF
--- a/server/chat-plugins/customavatars.js
+++ b/server/chat-plugins/customavatars.js
@@ -1,0 +1,106 @@
+//Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
+//Credits to CreaturePhil and the other listed contributors.
+//updated for the main server by Maxalexanderpi, with help from Hoeenhero.
+'use strict';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+const path = require('path');
+const fs = require('fs');
+
+/* eslint no-restricted-modules: [0] */
+const AVATAR_PATH = 'config/avatars/';
+
+const VALID_EXTENSIONS = ['.png'];
+
+const https = require('https');
+
+function downloadImage(image_url, name, extension) {
+	let req = https.get(image_url, res => {
+		 res.setEncoding('binary');
+		 res.on('response', response => {
+			 if (response.statusCode !== 200) return;
+			 let type = response.headers['content-type'].split('/');
+			 if (type[0] !== 'image') return;
+
+			response.pipe(FS(AVATAR_PATH + name + extension).createWriteStream());
+		 });
+	});
+	req.on('error', e => {
+		console.error(e);
+	});
+	req.end();
+}
+function loadCustomAvatars() {
+	fs.readdir(AVATAR_PATH, (err, files) => {
+		if (err) console.log("Error loading custom avatars: " + err);
+		if (!files) files = [];
+		files
+			.filter(file => VALID_EXTENSIONS.includes(path.extname(file)))
+			.forEach(file => {
+				let name = path.basename(file, path.extname(file));
+				Config.customavatars[name] = file;
+			});
+	});
+}
+
+loadCustomAvatars();
+
+exports.commands = {
+	customavatar: {
+		add(target, room, user) {
+			if (!this.can('avatar')) return false;
+			let parts = target.split(',').map(param => param.trim());
+			if (parts.length < 2) return this.parse('/help customavatar');
+
+			let name = toID(parts[0]);
+			let avatarUrl = parts[1];
+			if (!/^https?:\/\//i.test(avatarUrl)) avatarUrl = 'http://' + avatarUrl;
+			let ext = path.extname(avatarUrl);
+
+			if (!VALID_EXTENSIONS.includes(ext)) {
+				return this.errorReply("Image url must end in a .png extension.");
+			}
+
+			Config.customavatars[name] = name + ext;
+
+			downloadImage(avatarUrl, name, ext);
+			this.sendReply("|raw|" + name + "'s avatar was successfully set. Avatar:<br /><img src='" + avatarUrl + "' width='80' height='80'>");
+			Monitor.adminlog(name + "'s avatar was successfully set by " + user.name + ".");
+			if (Users(name)) Users(name).popup("|html|" + (user.name, 'Upper staff') + " have set your custom avatar.<br /><center><img src='" + avatarUrl + "' width='80' height='80'></center><br /> Refresh your page if you don't see it.");
+		},
+
+		remove(target, room, user) {
+			if (!this.can('avatar')) return false;
+
+			let userid = toID(target);
+			let image = Config.customavatars[userid];
+
+			if (!image) return this.errorReply(target + " does not have a custom avatar.");
+
+			delete Config.customavatars[userid];
+			fs.unlink(AVATAR_PATH + image, err => {
+				if (err && err.code === 'ENOENT') {
+					this.errorReply(target + "'s avatar does not exist.");
+				} else if (err) {
+					console.error(err);
+				}
+
+				if (Users(userid)) Users(userid).popup("|html|" + (user.name, 'Upper staff') + " have removed your custom avatar.");
+				this.sendReply(target + "'s avatar has been successfully removed.");
+				Monitor.adminlog(target + "'s avatar has been successfully removed by " + user.name + ".");
+			});
+		},
+
+		customavatarhelp: 'help',
+		help(target, room, user) {
+			this.parse('/help customavatar');
+		},
+	},
+
+	customavatarhelp: [
+		"Commands for /customavatar are:",
+		"/customavatar add [username], [image link] - Set a user's custom avatar. Requires: & ~",
+		"/customavatar remove [username] - Delete a user's custom avatar. Requires: & ~"],
+};
--- a/server/chat-plugins/customavatars.js
+++ b/server/chat-plugins/customavatars.js
@@ -1,20 +1,17 @@
-//Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
-//Credits to CreaturePhil and the other listed contributors.
-//updated for the main server by Maxalexanderpi, with help from Hoeenhero.
+// Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
+// Credits to CreaturePhil and the other listed contributors.
+// updated for the main server by Maxalexanderpi and Hoeenhero.
+*/
 'use strict';
 
 /** @type {typeof import('../../lib/fs').FS} */
-const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
-
 const path = require('path');
 const fs = require('fs');
+const https = require('https');
 
-/* eslint no-restricted-modules: [0] */
 const AVATAR_PATH = 'config/avatars/';
 
 const VALID_EXTENSIONS = ['.png'];
-
-const https = require('https');
 
 function downloadImage(image_url, name, extension) {
 	let req = https.get(image_url, res => {
@@ -24,7 +21,7 @@ function downloadImage(image_url, name, extension) {
 			 let type = response.headers['content-type'].split('/');
 			 if (type[0] !== 'image') return;
 
-			response.pipe(FS(AVATAR_PATH + name + extension).createWriteStream());
+			response.pipe(fs(AVATAR_PATH + name + extension).createWriteStream());
 		 });
 	});
 	req.on('error', e => {
@@ -44,7 +41,6 @@ function loadCustomAvatars() {
 			});
 	});
 }
-
 loadCustomAvatars();
 
 exports.commands = {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
+//Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
+//Credits to CreaturePhil and the other listed contributors.
+//updated for the main server by Maxalexanderpi, with help from Hoeenhero.
+'use strict';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+const path = require('path');
+const fs = require('fs');
```
scheibo: This should not be required when using `FS` (`FS` has `unlinkIfExists` and `readdir` already)

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+
+/* eslint no-restricted-modules: [0] */
+const AVATAR_PATH = 'config/avatars/';
+
+const VALID_EXTENSIONS = ['.png'];
+
+const https = require('https');
```
scheibo: should be grouped with the other imports

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+//updated for the main server by Maxalexanderpi, with help from Hoeenhero.
+'use strict';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+const path = require('path');
+const fs = require('fs');
+
+/* eslint no-restricted-modules: [0] */
```
scheibo: I don't think this is required?

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
+//Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
```
scheibo: Space between ``//` and the first character. Also, 'HoeenHero'.

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+	});
+	req.end();
+}
+function loadCustomAvatars() {
+	fs.readdir(AVATAR_PATH, (err, files) => {
+		if (err) console.log("Error loading custom avatars: " + err);
+		if (!files) files = [];
+		files
+			.filter(file => VALID_EXTENSIONS.includes(path.extname(file)))
+			.forEach(file => {
```
scheibo: re `forEach`: https://github.com/Zarel/Pokemon-Showdown/blob/master/CONTRIBUTING.md#es5-and-es6

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+
+function downloadImage(image_url, name, extension) {
+	let req = https.get(image_url, res => {
+		 res.setEncoding('binary');
+		 res.on('response', response => {
+			 if (response.statusCode !== 200) return;
+			 let type = response.headers['content-type'].split('/');
+			 if (type[0] !== 'image') return;
+
+			response.pipe(FS(AVATAR_PATH + name + extension).createWriteStream());
```
scheibo: all the string concatenations in this file can be template strings instead

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+
+/* eslint no-restricted-modules: [0] */
+const AVATAR_PATH = 'config/avatars/';
+
+const VALID_EXTENSIONS = ['.png'];
+
+const https = require('https');
+
+function downloadImage(image_url, name, extension) {
+	let req = https.get(image_url, res => {
```
scheibo: should probably be returning a promise

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+		 res.on('response', response => {
+			 if (response.statusCode !== 200) return;
+			 let type = response.headers['content-type'].split('/');
+			 if (type[0] !== 'image') return;
+
+			response.pipe(FS(AVATAR_PATH + name + extension).createWriteStream());
+		 });
+	});
+	req.on('error', e => {
+		console.error(e);
```
scheibo: not sure this is the correct error handling mechanism

HoeenCoder: Might be better to throw the error to crashlogger in hindsight.

I need to review the plugin as a whole when I have time, haven't done much other than offer advice on a few issues.

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+			if (!VALID_EXTENSIONS.includes(ext)) {
+				return this.errorReply("Image url must end in a .png extension.");
+			}
+
+			Config.customavatars[name] = name + ext;
+
+			downloadImage(avatarUrl, name, ext);
+			this.sendReply("|raw|" + name + "'s avatar was successfully set. Avatar:<br /><img src='" + avatarUrl + "' width='80' height='80'>");
+			Monitor.adminlog(name + "'s avatar was successfully set by " + user.name + ".");
+			if (Users(name)) Users(name).popup("|html|" + (user.name, 'Upper staff') + " have set your custom avatar.<br /><center><img src='" + avatarUrl + "' width='80' height='80'></center><br /> Refresh your page if you don't see it.");
```
TheJetOU: You should probably use `Chat.html` to sanitize the html. Same thing in L90 (the remove command).

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+			});
+	});
+}
+
+loadCustomAvatars();
+
+exports.commands = {
+	customavatar: {
+		add(target, room, user) {
+			if (!this.can('avatar')) return false;
```
TheJetOU: I don't think `avatar` is a permission in `config/config.js`?
```suggestion
			if (!this.can('declare')) return;
```

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+
+			let userid = toID(target);
+			let image = Config.customavatars[userid];
+
+			if (!image) return this.errorReply(target + " does not have a custom avatar.");
+
+			delete Config.customavatars[userid];
+			fs.unlink(AVATAR_PATH + image, err => {
+				if (err && err.code === 'ENOENT') {
+					this.errorReply(target + "'s avatar does not exist.");
```
TheJetOU: You should probably return this to prevent further execution of code.
```suggestion
					return this.errorReply(target + "'s avatar does not exist.");
```

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+const path = require('path');
+const fs = require('fs');
+
+/* eslint no-restricted-modules: [0] */
+const AVATAR_PATH = 'config/avatars/';
+
+const VALID_EXTENSIONS = ['.png'];
```
TheJetOU: Since it contains one value, is there any reason you're using an array instead of a string?

server/chat-plugins/customavatars.js
TheJetOU: Also, `<center>` is not supported in HTML, you can use CSS, but I don't see why this should be centered anyway. It also isn't centered in `/avatar`.

maxalexandderpi:
resolving as i'm fixing, will commit once everything's done so it's less spammy.

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,106 @@
...
+			let ext = path.extname(avatarUrl);
+
+			if (!VALID_EXTENSIONS.includes(ext)) {
+				return this.errorReply("Image url must end in a .png extension.");
+			}
+
+			Config.customavatars[name] = name + ext;
+
+			downloadImage(avatarUrl, name, ext);
+			this.sendReply("|raw|" + name + "'s avatar was successfully set. Avatar:<br /><img src='" + avatarUrl + "' width='80' height='80'>");
```
TheJetOU: This _could_ be wrong grammar. If I have the username "Ares" it would show "Ares's" which is wrong. If you don't want to implement the suggestion below, you can just change the wording to prevent that from occurring.
```suggestion
			this.sendReply(`|raw|${name}${name.endsWith('s') ? "'" : "'s"} avatar was successfully set. Avatar:<br /><img src="${avatarUrl}" width="80" height="80">);
```

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+exports.commands = {
+	customavatar: {
+		add(target, room, user) {
+			if (!this.can('avatar')) return false;
+			let parts = target.split(',').map(param => param.trim());
+			if (parts.length < 2) return this.parse('/help customavatar');
+
+			let name = toID(parts[0]);
+			let avatarUrl = parts[1];
+			if (!/^https?:\/\//i.test(avatarUrl)) avatarUrl = 'http://' + avatarUrl;
```
whalemer: URLs don't necessarily end in the extension of the actual file they point at

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+			let ext = path.extname(avatarUrl);
+
+			if (!VALID_EXTENSIONS.includes(ext)) {
+				return this.errorReply("Image url must end in a .png extension.");
+			}
+
+			Config.customavatars[name] = name + ext;
+
+			downloadImage(avatarUrl, name, ext);
+			this.sendReply(`|raw|${name}${name.endsWith('s') ? "'" : "'s"} avatar was successfully set. Avatar:<br /><img src="${avatarUrl}" width="80" height="80">);
```
TheJetOU:
```suggestion
			this.sendReply(`|raw|${name}${name.endsWith('s') ? "'" : "'s"} avatar was successfully set. Avatar:<br /><img src="${avatarUrl}" width="80" height="80">`);
```

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
+// Original plugin from https://github.com/CreaturePhil/Showdown-Boilerplate/blob/master/chat-plugins/customavatar.js.
+// Credits to CreaturePhil and the other listed contributors.
+// updated for the main server by Maxalexanderpi and Hoeenhero.
+*/
+'use strict';
+
+/** @type {typeof import('../../lib/fs').FS} */
```
TheJetOU: This is in the wrong place, it should be above the importing of 'FS'.

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+			Config.customavatars[name] = name + ext;
+
+			downloadImage(avatarUrl, name, ext);
+			this.sendReply(`|raw|${name}${name.endsWith('s') ? "'" : "'s"} avatar was successfully set. Avatar:<br /><img src="${avatarUrl}" width="80" height="80">);
+			Monitor.adminlog(name + "'s avatar was successfully set by " + user.name + ".");
+			if (Users(name)) Users(name).popup("|html|Upper staff have set your custom avatar.<br /><img src='" + avatarUrl + "' width='80' height='80'><br /> Refresh your page if you don't see it.");
+		},
+
+		remove(target, room, user) {
+			if (!this.can('')) return false;
```
TheJetOU: Why is this empty?

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+		console.error(e);
+	});
+	req.end();
+}
+function loadCustomAvatars() {
+	fs.readdir(AVATAR_PATH, (err, files) => {
+		if (err) console.log("Error loading custom avatars: " + err);
+		if (!files) files = [];
+		files
+			.filter(file => VALID_EXTENSIONS.includes(path.extname(file)))
```
Slayer95: Should probably blacklist ``__proto__`` and ``constructor``

server/chat-plugins/customavatars.js
Slayer95: Error handling and logging seems to be quite badly implemented.

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+const fs = require('fs');
+const https = require('https');
+
+const AVATAR_PATH = 'config/avatars/';
+
+const VALID_EXTENSIONS = ['.png'];
+
+function downloadImage(image_url, name, extension) {
+	let req = https.get(image_url, res => {
+		 res.setEncoding('binary');
```
Slayer95: No need to ``setEncoding()``. It actually hurts performance by converting the *binary* data to a ``latin-1``-encoded string.

Slayer95:
This implementation uses ``node-static``, so the comments should be disabled for ``Config.disablenodestatic == true``

server/chat-plugins/customavatars.js
```diff
@@ -0,0 +1,102 @@
...
+		 res.on('response', response => {
+			 if (response.statusCode !== 200) return;
+			 let type = response.headers['content-type'].split('/');
+			 if (type[0] !== 'image') return;
+
+			response.pipe(fs(AVATAR_PATH + name + extension).createWriteStream());
```
Slayer95: > should probably be returning a promise

Resolve the promise on ``finish`` event of the fs write stream.

Zarel:
> Ultimately, @Zarel is going to need to weigh in on whether the feature is desirable in the first place, as he's currently the only one uploading avatars to the best of my knowledge, and I believe that he'd probably want to have every avatar run by him anyway to be able to maintain quality standards and visual cohesiveness, so I don't know how much of an impact this would be on the custom avatar process because Zarel's attention would be required either way.

This is purely for side servers, which are allowed to set their own custom avatars without oversight. The main server already uses a completely different mechanism for custom avatars, mainly because we have access to more durable avatar storage (which lets avatars be visible in replays).

For side servers, yes, I'm completely OK with this.

Zarel:
This looks abandoned. Feel free to reopen if you're still planning to work on this, but otherwise I've moved it to #2444.

PR #5519, opened 2019-06-03
Include base power in |request| side data for Frustration/Return

scheibo:
Context: https://www.smogon.com/forums/threads/return-frustration-base-power-tooltip.3650730/

### DIFF
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -184,6 +184,12 @@ export class Side {
 					if (move === 'hiddenpower') {
 						return move + toID(pokemon.hpType) + (pokemon.hpPower === 70 ? '' : pokemon.hpPower);
 					}
+					if (move === 'frustration' || move === 'return') {
+						const m = this.battle.getMove(move)!;
+						// @ts-ignore - Frustration and Return only require the source Pokemon
+						const basePower = m.basePowerCallback(pokemon);
+						return `${m.name} ${basePower}`;
+					}
 					return move;
 				}),
 				baseAbility: pokemon.baseAbility,


sim/side.ts
```diff
@@ -190,6 +190,7 @@ export class Side {
 				item: pokemon.item,
 				pokeball: pokemon.pokeball,
 			};
+			if (pokemon.moves.some(move => ['frustration', 'return'].includes(move))) entry.happiness = pokemon.happiness;
```
```suggestion
			if (pokemon.moves.some(move => ['frustration', 'return'].includes(move))) {
				entry.happiness = pokemon.happiness;
			}
```

Zarel: I'm not _completely_ sure how I feel about this... It probably makes more sense to send the base power the way Hidden Power is, above.

Zarel: In fact, that's how Return/Frustration work in the move menu, too.

scheibo: Good point! I was unsure about this too, which is why I opened a PR for a one line change

Zarel:
Also, not a request for this specific PR (since apparently that's ambiguous whenever I say these things), but I'm going to register a plan to migrate `|request|` data to use names instead of IDs, to make it easier for server-side-only mods to have custom moves/etc that show up correctly.

scheibo:
>  I'm going to register a plan to migrate |request| data to use names instead of IDs, to make it easier for server-side-only mods to have custom moves/etc that show up correctly.

The `TODOs` in the client already indicated as much. How does it make things easier mods exactly? They'd still need to edit the data files, which then gets built into the client on the server, no?

sim/side.ts
```diff
@@ -184,6 +184,12 @@ export class Side {
 					if (move === 'hiddenpower') {
 						return move + toID(pokemon.hpType) + (pokemon.hpPower === 70 ? '' : pokemon.hpPower);
 					}
+					if (move === 'frustration' || move === 'return') {
```
scheibo: Possibly only want to include when not 102, but the code in `sim/pokemon.ts` includes it always so I figured I might as well match

Zarel:
> The `TODOs` in the client already indicated as much. How does it make things easier mods exactly? They'd still need to edit the data files, which then gets built into the client on the server, no?

I'm thinking about things like Staff Bros, which don't modify any client data files.

Although no one's reporting any issues with Staff Bros so far so either no one cares that the custom moves are sent as IDs, or we already have some other system to make them appear correctly?

sim/side.ts
Zarel: It makes sense to include when it _is_ 102. "Known to be 102" and "Unknown base power" would otherwise be ambiguous.

scheibo:
> or we already have some other system to make them appear correctly?

When we `build` in the client it copies the data files from the server already, so when we do `Dex.getMove` on the client we get the move name from the ID.

I think the current model where we pass around IDs as much as possible and only `Dex.getMove` on display is preferable - it doesn't *really* matter (though ID is going to be marginally less bandwidth), but switching just breaks a lot of things/causes a lot of migration work. One argument I can see in favor of name > ID is that downstream developers then dont need to munge or clone PS data files to get display versions of things (its a lot easily to implement `toID` in another language than it is to build up the data display tables required), but I'd prefer solving that rethinking PS's data to be easily ingestable by other languages (something I have a forthcoming design for).

**tl;dr:** not strongly opposed to the idea of using names instead of IDs, but think its probably less impactful than would justify the work involved

Zarel:
> When we `build` in the client it copies the data files from the server already, so when we do `Dex.getMove` on the client we get the move name from the ID.

It copies the data files from `data/`; it shouldn't touch mods except to build teambuilder tables, which explicitly don't include names.

PR #5522, opened 2019-06-04
Datasearch: Add ability to search for moves not in a Pokemon's moveset

jmclemo6:
Main use of this feature would be to compare the movesets of two Pokemon. For example, a user comes into Trainer Academy and asks why to use X mon over Z. Now room staff could show them the kinds of moves that X does get that Z doesn't.

I feel like this is also a readability improvement over the previous code.

### DIFF
--- a/server/chat-plugins/datasearch.js
+++ b/server/chat-plugins/datasearch.js
@@ -1084,10 +1084,7 @@ function runMovesearch(target, cmd, canAll, message) {
 
 	const getFullLearnsetOfPokemon = (template) => {
 		if (!template.learnset) template = Dex.getTemplate(template.baseSpecies);
-		let lsetData = new Set();
-		for (const move in template.learnset) {
-			lsetData.add(move);
-		}
+		const lsetData = new Set(Object.keys(template.learnset));
 
 		while (template.prevo) {
 			template = Dex.getTemplate(template.prevo);
@@ -1101,10 +1098,7 @@ function runMovesearch(target, cmd, canAll, message) {
 
 	// Since we assume we have no target mons at first
 	// then the valid moveset we can search is the set of all moves.
-	let validMoves = new Set();
-	for (const move in Dex.data.Movedex) {
-		validMoves.add(move);
-	}
+	const validMoves = new Set(Object.keys(Dex.data.Movedex));
 	validMoves.delete('magikarpsrevenge');
 	for (const mon of targetMons) {
 		const template = Dex.getTemplate(mon.name);
--- a/server/chat-plugins/datasearch.js
+++ b/server/chat-plugins/datasearch.js
@@ -806,7 +806,6 @@ function runMovesearch(target, cmd, canAll, message) {
 	}
 	let showAll = false;
 	let sort = null;
-	let lsetData = {};
 	let targetMons = [];
 	let randomOutput = 0;
 	for (const arg of target.split(',')) {
@@ -926,19 +925,10 @@ function runMovesearch(target, cmd, canAll, message) {
 
 			let template = Dex.getTemplate(target);
 			if (template.exists) {
-				if (targetMons.includes(template.name)) return {reply: "A search cannot include the same Pok\u00e9mon twice."};
 				if (parameters.length > 1) return {reply: "A Pok\u00e9mon learnset cannot have alternative parameters."};
-				if (!template.learnset) template = Dex.getTemplate(template.baseSpecies);
-				lsetData[template.name] = Object.assign({}, template.learnset);
-				targetMons.push(template.name);
-
-				let originalTemplateName = template.name;
-				while (template.prevo) {
-					template = Dex.getTemplate(template.prevo);
-					for (let move in template.learnset) {
-						if (!lsetData[originalTemplateName][move]) lsetData[originalTemplateName][move] = template.learnset[move];
-					}
-				}
+				if (targetMons.some(mon => mon.name === template.name && isNotSearch !== mon.shouldBeExcluded)) return {reply: "A search cannot both exclude and include the same Pok\u00e9mon."};
+				if (targetMons.some(mon => mon.name === template.name)) return {reply: "A search should not include a Pok\u00e9mon twice."};
+				targetMons.push({name: template.name, shouldBeExcluded: isNotSearch});
 				orGroup.skip = true;
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


server/chat-plugins/datasearch.js
```diff
@@ -1091,36 +1080,53 @@ function runMovesearch(target, cmd, canAll, message) {
...
-			if (intersectionOfLearnsets.length) {
-				intersectionOfLearnsets = intersectionOfLearnsets.filter(move => {
-					return lsetData[mon][move] !== undefined;
-				});
-			} else {
-				intersectionOfLearnsets = Object.keys(lsetData[mon]);
-			}
-		}
+	const dex = {};
+	for (let move in Dex.data.Movedex) {
```
scheibo: Wow, this seems super inefficient. Pretty confident you should only need the `Move` at the very end for display and should just be operating on `IDs` until then. You're creating 300+ objects with 20-30+ fields here.

server/chat-plugins/datasearch.js
```diff
@@ -925,20 +924,10 @@ function runMovesearch(target, cmd, canAll, message) {
...
-
-				let originalTemplateName = template.name;
-				while (template.prevo) {
-					template = Dex.getTemplate(template.prevo);
-					for (let move in template.learnset) {
-						if (!lsetData[originalTemplateName][move]) lsetData[originalTemplateName][move] = template.learnset[move];
-					}
-				}
-				orGroup.skip = true;
+				if (targetMons.some(mon => mon.name === template.name && isNotSearch !== mon.shouldBeExcluded)) return {reply: "A search cannot both exclude and include the same Pok\u00e9mon."};
```
scheibo: How large is `targetMons` here? Doing basically the same iteration over the list twice is kind of ugh. Isn't this just:
```ts
for (const mon of targetMons) {
  if (mon.name === template.name) {
    if (isNotSearch !== mon.shouldBeExcluded) return ...
    return ...
  }
}
```

server/chat-plugins/datasearch.js
```diff
@@ -1091,36 +1080,53 @@ function runMovesearch(target, cmd, canAll, message) {
...
+			lsetData = new Set([...lsetData, ...Object.keys(template.learnset)]);
 		}
 
-		for (let move in finalizedLearnset) {
-			dex[move] = Dex.getMove(move);
+		return lsetData;
+	};
+
+	// Since we assume we have no target mons at first
+	// then the valid moveset we can search is the entire dex.
```
scheibo: 'entire dex' confused me: entire movedex / set of all moves / etc would read better I think?

server/chat-plugins/datasearch.js
Zarel: I think most of Dexsearch is super inefficient. :( My dream is one day to be able to unify the client's teambuilder search and server dexsearch code.

(Actually partially underway! My latest refactor soon-to-be-pushed splits client Dexsearch code away from its UI code.)

server/chat-plugins/datasearch.js
jmclemo6: targetMons is only as big as the number of pokemon that the user enters, so practically 5-6 at most I would say.

jmclemo6:
Don't merge. Gonna modify to use only IDs until necessary.

server/chat-plugins/datasearch.js
jmclemo6: Changed to "set of all moves".

server/chat-plugins/datasearch.js
jmclemo6: Changed to work with IDs until the full data of the moves are needed.

jmclemo6:
Alright, uses only IDs until necessary now.

scheibo:
This is looking a lot better to me. My one remaining complaint is the `Set` operations - I know `/datasearch` is already a performance mess, but can we replace the cutesy one liners with `for` loops which are much more efficient? One example:

```ts
// lsetData = new Set([...lsetData, ...Object.keys(template.learnset)]);
for (const move of Object.keys(template.learnset)) {
  lsetData.add(move);
}
```

Spread syntax and all of the implicit array creation just ends up creating a lot of object churn IME.

Zarel:
`Object.keys` isn't implicit array creation? Maybe `for (const move in template.learnset)`?

scheibo:
True, it also is! I even missed one! Where's my performance dunce hat? :D

jmclemo6:
> This is looking a lot better to me. My one remaining complaint is the `Set` operations - I know `/datasearch` is already a performance mess, but can we replace the cutesy one liners with `for` loops which are much more efficient? One example:
> 
> ```ts
> // lsetData = new Set([...lsetData, ...Object.keys(template.learnset)]);
> for (const move of Object.keys(template.learnset)) {
>   lsetData.add(move);
> }
> ```
> 
> Spread syntax and all of the implicit array creation just ends up creating a lot of object churn IME.

Yeah, I can replace the one liners. I had benchmarked them beforehand, but made a call of readability (imo) over perf.

jmclemo6:
Got rid of all the implicit array creations..

server/chat-plugins/datasearch.js
```diff
@@ -1092,36 +1082,58 @@ function runMovesearch(target, cmd, canAll, message) {
...
-		}
-	} else {
-		for (let move in Dex.data.Movedex) {
-			dex[move] = Dex.getMove(move);
+		return lsetData;
+	};
+
+	// Since we assume we have no target mons at first
+	// then the valid moveset we can search is the set of all moves.
+	let validMoves = new Set();
```
scheibo: So, I probably wasn't clear, I apologize - I shouldn't have cited "implicit array creation". 

For initialization (like here and elsewhere below), something like `new Set(Object.keys(Dex.data.Movedex))` should be both more clear and pretty much as performant (yes, you create an array for `Object.keys`, but `Set` can theoretically use the `.length` hit to prealloc space better. Its also still `N` ops performance wise).

My complaint is with using the `set = new Set([...set, ...adds])` pattern for addition, especially when `set` is already large - you're re-adding all of `set`, making the operation `N + M` (which `N` = set.size and `M` = adds.size) when it should just be `M`.

scheibo:
```diff
diff --git a/server/chat-plugins/datasearch.js b/server/chat-plugins/datasearch.js
index 2af80d785..aec264509 100644
--- a/server/chat-plugins/datasearch.js
+++ b/server/chat-plugins/datasearch.js
@@ -1084,10 +1084,7 @@ function runMovesearch(target, cmd, canAll, message) {
 
        const getFullLearnsetOfPokemon = (template) => {
                if (!template.learnset) template = Dex.getTemplate(template.baseSpecies);
-         let lsetData = new Set();
-         for (const move in template.learnset) {
-                 lsetData.add(move);
-         }
+               const lsetData = new Set(Object.keys(template.learnset));
 
                while (template.prevo) {
                        template = Dex.getTemplate(template.prevo);
@@ -1101,10 +1098,7 @@ function runMovesearch(target, cmd, canAll, message) {
 
        // Since we assume we have no target mons at first
        // then the valid moveset we can search is the set of all moves.
- let validMoves = new Set();
- for (const move in Dex.data.Movedex) {
-         validMoves.add(move);
- }
+       const validMoves = new Set(Object.keys(Dex.data.Movedex));
        validMoves.delete('magikarpsrevenge');
        for (const mon of targetMons) {
                const template = Dex.getTemplate(mon.name);
```

For some reason I get:

```
error: src refspec movesearch-not-feature-implementation does not match any
error: failed to push some refs to 'git@github.com:jmclemo6/Pokemon-Showdown'
```

When trying to push to the branch.

PR #5523, opened 2019-06-04
Random Battle: Update Heliolisk

ascriptmaster:
- Change the Thunder/Thunderbolt rejection condition to a moveset
modification to improve consistency.
- Change the Rain Dance rejection condition on non-Water mons to
have a Water move.
- Give Heliolisk Waterium Z if it has Rain Dance + Surf.

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -7174,7 +7174,7 @@ let BattleFormatsData = {
 		tier: "LC",
 	},
 	heliolisk: {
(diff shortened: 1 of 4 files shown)


TheImmortal:
Hmm, the first changes are fine. But I'm not sure Waterium is a good idea. Life Orb is good.

ascriptmaster:
@TheImmortal I reverted the Waterium-Z change

PR #5526, opened 2019-06-05
Don't share self-replacing abilities

urkerab:
Reported [here](https://www.smogon.com/forums/posts/8066699/) (and with a couple of followup posts).

When these abilities replace themselves with another Ability, that will then get shared as normal.

I've implemented this by not cloning the effects for those Abilities (resulting in dummy no-op effects instead) so that the code dealing with adding and removing shared abilities is unaffected.

### DIFF
--- a/data/mods/pic/scripts.js
+++ b/data/mods/pic/scripts.js
@@ -7,7 +7,7 @@ exports.BattleScripts = {
 			return name;
 		}
 		let id = toID(name);
-		if (id.startsWith('ability')) return Object.assign(Object.create(this.getAbility(id.slice(7))), {id});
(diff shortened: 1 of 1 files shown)


Zarel:
I have no idea what pic is, but I'll trust you that this is a desirable fix.

data/mods/pic/scripts.js
```diff
@@ -7,7 +7,7 @@ exports.BattleScripts = {
 			return name;
 		}
 		let id = toID(name);
-		if (id.startsWith('ability')) return Object.assign(Object.create(this.getAbility(id.slice(7))), {id});
+		if (id.startsWith('ability') && !['abilitypowerofalchemy', 'abilityreceiver', 'abilitytrace'].includes(id)) return Object.assign(Object.create(this.getAbility(id.slice(7))), {id});
```
Zarel: In all honesty, I thought this was something the engine already handled.

Zarel: This is at least intended to be handled by the engine.

PR #5531, opened 2019-06-11
Support filtering by Z-move in movesearch

asgdf:


scheibo:
Thoughts on https://www.smogon.com/forums/threads/improvements-to-ds-and-ms.3651133/ while we're here?

asgdf:
Oh good point. Seems very straightforward to do.

PR #5532, opened 2019-06-11
Prevent duplicate options in polls

TheJetOU:


Zarel:
@bumbadadabum can you think of any use cases for duplicate options? I can only think of joke polls, which is probably not a good reason to prevent this.

bumbadadabum:
Nope can't think of any.

PR #3972, opened 2017-09-08
Refactor Modlog Formatting

sparkychildcharlie:
This PR's purpose has been changed.  Please refer to https://github.com/Zarel/Pokemon-Showdown/pull/3972#issuecomment-351908234 for the new purpose.

For the original purpose please refer to the branched PR https://github.com/Zarel/Pokemon-Showdown/pull/4254

PR #5535, opened 2019-06-11
Allow for viewing only punishment-related actions in the modlog

scheibo:
**Context:** [An option to split modlog into punishments and non punishments](https://www.smogon.com/forums/threads/an-option-to-split-modlog-into-punishments-and-non-punishments.3651100/) (3rd most popular)

- I'm not wedded to the idea of a `'/punishlog'` command vs. implementing some sort of flag (`'scope=punishments'`?) the way `maxLines` works, but this was the easiest approach and its relatively simple to change
- I may be missing or including actions which are or are not 'punishment related'. Something like `'NOTE'` is not necessarily 'punishment-related', but it often provides additional context for punishments so I figured it was useful to include. Ditto some others I've added. On the main server we may wish to add to `PUNISHMENTS` with actions from private code, potentially meaning the list should be defined in `config/` somewhere instead of here.
- Taking a step back - would it instead make sense to have the non-punishment actions show up in different log files entirely? The approach I've taken here relies on the existing design, but I'm not positive trivia actions belong in the same place as global bans to begin with, even if both require auth.

### DIFF
--- a/server/chat-plugins/modlog.js
+++ b/server/chat-plugins/modlog.js
@@ -31,6 +31,16 @@ const LINES_SEPARATOR = 'lines=';
 const MAX_RESULTS_LENGTH = MORE_BUTTON_INCREMENTS[MORE_BUTTON_INCREMENTS.length - 1];
 const LOG_PATH = 'logs/modlog/';
 
+const PUNISHMENTS = [
+	'ROOMBAN', 'UNROOMBAN', 'WARN', 'MUTE', 'HOURMUTE', 'UNMUTE', 'CRISISDEMOTE',
+	'WEEKLOCK', 'LOCK', 'UNLOCK', 'UNLOCKNAME', 'UNLOCKRANGE', 'UNLOCKIP', 'BAN',
+	'UNBAN', 'UNBANALL', 'DEROOMVOICEALL', 'RANGEBAN', 'UNRANGEBAN', 'RANGELOCK',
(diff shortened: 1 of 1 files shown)


scheibo:
@panpawn has brought up the fact that this depends on modlogs being uniformly formatted, @Zarel can you confirm that you've run @sparkychildcharlie's [script](https://pastebin.com/0UfyDmVu) from https://github.com/Zarel/Pokemon-Showdown/pull/3972?

sparkychildcharlie:
pretty sure it hasnt been run

Zarel:
I'm worried about how it'll work while the server is actively running...

AustinXII:
The script is missing battleban btw. Also perhaps try it on smogtours first?

sparkychildcharlie:
Im pretty sure at the time i wrote it battleban didnt exist on the server... sooo it wouldnt been in sentence form

AustinXII:
@sparkychildcharlie yeah you did, I was just pointing it out incase zarel were to run it right now since no one has mentioned it to him

panpawn:
Shouldn't matter, since the script's purpose is to only convert entries before the standard formatting was changed, and it only targets ones that need to be converted

scheibo:
@Zarel - are you OK with this approach? I think this change can still land without the script having been run, it just will be more accurate/effective once that is run as well.

Zarel:
It's fine I guess.

PR #5539, opened 2019-06-14
Fix input logs

HoeenCoder:
I wanted to also add tests for this since this is the 3rd/4th time I've fixed input logs, though I'm having issues with sockets crashing in the tests and haven't been able to find morfent to talk with them about whats going on. I'll add tests in a different commit, but I don't want to delay the fix any longer for tests.

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -434,6 +434,8 @@ class RoomBattle extends RoomGames.RoomGame {
 		 * @type {number}
 		 */
 		this.rated = options.rated || 0;
+		// true when onCreateBattleRoom has been called
+		this.missingBattleStartMessage = !!options.inputLog;
 		this.started = false;
 		this.ended = false;
 		this.active = false;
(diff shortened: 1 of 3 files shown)


server/room-battle.js
```diff
@@ -434,6 +434,8 @@ class RoomBattle extends RoomGames.RoomGame {
 		 * @type {number}
 		 */
 		this.rated = options.rated || 0;
+		// true when players have taken all open slots in an imported battle
+		this.missingPlayers = !!options.inputLog;
```
Zarel: Wait, why is this necessary?

Zarel: is this different from `!this.players.every(player => player.userid)`

HoeenCoder: Yes, after players are restored in an input log, we need to display the battle start message that is called in `Rooms.global.onCreateBattleRoom`. This flag is so that we only display that message once when a player is added.

This could still be simplified if `/addplayer` isn't supposed to support replacing players.

Zarel: This should probably get a better name, then. `.missingBattleStartMessage`?

HoeenCoder:
Bump @Zarel

PR #5543, opened 2019-06-16
Add juniper avatar

TapoCollino:
hopefully

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -172,6 +172,7 @@ const avatarTable = new Set([
 	'juan-gen3',
 	'juan',
 	'juggler-gen1', 'juggler-gen1rb', 'juggler-gen2', 'juggler-gen3', 'juggler',
+	'juniper',
 	'jupiter',
 	'karen-gen2', 'karen',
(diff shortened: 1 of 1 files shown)


Zarel:
Juniper was uploaded as an avatar to be specific to Help staff. I don't know how they'd feel about making it a generally usable avatar.

TapoCollino:
Oh didn't know about that. Someone brought it up and I thought you guys just forgot to add it.

scheibo:
> Juniper was uploaded as an avatar to be specific to Help staff

FWIW, Help Staff say they can't access it either - did you actually upload it?

Zarel:
> FWIW, Help Staff say they can't access it either - did you actually upload it?

I did, I even coded up an ACL for avatars, just for them, and then they wanted an ACL that automatically updated as help staff got promoted/demoted :| and I couldn't find time to do that.

scheibo:
Perhaps worth just going with the manual list for now as opposed to blocking indefinitely on the 'best' solution?

Zarel:
Good point. Done.

PR #5546, opened 2019-06-19
Allow roomauth to /hidetext even when there's no user object

scheibo:


Zarel:
Let it be known that I don't actually _like_ this, and part of me wants to make this command also set the room to hidden since they clearly don't have the auth to keep the room moderated.

I don't think it's a high expectation for a room to either have someone check in once an hour, or turn on modchat for the hours they don't have any staff. i.e. "The last person to leave should lock the door".

Maybe the correct solution is to just do it automatically? If the room doesn't have staff, modchat + automatically turns on?

AustinXII:
That would be nice. Maybe even check if there are staff that have been afk for a while since we recently got the afk timer ;)

scheibo:
> Maybe the correct solution is to just do it automatically? If the room doesn't have staff, modchat + automatically turns on?

I suggested this in the discussion on Discord, would be more than willing to implement. I think there needs to be a timer component (at least allow like a 5 minute grace or something before turning it on)?

PR #5547, opened 2019-06-20
Always inflict external staleness from Heal Pulse.

scheibo:
Currently the specification claims this to be the case, when the code actually only sets external staleness if HP is healed.

Context: https://replay.pokemonshowdown.com/gen7ou-931202355

It's not entirely clear whether we want this fix as opposed to simply changing the spec to specify that HP must be healed. The troll Funbro vs. Funbro battle exhibited in the replay only ends before 1000 turns if they both use Heal Pulse.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -5605,7 +5605,7 @@ let BattleMovedex = {
 			} else {
 				success = !!this.heal(Math.ceil(target.maxhp * 0.5));
 			}
-			if (success && target.side.id !== source.side.id) {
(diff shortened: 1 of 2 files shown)


scheibo:
I've decided to just [change one word of the spec](https://github.com/Zarel/Pokemon-Showdown-Dex/commit/bb0d4e4fb712eb9e34277dd75ccb490ef2ff8523) because I don't think this edge case is really valid anyway:

PR #5548, opened 2019-06-20
Make help ticket ignoring persistent

Kaiepi:


Zarel:
Making them temporary was a compromise, I don't want to change the deal at the moment.

scheibo:
What is 'the deal'? Having to ignore tickets several times a day if pretty obnoxious. I understand we want to encourage policy staff to accept the tickets, but I don't know that nagging others (the 'stick' approach) is the best way to achieve this. Adding metrics around tickets (@HoeenCoder) and rewarding/acknowledging those that are doing the work (the 'carrot' approach) seems like a better solution?

Zarel:
In general, the UI for tickets needs to be made nicer, so people will accept them more.

The problem is that the more people _don't_ take tickets, the more there'll be for others.

Tickets are, for the most part, better than the old approach where everyone would just get random PMs from users. So to the extent that people don't take tickets, the ticket UI should probably be changed to be more and more similar to "get random PMs from users".

AustinXII:
Adding on to this, being a staff roomvoice and having to constantly see helptickets that you can’t do anything about (or even help with) is very annoying

Zarel:
Helptickets shouldn't be shown to staff roomvoices in the first place, tbh.

Zarel:
I thought that was already implemented, actually. If not, it needs to be.

AustinXII:
> Helptickets shouldn't be shown to staff roomvoices in the first place, tbh.

Sounds good to me! :)

scheibo:
Hello @Zarel, can we revisit this, with one change: we also add a list to `Config` to restrict who is allowed to use it, and make the list include all tech gstaff? Tech gstaff are not expected to do policy, and it doesnt make sense that they should be forced to `/ht ignore` every time they log in. Ive chatted with @Lusamine  about it and she agrees, but I'd be more that willing to suggest this on the forums if you feel this change would be controversial in any way.

AustinXII:
I don’t think restricting tech gstaff from help tickets entirely. If that is added I think there should be an option to at least allow some tech gstaff to turn them on. Iirc some tech staff do help tickets

scheibo:
We have stats on this, and tech staff really doesnt do help tickets, but also:

> to restrict who is **allowed** to use it

Emphasis on allowed. Allow the tech gstaff to opt out. Dont opt them out by default, dont make them never be able to opt in again, just let them opt out permanently and not force them to click a button or run a command every session

Zarel:
I thought we already did this with a hardcoded list?

scheibo:
Well, I've done it for myself and chaos in private code, but thats not very flexible or fair to non-admin gstaff. I would like for us to give all technical gstaff the ability to perma opt-out (and then opt back in themselves without needing to bug someone with access to private code to change it for them)

Zarel:
I suppose I'm fine with that.

AnnikaCodes:
I fixed this last fall.

PR #5550, opened 2019-06-21
Update link in !suggestions

maxalexandderpi:
link was still to the old thread.


server/chat-plugins/info.js
```diff
@@ -4,7 +4,6 @@
  *
  * These are informational commands. For instance, you can define the command
  * 'whois' here, then use it by typing /whois into Pokemon Showdown.
- *
```
panpawn: Why was this removed o.o

PR #5551, opened 2019-06-21
Add a shorter alias for /offlinewhois

maxalexandderpi:
both /offlinewhois and /checkpunishment are rather inconvenient to type over and over - and I've been hearing this echoed a lot, so this couldn't hurt.

### DIFF
--- a/server/chat-plugins/info.js
+++ b/server/chat-plugins/info.js
@@ -231,6 +231,7 @@ const commands = {
 
 	'!offlinewhois': true,
 	checkpunishment: 'offlinewhois',
+	cpm: 'offlinewhois',
 	offlinewhois(target, room, user) {
 		if (!user.trusted) {
 			return this.errorReply("/offlinewhois - Access denied.");


scheibo:
`/cpm` seems kind of cryptic, but I don't necessarily have a better naming suggestion

Stupid question, why does `/offlinewhois` even exist - why can't its functionality be collapsed into `/whois` to begin with? https://github.com/Zarel/Pokemon-Showdown/commit/22c6342c0077dea91f1a6c0ca2fc38380059dff1 seems to imply its actually called by `/whois` anyway?

The existing `/alts` or `/ip` commands seem to accomplish the same thing as `/offlinewhois` already? Or does this only happen to be the case because I'm `&`?

AustinXII:
> `/cpm` seems kind of cryptic, but I don't necessarily have a better naming suggestion
> 
> Stupid question, why does `/offlinewhois` even exist - why can't its functionality be collapsed into `/whois` to begin with? [22c6342](https://github.com/Zarel/Pokemon-Showdown/commit/22c6342c0077dea91f1a6c0ca2fc38380059dff1) seems to imply its actually called by `/whois` anyway?
> 
> The existing `/alts` or `/ip` commands seem to accomplish the same thing as `/offlinewhois` already? Or does this only happen to be the case because I'm `&`?

Unless it’s been changed since I was staff, alts and ip does accomplish this for lower global staffers

Zarel:
`/ip` is the standard short name for `/offlinewhois`. The underlying command name is intentionally long because there's little need to use it directly.

PR #5552, opened 2019-06-22
Allow for namelocking without their exact current name

whalemer:


server/chat-commands.js
```diff
@@ -2582,10 +2582,11 @@ const commands = {
 	],
 
 	nl: 'namelock',
-	namelock(target, room, user) {
+	forcenamelock: 'namelock',
+	namelock(target, room, user, connection, cmd) {
 		if (!target) return this.parse('/help namelock');
 
-		let reason = this.splitTarget(target, true);
+		let reason = this.splitTarget(target, cmd !== 'forcenamelock');
```
Zarel: Could you make a follow-up to actually send a reminder to use `/forcenamelock` if someone's renamed?

Zarel:
Apparently I forgot to merge this. I said "could you make a follow-up" because I was going to merge it as-is.

Zarel:
Anyway, thanks! This is exactly what I was looking for!

scheibo:
> Apparently I forgot to merge this. I said "could you make a follow-up" because I was going to merge it as-is.

This is what I figured.

PR #5555, opened 2019-06-23
Change dexsearchhelp

anAwsomePerson:
I want to make it clear that "/ds nfe" doesn't find all not-fully-evolved Pokemon, only those that aren't in another tier, like NU or LC.


Zarel:
Has this been tested?

Unrelatedly, is this a bug in the other direction? The previous documentation certainly implies so. I vaguely remember someone intentionally changed it to search all NFE Pokémon and not just NFE-tier Pokémon.

PR #5556, opened 2019-06-25
Ensure random team generation always ends up with 6 Pokemon

scheibo:
Shoehorning monotype generation into normal team generation results in several issues, especially for mono-ice where there are only 28 possible Pokemon in the `pokemonPool` to begin with:

- Rotom-F, Ninetales-A and Sandslash-A are all going to be unfairly downsampled even though their alternate formes are not legal options on mono ice (Kyurem downsampling should be OK, because all are ice)
- We limit to 2 Pokemon per tier even though most Ice types are likely to be clustered in the lower tiers (or just Untiered) because theyre kind of trash
- We limit the number of type combinations even though a large fraction of Ice types are actually just pure Ice. We increase the 'allowable number of type combos' from 1 to 2 for monotype, but this is probably still too low for Ice.
- We only consider a single set for a Pokemon and dispose of the entire Pokemon if the set doesn't fit instead of rerolling the set

I'm merely documenting these issues here, but instead going with the simplest possible fix for having < 6 members on the team which is to just throw in another loop (with an `EXHAUSTION` constant to prevent infinite iteration). I'm also utilizing our `fulltest` tagging better - even with 1000 iterations it still only takes 3s on my notebook which is still substantially less than some of our slow tests which take 1 minute on our machine. I've also preseeded to make the first team generated one with only 5 mons before this change.

### DIFF
--- a/data/random-teams.js
+++ b/data/random-teams.js
@@ -1690,19 +1690,18 @@ class RandomTeams extends Dex.ModdedDex {
 		};
 	}
 
-	randomTeam() {
-		let pokemon = [];
-
+	/**
+	 * @param {string} type
+	 * @param {RandomTeamsTypes.RandomSet[]} pokemon
+	 * @param {boolean=} isMonotype
+	 */
+	getPokemonPool(type, pokemon = [], isMonotype = false) {
 		const allowedNFE = ['Chansey', 'Doublade', 'Gligar', 'Pikachu', 'Porygon2', 'Scyther', 'Type: Null'];
-
-		// For Monotype
-		let isMonotype = this.format.id === 'gen7monotyperandombattle';
-		let typePool = Object.keys(this.data.TypeChart);
-		let type = this.sample(typePool);
-
-		let pokemonPool = [];
+		const exclude = pokemon.map(p => toID(p.species));
+		const pokemonPool = [];
 		for (let id in this.data.FormatsData) {
 			let template = this.getTemplate(id);
+			if (exclude.includes(template.id)) continue;
 			if (isMonotype) {
 				let types = template.types;
 				if (template.battleOnly) types = this.getTemplate(template.baseSpecies).types;
@@ -1712,6 +1711,16 @@ class RandomTeams extends Dex.ModdedDex {
 				pokemonPool.push(id);
 			}
 		}
+		return pokemonPool;
+	}
+
+	randomTeam() {
+		let pokemon = [];
+
+		// For Monotype
+		let isMonotype = this.format.id === 'gen7monotyperandombattle';
+		let typePool = Object.keys(this.data.TypeChart);
+		let type = this.sample(typePool);
 
 		// PotD stuff
 		let potd;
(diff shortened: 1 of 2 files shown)


Zarel:
In theory, you should just need two loops, with the second one relaxing the "2 pokemon per tier" and type combo restrictions. Unless the randbat algorithm has massively changed since I originally wrote it, nothing else should be necessary.

Zarel:
I'm having trouble understanding how the "2 pokemon per tier" and "2 allowed type comboes" can fix themselves with multiple passes, though?

TheImmortal:
I mean the simplest fix is removing the tier restriction for Monotype. Monotype doesn’t really follow standard tiers anyway; something in RU could be OU in Monotype.

scheibo:
> I'm merely documenting these issues here...

Was meant to be a hint that this was just a strawman proposal, I'm glad its garnered discussion.

> I'm having trouble understanding how the "2 pokemon per tier" and "2 allowed type comboes" can fix themselves with multiple passes, though?

The multiple loops works around the 'species downsampling' and 'only evaluating on set' issues. I didn't solve 2 pokemon per tier or 2 allowed type combos, though actually one easy way to do so would just be to use the outer loop counter as a multiplier on the number per tier or per type combo.

> In theory, you should just need two loops, with the second one relaxing the "2 pokemon per tier" and type combo restrictions.

`EXHAUSTION = 2` plus the multiplier I discussed above then. I think you'll still want more loops though, I don't see why we should be removing species entirely just because we got unlucky with `randomSet`. 

> I mean the simplest fix is removing the tier restriction for Monotype. 

Completely?

Zarel:
I don't even know in what sense a set "wouldn't qualify". Clearly the randbat code has changed drastically since I originally wrote it.

Why do we have a tier limit, anyway? Clearly that just means the level offsets need to be rebalanced?

TheImmortal:
> Why do we have a tier limit, anyway?

For more balanced teams. Without a tier limit, teams with multiple shit mons were too common, and the same can be said for broken mons. Literally everyone would agree Random Battle became significantly better after the tier limit was added.

> Completely?

Not necessarily. I'll make it max 3 or something I guess for Monotype.

scheibo:
9f8e20 unforunately does not cause my test case to pass, nor does increasing the type combo limit from 2 -> 3 in the same way that tier was increased. 

I've updated this PR to gradually relax these two limits as the outer loop progresses.

scheibo:
Oh yeah, worth noting that I'm still very open to alternative solutions, I just think that they should pass the test introduced by this PR. :)

Zarel:
> For more balanced teams. Without a tier limit, teams with multiple shit mons were too common, and the same can be said for broken mons. Literally everyone would agree Random Battle became significantly better after the tier limit was added.

Maybe we should limit untiered mons specifically? For everything else, level balance should be enough.

scheibo:
_Friendly ping_ - @TheImmortal / @Zarel, how do you feel about the specific solution introduced in this PR which has an outer loop that gradually relaxed restrictions per iteration and actually solves the immediate 5-Pokemon monoice issue?

Zarel:
Yes, I can merge as-is and we can improve it later.

PR #5557, opened 2019-06-26
Remove hardcoded URLs in favor of Config.routes

scheibo:
As discussed previously, the primary motivation here is to be able to host a staging/testing environment for Pokemon Showdown's current set up. Supporting arbitrary client configurations and truly generic hosting setups is a non-goal. Starting with the server side changes here because the client changes are more involved.

---
After this change, running `git grep pokemonshowdown | grep -v "* Pokemon Showdown - http://pokemonshowdown.com/"`:

```
LICENSE:Copyright (c) 2011-2019 Guangcong Luo and other contributors http://pokemonshowdown.com/
PROTOCOL.md:> `http://play.pokemonshowdown.com/action.php?act=upkeep&challstr=CHALLSTR`
PROTOCOL.md:> Otherwise, you'll need to make an HTTP POST request to `http://play.pokemonshowdown.com/action.php`
README.md:  [1]: http://pokemonshowdown.com/
README.md:  [4]: http://pokemonshowdown.com/
README.md:- See http://pokemonshowdown.com/credits
```
Docs

```
config/config-example.js:exports.loginserver = 'http://play.pokemonshowdown.com/';
config/config-example.js:	root: 'pokemonshowdown.com',
config/config-example.js:	client: 'play.pokemonshowdown.com',
config/config-example.js:	dex: 'dex.pokemonshowdown.com',
config/config-example.js:	replays: 'replay.pokemonshowdown.com',
package.json:  "homepage": "http://pokemonshowdown.com",
```
Config

```
data/rulesets.js:		// implemented in sim/battle.js, see https://dex.pokemonshowdown.com/articles/battlerules#endlessbattleclause for the specification.
server/chat-plugins/info.js: * Pokemon Showdown - https://pokemonshowdown.com/
server/ladders-local.ts: * play.pokemonshowdown.com.
server/ladders-remote.ts: * play.pokemonshowdown.com.
```
Comments

```
server/chat-plugins/info.js:			`- <a href="https://replay.pokemonshowdown.com/gennextou-120689854">Zergo vs Mr Weegle Snarf</a><br />` +
server/chat-plugins/info.js:			`- <a href="https://replay.pokemonshowdown.com/gennextou-130756055">NickMP vs Khalogie</a>`
```
Specific replays which wouldnt exist on another server

```
server/chat.js:const LINK_WHITELIST = ['*.pokemonshowdown.com', 'psim.us', 'smogtours.psim.us', '*.smogon.com', '*.pastebin.com', '*.hastebin.com'];
server/chat.js:			'pokemonshowdown.com': 1,
```
Whitelisting

```
server/ladders-remote.ts:		return [`<tr><td><strong>Please use the official client at play.pokemonshowdown.com</strong></td></tr>`];
```
Only 'play.pokemonshowdown.com' is *official*, so this is still correct

```
server/static/404.html:<p>This is a <a href="http://www.pokemonshowdown.com/">Pok&eacute;mon Showdown</a> server.</p>
server/static/index.html:<p>This is a <a href="http://www.pokemonshowdown.com">Pok&eacute;mon Showdown</a> server!</p>
```
PITA to change without introducing a build for these files to replay the constants. Will need to do that on the client, but seems like overkill here. Also contains a reference to psim.us, once again, not sure worth what it would take to generalize

```
test/server/chat.js:			`<a href="//dex.pokemonshowdown.com/pokemon/oshawott" target="_blank"><psicon pokemon="Oshawott"/></a> &gt;w&lt;`
test/server/chat.js:			`<a href="//dex.pokemonshowdown.com/items/beastball" target="_blank">[Beast ball]</a> &gt;w&lt;`
```
Tests

```
translations/dutch.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-nl",
translations/english.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "",
translations/french.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-fr",
translations/german.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-de",
translations/hindi.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-hi",
translations/italian.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-it",
translations/japanese.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-ja",
translations/portuguese.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-pt",
translations/simplifiedchinese.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-zh",
translations/spanish.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-es",
translations/traditionalchinese.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-tw",
translations/turkish.json:		"[TN: Link to the PS rules for your language]https://pokemonshowdown.com/rules": "https://pokemonshowdown.com/pages/rules-tr",
```
Should probably also not be hardcoded, but translations are half baked anyway


config/config-example.js
```diff
@@ -136,6 +135,17 @@ Y929lRybWEiKUr+4Yw2O1W0CAwEAAQ==
...
+/**
+ * routes - where Pokemon Showdown is hosted.
+ *   Don't change this setting - there aren't any other options right now
+ */
+exports.routes = {
+	root: 'pokemonshowdown.com',
+	client: 'play.pokemonshowdown.com',
+	dex: 'dex.pokemonshowdown.com',
+	replays: 'replay.pokemonshowdown.com',
+};
```
Slayer95: I'd suggest ``app: 'sim.smogon.com',`` for the simulator server.

Slayer95: Also ``logs: 'logs2.psim.us:8080,'``.

scheibo: Neither of these are actually used anywhere in either this or the client repo? Also, actually hosting a logs server is also pretty esoteric currently (should https://github.com/Zarel/Pokemon-Showdown-Client/tree/log-viewer be merged into HEAD instead of being an orphan branch?). I can see adding those two for 'completeness' and so that if anyone *does* want to link to them from the repo in the future its more obvious that they should instead be using `${Config.routes.logs}` for example, but I'm also not sure its not 'YAGNI'?

Zarel: I would suggest staging server URLs here, instead of the official URLs.

server/chat-formatter.ts
```diff
@@ -307,7 +307,7 @@ class TextFormatter {
 						let dir = key;
 						if (key === 'item') dir += 's';
 
-						uri = `//dex.pokemonshowdown.com/${dir}/${toID(term)}`;
+						uri = `//${Config.routes.dex}/${dir}/${toID(term)}`;
```
Zarel: This is used by client, and can't really have external dependencies.

config/config-example.js
scheibo: > I would suggest staging server URLs here, instead of the official URLs.

Can you elaborate here? Are you talking about just the proposed `logs` and `app` routes? Things will break currently if `Config.routes` doesn't point to eg. the correct official `client` or `root` URLs otherwise?

Zarel: I'm pretty sure logs only appear in private code (and the config file?) anyway.

Zarel: > Can you elaborate here?

Nevermind, I thought this was the client. These are good defaults (although you'll break every existing server on update, forcing them to manually update their config).

scheibo: > (although you'll break every existing server on update, forcing them to manually update their config)

Ugh, you're right. Is there a way around this?

Zarel: Set the default if it doesn't exist.

`if (!Config.routes) Config.routes = ...`

scheibo: Probably would want to check each route TBH, in case only some are specified to be as safe as possible? Is `server/index.js` the place for this?

server/chat-formatter.ts
scheibo: Hmm, yes, I see that its cross compiled there. I'll punt on this for now. Reverted

config/config-example.js
Zarel: Possibly. I've wanted a config loader for a while (client has one now!) but haven't made one yet.

Slayer95: ``app`` would be used by #5515, and yeah ``logs`` would be for the private chat plugin.

config/config-example.js
```diff
@@ -136,6 +135,19 @@ Y929lRybWEiKUr+4Yw2O1W0CAwEAAQ==
...
+ * routes - where Pokemon Showdown is hosted.
+ *   Don't change this setting - there aren't any other options right now
+ */
+exports.routes = {
+	root: 'pokemonshowdown.com',
+	client: 'play.pokemonshowdown.com',
+	dex: 'dex.pokemonshowdown.com',
+	replays: 'replay.pokemonshowdown.com',
+	app: 'sim.smogon.com',
+	logs: 'logs2.psim.us:8080',
```
Zarel: If `logs` is for the private chat plugin, it probably shouldn't be in the example config. We can manually add it to Main's config.

config/config-example.js
```diff
@@ -136,6 +135,19 @@ Y929lRybWEiKUr+4Yw2O1W0CAwEAAQ==
...
+/**
+ * routes - where Pokemon Showdown is hosted.
+ *   Don't change this setting - there aren't any other options right now
+ */
+exports.routes = {
+	root: 'pokemonshowdown.com',
+	client: 'play.pokemonshowdown.com',
+	dex: 'dex.pokemonshowdown.com',
+	replays: 'replay.pokemonshowdown.com',
+	app: 'sim.smogon.com',
```
Zarel: Similarly, `sim.smogon.com` is explicit to Main, and should not be in an example config for side servers.

server/index.js
```diff
@@ -75,7 +76,7 @@ if (Config.watchconfig) {
 	FS(configPath).onModify(() => {
 		try {
 			delete require.cache[configPath];
-			global.Config = require('../config/config');
+			global.Config = ConfigLoader.load(require('../config/config'));
```
Zarel: In theory this shouldn't need an argument. `delete require.cache[configPath]` should probably be part of `load`

Zarel:
Looks good. One day, though, we should strongly type the Config object exported from config-loader.

Maybe it should extract the type exported from `config-example`.

Zarel:
There's a pending review request from Hoeen so I guess I'll leave this unmerged for now?

scheibo:
The pending review request is automatic due to `CODEOWNERS`, but I dont think this is a change Hoeen would care about...

PR #5558, opened 2019-06-27
Rework random teams to only perform 2 iterations

scheibo:
After this, the only remaining restrictions on the second pass are:

1) Species clause
2) 6th slot Zoroak
3) bellydrum/swordsdance/calmmind/nastyplot sets
4) multiple zmoves

Not sure if we want to remove the restrictions from 2-4 as well, LMK.

### DIFF
--- a/data/random-teams.js
+++ b/data/random-teams.js
@@ -1833,21 +1833,10 @@ class RandomTeams extends Dex.ModdedDex {
 
 				// Okay, the set passes, add it to our team
 				pokemon.push(set);
-
-				if (pokemon.length === 6) {
-					let illusion = teamDetails['illusion'];
-					if (illusion) {
-						// Make sure Zoroark isn't in the last slot
-						// (It can't use Illusion if it switches in from the last slot.)
-						if (illusion === 5) {
-							[pokemon[5], pokemon[4]] = [pokemon[4], pokemon[5]];
-							illusion = 4;
-						}
-						// Set Zoroark's level to be the same as the last Pokemon
-						pokemon[illusion - 1].level = pokemon[5].level;
-					}
-					break;
-				}
+				// For setting Zoroark's level and slot
+				if (set.ability === 'Illusion') teamDetails['illusion'] = pokemon.length;
+				// Don't bother performing accounting/tracking other teamDetails on the last Pokemon.
+				if (pokemon.length === 6) break;
 
 				// Now that our Pokemon has passed all checks, we can increment our counters
 				baseFormes[template.baseSpecies] = 1;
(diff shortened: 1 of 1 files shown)


Zarel:
Why is 6th slot Zoroark a restriction, rather than just a thing you do after generating the team?

scheibo:
> Why is 6th slot Zoroark a restriction, rather than just a thing you do after generating the team?

_Je sais pas_

data/random-teams.js
```diff
@@ -1835,9 +1835,16 @@ class RandomTeams extends Dex.ModdedDex {
 				pokemon.push(set);
 
 				if (pokemon.length === 6) {
-					// Set Zoroark's level to be the same as the last Pokemon
 					let illusion = teamDetails['illusion'];
```
Zarel: Why isn't all this just after the `break`?

data/random-teams.js
```diff
@@ -1835,9 +1835,16 @@ class RandomTeams extends Dex.ModdedDex {
...
-					if (illusion) pokemon[illusion - 1].level = pokemon[5].level;
+					if (illusion) {
+						// Swap Zoroark's slot if it is the last Pokemon
+						if (illusion === 5) {
```
Zarel: I'm confused, what's going on here? Can't Zoroark be in any slot?

scheibo: *if* it happens to be in the last slot, swap it with the second-to-last slot so that Illusion actually *does* something

data/random-teams.js
scheibo: I don't know, man. I'm not in this part of the codebase to ask questions, I'm here to fix crashes and help out the poor souls stuck with a 5-mon monoice team

data/random-teams.js
Zarel: Ohhhh. Maybe that comment could be rewritten to be clearer.

data/random-teams.js
Zarel: I mean, fair, but since you're rewriting this block anyway, you might as well move it to be after the iteration. The less heavily-indented code we have, the better.

data/random-teams.js
```diff
@@ -1835,9 +1835,16 @@ class RandomTeams extends Dex.ModdedDex {
 				pokemon.push(set);
 
 				if (pokemon.length === 6) {
-					// Set Zoroark's level to be the same as the last Pokemon
 					let illusion = teamDetails['illusion'];
-					if (illusion) pokemon[illusion - 1].level = pokemon[5].level;
+					if (illusion) {
+						// Swap Zoroark's slot if it is the last Pokemon
```
Zarel:
```suggestion
						// Make sure Zoroark isn't in the last slot
						// (It can't use Illusion if it switches in from the last slot.)
```

data/random-teams.js
```diff
@@ -1869,12 +1866,21 @@ class RandomTeams extends Dex.ModdedDex {
...
+			// Make sure Zoroark isn't in the last slot
+			// (It can't use Illusion if it switches in from the last slot)
+			if (illusion === 5) {
+				[pokemon[5], pokemon[4]] = [pokemon[4], pokemon[5]];
+				illusion = 4;
 			}
+			// Set Zoroark's level to be the same as the last Pokemon
+			pokemon[illusion - 1].level = pokemon[5].level;
 		}
-		if (pokemon.length < 6) throw new Error(`Could not build a random team for ${this.format} (type=${type})`);
```
Zarel: Why was this removed? It seems useful to leave in.

PR #5562, opened 2019-06-30
Change invite only modjoin from + -> %

scheibo:


TheJetOU:
IDK the context behind this, but shouldn't players also have the permission to use `/modjoin %` to keep things consistent?

server/chat-plugins/roomsettings.js
```diff
@@ -349,7 +349,7 @@ exports.commands = {
 			this.addModAction(`${user.name} set modjoin to autoconfirmed.`);
 			this.modlog('MODJOIN', null, 'autoconfirmed');
 		} else if (target in Config.groups || target === 'trusted') {
-			if (room.battle && !user.can('makeroom') && target !== '+') return this.errorReply(`/modjoin - Access denied from setting modjoin past + in battles.`);
+			if (room.battle && !user.can('makeroom') && (target !== '+' || target !== '%')) return this.errorReply(`/modjoin - Access denied from setting modjoin past % in battles.`);
```
scheibo: Should line 277 and or 302 (fuck GH for not letting me comment on them) be `/modjoin %` instead of `+`?

Zarel: Yes, assuming those are the `/inviteonly` implementations.

Zarel: This should be `&&`, not `||`.

`(target !== '+' || target !== '%')` always evaluates to `true`. :p

Zarel: Fortunately irrelevant, because there's a better way to write this:

```suggestion
			if (room.battle && !user.can('makeroom') && !'+%'.includes(target)) {
				return this.errorReply(`/modjoin - Access denied from setting modjoin past % in battles.`);
			}
```

server/chat-plugins/roomsettings.js
```diff
@@ -349,7 +349,7 @@ exports.commands = {
...
 			if (room.isPersonal && !user.can('makeroom') && target !== '+') return this.errorReply(`/modjoin - Access denied from setting modjoin past + in group chats.`);
```
Zarel: This should probably also be allowed.

```suggestion
			if (room.isPersonal && !user.can('makeroom') && !'+%'.includes(target)) {
				return this.errorReply(`/modjoin - Access denied from setting modjoin past + in group chats.`);
			}
```

server/rooms.js
```diff
@@ -1690,10 +1690,10 @@ let Rooms = Object.assign(getRoom, {
 		}
 		if (options.tour && !room.tour.modjoin) inviteOnly = [];
 		if (inviteOnly.length) {
-			room.modjoin = '+';
+			room.modjoin = '%';
 			room.isPrivate = 'hidden';
 			room.privacySetter = new Set(inviteOnly);
-			room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be rank + or invited with <code>/invite</code> to join</div>`);
+			room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be rank % or invited with <code>/invite</code> to join</div>`);
```
Zarel:
```suggestion
			room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be invited with <code>/invite</code> (or staff) to join</div>`);
```

Zarel:
Note that this doesn't fix the original bug that prompted this discussion (users below gdriver can apparently view and join hidden battles???)

PR #5564, opened 2019-07-01
Allow global drivers to set modchat in battles

bumbadadabum:
This was requested by global staff, not sure how to implement this.

### DIFF
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -208,9 +208,12 @@ exports.commands = {
 			return this.sendReply(`Moderated chat is currently set to: ${modchatSetting}`);
 		}
 		if (!this.canTalk()) return;
-		if (!this.can('modchat', null, room)) return false;
 
-		if (room.modchat && room.modchat.length <= 1 && Config.groupsranking.indexOf(room.modchat) > 1 && !user.can('modchatall', null, room)) {
+		const canSetAll = user.can('modchatall', null, room) || (room.battle && user.can('lock'));
+
+		if (!this.can('modchat', null, room) && !canSetAll) return false;
+
+		if (room.modchat && room.modchat.length <= 1 && Config.groupsranking.indexOf(room.modchat) > 1 && !canSetAll) {
 			return this.errorReply(`/modchat - Access denied for removing a setting higher than ${Config.groupsranking[1]}.`);
 		}
 		if (room.requestModchat) {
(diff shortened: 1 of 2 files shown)


Zarel:
-_-

Zarel:
Can you link to the staff discussion?

Zarel:
It's just a really hardcoded change.

I'm thinking, maybe we just want to give drivers all mod powers in battles. That would be also hardcoded, but probably a lot simpler to implement, and would prevent us having the same conversation over and over for every other mod power.

Zarel:
In particular, I'm thinking there's no reason not to give gdrivers `/roomban`, too.

bumbadadabum:
Honestly, how I'd ideally do it is adding a flag for permissions to indicate "for battles only", and just give drivers the mod perms only in battles directly in config.js

Zarel:
I discussed that with @scheibo, too. It seems like a lot of work if we're just going to give drivers every roommod permission in battles, anyway.

Also we probably want to treat all temporary rooms the same (including groupchats, help rooms, battles, etc). Instead of `isPersonal`, a new `isUnregistered` designation.

DaWoblefet:
Currently, I'm unable as a global mod to set modchat for driver+ in battles. It'd be really nice for champ battles in particular; it's a great way to corral everyone and have time to actually analyze the battle to see who needs to get locked while immediately solving the problem for the guy on the receiving end. Currently, modchatting to voice is a joke for champ battles and is easily evaded.

Zarel:
> Currently, I'm unable as a global mod to set modchat for driver+ in battles. [...] Currently, modchatting to voice is a joke for champ battles and is easily avoided by the player voicing everyone.

This wouldn't actually change anything; players can just change the modchat back.

What would make much more sense is if you could demote players, which is probably also one of those things mods/drivers should be able to do.

scheibo:
I think what is really being asked for is a 'battle lockdown' command in some way. Maybe this both sets modchat and demotes players to prevent them for changing it, I don't know. The usecase we're trying to solve is the one DaWoblefet laid out - slowing things down so that mods can sort out who to actually punish

scheibo:
> This wouldn't actually change anything; players can just change the modchat back.

Staff just tested this and it doesn't appear to be the case? Player's can't change the modchat back once its been set to %?

Zarel:
> Staff just tested this and it doesn't appear to be the case? Player's can't change the modchat back once its been set to %?

Oh yeah, I got modchat and modjoin mixed up.

In which case yes, giving gdrivers mod powers in battles would in fact solve this problem, too.

PR #5566, opened 2019-07-01
Fix Follow the Leader's validation

urkerab:


Zarel:
I have no clue how this format works, but I assume I can trust you on this.

PR #5567, opened 2019-07-02
Update Nature Swap

Spandamn:
@TheImmortal 
I think this would be useful.
 
ns

### DIFF
--- a/server/chat-plugins/othermetas.js
+++ b/server/chat-plugins/othermetas.js
@@ -347,13 +347,13 @@ const commands = {
 		let nature = target.trim().split(' ')[0];
 		let pokemon = target.trim().split(' ')[1];
 		if (!toID(nature) || !toID(pokemon)) return this.parse(`/help natureswap`);
-		nature = Dex.getNature(nature);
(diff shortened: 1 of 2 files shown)


server/chat-plugins/othermetas.js
```diff
@@ -339,6 +339,28 @@ const commands = {
...
+	natureswap(target, room, user) {
+		if (!this.runBroadcast()) return;
+		let nature = target.trim().split(' ')[0];
+		let pokemon = target.trim().split(' ')[1];
+		if (!toID(nature) || !toID(pokemon)) return this.parse(`/help natureswap`);
+		let natureObj = Dex.getNature(nature);
+		if (!natureObj.exists) return this.errorReply(`Error: Pokemon ${natureObj} not found.`);
+		let template = Dex.deepClone(Dex.getTemplate(pokemon));
+		if (!template.exists) return this.errorReply(`Error: Pokemon ${pokemon} not found.`);
+		// @ts-ignore
```
scheibo: You should not require `@ts-ignore` at all here, please fix the type errors instead.

Spandamn:
I also added a fix for the EVs bug

PR #5570, opened 2019-07-02
Actually filter statuses differently

whalemer:


whalemer:
dealt with in 48c9304a4a33499453e2e05222c0c5a9b3dc66a5

PR #5571, opened 2019-07-02
Notify users when trying to use a banned status

whalemer:


bumbadadabum:
To be nitpicky I would flip the condition and make it if forStatus else

PR #5573, opened 2019-07-02
Various status improvements

whalemer:


server/chat-commands.js
```diff
@@ -716,6 +714,23 @@ const commands = {
 	unaway: 'back',
 	unafk: 'back',
 	back(target, room, user) {
+		if (target) {
+			// Clearing another user's status
```
Zarel: This should probably be a different command?

Zarel: Especially since it has exactly nothing in common with the regular command.

server/chat-commands.js
```diff
@@ -716,6 +714,23 @@ const commands = {
...
+			let reason = this.splitTarget(target);
+			let targetUser = this.targetUser;
+			if (!targetUser) return this.errorReply(`User '${target}' not found.`);
+			if (!targetUser.status) return this.errorReply(`${targetUser.name} does not have a status set.`);
+			if (!this.can('forcerename', targetUser)) return false;
+
+			let bracketIndex = targetUser.status.indexOf(')');
+			if (bracketIndex < 0) bracketIndex = -2; // Someone's trying to clear the "Idle" status, for some reason
```
Zarel: Why is it `Idle` and not `(Idle)`, anyway?

scheibo: Check commits, its been fixed to be consistent

PR #5574, opened 2019-07-02
Record statistics for help tickets

HoeenCoder:
This Pull Request adds support for logging statistics about help tickets. These statistics are stored in tsv files located in `logs/tickets/YYYY-MM.tsv`.

When a ticket is closed, a line is added to the current month's tsv file containing information about the ticket including its type, how long it was open, how long it took staff to claim it for the first time, how long it was waiting on staff in total, what staff members participated (to our best knowledge) in the ticket's resolution, if the ticket was resolved, and its result. The possible ticket resolutions are:

- "dead" - The ticket was never activated by the user and was closed by the user (staff were not contacted).
- "unresolved" - The ticket was activated and staff were contacted, but no staff member came before the ticket was closed by the user.
- "resolved" - The ticket was activated and claimed by staff before it was closed.

Ticket results are either positive or negative, and the result is determined by which close button a global staff member uses.

- Appeal-type tickets ask global staff to close the ticket as "Approved" or "Denied". 
- Report-type tickets ask global staff to close the ticket as "Valid" or "Invalid".
- Other tickets ask global staff to close the ticket as "Assisted" or "Unassisted".

This information is formatted into two different tables. The first one shows average stats for each ticket type. This shows the total number of tickets of each type, average wait times for each type of ticket, and the percentage of ticket that closed with a specific resolution/positively.
Type Table

The second table shows stats for each staff member who has done at least one ticket in the month. The information shown is the number of tickets done, and the average time spent on a single ticket.
Staff Table

Of course none of this is perfectly accurate, outliers in averages can cause the average to jump higher/lower than it should be, and sometimes staff may help with far more tickets than shown in stats since they do it through communication in the staff room. But considering what we have to work with I think its setup pretty well.

Something to consider, If you have any suggestions for compressing the information in the % Resolutions, and % Positive Result cols (including the headers) please let me know, even something as simple as a better header to shrink that row is appreciated.

@scheibo helped with the design of this, thanks pre!

### DIFF
--- a/logs/tickets/README.md
+++ b/logs/tickets/README.md
@@ -1,0 +1,1 @@
+Logs of ticket stats are stored in this directory.
--- a/server/chat-plugins/helptickets.js
+++ b/server/chat-plugins/helptickets.js
@@ -36,6 +36,7 @@ const TICKET_BAN_DURATION = 48 * 60 * 60 * 1000; // 48 hours
  * @property {string} reason
  * @property {number} expires
  */
+/** @typedef {'approved' | 'valid' | 'assisted' | 'denied' | 'invalid' | 'unassisted' | 'ticketban' | 'deleted'} TicketResult */
 
 /** @type {{[k: string]: TicketState}} */
 let tickets = {};
@@ -69,6 +70,20 @@ function writeTickets() {
 	));
 }
 
+/**
+ * @param {string} line
+ */
+function writeStats(line) {
+	// ticketType\ttotalTime\ttimeToFirstClaim\tinactiveTime\tresolution\tresult\tstaff,userids,seperated,with,commas
+	const date = new Date();
+	const month = Chat.toTimestamp(date).split(' ')[0].split('-', 2).join('-');
+	try {
+		FS(`logs/tickets/${month}.tsv`).appendSync(line + '\n');
+	} catch (e) {
+		if (e.code !== 'ENOENT') throw e;
+	}
+}
+
 class HelpTicket extends Rooms.RoomGame {
 	/**
 	 * @param {ChatRoom} room
@@ -82,6 +97,19 @@ class HelpTicket extends Rooms.RoomGame {
 		this.ticket = ticket;
 		/** @type {string[]} */
 		this.claimQueue = [];
+		/* Stats */
+		/** @type {Set<ID>} */
+		this.involvedStaff = new Set();
+		this.createTime = Date.now();
+		this.activationTime = (ticket.active ? this.createTime : 0);
+		this.firstClaimTime = 0;
+		this.unclaimedTime = 0;
+		this.lastUnclaimedStart = (ticket.active ? this.createTime : 0);
+		this.closeTime = 0;
+		/** @type {'unknown' | 'dead' | 'unresolved' | 'resolved'} */
+		this.resolution = 'unknown';
+		/** @type {TicketResult?} */
+		this.result = null;
 	}
 
 	/**
@@ -97,6 +125,9 @@ class HelpTicket extends Rooms.RoomGame {
 		if (this.ticket.escalated && !user.can('declare')) return false;
 		if (!this.ticket.claimed) {
 			this.ticket.claimed = user.name;
+			if (!this.firstClaimTime) this.firstClaimTime = Date.now();
+			this.unclaimedTime += Date.now() - this.lastUnclaimedStart;
+			this.lastUnclaimedStart = 0; // Set back to 0 so we know that it was active when closed
 			tickets[this.ticket.userid] = this.ticket;
 			writeTickets();
 			this.modnote(user, `${user.name} claimed this ticket.`);
@@ -123,6 +154,7 @@ class HelpTicket extends Rooms.RoomGame {
 				this.modnote(user, `This ticket is now claimed by ${this.ticket.claimed}.`);
 			} else {
 				this.ticket.claimed = null;
+				this.lastUnclaimedStart = Date.now();
 				this.modnote(user, `This ticket is no longer claimed.`);
 				notifyStaff(this.ticket.escalated);
 			}
@@ -139,6 +171,8 @@ class HelpTicket extends Rooms.RoomGame {
 	 * @param {User} user
 	 */
 	onLogMessage(message, user) {
+		if (!this.ticket.open) return;
+		if (user.isStaff && this.ticket.userid !== user.userid) this.involvedStaff.add(user.userid);
 		if (this.ticket.active) return;
 		const blockedMessages = [
 			'hi', 'hello', 'hullo', 'hey', 'yo', 'ok',
@@ -152,6 +186,8 @@ class HelpTicket extends Rooms.RoomGame {
 		}
 		if ((!user.isStaff || this.ticket.userid === user.userid) && !this.ticket.active) {
 			this.ticket.active = true;
+			this.activationTime = Date.now();
+			this.lastUnclaimedStart = Date.now();
 			notifyStaff(this.ticket.escalated);
 		}
 	}
@@ -165,7 +201,7 @@ class HelpTicket extends Rooms.RoomGame {
 		if (!this.ticket.open) return;
 		this.modnote(user, `${user.name} is no longer interested in this ticket.`);
 		if (this.playerCount - 1 > 0) return; // There are still users in the ticket room, dont close the ticket
-		this.close(user);
+		this.close(user, !!(this.firstClaimTime));
 		return true;
 	}
 
@@ -223,8 +259,9 @@ class HelpTicket extends Rooms.RoomGame {
 
 	/**
 	 * @param {User} staff
+	 * @param {boolean | 'ticketban' | 'deleted'} result
 	 */
-	close(staff) {
+	close(staff, result) {
 		this.room.isHelp = 'closed';
 		this.ticket.open = false;
 		tickets[this.ticket.userid] = this.ticket;
@@ -237,13 +274,65 @@ class HelpTicket extends Rooms.RoomGame {
 			const user = Users.get(ticketGameUser.userid);
 			if (user) user.updateSearch();
 		}
+		if (!this.involvedStaff.size) {
+			if (staff.isStaff && staff.userid !== this.ticket.userid) {
+				this.involvedStaff.add(staff.userid);
+			} else {
+				this.involvedStaff.add(toID(this.ticket.claimed));
+			}
+		}
+		this.writeStats(result);
+	}
+
+	/**
+	 * @param {boolean | 'ticketban' | 'deleted'} result
+	 */
+	writeStats(result) {
+		// Only run when a ticket is closed/banned/deleted
+		this.closeTime = Date.now();
+		if (this.lastUnclaimedStart) this.unclaimedTime += this.closeTime - this.lastUnclaimedStart;
+		if (!this.ticket.active) {
+			this.resolution = "dead";
+		} else if (!this.firstClaimTime) {
+			this.resolution = "unresolved";
+		} else {
+			this.resolution = "resolved";
+		}
+		if (typeof result === 'boolean') {
+			switch (this.ticket.type) {
+			case 'Appeal':
+			case 'IP-Appeal':
+			case 'ISP-Appeal':
+				this.result = (result ? 'approved' : 'denied');
+				break;
+			case 'PM Harassment':
+			case 'Battle Harassment':
+			case 'Inappropriate Username':
+			case 'Inappropriate Pokemon Nicknames':
+				this.result = (result ? 'valid' : 'invalid');
+				break;
+			case 'Public Room Assistance Request':
+			case 'Other':
+			default:
+				this.result = (result ? 'assisted' : 'unassisted');
+				break;
+			}
+		} else {
+			this.result = result;
+		}
+		let firstClaimWait = 0;
+		if (this.activationTime) firstClaimWait = (this.firstClaimTime ? this.firstClaimTime : this.closeTime) - this.activationTime;
+		// Write to TSV
+		// ticketType\ttotalTime\ttimeToFirstClaim\tinactiveTime\tresolution\tresult\tstaff,userids,seperated,with,commas
+		const line = `${this.ticket.type}\t${(this.closeTime - this.createTime)}\t${firstClaimWait}\t${this.unclaimedTime}\t${this.resolution}\t${this.result}\t${Array.from(this.involvedStaff.entries()).map(s => s[0]).join(',')}`;
+		writeStats(line);
 	}
 
 	/**
 	 * @param {User} staff
 	 */
 	deleteTicket(staff) {
-		this.close(staff);
+		this.close(staff, 'deleted');
 		this.modnote(staff, `${staff.name} deleted this ticket.`);
 		delete tickets[this.ticket.userid];
 		writeTickets();
@@ -437,21 +526,14 @@ function checkTicketBanned(user) {
 for (const room of Rooms.rooms.values()) {
 	if (!room.isHelp || !room.game) continue;
 	let game = /** @type {HelpTicket} */ (room.game);
-	const queue = game.claimQueue;
-	const ticket = game.ticket;
-	room.game.destroy();
-	room.game = null;
-	if (!ticket) continue;
-	game = new HelpTicket(/** @type {ChatRoom} */ (room), tickets[ticket.userid]);
-	game.claimQueue = queue;
-	room.game = game;
+	game.ticket = tickets[game.ticket.userid];
 }
 
 /** @type {{[k: string]: string}} */
 const ticketTitles = Object.assign(Object.create(null), {
 	pmharassment: `PM Harassment`,
 	battleharassment: `Battle Harassment`,
-	inapname: `Inappropriate Username/Status Message`,
+	inapname: `Inappropriate Username / Status Message`,
 	inappokemon: `Inappropriate Pokemon Nicknames`,
 	appeal: `Appeal`,
 	ipappeal: `IP-Appeal`,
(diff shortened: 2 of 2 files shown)


server/chat-plugins/helptickets.js
```diff
@@ -69,6 +70,25 @@ function writeTickets() {
...
+	// ticketType,totalTime,timeToFirstClaim,inactiveTime,resolution,result,staff|userids|seperated|with|bars
+	const date = new Date();
+	const month = Chat.toTimestamp(date).split(' ')[0].split('-', 2).join('-');
+	let ticketStats = '';
+	try {
+		ticketStats = FS(`logs/tickets/${month}.csv`).readSync();
+	} catch (e) {
+		if (e.code !== 'ENOENT') throw e;
+	}
+	ticketStats += (ticketStats ? '\n' : '') + line;
```
TheJetOU: Writing the entire file contents to it every time sounds a bit inefficient (unless if I'm misunderstanding the code here), why not use something like `FS.append` since we're only adding one line at at time?

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+			} else {
+				buttonBar += `<a class="button disabled" style="float: right">Next Month &gt;</a>`;
+			}
+			let col = query.shift() || '';
+
+			let buf = `<div class="pad ladder"><div style="text-align: center">${buttonBar}</div><br />`;
+			buf += `<table style="margin-left: auto; margin-right: auto"><tbody><tr><th colspan="${table === 'tickets' ? 7 : 3}"><h2 style="margin: 5px auto">Help Ticket Stats - ${date.toLocaleString('en-us', {month: 'long', year: 'numeric'})}</h1></th></tr>`;
+			if (table === 'tickets') {
+				if (!['type', 'totaltickets', 'total', 'initwait', 'wait', 'resolution', 'result'].includes(col)) col = 'type';
+				buf += `<tr><th><Button>type</Button></th><th><Button>totaltickets</Button></th><th><Button>total</Button></th><th><Button>initwait</Button></th><th><Button>wait</Button></th><th><Button>resolution</Button></th><th><Button>result</Button></th></tr>`;
```
TheJetOU: Not really an 'issue' per se, but HTML conventions is to use lowercase for tags (see `<Button>`).

HoeenCoder: This isn't a normal button tag, check down at the end of the page defenition and you will see a replacement regexp that replaces `<Button>text</Button>` with some `<a>` tags.

scheibo: Out of scope for this change, but why?? Why is this not just something we handle with CSS?

scheibo:
> Something to consider, If you have any suggestions for compressing the information in the % Resolutions, and % Positive Result cols (including the headers) please let me know, even something as simple as a better header to shrink that row is appreciated.

I don't think '%' is necessary in the header - when its already included in each and every cell, its fairly clear that its talking about percentages.

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
 		},
+		stats(query, user, connection) {
+			// view-help-stats-TABLE-YYYY-MM-COL
+			if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
+			this.title = 'Ticket Stats';
+			if (!this.can('lock')) return;
+
+			let table = query.shift() || '';
+			if (!['staff', 'tickets'].includes(table)) table = 'tickets';
+			let monthString = `${query.shift()}-${query.shift()}`;
```
scheibo: I think using a destructuring assignment to create some variables would be more readable

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+
+			// Calculate next/previous month for stats and validate stats exist for the month
+			const prevDate = new Date(monthString);
+			if (!date) date = new Date((prevDate.getMonth() === 11 ? prevDate.getFullYear() + 1 : prevDate.getFullYear()), (prevDate.getMonth() === 11 ? 0 : prevDate.getMonth() + 1));
+			const nextDate = new Date((prevDate.getMonth() >= 10 ? prevDate.getFullYear() + 1 : prevDate.getFullYear()), (prevDate.getMonth() + 2 > 11 ? (prevDate.getMonth() + 2) - 12 : prevDate.getMonth() + 2));
+			let prevString = Chat.toTimestamp(prevDate).split(' ')[0].split('-', 2).join('-');
+			let nextString = Chat.toTimestamp(nextDate).split(' ')[0].split('-', 2).join('-');
+
+			let buttonBar = '';
+			if (FS(`logs/tickets/${prevString}.csv`).readIfExistsSync()) {
```
scheibo: Why are we reading in the whole file just to check if it exists :O? just `lstat`??

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+				buttonBar += `<a class="button" href="/view-help-stats-${table}-${prevString}" target="replace" style="float: left">&lt; Previous Month</a>`;
+			} else {
+				buttonBar += `<a class="button disabled" style="float: left">&lt; Previous Month</a>`;
+			}
+			buttonBar += `<a class="button${table === 'tickets' ? ' disabled"' : `" href="/view-help-stats-tickets-${monthString}" target="replace"`}>Ticket Stats</a> <a class="button ${table === 'staff' ? ' disabled"' : `" href="/view-help-stats-staff-${monthString}" target="replace"`}>Staff Stats</a>`;
+			if (FS(`logs/tickets/${nextString}.csv`).readIfExistsSync()) {
```
scheibo: ditto

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+					initwait: parseInt(splitLine[2]),
+					wait: parseInt(splitLine[3]),
+					resolution: splitLine[4],
+					result: splitLine[5],
+					staff: splitLine[6],
+				};
+			});
+			if (table === 'tickets') {
+				// {total: num, initwait: num, wait: num, resolution: {dead: num, unresolved: num, resolved: num}, result: num positive, totaltickets: num}
+				/** @type {any} */
```
scheibo: huh? You type this as `any` and add a comment explaining its real type? Why not just use the real type in the first place?

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+					if (['dead', 'unresolved', 'resolved'].includes(stats.resolution.toString())) {
+						if (!type.resolution[stats.resolution.toString()]) type.resolution[stats.resolution.toString()] = 0;
+						type.resolution[stats.resolution.toString()]++;
+					}
+				}
+
+				// Calculate averages
+				for (let t in typeStats) {
+					let type = typeStats[t];
+					type.total = Math.round(type.total / type.totaltickets);
```
scheibo: Why are you rounding at each step instead of only at the end?

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+					type.wait = Math.round(type.wait / type.totaltickets);
+					type.result = Math.round((type.result / type.totaltickets) * 100);
+					for (let resolution in type.resolution) {
+						type.resolution[resolution] = Math.round((type.resolution[resolution] / type.totaltickets) * 100);
+					}
+				}
+
+				let sortedStats = Object.keys(typeStats).sort((a, b) => {
+					if (col === 'type') {
+						// Alphabatize strings
```
scheibo: nit: Alphabetize

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+			} else {
+				/** @type {{[key: string]: {[key: string]: number}}} */
+				let staffStats = {};
+				for (let stats of ticketStats) {
+					let staff = (typeof stats.staff === 'string' ? stats.staff.split(';') : []);
+					for (let s = 0; s < staff.length; s++) {
+						if (!staff[s]) continue;
+						if (!staffStats[staff[s]]) staffStats[staff[s]] = {num: 0, time: 0};
+						staffStats[staff[s]].num++;
+						// @ts-ignore
```
scheibo: what are you ignoring? please either add a comment explaining why necessary or remove the ignore

server/chat-plugins/helptickets.js
```diff
@@ -758,6 +848,165 @@ const pages = {
...
+						staffStats[staff[s]].time += (stats.total - stats.initwait);
+					}
+				}
+				for (let staff in staffStats) {
+					staffStats[staff].time = Math.round(staffStats[staff].time / staffStats[staff].num);
+				}
+				let sortedStaff = Object.keys(staffStats).sort((a, b) => {
+					if (col === 'staff') {
+						// Alphabatize strings
```
scheibo: ditto

scheibo: also, why not just `localeCompare`?

server/chat-plugins/helptickets.js
HoeenCoder: Reduces the number of lines and amount of various checks I need to get this to work. Its also used for the help ticket form.

Lusamine:
By the way, you have a typo in "% Postive Result" in the screenshot.

server/chat-plugins/helptickets.js
```diff
@@ -235,13 +280,65 @@ class HelpTicket extends Rooms.RoomGame {
...
+			default:
+				this.result = (result ? 'assisted' : 'unassisted');
+				break;
+			}
+		} else {
+			this.result = result;
+		}
+		let firstClaimWait = 0;
+		if (this.activationTime) firstClaimWait = (this.firstClaimTime ? this.firstClaimTime : this.closeTime) - this.activationTime;
+		// Write to CSV
```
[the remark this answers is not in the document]
HoeenCoder: It still uses CSV (unless you meant something else)

server/chat-plugins/helptickets.js
```diff
@@ -82,6 +102,20 @@ class HelpTicket extends Rooms.RoomGame {
 		this.ticket = ticket;
 		/** @type {string[]} */
 		this.claimQueue = [];
+		/* Stats */
+		let createTime = Date.now();
+		/** @type {string[]} */
+		this.involvedStaff = [];
```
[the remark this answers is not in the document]
HoeenCoder: It should, but like I mentioned in our PMs, I cant push userid's then because they are considered strings.

server/chat-plugins/helptickets.js
HoeenCoder: Mostly because its a huge pain to get this typed properly, I think I just decided to stop for a bit and wanted to come back to it later but forgot. I'll give it another shot.

server/chat-plugins/helptickets.js
HoeenCoder: You'll have to clarify what you want me to do here since I dont know what you mean.

server/chat-plugins/helptickets.js
HoeenCoder: I don't really have a problem with the way it is atm. I wanted to round before sorting to prevent any possible issues there, and the only other place I can think of for rounding would be when building the table itself, which really isnt much different than this, its just all in one line.

server/chat-plugins/helptickets.js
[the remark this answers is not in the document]
scheibo: > separated with ; now, no?

server/chat-plugins/helptickets.js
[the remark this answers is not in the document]
scheibo: Ah OK

server/chat-plugins/helptickets.js
scheibo:
```ts
let [year, month] = query;
let monthString = `${year}-$month}`;
```

server/chat-plugins/helptickets.js
scheibo: My concern was that rounding as you calculate the average changes the semantics:

0.49 + 4.49 + 6.49 = 8.98/2 = 3.82, round(3.82) = 4
round(0.49) + round(4.49) + round(6.49) = 10/3 = 3

But I was led astray by the loop and the names (`type.total` really appears like a _total_, not an _average_) - youre doing the sums above (and not rounding) and only computing the averages here (where its correct to round).

server/chat-plugins/helptickets.js
HoeenCoder: What about windows/mac support? Also would it potentially be possible to implement something on FS to check if a file exists without reading it maybe? @Zarel

server/chat-plugins/helptickets.js
HoeenCoder: Apparently its ok now? TS was throwing for a weird reason when I coded this but removing the check doesn't cause an error.

server/chat-plugins/helptickets.js
```diff
@@ -69,6 +70,21 @@ function writeTickets() {
...
 
+/**
+ * @param {string} line
+ */
+function writeStats(line) {
+	// ticketType,totalTime,timeToFirstClaim,inactiveTime,resolution,result,staff;userids;seperated;with;semicolons
+	const date = new Date();
+	const month = Chat.toTimestamp(date).split(' ')[0].split('-', 2).join('-');
+	try {
+		// TODO dont start with \n if file is empty
```
HoeenCoder: Note to self.

Worst case, I can just have the parser handle a blank line at the end of the file.

logs/tickets/README.md
```diff
@@ -0,0 +1 @@
+Logs of ticket stats are stored in this directory.
```
Zarel: Are we still using flat files? How do we feel about SQLite here?

I suppose `JSON.stringify` is much easier than writing an ORM for SQLite. But it looks like you're doing CSV. At which point, we might want to seriously consider SQLite.

Zarel: (If SQLite is too much work, I would recommend at least TSV over CSV; it has less escaping concerns because you can just sanitize tabs out.)

scheibo: I had recommended TSV as well. I think its the way to go for this.

HoeenCoder: re: TSV, I feel like I'm missing something on the use of TSV, It can't be as simple as just seperating values with tabs rather than spaces can it? None of the values used here should have commas in them so it seems wasteful to just do this. Correct me if i'm wrong about how a TSV file is setup/its use here vs CSV.

re: SQLite, have never done that but would be willing to look into it potentionally.

scheibo: > re: TSV, I feel like I'm missing something on the use of TSV, It can't be as simple as just seperating values with tabs rather than spaces can it?

It is exactly as simple as separating values with tabs rather than _commas_. 

> None of the values used here should have commas in them so it seems wasteful to just do this.

Yeah, because you changed what would more logically be commas into `;` because youre using CSV. Zarel and I are saying: please use `\t` and `,` instead of `,` and `;`.

server/chat-plugins/helptickets.js
```diff
@@ -82,6 +97,19 @@ class HelpTicket extends Rooms.RoomGame {
 		this.ticket = ticket;
 		/** @type {string[]} */
 		this.claimQueue = [];
+		/* Stats */
+		/** @type {string[]} */
+		this.involvedStaff = [];
```
TheJetOU: This can now be `ID[]` because `User#userid` is `ID`!

server/chat-plugins/helptickets.js
```diff
@@ -235,13 +274,65 @@ class HelpTicket extends Rooms.RoomGame {
...
+				break;
+			}
+		} else {
+			this.result = result;
+		}
+		let firstClaimWait = 0;
+		if (this.activationTime) firstClaimWait = (this.firstClaimTime ? this.firstClaimTime : this.closeTime) - this.activationTime;
+		// Write to TSV
+		// ticketType\ttotalTime\ttimeToFirstClaim\tinactiveTime\tresolution\tresult\tstaff,userids,seperated,with,commas
+		let line = `${this.ticket.type}\t${(this.closeTime - this.createTime)}\t${firstClaimWait}\t${this.unclaimedTime}\t${this.resolution}\t${this.result}\t${this.involvedStaff.join(',')}`;
```
TheJetOU: Nitpick.
```suggestion
		const line = `${this.ticket.type}\t${(this.closeTime - this.createTime)}\t${firstClaimWait}\t${this.unclaimedTime}\t${this.resolution}\t${this.result}\t${this.involvedStaff.join(',')}`;
```

server/chat-plugins/helptickets.js
TheJetOU: Also, why not just use a set? It would remove the logic needed to check if a staff member is already there.

server/chat-plugins/helptickets.js
TheJetOU: (There's a lot more of these actually, these will all be fixed when this is TypeScript-ed anyway, so it's whatever I guess.)

server/chat-plugins/helptickets.js
```diff
@@ -97,6 +125,9 @@ class HelpTicket extends Rooms.RoomGame {
 		if (this.ticket.escalated && !user.can('declare')) return false;
 		if (!this.ticket.claimed) {
 			this.ticket.claimed = user.name;
+			if (!this.firstClaimTime) this.firstClaimTime = Date.now();
+			this.unclaimedTime += Date.now() - this.lastUnclaimedStart;
```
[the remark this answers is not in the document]
HoeenCoder: `this.unclaimedTime` contains the total time the ticket was unclaimed, even over multiple periods of time. For example, if it takes 2 minutes for the ticket to be claimed at first, and then it is unclaimed for another minute, this propertly will be 3 minutes (in milliseconds ofc).

Zarel:
There's apparently a merge conflict in here, probably from @whalemer's recent work on Help tickets.

HoeenCoder:
Yeah I have them fixed locally, I need to test again to make sure nothing new broke before I push it out.

server/chat-plugins/helptickets.js
```diff
@@ -237,13 +276,65 @@ class HelpTicket extends Rooms.RoomGame {
...
+				this.involvedStaff.add(staff.userid);
+			} else {
+				this.involvedStaff.add(toID(this.ticket.claimed));
+			}
+		}
+		this.writeStats(result);
+	}
+
+	/**
+	 * @param {(boolean | 'ticketban' | 'deleted')} result
```
Zarel:
```suggestion
	 * @param {boolean | 'ticketban' | 'deleted'} result
```

don't mind me, just nitpicking

server/chat-plugins/helptickets.js
```diff
@@ -223,8 +261,9 @@ class HelpTicket extends Rooms.RoomGame {
 
 	/**
 	 * @param {User} staff
+	 * @param {(boolean | 'ticketban' | 'deleted')} result
```
Zarel:
```suggestion
	 * @param {boolean | 'ticketban' | 'deleted'} result
```

server/chat-plugins/helptickets.js
```diff
@@ -139,6 +171,10 @@ class HelpTicket extends Rooms.RoomGame {
 	 * @param {User} user
 	 */
 	onLogMessage(message, user) {
+		if (!this.ticket.open) return;
+		if (user.isStaff && this.ticket.userid !== user.userid) {
+			if (!this.involvedStaff.has(user.userid)) this.involvedStaff.add(user.userid);
```
TheJetOU: That's the point of using a Set, so you don't get duplicate values. The only reason you check if the value is already there is if you want to show an error message.
```suggestion
			this.involvedStaff.add(user.userid);
```

PR #5576, opened 2019-07-02
Refactor user.isAway to a getter function

TheJetOU:
It makes sense semantically. It's also a pure function (has not side effects).

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -731,7 +731,7 @@ const commands = {
 			return;
 		}
 		if (!user.status) return;
-		const statusType = user.isAway() ? 'away' : user.status.startsWith('(Busy)') ? 'busy' : null;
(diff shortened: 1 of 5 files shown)


Zarel:
No, I actually already had this argument with Bumba.

Getters are just bad in general.

Zarel:
There are a number of arguments here, but the main one is that variables should look like variables, and functions should look like functions. Having a function that looks like a variable is bad for readability.

Zarel:
Unrelatedly, having a function that looks like a variable is also bad for performance. Like, 10x slower.

But also, it's just bad because it misleads you into thinking the way to set a user back is to use `.isAway = false`, but it's not.

I think it's important to be clear about what's state and what's computed.

TheJetOU:
Fair enough.

PR #5577, opened 2019-07-03
Check for obvious invalid statuses

whalemer:
This won't catch everything that namefilter does, but I think it's /close enough/ to work

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -656,7 +656,7 @@ const commands = {
 
 	status(target, room, user, connection, cmd) {
 		if (!this.canTalk()) return;
-		if (!target) return this.parse('/help status');
+		if (!toID(target)) return this.parse('/help status');
 		target = Chat.namefilter(target, user, true);
 		if (!target) return user.popup("Your status contains a banned word.");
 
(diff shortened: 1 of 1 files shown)


server/chat-commands.js
```diff
@@ -676,7 +676,7 @@ const commands = {
 		if (!this.canTalk()) return;
 
 		let message = Chat.namefilter(target, user, true);
-		if (!message && target) return user.popup("Your status contains a banned word.");
+		if (!message && toID(target)) return user.popup("Your status contains a banned word.");
```
Zarel: Should probably be more like `target.trim()`?

whalemer: toID also catches statuses made up entirely of symbols (eg, ``/away .``)

Zarel: Yes, but statuses made up entirely of banned symbols are different from calling the function without a status.

Zarel: Like, I don't want a message consisting entirely of swastikas to be allowed, just because it doesn't contain a letter.

whalemer: The status full of swastikas is still going through the namefilter and getting cleared, it just fails silently instead of telling the user their status is banned.

Zarel: It seems like it shouldn't fail silently.

PR #5579, opened 2019-07-03
Add Empty Learnsets for Pokestar Pokemon

jmclemo6:
Discussed in the dev room. I think actually adding empty learnsets for the Pokestar Pokemon is a better design than having to check if `Template.learnset` exists. 

Since Pokestar pokemon were the only things I could find without a learnset defined, I also made `learnset` a required property on `Template` to avoid a bug like this in the future.

These changes make ff911e2 unnecessary.

### DIFF
--- a/data/learnsets.js
+++ b/data/learnsets.js
@@ -64993,6 +64993,57 @@ let BattleLearnsets = {
 		wish: ["7L22"],
 		worryseed: ["7T"],
 	}},
+	pokestarsmeargle: {learnset: {
+
+	}},
+	pokestarufo: {learnset: {
+
+	}},
+	pokestarufo2: {learnset: {
+
+	}},
+	pokestarbrycenman: {learnset: {
+
+	}},
+	pokestarmt: {learnset: {
+
+	}},
+	pokestarmt2: {learnset: {
+
+	}},
+	pokestartransport: {learnset: {
+
+	}},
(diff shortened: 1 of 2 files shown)


Zarel:
The long-term plan here is to have `Dex.getLearnset` rather than shoving learnsets into a template file.

Events should probably also be bundled here, so it wouldn't directly return the learnsets table, but rather a `{learnsets, events}` object.

data/learnsets.js
```diff
@@ -64993,6 +64993,57 @@ let BattleLearnsets = {
 		wish: ["7L22"],
 		worryseed: ["7T"],
 	}},
+	pokestarsmeargle: {learnset: {
+
```
Zarel: Style has never been to add empty lines here. If we were to do these entries, we should do them as one-liners:

```
pokestarsmeargle: {learnset: {}},
```

Zarel: But this is irrelevant. I would rather autofill these in the `Data.Template` constructor, than add empty entries here.

jmclemo6: Alright, sounds good.

Zarel:
Out of curiosity, does this mean you're planning on submitting a PR for the Template constructor, or should I do that myself?

jmclemo6:
I can do that.

PR #5580, opened 2019-07-03
Allow for presetting settings after a rename with /trn

scheibo:
Required for Zarel/Pokemon-Showdown-Client#1271

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4488,6 +4488,7 @@ const commands = {
 		let targetName = target;
 		let targetRegistered = false;
 		let targetToken = '';
+		let settings = {};
 		if (commaIndex >= 0) {
 			targetName = target.substr(0, commaIndex);
 			target = target.substr(commaIndex + 1);
@@ -4495,10 +4496,16 @@ const commands = {
 			targetRegistered = target;
 			if (commaIndex >= 0) {
 				targetRegistered = !!parseInt(target.substr(0, commaIndex));
-				targetToken = target.substr(commaIndex + 1);
+				target = target.substr(commaIndex + 1);
+				commaIndex = target.indexOf(',');
+				targetToken = target;
+				if (commaIndex >= 0) {
+					targetToken = target.substr(0, commaIndex);
+					settings = JSON.parse(target.substr(commaIndex + 1));
+				}
 			}
 		}
-		user.rename(targetName, targetToken, targetRegistered, connection);
+		user.rename(targetName, targetToken, targetRegistered, connection, settings);
 	},
 
 	a(target, room, user) {
--- a/server/users.js
+++ b/server/users.js
@@ -546,7 +546,7 @@ class User extends Chat.MessageContext {
 	}
 
 	/**
-	 * @param {UserSettings} [settings]
+	 * @param {UserSettings?}
 	 */
 	applySettings(settings) {
 		if (!settings) return;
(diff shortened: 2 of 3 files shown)


scheibo:
The problem with this change is that an updated client (sending settings) talking to an old server (whose `/trn` method doesn't understand settings and thinks they're part of the token) isn't going to work. I could smuggle in settings as part of the second parameter (encode settings to it has no commas, and then `parseInt('0asfjhsajf')` still works on old servers etc), but that's pretty hideous. Is my only option here to add a new `/trn++` command, or is there a better way to negotiate with old servers/we don't care?

server/users.js
```diff
@@ -544,6 +545,16 @@ class User extends Chat.MessageContext {
 		Users.add(this);
 	}
 
+	/**
+	 * @param {UserSettings} [settings]
+	 */
+	applySettings(settings) {
+		if (!settings) return;
+		if ('blockChallenges' in settings) this.blockChallenges = !!settings.blockChallenges;
```
TheJetOU: Why the `!!`? Won't the object be something like `{blockChallenges: true, ignoreTickets: false}`? In which case it's the same without the `!!`.

server/users.js
```diff
@@ -544,6 +545,16 @@ class User extends Chat.MessageContext {
 		Users.add(this);
 	}
 
+	/**
+	 * @param {UserSettings} [settings]
```
TheJetOU: `settings` isn't an optional parameter though. You shouldn't be running this function with no parameters. It should probably be something like 
```suggestion
	 * @param {UserSettings?}
```

server/users.js
scheibo: Because the compiler complains otherwise. I think its stupid as well.

scheibo:
> Assertions can have commas in them which will ruin the parsing so you should probably move the settings before the assertion.

This is what I asked in `<<dev>>` - I figured that would probably be the case given we're not just doing a `target.split(',')`. 

---

The only way to move before the parsing and not break is to like base64 encode the JSON (or not use json and do a different scheme), but also to make it so that it can be `parseInt`-ed. I think a `/trns` alternative is probably the answer here

server/users.js
scheibo: Ugh. Of course you're missing `settings` now

TheJetOU: my bad

server/chat-commands.js
```diff
@@ -4488,17 +4488,24 @@ const commands = {
...
 			targetRegistered = target;
 			if (commaIndex >= 0) {
 				targetRegistered = !!parseInt(target.substr(0, commaIndex));
-				targetToken = target.substr(commaIndex + 1);
+				target = target.substr(commaIndex + 1);
+				commaIndex = target.indexOf(',');
+				targetToken = target;
+				if (commaIndex >= 0) {
+					targetToken = target.substr(0, commaIndex);
+					settings = JSON.parse(target.substr(commaIndex + 1));
```
whalemer: Bit of a nitpick but this should be in a try ... catch block and you need to make sure it's an object so you don't crash when using `in` on it.

whalemer:
How do you plan on detecting if a server supports `/trn+`? If you know if a server is updated or not, can't you just include settings based off that?

scheibo:
> How do you plan on detecting if a server supports `/trn+`? If you know if a server is updated or not, can't you just include settings based off that?

Ugh yeah, youre right. Simply adding the new command has its own host of issues -_-. Really dont want to deal with `CAP` negotiation...

PR #5583, opened 2019-07-04
Don't report name after forcerename for namelocked users

whalemer:


server/chat.js
```diff
@@ -1984,10 +1984,10 @@ Chat.monitors = {};
 /** @type {Map<string, string>} */
 Chat.namefilterwhitelist = new Map();
 /**
- * Inappropriate userid : forcerenaming staff member's userid
- * @type {Map<string, string>}
+ * userids that have been previously forcerenamed
+ * @type {Set<string>}
  */
-Chat.forceRenames = new Map();
+Chat.forceRenames = new Set();
```
Zarel: There's no need not to track these just because they're unused.

Zarel:
It looks like this removes the mention of the staff member who did the forcerename? Why?

whalemer:
The mention was removed a while back because it was annoying and you can just check modlog if you need to know. If you want me to still keep the data around then I can, I guess, but afaik it's not used anywhere.

Zarel:
I'd prefer for the data to still be kept around (if only to make it easier to hotpatch).

PR #5584, opened 2019-07-04
Mafia: Implement silences

whalemer:


server/chat-plugins/mafia.js
```diff
@@ -1483,6 +1484,12 @@ class MafiaTracker extends Rooms.RoomGame {
 			}
 		}
 
+		// Silenced check bypasses staff
```
HoeenCoder: This may not be a wise idea, for example silenced staff will be unable to moderate with some commands since `this.canTalk()` will return a false. `/mute`, `/unmute`, `/roomban`, `/roomunban` for example.

whalemer: `canTalk` doesn't check the game filter if there's no message given, so basically all commands will go through.

PR #5587, opened 2019-07-05
Clarify /away error messages

whalemer:


server/chat-commands.js
```diff
@@ -656,9 +656,10 @@ const commands = {
 
 	status(target, room, user, connection, cmd) {
 		if (!this.canTalk()) return;
-		if (!toID(target)) return this.parse('/help status');
+		if (!target) return this.parse('/help status');
+		if (!/[a-zA-Z0-9]/.test(target)) return this.errorReply("Your status must contain at least one letter or number.");
```
Zarel: Why?

whalemer: Which part are you asking about?

server/chat-commands.js
```diff
@@ -676,8 +677,10 @@ const commands = {
 		if (!this.canTalk()) return;
 
 		let message = Chat.namefilter(target, user, true);
-		if (!message && toID(target)) return user.popup("Your status contains a banned word.");
-
+		if (target) {
+			if (!/[a-zA-Z0-9]/.test(target)) return this.errorReply("Your status must contain at least one letter or number.");
```
Zarel: What is this testing?

server/chat-commands.js
```diff
@@ -698,11 +701,13 @@ const commands = {
 		} else {
 			awayType = `${awayType[0].toUpperCase()}${awayType.slice(1)}`;
 		}
-		if (toID(target)) {
+
+		if (target) {
+			if (!/[a-zA-Z0-9]/.test(target)) return this.errorReply("Your status must contain at least one letter or number.");
```
Zarel: Ditto

server/chat-commands.js
Zarel: Why must statuses contain at least one letter or number?

whalemer:
Statuses consisting only of symbols don't make it through the namefilter, because of
 ``	name = name.replace(/^[^A-Za-z0-9]+/, ""); // remove symbols from start`` 

This just catches that and gives a more friendly message.

server/chat-commands.js
Zarel: Like, that's not actually an intended restriction; it was a restriction you seemingly accidentally added in #5577.

Zarel:
Sounds like a problem with the namefilter.

Zarel:
Although I do see the problem; yes, I would deal with this by fixing the namefilter.

Zarel:
(This is a good next step in addition to fixing the namefilter)

PR #5588, opened 2019-07-05
Add a 15 second timeout when warning individual users

Kaiepi:
It's possible to spam /warn in groupchats, preventing users from being
able to close the warn modal. This mitigates that by limiting the rate
you can warn a user to once every 15 seconds, which is longer than the
amount of time it takes for the warn modal to allow you to close it.


Zarel:
Relatedly, why is `globalWarn` defined so weirdly? Is there any reason not to define it as just "used by global staff"?

PR #5589, opened 2019-07-05
Data: Make Template constructor autofill Pokestar learnsets

jmclemo6:
Changes as discussed in #5579. Makes ff911e2 unnecessary.

Couldn't just make `this.learnset` default to `{}` because so much existing code checks `!template.learnset` or `template.learnset` directly, so went with a check for the template being of a Pokestar mon and then setting the learnset based on that.

### DIFF
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -644,6 +644,9 @@ export class Template extends BasicEffect implements Readonly<BasicEffect & Temp
 		this.maleOnlyHidden = !!data.maleOnlyHidden;
 		this.maxHP = data.maxHP || undefined;
 		this.learnset = data.learnset || undefined;
+		if (this.isNonstandard === 'Pokestar') {
+			this.learnset = {};
+		}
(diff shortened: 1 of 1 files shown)


Zarel:
Oh, if existing code checks `!template.learnset`, this isn't really a viable fix.

Zarel:
We might have to migrate directly to `Dex.getLearnset`.

Zarel:
Also, none of these changes _really_ make ff911e22303 unnecessary

jmclemo6:
> Also, none of these changes _really_ make [ff911e2](https://github.com/Zarel/Pokemon-Showdown/commit/ff911e2230349d7af885643ad23aa1a8d29c33dc) unnecessary

Unless more pokemon are added that have the combination of no learnset themselves + no base species that has a learnset then I believe that commit is unnecessary. 

Also yeah. As far as I know, any code that deals with forms has to check `!template.learnset` because some forms don't have a learnset identified.

Issue #5590, opened 2019-07-06
AFK/Status Follow-up


scheibo:
Proposal
---

The initial launch of AFK/statuses has revealed some design inconsistencies and opportunities for improvements.

- A user can be in one of 4 states on the server: `Online`, `Idle`, `Away`, `Offline`. A user becomes `Idle` after a period of inactivity as determined by the server and can return from the `Idle` state by performing any action whatsoever. The `Away` state only occurs manually, through use of the `/away` command (and its aliases: `/afk`, `/brb`, `/busy`), and a user *only*  leaves this state through the use of `/back`. No other action in the `Away` state causes a user to leave `Away` (including joining/leaving a room or _chatting in a chat room*_, though these would cause you to leave the `Idle` state). 
- The server protocol will reserve the first character after the `@` to be a symbol denoting the state of the user, which will currently either be `'!'` for `Idle`/`Away` and `' '` for `Online`, followed by the user's status message. The user's status is either inferred from the command used to mark the user as `Away` (`/brb` -> `!BRB`) or is whatever the user specified (`/away eating` -> `!eating`).
- The client will display statuses based on the 'state' symbol (currently `'!'` or `' '`), and the message. The state character will currently be displayed on the client as a tag (`'!'` -> `'(Away)'`, `' '` -> `'Online'`), so `'!Idle'` -> `'(Away) Idle'` and `' LF shroomish'` -> `'(Online) LF shroomish'`. Instead of tags, the client may display the state symbol as a orb  or in a different color etc. The state symbol also determines the color and user list position on the client - `'!'` grays out the user and sorts it to the bottom of the user's respective rank.
- `'/busy'` no longer has special meaning and is now simply an alias for `'/away Busy'` (meaning it also sets the `'!'` bit and grays the username). Blocking PMs and Challenges are better accomplished through the persistent toggle in the client, and `/back` should only clear the users status/away state, not unblock anything.
- Statuses should not be filtered using the name filter (which disallows things like commas), and should probably have a more lenient filter based on the chat filter which pretty much only excludes `'|'` for protocol reasons, banwords/spoilers, and anything which could be confused with a rank (eg. `/(global|room|upper|senior)\s*(staff|admin|administrator|owner|operator|mod|moderator|driver|voice|creator)/gi`)

*_This is up for debate. PMs currently work this way with the `Away` state - users can still chat and be colored in the chat but show up as gray in the userlist. This proposal would extend this to regular chat rooms as well._

[...]

PR #5591, opened 2019-07-06
Modify status protocol and afk types

whalemer:
As per the first part of #5590

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -153,18 +153,18 @@ represented by a space), and the rest of the string being their username.
 `|join|USER@STATUS` or `|j|USER@STATUS`
 
 > `USER` joined the room. Optionally, `USER` can end in `@` followed by a user status message.
-> A `STATUS` starting in `!` indicates the user is away.
+> `STATUS` will start in `!` if the user is away, or '` `' otherwise.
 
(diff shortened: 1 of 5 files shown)


whalemer:
Client PR: zarel/pokemon-showdown-client#1317

PROTOCOL.md
```diff
@@ -153,18 +153,18 @@ represented by a space), and the rest of the string being their username.
 `|join|USER@STATUS` or `|j|USER@STATUS`
 
 > `USER` joined the room. Optionally, `USER` can end in `@` followed by a user status message.
-> A `STATUS` starting in `!` indicates the user is away.
+> `STATUS` will start in `!` if the user is away, or '` `' otherwise.
```
Zarel: I don't like this change.

I think there's a sense in which it's good that a user would be able to manually set an away status by adding `!` to the beginning of a status message.

Zarel:
I think it's a little bit premature to write this much code for a proposal that's only just gone up and I haven't taken a look at yet.

Zarel:
Part of it's, I'm not sure if I should write my objections in your PRs or in #5590. These aren't even draft PRs!

whalemer:
superseded by #5597

PR #5592, opened 2019-07-06
Prevent groupchatuptime from being used without a room

whalemer:


server/chat-commands.js
```diff
@@ -896,10 +896,9 @@ const commands = {
 		`/subroomgroupchat [roomname] - Creates a subroom groupchat of the current room. Can only be used in a public room you have staff in.`,
 	],
 
-	'!groupchatuptime': true,
 	groupchatuptime(target, room, user) {
-		if (!this.runBroadcast()) return;
 		if (!room.uptime) return this.errorReply("Can only be used in a groupchat.");
```
```suggestion
		if (!room || !room.uptime) return this.errorReply("Can only be used in a groupchat.");
```

Seems like a better way to implement this.

whalemer: The command shouldn't work at all without a room since it's not flagged. I feel like it's pretty obvious that `/groupchatuptime` needs to be used in the groupchat.

Zarel: Yes. `this.errorReply("Can only be used in a groupchat.");` qualifies as "shouldn't work at all", but with a more informative error message.

server/chat-commands.js
```diff
@@ -898,8 +898,8 @@ const commands = {
 
 	'!groupchatuptime': true,
 	groupchatuptime(target, room, user) {
+		if (!room || !room.uptime) return this.errorReply("Can only be used in a groupchat.");
```
panpawn: This isn't a complete sentence; "This command can only be used in a groupchat." sounds better

Zarel: It doesn't need to be a complete sentence.

PR #5593, opened 2019-07-06
Filter statuses properly

whalemer:


whalemer:
Also did you mean to bump the max status length from 18 to 32 chars?

server/chat.js
```diff
@@ -272,6 +272,7 @@ Chat.nicknamefilters = [];
  * @param {boolean} [forStatus]
  */
 Chat.nicknamefilter = function (nickname, user, forStatus = false) {
+	if (forStatus) nickname = nickname.replace(/\|/g, '');
```
Zarel: Nicknames shouldn't contain `\|` either.

Zarel:
And yes, there's no reason to restrict statuses as severely as usernames.

server/chat.js
Zarel: (It's already enforced elsewhere, but there's no harm in also enforcing it here.)

scheibo:
It doesn't fit on the usercard...

PR #5595, opened 2019-07-07
Display Elo in rated battles

TheJetOU:
Client Pull Request: Zarel/Pokemon-Showdown-Client#1318

Suggested and approved here: https://www.smogon.com/forums/threads/3651687/

I did not implement the "make `/rank` only display the tier of the battle you're playing in" suggestion because you can use `/rank [player], [format]` or with this change it's even easier because you can just scroll up.

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -505,11 +505,11 @@ class RoomBattle extends RoomGames.RoomGame {
 
 		this.listen();
 
-		this.addPlayer(options.p1, options.p1team || '');
-		this.addPlayer(options.p2, options.p2team || '');
+		this.addPlayer(options.p1, options.p1team || '', options.p1rating);
+		this.addPlayer(options.p2, options.p2team || '', options.p2rating);
 		if (this.playerCap > 2) {
-			this.addPlayer(options.p3, options.p3team || '');
-			this.addPlayer(options.p4, options.p4team || '');
+			this.addPlayer(options.p3, options.p3team || '', options.p3rating);
+			this.addPlayer(options.p4, options.p4team || '', options.p4rating);
 		}
 		this.timer = new RoomBattleTimer(this);
 		if (Config.forcetimer) this.timer.start();
@@ -960,8 +960,9 @@ class RoomBattle extends RoomGames.RoomGame {
 	 *
 	 * @param {User | null} user
 	 * @param {string?} team
+	 * @param {number} elo
 	 */
-	addPlayer(user, team) {
+	addPlayer(user, team, elo = 0) {
 		// TypeScript bug: no `T extends RoomGamePlayer`
 		const player = /** @type {RoomBattlePlayer} */ (super.addPlayer(user));
 		if (!player) return null;
@@ -973,14 +974,9 @@ class RoomBattle extends RoomGames.RoomGame {
 				name: player.name,
 				avatar: user ? '' + user.avatar : '',
 				team: team,
+				elo: elo,
 			};
 			this.stream.write(`>player ${slot} ${JSON.stringify(options)}`);
-			if (this.rated) {
-				new Ladders.LadderStore(this.format).getRating(player.userid).then(response => {
-					const elo = Math.round(response);
-					if (this.room) this.room.add(`|elo|${player.name}|${elo}`);
-				});
-			}
 		}
 
 		if (user) this.room.auth[user.userid] = Users.PLAYER_SYMBOL;
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -974,7 +974,7 @@ class RoomBattle extends RoomGames.RoomGame {
 				name: player.name,
 				avatar: user ? '' + user.avatar : '',
 				team: team,
-				rating: rating,
+				rating: Math.round(rating),
 			};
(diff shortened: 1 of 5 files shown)


sim/SIM-PROTOCOL.md
```diff
@@ -11,6 +11,8 @@ Receiving messages
 
 The beginning of a battle will look something like this:
 
+    |elo|Anonycat|1200
```
scheibo: I feel like this should come after `|player|` introduces the player, or perhaps it should be combined with `|player|` (but turned into a separate line in the output logs)

server/room-battle.js
```diff
@@ -974,7 +974,13 @@ class RoomBattle extends RoomGames.RoomGame {
 				avatar: user ? '' + user.avatar : '',
 				team: team,
 			};
-			this.stream.write(`>player ${slot} ` + JSON.stringify(options));
+			this.stream.write(`>player ${slot} ${JSON.stringify(options)}`);
+			if (this.rated) {
+				new Ladders.LadderStore(this.format).getRating(player.userid).then(response => {
```
Zarel: The rating should already be available from the `Ready` objects. Maybe forward those, instead of redoing them here.

Zarel: Especially since this is actively creating a new `LadderStore`, which is not the intended API use and I think might mess up the ladder database.

Zarel:
I'm tentatively okay with this implementation, assuming we decide on this feature being desirable.

server/room-battle.js
```diff
@@ -973,8 +974,9 @@ class RoomBattle extends RoomGames.RoomGame {
 				name: player.name,
 				avatar: user ? '' + user.avatar : '',
 				team: team,
+				elo: elo,
```
Zarel: I would call this `rating` instead of `elo`. Notice how the entire rest of the codebase calls it `rating`.

server/room-battle.js
```diff
@@ -960,8 +960,9 @@ class RoomBattle extends RoomGames.RoomGame {
 	 *
 	 * @param {User | null} user
 	 * @param {string?} team
+	 * @param {number} elo
 	 */
-	addPlayer(user, team) {
+	addPlayer(user, team, elo = 0) {
```
```suggestion
	addPlayer(user, team, rating = 0) {
```

server/room-battle.js
```diff
@@ -960,8 +960,9 @@ class RoomBattle extends RoomGames.RoomGame {
 	 *
 	 * @param {User | null} user
 	 * @param {string?} team
+	 * @param {number} elo
```
```suggestion
	 * @param {number} rating
```

sim/SIM-PROTOCOL.md
```diff
@@ -48,6 +48,7 @@ The beginning of a battle will look something like this:
 > - `USERNAME` is the username
 > - `AVATAR` is the player's avatar identifier (usually a number, but other
 >    values can be used for custom avatars)
+> - `ELO` is the player's elo in the format they're playing. This will be `0` in unrated battles
```
Zarel:
```suggestion
> - `RATING` is the player's Elo rating in the format they're playing. This will be `0` in unrated battles
```

sim/SIM-PROTOCOL.md
```diff
@@ -39,7 +39,7 @@ The beginning of a battle will look something like this:
     |
     |start
 
-`|player|PLAYER|USERNAME|AVATAR`
+`|player|PLAYER|USERNAME|AVATAR|ELO`
```
Zarel: I'm not sure if `|player|` is the correct message type to add it, though. Note that `|player|` is also sent every time a player goes online/offline or is substituted in. It's not the sort of message type for specific information intended to be sent only once at the beginning of a battle.

TheJetOU: The only other option is a new`|rating|username|elo`. Or maybe a `|battleinfo|JSON` for data that are absolute (like ratings in this case)?

Zarel: I wonder if we already have a `rating` message type. `battle.js` _did_ start life as a PO log viewer, after all.

Zarel: Looks like no.

I'm not horribly against the rating being here. As long as it's optional. I don't really want it to be sent for every reconnect or something.

TheJetOU: It's sent as `0` on everything that's not `addPlayer`, so the client will ignore it.

TheJetOU: Or are you worried about bandwidth?

Zarel: I'm worried about readability, partially. It should be sent blank instead of as `0`.

TheJetOU: This is done here https://github.com/Zarel/Pokemon-Showdown/pull/5595/commits/247f1de42d80d3a6cdaf310dab589185bfd25cd5.

Zarel:
I'm tentatively okay with this. Anyone else have feedback?

Zarel:
I'm not super happy with committing to this protocol, but I doubt I'll find a better one over the next few days. This can be allowed to live.

PR #5596, opened 2019-07-07
Allow bots to set statuses in the global room

Kaiepi:
This makes it easier to automatically set the status on connect. Instead
of making it so users can't set their status if they can't talk at all,
make it so they can't set it if they're locked or semilocked. Also,
don't send a success response if the command was sent in the global
room.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -654,8 +654,9 @@ const commands = {
 	},
 	unblockpmshelp: [`/unblockpms - Unblocks private messages. Block them with /blockpms.`],
 
+	'!status': true,
 	status(target, room, user, connection, cmd) {
-		if (!this.canTalk()) return;
+		if (user.locked || user.semilocked) return this.errorReply("Your status cannot be updated while you are locked or semilocked.");
 		if (!target) return this.parse('/help status');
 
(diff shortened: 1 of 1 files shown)


Kaiepi:
Removed not returning any message when the status was set in the global room since that was dumb.

whalemer:
What does changing the locked check achieve, besides letting guests set statuses?

Kaiepi:
It helps prevent abuse, like putting imgur filenames or YouTube video IDs in there. It's probably not enough of an issue to merit putting it in though.

server/chat-commands.js
```diff
@@ -655,7 +655,7 @@ const commands = {
 	unblockpmshelp: [`/unblockpms - Unblocks private messages. Block them with /blockpms.`],
 
 	status(target, room, user, connection, cmd) {
```
```suggestion
	'!status': true,
	status(target, room, user, connection, cmd) {
```

I think this is also necessary for setting status in the global room.

Kaiepi: Why? It worked when I tested it

Zarel: It makes the command usable in PMs, and commands in the global room are implemented as PMs to the server.

Even if some other weirdness makes it optional, there's no reason _not_ to add this.

Kaiepi:
Updated

PR #5597, opened 2019-07-08
Refactor statuses into statusType and userMessage and improve persistence

bumbadadabum:
Taking inspiration from #5591 here. Trying not to focus on implementing any of the other changes we discussed and instead focusing entirely on how we want them to be stored on the server. Now I did take some client changes into account, but those could either be (easily) implemented before merging, or I could temporarily edit the code to work with the current client before we overhaul anything. It just makes the code uglier to keep it working for things we are 100% changing very soon, which I feel is less than ideal to show what I'm thinking.


server/chat-commands.js
```diff
@@ -662,28 +662,16 @@ const commands = {
...
 
-		let statusType = '(Online)';
-		// Should work even if users use /status with a message containing ()
-		if (user.status && user.status.includes('(') && user.status.includes(')')) {
-			statusType = user.status.slice(0, user.status.indexOf(')') + 1);
-		}
-
-		user.setStatus(`${statusType} ${target}`);
-		this.sendReply(`Your status has been set to: ${target}`);
+		user.setStatus(null, target);
```
TheJetOU: I don't think `null` triggers default parameters in functions.
```suggestion
		user.setStatus(undefined, target);
```

server/users.js
```diff
@@ -1568,26 +1571,22 @@ class User extends Chat.MessageContext {
...
-	}
 	/**
-	 * @param {string} message
+	 * @param {StatusType} type
+	 * @param {string} [message]
 	 */
-	setStatus(message) {
-		if (message === this.status) return;
-		this.status = message;
+	setStatus(type = this.statusType, message) {
```
Zarel: `setStatusType` and `setStatusMessage` should be separate functions; nothing should set both at the same time.

server/users.js
```diff
@@ -538,8 +539,10 @@ class User extends Chat.MessageContext {
 		// Used in punishments
 		/** @type {string} */
 		this.trackRename = '';
+		/** @type {StatusType} */
+		this.statusType = 'online';
```
TheJetOU: Maybe `statusType` and `userMessage` should be merged into a singular `status = {type: '', message: ''}`?

server/users.js
```diff
@@ -606,8 +609,9 @@ class User extends Chat.MessageContext {
 	 */
 	getIdentityWithStatus(roomid = '') {
 		const identity = this.getIdentity(roomid);
-		if (!this.status) return identity;
-		return `${identity}@${this.status}`;
+		const statusSymbol = this.statusType === 'busy' ? '!' : this.statusType === 'idle' ? '?' : '';
+		const status = statusSymbol + (this.userMessage || '');
```
TheJetOU: When can `this.userMessage` be a false-y value?

server/users.js
bumbadadabum: But why?

server/users.js
bumbadadabum: I suppose it can only be ''. Decided to do this just to be safe. I guess it can be removed.

server/users.js
TheJetOU: They're a bit related; a status has two properties: type and message. It feels weird to have them as different variables.

bumbadadabum: but you're putting an object inside an object with no intention to use it as anything but a tuple

Zarel: Yeah, that would decrease usability for zero advantage, as far as I can tell.

server/users.js
Zarel: I like the safety here.

server/chat-commands.js
```diff
@@ -695,51 +683,46 @@ const commands = {
...
-		if (target) {
-			if (target.length > 32) return this.errorReply(`Your status is too long; it must be under 32 characters.`);
+		if (target) this.errorReply("Setting status messages in /away is no longer supported. Set a status using /status.");
 
-			awayMessage = Chat.nicknamefilter(target, user, true);
-			if (!awayMessage) return this.errorReply("Your status contains a banned word.");
-		}
-		awayMessage = `(${awayType})${awayMessage ? ` ${awayMessage}` : ''}`;
-		user.setAway(awayMessage);
+		user.setstatusType('idle');
```
Zarel: Capitalization...

server/users.js
```diff
@@ -606,8 +609,9 @@ class User extends Chat.MessageContext {
 	 */
 	getIdentityWithStatus(roomid = '') {
 		const identity = this.getIdentity(roomid);
-		if (!this.status) return identity;
-		return `${identity}@${this.status}`;
+		const statusSymbol = this.statusType === 'busy' ? '!' : this.statusType === 'idle' ? '?' : '';
```
Zarel: This is the new protocol, and would need a client change before it could be implemented.

server/users.js
scheibo: TBH, @TheJetOU's suggestion is something I might suggest as well, and I don't think the suggestion is necessarily one to be dismissed completely out of hand. The common prefix definitely makes it seems like there is a common 'status' entity that should be treated as a unit. In fact, we basically _were_ treating status as one entity in the past when it was all encode in a string (and `!` was used as the type).

Outside of the small performance overhead*, `this.status.type` vs. `this.statusType` is mostly a matter of style. You could also have a `toString()` on the `status` object Jet proposes that would encapsulate what you're doing in `getIdentityWithStatus`. Maybe my brain is warped from 5 years of writing enterprise Java, but I don't really think this is a black and white issue. I definitely use small types like this when I see common prefixes in my own code.

*_Even the performance overhead is not always cut and dry. For a class with a lot of fields always going to be megamorphic anyway, so you actually *want*  to group the fields into small types. However, performance is not really worth considering here because this is hardly a hotspot, just hard to stop myself from always thinking about it :P_

server/chat-commands.js
```diff
@@ -662,28 +662,16 @@ const commands = {
 		target = Chat.nicknamefilter(target, user, true);
```
scheibo: Unrelated to this change, but @Zarel, how did you come up with 32? You can fit 19 W's (widest character) before exceeding the user card.

server/chat-commands.js
```diff
@@ -695,51 +683,46 @@ const commands = {
...
 		this.sendReply("You are now marked as away. Send a message or use /back to indicate you are back.");
 	},
 	awayhelp: [`/away - Marks you as away. Send a message or use /back to indicate you are back.`],
 
-	'!back': true,
-	clearstatus: 'back',
-	unaway: 'back',
-	unafk: 'back',
-	back(target, room, user) {
+	clearstatus(target, room, user) {
```
scheibo: unrelated, but Flare asked for an `/cs` alias for this.

server/chat-commands.js
```diff
@@ -752,10 +735,7 @@ const commands = {
 
 		return this.sendReply("You have cleared your status message.");
 	},
-	backhelp: [
-		`/back - Marks you as back if you are away.`,
-		`/clearstatus - Clears your status message.`,
-	],
+	backhelp: [`/clearstatus - Clears your status message.`],
```
scheibo: ??? Shouldn't this be the `clearstatushelp` and `backhelp` should be the old 'Marks you back if you are away'?

server/chat-commands.js
Zarel: I vaguely remember talking about this, but it's possible it got lost in the internet trouble.

In any case, I think there's no need to restrict statuses to a single line, and I also think there's no need to optimize for the widest characters rather than average uses.

server/users.js
Zarel: You know, if you move the protocol changes to a different PR, this could get merged a lot sooner. It should probably be done in a different PR anyway.

Zarel:
```suggestion
		const statusSymbol = this.statusType === 'busy' ? '!(Busy) ' : this.statusType === 'idle' ? '!(Idle) ' : '';
```

server/users.js
TheJetOU: If they're completely different, which I didn't know. I can see why they are different variables.

In hindsight, I could have inferred that from the code, since setting your status and adding a message is two different commands.

Zarel: @scheibo it would make sense if they were `statusType` and `statusMessage`, but they're `statusType` and `userMessage` specifically so people don't mistake `userMessage` as having anything to do with statuses – `userMessage` is a textual avatar, and makes more sense being grouped with `avatar` than `statusType`.

Zarel: I don't have a problem with holding related state in a single object per se (we do it for stats all the time), but it would take more than two members and/or some level of reuse (even just functions that take the combined object) for me to consider it.

server/users.js
```diff
@@ -606,8 +609,9 @@ class User extends Chat.MessageContext {
 	 */
 	getIdentityWithStatus(roomid = '') {
 		const identity = this.getIdentity(roomid);
-		if (!this.status) return identity;
-		return `${identity}@${this.status}`;
+		const statusMessage = this.statusType === 'busy' ? '!(Busy) ' : this.statusType === 'idle' ? '!(Idle) ' : '';
```
Zarel: Nice, is this ready to be merged?

server/users.js
```diff
@@ -606,8 +609,9 @@ class User extends Chat.MessageContext {
...
+		const status = statusMessage + (this.userMessage || '');
+		return `${identity}${status ? `@${status}` : ''}`;
```
Zarel: Nesting `` ` `` D:

```suggestion
		return `${identity}${status ? '@' + status : ''}`;
```

server/chat-commands.js
Zarel: Like, anyone literally having a status of `WWWWWWWWWWWW` is not going to be particularly concerned about readability.

server/users.js
scheibo: I believe when Jet's comment was made it was still named `statusMessage` and there was no public announcement of our decision to detach the two concepts, which is why I felt the need to elaborate. And I also made the same point in my third paragraph. I think we're all on the same page here - two fields is the way to go here.

whalemer:
You need to update the code that broadcasts room join messages to do it with status to stop desyncs.

server/users.js
Zarel: Oh, sorry, I think I missed your third paragraph. My other comment still applies – I'd really only want to group things if there were more than two things or they were obviously linked in some way, such as functions needing to pass both of them.

bumbadadabum:
I've been very busy the past few days but will try to test this PR tonight
and make some of the final changes suggested.

bumbadadabum:
I tested this and it should work nicely. More eyes are welcome.

scheibo:
shipit

server/users.js
```diff
@@ -1568,26 +1571,28 @@ class User extends Chat.MessageContext {
...
+		this.userMessage = message;
+		this.updateIdentity();
 	}
-	clearStatus() {
-		if (!this.status) return;
-		this.status = '';
+	/**
+	 * @param {StatusType} [type]
+	 */
+	clearStatus(type = this.statusType) {
```
whalemer: bit nitpicky but do we need ``setUserMessage('')`` as well as ``clearStatus()``?

Zarel: Good point, we can migrate to just using `clearStatus`.

PR #5599, opened 2019-07-09
Refactor punishments

TheJetOU:
**Changes**:
- `PunishmentEntry` now has `ips` and `userids` instead of `keys`.
- `PunishmentEntry` now has `reason` and `expireTime` instead of `rest`.
- `rest` was still kept as Zarel noted it was used for extensibility. 
- `Punishments.renderEntry` converts `PunishmentEntry` into the old format for to not ruin the databases.
- ~Removed `type` from sharedIPs because it's _always_ the same.~ Turns out it's in case there are more types in the future, so I reverted that change.

This will make my life easier when I work on migrating punishments to SQLite3

It has also another advantage in making the code more explicit for future readers.

### DIFF
--- a/server/punishments.js
+++ b/server/punishments.js
@@ -1677,16 +1677,16 @@ const Punishments = new (class {
 			let [punishType, id, expireTime, reason] = punishment;
 			let expiresIn = new Date(expireTime).getTime() - Date.now();
 			if (expiresIn < 1000) continue;
-			let userids = [...data[0]].filter(user => user !== id);
+			let alts = [...data[0]].filter(user => user !== id);
 			let ips = [...data[1]];
-			output.push({"punishType": punishType, "id": id, "expiresIn": expiresIn, "reason": reason, "userids": userids, "ips": ips});
+			output.push({"punishType": punishType, "id": id, "expiresIn": expiresIn, "reason": reason, "alts": alts, "ips": ips});
 		}
 
 		if (room.muteQueue) {
 			for (const entry of room.muteQueue) {
 				let expiresIn = new Date(entry.time).getTime() - Date.now();
 				if (expiresIn < 1000) continue;
-				output.push({"punishType": 'MUTE', "id": entry.userid, "expiresIn": expiresIn, "reason": '', "userids": [], "ips": []});
+				output.push({"punishType": 'MUTE', "id": entry.userid, "expiresIn": expiresIn, "reason": '', "alts": [], "ips": []});
 			}
 		}
 		return output;
--- a/server/punishments.js
+++ b/server/punishments.js
@@ -230,20 +230,20 @@ const Punishments = new (class {
 	}
 
 	// punishments.tsv is in the format:
-	// punishType, userid, ips, userids, expiration time, reason
+	// punishType, userid, ips/usernames, expiration time, reason
 	// room-punishments.tsv is in the format:
-	// punishType, roomid, userid, ips, userids, expiration time, reason
+	// punishType, roomid:userid, ips/usernames, expiration time, reason
 	async loadPunishments() {
 		const data = await FS(PUNISHMENT_FILE).readIfExists();
 		if (!data) return;
 		for (const row of data.split("\n")) {
 			if (!row || row === '\r') continue;
-			const [punishType, userid, ips, userids, expireTimeStr, ...reason] = row.trim().split("\t");
+			const [punishType, id, altKeys, expireTimeStr, ...reason] = row.trim().split("\t");
 			const expireTime = Number(expireTimeStr);
 			if (punishType === "Punishment") continue;
-			const keys = userids.split(',').concat(ips.split(',')).concat(userid);
+			const keys = altKeys.split(',').concat(id);
 
-			const punishment = /** @type {Punishment} */ ([punishType, userid, expireTime, reason.join('\t')]);
+			const punishment = /** @type {Punishment} */ ([punishType, id, expireTime, reason.join(`\t`)]);
 			if (Date.now() >= expireTime) {
 				continue;
 			}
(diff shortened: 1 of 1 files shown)


server/punishments.js
```diff
@@ -42,9 +42,11 @@ const AUTOLOCK_POINT_THRESHOLD = 8;
 
 /**
  * @typedef {object} PunishmentEntry
- * @property {string[]} keys
+ * @property {string[]} ips
+ * @property {string[]} alts
```
Zarel: `userids` is probably a better name than `alts` here. It's clearer in terms of _what_ about the alts are being stored, as well as in terms of including the user's original userid as well as alts'.

server/punishments.js
```diff
@@ -42,9 +42,11 @@ const AUTOLOCK_POINT_THRESHOLD = 8;
...
  * @property {string} punishType
- * @property {[number, string]} rest
```
Zarel: We should still retain `rest` as an `any[]` array. It's intended to let the punishment system be extensible.

server/punishments.js
```diff
@@ -227,20 +229,20 @@ const Punishments = new (class {
 	}
 
 	// punishments.tsv is in the format:
-	// punishType, userid, ips/usernames, expiration time
+	// punishType, userid, ips, alts, expiration time, reason
```
Zarel: You're going to change the TSV format? Without even converting the old entries?

Zarel:
You should really specify _what_ you did in your refactor.

As far as I can tell, the changes are:

- you split `keys` apart into `ips` and `userids`
- also split apart the `rest` array with names of the things you changed
- also removed the `SHARED` column of the sharedips file

server/punishments.js
```diff
@@ -469,20 +480,19 @@ const Punishments = new (class {
 	 * @param {string} note
 	 */
 	appendSharedIp(ip, note) {
-		let buf = `${ip}\tSHARED\t${note}\r\n`;
```
Zarel: This is useful futureproofing.

Zarel:
(In case it was unclear, Punishments is approximately as much my code as bumba's. We essentially worked on it together, talked to each other about most design decisions, and I did a decent amount of the writing/maintenance.)

scheibo:
> (In case it was unclear, Punishments is approximately as much my code as bumba's. We essentially worked on it together, talked to each other about most design decisions, and I did a decent amount of the writing/maintenance.)

If this was directed at me, I understand you worked on punishments and would obviously be able to review as well, I just know bumba is confident in this area of the codebase and asked for his review in case you were busy or wanted to focus on other things and delegate.

server/punishments.js
Zarel: (But probably just don't change the format at all.)

server/punishments.js
TheJetOU: You mean if `sharedips.tsv` has more types in the future? I guess that makes.

server/punishments.js
TheJetOU: Done here: https://github.com/Zarel/Pokemon-Showdown/pull/5599/commits/4cd73ac99f8aab1eaeb18fc3c4deadbb56bc9718.

server/punishments.js
TheJetOU: I'm not sure if I completely understood what you mean but I think https://github.com/Zarel/Pokemon-Showdown/pull/5599/commits/e47f4598312b804e988759f4e259c9b07b100bde solves it.

server/punishments.js
TheJetOU: Yeah, that wasn't a good call on my part. I changed `renderEntry` to convert it into the old format. https://github.com/Zarel/Pokemon-Showdown/pull/5599/commits/906a666d6e697d9d002340aeee7442a6df0d5ccc.

TheJetOU:
> I think in general you should specify what you do in refactor PRs. It would make it easier to review.

Changed the PR description to includes the changes

server/punishments.js
TheJetOU: Reverted,

server/punishments.js
```diff
@@ -227,20 +230,20 @@ const Punishments = new (class {
...
 		for (const row of data.split("\n")) {
 			if (!row || row === '\r') continue;
-			const [punishType, id, altKeys, expireTimeStr, ...rest] = row.trim().split("\t");
+			const [punishType, id, altKeys, expireTimeStr, ...reason] = row.trim().split("\t");
 			const expireTime = Number(expireTimeStr);
 			if (punishType === "Punishment") continue;
 			const keys = altKeys.split(',').concat(id);
 
-			const punishment = /** @type {Punishment} */ ([punishType, id, expireTime, ...rest]);
+			const punishment = /** @type {Punishment} */ ([punishType, id, expireTime, reason.join(`\t`)]);
```
Zarel: ?

server/punishments.js
```diff
@@ -428,7 +449,7 @@ const Punishments = new (class {
 	 * @return {string}
 	 */
 	renderEntry(entry, id) {
-		let row = [entry.punishType, id, entry.keys.join(','), ...entry.rest];
+		let row = [entry.punishType, id, [entry.ips.join(','), entry.userids.join(',')].filter(el => el), entry.expireTime, entry.reason, ...entry.rest];
```
Zarel: This converts an explicit cast into an implicit cast, and only works because the default behavior of implicitly casting an array to a string is to join by `,`.

```suggestion
		const keys = entry.ips.concat(entry.userids);
		let row = [entry.punishType, id, keys.join(','), entry.expireTime, entry.reason, ...entry.rest];
```

Zarel:
Hm, so this is currently:

- splitting `keys` into `userids` and `ips`
- explicitly listing the `reason` and `expireTime` columns in functions that don't use them

These make things more readable/explicit, but they feel weird to me and I'm not sure why. I'll defer this to @bumbadadabum.

bumbadadabum:
Looks good to me.

PR #5600, opened 2019-07-12
Tournaments: Support replacing players

whalemer:


Zarel:
Remind me to merge this as of next restart.

whalemer:
[12:58:28] @jumbowhales: though it looks like the client's been updated?
[13:03:52] &pre: yeah, Config.version = 4c4e7aad for me
[13:03:56] &pre: which is the tournament chanfge

Zarel:
Yes, I updated the client upon merging.

The client does not exactly kick everyone off PS upon updating, though. To make sure everyone's using the new client, my policy is usually to wait until next restart, because restarts force everyone to update.

PR #5601, opened 2019-07-12
Use .tour property for reporting tournament wins consistently

whalemer:


Slayer95:
? What's wrong with the current approach?

whalemer:
https://github.com/Zarel/Pokemon-Showdown/blob/5f6af9049f2ec997031783c42c31ae85ed7a32b9/server/rooms.js#L1365-L1370

rooms.js does it differently and it annoys me

It also lets you keep the room as a subroom, so roomstaff can still moderate and the like (which I forgot to implement).

whalemer:
Though on second thoughts I think I'd prefer to do this the other way, and also it's probably a bit less important than I was thinking before.

server/tournaments/index.js
```diff
@@ -991,7 +991,6 @@ class Tournament extends Rooms.RoomGame {
 	 */
 	onBattleWin(room, winnerid) {
 		room.tour = null;
-		room.parent = null;
```
Zarel: Actually, why did the old code set this to `null`, anyway? It seems unnecessary.

Slayer95:
> Though on second thoughts I think I'd prefer to do this the other way, and also it's probably a bit less important than I was thinking before.

I understand from your words that you'd rather we standardize on using ``.parent``. And I couldn't agree more to that.

PR #5602, opened 2019-07-14
Migrate Punishments to SQLite

TheJetOU:


scheibo:
```
[17:02:55] %jetou: I wanted to make it draft.
```

database/schema/punishments.sql
```diff
@@ -0,0 +1,29 @@
+-- The column rest is used to make the punishments database extensible
+
```
scheibo: primary/foreign keys?

```sql
PRAGMA foreign_keys = ON;
```

build
```diff
@@ -66,6 +66,32 @@ replace(path.join(__dirname, '.server-dist'), [
...
+try {
+	require.resolve('./database/sqlite.db');
+} catch (err) {
+	if (err.code !== 'MODULE_NOT_FOUND') throw err; // should never happen
+	databaseExisted = false;
+	console.log('Databse does not exist. Creating one...');
+	fs.writeFileSync(path.resolve(__dirname, 'database/sqlite.db'), '');
+}
+
+var sqlite = require('sqlite');
```
panpawn: ``const``

scheibo: we deliberately use es3 in this file because apparently on old versions of node it would otherwise crash instead of printing the 'we require 10+' message

TheJetOU: In another note, I just realized I was using async/await, which is newer than `.then`, so I refactored it to `.then`.

database/schema/punishments.sql
TheJetOU: What do we need primary/foreign keys in this schema for?

TheJetOU: Actually, `ip` in `shared_ips` should be a primary key, but I don't see a need for foreign keys here?

build
```diff
@@ -66,6 +66,32 @@ replace(path.join(__dirname, '.server-dist'), [
...
+
+if (!databaseExisted) {
```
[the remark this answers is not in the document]
TheJetOU: I think https://github.com/Zarel/Pokemon-Showdown/pull/5602/commits/b36608ed1546522806ec8f90fbdff9c83dfb7967 should support migrations? Though I might have missed something so feel free to look over it.

build
```diff
@@ -66,6 +66,32 @@ replace(path.join(__dirname, '.server-dist'), [
...
+
+sqlite.open('./database/sqlite.db').then(function (database) {
+	if (!databaseExisted) {
+		var schemas = fs.readdirSync('./database/schema/');
+		for (var i = 0; i < schemas.length; i++) {
+			var sql = fs.readFileSync('./database/schema/' + schemas[i]).toString();
+			database.exec(sql);
+		}
+	}
+	if (fs.readdirSync('./database/migrations/').length) {
```
scheibo: it seems like this expects `-- Up` and `-- Down` comments?

TheJetOU: Yeah, it does. But the parser is built in.

TheJetOU: I should probably add a `DATABASE.md` to document the design.

scheibo: My point is that your migrations files does not have one. I noticed thats because you consider it a 'schema', but you must migrate all your schemas, so database/schema should not exist...

TheJetOU: right.

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,32 @@
+-- Up
```
scheibo: I'd make `-- Up` and `-- Down` stand out a lot more

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,32 @@
...
+	userids TEXT,
+	expireTime INTEGER,
+	reason TEXT,
+	rest BLOB
+);
+
+CREATE TABLE IF NOT EXISTS room_punishments (
+	punishType TEXT,
+	-- roomid:userid
+	id TEXT,
```
scheibo: is this not the primary key?

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,32 @@
+-- Up
+-- The column rest is used to make the punishments database extensible
+
+CREATE TABLE IF NOT EXISTS punishments (
```
scheibo: is the key for both these tables really all fields? also, there should probably be indexes here?

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,32 @@
...
+);
+
+CREATE TABLE IF NOT EXISTS shared_ips (
+	ip TEXT PRIMARY KEY,
+	-- type is always "SHARED", but we will store it anyway in case there are more types in the future
+	type TEXT,
+	note TEXT
+);
+
+-- Down
```
scheibo: this needs to reverse `-- Up`. where are the `DROP`s?

scheibo:
As you said yourself, a `DATABASE.md` is probably a good idea to include.

---

Taking a step back though, I'm not sure this is the approach I would take. I would have structured it so that persistence is configurable (eg. options for database/file/memory) and then just have a sqlite adapter instead. cf. https://github.com/Pokemon-Showdown/Pokemon-Showdown-Stats/blob/master/src/storage.ts#L9 (doesn't have a sql implementation yet, but it was written with one in mind). I dont think its substantially more work to make sqlite an optional dep, though obviously things become simpler if you can just force everything to be sqlite. However, abstracting things behind an interface makes it more flexible in the future if we ever wanted to change databases etc.

scheibo:
Also, I should stress that abstracting out an interface for persistence will make testing much faster and easier, as you can test against a memory implementation instead of needing to initialize a database (create the file, perform all migrations) and perform expensive tear down between test case.

Zarel:
We'd have to also abstract out migrations. Which would be interesting.

Zarel:
Maybe we could design a migrations interface that doesn't suck! :D

scheibo:
> We'd have to also abstract out migrations. Which would be interesting.

Yes, but not impossible.

> Maybe we could design a migrations interface that doesn't suck! :D

Exactly.

I think this is a lot to heap on @TheJetOU (_"I just want to get Punishments off of text files and they're making me design a sensible way to handle database operations across the entire application??"_), but adding a database obviously adds complexity, and I'd like us to do it in the most thoughtful way possible.

One compromise could be to just switch Punishments to sqlite, notice the painpoints and use that experience to help design a new abstraction before moving anything else to sqlite (and redoing punishments under the new design).

Zarel:
I mean I only want to get Punishments off text files if the replacement is actually _better_. Without all that fanciness, it's just really non-obvious to me that the replacement is actually better.

TheJetOU:
After doing some thinking, this is my current solution: 

- Add a `Config.databse` option which can be `sqlite` (`database/sqlite.db`), `flat` (`.tsv`, or `.csv` etc.), or `memory` (objects and the like, only to be used for a testing server and not in prod).

- Add a `PunishmentsDatabase`  class (referenced in the `Punishments` class as `this.database`), which in each of its methods will call the correct method in `PunishmentsFlatStorage` or `PunishmentsSqlStorage` class to do the job, with respect to the `Config.database` option set by the sysop, essentially abstracting out all the logic for which database class / method to use.

- As for database conversions, there will be a `database-converter` script that will convert punishments from `tsv` to `sqlite` (or vice versa) as specified by the command line arguments (`node convert --database punishments --from tsv --to sqlite`), this will be called in `server/punishments` whenever `Config.database` is set to `sqlite` / `flat`. This intentionally has a `--database` argument should we want to convert more database to SQLite in the future.

This will allow the sysop to seamlessly switch from `tsv` to `sqlite` and use whichever with ease and to use an in-memory database for a test server.

bumbadadabum:
I really like the ideas being brought up around the database interface. I think the flexibility here is absolutely essential given the wide variety of setups people have for their PS servers and test environments, as well as our needs when it comes to persistence in different areas of the server. I'll work out some actual ideas soon of what I think but for now just want to acknowledge that this is good stuff.

database/migrations/001-punishments-schema.sql
TheJetOU: > is the key for both these tables really all fields? 

I think so? I don't really know if a user can have two punishments globally or in a room? Maybe @bumbadadabum or @Zarel can weigh in.

>  there should probably be indexes here?

The only time where we query the database we retrieve _all_ the database (to cache them in maps), so I don't really see the use of indexes here.

TheJetOU:
RE: running the conversion script automatically

Running the conversion is unfortunately more nuanced than just "convert the database to the config option set", because if I was always running the TSV database, converting the TSV database to the SQLite one doesn't really make sense because the SQLite database is not the updated one in this case. It only makes sense to run the conversion script the first time you change databases. So, it will have to be done manually by the sysop.

Zarel:
I'm confused. You should always have only the TSV database or only the SQLite database, so it's obvious which one is the updated one. I can support requiring manual conversion if you have both (or just telling the server op to delete one of them).

TheJetOU:
> I'm confused. You should always have only the TSV database or only the SQLite database, so it's obvious which one is the updated one. I can support requiring manual conversion if you have both (or just telling the server op to delete one of them).

So from what I gather, you're saying we should only be running the conversion script if they're using a TSV database but there's only a SQLite database or vice versa?

The issue with that is when the Punishments code is ran the SQLite database already exists (from `node build`), and it can't tell if it was just generated. So, should it run the conversion anyway? I think no, because on the second run of using SQLite and not deleting the TSV database, the sysop will be fucked if they forgot to delete the TSV database, because they now have two sets of wrong information.

Zarel:
> The issue with that is when the Punishments code is ran the SQLite database already exists (from `node build`), and it can't tell if it was just generated.

Whatever creates the database in `node build` should check for a TSV database and do the conversion instead if it exists.

Zarel:
It doesn't matter whether you want to create the database in `node build` or on first launch of the Punishments code, but it should be the same place you do the conversion.

build
```diff
@@ -69,6 +70,32 @@ replace(path.join(__dirname, '.server-dist'), [
 replace(path.join(__dirname, './tools/set-import'), [
 	{regex: new RegExp(`(require\\\(.*?)(lib|sim)(.*?\\\))`, 'g'), replace: `$1.$2-dist$3`},
 ]);
+replace(path.join(__dirname, './tools/database-converter'), [
+	{regex: new RegExp(`(require\\\(.*?)(lib)(.*?\\\))`, 'g'), replace: `$1.lib-dist$3`},
+]);
+
+var databaseExisted = true;
```
scheibo: Does this logic and below belong in `build` and not `server/index`?

config/config-example.js
```diff
@@ -172,6 +172,14 @@ exports.crashguardemail = null;
 	subject: 'Pokemon Showdown has crashed!'
 }; */
 
+/**
+ * Databases.
+ * This is a map of section in codebase -> database engine.
+ * Use `memory` as the database engine if this is a test server.
+ * @type {{[key: string]: string}}
```
scheibo: should the database engine be an enumeration instead of `string`?

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,46 @@
...
+--------------------------------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS punishments (
+	punishType TEXT,
+	userid TEXT,
+	ips TEXT,
+	userids TEXT,
+	expireTime INTEGER,
+	reason TEXT,
+	-- The column rest is used to make the punishments database extensible
```
scheibo: I understand this is legacy, but in a database world would we not just create a migration to change the schema if we need to extend it?

tools/database-converter/index.js
```diff
@@ -0,0 +1,60 @@
...
+child_process.execSync('node build');
+
+const PunishmentsConverter = require('./punishments-converter').PunishmentsConverter;
+
+/** @type {{[k: string]: string[]}} */
+const DATABASES = {
+	punishments: ['tsv', 'sqlite'],
+};
+
+function parseFlags() {
```
scheibo: Given the increased need for options processing I think we should just add a devDependency on minimist or similar (minimist just happens to have a [great dependency graph](http://npm.broofa.com/?q=minimist)...)

tools/database-converter/index.js
```diff
@@ -0,0 +1,60 @@
...
+	throw new Error(`Invalid database specified.\nValid databases: ${Object.keys(DATABASES).join(', ')}.`);
+}
+
+if (!DATABASES[database].includes(from) || !DATABASES[database].includes(to)) {
+	throw new Error(`Invalid database format specified.\nValid database formats: ${DATABASES[database].join(', ')}.`);
+}
+
+switch (database) {
+case 'punishments':
+	new PunishmentsConverter(from, to).convert().catch(err => {
```
scheibo: catch err throw err :|

server/punishments.ts
```diff
@@ -46,6 +51,19 @@ interface PunishmentEntry {
 	rest: any[];
 }
 
+export type PunishmentsDatabaseResponse = {
```
scheibo: These are Tables, not Databases, no?

server/punishments.ts
```diff
@@ -108,87 +126,79 @@ class NestedPunishmentMap extends Map<RoomID, Map<string, Punishment>> {
...
-		['BATTLEBAN', 'battlebanned'],
-		['MUTE', 'muted'],
-	]);
-	constructor() {
-		setImmediate(() => {
+class PunishmentsStorage {
+	static connect(type: string) {
+		let storage;
+		if (type === 'sqlite') {
+			storage = PunishmentsStorageSqlite;
```
scheibo: SqlitePunishmentsStorage or PunishmentSqliteStorage etc. Both read better

server/punishments.ts
```diff
@@ -108,87 +126,79 @@ class NestedPunishmentMap extends Map<RoomID, Map<string, Punishment>> {
...
+			storage = PunishmentsStorageTsv;
+		} else {
+			storage = PunishmentsStorageMemory;
+		}
+		storage.load();
+		return storage;
+	}
+}
+
+const PunishmentsStorageMemory = new class {
```
scheibo: This is a NoopPunishmentsStorage, not Memory. Though really, 'Storage' is a misnomer - nothing is being stored (yes, I understand we're already storing the information in JS elsewhere).

At bare minimum, the memory storage needs to store things (eg. you should be able to query in a test that the storage actually has the data stored).

Basically, right now the `Punishments` class *is* memory storage + additional logic on top. I'd separate out the storage bit, and thats the MemoryPunishmentsStorage implementation. You can then _compose_ the storage classes to get the same behavior regarding storing things in Memory and TSV/Sqlite

tools/database-converter/punishments-converter.ts
```diff
@@ -0,0 +1,241 @@
...
+		if (this.to === 'sqlite') {
+			const database = await PunishmentsConverterSqlite.databasePromise;
+			database.exec(`DELETE FROM punishments; DELETE FROM room_punishments; DELETE FROM shared_ips; DELETE FROM ip_banlist`).then(async () => {
+				// tslint:disable-next-line: no-floating-promises
+				PunishmentsConverterSqlite.convert();
+			}).catch(err => {
+				throw err;
+			});
+		} else if (this.to === 'tsv') {
+			PunishmentsConverterTsv.convert().catch(err => {
```
scheibo: yeah... why are you not just returning `PunishmentsConverterTsv.convert()` and letting the caller handle the error as desired?

build
TheJetOU: I think so, yes. I feel like migrating and running conversion is part of the 'build' process and not so much 'starting' up the server process.

scheibo: There are multiple instances where we run build but have never start a server. see `pokemon-showdown` for example. Why do I need to migrate a database to `validate-team`?

Zarel: I think that's a valid part of the build. In practice, database migrations should be rare so it shouldn't come up much.

PR #5604, opened 2019-07-14
Allow starting tournaments with a custom name

whalemer:


server/tournaments/index.js
```diff
@@ -83,7 +84,7 @@ class Tournament extends Rooms.RoomGame {
 		this.playerCap = (playerCap ? parseInt(playerCap) : Config.tourdefaultplayercap) || 0;
 
 		/** @type {string} */
-		this.format = formatId;
+		this.format = name || formatId;
```
Zarel: This seems like a bad name for this property.

whalemer: What name would you prefer?

whalemer: Also, the property is called `format` in the update message for whatever reason, and I'd _really_ prefer to not change that.

Zarel: It's fine not to change the property in the protocol, but at least changing it here seems valid. Something like `this.name`?

whalemer: But then you need to send updates like ``{format: this.name}``, which looks weird and is more confusing IMO.

Zarel: To me, it's much more obvious that `{format: this.name}` isn't sending a format, than that `{format: this.format}` isn't sending a format.

Zarel: You should also comment it, to make it extra clear.

    // note: this is not a format, it's just called format for historical reasons
    {format: this.name}

server/tournaments/index.js
```diff
@@ -83,7 +84,8 @@ class Tournament extends Rooms.RoomGame {
 		this.playerCap = (playerCap ? parseInt(playerCap) : Config.tourdefaultplayercap) || 0;
 
 		/** @type {string} */
-		this.format = formatId;
+		// This will sometimes be sent alone in updates as "format", if the tour doesn't have a custom name
+		this.name = name || formatId;
 		this.originalFormat = formatId;
```
Zarel: Apparently there's already a `format` and `originalFormat`? How did they differ in the past?

server/tournaments/index.js
```diff
@@ -83,7 +84,8 @@ class Tournament extends Rooms.RoomGame {
 		this.playerCap = (playerCap ? parseInt(playerCap) : Config.tourdefaultplayercap) || 0;
 
 		/** @type {string} */
-		this.format = formatId;
+		// This will sometimes be sent alone in updates as "format", if the tour doesn't have a custom name
+		this.name = name || formatId;
```
Zarel: In theory we might want name to default to e.g. "OU (with custom rules)" for formats with custom rules.

server/tournaments/index.js
whalemer: As far as I know, `format` will be the format unless it isn't, in which case `originalFormat` will get sent in updates as `teambuilderFormat` (not to be confused with Tournament#teambuilderFormat, which is used to store custom rules)  so the client knows what teambuilder to use.

server/tournaments/index.js
whalemer: You can't actually create a tournament with custom rules, so that'd have to be done when setting, but I don't think that's a terrible idea.

whalemer: and / or allowing you to start a tournament with custom rules

Zarel: Yeah, that's what I mean, when setting.

server/tournaments/index.js
Zarel: Oh, I see, `originalFormat` doesn't include custom rules. That's the only difference, right?

Zarel: (If so, can you add doc comments to that effect, so I don't get confused again in the future?)

Zarel:
I'm going to merge this for the restart tonight, probably. The doc comments can wait for a future push.

server/tournaments/index.js
Zarel: It looks like `teambuilderFormat` is `"gen7ou@@@!speciesclause"`, `originalFormat` is `"gen7ou"`. I _still_ don't know what `format` is for.

PR #5608, opened 2019-07-15
Stop status desyncing when joining rooms or renaming

whalemer:


Zarel:
I'm... going to be too lazy to hotpatch this. It can wait until next restart.

Zarel:
This is at least partly my fault for merging bumba's refactor so hastily.

Zarel:
Also, `getIdentityWithStatus` can now just be `name` or `name + '@!'`, right?

whalemer:
Given that (afaik) we only use the status message in the userdetails popup, yes. If we want to include them in join/leave messages then we'd need to keep it.

PR #5610, opened 2019-07-15
Create Chat.statusfilter which also performs basic rank impersonation filtering

scheibo:
- separate `Chat.statusfilter` from `Chat.nicknamefilter` instead of just using `forStatus` for more flexibility going forward. Eliminating the duplicated code present in `namefilter`, `nicknamefilter` and `statusfilter` is out of scope, but would likely be beneficial as a followup
- do basic rank impersonation filtering, which when coupled with a UI which better differentiates rank in the user card (eg. Zarel/Pokemon-Showdown-Client#1319) should help with abuse

### DIFF
--- a/server/chat-plugins/chat-monitor.js
+++ b/server/chat-plugins/chat-monitor.js
@@ -270,8 +270,7 @@ let loginfilter = function (user) {
 	if (forceRenamed) Monitor.log(`[NameMonitor] Forcerenamed name being reused: ${user.name}`);
 };
 /** @type {NameFilter} */
-let nicknamefilter = function (name, user, forStatus) {
-	const nameType = forStatus ? 'status message' : 'Pokémon nickname';
+let nicknamefilter = function (name, user) {
 	let lcName = name.replace(/\u039d/g, 'N').toLowerCase().replace(/[\u200b\u007F\u00AD]/g, '').replace(/\u03bf/g, 'o').replace(/\u043e/g, 'o').replace(/\u0430/g, 'a').replace(/\u0435/g, 'e').replace(/\u039d/g, 'e');
 	// Remove false positives.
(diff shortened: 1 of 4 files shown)


bumbadadabum:
Looks good to me!

scheibo:
It wasn't working when I actually tested - instead of replacing I do a `test` and return so 'your status contains a banned word' is reported, which I think is the behavior we want.

server/chat-plugins/chat-monitor.js
```diff
@@ -301,6 +300,42 @@ let nicknamefilter = function (name, user, forStatus) {
...
 	return name;
 };
+/** @type {StatusFilter} */
+let statusfilter = function (status, user) {
+	let lcStatus = status.replace(/\u039d/g, 'N').toLowerCase().replace(/[\u200b\u007F\u00AD]/g, '').replace(/\u03bf/g, 'o').replace(/\u043e/g, 'o').replace(/\u0430/g, 'a').replace(/\u0435/g, 'e').replace(/\u039d/g, 'e');
+	// Remove false positives.
+	lcStatus = lcStatus.replace('herapist', '').replace('grape', '').replace('scrape', '');
+	// Check for blatant staff impersonation attempts. Ideally this could be completely generated from Config.grouplist
+	// for better support for side servers, but not all ranks are staff ranks or should necessarily be filted.
+	if (/(global|room|upper|senior)?\s*(staff|admin|adminstrator|leader|owner|mod|moderator|driver|voice|operator|sysop|creator)/gi.test(lcStatus)) {
```
Slayer95: ``s/adminstrator/administrator``

Since there is "creator", please add "founder" as well, which is common in side servers.

scheibo: But do we really want to allow people to typo administrator? ;)

PR #5617, opened 2019-07-16
Typescript users

whalemer:
I'm still not 100% on what's getting properly checked or not, but it seems to look fine in my editor and tsc isn't complaining.


whalemer:
(I also took this opportunity to make user#userid an actual ID, hence the ignores when getting users in a few places)

server/punishments.js
```diff
@@ -604,6 +604,7 @@ const Punishments = new (class {
 			Punishments.ips.set(ip, punishment);
 		}
 		const [punishType, id, expireTime, reason, ...rest] = punishment;
+		// @ts-ignore
```
Zarel: ideally, the ignores should all say `cast to ID` or something, so when those files subsequently get refactored to `.ts`, we know why.

server/users.ts
```diff
@@ -407,48 +377,102 @@ class Connection {
 	onDisconnect() {
 		connections.delete(this.id);
 		if (this.user) this.user.onDisconnect(this);
-		this.user = /** @type {any} */ (null);
+		this.user = null as any as User;
```
```suggestion
		this.user = null!;
```

server/users.ts
```diff
@@ -521,74 +536,53 @@ class User extends Chat.MessageContext {
...
 
-	/**
-	 * @param {string | BasicRoom?} roomid
-	 * @param {string} data
-	 */
-	sendTo(roomid, data) {
-		// @ts-ignore
-		if (roomid && roomid.id) roomid = roomid.id;
+	sendTo(roomid: string | BasicRoom | null, data: string) {
+		if (roomid && (roomid as BasicRoom).id) roomid = (roomid as BasicRoom).id;
```
Zarel: Ideally you don't do `roomid = roomid` ever.

```suggestion
		if (roomid && typeof roomid !== 'string') roomid = (roomid as BasicRoom).id;
```

Zarel:
Wow, this is great! I'm impressed by the trick to export types from `Rooms`; @scheibo and I tried for hours and couldn't figure it out.

server/users.ts
whalemer: I think I prefer the explicit cast here because it makes it clearer you're asserting it as a type it's not

Zarel: I feel like `null!` is obviously asserting a type it's not, though.

Zarel: This is approximately the "implicitly unwrapped optional" feature in Swift – it's a property asserted non-null because it's only null during initialization and after destruction.

If you wanted to be any more explicit, I would comment that it can be null in those cases in a doc comment on the property. Doing it here isn't adding any safety/readability to my eyes.

server/users.ts
```diff
@@ -521,74 +536,53 @@ class User extends Chat.MessageContext {
...
-	/**
-	 * @param {string | BasicRoom?} roomid
-	 * @param {string} data
-	 */
-	sendTo(roomid, data) {
-		// @ts-ignore
-		if (roomid && roomid.id) roomid = roomid.id;
+	sendTo(roomid: string | BasicRoom | null, data: string) {
+		if (roomid && typeof roomid !== 'string') roomid = roomid.id;
+		roomid = roomid as string | null;
```
```suggestion
```

(The purpose of that change was for this change.)

server/users.ts
whalemer: sure, I'm not too strongly opinionated on this

server/users.ts
```diff
@@ -115,20 +108,15 @@ function merge(user1, user2) {
  * Users.get("Some dude") will give you "Some guy"s user object.
  *
  * If this behavior is undesirable, use Users.getExact.
- * @param {?string | User} name
- * @param {boolean} exactName
- * @return {?User}
  */
-function getUser(name, exactName = false) {
+function getUser(name: string | User | null, exactName: boolean = false) {
```
scheibo: nit
```suggestion
function getUser(name: string | User | null, exactName = false) {
```

server/users.ts
```diff
@@ -186,24 +169,25 @@ function findUsers(userids, ips, options = {}) {
...
  *********************************************************/
 
-let usergroups = Object.create(null);
+const usergroups = Object.create(null);
 function importUsergroups() {
 	// can't just say usergroups = {} because it's exported
-	for (let i in usergroups) delete usergroups[i];
-
+	for (const i in usergroups) delete usergroups[i];
+	// tslint:disable-next-line:no-floating-promises
```
scheibo: This should actually be `return FS(....)`, and then at the callsite on lint 193 is where we should be ignoring

server/users.ts
```diff
@@ -186,24 +169,25 @@ function findUsers(userids, ips, options = {}) {
...
 		}
 	});
 }
 function exportUsergroups() {
 	let buffer = '';
-	for (let i in usergroups) {
+	for (const i in usergroups) {
 		buffer += usergroups[i].substr(1).replace(/,/g, '') + ',' + usergroups[i].charAt(0) + "\n";
 	}
+	// tslint:disable-next-line:no-floating-promises
```
scheibo: ditto. `return` here, and then have the caller ignore (or also `return` etc)

server/users.ts
```diff
@@ -407,48 +376,102 @@ class Connection {
...
+		buf += `:${this.ip}`;
+		if (this.protocol !== 'websocket') buf += `:${this.protocol}`;
+		return buf;
 	}
 }
 
-/** @typedef {[string, string, Connection]} ChatQueueEntry */
-
-const SETTINGS = ['isSysop', 'isStaff', 'blockChallenges', 'blockPMs', 'ignoreTickets', 'lastConnected', 'inviteOnlyNextBattle'];
+type ChatQueueEntry = [string, string, Connection];
```
scheibo: super nit: i'd put a blank lint after this type defintion because otherwise it looks like its somehow relevant to `SETTINGS` when theyre both grouped together like this

server/users.ts
```diff
@@ -407,48 +376,102 @@ class Connection {
...
+	connections: Connection[];
+	latestHost: string;
+	latestHostType: string;
+	ips: {[k: string]: number};
+	latestIp: string;
+	locked: string | ID | null;
+	semilocked: string | null;
+	namelocked: string | ID | null;
+	permalocked: string | ID | null;
+	// actually {[k: ID]: string}, but typescript doesn't like that
```
scheibo: i like doing `{[id: string]: string}` to add slightly more documentation that the key is intended to be an ID, or even `{[id: /* ID */ string]: string}`.

server/users.ts
```diff
@@ -1317,12 +1268,11 @@ class User extends Chat.MessageContext {
...
 	/**
-	 * If this user is included in the returned list of alts (i.e. when forPunishment is true), they will always be the first element of that list.
-	 * @param {boolean} includeTrusted
-	 * @param {boolean} forPunishment
+	 * If this user is included in the returned list of
+	 * alts (i.e. when forPunishment is true), they will always be the first element of that list.
 	 */
-	getAltUsers(includeTrusted = false, forPunishment = false) {
-		let alts = findUsers([this.getLastId()], Object.keys(this.ips), {includeTrusted: includeTrusted, forPunishment: forPunishment});
+	getAltUsers(includeTrusted: boolean = false, forPunishment: boolean = false) {
```
scheibo: `: boolean` is probably redundant?

server/users.ts
```diff
@@ -1196,16 +1152,14 @@ class User extends Chat.MessageContext {
 	/**
 	 * Set a user's group. Pass (' ', true) to force trusted
 	 * status without giving the user a group.
-	 * @param {string} group
-	 * @param {boolean} forceTrusted
 	 */
-	setGroup(group, forceTrusted = false) {
+	setGroup(group: string, forceTrusted: boolean = false) {
```
scheibo:
```suggestion
	setGroup(group: string, forceTrusted = false) {
```

server/users.ts
```diff
@@ -969,21 +933,16 @@ class User extends Chat.MessageContext {
 		Chat.loginfilter(this, null, userType);
 		return true;
 	}
-	/**
-	 * @param {string} name
-	 * @param {boolean} registered
-	 * @param {boolean} isForceRenamed
-	 */
-	forceRename(name, registered, isForceRenamed = false) {
+	forceRename(name: string, registered: boolean, isForceRenamed: boolean = false) {
```
scheibo:
```suggestion
	forceRename(name: string, registered: boolean, isForceRenamed = false) {
```

My old conversions probably used the explicit `foo: boolean = false` pattern, but I now try to leave as much implicit as possible (like return values). Basically, my new philosophy is 'only add explicit types when the inferred type isn't going to be the same so that its clear in which cases its required'.

FWIW I usually prefer `isForceRenamed?: boolean` personally for my *new* APIs, which isnt the case here.

server/users.ts
```diff
@@ -1689,35 +1597,24 @@ function socketConnect(worker, workerid, socketid, ip, protocol) {
...
-/**
- * @param {any} worker
- * @param {number} workerid
- * @param {string} socketid
- * @param {string} message
- */
-function socketReceive(worker, workerid, socketid, message) {
-	let id = '' + workerid + '-' + socketid;
+function socketReceive(worker: Worker, workerid: number, socketid: string, message: string) {
+	const id = '' + workerid + '-' + socketid;
```
scheibo:
```suggestion
	const id = `${workerid}-{socketid}`;
```

server/users.ts
```diff
@@ -1750,59 +1647,58 @@ function socketReceive(worker, workerid, socketid, message) {
...
-	if (lines.length > (user.isStaff || (room.auth && room.auth[user.userid] && room.auth[user.userid] !== '+') ? THROTTLE_MULTILINE_WARN_STAFF : THROTTLE_MULTILINE_WARN)) {
+	if (lines.length > (user.isStaff ||
+		(room.auth && room.auth[user.userid] && room.auth[user.userid] !== '+') ? THROTTLE_MULTILINE_WARN_STAFF
+			: THROTTLE_MULTILINE_WARN)) {
 		connection.popup(`You're sending too many lines at once. Try using a paste service like [[Pastebin]].`);
 		return;
 	}
 	// Emergency logging
 	if (Config.emergency) {
+		// tslint:disable-next-line:no-floating-promises
```
scheibo: This is probably the right place to suppress, FWIW.

Issue #5569, opened 2019-07-02
Links Not Filtered in Statuses


Zyresus:
![Screenshot_20190702-133533](https://user-images.githubusercontent.com/49875377/60513138-597bea80-9cce-11e9-9fff-7b3687b5f5f5.jpg)

Zarel:
I don't see any reason why they should be.

PR #5618, opened 2019-07-16
Improve status filter

whalemer:


server/chat-plugins/chat-monitor.js
```diff
@@ -307,7 +307,7 @@ let statusfilter = function (status, user) {
 	lcStatus = lcStatus.replace('herapist', '').replace('grape', '').replace('scrape', '');
 	// Check for blatant staff impersonation attempts. Ideally this could be completely generated from Config.grouplist
 	// for better support for side servers, but not all ranks are staff ranks or should necessarily be filted.
-	if (/(global|room|upper|senior)?\s*(staff|admin|administrator|leader|owner|founder|mod|moderator|driver|voice|operator|sysop|creator)/gi.test(lcStatus)) {
+	if (/(?:\b|^)(?:staff|admin|administrator|leader|owner|founder|mod|moderator|driver|voice|operator|sysop|creator)(?:\b|$)/gi.test(lcStatus)) {
```
Zarel: is `(?:\b|^)` different from `\b`?

Zarel:
```suggestion
	if (/\b(?:staff|admin|administrator|leader|owner|founder|mod|moderator|driver|voice|operator|sysop|creator)\b/gi.test(lcStatus)) {
```

I'm _pretty_ sure this is the exact same thing.

server/chat.js
```diff
@@ -286,7 +286,17 @@ Chat.statusfilters = [];
  * @param {User} user
  */
 Chat.statusfilter = function (status, user) {
+	// Protocol character
 	status = status.replace(/\|/g, '');
+	if (!Config.disablebasicnamefilter) {
+		// e-mail address
+		if (status.includes('@') && status.includes('.')) return '';
```
Zarel: Why not?

server/chat.js
```diff
@@ -286,7 +286,17 @@ Chat.statusfilters = [];
...
+		// url
+		if (/[a-z0-9]\.(com|net|org|us|uk|co|gg|tk|ml|gq|ga|xxx|download|stream|)\b/.test(status)) status = status.replace(/\./g, '');
```
Zarel: Also why not?

server/chat.js
```diff
@@ -286,7 +286,17 @@ Chat.statusfilters = [];
...
+		// RTL control
+		status = status.replace(/\u202e/g, '');
```
Zarel: Also why not?

server/chat-plugins/chat-monitor.js
scheibo: It is.

server/chat.js
whalemer: people have asked for it and i dont really see the legitimate reason for including emails in your status

server/chat.js
whalemer: ditto, see the guy in #5569

Zarel: I don't really have a problem with people advertising in a status. It's meant for you to talk about yourself.

server/chat-plugins/chat-monitor.js
```diff
@@ -307,7 +307,7 @@ let statusfilter = function (status, user) {
 	lcStatus = lcStatus.replace('herapist', '').replace('grape', '').replace('scrape', '');
 	// Check for blatant staff impersonation attempts. Ideally this could be completely generated from Config.grouplist
 	// for better support for side servers, but not all ranks are staff ranks or should necessarily be filted.
-	if (/(global|room|upper|senior)?\s*(staff|admin|administrator|leader|owner|founder|mod|moderator|driver|voice|operator|sysop|creator)/gi.test(lcStatus)) {
+	if (/\b(?:staff|admin|administrator|leader|owner|founder|mod|moderator|driver|voice|operator|sysop|creator)\b/gi.test(lcStatus)) {
```
Zarel: I'm going to split this into a separate commit because the rest of this PR is controversial to me.

server/chat.js
AustinXII: Isn’t it odd that advertising in PMs and rooms will get you locked but not statuses though? It seems like it should be consistent

scheibo: passive vs. active

whalemer: Advertising in your name will get you filtered or FRd, but not statuses?

scheibo: I mean, I think its slightly inconsistent as well, but I was just trying to provide what I believe Zarel's argument would be. However, name/PMs/rooms are all in your face advertisements, you need to go out of your way to see someones status and so its not necessarily the same thing as the others.

`.xxx` domains though... - are ANY of those ones we would want in someones status lol?

Zarel: `.xxx` domains should already be filtered out by the other filters.

Zarel: Advertising in private rooms will not get you locked.

Neither will advertising to your friends in PMs.

I consider status messages roughly on the level of those.

Zarel: In particular, there's a difference between advertising _after_ someone asks, "I want to hear what you have to say", and advertising to someone who did not and has no intention to. The purpose of the status message is advertising. Basically any use of the status message is against chat rules. I would also lock you if you spammed "I'm going afk to eat dinner" in PM to a bunch of people who weren't talking to you, and you would at least get a warning if you said that in most public rooms you weren't actively participating in.

Heck, if I ask someone in a public room "what are you working on?" and they say "I'm programming for this private server asdf.psim.us", I would be an asshole if I locked them. That's not advertising.

scheibo:
Please reopen if this is still something you feel like pursuing.

PR #5620, opened 2019-07-17
Typescript chat

whalemer:
There's still a few things I want to do with this, I'm just making this PR so I can test it, since my tslint is doing some really weird things I don't understand.


whalemer:
I think that's everything I wanted to do

server/chat.ts
```diff
@@ -24,26 +24,59 @@ To reload chat commands:
...
-/** @typedef {(this: PageContext, query: string[], user: User, connection: Connection) => (Promise<string | null | void> | string | null | void)} PageHandler */
-/** @typedef {{[k: string]: PageHandler | PageTable}} PageTable */
+type Room = import('./rooms').Room;
+type ChatRoom = import('./rooms').ChatRoomType;
+type GameRoom = import('./rooms').GameRoomType;
+type BasicChatRoom = import('./rooms').BasicChatRoomType;
 
-/** @typedef {(this: CommandContext, target: string, room: ChatRoom | GameRoom, user: User, connection: Connection, cmd: string, message: string) => (void)} ChatHandler */
-/** @typedef {{[k: string]: ChatHandler | string | true | string[] | ChatCommands}} ChatCommands */
+type User = import('./users').User;
```
Zarel: Ideally we have `Users` as a global type, as before.

Zarel: But I guess that's something I can just personally fix later.

whalemer: That just involves putting that typedef in `globals.ts`, right?

Zarel: Yeah.

server/chat.ts
```diff
@@ -1449,46 +1323,36 @@ Chat.CommandContext = CommandContext;
...
 	if (onlyRecipient) return onlyRecipient.send(buf);
 	user.send(buf);
 	if (pmTarget !== user) pmTarget.send(buf);
 	pmTarget.lastPM = user.userid;
 	user.lastPM = pmTarget.userid;
-};
+}
 
-Chat.package = {};
+export let packageData = {};
```
Zarel: Wait, why was this renamed?

server/index.js
```diff
@@ -94,7 +94,7 @@ global.LoginServer = require('../.server-dist/loginserver').LoginServer;
 
 global.Ladders = require('./ladders');
 
-global.Chat = require('./chat');
+global.Chat = require('../.server-dist/chat');
```
Zarel: Perhaps export one big object named `Chat` instead of a bunch of little things?

server/chat.ts
whalemer: `package` is a restricted word, so you can't call a variable that.

server/index.js
whalemer: I'm not sure if that really achieves much, especially since (afaik) you'd still need to export the types separately. "Exporting a bunch of little things" doesn't really seem bad to me, personally.

server/chat.ts
Zarel: Fair.

server/index.js
Zarel: You wouldn't have to. I believe the correct approach is to export a namespace with the types.

Zarel: The reason it's bad is because it makes the APIs inconsistent. Which means you have to think while importing it, what the correct way to import is, instead of there always being one correct way you can rely upon.

whalemer: maybe putting the utility string functions (html, plural, count, etc) into a static class, à la tools?

whalemer: >The reason it's bad is because it makes the APIs inconsistent.

Don't we only use the way of exporting a big object elsewhere because we need it for the `Object.assign(function, object)` pattern?

Zarel: ...no? Chat never even used the `Object.assign(function, object)`. I just did it to make one API nicer. `Users` and `Rooms` were a big object even back when the API was `Users.get()` rather than `Users()`.

scheibo: The `Object.assign(function, object)` pattern can (should) be removed, I just havent gotten around to it yet

whalemer: I feel like `Users` is the exception here, since you need to import it as `require('./users').Users`

Zarel: I made every file work that way after they were converted to TypeScript, though.

Zarel: `Users` may be the exception compared to other `.js` files, but not to other `.ts` files.

whalemer:
>[11:24:07] #Zarel: the namespace can be exported from globals.ts
[11:24:12] #Zarel: sucrase doesn't need to know about it
[11:24:18] @ peach: what do you mean?
[11:24:41] #Zarel: um
[11:24:47] #Zarel: you can also just not do it
[11:24:52] #Zarel: I can just do it myself
[11:25:04] #Zarel: that might be easier because I'd also be able to see exactly what problems come up

cough cough, currently waiting on you to do something with this since I still don't know what you mean.

Zarel:
Yeah I was guessing from the lack of progress that this was blocked on that.

Zarel:
Can you rebase and update to the latest? I'll get to this when I have time.

whalemer:
rebased

scheibo:
@Zarel ping

whalemer:
bump please, I don't want to rebase this any more than absolutely necessary

Zarel:
Okay, I'll prioritize this.

whalemer:
Is there a reason we initialise Chat.commands and the like as undefined, and then reassign them to an object? It'd be a lot cleaner if we just initialised them as an object, imo.

server/chat.ts
```diff
@@ -153,23 +153,23 @@ class PatternTester {
...
  * Load command files
  *********************************************************/
 
-export let baseCommands: ChatCommands = undefined!;
-export let commands: ChatCommands = undefined!;
-export let basePages: PageTable = undefined!;
-export let pages: PageTable = undefined!;
-export const destroyHandlers: (() => void)[] = [];
+const baseCommands: ChatCommands = undefined!;
+const commands: ChatCommands = undefined!;
```
Zarel: Is this actually used anywhere anymore?

server/chat.ts
```diff
@@ -233,142 +239,114 @@ Chat.namefilter = function (name, user) {
...
-		return (function (strings, ...keys) {
-			Chat.tr(language, strings, ...keys);
+		return ((fStrings: TemplateStringsArray | string, ...fKeys: any) => {
+			tr(language, fStrings, ...fKeys);
 		});
 	}
 
 	// If strings is an array (normally the case), combine before translating.
-	let string = Array.isArray(strings) ? strings.join('${}') : strings;
+	const trString = Array.isArray(strings) ? strings.join('${}') : strings as string;
```
Zarel: Does this actually need to be cast? I thought TS supported doing it automatically.

whalemer: according to me deleting the cast and checking, yes

Zarel:
> Is there a reason we initialise Chat.commands and the like as undefined, and then reassign them to an object? It'd be a lot cleaner if we just initialised them as an object, imo.

I'd presume so bugged access-before-initialization would crash, or something?

Zarel:
Cool, this is definitely mergeable.

PR #5622, opened 2019-07-17
Remove Idle popup

scheibo:
This was originally added in the original design when we were actively logging Idle users out of their account. Currently it serves mostly to explain what Idle is, but mostly it just annoys users.


Zarel:
@bumbadadabum What was your plan here?

Zarel:
Honestly, it's correct that this popup definitely doesn't achieve the original discussed point, so it's probably safe to remove.

PR #5623, opened 2019-07-17
Introduce a blitz randbats format

scheibo:
_give the people what they want..._

I know your original concern was that its hard to be given a random team and have to learn the sets etc with a Blitz timer, but apparently thats 'part of the challenge' and what people actually want

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -24,6 +24,12 @@ let Formats = [
 		ruleset: ['PotD', 'Pokemon', 'Sleep Clause Mod', 'HP Percentage Mod', 'Cancel Mod'],
 	},
 	{
+		name: "[Gen 7] Random Battle (Blitz)",
+		mod: 'gen7',
+		team: 'random',
+		ruleset: ['PotD', 'Pokemon', 'Sleep Clause Mod', 'HP Percentage Mod', 'Cancel Mod', 'Blitz'],
+	},
+	{
 		name: "[Gen 7] Unrated Random Battle",
 
 		mod: 'gen7',
--- a/config/formats.js
+++ b/config/formats.js
@@ -27,7 +27,7 @@ let Formats = [
 		name: "[Gen 7] Random Battle (Blitz)",
 		mod: 'gen7',
 		team: 'random',
-		ruleset: ['PotD', 'Pokemon', 'Sleep Clause Mod', 'HP Percentage Mod', 'Cancel Mod', 'Blitz'],
+		ruleset: ['[Gen 7] Random Battle', 'Blitz'],
 	},
 	{
 		name: "[Gen 7] Unrated Random Battle",


Zarel:
It's probably a bad idea to give people a thing just because they ask for it... How much demand is there?

config/formats.js
```diff
@@ -23,6 +23,12 @@ let Formats = [
 		team: 'random',
 		ruleset: ['PotD', 'Pokemon', 'Sleep Clause Mod', 'HP Percentage Mod', 'Cancel Mod'],
 	},
+	{
+		name: "[Gen 7] Random Battle (Blitz)",
+		mod: 'gen7',
+		team: 'random',
```
Zarel: `team` should probably also be settable by `ruleset`.

config/formats.js
```diff
@@ -23,6 +23,12 @@ let Formats = [
...
+		ruleset: ['PotD', 'Pokemon', 'Sleep Clause Mod', 'HP Percentage Mod', 'Cancel Mod', 'Blitz'],
```
Zarel:
```suggestion
		ruleset: ['[Gen 7] Random Battle', 'Blitz'],
```

`ruleset` provides a very easy way to inherit rules from other formats, just sayin'.

config/formats.js
scheibo: I have a WIP branch to make more things ruleset-settable, but OOS for this

scheibo:
> It's probably a bad idea to give people a thing just because they ask for it... How much demand is there?

Any worse idea than adding a format which cannibalizes the OU playerbase on a whim? ;)

VineST:
I have heard many informal requests for this. I was anticipating it would wait on seeing how popular OU blitz is, but this works too!

Zarel:
> Any worse idea than adding a format which cannibalizes the OU playerbase on a whim? ;)

Hey, I'm the BDFL here, only I'm allowed to do stuff on a whim! Imagine the chaos if _anyone_ could just do stuff on a whim!

Zarel:
I don't even see a thread in Suggestions! I was going to let the OU Blitz format run for a bit more, polish up the implementation a bit more (it's already a lot more polished than launch), before adding other ladders.

scheibo:
What does polish entail? Server side its pretty much only timer tweaks (perhaps autostarting the time in Blitz?). Client side I'd much rather a 'minimal animations' mode than 'hyperfast', but we've had that discussion on https://github.com/Zarel/Pokemon-Showdown-Client/pull/1285

I think its worth questioning whether we want to bring this other ladders in the first place - it does definitely impact metagames and I think it bears more discussion. Random Battles seems like it would have been a better place to start with the Blitz trial given its already a PS-only meta without as wide reaching implications on things like usage stats.

Client UI wise, doubling our metas (`Foo` + `Foo (Blitz)`) seems poor - should '(Blitz)' not be a checkbox or something in the UI?

Zarel:
Yes, autostarting is planned. "Hyperfast" _is_ the minimal animations mode I was talking about – I'm not sure what else you'd be looking for in a minimal animations mode.

scheibo:
>  "Hyperfast" is the minimal animations mode I was talking about – I'm not sure what else you'd be looking for in a minimal animations mode.

The equivalent of turning off Battle Animations on the cartridge, but not having everything be completely static. IIRC, turning off Battle Animations doesnt make the sprites stop moving, and shows 'impact' so you can detect when a move hits or misses, but otherwise theres no frills

scheibo:
> I don't even see a thread in Suggestions!

We gotchu fam: https://www.smogon.com/forums/threads/introduce-a-blitz-randbats-format.3652592/ ;)

Slayer95:
+1 to adding Blitz version of the most popular PS format.
+1 to waiting on further feedback from OU (Blitz) players.

I'd expect the polishment process to involve allowing for up to 10-20 seconds more of Team Preview in Blitz random formats. That would be some sort of balancing the planning and blitz aspects of the format.

Zarel:
Apparently `forcetimer` forces the timer before the first turn, which will also need to be fixed.

TheImmortal:
OU was one thing, which I am fine with btw, but I don't agree with extending it to other formats. No format has the capacity of OU to be able to support two ladders.

Regarding Random Battle, the whole point of it is to jump on and instantly find games. With this we will have _three_ exactly the same ladders, but one ranked, one unranked, and one faster. To add on to the point that you yourself brought up @Zarel that Random Battle is about learning and working with the team you're given. Are we literally adding clicking ladders now? Are our standards that low? Strong no from me. Listening to the community is nice and all but it takes a good leader to say no.

Slayer95:
@TheImmortal , to clarify, are you opposed to adding the _format_? Or opposed to adding a Blitz Randbats _ladder_?

A ladder for this certainly feels premature. However I wonder on what basis you claim that OU will be _more_ capable of supporting a Blitz ladder than Randbats? Are randbats not the most popular format anymore?

Anyway, having this as a challenge-only format would be great IMO. Hell, it could be implemented as a challenge setting... (Something something custom custom games...)

Slayer95:
Note also that since Blitz games are expected to be shorter, the ladder will need a much lower amount of players in order to achieve the same average waiting time. As for how much shorter, and how much lower... Well, the need to figure that out would be another reason to wait for a while before adding the format and/or ladder.

scheibo:
Cards on the table: I don't personally like the Blitz format at all. I agree its 'clicking buttons', and completely different to the style of play I myself prefer. I'd prefer no ladders for Blitz period. But I am not necessarily 'the user', and I don't get to tell users what format to prefer.

I don't even necessarily agree that OU has the capacity for a Blitz ladder, but if it did, then Random Battles would have even more capacity because its a larger format. Agreeing with an OU Blitz and not a Random Battle Blitz ladder on the grounds of size seems to point to bias.

As for Random Battles being about "learning the team" - that's your interpretation of the format. Players apparently don't necessarily feel the same way. My interpretation of OU does not involve 15 seconds per turn, but hey, for some people this is in fact what they want.

Adding a `Foo (Blitz)` ladder could cause more users to play `Foo` than before, could completely wipe out `Foo` in favor of `Foo (Blitz)` (perhaps `Foo (Blitz)` is more enjoyable metagame for people than `Foo`, even though it doesn't appeal to me), or it could end up with a split playerbase where the same number of people are playing, just spread more thin. The third case is clearly the most worrisome, but its not really something we can necessarily predict ahead of time, and its something which can be fixed by simply removing one of the ladders.

TheImmortal:
Blitz ladder. If we removed unranked I’ll reluctantly allow it because every time someone mentions unranked it’s in the context of why is that a thing.

I meant OU is the only good choice amongst all real metas. My objection has to do with there already being two random battle ladders of the exact same meta plus what I mentioned in my post above.

Zarel:
Bring back Hackmons and you have a deal. ;)

TheImmortal:
Definitely not gonna ever be a ladder. I can consider a challenge format if that works.

Zarel:
I still think you're optimizing too much for "ladder quality" over "what the users want"

TheImmortal:
I’m not gonna give you a half ass answer as to why not because it’s complicated. Ladders may not matter to you but we don’t make decisions based on a whim. Should I consider a challenge format or not?

TheImmortal:
Also that is completely unrelated to this PR. Please explain why three ladders for the same format is a good idea. And why unrated needs to even exist in the first place.

Slayer95:
There is an issue affecting gamers called "[Ladder anxiety](https://www.areseia.com/what-is-ladder-anxiety-and-can-we-overcome-it/)y". While it's kinda unheard of in the competitive community of Pokémon, I am positive there is a minority of players that should be affected by it. Casual, unranked, formats prevent it.

That's why many games offer an unranked experience. However, there are different approaches to it. For example, in Starcraft 2 _both_ Ranked and Unranked players go into the same matchmaking pool, and their MMR values do get updated. The only difference is that there is no way in which Unranked players can view their rating.

IMO we can effectively replace the current system of randbats' split ladders with the Starcraft's approach. Then, players could opt-in to unranked mode through a checkbox the battle search menu, just like it's (planned?) to add a similar checkbox for "Search muted".

VineST:
Just putting it out there that the Smogon suggestion thread has reached almost 70 likes within 10 hours.

scheibo:
> IMO we can effectively replace the current system of randbats' split ladders with the Starcraft's approach. Then, players could opt-in to unranked mode through a checkbox in the battle search menu

How exactly does this work? If I play every other game alternating between ranked and unranked, the backend calculates my rankings the same way as if every game was ranked? In the same scenario, if I play every ranked match but immediately forfeit every unranked match, I'm going to be penalized for the unranked battles as theyll affect my ranked rating, no?

> Also that is completely unrelated to this PR. Please explain why three ladders for the same format is a good idea. And why unrated needs to even exist in the first place.

Unrated vs. rated is a separate conversation - Blitz vs. non Blitz are in fact not the same format in the same way Chess and Blitz Chess are not the same - different aspects of play are weighted more heavily. 

> I’m not gonna give you a half ass answer as to why not because it’s complicated.

Any chance you can 'full ass' an answer or provide a link to past discussion on this :)? I don't know the history here, but I don't know how I feel about being an arbiter of what formats people are allowed to enjoy? Anything Goes or Hackmons might be objectively less competitive etc than Ubers/BH, but if there is a sizeable player base it seems somewhat paternalistic to decide which format players should be able to play? We support a glorified rock-paper-scissors meta in 1v1...

Zarel:
> How exactly does this work? If I play every other game alternating between ranked and unranked, the backend calculates my rankings the same way as if every game was ranked? In the same scenario, if I play every ranked match but immediately forfeit every unranked match, I'm going to be penalized for the unranked battles as theyll affect my ranked rating, no?

It's the equivalent of having separate usernames for the ranked and unranked modes. The ranks are tracked separately.

> Any chance you can 'full ass' an answer or provide a link to past discussion on this :)? I don't know the history here, but I don't know how I feel about being an arbiter of what formats people are allowed to enjoy? Anything Goes or Hackmons might be objectively less competitive etc than Ubers/BH, but if there is a sizeable player base it seems somewhat paternalistic to decide which format players should be able to play? We support a glorified rock-paper-scissors meta in 1v1...

I don't remember getting a more detailed answer than "because Hackmons is a bad meta". Which I never personally liked as a reason for no ladder, but at the time I had enough other things to worry about that I didn't have time to fight it.

But yes, it should at least be in the Challenge menu.

scheibo:
> It's the equivalent of having separate usernames for the ranked and unranked modes. The ranks are tracked separately.

Great, big fan of this!

TheImmortal:
Okay @scheibo @Zarel. Hackmons does not have a ladder for three reasons.

1) It hasn't earned one
Ladders are earned and not just given on a whim. The userbase for Hackmons is quite small, a fraction of the Balanced Hackmons players. Ladders are earned through their popularity primarily. There are ways that can be seen, such as the activity on the forum, the demand for tours, etc...

2) There isn't a need for one
Some might call it "fun" but it _is_ a bad metagame. It's broken things checking broken things, such as No Guard OHKO vs Wonder Guard. Such a metagame would never be a part of the OM tournament circuit, thus there isn't a need for it to be playable on the main server.

3) It divides one of our already small communities
OMs combined can barely compete with the size of individual official tiers, let alone a single metagame such as BH. The only players that would be interested in playing Hackmons would be existing BH players. You're not going to be an AAA or STABmons or Mix and Mega, or any other OM player and suddenly decide to play Hackmons because it has no bans. Hackmons being a playable format takes a fraction of players away from our already small community, and we can't afford such a split.

Hypothetically if the ladder were added based on your whims, I have to answer to the OM community. Why was a ladder added and why was it Hackmons. I don't have any answer because that should never happen. You don't have anyone to answer to so it seems trivial to you, Zarel. You wake up and think something would be cool but don't care about its repercussions. I've been managing the formats ecosystem for years, not just Other Metagames, and I'd appreciate being kept in the loop if you plan to change anything. I didn't say anything earlier because OU Blitz was fine. But please don't add more without consulting the leader of the relevant section (for example: UU leader if you wanted to add UU Blitz). On the topic, I still say I'll reluctantly accept it if it replaced Unranked because nobody really cares about that (you can literally go an alt if points matter to you that much) and we'd retain only two of the same ladders, just like OU.

Zarel:
> Ladders are earned and not just given on a whim. The userbase for Hackmons is quite small, a fraction of the Balanced Hackmons players.

The userbase for Hackmons was larger than Balanced Hackmons at the point you removed it.

> It divides one of our already small communities

You said yourself that Hackmons is a "bad" metagame. I don't think it attracts the same kind of people who play "good" metagames.

We had the same problem issue with Monotype. Their audience is pretty different from the existing OM community, as far as I can tell, because people play it because it's fun, rather than because it's good.

TheImmortal:
What problem did we have with Monotype exactly? Monotype is a balanced metagame, and isn’t even an OM but an official Smogon metagame.

Zarel:
If I'm remembering correctly, Monotype wasn't accepted as a meta until way later.

TheImmortal:
That topic aside, I don’t believe you’ve commented on why we need three of the same formats, or on the suggestion of blitz simply replacing unranked. Slayer’s suggestion works too, I guess.

Zarel:
I'm fine with putting ranked and unranked on the same ladder.

Zarel:
Well, not _completely_ fine. I'm worried unranked players will goof off. It would probably lead to a lot more ladder inflation.

Zarel:
I'm also not remotely worried about ranked randbats not having enough players.

Zarel:
One thing I was thinking about was to have the "Unranked" and "Blitz" options separate from the Formats dropdown – it would just be an option available after clicking on the format.

But that might make them harder to find... the eternal dilemma...

scheibo:
> One thing I was thinking about was to have the "Unranked" and "Blitz" options separate from the Formats dropdown – it would just be an option available after clicking on the format.

Why not just Unranked and Blitz checkboxes like Slayer suggested, outside of the dropdown and always present? I assume youre worried about having the users make too many decisions up front?

Zarel:
> Why not just Unranked and Blitz checkboxes like Slayer suggested, outside of the dropdown and always present? I assume youre worried about having the users make too many decisions up front?

I'm afraid most ladders won't have enough activity to support a separate Blitz ladder.

Zarel:
(Unranked can be a checkbox no problem though.)

TheImmortal:
Just unranked as a checkbox is cool. Blitz can be for OU and Random Battle for now and if any tier leader thinks their tier is large enough to support a blitz, we can fulfill their request.

scheibo:
Is this blocked on #5648? Also, re "polish up the implementation a bit more" - what more did you want on this front?

Zarel:
The main thing I would block an blitz randbats on is an implementation for auto-timer. TI seems to also want to block on the unranked thing, but I don't particularly care.

scheibo:
When I was asking if it was blocked on unrated I actually meant "would we be willing to remove the existing unrated random battle ladder (to eventually be replaced by our more generalized implementation) to add blitz?" (no idea how the userbase would feel about one vs. the other). I have no intention of adding 3 ladders after TI's vocal and reasonable opposition.

Zarel:
Oh, I think I misread this comment last time I opened this. The I would not be willing to remove the existing unrated randbat ladder; removing features, even temporarily, is generally worse than delaying new features.

TheImmortal:
Shower thought: what if we had Blitz on the same ladder as the normal format? Just match two Blitz players with each other through a Blitz checkbox. Then, theoretically, every format would support Blitz. I mean, I don't see an issue if both players are agreeing to a shorter timer; it's still the same metagame.

Zarel:
@TheImmortal That sounds like the same thing as a separate ladder, except maybe with different UI.

I don't think I like it.

PR #5627, opened 2019-07-18
Change how status type is encoded in the protocol

scheibo:
Not to be merged until the client can handle the new protocol. We might need to also give bot developers a heads up as well? Discussion is all on #5590

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -598,13 +598,11 @@ class User extends Chat.MessageContext {
 	}
 	getIdentityWithStatus(roomid: string = '') {
 		const identity = this.getIdentity(roomid);
-		const status = this.statusType === 'online' ? '' : '@!';
+		const status = this.getStatus();
 		return `${identity}${status}`;
 	}
 	getStatus() {
-		const statusMessage = this.statusType === 'busy' ? '!(Busy) ' : this.statusType === 'idle' ? '!(Idle) ' : '';
-		const status = statusMessage + (this.userMessage || '');
-		return status;
+		return this.statusType === 'idle' ? '@!' : this.statusType === 'busy' ? '@' : /* online */ '';
 	}
 	authAtLeast(minAuth: string, room: BasicChatRoom | null = null) {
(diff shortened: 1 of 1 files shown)


server/users.ts
```diff
@@ -598,13 +598,11 @@ class User extends Chat.MessageContext {
...
 		const identity = this.getIdentity(roomid);
-		const status = this.statusType === 'online' ? '' : '@!';
+		const status = this.getStatus();
 		return `${identity}${status}`;
 	}
 	getStatus() {
-		const statusMessage = this.statusType === 'busy' ? '!(Busy) ' : this.statusType === 'idle' ? '!(Idle) ' : '';
-		const status = statusMessage + (this.userMessage || '');
-		return status;
+		return this.statusType === 'idle' ? '@!' : this.statusType === 'busy' ? '@' : /* online */ '';
```
TheJetOU: Whether Idle should be `@` or `@?` is up on the fence still, I think?
```suggestion
		return this.statusType === 'busy' ? '@!' : this.statusType === 'idle' ? '@' : /* online */ '';
```

scheibo: No, Zarel is pretty vehement about Busy as `@` and Idle as `@!`, so the code as written should correctly reflect his preference.

TheJetOU: Oh, I reverted it in my brain and thought he wanted `@!` for busy and `@` idle, my bad.

scheibo: Yeah, *we* want `@!` for Busy which is why its so confusing, but to Zarel, Busy is ⊥, and I'm tired to arguing.

TheJetOU:
Hm, looks like you'll also need to change:

-  `/cmd userdetails` https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-commands.js#L4485-L4486 (it's already broken anyway, because `targetUser.away` is not a valid property in `User`...) to use `targetUser.userMessage` in the `status` key because `getStatus` doesn't return the status anymore.

- `PROTOCOL.md` to reflect the newest changes.

Issue #5628, opened 2019-07-18
Fix Follow the Leader problem text


ColeFrench:
When validating Follow the Leader teams, problems are displayed with the leader's name, even if the problem is for a follower.

Related: #5566

TheImmortal:
Closing this for now. If the metagame returns, it'll probably have to be re-coded anyway as PS! has made many engine changes since then.

PR #5629, opened 2019-07-18
Improve Follow the Leader validation

ColeFrench:
Among fixing #5628, this PR removes a second team validation which I believe to be redundant—relevant lines are lines [638](https://github.com/Zarel/Pokemon-Showdown/commit/d7cc1dfec5d7dc8987f9efd9dab15bf043220a25#diff-51750d8b764d10b745cbaea560131ce8R638) and [641](https://github.com/Zarel/Pokemon-Showdown/commit/d7cc1dfec5d7dc8987f9efd9dab15bf043220a25#diff-51750d8b764d10b745cbaea560131ce8R641) of a previous commit.

@urkerab, maybe you can confirm.

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -656,16 +656,15 @@ let Formats = [
 			}
 			const setIds = [set.name, set.species];
 			[set.name, set.species] = teamHas.leaderIds;
-			let problems = this.validateSet(set, teamHas) || [];
-			problems = problems.map(problem => problem.replace(teamHas.leaderIds[0], setIds[0]));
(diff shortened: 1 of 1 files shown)


urkerab:
I think the second validation is necessary to make mega stones work correctly. Also, `checkLearnset` gets called once per move in the set, so I don't think that change will work.

urkerab:
(By which I mean that a set with a mega forme and matching stone should validate but the forme should be changed back to the non mega forme.)

ColeFrench:
Thanks, I understand now. I'll fix that.

PR #5630, opened 2019-07-18
Prevent Idle state from propagating

scheibo:
Currently users are complaining about being `Idle` right from startup until they click a username/chat.

Merging users seems active enough to wipe the `Idle` state - do merges ever happen without activity on the user's end?

We may also want to clear `Idle` in `markDisconnected`? WDYT?

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -1058,7 +1058,7 @@ class User extends Chat.MessageContext {
 		this.latestHost = oldUser.latestHost;
 		this.latestHostType = oldUser.latestHostType;
 		this.userMessage = oldUser.userMessage || this.userMessage || '';
-		this.statusType = oldUser.statusType !== 'online' ? oldUser.statusType : this.statusType;
+		this.statusType = oldUser.statusType === 'busy' ? oldUser.statusType : this.statusType;
 
 		oldUser.markDisconnected();
 	}


Zarel:
Clearing it on merge should be good enough.

Zarel:
(This code does not clear it on merge.)

scheibo:
> Clearing it on merge should be good enough.
> (This code does not clear it on merge.)

Little confused about this? We deliberately wanted to preserve `Busy` on merge, I've modified it now to _only_ preserve `Busy`.

Zarel:
This code would merge `Idle + Online -> Online`, which is good, but it would merge `Online + Idle -> Idle`, which is bad, and also the main case you were _trying_ to fix.

bumbadadabum:
how is the new user ever idle when you're merging?

Zarel:
The new user could have been offline for half an hour. That would do it.

Zarel:
`oldUser` was named sort of randomly; it essentially means `otherUser`.

Let me actually run through the login process.

Say you're logged into the user "bumbadadabum". You close PS, and open it again half an hour later.

1. When you open PS and connect to the server, you are a Guest.

2. After the login handshake, you merge.

3. The old bumbadadabum user session is `this`. The new Guest account is `oldUser`.

(The name `oldUser` is actually the _newer_ session. It's called `old` because it's the one that goes away after merge - it's not going to be the new, merged user.)

scheibo:
https://github.com/Zarel/Pokemon-Showdown/commit/ba4e0254d810d360dbe367c86deaea934b1505bd

PR #5631, opened 2019-07-19
Add easy method of undoing /trustuser

asgdf:
The change in users.js is so you can actually untrustuser and then trustuser again without requiring the user to do a logout cycle to reset their trusted status. I've run into issues like that and it was painful.

Hope it's ok to just re-use the trustuser command for the untrustusering as well. I originally had a separate one but duplicating all the boilerplate didn't feel very elegant.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -2463,8 +2463,10 @@ const commands = {
 	},
 	promotehelp: [`/promote [username], [group] - Promotes the user to the specified group. Requires: & ~`],
 
+	untrustuser: 'trustuser',
+	unconfirmuser: 'trustuser',
 	confirmuser: 'trustuser',
-	trustuser(target, room, user) {
+	trustuser(target, room, user, connection, cmd) {
 		if (!target) return this.parse('/help trustuser');
 		if (!this.can('promote')) return;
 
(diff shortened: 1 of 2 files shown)


server/chat-commands.js
```diff
@@ -2477,14 +2479,29 @@ const commands = {
...
+			targetUser.setGroup(Config.groupsranking[0], true);
+			this.sendReply(`User '${name}' is now trusted.`);
+			this.privateModAction(`${name} was set as a trusted user by ${user.name}.`);
+			this.modlog('TRUSTUSER', userid);
+		}
 	},
-	trustuserhelp: [`/trustuser [username] - Trusts the user (makes them immune to locks). Requires: & ~`],
+	trustuserhelp: [
+		`/trustuser [username] - Trusts the user (makes them immune to locks). Requires: & ~`,
+		`/ustrustuser [username] - Removes the trusted user status from the user. Requires: & ~`,
```
scheibo:
```suggestion
		`/untrustuser [username] - Removes the trusted user status from the user. Requires: & ~`,
```

asgdf: Oh whoops. Fixed.

server/users.ts
```diff
@@ -1171,6 +1171,7 @@ class User extends Chat.MessageContext {
 				this.autoconfirmed = this.userid;
 			} else {
 				delete usergroups[this.userid];
+				this.trusted = '';
```
Zarel: Should this not be `null`?

Zarel: Looks like `''` is correct.

asgdf: `this.trusted = '';` is the default value that's set in the `User` constructor (all the other string properties are also initialized as '')

PR #5637, opened 2019-07-20
Don't show "It's a one-hit KO!" if move did not OHKO the target

Danofares:


Zarel:
In what situation would this come up?

If it does, it sounds like a bug in a different part of the code.

Zarel:
I guess Focus Sash? @Marty-D?

Zarel:
Does it not display that message if the target's already taken other damage?

Danofares:
Yeah, focus sash/focus band causes the message to happen, according to https://github.com/Zarel/Pokemon-Showdown/projects/3#card-23557617

Zarel:
Ideally there'd be a link to research. That's almost more important than a replay depicting what PS currently does.

Zarel:
Anyway, does the message appear if the foe's already taken other damage before the OHKO move is used?

data/scripts.js
```diff
@@ -699,7 +699,7 @@ let BattleScripts = {
 			}
 		}
 
-		if (move.ohko) this.add('-ohko');
+		if (move.ohko && move.totalDamage === targets[0].hp) this.add('-ohko');
```
Zarel: If we're only trying to detect the case where the foe survives, this would be:

```suggestion
		if (move.ohko && !targets[0].hp) this.add('-ohko');
```

Danofares:
I'm not sure, I changed it accordingly to account for focus band having not tested that yet. let me see.

data/scripts.js
Zarel: If we also don't want to display the message if the foe starts at non-full HP, I would do:

```suggestion
		if (move.ohko && move.totalDamage >= targets[0].hp) this.add('-ohko');
```

Zarel:
Also @DaWoblefet?

Danofares:
Checked. it does show the message when the target takes damage

DaWoblefet:
I was the one who added that bug report, Zarel. Another example would be an OHKO move into Disguise, but surviving due to a Focus Sash is the most obvious one. If a target has already taken damage, an OHKO move KOing the target will display the "It's a one-hit KO!" message. Here's an example of that: https://youtu.be/uoJEROcq6BI?t=5m58s.

To clarify, if the target doesn't actually faint from the OHKO move, then the text "It's a one-hit KO!" shouldn't appear.

So in testing just now, apparently if Sheer Cold breaks a Substitute, it will still display the "It's a one-hit KO!" message (even prior to the message that the Substitute faded), so maybe this needs more research than I thought still.

Zarel:
I'd also like to know if it appears in the Unnerve 0hp survival situation.

DaWoblefet:
It's impossible to trigger that with OHKO moves; the Unnerve 0 HP survival requires that you use a spread move to KO your Unnerve ally as you reduce an opponent's HP to 0.

Zarel:
We'll need other code for the Substitute situation, but this is a good start. Thanks!

PR #5638, opened 2019-07-21
Update PROTOCOL.md WRT recent user status protocol changes

Kaiepi:
This needs a review, I'm only basing this off changes I can see while interacting with the server.


scheibo:
Thank you for updating this! I will also need to remember in #5627 to update it further, but this change accurately reflects the current protocol to the best of my understanding

PR #5639, opened 2019-07-21
Moves: Fix Flower Veil + Yawn interaction when ally is already under effect of Yawn

jmclemo6:
Fixes https://github.com/Zarel/Pokemon-Showdown/projects/3#card-24205020

Unit test included.


Slayer95:
Thanks!

PR #5641, opened 2019-07-21
Fix Encore + Pledge moves interaction

jmclemo6:
I believe this fixes https://github.com/Zarel/Pokemon-Showdown/projects/3#card-24204988.

I added a test for this, but there are probably better ways to check what I was checking. If so, let me know and I can fix them up.

### DIFF
--- a/data/scripts.js
+++ b/data/scripts.js
@@ -20,7 +20,7 @@ let BattleScripts = {
 		let target = this.getTarget(pokemon, zMove || moveOrMoveName, targetLoc);
 		let baseMove = this.getActiveMove(moveOrMoveName);
 		const pranksterBoosted = baseMove.pranksterBoosted;
-		if (!sourceEffect && baseMove.id !== 'struggle' && !zMove) {
+		if (baseMove.id !== 'struggle' && !zMove && !externalMove) {
 			let changedMove = this.runEvent('OverrideAction', pokemon, target, baseMove);
 			if (changedMove && changedMove !== true) {
 				baseMove = this.getActiveMove(changedMove);
--- a/test/sim/misc/pledgemoves.js
+++ b/test/sim/misc/pledgemoves.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const assert = require('./../../assert');
+const common = require('./../../common');
+
+let battle;
+
+describe('Pledge Moves', function () {
+	beforeEach(function () {
+		battle = common.createBattle({gameType: 'doubles'});
+	});
+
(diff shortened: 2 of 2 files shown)


Slayer95:
It would probably be better to fix this in the ``OverrideAction`` [implementation](https://github.com/Zarel/Pokemon-Showdown/blob/5fb57e6cfc33031cd66cc15a09644ec4a6b47907/data/scripts.js#L23).

It would also make sense to test whether a prioritized Round can be overriden into an encored move. cc @DaWoblefet

DaWoblefet:
Round still maintains the "you move next" property of double Round, but it won't gain a boosted BP or override the move with Round.

To give an illustration:
Previous turn Meloetta uses Return.
This turn, faster Smeargle and Meloetta select Round. The opposing Whimsicott Encores Meloetta into Round. Smeargle uses Round, then Meloetta uses Return before the opponent's Pokemon which was faster than Meloetta.

Slayer95:
Thanks! So Round has the same glitch.

Then it would go something like this... 
```js
if (!zMove && baseMove.id !== 'struggle' && (!sourceEffect || ['round', 'firepledge', 'waterpledge', 'grasspledge'].includes(baseMove.id))) {
}
```

At which point we must consider whether ``!sourceEffect`` makes sense at all? It seems it'd only be relevant for Instruct and Pursuit...

So, if A uses Instruct and then the foe uses Encore on A....??? I guess Encore would actually fail?

Marty-D:
Sorry about the radio silence; to Slayer95's point above I've been meaning to test the Pursuit interaction without checking for sourceEffects here but haven't had time yet.

How does Dancer react in this situation? While Encored into something it should attempt to copy dance moves and then fail unless it's the Encored move (just like Choice locking), but I don't think that's happening right now?

jmclemo6:
Dancer has never checked for that on PS as far as I know. At the very least that bug is also on main right now as I just tested it there.

Marty-D:
Okay it turns out Pursuit doesn't even pass a `sourceEffect`, so nothing that actually passes one wants to avoid having the move overridden by Encore. Except technically Dancer, but that's a special case. We can implement that later though.

This means the line can just be `if (baseMove.id !== 'struggle' && !zMove && !externalMove) {` for now, and then deal with Dancer separately.

jmclemo6:
Done @Marty-D. :D

DaWoblefet:
Research completed on oldgens @Slayer95;  this has occurred since the moves were introduced in Generation 5.

Slayer95:
Thanks!

PR #5645, opened 2019-07-23
Add a few kaomoji commands

Kaiepi:
This adds /shrug, /tableflip, and /tableunflip.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -419,6 +419,30 @@ const commands = {
 		return target;
 	},
 
+	'!shrug': true,
+	shrug(target) {
+		if (!this.canTalk()) return;
+		target = target ? target + ' ' : '';
+		return target + '¯\\_(ツ)_/¯';
+	},
+	shrughelp: ['/shrug [message] - Sends the given message, if any, appended with ¯\\_(ツ)_/¯'],
+
+	'!tableflip': true,
+	tableflip(target) {
+		if (!this.canTalk()) return;
+		target = target ? target + ' ' : '';
+		return target + '(╯°□°）╯︵ ┻━┻';
+	},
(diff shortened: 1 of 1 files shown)


server/chat-commands.js
```diff
@@ -419,6 +419,27 @@ const commands = {
 		return target;
 	},
 
+	'!shrug': true,
+	shrug() {
+		if (!this.canTalk()) return;
```
Zarel: Would this otherwise bypass `canTalk`?

Also I believe Discord's implementation is `target + ' ¯\_(ツ)_/¯'`

server/chat-commands.js
```diff
@@ -419,6 +419,27 @@ const commands = {
...
+		return '¯\\_(ツ)_/¯';
```
Zarel: You don't need the double-backslash, PS, unlike Discord, is actually designed _not_ to mangle kaomoji.

Kaiepi: JS tried escaping the `_` when I tested it

Slayer95: This is because JavaScript

Slayer95:
The strings should be ``\u``-escaped, right?

server/chat-commands.js
Kaiepi: Seems like it does

Kaiepi:
Do they? I think it's easier to tell what the commands are doing when it's unescaped. I'd expect people working on PS to know how to set up their fonts so it doesn't show up as tofu

Slayer95:
Well, it's already the ruling that we should not use ``"Pokémon"`` (not that I support that rule).

server/chat-commands.js
Zarel: Ohhhhh lol nvm

Zarel:
I'll allow it here.

PR #5656, opened 2019-07-25
Fix kaomoji canTalk checking for real

whalemer:


Slayer95:
Yup

Zarel:
Out of curiosity, would this work?

```js
shrug(target) {
  return this.parse(target + ' ¯\\_(ツ)_/¯');
}
```

It would fix `/shrug /me's test`, which this commit doesn't.

whalemer:
It would (I suggested it in dev ;p), though it does lead to some perhaps unintentional things like ``/shrug /lock zarel, `` working, though I really doubt that's an actual issue.

_i think_ it'd also open a crash by faking infinite recursion with ``/shrug /shrug /shrug /shrug /shrug /shrug /shrug /shrug /shrug /shrug``

Slayer95:
Hm? I think you are confused, @Zarel . With this commit:

```
> /shrug /me's test``
< [21:24] #Slayer95: /me's test ¯\_(ツ)_/¯
```

```
> /shrug /me 's test``
< [21:24] • #Slayer95 's test ¯\_(ツ)_/¯
```
```
> /shrug /mee 's test``
< [21:24] • #Slayer95's test ¯\_(ツ)_/¯
```

Zarel:
Yeah, that's what I'm saying.

`/me's test` normally outputs

`[21:24] • #Slayer95's test`

But that fails with this commit, since it hardcodes `/me` instead of going through the full command parser.

whalemer:
/me only works because it's a weird clientside command, iirc (`[21:24] • #Slayer95's test` is actually `|c|#Slayer95|/mee 's test`)

Slayer95:
> eah, that's what I'm saying.

/me's test normally outputs

[21:24] • #Slayer95's test

W-what... is that new?

Anyway, implementing this through recursivity doesn't look quite good to me, else we need to start adding safeguards for which commands are allowed in reparsing.......

It's also the case that a sequence of commands isn't intuitive at all. It'd be much better if users were to type ``/me test :shrug:``, and get the ``:shrug:`` replaced by the emoji client-side on-send.

Zarel:
It's there because it's more discoverable than `/mee`. It could probably stand to be documented, though.

Slayer95:
I don't have enough words to argue about how that feature is inconsistent, badly planned and has become now a technical debt source.

It would have been enough to remove the space from ``/me 's testing`` client-side. That would also automatically fix your raised issue re: shrug.

Slayer95:
I don't have enough words to argue about how that feature is inconsistent, badly planned and has become now a technical debt source.

It would have been enough to remove the space from ``/me 's testing`` client-side. That would also automatically fix your raised issue re: shrug.

PR #5657, opened 2019-07-25
Improve README attribution

scheibo:
Trying to standardize on `Real name [PS username]`.

Currently its inconsistent whether the username is included when its identical to one of a contributor's real name (Austin vs. Kris + McLemore + Spandan). Furthermore, GH username is only included in some cases (scheibo, Kaipei) and not in all (peach + whaelmer, McLemore + jmclemo6, jetou + TheJetOU, spandan + Spandanm etc).

I'm not sure which of `[moo, CheeseMuffin]` and `[Lyren, SolarisFox]` is the PS username (I assume the first?). If you're adamant on keeping GH username, I'd want us to include it for everyone where it differs, not just on a select few names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,8 +152,8 @@ Staff
 
 - Bär Halberkamp [bumbadadabum] - Development
 - Chris Monsanto [chaos] - Sysadmin
-- Mathieu Dias-Martins [Marty-D] - Research (game mechanics), Development
 - Kirk Scheibelhut [pre] - Development
+- Mathieu Dias-Martins [Marty-D] - Research (game mechanics), Development
 - [The Immortal] - Development
 
 Retired Staff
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -168,7 +168,7 @@ Major Contributors
 
 - Andrew Werner [HoeenHero] - Development
 - Austin Couturier [Austin] - Development (damage calculator)
-- [jumbowhales] - Development
+- [peach] - Development
```
whalemer: This is in the wrong place now :)

scheibo: Ah. There were a couple other places not alphabetized either, fixed.

Zarel:
I prefer to keep GH username whenever it's non-obvious.

scheibo:
Define 'obvious'. eg. you can easily see from my GH profile that lists my full name what my GH name is, having one is perfectly sufficient. 

I think the best option would be to allow the specific _contributor_ to decide how they would like their name/identity attributed.

Zarel:
I would say "guessable without additional information".

scheibo:
I'd argue that the same logic that gets you from 'jmclemo6' -> '**J**acob **McLemo**re' will take you from 'scheibo' -> 'Kirk **Scheib**elhut'

Zarel:
Sure.

PR #5262, opened 2019-03-06
Chat plugins: ignore links when determining room FAQ length

Kaiepi:
The current limit's much too short for Bot Development's room FAQ on
bots' nicks and git repositories.

PR #5658, opened 2019-07-27
Implement Chat.stripFormatting

HoeenCoder:
The idea here is that formatting such as `**`, `__`, `~~`, ect and links (bar the "text" in `[[text<url>]]`) do not count to the 500 character limit for daily spotlights/rfaqs. This has been requested by UnleashOurPassion (and iirc other staff members).

This Pull Request closes #5659 (I asked UOP if I could merge their changes into this).

### DIFF
--- a/server/chat-formatter.ts
+++ b/server/chat-formatter.ts
@@ -444,3 +444,14 @@ class TextFormatter {
 export function formatText(str: string, isTrusted = false) {
 	return new TextFormatter(str, isTrusted).get();
 }
+
+/**
+ * Takes a string and strips all standard chat formatting except greentext from it, the text of a link is kept.
+ */
+export function stripFormatting(str: string) {
+	// Doesn't match > meme arrows because the angle bracket appears in the chat still.
+	str = str.replace(/\*\*([^\s\*]+)\*\*|__([^\s_]+)__|~~([^\s~]+)~~|``([^\s`]+)``|\^\^([^\s\^]+)\^\^|\\([^\s\\]+)\\/g,
+		(match, $1, $2, $3, $4, $5, $6) => $1 || $2 || $3 || $4 || $5 || $6);
+	// Remove all of the link expect for the text in [[text<url>]]
+	return str.replace(/\[\[(?:([^<]*)\s*<[^>]+>|([^\]]+))\]\]/g, (match, $1, $2) => $1 || $2 || '');
+}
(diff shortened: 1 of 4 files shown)


whalemer:
implementation of a similar thing at #5262

TheMezStrikes:
Yeah #5262 is a much cleaner implementation than the code I wrote, and is functionally identical as far as I can tell (provided you fix the regex to actually work)

HoeenCoder:
Closes #5659

server/chat-formatter.ts
```diff
@@ -444,3 +444,14 @@ class TextFormatter {
 export function formatText(str: string, isTrusted = false) {
 	return new TextFormatter(str, isTrusted).get();
 }
+
+/**
+ * Takes a string and strips all standard chat formatting from it, the text of a link is kept.
+ */
+export function stripFormatting(str: string) {
+	// Doesn't match > meme arrows because the angle bracket appears in the chat still.
+	str = str.replace(/\*\*([^\s\*]+)\*\*|__([^\s_]+)__|~~([^\s~]+)~~|``([^\s`]+)``|\^\^([^\s\^]+)\^\^|\\([^\s\\]+)\\/g,
```
HoeenCoder: This is close to working, but still wont properly strip nested formatting such as `**~~text~~**`, which will return `~~text~~`. Any help on this?

TheJetOU: I guess checking if the RegExp still matches after the `.replace` and if it does running the function recursively should work.

server/chat-formatter.ts
```diff
@@ -444,3 +444,14 @@ class TextFormatter {
 export function formatText(str: string, isTrusted = false) {
 	return new TextFormatter(str, isTrusted).get();
 }
+
+/**
+ * Takes a string and strips all standard chat formatting from it, the text of a link is kept.
```
TheJetOU: Should probably add the comment about greentext here, otherwise it's kind of misleading.

server/chat-formatter.ts
Zarel: I think it's unnecessary to strip nested formatting. Links are the main thing, anyway.

Zarel: Like I would consider this "good enough" and merge despite that problem. No need to make it perfect.

HoeenCoder: Ok, lets go with this.

PR #5659, opened 2019-07-27
Properly ignore links when determining RFAQ length

TheMezStrikes:
- Use a multi-line regular expression for a multi-line command
- Add a `*` after the `\s` because the space in the middle of `[[foo <bar>]` is not mandatory for link formatting

### DIFF
--- a/server/chat-plugins/room-faqs.js
+++ b/server/chat-plugins/room-faqs.js
@@ -47,7 +47,7 @@ const commands = {
 		topic = toID(topic);
 		if (!(topic && rest.length)) return this.parse('/help roomfaq');
 		let text = rest.join(',').trim();
-		let filteredText = text.replace(/\[\[(?:([^<]+)\s<[^>]+>|([^\]]+))\]\]/g, (match, $1, $2) => $1 || $2);
+		let filteredText = text.replace(/\[\[(?:([^<]+)\s*<[^>]+>|([^\]]+))\]\]/gm, (match, $1, $2) => $1 || $2);
 		if (topic.length > 25) return this.errorReply("FAQ topics should not exceed 25 characters.");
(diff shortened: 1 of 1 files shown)


TheJetOU:
Maybe we should abstract this away into a function in `Chat`? `Chat.stripLinks`?

TheMezStrikes:
I support abstracting it but am not familiar enough with PS coding conventions to feel confident doing something like that myself, so someone else would have to

Zarel:
Can links themselves span multiple lines?

TheMezStrikes:
They cannot

Zarel:
So what is the `m` (multiline) flag supposed to achieve?

Zarel:
It looks like the `m` flag is useless here: It only affects what `^` and `$` do, and you don't use them in this regex.

server/chat-plugins/room-faqs.js
```diff
@@ -47,7 +47,7 @@ const commands = {
 		topic = toID(topic);
 		if (!(topic && rest.length)) return this.parse('/help roomfaq');
 		let text = rest.join(',').trim();
-		let filteredText = text.replace(/\[\[(?:([^<]+)\s<[^>]+>|([^\]]+))\]\]/g, (match, $1, $2) => $1 || $2);
+		let filteredText = text.replace(/\[\[(?:([^<]+)\s*<[^>]+>|([^\]]+))\]\]/gm, (match, $1, $2) => $1 || $2);
```
Zarel:
```suggestion
		let filteredText = text.replace(/\[\[(?:([^<]+)\s*<[^>]+>|([^\]]+))\]\]/g, (match, $1, $2) => $1 || $2);
```

PR #5660, opened 2019-07-27
Attribute creator in polls

scheibo:
**Context:** https://www.smogon.com/forums/threads/let-users-see-who-made-a-poll-or-hangman.3652444/

image

### DIFF
--- a/server/chat-plugins/poll.js
+++ b/server/chat-plugins/poll.js
@@ -12,12 +12,14 @@ class Poll {
 	/**
 	 * @param {ChatRoom | GameRoom} room
 	 * @param {QuestionData} questionData
+	 * @param {string|null} creator
 	 * @param {string[]} options
 	 */
-	constructor(room, questionData, options) {
+	constructor(room, questionData, creator, options) {
 		this.pollNumber = ++room.gameNumber;
 		this.room = room;
 		this.question = questionData.source;
+		this.creator = creator;
 		this.supportHTML = questionData.supportHTML;
 		/** @type {{[k: string]: number}} */
 		this.voters = {};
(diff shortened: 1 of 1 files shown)


server/chat-plugins/poll.js
```diff
@@ -218,6 +222,7 @@ exports.Poll = Poll;
 /** @type {ChatCommands} */
 const commands = {
 	poll: {
+		anoncreate: 'new',
```
TheJetOU: That means you can't create anonymous HTML polls.

You'll probably need to make another alias (`anonhtmlcreate`) and handle it accordingly.

bumbadadabum:
But why? What's the point?

Zarel:
I think a "Foo created a poll" public message would be nice at the start of a poll, but as-is this feels like a waste of screen real-estate. Polls are already too spammy as-is.

Slayer95:
There are rooms give attribution to the _user who submits the question_ for staff to publish. That already uses up some space. Implementing support for that _might_ be more useful? Still, staff already handle that with the currently available tools...

So, I don't think this is needed. +1 for a single notification when the poll starts.

server/chat-plugins/poll.js
scheibo: Yeah, I saw that, just didnt think it mattered.

scheibo:
From the suggestion it seemed the main support for the use case was not having a bunch of users asking "who created the poll?". I'm not sure a public message accomplishes that for users the join the room after the fact (because polls can persist much longer than default scrollback, no?). Anyway, whatever, idc. Someone else can try this.

Issue #5646, opened 2019-07-23
Add a way to flag 'tour' or 'suspect' accounts


scheibo:
> I'm not against adding a tour bit to accounts - one use case for flagging accounts that came up recently is that **we could disable `/offertie` for ladder tours** in the same way we do on Smogtours/room tours. Similarly, being able to **flag an account as participating in a suspect test** would also make some Tier Leaders happy, given there was concern over how they would treat users who `/offertie`.

_Originally posted by @scheibo in https://github.com/Zarel/Pokemon-Showdown-Client/pull/1326/threads/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTg2MTM5OTc2OnYy/unresolve_

Such a bit would also allow us to [disable `/ionext` in OLT](https://www.smogon.com/forums/threads/the-world-cup-of-pok%C3%A9mon-2019-finals-won-by-team-us-west.3652418/page-6#post-8191756) (policy pending)

TheJetOU:
How would we do this in a reasonable way?

One way would be for upper staff to be able to mark account that start with a certain string as `laddertour` / `suspect` (or just straight up hard-coded with the suspect test PRs), and in the `User` object have a `type` property which could be `regular`, `laddertour`, or `suspect`, depending on if the `userid` starts with the `laddertour`  or `suspect` values, otherwise, it's a `regular` account.

Then from there, we would use standard logic to enable / disable certain features.

We would also need to display it in the UI, perhaps in `/cmd userdetails`?

There's probably a better approach though.

scheibo:
> How would we do this in a reasonable way?

I don't have a solution, I probably should have made that clear with the 'discussion' label :)

> mark account that start with a certain string

In an ideal design, I'd imagine this to just be a toggle users could set, and it wouldnt need to affect their name at all. However, encoding it in the name as we currently do today seems to be the most obvious implementation which would result in the least changes to code/protocol.

TheJetOU:
> In an ideal design, I'd imagine this to just be a toggle users could set, and it wouldnt need to affect their name at all.

I'm not sure I agree with users being able to make any of their accounts a 'suspect' / 'ladder' account; that'd be really confusing to new users.

scheibo:
Current feeling: add a `Config.prefixlist`, something like:

```ts
Config.prefixlist = {
   // OLT Qualifier #2
[...]

PR #5661, opened 2019-07-28
Add forcedpublicprefixes for public OLT battles

scheibo:
Usernames prefixed by a special value decided ahead of time (eg. `'LT61RQ'`) will unable to hide their battles from spectators, as discussed in #5646. This is not a 'complete' OLT solution, we will probably want to followup with support for `/hidereplay`, as well as force unregistered accounts to only be able to ladder unrated to simplify coordination issues. However, simply coordinating  around updating `Config.forcedpublicprefixes` with the new prefix for the next qualifying periods is probably sufficient.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -402,6 +402,15 @@ exports.replsocketmode = 0o600;
 exports.disablehotpatchall = false;
 
 /**
+ * forcedpublicprefixes - user ID prefixes which will be forced to battle publicly.
+ * Battles involving user IDs which begin with one of the prefixes configured here
+ * will be unaffected by various battle privacy commands such as /modjoin, /hideroom
+ * or /ionext.
+ * @type {string[]}
+ */
+exports.forcedpublicprefixes = [];
+
+/**
  * permissions and groups:
  *   Each entry in `grouplist' is a seperate group. Some of the members are "special"
  *     while the rest is just a normal permission.
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4105,6 +4105,13 @@ const commands = {
 		}));
 	},
 
+	hidereplay: function (target, room, user, connection) {
+		if (!room || !room.battle || !this.can('joinbattle', null, room)) return;
+		if (room.hideReplay) return this.errorReply(`The replay for this battle is already set to hidden.`);
+		room.hideReplay = true;
+		this.addModAction(`${user.name} hid the replay of this battle.`);
+	},
+
 	addplayer(target, room, user) {
 		if (!target) return this.parse('/help addplayer');
 		if (!room.battle) return this.errorReply("You can only do this in battle rooms.");
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1008,8 +1008,8 @@ const commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
-			const prefix = room.forcedPublic();
-			if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			const prefix = room.battle.forcedPublic();
+			if (prefix && !user.can('lock')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			// registered chatrooms show up on the room list and so require
 			// higher permissions to modify privacy settings
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1008,6 +1008,10 @@ const commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
+			if (room.battle) {
+				const prefix = room.forcedPublic();
+				if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			}
 		} else {
 			// registered chatrooms show up on the room list and so require
 			// higher permissions to modify privacy settings
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1008,10 +1008,8 @@ const commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
-			if (room.battle) {
-				const prefix = room.forcedPublic();
-				if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
-			}
+			const prefix = room.forcedPublic();
+			if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			// registered chatrooms show up on the room list and so require
 			// higher permissions to modify privacy settings
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1009,7 +1009,7 @@ const commands = {
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
 			const prefix = room.battle.forcedPublic();
-			if (prefix && !user.can('lock')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			if (prefix && !user.can('editprivacy')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			// registered chatrooms show up on the room list and so require
 			// higher permissions to modify privacy settings
@@ -4105,7 +4105,7 @@ const commands = {
 		}));
 	},
 
-	hidereplay: function (target, room, user, connection) {
+	hidereplay(target, room, user, connection) {
 		if (!room || !room.battle || !this.can('joinbattle', null, room)) return;
 		if (room.hideReplay) return this.errorReply(`The replay for this battle is already set to hidden.`);
 		room.hideReplay = true;
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -328,6 +328,10 @@ exports.commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
+			if (room.battle) {
+				const prefix = room.forcedPublic();
+				if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			}
 		} else {
 			if (!this.can('makeroom')) return;
 		}
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -328,10 +328,8 @@ exports.commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
-			if (room.battle) {
-				const prefix = room.forcedPublic();
-				if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
-			}
+			const prefix = room.forcedPublic();
+			if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			if (!this.can('makeroom')) return;
 		}
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -330,7 +330,7 @@ exports.commands = {
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
 			const prefix = room.battle.forcedPublic();
-			if (prefix && !user.can('lock')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			if (prefix && !user.can('editprivacy')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			if (!this.can('makeroom')) return;
 		}
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -296,6 +296,7 @@ exports.commands = {
 		} else {
 			user.inviteOnlyNextBattle = true;
 			user.update('inviteOnlyNextBattle');
+			if (user.forcedPublic) return this.errorReply(`Your next battle will be invite-only provided it is not rated, otherwise your '${user.forcedPublic}' prefix will force the battle to be public.`);
 			this.sendReply("Your next battle will be invite-only.");
 		}
 	},
@@ -328,8 +329,8 @@ exports.commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
-			const prefix = room.forcedPublic();
-			if (prefix) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
+			const prefix = room.battle.forcedPublic();
+			if (prefix && !user.can('lock')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 		} else {
 			if (!this.can('makeroom')) return;
 		}
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -998,6 +998,7 @@ class RoomBattle extends RoomGames.RoomGame {
 	}
 
 	forcedPublic() {
+		if (!this.rated) return;
 		for (const userid in this.playerTable) {
 			const user = Users(userid);
 			if (user && user.forcedPublic) return user.forcedPublic;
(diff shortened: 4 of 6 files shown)


server/rooms.js
```diff
@@ -1693,8 +1693,12 @@ let Rooms = Object.assign(getRoom, {
 				user.inviteOnlyNextBattle = false;
 			}
 		}
-		if (options.tour && !room.tour.modjoin) inviteOnly = [];
-		if (inviteOnly.length) {
+		const prefix = /** @type {RoomBattle} */(room.game).forcedPublic();
+		if (prefix && inviteOnly.length) {
+			room.isPrivate = false;
+			room.modjoin = null;
+			room.add(`|raw|<div class="broadcast-red"><strong>This battle is required to be public due to a player having a name prefixed by '${prefix}'.</div>`);
```
scheibo: Apparently we don't provide a warning for this in `room.tour` situations, but I really think player's should be alerted if their `/ionext` doesn't work all of a sudden.

config/config-example.js
```diff
@@ -401,6 +401,15 @@ exports.replsocketmode = 0o600;
...
 exports.disablehotpatchall = false;
 
+/**
+ * forcedpublicprefixes - user ID prefixes which will be forced to battle publicly.
+ * Battles involving user IDs which begin with one of the prefixes configured here
+ * will be unaffected by various battle privacy commands such as /modjoin, /hideroom
+ * or /ionext.
+ * @type {string[]}
+ */
+exports.forcedpublicprefixes = [];
```
scheibo: I'm accepting `string` prefixes because a) its a `.js` file and you cant `as ID` stuff, and b) the uppercase prefix is prettier for display purposes.

I gave up on making prefixes more finely grained (lots of edge cases and weird interactions) - instead, these only apply to battles and only to a preset amount of commands involving privacy)

server/room-battle.js
```diff
@@ -997,6 +997,13 @@ class RoomBattle extends RoomGames.RoomGame {
 		return player;
 	}
 
+	forcedPublic() {
+		for (const userid in this.playerTable) {
+			const user = Users(userid);
+			if (user && user.forcedPublic) return user.forcedPublic;
```
scheibo: This is potentially all overkill - we might never have/need more than one 'forced public' prefix at the same time, and maybe we don't even care about reporting what the prefix is to users - I just figured the extra explanation would be helpful given we're kind of springing this privacy regression on them,

bumbadadabum: I think it's fine. I don't think it's 100% needed but it doesn't do harm, and is a nice addition.

bumbadadabum:
It looks good to me but I'm not entirely certain of the battle code either. I figure this is all you'd need for rated battles though.

server/rooms.js
TheJetOU: Why does this need to be a broadcast instead of an error reply (like in `/modjoin`)?

server/chat-plugins/roomsettings.js
```diff
@@ -328,6 +328,10 @@ exports.commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
+			if (room.battle) {
```
TheJetOU: The same condition is there 2 lines above.

server/rooms.js
```diff
@@ -1693,8 +1693,12 @@ let Rooms = Object.assign(getRoom, {
...
+		} else if (!options.tour || room.tour.modjoin) {
```
TheJetOU: This seems to be a faulty if condition? Like, this could run even if `inviteOnly.length` is empty.

server/rooms.js
```diff
@@ -1693,8 +1693,12 @@ let Rooms = Object.assign(getRoom, {
 				user.inviteOnlyNextBattle = false;
 			}
 		}
-		if (options.tour && !room.tour.modjoin) inviteOnly = [];
-		if (inviteOnly.length) {
+		const prefix = /** @type {RoomBattle} */(room.game).forcedPublic();
+		if (prefix && inviteOnly.length) {
```
whalemer: This should probably only get checked in rated battles, as long as I'm understanding how ladder tours work properly.

whalemer: Also, doesn't this totally break setting invite only for ionext + no prefix?

whalemer:
just incase you weren't aware, ``Config.fakeladder`` lets you bypass all the important matchmaking checks to make testing locally at least possible

server/chat-plugins/roomsettings.js
scheibo: Thanks!

server/rooms.js
scheibo: This is the is the same as what `/ionext` does below. It doesnt happen when you issue the command (which is when we `errorReply`, it happens after the command already succeeded, which is why I think the more aggressive messaging is important.

server/rooms.js
scheibo: right, I meant to `if (inviteOnly.length)` the whole thing, thanks.

server/rooms.js
scheibo: Added a `room.rated` check. `/ionext` without a prefix should be handled in the `else if`

server/chat-commands.js
```diff
@@ -1008,6 +1008,10 @@ const commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
+			if (room.battle) {
```
TheJetOU: Same issue here (dupe conditions).

server/rooms.js
```diff
@@ -1693,12 +1693,18 @@ let Rooms = Object.assign(getRoom, {
...
 			}
 		}
-		if (options.tour && !room.tour.modjoin) inviteOnly = [];
 		if (inviteOnly.length) {
-			room.modjoin = '%';
-			room.isPrivate = 'hidden';
-			room.privacySetter = new Set(inviteOnly);
-			room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be invited with <code>/invite</code> (or be staff) to join</div>`);
+			const prefix = /** @type {RoomBattle} */(room.game).forcedPublic();
+			if (prefix && room.rated) {
```
TheJetOU:
```suggestion
			if (prefix && room.game.rated) {
```

server/chat-commands.js
scheibo: Thanks. I think I copied this block from elsewhere when I was trying to make things more general

server/rooms.js
scheibo: I switched to this, but I dont think it should matter? `GameRoom` also has a `rated` field based on `options.rated`, no? It should be pretty much the exact same thing?

TheJetOU: It /shouldn't/ matter, but it's a bit more thinking to arrive at the same conclusion.

`room` is `createGameRoom` which calls `GameRoom` which has a `rated` property.

vs.

`room.game` is `RoomBattle` which has a rated property.

`room` also isn't usually associated with rated-ness.

Super nitpick, but yeah.

HoeenCoder:
Ok nvm I spoke too soon. Currently the commands for modjoin/ect will prevent your from editing the setting in all battles, there should be a `rated` check in there or in `room.forcedPublic`.

scheibo:
> just incase you weren't aware, `Config.fakeladder` lets you bypass all the important matchmaking checks to make testing locally at least possible

<3

Thanks for this. Found a bunch of issues, but now I can at least verify this appears to be WAI.

@Zarel - this requires a restart and I'm going to be going forward with it ~5-7AM CEST tomorrow to have it in place for the 3rd OLT qualifying round. I'd really appreciate if you could weigh in before then, but this has been up for a while and has had plenty of eyes on it.

server/users.ts
```diff
@@ -93,7 +95,16 @@ function merge(user1: User, user2: User) {
 	prevUsers.delete(user2.userid);
 	prevUsers.set(user1.userid, user2.userid);
 }
-
+function updateForcedPublic(user: User) {
```
whalemer: This doesn't clear `forcedPublic`, so `/nick`ing to a prefixed name and back will still force public.

server/users.ts
```diff
@@ -84,6 +85,7 @@ function add(user: User) {
 
 	if (users.has(user.userid)) throw new Error(`userid taken: ${user.userid}`);
 	users.set(user.userid, user);
+	updateForcedPublic(user);
```
whalemer: isn't the userid always going to be `guest<numbers>` here?

server/chat-plugins/roomsettings.js
```diff
@@ -328,6 +329,8 @@ exports.commands = {
 			if (!this.can('editroom', null, room)) return;
 		} else if (room.battle) {
 			if (!this.can('editprivacy', null, room)) return;
+			const prefix = room.battle.forcedPublic();
+			if (prefix && !user.can('lock')) return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
```
whalemer: `lock` seems like a weird permission for this, personally

server/users.ts
scheibo: OK, thats what I saw in testing, but wasn't confident. Removed.

server/users.ts
scheibo: Thank, clearing `forcedPublic` now.

server/chat-plugins/roomsettings.js
scheibo: Basically just wanted gstaff to be able to handle things. I guess global `editprivacy` makes the most sense, but its `&`+. Switched to that, don't know if I'll have people in staff getting angry with me.

server/chat-commands.js
```diff
@@ -4105,6 +4105,13 @@ const commands = {
 		}));
 	},
 
+	hidereplay: function (target, room, user, connection) {
```
scheibo: This comes from tours private code, at the request of the TDs. Its strictly less powerful than `/hideroom`, so doesnt seem to harmful to support for everyone, though it would likely only be of use to forced public users

scheibo:
>  going forward with it ~5-7AM EST tomorrow to have it in place

I've synced up with @QuiteQuiet and we don't want to do a restart between ~5AM-9AM tomorrow because those are the crucial finishing hours of the tournament and it could have a negative effect on the results. If the user numbers are still <~8k around 9AM EST  id consider doing the restart then, but its more likely I'll wait until Wednesday @ 5AM EST instead (being in CEST really helps for these stupidly early morning restarts :P)

server/chat-commands.js
```diff
@@ -4103,6 +4105,13 @@ const commands = {
 		}));
 	},
 
+	hidereplay(target, room, user, connection) {
+		if (!room || !room.battle || !this.can('joinbattle', null, room)) return;
```
Zarel: Should this also have a `forcedPublic` check?

scheibo: This is the command we specifically want OLT participants to have access to

server/room-battle.js
```diff
@@ -997,6 +997,14 @@ class RoomBattle extends RoomGames.RoomGame {
 		return player;
 	}
 
+	forcedPublic() {
```
Zarel: `isForcedPublic` reads better, I think? The codebase is kinda inconsistent about this, though.

scheibo: I agree, but `is` to me implies a `boolean` should be returned, not `string | undefined`.

server/room-battle.js
```diff
@@ -997,6 +997,14 @@ class RoomBattle extends RoomGames.RoomGame {
...
+		if (!this.rated) return;
+		for (const userid in this.playerTable) {
```
Zarel: Ideally don't iterate `playerTable` ever.

`for (const player of this.players) {` should be what you're looking for.

server/room-battle.js
```diff
@@ -997,6 +997,14 @@ class RoomBattle extends RoomGames.RoomGame {
...
+			const user = Users(userid);
+			if (user && user.forcedPublic) return user.forcedPublic;
```
Zarel: wtb optional chaining

server/room-battle.js
scheibo: Don't they not have a user id though?

server/room-battle.js
scheibo: Yeah :(

server/rooms.js
```diff
@@ -1693,12 +1694,18 @@ let Rooms = Object.assign(getRoom, {
...
 		if (inviteOnly.length) {
-			room.modjoin = '%';
-			room.isPrivate = 'hidden';
-			room.privacySetter = new Set(inviteOnly);
-			room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be invited with <code>/invite</code> (or be staff) to join</div>`);
+			const prefix = battle.forcedPublic();
+			if (prefix) {
+				room.isPrivate = false;
+				room.modjoin = null;
+				room.add(`|raw|<div class="broadcast-red"><strong>This battle is required to be public due to a player having a name prefixed by '${prefix}'.</div>`);
```
Zarel: I would color this non-red so it's clear it's not private. Maybe like blue?

Zarel:
```suggestion
				room.add(`|raw|<div class="broadcast-blue"><strong>This battle is required to be public due to a player having a name prefixed by '${prefix}'.</div>`);
```

suggestion for one-click application, but feel free to decline this one

server/users.ts
```diff
@@ -72,6 +72,16 @@ function move(user: User, newUserid: ID) {
 	user.userid = newUserid;
 	users.set(newUserid, user);
 
+	user.forcedPublic = undefined;
```
```suggestion
	user.forcedPublic = null;
```

And here I thought we agreed to use `null` for nullables :p

server/users.ts
```diff
@@ -440,6 +450,7 @@ class User extends Chat.MessageContext {
 	lastPM: string;
 	team: string;
 	lastMatch: string;
+	forcedPublic?: string;
```
```suggestion
	forcedPublic: string | null;
```

server/rooms.js
scheibo: I thought not being private when a user intended it to be is more critical to call out than it actually being private, which is why i went with red. But good point, if private already === red, then a user could just see the color and assume it means private. Blue is fine for me, even if its less scary/abnormal

server/room-battle.js
Zarel: They do.

```suggestion
		for (const player of this.players) {
```

server/room-battle.js
```diff
@@ -997,6 +997,14 @@ class RoomBattle extends RoomGames.RoomGame {
 		return player;
 	}
 
+	forcedPublic() {
+		if (!this.rated) return;
+		for (const userid in this.playerTable) {
+			const user = Users(userid);
```
```suggestion
			const user = player.getUser();
```

server/room-battle.js
Zarel: Fair, it can be left as-is.

But note that we use `is` as a prefix for a lot of other pseudo-booleans which can be treated as booleans but also contain other information.

scheibo: I'm fine to change to `isForcedPublic` if you'd prefer, it just seems like it was add to confusion to me. Its a trivial change, if you'd like me to `sed -i s/forcedPublic/isForcedPublic/g` im happy to do so

Zarel: I don't have a strong preference here. I'd say leave it alone for now.

server/users.ts
```diff
@@ -526,6 +526,7 @@ class User extends Chat.MessageContext {
 		this.lastPM = '';
 		this.team = '';
 		this.lastMatch = '';
+		this.forcedPublic = null;
```
Zarel: This is actually another reason to use `null` over `undefined`; according to various docs on how V8 works, forcing it to be set here will make it so `User` stays the same virtual class, instead of V8 needing to change the class when the property is added.

Zarel: It matters what order you add properties, too, because it affects stuff like the order of `Object.keys`.

If you set `forcedPublic` first, vs if you set `locked` first, would make them end up as different classes (as verified by `Object.keys`), leading to significantly more polymorphism needed by functions taking `User` objects than you would expect.

scheibo: The class is already megamorphic

Zarel: In theory, I want to fix that, and TypeScript can assist in doing so.

Zarel:
Looks fine.

server/chat-commands.js
Zarel: Please either reply or resolve conversations, but not both. When you do both, it hides your response from me.

Anyway, is there a link to discussion/reasoning?

scheibo: I left this comment unresolved for multiple review cycles after responding and your lack of activity on it made it seem resolved.

https://www.smogon.com/forums/threads/rated-games-and-modjoin.3652782/ is the thread. Below is confirmation from a TD:

```
16:38] pre: you wanted me to also steal /hidereplay from tours, right?
[16:40] Tony: let me double check if that's still what the other tds want
[16:41] Tony: ok yes we do
[16:41] Tony: if it's not too tricky to implement of course
[16:41] pre: Well, it's more a policy question
[16:42] Tony: i suggested it in the thread because i like the consistency it offers between smogon tour and olt qualification, while also giving the players some protection
[16:42] pre: Do we give it to all sim users? Just OLT prefixed users? Is it something we want to support at all etc?
[16:43] Tony: and the players responded positively to it, so did the tds
[16:43] Tony: well outside of olt is not really anything i have a say in i think
[16:43] Tony: it can't hurt right?
[16:43] Tony: if you have hideroom for non-olt games, then hidereplay seems harmless
```

PR #5662, opened 2019-08-01
Allow filtering by userid prefix in battle search

TheJetOU:
(Client PR: Zarel/Pokemon-Showdown-Client#1342)

Suggested and apporved here: https://www.smogon.com/forums/threads/3653073/

### DIFF
--- a/server/rooms.js
+++ b/server/rooms.js
@@ -623,12 +623,12 @@ class GlobalRoom extends BasicRoom {
 	}
 
 	/**
-	 * @param {string} filter "formatfilter, elofilter"
+	 * @param {string} filter "formatfilter, elofilter, usernamefilter"
 	 */
 	getBattles(filter) {
 		let rooms = /** @type {GameRoom[]} */ ([]);
 		let skipCount = 0;
-		const [formatFilter, eloFilterString] = filter.split(',');
+		const [formatFilter, eloFilterString, usernameFilter] = filter.split(',');
 		const eloFilter = +eloFilterString;
 		if (this.battleCount > 150 && !formatFilter && !eloFilter) {
 			skipCount = this.battleCount - 150;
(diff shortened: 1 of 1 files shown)


server/rooms.js
```diff
@@ -638,12 +638,17 @@ class GlobalRoom extends BasicRoom {
...
+			if (usernameFilter && ((p1userid && !p1userid.startsWith(usernameFilter)) && (p2userid && !p2userid.startsWith(usernameFilter)))) {
+				continue;
+			}
 			if (skipCount && skipCount--) continue;
 
 			rooms.push(room);
 		}
 
-		let roomTable = /** @type {{[roomid: string]: AnyObject}} */ ({});
+		let roomTable = /** @type {{[roomid: string]: {p1?: string, p2?: string, minElo?: string | number}}} */ ({});
```
scheibo:
```suggestion
		let roomTable = /** @type {{[roomid: string]: {p1?: string, p2?: string, minElo?: 'tour' | number}}} */ ({});
```

server/rooms.js
```diff
@@ -638,12 +638,17 @@ class GlobalRoom extends BasicRoom {
 			if (room.type !== 'battle') continue;
 			if (formatFilter && formatFilter !== room.format) continue;
 			if (eloFilter && (!room.rated || room.rated < eloFilter)) continue;
+			const p1userid = room.battle && room.battle.p1.userid;
+			const p2userid = room.battle && room.battle.p2.userid;
+			if (usernameFilter && ((p1userid && !p1userid.startsWith(usernameFilter)) && (p2userid && !p2userid.startsWith(usernameFilter)))) {
```
scheibo: confused by this conditional - isnt this saying both usernames need to start with the username filter?

TheJetOU: No, it will only skip the iteration if both don't start with the prefix.

TheJetOU: That conditional will only be `true` if both of them don't start with the prefix, if we used `||` it could be `true` even if one of them starts with the prefix. `true` in this case means 'skip' this battle.

scheibo: Maybe something like this would be more clear?

```ts
if (userfilter && room.battle) {
  if (!room.battle.p1.userid || !room.battle.p2.userid) continue;
  if (!(room.battle.p1.userid.startsWith(usernameFilter) || !room.battle.p2.userid.startsWith(usernameFilter)) continue;
}
```

PR #5664, opened 2019-08-01
OMotM: Fix crashes and oversights

KrisXV:


config/formats.js
```diff
@@ -689,7 +689,8 @@ let Formats = [
 			set.species = template.species;
 			return problems;
 		},
-		onModifyTemplate(template, target, effect) {
+		onModifyTemplate(template, target, source, effect) {
+			if (!effect || !target) return;
```
Zarel: I'm pretty sure `!effect` is unnecessary.

Also how does Cross Evolution work? Should `/dt` just ignore it entirely?

KrisXV: i dont really know how it works, its one of the few OMs i dont care enough to look into.

and yeah `!effect` is unnecessary, ill remove it

Zarel: Aha, it works on nicknames which are unavailable in `/dt`. That makes sense.

KrisXV: removed `!effect` @Zarel

config/formats.js
```diff
@@ -722,8 +723,8 @@ let Formats = [
 		},
 		onBegin() {
 			let allPokemon = this.p1.pokemon.concat(this.p2.pokemon);
-			for (let i = 0, len = allPokemon.length; i < len; i++) {
-				allPokemon[i].baseTemplate = allPokemon[i].template;
+			for (const pokemon of allPokemon) {
```
Slayer95:
```js
// stuff
}
```

PR #5665, opened 2019-08-01
Add UU suspect test

KrisXV:
https://www.smogon.com/forums/threads/np-sm-uu-stage-10-5-water-me-weather-test.3653094/


Zarel:
(This isn't just the notice, this is the entire suspect test itself.)

PR #5666, opened 2019-08-01
Scavengers: Add Recycled Hunts functionality

jmclemo6:
This adds a database of recycled hunts that Scavengers room staff can start and also auto-queue for times when the room is slow. This has been discussed with appropriate room staff.

The database itself is just JSON that is loaded and worked with in memory. Any changes that are made are immediately written back out to the JSON file.

The main way to interact with the database is through a new master command `/recycledhunts`. It is possible to add hunts, remove hunts, add hints, remove hints, and to enable/disable the auto-queue functionality through this command.

The existing queue and create commands were also updated to work with recycled hunts. Staff are able to queue and create either a random or specific recycled hunt from the database.

I think I updated and created all appropriate help text as well.

As always, let me know if anything needs changing and I'll take care of it.

### DIFF
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -253,21 +253,25 @@ class ScavengerHuntDatabase {
 		}
 
 		scavengersData.recycledHunts.push(huntSchema);
-		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+		this.updateDatabaseOnDisk();
 	}
 
 	static removeRecycledHuntFromDatabase(index) {
 		scavengersData.recycledHunts.splice(index - 1, 1);
-		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+		this.updateDatabaseOnDisk();
 	}
 
 	static addHintToRecycledHunt(huntNumber, questionNumber, hint) {
 		scavengersData.recycledHunts[huntNumber - 1].questions[questionNumber - 1].hints.push(hint);
-		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+		this.updateDatabaseOnDisk();
 	}
 
 	static removeHintToRecycledHunt(huntNumber, questionNumber, hintNumber) {
 		scavengersData.recycledHunts[huntNumber - 1].questions[questionNumber - 1].hints.splice(hintNumber - 1);
+		this.updateDatabaseOnDisk();
+	}
+
+	static updateDatabaseOnDisk() {
 		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
 	}
 
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -485,6 +485,26 @@ class ScavengerHunt extends Rooms.RoomGame {
 			reset = true;
 		}
 
+		if (this.room.addRecycledHuntsToQueueAutomatically) {
+			if (!this.room.scavQueue) {
+				this.room.scavQueue = [];
+			}
+
+			const next = ScavengerHuntDatabase.getRecycledHuntFromDatabase();
+			const correctlyFormattedQuestions = next.questions.reduce((alreadyFormatted, questionObject) => {
+				alreadyFormatted.push(questionObject.text);
+				alreadyFormatted.push(questionObject.answers);
+				return alreadyFormatted;
+			}, []);
+			this.room.scavQueue.push({
+				hosts: next.hosts,
+				questions: correctlyFormattedQuestions,
+				staffHostId: 'scavengermanager',
+				staffHostName: 'Scavenger Manager',
+				gameType: 'unrated',
+			});
+		}
+
 		if (!reset) {
 			let sliceIndex = this.gameType === 'official' ? 5 : 3;
 
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -35,9 +35,7 @@ const HISTORY_PERIOD = 6; // months
 const ScavengerGames = require("./scavenger-games");
 
 let scavengersData = {recycledHunts: []};
-try {
-	scavengersData = JSON.parse(FS.readIfExistsSync(DATABASE_FILE));
-} catch (e) {}
+scavengersData = JSON.parse(FS(DATABASE_FILE).readIfExistsSync());
 
 // convert points stored in the old format
 const scavsRoom = Rooms('scavengers');
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -1651,6 +1651,49 @@ let commands = {
 	},
 };
 
+const pages = {
+	recycledHunts(query, user, connection) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/scavengers.js
```diff
@@ -33,6 +34,11 @@ const HISTORY_PERIOD = 6; // months
 
 const ScavengerGames = require("./scavenger-games");
 
+let scavengersData = {};
+try {
+	scavengersData = require(`../../${DATABASE_FILE}`);
```
xfix: Please use `FS` instead, `require` caching will lead to confusion when hotpatching.

xfix: Please use `FS` instead, `require` caching will lead to confusion when hotpatching.

server/chat-plugins/scavengers.js
```diff
@@ -216,6 +222,73 @@ function formatOrder(place) {
...
+		// Return a random hunt from the database.
+		return scavengersData.recycledHunts[Math.floor(Math.random() * scavengersData.recycledHunts.length)];
+	}
+
+	static addRecycledHuntToDatabase(hosts, params) {
+		if (!scavengersData.recycledHunts) {
+			scavengersData.recycledHunts = [];
+		}
+
+
```
xfix: Replace two lines with a single one.

server/chat-plugins/scavengers.js
```diff
@@ -216,6 +222,73 @@ function formatOrder(place) {
...
+		const huntSchema = {
+			hosts: hosts,
+			questions: [],
+			get fullText() {
```
xfix: `JSON.stringify` serializes getters as retrieving the value. This is mostly fine, but will cause issues if `fullText` format ever changes, you may want to have `getFullText` outside of hunt to avoid storing it in JSON.

server/chat-plugins/scavengers.js
```diff
@@ -216,6 +222,73 @@ function formatOrder(place) {
...
+		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+	}
+
+	static removeHintToRecycledHunt(huntNumber, questionNumber, hintNumber) {
+		scavengersData.recycledHunts[huntNumber - 1].questions[questionNumber - 1].hints.splice(hintNumber - 1);
+		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+	}
+
+	static isEmpty() {
+		return scavengersData && scavengersData.recycledHunts && scavengersData.recycledHunts.length !== 0;
```
xfix: Get rid of those unnecessary null checks. `scavengersData` will always be non-null, and `scavengersData.recycledHunts` should be properly initialized to an empty array (it's not currently, but that would be a good idea to avoid bugs).

server/chat-plugins/scavengers.js
```diff
@@ -216,6 +222,73 @@ function formatOrder(place) {
...
+	}
+
+	static hasHunt(hunt_number) {
+		return !this.isEmpty() && !isNaN(hunt_number) && hunt_number > 0 && hunt_number <= scavengersData.recycledHunts.length;
```
xfix: Checking `this.isEmpty()` is an unnecessary special case.

server/chat-plugins/scavengers.js
```diff
@@ -416,6 +489,26 @@ class ScavengerHunt extends Rooms.RoomGame {
 			reset = true;
 		}
 
+		if (!ScavengerHuntDatabase.isEmpty() && this.room.addRecycledHuntsToQueueAutomatically) {
+			if (!this.room.scavQueue) {
+				this.room.scavQueue = [];
+			}
+
+			const next = ScavengerHuntDatabase.getRecycledHuntFromDatabase();
+			const correctlyFormattedQuestions = next.questions.reduce((alreadyFormatted, questionObject) => {
```
xfix: Please use `flatMap` instead. Using `reduce` like this is very confusing, and unnecessary.

server/chat-plugins/scavengers.js
jmclemo6: Should be good now.

server/chat-plugins/scavengers.js
jmclemo6: Now intializing `scavengersData.recycledHunts` to an empty array.

jmclemo6:
@xfix, changes done.

server/chat-plugins/scavengers.js
```diff
@@ -33,6 +34,11 @@ const HISTORY_PERIOD = 6; // months
 
 const ScavengerGames = require("./scavenger-games");
 
+let scavengersData = {recycledHunts: []};
+try {
+	scavengersData = JSON.parse(FS.readIfExistsSync(DATABASE_FILE));
+} catch (e) {}
```
xfix: Shouldn't be in `try`/`catch` anymore, `readIfExistsSync` should handle this for you.

server/chat-plugins/scavengers.js
```diff
@@ -216,6 +222,69 @@ function formatOrder(place) {
...
+	}
+
+	static addHintToRecycledHunt(huntNumber, questionNumber, hint) {
+		scavengersData.recycledHunts[huntNumber - 1].questions[questionNumber - 1].hints.push(hint);
+		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
+	}
+
+	static removeHintToRecycledHunt(huntNumber, questionNumber, hintNumber) {
+		scavengersData.recycledHunts[huntNumber - 1].questions[questionNumber - 1].hints.splice(hintNumber - 1);
+		FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));
```
xfix: All of `FS(DATABASE_FILE).writeUpdate(() => JSON.stringify(scavengersData));` probably should be its own function.

server/chat-plugins/scavengers.js
jmclemo6: `FS.readIfExistsSync` just returns a string and if it returns the empty string (what it does if the file doesn't exist) then should we catch the error thrown by `JSON.parse`?

server/chat-plugins/scavengers.js
```diff
@@ -33,6 +34,13 @@ const HISTORY_PERIOD = 6; // months
 
 const ScavengerGames = require("./scavenger-games");
 
+let scavengersData = {recycledHunts: []};
+try {
+	scavengersData = JSON.parse(FS(DATABASE_FILE).readIfExistsSync());
+} catch (err) {
+	console.log(err);
+}
```
xfix: Please don't abuse `try`/`catch` like this. I was thinking more along the lines of this.

```javascript
const contents = FS(DATABASE_FILE).readIfExistsSync();
const scavengersData = contents ? JSON.parse(contents) : {recycledHunts: []};
```

jmclemo6: I believe this would cause Scavengers to crash if the file can't be accessed for reasons other than it not existing. Is that alright?

xfix: Yes, and that's a good thing. If the database file couldn't be loaded, chances are things would break later on, and data would end up being lost.

PR #5669, opened 2019-08-03
 Prevent locked users from having Pokemon nicknames hidden in Cross Evolution

### LINKED ISSUE #5668: Prevent locked users from having Pokemon nicknames hidden in Cross Evolution
A Pokemon's nickname is required to determine its cross evolution in this month's OM, but #2639 causes locked users' Pokemons' nicknames to be unset.

ColeFrench:
This PR fixes #5668 by modifying the behavior of the `removeNicknames` flag for Cross Evolution: if a Pokemon's nickname represents the name of an actual Pokemon, it is not removed.

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -64,7 +64,11 @@ export class TeamValidator {
 			if (setProblems) {
 				problems = problems.concat(setProblems);
 			}
-			if (removeNicknames) set.name = dex.getTemplate(set.species).baseSpecies;
+			if (removeNicknames) {
+				if (format.name !== '[Gen 7] Cross Evolution' || !dex.getTemplate(set.name).exists) {
+					set.name = dex.getTemplate(set.species).baseSpecies;
+				}
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


sim/team-validator.ts
```diff
@@ -64,7 +64,11 @@ export class TeamValidator {
 			if (setProblems) {
 				problems = problems.concat(setProblems);
 			}
-			if (removeNicknames) set.name = dex.getTemplate(set.species).baseSpecies;
+			if (removeNicknames) {
+				if (format.name !== '[Gen 7] Cross Evolution' || !dex.getTemplate(set.name).exists) {
```
Zarel: This would allow any name that matches a template.

You might want to override with the exact template name, to prevent stuff like "pikachu 𝐅𝐔𝐂𝐊" from passing this check.

Zarel:
Other than that one issue, though, this looks great. Good catch!

Zarel:
Thanks!

PR #5670, opened 2019-08-03
Make linked moves target properly in Linked

ColeFrench:
Selecting a linked move which targets the user causes the follow-up move to fail if it targets a different Pokemon. This PR fixes that by causing each move to target its intended target, instead of using the target of the selected linked move.

### DIFF
--- a/data/mods/linked/scripts.js
+++ b/data/mods/linked/scripts.js
@@ -286,7 +286,7 @@ exports.BattleScripts = {
 				// @ts-ignore
 				let linkedMoves = action.linked;
 				for (let i = linkedMoves.length - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


Slayer95:
I guess this is a regression after a correctness update in the decisions engine?

If that's so, the patch to Linked that mimick former behavior would be to run ``this.validTargetLoc()``  for each sub-decision, and _only then_ autotarget if needed. (Relevant for Doubles). Could you confirm whether that still fixes the issue?

Slayer95:
Fixed in 045fe456. Note that:

- ``getTargetLoc()`` expects the targetted ``pokemon`` as its first argument, not a targetting type (like ``linkedMove.target`` is).
- By calling ``getTargetLoc()`` in other ways, it returns ``NaN``, which results in autotargetting happening by default. However, the proper way to do that would be with ``resolveTarget()`` as in the linked commit.

ColeFrench:
Forgive me; I'm relatively new to Showdown's API and thought `move.target` was a Pokemon, not a string describing the target type. Didn't catch it while testing since as you state, the `NaN` caused auto-targeting. Good going.

Slayer95:
You are welcome ~

PR #5671, opened 2019-08-04
Tournaments: always enforce scouting setting

sirDonovan:
Previously players were able to join battles as soon as they ended


Zarel:
Hm, what happens if a completed match expires?

sirDonovan:
I changed it slightly to avoid references to match rooms. Completed expiring matches already don't call `onBattleWin()` so now there shouldn't be any change in behavior.

sirDonovan:
@Zarel is this ok now?

Zarel:
Looks good, although maybe `completedMatches` should be an array?

PR #5672, opened 2019-08-05
Automatically consent to extracting your own battlelog

whalemer:


server/chat-commands.js
```diff
@@ -3895,6 +3895,7 @@ const commands = {
 		if (!battle.allowExtraction) {
 			battle.allowExtraction = ['', ''];
 		}
+		if (user.userid === toID(battle.p1.name) || user.userid === toID(battle.p2.name)) this.parse(`/allowexportinputlog ${user.userid}`);
```
```suggestion
		if (user.userid === toID(battle.p1.name) || user.userid === toID(battle.p2.name)) {
			this.parse(`/allowexportinputlog ${user.userid}`);
		}
```

Zarel:
Looks good, thanks!

PR #5678, opened 2019-08-05
Improve /hidereplay privacy

scheibo:
- hide the room as well so that it doesnt show up in a user's trainer card and end up auto-linking to replays
- upload a replay after the game to ensure the hidden setting is properly written

**NOTE:** this change does **not** fix private battles not properly getting passworded, but closes a few other loopholes that were found while investigating.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4095,6 +4095,7 @@ const commands = {
 			hidden: room.isPrivate || room.hideReplay ? '1' : '',
 			inputlog: battle.inputLog ? battle.inputLog.join('\n') : null,
 		});
+		if (success) battle.replySaved = true;
 		if (success && success.errorip) {
 			connection.popup(`This server's request IP ${success.errorip} is not a registered server.`);
 			return;
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4110,7 +4110,7 @@ const commands = {
 		if (!room || !room.battle || !this.can('joinbattle', null, room)) return;
 		if (room.hideReplay) return this.errorReply(`The replay for this battle is already set to hidden.`);
 		room.hideReplay = true;
-		this.parse('/savereplay');
+		if (room.battle.replaySaved) this.parse('/savereplay');
 		this.addModAction(`${user.name} hid the replay of this battle.`);
 	},
 
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4110,6 +4110,7 @@ const commands = {
 		if (!room || !room.battle || !this.can('joinbattle', null, room)) return;
 		if (room.hideReplay) return this.errorReply(`The replay for this battle is already set to hidden.`);
 		room.hideReplay = true;
+		this.parse('/savereplay');
 		this.addModAction(`${user.name} hid the replay of this battle.`);
 	},
 
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4095,7 +4095,7 @@ const commands = {
 			hidden: room.isPrivate || room.hideReplay ? '1' : '',
 			inputlog: battle.inputLog ? battle.inputLog.join('\n') : null,
 		});
-		if (success) battle.replySaved = true;
+		if (success) battle.replaySaved = true;
 		if (success && success.errorip) {
 			connection.popup(`This server's request IP ${success.errorip} is not a registered server.`);
 			return;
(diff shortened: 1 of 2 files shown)


server/room-battle.js
```diff
@@ -783,7 +783,15 @@ class RoomBattle extends RoomGames.RoomGame {
 			parentGame.onBattleWin(this.room, winnerid);
 		}
 		// If the room's replay was hidden, disable users from joining after the game is over
-		if (this.room.hideReplay) this.room.modjoin = '%';
+		if (this.room.hideReplay) {
+			this.room.modjoin = '%';
+			this.room.isPrivate = 'hidden';
```
scheibo: I'm going off the assumption that tracking `privateSetter` doesn't really matter here because it's not the end of the world if someone does `/publicroom` after - the main concern is the replay not being password protected.

server/room-battle.js
```diff
@@ -783,7 +783,15 @@ class RoomBattle extends RoomGames.RoomGame {
...
+			// We also save the replay to make sure the hidden setting gets reflected
```
Zarel: I would only do this if the replay already exists.

server/room-battle.js
Zarel: This shouldn't affect whether or not the replay is password protected – `hideReplay` should be enough.

scheibo: > NOTE: this change does not fix private battles not properly getting passworded, but closes a few other loopholes that were found while investigating.

Seriously, its like you don't read anything I write...

server/room-battle.js
scheibo: How do I determine if the replay already exists?

server/room-battle.js
Zarel: I suppose I understand which loophole this is.

Zarel: And yes, at this point I've been bombarded with too much information to closely read everything said to me. I might misread things.

server/room-battle.js
Zarel: Add a new property on the battle or room (probably battle) to track whether or not the replay has been saved.

server/room-battle.js
Zarel: I was originally opposed, but I guess it makes sense for the room to be hidden.

server/room-battle.js
```diff
@@ -783,7 +784,17 @@ class RoomBattle extends RoomGames.RoomGame {
 			parentGame.onBattleWin(this.room, winnerid);
 		}
 		// If the room's replay was hidden, disable users from joining after the game is over
-		if (this.room.hideReplay) this.room.modjoin = '%';
+		if (this.room.hideReplay) {
+			this.room.modjoin = '%';
+			this.room.isPrivate = 'hidden';
+			// We also resave the replay to make sure the hidden setting gets reflected
```
Zarel: I wonder if we should just do this as a general rule – save the replay at the end of a battle if `replaySaved` is true.

scheibo: Done.

scheibo:
My main concern with `replaySaved` is whether it covers all cases for uploading replays - can https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/client.js#L1257 upload a replay without setting `replaySaved`, thus causing us to fail to upload after `hideReplay`?

scheibo:
> My main concern with `replaySaved` is whether it covers all cases for uploading replays - can https://github.com/Zarel/Pokemon-Showdown-Client/blob/master/js/client.js#L1257 upload a replay without setting `replaySaved`, thus causing us to fail to upload after `hideReplay`?

```
[10:17:16] @Slayer95: scheibo, the client cannot upload replays without going through /svereplay
[10:18:04] @Slayer95: because /savereplay makes the sim server send a request to the replay server which tells it that a replay is gonna be uploaded, with a given MD5 hash
[10:18:31] @Slayer95: so it's the sim server authorizes the client to upload
```

Seems to resolve it.

server/chat-commands.js
```diff
@@ -4095,6 +4095,7 @@ const commands = {
 			hidden: room.isPrivate || room.hideReplay ? '1' : '',
 			inputlog: battle.inputLog ? battle.inputLog.join('\n') : null,
 		});
+		if (success) battle.replySaved = true;
```
Slayer95: ``s/replySaved/replaySaved``

scheibo: Thanks, I already caught this too.

Issue #2327, opened 2015-12-18
Genders can be hacked


TheImmortal:
It's possible to change any Pokemon's gender to male, female, or even genderless. The client (read: teambuilder) would also need to be updated to support this.

Definitely not important, but we might want to implement it eventually for the sake of following in-game limitations.

Proof: https://gyazo.com/5a6ae1f30ad090a52cdf1af2af8a1014

Slayer95:
Should we default Pokémon to genderless?

Zarel:
This isn't too hard to support, I think. Just stop validating gender in custom games.

TheImmortal:
I think the current defaulting to random legal ability is fine, even in hackmons.

hvk3:
Can I try and rectify this?

Zarel:
Yes...

TheImmortal:
This needs to be tested in previous gens. @Marty-D?

Marty-D:
I assume it's possible to do in Gen 5, and maybe Gen 3 and 4 too, but if it is I don't know how to do it. What I do know is gender is tied to PID, and messing with it may have other effects on a Pokemon in past gens where the PID is used for more than just gender (such as Ability or shininess).

Zarel:
This is now implemented for all gens.

Can you clarify re: gen 3-4? Presumably it's possible to generate random PIDs until you get one with the gender, ability, and shininess you want?

Slayer95:
Closing. If specific information arises regarding past gens and PIDs, we shall deal with it then.

PR #5679, opened 2019-08-05
Change packed encoding of abilities

scheibo:
0|1|H saves minimal space over the ability ID and requires parsers have the data files, in addition to limiting the ability to search abilities in the teambuilder.

---

Some thoughts:

**1)** I assume we need to keep the logic for decoding `0|1|H` in `fastUnpackTeam` for a while for backwards compatibility? 
**2)** Client side changes obviously need to be made as well (though maybe this is the only place we need the backwards compatibility?)
**3)** Can we change gender's encoding as well to make things completely data-free?

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -1097,29 +1097,14 @@ export class ModdedDex {
 			buf += (set.name || set.species);
 
 			// species
-			let id = toID(set.species || set.name);
+			const id = toID(set.species || set.name);
 			buf += '|' + (toID(set.name || set.species) === id ? '' : id);
 
 			// item
 			buf += '|' + toID(set.item);
 
 			// ability
-			const template = dexes['base'].getTemplate(set.species || set.name);
-			const abilities = template.abilities;
-			id = toID(set.ability);
-			if (abilities) {
-				if (id === toID(abilities['0'])) {
-					buf += '|';
-				} else if (id === toID(abilities['1'])) {
-					buf += '|1';
-				} else if (id === toID(abilities['H'])) {
-					buf += '|H';
-				} else {
-					buf += '|' + id;
-				}
-			} else {
-				buf += '|' + id;
-			}
+			buf += '|' + toID(set.ability);
 
 			// moves
 			buf += '|' + set.moves.map(toID).join(',');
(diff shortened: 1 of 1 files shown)


sim/dex.ts
```diff
@@ -1139,6 +1124,7 @@ export class ModdedDex {
 			}
 
 			// gender
+			const template = dexes['base'].getTemplate(set.species || set.name);
```
scheibo: If we just always encode gender (0-1 character...) we can remove the need for having to `getTemplate` at all

Slayer95: We support random gender.

Slayer95:
> I assume we need to keep the logic for decoding 0|1|H in fastUnpackTeam for a while for backwards compatibility?

Server-side, it would be helpful to keep backcompat for a while because of caching,
Client-side, we are stuck with supporting both formats forever, because of team storage.

> Re: Genders

Might need to ensure that we handle #2327 properly. Also, random genders support as commented inline.

sim/dex.ts
Slayer95: > We support random gender.

This is actually an argument _for_ doing as suggested by @scheibo .

Zarel: Sure.

Zarel:
I would keep backcompat forever in both.

sim/dex.ts
```diff
@@ -1139,6 +1124,7 @@ export class ModdedDex {
...
 			if (set.gender && set.gender !== template.gender) {
```
```suggestion
			if (set.gender) {
```

So yeah, I agree, let's do this.

sim/dex.ts
Zarel:
```suggestion
```

Zarel:
(This will also need to be done on client.)

scheibo:
> (This will also need to be done on client.)

> **2)** Client side changes obviously need to be made as well 

*smh*

Zarel:
Like I said in the other PR, I'm really swamped right now and might miss things.

scheibo:
I'd rather you wait to comment in those cases :(. This was not a high priority PR.

PR #5680, opened 2019-08-05
Let users see who made a hangman/poll

maxalexandderpi:
re: this approved suggestion. 
https://www.smogon.com/forums/threads/let-users-see-who-made-a-poll-or-hangman.3652444/


Zarel:
Looks good, although it doesn't really fix the problem described in the suggestion.

PR #5681, opened 2019-08-07
update blitz's description to proper timings

maxalexandderpi:
it should be saying 15 seconds per turn like it actually is, currently says 10 seconds.

### DIFF
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -418,7 +418,7 @@ let BattleFormats = {
 	blitz: {
 		effectType: 'Rule',
 		name: 'Blitz',
-		desc: "Super-fast 'Blitz' timer giving 30 second Team Preview and 10 seconds per turn.",
+		desc: "Super-fast 'Blitz' timer giving 30 second Team Preview and 15 seconds per turn.",
 		onBegin() {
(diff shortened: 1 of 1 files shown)


Zarel:
Ugh, this is wrong. It's 10 seconds per turn.

Zarel:
It's 15 seconds per turn burstable, 10 seconds per turn sustained.

The 15 seconds is there only to protect against internet hiccups. You can only ever use it once before it goes down to 10 seconds per turn.

scheibo:
`timer: {starting: 15, addPerTurn: 5, maxPerTurn: 15, maxFirstTurn: 30, grace: 30}`

???

Is there a magic `-5` somewhere applied to `maxPerTurn`?

scheibo:
I guess thats:

`addPerTurn: hasLongTurns ? 25 : 10`?

Zarel:
`maxPerTurn` is the maximum length a single turn can take.

The maximum length any turn can take is `min(maxPerTurn, current)`

`current` starts at `15`, and increases by `5` per turn. The timer has a resolution of 5, so `10` is the lowest `current` can be at the start of a turn (and is what it will consistently be if you max out the timer).

PR #5683, opened 2019-08-07
Fix when forcetimer starts the timer

scheibo:
Is this the most elegant way of solving things? No. But its an improvement over starting it at the wrong time.

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -520,7 +520,6 @@ class RoomBattle extends RoomGames.RoomGame {
 			this.addPlayer(options.p4, options.p4team || '', options.p4rating);
 		}
 		this.timer = new RoomBattleTimer(this);
-		if (Config.forcetimer) this.timer.start();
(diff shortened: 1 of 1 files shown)


server/room-battle.js
```diff
@@ -520,7 +520,6 @@ class RoomBattle extends RoomGames.RoomGame {
 			this.addPlayer(options.p4, options.p4team || '', options.p4rating);
 		}
 		this.timer = new RoomBattleTimer(this);
-		if (Config.forcetimer) this.timer.start();
```
Zarel: I would set an `autostart` setting in the timer or something?

scheibo: WDYM? Like `this.timer.autostart = 3` so it starts the 3rd time `this.timer.nextRequest` is called, and then revert the changes below?

Zarel: I'll write this myself.

PR #5685, opened 2019-08-08
Track turns in RoomBattle

TheJetOU:
Also use that new property in `/offertie`.

### DIFF
--- a/server/room-battle.js
+++ b/server/room-battle.js
@@ -705,10 +705,8 @@ class RoomBattle extends RoomGames.RoomGame {
 		switch (lines[0]) {
 		case 'update':
 			for (const line of lines.slice(1)) {
-				const tokens = line.slice(1).split("|");
-				const type = tokens.shift();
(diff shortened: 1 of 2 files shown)


server/room-battle.js
```diff
@@ -704,6 +705,11 @@ class RoomBattle extends RoomGames.RoomGame {
 		switch (lines[0]) {
 		case 'update':
 			for (const line of lines.slice(1)) {
+				const tokens = line.slice(1).split("|");
```
Zarel: This sounds like overkill for `line.startsWith('|turn|')`

Zarel: All the unnecessary tokenizing/parsing/allocation...

TheJetOU: Done here https://github.com/Zarel/Pokemon-Showdown/pull/5685/commits/7b2d824a11d1112fd352c1235ed048ddb826bd26.

PR #5686, opened 2019-08-08
Fix disabling mafia

whalemer:


Zarel:
I'm confused why this is necessary?

whalemer:
Deleting it removes it from the object, stopping it from getting saved.

Zarel:
Isn't that the point? To stop it from getting saved?

whalemer:
Mafia defaults to enabled, the property needs to be explicitly falsey to disable it, and it needs to be ``false`` to be saved.

Zarel:
Oh, that makes sense.

Zarel:
We should probably not have settings that work like that. I'd rather call it `mafiaDisabled` or something.

whalemer:
I thought about that but avoided it because it'd break backwards compatibility.

Though since disabling doesn't work, it's not like that'd change anything...

PR #5687, opened 2019-08-08
Use separate commands for banning literal strings and regex

whalemer:


bumbadadabum:
for the record I also told peach this is what I wanted.

PR #5688, opened 2019-08-08
Make /hotpatch fail if git history hasn't changed

scheibo:
/forcehotpatch can be used to perform the hotpatch regardless.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -90,7 +90,6 @@ const BROADCAST_TOKEN = '!';
 
 const TRANSLATION_DIRECTORY = 'translations/';
 
-import { exec, ExecException } from 'child_process';
 import { FS } from '../lib/fs';
 
 // @ts-ignore no typedef available
@@ -1758,8 +1757,6 @@ function stringify(value: any, depth = 0): string {
 }
 
 import { formatText, linkRegex, stripFormatting } from './chat-formatter';
-// tslint:disable-next-line: prefer-const exported
-let updateServerLock = false;
 
 /**
  * Gets the dimension of the image at url. Returns 0x0 if the image isn't found, as well as the relevant error.
@@ -1847,36 +1844,6 @@ function resolvePage(pageid: string, user: User, connection: Connection) {
 	return (new PageContext({pageid, user, connection})).resolve();
 }
 
-async function version() {
-	function sh(command: string, path: string): Promise<[number, string, string]> {
-		return new Promise((resolve, reject) => {
-			exec(command, {
-				cwd: __dirname,
-				env: {GIT_INDEX_FILE: path},
-			}, (error: ExecException | null, stdout: string | Buffer, stderr: string | Buffer) => {
-				resolve([error && error.code || 0, '' + stdout, '' + stderr]);
-			});
-		});
-	}
-
-	let hash;
-	try {
-		await FS('.git/index').copyFile('logs/.gitindex');
-		const index = FS('logs/.gitindex');
-
-		let [code, stdout, stderr] = await sh(`git add -A`, index.path);
-		if (code || stderr) return;
-		[code, stdout, stderr] = await sh(`git write-tree`, index.path);
-
-		if (code || stderr) return;
-		hash = stdout.trim();
-
-		await sh(`git reset`, index.path);
-		await index.unlinkIfExists();
-	} catch (err) {}
-	return hash;
-}
-
 export const Chat = {
 	multiLinePattern,
 	baseCommands,
@@ -1944,7 +1911,4 @@ export const Chat = {
 	formatText,
 	linkRegex,
 	stripFormatting,
-
-	updateServerLock,
-	version,
 };
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -90,6 +90,7 @@ const BROADCAST_TOKEN = '!';
 
 const TRANSLATION_DIRECTORY = 'translations/';
 
+import { ExecException, exec } from 'child_process';
 import { FS } from '../lib/fs';
 
 // @ts-ignore no typedef available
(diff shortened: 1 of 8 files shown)


scheibo:
Draft because this doesn't work properly and looking for suggestions, main issues:

1) `global.__version.head` isn't defined most of the time
2) `/hotpatch chat` wipes the `HOTPATCH_VERSIONS`

Slayer95:
- We should be disallowing ``/forcehotpatch`` if ``user.lastCommand`` isn't ``/hotpatch``, to prevent laziness (Note: ``user.lastCommand`` isn't automatically set by the parser)
- (Also, I previously suggested requiring sysop permissions for ``/forcehotpatch``, but that would unintendendly invalidate some workflows, so I retract it.)

> /hotpatch chat wipes the HOTPATCH_VERSIONS

``HOTPATCH_VERSIONS`` should probably go under ``Monitor`` - ``hotpatchLock`` is there already.

> global.__version.head isn't defined most of the time

I'd make it a getter which execs ``git`` if not cached yet, but we might want another approach.

scheibo:
> We should be disallowing `/forcehotpatch` if `user.lastCommand` isn't `/hotpatch`, to prevent laziness (Note: `user.lastCommand` isn't automatically set by the parser)

I don't personally care about this. If people want to be lazy about this they can `/forcehotpatch` 100% of the time. **I** won't use it, but if other people want to be more cavalier then they can go for it.

panpawn:
>global.__version.head isn't defined most of the time

You could probably store this in a JSON file for persistence

server/index.js
```diff
@@ -95,6 +95,7 @@ global.LoginServer = require('../.server-dist/loginserver').LoginServer;
 global.Ladders = require('./ladders');
 
 global.Chat = require('./chat');
+global.__version = {head: global.Chat.version()};
```
panpawn: Can't ``global.Chat.version()`` here just be ``Chat.version()``?

Slayer95: Triggers the linter IIRC

scheibo:
> You could probably store this in a JSON file for persistence

I'm OK with it just occasionally not being defined in some edge cases.

server/index.js
scheibo: I believe its usually `tsc` that complains abut this, but this file isnt typechecked so...

server/index.js
```diff
@@ -95,6 +95,7 @@ global.LoginServer = require('../.server-dist/loginserver').LoginServer;
 global.Ladders = require('./ladders');
 
 global.Chat = require('./chat');
+global.__version = {head: Chat.version()};
```
Zarel: I would put this in `Chat` instead of introducing another global.

Zarel:
I guess this is so you don't accidentally forget to `/updateserver` before `/hotpatch`? An interesting solution.

Zarel:
> I don't personally care about this. If people want to be lazy about this they can `/forcehotpatch` 100% of the time. **I** won't use it, but if other people want to be more cavalier then they can go for it.

I agree, unlike the other force commands, `/forcehotpatch` is the old hotpatch behavior, and I've never had a problem with it.

Zarel:
I'm not _particularly_ happy with this because I commonly use hotpatch to test uncommitted code, but I suppose five letters isn't a huge difference.

server/index.js
scheibo: it would be erased on `/hotpatch chat`

scheibo:
> I commonly use hotpatch to test uncommitted code

I mean, I think its better not to...

_"I commonly aim loaded guns at my feet..."_

server/index.js
scheibo: Also, I didn't introduce a new global, this global already exists, its just only set in room-battle and battle.ts.

Zarel: Oh, fair, I guess this makes sense, then.

Zarel:
> > I commonly use hotpatch to test uncommitted code
> 
> I mean, I think its better not to...
> 
> _"I commonly aim loaded guns at my feet..."_

I think it's better to test code before committing it... You do realize I mean test on a local server, right?

bumbadadabum:
Yeah this is going to make it so much harder to test code on a local server.

bumbadadabum:
I think there are better ways to remind you to updateserver

panpawn:
>Yeah this is going to make it so much harder to test code on a local server.

We could make this a ``Config`` setting. Maybe ``Config.hotpatchgitcheck`` or something? This way, people can easily disable it if they're testing code locally

scheibo:
> Yeah this is going to make it so much harder to test code on a local server.

So much harder = 5 extra characters.... get real

Also, we could change the commit hash to also include local code, so local code changes would not require a `/forcehotpatch`?

Zarel:
I like the idea of including local code in the commit hash, but how would you do it?

scheibo:
> I like the idea of including local code in the commit hash, but how would you do it?

The following seems to work:

```sh
cp .git/index /tmp/git_index
GIT_INDEX_FILE=/tmp/git_index git add -A
GIT_INDEX_FILE=/tmp/git_index git write-tree
GIT_INDEX_FILE=/tmp/git_index git reset
rm /tmp/git_index
```

Basically, fork the index, add everything, get the SHA and then revert. The last two lines are kind of optional - no real need to reset the temp index, and no need to delete the temp index if a subsequent run will just override it.

Zarel:
Condolences for the TypeScript refactor merge conflict.

server/index.js
```diff
@@ -95,6 +95,9 @@ global.LoginServer = require('../.server-dist/loginserver').LoginServer;
 global.Ladders = require('./ladders');
 
 global.Chat = require('./chat');
+Chat.version().then(function (hash) {
```
Zarel: This sounds like a race condition in the making?

Maybe you should do something like

```suggestion
global.__verison = {};
Chat.version().then(function (hash) {
```

scheibo:
FUCK

server/chat.ts
```diff
@@ -1914,4 +1946,5 @@ export const Chat = {
 	stripFormatting,
 
 	updateServerLock,
+	version,
```
Slayer95: Can we move these to ``Monitor``?

Slayer95: https://github.com/Zarel/Pokemon-Showdown/search?q=updateServerLock&unscoped_q=updateServerLock

scheibo: Done.

server/monitor.ts
```diff
@@ -274,8 +276,39 @@ export const Monitor = {
 		return bytes;
 	},
 
+	async version() {
+		function sh(command: string, path: string): Promise<[number, string, string]> {
```
Zarel: Maybe pull out this function?

Zarel: Into like `Monitor.sh` at the very least.

Slayer95: Then I'd suggest making its signature ``sh(command, options)``, since the ``options`` passed to ``exec`` (including the current ``path``) are the same in all of its calls,

PR #5690, opened 2019-08-09
Set prevUsers properly when merging

whalemer:
As far as I can tell this isn't intended behaviour.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -99,9 +99,9 @@ function deleteUser(user: User) {
 	prevUsers.delete('guest' + user.guestNum as ID);
 	users.delete(user.userid);
 }
-function merge(user1: User, user2: User) {
-	prevUsers.delete(user2.userid);
-	prevUsers.set(user1.userid, user2.userid);
+function merge(fromUser: User, intoUser: User) {
+	prevUsers.delete(intoUser.userid);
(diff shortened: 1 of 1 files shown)


Slayer95:
LGTM :+1:

Slayer95:
Hmm.... Actually, ``merge``'s API should probably be ``(fromUser, intoUser)``, matching the API of ``Object.assign`` and the like.

whalemer:
>...(fromUser, intoUser), matching the API of Object.assign and the like.

Isn't `Object.assign` (destination, source)?

Slayer95:
Uhm... Yea, that's what I mean, ``(intoUser, fromUser)`` :sweat_smile:

Zarel:
Why is there a function just called `merge` here, anyway?

Zarel:
I think `destination` and `source` are just not very clear names. They're better than `1` and `2` but I'd rather think of better names.

Maybe `toRemain`, `toDestroy`?

whalemer:
>Why is there a function just called merge here, anyway?

```
// Low-level functions for manipulating Users.users and Users.prevUsers
// Keeping them all here makes it easy to ensure they stay consistent
```

formerly assigned as Users.merge but typescript doesn't like that

PR #5692, opened 2019-08-11
TypeScript RoomGame and RoomBattle

TheJetOU:


server/room-battle.ts
```diff
@@ -432,25 +447,39 @@ class RoomBattleTimer {
...
+	p4: RoomBattlePlayer;
+	allowExtraction: [ID, ID] | null;
+	logData: AnyObject | null;
+	endType: string;
+	score: number[] | null;
+	inputLog: string | null;
+	turn: number;
+	rqid: number;
+	requestCount: number;
+	stream: import("../lib/streams").ObjectReadWriteStream<string>;
```
scheibo: please move this to the top with the other imports

server/room-battle.ts
```diff
@@ -1211,42 +1173,28 @@ class RoomBattleStream extends BattleStream {
...
-const PM = new StreamProcessManager(module, () => {
+export const PM = new StreamProcessManager(module, () => {
 	return new RoomBattleStream();
 });
 
 if (!PM.isParentProcess) {
 	// This is a child process!
-	global.Config = require(/** @type {any} */('../.server-dist/config-loader')).Config;
-	global.Chat = require(/** @type {any} */('../.server-dist/chat'));
+	global.Config = (require('./config-loader') as any).Config;
```
scheibo: i don't think this or below still needs to be `as any`? these have been typescripted?

TheJetOU: As discussed in https://psim.us/dev, I initially did this to prevent TSLint from complaining about this not using `import`, but it seems like disabling it is the preferred action here.

PR #5695, opened 2019-08-12
Don't clear player.userid upon unlinking

TheJetOU:
Approved by @Zarel in https://github.com/Zarel/Pokemon-Showdown/pull/5456#discussion_r312765366 and needed for #5456.


Zarel:
Wait, nothing else needs to happen for this change? Are you sure?

TheJetOU:
I don't think so?

What else do you think needs to happen?

Zarel:
I suppose there isn't; I'm just surprised there isn't.

PR #5697, opened 2019-08-12
Update a link in the /intro command

RyotaMitaraiWeb:
The link in /intro about "What do 'OU' and etc. mean" is from BW era so I am pushing an edit to replace it with the SM one.


HoeenCoder:
Thanks!

PR #5698, opened 2019-08-13
Updated /intro

ayiadvance:
This splits the "What are the rules for each format? What is 'Sleep Clause'?" part of /intro into two questions, both with updated links.

### DIFF
--- a/server/chat-plugins/info.js
+++ b/server/chat-plugins/info.js
@@ -1512,7 +1512,7 @@ const commands = {
 			`- <a href="https://www.smogon.com/forums/threads/3496279/">Beginner's Guide to Pok&eacute;mon Showdown</a><br />` +
 			`- <a href="https://www.smogon.com/dp/articles/intro_comp_pokemon">An introduction to competitive Pok&eacute;mon</a><br />` +
 			`- <a href="https://www.smogon.com/sm/articles/sm_tiers">What do 'OU', 'UU', etc mean?</a><br />` +
-      `- <a href="https://www.smogon.com/dex/sm/formats/">What are the rules for each format?</a><br />` +
+			`- <a href="https://www.smogon.com/dex/sm/formats/">What are the rules for each format?</a><br />` +
 			`- <a href="https://www.smogon.com/sm/articles/clauses">What is 'Sleep Clause' and other clauses?</a>`
 		);
 	},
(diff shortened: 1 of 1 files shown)


Zarel:
This is intentional. "What is 'sleep clause'?" is an add-on question, intended to clarify what sorts of rules are being asked about. The rules _are_ explained in the tier explanation pages.

TheJetOU:
The page is a bit outdated though since it's from XY, and I don't think there's a page like that for SM? So maybe splitting it is the only available option.

Zarel:
I do agree with updating the link to https://www.smogon.com/dex/sm/formats/. I just don't see a reason to split it.

TheJetOU:
Yeah, but that page doesn't really explain the clauses. Well, I guess if you click one of the formats and scroll down a bit, the clauses are explained. But the other page is way more straight forward.

Zarel:
It literally does explain the clauses, you said so yourself.

The other page is _less_ straightforward. It doesn't explain the formats, only the clauses. The link is about the formats.

TheJetOU:
My only gripe with your suggestion is that it says "What is Sleep Clause?" but you have to do a little bit of digging to find out what it means, but I suppose the "digging" here is minuscule, so it's not a big deal.

HoeenCoder:
FYI:
1. This is another one of my students (so still a very to JS in general)
2. Iyarito actually asked me to make this split, I personally have no problem with the split.

server/chat-plugins/info.js
```diff
@@ -1512,7 +1512,8 @@ const commands = {
 			`- <a href="https://www.smogon.com/forums/threads/3496279/">Beginner's Guide to Pok&eacute;mon Showdown</a><br />` +
 			`- <a href="https://www.smogon.com/dp/articles/intro_comp_pokemon">An introduction to competitive Pok&eacute;mon</a><br />` +
 			`- <a href="https://www.smogon.com/sm/articles/sm_tiers">What do 'OU', 'UU', etc mean?</a><br />` +
-			`- <a href="https://www.smogon.com/xyhub/tiers">What are the rules for each format? What is 'Sleep Clause'?</a>`
+      `- <a href="https://www.smogon.com/dex/sm/formats/">What are the rules for each format?</a><br />` +
```
HoeenCoder: Indentation appears to be off here actually, be sure to correct that.

Zarel:
I suppose I'm fine with the split...

PR #5699, opened 2019-08-13
TypeScript Rooms and Roomlogs

TheJetOU:


server/roomlogs.ts
```diff
@@ -30,16 +30,21 @@ const FS = require(/** @type {any} */('../.lib-dist/fs')).FS;
...
+	isMultichannel: boolean;
+	autoTruncate: boolean;
+	logTimes: boolean;
+	modlogStream?: WriteStream | null;
+	roomlogStream?: WriteStream | null;
+	sharedModlog: boolean;
+	roomlogFilename: string;
+	constructor(room: BasicChatRoom, options: {isMultichannel?: any, autoTruncate?: any, logTimes?: any} = {}) {
 		this.id = room.id;
 		/**
```
Zarel: Doc comments go on the definition.

server/roomlogs.ts
```diff
@@ -276,33 +262,26 @@ class Roomlog {
 	}
 }
 
-/** @type {Map<string, WriteStream>} */
-const sharedModlogs = new Map();
+const sharedModlogs: Map<string, WriteStream> = new Map();
```
Zarel:
```suggestion
const sharedModlogs = new Map<string, WriteStream>();
```

server/roomlogs.ts
```diff
@@ -276,33 +262,26 @@ class Roomlog {
...
 
-/** @type {Map<string, Roomlog>} */
-const roomlogs = new Map();
+const roomlogs: Map<string, Roomlog> = new Map();
```
Zarel:
```suggestion
const roomlogs = new Map<string, Roomlog>();
```

server/rooms.ts
```diff
@@ -19,63 +19,113 @@ const CRASH_REPORT_THROTTLE = 60 * 60 * 1000;
...
-		/** @type {{[userid: string]: User}} */
 		this.users = Object.create(null);
 		this.userCount = 0;
 
-		/** @type {'chat' | 'battle' | 'global'} */
 		this.type = 'chat';
-		/** @type {?{[userid: string]: string}} */
 		this.auth = null;
 
 		/**
```
Zarel: Again, doc comments go on the declaration, not the assignment.

server/roomlogs.ts
scheibo: Did not know that worked! Cool!

server/users.ts
```diff
@@ -1156,7 +1156,7 @@ export class User extends Chat.MessageContext {
 		this.isStaff = Config.groups[this.group] && (Config.groups[this.group].lock || Config.groups[this.group].root);
 		if (!this.isStaff) {
 			const staffRoom = Rooms('staff');
-			this.isStaff = (staffRoom && staffRoom.auth && staffRoom.auth[this.userid]);
+			this.isStaff = Boolean(staffRoom && staffRoom.auth && staffRoom.auth[this.userid]);
```
```suggestion
			this.isStaff = !!(staffRoom && staffRoom.auth && staffRoom.auth[this.userid]);
```

`Boolean` may be more explicit but all the rest of our code uses `!!` – may as well stay consistent until we ever decide to change it all at once.

server/users.ts
```diff
@@ -1187,7 +1187,7 @@ export class User extends Chat.MessageContext {
...
```
```suggestion
			this.isStaff = !!(staffRoom && staffRoom.auth && staffRoom.auth[this.userid]);
```

server/tournaments/index.js
```diff
@@ -1073,7 +1074,7 @@ class Tournament extends Rooms.RoomGame {
 		this.completedMatches.add(room.id);
 		room.parent = null;
 		if (!room.battle) throw new Error("onBattleWin called without a battle");
-
+		if (!room.p1 || !room.p2) throw new Error("onBattleWin called with missing players");
```
Zarel: I think it makes sense to use `!` assertions, rather than add throws that don't add any information than the crash that would otherwise happen here.

I would even use `!` in place of the above `room.battle` check.

server/rooms.ts
```diff
@@ -19,63 +19,113 @@ const CRASH_REPORT_THROTTLE = 60 * 60 * 1000;
...
 const LAST_BATTLE_WRITE_THROTTLE = 10;
 
-/** @type {null} */
 const RETRY_AFTER_LOGIN = null;
 
-/** @type {typeof import('../lib/fs').FS} */
-const FS = require(/** @type {any} */('../.lib-dist/fs')).FS;
-const Roomlogs = require('./roomlogs');
+import {FS} from '../lib/fs';
+import {WriteStream} from '../lib/streams';
```
Zarel: Is this just being used as a type? I'd rather import it as a type if so.

```suggestion
type WriteStream = import('../lib/streams').WriteStream;
```

server/rooms.ts
```diff
@@ -19,63 +19,113 @@ const CRASH_REPORT_THROTTLE = 60 * 60 * 1000;
...
+	minElo?: 'tour' | number;
+}
+
+export type Room = GlobalRoom | GameRoom | ChatRoom;
+type User = import('./users').User;
+type Connection = import('./users').Connection;
+type Roomlog = import('./roomlogs').Roomlog;
+type Poll = import('./chat-plugins/poll').PollType;
+
+export abstract class BasicRoom {
```
Zarel: nice

Zarel: (I'm happy this can be an abstract class now that we're using TypeScript)

server/rooms.ts
```diff
@@ -129,91 +167,71 @@ class BasicRoom {
...
-	add(message) { throw new Error(`should be implemented by subclass`); }
-	roomlog(/** @type {string} */ message) { throw new Error(`should be implemented by subclass`); }
-	modlog(/** @type {string} */ message) { throw new Error(`should be implemented by subclass`); }
+	add(message: string): this { throw new Error(`should be implemented by subclass`); }
+	roomlog(message: string) { throw new Error(`should be implemented by subclass`); }
+	modlog(message: string) { throw new Error(`should be implemented by subclass`); }
 	logEntry() { throw new Error(`room.logEntry has been renamed room.roomlog`); }
 	addLogMessage() { throw new Error(`room.addLogMessage has been renamed room.addByUser`); }
-	/**
-	 * Inserts (sanitized) HTML into the room log.
```
Zarel: You accidentally deleted this doc comment.

server/rooms.ts
```diff
@@ -225,14 +243,14 @@ class BasicRoom {
...
-		for (let i in this.users) {
+		for (const i in this.users) {
 			if (!this.users[i].named) {
 				continue;
 			}
 			counter++;
 			buffer += ',' + this.users[i].getIdentityWithStatus(this.id);
 		}
-		let msg = '|users|' + counter + buffer;
+		const msg = '|users|' + counter + buffer;
```
```suggestion
		const msg = `|users|${counter}${buffer}`;
```

If you're going to change this line, you might as well convert it to modern string standards.

server/rooms.ts
```diff
@@ -697,49 +707,38 @@ class GlobalRoom extends BasicRoom {
...
-	/**
-	 * @param {string} title
-	 */
-	addChatRoom(title) {
-		let id = toID(title);
-		if (id === 'battles' || id === 'rooms' || id === 'ladder' || id === 'teambuilder' || id === 'home' || id === 'all' || id === 'public') return false;
+	addChatRoom(title: string) {
+		const id = toID(title);
+		if (id === 'battles' || id === 'rooms' || id === 'ladder' || id === 'teambuilder' ||
+			id === 'home' || id === 'all' || id === 'public') return false;
```
```ts
if (['battles', 'rooms', 'ladder', 'teambuilder', 'home', 'all', 'public'].includes(id)) {
  return false;
}
```

server/rooms.ts
TheJetOU: Yes, its value is being used in `ladderIpLog`.

server/tournaments/index.js
Zarel: Ohhh. Yeah, this is fine, then.

Zarel: (I might prefer `@ts-ignore`)

server/rooms.ts
```diff
@@ -1054,103 +1037,85 @@ class BasicChatRoom extends BasicRoom {
...
 		}
 
-		/** @type {Map<string, ChatRoom>?} */
 		this.subRooms = null;
 
-		/** @type {'chat' | 'battle'} */
 		this.type = 'chat';
-		/** @type {boolean} */
 		this.active = false;
 		// TypeScript bug: subclass null
```
```suggestion
```

This isn't a problem in real TypeScript, as you can tell by how you could remove the doc comment.

server/rooms.ts
```diff
@@ -1054,103 +1037,85 @@ class BasicChatRoom extends BasicRoom {
...
 			if (Config.loguserstats) {
 				this.logUserStatsInterval = setInterval(() => this.logUserStats(), Config.loguserstats);
 			}
 		}
 
+		this.userList = '';
 		if (this.batchJoins) {
 			this.userList = this.getUserList();
 		}
 		// TypeScript bug: subclass member
```
```suggestion
```

server/tournaments/index.js
TheJetOU: I mean, `@ts-ignore` over 6 lines of code, or just an if-condition crash. I suppose you could reduce the amount of `@ts-ignore` by declaring a variable with `room.battle` as the value and disabling the compiler there, but that looks a bit awkward since there's already a `p1` variable.

Zarel: I thought it was only one line.

Zarel: I'd prefer an if condition over six lines of `ts-ignore`

TheJetOU: Wait, I was wrong, it would need 3 ``@ts-ignore``, which is still a line more than an if condition.

server/chat-plugins/poll.js
```diff
@@ -306,7 +307,7 @@ const commands = {
 				if (room.poll.timeout) clearTimeout(room.poll.timeout);
 				room.poll.timeoutMins = timeout;
 				room.poll.timeout = setTimeout(() => {
-					room.poll.end();
+					if (room.poll) room.poll.end();
 					delete room.poll;
```
Zarel: `room.poll = null` probably?

server/punishments.js
```diff
@@ -1526,8 +1527,9 @@ const Punishments = new (class {
 	checkNewNameInRoom(user, userid, roomid) {
 		/** @type {Punishment?} */
 		let punishment = Punishments.roomUserids.nestedGet(roomid, userid) || null;
-		if (!punishment && Rooms(roomid).parent) {
-			punishment = Punishments.checkNewNameInRoom(user, userid, Rooms(roomid).parent.id);
+		const room = Rooms(roomid);
```
Zarel:
```suggestion
		const room = !punishment && Rooms(roomid);
```

Save a call if the punishment doesn't exist.

Zarel: Actually, the right way probably involves multi-nesting

```
if (!punishment) {
  room = ...
  if (room.parent) {
```
etc

server/chat-plugins/poll.js
Zarel: Wow, yeah, I'm surprised this passed the linter.

```suggestion
					room.poll = null;
```

server/chat-plugins/poll.js
```diff
@@ -306,7 +307,7 @@ const commands = {
...
 				}, (timeout * 60000));
```
Zarel:
```suggestion
				}, timeout * 60000);
```

server/rooms.ts
```diff
@@ -602,33 +611,34 @@ class GlobalRoom extends BasicRoom {
...
 		Config.rankList = '|customgroups|' + JSON.stringify(rankList) + '\n';
 		return Config.rankList;
 	}
 
-	/**
-	 * @param {string} filter "formatfilter, elofilter, usernamefilter"
-	 */
-	getBattles(filter) {
-		let rooms = /** @type {GameRoom[]} */ ([]);
+	getBattles(/** "formatfilter, elofilter, usernamefilter */ filter: string) {
```
scheibo:
```suggestion
	getBattles(/** formatfilter, elofilter, usernamefilter */ filter: string) {
```

scheibo: Whatever, fuck it.

PR #5702, opened 2019-08-13
Simplify Map declarations

TheJetOU:


Zarel:
I wish there were a lint rule for this, tbh.

I do appreciate you going back and fixing these things when I notice them on your most recent PR.

PR #5705, opened 2019-08-14
Role Clarifications in /mafia dt

ayiadvance:
Created distinctions between:
Role Cop and Role Cop (Classic)
Red Goo and Red Goo (Cracking Idea)
Neapolitan and Vanilla Cop

Also clarified some roles and standardized how Idea Variant roles are listed.


YashGreninja:
If we are doing distinctions within /mafia dt, can we also implement Silencer and Silencer GI (since those are distinctly different)?

server/chat-plugins/mafia-data.js
```diff
@@ -390,7 +402,7 @@ const roles = {
 	coroner_cracking_idea: `coronercrackingidea`,
 	coronercrackingidea: {
 		name: `Coroner (Cracking Idea)`,
-		id: `coronercrackingidea`,
+		id: `coronerci`,
```
HoeenCoder: Why is this ID changing? it dosent match the property now.

server/chat-plugins/mafia-data.js
```diff
@@ -530,12 +542,24 @@ const roles = {
...
+	graygooci: `greygoocrackingidea`,
+	gray_goo_ci: `greygoocrackingidea`,
+	gray_goo_cracking_idea: `greygoocrackingidea`,
+	graygoocrackingidea: `greygoocrackingidea`,
+	greygooci: `greygoocrackingidea`,
+	grey_goo_ci: `greygoocrackingidea`,
+	grey_goo_cracking_idea: `greygoocrackingidea`,
+	greygoocrackingidea: {
+		name: `Grey Goo (Cracking Idea)`,
+		id: `greygooci`,
```
HoeenCoder: ID should not match an alias, but rather the main property eg: `greygoocrackingidea`

server/chat-plugins/mafia-data.js
```diff
@@ -844,9 +868,12 @@ const roles = {
...
 	},
-	red_goo: {
-		name: `Red Goo`,
-		id: `redgoo`,
+	redgooci: `redgoocrackingidea`,
+	red_goo_ci: `redgoocrackingidea`,
+	red_goo_cracking_idea: `redgoocrackingidea`,
+	redgoocrackingidea: {
+		name: `Red Goo (Cracking Idea)`,
+		id: `redgooci`,
```
HoeenCoder: `id` should match the property

server/chat-plugins/mafia-data.js
HoeenCoder: `id` should be the ID version (all lowercase, no spaces or symbols, keep numbers) of the property (eg: `coronercrackingidea`).

server/chat-plugins/mafia-data.js
HoeenCoder: Same ID issue here.

server/chat-plugins/mafia-data.js
HoeenCoder: Same ID issue here.

server/chat-plugins/mafia-data.js
```diff
@@ -924,6 +951,14 @@ const roles = {
 		id: `silencer`,
 		memo: [`Silencer: During the Night, you may PM the host the name of another player. This player will not be able to talk during the next day. Ask the host whether this player will be able to lynch or not.`],
 	},
+	silencer_gi: `silencergreateridea`,
+	silencergi: `silencergreateridea`,
+	silencer_cracking_idea: `silencergreateridea`,
+	silencergreateridea: {
+		name: `Silencer (Greater Idea)`,
+		id: `silencergi`,
```
HoeenCoder: Same ID issue here.

server/chat-plugins/mafia-data.js
```diff
@@ -359,10 +359,22 @@ const roles = {
...
+		image: `<img width="75" height="75" src="//${Config.routes.client}/fx/mafia-cop.png"/>`,
+	},
+	role_cop_classic: `rolecopclassic`,
+	rolecopclassic: {
+		name: `Role Cop (Classic)`,
+		id: `rolecopclassic`,
+		memo: [`Role Cop: Each night you can PM the host the name of another player. You will be told their exact role and alignment, or NO RESULT if your investigation failed.`],
+		image: `<img width="75" height="75" src="//${Config.routes.client}/fx/mafia-cop.png"/>`,
+	},
+	vanilla_cop: {
```
HoeenCoder: Isn't this the same as the neapolitan but under a different name?

ayiadvance: No, Vanilla Cop checks Vanilla / Not Vanilla, while Neapolitan specifically checks VT / not VT. It's subtle but it has very different uses in setups.

PR #5706, opened 2019-08-14
Auto-accept challenges from challenge target

scheibo:
**Context:** https://www.smogon.com/forums/threads/fixing-simultaneous-challenges.3653287/

I don't think this precludes us from *also* improving the UI, but defense in depth is never a bad thing.

```
[09:51] Zarel: we might also want to think more about this, to see if we can think of better solutions
[09:51] Zarel: simultaneously displaying a challenge box and a being-challenged screen would help
[09:52] Zarel: so when you saw you were being challenged, you would stop challenging your opponent
```

### DIFF
--- a/server/ladders.ts
+++ b/server/ladders.ts
@@ -211,6 +211,21 @@ class Ladder extends LadderStore {
 		}
 		const ready = await this.prepBattle(connection);
 		if (!ready) return false;
+		// If our target is already challenging us in the same format,
+		// simply accept the pending challenge instead of creating a new one.
+		const targetChalls = Ladders.challenges.get(targetUser.userid);
+		if (targetChalls) {
+			for (const chall of targetChalls) {
+				if (chall.from === targetUser.userid &&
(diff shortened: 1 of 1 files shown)


TheImmortal:
I have a feeling this will create complaints such as "challenged with the wrong team and it auto-accepted before I could cancel".

Personally, I would have just sent the 2nd challenger an error popup saying X is already challenging you to a Y game.

scheibo:
> I have a feeling this will create complaints such as "challenged with the wrong team and it auto-accepted before I could cancel".

This is always a risk with any 'cancel' mechanism - there's no guarantee your opponent wont make a decision before you decide to cancel

> Personally, I would have just sent the 2nd challenger an error popup saying X is already challenging you to a Y game.

This is also an option. Feel free to change L222-L224 where we accept the challenge and instead return a popup. The tradeoff is then you force the player to navigate through their teams and find the exact same team they just issued the challenge with before accepting. Both options have tradeoffs.

PR #5711, opened 2019-08-17
Add Aerodactyl Event

tmagicturtle:
Being distributed as of today. Notable change is that it now has access to Celebrate.
https://www.pokemon.com/us/play-pokemon/worlds/2019/pokemon-distribution/


Marty-D:
Thanks!

PR #5713, opened 2019-08-17
Add margin in helptickets

TheJetOU:
They look a bit too cluttered when they wrap into more than one line IMO.


Zarel:
`inline-block` doesn't add a margin?

PR #5718, opened 2019-08-19
Mafia: Implement forced selflynch, disabling hammer

whalemer:
The latter feature relies on the nature of arithmetic with NaN to propagate the NaN value, so hammer counts and the like should always remain at NaN and fail.


HoeenCoder:
Why aren't we just setting hammerCount to `false` or `null` or instead of NaN? Seems like that would make more sense.

HoeenCoder:
Does the fact that hammerCount can be NaN need to be set in typescript?

Zarel:
`NaN` is a `Number`.

I _wish_ TypeScript were sophisticated enough to have "non-NaN number" as a separate type, but no.

PR #5719, opened 2019-08-19
Bump max room FAQ length to 8KB

Kaiepi:
Bot Development needs a much longer max room FAQ length limit because of its bot list in /rfaq bots. Even when excluding the formatting in the length limit, it's simply too short.


Kaiepi:
I forgot I tried bumping it to 2500 characters in the past and got turned down. Would 1KB be more acceptable?

scheibo:
I feel @bumbadadabum would be pretty hypocritical to suggest 500 characters is enough for room faqs.

bumbadadabum:
Shit yeah I missed all of these tags horrible dev yeah

PR #5722, opened 2019-08-21
Add lottery plugin

TheJetOU:


server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,213 @@
+'use strict';
+
+const LOTTERY_FILE = 'config/chat-plugins/lottery.json';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+/** @type {{[roomid: string]: {name: string, markup: string, participants: {[ip: string]: string}}}} */
+let lotteries = {};
```
Kaiepi: This should be `Object.create(null)` or `new Map()`, not `{}`. `{}` isn't actually an empty object

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,213 @@
...
+	constructor(roomid) {
+		this.roomid = roomid;
+	}
+	/**
+	 * @param {string} name
+	 * @param {string} markup
+	 */
+	create(name, markup) {
+		if (!lotteries[this.roomid]) {
+			lotteries[this.roomid] = {name, markup, participants: {}};
```
Kaiepi: It doesn't make a whole lot of sense to both have a `Lottery` class and keep state it depends on outside of it. `lotteries` should be a map of roomids to `Lottery` instances and `Lottery` itself should be keeping track of `name`, `markup`, and `participants` instead.

Also, same deal with not using `{}` here for holding on to participants.

Kaiepi: I'd also make it so `lotteries` is only used from commands instead, since it's just the commands that really need to care about this state.

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+	}
+	getWinner() {
+		const winners = [];
+		const lotteryData = lotteries[this.roomid];
+		let participantsKeys = Object.keys(lotteryData.participants);
+		for (let i = 0; i < lotteryData.maxWinners; i++) {
+			const randomKey = participantsKeys[participantsKeys.length * Math.random() << 0];
+			const winner = lotteryData.participants[randomKey];
+			winners.push(winner);
+			participantsKeys = participantsKeys.filter(username => username !== winner);
```
whalemer: this looks like a more expensive ``splice``

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			if (!this.can('declare', null, room)) return;
+			if (!room) {
+				return this.errorReply('This can only be used in rooms.');
+			}
+			const lotteryData = lotteries[room.id];
+			if (!lotteryData) {
+				return this.errorReply('This room does not have a lottery running.');
+			}
+			const participants = Object.entries(lotteryData.participants).map(([ip, participant]) => {
+				return `${participant}${user.can('ban') ? ' (IP: ' + ip + ')' : ''}`;
```
whalemer: nit but I'd pull this check out of the loop

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+	}
+	destroy() {
+		delete lotteries[this.roomid];
+		this.write();
+	}
+	/**
+     * @param {User} user
+     */
+	addUser(user) {
+		if (!lotteries[this.roomid].participants[user.latestIp]) {
```
whalemer: doesn't this mean that if I have two IPs, I'll be able to join twice under the same name?

server/chat-plugins/lottery.js
whalemer: also isn't this comparison between the key (user's latest IP) and the value (user's name)?

Zarel:
Out of curiosity, was there a request for this or something? What is this intended to be used for?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+/** @type {{[roomid: string]: {maxWinners: number, name: string, markup: string, participants: {[ip: string]: string}}}} */
+let lotteries = {};
+try {
+	lotteries = require(`../../${LOTTERY_FILE}`);
+} catch (e) {
+	if (e.code !== 'MODULE_NOT_FOUND' && e.code !== 'ENOENT') throw e;
+}
+if (!lotteries || typeof lotteries !== 'object') lotteries = {};
```
panpawn: Also here, as @Kaiepi said:
>This should be Object.create(null) or new Map(), not {}. {} isn't actually an empty object

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+		create(target, room, user) {
+			if (!this.can('declare', null, room)) return;
+			if (!room) {
+				return this.errorReply("This can only be used in rooms.");
+			}
+			if (lotteries[room.id]) {
+				return this.errorReply(`There is already a lottery running.`);
+			}
+			const lottery = new Lottery(room.id);
+			let [maxWinners, name, ...markup] = target.split(',').map(val => val.trim());
```
panpawn: It doesn't appear that ``markup`` ever gets ran past ``this.canHTML(foo)``, unless I'm mistaken

server/chat-plugins/lottery.js
bumbadadabum: But wouldn't that just not work once you load from the file and don't make a new object?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+}
+
+/** @type {ChatCommands} */
+exports.commands = {
+	lottery: {
+		'!': true,
+		''(target, room) {
+			if (!room) {
+				return this.errorReply("This can only be used in rooms.");
+			}
```
bumbadadabum: Why is this both marked as ! and has a filter for whether you're using it in a room?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			const maxWinnersNum = parseInt(maxWinners);
+			if (isNaN(maxWinnersNum)) {
+				return this.errorReply(`${maxWinners} is not a valid number.`);
+			}
+			if (maxWinnersNum < 1) {
+				return this.errorReply('The maximum winners should be at least 1.');
+			}
+			lottery.create(maxWinnersNum, name, markup.join(','));
+			this.sendReply('The lottery was successfully created.');
+			this.add(Chat.html`|raw|<div class="broadcast-blue"><b>${user.name} created the "${name}" lottery!</b></div>`);
```
bumbadadabum: name needs to be html escaped and also needs a max length set.

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			this.modlog('LOTTERY CREATE');
```
bumbadadabum: Maybe specify WHICH lottery?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+		delete(target, room, user) {
+			if (!this.can('declare', null, room)) return;
+			if (!room) {
+				return this.errorReply('This can only be used in rooms.');
+			}
+			const lotteryData = lotteries[room.id];
+			if (!lotteryData) {
+				return this.errorReply('This room does not have a lottery running.');
+			}
+			const lottery = new Lottery(room.id);
```
bumbadadabum: I really fail to see why you're making a class for Lottery if you're not actually storing any information in it. You're keeping the state in a javascript object, and this class just provides a few methods without keeping any state. I would either keep track of a single Lottery object you use to store all the data in, or get rid of the OOP entirely.

server/chat-plugins/lottery.js
panpawn: Isn't ``Chat.html`` here escaping ``name`` already?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			}
+			const maxWinnersNum = parseInt(maxWinners);
+			if (isNaN(maxWinnersNum)) {
+				return this.errorReply(`${maxWinners} is not a valid number.`);
+			}
+			if (maxWinnersNum < 1) {
+				return this.errorReply('The maximum winners should be at least 1.');
+			}
+			lottery.create(maxWinnersNum, name, markup.join(','));
+			this.sendReply('The lottery was successfully created.');
```
panpawn: Considering this sends a declare-like message right after with the user's username, I wonder if this reply is necessary or not

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+	removeUser(user) {
+		if (lotteries[this.roomid].participants[user.latestIp]) {
+			delete lotteries[this.roomid].participants[user.latestIp];
+			this.write();
+			return true;
+		}
+		return false;
+	}
+	getWinner() {
+		const winners = [];
```
panpawn: I know that you _can_ push to arrays that are delcared as a ``const``, but I think I was told a while ago that we should still declare them using ``let`` because their value is subject to change

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			return this.parse(`/join view-lottery-${room.id}`);
+		},
+		'!create': true,
+		create(target, room, user) {
+			if (!this.can('declare', null, room)) return;
+			if (!room) {
+				return this.errorReply("This can only be used in rooms.");
+			}
+			if (lotteries[room.id]) {
+				return this.errorReply(`There is already a lottery running.`);
```
panpawn:
```suggestion
				return this.errorReply(`There is already a lottery running in this room.`);
```

server/chat-plugins/lottery.js
TheJetOU: Refactored to low-level functions that manipulate lotteries.

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,248 @@
...
+			const lotteryData = lotteries[room.id];
+			if (!lotteryData) {
+				return this.errorReply("This room doesn't have a lottery running.");
+			}
+			return this.parse(`/join view-lottery-${room.id}`);
+		},
+		'!create': true,
+		create(target, room, user) {
+			if (!this.can('declare', null, room)) return;
+			if (!room) {
```
panpawn: Maybe there should be a check if the room is a battle room?

server/chat-plugins/lottery.js
TheJetOU: ...I'm not actually sure why I was not just generating it from ``Object.values(lottery)``, easier. Also yeah, `splice` is probably just simpler here.

server/chat-plugins/lottery.js
TheJetOU: Fixed, but you can still just log in to another account, but nothing we could do about that, I think.

server/chat-plugins/lottery.js
TheJetOU: Yeah, most of these don't need to be global commands, fixed.

server/chat-plugins/lottery.js
TheJetOU: Refactored to low-level functions to manipulate the lottery.

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,255 @@
...
+}
+/**
+ * @param {string} roomid
+ * @param {User} user
+ */
+function removeUserFromLottery(roomid, user) {
+	const lottery = lotteries[roomid];
+	if (!lottery) return;
+	const participants = lottery.participants;
+	const userSignedup = participants[user.latestIp] || Object.values(participants).map(username => toID(username)).includes(user.userid);
```
whalemer: you're duplicating a lot of searching here

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,255 @@
...
+ */
+function getWinnersInLottery(roomid) {
+	const lottery = lotteries[roomid];
+	if (!lottery) return;
+	const winners = [];
+	const participants = Object.values(lottery.participants);
+	for (let i = 0; i < lottery.maxWinners; i++) {
+		const winner = participants[participants.length * Math.random() << 0];
+		winners.push(winner);
+		const winnerIdx = participants.indexOf(winner);
```
whalemer: why not just hold onto the index after you generate it?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,258 @@
...
+
+const LOTTERY_FILE = 'config/chat-plugins/lottery.json';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+/** @type {{[roomid: string]: {maxWinners: number, name: string, markup: string, participants: {[ip: string]: string}}}} */
+let lotteries = Object.create(null);
+try {
+	lotteries = require(`../../${LOTTERY_FILE}`);
```
whalemer: you need to modify the prototype of this to make sure it doesn't have `constructor` and the like

server/chat-plugins/lottery.js
TheJetOU: It is. I added a length limit of 50.

server/chat-plugins/lottery.js
TheJetOU: I feel like feedback other than the message the is sent to everybody is nice.

server/chat-plugins/lottery.js
TheJetOU: I feel like I've seen this style of coding before though. I also can't find it in ``CONTRIBUTING.md``, do you have a link?

server/chat-plugins/lottery.js
TheJetOU: The lottery loading was a bit verbose so I just re-wrote it.

I also don't really understand what you were saying, so can you check if this is still a problem in the new system?

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,249 @@
+'use strict';
+
+const LOTTERY_FILE = 'config/chat-plugins/lottery.json';
+
+/** @type {typeof import('../../lib/fs').FS} */
+const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
+
+const lotteriesContents = FS(LOTTERY_FILE).readIfExistsSync();
+/** @type {{[roomid: string]: {maxWinners: number, name: string, markup: string, participants: {[ip: string]: string}}}} */
+const lotteries = lotteriesContents ? JSON.parse(lotteriesContents) : Object.create(null);
```
whalemer: you still need to fix the prototype when you're loading from json

server/chat-plugins/lottery.js
TheJetOU: If the consensus is against me,  though I'll remove it, I don't really mind that much.

server/chat-plugins/lottery.js
```diff
@@ -0,0 +1,252 @@
...
+			if (!room) {
+				return this.errorReply("This can only be used in rooms.");
+			}
+			if (room.battle || !room.chatRoomData) {
+				return this.errorReply('This room does not support the creation of lotteries.');
+			}
+			if (lotteries[room.id]) {
+				return this.errorReply(`There is already a lottery running in this room.`);
+			}
+			let [maxWinners, name, ...markup] = target.split(',').map(val => val.trim());
```
whalemer: Won't this trim spaces off the html text? `Chat.splitFirst` is probably more what you're looking for here

server/chat-plugins/lottery.js
panpawn: Unfortunately, I do not. I think it was @Zarel who told me this, though.

panpawn:
Does this plugin handle garbage data at all? More specifically, what happens to a lottery's data if a room gets deleted while there was an ongoing lottery?  Or is this not really a concern?

server/chat-plugins/lottery.js
Zarel: I never put it in CONTRIBUTING, but I did _write_ in that style for a while.

I've given up on that style of coding, because it conflicts with `prefer-const`. And I've never actually inflicted it on other people in code review.

Zarel: If I told you, it's probably because someone asked what I was doing in my own code I wrote myself, and I was explaining that. I have no memory of telling anyone else to do it in that style.

panpawn: I found the discussion for reference, it was on https://github.com/Zarel/Pokemon-Showdown/commit/1a513ed585f755200d2777c2e45439672ce16223

TheJetOU:
> Does this plugin handle garbage data at all? More specifically, what happens to a lottery's data if a room gets deleted while there was an ongoing lottery? Or is this not really a concern?

I can probably just check if the rooms exist upon loading them and if they don't it would delete it.

server/chat-plugins/lottery.js
bumbadadabum: yeah whoops mb

bumbadadabum:
Still don't know what I really think of the functions that are only called once. I would probably just move all the logic to the commands itself, or track everything in a class. Right now your logic is kinda spread all over the file.

TheJetOU:
> Still don't know what I really think of the functions that are only called once. I would probably just move all the logic to the commands itself, or track everything in a class. Right now your logic is kinda spread all over the file.

I don't know, I kind of like the currently way it's written. The command itself validates the input and if the user has necessary perms; the functions manipulate the lottery object and write directly to the file.

I don't necessarily disagree with tracking them in a class. I just don't see what's wrong with only using a function once.

TheJetOU:
Using a `Map` wouldn't really work. `JSON.stringify` doesn't support maps. I guess there could be a function which would parse the map into JSON but it doesn't seem worth it.

PR #5723, opened 2019-08-23
TypeScript Tournaments

TheJetOU:


TheJetOU:
> Why not use the same export format as all the other files?

WDYM?

Zarel:
Most other files are imported as `import {Tournaments} from './tournaments';`

You're doing `import * as Tournaments from './tournaments';`

I'd prefer it all to work the same way so importing things is easier.

server/tournaments/index.ts
```diff
@@ -1080,7 +1080,7 @@ export class Tournament extends Rooms.RoomGame {
 		this.room.add(`|tournament|end|${JSON.stringify(update)}`);
 		this.isEnded = true;
 		if (this.autoDisqualifyTimer) clearTimeout(this.autoDisqualifyTimer);
-		delete exports.tournaments[this.room.id];
+		delete tournaments[this.room.id];
```
Zarel: I would actually use `Tournaments.tournaments` here. Makes hotpatching easier.

server/tournaments/generator-elimination.ts
```diff
@@ -66,21 +73,19 @@ class ElimNode {
 		}
 		this.children = children;
 	}
-	traverse(/** @type {(node: ElimNode) => void} */ callback) {
-		/** @type {ElimNode[]} */
-		const queue = [this];
+	traverse(callback: (node: ElimNode) => void) {
+		const queue: ElimNode[] = [this];
 		let node;
+		// tslint:disable-next-line: no-conditional-assignment
```
Zarel: Man, tslint doesn't have the double parens to get around conditional assignment? Rude.

server/tournaments/index.ts
```diff
@@ -1,4 +1,19 @@
...
+
+type TourCommand = (
+	this: CommandContext, tournament: Tournament, user: User, params: string[], cmd: string, connection: Connection
+) => void;
+type Generator = RoundRobin | Elimination;
+type GameRoom = import('../rooms').GameRoom;
+type ChatRoom = import('../rooms').ChatRoom;
+type CommandContext = import('../chat').CommandContext;
+type Connection = import('../users').Connection;
+type User = import('../users').User;
```
```suggestion
```

Everything from `GameRoom` down should also be deleted.

server/tournaments/index.ts
```diff
@@ -7,42 +22,38 @@ const MAX_REASON_LENGTH = 300;
...
+
 Punishments.roomPunishmentTypes.set('TOURBAN', 'banned from tournaments');
 
-/** @type {{[k: string]: Object}} */
-let TournamentGenerators = Object.create(null);
-/** @type {{roundrobin: string, elimination: string}} */
-const generatorFiles = {
-	'roundrobin': 'generator-round-robin',
-	'elimination': 'generator-elimination',
+const TournamentGenerators = {
```
```suggestion
const TournamentGenerators = {
	__proto__: null,
```

I vaguely remember a `constructor` crash if you don't do this.

server/tournaments/index.ts
```diff
@@ -7,42 +22,38 @@ const MAX_REASON_LENGTH = 300;
...
 }
 
-class TournamentPlayer extends Rooms.RoomGamePlayer {
-	/**
-	 * @param {User | string | null} user
-	 * @param {Tournament} game
-	 * @param {number} num
-	 */
-	constructor(user, game, num) {
+const tournaments: {[k: string]: Tournament} = {};
```
```suggestion
const tournaments: {[k: string]: Tournament} = Object.create(null);
```

server/tournaments/index.ts
```diff
@@ -1115,7 +1048,7 @@ class Tournament extends Rooms.RoomGame {
 		p2.games += 1;
 		if (!(p1.isDisqualified || p2.isDisqualified)) {
 			// If a player was disqualified, handle the results there
-			const error = this.generator.setMatchResult([p1, p2], result, score);
+			const error = this.generator.setMatchResult([p1, p2], result as 'win' | 'loss', score);
```
Zarel: Should this cast be necessary? Should `setMatchResult` take `'draw'`?

TheJetOU: If `setMatchResult` takes anything other than `win` and `loss` it returns `InvalidMatchResult`, so I don't think it should.

But if there's a winner (by virtue of there being a disqualified player) `result` definitely isn't `draw`, right?

Zarel:
Looks good.

PR #5724, opened 2019-08-24
Improve hidetext permissions

whalemer:
seemed to be a solution that at least a few people liked

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -2818,7 +2818,7 @@ const commands = {
 			}
 		}
 
-		if (targetUser && (cmd.includes('alt'))) {
+		if (targetUser && cmd.includes('alt')) {
 			room.send(`|c|~|${name}'s alts messages were cleared from ${room.title} by ${user.name}.`);
 			this.modlog('HIDEALTSTEXT', targetUser, null, {noip: 1});
 			room.hideText([
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -2806,17 +2806,15 @@ const commands = {
 		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${target} not found or has no roomlogs.`);
 		let userid = toID(this.inputUsername);
 
-		if (!user.can('mute', targetUser, room)) {
-			if (!this.can('mute', null, room)) return;
-			if (targetUser !== user) {
-				if (!cmd.includes('force')) {
-					return this.errorReply(`${target} is a higher rank than you, are you sure you want to hide their messages? Use /forcehidetext if you're sure.`);
-				}
-				// Notify staff when someone's hiding text outside of their jurisdiction
-				if (!room.isPrivate) {
-					const staffRoom = Rooms('staff');
-					if (staffRoom) staffRoom.addByUser(user, `<<${room.id}>> ${user.getIdentity(room)} hid ${targetUser ? targetUser.getIdentity(room) : target}'s messages`);
-				}
+		if (!this.can('mute', null, room)) return;
+		if (targetUser && targetUser.trusted && targetUser !== user) {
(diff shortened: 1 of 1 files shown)


server/chat-commands.js
```diff
@@ -2803,9 +2805,22 @@ const commands = {
...
-		if (!this.can('mute', null, room)) return;
 
-		if (targetUser && (cmd === 'hidealtstext' || cmd === 'hidetextalts' || cmd === 'hidealttext')) {
+		if (!user.can('mute', targetUser, room)) {
+			if (!this.can('mute', null, room)) return;
+			if (targetUser !== user) {
+				if (!cmd.includes('force')) {
+					return this.errorReply(`${target} is a higher rank than you, are you sure you want to hide their messages? Use /forcehidetext if you're sure.`);
+				}
+				// Notify staff when someone's abusing hidetext in a public room
```
scheibo: I think this should be rephrased (to at least include the word 'potentially') - simply using `forcehidetext` is not necessarily 'abuse', or why would we even be allowing it?

server/chat-commands.js
```diff
@@ -2803,9 +2805,22 @@ const commands = {
...
 		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${target} not found or has no roomlogs.`);
 		let userid = toID(this.inputUsername);
-		if (!this.can('mute', null, room)) return;
 
-		if (targetUser && (cmd === 'hidealtstext' || cmd === 'hidetextalts' || cmd === 'hidealttext')) {
+		if (!user.can('mute', targetUser, room)) {
+			if (!this.can('mute', null, room)) return;
+			if (targetUser !== user) {
+				if (!cmd.includes('force')) {
+					return this.errorReply(`${target} is a higher rank than you, are you sure you want to hide their messages? Use /forcehidetext if you're sure.`);
```
Zarel: Why???

Why not just any auth? I don't see why relative rank should matter.

whalemer: Mirroring how every other punishment on auth works, mostly.

I wouldn't be opposed to a monitor for if an auth was punished in any capacity, but that's definitely something I'd want to put through policy staff first.

Zarel: That's not how `/force`-anything works. It's not a permission check, it's a typo guard.

Zarel: `/lock` vs `/forcelock` works the same way.

whalemer: Would you want the monitor to be on any use on auth, then?

Zarel: It depends on how common people use it on auth. I would exempt using it on themselves, but otherwise yes. We can update it if it's too spammy.

server/chat-commands.js
```diff
@@ -2803,9 +2805,20 @@ const commands = {
...
+			}
+			// Notify staff when a trusted user gets their messages hidden
+			if (!room.isPrivate) {
+				const staffRoom = Rooms('staff');
+				if (staffRoom) staffRoom.addByUser(user, `<<${room.id}>> ${user.name} hid trusted user ${targetUser.name}'s messages.`);
+			}
+		}
 
-		if (targetUser && (cmd === 'hidealtstext' || cmd === 'hidetextalts' || cmd === 'hidealttext')) {
+		if (targetUser && (cmd.includes('alt'))) {
```
panpawn:
```suggestion
		if (targetUser && cmd.includes('alt')) {
```

panpawn: Also, what if you're trying to hide all of the alt's messages of a trusted user? Is this too niche of a situation to worry about?

whalemer: I would say it's too niche to care about

server/chat-commands.js
whalemer: Made a `trusted` check a la forcelock.

PR #5725, opened 2019-08-24
Deprecate Rooms(), Users() in favor of Rooms.get(), Users.get()

TheJetOU:


Zarel:
Sure whatever I give up...

Zarel:
I have trouble understanding why the build keeps running out of memory at the same spot. I wouldn't expect _not_ using this pattern to make performance worse.

xfix:
This change looks fine, and because it did change Scavengers technically, it triggered a review from me.

scheibo:
> and because it did change Scavengers technically, it triggered a review from me.

Which is sort of obnoxious. How strongly do you (and @HoeenCoder) feel about `CODEOWNERS`? I think any substantial changes to the areas you have declared yourself owners for would already cause another GH collaborator to 'Request Review' from you anyway, the automated approach is a little heavy handed IMO?

xfix:
@scheibo 

Hm, the issue I feel like is that this triggers for global refactorings, which are way more common than I would have expected them to be. That said, I do want to be informed about any changes to Scavengers and Trivia scripts, but at the same time, I don't expect them to be all that common, so I don't know.

whalemer:
Given how infrequently scavs and trivia (and mafia) are edited, I feel like it's simple enough for people to just tag you.

Zarel:
I mean, global refactors aren't _that_ common. I don't think it'd be necessary to turn off the CODEOWNERS thing. Aren't you already subscribed to the repo and get notified for all PRs anyway?

scheibo:
>  I don't think it'd be necessary to turn off the CODEOWNERS thing

It has been raised as an annoyance numerous times in `<dev>` which is why I brought it up.

> I do want to be informed about any real changes to Scavengers and Trivia scripts

And I really wouldn't feel comfortable reviewing such changes anyway without the context, which is why I'd request a review from you in the first place.

PR #5727, opened 2019-08-25
Make /events view visually match /events

TheMezStrikes:
Lack of consistency was bothering me

### DIFF
--- a/server/chat-plugins/room-events.js
+++ b/server/chat-plugins/room-events.js
@@ -9,6 +9,18 @@
 
 'use strict';
 
+let formatEvent = function(event) {
+	let timeRemaining = new Date(event.date).getTime() - new Date().getTime();
+	if (!timeRemaining) timeRemaining = "The time remaining for this event is not available";
+	if (timeRemaining < 0) timeRemaining = "This event will start soon";
+	if (event.started) timeRemaining = "This event has started";
+	let ret = `<tr title="${isNaN(timeRemaining) ? timeRemaining : `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`}">`;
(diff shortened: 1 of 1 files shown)


Zarel:
tbh anyone can review this, I just don't use events at all

TheJetOU:
You should probably also add the 'Time Renaming \<time\>' on-hover from `/events`.

Which would make the view code be essentially written twice, so we might just want to add a `visualizeEvent` function in the same file to abstract that logic.

bumbadadabum:
@TheMezStrikes Sorry for not getting to this for 3 weeks but I agree with the above comment, could you still fix that (merge the 2 into one function)?

server/chat-plugins/room-events.js
```diff
@@ -9,6 +9,18 @@
 
 'use strict';
 
+let formatEvent = function(event) {
+	let timeRemaining = new Date(event.date).getTime() - new Date().getTime();
```
Asheviere: Indent?

server/chat-plugins/room-events.js
```diff
@@ -9,6 +9,18 @@
...
+	if (!timeRemaining) timeRemaining = "The time remaining for this event is not available";
+	if (timeRemaining < 0) timeRemaining = "This event will start soon";
+	if (event.started) timeRemaining = "This event has started";
+	let ret = `<tr title="${isNaN(timeRemaining) ? timeRemaining : `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`}">`;
```
Asheviere: Nested template strings aren't very readable. If you're going to split up the string buffer into multiple lines away, construct the string the line before.

TheMezStrikes: Done

PR #5728, opened 2019-08-25
Typescript punishments

whalemer:


server/punishments.ts
```diff
@@ -380,22 +339,23 @@ const Punishments = new (class {
...
-		let rangebans = [];
+		const rangebans = [];
 		for (const row of data.split("\n")) {
 			const ip = row.split('#')[0].trim();
 			if (!ip) continue;
 			if (ip.includes('/')) {
 				rangebans.push(ip);
 			} else if (!Punishments.ips.has(ip)) {
-				Punishments.ips.set(ip, ['BAN', '#ipban', Infinity, '']);
+				Punishments.ips.set(ip, ['BAN', '#ipban' as ID, Infinity, '']);
```
TheJetOU: Why are we casting this to `ID`? It's not an ID.

If the `Punishment` array takes non-ID string, you should probably just change the type itself, instead of abusing type casting.

server/punishments.ts
```diff
@@ -1413,20 +1205,15 @@ const Punishments = new (class {
...
 		return banned;
 	}
 
-	/**
-	 * @param {User} user
-	 * @param {string} roomid
-	 * @return {boolean}
-	 */
-	checkNameInRoom(user, roomid) {
+	checkNameInRoom(user: User, roomid: string): boolean {
```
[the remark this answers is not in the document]
whalemer: No, typescript seems to give up entirely when it recurses.

server/punishments.ts
```diff
@@ -424,26 +380,20 @@ const Punishments = new (class {
...
-	punish(user, punishment) {
+	punish(user: User | ID, punishment: Punishment) {
 		if (typeof user === 'string') return Punishments.punishName(user, punishment);
 
-		/** @type {Set<string>} */
-		const userids = new Set();
-		/** @type {Set<string>} */
-		const ips = new Set();
-		const affected = user.getAltUsers(PUNISH_TRUSTED, true);
+		const userids: Set<string> = new Set();
```
TheJetOU: Same thing for set declarations.
```suggestion
		const userids = new Set<string>();
```

TheJetOU:
You'll also need to update `/hotpatch punishments`.

server/chat-commands.js
```diff
@@ -3268,8 +3268,8 @@ const commands = {
 				if (lock['punishments']) return this.errorReply(`Hot-patching punishments has been disabled by ${lock['punishments'].by} (${lock['punishments'].reason})`);
 				if (requiresForce(patch)) return this.errorReply(requiresForceMessage);
 
-				Chat.uncache('./server/punishments');
-				global.Punishments = require('./punishments');
+				Chat.uncache('./.server-dist/punishments');
+				global.Punishments = require('./punishments').Punishments;
```
TheJetOU:
```suggestion
				global.Punishments = require('../.server-dist/punishments').Punishments;
```

server/punishments.ts
whalemer: (it's being asserted as that for ease of use since all the relevant code checks for it, left a comment to that extent)

server/punishments.ts
```diff
@@ -604,17 +587,17 @@ export const Punishments = new (class {
...
 					ipSubMap.delete(key);
 					success = id;
 				}
 			}
 		}
 		const useridSubMap = Punishments.roomUserids.get(roomid);
 		if (useridSubMap) {
-			for (const [key, curPunishment] of useridSubMap) {
-				if (curPunishment[1] === id && curPunishment[0] === punishType) {
+			for (const [key, [curPunishmentType, curId]] of useridSubMap) {
```
whalemer: not sure if this is actually an improvement in terms of readability

server/punishments.ts
whalemer: We could also type Punishments#ips as `PunishmentMap<string, punishment with second key string>` and the userid punishments as `PunishmentMap<ID, ID>` (and the same for nested maps), but that turns the earlier code into ] `>><><<>` soup, which I'm not sure is an improvement overall.

server/punishments.ts
```diff
@@ -1151,12 +952,11 @@ const Punishments = new (class {
...
-	 * @param {string} searchId userid or IP
-	 * @return {[string, string, Punishment][]}
+	 * @param searchId userid or IP
 	 */
-	search(searchId) {
-		/** @type {[string, string, Punishment][]} */
-		let results = [];
+	search(searchId: string) {
+		/** [key, roomid, punishment][] */
+		const results: [string, string, PunishmentG<string | ID>][] = [];
```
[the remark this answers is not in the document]
whalemer: typed it as `'#ipban' | ID` where applicable

server/punishments.ts
```diff
@@ -31,47 +32,39 @@ const BLACKLIST_DURATION = 365 * 24 * 60 * 60 * 1000; // 1 year
...
- * @typedef {object} PunishmentEntry
- * @property {string[]} ips
- * @property {string[]} userids
- * @property {string} punishType
- * @property {number} expireTime
- * @property {string} reason
- * @property {any[]} rest
  */
+type PunishmentG<T> = [string, T, number, string];
+type Punishment = PunishmentG<ID>;
```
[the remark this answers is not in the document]
whalemer: `PunishmentG`eneric, but also because I couldn't think of a better name at the time.

I'm specifically typing Punishments#ips as `'#ipban' | ID` because it's the only one that can be a non ID. If every punishment was typed that way you'd need a lot more unnecessary checks for `'#ipban'`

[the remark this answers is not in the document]
whalemer: Renamed it, I still want `Punishment` to be based off the generic because `PunishmentMap` needs to be typed in terms of the generic.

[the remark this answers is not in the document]
whalemer: me too, but I can't think of any other solutions that don't involve a ton of asserts / unnecessary checks.

scheibo:
@peach - This breaks things, please see the staff room logs. You're calling `toID` in 4 additional places here, in particular line 1062 looks very problematic as `user.locked` should definitely not have `toID` called on it. (lines 307, 1058, 1381 are the other 3 places where you're now calling `toID` and could also be problematic).

PR #5729, opened 2019-08-25
Refactor Chat to a big class

whalemer:
Not sure if this is an improvement but at least we don't have a massive object at the bottom anymore.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -83,6 +83,7 @@ const BROADCAST_TOKEN = '!';
 const TRANSLATION_DIRECTORY = 'translations/';
 
 import { FS } from '../lib/fs';
+import { formatText, linkRegex, stripFormatting } from './chat-formatter';
 
 // @ts-ignore no typedef available
 import ProbeModule = require('probe-image-size');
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -1293,509 +1088,763 @@ export class CommandContext extends MessageContext {
...
- *
- * @param message - the message the user is trying to say
- * @param room - the room the user is trying to say it in
- * @param user - the user that sent the message
- * @param connection - the connection the user sent the message from
- */
-function parse(message: string, room: Room, user: User, connection: Connection) {
-	Chat.loadPlugins();
-	const context = new CommandContext({message, room, user, connection});
+export const Chat = new (class {
```
Zarel:
```suggestion
export const Chat = new class {
```

server/chat.ts
```diff
@@ -1293,509 +1088,763 @@ export class CommandContext extends MessageContext {
...
+
+	registerMonitor(id: string, entry: Monitor) {
+		if (!Chat.filterWords[id]) Chat.filterWords[id] = [];
+		Chat.monitors[id] = entry;
+	}
+
+	resolvePage(pageid: string, user: User, connection: Connection) {
+		return (new PageContext({pageid, user, connection})).resolve();
+	}
+});
```
Zarel:
```suggestion
};
```

server/chat.ts
```diff
@@ -1293,509 +1088,763 @@ export class CommandContext extends MessageContext {
...
- * @param room - the room the user is trying to say it in
- * @param user - the user that sent the message
- * @param connection - the connection the user sent the message from
- */
-function parse(message: string, room: Room, user: User, connection: Connection) {
-	Chat.loadPlugins();
-	const context = new CommandContext({message, room, user, connection});
+export const Chat = new class {
+	constructor() {
+		this.multiLinePattern = new PatternTester();
```
Zarel: I would rather define these in the declaration than in the constructor. We don't usually do it because of some minor overhead from sucrase, but for literal one-time-use `new class` classes, I'd prioritize readability.

server/chat.ts
```diff
@@ -1293,509 +1088,738 @@ export class CommandContext extends MessageContext {
...
+		for (const curFilter of Chat.statusfilters) {
+			status = curFilter(status, user);
+			if (!status) return '';
+		}
+		return status;
+	}
+	/*********************************************************
+	 * Translations
+	 *********************************************************/
+	// language id -> language name
```
Zarel:
```suggestion
	/** language id -> language name */
```

Documenting properties should be in doc comments so they can show up on hover.

server/chat.ts
```diff
@@ -1293,509 +1088,738 @@ export class CommandContext extends MessageContext {
...
+	languages = new Map<string, string>();
+	// language id -> (english string -> translated string)
```
Zarel:
```suggestion
	/** language id -> (english string -> translated string) */
```

Zarel:
I haven't said this yet, but yes, I consider this a huge improvement.

whalemer:
Is this something you want done to Rooms and Users too, then? (now that we don't need Rooms/Users to be callable)

Zarel:
Sure.

Zarel:
I think I have a preference for "all properties before all methods".

I haven't enforced that in the past, but let's do that for all future refactors going forward, including this one.

whalemer:
Chat is mostly in groups by functionality, though. IMO it's clearer that way than having everything at the top.

PR #5735, opened 2019-08-27
Fix missed Rooms()

TheJetOU:


bumbadadabum:
essing my aych

bumbadadabum:
essing my aych

PR #5739, opened 2019-08-30
Typescript lottery

whalemer:
This is mostly just a test for typescripting a chat plugin.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1409,9 +1409,9 @@ export const Chat = new class {
 
 		for (const file of files) {
 			let plugin;
-			if (file.substr(-3) === '.ts') {
+			if (file.endsWith('.ts')) {
 				plugin = require(`./chat-plugins/${file.slice(0, -3)}`);
-			} else if (file.substr(-3) === '.js') {
+			} else if (file.endsWith('.js')) {
 				plugin = require(`../server/chat-plugins/${file}`);
 			} else {
 				continue;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1408,9 +1408,14 @@ export const Chat = new class {
 		files.unshift('info.js');
 
 		for (const file of files) {
-			if (file.substr(-3) !== '.js') continue;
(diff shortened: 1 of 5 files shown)


server/global.d.ts
```diff
@@ -12,6 +12,8 @@ import * as StreamsType from './../lib/streams';
 import * as UsersType from './users';
 import * as ChatType from './chat'
 
+import * as LotteryType from './chat-plugins/lottery';
```
Zarel: This seems unnecessary?

server/chat.ts
```diff
@@ -1408,9 +1408,14 @@ export const Chat = new class {
 		files.unshift('info.js');
 
 		for (const file of files) {
-			if (file.substr(-3) !== '.js') continue;
-			const plugin = require(`../server/chat-plugins/${file}`);
-
+			let plugin;
+			if (file.substr(-3) === '.ts') {
```
Zarel:
```suggestion
			if (file.endsWith('.ts')) {
```

server/chat.ts
```diff
@@ -1408,9 +1408,14 @@ export const Chat = new class {
...
+				plugin = require(`./chat-plugins/${file.slice(0, -3)}`);
+			} else if (file.substr(-3) === '.js') {
```
Zarel:
```suggestion
			} else if (file.endsWith('.js')) {
```

Zarel:
Looks great!

server/global.d.ts
whalemer: As far as I can tell it's necessary for typescript to recognise it, since the file's never directly included elsewhere.

TheJetOU: I'm pretty sure that's because you did not include it in `tsconfig.json`.

Zarel: Yeah, I would manually include it in `tsconfig` like JetOU is suggesting.

whalemer: ...right, that works

PR #5743, opened 2019-08-31
Improve username punishments

whalemer:
Track how many times a name has been forcerenamed, note differently for offline warns and forcerenames


server/users.ts
```diff
@@ -1535,6 +1535,12 @@ export class User extends Chat.MessageContext {
 		this.userMessage = '';
 		this.updateIdentity();
 	}
+	getPseudorankString() {
```
bumbadadabum: I'm not sold on the name of this. Maybe `getAccountStatus()`?

PR #5744, opened 2019-09-01
Wonder Guard should take effect on turn 2 of Sky Drop

Volco:
https://github.com/Zarel/Pokemon-Showdown/projects/3#card-25870318

### DIFF
--- a/test/sim/abilities/wonderguard.js
+++ b/test/sim/abilities/wonderguard.js
@@ -46,9 +46,9 @@ describe('Wonder Guard', function () {
 
 	it('should allow first turn of Sky Drop but not the second', function () {
 		battle = common.createBattle();
-		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['sleeptalk']}]});
+		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['howl']}]});
 		battle.setPlayer('p2', {team: [{species: "Aerodactyl", ability: 'owntempo', moves: ['skydrop']}]});
-		battle.makeChoices('move sleeptalk', 'move skydrop');
-		assert.false.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move skydrop'));
+		battle.makeChoices('move howl', 'move skydrop');
+		assert.fullHP(battle.p1.active[0]);
 	});
 });
--- a/test/sim/abilities/wonderguard.js
+++ b/test/sim/abilities/wonderguard.js
@@ -43,4 +43,12 @@ describe('Wonder Guard', function () {
 		battle.setPlayer('p2', {team: [{species: "Reshiram", ability: 'turboblaze', moves: ['fusionflare']}]});
 		assert.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move fusionflare'));
 	});
+
+	it('should allow first turn of Sky Drop but not the second', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['sleeptalk']}]});
+		battle.setPlayer('p2', {team: [{species: "Aerodactyl", ability: 'owntempo', moves: ['skydrop']}]});
+		battle.makeChoices('move sleeptalk', 'move skydrop');
+		assert.false.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move skydrop'));
+	});
 });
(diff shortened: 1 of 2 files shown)


Zarel:
I wonder if this should work like Kee Berry, where we should consider Sky Drop a status move on its first turn?

Incidentally: Is Sky Drop Kee Berry bugged?

Volco:
Good question, i could test that

Volco:
after testing it does not seem to trigger 
image

scheibo:
Would you please consider adding new unit tests for this change and the others you are proposing? Unit testing mechanics issues should be quite straightforward and helps prevent regressions going forward.

Volco:
okay, i'll take a look at how those work and then update the prs

Volco:
I'm not quite sure if i did the test correctly so if someone could take a look at if and tell me if anything needs to be changed i'd appreciate it

test/sim/abilities/wonderguard.js
```diff
@@ -43,4 +43,12 @@ describe('Wonder Guard', function () {
 		battle.setPlayer('p2', {team: [{species: "Reshiram", ability: 'turboblaze', moves: ['fusionflare']}]});
 		assert.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move fusionflare'));
 	});
+
+	it('should allow first turn of Sky Drop but not the second', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['sleeptalk']}]});
+		battle.setPlayer('p2', {team: [{species: "Aerodactyl", ability: 'owntempo', moves: ['skydrop']}]});
+		battle.makeChoices('move sleeptalk', 'move skydrop');
+		assert.false.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move skydrop'));
```
Zarel: I prefer not to use `assert.hurts` in general. You can just check if HP equals max HP.

test/sim/abilities/wonderguard.js
```diff
@@ -43,4 +43,12 @@ describe('Wonder Guard', function () {
 		battle.setPlayer('p2', {team: [{species: "Reshiram", ability: 'turboblaze', moves: ['fusionflare']}]});
 		assert.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move fusionflare'));
 	});
+
+	it('should allow first turn of Sky Drop but not the second', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['sleeptalk']}]});
+		battle.setPlayer('p2', {team: [{species: "Aerodactyl", ability: 'owntempo', moves: ['skydrop']}]});
+		battle.makeChoices('move sleeptalk', 'move skydrop');
```
Zarel: You should probably have Smeargle use a move with a side effect (like Howl), so you can make sure it didn't take effect.

Zarel: This is, after all, the most important part of this test: You need to write a test that fails before your test, and succeeds now.

This is one of the most important parts of writing tests: Fixing confirmation bias. It's not enough for a test to pass for things you want, it also needs to fail for things you don't want.

Volco: Okay, that's good to know for the future

test/sim/abilities/wonderguard.js
Volco: Okay, i'll do the changes tomorrow after classes, adding the side effect move (howl) and this change

Volco:
Okay, hopefully it's good now. if changes are needed in the test file lmk. Still not great with the test files, first time i've editted them

test/sim/abilities/wonderguard.js
```diff
@@ -43,4 +43,12 @@ describe('Wonder Guard', function () {
 		battle.setPlayer('p2', {team: [{species: "Reshiram", ability: 'turboblaze', moves: ['fusionflare']}]});
 		assert.hurts(battle.p1.active[0], () => battle.makeChoices('move sleeptalk', 'move fusionflare'));
 	});
+
+	it('should allow first turn of Sky Drop but not the second', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: "Smeargle", ability: 'wonderguard', moves: ['howl']}]});
+		battle.setPlayer('p2', {team: [{species: "Aerodactyl", ability: 'owntempo', moves: ['skydrop']}]});
+		battle.makeChoices('move howl', 'move skydrop');
+		assert.fullHP(battle.p1.active[0]);
```
Zarel: You're not testing for the side effects of the Howl. You need to test that the Howl didn't happen (that Smeargle's Atk isn't +1).

Zarel: You're also not testing that the second turn isn't allowed: this is only a test for the first turn.

AnnikaCodes:
This is marked as "Done" in the kanban board, so I assume it's been fixed.

PR #5745, opened 2019-09-01
Fix keeberry activiating due to present healing

Volco:
https://github.com/Zarel/Pokemon-Showdown/projects/3#card-25870237

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -12489,8 +12489,8 @@ let BattleMovedex = {
 			let rand = this.random(10);
 			if (rand < 2) {
 				move.heal = [1, 4];
-				// For the purposes of kee berry
-				move.category = "Status";
+				// Shouldn't activate Kee Berry
+				move.category = 'Status';
 			} else if (rand < 6) {
 				move.basePower = 40;
 			} else if (rand < 9) {
--- a/data/moves.js
+++ b/data/moves.js
@@ -12489,6 +12489,8 @@ let BattleMovedex = {
 			let rand = this.random(10);
 			if (rand < 2) {
 				move.heal = [1, 4];
+				// For the purposes of kee berry
+				move.category = "???";
 			} else if (rand < 6) {
 				move.basePower = 40;
 			} else if (rand < 9) {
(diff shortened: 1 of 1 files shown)


Volco:
Probably not the best possible fix for this but for now it should work fine for it's purpose

Zarel:
Genius. But yes I'm somewhat afraid of side effects. @Marty-D?

Marty-D:
I believe this would break interaction with King's Shield.

Zarel:
Should Kee Berry check for 0bp attacks, then?

Volco:
it would also break the baneful bunker interaction now that you mention king's shield

Volco:
wait actually no, present isn't contact

Marty-D:
It's not about being contact; King's Shield purposely doesn't block Status moves.

Volco:
oh, i forgot it did that

Volco:
okay, i tested it by forcing keeberry to always give hp for the test specifically. it is not an issue. the move still gets blocked by kings shield

Volco:
instead of checking for 0bp why not check to see if it dealt damage. there was something added a while back that would allow checking if a mon took damage that turn and by who iirc.

Zarel:
I'm pretty sure that this is wrong because other 0-damage moves would activate Kee Berry, such as False Swipe against a foe with 1 HP left.

Volco:
Oh, hmm. Well i did test the interaction of healing present (Status) on king's shield and it was blocked by king's shield. Likely due to ``onModifyMove()`` runs after ``onTryHit()``

Volco:
Once something is figured out what to do instead i will adjust the pr accordingly

Volco:
so how should this be handled?

Zarel:
We could keep track of this sort of thing in the `move.damage` property, which could treat `.damage === 0` separately from `.damage === null` or something.

Volco:
sorry still been busy, will pick this back up once i have time
same for my other 2 prs

Zarel:
Why was this merged? I thought Marty and I both pointed out this would introduce more bugs than it fixes.

Zarel:
Especially since I was maintaining this part of the code and could easily have fixed it properly. :[

TheImmortal:
It does not break King’s Shield as the user mentioned. So what bug are you referring to?

Zarel:
There are too many tiny interactions. I'd rather just hardcode Kee Berry.

PR #5746, opened 2019-09-01
Fix resting with full hp and a status effect message

Volco:
Resting with full hp + a status now displays "'s HP is full!" and not "But it failed!"

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -13516,10 +13516,8 @@ let BattleMovedex = {
 		flags: {snatch: 1, heal: 1},
 		onTryMove(pokemon) {
 			if (pokemon.hp < pokemon.maxhp && pokemon.status !== 'slp' && !pokemon.hasAbility('comatose')) return;
-			if (pokemon.hp === pokemon.maxhp && pokemon.status) {
-				this.add('-message', pokemon.name + "'s HP is full!");
-			} else {
-				this.add('-fail', pokemon);
+			if (pokemon.hp === pokemon.maxhp) {
+				this.add('-fail', pokemon, 'heal');
 			}
 			return null;
 		},
--- a/data/moves.js
+++ b/data/moves.js
@@ -13516,7 +13516,11 @@ let BattleMovedex = {
 		flags: {snatch: 1, heal: 1},
 		onTryMove(pokemon) {
 			if (pokemon.hp < pokemon.maxhp && pokemon.status !== 'slp' && !pokemon.hasAbility('comatose')) return;
-			this.add('-fail', pokemon);
+			if (pokemon.hp === pokemon.maxhp && pokemon.status) {
+				this.add('-message', pokemon.name + "'s HP is full!");
+			} else {
+				this.add('-fail', pokemon);
+			}
 			return null;
 		},
 		onHit(target, source, move) {


Zarel:
Messages are implemented client-side (so they can be translated).

data/moves.js
```diff
@@ -13516,7 +13516,11 @@ let BattleMovedex = {
 		flags: {snatch: 1, heal: 1},
 		onTryMove(pokemon) {
 			if (pokemon.hp < pokemon.maxhp && pokemon.status !== 'slp' && !pokemon.hasAbility('comatose')) return;
-			this.add('-fail', pokemon);
+			if (pokemon.hp === pokemon.maxhp && pokemon.status) {
+				this.add('-message', pokemon.name + "'s HP is full!");
```
Zarel: In other words, this should be:

```suggestion
				this.add('-fail', pokemon, 'move: Rest');
```

Zarel:
All this is assuming this is the correct error message. @Marty-D? @DaWoblefet?

data/moves.js
Volco: Oh there was a fail message for rest added to the client? Good to know

Volco:
I got it from the known bug mechanics project so it should be

scheibo:
https://github.com/Zarel/Pokemon-Showdown/projects/3#card-24620941

data/moves.js
Zarel: No, but you (or someone else) should add it to the client.

Zarel:
Sounds good. I wonder if there's some way to remind users to link Known Mechanics Bugs in relevant PRs.

data/moves.js
Volco: ah

Volco: i'll look at the client and see if i can figure out how to do that change. shouldn't be too difficult i'd imagine

scheibo:
You could edit it into the main post, but you could also just do a search in the mechanics bug project if you see a PR around mechanics and are unsure of the context...

Volco:
ill edit the prs i have with a link to the relevant mechanic bugs

data/moves.js
Zarel: You just need to add

```
fail: "[POKEMON]'s HP is full!",
```

to `data/text.js`.

Volco: oh, okay thanks. that makes it easier since i'd have to have found the correct file

Marty-D:
The HP is full message isn't exclusive to Rest failing, it's used for all recovery moves at full HP.

Volco:
currently all recovery moves used at full hp just say "but it failed!" don't they?

Zarel:
Oh, is it also displayed when Rest is used at full HP with no status?

Volco:
i can check on my moon save file real fast

Volco:
okay, i tested it on pokemon moon, recover does say it even when used at full HP with no status. meaning this becomes an entirely different issue

DaWoblefet:
Not only should all recovery moves at full HP display "<Pokemon>'s HP is full!", including Heal Pulse, Pollen Puff, and Floral Healing, but Rest gets its message at a different point than all the other recovery moves. See SadisticMystic's [list of checks done before move execution](https://www.smogon.com/forums/threads/ultra-sun-ultra-moon-battle-mechanics-research-read-post-2.3620030/page-10#post-8096570); Rest at full HP is checked at step 28, while the other healing moves are checked at step 40.

Volco:
The " [Pokemon]'s HP is Full!" message should definitely be added to to client regardless but not as a rest specific fail

Zarel:
Agreed.

Volco:
Tomorrow after classes i'll add the "fullHpFail" to the client and change rest to use it then make a new pr for the ones that also use it in-game, which i'll have to test before i make the changes to other moves

Zarel:
`fullHpFail` is a bad name. Look at how the other messages are named.

Zarel:
It should probably start with `fail`. `failFullHp` might work, but I think I'd rather put it in "misc effects" under `heal:fail`.

Volco:
Sorry, i said i'd do the pr updates yesterday. I might not be until the weekend at this rate since i've had too get up earlier than normal this entire week so i've been too tired to get the updates done once i got home

scheibo:
Don't worry about it - there isn't a timer on this! Take your time, thank you for your contributions. :)

Zarel:
Should we put the other heal failure messages in this PR, too? What situations do they occur in?

Volco:
I'm going to test that ingame when i have time so that may take a bit longer. I'll probably make a new pr for all of those

PR #5747, opened 2019-09-02
Allow staff room voices to ignore help tickets

AustinXII:
even though the functionality apparently doesn't work, this would still be useful in the future.

### DIFF
--- a/server/chat-plugins/helptickets.js
+++ b/server/chat-plugins/helptickets.js
@@ -1066,7 +1066,7 @@ let commands = {
 		unbanhelp: [`/helpticket unban [user] - Ticket unbans a user. Requires: % @ & ~`],
 
 		ignore(target, room, user) {
-			if (!this.can('lock')) return;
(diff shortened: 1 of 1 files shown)


Zarel:
...why are staff roomvoices notified of help tickets in the first place?

I'm also not sure a PR that doesn't do the thing it says it does is all that useful.

xfix:
Staff roomvoices shouldn't be notified about the tickets, unless the code is buggy.

TheJetOU:
It seems like they shouldn't be notified (https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-plugins/helptickets.js#L298 and https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-plugins/helptickets.js#L377)? 

Maybe something is wrong with the permission system that makes it think Staff room voices can mute?

AustinXII:
Not showing help tickets to voices seems like the best route

PR #5748, opened 2019-09-02
Improve Lottery

TheJetOU:
- Anyone can see the amount of participants
- Max winners can't go above `Number.MAX_SAFE_INTEGER`
- Max winners is added in modlog
- `/lottery edit <max winners>, <name>, <markup>` can be used to edit the markup
- `/lottery edit` shows the current configuration of the lottery
- "Page unavailable" is not shown to guest users, instead the not "allowing guests to join" logic is done in `/lottery join`.

### DIFF
--- a/server/chat-plugins/lottery.ts
+++ b/server/chat-plugins/lottery.ts
@@ -15,6 +15,9 @@ const lotteries: {
 } = lotteriesContents ? Object.assign(Object.create(null), JSON.parse(lotteriesContents)) : Object.create(null);
 
 function createLottery(roomid: string, maxWinners: number, name: string, markup: string) {
+	if (lotteries[roomid] && !lotteries[roomid].running) {
+		delete lotteries[roomid];
+	}
 	if (!lotteries[roomid]) {
 		lotteries[roomid] = {maxWinners, name, markup, participants: Object.create(null), winners: [], running: true};
 		writeLotteries();
@@ -140,6 +143,14 @@ export const commands: ChatCommands = {
 			}
 			if (!lottery.running) {
 				return this.errorReply(`The "${lottery.name}" lottery already ended.`);
+			}
+			for (const [ip, participant] of Object.entries(lottery.participants)) {
+				const userid = toID(participant);
+				const pUser = Users.get(userid);
+				if (Punishments.userids.get(userid)
+					|| Punishments.getRoomPunishments(pUser || userid, {publicOnly: true, checkIps: true}).length) {
+					delete lottery.participants[ip];
+				}
 			}
 			if (lottery.maxWinners >= Object.keys(lottery.participants).length) {
 				return this.errorReply('There have been not enough participants for you to be able to end this. If you wish to end it anyway use /lottery delete.');
--- a/server/chat-plugins/lottery.ts
+++ b/server/chat-plugins/lottery.ts
@@ -222,7 +222,6 @@ export const commands: ChatCommands = {
 			}
 		},
 		participants(target, room, user) {
-			if (!this.can('mute', null, room)) return;
 			const lottery = lotteries[room.id];
 			if (!lottery) {
 				return this.errorReply('This room does not have a lottery running.');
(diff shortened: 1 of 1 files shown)


scheibo:
Can you please add the ability to edit `maxWinners` as well? Probably even title. Like, instead of `editMarkup`, can a subsequent `create` when a lottery already exists just edit the current one, in the same way roomintro/staffintro etc works?  Maybe an `edit` alias for `create`?

Also, might be worth modlogging more information about the lottery creation, like the number of winners. We were unsure that the PS day one was actually set up correctly because it kind of went live with a test setup and I had to open the file on the server to inspect that the number of winners was set correctly.

server/chat-plugins/lottery.ts
```diff
@@ -209,7 +231,12 @@ export const commands: ChatCommands = {
 			const participants = Object.entries(lottery.participants).map(([ip, participant]) => {
 				return `${participant}${canSeeIps ? ' (IP: ' + ip + ')' : ''}`;
 			});
-			const buf = `<b>List of participants (${participants.length}):</b><p>${participants.join(', ')}</p>`;
+			let buf = '';
+			if (user.can('declare', null, room)) {
+				buf += `<b>List of participants (${participants.length}):</b><p>${participants.join(', ')}</p>`;
+			} else {
+				buf += `${participants.length} participant(s) joined this lottery.`;
```
scheibo: Does just the count even need to be gated on staff? I see this as being like a poll where you can see the number of people who have voted in the poll?

TheJetOU: Sure, I was thinking of adding the information (max winners and participants numbers) in the page, but I couldn't find a sensible place to add it.

TheJetOU:
> Can you please add the ability to edit maxWinners as well? Probably even title. Like, instead of editMarkup, can a subsequent create when a lottery already exists just edit the current one, in the same way roomintro/staffintro etc works? Maybe an edit alias for create?

Hm. The problem with that is it would require re-entering the markup which is not as easily accessible as the description in room events or the like. I think a `/lottery edit <name|max winners|markup> <edited version>` would be better.

> Also, might be worth modlogging more information about the lottery creation, like the number of winners. We were unsure that the PS day one was actually set up correctly because it kind of went live with a test setup and I had to open the file on the server to inspect that the number of winners was set correctly.

Done, but maybe the 'better' solution is adding that information in the lottery page, though the modlog is still useful for future reference.

scheibo:
I think the answer is to then make the markup accessible for editing. Perhaps `/lottery edit` with no args just spits out the current configuration which can then be edited to something, to make it easy to then modify

TheJetOU:
@scheibo

server/chat-plugins/lottery.ts
```diff
@@ -89,14 +93,16 @@ export const commands: ChatCommands = {
...
 			if (!this.can('declare', null, room)) return;
 			if (room.battle || !room.chatRoomData) {
 				return this.errorReply('This room does not support the creation of lotteries.');
 			}
 			const lottery = lotteries[room.id];
-			if (lottery && lottery.running) {
-				return this.errorReply(`There is already a lottery running in this room.`);
+			const edited = lottery && lottery.running;
+			if (cmd === 'edit' && !target && lottery) {
+				return this.sendReplyBox(`<strong>Max winners: </strong>${lottery.maxWinners}<br /><strong>Name: </strong>${lottery.name}<br /><strong>Markup: </strong>${Chat.html`${lottery.markup}`}`);
```
scheibo: This is in the wrong format - ideally you should be able to copy and paste the exact contents that are returned from a `code` block to be able to reproduce the existing lottery like with other commands that allow for editing. cf https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-commands.js#L1287  and https://github.com/Zarel/Pokemon-Showdown/blob/master/server/chat-commands.js#L1341

scheibo:
LG modulo formatting

scheibo:
No, use code styling for the whole thing please.

PR #5749, opened 2019-09-03
Introduce set importer logic

### LINKED ISSUE #5732: Move set import logic from the calc to Pokemon-Showdown
In order to support easily accessing sets in the client teambuilder, the [newly committed `sets` package from the damage calc](https://github.com/Zarel/honko-damagecalc/tree/master/sets) will be moved to this repository.

1. `@pokemon-showdown/sets` package will be hosted in the main https://github.com/Pokemon-Showdown/Pokemon-Showdown repo. The code will be modified to drop the `@pokemon-showdown/calc` dep and will depend directly on `Dex` without a shim. Because the `@pokemon-showdown/sets` package will only be exporting the data (see 3 below), isolating the generation code into a subpackage is less important (though can be done at a later date once https://pkmn.cc/ps-core-design is complete).
2. The data produced will include both sets and weights (adds negligible size overhead but allows for https://calc.pokemonshowdown.com to depend on https://play.pokemonshowdown.com assets for better caching, also facilitates integrating usage statistics into the client). The set data will follow `PokemonSet` conventions instead of those used by the damage calc (ie. standard as opposed to abbreviated `StatName`)
3. `sets` package will actually publish the **data**, not the code for fetching the data (ie/ two 'build' folders, compiled TS which can then be run to fetch the script, and the actual data files which is going to be what developers importing  `@pokemon-showdown/sets` will be accessing). The npm package's will also include an `index.js` and `index.d.ts` for loading logic and typings, but Pokemon-Showdown-Client will likely just pull the raw JSON files out of the `dist` folder.
4. Because the `sets` package pushes data, sets are updated before `npm publish`. `node build full` on the client will simply copy the data out of `node_modules/`, so updating sets actually involves running the set importer locally, bumping the version, publishing, updating the client's package.json and running `node build full`
5. The new UI of the calc will depend on `@pokemon-showdown/sets` and Parcel will handle bundling and lazy loading for us. The current UI can simply copy the files like the client.

Originally the `@pokemon-showdown/sets` package will support slicing by generation only (`gen1.json` etc), but I would like to also support supporting by generation + format (`gen1lgpeou.json`) as well.

scheibo:
Fixes #5732


scheibo:
Merging based on

> [19:20] Zarel: I don't see anything obviously wrong

PR #5753, opened 2019-09-04
Fix modlog entries for autolocks in groupchats

whalemer:


server/punishments.ts
```diff
@@ -678,7 +678,8 @@ export const Punishments = new class {
 		}
 		Monitor.log(`[${source}] ${punishment}: ${message}`);
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
-		Rooms.global.modlog(`(${toID(room)}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+		const roomid = (room as Room).id ? (room as Room).id : room;
```
Zarel: Wait... `toID` doesn't do this?

Oh, roomids aren't IDs.

PR #5754, opened 2019-09-04
Improve subroom checks

whalemer:
Allow subrooms of public rooms to have subrooms, allow making a room with subroomgroupchats private

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1061,6 +1061,9 @@ const commands = {
 			if (!room.isPrivate) {
 				return this.errorReply(`This room is already public.`);
 			}
+			if (room.parent && room.parent.isPrivate) {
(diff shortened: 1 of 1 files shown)


server/chat-commands.js
```diff
@@ -1080,7 +1080,15 @@ const commands = {
 			}
 		} else {
 			const settingName = (setting === true ? 'secret' : setting);
-			if (room.subRooms) return this.errorReply("Private rooms cannot have subrooms.");
+			if (room.subRooms) {
+				// As long as the overall parent is public, subrooms can be private
```
Zarel: Is this restriction necessary at all? Do we know why it was put into place?

bumbadadabum:
Talked this through on Discord with whales and upper staff

Zarel:
I still think this is way too complicated to try to enforce in code. :/

PR #5755, opened 2019-09-04
Destroy groupchats on all global punishments

whalemer:


server/rooms.ts
```diff
@@ -994,6 +994,25 @@ export class GlobalRoom extends BasicRoom {
 			if (staffRoom) staffRoom.add(crashMessage).update();
 		}
 	}
+	/**
+	 * Destroys personal rooms of a (punished) user
+	 * Returns a list of the user's remaining public auth
+	 */
+	destroyPersonalRooms(userid: ID) {
```
Zarel: This is going to need to be manually hotpatched or someone needs to `/nohotpatch`

PR #5759, opened 2019-09-05
Add a RoomID type

TheJetOU:
Next: Rename `Room#id` -> `Room#roomid` and `User#userid` -> `User#id`.

### DIFF
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -1,5 +1,3 @@
-'use strict';
-/** @typedef {GlobalRoom | GameRoom | ChatRoom} Room */
 /**
  * Punishments
  * Pokemon Showdown - http://pokemonshowdown.com/
@@ -14,7 +12,6 @@
  */
 
 import {FS} from '../lib/fs';
-import {BasicRoom} from './rooms';
 
 type Tournament = import('./tournaments').Tournament;
 
@@ -512,7 +509,8 @@ export const Punishments = new class {
 			rest,
 		}, roomid + ':' + id, ROOM_PUNISHMENT_FILE);
 
-		if (room instanceof BasicRoom && !(room.isPrivate === true || room.isPersonal || room.battle)) {
+		if (typeof room !== 'string' && !((room as BasicRoom).isPrivate === true || 
+			(room as BasicRoom).isPersonal || (room as BasicRoom).battle)) {
 			Punishments.monitorRoomPunishments(user);
 		}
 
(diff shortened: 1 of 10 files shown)


server/rooms.ts
```diff
@@ -480,7 +480,7 @@ export class GlobalRoom extends BasicRoom {
 				Monitor.warn(`ERROR: Room number ${i} has no data and could not be loaded.`);
 				continue;
 			}
-			const id = toID(chatRoomData.title);
+			const id = Rooms.toRoomID(chatRoomData.title);
```
Zarel: That is possibly a bug there.

"Wi-Fi" has the roomid `wifi`, not `wi-fi`.

Zarel: (Ditto everywhere else you use `toRoomID`.)

server/punishments.ts
```diff
@@ -491,8 +491,8 @@ export const Punishments = new class {
 		return success;
 	}
 
-	roomPunish(room: Room | string, user: User, punishment: Punishment) {
-		const roomid = typeof room === 'string' ? room : room.id;
+	roomPunish(room: Room | RoomID, user: User, punishment: Punishment) {
+		const roomid = room instanceof Rooms.BasicRoom ? room.id : room;
```
Zarel: Please don't; we intentionally try to avoid `instanceof` so we can duck-type where necessary.

Zarel: (That goes for this and every other `typeof` you converted to `instanceof`.)

server/users.ts
```diff
@@ -579,7 +579,7 @@ export class User extends Chat.MessageContext {
 	popup(message: string) {
 		this.send(`|popup|` + message.replace(/\n/g, '||'));
 	}
-	getIdentity(roomid: string = '') {
+	getIdentity(roomid: RoomID = '' as RoomID) {
```
```suggestion
	getIdentity(roomid = '' as RoomID) {
```

If you ever find yourself typing a type twice, you're probably underestimating TypeScript's ability to infer.

server/users.ts
```diff
@@ -602,7 +602,7 @@ export class User extends Chat.MessageContext {
 		}
 		return this.group + this.name;
 	}
-	getIdentityWithStatus(roomid: string = '') {
+	getIdentityWithStatus(roomid: RoomID = '' as RoomID) {
```
```suggestion
	getIdentityWithStatus(roomid = '' as RoomID) {
```

You'd think TSLint would catch this.

PR #5761, opened 2019-09-05
Helptickets: Show the user or room being reported

whalemer:


server/chat-plugins/helptickets.js
```diff
@@ -498,7 +498,11 @@ const pages = {
 			}
 
 			const isStaff = user.can('lock');
+			// room / user being reported
+			let meta = '';
 			if (!query.length) query = [''];
+			const targetTypeIndex = query.findIndex(value => value === 'user' || value === 'room');
```
Zarel:
```suggestion
			const targetTypeIndex = Math.max(query.indexOf('user'), query.indexOf('room'));
```

Really just a nitpick, but this is slightly more readable and better perf.

PR #5762, opened 2019-09-06
Update datacenters

Lusamine:


config/datacenters.csv
```diff
@@ -47,6 +47,10 @@
 5.199.173.240,5.199.173.247,OneProvider,http://oneprovider.com/
 5.199.174.104,5.199.174.111,OneProvider,http://oneprovider.com/
 5.252.68.0,5.252.69.255,Host1Plus UK,https://www.host1plus.com/
+5.253.204.0,5.253.204.255,M247,https://www.m247.com/
```
scheibo: These all appear consecutive - could it be combined into one line? ie, would something like

```
5.253.204.0,5.253.207.255,M247,https://www.m247.com/
```

work?

config/datacenters.csv
```diff
@@ -380,6 +386,10 @@
 45.62.222.0,45.62.255.255,KW Datacenter CA,http://www.kwdatacentre.com/
 45.63.0.0,45.63.127.255,Choopa,https://www.choopa.com/
 45.76.0.0,45.77.255.255,Choopa,https://www.choopa.com/
+45.114.116.0,45.114.116.255,DigitalFyre,https://www.digitalfyre.com/
```
scheibo: ditto

config/datacenters.csv
```diff
@@ -1199,8 +1212,7 @@
 84.40.0.0,84.40.63.255,Hostway Corporation,http://hostway.co.uk/
 84.53.128.0,84.53.191.255,Akamai,http://akamai.com/
 84.200.0.0,84.201.63.255,accelerated.de,http://accelerated.de/
-84.232.0.0,84.232.127.255,ServiHosting Networks S.L.,http://servihosting.es
-84.236.128.0,84.236.255.255,ServiHosting Networks S.L.,http://servihosting.es
+84.232.0.0,84.232.7.255,ServiHosting Networks S.L.,http://servihosting.es
```
Zarel: Servihosting contains a lot regional Spanish ISPs that serve residential customers. I suppose that's why you're unblocking them here?

Lusamine: 84.236.248.81 was reported as a false positive.  I went through much of the two blocks and found out most looked like legitimate ISPs.  I pared down the original ranges based on this
image

config/datacenters.csv
Lusamine: We discussed this in the chatfilters channel: m247 typically will have 4 blocks in a row, all in different countries.  Marty said that he usually looks up all 4 and puts them in separately since they can be registered to different providers and it makes it easier to rearrange later.  If we want to start combining blocks, there are lots of entries that have already been added this way.
image

scheibo: Thanks for the explanation!

config/datacenters.csv
Zarel: Nice, thanks!

PR #5764, opened 2019-09-08
Super Staff Bros Brawl Update

HoeenCoder:
This is a large update for SSB that has been in the work for a while. It replaces a large number of retired staff with new sets, updates a number of other sets, and brings the total up to 120.

Still TODO:
- [x] Remove debug code
- [x] Grammar checks on descriptions
- [x] Uniformity checks on descriptions.
- [x] ~~Look into renaming the format to "Super Staff Bros Project M" since people seem to like that. Unsure myself.~~ no
- [x] Code review (thats this, I need to do a full code review myself too).

Theres no set release date on this yet, so theres no huge rush on reviewing it. The goal is to have this out before the end of the month. When we go live, theres a number of things that I will need to do so please do not merge this until were ready.

PR #5765, opened 2019-09-09
Change weight units to hectograms

urkerab:
Mechanics implemented:

- Weight is calculated in positive integer multiples of hectograms, in particular for Light Metal
- Heavy Metal always applies before Float Stone
- Since Gen 6, weight is a property of the forme that is directly modified by Autotomize

My attempt to implement the mechanics described here:

https://www.smogon.com/forums/posts/8095060

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -1342,8 +1342,9 @@ let BattleAbilities = {
 	},
 	"heavymetal": {
 		shortDesc: "This Pokemon's weight is doubled.",
-		onModifyWeight(weight) {
-			return weight * 2;
+		onModifyWeightPriority: 1,
+		onModifyWeight(weighthg) {
+			return weighthg * 2;
 		},
 		id: "heavymetal",
 		name: "Heavy Metal",
@@ -1690,8 +1691,8 @@ let BattleAbilities = {
 	},
 	"lightmetal": {
 		shortDesc: "This Pokemon's weight is halved.",
-		onModifyWeight(weight) {
-			return weight / 2;
+		onModifyWeight(weighthg) {
+			return this.trunc(weighthg / 2);
 		},
 		id: "lightmetal",
 		name: "Light Metal",
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -550,8 +550,10 @@ export class Template extends BasicEffect implements Readonly<BasicEffect & Temp
 	readonly baseStats: StatsTable;
 	/** Max HP. Overrides usual HP calculations (for Shedinja). */
 	readonly maxHP?: number;
-	/** Weight (in kg). */
+	/** Weight (in kg). Not valid for OMs; use weighthg / 10 instead. */
 	readonly weightkg: number;
+	/** Weight (in integer multiples of 0.1kg). */
+	readonly weighthg: number;
 	/** Height (in m). */
 	readonly heightm: number;
 	/** Color. */
@@ -643,6 +645,7 @@ export class Template extends BasicEffect implements Readonly<BasicEffect & Temp
 		this.requiredItems = this.requiredItems || (this.requiredItem ? [this.requiredItem] : undefined);
 		this.baseStats = data.baseStats!;
 		this.weightkg = data.weightkg!;
+		this.weighthg = this.weightkg * 10;
 		this.heightm = data.heightm!;
 		this.color = data.color || '';
 		this.unreleasedHidden = !!data.unreleasedHidden;
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -192,6 +192,7 @@ export class Pokemon {
 	isStarted: boolean;
 	duringMove: boolean;
 
+	weighthg: number;
 	speed: number;
 	abilityOrder: number;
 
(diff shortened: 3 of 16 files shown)


Zarel:
I would still want the dex to store weight in kg...

Zarel:
Like, yes, the rest of the sim should track weight in hg (using variable name `weighthg` so it's clear what unit it is), but I think the pokedex files should keep it in kg for clarity.

data/abilities.js
```diff
@@ -1691,7 +1692,7 @@ let BattleAbilities = {
 	"lightmetal": {
 		shortDesc: "This Pokemon's weight is halved.",
 		onModifyWeight(weight) {
-			return weight / 2;
+			return weight >> 1;
```
Zarel: ಠ_ಠ

Zarel: I'm actually somewhat inclined to write this as `Math.trunc(weight / 2)` for readability.

urkerab:
- Changed data files to weightkg
- Changed everything else to weighthg
- Used `trunc` for integer halve

sim/dex-data.ts
```diff
@@ -550,8 +550,8 @@ export class Template extends BasicEffect implements Readonly<BasicEffect & Temp
 	readonly baseStats: StatsTable;
 	/** Max HP. Overrides usual HP calculations (for Shedinja). */
 	readonly maxHP?: number;
-	/** Weight (in kg). */
-	readonly weightkg: number;
+	/** Weight (in multiples of 0.1kg). */
+	readonly weighthg: number;
```
Zarel: I would also include `weightkg` here, for outside consumers.

HoeenCoder:
SSB changes seem fine, If this gets merged before the SSB update does I'll have to update that with trickster's changes. (Tag for self: #5764 )

urkerab:
Kept `weightkg` on `Template`, although I tweaked the comment on `weighthg` to make it clearer.

sim/pokemon.ts
```diff
@@ -966,7 +968,7 @@ export class Pokemon {
 		this.apparentType = rawTemplate.types.join('/');
 		this.addedType = template.addedType || '';
 		this.knownType = true;
-		if (this.battle.gen >= 7) this.removeVolatile('autotomize');
```
Zarel: um is this ok?

sim/pokemon.ts
```diff
@@ -192,6 +192,7 @@ export class Pokemon {
 	isStarted: boolean;
 	duringMove: boolean;
 
+	weighthg: number;
```
Zarel: Oh, I see, we're caching this value? Should this be cached, though?

urkerab: I believe this more accurately represents the way Autotomize affects weight in Gen 7; while in Gen 6 it was a weight modification effect which persisted across forme changes, in Gen 7 it applies as if it's property of the current forme, much like the type.

DaWoblefet: Weight also resets on forme changes in Generation 6. An Aggron that used Autotomize twice in pre-Mega can be picked up by Sky Drop, but Mega Evolving causes Mega Aggron to retain the Speed boosts, yet not be picked up by Sky Drop, because its weight reset.

Zarel: We should maybe add a test for that. This also should maybe not be in the same commit, or at least be documented in the commit summary if it's a lot of work to pull them apart.

Zarel: I know your PR description says "implement the mechanics described here:", but I'd rather you explicitly list which mechanics you implement.

DaWoblefet: Autotomize also stacks in Generation 5. Weight does not reset on forme changes in Generation 5.

> Generation 5: a Shaymin-Sky that uses Autotomize, then is frozen to revert to Land Forme Shaymin, gets hit with a Heavy Slam from Tynamo.  It deals 65 damage to a level 100, 229 Attack move against the 213 Defense Shaymin. This implies Heavy Slam was 80 BP rather than 40 BP, notably implying that weight does not reset on forme changes in this generation (Shaymin is the only candidate anyway, which also can't learn Autotomize in any case naturally).

Autotomize also does stack in Generation 5. After 3 Autotomizes, a level 17 Timburr with 35 Attack dealt 16 damage with Low Kick to a 185 Defense Aggron, which implies Low Kick was 80 HP, which would not be possible if only one Autotomize were being applied.

urkerab: - Updated for SSB changes
- Moved Autotomize weight not resetting to Gen 5
- Some tests already exist in `test/sim/misc/weight.js`, do you need others?

Actual alterations to mechanics implemented:
- Weight is calculated in positive integer multiples of hectograms, in particular for Light Metal
- Heavy Metal always applies before Float Stone
- Since Gen 6, weight is a property of the forme that is directly modified by Autotomize

sim/pokemon.ts
scheibo: @urkerab - can you confirm this was intended?

Zarel: We had a conversation about it since then. Autotomize is no longer a volatile, it now just manually changes volatile weight.

Zarel:
I'd merge this without a test, but I do want tests if you can provide them.

urkerab:
(Rebased for SSB conflict so trivial Meld autoresolved it)

I've added a test which makes Aegislash attack after using Autotomize thus resetting its weight.

PR #5766, opened 2019-09-09
Make a couple improvements to /cmd

Kaiepi:
PoS-Bot uses `/cmd userdetails` when PM commands are used and it is not actually aware of the user existing because they're not in any of its rooms. It expects a userid, name, group, and status of some sort when creating user objects, but `/cmd userdetails` doesn't include a name at the moment. It'd be nice if it did so I don't have to pass both a username and userid to the routine for getting users from the bot's state machine.

I'm not explaining the second commit. You should already know why it's included and why it wouldn't be a good idea to do so.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -4501,6 +4501,11 @@ const commands = {
 			target = '';
 		}
 		if (cmd === 'userdetails') {
+			if (target.length > 18) {
+				connection.send('|queryrespone|userdetails|null');
+				return false;
+			}
+
 			let targetUser = Users.get(target);
 			if (!trustable || !targetUser) {
 				connection.send('|queryresponse|userdetails|' + JSON.stringify({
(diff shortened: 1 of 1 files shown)


server/chat-commands.js
```diff
@@ -4501,11 +4501,17 @@ const commands = {
...
+				connection.send('|queryrespone|userdetails|null');
+				return false;
+			}
+
 			let targetUser = Users.get(target);
 			if (!trustable || !targetUser) {
 				connection.send('|queryresponse|userdetails|' + JSON.stringify({
 					id: target,
 					userid: toID(target),
+					name: target,
```
Zarel: Shouldn't this be the user's real name?

Zarel: Honestly I'm also kind of not particularly happy with increasing PS's overall bandwidth use for the benefit of one bot...

Zarel: Oh, this isn't the user's real name because the user's offline.

PR #5767, opened 2019-09-09
Fix typo

Nolnis:
Fixes typo in endless battle warning


Zarel:
...in what sense is this a typo?

Zarel:
Ohhh, `turnsLeftText` already has "turns" in it.

Zarel:
Thanks!

PR #5768, opened 2019-09-10
Sockets: track when connections get made and log probable ghost connections

Kaiepi:


server/sockets.js
```diff
@@ -41,26 +41,28 @@ if (cluster.isMaster) {
 			// console.log('master received: ' + data);
 			switch (data.charAt(0)) {
 			case '*': {
-				// *socketid, ip, protocol
+				// *socketid, ip, protocol, connectedAt
```
Zarel: Is this really necessary? You can just use `Date.now` instead of sending `connectedAt` over IPC.

Kaiepi: The time would be less accurate if I did that because of the overhead of IPC and the work done in `sockets.js` before sending information about sockets to the main process.

server/users.ts
```diff
@@ -454,7 +457,10 @@ export class User extends Chat.MessageContext {
 	blockChallenges: boolean;
 	blockPMs: boolean | string;
 	ignoreTickets: boolean;
+	/** When the user last disconnected. */
 	lastConnected: number;
+	/** When the user last connected. */
+	lastConnectedAt: number;
```
Zarel: Why not call these `lastDisconnected`/`lastConnected`?

server/sockets.js
Zarel: Your threshold is seven days. A few milliseconds isn't going to affect it at all. I don't see that as worth complicating the IPC protocol for.

server/users.ts
Kaiepi: I thought about this, but didn't rename it because `lastConnected` is included in `|userdetails|` and there may be bots out there that depend on it being there and it being the time a user last disconnected.

Kaiepi:
FTR, I haven't forgotten about this, I just haven't had much spare time to work on it. I'll probably be able to get this ready to merge on the weekend.

server/users.ts
Zarel: You can change it here without changing it in `|userdetails|`.

Kaiepi:
> on the weekend

Or now.

server/sockets.js
Kaiepi: There are no results for bots when I search for `lastConnected` and `play.pokemonshowdown.com`, so I think it should be OK to change the property name without needing to do anything hacky to keep `lastDisconnected` named `lastConnected` in `|userdetails|`.

server/users.ts
Kaiepi: There are no results for bots when I search for `play.pokemonshowdown.com` and `lastConnected`, so I think it should be OK to just change the name of the property to `lastDisconnected` without needing to do anything hacky to keep it named `lastConnected` in `|userdetails|`.

server/sockets.js
Zarel: You marked this as resolved but you've neither changed this nor responded...

Kaiepi: It is changed, I just forgot to update the comment.

Zarel:
@scheibo if you want to give this a once-over, feel free, otherwise just merge

Issue #5684, opened 2019-08-07
Tournaments Improvements


scheibo:
A collection of tournament related features/bugs suggestion on the forums. The links to the original threads provide much more in-depth explanations as well as implementation hints.

## Server

> May also require some small client changes, not blocked on Preact

- [ ] Flag battles that went unplayed (state walkover instead of an arbitrary score).
- [ ] Command to export the brackets of the recently finished tournament, to allow for it to be posted on the forums.

_Originally posted by **Lutra** on [Smogon](https://www.smogon.com/forums/threads/tournament-suggestions.3650627)_

- [ ] Support `/tour latejoin <minutes/off>` to allow players to join tournaments after they've started (ie. subbing in for what would otherwise be byes)

_Originally posted by **Lutra** on [Smogon](https://www.smogon.com/forums/threads/tournament-late-join.3652559/post-8187479)_

- [ ]  Add a `'soft'` option to `/tour autodq <minutes/off>` which marks inactive players as ready to either disqualify or substitute out. 

_Originally posted by **Lutra** on [Smogon](https://www.smogon.com/forums/threads/tournament-late-join.3652559/post-8187479)_

- [ ] Add `/tour pause` and `/tour unpause`, with second-precision, that would pause/unpause autostart timers in signups, and pause/unpause dqtimers when the tournament has started.

_Originally posted by **Lutra** on [Smogon](https://www.smogon.com/forums/threads/tournament-changes.3652948/post-8197295)_

## Client

> None of these client changes should be blocked on Preact

- [ ] Don't show "tournament created" browser notifications if the room tournament has started/ended.

_Originally posted by @DaWoblefet on [Smogon](https://www.smogon.com/forums/threads/dont-show-tournament-created-browser-notifications-if-the-room-tournament-has-started-ended.3652427)_

- [ ] Dynamically resizing font in room tournament titles

_Originally posted by @DaWoblefet on [Smogon](https://www.smogon.com/forums/threads/dynamically-resizing-font-in-room-tournament-titles.3652587/)_

- [ ] Disable `/tour rules` display when tournaments are ignored

_Originally posted by @DaWoblefet on [Smogon](https://www.smogon.com/forums/threads/disable-tour-rules-display-when-tournaments-are-ignored.3651272/)_

scheibo:
@kanmingxu00

kanmingxu00:
@scheibo can you make me an assignee so no one takes this

Jakilutra:
[...]

PR #5769, opened 2019-09-10
Support exporting elim tournaments

kanmingxu00:
https://github.com/Zarel/Pokemon-Showdown/issues/5684

Export works for elimination tournaments only, after they're done.


server/chat-commands.js
```diff
@@ -609,6 +609,41 @@ const commands = {
 		`(in a PM) /invite [roomname] - Invites the player you're PMing to join the room [roomname].`,
 	],
 
+	tourexport: 'tournamentexport',
+	export: 'tournamentexport',
+	tournamentexport(target, room) {
```
scheibo: This should be a subcommand of the tournament command, not in `chat-commands`

kanmingxu00: ok, can you run the tournament command after a tournament has ended?

kanmingxu00: WIll come back to this, am so busy.

server/tournaments/index.ts
```diff
@@ -1701,7 +1748,7 @@ const chatCommands: ChatCommands = {
 			}
 		} else {
 			const tournament = (room.game && room.game.gameid === 'tournament') ? room.game as Tournament : null;
-			if (!tournament) {
+			if (!tournament && cmd !== 'export') {
```
whalemer: just put the command code above here, look at how ``create`` is implemented.

kanmingxu00: Ok, I wasn't sure abt that, the other actions seems more moderator-y. Thanks.

server/tournaments/index.ts
```diff
@@ -1211,6 +1226,34 @@ const commands: {basic: TourCommands, creation: TourCommands, moderation: TourCo
 			}
 			this.sendReply(`|html|<div class='infobox infobox-limited'>This tournament includes:<br />${tournament.getCustomRules()}</div>`);
 		},
+		export(tournament) {
+			if (this.room.lastTournament && this.room.lastTournament.type === 'tree') {
```
whalemer: You initialize `room#lastTournament` as `{}`, so check isn't right. I'd probably initialize it as `null`, so if it's present at all it'll have the correct data.

kanmingxu00: OK thanks.

server/rooms.ts
```diff
@@ -1419,6 +1445,7 @@ export class GameRoom extends BasicChatRoom {
 	active: boolean;
 	format: string;
 	auth: {[userid: string]: string};
+	lastTournament: LastTournamentData | null;
```
Zarel: `lastTournament` is a bad name for this; I'd suggest `lastTournamentData`.

server/tournaments/index.ts
```diff
@@ -1,5 +1,6 @@
 import {Elimination} from './generator-elimination';
 import {RoundRobin} from './generator-round-robin';
+import {LastTournamentDataNode} from '../rooms';
```
Zarel: This should be `type LastTournamentDataNode = import(...`

Zarel:
```suggestion
type LastTournamentDataNode = import('../rooms').LastTournamentDataNode;
```

server/rooms.ts
```diff
@@ -51,6 +51,29 @@ interface BattleRoomTable {
...
+	rootNode: LastTournamentDataNode;
+	playersLength: number;
+}
+
+export type LastTournamentDataNode = LastTournamentDataBattle | LastTournamentDataTeam;
+
+interface LastTournamentDataBattle {
+	children: LastTournamentDataNode[];
+	state: string;
+	team: string;
```
Zarel: `team` doesn't actually mean "team", does it? This should maybe be documented.

Also this should probably be typed in Tours itself, if we're going to type it.

Zarel: Also I'd just call it `TournamentData`; there's nothing "last" about the type.

kanmingxu00: It's not the data of the current Tournament though, it's the data of the Last one that was completed. I can change it to TournamentData though. 
Team is the term that was already used in the Object, I'm not changing anything here, just reflecting the data I'm given.

server/rooms.ts
kanmingxu00: Got it, made changes.

server/tournaments/index.ts
kanmingxu00: Got it, thanks!

Zarel:
Don't worry about the LGTM error, by the way; it looks like their bug, not yours.

server/rooms.ts
Zarel: No, it's the data of any tournament. You're mixing up the data type with the data itself.

Zarel: I know you can't rename `team`, but you can document that it's poorly named.

kanmingxu00: Oh I see what you're saying. There is only one use case for this data type though. But I can make the change. And yes, I'll comment

Zarel:
Is there a reason you can't run `npm test` locally?

server/rooms.ts
Zarel: No, there's another use: as the return type of `getBracketData`. That's also why I think this type should be defined in `tournaments/index`.

server/rooms.ts
```diff
@@ -51,6 +51,29 @@ interface BattleRoomTable {
...
+	rootNode: TournamentDataNode;
+	playersLength: number;
+}
+
+export type TournamentDataNode = TournamentDataBattle | TournamentDataTeam;
+
+interface TournamentDataBattle {
+	children: TournamentDataNode[];
+	state: string;
+	// "team" is name of winning player of battle
```
```suggestion
	/** actually the username of the winner */
```

kanmingxu00:
yea i've been being an idiot... i am running fulltest locally now. Before it was bugging because I hadn't npm installed but I thought it was something wrong with my VM.

server/rooms.ts
kanmingxu00: It's not quite the return type of getBracketData, I think I'm also including a PlayerCount in there for ease of calculation (I didn't want to have to traverse the tree to count the player numbers). I manually insert this playercount in onTournamentEnd()

Zarel:
You don't need to fulltest, just a regular `npm test` command is fine and will be a lot faster.

server/rooms.ts
Zarel: Maybe it should be. But the other types _are_ used in `getBracketData`, anyway.

Zarel: I would probably just drop `playerCount` from the type, and add it back with `& {playerCount: number}` in the variable definition.

kanmingxu00: Not sure I know what you mean exactly? sorry

Zarel: Um, I'd just do this for you but I'm really backlogged right now so it might take a dayish.

scheibo: see https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types

Remove the `playerCount` field from the type and instead declare the variable to be `LastTournamentDataBattle & {playerCount: number}`

kanmingxu00: thanks!

kanmingxu00:
@Zarel i moved it over to tournaments index and did the thing you suggested with the intersection types.

server/tournaments/index.ts
```diff
@@ -1142,6 +1170,17 @@ function createTournament(
 	return tour;
 }
 
+function moreRound(target: any) {
```
Zarel: Why `any`?

Zarel:
```suggestion
function moreRound(target: TournamentDataNode[]) {
```

server/tournaments/index.ts
```diff
@@ -1142,6 +1170,17 @@ function createTournament(
...
+	if (target && target.length) {
```
```suggestion
```

(you don't need either of these checks.)

server/tournaments/index.ts
```diff
@@ -1699,6 +1738,33 @@ const chatCommands: ChatCommands = {
...
 				}
 			}
+		} else if (cmd === 'export') {
+			if (room.lastTournamentData && room.lastTournamentData.type === 'tree') {
+				let tourString = '';
+				let tourData = [];
+				const tourWinner = room.lastTournamentData.rootNode.team;
+				let roundCounter = Math.ceil(Math.log(room.lastTournamentData.playersLength) / Math.log(2));
+				tourData.push(room.lastTournamentData.rootNode);
+				while (moreRound(tourData)) {
```
Zarel: Actually, `moreRound` in general is unnecessary here.

```suggestion
				while (tourData.some(node => node.children)) {
```

Zarel: You might need a `!!` cast, if TypeScript complains about `some`'s return type.

server/tournaments/index.ts
```diff
@@ -1078,6 +1100,12 @@ export class Tournament extends Rooms.RoomGame {
 			bracketData: this.getBracketData(),
 		};
 		this.room.add(`|tournament|end|${JSON.stringify(update)}`);
+		if (update.bracketData.type === 'tree') {
+			this.room.lastTournamentData = { ...update.bracketData };
+			if (this.room.lastTournamentData) {
```
Zarel: This will never be falsy...

server/tournaments/index.ts
```diff
@@ -1078,6 +1100,12 @@ export class Tournament extends Rooms.RoomGame {
 			bracketData: this.getBracketData(),
 		};
 		this.room.add(`|tournament|end|${JSON.stringify(update)}`);
+		if (update.bracketData.type === 'tree') {
+			this.room.lastTournamentData = { ...update.bracketData };
```
```suggestion
			this.room.lastTournamentData = {
				...update.bracketData,
				playersLength: this.players.length,
			};
```

Zarel:
Wow, in hindsight, LGTM was right; it was just looking at the actual context in which the function was used. I just didn't review that closely until now, but my fix also stopped LGTM's complaining.

server/tournaments/index.ts
```diff
@@ -30,6 +30,28 @@ function usersToNames(users: TournamentPlayer[]) {
 	return users.map(user => user.name);
 }
 
+export interface TournamentData {
```
Zarel: You haven't added this type to the return type of `getBracketData` yet.

kanmingxu00: done!

scheibo:
Context from a while back, posted here for huhst's benefit:

```
21:08:40 |c|&pre|i tried dropping a link to my server in the staff room and got no takers
21:08:46 |c|&pre|was faster to just log in all the account smyself
21:08:53 |c|&pre|forever alone
21:09:56 |c|&pre|i was trying to work on the feature which would let you export the final tournament state as BBCode after the fact
21:10:29 |c|&pre|which seems like it would be easy (just walk over the JSON and turn it into BBCode), only once the tour is over the tour object no longer exists
21:10:39 |c|&pre|so ``/tour export`` doesnt have context anymore
21:10:46 |c|&pre|am i missing something basic?
21:11:20 |c|&pre|or should i be like keeping around the last tour object until the next ``/tour create`` or something?
21:11:29 |c|@jumbowhales|i mean if you want to keep the data then youd need to
21:11:44 |c|@Slayer95|that's what i'd suggest we do yea
21:11:47 |c|@jumbowhales|or just attach the json to the room somewhere when it ends
21:12:00 |c|@jumbowhales|since keeping an ended room game around seems like kinda a bad idea
21:12:13 |c|@Slayer95|with a "gc" timeout
21:12:23 |c|&pre|how long do you think?
21:12:35 |c|&pre|i mean, obviously configurable as a constant
21:12:43 |c|@Slayer95|oh the bbcode could be posted to the chat logs
21:13:04 |c|&pre|to everyone?
21:13:11 |c|&pre|they already get the pretty UI
21:13:19 |c|@Slayer95|no, staff chat log
21:13:27 |c|@Slayer95|//roomlog
21:13:36 |c|@jumbowhales|as in write it to roomlog without sending?
21:13:42 |c|&pre|did not know that exists
21:13:43 |c|@Slayer95|^
21:13:45 |c|&pre|how do you see the roomlogs?
21:13:51 |c|&pre|oh
21:13:52 |c|@jumbowhales|that has the issue of not working in groupchats for non gstaff though
21:14:02 |c|&pre|like it will show up in the logs viewer?
21:14:07 |c|&pre|but not to users?
21:14:32 |c| unrankedwhales|im assuming that's what slayer was suggesting
21:14:46 |c| unrankedwhales|but you cant see groupchat logs without being staff, i think
21:14:49 |c|&pre|i mean, its kind of an attactive idea because then i dont need to hold around the object
21:15:08 |c|@jumbowhales|yeah, i think it's a fine place to keep it
21:15:15 |c|@jumbowhales|or even a separate tours log file
21:15:31 |c|@jumbowhales|and then have a way to search that file
21:16:01 |c|@Slayer95|yes
21:16:17 |c|@Slayer95|idk what the api for that is nowadays
21:16:28 |c|@Slayer95|used to be room.logEntry
21:16:41 |c|@jumbowhales|room.log.roomlog iirc
21:16:58 |c|@jumbowhales|adds to log without broadcasting
```

server/tournaments/index.ts
```diff
@@ -1699,6 +1727,33 @@ const chatCommands: ChatCommands = {
...
+							tourString = '<span' + (winner ? ' style="font-weight:bold;"' : '') + `>${copyChild.children[0].team}</span> vs <span` + (!winner ? ' style="font-weight:bold;"' : '') + `>${copyChild.children[1].team}</span><br>` + tourString;
+						}
+					}
+					tourString = `<br><strong>Round ${roundCounter}</strong><br>` + tourString;
+					roundCounter -= 1;
+				}
+				tourString = `<p> Tournament winner: <strong>${tourWinner}</strong></p>` + tourString;
+				return this.sendReplyBox(`<div style="max-height: 250px; overflow-y: auto"><p style="font-weight:bold;">${Chat.escapeHTML(this.room.title)} elimination tournament paste</p>` + tourString + '</div>');
+			} else {
+				return this.errorReply("You can only use the tournamentexport command after an elimination tour in this room has ended");
```
scheibo: tournamentexport -> tournament export?

scheibo: also, why? why not support exporting the current state of the tournament if someone ran it in the middle?

server/tournaments/index.ts
```diff
@@ -1699,6 +1727,33 @@ const chatCommands: ChatCommands = {
...
 			}
+		} else if (cmd === 'export') {
+			if (room.lastTournamentData && room.lastTournamentData.type === 'tree') {
+				let tourString = '';
+				let tourData = [];
+				const tourWinner = room.lastTournamentData.rootNode.team;
+				let roundCounter = Math.ceil(Math.log(room.lastTournamentData.playersLength) / Math.log(2));
+				tourData.push(room.lastTournamentData.rootNode);
+				while (tourData.some(node => node.children)) {
+					const tourDataCopy: TournamentDataNode[] = tourData;
```
scheibo: Naming here is kind of a mess, IMO. `Copy` is a smell. this also feels backwards = why not name the new variable below something other than `tourData` (maybe something to do with `round`, that is what this is iterating over, no?) and reassign `tourData` at the bottom of the loop as well? Also, `copyChild` seems like it should just be `child` followed by `grandChild` instead of `copyChild` and `child`?

Zarel:
I'm fine with sticking it in the roomlog, but it would be a bit less ergonomic.

server/tournaments/index.ts
```diff
@@ -1699,6 +1727,33 @@ const chatCommands: ChatCommands = {
 					}
 				}
 			}
+		} else if (cmd === 'export') {
+			if (room.lastTournamentData && room.lastTournamentData.type === 'tree') {
+				let tourString = '';
+				let tourData = [];
+				const tourWinner = room.lastTournamentData.rootNode.team;
+				let roundCounter = Math.ceil(Math.log(room.lastTournamentData.playersLength) / Math.log(2));
```
whalemer: (afaik) this doesn't work for multi-elim tours

server/tournaments/index.ts
kanmingxu00: Oh, the function used to be called tournamentexport, forgot to change.
Is it useful to show the current state of the tournament? I thought the paste would be for finished tours.

server/tournaments/index.ts
kanmingxu00: wait, do we support multielim tours?

server/tournaments/index.ts
scheibo: I dont know if its 'useful' per se, but in my mind the 'tour' command should be implicitly scoped to the current tour (which is how all the other subcommands work), and only if there is no current tour should it do something different. If we named this new command `lastexport` or `latestexport` or something it would make sense to always point to the past, but to me, the `tour` command should operate on the current tour by default.

scheibo:
> I'm fine with sticking it in the roomlog, but it would be a bit less ergonomic.

I don't actually have a strong opinion here, as I added after posting the log, "I also dont know how much we need to worry about keeping around the object". I'd probably stick with saving it on room for now and if we feel its an issue we can move it to logs.

server/tournaments/index.ts
kanmingxu00: that makes sense, and that's why i had it as a separate chatcommand or something before. 
If you only want me to act on the current tour, are you saying that I shouldn't be able to access the tour after the tour ends?
thanks.

scheibo: I'd have written this so that if there is a current tour, just use the current (unfinished bracket data) bracket data, otherwise use the last saved bracket data

server/tournaments/index.ts
Zarel: Yes.

whalemer:
@kanmingxu00 do you plan on finishing this?

AnnikaCodes:

PR #5770, opened 2019-09-11
Only validate mega evolutions in gen 6 or later

HoeenCoder:
This fixes an issue with pokemon having mega stones during validation in gen 1. This most notably allows mewtwo to be used in gen 1 OU.

Would appreciate feedback on this.

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -222,18 +222,20 @@ export class TeamValidator {
 		}
 		templateOverride = templateOverride || ruleTable.has('+basepokemon:' + toID(template.baseSpecies));
 		let postMegaTemplate = template;
-		if (item.megaEvolves === template.species) {
-			if (!item.megaStone) throw new Error(`Item ${item.name} has no base form for mega evolution`);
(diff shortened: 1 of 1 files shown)


Zarel:
This makes sense, but why is it necessary?

Zarel:
Mega Mewtwo X and Mega Mewtwo Y are both `Illegal` tier, so they shouldn't be allowed, either. The Mewtwonites are also nonstandard. There's way more than one bug here.

whalemer:
The mewtwonites don't get checked because gen 1 silently strips items without any legality checks.

Mega Mewtwo is in the `Illegal` tier, which doesn't get blocked by the rule table for some reason.

Zarel:
The latter should be fixed; the former can stay for now.

Zarel:
Shouldn't the `Pokemon` rule ban the Illegal tier? How is it banned in later gens?

whalemer:
Incidentally, why isn't the item stripping handled in `onChangeSet`? It seems like a better place for it.

PR #5774, opened 2019-09-13
Add punishments database schema

TheJetOU:


Zarel:
idk if I like splitting it apart? I'm not sure it makes sense to merge in piecemeal...

scheibo:
I do not see the previous PR landing given its size and scope. Committing in manageable chunks makes reviewing easier, reduces review latency and make it easier to detect bugs. In an ideal world, the chain of changes could be structured such that they provide functionality after each step, but that isn't always the case (how useful are the friends list tables and code that were committed but were never used?). One option here would be to find a more trivial example than punishments to convert to SQLite, but that would still likely be a fairly large PR.

database/migrate.js
```diff
@@ -0,0 +1,34 @@
...
+		if (err.code !== 'MODULE_NOT_FOUND') throw err;
+		return true;
+	}
+}
+
+function shell(cmd) {
+	child_process.execSync(cmd, {stdio: 'inherit', cwd: __dirname});
+}
+
+if (missing('sqlite')) {
```
scheibo: I would have just inlined `missing` and `shell` here because they're not used more than once, but I guess it reads better? It also makes me wonder if we should have a common tool-utils library somewhere given I obviously have copied and pasted these functions all over the place at this point?

database/migrate.js
```diff
@@ -0,0 +1,34 @@
...
+	shell('npm install sqlite');
+}
+
+const child_process = require('child_process');
+const fs = require('fs');
+const sqlite = require('sqlite');
+
+if (missing('./database/sqlite.db')) {
```
scheibo: Oh, you're actually using `missing` here... Its not a module though? Does this work? Is this to avoid using `fs.existsSync` or whatever? I'm not sure about this trick

database/migrate.js
```diff
@@ -0,0 +1,34 @@
...
+	console.log('No SQLite database found. Creating one...');
+	fs.writeFileSync('./database/sqlite.db', '');
+}
+
+if (fs.readdirSync('./database/migrations/').length) {
+	sqlite.open('./sqlite.db').then(function (database) {
```
scheibo: shouldnt this be `./database/sqlite.db`?

server/index.js
```diff
@@ -87,6 +88,14 @@ if (Config.watchconfig) {
 	});
 }
 
+/*********************************************************
+ * Run database migrations and conversions
+ *********************************************************/
+
+if (Object.values(Config.storage || {}).includes('sqlite')) {
+	child_process.execSync('node database/migrate');
```
scheibo: `../ ` and  `{stdio: 'inherit', cwd: __dirname}` (or `path.resolve(__dirname, '..')`), no?

database/migrate.js
TheJetOU: > It also makes me wonder if we should have a common tool-utils library somewhere

Possibly. That's out of scope for this PR though.

database/migrate.js
TheJetOU: (Using `missing` didn't actually work here.)

database/migrate.js
TheJetOU: Yes, `path.resolve(__dirname, './sqlite.db')` to be exact for it work when loading it in `server/index.js`.

TheJetOU:
(Tagging @Zarel to review the schema.)

database/migrate.js
```diff
@@ -0,0 +1,30 @@
...
+	shell('npm install sqlite');
+}
+
+const child_process = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const sqlite = require('sqlite');
+
+if (fs.readdirSync(path.resolve(__dirname, './migrations/')).length) {
+	sqlite.open(path.resolve(__dirname, './sqlite.db')).then(function (database) {
```
scheibo: I think we should be able to use `=>` inthis code - we only avoid it in specific top level scripts

TheJetOU: Oh yeah, I copied this script from my last PR where it was in the build script, so I was trying to make it pre ES6-compliant.

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,46 @@
+--------------------------------------------------------------------------------
+-- Up
+--------------------------------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS punishments (
+	punishType TEXT NOT NULL,
+	userid TEXT NOT NULL,
+	ips TEXT NOT NULL,
+	userids TEXT NOT NULL,
```
kotarou3: Are you using this field as an text "array" (e.g., multiple elements separated with a delimiter)? The plural gives that impression. Since this is not the way to do arrays in SQL databases.

You're meant to represent arrays by having each element be a separate row, with elements belonging to the same array sharing a key. Aka a [join table](https://en.wikipedia.org/wiki/Associative_entity). See [this](https://stackoverflow.com/questions/17371639/how-to-store-arrays-in-mysql) for a concrete example

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,46 @@
...
+	expireTime INTEGER NOT NULL,
+	reason TEXT NOT NULL,
+	PRIMARY KEY (punishType, id)
+);
+
+CREATE TABLE IF NOT EXISTS shared_ips (
+	ip TEXT NOT NULL,
+	-- type is always "SHARED", but we will store it anyway in case there are more types in the future
```
kotarou3: Just leave it out - if more types are added in the future, old rows can be migrated to add "SHARED" in

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,46 @@
...
+	ips TEXT NOT NULL,
+	userids TEXT NOT NULL,
+	expireTime INTEGER NOT NULL,
+	reason TEXT NOT NULL,
+	PRIMARY KEY (punishType, userid)
+);
+
+CREATE TABLE IF NOT EXISTS room_punishments (
+	punishType TEXT NOT NULL,
+	-- roomid:userid
```
kotarou3: Are you implying `id` is a concatenation of `roomid` and `userid`? Why not just have them as separate columns? If those columns are then indexed, this lets us look up by `roomid` and `userid` very quickly

kotarou3:
Additionally, considering these are punishments, they lack columns for accountability. E.g., the user who added the punishment, and when it was added.

database/migrations/001-punishments-schema.sql
TheJetOU: Hm, I'm not sure if it's worth having extra tables, adding data to two extra tables, etc.

Are there any benefits with going full relational database rather than just serializing / deserializing?

TheJetOU:
> Additionally, considering these are punishments, they lack columns for accountability. E.g., the user who added the punishment, and when it was added.

We don't currently store that data in the TSV database, so I'm not sure how much we care about them. We also don't need it to enforce punishments, if you need to know who the punisher is and when it happened, use modlog.

database/migrations/001-punishments-schema.sql
Zarel: Does SQLite just have TEXT and VARCHAR and all those just work exactly the same?

Zarel: I wonder if this should be a JSON column. Apparently SQLite supports those.

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,44 @@
...
+
+CREATE TABLE IF NOT EXISTS room_punishments (
+	punishType TEXT NOT NULL,
+	roomid TEXT NOT NULL,
+	userid TEXT NOT NULL,
+	ips TEXT NOT NULL,
+	userids TEXT NOT NULL,
+	expireTime INTEGER NOT NULL,
+	reason TEXT NOT NULL,
+	PRIMARY KEY (punishType, id)
```
Zarel: `id`?

database/migrations/001-punishments-schema.sql
```diff
@@ -0,0 +1,44 @@
...
+	user_id TEXT NOT NULL,
+	ips TEXT NOT NULL,
+	user_ids TEXT NOT NULL,
+	expire_time INTEGER NOT NULL,
+	reason TEXT NOT NULL,
+	PRIMARY KEY (punishType, userid)
+);
+
+CREATE TABLE IF NOT EXISTS room_punishments (
+	punish_type TEXT NOT NULL,
```
Zarel: Is the `_` delimiter standard? I hope this sort of thing is standardized somewhere.

TheJetOU: Isn't a `_` delimiter standard SQL naming?  I'm not sure what you mean by 'I hope this sort of thing is standardized somewhere' though, standardized where exactly?

Zarel: It might be? I don't know what standard SQL naming actually is.

Zarel: https://stackoverflow.com/questions/7662/database-table-and-column-naming-conventions

StackOverflow seems to want UpperCamelCase, which I don't think I've seen much of.

Zarel: @scheibo?

TheJetOU: Yeah, I've mostly seen `_`-delimited in SQL databases, so I just assumed that it was the standard.

TheJetOU: SQLite docs seems to follow that standard (`_`-delimited).

database/migrations/001-punishments-schema.sql
scheibo: > Does SQLite just have TEXT and VARCHAR and all those just work exactly the same?

https://www.sqlite.org/datatype3.html

database/migrations/001-punishments-schema.sql
scheibo: https://www.sqlite.org/foreignkeys.html doesnt seem to show `_`, it just does `songartist`?

scheibo: Google internal SQL styleguide favors `UpperCamelCase` for table names and `snake_case` for column names, FWIW.

TheJetOU: https://www.sqlite.org/cli.html

They use `snake_case` here, it seems like they can't even decide on one standard :p.

TheJetOU: I think either of these options are fine are:

\- `UpperCamelCase` for tables and `snake_case` for columns 

\- `snake_case` for both.

Zarel: SQLite's docs seems to be using snake_case for table names and nospaces for column names. This sucks.

The docs I've seen that recommend snake_case seem to be for databases that require quoting for identifiers that aren't all-lowercase, which is a good reason but not necessarily relevant to us. :[

Zarel: I'm leaning towards snake_case here. I'll trust JetOU that it's standard.

.gitignore
```diff
@@ -9,6 +9,7 @@ npm-debug.log
 .eslintcache
 package-lock.json
 /tools/set-import/sets
+database/sqlite.db
```
Zarel: I think we agreed on `databases`?

Also should we just ignore a `databases/` and force important files into Git?

database/migrations/001-punishments-schema.sql
scheibo: I'm fine with everything being snake case

.gitignore
scheibo: oh, I might have messed that up. I think it should be plural though because logs is plural as well. And yes, ignore databases/*.db* or ignore databases and !databases/migrations or whatever

TheJetOU: > I think we agreed on databases?

There was never any discussion on it as far as I'm aware. I don't particularly care about it.

> Also should we just ignore a databases/ and force important files into Git?

I think `database(s)/*.db` is a better suggestion than just ignoring everything.

Zarel: Sure. But there's nothing wrong with ignoring everything; it's what we do in `config` and `logs` after all.

.gitignore
```diff
@@ -9,7 +9,7 @@ npm-debug.log
 .eslintcache
 package-lock.json
 /tools/set-import/sets
-databases/sqlite.db
+databases/*.db
```
scheibo: This needs to be `databases/*.db*` as I said earlier. If you enable [WAL](https://www.sqlite.org/wal.html) (which PS does for certain databases with high write traffic) you get `.db.shm` and `.db.wal` files as well.

TheJetOU:
I think this should be good to go. @Zarel?

database/migrations/001-punishments-schema.sql
kotarou3: The benefits of going full relational is being able to query individual elements. E.g., Look up which punishments apply to a given IP

kotarou3:
Putting message here since Github isn't letting me unresolve comments:

>Hm, I'm not sure if it's worth having extra tables, adding data to two extra tables, etc.
>
>Are there any benefits with going full relational database rather than just serializing / deserializing?

The benefits of going full relational is being able to query individual elements. E.g., Look up which punishments apply to a given IP

Zarel:
I did see your response. I'd agree that's useful.

Zarel:
Although the database _is_ designed to just get loaded completely into RAM on start, so I'm not sure if that necessarily gives us anything over that.

TheJetOU:
We technically can query IPs with `SELECT * FROM punishments WHERE ips LIKE "%ip%"` because IPs _cannot_ be a substring of one another (false positive). On the other hand, userids _can_ be a substring of another userid.

My personal opinion is that as Zarel said: we load it all into memory anyway.

If we do stop loading into RAM, we can just migrate the tables to be full relational. Currently now, it will only take up extra memory and extra insertions.

Zarel:
`LIKE "%ip%"` isn't indexable, but yes the database is too small for that to matter.

You can add a delimiter like `:userid1:userid2:` and then search for `LIKE "%:userid1:%`. I do that trick a lot. But all this is probably overkill.

Zarel:
@scheibo Futureproofing or conservative changes? Schema changes _are_ annoying so futureproofing might be a legitimate consideration.

scheibo:
Are schema changes that much of a concern when we're talking about a 52K file? You can literally dump and reload the whole thing in seconds, and even boneheaded schema decisions will still execute quickly enough. I'd optimize for simplicity and convenience here - what makes the queries the most readable?

I don't want to diminish the importance of choosing the correct schema, but for the amount of data we're talking about in this instance it's probably not a good use of our time. The schemas for the `battles`, `ps_replays`, `ntbb_users` etc tables are infinitely more important, and on the sim server, a schema for chat/mod logs or a sim-side users table seem to be higher value to debate.

I'd also like to note that when choosing schemas, the answer is not always to implement things in the fully normalized way which is often taught in schools (cf. [denormalization](https://en.wikipedia.org/wiki/Denormalization)). There are tradeoffs with every schema design, and given my lack of context on punishments I had asked Jet to have Zarel or Ashe review it given they have more interest and understanding regarding the types of operations required and potential future use cases.

Zarel:
There's an argument that if the database _is_ designed to get loaded entirely into RAM on startup, what even is the point of moving to SQLite?

scheibo:
As I've been saying this whole time, I'm less interested in the specifics of what gets migrated to alternative storage backends and more interested in the infrastructure and patterns that we come up with to make it possible. I think there are valid use cases on the sim server which would benefit from being moved out of text files, and was fine with punishments being chosen to be the canary in the coal mine if that was what Jet felt needed converting.

TheJetOU:
> There's an argument that if the database _is_ designed to get loaded entirely into RAM on startup, what even is the point of moving to SQLite?

Moving to SQLite makes it more efficient since we're not writing the entire file onto disk every time a user is (un)punished.

It will also serve as a future reference for how other conversions architecture should be.

TheJetOU:
As scheibo said, I just picked punishments because it's low-risk (and mostly low-effort, because optimizing isn't too important as it's dumped onto RAM) to convert, where after it we could work  
on more important conversion like the users table on the sim server (persistent setting), modlog (the current setup is sluggish), etc.

Zarel:
That makes sense. I think this is fine, then.

TheJetOU:
@Zarel, is there anything else that needs to be done for this to be ready for merge?

monsanto:
@TheJetOU you will need to remove your changes to .gitignore & package.json

TheJetOU:
New years bump?

PR #5775, opened 2019-09-13
Clearstatus Modlog Fixes

Lusamine:
Fixed a missing space in clearstatus modlogging and added quotes around the bad status.


Zarel:
Thanks!

PR #5777, opened 2019-09-13
Add more hunt management options

sparkychildcharlie:
- Hints: add a command to display hints that can be recalled using /scav hint and /viewhunt for hunts that last several hours.  This allows previously given hints to still be visible even after it scrolls out of chat, or the user uses /clear after completing some difficult param question.  This command cannot be used until after 10 minutes of th
- /scav status is enhanced to show users who have tried to guess in the last 5 minutes in bold (&lt;strong&gt; for you HTML purists)

This has been approved by all 4 roomowners.

### DIFF
--- a/server/chat-plugins/scavengers.js
+++ b/server/chat-plugins/scavengers.js
@@ -449,7 +449,7 @@ class ScavengerHunt extends Rooms.RoomGame {
 
 		let current = player.getCurrentQuestion();
 
-		player.sendRoom(`|raw|You are on ${(current.number === this.questions.length ? "final " : "")}hint #${current.number}: ${Chat.formatText(current.question.hint)}${showHints && current.question.spoilers ? `<details><summary>Extra Hints:</summary>${current.question.spoilers.map(p => `- ${p}`).join('<br />')}</details>` : ''}`);
+		player.sendRoom(`|raw|You are on ${(current.number === this.questions.length ? "final " : "")}hint #${current.number}: ${Chat.formatText(current.question.hint)}${showHints && current.question.spoilers.length ? `<details><summary>Extra Hints:</summary>${current.question.spoilers.map(p => `- ${p}`).join('<br />')}</details>` : ''}`);
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


server/chat-plugins/scavengers.js
```diff
@@ -440,15 +441,15 @@ class ScavengerHunt extends Rooms.RoomGame {
...
+	onSendQuestion(user, showHints) {
 		if (!(user.userid in this.playerTable) || this.hosts.some(h => h.userid === user.userid)) return false;
 
 		let player = this.playerTable[user.userid];
 		if (player.completed) return false;
 
 		let current = player.getCurrentQuestion();
 
-		player.sendRoom(`|raw|You are on ${(current.number === this.questions.length ? "final " : "")}hint #${current.number}: ${Chat.formatText(current.question.hint)}`);
+		player.sendRoom(`|raw|You are on ${(current.number === this.questions.length ? "final " : "")}hint #${current.number}: ${Chat.formatText(current.question.hint)}${showHints && current.question.spoilers ? `<details><summary>Extra Hints:</summary>${current.question.spoilers.map(p => `- ${p}`).join('<br />')}</details>` : ''}`);
```
xfix: I think this should check whether the hunt has hints, otherwise you end up with confusing `<details>` that do nothing when clicked.

server/chat-plugins/scavengers.js
```diff
@@ -1130,6 +1132,25 @@ let commands = {
...
+		let elapsedTime = Date.now() - game.startTime;
+		if (elapsedTime < 600000 /* 10 minutes */) return this.errorReply("You can only use this command 10 minutes after the hunt starts.");
+
+		let [question, ...hint] = target.split(',');
+		question = parseInt(question) - 1;
+
+		if (!game.questions[question]) return this.errorReply(`Invalid question number.`);
+		game.questions[question].spoilers.push(hint.join(','));
+
+		room.addByUser(user, `Question #${question + 1} hint - spoiler: ${hint.join(',')}`);
```
xfix: It may be a good idea to disallow empty hints. A command like `/scav spoiler 1` will send an empty spoiler to the room, which isn't really useful.

sparkychildcharlie:
nyaa~ aaand fixed! :3

PR #5778, opened 2019-09-14
Show an access denied for modjoined battle rooms

### LINKED ISSUE #5676: Modjoined battles send |noinit|nonexistent|, client shows "battle expired"
- We probably don't need to lie about the existence of battle rooms.
- Causes the client to GET the corresponding replay data needlessly and fail with confusing message.

whalemer:
Fixes #5676

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -1315,14 +1315,14 @@ export class User extends Chat.MessageContext {
 			return Chat.resolvePage(roomid, this, connection);
 		}
 		if (!room || !room.checkModjoin(this)) {
-			if (room && room.battle) {
-				connection.sendTo(roomid, `|noinit|joinfailed|You do not have permission to join the battle "${roomid}"`);
-				return false;
-			}
 			if (!this.named) {
 				return Rooms.RETRY_AFTER_LOGIN;
 			} else {
-				connection.sendTo(roomid, `|noinit|nonexistent|The room "${roomid}" does not exist.`);
+				if (room && room.battle) {
+					connection.sendTo(roomid, `|noinit|joinfailed|You do not have permission to join the battle "${roomid}"`);
+				} else {
+					connection.sendTo(roomid, `|noinit|nonexistent|The room "${roomid}" does not exist.`);
+				}
 				return false;
 			}
 		}
--- a/server/users.ts
+++ b/server/users.ts
@@ -1315,6 +1315,10 @@ export class User extends Chat.MessageContext {
 			return Chat.resolvePage(roomid, this, connection);
 		}
 		if (!room || !room.checkModjoin(this)) {
+			if (room && room.battle) {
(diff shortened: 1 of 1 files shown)


server/users.ts
```diff
@@ -1318,7 +1318,11 @@ export class User extends Chat.MessageContext {
 			if (!this.named) {
 				return Rooms.RETRY_AFTER_LOGIN;
 			} else {
-				connection.sendTo(roomid, `|noinit|nonexistent|The room "${roomid}" does not exist.`);
+				if (room && room.battle) {
+					connection.sendTo(roomid, `|noinit|joinfailed|You do not have permission to join the battle "${roomid}"`);
```
Zarel: Wait, do we have a `joinfailed`?

Zarel: Apparently yes, we do.

I would probably phrase this as "This room is invite-only and you have not been invited."

Also I would probably apply this to all rooms, not just battles.

asgdf: Is there a reason to apply it to all private rooms instead of just battles?

Zarel: Is there a reason not to apply it to all private rooms?

Zarel: (My main reason for applying it to private rooms is "consistency" and "better UX".)

Lusamine: It's not a _huge_ deal to confirm the existence of private rooms through a failed join message now that modjoin exists to prevent disruption.  In general, we discourage people from mentioning private rooms in public; doing so innocently might get your text hidden, and doing so maliciously gets you locked.  So one might say that confirming the existence of a private room is somewhat contrary to how we handle private rooms.

Zarel: "Now that modjoin exists" is kind of confusing: The failed join message could only ever appear as a result of modjoin existing.

Without modjoin, people would _definitely_ be able to confirm the existence of a private room by test-joining it.

Zarel: This is the same issue as "Incorrect password or username" – there are often other ways to figure out if a username exists, so this makes the UX more confusing, for little practical benefit.

Someone knowing about the existence of a modjoined room but not being able to join is relatively harmless. We might as well make it easier for people who should be able to join it to understand that they're on the wrong account and didn't misspell it.

whalemer: sure

server/users.ts
```diff
@@ -1318,7 +1318,11 @@ export class User extends Chat.MessageContext {
 			if (!this.named) {
 				return Rooms.RETRY_AFTER_LOGIN;
 			} else {
-				connection.sendTo(roomid, `|noinit|nonexistent|The room "${roomid}" does not exist.`);
+				if (room) {
+					connection.sendTo(roomid, `|noinit|joinfailed|You must be invited to join room "${roomid}"`);
```
```suggestion
					connection.sendTo(roomid, `|noinit|joinfailed|The room "${roomid}" is invite-only, and you haven't been invited.`);
```

PR #5779, opened 2019-09-14
Linkify autoconfirmed alt display

asgdf:
This is a QoL change for easier access to the user page for cases where the active name isn't the autoconfirmed one.


Zarel:
Why not do it for trusted, etc, too?

asgdf:
Because that use case is much much rarer and I didn't bother before. But sure, I added it there as well now. Consistency!

PR #5781, opened 2019-09-14
Run tsc before mocha in `npm run fulltest`

TheJetOU:
It's kind of annoying to run 5 minutes of mocha test just for the test to fail in the TypeScript compiler which is at least 5x times faster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,7 @@
     "lint": "eslint --cache . && tslint --project .",
     "pretest": "npm run lint && npm run build",
     "test": "mocha",
-    "posttest": "npm run tsc",
-    "fulltest": "npm run pretest && mocha -g \".*\" && npm run posttest"
+    "fulltest": "npm run pretest && npm run tsc && mocha -g \".*\" "
   },
   "husky": {
     "hooks": {
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "lint": "eslint --cache . && tslint --project .",
     "pretest": "npm run lint && npm run build",
     "test": "mocha",
-    "fulltest": "npm run pretest && npm run tsc && mocha -g \".*\" "
+    "posttest": "npm run tsc",
+    "fulltest": "npm run pretest && npm run tsc && mocha -g \".*\""
   },
   "husky": {
     "hooks": {


scheibo:
Use `mocha run test` instead of `mocha run fulltest` during iterative development and then its like 10s tests vs. 20s `tsc` (on my box). And if you want to run `tsc`, then run `tsc`, not `mocha run fulltest`.

Zarel:
The difference in timing is so weird. I moved Mocha first because Mocha was taking ~10s and TypeScript was taking ~20s. But I guess a Mocha fulltest takes longer.

Zarel:
What are you doing using `npm run fulltest` anyway? You should be doing `npm test`. `npm run fulltest` is really only to be done once.

TheJetOU:
My theory was: run `npm run fulltest` -> mocha passes -> `tsc` throws -> fix it -> re-run to make sure everything passes. I could shave off 5 minutes by making `tsc` throw before mocha passes.

TheJetOU:
Oh wow, I wasn't aware of what `npm run test`, where it seems to take sub 90 seconds for the entire test suite. I was running 6 minute tests like a manic the whole fucking time.

I guess, yeah, there's no reason to use `npm run fulltest` most of the time, but I think there's only upsides to this change no matter how minimal.

Zarel:
I'm saying:

1. You should be using `npm test`, not `npm run fulltest`. The full test is just a sanity check; having Travis run it is plenty enough safety for my tastes.

2. You should be using VS Code with live-highlighted `tsc` errors.

TheJetOU:
> You should be using npm test, not npm run fulltest. The full test is just a sanity check; having Travis run it is plenty enough safety for my tastes.

(I'm assuming you haven't seen my previous comment.)

> You should be using VS Code with live-highlighted tsc errors.

VSCode doesn't report errors for _all_ files, only the ones you have currently open, or have opened recently.

Zarel:
Oh, yeah, also it's just `npm test`, the `run` part is optional.

Zarel:
As I said: The reason we currently do `mocha` before `tsc` is because on my laptop (and pre's, apparently), `mocha` takes 10 seconds and `tsc` takes 20 seconds.

package.json
```diff
@@ -25,8 +25,7 @@
     "lint": "eslint --cache . && tslint --project .",
     "pretest": "npm run lint && npm run build",
     "test": "mocha",
```
Zarel: Wait, wait, wait, `test` no longer runs `tsc`????

TheJetOU:
> As I said: The reason we currently do `mocha` before `tsc` is because on my laptop (and pre's, apparently), `mocha` takes 10 seconds and `tsc` takes 20 seconds.

Not for `fulltest` though..

package.json
whalemer: `npm test` runs the pretest script too

Zarel: Okay, it does, `npm test` is different from `npm run test`.

TheJetOU: You are correct, it does run the `pretest` script.

Zarel:
Your change affects `test`, though.

Zarel:
I suppose I have no opinion on this in its new state.

@scheibo?

PR #5784, opened 2019-09-15
Helptickets: Don't drop metadata when going back

whalemer:


Zarel:
Shouldn't the metadata be one of the things you get rid of when going back?

Zarel:
Like, at least when going back to the top?

whalemer:
when going back to the top, sure. Other situations I wouldn't say so, you still want to report the same person even if you can't decide if it's room or PM harassment.

PR #5787, opened 2019-09-16
Gen 3 Random Battle: Improve Movesets

TheTiksiBranch:


TheTiksiBranch:
(Reverted Shadow Ball removal because I forgot that "this Pokemon doesn't need STAB" club exists and that Gengar is in it. Guess I just got unlucky with my generated sets that always had SBall.)

PR #5788, opened 2019-09-16
Gen 3 Random Battle: Minor Move Generation Improvements

TheTiksiBranch:


TheTiksiBranch:
Being forced to setup before being able to attack _period_ stinks, and Encore is redundant when you can reliably (I chose not to include Hypnosis) sleep them and prevent them from moving at all.

Ninjask should (at least imo) be forced to have Baton Pass like how certain mons are forced to have recovery, but I'm not sure what way of doing that would be best.

PR #5789, opened 2019-09-17
Allow users to set Modchat for their PMs

ValkyrieMajorly:
On Smogon, XpRienzo wanted this to happen and plus, it was Approved so I would like to do this please. If I can't do it please explain why and I'll understand. P.S: uh, I got better at coding now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ Retired Staff
 
 Major Contributors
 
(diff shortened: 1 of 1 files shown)


Zarel:
Uh, this is an outdated credits update, not the feature you said it was.

PR #5791, opened 2019-09-17
Camomons: Show correct typings for Silvally and Arceus

WeWuzNidokangz:
* See discussion here: https://www.smogon.com/forums/threads/camomons.3598418/page-11#post-8213374
* Checked intended behavior with Camomons leader In The Hills.
* Used code is due to Urkerab.
* Tested on side-servers.


TheImmortal:
I don't think it'd be wrong for Camomons to affect Arceus/Silvally. Think of it this way, the Camomons mechanic is applied last. So you're not touching Arceus mechanics in the definition of the OM.

Zarel:
This code seems correct to me as an interim fix, regardless of what your further plans are.

PR #5793, opened 2019-09-18
Remove excess "item ="

tmagicturtle:


Zarel:
Yay, LGTM has its own bot instead of doing things through my account!

PR #5795, opened 2019-09-18
TypeScript: port server/sockets.js

Kaiepi:
This also splits server/sockets.js into two files: server/sockets.ts and
server/sockets-child.ts. This is not an important change for now, but
will be once socket handling gets refactored so it can actually be unit
tested.


Kaiepi:
Closing this in favour of https://github.com/smogon/pokemon-showdown/pull/6584.

PR #5800, opened 2019-09-19
Fix roomevents help messages

Dragonmirror27262:
Added bots to the Requires part of the help message
Added Requires to /roomevents start


Zarel:
We have never included bots in Require lists. Anyone running a bot should know that it has the same permissions as `@` unless otherwise specified.

PR #5801, opened 2019-09-20
Update datacenters and ip-tools

Lusamine:


config/datacenters.csv
```diff
@@ -223,7 +223,6 @@
 31.207.88.0,31.207.95.255,Hispaweb Network,http://hispaweb.net/
 31.210.64.0,31.210.127.255,Radore Hosting,http://as42926.net/
 31.210.232.0,31.210.239.255,crosspointcolo.co.uk,http://crosspointcolo.co.uk/
-31.214.128.0,31.214.255.255,Optimate Server,http://optimate-server.de/
```
Zarel: Why this removal? I haven't seen false positives for it...

Lusamine: asgdf reported 31.214.190.12 as a false positive in chatfilters channel and it does appear to belong to conectabalear.com now.  I'll go through the rest of the range and see if anything else looks like it should be added separately.
image

PR #5803, opened 2019-09-21
Allow locked players to set modchat in battle

scheibo:


Zarel:
There's an argument that the locked player would then be able to spam turning modchat on and off... but I suppose that's minor?

scheibo:
> [20:16:18] ‽m00la: i can already do that with modjoin

Which maybe is an argument for also turning off `/modjoin`...

But at the same time, I think the point is that people will find a way if they want. Like the namespamming people do as well.

My thought, if people are malicious enough to spam this, just gban them?

Zarel:
Fine by me, although I should point out that the modchat message is significantly more red and scary and spammy than the join/leave message (which is easy to mute).

PR #5805, opened 2019-09-22
Fix Gyro Ball/Electro Ball divide-by-0 edge cases

SadisticMystic:
The move documentation was accurate here, but the behavior was never actually implemented. So there's that.

### DIFF
--- a/data/mods/gen5/moves.js
+++ b/data/mods/gen5/moves.js
@@ -266,6 +266,23 @@ let BattleMovedex = {
 	},
 	electroball: {
 		inherit: true,
+		basePowerCallback(pokemon, target) {
+			let ratio = (pokemon.getStat('spe') / Math.max(1, target.getStat('spe')));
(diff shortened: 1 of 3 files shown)


data/mods/gen5/moves.js
```diff
@@ -266,6 +266,23 @@ let BattleMovedex = {
 	},
 	electroball: {
 		inherit: true,
+		basePowerCallback(pokemon, target) {
+			let ratio = (pokemon.getStat('spe') / Math.max(1, target.getStat('spe')));
+			this.debug([40, 60, 80, 120, 150][(Math.floor(ratio) > 4 ? 4 : Math.floor(ratio))] + ' bp');
```
scheibo: ???

why not just return this instead of using 12 lines of if blocks to do the same thing?

SadisticMystic: Because that's how the existing copy of that code in the current-gen section is structured.

scheibo: RIP. Can we please modify it in both places then? The `if` blocks are really ugh...

PR #5809, opened 2019-09-23
Tweak SSB text

Lusamine:
shortens a couple of descs that were cut off and adds a line to easter egg


data/mods/ssb/moves.js
```diff
@@ -412,7 +412,7 @@ let BattleMovedex = {
 		basePower: 0,
 		category: "Status",
 		desc: "For 5 turns, the terrain becomes Lava Terrain. During the effect, the power of Fire-type attacks made by Pokemon is multiplied by 1.5 and the power of Water-type attacks made by Pokemon is halved. Quilava's defense is doubled under Lava Terrain. Fire-type Pokemon have 1/16 of their maximum HP restored at the end of each turn, and all other Pokemon lose 1/16 of their maximum HP at the end of each turn.",
-		shortDesc: "5 turns. Damages non-Fire-types. Quilava: 2x Def.",
+		shortDesc: "Damages non-Fire-type. +Fire, -Water power.",
```
Torracat: There really is no way to fit all of that in that tiny description, is there...

Zarel: I would leave off "type": "Damages non-Fire"

PR #5811, opened 2019-09-23
Punishments: Abstract persistent storage

TheJetOU:
This locks persistent storage behind an interface so that other database adapters can be added in like SQLite.

Some notes: 
- `appendPunishment` has been split up to `appendPunishment` and `appendRoomPunishments` because the `filename` argument will get confusing with other database adapters


server/punishments.ts
```diff
@@ -108,87 +108,123 @@ class NestedPunishmentMap extends Map<RoomID, Map<string, Punishment>> {
...
-		});
+class PunishmentsStorage {
+	static connect(type: string) {
+		let storage;
+		if (type === 'memory') {
+			storage = PunishmentsMemoryStorage;
+		} else if (type === 'tsv') {
+			storage = PunishmentsTsvStorage;
+		} else {
+			Monitor.log(`Unrecongized punishments database type: ${type}, defaulting to memory storage.`);
```
TheJetOU: I wanted to crash here, but it'd put the server under lockdown because it happens in the first thirty minutes of the server starting up, which isn't ideal, so I compromised with logging to `stdout` and the Staff room.

TheJetOU:
Hm, I'm thinking of moving `Punishments.(room)?(userids|ips)` to `Punishments.storage`. I don't really like the current architecture where a class (`Punishments.storage`) manipulates state in another class (`Punishments`).

(By the way, someone please add the 'Do Not Merge' label as the design is far from final. I should remember to make my PRs draft...)

TheJetOU:
> I'm not a fan of calling PunishmentsMemoryStorage directly in PunishmentsTsvStorage (and subsequently in the implementation of other database adapters...) but I can't think of a better solution.

I think a 'better' solution would be to move `PunishmentsMemoryStorage` back to `Punishments` as it's used regardless of what you pick in `Config.storage.punishments` so it doesn't make sense to have it as a storage method, then call it directly rather than `Punishments.storage`.

TheJetOU:
Okay, I'm happy with the current setup.

(Tagging @bumbadadabum and @Zarel for a review. A poor man's assignation...)

scheibo:
> > I'm not a fan of calling PunishmentsMemoryStorage directly in PunishmentsTsvStorage (and subsequently in the implementation of other database adapters...) but I can't think of a better solution.
> 
> I think a 'better' solution would be to move `PunishmentsMemoryStorage` back to `Punishments` as it's used regardless of what you pick in `Config.storage.punishments` so it doesn't make sense to have it as a storage method, then call it directly rather than `Punishments.storage`.

The point is not the manipulate state in another class, but that the `PunishmentsStorage` *is* where the state is stored. Calling `PunishmentsMemoryStorage` in `PunishmentsTsvStorage` felt wrong because it *is* wrong, they are two different things, you just need to then go to a higher level and make a `CachingPunishmentsStorage<T extends PunishementStorage>` (or something) which delegates to a `PunishmentStorage` and but also writes to `PunishmentMemoryStorage`

TheJetOU:
@Asheviere and @Zarel: please review these changes.

TheJetOU:
New years bump??

AnnikaCodes:
This is really cool, but I don't think JetOU is planning to finish this...

PR #5815, opened 2019-09-25
Rework automatic replay uploading, upload replays on warn and autolocks

whalemer:


server/rooms.ts
```diff
@@ -1517,6 +1518,42 @@ export class GameRoom extends BasicChatRoom {
 		this.sendUser(connection, '|init|battle\n|title|' + this.title + '\n' + this.getLogForUser(user));
 		if (this.game && this.game.onConnect) this.game.onConnect(user, connection);
 	}
+	async saveReplay(forceUnlisted = false) {
+		const battle = this.battle;
+		if (!battle) return false;
```
Zarel: Throw?

server/rooms.ts
```diff
@@ -1517,6 +1518,42 @@ export class GameRoom extends BasicChatRoom {
...
+		// otherwise, don't show full details
+		let hideDetails = !format.id.includes('customgame');
+		if (format.team && battle.ended) hideDetails = false;
+
+		const data = this.getLog(hideDetails ? 0 : -1);
+		const datahash = crypto.createHash('md5').update(data.replace(/[^(\x20-\x7F)]+/g, '')).digest('hex');
+		let rating = 0;
+		if (battle.ended && this.rated) rating = this.rated;
+		const [success] = await LoginServer.request('prepreplay', {
+			id: this.id.substr(7),
```
```suggestion
			id: this.id.slice(7),
```

server/rooms.ts
whalemer: Why is battle nullable anyway?

Zarel: In GameRoom? Because some games aren't battles.

whalemer: which is why I didn't throw there, because it's not guaranteed to be a state that doesnt exist.

That being said, I think I prefer using room.battle as the check instead of just checking to see if it's a `GameRoom`

server/chat-commands.js
```diff
@@ -4156,39 +4156,9 @@ const commands = {
...
-			connection.popup(`This server's request IP ${success.errorip} is not a registered server.`);
-			return;
-		}
-		connection.send('|queryresponse|savereplay|' + JSON.stringify({
-			log: data,
-			id: room.id.substr(7),
-		}));
+		const result = await room.saveReplay(forPunishment);
+		if (result.error) return this.popupReply(result.error);
+		connection.send('|queryresponse|savereplay|' + JSON.stringify(result));
```
```suggestion
		connection.send(`|queryresponse|savereplay|` + JSON.stringify(result));
```

whalemer: Should we convert this all the way? `` `|queryresponse|savereplay|${JSON.stringify(result)` ``?

Zarel: No strong opinion. String standards just say `` ` `` for protocol code.

server/rooms.ts
```diff
@@ -1517,6 +1518,42 @@ export class GameRoom extends BasicChatRoom {
 		this.sendUser(connection, '|init|battle\n|title|' + this.title + '\n' + this.getLogForUser(user));
 		if (this.game && this.game.onConnect) this.game.onConnect(user, connection);
 	}
+	async saveReplay(forceUnlisted = false) {
```
Zarel: This is a bad name. It doesn't save the replay, only preps it.

server/punishments.ts
```diff
@@ -679,6 +679,8 @@ export const Punishments = new class {
 		const roomauth = Rooms.global.destroyPersonalRooms(userid);
 		// tslint:disable-next-line: max-line-length
 		if (roomauth.length) Monitor.log(`[CrisisMonitor] Autolocked ${name} has public roomauth (${roomauth.join(', ')}), and should probably be demoted.`);
+		// tslint:disable-next-line: no-floating-promises
+		if ((room as GameRoom).battle) (room as GameRoom).saveReplay(true);
```
Zarel: Your bad function naming also messes you up here. This doesn't save the replay, it only preps the replay.

`|queryresponse|savereplay|` is what saves the replay.

whalemer:
>|queryresponse|savereplay| is what saves the replay.

ugh, forgot about this

This refactor is pretty much impossible (or at least, won't improve anything) in it's current state.

PR #5816, opened 2019-09-26
Change Acid Rain so Steel/Poison retain immunities; Rewrite Corrosive Toxic

Lusamine:
[10:18 AM] A Cake Wearing A Hat: Okay yeah @ Programmer acid rain will no longer allow poison types to be poisoned, and the description should also not erroneously state that steels are not immune to poison under acid rain

Paradise's Corrosive Toxic was acting up with Distortion Blast because of the hacky way it was implemented, so with the help of @whalemer, it's been changed over to use setStatus.

### DIFF
--- a/data/mods/ssb/moves.js
+++ b/data/mods/ssb/moves.js
@@ -3341,7 +3341,7 @@ let BattleMovedex = {
 		accuracy: true,
 		basePower: 0,
 		category: "Status",
-		desc: "For 5 turns, the weather becomes Acid Rain.  Pokemon that are not Poison-type take damage every turn.  Special Defense of Poison-type pokemon is multiplied by 1.5.",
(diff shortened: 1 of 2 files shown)


data/mods/ssb/moves.js
```diff
@@ -3352,7 +3341,7 @@ let BattleMovedex = {
 		accuracy: true,
 		basePower: 0,
 		category: "Status",
-		desc: "For 5 turns, the weather becomes Acid Rain.  Pokemon that are not Poison-type take damage every turn.  Special Defense of Poison-type pokemon is multiplied by 1.5.  Poison moves ignore Steel immunity and Poison-type Pokemon can be poisoned.",
+		desc: "For 5 turns, the weather becomes Acid Rain.  Pokemon that are not Poison-type take damage every turn.  Special Defense of Poison-type pokemon is multiplied by 1.5.",
```
HoeenCoder: Extra spaces between sentences.

PR #5818, opened 2019-09-26
Fix multiple places in SSB leaking Illusion

Lusamine:
Lava Terrain leaks Illusion; this fix shows the Illusion mon's name if it's in effect.


TheImmortal:
Speaking of Lava Terrain, is it intentional that it doesn't override other Terrains @HoeenCoder ?

Torracat:
@TheImmortal it does override terrains. There is a visual bug that makes it look like terrains are stacking on top of each other

HoeenCoder:
Its a visual bug yeah, I'm going to do a client PR for that / patch setTerrain again soon(tm)

HoeenCoder:
Looks like scripted terrain has this issue too, can you patch that and check the rest just in case?

PR #5821, opened 2019-09-27
Fix BSS17 banlist

LegoFigure11:
English PGL was wrong again...


Zarel:
I already fixed this server-side, but apparently the push failed due to some corruption in `.git`.

Zarel:
What's this about the BSS17 banlist?

LegoFigure11:
We mistakenly had Wigglytuff allowed in BSS Special 17 when it should be banned, and Wishiwashi banned when it should be allowed (there were no official regulations uploaded to the PGL for us to go off, so there were a few mistakes in our original banlist). It's been quadruple-checked by the community now and after this change it will be 100% correct.

Zarel:
Was this done publicly? A link to a source would be nice.

Zarel:
Not necessarily the forum thread although that would also be nice, but more important would be a link to an official source where the banlist is.

LegoFigure11:
Most of the checking was done in the [Battle Spot discord](http://spo.ink/bsdiscord), and the results are in [this forum post](https://www.smogon.com/forums/threads/season-17-level-1.3653945/#post-8221288). It was also cross-referenced with [some members of of the japanese community on twitter](https://twitter.com/madari_poke/status/1166498977144901633?s=12), and is correct now.

LegoFigure11:
There's no quote-unquote "official" source, according to the PGL site _everything_ is legal, and the in-game hack checks will in fact let [anything through](https://twitter.com/UltraTerryy/status/1175175089882574848) (Lucario and Gliscor are not available in the games at Lv. 1). However as this isn't doable in-game without hacking, the banlist we've compiled here is the stuff that's legally obtainable at Lv. 1 in the games, which is the banlist that Nintendo presumably intended to use.

PR #5822, opened 2019-09-27
Use global modlog for queries in help tickets instead of failing

asgdf:
Based on a staff suggestion that turned out rather popular. Looking up modlogs in help tickets is somewhat of a common occurrence, so this'll save switching rooms or adding "global" manually.

### DIFF
--- a/server/chat-plugins/modlog.js
+++ b/server/chat-plugins/modlog.js
@@ -522,7 +522,15 @@ exports.commands = {
 		let targetRoom = Rooms.search(roomid);
 		// if a room alias was used, replace alias with actual id
 		if (targetRoom) roomid = targetRoom.id;
-		if (roomid.includes('-')) return this.errorReply(`Battles and groupchats (and other rooms with - in their ID) don't have individual modlogs.`);
+
+		if (roomid.includes('-')) {
+			if (user.can('modlog')) {
+				// default to global modlog for staff convenience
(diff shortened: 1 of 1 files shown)


whalemer:
Should this apply to battles too?

perhaps only automatically if you're gstaff, so as to not confuse people trying to moderate tours and the like.

asgdf:
Hmm I wanted to avoid making it look at different modlogs for different users.

Zarel:
For tour games specifically it might make sense to go up to the parent room?

asgdf:
I feel like that might end up tripping people up - both global staff being called to a tour game that expects the global modlog, as well as room staff that expect the parent room's modlog to actually contain entries from room tournament battles, which it doesn't.

Zarel:
Okay. It sounds it would be best just to use the global modlog at all times, and just display a permission error to non-global auth?

asgdf:
We could merge the two concepts - default to global, and if they don't have permission, explain how these rooms don't have individual modlogs and that's why they can't use it.
That way, roomstaff in battles and hosts in groupchats don't get confused that they can't see the modlog of a room that they have staff powers in.

Zarel:
Yes, that's what I was suggesting.

asgdf:
Does this look ok?

Zarel:
Looks good to me

PR #5824, opened 2019-09-27
Prevent overzealous spead of locks under some circumstances

asgdf:
This is for the situations "permalocked user logging in" and "lock on shared IP". In both cases, existing alts are also locked by name right now. This tends to be overkill.
Unfortunately this had the side effect of not locking existing alts at all (not even based on the locked IP). 

Do we want that behavior? Do we need to re-run checkIP for every alt so it locks them properly (just without punishing their usernames)?

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -2069,9 +2069,10 @@ const commands = {
 		let affected = [];
 
 		if (targetUser) {
-			affected = Punishments.lock(targetUser, duration, targetUser.locked, userReason);
+			const ignoreAlts = Punishments.sharedIps.has(targetUser.latestIP);
+			affected = Punishments.lock(targetUser, duration, targetUser.locked, ignoreAlts, userReason);
 		} else {
-			affected = Punishments.lock(null, duration, userid, userReason);
+			affected = Punishments.lock(null, duration, userid, false, userReason);
 		}
 
 		const globalReason = (target ? `: ${userReason} ${proof}` : '');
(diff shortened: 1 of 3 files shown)


bumbadadabum:
I think this is definitely a needed change. We could look at fine tuning what we want to do in these edge cases but I think this is a needed improvement for now.

Zarel:
Does this need any special hotpatching?

Zarel:
I forget if hotpatching chat automatically also hotpatches punishments

TheJetOU:
> I forget if hotpatching chat automatically also hotpatches punishments

It doesn't. I think you'd need to hotpatch chat, punishments, and monkey patch `User#handleRename`.

Zarel:
`handleRename`? Maybe we should wait for next restart?

TheJetOU:
> `handleRename`? Maybe we should wait for next restart?

Perhaps. Monkey patching `User` is can of worms in itself because if you change the prototype, the old user objects will still have the broken function.

Zarel:
> Perhaps. Monkey patching `User` is can of worms in itself because if you change the prototype, the old user objects will still have the broken function.

That's not how prototypes work.............

TheJetOU:
> That's not how prototypes work.............

Is it not? I'm not too sure how JavaScript prototypes work, but I thought I had that right, oh well.

Zarel:
image

asgdf:
(bump, just wanting to make sure nobody forgets to actually commit this before the next restart)

PR #5826, opened 2019-09-28
Rename Room#id -> Room#roomid, User#userid -> User#id

TheJetOU:
This rename makes the properties match their respective types.


TheJetOU:
(Whoever merges and restarts the server must also rename them in private chat plugins.)

sim/dex-data.ts
```diff
@@ -40,8 +40,8 @@ export class Tools {
 	static getId(text: any): ID {
 		if (text && text.id) {
 			text = text.id;
-		} else if (text && text.userid) {
```
whalemer: I'm not sure if it matters anywhere but we might want to keep the .userid check

TheJetOU: I guess it's still used for `muteQueue`, `BattleReady`, etc. Re-added.

Zarel:
Okay let's do this tomorrow.

PR #5827, opened 2019-09-29
Use void in promises to appease TSLint

TheJetOU:
We obviously don't care about floating promises if we're disabling them.

The server checks checks for `unhandledRejection`s anyway.

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -164,7 +164,6 @@ class FSPath {
 			return;
 		}
 
-		// tslint:disable-next-line:no-floating-promises
 		this.writeUpdateNow(dataFetcher, options);
 	}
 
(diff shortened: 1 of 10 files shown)


Zarel:
We had this discussion when it was first turned on.

@scheibo's idea was that floating promises should be explicit. It's kind of like `// falls through` in switch-cases: it's not about "we shouldn't do it", but "we shouldn't do it accidentally".

scheibo:
I gave up on my mission to civilize.

Zarel:
What's wrong with my `void` idea, anyway?

scheibo:
The point is to propagate the future up to the correct level and handle it there (or `// tslint:disable`) it and then probably half of these ignores would go away, but I gave up trying to tell people to fix the chain in `server/`.

scheibo:
If you want to use `void`, do what you want. I will continue to fix the chains in the code I need and disregard everything in `server/` anyway

PR #5831, opened 2019-09-30
Mafia: Fix IDEA data

ZestOfLife:
Since the "Greater Idea" etc bit gets ignored by the role praser (since the praser ignores parentheses), I have written those roles in their shortform without parentheses. Lmk if there's a better way to do it.


HoeenCoder:
Yeah this seems like the best way after further review.

PR #5836, opened 2019-10-02
SSB: Fix Nap Time with existing status

420Blazeitt:


data/mods/ssb/moves.js
```diff
@@ -2726,7 +2726,12 @@ let BattleMovedex = {
 		onHit(target, source, move) {
 			let napWeather = this.field.pseudoWeather['naptime'];
 			// Trigger sleep clause if not the original user
-			if (!target.trySetStatus('slp', napWeather.source, move)) return false;
+			if (!(target === napWeather.source)) {
+				for (const ally of target.side.pokemon) {
+					if (ally.status === 'slp') return false;
```
HoeenCoder: This isn't a proper sleep clause check, what if the only mon was put to sleep by its user? (rest, ect)

PR #5842, opened 2019-10-03
Add battle filter and evasion detection to chat monitor

bumbadadabum:
…tection

### DIFF
--- a/server/chat-plugins/chat-monitor.js
+++ b/server/chat-plugins/chat-monitor.js
@@ -40,7 +40,7 @@ const EVASION_DETECTION_SUBSTITUTIONS = {
 /** @type {{[k: string]: string}} */
 const EVASION_DETECTION_SUB_STRINGS = {};
 
-for (const letter in EVASION_DETECTION_SUBSTITUTIONS)  {
+for (const letter in EVASION_DETECTION_SUBSTITUTIONS) {
 	EVASION_DETECTION_SUB_STRINGS[letter] = `[${EVASION_DETECTION_SUBSTITUTIONS[letter].join('')}]`;
 }
 
@@ -179,6 +179,9 @@ Chat.registerMonitor('evasion', {
 			regex = constructEvasionRegex(word);
 			evasionFilterCache.set(word, regex);
 		}
+		// Remove spaces and obvious false positives
+		lcMessage = lcMessage.replace(/\bniger\b/g, '').replace(/\bnigeria/g, '');
+		lcMessage = lcMessage.replace(/[\s-_]/g, '');
 		const match = lcMessage.replace(/[\s-_]/g, '').match(regex);
 		if (match && (match[0] !== word || !regex.test(message))) {
 			if (isStaff) return `${message} __[would be locked for filter evading: ${match[0]} (${word})]__`;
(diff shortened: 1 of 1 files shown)


bumbadadabum:
well ok the message messed up whoops

bumbadadabum:
Two things that still need to be solved before I merge this:

- You can evade by adding spaces in the words, is there a good way we can mitigate that?
- Currently it locks you if you use a wordfiltered word because it doesn't match the original message. However, this is the only way I can discover to easily check if formatting (or other things stripped in lcMessage) was used to evade the filter.

server/chat-plugins/chat-monitor.js
```diff
@@ -6,9 +6,59 @@ const FS = require(/** @type {any} */('../../.lib-dist/fs')).FS;
...
+ * @param {string} str
+ */
+function constructEvasionRegex(str) {
+	// Handle duplicate letters
+	str = str.replace(/(\w)\1*/g, '$1+');
+
+	// substitutions
+	for (let letter of str) {
+		if (!(letter in EVASION_DETECTION_SUBSTITUTIONS)) continue;
+		str = str.replace(letter, `[${EVASION_DETECTION_SUBSTITUTIONS[letter].join('')}]`);
```
scheibo: We should pre-join these where theyre defined instead of joining the arrays on demand repetitively

scheibo: Or really pre-define the entire character class (ie/including `[` and `]`) so do no work here whatsoever pretty much

bumbadadabum: i suppose we could? I did this for readability (and because i didn't want to have to remake this entire fucking table) but I'm down to change it?

scheibo: You dont need to rewrite anyway, you just need to add a second `const` which takes the existing readable table and performs the transforms on it once and then just refer to that new variable instead

scheibo:
```ts
Object.fromEntries(Object.entries(obj, e => [e[0], `[${e[1].join('')}]`));
```

or something, though I dont know Object.fromEntries exists yet

bumbadadabum: only node.js 12 so no one liner :(

server/chat-plugins/chat-monitor.js
```diff
@@ -110,6 +169,36 @@ Chat.registerMonitor('warn', {
...
 });
 
+Chat.registerMonitor('evasion', {
+	location: 'EVERYWHERE',
+	punishment: 'EVASION',
+	label: 'Filter Evasion Detection',
+	monitor(line, room, user, message, lcMessage, isStaff) {
+		let [word, reason] = line;
+		if (typeof word !== 'string') throw new Error(`evader monitor filters should only have strings`);
+		let regex = evasionFilterCache.get(word);
```
whalemer: this may be really premature, but why don't we just generate every regex on load and store it with the line instead of requiring a separate map lookup every time?

bumbadadabum: Oh yeah we discussed this in DMs, iirc my answer was something along the lines of "i don't want to have to have cases for adding and changing filters, and complicate file loading"

PR #5845, opened 2019-10-04
Disallow regs from seeing their own IPs

whalemer:
Certain bad users are convincing people to merge their accounts as a way to find their IP, and most people don't have a real need to see their own IP through PS.

### DIFF
--- a/server/chat-plugins/info.js
+++ b/server/chat-plugins/info.js
@@ -158,7 +158,7 @@ const commands = {
 				buf += `<br />Semilocked: ${targetUser.semilocked}`;
 			}
 		}
-		if ((user.can('ip', targetUser) || user === targetUser)) {
+		if (user.can('ip', targetUser)) {
 			let ips = Object.keys(targetUser.ips);
 			ips = ips.map(ip => {
 				let status = [];
--- a/server/chat-plugins/info.js
+++ b/server/chat-plugins/info.js
@@ -182,6 +182,8 @@ const commands = {
 			if (user.group !== ' ' && targetUser.latestHost) {
 				buf += Chat.html`<br />Host: ${targetUser.latestHost} [${targetUser.latestHostType}]`;
 			}
+		} else {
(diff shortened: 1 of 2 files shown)


Zarel:
I'd at least show the connection's IP.

scheibo:
> I'd at least show the connection's IP.

Why?

kotarou3:
Surely the bigger problem here is letting the merging happen in the first place?

Lusamine:
> 
> 
> Surely the bigger problem here is letting the merging happen in the first place?

If you disallow merging, you can penalize people who are simply logging into their own account from different devices or networks.

Zarel:
> > I'd at least show the connection's IP.
> 
> Why?

It's easier than typing `whatismyip.com` into the URL bar.

> If you disallow merging, you can penalize people who are simply logging into their own account from different devices or networks.

In theory you could log out the previous device if you log into a new device. It would prevent PS from being used in multiple tabs/windows, though, and the pop-out window was something I always wanted to support.

> I recall discussion in the past about handling merging unregged alts, but that's easier said than done.

Merging unregged alts is currently not possible. I was thinking about making it possible, but it would be cookie-regulated, so it couldn't be used for trickery.

server/chat-plugins/info.js
```diff
@@ -182,6 +182,8 @@ const commands = {
 			if (user.group !== ' ' && targetUser.latestHost) {
 				buf += Chat.html`<br />Host: ${targetUser.latestHost} [${targetUser.latestHostType}]`;
 			}
+		} else {
```
Zarel:
```suggestion
		} else if (user === targetUser) {
```

scheibo:
> It's easier than typing `whatismyip.com` into the URL bar.

Sure, but why the heck do we care about supporting that anyway??? Why are we purposefully making a security compromise to cover a niche use case which has nothing to do with our core value?

Zarel:
I don't understand in what sense this is a security compromise.

PS has several IP-related features for which it makes sense to see your own IP, and having a built-in command makes it more convenient.

PR #5848, opened 2019-10-04
SSB: Protect blocks Soup Stealing 7 star strike

Torracat:


Zarel:
Why are these all `attrLastMove('[still]')` instead of `attrLastMove('[anim]')`?

HoeenCoder:
Thats how it was in Joim's SSB and I never bothered to look into if there was a better way since it was working.

PR #5850, opened 2019-10-05
Don't duplicate some mod actions in room modlogs

asgdf:
`globalModlog` writes to both the global modlog and the room one. Calling `modlog` in addition to this means stuff is logged twice.
I suppose `modlog` can have some slightly different behavior with the noip ad noalts options, but we have been logging the version with IP and alts in the roomlog for years as well now so this doesn't introduce any _new_ behavior. This is also only for stuff that's globally relevant enough to get put in the global modlog, so incuding alts/IPs should not be an issue.

### DIFF
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1866,7 +1866,6 @@ const commands = {
 			if (!targetUser || !globalWarn) return this.errorReply(`User '${this.targetUsername}' not found.`);
 
 			this.addModAction(`${targetUser.name} would be warned by ${user.name} but is offline.${(target ? ` (${target})` : ``)}`);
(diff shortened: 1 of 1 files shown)


Zarel:
Looks good to me. @bumbadadabum, how do you feel about including alts/IPs?

bumbadadabum:
Oh shit never saw this.

I personal think that while leaking IPs and alts to non staff is a serious issue, this is probably not the place to address the issue. Better to solve this in part of the modlog parsing code.

bumbadadabum:
So yeah seems fine

asgdf:
We already filter out all IPs for non-global staff in the modlog command. But we can't really filter out all alts there, because for some punishments, like roombans, roomstaff should be able to see them.

PR #5851, opened 2019-10-05
Split Dex out of Battle

Zarel:
In most other similar systems, like TeamValidator, we use `thing.dex` instead of having it extend `ModdedDex`. Battle has always extended `ModdedDex`, though. This changes Battle to match the others.

This should fix an issue with `Battle.data` not being cached.

This also frees up Battle to extend ObjectReadWriteStream<string> in a future update.


Zarel:
While writing the commit message, I thought a long time about why I now prefer `.dex` to extending `ModdedDex`, but I came up mostly blank. It definitely feels less "magical", but that's about it.

scheibo:
> While writing the commit message, I thought a long time about why I now prefer .dex to extending ModdedDex, but I came up mostly blank. It definitely feels less "magical", but that's about it.

Because perf. Perf is why.

Zarel:
> Because perf. Perf is why.

_Is_ it a perf issue? We can solve that one just by adding `this.data = Dex.forFormat(format).data` to the bottom of `Battle#constructor`, right?

urkerab:
(This broke the Anti-Vaxxers OM, which relied on being able to override `Battle.getImmunity()` in its mod...)

Zarel:
Why not override `runImmunity`?

urkerab:
There are a lot of callers to `getImmunity` other than just `runImmunity`. (And you can just use `onNegateImmunity` in that case anyway.) But don't worry too much, it was last played in Gen 6.

KrisXV:
This actually broke a ton of OMs to the point that the list is too big

urkerab:
I don't know about a ton of OMs, but I've just realised it broke Partners in Crime, Pokebilities and Shared Power, who were expecting to be able to override `getEffect`. I think I can work around it with a volatile named `ability:intimidate`, but so far I haven't been able to successfully add it yet.

KrisXV:
those were definitely the most notable ones for sure yeah

urkerab:
Yes, that works if I do it properly. (Partners in Crime is special and doesn't want you having Trace, Receiver or Power of Alchemy as a partner ability, and my attempt to fix that completely broke setting `ability:` volatiles. Oops.)

urkerab:
I've since discovered that Dual Wielding and Trademarked were also affected, which are a little harder to fix, but PR #6293 will actually make it easier than their current code, which is nice.

urkerab:
I've since discovered that Let's Go! was affected, but only where AVs were allowed, which they no longer are on main.

Zarel:
Is that easily fixable?

urkerab:
I think commit 02916c1 (on PR #7540) fixes Let's Go!

PR #5852, opened 2019-10-06
Require normal targets to be adjacent in Triples battles

urkerab:
If in a triples battle you use a move that would target an adjacent foe, but by the time it is your turn the only active foe is on the far side of the field, then this should causes the move to fail. Note that it's possible for your adjacent ally to redirect the move, so we can't get claim there is no target until after that check. Otherwise we get stuck in an infinite loop for some reason.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2220,6 +2220,7 @@ export class Battle {
 				adjacentFoes = adjacentFoes.filter(active => active && !active.fainted);
 				if (adjacentFoes.length) return this.sample(adjacentFoes);
 				// no valid target at all, return a foe for any possible redirection
+				return foeActives[frontPosition];
 			}
 		}
 		return pokemon.side.foe.randomActive() || pokemon.side.foe.active[0];


Zarel:
This seems like the wrong line to fix the bug you describe.

data/scripts.js
```diff
@@ -247,7 +247,11 @@ let BattleScripts = {
 			if (damage === this.NOT_FAIL) pokemon.moveThisTurnResult = null;
 			if (damage || damage === 0 || damage === undefined) moveResult = true;
 		} else {
-			if (!targets.length) {
+			let notarget = !targets.length;
+			if (!notarget && this.gameType === 'triples' && ['adjacentFoe', 'adjacentAlly', 'normal', 'randomNormal'].includes(move.target)) {
+				notarget = !this.isAdjacent(targets[0], pokemon);
```
Zarel: My concern here is: How could this possibly be true?

Zarel: `targets` is assigned from

```js
		const {targets, pressureTargets} = pokemon.getMoveTargets(move, target);
```

This should not return a non-adjacent target if the move target-type requires adjacent.

urkerab: If the selected target has already fainted by the time the move executes then `getMoveTarget` calls `resolveTarget`, which tries to find a non-fainted adjacent foe. However, if this also fails then `resolveTarget` just returns any foe, preferring one that is not fainted. I tried changing that behaviour but it just made the tests hang so I went back to how it was done before the spread move refactor.

Zarel: It sounds like we should investigate why the change in `resolveTarget` makes tests hang.

urkerab: I've changed it now so that `resolveTarget` always returns an adjacent foe for normal targeting, preferring a non-fainted one, otherwise just using your opposite foe. This seems to fix Triples without breaking anything else.

Zarel: This is an improvement, but we should probably investigate why we can't return `null` here.

Slayer95: #8735 for a [discussion](https://github.com/smogon/pokemon-showdown/pull/8735#discussion_r857179601) on ``null``

PR #5853, opened 2019-10-06
Movesearch: fix search for stat boosts

Parnassius:
E.g. Close Combat is now shown for `/ms lowers def, lowers spd` and
Mud-Slap for `/ms lowers accuracy`


AnnikaCodes:
Mud-Slap shows up in lowers accuracy searches; Close Combat doesn't, presumably because it's a negative side effect of the move.

Also, this is a pretty dead PR.

PR #5857, opened 2019-10-08
Automatically save replays for global warnings and autolocks

asgdf:
There is of course the problem that the savereplay command requires a user, which does not mesh well with the concept of an autolock. Current convention seems to be to just use the user who is getting locked, which works, but means they'll get the replay sent to them with the popup and all. I don't really like that but I'm not sure what else to do.

### DIFF
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -679,6 +679,12 @@ export const Punishments = new class {
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
 		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
 		Rooms.global.modlog(`(${roomid}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+
+		const roomObject = Rooms.get(room);
+		const userObject = Users.get(user);
+		if (roomObject && roomObject.battle && userObject && userObject.connections[0]) {
+			Chat.parse('/savereplay', roomObject, userObject, userObject.connections[0]);
+		}
 	}
 	unlock(name: string) {
 		const user = Users.get(name);
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -683,7 +683,7 @@ export const Punishments = new class {
 		const roomObject = Rooms.get(room);
 		const userObject = Users.get(user);
 		if (roomObject && roomObject.battle && userObject && userObject.connections[0]) {
-			Chat.parse('/savereplay', roomObject, userObject, userObject.connections[0]);
+			Chat.parse('/savereplay forpunishment', roomObject, userObject, userObject.connections[0]);
 		}
 	}
 	unlock(name: string) {
--- a/server/chat-commands.js
+++ b/server/chat-commands.js
@@ -1896,6 +1896,9 @@ const commands = {
 		if (userid !== toID(this.inputUsername)) this.add(`|unlink|${toID(this.inputUsername)}`);
 
 		targetUser.lastWarnedAt = now;
+
+		// Automatically upload replays as evidence/reference to the punishment
+		if (globalWarn && room.battle) this.parse('/savereplay forpunishment');
 	},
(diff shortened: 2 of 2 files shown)


server/punishments.ts
```diff
@@ -679,6 +679,7 @@ export const Punishments = new class {
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
 		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
 		Rooms.global.modlog(`(${roomid}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+		Chat.parse('/savereplay', room as Room, user as User, (user as User).connections[0]);
```
whalemer: this cast isn't always right, see: https://github.com/Zarel/Pokemon-Showdown/blob/c990e015dbff2cece541833e77cb4a903a61cd2a/server/punishments.ts#L1452

asgdf: Oh that's fair.

server/punishments.ts
```diff
@@ -679,6 +679,7 @@ export const Punishments = new class {
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
 		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
 		Rooms.global.modlog(`(${roomid}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+		Chat.parse('/savereplay', Rooms.get(room), user as User, (user as User).connections[0]);
```
whalemer: This still isn't right, you can trigger this with a userid (ie if you nameblacklist someone with 2 RBs). You also need to make sure that `connections[0]` is actually defined (ie, punishing an offline user)

asgdf: I didn't even check whether the room was a battle at all. What was I thinking? I guess I'm dumb.

server/punishments.ts
```diff
@@ -679,6 +679,12 @@ export const Punishments = new class {
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
 		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
 		Rooms.global.modlog(`(${roomid}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+
+		const roomObject = Rooms.get(room);
+		const userObject = Users.get(user);
+		if (roomObject.battle && userObject && userObject.connections[0]) {
```
whalemer: Not checking if `roomObject` exists (the return type for getRoom is wrong) makes me nervous, but I think it's always going to be fine in practice.

Zarel: Well, it's going to need a TS override now that we merged your PR. :p

server/punishments.ts
```diff
@@ -679,6 +679,12 @@ export const Punishments = new class {
 		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
 		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
 		Rooms.global.modlog(`(${roomid}) AUTO${namelock ? `NAME` : ''}LOCK: [${userid}]${ipStr}: ${reason}`);
+
+		const roomObject = Rooms.get(room);
+		const userObject = Users.get(user);
+		if (roomObject && roomObject.battle && userObject && userObject.connections[0]) {
```
Zarel:
```js
		if (roomObject?.battle && userObject?.connections[0]) {
```

wtb optional chaining ;_;

Zarel:
> Current convention seems to be to just use the user who is getting locked, which works, but means they'll get the replay sent to them with the popup and all.

Wait, what???? I thought current convention was to use the mod who did the locking........

server/punishments.ts
```diff
@@ -679,6 +679,12 @@ export const Punishments = new class {
...
+			Chat.parse('/savereplay', roomObject, userObject, userObject.connections[0]);
```
Zarel: Why not `/savereplay forPunishment` like locks?

Zarel: Without it, hidden room replays are inaccessible to mods.

asgdf:
Okay, removed the auth check in /savereplay forpunishment, and made autolocks use them.

PR #5860, opened 2019-10-09
Rooms: Rooms.get can return null

whalemer:
Now that we have typescript's `!` in most places, I much prefer using that to asserting `getRoom`'s return value


Zarel:
What about `Users.get`?

whalemer:
>What about `Users.get`?

Properly typed as returning `User | null`

PR #5863, opened 2019-10-10
Modnote Tours Player Count Upon Start

Aurolux:


server/tournaments/index.ts
```diff
@@ -607,6 +607,7 @@ export class Tournament extends Rooms.RoomGame {
 		}
 		this.isBracketInvalidated = true;
 		this.room.add(`|tournament|start|${this.players.length}`);
+		this.modlog('TOUR START', null, `The tournament has started with ${this.players.length} users.`);	// Modnote number of players and when a tour starts
```
```suggestion
		this.modlog('TOUR START', null, `The tournament has started with ${this.players.length} users.`);
```

This comment is useless.

server/tournaments/index.ts
```diff
@@ -607,6 +607,7 @@ export class Tournament extends Rooms.RoomGame {
 		}
 		this.isBracketInvalidated = true;
 		this.room.add(`|tournament|start|${this.players.length}`);
+		this.modlog('TOUR START', null, `The tournament has started with ${this.players.length} users.`);
```
```suggestion
		output.modlog('TOUR START', null, `The tournament has started with ${this.players.length} users.`);
```

Your build error is because your CommandContext is `output`, not `this`.

Aurolux: Thanks. Coming from Java I struggle a bit with the CommandContext element of JS. That solved pretty much all the problems that I ran into. Do you think this line of code would be better in the individual start and autostart commands or here?

Zarel: I think it's better to do it here; in general, less repetition is better. It's called "DRY" – "Don't Repeat Yourself", and it's a standard programming principle.

(Also, ideally don't click "Resolve conversation" after asking a question but before getting an answer for it. "Resolve" means that the conversation is finished.)

Aurolux: Alright noted about "Resolve conversation". After shortening the modlog entry message, I see no further changes currently.

The code has been tested on a dev server, and it works just fine.

Zarel:
I guess the main question is, did someone request this?

Aurolux:
No. I wanted to PR this as an aid to log Tournament Participation in wifi, and saw that Tournament Participation numbers were not logged by Tours. Other rooms may have some functionality for this, and I don't see it as modlog clutter, but those are just my reasons for PR as of now.

Zarel:
Interesting. Roomlog might be a good place, too.

@bumbadadabum, opinions?

Aurolux:
I believe it's already noted in the roomlog, though I just wanted to add a modlog line for it too because it's hard to access and has nothing linking to the roomlog line so you have to fish it up in roomlog.

bumbadadabum:
Sorry for taking 2 weeks whoops

PR #5867, opened 2019-10-10
Allow loading chat plugins from subdirectories

asgdf:
As requested by @scheibo. This looks through all files in all subdirectories and tries to load them as chat plugins. I excluded hidden directories so it doesn't bother looking through .git in the case that someone checks out a chat plugin from a separate repository here.

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -400,7 +400,7 @@ class FSPath {
 		return fs.statSync(this.path).isFile();
 	}
 
-	async isDir(): Promise<boolean> {
+	async isDirectory(): Promise<boolean> {
 		return new Promise((resolve, reject) => {
 			fs.stat(this.path, (err, stats) => {
 				err ? reject(err) : resolve(stats.isDirectory());
@@ -408,7 +408,7 @@ class FSPath {
 		});
 	}
 
-	isDirSync(): boolean {
+	isDirectorySync(): boolean {
 		return fs.statSync(this.path).isDirectory();
 	}
 }
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -388,27 +388,27 @@ class FSPath {
 		fs.unwatchFile(this.path);
 	}
 
-	async isFile(): Promise<boolean> {
-		return new Promise((resolve, reject) => {
+	async isFile() {
+		return new Promise<boolean>((resolve, reject) => {
 			fs.stat(this.path, (err, stats) => {
 				err ? reject(err) : resolve(stats.isFile());
 			});
 		});
 	}
 
-	isFileSync(): boolean {
+	isFileSync() {
 		return fs.statSync(this.path).isFile();
 	}
 
-	async isDirectory(): Promise<boolean> {
-		return new Promise((resolve, reject) => {
+	async isDirectory() {
+		return new Promise<boolean>((resolve, reject) => {
 			fs.stat(this.path, (err, stats) => {
 				err ? reject(err) : resolve(stats.isDirectory());
 			});
 		});
 	}
 
-	isDirectorySync(): boolean {
+	isDirectorySync() {
 		return fs.statSync(this.path).isDirectory();
 	}
 }
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -387,6 +387,30 @@ class FSPath {
 	unwatch() {
 		fs.unwatchFile(this.path);
 	}
+
+	async isFile(): Promise<boolean> {
+		return new Promise((resolve, reject) => {
+			fs.stat(this.path, (err, stats) => {
+				err ? reject(err) : resolve(stats.isFile());
+			});
+		});
+	}
+
+	isFileSync(): boolean {
+		return fs.statSync(this.path).isFile();
+	}
+
+	async isDir(): Promise<boolean> {
+		return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


server/chat.ts
```diff
@@ -1417,8 +1417,24 @@ export const Chat = new class {
 
 		// Install plug-in commands and chat filters
 
+		// All resulting filenames will be relative to basePath
+		const getFiles = (basePath: string, path: string): string[] => {
+			const filesInThisDir = FS(basePath + '/' + path).readdirSync();
```
scheibo:
```suggestion
			const filesInThisDir = FS(`${basePath}/${path}`).readdirSync();
```

server/chat.ts
```diff
@@ -1417,8 +1417,24 @@ export const Chat = new class {
...
+			let allFiles: string[] = [];
+			for (const file of filesInThisDir) {
+				const fileWithPath = path + (path ? '/' : '') + file;
+				if (FS(basePath + '/' + fileWithPath).isDirSync()) {
```
scheibo:
```suggestion
				if (FS(`${basePath}/${fileWithPath}`).isDirSync()) {
```

Slayer95:
Is this deep-recursive? I oppose. We should support the existence of files acting as modular dependencies for chat plugin main files.

Zarel:
Yeah, just make `private.js` include everything in `private/`. I don't like this as a generalized solution.

monsanto:
If this route is chosen then `loadPlugins` should be refactored into a single function `loadPlugin(fileName)`.

asgdf:
> Yeah, just make `private.js` include everything in `private/`. I don't like this as a generalized solution.

As in, have a public private.js file at the top level that does nothing but inlcude every file in `private/`? Since the whole point of this was to be able to check out chat plugins in `private/` and having it work out of the box

Zarel:
Maybe hardcode `chat-plugins/private/`?

asgdf:
Okay, now it does things the old way for everything except `chat-plugins/private/`, which will have every js/ts file in it loaded individually.

lib/fs.ts
```diff
@@ -387,6 +387,30 @@ class FSPath {
...
+				err ? reject(err) : resolve(stats.isFile());
+			});
+		});
+	}
+
+	isFileSync(): boolean {
+		return fs.statSync(this.path).isFile();
+	}
+
+	async isDir(): Promise<boolean> {
```
Zarel: Hm, if `stat` uses `isDirectory`, maybe we should too?

server/chat.ts
```diff
@@ -1417,16 +1417,39 @@ export const Chat = new class {
...
+		try {
+			if (FS('server/chat-plugins/private').isDirSync()) {
+				files = files.concat(getFiles('server/chat-plugins', 'private'));
+			}
+		} catch (err) {
+			if (err.code !== 'ENOENT') throw err;
+		}
 
 		for (const file of files) {
 			let plugin;
```
Zarel: Maybe the contents of this loop should be refactored out into a `loadPlugin` function, which could then be called inline.

asgdf:
Sure, did both of these.

lib/fs.ts
```diff
@@ -387,6 +387,30 @@ class FSPath {
...
+
+	async isFile(): Promise<boolean> {
+		return new Promise((resolve, reject) => {
+			fs.stat(this.path, (err, stats) => {
+				err ? reject(err) : resolve(stats.isFile());
+			});
+		});
+	}
+
+	isFileSync(): boolean {
```
Zarel: I think this can be inferred?

lib/fs.ts
```diff
@@ -387,6 +387,30 @@ class FSPath {
 	unwatch() {
 		fs.unwatchFile(this.path);
 	}
+
+	async isFile(): Promise<boolean> {
+		return new Promise((resolve, reject) => {
```
Zarel:
```suggestion
	async isFile() {
		return new Promise<boolean>((resolve, reject) => {
```

Zarel: I _figured_ this wouldn't work but it was worth a try.

asgdf: I'm not sure what you mean - doing that works for me. But is there a point in doing it?

Zarel: Oh, I mean the two-line suggested change didn't work; you'd have to commit it manually.

As for the point, the main point is that it's DRYer, it uses `Promise` once instead of twice.

asgdf: Ah. Sure.

Zarel:
I suppose there's an argument for explicitly documenting return types, especially for the purposes of readability and catching bugs. But we don't do this for the rest of the codebase. I'm willing to do it for things in `lib/` specifically, though.

asgdf:
Is that a "you should add them back", or an "I guess if you really want to you can add them back"?

Zarel:
That's a "we should discuss whether or not we should add them back". @scheibo, opinions?

asgdf:
Personally I have a slight preference for types in method signatures even when not strictly necessary. Just not strong enough of a preference to really argue.

Zarel:
Anyway, this looks good to me, I won't block it on bikeshedding.

Zarel:
I'm fine with adding types, but it should be a linter rule.

I think it's noticeably not very DRY on short functions, but it might be worth it overall.

PR #5873, opened 2019-10-11
Add a full ability description for Contrary + Simple

Alumn2000:
https://www.smogon.com/forums/threads/add-a-full-ability-description-for-contrary.3652011/

@Marty-D , if you want add something or change, say it.

https://bulbapedia.bulbagarden.net/wiki/Contrary_(Ability)

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -497,7 +497,7 @@ let BattleAbilities = {
 		num: 14,
 	},
 	"contrary": {
-		desc: "If this Pokemon has a stat stage raised it is lowered instead, and vice versa. Stat stage increases caused by the Z-Power effects of status Z-Moves are not inverted by Contrary, except for Extreme Evoboost.",
+		desc: "If this Pokemon has a stat stage raised it is lowered instead, and vice versa. This Ability does not affect the stat stages received from Z-Power effects.",
 		shortDesc: "If this Pokemon has a stat stage raised it is lowered instead, and vice versa.",
 		onBoost(boost, target, source, effect) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
I think it would be simpler to for the second sentence to read something like "Stat stage increases caused by the Z-Power effects of status Z-Moves are not inverted by Contrary, except for Extreme Evoboost".

Alumn2000:
I agree. Added

Marty-D:
If you clarify it like that then you're implying Extreme Evoboost has a Z-Power effect that somehow doesn't apply here. It doesn't; Extreme Evoboost is like Swords Dance, the stat stage increases are what the move does. Copying Bulbapedia is bad!

I was just going to add something like "This Ability does not affect the stat stages received from Z-Power effects." to both Contrary and Simple. You also have to override the description in the Gen 6 mod to remove Gen 7-specific stuff.

Alumn2000:
I would be glad if someone could check that I didn't crashed something.

Zarel:
I would probably still clarify like `(boosting Z-moves like Extreme Evoboost are not Z-power effects)` or something.

Zarel:
> I would be glad if someone could check that I didn't crashed something.

It's a wording change; unless the checks fail (this thing), there's basically no way you could mess it up.

image

Alumn2000:
ok. Thanks Zarel.

Alumn2000:
@Marty-D , say if it's ok. If not, then just put the sentence in the comment and I will just c/c.

Marty-D:
Thanks!

PR #5877, opened 2019-10-12
Rename TryImmunity to TryInvulnerability

urkerab:
It's a check for semi-invulnerability rather than any sort of immunity.

### DIFF
--- a/data/mods/gen1/moves.js
+++ b/data/mods/gen1/moves.js
@@ -276,7 +276,7 @@ let BattleMovedex = {
 		effect: {
 			duration: 2,
 			onLockMove: 'dig',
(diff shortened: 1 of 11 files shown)


Zarel:
Maybe just `onInvulnerability`?

Zarel:
I did say "maybe". What does everyone else think?

Zarel:
Mostly, I feel like `onTryBlah` is only used if we already have an `onBlah` that does something different.

urkerab:
... that's why it's a separate commit so it can be easily reverted.

Zarel:
I didn't say you did anything wrong. :p

Marty-D:
Yes! Thank you!

PR #5882, opened 2019-10-13
Fix nameblacklist target auth check

asgdf:


Zarel:
Ugh, at this point I'm going to say no, fix the call signatures of these functions.

Zarel:
No function should be taking an `id` parameter instead of a `userid` parameter.

Zarel:
Heck, I'd nearly advocate changing `id` back to `userid` everywhere.

We should maybe even release a `UserID` type.

Zarel:
`toID` should return `ID & UserID`.

asgdf:
The function takes a user object as an argument, but because it just uses the ID from that, we mock up a fake user object that just contains an ID. That's my understanding what's happening here.

Zarel:
Ohhhh. That's fine, then.

Zarel:
I don't like the "fake user" pattern, though; we should maybe do something about that.

PR #5883, opened 2019-10-13
Use npx instead of relying on globally installed tools 

danielmayorga:
Now you can manage the versioning of these packages in the project and not have to worry about some dev not having the right version of the typescript compiler (or any of the other tools) globally installed.

Resource:
https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,12 +23,12 @@
   "scripts": {
     "start": "node pokemon-showdown start",
     "build": "node build",
-    "tsc": "tsc",
-    "lint": "eslint --cache . && tslint --project .",
+    "tsc": "npx tsc",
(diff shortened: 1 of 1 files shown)


monsanto:
This is redundant, scripts run from `package.json` have the local `node_modules/.bin` prepended to `PATH`

xfix:
I don't see the point, those libraries are installed by [package.json](https://github.com/Zarel/Pokemon-Showdown/blob/master/package.json). If you use `npm install`, those tools should install themselves, and by using `npx`, we lose versioning.

danielmayorga:
Good point @monsanto. I didnt realize npm scripts had that property. TIL

PR #5885, opened 2019-10-14
Moves: Fix Rest at full HP

LuisF3:
Fix Rest at full HP should display "'s HP is full!", not "But it failed!".
https://github.com/Zarel/Pokemon-Showdown/projects/3#card-24620941

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -13501,7 +13501,8 @@ let BattleMovedex = {
 		flags: {snatch: 1, heal: 1},
 		onTryMove(pokemon) {
 			if (pokemon.hp < pokemon.maxhp && pokemon.status !== 'slp' && !pokemon.hasAbility('comatose')) return;
-			this.add('-fail', pokemon);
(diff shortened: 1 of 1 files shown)


Zarel:
I remember the last time someone tried to fix this, we learned this was really complicated. @Marty-D? @DaWoblefet?

Zarel:
Also, `-message` is incorrect; you should actually add the message to the client. Which I also thought we already did...

DaWoblefet:
Yeah, Volco was [working on it](https://github.com/Zarel/Pokemon-Showdown/pull/5746) but hasn't updated it in over a month. See my [specific comment](https://github.com/Zarel/Pokemon-Showdown/pull/5746#issuecomment-527264803) about healing messages.

Zarel:
I guess I'm just fixing this myself. :/

Zarel:
It looks like the `heal.fail` message is already in the code.

PR #5890, opened 2019-10-17
Fix duplicate globalModlog entries in some cases

Lusamine:
Stops globalModlog from duplicating entries in the global modlog in specific cases.


Zarel:
Thanks!

PR #5894, opened 2019-10-17
Centralise invulnerability checks for No Guard and Lock On

urkerab:
if (source.hasAbility('noguard') || target.hasAbility('noguard')) {
       return;
    }
    if (source.volatiles['lockon'] && target === source.volatiles['lockon'].source) return;

Rather than copy/pasting this code block for every semi-invulnerability move, I thought that perhaps the tests could all be done centrally (except for Gen 2, which is a special snowflake), so I made a zero return from `onInvulnerability` mean not invulnerable due to No Guard or Lock On, while false continues to mean semi-invulnerable and true means neither of the above.

I wasn't sure what was going on with Lock On in Gen 4 though; the description differs from later gens in that Baton Pass is not specifically exempted, but there's no relevant code change? I added one in, but...

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -2170,6 +2170,10 @@ let BattleAbilities = {
 	},
 	"noguard": {
 		shortDesc: "Every move used by or against this Pokemon will always hit.",
+		onAnyInvulnerabilityPriority: 1,
+		onAnyInvulnerability(target, source, move) {
+			if (move && (source === this.effectData.target || target === this.effectData.target)) return 0;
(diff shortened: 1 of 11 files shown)


Zarel:
Looks good to me. @Marty-D?

Marty-D:
Gen 2 and 4 Lock-On/Mind Reader don't appear to be implemented as written but I won't block on that since it's out of scope for this change and unlike any other effects. This is a great improvement, nice work! 👍

PR #5895, opened 2019-10-17
Fix activation of Defiant and Competitive when boosts are maxed

urkerab:
Currently if you're already at `+6` when your Defiant or Competitive activates, you do get the "[POKEMON]'s [STAT] won't go any higher!" message, but without the Defiant or Competitive activation that showed why the stat tried to rise.

Compare Gooey which shows the activation and then the stat change, whether it succeeds or fails. It is granted a special case in the `boost` function to avoid the ability activation therein allowing it to enforce it prior to the invocation (necessary for Clear Body to function), so I could have just copied that hack for Defiant and Competitive.

But rather than having more special cases, I thought it was better to use the flag that Gooey was already passing in to `boost` to get special behaviour in the failure case and grant all the relevant Abilities special behaviour in the success case too.

Tangling Hair is a clone of Gooey, and so should be passing the same parameters to `boost`. I didn't bother to check how the invalid parameters make it misbehave. The change allows the removal of its special case from `boost` too.

Intimidate is supposed to display the "won't go any lower!" message if the victim is already at -6. I added the Gooey flag to fix that and this also allows the removal of its special case from `boost`.

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -476,6 +476,7 @@ let BattleAbilities = {
 				}
 			}
 			if (statsLowered) {
+				this.add('-ability', target, 'Competitive');
 				this.boost({spa: 2}, target, target, null, true);
 			}
 		},
@@ -647,6 +648,7 @@ let BattleAbilities = {
 				}
 			}
 			if (statsLowered) {
+				this.add('-ability', target, 'Defiant');
 				this.boost({atk: 2}, target, target, null, true);
 			}
 		},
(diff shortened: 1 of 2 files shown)


Zarel:
We should maybe use `-activate` rather than `-ability`?

urkerab:
So that's `this.add('-activate', target, 'ability: Competitive');` etc? As there are ~20 of those, should that really be done in a separate PR?

Zarel:
Possibly.

Zarel:
I do like this as-is.

PR #5898, opened 2019-10-19
Hide stacktraces for private code

xfix:
Errors in private code aren't actionable for Showdown contributors without access to private code, so no need to show them the details of those, only for them to realize those files aren't in the repo. This unnecessarily wastes time.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -987,7 +987,7 @@ export class GlobalRoom extends BasicRoom {
 		}
 		const stack = stackLines.slice(0, 2).join(`<br />`);
 		let crashMessage;
-		if (/private|chat-plugins&#x2f;_/.test(stack)) {
+		if (/private/.test(stack)) {
 			crashMessage = `|html|<div class="broadcast-red"><b>${crasher} has crashed in private code</b></div>`;
 		} else {
 			crashMessage = `|html|<div class="broadcast-red"><b>${crasher} has crashed:</b> ${stack}</div>`;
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -986,7 +986,12 @@ export class GlobalRoom extends BasicRoom {
 			}
 		}
 		const stack = stackLines.slice(0, 2).join(`<br />`);
-		const crashMessage = `|html|<div class="broadcast-red"><b>${crasher} has crashed:</b> ${stack}</div>`;
+		let crashMessage;
+		if (/private|chat-plugins&#x2f;_/.test(stack)) {
```
Zarel: `chat-plugins&#x2f;_`?

xfix: Yeah, there is `Chat.escapeHTML` used before for whatever reason.

Zarel: Yes, but we have plenty of chat plugins that _aren't_ private code. All the chat plugins that _are_ private code will have `private` in their name.

bumbadadabum: I would codify that somewhere in the plugin loader, and also maybe make sure something like "private-messages" wouldn't be parsed as a private plugin either.

Zarel: It's already codified by `.gitignore`. :|

Zarel: It's chat plugins ending in `-private`.

```suggestion
		if (/private/.test(stack)) {
```
If `_` hasn't been fixed yet, it will be. Either way, `_` as a privacy marker is not something I want to support.

Zarel: I'd also probably move the notification to the staff or upper staff room instead of hiding it completely.

PR #5899, opened 2019-10-19
Support custom Max Level and Team Length in tours

TheMezStrikes:
I tried my absolute best to keep code standards good, but truth is I'm just really not good at things like variable naming. Adjustments will most likely need to be made by someone who's actually good at those things

I do know however that this code does what it needs to (`/tour rules Max Level:5` sets max level to 5, and `/tour rules Team Length:3-1` sets the teamLength to set max team length to 3, and allow one of those to be picked into battle)
As far as I can tell the syntax I used does not interfere with anything already in place, and shouldn't interfere with anything that could be added in the future

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -76,40 +76,40 @@ const DATA_FILES = {
 };
 
 interface BasicRule {
-	ruleType: 'basicRule',
-	rule: string,
+	ruleType: 'basicRule';
+	rule: string;
 }
 
 interface ComplexTeamRule {
-	ruleType: 'complexTeamRule',
-	innerRule: number,
-	source: number,
-	limit: string,
-	bans: string[],
+	ruleType: 'complexTeamRule';
+	innerRule: number;
+	source: number;
+	limit: string;
+	bans: string[];
 }
 
 interface ComplexRule {
-	ruleType: 'complexRule',
-	innerRule: number,
-	source: number,
-	limit: string,
-	bans: string[],
+	ruleType: 'complexRule';
+	innerRule: number;
+	source: number;
+	limit: string;
+	bans: string[];
 }
 
 interface BanRule {
-	ruleType: 'banRule',
-	ban: string,
+	ruleType: 'banRule';
+	ban: string;
 }
 
 interface TeamLengthRule {
-	ruleType: 'teamLengthRule',
-	onBattle: number,
+	ruleType: 'teamLengthRule';
+	onBattle: number;
 	onValidate: number,
 }
 
 interface LevelRule {
-	ruleType: 'levelRule',
-	level: number,
+	ruleType: 'levelRule';
+	level: number;
 }
 
 type ValidatedRule = BasicRule | ComplexTeamRule | ComplexRule | BanRule | TeamLengthRule | LevelRule;
@@ -847,6 +847,7 @@ export class ModdedDex {
 			if (rule.startsWith('!')) continue;
 
 			const ruleSpec = this.validateRule(rule, format);
+
 			switch(ruleSpec.ruleType) {
 				case 'complexTeamRule':
 					const complexTeamBan: Data.ComplexTeamBan = [ruleSpec.innerRule, ruleSpec.source, ruleSpec.limit, ruleSpec.bans] as Data.ComplexTeamBan;
@@ -935,19 +936,19 @@ export class ModdedDex {
 				if (checkTeam) {
 					return {
 						ruleType: 'complexTeamBan'
-						innerRule: innerRule,
+						innerRule,
 						source: '',
-						limit: limit,
-						bans: bans,
+						limit,
+						bans,
 					};
 				}
 				if (bans.length > 1 || limit > 0) {
 					return {
 						ruleType: 'complexBan'
-						innerRule: innerRule,
+						innerRule,
 						source: '',
-						limit: limit,
-						bans: bans,
+						limit,
+						bans,
 					};
 				}
 				throw new Error(`Confusing rule ${rule}`);
@@ -974,8 +975,8 @@ export class ModdedDex {
 						}
 						return {
 							ruleType: 'teamLengthRule',
-							onBattle: onBattle,
-							onValidate: onValidate,
+							onBattle,
+							onValidate,
 						};
 					}
 				} else if (type === "level" || type === "maxlevel") {
@@ -983,7 +984,7 @@ export class ModdedDex {
 					if (!isNaN(level)) {
 						return {
 							ruleType: 'levelRule'
-							level: level
+							level,
 						};
 					}
 				}
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -832,12 +832,12 @@ export class ModdedDex {
 					const vals = ruleValue.split('-');
 					const onValidate = parseInt(vals[1]);
 					const onBattle = parseInt(vals[0]);
-					format.teamLength = {
+					ruleTable.teamLength = {
 						validate: [onBattle, onValidate],
 						battle: onBattle,
 					};
 				} else if (ruleName === 'level') {
-					format.maxLevel = parseInt(ruleValue);
+					ruleTable.maxLevel = parseInt(ruleValue);
 				}
 				continue;
 			}
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -982,7 +982,7 @@ export class ModdedDex {
 					const level = parseInt(parts[1]);
 					if (!isNaN(level)) {
 						return {
-							ruleType: 'levelRule'
+							ruleType: 'levelRule',
 							level,
 						};
 					}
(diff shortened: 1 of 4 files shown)


sim/dex.ts
```diff
@@ -826,6 +826,22 @@ export class ModdedDex {
 				ruleTable.set(ruleSpec, '');
 				continue;
 			}
+			if (ruleSpec.split(":").length > 1) {
```
Zarel: This is an unnecessarily inefficient way to write `ruleSpec.includes(":")`

sim/dex.ts
```diff
@@ -826,6 +826,22 @@ export class ModdedDex {
...
+				if (parts[0] === 'teamlength') {
+					const vals = ruleSpec.split(":")[1].split('-');
+					const onValidate = parseInt(vals[1]);
+					const onBattle = parseInt(vals[0]);
+					format.teamLength = {
+						validate: [onBattle, onValidate],
+						battle: onBattle
+					}
+				}
+				else if (parts[0] === 'level') {
```
Zarel: Ugh I hate that tslint doesn't have whitespace validation. PS requires `} else ...`

sim/dex.ts
TheMezStrikes: Didn't know .includes existed, will change along with fixing build errors

sim/dex.ts
```diff
@@ -826,6 +826,22 @@ export class ModdedDex {
...
+					format.maxLevel = parseInt(parts[1]);
+				}
+				continue;
```
Zarel: The `continue` combined with the check above would eliminate `:` from being supported in any other rule name. This does happen after `+` and `-` so you're at least not filtering out actual legitimate bans, but I'm still not super comfortable with this.

sim/dex.ts
```diff
@@ -826,6 +826,22 @@ export class ModdedDex {
 				ruleTable.set(ruleSpec, '');
 				continue;
 			}
+			if (ruleSpec.split(":").length > 1) {
+				const parts = ruleSpec.split(":");
```
```suggestion
				const [ruleName, ruleValue] = ruleSpec.split(":");
```

We generally prefer named parts. Makes it clearer what's going on.

sim/dex.ts
```diff
@@ -826,6 +826,22 @@ export class ModdedDex {
...
+					const vals = ruleSpec.split(":")[1].split('-');
+					const onValidate = parseInt(vals[1]);
+					const onBattle = parseInt(vals[0]);
+					format.teamLength = {
+						validate: [onBattle, onValidate],
+						battle: onBattle
+					}
+				}
+				else if (parts[0] === 'level') {
+					format.maxLevel = parseInt(parts[1]);
```
Zarel: It might be useful to consider `maxForcedLevel` and `forcedLevel`. These are usually what you want for VGC-likes.

sim/dex.ts
TheMezStrikes: I figured any future rules using `:` would most likely end up in the same code block (before the continue) anyway as it doesn't make much sense to split by `:` twice

sim/dex.ts
Zarel: Oh, wait, Travis _did_ complain about it. I take back my complaints.

Zarel:
I wonder if `=` might be a better choice than `:`.

TheMezStrikes:
In a very hacky fashion, I made it support both

sim/dex.ts
TheMezStrikes: Yea Travis complained and I fixed

sim/dex.ts
TheMezStrikes: Hmm but that conflicts with trying to do past generation LC, where maxLevel is what you're looking for. A compromise would be to separate the rules but I think the average PS staff member wouldn't be able to figure out the difference, so you lose intuitiveness there

TheMezStrikes:
Okay now I'm just making mistakes all over the place, I need to take a step back and figure out what I'm doing

TheMezStrikes:
Going to need someone to go over that failed Travis build and explain to me how to make it not fail. From what I can tell you're working on a copy of `format` so any changes you make are non-permanent and don't need to be read-only

Zarel:
I think your main problem is that `format` is immutable here.

When I tried to tackle a similar problem in the past, my solution was to shove these values (`maxLevel` and `teamLength`) _inside_ the `ruleTable`. That might work for you?

TheMezStrikes:
That'd require some changes to `battle.ts` and `team-validator.ts` but seems possible. Gonna try getting that to work

sim/team-validator.ts
```diff
@@ -309,8 +309,8 @@ export class TeamValidator {
 		set.nature = dex.getNature(Dex.getString(set.nature)).name;
 		if (!Array.isArray(set.moves)) set.moves = [];
 
-		const maxLevel = format.maxLevel || 100;
-		const maxForcedLevel = format.maxForcedLevel || maxLevel;
+		const maxLevel = this.ruleTable.maxLevel || format.maxLevel || 100;
```
Zarel:
```suggestion
		const maxLevel = ruleTable.maxLevel;
```

I would implement the fallbacks to `format.maxLevel` and `100` inside `ruleTable.maxLevel`, rather than out here.

TheMezStrikes: I'm sorry, I don't know how to work with classes well enough to do that

Zarel: Although it makes sense to at least have the `100` fallback out here, so you can detect _conflicting_ `maxLevel` rules.

Zarel: That's fair.

sim/dex.ts
```diff
@@ -897,7 +912,31 @@ export class ModdedDex {
...
+					let onBattle = parseInt(options[1]);
+					if (options.length === 1) onBattle = onValidate;
+					if (!isNaN(onValidate) && !isNaN(onBattle)) {
+						if (onValidate < onBattle) {
+							throw new Error("You can't bring more Pokémon to a battle than the maximum team length");
+						}
+						if (onValidate > 24) {
+							throw new Error("You may not have more than 24 Pokémon on a team");
+						}
+						return `teamlength:${onBattle}-${onValidate}`;
```
Zarel: It would probably make sense to return an array like `complexBan` and `complexTeamBan`, instead of needing to parse this twice?

TheMezStrikes: The return value is added to the format string so needs to return a string unless I'm really stupid

Zarel:
```ts
			const ruleSpec = this.validateRule(rule, format);
			if (typeof ruleSpec !== 'string') {
```

Look at line 810: there's a test for the return value (for `complexBan` etc).

TheMezStrikes:
Sorry got sick and couldn't update this for a while. I've implemented all I could as you suggested, from here it can either be merged or be taken over by someone more capable

TheMezStrikes:
I don't have the slightest clue what I'm doing wrong at this point

sim/dex.ts
Zarel: I'm confused why you resolved this. This doesn't seem to have happened at all? You're still returning a string here.

sim/dex.ts
```diff
@@ -815,6 +815,16 @@ export class ModdedDex {
 				} else if (ruleSpec[0] === 'complexBan') {
 					const complexBan: Data.ComplexBan = ruleSpec.slice(1) as Data.ComplexBan;
 					ruleTable.addComplexBan(complexBan[0], complexBan[1], complexBan[2], complexBan[3]);
+				} else if (ruleSpec[0] === 'teamlength') {
+					const onValidate = ruleSpec[2] as number;
```
Zarel: You shouldn't need to cast this; you should just define `ruleSpec` as a union type such that this will automatically be true.

monsanto: Akasianse was looking for help on their PR and mentioned this comment. `ruleSpec[2]` is already the union type `string | number | number[]`, due to the return type of `validateRule`. It is forced to be a `number` by `rule.teamLength.validate : [number, number]`. There is no way to avoid the cast without modifying `validateRule` to return a discriminated union with tags `'complexBan'`, `'teamlength'`, `'level'`, etc. I think that's a good idea but I said that they should talk to you first before modifying functions outside the scope of their PR.

sim/dex.ts
TheMezStrikes: You're absolutely correct, let me check how it's possible because I could've sworn I changed that

TheMezStrikes: Apologies, I did indeed forget to commit that edit

sim/dex.ts
Zarel: Yes, the discriminated union was indeed what I was talking about.

Zarel: Also, `validateRule`'s return signature is already modified in this PR.

TheMezStrikes:
Will start fixing that later

PR #5902, opened 2019-10-20
Server: Use readonly where appropriate

TheJetOU:


server/chat.ts
```diff
@@ -152,14 +152,13 @@ class PatternTester {
 
 // These classes need to be declared here because they aren't hoisted
 class MessageContext {
+	readonly user: User;
+	readonly language: string | null;
```
Zarel: idk it makes sense that a command would want to modify this.

Zarel: Like if someone does `/setlanguage English` the response should probably be in English.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
 	 * joining the room. Should roughly match what's on everyone's
 	 * screen.
 	 */
-	log: Roomlog | null;
-	game: RoomGame | null;
-	battle: RoomBattle | null;
-	active: boolean;
-	muteQueue: MuteEntry[];
+	readonly log: Roomlog | null;
+	readonly game: RoomGame | null;
```
Zarel: This is patently false, starting a roomgame (like `/hangman start`) will modify this.

If you're doing this based on what currently mutates these, keep in mind that a lot of chat plugins haven't been TypeScripted yet.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
-	roomid: RoomID;
-	title: string;
-	parent: Room | null;
-	aliases: string[] | null;
-	users: {[userid: string]: User};
-	userCount: number;
-	type: 'chat' | 'battle' | 'global';
-	auth: {[userid: string]: string} | null;
+	readonly roomid: RoomID;
+	readonly parent: Room | null;
```
Zarel: Modified by chat plugins.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
 export abstract class BasicRoom {
-	roomid: RoomID;
-	title: string;
-	parent: Room | null;
-	aliases: string[] | null;
-	users: {[userid: string]: User};
-	userCount: number;
-	type: 'chat' | 'battle' | 'global';
-	auth: {[userid: string]: string} | null;
+	readonly roomid: RoomID;
```
Zarel: Planned to be modifiable – it's one of the suggestions for new devs iirc.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
+	readonly parent: Room | null;
+	readonly aliases: string[] | null;
```
Zarel: Modified by chat plugins.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
 	 */
-	log: Roomlog | null;
-	game: RoomGame | null;
-	battle: RoomBattle | null;
-	active: boolean;
-	muteQueue: MuteEntry[];
+	readonly log: Roomlog | null;
+	readonly game: RoomGame | null;
+	readonly battle: RoomBattle | null;
+	readonly active: boolean;
```
Zarel: ...should be actually modified by the server; I'm confused why this isn't a TypeScript error.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
+	readonly muteQueue: MuteEntry[];
+	readonly chatRoomData: AnyObject | null;
```
Zarel: Modified by core chat commands.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
+	readonly reportJoins: boolean;
```
Zarel: Modified by core chat commands.

server/rooms.ts
```diff
@@ -57,41 +57,41 @@ type Announcement = import('./chat-plugins/announcements').AnnouncementType;
...
+	readonly staffRoom: boolean;
```
Zarel: Modified by core chat commands.

server/rooms.ts
```diff
@@ -1018,7 +1018,12 @@ export class GlobalRoom extends BasicRoom {
 }
 
 export class BasicChatRoom extends BasicRoom {
-	log: Roomlog;
+	readonly log: Roomlog;
+	readonly autojoin: boolean;
+	readonly staffAutojoin: string | boolean;
+	readonly banwords: string[];
+	readonly uptime: number | null;
```
Zarel: I'm surprised this isn't modified. This should probably be renamed to like `creationTime` or something.

server/rooms.ts
```diff
@@ -1413,22 +1410,22 @@ export class ChatRoom extends BasicChatRoom {
...
-	p1: AnyObject | null;
-	p2: AnyObject | null;
-	p3: AnyObject | null;
-	p4: AnyObject | null;
+	readonly type: 'battle';
+	readonly format: string;
+	readonly p1: AnyObject | null;
+	readonly p2: AnyObject | null;
+	readonly p3: AnyObject | null;
+	readonly p4: AnyObject | null;
```
Zarel: Modified by core chat commands.

server/users.ts
```diff
@@ -418,7 +419,7 @@ const SETTINGS = [
 
 // User
 export class User extends Chat.MessageContext {
```
Zarel: Wait, since when did this extend MessageContext? I'm going to need to investigate this.

Zarel:
This is actually mostly good besides the whole list of things in Rooms.

server/rooms.ts
TheJetOU: Why are you surprised that it isn't modified? It's used in groupchats (for `/groupchatuptime`) and is just `Date.now()` (or `null` in the case that the room isn't a groupchat). I see no reason anyone should be mutating it.

server/users.ts
TheJetOU: `git blame` reveals that it was added here: https://github.com/Zarel/Pokemon-Showdown/commit/eb67a12ad6c0f4021667468feedb3099ea255f1c#diff-68bb721f039d068f9977098b9f6c2ac2R437. I can't really tell _why_ it was needed but maybe @bumbadadabum can tell us.

server/rooms.ts
Zarel: `now` isn't the uptime, `now - creationTime` is the uptime. The main thing that surprises me is that it's named `uptime` when it very clearly isn't the uptime but the creation time.

server/users.ts
Zarel: Oh, for translations, that makes sense.

PR #5903, opened 2019-10-20
Server: Reorganize TypeScript globals

TheJetOU:
This commit keeps modules in `server/global.d.ts` and moves types to `globals.ts` (a la `sim/`'s structure).


Zarel:
Why not merge the two completely? I don't see a strong reason for them to be separate.

Zarel:
(I think the original reason they were separate was because of a limitation with TypeScripted jsdoc-`.js` files which no longer exist.)

TheJetOU:
> Why not merge the two completely? I don't see a strong reason for them to be separate.

I think a good reason to seperate them is to avoid dumping all the global types, interfaces, namespaces, into the `declare global { ... }` block which locks a bit cluttered IMO.

Zarel:
I don't think we'd necessarily need to?

Zarel:
What lines are we drawing, and can we rename these files to match?

TheJetOU:
> What lines are we drawing, and can we rename these files to match?

I'm not sure what you mean by that. Why would we rename the files? ``globals.ts`` matches the file's use, so does `global.d.ts`.

Zarel:
It looks like it should be `global-variables.d.ts` and `global-types.ts`?

TheJetOU:
> It looks like it should be `global-variables.d.ts` and `global-types.ts`?

That's fine by me. +1.

I suppose `sim/` should also follow suit.

scheibo:
Oh man, no more having to remember when autocompleting which one requires an `'s'` or not. Lovely!

Zarel:
Looks good! Thanks!

PR #5216, opened 2019-02-27
Fix spread move effect order

MacChaeger:
TODO: 

- [x] fix the 74 failing tests
- [x] remove the old `moveHit` function and replace it with `spreadMoveHit` (probably should rename `spreadMoveHit` to `moveHit`)
- [x] rewrite `tryMoveHit` to use new functions
- [x] add tests for things
- [x] old gens...

PR #5904, opened 2019-10-20
Add a single event for a move's own immunity to a specific target

urkerab:
Captivate, Dream Eater, Leech Seed and Synchronoise have a target-specific immunity calculation. These moves used to use the `TryHit` event, as this used to run once per target. However, this does not properly recreate cart mechanics, as the check should happen before the accuracy check. Additionally, the spread move effect order refactor (#5216) changed the mechanics of the `TryHit` event so that it only runs once (for the first target) thus breaking Captivate and Synchronoise.

This new `TryImmunity` (or I guess you want it just called `Immunity` right?) event allows the move to veto each target independently by returning `false` for immune targets. This also avoids having to specifically add the `-immune` message in each of those moves.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -2102,11 +2102,8 @@ let BattleMovedex = {
 		pp: 20,
 		priority: 0,
 		flags: {protect: 1, reflectable: 1, mirror: 1},
-		onTryHit(pokemon, source) {
-			if ((pokemon.gender === 'M' && source.gender === 'F') || (pokemon.gender === 'F' && source.gender === 'M')) {
-				return;
-			}
-			return false;
+		onTryImmunity(pokemon, source) {
+			return (pokemon.gender === 'M' && source.gender === 'F') || (pokemon.gender === 'F' && source.gender === 'M');
 		},
 		boosts: {
 			spa: -2,
@@ -4015,11 +4012,8 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {protect: 1, mirror: 1, heal: 1},
 		drain: [1, 2],
-		onTryHit(target) {
-			if (target.status !== 'slp' && !target.hasAbility('comatose')) {
-				this.add('-immune', target);
-				return null;
-			}
+		onTryImmunity(target) {
+			return target.status === 'slp' || target.hasAbility('comatose');
 		},
 		secondary: null,
 		target: "normal",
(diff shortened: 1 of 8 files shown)


Zarel:
Is this at all specific to Invulnerability?

data/moves.js
```diff
@@ -17190,10 +17181,7 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {protect: 1, mirror: 1, mystery: 1},
 		onTryHit(target) {
```
Marty-D: `onTryImmunity`?

I think you missed Trick too.

urkerab: Whoops, I had it in my testing tree, just forgot to transfer it to my clean pull. Sorry about that. But hey, that reminded me to add Endeavo(u)r, which is also supposed to check the target's HP before accuracy.

urkerab:
> Is this at all specific to Invulnerability?

It's just another form of immunity, it's not related to two-turn move semi-invulnerability.

data/scripts.js
```diff
@@ -291,19 +291,22 @@ let BattleScripts = {
 			// 3. check for powder immunity (gen 6+ only)
 			this.hitStepPowderImmunity,
 
-			// 4. check for prankster immunity (gen 6+ only)
+			// 4. check for move-specific immunity
+			this.hitStepTryImmunityEvent,
+
+			// 5. check for prankster immunity (gen 7+ only)
 			this.hitStepPranksterImmunity,
```
Zarel: Is th4ere any reason for these not to be the same hitstep?

urkerab:
"This branch has conflicts that must be resolved"

Actually an auto-merge...

data/scripts.js
Zarel: poke

urkerab: If you moved the move eligibility checks inside the loop, then you could combine the powder, move-specific and prankster immunities into one function, yes.

Zarel: Like it wouldn't affect turn order for these to be the same hitstep, right?

urkerab: I don't know what the turn order is for spread move targets on cart, but there would be a minor subtle difference in that if Prankster Whimsicott used Cotton Spore against both a Dark type and a Grass type, then under the previous version the Grass type would be immune first (because Powder immunity is checked first) while under the revised version the Dark type would be immune first (because all of its immunities are checked at once).

Zarel: hrm

PR #5905, opened 2019-10-20
Normalize messages before detecting filter evasion (quick fix)

Kaiepi:
Refer to the code's comment for what this means.

This hasn't been tested.

### DIFF
--- a/server/chat-plugins/chat-monitor.js
+++ b/server/chat-plugins/chat-monitor.js
@@ -139,6 +139,12 @@ Chat.registerMonitor('evasion', {
 	punishment: 'EVASION',
 	label: 'Filter Evasion Detection',
 	monitor(line, room, user, message, lcMessage, isStaff) {
+		// Many codepoints used in filter evasion detection can be decomposed
+		// into multiple codepoints that are canonically equivalent to the
+		// original. Perform a canonical composition on the message to detect
+		// when people attempt to evade by abusing this behaviour of Unicode.
+		lcMessage = lcMessage.normalize('NFKC');
+
 		let [regex, word, reason] = line;
 		// Remove spaces and obvious false positives
 		lcMessage = lcMessage.replace(/[\s-_,.]/g, '');


Zarel:
Which one is NFC again?

Zarel:
That's composed, right? Wouldn't decomposed be easier? Also shouldn't we be using the destructive normalization, considering this is a filter?

Zarel:
Yeah, it should be NFKD, right?

Zarel:
Also shouldn't NFKD be done to `lcMessage` before passing it to the monitor function? And the regular `message` probably shouldn't be messed with at all.

Kaiepi:
This isn't really intended to be permanent because of how much work filters need to handle Unicode properly; this is more of a quick fix for one way people can evade at the moment. NFKD was what I originally planned on using for filters, but for this NFC makes more sense because evasion detection only works with graphemes in their NFC form for now.

Zarel:
Why not NFKC then?

Kaiepi:
NFKC's used when compatibility is more important than actually being equivalent to the original string. It could break the filter.

Kaiepi:
I was wrong in the last comment. NFKC would make more sense with the code that exists now because it would split graphemes representing more than one character into its separate characters, which NFC wouldn't do, so the filter would be able to detect graphemes like `ﬁ` being used to evade.

Zarel:
I also still think we should NFKD and update the filters to match. That sounds in-scope for this PR, actually.

Kaiepi:
I would if I could get to work on it straight away, but it's getting kind of late here and I won't have much time to deal with PS on weekdays until after I complete my grant.

Zarel:
I'll take it.

server/chat-plugins/chat-monitor.js
```diff
@@ -139,6 +139,13 @@ Chat.registerMonitor('evasion', {
 	punishment: 'EVASION',
 	label: 'Filter Evasion Detection',
 	monitor(line, room, user, message, lcMessage, isStaff) {
+		// Many codepoints used in filter evasion detection can be decomposed
+		// into multiple codepoints that are canonically equivalent to the
+		// original. Perform a canonical composition on the message to detect
+		// when people attempt to evade by abusing this behaviour of Unicode.
+		message = message.normalize('NFKC');
```
Zarel:
```suggestion
```
I would at least remove this one; it shouldn't be used for filtering in the first place.

bumbadadabum: Yeah this doesnt seem necessary. Message in this function is only used to check if the exact word was used, as opposed to evading using something stripped in a message. I think this only harms that behavior.

Kaiepi:
Updated

Slayer95:
Does the main server's build of Node.js support ``String#normalize()`` at all?

https://nodejs.org/api/intl.html#intl_options_for_building_node_js

Zarel:
```
>> "ﬁ".normalize("NFKC")
<< "fi"
```

Looks like yes, but that's definitely a good question to have asked earlier.

PR #5906, opened 2019-10-23
Implement /sharedbattles for finding shared battles

whalemer:


Zarel:
Weird name for it. We can't think of anything better?

Zarel:
Also the naming definitely doesn't suggest that the relationship is directional. Does it need to be?

panpawn:
Why not just ``/sharedbattles``?

server/chat-plugins/info.js
```diff
@@ -306,6 +306,33 @@ const commands = {
...
 	},
 
+	findsharedbattles(target, room) {
+		if (!this.can('lock')) return false;
+
+		const [targetUsername1, targetUsername2] = target.split(',');
+		const user1 = Users.get(targetUsername1);
+		const user2 = Users.get(targetUsername2);
+		if (!user1) return this.errorReply(`User ${targetUsername1} not found.`);
+		if (!user2) return this.errorReply(`User ${targetUsername2} not found.`);
```
Slayer95: This condition shouldn't be required to account for the case when the user has expired (either in the exhaustive version suggested in the comment below, or in the only version if said comment is dismissed).

Slayer95: I can't figure out whether tickets outlive the submitter. In that case, ``user1`` shouldn't be checked either.

whalemer:
I've changed the intersection check to both users (have auth or are in the room), as I feel like this covers most situations well enough.

Zarel:
_Is_ it a very expensive loop? Generally an O(n) loop over all rooms is fine.

Zarel:
I just took a look. Am I missing something? It looks O(n) to me.

whalemer:
>Is it a very expensive loop? Generally an O(n) loop over all rooms is fine.

The previous version, checking if a battle log contained someone could get pretty expensive (tens of rooms by hundreds of chat lines). This one is just O(n), which I think is fine (it's basically the same loop from `/whois`'s "recent games").

whalemer:
Regarding what Astrid said, the usecase I envisioned for this command was just as a QoL improvement over having to `/whois` two users and manually search for common battles when handling reports. A more exhaustive search of battle logs might be useful, but I think it's outside the scope of this.

Zarel:
I think we probably would want to just track battle history on users.

Zarel:
That would allow for the more exhaustive search.

Asheviere:
I think a battle history is a good idea but not very feasible until we have proper serverside database support and a users table. I think the way the PR has it now covers the intentional usecase well enough for now.

PR #5910, opened 2019-10-25
Add information about permalocks in help tickets

Lusamine:


server/chat-plugins/helptickets.js
```diff
@@ -685,6 +685,7 @@ const pages = {
 					buf += `<p><Button>other</Button></p>`;
 					break;
 				case 'permalock':
+					buf += `<p>Permalocks are usually for repeated incidents of poor behavior over an extended period of time, and rarely for a single severe infraction.</p>`;
```
HoeenCoder:
```suggestion
					buf += `<p>Permalocks are usually for repeated incidents of poor behavior over an extended period of time, and rarely for a single severe infraction. Please keep this in mind when appealing a permalock.</p>`;
```

PR #5912, opened 2019-10-27
Fix crash from RoomGame refactor

sparkychildcharlie:
RoomGame's constructor currently sets the room's game to whatever roomgame is being created.

This sets ``self`` as the parent game, lets the child game overwrite ``room.game`` and sets it back to the parent game after creating a hunt.

### DIFF
--- a/server/chat-plugins/scavenger-games.js
+++ b/server/chat-plugins/scavenger-games.js
@@ -188,8 +188,14 @@ class ScavGame extends Rooms.RoomGame {
 	createHunt(room, staffHost, host, gameType, questions) {
 		if (this.childGame) return staffHost.sendTo(this.room, "There is already a scavenger hunt in progress.");
 		this.onStartEvent();
-		this.childGame = new Rooms.ScavengerHunt(room, staffHost, host, gameType, questions, this);
+		this.startHunt(room, staffHost, host, gameType, questions, this);
 		return true;
+	}
+
+	startHunt(...args) {
+		let self = this;
+		self.childGame = new Rooms.ScavengerHunt(...args);
+		this.room.game = self;
 	}
 
 	announce(msg) {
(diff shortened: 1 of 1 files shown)


sparkychildcharlie:
@Zarel I hope this way of fixing is acceptable.  I've explained on discord why I need both things to be a roomgame (to track renames).  Having them both in the same room because one of the 2 serves as a "control device" and not an actual game.

Zarel:
I've explained on Discord that you can't just create a "free-floating" roomgame. All our code expects the roomgame to be `room.game`.

https://github.com/Zarel/Pokemon-Showdown/blob/0c1e1087397187a6af8a3cd6e10a75f8624189db/server/users.ts#L765

For instance, this crash, which I've had to suppress, is nearly definitely caused by your code.

https://github.com/Zarel/Pokemon-Showdown/blob/0c1e1087397187a6af8a3cd6e10a75f8624189db/server/users.ts#L1004

As is this.

server/chat-plugins/scavenger-games.js
```diff
@@ -188,10 +188,16 @@ class ScavGame extends Rooms.RoomGame {
...
 		this.onStartEvent();
-		this.childGame = new Rooms.ScavengerHunt(room, staffHost, host, gameType, questions, this);
+		this.startHunt(room, staffHost, host, gameType, questions, this);
 		return true;
 	}
 
+	startHunt(...args) {
+		let self = this;
+		self.childGame = new Rooms.ScavengerHunt(...args);
+		this.room.game = self;
```
xfix: I'm not sure if I like this, seems very hacky to me.

Zarel:
This is still not fixed. Should we just remove the Scavengers plugin?

KrisXV:
No. Absolutely not. It'll be fixed if this gets typescripted and uses room.getGame i presume.

Zarel:
The problem isn't that simple, or I would have fixed it myself.

Zarel:
It's not like `.js` files _can't_ use `room.getGame`

Zarel:
I think the correct interim solution is to make one of these games a `minorActivity`.

sparkychildcharlie:
This is fixed in #6507 simply by making the scavenger game into a non-roomgame activity simply used to keep a list of userids of players who are still in the game

PR #5913, opened 2019-10-27
 Validate pre-Battle Bond Greninja correctly regardless of whether 'loose' Greninja-Ash is enabled though !Obtainable Formes

WeWuzNidokangz:
This is a small nitpick to the recent validation improvement of Greninja-Ash, and a step towards OMs like Megamons being possible without proprietary mods.

Currently, if we apply rules like this,

/tour rules !Obtainable Formes, -Greninja-Ash, -Zygarde-Complete (, ...)

to legalize a subset of !Obtainable Formes, pre-Battle Bond Greninja will be able to be validated with with a gender that should be illegal (F) and moves that should be exclusive to regular Greninja (e.g. Dig). This change prevents that.

* The special-case gender legality here seemed like it should be controlled by Obtainable Misc in the same way that general gender legality is.

* There should be no effect on hackmons metas, as the differences will still be covered under !Obtainable through !Obtainable Formes, !Obtainable Misc and !Obtainable Moves.

Please consider, thank-you!


sim/team-validator.ts
```diff
@@ -372,12 +372,18 @@ export class TeamValidator {
...
-		if (ability.id === 'battlebond' && template.id === 'greninja' && ruleTable.has('obtainableformes')) {
-			tierTemplate = learnsetTemplate = dex.getTemplate('greninjaash');
-			if (set.gender && set.gender !== 'M') {
-				problems.push(`Battle Bond Greninja must be male.`);
+		if (ability.id === 'battlebond' && template.id === 'greninja') {
+			let ashTemplate = dex.getTemplate('greninjaash');
+			if (ruleTable.has('obtainableformes')) {
+				tierTemplate = ashTemplate;
+			}
+			learnsetTemplate = ashTemplate;
```
Zarel: This seems like an unnecessarily complicated way to write

```ts
			learnsetTemplate = dex.getTemplate('greninjaash');
			if (ruleTable.has('obtainableformes')) {
				tierTemplate = learnsetTemplate;
			}
```

WeWuzNidokangz: I think I was considering putting the learnset assignment behind another condition like if (ruleTable.has('obtainablemoves')) but it seemed redundant, so it ended up like that. What you wrote seems like it should be slightly more performative for what it's doing now, so I updated it to that.

PR #5914, opened 2019-10-27
 Implement warning on level 50 Pokemon in non-cartridge formats

DaWoblefet:
Implements [this suggestion](https://www.smogon.com/forums/threads/warn-on-using-level-50-pokemon-in-non-vgc-battle-spot-formats.3655675/) from myself.
image

Since Pokemon in cartridge formats are temporarily set to level 100 when being validated, they will never prompt this check. It is useful for players who copy -> paste in sets from, say, VGC 2018 into Gen 7 Doubles OU and load up for a battle without fixing the level. It does not apply this check for custom game.

This technically prohibits anyone from actually using level 50 Pokemon in non-cartridge formats, but I can't think of a reason anyone would actually want to do that for. I couldn't think of a sanity check like what we use for the EV cases.

This is my first real PR for Showdown, so if I did anything wrong let me know. I was having issues with npm test failing but asheviere said someone else broke the Travis build so I used --no-verify to commit, I don't intend to do that normally. The change I made actually works locally and no new errors were introduced, so I assume it's ok.

EDIT: Marty said it was some setting in eslint getting changed that was causing my initial issues.


asgdf:
How would this interact with https://github.com/Zarel/Pokemon-Showdown/pull/5899? If the max level of a format is set to 50, and someone set their pokemon's level to 50, this would reject them, right? Is it possible to just add a check for the format's max allowed level before doing this check?

Lusamine:
I see that there's currently a PR to support [max level caps in tours](https://github.com/Zarel/Pokemon-Showdown/pull/5899).  One use I could think of would be if WiFi wanted to hold a tour capped at level 50 to simulate playing OU on a retail device.  Would this interfere with people setting mons in their team to level 50 to participate in this kind of tournament? If the other PR goes through, would it be possible to check for the max allowed level before straight up banning all level 50 mons entirely?

asgdf:
If the other PR goes through, the custom max level would be in ``this.ruleTable.maxLevel``. Though that might still change since the PR's still open.

DaWoblefet:
I'll wait for Akainse's PR to go through before making any further changes then, since ``this.ruleTable.maxLevel`` won't exist until after it goes through.

Zarel:
A simple sanity cap would be to do the same thing: Ask the user to add `1` to one of their EVs.

Slayer95:
Client-side teambuilder suggestions pls.

Slayer95:
Changed title to mirror what the PR currently does.

Zarel:
Yeah, I'd strongly prefer client-side suggestions.

Maybe we can compromise with a flag of "ignore suggestions" somewhere.

Zarel:
We could even do it without much change to the current system: Just have the checkbox add `1` to the EVs.

DaWoblefet:
@TheMezStrikes said they don't anticipate making significant process on their PR and would evaluate any tournament-related implications of this PR in what they're doing later. I followed Zarel's advice on the server-side EV sanity check. No client side options, but that's beyond the scope of what I wanted to do.
image
After an EV change:
image

DaWoblefet:
@Zarel look fine?

Zarel:
Looks good.

Zarel:
Thanks!

Zarel:
Wait wait this is the wrong place for the check, let me move it to the correct place.

DaWoblefet:
I wanted it not to throw in custom game, but so long as it doesn't do that I don't care where the check is placed.

Zarel:
Oh, I see how this works; no format has a `maxLevel` of 50, only a `maxForcedLevel`, so `set.level === 50` only happens if `maxForcedLevel` isn't 50 (because otherwise it would be set to 100).

That's subtle enough to deserve a comment, although I'll just add a manual check for `maxLevel` just to make the code easier to read (and just in case one day it actually happens).

PR #5923, opened 2019-10-31
Allow /blockpms to block locked and semilocked users

Lusamine:
Per global staff discussion, allow gstaff to opt into blocking PMs from locked and semilocked users due to harassment issues. They're directed to a help ticket if they truly need help.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -622,6 +622,7 @@ export class User extends Chat.MessageContext {
 	}
 	authAtLeast(minAuth: string, room: BasicChatRoom | null = null) {
 		if (!minAuth || minAuth === ' ') return true;
+		if (minAuth === 'unlocked') return !(this.locked || this.semilocked);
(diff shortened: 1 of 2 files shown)


server/users.ts
```diff
@@ -622,6 +622,7 @@ export class User extends Chat.MessageContext {
 	}
 	authAtLeast(minAuth: string, room: BasicChatRoom | null = null) {
 		if (!minAuth || minAuth === ' ') return true;
+		if (minAuth === 'unlocked' && !this.locked && !this.semilocked) return true;
```
panpawn: I think this should be
```suggestion
		if (minAuth === 'unlocked' && (!this.locked || !this.semilocked)) return true;
```
right?

Zarel:
I've definitely been mulling "locked users can't PM staff" for a while now. I suppose this is a decent compromise.

PR #5927, opened 2019-11-02
Fix Blast Burn wording for consistency

pokemonvortex:
Blast Burn's desc changed from "cannot make a move" to "cannot select a move" to match with all other recharge moves


Marty-D:
Guess I missed one, thanks!

PR #5929, opened 2019-11-03
Use Chat.html`` a bit more

urkerab:


server/chat-commands.js
```diff
@@ -798,7 +798,7 @@ const commands = {
 
 		Ladders.visualizeAll(target).then(values => {
 			let buffer = `<div class="ladder"><table>`;
-			buffer += `<tr><td colspan="8">User: <strong>${Chat.escapeHTML(target)}</strong></td></tr>`;
+			buffer += Chat.html`<tr><td colspan="8">User: <strong>${target}</strong></td></tr>`;
```
TheJetOU: Won't this cause real HTML to get escaped?

Zarel: There's no real HTML in here...

server/chat-commands.js
```diff
@@ -828,26 +828,26 @@ const commands = {
 		if (!id) return this.parse('/help makechatroom');
 		if (id.length > MAX_CHATROOM_ID_LENGTH) return this.errorReply("The given room title is too long.");
 		// Check if the name already exists as a room or alias
-		if (Rooms.search(id)) return this.errorReply(`The room '${target}' already exists.`);
-		if (!Rooms.global.addChatRoom(target)) return this.errorReply(`An error occurred while trying to create the room '${target}'.`);
+		if (Rooms.search(id)) return this.errorReply(Chat.html`The room '${target}' already exists.`);
```
Zarel: `errorReply` does not take HTML; this will lead to double-escaping.

urkerab: Ah, that goes for `sendReply` as well I take it. Thanks, that had confused me...

Zarel: `sendReply` sends a protocol message. If you do something like ``sendReply(`|html|...`)``, that will need escaping.

server/chat-plugins/mafia.js
```diff
@@ -1044,7 +1044,7 @@ class MafiaTracker extends Rooms.RoomGame {
 			if (this.subs.includes(targetUser.id)) this.subs.splice(this.subs.indexOf(targetUser.id), 1);
 			this.played.push(targetUser.id);
 			this.playerTable[targetUser.id] = player;
-			this.sendRoom(`${Chat.escapeHTML(targetUser.name)} has been added to the game by ${Chat.escapeHTML(user.name)}!`, {declare: true});
+			this.sendRoom(Chat.html`${targetUser.name} has been added to the game by ${user.name}!`, {declare: true});
```
Zarel: `{declare: boolean}` as an option? Ick.

server/chat-plugins/scavengers.js
```diff
@@ -193,7 +193,7 @@ function formatQueue(queue = [], viewer, room, broadcasting) {
 			const queuedBy = item.hosts.every(h => h.id !== item.staffHostId) ? ` / ${item.staffHostId}` : '';
 			let questions;
 			if (!broadcasting && (item.hosts.some(h => h.id === viewer.id) || viewer.id === item.staffHostId)) {
-				questions = item.questions.map((q, i) => i % 2 ? `<span style="color: green"><em>[${Chat.escapeHTML(q.join(' / '))}]</em></span><br />` : Chat.escapeHTML(q)).join(" ");
+				questions = item.questions.map((q, i) => i % 2 ? Chat.html`<span style="color: green"><em>[${q.join(' / ')}]</em></span><br />` : Chat.escapeHTML(q)).join(" ");
```
Zarel:
```suggestion
				questions = item.questions.map((q, i) => i % 2 ? (
					Chat.html`<span style="color: green"><em>[${q.join(' / ')}]</em></span><br />`
				) : (
					Chat.escapeHTML(q)
				)).join(" ");
```

This is pretty unreadable without this.

server/chat-plugins/scavengers.js
```diff
@@ -1034,7 +1034,7 @@ let commands = {
 				if (!players.length) {
 					str += `<tr><td>${questionNum}</td><td>None</td>`;
 				} else {
-					str += `<tr><td>${questionNum}</td><td>${players.map(pl => pl.lastGuess > Date.now() - 1000 * 300 ? `<strong>${Chat.escapeHTML(pl.name)}</strong>` : Chat.escapeHTML(pl.name)).join(", ")}`;
+					str += `<tr><td>${questionNum}</td><td>${players.map(pl => pl.lastGuess > Date.now() - 1000 * 300 ? Chat.html`<strong>${pl.name}</strong>` : Chat.escapeHTML(pl.name)).join(", ")}`;
```
Zarel:
```suggestion
					str += `<tr><td>${questionNum}</td><td>`;
					str += players.map(pl => pl.lastGuess > Date.now() - 1000 * 300 ?
						Chat.html`<strong>${pl.name}</strong>`
					:
						Chat.escapeHTML(pl.name)
					).join(", ");
```

(readability pass)

Zarel:
Everything else looks good.

server/chat-plugins/mafia.js
Zarel: (This should be refactored into two different functions `sendRoom` and `declareRoom`, but that doesn't need to be in this PR.)

server/chat-commands.js
urkerab: Right, but none of the cases I touched did that, which is why I reverted all of those changes.

server/chat-plugins/scavengers.js
urkerab:
```
   198:1  error  Expected indentation of 5 tabs but found 4  indent
   198:6  error  Bad line breaking before and after ':'      operator-linebreak
```

server/chat-plugins/scavengers.js
urkerab:
```
  1044:1  error  Expected indentation of 6 tabs but found 5  indent
  1044:7  error  Bad line breaking before and after ':'      operator-linebreak
```

server/chat-plugins/mafia.js
urkerab: It actually accepts 4 boolean options, `uhtml`, `declare`, `strong`, and `timestamp` (in order of priority).

server/chat-plugins/scavengers.js
Zarel: Can you play around with it until it stops complaining? Maybe it needs parens?

Zarel:
```suggestion
				questions = item.questions.map((q, i) => (i % 2 ?
					Chat.html`<span style="color: green"><em>[${q.join(' / ')}]</em></span><br />`
				:
					Chat.escapeHTML(q)
				)).join(" ");
```

Try this?

urkerab: I looked into this and I don't think `eslint` allows a bare `:` at all; you either need to have it all on one line (if it fits), or you can (as currently configured in `.eslintrc`) only allow a newline after the `:` to break the line up, or the other setting would only allow a newline before the `:` to break the line up. (Unless you override `eslint` for this line of course.)

Zarel: Can you override `eslint` to permit it? Feel free to do that in a separate PR; I won't block this one on that.

Zarel:
Sorry, this conflicted with the `chat-commands/` refactor. Can you rebase?

Zarel:
I think you've fixed everything important, so I can merge immediately upon your rebase.

urkerab:
Ugh, that rebase was hard.

PR #5930, opened 2019-11-03
Centralise semi-invulnerability checks for Poison type using Toxic in Gen6+

urkerab:
Also optionally centralise the check for Helping Hand.


Marty-D:
Maybe I'm missing something but how are Phantom Force/Shadow Force supposed to work if you remove their `onInvulnerability` entirely?

urkerab:
No, you're not missing anything, I'd just forgotten to test properly after the Helping Hand change. Sorry about that.

PR #5934, opened 2019-11-05
Introduce /modchatnext command

scheibo:
The TDs have found that setting the default modchat on smogtours to player is undesirable because a lot more casual games happen there than they anticipated.

```
[14:19]Tony:i was wondering how possible it was for to create a /ionext type of command for modchat on smogtours
[14:20]Tony:as in, one of the two players (or both players) enters /modchaton or something in the lobby or their private messages
[14:20]Tony:and then the game starts with modchat immediately on
[14:20]Tony:and if neither player does, then modchat is off
[17:14]pre:we would need to do it in the main code, because its somewhat tricky to implement a /modchatnext as a private plugin
[17:15]pre:whats the motivation?
[17:16]pre:ionext needs to work the way it does to ensure the battle stays private, but is /modchat at the start of the battle not sufficient already? are we trying to guard against some errant 'first!'s or something
[05:19]Tony:the motivation is that we underestimated how many friendly games are played on smogtours as well
[05:20]Tony:and by changing our stance from opt-out to opt-in we still make it possible to instantly have modchat on for those who want that during their games, especially official tournament games, while also catering to those that use the server more casually and wanna mess around with their friends
[05:21]Tony: "are we trying to guard against some errant 'first!'s or something" = we are essentially. people love spamming battlechats of official games like that until a moderator sets modchat
```

I ran into a lot of questions when writing this draft PR:

1) Does it make sense to try and merge next-battle related settings in some way as opposed to having `modchatNextBattle` and `inviteOnlyNextBattle` (do we expect more in the future? does it make sense to consolidate/generalize)?
2) Do we want to support multiple levels for `modchatNextBattle` instead of just 'on' or 'off'? I was looking into going that route in this change, but it increases complexity quite a bit.
3) Completely alternative solution: some sort of default slowchat-esque period at the start of a battle which would provide players to opportunity to set up this modchat normally

More thoughts appreciated.

### DIFF
--- a/server/chat-plugins/roomsettings.js
+++ b/server/chat-plugins/roomsettings.js
@@ -211,30 +211,6 @@ exports.commands = {
 	},
 	roomsettingshelp: [`/roomsettings - Shows current room settings with buttons to change them (if you can).`],
 
-	// TODO: how to indicate this helper method is not a command?
-	targetToModchat(target, threshold = 1) {
-		if (this.meansNo(target)) return false;
-		switch (target) {
-		case 'ac':
-		case 'autoconfirmed':
-			return 'autoconfirmed';
-		case 'trusted':
-			return 'trusted';
-		case 'player':
-			target = Users.PLAYER_SYMBOL;
-			/* falls through */
-		default:
-			if (!Config.groups[target]) {
-				this.errorReply(`The rank '${target}' was unrecognized as a modchat level.`);
-				return this.parse('/help modchat');
-			}
-			if (Config.groupsranking.indexOf(target) > threshold) {
-				return this.errorReply(`/modchat - Access denied for setting higher than ${Config.groupsranking[threshold]}.`);
-			}
-			return target;
-		}
-	},
-
 	modchat(target, room, user) {
 		if (!target) {
 			const modchatSetting = (room.modchat || "OFF");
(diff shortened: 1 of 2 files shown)


Zarel:
Maybe a general setting-for-next-battle command, with some UI for it?

Zarel:
Also don't we already have specialnextbattle? Or whatever the thing is that enables SPL avatars? Let's reuse that.

scheibo:
Not interested in adding a UI, but someone is more than welcome to do that as a followup.

`specialNextBattle` is just a `boolean` and doesn't seem useful here. But yes, its another 'next battle' type of thing - I think it points to us instead creating a `nextBattle` field (or `nextBattleSettings` or similar) on `User` where invite only, special and modchat can be grouped under.

```ts
interface NextBattleSettings {
  inviteOnly: boolean;
  special: boolean;
  modchat: boolean; // or should this instead be an enum?
}
```

Zarel:
special should imply modchat, there's no need for them to be separate.

scheibo:
> special should imply modchat, there's no need for them to be separate

These seem completely orthogonal to me. Special was/is just used to opt into an SPL background, no (and needed both players to opt-in)? 

I can see something like `/officialnext` to indicate the next battle is an 'official tournmanent battle' that will opt into special background/modchat player automatically etc and which replaces/repurposes `specialNextBattle` and obviates the need for `modchatnext`? However, I'm unclear that both players should need to do it (this seems easy to forget/harder to coordinate) and that its usage should be restricted?

Zarel:
> unclear that both players should need to do it (this seems easy to forget/harder to coordinate)

If one does and one doesn't, it'll show them an error message, letting the players coordinate via PM about whether or not it should be official.

> and that its usage should be restricted

I agree, the modchat part doesn't need to be restricted, just the avatar part.

scheibo:
OK, so 'special' battles will set modchat player (this PR). Then we can have `/officialnext` (below) to mirror `/splnext` but not place stipulations on the avatar (ie. ACLs)

```js
{
  ...
  splmode: 'splnext',
  splnext: function (target, room, user) {
    if (!('' + user.avatar).startsWith('#splx')) {
      return this.errorReply("You need to be using your SPL avatar.");
    }
    user.specialNextBattle = "Smogon Premier League X";
    this.sendReply("Your next battle will be in SPL mode.");
  },
  officialmode: 'officialnext',
  officialnext: function (target, room, user) {
    user.specialNextBattle = "Official Tournament";
    this.sendReply("Your next battle will be an Official Tournament battle.");
  },
};
```

I think this works as an MVP. Additional QoL improvements which could be made would be **(1)** persisting the `specialNextBattle` bit until a battle is created (currently if you said `specialNextBattle` but your opponent doesn't, the bit gets cleared because otherwise theres no way to turn it off in case you and your opponent negotiate for it not to be an official tournament battle) and **(2)** doing some sort of reconciliation with the SPL `specialNextBattle` setting (ie. SPL battles are Official tournament battles, so should SPL vs. Official just degrade to Official?) I'm punting on both of these.

This PR fixes a bug where `specialNextBattle` was never cleared (which sees us trade in the `some(...)` and `break` partial loops for a single full loop) and updates the type signature for `specialNextBattle` as its not a `boolean`.

PR #5935, opened 2019-11-05
Upgrade to TypeScript 3.7

TheJetOU:
https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/

Unfortunately, we will not be able to use the exciting new features
such as optional chaining and nullish coalescing until Node.js
supports them natively as we use a third-party compiler (sucrase)
rather than tsc to ensure fast build times.

### DIFF
--- a/data/random-teams.js
+++ b/data/random-teams.js
@@ -1,5 +1,7 @@
 'use strict';
 
+/** @typedef {import('../sim/prng').PRNG} PRNGType */
+
 /** @type {typeof import('../sim/dex').Dex} */
 const Dex = require(/** @type {any} */ ('../.sim-dist/dex')).Dex;
 /** @type {typeof import('../sim/prng').PRNG} */
@@ -30,7 +32,7 @@ const randomFactorySets = require('./factory-sets.json');
 class RandomTeams {
 	/**
 	 * @param {Format | string} format
-	 * @param {?PRNG | [number, number, number, number]} [prng]
+	 * @param {?PRNGType | [number, number, number, number]} [prng]
 	 */
 	constructor(format, prng) {
 		format = Dex.getFormat(format);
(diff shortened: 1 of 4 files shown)


data/random-teams.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+/** @typedef {import('../sim/prng').PRNG} PRNGType */
```
Zarel: Can we just typedef this `PRNG`?

Zarel:
Is that necessary? I just meant `/** @typedef {import('../sim/prng').PRNG} PRNG */`

TheJetOU:
> Is that necessary?

No, that was just what I understood from your review.

data/random-teams.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+/** @typedef {import('../sim/prng').PRNG} PRNG */
+
 /** @type {typeof import('../sim/dex').Dex} */
 const Dex = require(/** @type {any} */ ('../.sim-dist/dex')).Dex;
 /** @type {typeof import('../sim/prng').PRNG} */
```
urkerab: Worth simplifying this type annotation at some point, given the typedef?

Zarel: Sadly, no, this can't be simplified.

Zarel: Maybe we could change to `.ts`, though.

TheJetOU: I tried that but TypeScript complained because you can't assign a type to a value.

Zarel: Oh, wait, I see what you're saying. Yes, this can be `/** @type {typeof PRNG} */`

Zarel: It's such a minor optimization, though.

TheJetOU: You can't do `/** @type {typeof PRNG} /*` because the variable itself is called `PRNG`.

So you need to rename to something like `PRNGType` (which is why it was called that initially that), but you'll arrive to this error:

```
data/random-teams.js:7:19 - error TS2693: 'PRNGType' only refers to a type, but is being used as a value here.
```

Zarel: Oh, yeah, that makes sense too. Oh well, it works as-is.

urkerab: Thanks for explaining.

PR #5938, opened 2019-11-08
Fix handling of locked and proxy connections

Lusamine:
A few changes to the popup for users on a proxy:
1) Link a user on a proxy to information on disabling it within the help ticket system.
2) Handles the case where a user turns off the proxy but doesn't log out by telling them they need to log out fully.  These people frequently end up in help tickets.  Ideally, I'd like to have a "Log out" button here but I can't get it to do anything right now.

Adds the case in help ticket appeals where a user on a "[proxy]" IP that is locked as another user is directed to how to turn off their proxy rather than to IP-appeals, where they would normally get the IP markshared or unlocked. Proxy IPs should remain locked by default, even if another user has a punishment on it.

### DIFF
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -1122,10 +1122,10 @@ export const Punishments = new class {
 			user.resetName();
 			user.updateIdentity();
 		} else {
-			if (punishUserid === '#hostfilter') {
-				user.popup(`Due to spam, you can't chat using a proxy. (Your IP ${user.latestIp} appears to be a proxy.)`);
-			} else if (punishUserid === '#ipban') {
-				user.popup(`Your IP (${user.latestIp}) is not allowed to chat on PS, because it has been used to spam, hack, or otherwise attack our server. Make sure you are not using any proxies to connect to PS.`);
+			if (punishUserid === '#hostfilter' || punishUserid === '#ipban') {
+				user.send(`|popup||html|Your IP (${user.latestIp}) is currently locked due to being a proxy. We automatically lock these connections since they are used to spam, hack, or otherwise attack our server. Disable any proxies you are using to connect to PS.\n\n<a href="view-help-request--appeal"><button class="button">Help me with a lock from a proxy</button></a>`);
+			} else if (user.latestHostType === 'proxy' && user.locked !== user.id) {
+				user.send(`|popup||html|You are locked${bannedUnder} on the IP (${user.latestIp}), which is a proxy. We automatically lock these connections since they are used to spam, hack, or otherwise attack our server. Disable any proxies you are using to connect to PS.\n\n<a href="view-help-request--appeal"><button class="button">Help me with a lock from a proxy</button></a>`);
 			} else if (!user.lockNotified) {
 				user.send(`|popup||html|You are locked${bannedUnder}. ${user.permalocked ? `This lock is permanent.` : `Your lock will expire in a few days.`}${reason}${appeal}`);
 			}
@@ -1158,7 +1158,6 @@ export const Punishments = new class {
 
 		return IPTools.lookup(ip).then(({dnsbl, host, hostType}) => {
 			user = connection.user || user;
-			if (user.locked === '#hostfilter') user.locked = null;
 
 			if (hostType === 'proxy' && !user.trusted && !user.locked) {
 				user.locked = '#hostfilter';
--- a/server/users.ts
+++ b/server/users.ts
@@ -1040,12 +1040,23 @@ export class User extends Chat.MessageContext {
 			Rooms.get(roomid)!.onLeave(oldUser);
 		}
 
-		if (this.locked === '#dnsbl' && !oldUser.locked) this.locked = null;
-		if (!this.locked && oldUser.locked === '#dnsbl') oldUser.locked = null;
-		if (oldUser.locked) this.locked = oldUser.locked;
+		const oldLocked = this.locked;
+		const oldSemilocked = this.semilocked;
+
+		if (!oldUser.semilocked) this.semilocked = null;
+
+		// If either user is unlocked and neither is locked by name, remove the lock.
+		// Otherwise, keep any locks that were by name.
+		if ((!oldUser.locked || !this.locked) && oldUser.locked !== oldUser.id && this.locked !== this.id) {
+			this.locked = null;
+		} else if (this.locked !== this.id) {
+			this.locked = oldUser.locked;
+		}
 		if (oldUser.autoconfirmed) this.autoconfirmed = oldUser.autoconfirmed;
 
 		this.updateGroup(this.registered);
+		if (oldLocked !== this.locked || oldSemilocked !== this.semilocked) this.updateIdentity();
+
 		// We only propagate the 'busy' statusType through merging - merging is
 		// active enough that the user should no longer be in the 'idle' state.
 		// Doing this before merging connections ensures the updateuser message
--- a/server/chat-plugins/helptickets.js
+++ b/server/chat-plugins/helptickets.js
@@ -671,10 +671,10 @@ const pages = {
 								buf += `<p><Button>lock</Button></p>`;
 							}
 						}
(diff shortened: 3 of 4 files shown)


server/punishments.ts
```diff
@@ -1122,10 +1122,13 @@ export const Punishments = new class {
...
 			user.updateIdentity();
 		} else {
-			if (punishUserid === '#hostfilter') {
-				user.popup(`Due to spam, you can't chat using a proxy. (Your IP ${user.latestIp} appears to be a proxy.)`);
-			} else if (punishUserid === '#ipban') {
-				user.popup(`Your IP (${user.latestIp}) is not allowed to chat on PS, because it has been used to spam, hack, or otherwise attack our server. Make sure you are not using any proxies to connect to PS.`);
+			if (punishUserid === '#hostfilter' || punishUserid === '#ipban') {
+				if (user.latestHostType !== 'proxy') {
+					// Disabled their proxy but haven't logged out.
+					user.send(`|popup||html|You are currently locked due to being on a proxy. If you have disabled it, log out fully by typing <code>/logout</code> in any chatroom.`);
```
panpawn: I think you could put the logout button in this popup, if that'd be more helpful?

Lusamine: I tried [these](https://hastebin.com/dijeziviro.cs) and it seems a client change is needed.  I see that [this](https://github.com/smogon/Pokemon-Showdown-Client/commit/52973f240affcbf49bcb1d296f1511f96809d4fc) exists for Preact.  If you can get it to work, feel free to suggest how.

Zarel:
We should instead just fix it by making it so you don't need to log out to fix the proxy issue.

In fact, you already _shouldn't_ need to.

panpawn:
Could check ``onConnection`` or something, right?

Zarel:
Err, `onConnection` is for rooms/etc. It should already be implemented directly in `user.merge`.

Lusamine:
I believe this fixes the bug in ``user.merge`` that fails to remove the ``#hostfilter`` lock.  I updated the help ticket information and popup information accordingly.

server/users.ts
```diff
@@ -1040,12 +1040,19 @@ export class User extends Chat.MessageContext {
...
 		}
 
-		if (this.locked === '#dnsbl' && !oldUser.locked) this.locked = null;
-		if (!this.locked && oldUser.locked === '#dnsbl') oldUser.locked = null;
+		const startLock = this.locked;
+		const startSemilock = this.semilocked;
+
+		if (this.semilocked === '#dnsbl' && !oldUser.semilocked) this.semilocked = null;
+		if (!this.semilocked && oldUser.semilocked === '#dnsbl') oldUser.semilocked = null;
+		if ((this.locked === '#hostfilter' || this.locked === '#ipban') && !oldUser.locked) this.locked = null;
```
Zarel: Maybe just check for the first character to be `#`.

server/users.ts
```diff
@@ -1040,12 +1040,19 @@ export class User extends Chat.MessageContext {
 			Rooms.get(roomid)!.onLeave(oldUser);
 		}
 
-		if (this.locked === '#dnsbl' && !oldUser.locked) this.locked = null;
-		if (!this.locked && oldUser.locked === '#dnsbl') oldUser.locked = null;
+		const startLock = this.locked;
```
Zarel: These should be `oldLocked` and `oldSemilocked` to be consistent with the rest of our code.

server/users.ts
```diff
@@ -1040,12 +1040,19 @@ export class User extends Chat.MessageContext {
...
+		const startSemilock = this.semilocked;
+
+		if (this.semilocked === '#dnsbl' && !oldUser.semilocked) this.semilocked = null;
```
Zarel: I wouldn't propagate any semilock, actually.

```suggestion
		if (!oldUser.semilocked) this.semilocked = null;
```

server/users.ts
```diff
@@ -1040,12 +1040,19 @@ export class User extends Chat.MessageContext {
...
+		if (!this.semilocked && oldUser.semilocked === '#dnsbl') oldUser.semilocked = null;
```
Zarel: After reflection, I wouldn't bother updating `oldUser`.

```suggestion
```

server/users.ts
Zarel: Actually, maybe we should just forgive all locks that aren't username locks? We've definitely had problems with people in schools being perpetually locked. Is there a way to tell if a lock comes from an IP or a username?

Lusamine: I'm somewhat wary of special locks starting with ``#`` and whether they should be treated the same way.  I know ``#proxy`` and ``#range`` are other ones.

Lusamine: > Actually, maybe we should just forgive all locks that aren't username locks? We've definitely had problems with people in schools being perpetually locked. Is there a way to tell if a lock comes from an IP or a username?

I think this would be too forgiving.  We already markshared schools, but if we forgave everyone who is locked on the same IP as an alt, we'd be letting a lot of actual alts go free.

Zarel: > I'm somewhat wary of special locks starting with ``#`` and whether they should be treated the same way.  I know ``#proxy`` and ``#range`` are other ones.  I think overall, it should be safe.

I can't think of one that shouldn't be freed, at least.

> I think this would be too forgiving. We already markshared schools, but if we forgave everyone who is locked on the same IP as an alt, we'd be letting a lot of actual alts go free.

Would we? The user would still need a new account _and_ a new IP, and if they have those, they could have evaded just by being more cautious anyway. I think a lot of schools aren't markshared, so I think it would be worth having a better experience for people on shared IPs, in exchange for slightly more evasion.

I don't have a good sense of how markshared works. Are schools locked with `#shared` or something?

Zarel:
By the way, thanks for tackling this! I've been wanting to get punishment propagation fixed for a long time.

server/users.ts
Lusamine: I get what you're talking about now.
> Is your idea that if an uninvolved name logs into a bad IP, it'll instantly forgive them if they log into a clean IP? While alts who were present during the lock are locked by name and remain locked?
I think this is fine since we already forgive random names that log into punished IPs; they simply have to log out and go onto another network, and we can remove the logout step.

server/users.ts
```diff
@@ -1040,12 +1040,23 @@ export class User extends Chat.MessageContext {
 			Rooms.get(roomid)!.onLeave(oldUser);
 		}
 
-		if (this.locked === '#dnsbl' && !oldUser.locked) this.locked = null;
-		if (!this.locked && oldUser.locked === '#dnsbl') oldUser.locked = null;
-		if (oldUser.locked) this.locked = oldUser.locked;
+		const oldLocked = this.locked;
+
+		if (this.locked && this.locked.startsWith('#') && !oldUser.locked) this.locked = null;
+		if (!this.locked && oldUser.locked && oldUser.locked.startsWith('#')) oldUser.locked = null;
```
Zarel: Wait, what happened to doing this with semilocks, too? Did I miscommunicate in a previous review?

Lusamine: Oops, I misread the line that was being deleted.  Fixing now.

server/punishments.ts
```diff
@@ -1158,11 +1158,10 @@ export const Punishments = new class {
 
 		return IPTools.lookup(ip).then(({dnsbl, host, hostType}) => {
 			user = connection.user || user;
-			if (user.locked === '#hostfilter') user.locked = null;
 
-			if (hostType === 'proxy' && !user.trusted && !user.locked) {
+			if (hostType === 'proxy' && !user.locked) {
```
Zarel: Why was `!user.trusted` removed?

Lusamine: The only place I see ``Punishments.checkIp`` called is in ``socketConnect`` and the user can't be trusted at this point since it is before ``handleRename``.

Lusamine: I put the sanity checks back in in case this function is used differently in the future.

PR #5939, opened 2019-11-08
Make config-loader return type `Config`

TheJetOU:


Zarel:
What is `Config.routes`?

Zarel:
Oh, that. Why is it in config-example if changing it is pointless?

server/config-loader.ts
```diff
@@ -9,13 +9,13 @@ import * as defaults from '../config/config-example';
...
-export function load(invalidate: boolean = false) {
+export function load(invalidate = false) {
 	if (invalidate) delete require.cache[CONFIG_PATH];
-	const config = Object.assign({}, defaults, require('../config/config'));
+	const config: Config = Object.assign({}, defaults, require('../config/config'));
 	// config.routes is nested - we need to ensure values are set for its keys as well.
 	const routes = config.routes;
 	for (const [key, value] of Object.entries(defaults.routes)) {
-		if (!(key in routes)) routes[key] = value;
+		if (!(key in routes)) routes[key as keyof Config['routes']] = value;
```
Zarel: Does `keyof Config.routes` not work?

Zarel: In theory, an `Object.assign` here might be easier than the for-loop. Is there any reason why you didn't use `Object.assign`?

TheJetOU: > Does keyof Config.routes not work?

No, because `Config` is a type and not a namespace.

> In theory, an Object.assign here might be easier than the for-loop. Is there any reason why you didn't use Object.assign?

I don't know. I'm not the original code author. I'm just trying to make it return `Config` and not `any`.

server/config-loader.ts
```diff
@@ -9,13 +9,13 @@ import * as defaults from '../config/config-example';
 
 const CONFIG_PATH = require.resolve('../config/config');
 
-export function load(invalidate: boolean = false) {
+export function load(invalidate = false) {
 	if (invalidate) delete require.cache[CONFIG_PATH];
-	const config = Object.assign({}, defaults, require('../config/config'));
+	const config: Config = Object.assign({}, defaults, require('../config/config'));
 	// config.routes is nested - we need to ensure values are set for its keys as well.
 	const routes = config.routes;
```
Zarel:
```suggestion
	config.routes = Object.assign({}, defaults.routes, config.routes);
```

Zarel: In that case, let's just do this.

Zarel:
That Travis error is very mysterious to me.

Zarel:
Thanks!

PR #5940, opened 2019-11-08
Add a RoomBattle#challengeType property

TheJetOU:
This commit adds a `challengeType` property to `RoomBattle` of
type `rated | unrated | challenge | tour`.

Previously, there was no way to programatically differ an unrated
battle from a challenge, which is useful in places like filters.

(This pull request requires a restart.)

### DIFF
--- a/server/chat-plugins/chat-monitor.js
+++ b/server/chat-plugins/chat-monitor.js
@@ -287,7 +287,7 @@ let chatfilter = function (message, user, room) {
 		const {location, condition, monitor} = Chat.monitors[list];
 		if (!monitor) continue;
 		// Ignore challenge games, which are unrated and not part of roomtours.
-		if (location === 'BATTLES' && !(room && room.battle && room.battle.type !== 'challenge')) continue;
+		if (location === 'BATTLES' && !(room && room.battle && room.battle.challengeType !== 'challenge')) continue;
 		if (location === 'PUBLIC' && room && room.isPrivate === true) continue;
 
 		switch (condition) {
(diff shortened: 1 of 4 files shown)


Zarel:
Calling it `type` would probably make it confused with `Battle#gametype`. We should at least think of a better name.

Zarel:
Two ideas, pick one:

- Just overload `rated` with `"Tournament" | "Unrated" | "Challenge" | number`
- Name this `challengeType`

TheJetOU:
I'll go with `challengeType` because ``>> room.battle.rated`` returning `"Tournament"` feels weird.

Zarel:
It actually already returns "Tournament".

Zarel:
Well, `Battle#rated` does. `RoomBattle#rated` currently doesn't. But they should probably be updated to match, anyway.

TheJetOU:
> Well, `Battle#rated` does. `RoomBattle#rated` currently doesn't. But they should probably be updated to match, anyway.

It doesn't. Am I doing something wrong?

```
>> room
<< GameRoom(battle-gen7randombattle-1)
>> !!room.tour
<< true
>> room.rated
<< 0
```

Zarel:
I'm saying `Battle#rated` i.e. the `battle` object from `sim/`.

https://github.com/smogon/pokemon-showdown/blob/f6d5f9ddb8e03a05c7179d19214ce12f1d4d2ba9/sim/battle.ts#L65

This one.

TheJetOU:
Hm, I am leaning more towards renaming it to something more sensible (maybe even renaming the protocol message?), because a `rated` property holding the string "Tournament Battle" is illogical and unexpected, for me at least.

I'm not too interested in taking that on though currently.

Zarel:
That's fair. I should merge this soonish.

PR #5942, opened 2019-11-09
Generalize NFE bans into validator rule

TheMezStrikes:
so various NFE tournaments like old generations or Mix & Mega can be played without gentleman's rule or complex banlists. Tried to ask @TheImmortal for permission beforehand but didn't get an answer so figured I'd just propose it and see if it gets approved

### DIFF
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -535,6 +535,20 @@ let BattleFormats = {
 			this.add('rule', 'Z-Move Clause: Z-Moves are banned');
 		},
 	},
+	nfeclause: {
+		effectType: 'ValidatorRule',
+		name: 'NFE Clause',
+		desc: "Bans Pok&eacute;mon that cannot evolve",
+		onValidateSet(set) {
+			const template = this.dex.getTemplate(set.species || set.name);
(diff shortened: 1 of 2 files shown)


data/rulesets.js
```diff
@@ -535,6 +535,20 @@ let BattleFormats = {
 			this.add('rule', 'Z-Move Clause: Z-Moves are banned');
 		},
 	},
+	nfeclause: {
+		effectType: 'ValidatorRule',
+		name: 'NFE Clause',
+		desc: "Bans Pok&eacute;mon that cannot evolve",
```
Zarel: I feel like "cannot evolve" is a bit vague. I would say either "Bans Pokemon that are fully evolved", or "Bans Pokemon that are fully evolved or can't evolve".

TheImmortal: Agree. Should be the latter.

TheMezStrikes: Should be good now

Zarel:
I'm not super sold on the name, because clause names are usually what's banned, not what's not banned, but whatever.

PR #5943, opened 2019-11-11
Split up server/chat-commands/

Zarel:
`server/chat-commands.js` is now a directory. It's been split into `core`, `moderation`, and `admin`. `info` and `roomsettings` from `chat-plugins` have also moved to `chat-commands`.

Bot commands for inserting HTML into rooms have been moved from `info` into `admin`.

`/a` has been renamed `/addline`, for clarity (and also moved from `info` into `admin`).


Zarel:
(I'll be honest, I did this mostly because I wanted chat-commands to be TypeScripted, but VS Code was lagging on typechecking it because it was too big; this will let the parts pe TypeScripted piecemeal.)

panpawn:
Semi-unrelated, but I'm starting to wonder if ``server/`` should have a sub-directory called ``chat/``

server/chat-commands/core.js
```diff
@@ -0,0 +1,1629 @@
...
+		if (targetUser.id in targetRoom.users) return this.errorReply(`This user is already in "${targetRoom.title}".`);
+
+		return `/invite ${targetRoom.roomid}`;
+	},
+	invitehelp: [
+		`/invite [username] - Invites the player [username] to join the room you sent the command to.`,
+		`(in a PM) /invite [roomname] - Invites the player you're PMing to join the room [roomname].`,
+	],
+
+	pminfobox(target, room, user, connection) {
```
whalemer: Should these be in admin with the rest of the html commands?

scheibo:
> `/a` has been renamed `/addline`, for clarity (and also moved from info into admin).

~Hmm... seems like it could use an `/a` alias to continue to facilitate easy admin-abuse~

server/chat-commands/core.js
Zarel: I actually do like them in html commands, actually.

Zarel:
> ~Hmm... seems like it could use an `/a` alias to continue to facilitate easy admin-abuse~

In case you're half-serious, I think the up key is easy enough that I don't need to make it any easier.

PR #5944, opened 2019-11-15
Allow viewing roomfaqs without showing source

whalemer:
It's annoying and floody having the source come up for roomfaqs every time if you're a mod.

wtb better command name


Zarel:
Or you could put the source behind a collapsey thing?

whalemer:
I think I prefer the two commands, a la staffintro / stafftopic.

PR #5945, opened 2019-11-15
Fix build

KrisXV:
and add event gigantamax meowth and fix gigantamax legality, but thats not as important


TheImmortal:
Please don't merge this until I'm done cleaning up FormatsData.

PR #5947, opened 2019-11-15
Remove Eternatus-Eternamax

KrisXV:


sim/team-validator.ts
```diff
@@ -1186,7 +1186,7 @@ export class TeamValidator {
 		}
 
 		if (
-			tierTemplate.isNonstandard === 'Pokestar' && dex.gen === 5 ||
+			tierTemplate.isNonstandard === 'Unusable' && dex.gen === 5 ||
```
TheImmortal: Eternatus isn't gen 5

KrisXV: fixed

sim/team-validator.ts
```diff
@@ -1186,7 +1186,7 @@ export class TeamValidator {
 		}
 
 		if (
-			tierTemplate.isNonstandard === 'Pokestar' && dex.gen === 5 ||
```
xfix: This allows using Pokestar mons in Gen6.

KrisXV: done i think

PR #5949, opened 2019-11-15
Fix certain Gen 8 Pokemon's stats

420Blazeitt:
Just some discrepancies I found with the doc that I wanted to fix


data/pokedex.js
```diff
@@ -2298,7 +2298,7 @@ let BattlePokedex = {
 		species: "Weezing",
 		types: ["Poison"],
 		baseStats: {hp: 65, atk: 90, def: 120, spa: 85, spd: 70, spe: 60},
-		abilities: {0: "Levitate", 1: "Neutralizing Gas", H: "Misty Surge"},
+		abilities: {0: "Levitate", 1: "Neutralizing Gas", H: "Stench"},
```
Zarel: Regular Weezing gets Neutralizing Gas? Are we sure that's correct?

TheImmortal: Yes, that's correct according to the datamine.

PR #5961, opened 2019-11-16
Fix Gmax forme validation

KrisXV:


Zarel:
Wait, why not update the `isGigantamax` definition, instead of here?

KrisXV:
because i edited this on the website and im not sure where else hoeen might have used this and dont want to break anything

Zarel:
(Go up four lines.)

KrisXV:
ok fine

Zarel:
Oh, that's why Hoeen changed it; he wanted `isGigantamax` to be a string. Unfortunately, it's not set up yet, so you'll have to change it back to a boolean, too.

KrisXV:
ok done

Zarel:
It turns out you were right the first time, but now we know why. Progress, I guess?

PR #5965, opened 2019-11-16
Update datasearch for gen 8

TheMezStrikes:
- added `ds8` alias for `dexsearch` and `usumlearn` for `learn` (should maybe just be sm or both)
- added `gmaxSearch` and `tierSearch` variables to exclude gmax formes unless relevant
- increased every generation cap by one
- added `gmax` parameter to search for gmax formes
- (temporarily) made pokemon tiered `New` show up in `/ds ou`
- Added life dew to recovery moves

Might be wise to not merge until gen8 is more mainstream competitively so people for now just get their gen 7 default datasearch.
Also, `/ds all` excludes pokemon that aren't in the galar dex, but do when you `/ds all, maxgen7`. Might be an unwanted inconsistency, might be the way it should be, I'm not quite sure what to do with it.

### DIFF
--- a/server/chat-plugins/datasearch.js
+++ b/server/chat-plugins/datasearch.js
@@ -28,6 +28,7 @@ exports.commands = {
 	ds5: 'dexsearch',
 	ds6: 'dexsearch',
 	ds7: 'dexsearch',
+	ds8: 'dexsearch',
 	dsearch: 'dexsearch',
 	dexsearch(target, room, user, connection, cmd, message) {
 		if (!this.canBroadcast()) return;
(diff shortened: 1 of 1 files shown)


TheMezStrikes:
Crashes on `/ds [move]` do not merge

TheMezStrikes:
I'm not sure what happened, but after changing nothing and hotpatching everything works again. Will look into this more

TheMezStrikes:
Not clean but it works

server/chat-plugins/datasearch.js
```diff
@@ -614,16 +630,16 @@ function runDexsearch(target, cmd, canAll, message) {
...
 				let tier = dex[mon].tier;
 				if (tier[0] === '(' && tier !== '(PU)') tier = tier.slice(1, -1);
+				if (tier === 'New') tier = 'OU';
 				if (alts.tiers[tier]) continue;
 				if (Object.values(alts.tiers).includes(false) && alts.tiers[tier] !== false) continue;
 				// LC handling, checks for LC Pokemon in higher tiers that need to be handled separately,
 				// as well as event-only Pokemon that are not eligible for LC despite being the first stage
 				let format = Dex.getFormat('gen' + maxGen + 'lc');
-				if (!format.exists) format = Dex.getFormat('gen7lc');
+				if (!format.exists) format = Dex.getFormat('gen8lc');
```
xfix: There is no format called `gen8lc` currently.

TheMezStrikes: Since as far as I can tell, this code is supposed to handle cases that won't exist before gen8lc exists, this change should be harmless

TheMezStrikes: lc exists now

PR #5966, opened 2019-11-16
Add shortDesc to abilities

TheMezStrikes:
Some of these are really bad, particularly Gulp Missile, but I didn't know how to condense it properly

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -528,7 +528,8 @@ let BattleAbilities = {
 		num: 212,
 	},
 	"cottondown": {
-		shortDesc: "When the Pokémon is hit by an attack, it scatters cotton fluff around and lowers the Speed stat of all Pokémon except itself.",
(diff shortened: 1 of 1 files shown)


data/abilities.js
```diff
@@ -1344,7 +1345,8 @@ let BattleAbilities = {
 		num: 229,
 	},
 	"gulpmissile": {
-		shortDesc: "When the Pokémon uses Surf or Dive, it will come back with prey. When it takes damage, it will spit out the prey to attack.",
+		desc: "When the Pokémon uses Surf or Dive, it will come back with prey. When it takes damage, it will spit out the prey to attack.",
+		shortDesc: "Using Surf/Dive gives prey. When taking damage, prey is used to attack.",
```
```suggestion
		shortDesc: "Get prey with Surf/Dive. When taking damage, prey is used to attack.",
```

PR #5967, opened 2019-11-16
Add Gen 8 Doubles OU Beta

DaWoblefet:
Bans for Gen 8 Doubles OU, approved from tier leader MajorBowman:

- Zacian
- Zamazenta
- Eternatus
- Gravnosis clause

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -648,6 +648,17 @@ let Formats = [
 		banlist: ['Uber', 'Baton Pass'],
 	},
 	{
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Well it would help if I could recognize what the new gen number was, but it is actually Gen 8.

Zarel:
I'm pretty sure Grass Whistle no longer exists?

DaWoblefet:
Fixed. I made sure to check I wasn't missing any new sleep moves either.

PR #5968, opened 2019-11-16
Octolock fix

PrincessScepti:
Fixing Octolock to add the def/spd drop

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -13002,10 +13002,14 @@ let BattleMovedex = {
 		onHit(target, source, move) {
 			return target.addVolatile('trapped', source, move, 'trapper');
 		},
+		volatileStatus: 'octolock',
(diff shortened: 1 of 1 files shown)


data/moves.js
```diff
@@ -13002,10 +13002,14 @@ let BattleMovedex = {
...
+		volatileStatus: 'octolock',
 		effect: {
+			onStart(target) {
+				this.add('-start', target, 'Octolock');
+			},
 			onResidualOrder: 13,
-			onResidual(target, source) {
-				this.boost({def: -1, spd: -1}, target, source, this.dex.getActiveMove("Octolock"));
+			onResidual(target) {
+				this.boost({def: -1, spd: -1});
```
Zarel: Why did you remove the effect name of Octolock?

PR #5969, opened 2019-11-16
Fix data loading for 'generate-team'

asgdf:
Since the data files are only loaded in the getter for Dex.data, which RandomTeam does not actually access, Dex.gen was 0, which led to an empty pokemon pool for random team generation.
This way, Dex.forFormat loads the data required for that format on its own.

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -222,7 +222,7 @@ export class ModdedDex {
 	forFormat(format: Format | string): ModdedDex {
 		if (!this.modsLoaded) this.includeMods();
 		const mod = this.getFormat(format).mod;
-		return dexes[mod || 'gen7'].includeData();
+		return dexes[mod || 'gen8'].includeData();
 	}
 
 	modData(dataType: DataType, id: string) {


Zarel:
Half-implemented in 43ef4d978548b0076686. Should we default to gen 8?

asgdf:
Looking at our formats, I figure it doesn't make that big of a difference: all formats have a mod specified anyway. But I assumed we might as well default to the most recent gen?

Zarel:
Fair. Feel free to rebase.

asgdf:
Okay.

Zarel:
Can you update all our existing mods to point to Gen 7? Otherwise that's just asking for bugs.

TheImmortal:
Not sure what mod SSB should be using. @HoeenCoder ?

Zarel:
Someone reported Tier Shift was messed up?

asgdf:
Tier Shift is explicitly set to use the gen7 mod. I think they might have been referring to the /ts command, which would be unrelated to this. But, fair, I can look at it.

PR #5970, opened 2019-11-16
Fix Pokemon in pastgen randbats sometimes having no moves

asgdf:
...by taking away their Gmax formes in the oldgen pokedex, so those don't get randomly chosen anymore.

### DIFF
--- a/data/mods/gen7/pokedex.js
+++ b/data/mods/gen7/pokedex.js
@@ -2,6 +2,18 @@
 
 /**@type {{[k: string]: ModdedTemplateData}} */
 let BattlePokedex = {
+	charizard: {
+		inherit: true,
+		otherFormes: ["charizardmegax", "charizardmegay"],
+	},
+	butterfree: {
+		inherit: true,
+		otherFormes: undefined,
+	},
+	pikachu: {
+		inherit: true,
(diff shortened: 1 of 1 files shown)


Zarel:
`getPokemonPool` has `if (!species.randomBattleMoves) continue;`

How did this get broken?

Zarel:
In the future, do not ever remove pokemon from oldgen pokedexes. This is what `isNonstandard` is for.

Zarel:
The relevant functions should be checking `isNonstandard`.

TheImmortal:
I don't recall what happened here but ``getPokePool`` has ``if (species.gen > this.gen) continue;`` so rolling Gmax Pokemon should never occur regardless.

I'd assume it was an issue with Gmax Pokemon not being identified as gen 8 at the time.

Zarel:
I found it, the problem is in `gen7/random-teams.ts`:

```
species = this.dex.getSpecies(
	species.otherFormes.length >= 2 ? this.sample(species.otherFormes) : species.otherFormes[0]
);
```

This doesn't have an `isNonstandard` check (and also seems to think that the base forme is part of the formes list).

TheImmortal:
Ah. Tbh, I've never liked how that was implemented. Slayer or someone did it during a period when I was inactive. All formes should be rolled using ``getPokemonPool`` in ``randomTeam``. Formes shouldn't be altered in ``randomSet`` imo. I will rewrite it when I find time.

Zarel:
The last relevant change seems to be:

4410843e5ac33bb08668affef1719cf1e5ee7aac

Holy shit, five years ago.

Anyway, it seems outright incorrect to me, and `getPokemonPool` should already handle it.

PR #5974, opened 2019-11-16
Update dexsearch for gen 8

tmagicturtle:


tmagicturtle:
Just noticed that this already has a PR

PR #5976, opened 2019-11-17
Support gen 8 challenge cup

whalemer:
(editor's note: this doesn't do anything right now, as we don't have a gen 8 cc format)

### DIFF
--- a/data/random-teams.js
+++ b/data/random-teams.js
@@ -22,6 +22,9 @@ const PRNG = require(/** @type {any} */ ('../.sim-dist/prng')).PRNG;
  * @property {number} [eeveeLimCount]
  */
 
+/** @type {string[][]} */
+let gen8ValidPoolCache;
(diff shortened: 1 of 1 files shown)


TheImmortal:
Let me see if I can do this simpler

TheImmortal:
Gen 8 CC actually works correctly without this so I will assume it is purely for the cache.

@Zarel is this necessary

whalemer:
I'm not sure what you're doing, but for me (verified by kris), gen 8 CC does not work on the latest commit, crashing with something like ``CRASH: Error: Invalid pokemon gen 8: [[],[],[],[],[],[]] numbers {"1.6900334125384688":0,"1.629456696799025":1,"1.301599362399429":2,"1.3562490125186741":3,"1.9336451017297804":4,"1.0593001677189022":5}``

Issue #5975, opened 2019-11-17
Body Press is broken


Fikou:
body press is supposed to use your defense stat as the attacking stat (at least i assume that from the move description)
but it uses your defense stat as the defending stat which makes it only good for glass cannons
https://replay.pokemonshowdown.com/gen8customgame-1011023106

Shiny-andKittens:
I looked under the move’s code, it has a flag called `useSourceDefensive:true`.

This may be the root of the issue.

It looks like that could be used for confusion damage purposes and was applied to the move for some reason

Shiny-andKittens:
Turns out, it is indeed the root of the issue.

I’ll try to make a fix for it.

Zarel:
Oh, the move description was written so vaguely that I didn't realize that was what it was supposed to do. We had a confusing discussion about it on the Dev chatroom and it's not until now that I realized what the actual problem was.

PR #5979, opened 2019-11-17
Fix useSourceDefensive

Shiny-andKittens:
Explanation:

This move was supposed to use the User’s Defense stat in place of the Attack Stat, but instead replaces the Target’s Defense stat with the user’s, having the opposite effect.

The original code just gives it the exact opposite effect of `useTargetOffensive`:

useTargetOffensive replaces the user’s attack with the target’s
useSourceDefensive replaces the target’s defense with the user’s.

I changed it so the target replaces it’s own attack with it’s own defense (and otherwise has a normal effect when it’s `false`).

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2017,9 +2017,9 @@ export class Battle {
 		}
 
 		if (move.useSourceDefensive) {
-			defense = attacker.calculateStat(defenseStat, defBoosts);
+			attack = attacker.calculateStat(defenseStat, defBoosts);
 		} else {
-			defense = defender.calculateStat(defenseStat, defBoosts);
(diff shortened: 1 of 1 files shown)


Shiny-andKittens:
This was written to fix issue #5975

Zarel:
The name `useSourceDefensive` is what caused this problem in the first place. I'd name it `useSourceDefensiveAsOffensive`.

Shiny-andKittens:
Was my fix correct other than that?

Zarel:
No. You removed all code for calculating Defense at all, causing the build failure you see. My commit 4fc49bc is the correct way to do it.

PR #5981, opened 2019-11-17
Make Snipe Shot ignore redirection

KrisXV:


data/moves.js
```diff
@@ -17355,14 +17355,20 @@ let BattleMovedex = {
...
 		pp: 15,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
 		critRatio: 2,
+		effect: {
+			onRedirectTargetPriority: 3,
+			onRedirectTarget(target, source, source2, move) {
+				return target;
+			},
+		},
```
urkerab: Don't you need to add the volatile? (Also remove it again, in case this was executed due to Instruct.)

Zarel: (yes)

Zarel: That, or try a different approach. I like a move flag saying "ignore redirects".

KrisXV: I'm confused. What volatile are you talking about?

urkerab: When you add an `effect` block, you're creating an effect, but you have to add it as a volatile (or status, weather, etc.) for it to actually do anything. Because you never add the effect, it doesn't have any... effect... (sorry I couldn't think of a way out of the pun)

Zarel: This is all irrelevant, because this should be a move-property-boolean rather than a custom effect.

KrisXV:
this was a mistake oops

Zarel:
Closing it was? I think if you close and forcepush, there's no way to reopen. You'd have to make a new PR (if you haven't yet).

urkerab:
I think you can maybe reopen if you force push back to when you closed it?

PR #5982, opened 2019-11-17
Fix the evolution method of new Pokemon

KrisXV:


data/pokedex.js
```diff
@@ -14789,7 +14789,7 @@ let BattlePokedex = {
 		weightkg: 117,
 		color: "White",
 		prevo: "farfetchdgalar",
-		evoType: "levelExtra",
```
Zarel: Does it need to level-up at all? If so, `evoCondition` should be reworded with `" after landing"` or something; If not, we need a new `evoType: "other"`.

KrisXV: it doesnt and it appears theres other mons that can evolve without leveling too, so ill modify this PR in a bit

Zarel: Thanks!

KrisXV:
done @Zarel

PR #5983, opened 2019-11-17
Effects of (G-)Max Moves should not be secondaries

urkerab:


KrisXV:
There should be a `secondary: null,` property on all of these moves if this is the case

Zarel:
Huh, I'm surprised `secondary: null` is required.

data/moves.js
```diff
@@ -7146,14 +7143,7 @@ let BattleMovedex = {
...
 		isMax: "Orbeetle",
-		secondary: {
-			chance: 100,
-			self: {
-				onHit() {
-					this.field.addPseudoWeather('gravity');
-				},
-			},
-		},
+		pseudoweather: 'gravity',
```
```suggestion
		pseudoWeather: 'gravity',
```

urkerab: Huh, and I thought I'd tested all of the moves. I must be blind...

urkerab:
`secondary: null` (or its predecessor `secondary: false`) doesn't seem to have been necessary for years. Commit 21be1f4 centralised secondary handling, but even back then, I can't see anything checking for a specific value.

PR #5984, opened 2019-11-17
Make Pyro Ball viable

Alumn2000:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8285488


Alumn2000:
The build says that the error begins on "data/abilities.js:1566:3" (so I don't think that I broke it).

PR #5986, opened 2019-11-17
Make Libero descriptions match Protean

PartMan7:
(same as Protean)


Zarel:
Is there a long desc to also copy from Protean?

PartMan7:
Missed that, sorry. Adding.

PR #5987, opened 2019-11-17
Convert index.js and team-validator-async.js to typescript

not-a-seagull:


server/index.ts
```diff
@@ -35,13 +35,10 @@
  * Sockets - from sockets.js
  *
  *   Used to abstract out network connections. sockets.js handles
- *   the actual server and connection set-up.
```
Zarel: ? I assume this was a mistake

server/index.ts
```diff
@@ -54,24 +51,29 @@ try {
...
 }
+
 try {
 	require.resolve('../.sim-dist/index');
 } catch (e) {
 	throw new Error("Dependencies are unmet; run `node build` before launching Pokemon Showdown again.");
 }
 
-const FS = require('../.lib-dist/fs').FS;
+import { FS } from '../lib/fs';
```
Zarel:
```suggestion
import {FS} from '../lib/fs';
```
(style)

server/index.ts
not-a-seagull: Yeah, I don't know what that's doing there.

server/index.ts
```diff
@@ -54,24 +52,29 @@ try {
...
-const FS = require('../.lib-dist/fs').FS;
+import {FS} from '../lib/fs';
 
 /*********************************************************
  * Load configuration
  *********************************************************/
 
-const ConfigLoader = require('../.server-dist/config-loader');
+// global becomes much easier to use if declared as an object
+declare var global: any;
```
Zarel:
```suggestion
declare const global: any;
```

We shouldn't be changing where it points to, anyway.

server/team-validator-async.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const get = TeamValidatorAsync.get;
+
+/*********************************************************
+ * Process manager
+ *********************************************************/
+
+import { QueryProcessManager } from '../lib/process-manager';
+
+// @ts-ignore
```
Zarel: Wait, what's being suppressed here?

server/team-validator-async.ts
```diff
@@ -0,0 +1,106 @@
...
+export const PM = new QueryProcessManager(module, async message => {
```
Zarel: Why `async`?

Zarel: Looks like it was in the original code. I'm fine with leaving it alone.

server/team-validator-async.ts
not-a-seagull: It was there in the original code. When I removed it, TSC didn't throw any errors.

server/index.ts
```diff
@@ -54,24 +52,29 @@ try {
...
 
-const ConfigLoader = require('../.server-dist/config-loader');
+// global becomes much easier to use if declared as an object
+declare const global: any;
+
+import * as ConfigLoader from './config-loader';
 global.Config = ConfigLoader.Config;
 
-global.Monitor = require('../.server-dist/monitor').Monitor;
+import { Monitor } from './monitor';
```
Zarel:
```suggestion
import {Monitor} from './monitor';
```

Could you grep `import { ` (with the space) and fix those, too?

not-a-seagull: Resolved this in the latest commit

TheJetOU:
You'll also need to edit the `pokemon-showdown` script to refer to `.server-dist` when starting up the server.

Zarel:
This will also be awkward for the next time we restart the server.

Might as well update to Node 13, right? Right???

Zarel:
Ugh, I can't merge this for today's restart because of conflicts. :(

TheJetOU:
> Ugh, I can't merge this for today's restart because of conflicts. :(

Thankfully, maintainers can push to branches that are being PR'ed :).

Zarel:
@TheJetOU Unless it's merging suggestions, because of a GitHub bug. Anyway, I don't have time for complicated fixes during restarts anyway.

not-a-seagull:
I fixed the merge conflicts.

Zarel:
@TheJetOU brings up a good point, this might not require a restart. The hotpatchable/unhotpatchable interface doesn't seem to be crossed much here.

PR #5991, opened 2019-11-18
Require a break on either side of evaded words

whalemer:
This doesn't exactly feel like a /good/ way to do this to me, but I can't think of any better way.


server/chat-plugins/chat-monitor.js
```diff
@@ -151,8 +144,8 @@ Chat.registerMonitor('evasion', {
 		lcMessage = lcMessage.normalize('NFKC');
 
 		let [regex, word, reason] = line;
-		// Remove spaces and obvious false positives
-		lcMessage = lcMessage.replace(/[\s-_,.]/g, '');
+		// Normalise spaces and other common evasion characters to a period
+		lcMessage = lcMessage.replace(/[\s-_,.]+/g, '.');
```
Zarel: I would not call this `lcMessage` if you're doing such a drastic change.

PR #5992, opened 2019-11-18
RoomGame: Uncomment optional functions

TheJetOU:
TypeScript will now recognizes these functions existence so
we don't have to supress the compiler whenever they come up.

### DIFF
--- a/server/room-game.ts
+++ b/server/room-game.ts
@@ -190,27 +190,37 @@ export class RoomGame {
 
 	// These are all optional to implement:
 
-	// forfeit(user)
-	//   Called when a user uses /forfeit
-	//   Also planned to be used for some force-forfeit situations, such
-	//   as when a user changes their name and .allowRenames === false
-	//   This is strongly recommended to be supported, as the user is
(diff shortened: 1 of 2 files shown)


whalemer:
There's stuff like `/forfeit` and friends that check for if these functions exist to return a "not supported" message, you might want to return a similar message from the default implementation instead.

TheJetOU:
> There's stuff like `/forfeit` and friends that check for if these functions exist to return a "not supported" message, you might want to return a similar message from the default implementation instead.

Those commands don't return the function though, they just run them. So, you'd need to refactor them in some way which doesn't seem worth the effort.

Zarel:
Having these be nullable is nice especially since we're switching to allowing multiple games in the same room; if only one of them has a `choose` function, `/choose` is unambiguous.

whalemer:
Should these default to null, then? Right now you'd need to explicitly null them for that to work.

Zarel:
Null and undefined work the same way. Feel free, but it's not a big deal here.

whalemer:
Oh, my bad, was misunderstanding syntax.

PR #5995, opened 2019-11-18
Fix Tar Shot

Alumn2000:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8286082

1) The short description "-1 Spe" was misleading. We used that for Curse who lowers the user's Speed but Tar Shot lowers the target Speed.
2) It's not a contact move. It's a bounceable move (so reflectable)


data/moves.js
```diff
@@ -19157,12 +19157,12 @@ let BattleMovedex = {
 		basePower: 0,
 		category: "Status",
 		desc: "Lowers the target's Speed by 1 stage. Until the target switches out, it takes 2x damage from Fire moves.",
-		shortDesc: "-1 Spe. Target takes 2x damage from Fire moves.",
+		shortDesc: "Lowers the target's Speed by 1 stage. Target takes 2x damage from Fire moves.",
```
HoeenCoder: This is too long, it cuts off. 
image

Alumn2000: Maybe "-1 target Spe" ?

Alumn2000: Opera Instantané_2019-11-18_151644_localhost psim us
Maybe that's ok 
I don't know which verb will be better [takes, gets, receives (too long)], because it doesn't sound english for me.
I have another wording who's good too
Opera Instantané_2019-11-18_152437_localhost psim us

Zarel: "Target gets -1 Spe and takes 2x damage from Fire."

Zarel:
> The short description "-1 Spe" was misleading. We used that for Curse who lowers the user's Speed but Tar Shot lowers the target Speed.

It's reasonably accurate, I think. In both moves, the -1 Spe happens to the move target. But sure, more clarity is always better.

PR #5997, opened 2019-11-18
Fix Max Move effects not activating against substitute

420Blazeitt:


data/moves.js
```diff
@@ -7143,7 +7147,9 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {},
 		isMax: "Orbeetle",
-		pseudoWeather: 'gravity',
+		self: {
+			pseudoWeather: 'gravity',
```
Zarel: In theory, this one shouldn't be necessary (nor the others).

420Blazeitt: it is necessary, I tested the current code on PS and here are the results: https://replay.pokemonshowdown.com/gen8customgame-1012218343

data/moves.js
```diff
@@ -7074,10 +7076,12 @@ let BattleMovedex = {
...
 		flags: {},
 		isMax: "Alcremie",
-		onAfterHit(target, source) {
-			for (let pokemon of source.side.active) {
-				pokemon.heal(pokemon.maxhp);
-			}
+		self: {
+			onAfterHit(target, source) {
+				for (let pokemon of source.side.active) {
+					pokemon.heal(pokemon.maxhp);
```
urkerab: While you're here, this should be `this.heal(pokemon.maxhp, pokemon, pokemon);`. See https://www.smogon.com/forums/posts/8287090/

420Blazeitt: thanks!

420Blazeitt: thanks!

data/moves.js
Zarel: Awkward. I guess this will need to be fixed later.

Zarel: Nice creative hack, though.

PR #5998, opened 2019-11-18
Fix missing egg groups

Nolnis:
Add Human-Like egg group to Ralts line and Dragon egg group to Trapinch line
Source: https://pastebin.com/mBKX1Byd


TheImmortal:
Is this new? If it is then it should be removed in the gen 7 mod.

Zarel:
Yeah, these are new, and they should also definitely be removed from the gen 7 mod.

Nolnis:
https://github.com/smogon/pokemon-showdown/pull/6003

PR #6000, opened 2019-11-18
Fix Dragon Darts interaction in Doubles

Alumn2000:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8287014


Alumn2000:
Btw it's nice that I have the privilege to have the #6000 pr :)

TheImmortal:
This still isn't accurate mechanics. It hits only a single target if the other target is a Fairy.

Source: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8286618

PR #6002, opened 2019-11-18
Random Battle: Update Greedent

ascriptmaster:


TheImmortal:
Uh, I was going to give it Iapapa Berry. It doesn't have Sub, and it has SD so Liechi isn't helpful. Iapapa + Cheek Pouch was the plan.

ascriptmaster:
Isn't it hard to hit the 25% threshold? That's why I went Iapapa + Gluttony. But I guess your idea works too.

ascriptmaster:
Changed the code to allow for Iapapa with EITHER ability, since I think they're both situationally useful. Gluttony changes a 2HKO to a 3HKO, but Cheek Pouch heals more in other situations.

TheImmortal:
I don't think the future-proofing is necessary. I've been actually cleaning the file of things irrelevant to Galar mons.

PR #6003, opened 2019-11-18
Fix egg groups in gen 7 mod

Nolnis:
Ralts and Trapinch families should only have Amorphous and Bug egg groups respectively until gen 8


Zarel:
Can this and #5998 be the same PR? Do you need help editing PRs?

Nolnis:
@Zarel Yeah, they can be the same PR. I'm unsure of how to merge the two.

PR #6006, opened 2019-11-19
Update Intimidate Interactions

PartMan7:
Added Speed boost due to Rattled being hit by Intimidate (source: https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8285401).
Edited desc and shortDesc of Rattled to include new interaction.

### DIFF
--- a/data/mods/gen7/abilities.js
+++ b/data/mods/gen7/abilities.js
@@ -14,6 +14,7 @@ let BattleAbilities = {
 	},
 	"innerfocus": {
 		inherit: true,
+		shortDesc: "This Pokemon cannot be made to flinch.",
 		rating: 1,
 	},
 	"intimidate": {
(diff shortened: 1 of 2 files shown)


PartMan7:
Sorry about that mix-up.

data/mods/gen7/abilities.js
```diff
@@ -34,6 +35,35 @@ let BattleAbilities = {
...
+		shortDesc: "This Pokemon's Speed is raised 1 stage if hit by a Bug-, Dark-, or Ghost-type attack.",
+		onAfterDamage(damage, target, source, effect) {
+			if (effect && (effect.type === 'Dark' || effect.type === 'Bug' || effect.type === 'Ghost')) {
+				this.boost({spe: 1});
+			}
+		},
+		id: "rattled",
+		name: "Rattled",
+		rating: 1.5,
+		num: 155,
```
urkerab: I guess this could have been inherited and then the onAfterBoost function nulled out?

TheImmortal: Would have had to add the descs too.

urkerab: The idea was that I was commenting on lines +53 to +61 as being the properties that would have been inherited.

PR #6009, opened 2019-11-19
Clarify misleading move/ability descriptions

PartMan7:
Current Body Press shortDesc: `Uses user's Defense stat in damage calculation.` (Implies that the damage calculation uses the user's Def instead of the opponent's; unsure of whether this should be changed to `Uses user's Def instead of Atk in damage calculation.`, since the exact mechanics for the move aren't yet available. For now, I've used a description similar to Electro Ball.)
Rapid Spin's current shortDesc is three characters too long.

// Done - Clarify Propeller Tail.
Propeller Tail - Slightly reworded to avoid confusion with Mold Breaker-like effects.


data/abilities.js
```diff
@@ -2880,7 +2880,7 @@ let BattleAbilities = {
 		num: 232,
 	},
 	"propellertail": {
-		shortDesc: "Ignores the effects of opposing Pokémon's Abilities and moves that draw in moves.",
+		shortDesc: "Ignores the effects of opposing Pokémon's moves / Abilities that draw in moves.",
```
```suggestion
		shortDesc: "Ignores the effects of opposing Pokémon's moves/Abilities that redirect moves.",
```

data/moves.js
```diff
@@ -1622,7 +1622,7 @@ let BattleMovedex = {
 		basePower: 80,
 		category: "Physical",
 		desc: "Damage is calculated using the user's Defense stat, including stat stage changes. The user's Ability is used as normal.",
```
```suggestion
		desc: "Damage is calculated using the user's Defense stat instead of Attack, including stat stage changes. The user's Ability is used as normal.",
```

data/moves.js
```diff
@@ -1622,7 +1622,7 @@ let BattleMovedex = {
...
-		shortDesc: "Uses user's Defense stat in damage calculation.",
+		shortDesc: "More power the higher the user's Def.",
```
```suggestion
		shortDesc: "Uses Def instead of Atk in damage calculation.",
```

PR #6014, opened 2019-11-19
Fix previous gen Pikachu/Gumshoos formes

monsanto:
This issue was caught by smogon/data's test suite when upgrading from 7 -> 8. I'm not 100% sure about the other attributes, I'm just following the lead on Vikavolt-Totem's entry.


TheImmortal:
No need for Pichu/Pikachu. They are still illegal in gen 7. The gen 6 mod frees them.

monsanto:
`isNonstandard: "Past"` is incorrect for them.

Issue #6005, opened 2019-11-19
Move Stuff Cheeks only raises Defense


jshenfield:
Currently, Stuff Cheeks raises the Defense stat by two stages without consuming whichever berry the user holds

jshenfield:
So actually, this didn't look like a difficult fix, so I had a try at writing the fix. Would I just submit a Pull Request after local testing/linting?

Zarel:
Yeah, sure!

KrisXV:
Fixed in #6023

PR #6023, opened 2019-11-20
Fix Stuffed Cheeks

jshenfield:


Zarel:
In the future, please fix pull requests instead of creating new ones. People in the Dev chatroom or Discord will generally be happy to help you.

Zarel:
(Re: #6021)

Zarel:
I assume this is the new fix for #6005

Zarel:
Anyway, it looks like we should have an `eat` script, instead of neeing to reimplement it in all of these.

Zarel:
Wait, we do. `pokemon.eatItem()` already exists. Is there a reason you're not using it?

jshenfield:
Sorry about the closing and re-opening MR. Yeah, this is for #6005 
I'll need a moment to look at `pokemon.eatItem()`, but I can use that instead.

jshenfield:
I did a search across moves.js, and it looks like other places are using `runEvent('EatItem')...`
Should the other moves be refactored to use `pokemon.eatItem()`? I'm guessing this is the kind of question I can ask in dev chatroom/Discord

Zarel:
Yes please!

TZubiri:
A user just [reported this bug](https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8288667), good foresifght!  
Is this ready to be merged? Perhaps the refactoring of the other EatItem moves could be done in another PR

jshenfield:
So I need to push one more commit. Currently stuff Cheeks will eat ANY
item, so I need to add a check if the item is a berry. If it is alright,
can I make the refactor another PR?

I also have one more bug to write up relating to cheek pouch :)

jshenfield:
Alright, assuming CI passes, this one should be good to go

Zarel:
CI did not pass.

jshenfield:
Walp. Let me look into that.
BTW, I'll post into the discord, but I do have some questions

data/moves.js
```diff
@@ -18573,15 +18573,20 @@ let BattleMovedex = {
...
+		shortDesc: "Consumes berry and raises the user's Def. by 2 stages.",
 		id: "stuffcheeks",
 		name: "Stuff Cheeks",
 		pp: 10,
 		priority: 0,
 		flags: {snatch: 1},
-		boosts: {
-			def: 2,
+		onTryHit(target, source, move) {
+            let item = source.getItem();
```
```suggestion
			let item = source.getItem();
```

jshenfield:
False alarm on the other bug.
I tested Stuff Cheeks while holding a mental herb, and consuming the herb with the move triggered Cheek Pouch healing. I didn't see the issue when the herb was consumed with plain old taunt.

data/moves.js
```diff
@@ -18573,15 +18573,20 @@ let BattleMovedex = {
...
-		shortDesc: "Raises the user's Def. by 2 stages.",
+		desc: "Consumes berry and raises the user's Defense by 2 stages.",
+		shortDesc: "Consumes berry and raises the user's Def. by 2 stages.",
 		id: "stuffcheeks",
 		name: "Stuff Cheeks",
 		pp: 10,
 		priority: 0,
 		flags: {snatch: 1},
-		boosts: {
-			def: 2,
```
urkerab: You could probably have left the boosts in as the primary effect of the move, and then just used the `TryHit` event to make the move fail if the user could not eat a berry.

PR #6029, opened 2019-11-20
Fix Tar Shot

KrisXV:
Tar Shot changes the target's effectiveness to become weaker against Fire-type moves instead of a set modifier.


data/moves.js
```diff
@@ -19279,11 +19279,10 @@ let BattleMovedex = {
...
 				this.add('-start', pokemon, 'Tar Shot');
 			},
-			onSourceModifyDamage(damage, source, target, move) {
-				if (target.volatiles['tarshot'] && this.dex.getActiveMove(move).type === 'Fire') {
-					// TODO: Figure out damage modifier
-					return this.chainModify(2);
-				}
+			onEffectiveness(typeMod, target, type, move) {
+				if (move.type !== 'Fire') return;
+				if (!target || !target.volatiles['tarshot']) return;
```
Zarel: I'm curious what situations this would be true in.

KrisXV: i used the code for thousand arrows as a base and it used it as a crash prevention in chat commands. ig we dont really need the `!target` part

Zarel: Do we need the other part?

urkerab: We don't need either. This is running as an effect, so must be in a battle, while Thousand Arrows and Flying Press have the event directly on the move itself, which is why Thousand Arrows needs the chat command protection.

data/moves.js
```diff
@@ -19279,11 +19279,10 @@ let BattleMovedex = {
...
+				return this.dex.getEffectiveness('Fire', target) + 1;
```
urkerab: Could this be just `typeMod + 1`?

PR #6032, opened 2019-11-20
Ban Corsola-Galar from Gen 8 LC

TheRealQuote:
Corsola-Galar into LC Ubers


KrisXV:
Why are none of the banned Pokemon tiered as "LC Uber" in formats-data?

PR #6043, opened 2019-11-21
Cramorant always changes to Cramorant-Gorging if its HP is 50% or less

KrisXV:


data/abilities.js
```diff
@@ -1373,7 +1373,7 @@ let BattleAbilities = {
 			if (!['surf', 'dive'].includes(move.id) || source.volatiles['dive'] || source.speciesid !== 'cramorant' || source.transformed) return;
 
 			// PLACEHOLDER Estimated 10% of the time you get pikachu
-			const forme = this.random(10) === 1 ? 'cramorantgorging' : 'cramorantgulping';
+			const forme = this.random(10) === 1 || source.hp <= Math.floor(source.maxhp / 2) ? 'cramorantgorging' : 'cramorantgulping';
```
TheImmortal: I don't understand. If it's 50%, why did you keep the 10% roll?

KrisXV: it still has a random chance to change into Cramorant-Gorging above 50%

HoeenCoder: Some tests that my brother did for me confirms its possible to get Gorging above 50%. Going to merge as is and get the exact percent later.

HoeenCoder: After testing about 50 uses of surf cramorant with > 50% HP, Pikachu never appeared. So I'd say its supposed to be the fish when > 50%, and pikachu when < 50%.

data/abilities.js
```diff
@@ -1371,9 +1371,7 @@ let BattleAbilities = {
 		},
 		onAfterMove(source, target, move) {
 			if (!['surf', 'dive'].includes(move.id) || source.volatiles['dive'] || source.speciesid !== 'cramorant' || source.transformed) return;
-
-			// PLACEHOLDER Estimated 10% of the time you get pikachu
-			const forme = this.random(10) === 1 ? 'cramorantgorging' : 'cramorantgulping';
+			const forme = source.hp <= Math.floor(source.maxhp / 2) ? 'cramorantgorging' : 'cramorantgulping';
```
urkerab: Note that Belly Drum doesn't bother using `Math.floor`.

HoeenCoder: Any reason this was marked as resolved?

KrisXV: I dont remember doing this

KrisXV: ok now it is actually resolved

PR #6045, opened 2019-11-21
Add missing G-Max moves

KrisXV:


data/moves.js
```diff
@@ -7368,6 +7368,279 @@ let BattleMovedex = {
...
+		id: "gmaxvolcalith",
+		isNonstandard: "Custom",
+		name: "G-Max Volcalith",
+		pp: 10,
+		priority: 0,
+		flags: {},
+		isMax: "Coalossal",
+		onHit(target, source, move) {
+			for (const pokemon of target.side.active) {
+				pokemon.addVolatile('gmaxvolcalith');
```
urkerab: Current thinking seems to be that this is a side condition. You do at least expect your effect to be called on a side, rather than a Pokémon, so you should add it to the side too. (And then remove everything to do with checking the volatile of course.)

KrisXV: Someone can do testing and confirm if that's true. For now I'm treating it individually since Pokemon can still switch and switch ins are not affected (from what I know)

PR #6049, opened 2019-11-22
Eternabeam is one word

PartMan7:
Source: https://bulbapedia.bulbagarden.net/wiki/Eternabeam_(move)


Zarel:
Are we actually using Bulbapedia as a source? I remember they were one of the worst recently.

PartMan7:
PokemonDB, pokemon.fandom, and Serebii all say the same thing, too.

Zarel:
Yes, I know, I'm just saying you shouldn't cite Bulbapedia.

PartMan7:
Noted.

PR #6050, opened 2019-11-22
Fix lockedmove volatile interaction with Max moves

KrisXV:
Currently, moves that cause fatigue such as Thrash and Outrage will cause the battle to pause after the first turn of dynamaxing. in [this battle](https://replay.pokemonshowdown.com/gen8doublesou-1014430299), n10sit uses Max Wyrmwind based off of Outrage to attack, and it works fine the first turn. However, the battle falsely emits an Invalid choice on the following turn.

### DIFF
--- a/data/statuses.js
+++ b/data/statuses.js
@@ -291,6 +291,7 @@ let BattleStatuses = {
 			target.addVolatile('confusion');
 		},
 		onLockMove(pokemon) {
+			if (pokemon.volatiles['dynamax']) return;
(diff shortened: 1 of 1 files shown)


Zarel:
Is this true? Unless GF changed multi-turn move mechanics, there should be no way to Dynamax halfway through Outrage or Dive.

KrisXV:
Not sure but it's a temporary fix to battles that cannot progress

PR #6052, opened 2019-11-22
Update format aliases for gen 8

TheMezStrikes:


TheImmortal:
There are more aliases that could be added. I'll edit this later.

PR #6054, opened 2019-11-23
G-Max Snooze shouldn't affect already statused foes

urkerab:
Reported somewhere, but I forget where, sorry...


Zarel:
Maybe this check should go into the Yawn `onStart`?

urkerab:
I think it needs to happen before Abilities trigger, unfortunately.

Zarel:
Fair.

PR #6055, opened 2019-11-23
Update moves' shortDesc

PartMan7:
NOTE: There are a lot of moves that have inconsistent / overflowing shortDescs, will be fixing them as I find them. 

Gonna try and tackle most of the overflowing G-Max move descriptions.


PartMan7:
At the moment, G-Max Steelsurge is the only one I haven't been able to fit in the bounds. Been wondering if ``Sets Steel entry hazard. BP scales w/ base move.`` is acceptable.

Zarel:
Works for me.

data/moves.js
```diff
@@ -7344,7 +7344,7 @@ let BattleMovedex = {
 		accuracy: true,
 		basePower: 10,
 		category: "Physical",
-		shortDesc: "Sets Steel-type entry hazard. BP scales with base move's BP.",
+		shortDesc: "Sets Steel entry hazard. BP scales with base move's BP.",
```
```suggestion
		shortDesc: "Sets Steel entry hazard. BP scales w/ base move.",
```

PR #6064, opened 2019-11-24
Update disguise shortDesc

Enrii17:


Zarel:
We still need the old desc for Gen 7. Can you add it in there?

data/abilities.js
```diff
@@ -744,7 +744,7 @@ let BattleAbilities = {
 	},
 	"disguise": {
 		desc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage. Its disguise is then broken, it changes to Busted Form, and it loses 1/10 of its max HP. Confusion damage also breaks the disguise.",
-		shortDesc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage.",
+		shortDesc: "If this Pokemon is a Mimikyu, first hit 0 damage. Breaks disguise losing 10% HP",
```
```suggestion
		shortDesc: "(Mimikyu only) The first hit it takes is blocked, and it takes 10% HP damage instead.",
```

Enrii17: Sure, do we need it for every ability that changed in Gen 8 or just Disguise?

Zarel: We've already done it for other Abilities:

https://dex.pokemonshowdown.com/abilities/moody

Zarel: Feel free to do it for any ability we've missed, though?

Zarel:
You still haven't fixed it in past gens...

Enrii17:
> You still haven't fixed it in Gen 7...

oh sorry, my internet was extremely bad past few days and I forgot about past gens, will do it now

data/mods/gen7/abilities.js
```diff
@@ -5,6 +5,7 @@ let BattleAbilities = {
 	"disguise": {
 		inherit: true,
 		desc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage. Its disguise is then broken and it changes to Busted Form. Confusion damage also breaks the disguise.",
+		shortDesc: "(Mimikyu only) First hit deals 0 damage, breaks disguise.",
```
Zarel:
```suggestion
		shortDesc: "(Mimikyu only) The first hit it takes is blocked and deals 0 damage instead.",
```

data/abilities.js
```diff
@@ -744,7 +744,7 @@ let BattleAbilities = {
 	},
 	"disguise": {
 		desc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage. Its disguise is then broken, it changes to Busted Form, and it loses 1/10 of its max HP. Confusion damage also breaks the disguise.",
```
```suggestion
		desc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage. Its disguise is then broken, it changes to Busted Form, and it loses 1/8 of its max HP. Confusion damage also breaks the disguise.",
```

data/abilities.js
```diff
@@ -744,7 +744,7 @@ let BattleAbilities = {
...
-		shortDesc: "If this Pokemon is a Mimikyu, the first hit it takes in battle deals 0 neutral damage.",
+		shortDesc: "(Mimikyu only) The first hit it takes is blocked, and it takes 10% HP damage instead.",
```
```suggestion
		shortDesc: "(Mimikyu only) The first hit it takes is blocked, and it takes 1/8 HP damage instead.",
```

PR #6067, opened 2019-11-25
Shorten Ice Scales shortDesc

PartMan7:
Overflowing at the moment. 

Also, what's the policy on descriptive abilities?
(stuff like ``This Pokemon is protected by ice scales, which`` which provide a description and can be removed without leaving the effects out of context)

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -1597,7 +1597,7 @@ let BattleAbilities = {
 		num: 248,
 	},
 	"icescales": {
-		shortDesc: "This Pokémon's Sp. Defense is doubled.",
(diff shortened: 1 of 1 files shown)


data/abilities.js
```diff
@@ -1597,7 +1597,7 @@ let BattleAbilities = {
 		num: 248,
 	},
 	"icescales": {
-		shortDesc: "The Pokémon is protected by ice scales, which halve the damage taken from special moves.",
+		shortDesc: "This Pokémon's Sp. Defense is doubled.",
```
```suggestion
		shortDesc: "This Pokémon's Special Defense is doubled.",
```

PR #6068, opened 2019-11-25
Random Battle: Update Weezing-Galar

ascriptmaster:


TheImmortal:
I mean, it could probably just be rejected outright because Fini is dead.

TheImmortal:
I'll do it in my big SwSh update.

PR #6069, opened 2019-11-25
Add some Gen 8 aliases

Spandamn:


Zarel:
This should be hardcoded, like "Mega".

sim/dex.ts
```diff
@@ -363,15 +363,25 @@ export class ModdedDex {
...
-				aliasTo = id.slice(4) + 'mega';
-			} else if (id.startsWith('m') && this.data.Pokedex[id.slice(1) + 'mega']) {
-				aliasTo = id.slice(1) + 'mega';
-			} else if (id.startsWith('primal') && this.data.Pokedex[id.slice(6) + 'primal']) {
-				aliasTo = id.slice(6) + 'primal';
-			} else if (id.startsWith('p') && this.data.Pokedex[id.slice(1) + 'primal']) {
-				aliasTo = id.slice(1) + 'primal';
-			}
+			const formeNames = {
+				a: 'alola',
```
KrisXV: why not just do `alola: ['a', 'alolan']` etc

sim/dex.ts
```diff
@@ -363,15 +363,25 @@ export class ModdedDex {
...
+				g: 'galar',
+				galarian: 'galar',
+				gigantamax: 'gmax',
+				gmax: 'gmax',
+				m: 'mega',
+				mega: 'mega',
+				p: 'primal',
+				primal: 'primal',
+			};
+			Object.keys(formeNames).forEach(forme => {
```
KrisXV: theres no reason to use forEach

Spandamn: i thought it was simpler, how would you do it

KrisXV: `for (const formeName of Object.keys(formeNames))` or just using `for (const i in formeNames)`

sim/dex.ts
Spandamn: isolating the pokemon name if i do that would be complicated

Spandamn:
Halp with TS

sim/dex.ts
```diff
@@ -363,14 +363,26 @@ export class ModdedDex {
...
-				aliasTo = id.slice(6) + 'primal';
-			} else if (id.startsWith('p') && this.data.Pokedex[id.slice(1) + 'primal']) {
-				aliasTo = id.slice(1) + 'primal';
+			const formeNames = {
+				a: 'alola'
+				alolan: 'alola',
+				g: 'galar',
+				galarian: 'galar',
+				gigantamax: 'gmax',
+				gmax: 'gmax',
```
KrisXV: gmax === gmax ???

sim/dex.ts
```diff
@@ -363,14 +363,26 @@ export class ModdedDex {
...
+				m: 'mega',
+				mega: 'mega',
```
KrisXV: mega === mega ???

sim/dex.ts
```diff
@@ -363,14 +363,26 @@ export class ModdedDex {
...
+				p: 'primal',
+				primal: 'primal',
```
KrisXV: primal === primal ???

sim/dex.ts
Spandamn: wrote that for startsWith
currently if you use `/dt gmax charizard` it wont work

Spandamn:
image

sim/dex.ts
```diff
@@ -363,14 +363,29 @@ export class ModdedDex {
...
+			};
+			for (const forme in formeNames) {
+				if (id.startsWith(forme)) {
+					let pokeName = id.slice(forme.length);
```
KrisXV: the errors are because you dont have `pokeName` declared anywhere outside of this if statement, declare it above the if statement as
```typescript
let pokeName = '' as string;
```

Spandamn: ty i missed that

sim/dex.ts
Zarel: ?

Zarel: It shouldn't make a difference for that.

Spandamn:
That was the last commit i swear. This okay now @Zarel ?

sim/dex.ts
```diff
@@ -363,14 +363,27 @@ export class ModdedDex {
...
+				mega: ['m', 'mega'],
+				primal: ['p', 'primal'],
+			};
+			for (const forme in formeNames) {
+				let pokeName = '';
+				for (const i of formeNames[forme]) {
+					if (id.startsWith(i)) {
+						pokeName = id.slice(i.length);
+					} else if (id.endsWith(i)) {
+						pokeName = id.slice(0, id.length - i.length);
```
```suggestion
						pokeName = id.slice(0, -i.length);
```

sim/dex.ts
```diff
@@ -363,14 +363,27 @@ export class ModdedDex {
...
-			} else if (id.startsWith('p') && this.data.Pokedex[id.slice(1) + 'primal']) {
-				aliasTo = id.slice(1) + 'primal';
+			const formeNames: {[k: string]: string[]} = {
+				alola: ['a', 'alola', 'alolan'],
+				galar: ['g', 'galar', 'galarian'],
+				gmax: ['gigantamax', 'gmax'],
+				mega: ['m', 'mega'],
+				primal: ['p', 'primal'],
+			};
+			for (const forme in formeNames) {
```
Zarel: I don't like this algorithm. It fails on Mega Altaria, since that will be converted to `megaaltarialola` and then fail to show up.

Zarel: Oh, wait, you don't have early breaks. You have the opposite problem: `alolangolem` will be converted to `alolangolemega` and then fail.

Zarel: Oh, wait, you are testing them all separately.

Zarel: Yeah, this is fine. Non-ideal, but it won't have any weird misbehavior.

Spandamn: So this is good to go right

PR #6071, opened 2019-11-25
Some Movesearch Changes

Spandamn:
- Allow searching for max and gmax moves
- Allow searching for protection moves
image


Spandamn:
@Zarel

PR #6072, opened 2019-11-25
Do not give Clangorous Soul boost below 33%

tmagicturtle:
Clangorous Soul is supposed to fail when used below 1/3 max HP but doesn't. Sorry - I'm not sure this fixes it; I'm unable to test.

https://replay.pokemonshowdown.com/gen8customgame-1016802018

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -2556,18 +2556,12 @@ let BattleMovedex = {
 		pp: 5,
 		priority: 0,
 		flags: {snatch: 1, sound: 1, dance: 1},
-		onHit(target) {
+		onHit(target, source, move) {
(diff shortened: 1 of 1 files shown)


Zarel:
This should definitely be fixed in the engine instead of here.

Zarel:
Also, why can't you test?

The simplest test for this kind of thing is just to export your input log, and then import it to `./pokemon-showdown simulate-battle`.

tmagicturtle:
The latest Node package available for my OS is v8.10.0. A lot of PS dependencies don't work on my laptop. I could look into using a VPS, I suppose.

Zarel:
Is your OS a Linux? If so, this will get you an up-to-date copy of Node:

```
npm i n
n latest
```

Zarel:
(BSDs are more awkward, but I think BSDs have a more recent version of Node anyway.)

PR #6076, opened 2019-11-26
Clarify doubles interactions for Max moves

ZardMX:


ZardMX:
PS: PartMan helped because my computer died.

Zarel:
Should probably have a regular description that isn't so compressed. Can you do that for me?

PR #6079, opened 2019-11-26
Add aliases for max moves

TheMezStrikes:


TheMezStrikes:
Oops that's probably some trailing whitespace

PR #6080, opened 2019-11-26
Fix type-changing moves & abilities change while Dynamaxed

420Blazeitt:
Probably not the cleanest way to do it but it works ^^'


Zarel:
I'm not particularly fond of how hardcoded this is.

Zarel:
Do these all run the same type-changing event?

420Blazeitt:
yeah i’m sorry, uh. They all have onModifyMove. But that’s about it

data/scripts.js
```diff
@@ -161,6 +161,15 @@ let BattleScripts = {
...
 			move = this.getActiveZMove(move, pokemon);
 		}
+		if (pokemon.hasAbility('pixilate') || pokemon.hasAbility('refrigerate') || pokemon.hasAbility('galvanize') || pokemon.hasAbility('normalize') || pokemon.hasAbility('liquidvoice') && maxMove) {
+			this.singleEvent('ModifyMove', pokemon.getAbility(), null, pokemon, target, pokemon.getAbility(), move);
+			sourceEffect = move;
+		}
+		if (['aurawheel', 'hiddenpower', 'judgment', 'multiattack', 'naturalgift', 'revelationdance', 'technoblast', 'weatherball'].includes(move.id) && maxMove) {
+			// Max moves change type when used directly
+			this.singleEvent('ModifyMove', move, null, pokemon, target, move, move);
+			if (move.type !== 'Normal') sourceEffect = move;
```
KrisXV: I didn't know Aura Wheel was a Normal-type type-changing move

420Blazeitt:
Alright there is no hardcoding used in this way, but still works just as well!

Zarel:
I'd prefer a separate `ModifyType` event, so we don't have to go through the full `ModifyMove` rigamarole.

420Blazeitt:
yeah that’s a good idea

420Blazeitt:
I haven't typed anything properly, will get to that later

data/scripts.js
```diff
@@ -155,12 +155,19 @@ let BattleScripts = {
...
-			this.singleEvent('ModifyMove', move, null, pokemon, target, move, move);
+			this.singleEvent('ModifyType', move, null, pokemon, target, move, move);
 			if (move.type !== 'Normal') sourceEffect = move;
 		}
 		if (zMove || (move.category !== 'Status' && sourceEffect && /** @type {ActiveMove} */(sourceEffect).isZ)) {
 			move = this.getActiveZMove(move, pokemon);
 		}
+		if (maxMove && move.category !== 'Status') {
+			let moveType = move.type;
+			// Max move outcome is dependant on the move type after type modifications from ability and the move itself
```
Zarel:
```suggestion
			// Max move outcome is dependent on the move type after type modifications from ability and the move itself
```

Zarel:
This isn't the signature I'd use, but it's a good first step. I'd merge it after the spelling fix.

420Blazeitt:
So, I made the typo fix. ;w;

Amaterasu5:
Hey! sorry to revive an old thread but this doesn't appear to be fixed in the damage calc. Using max weather ball in the sun from torkoal, for example, registers in the calc as max strike without the damage or sun boost. I've attached an image:
Screen Shot 2020-08-07 at 9 53 02 PM

Zarel:
@Amaterasu5 you should report that in the damagecalc:

https://github.com/smogon/damage-calc

Amaterasu5:
@Zarel will do, thanks!

PR #6081, opened 2019-11-26
Add Eternamax forme for Eternatus

Spandamn:


Spandamn:
The Eternamax forme will act like a permanent Dynamax. Behemoth blade, Behemoth Bash and Dynamax Cannon will deal double damage like they do to normal dynamax pokemon. The Dynamax volatile is not cleared when it is switched out. Eternatus-Eternamax can only use Max Moves.

Spandamn:
This can only be used in custom game btw

data/statuses.js
```diff
@@ -718,6 +718,14 @@ let BattleStatuses = {
...
 		duration: 3,
 		onStart(pokemon) {
+			if (pokemon.species === 'Eternatus-Eternamax') { // Special for Eternatus' Eternamax forme
+				this.add('-start', pokemon, 'Eternamax');
+				pokemon.maxhp = Math.floor(pokemon.maxhp * 2); // Assuming Eternamax is max level
+				pokemon.hp = Math.floor(pokemon.hp * 2);
+				this.add('-heal', pokemon, pokemon.getHealth, '[silent]');
+				this.effectData.duration = 0;
+				return;
+			}
```
HoeenCoder: This can be simplified alot if we even want to do this. Would be easy enough to just so something like
if (isEternamax) {
  add start message
  set duration to 0
} else {
  standard start message
}
rest of dynamax code which you have duplicated.

data/statuses.js
```diff
@@ -718,6 +718,14 @@ let BattleStatuses = {
 		num: 0,
 		duration: 3,
 		onStart(pokemon) {
+			if (pokemon.species === 'Eternatus-Eternamax') { // Special for Eternatus' Eternamax forme
+				this.add('-start', pokemon, 'Eternamax');
```
HoeenCoder: Do we need a special start message for emax?

data/formats-data.js
```diff
@@ -7854,6 +7854,10 @@ let BattleFormatsData = {
 		tier: "Uber",
 		doublesTier: "DUber",
 	},
+	eternatuseternamax: {
+		isNonstandard: "Eternamax",
```
TheImmortal: Just use Custom

data/learnsets.js
```diff
@@ -68873,6 +68873,50 @@ let BattleLearnsets = {
 		venomdrench: ["8M"],
 		venoshock: ["8M", "8L16"],
 	}},
+	eternatuseternamax: {learnset: {
```
TheImmortal: Does it actually have learnset differences

KrisXV: No

Zarel:
Is this how it works in Hackmons? If not, it shouldn't work this way.

data/statuses.js
Spandamn: In game it just shows "Eternatus appeared" iirc, ill remove this

Spandamn:
> Is this how it works in Hackmons? If not, it shouldn't work this way.
It can't be hacked ingame, its only supposed to be allowed for Custom Games

data/statuses.js
Spandamn: I'm adding one just so it can differentiate

Spandamn:
Tested it out, should work

Spandamn:
What the FUCK is that supposed to mean lmao

Spandamn:
Okay that _should_ be it

data/statuses.js
```diff
@@ -728,8 +729,23 @@ let BattleStatuses = {
...
+			this.add('-heal', pokemon, pokemon.getHealth, '[silent]');
+		},
 		onFlinch: false,
 		onBeforeSwitchOut(pokemon) {
+			if (pokemon.species === 'Eternatus-Eternamax') {
+				pokemon.maxhp = Math.floor(pokemon.maxhp / 2); // TODO prevent maxhp loss
+				pokemon.hp = Math.floor(pokemon.hp / 2);
+				if (pokemon.hp <= 0) pokemon.hp = 1;
+				return;
+			}
```
HoeenCoder: Why not just do `if (pokemon.species === 'Eternatus-Eternamax') return;` right before `this.hint("Dynamax ended.");`?

data/statuses.js
```diff
@@ -744,6 +760,12 @@ let BattleStatuses = {
 			return null;
 		},
 		onEnd(pokemon) {
+			if (pokemon.species === 'Eternatus-Eternamax') {
+				pokemon.maxhp = Math.floor(pokemon.maxhp / 2);
+				pokemon.hp = Math.floor(pokemon.hp / 2);
+				if (pokemon.hp <= 0) pokemon.hp = 1;
+				return;
+			}
 			this.add('-end', pokemon, 'Dynamax');
```
HoeenCoder: Again, remove the custom HP code and just skip this step if its emax.

data/learnsets.js
TheImmortal: Then there's no need for a learnset. Especially since this is only usable in Custom Game and everything is legal.

Zarel:
@Spandamn You forgot to remove the learnset.

data/statuses.js
```diff
@@ -728,6 +729,15 @@ let BattleStatuses = {
 			// TODO work on display for HP
 			this.add('-heal', pokemon, pokemon.getHealth, '[silent]');
 		},
+		onSwitchIn(pokemon) { // Putting Eternamax in onSwitchIn so it shows up eveytime Eternatus switches in.
```
TheImmortal: everytime*

Zarel: every time*

PR #6090, opened 2019-11-26
Fix CAP banlist

TheMezStrikes:


Zarel:
Should we just remove these moves from their learnsets, or is there a reason we're doing it this way?

TheMezStrikes:
I'm pretty sure CAP staff doesn't want to remove moves from learnsets because they're historically part of the pokemon, but I can talk to them about it. I'll get back to you

TheMezStrikes:
I have returned with an answer. It's for past gen compatibility. The moves were and are legal in gen 6 cap which is still played (I'd assume through custom game).

I think PS code supports doing this in learnsets.js nowadays though, so I'll port that over sooner or later. For now, I'd just merge this PR so the CAP players at least have their moves banned

Zarel:
No, there's no easy way to do this. The bans are probably for the best.

PR #6091, opened 2019-11-26
Don't display Weight for GMax Pokemon

PartMan7:
Alternatively, I can make it display `` Weight: 0 kg`` over `` Weight: 0 kg (20 BP)``.

### DIFF
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -558,8 +558,8 @@ const commands = {
 						"Dex#": pokemon.num,
 						"Gen": pokemon.gen || 'CAP',
 						"Height": pokemon.heightm + " m",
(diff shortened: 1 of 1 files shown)


KrisXV:
It would probably be better to display the weight as `Weight: 0 kg` since they still have a weight

Zarel:
maybe "(GK/LK fail)"

PartMan7:
Implemented. On the other hand, what are your thoughts on using something like ``Weightless`` instead of the Weight flag, and accordingly modifying the ``desc``s for weight-based moves to specify that they fail against Weightless Pokemon?

Zarel:
That seems uglier, to me.

PR #6092, opened 2019-11-26
Fix vicegrip regression

TZubiri:
Fixes https://github.com/smogon/pokemon-showdown/issues/6010

Do these files support comments? If so, a comment warning should be added to avoid this bug being reintroduced.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -20843,13 +20843,13 @@ let BattleMovedex = {
 		zMovePower: 100,
 		contestType: "Cool",
 	},
-	"visegrip": {
+	"vicegrip": {
 		num: 11,
 		accuracy: 100,
 		basePower: 55,
 		category: "Physical",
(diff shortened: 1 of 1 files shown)


PartMan7:
Just my thoughts, but reverting a fix to fix something else isn't really the best course of action. I'm trying to think of a fix to the matter without going back to how it was, since that doesn't solve anything.

TZubiri:
Maybe I'm missing something, but I've kept all of the changes you introduced, including keeping the display name as "Vise Grip". It's only the id that remains as vicegrip.  Is there some reason the id needs to be visegrip?

TZubiri:
Ah you are right, my bad. Even the manual smoke test I did was wrong, setting the team to Gen 8 reveals it doesn't even fix the original bug.
I have to test my pull requests more thoroughly. And run the npm test and fulltests locally.

PR #6096, opened 2019-11-27
Fix Aura Wheel's interaction with dynamaxing

KrisXV:
This will need client side adjustment as well


TZubiri:
This is also fixed here I believe.https://github.com/smogon/pokemon-showdown/pull/6080/commits/95b262051cb5daa4b575c86c11cb8362a3156711

KrisXV:
That code doesn't exist in scripts.js...

Zarel:
Specifically, it's fixed by #6080.

Zarel:
I'm not even accepting #6080; I'm definitely not accepting the hardcode of a single move.

PR #6097, opened 2019-11-27
Initialise the m property on Pokemon earlier

urkerab:
I was converting some old OMs to use the `m` property and some of them need it in `onModifyTemplate` which is called during the constructor, and theoretically OMs could override other Pokemon functions and would want `m` to already exist there too.


sim/pokemon.ts
```diff
@@ -214,6 +214,11 @@ export class Pokemon {
 		this.side = side;
 		this.battle = side.battle;
 
+		/**
+		 * An object for storing untyped data, for mods to use.
+		 */
```
Zarel: This should be on the `m: {...}` property description.

urkerab: Replacing the previous description I assume...

PR #6098, opened 2019-11-27
Add full desc for Max moves

ZardMX:
Question: Looks like Obstruct (protection move introduced in gen 8) is not included in the list of moves listed in other protection moves' full descriptions, do I add Obstruct in those lists in this PR as well or do I create a new one for it after this. If I should add it in this PR, what do I title this?


Zarel:
Thanks! Yeah, adding Obstruct is probably a good idea, but asking @Marty-D to make sure.

Issue #6070, opened 2019-11-25
Speed calculation under weather triggered abilities?


betairylia:
Hi, I know gen8 is currently under test. here's probably a problem:

https://replay.pokemonshowdown.com/gen8battlestadiumdoubles-1016567180
Turn 4 (I am sunny side)

Not checked the in-game behavior though, so if I have expected something weird please forgive me...
( I will try to check that in the game later.

Vanilluxe: scarf (guessed), maximum speed should be 216 (144 * 1.5);
However the Leafeon has the ability Chlorophyll, with 252EV (neutral nature) on Spe, resulting 294 Spe stat (147 * 2):

Leafeon @ Choice Band  
Ability: Chlorophyll  
Level: 50  
EVs: 252 Atk / 4 SpD / 252 Spe  
Adamant Nature  
- Solar Blade  
- Leaf Blade  
- Double Kick  
- Quick Attack

In this turn leafeon chose the move Solar Blade. ( not sure if it was related to this specific move, bcz it works fine in other match
So the expected behavior should be leafeon uses its solar blade before the blizzard, if I understood this generation's speed calculation correctly (it was calculated dynamically)
Not sure where goes wrong, just want to discuss (perhaps my head goes wrong sry)

Or, is it just because the lack of notication when Quick Claw was triggered?
I think it will play an animation in-game...

Thanks!

Zarel:
I'm surprised there wasn't already an issue about this. But yes, Speed is updated dynamically this gen, which has yet to be implemented.

betairylia:
Thanks for your reply! It was not implemented yet?
Do you mean weather abailities or including weather etc. ? Sometimes it feels (honestly at most time) that it was already implemented while using prankster tailwind.

HoeenCoder:
This was fixed in #6100

PR #6100, opened 2019-11-28
Dynamic speed updates for Gen 8

HoeenCoder:
This makes generation 8 use dynamically updating speed. Specifically speeds are updated after each action in the queue completes. Changes up to this point tested with @DaWoblefet.

TODO:
- [x] Fix failed tests (fusion bolt/flare, redirection in doubles)
- [x] Add tests to ensure dynamic speed works in gen 8 and prior speed system is used in gen 7.
- [x] Figure out whats going on with this line is doing. https://github.com/smogon/pokemon-showdown/blob/master/sim/side.ts#L694 Specifically the priority being lower (<= 0) than start (102) causes problems if the queue is dynamically updated when `battle.turn === 0`.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -284,6 +284,7 @@ let BattleMovedex = {
 			let action = this.willMove(target);
 			if (action) {
 				this.cancelMove(target);
+				action.priority = 7.1;
 				this.queue.unshift(action);
 				this.add('-activate', target, 'move: After You');
 			} else {
--- a/data/moves.js
+++ b/data/moves.js
@@ -283,9 +283,7 @@ let BattleMovedex = {
 			if (target.side.active.length < 2) return false; // fails in singles
 			let action = this.willMove(target);
 			if (action) {
-				this.cancelMove(target);
-				action.priority = 7.1;
-				this.queue.unshift(action);
+				this.prioritizeAction(action);
 				this.add('-activate', target, 'move: After You');
 			} else {
 				return false;
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2502,7 +2502,7 @@ export class Battle {
 			}
 		}
 		action.sourceEffect = sourceEffect;
-		action.priority = 7.1;
+		action.priority = 10;
 		this.queue.unshift(action);
 	}
 
@@ -2738,7 +2738,7 @@ export class Battle {
 			// in gen 3 or earlier, switching in fainted pokemon is done after
 			// every move, rather than only at the end of the turn.
 			this.checkFainted();
-		} else if (['megaEvo', 'runDynamax'].includes(action.choice) && this.gen >= 7) {
+		} else if (action.choice === 'megaEvo' && this.gen === 7) {
 			this.eachEvent('Update');
 			// In Gen 7, the action order is recalculated for a Pokémon that mega evolves.
 			const moveIndex = this.queue.findIndex(queuedAction =>
@@ -2778,6 +2778,10 @@ export class Battle {
 		}
 
 		this.eachEvent('Update');
+		if (this.gen >= 8 && this.queue.length && this.queue[0].choice === 'move') {
+			this.updateSpeed();
+			this.sortQueue();
+		}
 
 		return false;
 	}
@@ -2797,12 +2801,6 @@ export class Battle {
 			this.queue.shift();
 			this.runAction(action);
 			if (this.requestState || this.ended) return;
-
-			if (this.gen >= 8 && this.turn > 0) {
-				// Dynamically updated speed mid-turn except during team selection.
-				this.updateSpeed();
-				this.sortQueue();
-			}
 		}
 
 		this.nextTurn();
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -308,13 +308,6 @@ export class Battle {
 		for (const pokemon of this.getAllActive()) {
 			pokemon.updateSpeed();
 		}
-
-		if (this.gen < 8) return;
-
-		// In gen 8, speed is updated dynamically so update the queue's speed properties.
-		for (const action of this.queue) {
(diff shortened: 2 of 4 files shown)


DaWoblefet:
Interactions we tested:

- Switch to rain -> Swift Swim 201 Speed moves before 401 Speed.
- Prankster Tailwind -> partner 201 Speed moves before 401 Speed.
- Speed tie 200 Speed vs. opposing 200 Speed -> win tie -> partner 101 Speed moves before opponent 200 Speed.
- After You -> Trick Room -> 200 Speed moves before 400 Speed.
- Round/Pledges/Quash/After You/Pursuit is functioning in Gen 7/8
- Faster Intimidate -> proc 156 Speed Adrenaline Orb -> 156 Snow Warning -> 157 Intimidate (I assume this is how this one would behave on cartridge, have not tested / seen footage).

More that should be tested that I of course am only thinking of after the fact:
- Choice Scarf Dynamax uses Choice Scarf for the purposes of Dynamaxing order but not turn order (need to test on cart to confirm)
- Insert other ideas here

Zarel:
...I'm working on this right now.

Zarel:
```diff
diff --git a/sim/battle.ts b/sim/battle.ts
index b80bfa574..ed64d0bac 100644
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2731,7 +2731,7 @@ export class Battle {
                        // in gen 3 or earlier, switching in fainted pokemon is done after
                        // every move, rather than only at the end of the turn.
                        this.checkFainted();
-               } else if (['megaEvo', 'runDynamax'].includes(action.choice) && this.gen >= 7) {
+               } else if (action.choice === 'megaEvo' && this.gen === 7) {
                        this.eachEvent('Update');
                        // In Gen 7, the action order is recalculated for a Pokémon that mega evolves.
                        const moveIndex = this.queue.findIndex(queuedAction =>
@@ -2771,6 +2771,9 @@ export class Battle {
                }

                this.eachEvent('Update');
+               if (this.gen >= 8 && this.queue.length && this.queue[0].choice === 'move') {
+                       this.sortQueue();
+               }

                return false;
        }
```

is all it needs

sim/battle.ts
```diff
@@ -2790,6 +2797,12 @@ export class Battle {
 			this.queue.shift();
 			this.runAction(action);
 			if (this.requestState || this.ended) return;
+
+			if (this.gen >= 8 && this.turn > 0) {
+				// Dynamically updated speed mid-turn except during team selection.
+				this.updateSpeed();
+				this.sortQueue();
+			}
```
Zarel: This should be moved to the end of `runAction`.

Also, for performance reasons, you might as well just do this:

```
 if (this.gen >= 8 && this.queue.length && this.queue[0].choice === 'move') {
```

sim/battle.ts
```diff
@@ -2496,6 +2502,7 @@ export class Battle {
 			}
 		}
 		action.sourceEffect = sourceEffect;
+		action.priority = 7.1;
```
Zarel: Maybe something significantly higher, like 10? Unless there's a reason for this.

Zarel: Also, with my optimization, this doesn't actually need to happen afaict.

Zarel: Oh, nvm, it is move actions being prioritized. Yeah, this is necessary.

data/moves.js
```diff
@@ -284,6 +284,7 @@ let BattleMovedex = {
 			let action = this.willMove(target);
 			if (action) {
 				this.cancelMove(target);
+				action.priority = 7.1;
 				this.queue.unshift(action);
```
Zarel: Should this just use `prioritizeAction`?

Zarel:
>  Figure out whats going on with this line is doing. https://github.com/smogon/pokemon-showdown/blob/master/sim/side.ts#L694 Specifically the priority being lower (<= 0) than start (102) causes problems if the queue is dynamically updated when battle.turn === 0.

It puts the Pokémon in the order specified by Team Preview.

Zarel:
The problem with the Fusion Bolt/Flare test is that the RNG rerolls cause Zekrom rather than Reshiram to win the Speed tie.

Zarel:
Fixes #6070

Should probably note that.

Zarel:
The problem with the redirection test is that one of the Wailords gets Confused from Water Pulse, changing the test results.

Zarel:
I fixed your tests in a00d64a092f36f2728a3; just rebase on top of master.

Zarel:
Your last failing test is annoying so I'll fix it after you rebase.

data/moves.js
HoeenCoder: Probably, I'll test it.

sim/battle.ts
HoeenCoder: I set the priority level to 10 in prioritizeAction

sim/battle.ts
```diff
@@ -2771,6 +2778,10 @@ export class Battle {
 		}
 
 		this.eachEvent('Update');
+		if (this.gen >= 8 && this.queue.length && this.queue[0].choice === 'move') {
+			this.updateSpeed();
```
Zarel: I would do your `action.speed = action.pokemon.speed` loop here rather than inside `updateSpeed`.

Zarel:
I like the Water Gun fix more than the Own Tempo fix; feel free to fix mine to also use Water Gun.

urkerab:
Well, that didn't quite work out, due to the Triples test, but whatever.

HoeenCoder:
I think its overall OK. That one should be safe from RNG due to it only checking if gastrodon's spa stat was boosted (storm drain).

urkerab:
A bug report regarding After You:

https://www.smogon.com/forums/posts/8300643

PR #6102, opened 2019-11-28
Update Repo Links

Spandamn:
Obviously had to write a script for this


Zarel:
> write a script

For the record, find-and-replace is built into VS Code, no scripting necessary. :p

PR #6104, opened 2019-11-28
Update G-Max Wildfire' desc

DeltaCoderr:
It shows in Bulbapedia that G-Max Wildfire only deals damage on non-fire type pokemon.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -7423,7 +7423,7 @@ let BattleMovedex = {
 		accuracy: true,
 		basePower: 10,
 		category: "Physical",
-		desc: "Damages opponent(s) by 1/8 of their maximum HP for four turns. Base Power scales with the base move's Base Power.",
+		desc: "Damages opponent(s) by 1/8 of their maximum HP for four turns on non-Fire-type Pokemon. Base Power scales with the base move's Base Power.",
 		shortDesc: "Damages foes for 4 turns. BP scales w/ base move.",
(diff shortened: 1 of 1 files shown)


DeltaCoderr:
Here - https://m.bulbapedia.bulbagarden.net/wiki/G-Max_Move

whalemer:
If we have proper confirmation this is correct, we should probably update the effect too.

DeltaCoderr:
Yes.

KrisXV:
I would just copy the descriptions from Fire Pledge and edit accordingly to remove the grass pledge and water pledge events.

And also the effect.

Zarel:
> Yes.

I can't tell if this means you're planning on doing it or not?

Zarel:
Also yes I'd like proper confirmation, too.

DeltaCoderr:
Sorry to say but, I am not that familiar with adding the code for effects and I am a new person in the development of PS and still learning. Perhaps can you guys do it please?

Marty-D:
It's not the same effect as the Pledge combo, it has a different message and deals 1/6 instead of 1/8. Check the research thread.

Zarel:
@Marty-D the research thread is like 20 pages long; can you just tell me?

Marty-D:
I mean the first post, which I keep updating with relevant info and a link to the post where it was confirmed. https://www.smogon.com/forums/posts/8283555/

PR #6108, opened 2019-11-28
Zmove, Megas, and Primals Pokemon Should Not Dynamax

YaBoiJD:


data/scripts.js
```diff
@@ -1172,6 +1172,10 @@ let BattleScripts = {
 			if (!pokemon.side.canDynamax) return;
 			if (this.canZMove(pokemon)) return;
 			if (this.canMegaEvo(pokemon)) return;
+			if (pokemon.getItem().zMove) return;
```
Zarel: This is redundant with `this.canZMove(pokemon)`

data/scripts.js
```diff
@@ -1172,6 +1172,10 @@ let BattleScripts = {
 			if (!pokemon.side.canDynamax) return;
 			if (this.canZMove(pokemon)) return;
```
Zarel: This does reveal the actual problem, though: Once a Pokémon has used its Z-move or mega evo, it's no longer governed by these checks and can Dynamax.

data/scripts.js
YaBoiJD: `this.canZMove(pokemon)` does not check after it has used its Z-Move

HoeenCoder: Get rid of `canZMove` basically, you dont need to check if they are still able to use a Z move in general, just if they have a z crystal.

Zarel: This seems wrong to me; we might want to code the restriction in rulesets if there's no engine reason for it.

data/scripts.js
```diff
@@ -1174,7 +1174,6 @@ let BattleScripts = {
 			if (pokemon.getItem().zMove) return;
 			if (pokemon.template.isMega) return;
 			if (pokemon.template.isPrimal) return;
-			if (pokemon.transformed) return;
```
YaBoiJD: Transformed pokemon can only Dynamax, it can't Gigantamax, in a result of it not being able to Gigantamax, it cannot use the G-Max moves, just the regular Max Moves

data/statuses.js
```diff
@@ -716,6 +716,7 @@ let BattleStatuses = {
 		name: 'Dynamax',
 		id: 'dynamax',
 		num: 0,
+		noCopy: true,
```
YaBoiJD: prevents copying by transform/baton pass

HoeenCoder:
Also refactored side.canDynamax -> pokemon.canDynamax to support these changes.

data/rulesets.js
```diff
@@ -714,9 +714,9 @@ let BattleFormats = {
 			return -typeMod;
 		},
 	},
-	natdex: {
+	natdexrule: {
 		effectType: 'Rule',
-		name: 'NatDex',
+		name: 'NatDex Rule',
```
Zarel: Did this really need to be renamed?

HoeenCoder: I made the decision to rename this instead of remove the alias since I figured "natdex" would be more useful as a format alias than as a ruleset name.

Zarel: We've never called anything "rule", though. Why not NatDex Clause or NatDex Mod?

HoeenCoder: No specific reason other than I thought that would be fine, I don't mind it being changed.

PR #6110, opened 2019-11-29
G-Max moves should run their effects against a substitute

urkerab:
Reported here: https://www.smogon.com/forums/posts/8297191 and other places.

All (G-)Max effects need to run as `self` so that they're unaffected by the target's Substitute.

Since self effects don't identify the target, this means that anything that effects the foes has to explicitly use `source.side.foe`. To make this clearer, I've updated all of the effects for all (G-)Max moves to only use a single `source` parameter, as I'm concerned that someone might subsequently try to use `target` by mistake.

Sandaconda's G-Max move appears not to match its description; Anubis has it on his to-do list of things to check. In the mean time I changed it so something nearer its description.

I also changed G-Max Wind Rage to use itself as the effect that's removing hazards, as that seemed wrong to me.

Let me know if you want any of those changes reverted.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -6818,17 +6818,19 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {},
 		isMax: "Butterfree",
-		onHit(target, source) {
-			for (let pokemon of target.side.active) {
-				let result = this.random(3);
-				if (result === 0) {
-					pokemon.trySetStatus('slp', source);
-				} else if (result === 1) {
-					pokemon.trySetStatus('par', source);
-				} else {
-					pokemon.trySetStatus('psn', source);
+		self: {
+			onHit(source) {
+				for (let pokemon of source.side.foe.active) {
+					let result = this.random(3);
+					if (result === 0) {
+						pokemon.trySetStatus('slp', source);
+					} else if (result === 1) {
+						pokemon.trySetStatus('par', source);
+					} else {
+						pokemon.trySetStatus('psn', source);
+					}
 				}
-			}
+			},
 		},
 		target: "adjacentFoe",
 		type: "Bug",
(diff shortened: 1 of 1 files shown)


data/moves.js
```diff
@@ -6818,17 +6818,19 @@ let BattleMovedex = {
...
-				let result = this.random(3);
-				if (result === 0) {
-					pokemon.trySetStatus('slp', source);
-				} else if (result === 1) {
-					pokemon.trySetStatus('par', source);
-				} else {
-					pokemon.trySetStatus('psn', source);
+		self: {
+			onHit(source) {
+				for (let pokemon of source.side.foe.active) {
```
Zarel: Is this true? Has anyone tried using this on an ally?

data/moves.js
```diff
@@ -6848,10 +6850,12 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {},
 		isMax: "Centiskorch",
-		onHit(target, source) {
-			for (let pokemon of target.side.active) {
-				pokemon.addVolatile('partiallytrapped');
-			}
+		self: {
+			onHit(source) {
+				for (let pokemon of source.side.foe.active) {
```
Zarel: ditto

data/moves.js
urkerab: Max Moves can't target allies.

TheImmortal:
The ones that set status such as Snooze only do so on the target and not every foe. How does one implement that under self? I tried a few ideas but didn’t work.

urkerab:
@TheImmortal Sorry for the late reply. Zarel got around to fixing Snooze in 8e5fe17, but for the record, you basically have to duplicate the code in `onHit` and `onAfterSubDamage`.

PR #6112, opened 2019-11-29
Update G-Max Gravitas' shortDesc

DeltaCoderr:


KrisXV:
I don't think this is necessary

Zarel:
Does this message even fit?

Zarel:
No, but this fits:

image

I don't know if we care about this, though.

KrisXV:
If we wanted to be sticklers about it, we would also need to update Max Flare's description on top of all other moves that call terrains/weathers/field effects

DeltaCoderr:
So  do you want me to change it to `Summons Gravity (5 turns)`?

Zarel:
It depends on what we agree on. @Marty-D, opinions?

Marty-D:
I also don't think this is necessary. You're supposed to read it as "this starts the effect of Gravity, the move, which lasts 5 turns", just like effects referencing Rain Dance are implied to last 5 turns or 8 turns if the effect starter was holding Damp Rock, because that's how Rain Dance the move works. There's no space in short descriptions to be wordy like this.

PR #6115, opened 2019-11-30
Natdex: Power Construct is banned

asgdf:
As per https://www.smogon.com/forums/threads/national-dex-ou-metagame-discussion.3656899/.

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -324,7 +324,7 @@ let Formats = [
 
 		mod: 'gen8',
 		ruleset: ['Obtainable', 'Standard', 'Team Preview', '+Past', '+Unreleased', 'NatDex'],
-		banlist: ['Uber', 'Moody', 'Power Construct', 'Shadow Tag', 'Baton Pass',
+		banlist: ['Uber', 'Power Construct', 'Shadow Tag', 'Baton Pass',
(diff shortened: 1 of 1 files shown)


asgdf:
Ok maybe Moody is not supposed to be baned, the OP just still says it is because it wasn't updated? Don't merge quite yet I guess.

Zarel:
You should be banning Zygarde-C, not Power Construct.

asgdf:
Gen 7 OU, Gen 7 DOU and Metronome Battles all ban Power Construct, while no format bans Zygarde-C, so I went with consistency.

Zarel:
:[ Fine, I guess.

Also I thought Zygarde was banned in Gen 7 OU? The Power Construct ban seems unnecessary.

Marty-D:
Zygarde 10% also exists.

asgdf:
Okay, so they didn't want Moody banned after all, and updated the OP accordingly, so I removed that ban again.
I did this through the online interface so I can't amend the commit easily, but that can be quashed when merging right?

Zarel:
This conflicts with the Dyna+Mega+Z fix I think.

asgdf:
Not anymore.

PR #6116, opened 2019-11-30
Galar Beginnings: Allow Snorlax-Gmax

LegoFigure11:
Snorlax-Gmax will be released before the competition starts on cartridge and therefore allowed in the competition, so it'd be great if we could get it unbanned here for testing before the comp starts.

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -170,6 +170,7 @@ let Formats = [
 			battle: 3,
 		},
 		ruleset: ['Obtainable', 'Standard GBU'],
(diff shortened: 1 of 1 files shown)


TheImmortal:
You can unban unreleased Pokemon? That doesn't seem right.

Zarel:
> You can unban unreleased Pokemon? That doesn't seem right.

What's wrong with that? I added that feature a few months ago with the validator refactor.

TheImmortal:
TIL.

PR #6117, opened 2019-11-30
Typescript roomsettings

whalemer:


server/chat.ts
```diff
@@ -555,7 +555,7 @@ export class CommandContext extends MessageContext {
 	checkBanwords(room: BasicChatRoom | null | undefined, message: string): boolean {
 		if (!room) return true;
 		if (!room.banwordRegex) {
-			if (room.banwords && room.banwords.length) {
+			if (room.banwords.length) {
```
Zarel: :[

I'm not sure _enough_ of PS is TypeScripted for me not to be worried about this even if it's unnecessary.

whalemer: Well, if we make this change, we can find out!

(agreed, reverted that)

Zarel: I mean, if it weren't right after the release of a new gen, I'd be interested in finding out. I do want to make an "unnecessary check" pass, but unfortunately this isn't Swift and "auto-force-unwrapped optionals" isn't a thing.

whalemer: tbh I'm half surprised typescript isn't yelling at me for the `room.banwords` check, given that as far as it knows that should always be true.

Zarel: I'm more surprised LGTM and tslint aren't. I guess "automatically force-unwrapped optionals" are common enough that you can't just blanket-ID all such cases as useless.

It _is_ required for writing polyfills, after all.

PR #6118, opened 2019-11-30
Refactor tournament plugin loading

whalemer:


Zarel:
I'm confused what this achieves?

Zarel:
It doesn't seem simpler at a glance. Can you clarify what improvement this makes?

whalemer:
Feels less hacky than the current situation to me, makes it a lot easier to load things other than commands (ie, roomsettings) from tours.

Zarel:
Oh, I see. Yeah, this is good.

PR #6119, opened 2019-11-30
Allow for plugins to add their own roomsettings

whalemer:


server/chat-commands/room-settings.ts
```diff
@@ -205,13 +157,12 @@ class RoomSettings {
 		output += `<strong>Caps filter:</strong> <br />${this.capitals()}<br />`;
 		output += `<strong>Emoji filter:</strong> <br />${this.emojis()}<br />`;
 		output += `<strong>Slowchat:</strong> <br />${this.slowchat()}<br />`;
```
Zarel: Perhaps all settings should use the same syntax? Maybe official settings should also be registered in `Chat.roomSettings`.

server/chat-commands/room-settings.ts
```diff
@@ -17,218 +17,32 @@ const SLOWCHAT_USER_REQUIREMENT = 10;
...
+		function button(setting: string, disable: boolean, command = ' ') {
+			if (disable) {
+				return Chat.html`<button class="button disabled" style="font-weight:bold; color:#575757; font-weight:bold; background-color:#d3d3d3;">${setting}</button> `;
+			}
+			return Chat.html`<button class="button" name="send" value="/roomsetting ${command}">${setting}</button> `;
+		}
+
+		let output = Chat.html`<div class="infobox">Room Settings for ${room.title}<br />`;
+		for (const handler of Chat.roomSettings) {
+			output += handler.call({button}, room, user, connection);
```
Zarel: Hm, I wonder if `{button}` should be replaced with a nicer API that had less code duplication...

It might be useful to nail down API now instead of later.

Zarel:
```ts
class SettingsHandler {
  buf: string;
  constructor() {
    this.buf = ``;
  }
  label(text: string) {
    buf += `<strong>${text}:</strong> <br />`;
  }
  button(...etc
```

whalemer: Given that the structure for settings is always the same, what about something like
```ts
type SettingsHandler = (room, user, connection) => {
    label: string, // ie, "Mafia"
    buttons: [string /** button label */, string | null /** button command */][],
}
```

whalemer: something like https://github.com/smogon/pokemon-showdown/compare/master...whalemer:wbi

whalemer: Though, having implemented a few of these, basically everything is either a radio button or an on/off toggle, so maybe we should just have helpers for those.

Zarel: I'm not a huge fan of how that looks in practice. A lot of intermediate variables.

Zarel:
```ts
(room, user) => {
	const label = "Language";
	if (!user.can('editroom', null, room)) {
		return {
			label,
			buttons: [[room.language ? Chat.languages.get(room.language)! : 'English', true]],
		};
	}
	const buttons: [string, string | true][] = [];
	buttons.push([`English`, !room.language || 'roomlanguage english']);
	for (const [id, text] of Chat.languages) {
		if (id !== 'english') buttons.push([text, room.language === id || `roomlanguage ${id}`]);
	}
	return {label, buttons};
},
```

vs

```ts
(room, user) => ({
	label: "Language",
	buttons: user.can('editroom', null, room)) ? [
		[`English`, !room.language || 'roomlanguage english'],
		...([...Chat.languages].filter(([id]) => id !== 'english').map(([id, language]) =>
			[language, room.language === id || `roomlanguage ${id}`]
		))
	] : [
		[room.language ? Chat.languages.get(room.language)! : 'English', true]
	],
}),
```

Zarel: I think I would then think about how to design the API to make that look even simpler than it currently is.

I think it would help to just force `Chat.languages` to have English first, so we don't get that messy `...([...Chat.languages])` mess, and then we can just do:

```ts
(room, user) => ({
	label: "Language",
	buttons: (user.can('editroom', null, room)) ?
		[...Chat.languages].map(([id, language]) =>
			[language, room.language === id || `roomlanguage ${id}`]
		)
	:
		[[room.language ? Chat.languages.get(room.language)! : 'English', true]]
	,
}),
```

Zarel: And then we'd just have a `permission` field that, when false, would only show the currently selected option, so it could just be:

```ts
(room, user) => ({
	label: "Language",
	permission: user.can('editroom', null, room),
	buttons: [...Chat.languages].map(([id, language]) =>
		[language, room.language === id || `roomlanguage ${id}`]
	),
}),
```

Asheviere:
I have wanted to dynamically generate the roomsettings window since it was introduced. 
Good stuff!!!!

server/chat-commands/room-settings.ts
```diff
@@ -1382,3 +1204,82 @@ export const commands: ChatCommands = {
...
+			// groupchat ROs can set modjoin, but only to +
+			// first rank is for modjoin off
+			...RANKS.slice(1, room.isPersonal && !user.can('makeroom') ? 2 : undefined),
+		].map(rank =>
+			[rank, (rank === 'off' ? !room.modjoin : rank === room.modjoin) || `modjoin ${rank || 'off'}`]
+		),
+	}),
+	(room, user) => ({
+		label: "Language",
+		permission: user.can('editroom', null, room),
```
whalemer: ...wonder if it's a good idea to let this return `bool | string` to save on the `user.can(` every time...

Zarel: That's a good point. I'd support that.

server/chat-commands/room-settings.ts
```diff
@@ -1232,7 +1233,7 @@ export const roomSettings: SettingsHandler[] = [
 	},
 	(room, user) => ({
 		label: "Modjoin",
-		permission: room.isPersonal ? user.can('editroom', null, room) : user.can('makeroom'),
+		permission: room.isPersonal ? 'editroom' : 'makeroom',
```
Zarel: This should probably still be `user.can('makeroom')`. Roomvoicing a Leader should not remove their ability to control rooms.

Zarel:
Is this ready?

whalemer:
I think so.

PR #6124, opened 2019-11-30
Add Obstruct in list of protection moves

ZardMX:


Marty-D:
Please put the previous descriptions in the Gen 7 mod. Things that don't exist in old gens shouldn't be mentioned there.

ZardMX:
I'm a little bit confused here, where do we see those descriptions except in the code? I mean for example the move Protect has a different desc in gen 2 than in gen 8 but it shows the shortDesc to me when I am in a gen 2 battle.

Zarel:
They're not currently used, but future-proofing is probably a good idea.

ZardMX:
Why is Travis not checking the changes from my latest commit?

Zarel:
Because it has a merge conflict. You'll need to fix the merge conflict before Travis can check it.

ZardMX:
Let me know if I messed up anything while fixing the merge conflict.

PR #6126, opened 2019-11-30
Defog now removes terrains and G-Max Steelsurge

420Blazeitt:
https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-29#post-8298814

also allows gmax steelsurge to be removed on both sides. Sorry >.>


420Blazeitt:
oh uh, Im having a bit of trouble rewriting the current description without sounding too clunky. Sorry about that

Zarel:
We can let someone else handle the description.

Issue #6121, opened 2019-11-30
Utility Umbrella Mechanics


KrisXV:
- [x] Does Utility Umbrella allow the holder to use Water moves in Desolate Land/Fire moves in Primordial Sea?
- [x] Does Utility Umbrella allow freezing to happen in the sun?
- [x] How does Solar Power interact with Utility Umbrella?
- [x] Does Utility Umbrella nullify the 50% Sp. Def boost to Rock types in Sandstorm?
- [x] Is Castform holding Utility Umbrella prevented from transforming into Castform-Sunny/Rainy in the respective weathers (this one will have to be an assumed mechanic for the time being since Castform isn't in Sword/Shield)?
- [x] Are weather-dependent accuracy moves (Thunder, Hurricane) still able to miss when used against a Pokemon holding Utility Umbrella (Hurricane Gyarados vs. Utility Umbrella Pokemon)?
- [x] Is a statused Pokemon with Hydration holding Utility Umbrella cured in the rain or not?
- [x] Does Growth still boost the user's Atk and Sp. Atk 2 stages in the sun with Utility Umbrella equipped?
- [x] Does Utility Umbrella prevent the holder from using Aurora Veil in Hail?
- [x] What happens if you use Trick to give Cherrim a Utility Umbrella when Sun is active?

Feel free to add more/post answers to listed mechanics questions (with proof)

KrisXV:
> Does Utility Umbrella allow the holder to use Water moves in Desolate Land/Fire moves in Primordial Sea?

no

> Does Utility Umbrella allow freezing to happen in the sun?

yes

> How does Solar Power interact with Utility Umbrella?

As expected; the holder does not lose HP and its Sp. Atk is not boosted.

> Does Utility Umbrella nullify the 50% Sp. Def boost to Rock types in Sandstorm?

no

> Is Castform holding Utility Umbrella prevented from transforming into Castform-Sunny/Rainy in the respective weathers (this one will have to be an assumed mechanic for the time being since Castform isn't in Sword/Shield)?

I've implemented it with Flower Gift as a basis; if the weather is rain/sun and the holder is a Castform, it will not transform.

> Are weather-dependent accuracy moves (Thunder, Hurricane) still able to miss when used against a Pokemon holding Utility Umbrella (Hurricane Gyarados vs. Utility Umbrella Pokemon)?

yes!

> Is a statused Pokemon with Hydration holding Utility Umbrella cured in the rain or not?

no, it is not cured.

[...]

PR #6127, opened 2019-11-30
Update past gen description of effects affected by Utility Umbrella and update item checking

KrisXV:


Zarel:
I think we should wait for #6121 for this one.

KrisXV:
all of the questions in #6121 have been answered

data/abilities.js
```diff
@@ -1022,16 +1022,18 @@ let BattleAbilities = {
...
 			delete this.effectData.forme;
 		},
 		onUpdate(pokemon) {
-			if (!pokemon.isActive || pokemon.baseTemplate.baseSpecies !== 'Cherrim' || pokemon.transformed || pokemon.item === "utilityumbrella") return;
+			if (!pokemon.isActive || pokemon.baseTemplate.baseSpecies !== 'Cherrim' || pokemon.transformed || pokemon.hasItem('utilityumbrella')) return;
 			if (this.field.isWeather(['sunnyday', 'desolateland'])) {
 				if (pokemon.template.speciesid !== 'cherrimsunshine') {
 					pokemon.formeChange('Cherrim-Sunshine', this.effect, false, '[msg]');
+				} else if (pokemon.template.speciesid === 'cherrimsunshine' && pokemon.hasItem('utilityumbrella')) {
+					pokemon.formeChange('Cherrim', this.effect, false, '[msg]');
```
urkerab: This check is too late, you need to check the item when you're checking the weather.

data/abilities.js
```diff
@@ -1114,18 +1116,20 @@ let BattleAbilities = {
...
-		desc: "If this Pokemon is a Castform, its type changes to the current weather condition's type, except Sandstorm.",
+		desc: "If this Pokemon is a Castform, its type changes to the current weather condition's type, except Sandstorm. If this Pokemon is holding Utility Umbrella and the weather condition is Sunny Day, Desolate Land, Rain Dance, or Primordial Sea, it will not change types.",
 		shortDesc: "Castform's type changes to the current weather condition's type, except Sandstorm.",
 		onUpdate(pokemon) {
 			if (pokemon.baseTemplate.baseSpecies !== 'Castform' || pokemon.transformed) return;
 			let forme = null;
 			switch (this.field.effectiveWeather()) {
 			case 'sunnyday':
 			case 'desolateland':
+				if (pokemon.hasItem('utilityumbrella')) break;
```
urkerab: There's a similar problem here: holding the Umbrella should process as if there is no weather (or Delta Stream) rather than not process at all.

data/statuses.js
```diff
@@ -479,7 +479,7 @@ let BattleStatuses = {
 		},
 		onWeatherModifyDamage(damage, attacker, defender, move) {
 			if (move.type === 'Water') {
-				if (defender.item === "utilityumbrella") return;
+				if (defender.hasItem('utilityumbrella')) return;
```
urkerab: The mechanics report says that the nerf is also suppressed, so you could just move the check to the beginning of the function.

data/statuses.js
```diff
@@ -701,7 +701,7 @@ let BattleStatuses = {
 		effectType: 'Weather',
 		duration: 0,
 		onEffectiveness(typeMod, target, type, move) {
-			if (target && target.item === "utilityumbrella") return;
+			if (target && target.hasItem('utilityumbrella')) return;
```
urkerab: I think Anubis determined that the Umbrella does not interact with Delta Stream?

urkerab:
I still see direct comparisons to Utility Umbrella in the Solar moves. Maybe you missed them because they were using the wrong sort of quotes?

I believe Moonlight, Morning Sun and Synthesis are also affected by the Umbrella.

data/statuses.js
KrisXV: It doesnt, not sure what that has to do with this line though?

data/abilities.js
KrisXV: No? This is for if the item is tricked onto cherrim-s

KrisXV:
@Zarel

KrisXV:
Why is the build failing????

```
/home/travis/build/smogon/pokemon-showdown/data/mods/gen7/abilities.js
  106:1  error  Parsing error: Unexpected token <<
✖ 1 problem (1 error, 0 warnings)
```
no?????? that token doesnt exist????

KrisXV:
Bump

PR #6132, opened 2019-12-02
Ladders: Delete rating cache if no response.

HoeenCoder:
If the remote ladder's update rating method dosen't receive
a response from the login server, it will not change the user's
mmrCache for the format, resulting in future ladder matches using
an outdated ELO until the server receives a response at the end of
a future match.

This is the cause of issues such as users with 1400 ELO being matched
with users with 1000 ELO with a 10 second wait time in current gen OU.

HIGH PRIORITY due to an upcoming OU suspect test. Ideally we merge this and reboot the server tonight.

### DIFF
--- a/server/ladders-remote.ts
+++ b/server/ladders-remote.ts
@@ -67,6 +67,8 @@ export class LadderStore {
 		}
 
 		const formatid = this.formatid;
+		const p1 = Users.getExact(p1name);
+		const p2 = Users.getExact(p2name);
 		room.update();
 		room.send(`||Ladder updating...`);
 		const [data, , error] = await LoginServer.request('ladderupdate', {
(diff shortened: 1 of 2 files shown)


server/ladders-remote.ts
```diff
@@ -81,6 +83,9 @@ export class LadderStore {
 			} else {
 				room.add(`||Ladder (probably) updated, but score could not be retrieved (${error.message}).`);
 			}
+			// Clear mmrCache for the format to get the users updated rating next search
+			if (p1 && p1.mmrCache[formatid]) delete p1.mmrCache[formatid];
```
monsanto: The test that `p1.mmrCache[formatid]` exists is redundant; delete is a no-op in this case

server/ladders-remote.ts
```diff
@@ -81,6 +83,9 @@ export class LadderStore {
...
+			if (p2 && p2.mmrCache[formatid]) delete p2.mmrCache[formatid];
 			return [p1score, null, null];
 		}
 		if (!room.battle) {
```
monsanto: I would expect the deletion to be required in every error case

monsanto:
X-posting here for the record:

> ~HoeenHero: specific line comments are easy enough, though for your overall comment I'm not sure of the best way to do that since you wouldnt be able to search for a battle while an update is pending.
> ~HoeenHero: cant really start a match w/o a rating and without being able to request an update
> ~chaos: well, its mostly me trying to understand what the invariant needed for correctness is
> ~chaos: if you only clear on error, then that allows bad matches to start while waiting for an update
> ~chaos: but maybe thats ok
> ~chaos: or am i misunderstanding something here?
> ~HoeenHero: I get your point yeah, the update can take a while though so if we dont have a rating of some kind - outdated or not we would have to to tell the user to wait on their next match (I see alot of users close the match as soon as it ends and start a new search).
> ~chaos: that makes sense, then perhaps a comment explaining the situation would be good enough
> ~chaos: perhaps the invariant that you are looking for is that pairing is accurate within 1 match? not clearing on error means that you can have an unbounded # of matches with an unknown ELO
> ~HoeenHero: Yeah I'm ok with accurate within 1 match, its far better than potentially accurate as of 25 matches ago

Zarel:
This will add even more ladder load...

Zarel:
Someone mentioned doing the Elo update sim-side, which might be a lot better.

PR #6134, opened 2019-12-02
Gmax Eevee, Meowth, and Pikachu shouldn't be LC/NFE

KrisXV:
All three can't evolve, and the latter one isn't capable of getting evolved into.

It's the same situation as the other arbitrary Pikachu formes from previous generations.

### DIFF
--- a/data/formats-data.js
+++ b/data/formats-data.js
@@ -518,6 +518,8 @@ let BattleFormatsData = {
 		eventOnly: true,
 		unreleasedHidden: true,
 		isGigantamax: "G-Max Volt Crash",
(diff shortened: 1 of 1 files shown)


Zarel:
I... um... will leave this to someone else.

Zarel:
Do they benefit from Eviolite? I would assume yes, so I think making them LC/NFE makes sense.

KrisXV:
They do not

Zarel:
Oh wow. Maybe they shouldn't. I would let Smogon make the call, though. It seems weird to make it for them.

PR #6137, opened 2019-12-03
Octolock can't hit ghost types

whalemer:


Marty-D:
I imagine this is a trapping immunity and not a type immunity like Thunder Wave. If you Electrify Octolock it should be able to hit Ground types, probably.

whalemer:
OK yeah, electrified octolock still fails against ghost types.

data/moves.js
```diff
@@ -13199,6 +13199,9 @@ let BattleMovedex = {
 		pp: 15,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
+		onTryImmunity(target) {
+			if (!this.dex.getImmunity('trapped', target)) return false;
```
Marty-D:
```suggestion
			return this.dex.getImmunity('trapped', target);
```

PR #6139, opened 2019-12-03
Fix Ditto from Dynamaxing as a Mega/Primal etc

YaBoiJD:
Ditto is still able to Dynamax while transformed as a Megas, Primals, and Ultra Necrozma.  Since Ditto cannot Dynamax as Zacian, Zamazenta, and Eternatus, Megazard has asked me to fix this for more Balance within the National Dex formats.

### DIFF
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -733,16 +733,6 @@ let BattleFormats = {
 				return [`${set.name}'s item ${item.name} does not exist in Gen ${this.dex.gen}.`];
 			}
 		},
-		onBegin() {
-			// if you have a mega/primal or z, you can't dynamax
-			for (const pokemon of this.getAllPokemon()) {
-				const item = pokemon.getItem();
-				// this.canMegaEvo check is for Rayquaza.
-				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove) {
-					pokemon.canDynamax = false;
-				}
-			}
-		},
 	},
 	ignoreillegalabilities: {
 		effectType: 'ValidatorRule',
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -738,7 +738,7 @@ let BattleFormats = {
 			for (const pokemon of this.getAllPokemon()) {
 				const item = pokemon.getItem();
 				// this.canMegaEvo check is for Rayquaza.
-				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove || (pokemon.transformed && (pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra"))) {
+				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove) {
 					pokemon.canDynamax = false;
 				}
 			}
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -738,7 +738,7 @@ let BattleFormats = {
 			for (const pokemon of this.getAllPokemon()) {
 				const item = pokemon.getItem();
 				// this.canMegaEvo check is for Rayquaza.
-				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove) {
+				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove  || (pokemon.transformed && (pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra" || pokemon.template === 'zacian'|| pokemon.template === 'zamazenta' || pokemon.template === 'eternatus'))) {
 					pokemon.canDynamax = false;
 				}
 			}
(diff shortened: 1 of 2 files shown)


data/rulesets.js
```diff
@@ -737,7 +737,7 @@ let BattleFormats = {
 			for (const pokemon of this.getAllPokemon()) {
 				const item = pokemon.getItem();
 				// this.canMegaEvo check is for Rayquaza.
-				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove) {
+				if (item.megaStone || this.canMegaEvo(pokemon) || item.onPrimal || item.zMove || (pokemon.transformed && (pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra"))) {
```
urkerab: This only runs at the start of the battle, so it has no effect on Ditto.

data/scripts.js
```diff
@@ -1170,6 +1170,7 @@ let BattleScripts = {
 		// {gigantamax?: string, maxMoves: {[k: string]: string} | null}[]
 		if (!skipChecks) {
 			if (!pokemon.canDynamax) return;
+			if (pokemon.transformed && (pokemon.template.baseSpecies === 'zacian' || pokemon.template.baseSpecies === 'zamazenta' || pokemon.template.baseSpecies === 'eternatus')) return;
```
urkerab: This doesn't do anything because the base species is a name, not an ID. (But by my reading of the following code, Ditto transformed into those legendaries should already be blocked from Dynamaxing.)

data/rulesets.js
YaBoiJD: thanks for the feedback @urkerab, much appreciated

data/scripts.js
YaBoiJD: I did tests on this before and after, before this ditto was able to dynamax if it transformed into those legendaries, and after with these changes, and ditto was not able to

Zarel:
???

Did you even bother to ask NatDex tier leaders whether or not this was intentional?

This is intentional, and it matches the closest analogue in Sword/Shield: Zacian and Zamazenta can't Dynamax, but a Ditto transformed into them can.

Zarel:
If you want to be helpful, add a `-hint` message reminding people this is intentionally allowed, so we get fewer bug reports about it.

YaBoiJD:
i was told by Sadisctic and DaWoblefet that a ditto that has transformed into the three legends it couldnt dynamax @Zarel, plus i did also talk to megazard about this too

data/scripts.js
```diff
@@ -1170,6 +1170,7 @@ let BattleScripts = {
 		// {gigantamax?: string, maxMoves: {[k: string]: string} | null}[]
 		if (!skipChecks) {
 			if (!pokemon.canDynamax) return;
+			if (pokemon.transformed && (pokemon.template.baseSpecies === 'zacian' || pokemon.template.baseSpecies === 'zamazenta' || pokemon.template.baseSpecies === 'eternatus' || pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra")) return;
```
Zarel:
```suggestion
			if (['zacian', 'zamazenta', 'eternatus'].includes(pokemon.template.baseSpecies) || pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra") {
				return;
			}
```

data/scripts.js
```diff
@@ -1170,7 +1170,9 @@ let BattleScripts = {
 		// {gigantamax?: string, maxMoves: {[k: string]: string} | null}[]
 		if (!skipChecks) {
 			if (!pokemon.canDynamax) return;
-			if (pokemon.transformed && (pokemon.template.baseSpecies === 'zacian' || pokemon.template.baseSpecies === 'zamazenta' || pokemon.template.baseSpecies === 'eternatus')) return;
+			if (['zacian', 'zamazenta', 'eternatus'].includes(pokemon.template.baseSpecies) || pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra") {
+				return;
+			}
 			// Some pokemon species are unable to dynamax
 			const cannotDynamax = ['zacian', 'zamazenta', 'eternatus'];
 			if (cannotDynamax.includes(toID(pokemon.template.baseSpecies))) {
```
urkerab: Well in that case I'm missing how the three lines after your code don't duplicate the species checks...

YaBoiJD:
uhhh shit
@HoeenCoder or @zarel pm me with halp

data/scripts.js
YaBoiJD: i have no clue, i'm about to look more into it

YaBoiJD:
@Zarel do we necessarily need the Dynamax checks in data/rulesets.js? because with what you suggested, with just doing `pokemon.template.isMega/isPrimal` etc.

data/scripts.js
```diff
@@ -1170,6 +1170,9 @@ let BattleScripts = {
 		// {gigantamax?: string, maxMoves: {[k: string]: string} | null}[]
 		if (!skipChecks) {
 			if (!pokemon.canDynamax) return;
+			if (pokemon.template.isMega || pokemon.template.isPrimal || pokemon.template.forme === "Ultra") {
```
YaBoiJD: I ended up getting rid of the check for ditto when it has transformed into Zacian, Zamazenta, or Eternatus. because when I did further testing, ditto was not able to Dynamax if it had transformed into those three.

PR #6140, opened 2019-12-03
Add chat commands to interface with teams database

marksg07:
Game server portion of smogon/pokemon-showdown-client#1417. See also the related PR for the login server/backend part that this PR is interfacing with: smogon/pokemon-showdown-client#1418.

This PR adds two new chat commands that interface directly with the new teams database. Specifically, the new `uploadteam` and `shareteam` actions are server-only, because they cause a write to the PS database. The two new chat commands are: 

- `/ulteam <format> <teamname> <packedteam> <ispublic: 0|1>`: Uploads a new team to the teams DB by using the `uploadteam` action. Validates the team provided for the given format before doing this.
- `/shareteam <user>, <teamid>`: Shares a team owned by the user with given other user. The login server validates that the given team is actually owned by the user.

Possible improvements: some form of request limiting? Not sure of the best way for this to be implemented. Also, not sure if there's a possible attack on the team validator where it could validate a super long packed team as correct -- it might be better to validate the team, import it, and then export it and upload that in case the user screwed with the team in a way that would be ignored by the validator.

### DIFF
--- a/server/chat-commands/core.js
+++ b/server/chat-commands/core.js
@@ -18,6 +18,7 @@
 /* eslint no-else-return: "error" */
 
 const crypto = require('crypto');
+const assert = require('assert');
 
 const avatarTable = new Set([
 	'aaron',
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
This is blocked on smogon/pokemon-showdown-client#1418, which depends on a Postgres loginserver (which depends on either a security review of the new loginserver or major changes to a legacy codebase).

mia-pi-git:
This is both very inactive and has been implemented already, albeit in a different form. Closing.

PR #6146, opened 2019-12-04
Add VGC 2020 format

DaWoblefet:
Rules announced by TPCi: https://www.pokemon.com/us/pokemon-news/2020-pokemon-video-game-championships-vgc-format-rules/
- Still has 7 minute Your Time
- Only legal Gmax Pokemon are the following: Butterfree, Corviknight, Drednaw, Centiskorch, Meowth, Pikachu, Eevee, Snorlax, Sandaconda, Charizard with Blaze. I implemented this by just hard banning every other Gmax mon.

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -254,7 +254,6 @@ let Formats = [
 		},
 		ruleset: ['Obtainable', 'Standard GBU', 'VGC Timer'],
 		minSourceGen: 8,
-		
 	},
 	{
 		name: "[Gen 8] 2v2 Doubles",
--- a/config/formats.js
+++ b/config/formats.js
@@ -232,13 +232,13 @@ let Formats = [
 
 		mod: 'gen8',
 		gameType: 'doubles',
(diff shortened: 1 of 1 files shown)


DaWoblefet:
More thorough rules update here (nothing contradictory with the above, but also for reference): https://assets.pokemon.com//assets/cms2/pdf/play-pokemon/rules/play-pokemon-vg-rules-formats-and-penalty-guidelines-12042019-en.pdf

Apparently TPCi intends to allow more Gmax Pokemon as the season progresses. I highly, highly doubt this will be comparable to Sun/Moon/Ultra in VGC 2019, so that means we'll just need to periodically unban Gmax formes as they allow them. Even though the majority of these Gmax mons are released, they're apparently hard to find or something (?), so that seems to be behind the motivation for banning most Gmax mons.

DaWoblefet:
Oh yeah, @TheImmortal or whomever

PR #6147, opened 2019-12-04
Ice Face restores on switch-in when hail is active

ZardMX:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8299187

The shortDesc was kind of misleading with the "Restored in hail" part, but I am not sure what to change to make it more clear when it actually restores.

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -1567,9 +1567,15 @@ let BattleAbilities = {
 		num: 115,
 	},
 	"iceface": {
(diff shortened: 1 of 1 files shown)


TheImmortal:
That's not really a reliable source.

ZardMX:
What do you mean? https://www.youtube.com/watch?v=QWKJb3Czy8Q shows that is how it is in the games.

TheImmortal:
Oh, the video didn’t show when I clicked there before.

PR #6152, opened 2019-12-06
Allow drivers to modchat by default

asgdf:
As was discussed.
The roomban permission thing is probably not great. There is a ban permission in the config, but it will apply to all drivers (and includes gban for gdrivers), which I don't want. I'm open to other suggestions of how to implement that.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -543,7 +543,6 @@ exports.grouplist = [
 		inherit: '%',
 		jurisdiction: 'u',
 		ban: true,
-		modchat: true,
 		modchatall: true,
 		roomvoice: true,
 		forcerename: true,
@@ -573,6 +572,7 @@ exports.grouplist = [
 		jeopardy: true,
 		joinbattle: true,
 		minigame: true,
+		modchat: true,
 	},
 	{
 		symbol: '\u2606',


asgdf:
Bump, any opinions?

@Zarel since the permission system is mostly your work afaik, what do you think would be best here?

Zarel:
:[

asgdf:
Which part of this makes you :[? Maybe we can find another option.

Zarel:
The permission system doesn't support "room driver" and "global driver" being this different.

I think we need to just create different symbols for the two roles.

Zarel:
It sounds like we should just rename gdriver to gmod, and gmod to supermod or something?

Lusamine:
> 
> 
> It sounds like we should just rename gdriver to gmod, and gmod to supermod or something?

I feel like changing names/ranks people are already used to would cause more confusion than help.  But if you do that, pick something other than "supermod" because that'll be confusing as hell when Smogon has supermods.

asgdf:
Hmm. If it came down to it I'd prefer keeping the ranks as they are to allowing gdrivers roomban.
So I assume bypassing the permission system and doing a special case (like it is right now) is not a preferable option?

asgdf:
Alright ok we're not doing the roomban thing (for now) because the permissions would be super messy. So I guess I'm just doing the modchat change, that fits into the system.

TheJetOU:
What about having a `globalperm` field? It would allow for overriding existing perms or add new perms if the user has the global form of the rank.
```javascript
  symbol: "%",
  ...
  globalperm: {
    ban: true,
  },
````

asgdf:
The thing is that "ban" can refer to both roomban and globalban, depending on whether it's a room or global driver

TheJetOU:
> The thing is that "ban" can refer to both roomban and globalban, depending on whether it's a room or global driver

You could split `ban` into `roomban` and `globalban`.

asgdf:
It would be an option, but that seemed like a refactor that would end up going wildly out of scope, and I didn't want to go down that rabbit hole.

TheJetOU:
Fair enough, I suppose.

@Zarel - thoughts on the proposed idea?

Zarel:
I support splitting `ban` into `ban` and `globalban`.

But that doesn't solve the problem of "we want gdrivers to be able to roomban but not roomdrivers".

TheJetOU:
> But that doesn't solve the problem of "we want gdrivers to be able to roomban but not roomdrivers".

https://github.com/smogon/pokemon-showdown/pull/6152#issuecomment-565873839

Zarel:
The problem isn't technical, the problem is user confusion due to the symbols meaning massively different thing.

I said in the Staff chatroom that I want a better compromise. My suggestion is: Global drivers should get `@` and be called Global Mods, Global Mods get `&` and get some new name, and Leaders/Admins can share `~`.

KrisXV:
Sorry and no offense but that really doesn't seem like a good idea. Aside from the fact that changing the symbols [and names] of all the global staff would confuse a majority of the user base even more than you seem to believe it confuses them already, it just doesn't come off as a rational approach to giving global staff more flexible permissions. I can somewhat understand some changes like combining Leader and Admin since they already share almost all of the same permissions (or making `~` Technical admin and `&` Policy admin could work), but most of the PS user base already understands the difference between a Room and Global Mod/Driver and that they even come with different permissions. I don't see how this issue is anything more than just a technical one because of the number of already established articles (that explain the difference) and the general widespread understanding of room vs global staff throughout the sim.

AeonicX7:
I'm with Kris on this one. I think a shakeup to ranks on this scale would confuse the userbase too much and isn't worth the small benefits we would get from this move. I think the idea absolutely has merit in theory, and maybe it can be reworked a bit for an implementation, but right now I can't see it going smoothly. 

Sort of a splinter topic, but reworking leader/admin to have more equal footing but separate ranks is something I've thought would make sense. I'm all for that idea.

TheRealQuote:
At this point we're better off abandoning global driver roombans until we can come to a solution that works for all of us. Changing the ranks is not happening. It'll cause too much confusion and doesn't address the problem. If you don't want to implement this because of technical limitations I understand, but if that's not actually the issue...then why not?

Zarel:
The issue is that the English words "global driver" mean "roomdriver in every room". The symbol system even supports this interpretation!

If "global driver" on PS actually means "roommod in every room", then the name is clearly misleading. It may not be a problem to you, but it's probably a problem for people who are worse at English, and people who are not heavily experienced with the PS auth system. You always get too blind to problems like these from the inside where you're used to the massive gulf between roomdrivers and gdrivers, but I'm putting my foot down because this is a horrible system to people who aren't people like you.

TheRealQuote:
What if we start referring to Drivers as Junior Mods? Would that work?

Zarel:
Possibly. The symbols, though?

TheRealQuote:
The symbols? $, ^ and = are all international symbols that can be used.
Also, 
"You always get too blind to problems like these from the inside where you've spent years here and gotten used to the massive gulf between roomdrivers and gdrivers"

Believe it or not, I was a reg at one point, and understanding the difference between a reg and a global was very easy. What Lowest Common Denominator are you trying to appeal to by making this For Dummies? Anyone that spends enough time in the chats or reads the guides can understand the difference.

Asheviere:
To me it's about 6 years too late to completely overhaul the rank system. Changing the whole system at this point is not going to be worth making the switch unless there are absolutely huge benefits and i dont think it applies here at all.

PR #6159, opened 2019-12-07
Add Gen 7 Hackmons Cup

hilts-vaughan:
Is it possible to get this in the server since there are still other Gen7 metas? 

FYI I did test locally.


TheImmortal:
Perhaps. Not as a ladder tho.

TheImmortal:
And not doubles because that's even more niche.

hilts-vaughan:
I know it's niche but I think the main gripe I had is that now it has a much larger Dex pool. I can always spin up my own server with this change (and I have) so it's not a deal breaker but I hope you will at least reconsider in the future. :)

hilts-vaughan:
Thanks for merging my PR!

PR #6160, opened 2019-12-07
Fix various evolution methods

KrisXV:


server/chat-commands/info.js
```diff
@@ -585,7 +585,7 @@ const commands = {
 								evos.push(`${evo.name} (level-up with ${evo.evoMove}${condition})`);
 								break;
 							case 'other':
-								evos.push(`${evo.name} (${condition})`);
+								evos.push(`${evo.name} (${evo.evoCondition})`);
```
KrisXV: I left this as evo.evoCondition instead of using the defined `const` because the condition constant adds an unnecessary space

PR #6161, opened 2019-12-07
Make tour battles inherit mutes

whalemer:


server/rooms.ts
```diff
@@ -1447,6 +1447,12 @@ export class GameRoom extends BasicChatRoom {
 
 		this.tour = options.tour || null;
 		this.parent = options.parent || (this.tour && this.tour.room) || null;
+		// copy mutes from parent room
```
Zarel: I would rather inherit them (search parents for mutes when checking for mutes) than copy them. That way, new mutes would also apply to subrooms.

whalemer: .....yeah, that's a much better way to do this

server/rooms.ts
```diff
@@ -373,7 +375,7 @@ export abstract class BasicRoom {
 		}
 		this.runMuteTimer();
 
-		user.updateIdentity(this.roomid);
+		user.updateIdentity();
```
Zarel: ?

whalemer: You need to update it globally now, since you might also be getting muted / unmuted in a subroom.

Zarel: ohhhh

PR #6167, opened 2019-12-09
Make G-Max Centiferno/Sandblast mechanics more accurate

KrisXV:


data/statuses.js
```diff
@@ -247,6 +247,8 @@ let BattleStatuses = {
 		onResidual(pokemon) {
 			const source = this.effectData.source;
 			if (source && (!source.isActive || source.hp <= 0 || !source.activeTurns)) {
+				// G-Max Centiferno and G-Max Sandblast continue even after the user leaves the field
+				if (['gmaxcentiferno', 'gmaxsandblast'].includes(this.effectData.sourceEffect.id)) return;
```
urkerab: Does the damage continue or just the trapping effect? This only makes the trapping effect continue.

KrisXV: oh crap

PR #6168, opened 2019-12-09
Change inheritsLearnsetFrom to inheritsFrom and fix Necrozma-Ultra's inheritance

KrisXV:


data/rulesets.js
```diff
@@ -118,7 +118,12 @@ let BattleFormats = {
 		effectType: 'ValidatorRule',
 		name: 'Obtainable Moves',
 		desc: "Makes sure moves are learnable by the species.",
-		// Hardcoded in team-validator.ts
+		// Mostly hardcoded in team-validator.ts
+		banlist: [
+			// Sunsteel Strike: only obtainable from Necrozma-Ultra that start as Necrozma-Dusk-Mane
+			// Moongeist Beam: only obtainable from Necrozma-Ultra that start as Necrozma-Dawn-Wings
+			'Necrozma-Ultra + Moongeist Beam + Sunsteel Strike',
+		],
```
urkerab: According to commit 71498d4,  `R` moves are already supposed to be only allowed once per set, so this change shouldn't be necessary.

TheImmortal: I believe you can't select Necrozma-U in the teambuilder anyway because it doesn't validate

urkerab: Well, we could always fix the validator to allow it to validate if it has exactly one of Moongeist Beam or Sunsteel Strike...

PR #6169, opened 2019-12-09
Update /smogdex to gen 8

420Blazeitt:


TheImmortal:
It should take you to SM for deleted Pokémon imo

420Blazeitt:
true, good point!

>

TheImmortal:
Deleted moves too!

Zarel:
That can be in a subsequent PR.

PR #6176, opened 2019-12-11
Fix Necrozma-Ultra's inheritance

KrisXV:


Zarel:
Shouldn't the correct answer be to just reject Necrozma-Ultra as an in-battle forme only?

Zarel:
I don't like the amount of complexity this adds.

KrisXV:
Blame chaos then

sim/team-validator.ts
```diff
@@ -1848,7 +1848,7 @@ export class TeamValidator {
 			return template;
 		} else if (template.inheritsFrom) {
 			// For Pokemon like Rotom, Necrozma, and Gmax formes whose movesets are extensions are their base formes
-			return this.dex.getTemplate(template.inheritsFrom);
+			return this.dex.getTemplate(Array.isArray(template.inheritsFrom) ? template.inheritsFrom[0] : template.inheritsFrom);
```
Zarel: I don't see how this can be blamed on chaos.

KrisXV: Read the sim-dev channel on discord, chaos and I have had some conversations about this.

Zarel: @monsanto? I don't like this solution.

Zarel: I think in general you should just give an error "Please choose which Necrozma to inherit from". Maybe if you have a signature move, you can just inherit from that one.

KrisXV: I’ll look into it in a bit

monsanto: @Zarel Sorry, your GH mention got lost in my notifications. I agree with your first take that validating learnsets on in-battle only formes doesn't make much sense. I said as much in the convo and so did @TheImmortal. 

@KrisXV said he wanted this behavior anyway, so I suggested a minimally invasive change. Picking either element is correct for Necro-U because learnsets.js already has an explicit entry for it containing the union of the moves of the formes it inherits from. 

The "right thing" would be to loop over all parents and union the results. But this adds a bunch of complexity to the validator for a case that shouldn't exist in the first place.

Zarel: Yeah, I remember you also disagreed with validating Necrozma-U, so I was confused by what Kris meant. I'll just implement it the way I want it.

TheImmortal: urkerab's idea of validating Necrozma-U if it has either of the special moves seems good

PR #6178, opened 2019-12-12
Update the LGPE moves

KrisXV:
https://www.smogon.com/forums/threads/pokemon-sword-and-shield-datamine-thread.3656143/page-69#post-8312138


Zarel:
Are these moves even usable at all? Should we implement these changes?

Zarel:
We usually don't bother implementing unreleased changes if they don't affect Hackmons, afaik.

TheImmortal:
I agree. But gen 8 custom game I guess? I don’t care either way.

PR #6179, opened 2019-12-12
Allow Melmetal and Meltan in NatDex Formats

YaBoiJD:


Zarel:
These PRs should come with a link to the announcement thread on Smogon.

Zarel:
Hm, well, if we assume LGPE has Pokémon Home access, which is probably a valid assumption, I suppose this makes sense.

PR #6180, opened 2019-12-12
Some Utility Umbrella fixes

urkerab:
- Cherrim should revert if given a Utility Umbrella in Sun.
- Castform shouldn't keep transforming if it holds a Utility Umbrella.
- Leaf guard Utility Umbrella check works but is just inconsistent placement.
- Sun/Rain affected recovery moves should heal 50% with Utility Umbrella.
- Solar moves power should halve in sand or hail even with Utility Umbrella.
- Weather Ball should double in power if Magic Room blocks Utility Umbrella.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -12543,12 +12543,10 @@ let BattleMovedex = {
 		flags: {snatch: 1, heal: 1},
 		onHit(pokemon) {
 			let factor = 0.5;
-			if (this.field.isWeather(['sunnyday', 'desolateland'])) {
+			if (!pokemon.hasItem('utilityumbrella') && this.field.isWeather(['sunnyday', 'desolateland'])) {
 				factor = 0.667;
-			} else if (this.field.isWeather(['raindance', 'primordialsea', 'sandstorm', 'hail'])) {
+			} else if (this.field.isWeather(pokemon.hasItem('utilityumbrella') ? ['sandstorm', 'hail'] : ['raindance', 'primordialsea', 'sandstorm', 'hail'])) {
 				factor = 0.25;
-			} else if (pokemon.hasItem('utilityumbrella')) {
-				factor = 0.5;
 			}
 			return !!this.heal(this.modify(pokemon.maxhp, factor));
 		},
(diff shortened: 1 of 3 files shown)


urkerab:
This started getting repetitive and I was wondering whether it was worth having a `pokemon.effectiveWeather()` function which checks for the umbrella and if so ignores sun and rain.

data/moves.js
```diff
@@ -17576,8 +17572,7 @@ let BattleMovedex = {
 			return null;
 		},
 		onBasePower(basePower, pokemon, target) {
-			if (pokemon.hasItem('utilityumbrella')) return;
-			if (this.field.isWeather(['raindance', 'primordialsea', 'sandstorm', 'hail'])) {
+			if (this.field.isWeather(pokemon.hasItem('utilityumbrella') ? ['sandstorm', 'hail'] : ['raindance', 'primordialsea', 'sandstorm', 'hail'])) {
```
KrisXV: I don't really like how long this line is, maybe changing it to a variable? it functions the same either way but using a separate function has more readability

data/moves.js
```diff
@@ -20904,12 +20896,12 @@ let BattleMovedex = {
...
 			case 'sunnyday':
 			case 'desolateland':
-				if (pokemon.item === "utilityumbrella") break;
+				if (pokemon.hasItem('utilityumbrella')) break;
 				move.basePower *= 2;
 				break;
 			case 'raindance':
 			case 'primordialsea':
-				if (pokemon.item === "utilityumbrella") break;
+				if (pokemon.hasItem('utilityumbrella')) break;
```
KrisXV: i swear i fixed these i dont know what happened when i rebased last time

Zarel:
I definitely approve of `pokemon.hasWeather`

Zarel:
Although I would name it `pokemon.hasWeather`

urkerab:
Still using `effectiveWeather` for now since some of the `switch` statements are tricky to implement using `hasWeather`.

urkerab:
Ugh, accidentally added the comment with spaces instead of tabs...

Zarel:
Looks good. Thanks!

PR #6181, opened 2019-12-13
Gigantamax forme change is client-only

urkerab:
Despite accepting that his other reports were invalid the reporter here https://www.smogon.com/forums/posts/8312554 is defiant(!) about this one...

### DIFF
--- a/data/statuses.js
+++ b/data/statuses.js
@@ -730,7 +730,7 @@ let BattleStatuses = {
 			if (pokemon.species === 'Eternatus-Eternamax') return;
 			pokemon.removeVolatile('substitute');
 			this.add('-start', pokemon, 'Dynamax');
-			if (pokemon.canGigantamax) pokemon.formeChange(pokemon.canGigantamax);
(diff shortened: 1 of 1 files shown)


urkerab:
DaWoblefet confirmed this for me.

Zarel:
Honestly, I was going to have it governed by a flag on the Dynamax thingy.

urkerab:
Oh, you mean that when you tell the client that you're Dynamaxing, you include the Gigantamax flag? I was worried that would confuse the client in the case where a natural forme change happens during Dynamaxing, which shouldn't wear off at the end of Dynamax.

Zarel:
I'm pretty sure those wouldn't interfere.

But yes, I mean `|-start|POKEMON|Dynamax|Gigantamax`

Zarel:
But I wonder if this should also apply to other cosmetic forme changes? Like idk, Mimikyu?

urkerab:
That's not cosmetic; when Mimikyu's Disguise is busted, its speed, type, weight, etc. all reset, while Dynamaxing doesn't reset speed, type or weight.

PR #6183, opened 2019-12-13
NFE: Don't inherit from OU

TheMezStrikes:


Zarel:
Why not?

Zarel:
...why not?

TheImmortal:
It's a unique metagame? It doesn't follow OU. Just because it shared similar clauses doesn't mean it should inherit.

Zarel:
I guess...

TheMezStrikes:
Very similar case to why LC doesn't inherit

PR #6184, opened 2019-12-14
Added Dockerfile for containerization

hilts-vaughan:
You can use this to run Pokemon Showdown in Docker easily. I use this for running a personal server with a custom config on my server easily -- without needing to worry about versioning of Node etc for other software. 

You can use the following to test it out:

```
docker image build -t pokemon-showdown-server:latest .
docker container run --network host pokemon-showdown-server
```

I run this in my own branch right now -- wondering if we could get this upstream? I would be fine to maintain this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
 # Dockerfile for Pokemon Showdown Server
-FROM node:10
+FROM node:13
 
 # Create app directory
 WORKDIR /usr/src/app
 
-# Copy pacakge and cache things
+# Copy package and cache things
 COPY package*.json ./
 RUN npm install
 
@@ -15,4 +15,4 @@ COPY . .
 EXPOSE 8000
 
 # Start the server
-CMD ["node", "pokemon-showdown"]
+CMD ["pokemon-showdown"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# Dockerfile for Pokemon Showdown Server
+FROM node:10
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Copy pacakge and cache things
+COPY package*.json ./
+RUN npm install
+
+# Copy the application into the docker image
+COPY . .
+
+# Expose the server on port 8000
+EXPOSE 8000
+
+# Start the server
+CMD ["node", "pokemon-showdown"]


Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Dockerfile for Pokemon Showdown Server
+FROM node:10
```
Zarel: We're about to upgrade to requiring Node 13, btw

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN npm install
+
+# Copy the application into the docker image
+COPY . .
+
+# Expose the server on port 8000
+EXPOSE 8000
+
+# Start the server
+CMD ["node", "pokemon-showdown"]
```
Zarel: You can't just do `["pokemon-showdown"]`? The `node pokemon-showdown` invocation is a Windows-specific workaround because Windows doesn't support Unix-style shell scripts.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Dockerfile for Pokemon Showdown Server
+FROM node:10
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Copy pacakge and cache things
```
Zarel:
```suggestion
# Copy package and cache things
```

Dockerfile
hilts-vaughan: I can swap this to Node 13.

Dockerfile
hilts-vaughan: You can; though I'm not in the habit of prefixing `node` since it works without the hashbang. Looks like you have it in the script, though.

Dockerfile
hilts-vaughan: Done.

hilts-vaughan:
Done. Are you open to this change, then since you took the time to comment on it?

Zarel:
I'm neutral towards the change. The rest of the staff seem generally against it, though?

hilts-vaughan:
What makes you say that?

Zarel:
Let me just quote them:

> unfortunately, Showdown currently isn't really designed to work in platforms like this
this is not a fault of PR author, the problem is that Pokemon Showdown has certain assumptions about file system it is running in that Docker violates
this can be dealt with, but likely would necessitate some changes outside of Dockerfile
to be specific, what i mean by that is state, specifically that i don't see VOLUME instruction in Dockerfile, which means Docker is free to remove files when it feels like doing so (the application is declared to be stateless when it's very much not so)
which makes Dockerfile as proposed dangerous to use in production
having VOLUME would require having to have a data directory, you cannot put state anywhere, and we currently put it anywhere

hilts-vaughan:
I run this with some custom mount points for things like `config`

It's true -- you need more configuration to make it truly portable. Though, need a bit more info than "anywhere". Does it ever write outside of the directory? Can the entire thing be mounted? Is there an interest in doing this _at all_? Since it's just a single file in my fork, I don't really have a strong feeling about it one way or the other.

Zarel:
PS only writes state into `config/` and `logs/lastbattle.txt` (the latter of which should probably be moved to config because it's state). The rest of `logs/` is logs, and everything else is build artifacts that can be freely cleared.

hilts-vaughan:
Then the statements made by the team above are quite odd -- these are simple changes that can be made if you want to formalize these (I already mount config as a VOLUME on my server and I guess I never cared about logs but that's trivial to mount as well)

I'll make the adjustments.

monsanto:
> Is there an interest in doing this at all?

I doubt any of our current developers will use it and we won't use it in production, so there's no guarantee that future changes won't break your script. 

> Can the entire thing be mounted?

I think this is the safest thing to do for now. Limiting it to `logs`/`config` would break the proposed change of adding a `databases` folder (already used on the main server). It might be a good idea for the future to have an explicitly documented guarantee to only write in a particular (perhaps configurable) directory.

Zarel:
Why wasn't `databases/` added to `config/`? It can still be moved, right?

monsanto:
Sure, it can be moved. It's unclear to me why `config/` would be preferred over `logs/`, for the pmlog at least. I just put it in the `databases` folder because that's what I was told worked with what others were working on.

Zarel:
Oh, hm. I think some databases belong in `logs/` and some in `config/`.

Is there a need for the directory explicitly?

monsanto:
I don't personally think it's necessary to have a specific `databases` directory. The only wrinkle I see is that `databases/` is on SSD and `logs/` is on HDD. If each database is in its own folder (like `logs/pms/`) then it is easy to pick and choose which disk it is stored on.

Zarel:
I don't think `logs/` being on HDD is a big deal; we probably shouldn't compromise our folder structure.

scheibo:
call me crazy, but i think configs/ should store configs (which i wouldnt expect the application to change, though may be changed by the user - i dont think lastbattle.txt belongs here), logs/ should store chat/battle logs, databases/ should store databases.

Zarel:
Call _me_ crazy, but chat/battle logs are a type of database, and configuration is also a type of database. Heck, `data/` is a read-only database. Having a folder named `databases/` is really weird to me if you're not going to put all your databases in it.

scheibo:
Don't get me started on `data/` - its terribly named given it actually contains the bulk of _code_/functionality in the applicaiton

Zarel:
I mean, valid. I've been trying to refactor `scripts.js` back into `sim/` for a while now.

scheibo:
To return to the original naming question:

I agree there different interpretations of 'databases'. What makes sense to me:

- `logs/` - append only files where entries to the end of the file are newer than entries at the top. this includes chat logs/mod logs/battle logs (which can include chats)/error or operational logs. I don't believe `lastbattle.txt` or `repl/` belongs here, nor anything with a `.csv`/`.tsv` suffix (`tickets/`)

- `config/` - information configured by the sysadmin, *not written by the server*. This data is not immutable, it can be changed at runtime by the sysadmin, but the server doesnt write here. By this definition I dont think a lot of the stuff we write here belongs elsewhere (`.tsv` and `.json` files which are updated all the time via PS), but this seems at odds with your view that PS state *does* belong here. I'd suggest they all instead belong under a third directory (`databases/`, `state/`, w/e)

From writing this up, perhaps the real distinction I'm thinking of is actually 'output', 'input', 'input/output'. `logs/` is what I view as *output*, `config/` is what I see as *input* (though currently its *input/output*), and some third directory is 'input/output'). TBH i'm OK with having just 'output' and 'input/output' dirs like we have today, i just think `config/` isn't really a great name for the 'input/output' folder as config implies something different to me. I think this distinction is also somewhat helpful for deciding what should go on HDD vs. SSD, though as with everything there are nuances and exceptions

Zarel:
I do agree that your distinction of "output-only" vs "input-only" vs "input/output" makes sense. It's definitely a logical way to categorize things.

But I have two objections to adopting those categories as directory structure:

1. I don't think it's a very ergonomic way to categorize things; it requires familiarity with the codebase. I think categorizing by use-case makes more sense.

2. We don't actually have any "output-only" or "input-only" files. Logs? Log viewers and searchers takes them as input. Config files? PS outputs them if they don't exist on launch, and nothing precludes us from releasing commands/UI to edit them from inside PS in the future.

Zarel:
I do agree that the current directory structure could use rearranging – `config/` is probably not a great name for the directory with most of our state. I don't know what we should call it, though. Windows calls it `Application Data` and macOS calls it `Library/Application Support` which doesn't exactly help us here.

Zarel:
I think the main problem is that config and state get kind of blurry. To me, config is anything you would expect a sysadmin to want to configure by editing files. But then PS writes to a lot of those files.

The best example is probably `usergroups.csv`. You set up the server by adding `your username,~` to it. And after that, PS writes to it, and you never touch it again. Is it a config file? I'd argue yes.

AnnikaCodes:
I don't know if we're still interested in this. `databases/` is also a directory that PS writes to, if SQLite modlogs are in use.

Zarel:
Presumably we should designate all of `config`, `logs`, and `databases` as folders containing application state. The rest of this seems fine for Docker? Really, anyway who uses Docker could probably finish this up.

Zarel:
> I mean, valid. I've been trying to refactor `scripts.js` back into `sim/` for a while now.

Side note, but this is done! 🎊 

> I do agree that your distinction of "output-only" vs "input-only" vs "input/output" makes sense. It's definitely a logical way to categorize things.

I continue to think it's a poor fit for PS, though. There's no reason PS _shouldn't_ write to `config.js` except that we don't currently do it. In theory, all we need is for someone to set up `usergroups.csv`, and then there's no reason they shouldn't be able to set the rest of the configuration from inside PS.

In other words, we only have one input-only file, and there's no reason it needs to be input-only. If `config/` is a bad name for an input/output directory, I'm open to a better name.

AnnikaCodes:
I'm closing this -- we're already discussing containers at #8693 and #8824; I don't think we need a third discussion. @hilts-vaughan, we'd be happy to hear your input in the other threads though :)

PR #6195, opened 2019-12-19
Take National Dex out of beta

KrisXV:


config/formats.js
```diff
@@ -333,26 +333,18 @@ let Formats = [
 	{
 		name: "[Gen 8] National Dex (beta)",
 
-		mod: 'gen8',
+		mod: 'natdex',
```
Zarel: Is this necessary?

Zarel:
I don't know why we need a mod for NatDex. I think you can just detect the prefix, same as Let's Go.

KrisXV:
The main reason why I added a mod file was because if we tried coding it in the client’s build index like we did with doubles before Doubles tiers were added, there would be a ton of misplaced formes and Pokémon

Zarel:
`natDexTier` entry in formats-data?

KrisXV:
sure

KrisXV:
there are validation issues that i can fix later

TheImmortal:
I vote against filling formats-data with unnecessary data. Like, there aren't even NatDex tiers. Just hardcode the Ubers on the client if you must.

config/formats.js
```diff
@@ -335,14 +335,12 @@ let Formats = [
...
-		banlist: [
-			'Arceus', 'Blaziken', 'Darkrai', 'Deoxys-Attack', 'Deoxys-Base', 'Deoxys-Speed', 'Dialga', 'Eternatus', 'Gengar-Mega',
-			'Giratina', 'Groudon', 'Ho-Oh', 'Kangaskhan-Mega', 'Kyogre', 'Kyurem-Black', 'Kyurem-White', 'Landorus-Base', 'Lucario-Mega',
-			'Lugia', 'Lunala', 'Marshadow', 'Mewtwo', 'Naganadel', 'Necrozma-Dawn-Wings', 'Necrozma-Dusk-Mane', 'Palkia', 'Pheromosa',
-			'Rayquaza', 'Reshiram', 'Salamence-Mega', 'Shaymin-Sky', 'Solgaleo', 'Xerneas', 'Yveltal', 'Zacian', 'Zamazenta', 'Zekrom',
-			'Moody', 'Power Construct', 'Shadow Tag', 'Baton Pass',
-		],
-		unbanlist: ['Melmetal', 'Meltan'],
+		banlist: ['Uber', 'Moody', 'Power Construct', 'Shadow Tag', 'Baton Pass'],
+		onValidateSet(set, format) {
```
TheImmortal: This is dumb. Users should be able to instantly see the banlist with ``/tier nationaldex``.

TheImmortal:
Not just formats-data, touching validator/dex/etc is unnecessary. Let it be a mod so this can be easily wiped when NatDex's time ends.

Zarel:
Ugh, no, do not make it a mod. Mods are way too heavyweight for what's literally just a validator change.

PR #6198, opened 2019-12-19
Add CommandContext#shouldBroadcast and fix !code permission checks

Kaiepi:


server/chat-commands/info.js
```diff
@@ -2401,16 +2401,17 @@ const commands = {
...
-		if (!this.canTalk()) return;
-		if (!this.runBroadcast(true, '!code')) return;
 		if (target.length >= 8192) return this.errorReply("Your code must be under 8192 characters long!");
 
 		const params = target.split('\n');
 		if (!params[0]) params.unshift();
 		if (!params[params.length - 1]) params.pop();
-		if (params.length === 1 && params[0].length < 80 && !target.includes('```') && this.message.startsWith('/')) {
-			return `\`\`\`${params[0]}\`\`\``;
+		if (params.length === 1 && params[0].length < 80 && !params[0].includes('```') && this.willBroadcast) {
```
Zarel: `this.willBroadcast` is always `false` here, isn't it?

Zarel: Also changing `target` to `params[0]` is pointless.

Kaiepi: `this.willBroadcast` is `false` with `/code`, but `true` with `!code`. `params[0]` may have fewer characters if there was a leading or trailing newline.

Zarel: I don't believe you.

Zarel: image

server/chat.ts
```diff
@@ -733,8 +733,11 @@ export class CommandContext extends MessageContext {
 		}
 		return true;
 	}
+	get willBroadcast() {
```
Zarel: I hate getters. :[

Zarel: Just make it a function.

server/chat-commands/info.js
Zarel: Oh, you introduced it in the PR. It shouldn't be a getter; getters are mostly banned.

server/chat.ts
Zarel: Also I'd probably name it something like `shouldBroadcast()` for clarity?

Kaiepi: That sounds reasonable. What's wrong with getters though? I'm not attached to this being a getter or anything, just for reference in the future

Kaiepi:
Rebased

server/chat.ts
Zarel: I just prefer code execution to be explicit.

PR #6199, opened 2019-12-20
Fix spelling for punishment list

Claire238:
Changed "FORCRERENAME" and "TOURBAN" to "FORCERENAME" and "TOUR BAN" to be consistent with modlog


Zarel:
Thanks!

PR #6207, opened 2019-12-22
Chat: unit test filters

Kaiepi:
There's still quite a bit of work that needs to be done before this will be finished, so I wanna make sure what I have now is testing them properly.


AnnikaCodes:
Superseded by #7710.

PR #6208, opened 2019-12-23
(WIP) Implement gen 8 illusion

KrisXV:
i cannot figure out how to edit a Pokemon's nickname from pokemon.ts

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -1624,8 +1624,8 @@ let BattleAbilities = {
 		num: 35,
 	},
 	"illusion": {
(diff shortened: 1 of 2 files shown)


data/abilities.js
```diff
@@ -1624,8 +1624,8 @@ let BattleAbilities = {
 		num: 35,
 	},
 	"illusion": {
```
Zarel: Copy the entire old implementation to `gen7/abilities.js`.

Zarel:
`toString()` returns the fullname; that will need to be adjusted.

`getDetails()` returns the full details; that will need to be adjusted.

Zarel:
You will need a silent formechange on switch-in, similar to how Minior etc work.

PR #6211, opened 2019-12-23
Update item descriptions

YaBoiJD:


data/items.js
```diff
@@ -378,7 +378,7 @@ let BattleItems = {
 		num: 851,
 		gen: 7,
 		isPokeball: true,
-		desc: "A special Poke Ball designed to catch Ultra Beasts.",
+		desc: "A Poké Ball that has a low success rate for catching a Pokémon.",
```
Zarel: Descriptions are supposed to be clear and useful, you're not supposed to just copy the in-game description.

Zarel: This should probably just be changed back.

KrisXV: You’re also not supposed to use accented e because smogon standard is to not use them

Zarel: Do we need to follow Smogon standard?

data/items.js
```diff
@@ -1705,7 +1705,7 @@ let BattleItems = {
 		},
 		num: 34,
 		gen: 2,
-		desc: "Restores 50 HP to one Pokemon but lowers Happiness.",
+		desc: "Restores 60 HP to one Pokemon",
```
KrisXV: This is missing a period

YaBoiJD:
@Zarel, Should I make a better description for the Utility umbrella?

Zarel:
If you want, sure.

KrisXV:
the shortdesc of Utility Umbrella should just be `Blocks the holder from the effects of sun/rain`, but the normal description should go into all of the details

YaBoiJD:
@Zarel should this be closed, or should I leave open?

PR #6213, opened 2019-12-23
CAP: Prevent unreleased Pokemon from being used

KrisXV:


Zarel:
Why not just set `minSourceGen: 8` like OU does?

Zarel:
Also why is everyone so inconsistent every time? Last gen everyone was scrambling to make the one with unreleased Pokémon the official OU tier.

KrisXV:
because none of the CAP pokemon have gen 8 learnsets so theyd be completely unusable, which defeats the point of cap

PR #6215, opened 2019-12-23
Make semi-invulnerability work in Gen 1 again

urkerab:
The changes I made to simplify the No Guard checks in later gens mean that this function is assumed to return one of `true`, `false` or `0`, but not `null`, which I overlooked at the time...


Zarel:
Maybe add tests?

Issue #6212, opened 2019-12-23
RandomPlayerAI wrong teamsize assumption


Liminalia:
Running a battle versus a RandomPlayerAI with a team of size 1 causes an error:
```
(node:5460) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'condition' of undefined
    at request.active.map (/usr/src/pokemon-showdown-master/.sim-dist/tools/random-player-ai.js:103:37)
    at Array.map (<anonymous>)
    at RandomPlayerAI.receiveRequest (/usr/src/pokemon-showdown-master/.sim-dist/tools/random-player-ai.js:70:35)
    at RandomPlayerAI.receiveLine (/usr/src/pokemon-showdown-master/.sim-dist/battle-stream.js:232:38)
    at RandomPlayerAI.receive (/usr/src/pokemon-showdown-master/.sim-dist/battle-stream.js:224:9)
    at RandomPlayerAI.start (/usr/src/pokemon-showdown-master/.sim-dist/battle-stream.js:218:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:5460) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:5460) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
At [/sim/tools/random-player-ai.ts#L102](https://github.com/smogon/pokemon-showdown/blob/master/sim/tools/random-player-ai.ts#L102) RandomPlayerAI assumes teamsize is at least 2, which may be not necessarily true.

This can be fixed by checking the teamsize before running the status check on allies.
```
let hasAlly = false;
if(pokemon.length>1){
  hasAlly = !pokemon[i ^ 1].condition.endsWith(` fnt`);
}
```

Zarel:
Do you want to PR that or should I?

Also @scheibo do you have any opinions on this?

Liminalia:
I can do a PR if noone is against it

Zarel:
(Fixed in #6216)

scheibo:
> Also @scheibo do you have any opinions on this?

Sorry, missed this bug (though apparently I had seen the PR). Thank you for finding and fixing.

PR #6216, opened 2019-12-24
Fix RandomPlayerAI wrong teamsize assumption

Liminalia:
#6212


sim/tools/random-player-ai.ts
```diff
@@ -99,7 +99,10 @@ export class RandomPlayerAI extends BattlePlayer {
 
 				// Filter out adjacentAlly moves if we have no allies left, unless they're our
 				// only possible move options.
-				const hasAlly = !pokemon[i ^ 1].condition.endsWith(` fnt`);
+				let hasAlly = false;
```
scheibo: this can all be done on one line - just `pokemon.length > 1 && ...`

PR #6218, opened 2019-12-24
Fix Eternamax Validation

YaBoiJD:


TheImmortal:
This makes it illegal in hackmons (and hackmons cup).

TheImmortal:
I'll fix this.

PR #6220, opened 2019-12-25
(wip) Tournament exporting

whalemer:
This is still a little dodgy, mostly with regards to multi-elim tournaments and how smogon formats tour rounds, but I want to put this into PRs for bikeshedding over where to put the data / other feedback.

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -1071,12 +1071,14 @@ export class Tournament extends Rooms.RoomGame {
 		this.room.update();
 	}
 	onTournamentEnd() {
+		const bracketData = this.getBracketData();
 		const update = {
 			results: (this.generator.getResults() as TournamentPlayer[][]).map(usersToNames),
 			format: this.name,
 			generator: this.generator.name,
-			bracketData: this.getBracketData(),
+			bracketData,
 		};
+		this.room.roomlog(this.generator.getBBCode());
 		this.room.add(`|tournament|end|${JSON.stringify(update)}`);
 		this.remove();
 	}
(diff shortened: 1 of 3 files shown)


server/tournaments/index.ts
```diff
@@ -1695,6 +1697,17 @@ export const commands: ChatCommands = {
 					}
 				}
 			}
+		} else if (cmd === 'export') {
```
TheJetOU: Possibly seek out the argument after `export` for future-proofing with HTML, Markdown, et al?

server/tournaments/index.ts
```diff
@@ -1695,6 +1697,17 @@ export const commands: ChatCommands = {
...
+			if (room.game && room.game.gameid === 'tournament') {
+				if (!this.can('gamemoderation', null, room)) return;
```
TheJetOU: Why do a user need auth to export the tournament? They can already do so manually and it contains no sensitive data,

Jakilutra:
> I'm not sure about the table format. Smogon conventionally formats its tournaments as so:
> 
> ```
> [B]Round 1[/B]
> 
> [B]W vs. L[B]
> ...
> ```

Well the round robin results right triangle table (or rectangle if you double it up and blank the self-pairings) is like the complement to elimination brackets I'd say. A round robin can't practically be represented by some kind of flow graph, since you are just travelling to nodes nearer or further away regardless of wins/losses  so it gets cluttered with all the paths. Likewise, an elimination brackets put into a potentially giant right triangle would have lots of blank spaces for matches that aren't allowed to happen.

I view the brackets/results table as one form of representation, and listing the matchups under each round in a single colum another. It's not practical to export an editable brackets, so there's only 1 option for the elimination.

What I personally do at the moment for round robin tournaments on the forums is 

```Player A (3-1), Player B (0-4), Player C (3-1) ```

and then just put the 6 matchups in a column under hide tags.

I think I would do the results right triangle if I really wanted to make it look nice. It's just effort to set it up. So I would request both representations to be outputted at the same time if possible.

whalemer:
This is blocked on me figuring out how smogon decides if a round consists of only byes, which realistically I'm never going to do at this point.

PR #6221, opened 2019-12-25
Body Press uses SpD boosts under Wonder Room

urkerab:
Body Press tries really hard to use the attacker's Def, and it's almost clever enough to compensate for Wonder Room, but it turns out that it then goes on to use the attacker's SpD boosts.

Sorry, but I can't remember which researcher discovered this quirk.


Zarel:
We can add a hint here, right? "Wonder room, Body Press used" is public information, right?

Zarel:
(@DaWoblefet, @Marty-D)

Marty-D:
You're right, it wouldn't be an info leak.

Zarel:
@urkerab pls do

urkerab:
Something like this?

Zarel:
In hindsight, you should probably check boosts for both Def and SpD.

PR #6223, opened 2019-12-26
Sunflora Gen 2 move legality

SamanthaLola:
Sunflora can only learn Razor Leaf at Level 10, after it evolves. It can only learn Synthesis at Level 31, before it evolves. So it can't have both.


Zarel:
Thanks!

PR #6224, opened 2019-12-27
Calculate fractional priority once per turn

urkerab:
This fixes Custap Berry, Quick Claw and Stall for Gen 8.
In particular, Skill Swap on Stall doesn't affect turn order this turn.


Zarel:
Ugh, come on, Game Freak. :/

PR #6229, opened 2019-12-28
Typescript uno/othermetas/hangman

KrisXV:


server/chat-plugins/uno.ts
```diff
@@ -216,29 +206,21 @@ class UnoGame extends Rooms.RoomGame {
 
 	/**
 	 * Overwrite the default makePlayer so it makes an UnoGamePlayer instead.
-	 * @param {User} user
-	 * @return {UnoGamePlayer}
 	 */
-	makePlayer(user) {
+	makePlayer(user: User): UnoGamePlayer {
```
whalemer: Generally we don't annotate return types unless typescript can't figure it out itself.

server/chat-plugins/uno.ts
```diff
@@ -286,46 +264,39 @@ class UnoGame extends Rooms.RoomGame {
 		return name;
 	}
 
-	/**
-	 * @param {string} msg
-	 * @param {boolean} [overrideSuppress]
-	 */
-	sendToRoom(msg, overrideSuppress = false) {
+	sendToRoom(msg: string, overrideSuppress: boolean = false) {
```
TheJetOU:
```suggestion
	sendToRoom(msg: string, overrideSuppress = false) {
```

This is inferrable by the compiler.

server/chat-plugins/uno.ts
```diff
@@ -365,53 +333,41 @@ class UnoGame extends Rooms.RoomGame {
...
-		let userList = Object.keys(this.playerTable);
+		const userList = Object.keys(this.playerTable);
 
 		let player = userList[(userList.indexOf(this.currentPlayerid) + this.direction)];
 
 		if (!player) {
 			player = this.direction === 1 ? userList[0] : userList[this.playerCount - 1];
 		}
-		return player;
+		return player as ID;
```
TheJetOU: Why not just define `player` as `let player: ID` for extra type safeness ?

server/chat-plugins/uno.ts
```diff
@@ -435,7 +401,7 @@ class UnoGame extends Rooms.RoomGame {
 		// update the unoId here, so when the display is sent to the player when the play is made
 		if (player.hand.length === 1) {
 			this.awaitUno = player.id;
-			this.unoId = Math.floor(Math.random() * 100).toString();
+			this.unoId = toID(Math.floor(Math.random() * 100).toString());
```
TheJetOU:
```suggestion
			this.unoId = Math.floor(Math.random() * 100).toString() as ID;
```

server/chat-plugins/uno.ts
```diff
@@ -526,33 +490,28 @@ class UnoGame extends Rooms.RoomGame {
...
-	 * @param {UnoGamePlayer} player
-	 * @param {number} count
-	 * @return {Card[]}
-	 */
-	onDrawCard(player, count) {
+	onDrawCard(player: UnoGamePlayer, count: number) {
 		if (typeof count === 'string') count = parseInt(count);
 		if (!count || isNaN(count) || count < 1) count = 1;
-		let drawnCards = /** @type {Card[]} */ (this.drawCard(count));
+		const drawnCards: Card[] = this.drawCard(count);
```
TheJetOU:
```suggestion
		const drawnCards = this.drawCard(count);
```

I think this is inferrable too.

server/chat-plugins/uno.ts
```diff
@@ -561,12 +520,7 @@ class UnoGame extends Rooms.RoomGame {
 		return drawnCards;
 	}
 
-	/**
-	 * @param {UnoGamePlayer} player
-	 * @param {string} unoId
-	 * @return {false | void}
-	 */
-	onUno(player, unoId) {
+	onUno(player: UnoGamePlayer, unoId: string) {
```
TheJetOU:
```suggestion
	onUno(player: UnoGamePlayer, unoId: ID) {
```

server/chat-plugins/uno.ts
KrisXV: im not sure what difference this makes

server/chat-plugins/uno.ts
KrisXV: i know but id prefer to keep it defined in the case that something is added later on that introduces potential complexities

server/chat-plugins/uno.ts
TheJetOU: You're changing code that's used in run-time to appease TypeScript, which only runs at compile-time. There's no reason to do that unless it's the only solution.

server/rooms.ts
```diff
@@ -157,6 +158,7 @@ export abstract class BasicRoom {
 		this.unoDisabled = false;
 		this.blackjackDisabled = false;
 		this.hangmanDisabled = false;
+		this.giveaways = Object.create(null);
```
Zarel: ??? no!

Zarel:
```suggestion
		this.giveaways = null;
```

server/rooms.ts
```diff
@@ -101,6 +101,7 @@ export abstract class BasicRoom {
 	unoDisabled: boolean;
 	blackjackDisabled: boolean;
 	hangmanDisabled: boolean;
+	giveaways: {wifi: any, gts: any};
```
```suggestion
	giveaways: {wifi: any, gts: any} | null;
```

server/rooms.ts
Zarel: Actually I'm not sure this should be in the main room object at all.

Zarel:
Going to just merge because why not

PR #6231, opened 2019-12-29
Shell trap takes effect directly after a physical hit

urkerab:
In Doubles, Shell Trap fires directly after the physical hit that triggers hit, before anyone else would normally move. Seen in this video: https://www.youtube.com/watch?v=zFnIvJfgvVk&t=270 where Grimmsnarl hits Turtonator with Sucker Punch and Shell Trap then fires before Drednaw or Ninetales can move. Behaviour confirmed to apply also to Gen 7 by @SadisticMystic.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -16594,12 +16594,11 @@ let BattleMovedex = {
 		beforeTurnCallback(pokemon) {
 			pokemon.addVolatile('shelltrap');
 		},
-		// TODO: In order to correct PP usage, after spread move order has been reworked,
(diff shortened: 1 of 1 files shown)


Zarel:
Nice, thanks!

Zarel:
Does `prioritizeAction` still work in combination with Gen 8's "constantly recalc speed" though?

urkerab:
I was assuming #6143 had fixed it to work properly, given that it has to work with After You, Pledge moves, and Round.

Zarel:
Oh, yes, it does.

PR #6232, opened 2019-12-29
Fix Neutralizing Gas

ZardMX:
If you switch out the Neutralizing Gas Pokemon once, the ability will simply never work again in the battle.
https://replay.pokemonshowdown.com/gen8customgame-1037347095-5gmu42swmzhcxgzpyfytlk4kkzzkf3gpw
This fixes that.


ZardMX:
Neutralizing Gas doesn't nullify Gulp Missile.
https://www.youtube.com/watch?v=9SnsRMMqgYM&feature=share (Thanks KennedyLFC)

Zarel:
Why is `abilityData` even defined that way? o_o

Zarel:
I didn't because I was working on bf8ed1861de58 to fix type definitions.

PR #6237, opened 2019-12-30
Fix /statcalc without pokemon argument

asgdf:
Non-existant templates have a base stat object now (with every stat at 0), so just checking whether there is a base stat object isn't enough to filter out nonexistent templates.

### DIFF
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -1273,7 +1273,7 @@ const commands = {
 
 			if (!pokemon) {
 				let testPoke = Dex.getTemplate(arg);
-				if (testPoke.exists && testPoke.baseStats) {
+				if (testPoke.exists) {
 					pokemon = testPoke.baseStats;
 					baseSet = true;
 					continue;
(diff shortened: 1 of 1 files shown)


asgdf:
(bump - sorry, this command is really useful and I miss using it)

TheImmortal:
Is the ``testPoke.baseStats`` check necessary then

asgdf:
_Probably_ not but I don't know for sure if there are situations where there are no base stats

server/chat-commands/info.js
```diff
@@ -1273,7 +1273,7 @@ const commands = {
 
 			if (!pokemon) {
 				let testPoke = Dex.getTemplate(arg);
-				if (testPoke.baseStats) {
+				if (testPoke.exists && testPoke.baseStats) {
```
```suggestion
				if (testPoke.exists) {
```

Zarel: Oh, it looks like there was already discussion on this. No, the latter check should be removed. You're right that you don't know for sure if there are situations with no `.baseStats`, but I _do_ know for sure that if there are, they're bugs and should crash so we can fix them.

asgdf: Alright, fair.

PR #6240, opened 2019-12-31
Rename Vice Grip to Vise Grip

KrisXV:
Vise Grip was only renamed in learnsets.js before and not throughout the repository.


urkerab:
Does this need an alias? I think Feint Attack got one for Faint Attack.

Zarel:
@urkerab The alias already exists. It was mostly renamed already; this commit is just picking up a few stragglers.

PR #6244, opened 2019-12-31
Fix displayed target of move-specific immunity in Gen 1 and 2

urkerab:
Followup to #5904; Leech Seed into a Grass type and Dream Eater into anything awake are calculating the immunity correctly but then erroneously crediting it to the attacker instead of the defender.

### DIFF
--- a/data/mods/gen1/scripts.js
+++ b/data/mods/gen1/scripts.js
@@ -273,7 +273,7 @@ let BattleScripts = {
 		}
 		hitResult = this.singleEvent('TryImmunity', move, null, target, pokemon, move);
 		if (hitResult === false) {
-			this.add('-immune', pokemon);
+			this.add('-immune', target);
 			return false;
 		}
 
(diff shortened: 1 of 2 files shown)


Zarel:
@TheImmortal can you not merge PRs that don't pass Travis?

Zarel:
Like if it's a Travis bug, either fix it yourself or wait for me to approve it...

KrisXV:
Isn't the only reason the build failed because the build never got a chance to finish before being merged?

Zarel:
Yes, which is also something I don't want people to do.

TheImmortal:
I looked at the changes, Zarel. It was obvious that this wouldn’t fail.

TheImmortal:
Sounds like a Travis bug that it automatically fails merged PRs :^)

Zarel:
I mean, yes, it's 100% a Travis bug.

Zarel:
But I don't want you relying on "obvious that this wouldn’t fail". You've definitely merged PRs that have failed. I don't even trust my own judgment on this, much less yours.

PR #6245, opened 2020-01-01
Add Pet Mod of the Season

KrisXV:


Zarel:
Was the modified Species Clause removed? I don't see it in the diff.

KrisXV:
Yeah, i changed the dex numbers in the mod file so there's no need to worry about it @Zarel

Zarel:
This will still lead to client bugs. Can you just ask them to choose different mons than Rotom?

KrisXV:
No..., if it will lead to client bugs ill leave the numbers as normal and just modify species clause again

Zarel:
You'd need to call it something besides "Species Clause".

Specifically, client assumes that if there exists a rule named "Species Clause", there will be at most one Pokémon for every dex number.

KrisXV:
Does this work @Zarel ?

I didn't add it to the CUSTOM-RULES.md documentation because it wont work outside of this format

data/mods/petmod/rulesets.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**@type {{[k: string]: ModdedFormatsData}} */
+let BattleFormats = {
+	standardpetmod: {
+		inherit: true,
+		ruleset: ['[Gen 7] OU', '!Species Clause'],
+		banlist: ['All Pokemon'],
+		onBegin() {
+			// The only validator rule this currently modifies is Species Clause, so the added rule is just this
+			this.add('rule', 'Species Clause: Limit one of each Pokémon');
```
Zarel: No, this doesn't work. This is the line client has trouble with. You can't send a message starting with `|rule|Species Clause:`.

KrisXV: ok

data/mods/petmod/rulesets.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**@type {{[k: string]: ModdedFormatsData}} */
+let BattleFormats = {
+	standardpetmod: {
+		inherit: true,
+		ruleset: ['[Gen 7] OU', '!Species Clause'],
+		banlist: ['All Pokemon'],
+		onBegin() {
+			// The only validator rule this currently modifies is Species Clause, so the added rule is just this
+			this.add('rule', 'PM Species Clause: Limit one of each Pokémon');
```
Zarel: Why not Clean Slate? Will all pet mods have the same species clause?

scoopapa: No, all pet mods do not have the same species clause, it's just this one that needs the modified one.

Zarel: Then it should not called "PM". Also the mod should not be called "petmod".

KrisXV: I’m calling it petmod for ease of client access for future pet mods

KrisXV: anyways, fixed

PR #6246, opened 2020-01-01
Pet Mod of the Season ( Clean Slate: Micro )

scoopapa:
Updated config/formats.js and added data/mods/cleanslatemicro to the mods folder.

https://www.smogon.com/forums/threads/pet-mod-of-the-season-voting-thread.3658207/post-8326596

Is there a way to get teambuilder support for this? The list of pokemon is limited to 20 pokemon, and they all have custom stats, and modified movepools, which might get confusing in the teambuilder. If someone could help with this or tell me what I need to do ( I assume something in client-teambuilder.js ), it would be much appreciated. 

Thanks for hosting Pet Mod of the Season!


scoopapa:
Should I bother with this? I just noticed Kris posted the same thing just earlier ( and seemingly redid a lot of the code ). Just make sure they know you can use multiple rotom formes without violiating species clause. I wasn't aware of the formatting necessities but I can take care of them if need be.

TheImmortal:
Kris’s is fine.

PR #6247, opened 2020-01-01
Fix Howl not raising ally's Atk

ZardMX:
Specifically done this way so it is still searchable by ``/movesearch boosts atk``


Zarel:
Just change the target to `allies`...

ZardMX:
> Just change the target to `allies`...
That doesn't seem to work? It is already ``allyTeam``

PR #6253, opened 2020-01-02
Remove tightening focus mechanic when used by dynamaxed pokemon

CK-Yong:
PR for https://github.com/smogon/pokemon-showdown/projects/3#card-29862179

~~Check the current turn's queued messages to see if the pokemon in question will be Dynamaxing this turn or not. If it is Dynamaxing, do not tighten the focus. It will be converted into Max Knuckle later.~~ Now checks all moves during resolution, and lets the `runDynamax` mechanic handle move resolution if the move is considered a Max Move.

I've tested it locally of course. If there's anything else, feel free to leave a comment :)

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2390,7 +2390,7 @@ export class Battle {
 		}
 		if (!midTurn) {
 			if (action.choice === 'move') {
-				if (!action.zmove && action.move.beforeTurnCallback) {
+				if (!action.maxMove && !action.zmove && action.move.beforeTurnCallback) {
 					this.addToQueue({
 						choice: 'beforeTurnMove', pokemon: action.pokemon, move: action.move, targetLoc: action.targetLoc,
 					});
--- a/test/sim/moves/focuspunch.js
+++ b/test/sim/moves/focuspunch.js
@@ -86,4 +86,36 @@ describe('Focus Punch', function () {
 		battle.makeChoices('move focuspunch', 'move toxic');
 		assert.strictEqual(move.pp, move.maxpp - 2);
 	});
+
+	it('should not tighten the pokemon\'s focus when Dynamaxing', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Chansey', ability: 'naturalcure', moves: ['focuspunch']}]});
+		battle.setPlayer('p2', {team: [{species: 'Venusaur', ability: 'overgrow', moves: ['magicalleaf', 'toxic']}]});
+
+		battle.makeChoices('move focuspunch dynamax', 'move magicalleaf');
+		const tighteningFocusMessage = battle.log.filter(str => str === '|-singleturn|p1a: Chansey|move: Focus Punch');
+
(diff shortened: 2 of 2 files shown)


urkerab:
Out of interest, should this apply to Beak Blast as well, to support NatDex formats?

Also, what happens if you use Focus Punch on the second turn of Dynamaxing? Has it already changed to Max Knuckle by then?

KrisXV:
Yes

CK-Yong:
> Out of interest, should this apply to Beak Blast as well, to support NatDex formats?
> 
> Also, what happens if you use Focus Punch on the second turn of Dynamaxing? Has it already changed to Max Knuckle by then?

Well spotted! I'm not familiar with Beak Blast, but a bit of Googling shows that it has the same mechanic as Focus Punch. So perhaps we could add another ticket on the board to add it? Unless the maintainers here are ok with fixes that are sneaked in?

I've also updated the code to look at the actual move that is being performed, instead of looking at whether the Pokemon is Dynamaxing in the same turn or not. By looking at the `action.maxMove` property we can decide whether the attack will be changed into a max move or not.

urkerab:
It occurred to me that I previously fixed this before for Z-moves in issue #3057, so this fix should probably be along the same lines. (But tests are always nice. As you can see, I suck at writing them.)

data/moves.js
```diff
@@ -593,7 +593,7 @@ let BattleMovedex = {
 			let success = false;
 			for (const ally of pokemon.side.pokemon) {
 				if (ally !== source && ((ally.hasAbility('sapsipper')) ||
-						(ally.volatiles['substitute'] && !move.infiltrates))) {
+					(ally.volatiles['substitute'] && !move.infiltrates))) {
```
Zarel: I get that we're inconsistent about indentation, but let's not go reindenting everything in unrelated PRs. Feel free to submit a PR with a lint rule to make this sort of thing consistent, though.

CK-Yong: Ah. That's something my IDE did (Webstorm). It picked up the editor config and "fixed it" for the entire file while I was working on it. I will revert these lines.

CK-Yong:
@urkerab that's a nice solution. I'm quite new to the engine and still learning the architecture. Thanks for the heads up, I will make the change you suggested! :)

data/moves.js
Zarel: I guess it's not worth worrying about if it's difficult to configure your IDE not to do this.

CK-Yong: Probably the next person using Webstorm would be doing the same thing. Luckily it only fixes it in like 3 places or so. I'll keep it in mind for the future then.

CK-Yong:
I've made the change that @urkerab suggested, and tested it locally. Looks like it's working properly now, and it should also work for moves like Beak Blast (not explicitly tested).

data/moves.js
CK-Yong: Not applicable anymore, since adding @urkerab 's change I managed to leave the entire moves.js file alone.

Zarel:
Cool, thanks!

PR #6254, opened 2020-01-03
Fix Pickpocket applying when being switched out through Eject Button

CK-Yong:
Also makes sure that dragging moves like Dragon Tail do activate Pickpocket.

This fixes https://github.com/smogon/pokemon-showdown/projects/3#card-25870242


Zarel:
Thanks!

PR #6256, opened 2020-01-03
Add No Switch Mod

zisko:
This adds a No Switch Mod- requested and approved here:
https://www.smogon.com/forums/threads/no-switch-rule.3653057/

### DIFF
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -636,6 +636,21 @@ let BattleFormats = {
 			}
 		},
 	},
+	noswitchmod: {
+		effectType: 'Rule',
+		name: 'No Switch Mod',
+		desc: "Prevents players from switching pokemon, except via move i.e. U-turn",
+		onBegin() {
+			this.add('rule', 'No Switch Mod: No switching allowed');
+			this.add('-message', 'No Switch Mod activated.');
+		},
+		onBeforeSwitchOut(pokemon) {
+			if (!(typeof pokemon.switchFlag === 'string')) {
+				pokemon.switchFlag = false;
+				this.add('-message', 'No Switch Mod activated.');
+			}
+		},
+	},
 	switchpriorityclausemod: {
 		effectType: 'Rule',
 		name: 'Switch Priority Clause Mod',
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -645,7 +645,7 @@ let BattleFormats = {
 			this.add('-message', 'No Switch Mod activated.');
 		},
 		onBeforeSwitchOut(pokemon) {
-			if (!(typeof pokemon.switchFlag === 'string')) {
+			if (typeof pokemon.switchFlag !== 'string') {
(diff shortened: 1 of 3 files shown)


TheImmortal:
...but why

zisko:
https://www.smogon.com/forums/threads/no-switch-rule.3653057/

@TheImmortal It was approved here for metronome battles.

data/rulesets.js
```diff
@@ -636,6 +636,21 @@ let BattleFormats = {
...
+	noswitchmod: {
+		effectType: 'Rule',
+		name: 'No Switch Mod',
+		desc: "Prevents players from switching pokemon, except via move i.e. U-turn",
+		onBegin() {
+			this.add('rule', 'No Switch Mod: No switching allowed');
+			this.add('-message', 'No Switch Mod activated.');
+		},
+		onBeforeSwitchOut(pokemon) {
+			if (!(typeof pokemon.switchFlag === 'string')) {
```
KrisXV:
```suggestion
			if (typeof pokemon.switchFlag !== 'string') {
```

TheImmortal:
> https://www.smogon.com/forums/threads/no-switch-rule.3653057/
> 
> @TheImmortal It was approved here for metronome battles.

Metronome Battle itself doesn't have switching as it's a 2v2 format. This rule is useless no actual formats use it and if you wanted it for "fun" then you can literally choose to not switch in battles.

Jakilutra:
2v2 formats clearly aren't compatible with GSC and RBY. No Switches was the historical format as far as I'm aware. I can definitely recall playing it on PO main prior to PS and it being considered established. I am not sure of the exact history. Maybe ask older players to see if it was more a coyotte508 circle thing or not.

It's one of the most fundamental bans you can do, which you can apply it to pretty much every room tour (barring formats that force you to switch I suppose) akin to blitz, so I would consider it to have great utility for RoA Tour nights for example.

Jakilutra:
kland says this:

"it was better then metronome with switches, because stat drop moves are common, i'm not sure on the actual percentage but it wouldn't surprise if with stat drop moves are the most common in mons
one of the biggest issues was pp stall during metronome due to people pulling stat drop moves constantly"

Zarel:
It's a clause, not a mod. Same as Mega Rayquaza Clause.

Zarel:
(An agreement not to do something is not a mod.)

AnnikaCodes:
@zisko sorry about the lack of communication. 

I'm fine with this as a clause. Are you still interested in working on it?

Zarel:
I'm fine with this rule existing, although I don't understand the implementation. `switchFlag` is for forced switches only and shouldn't have any effect on switches that replace a move at all. You should be looking for a trapping flag instead...

KrisXV:
superseded by https://github.com/smogon/pokemon-showdown/pull/8241

zisko:
Sorry this went stale!

PR #6257, opened 2020-01-04
feat: Make alts configurable on a per server basis for tournaments

hilts-vaughan:
A server I use has to have this off because we often having people playing from the same household. It's not a large server (< 10 people) so it's fine for our use case.

Are there objections to making this configurable?

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -372,6 +372,7 @@ exports.customavatars = {
  * tourroom - specify a room to receive tournament announcements (defaults to
  * the room 'tournaments').
  * tourannouncements - announcements are only allowed in these rooms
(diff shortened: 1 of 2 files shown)


Zarel:
Hm, some of the features of `Config.fakeladder` might also be better here.

Zarel:
I got the flu and was very out of it last week, so sorry for the late reply. I need to go back to focusing on the Stalwart implementation. In the meantime, I think if this feature is to be merged, it should probably apply to the ladder as well as to tournaments.

I _am_ interested in this feature, though.

AnnikaCodes:
You can use `Config.noipchecks` for this.

PR #6260, opened 2020-01-04
Typescript move.target

urkerab:
Also allowing the various move targets to be documented.

This found a bug in SSB!

### DIFF
--- a/sim/global-types.ts
+++ b/sim/global-types.ts
@@ -23,38 +23,23 @@ type SparseBoostsTable = Partial<BoostsTable>;
 type Nonstandard = 'Past' | 'Future' | 'Unobtainable' | 'CAP' | 'LGPE' | 'Custom';
 /**
  * Describes the acceptable target(s) of a move.
+ * adjacentAlly - Only relevant to Doubles or Triples, the move only targets an ally of the user.
+ * adjacentAllyOrSelf - The move can target the user or its ally.
+ * adjacentFoe - The move can target a foe, but not (in Triples) a distant foe.
+ * all - The move targets the field or all Pokémon at once.
+ * allAdjacent - The move is a spread move that also hits the user's ally.
+ * allAdjacentFoes - The move is a spread move.
+ * allies - The move affects all active Pokémon on the user's team.
+ * allySide - The move adds a side condition on the user's side.
+ * allyTeam - The move affects all unfainted Pokémon on the user's team.
+ * any - The move can hit any other active Pokémon, not just those adjacent.
+ * foeSide - The move adds a side condition on the foe's side.
+ * normal - The move can hit one adjacent Pokémon of your choice.
+ * randomNormal - The move targets an adjacent foe at random.
+ * scripted - The move targets the foe that damaged the user.
+ * self - The move affects the user of the move.
  */
-type MoveTarget =
-	// Only relevant to Doubles or Triples, the move targets your ally only.
-	'adjacentAlly' |
-	// This move can target your ally or yourself.
-	'adjacentAllyOrSelf' |
-	// This move can only target a foe, but not (in Triples) a distant foe.
-	'adjacentFoe' |
-	// This move targets the field or all Pokémon at once.
-	'all' |
-	// This is a spread move that also hits your ally.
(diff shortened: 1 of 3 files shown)


sim/global-types.ts
```diff
@@ -21,6 +21,41 @@ type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
 type BoostsTable = {[boost in BoostName]: number };
 type SparseBoostsTable = Partial<BoostsTable>;
 type Nonstandard = 'Past' | 'Future' | 'Unobtainable' | 'CAP' | 'LGPE' | 'Custom';
+/**
+ * Describes the acceptable target(s) of a move.
+ */
+type MoveTarget =
+	// Only relevant to Doubles or Triples, the move targets your ally only.
```
Zarel: This documentation should be in the `/** */` block, so it shows up on hover.

sim/global-types.ts
```diff
@@ -21,6 +21,41 @@ type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
...
+	// This move targets the field or all Pokémon at once.
+	'all' |
+	// This is a spread move that also hits your ally.
+	'allAdjacent' |
+	// This is a spread move.
+	'allAdjacentFoes' |
+	// This move affects all active Pokémon on your team.
+	'allies' |
+	// This move is a side effect for your side.
+	'allySide' |
```
Zarel: > petition to rename
> `allySide` -> `allySideOfField`
> and
> `allyTeam` -> `alliesIncludingInactive`

sim/global-types.ts
```diff
@@ -21,6 +21,41 @@ type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
...
+	'adjacentFoe' |
+	// This move targets the field or all Pokémon at once.
+	'all' |
+	// This is a spread move that also hits your ally.
+	'allAdjacent' |
+	// This is a spread move.
+	'allAdjacentFoes' |
+	// This move affects all active Pokémon on your team.
+	'allies' |
+	// This move is a side effect for your side.
```
Zarel: "adds a side condition"

sim/global-types.ts
```diff
@@ -21,6 +21,41 @@ type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
...
+	'allySide' |
+	// This move affects all unfainted Pokémon on your team.
+	'allyTeam' |
+	// This move can hit any other active Pokémon, even in Triples.
+	'any' |
+	// This move is a side effect for your opponent.
```
Zarel: "adds a side condition"

sim/global-types.ts
```diff
@@ -21,6 +21,41 @@ type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
...
+	'allAdjacent' |
+	// This is a spread move.
+	'allAdjacentFoes' |
+	// This move affects all active Pokémon on your team.
+	'allies' |
+	// This move is a side effect for your side.
+	'allySide' |
+	// This move affects all unfainted Pokémon on your team.
+	'allyTeam' |
+	// This move can hit any other active Pokémon, even in Triples.
```
Zarel: "not just adjacent"

sim/global-types.ts
urkerab: I'll take your word for it. (I develop for PS on Linux so I don't use VS Code.)

Zarel:
I mentioned on Discord:

> petition to rename
> `allySide` -> `allySideOfField`
> and
> `allyTeam` -> `alliesIncludingInactive`

your thoughts?

urkerab:
If you're asking me, then I don't think the rename is helpful, what with the improved documentation (!)

Zarel:
I don't think the improved documentation is easily to accessible to anyone who doesn't use VS Code or some other IDE with deep TypeScript integration.

Zarel:
Which I'm pretty sure includes a lot of researchers like DaWob.

Zarel:
This is a strict improvement and it's been floating too long, though (sorry, I got sick a few days ago.)

PR #6261, opened 2020-01-04
Dynamax Cannon now targets both foes instead of just one in doubles

CK-Yong:
This fixes https://github.com/smogon/pokemon-showdown/projects/3#card-29789538

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -4369,7 +4369,7 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {protect: 1},
 		secondary: null,
-		target: "normal",
+		target: "allAdjacentFoes",
 		type: "Dragon",
 	},
 	"dynamicpunch": {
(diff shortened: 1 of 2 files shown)


Zarel:
...this does not need a test. The test adds nothing the change doesn't.

DaWoblefet:
Wait a second here, who said Dynamax Cannon was a spread move? This is blatantly false; it's just like any other single target attack.

Zarel:
Good to know.

Zarel:
@TZubiri I'm removing your write access to Mechanics Bugs. You _must_ ask researchers before adding a mechanics bug.

CK-Yong:
Thanks for verifying guys, I didn't test it myself, just assumed that someone tested this in-game and posted the ticket.

TZubiri:
@Zarel, Removing write access is a good idea since I'm no longer active in triaging issues from the smogon thread.

However I must vocalize my recommendation, in case someone chooses to take over triaging, that mechanics bugs be reported as github issues rather than a kanban board item, so that more detail can fit, like in this example: https://github.com/smogon/pokemon-showdown/issues/6089

In the case of a false bug, the source trail can then be audited, to identify the fault and address it systematically. In this case I think it was a misinterpreted third party website.

I wish you all the best in the project.

PR #6263, opened 2020-01-05
Create flag for abilities not affected by Gastro Acid

urkerab:
And use it everywhere to simplify checks for illegal copying, overwriting or swapping of Abilities.

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -273,6 +273,7 @@ let BattleAbilities = {
 				move.multihit = 3;
 			}
 		},
+		permanent: true,
 		id: "battlebond",
 		name: "Battle Bond",
 		rating: 4,
@@ -464,6 +465,7 @@ let BattleAbilities = {
 		},
 		// Permanent sleep "status" implemented in the relevant sleep-checking effects
 		isUnbreakable: true,
+		permanent: true,
 		id: "comatose",
 		name: "Comatose",
 		rating: 3.5,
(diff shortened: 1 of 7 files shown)


Marty-D:
I don't know if Flower Gift is on this list now, but it wasn't in any other gen and neither was Forecast. Zen Mode only got on this list in Gen 7.

I disagree with removing detail in the descriptions.

urkerab:
Sorry, despite knowing that Zen Mode got added to the list in Gen 7, I got the list mixed up with the list of Abilities that can't be copied (but can be suppressed or replaced) except for some reason by Skill Swap.

Zen Mode not being in the list in Gen 6 means that Simple Beam, Skill Swap and Worry Seed now automatically special-case it, avoiding the need for an override.

I've reverted all of the descriptions.

Zarel:
I think I like this, although I'm not sold on `permanent` as a flag name. `isUnchangeable`?

urkerab:
> I'm not sold on permanent as a flag name

https://github.com/smogon/pokemon-showdown/commit/d8bea75618b1382305b7f8071c204fba3ebb054c#commitcomment-36425902

**Zarel** replied on 14 Dec 2019

> I like that name. Let's do it.

Zarel:
I didn't imagine it would just be a property. As a property, I think `isPermanent` makes more sense.

Zarel:
Superseded by #7614

PR #6269, opened 2020-01-08
Clarify Wandering spirit's shortdesc

Dawnofares:
The way it is currently worded, the pokemon with the ability has to hit another pokemon with a contact move, when it should be the other way around.


data/abilities.js
```diff
@@ -4461,7 +4461,7 @@ let BattleAbilities = {
 	},
 	"wanderingspirit": {
 		desc: "The Pokémon exchanges Abilities with a Pokémon that hits it with a move that makes direct contact.",
-		shortDesc: "Exchanges abilities when hitting a Pokémon with a contact move.",
+		shortDesc: "Exchanges abilities when being hit with a contact move.",
```
KrisXV:
```suggestion
		shortDesc: "Exchanges abilities when hit with a contact move.",
```

PR #6270, opened 2020-01-09
Re-add Evasion Abilities Clause to Gen 6/7 DOU

DaWoblefet:
When @TheImmortal removed Evasion Abilities Clause, it wasn't re-added back to Gen 7 and Gen 6 DOU which should have the clause (Gen 5 already has it). DOU tier leaders confirm it should only be removed from Gen 8. According to Pokesartoolcay, 2v2 Doubles doesn't need Evasion Abilities Clause so I did not re-add it there.
image

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -1303,7 +1303,7 @@ let Formats = [
 
 		mod: 'gen7',
 		gameType: 'doubles',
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Never mind I'm an idiot.

PR #6274, opened 2020-01-09
Slowpoke-Galar can use Eviolite

KrisXV:
Despite not having a coded evolution method or evolutions that exist within the games, Eviolite still works on Slowpoke-Galar.


TheImmortal:
I don’t know if this is the correct fix. It should only be legal in NFE if it’s legal in LC.

KrisXV:
Should I just hardcode it?

PR #6279, opened 2020-01-12
Improve Dragon Darts

HoeenCoder:
This moves most of dragon dart's logic to `Pokemon.getMovetargets` by giving a new move target type called `allFoesSmart` (Feel free to bikeshed about the name, I'm not too attatched).

Put this up as a draft since I need to do something about being unable to tell if a move's target was forcibly changed during a `RedirectTarget` event. Looking for feedback/assistance on this.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -4121,9 +4121,6 @@ let BattleMovedex = {
 			// We need to do this before targeting is calculated.
 			pokemon.addVolatile('smarttarget');
 		},
-		onModifyMove(move, pokemon, target) {
-			if (target.side.active.length === 2) move.spreadModifier = 1;
-		},
 		target: "normal",
 		type: "Dragon",
 		gmaxPower: 130,
--- a/data/moves.js
+++ b/data/moves.js
@@ -4116,20 +4116,13 @@ let BattleMovedex = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
-		multihit: 2,
 		secondary: null,
+		beforeMoveCallback(pokemon) {
+			// We need to do this before targeting is calculated.
+			pokemon.addVolatile('smarttarget');
+		},
 		onModifyMove(move, pokemon, target) {
-			if (target.side.active.length === 2) {
-				move.multihit = 1;
-				move.spreadModifier = 1;
-				move.target = "allAdjacentFoes";
-				for (const currentTarget of target.side.active) {
-					if (currentTarget.volatiles['protect'] || !currentTarget.runImmunity('Dragon') || currentTarget.fainted) {
-						move.multihit = 2;
-						break;
-					}
-				}
-			}
+			if (target.side.active.length === 2) move.spreadModifier = 1;
 		},
 		target: "normal",
 		type: "Dragon",
--- a/data/statuses.js
+++ b/data/statuses.js
@@ -462,6 +462,30 @@ let BattleStatuses = {
 			return this.chainModify([0x14CD, 0x1000]);
 		},
 	},
+	smarttarget: {
+		name: 'smarttarget',
+		id: 'smarttarget',
+		num: 0,
+		duration: 1,
+		onRedirectTargetPriority: -1,
+		onRedirectTarget(target, source, source2, move) {
+			// Calculate smart targeting, goes last to allow redirection moves to override smart targeting
+			let targets = source.nearbyFoes();
+			// Handle singles/empty slots in doubles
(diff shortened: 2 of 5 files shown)


sim/pokemon.ts
```diff
@@ -628,6 +628,42 @@ export class Pokemon {
 				this.battle.retargetLastMove(targets[targets.length - 1]);
 			}
 			break;
+		// FIXME client support for this move target (allow user to select a target in doubles)
```
Zarel: Client PR should happen first.

data/moves.js
```diff
@@ -4116,22 +4116,11 @@ let BattleMovedex = {
...
-				move.spreadModifier = 1;
-				move.target = "allAdjacentFoes";
-				for (const currentTarget of target.side.active) {
-					if (currentTarget.volatiles['protect'] || !currentTarget.runImmunity('Dragon') || currentTarget.fainted) {
-						move.multihit = 2;
-						break;
-					}
-				}
-			}
+			if (target.side.active.length === 2) move.spreadModifier = 1;
```
```suggestion
			move.spreadModifier = 1;
```

sim/pokemon.ts
```diff
@@ -628,6 +628,42 @@ export class Pokemon {
...
+		case 'allFoesSmart':
+			// Check for redirection first
+			const newTarget = this.battle.priorityEvent('RedirectTarget', this, this, this.battle.dex.getActiveMove(move), target);
+			this.battle.debug(`Original Target: ${target.side.name} ${target.species}, New Target: ${newTarget.side.name} ${newTarget.species}`);
```
Zarel: remember to remove this

sim/pokemon.ts
```diff
@@ -628,6 +628,42 @@ export class Pokemon {
 				this.battle.retargetLastMove(targets[targets.length - 1]);
 			}
 			break;
+		// FIXME client support for this move target (allow user to select a target in doubles)
+		case 'allFoesSmart':
+			// Check for redirection first
+			const newTarget = this.battle.priorityEvent('RedirectTarget', this, this, this.battle.dex.getActiveMove(move), target);
```
Zarel: Is Dragon Darts even redirectable? I thought no.

Zarel: Ohhh I see from further on, interesting.

sim/pokemon.ts
Zarel: Does selecting a target even do anything? If no, it shouldn't matter.

HoeenCoder:
Made significant changes

data/moves.js
urkerab: Or you could just put the `spreadModifier` directly in the move data?

Zarel: I think this probably shouldn't be considered a spread move at all, but a multihit move that smart-targets every hit.

Zarel:
Wow, we had a Dragon Darts PR a monthish ago and I forgot about it?

HoeenCoder:
Yeah my plan is/was to resume work when I get time, though since its high priority if you want to work on it, be my guest.

sim/global-types.ts
```diff
@@ -250,7 +250,7 @@ interface EventMethods {
 	onNegateImmunity?: ((this: Battle, pokemon: Pokemon, type: string) => boolean | void) | boolean
 	onOverrideAction?: (this: Battle, pokemon: Pokemon, target: Pokemon, move: ActiveMove) => string | void
 	onPrepareHit?: CommonHandlers['ResultSourceMove']
-	onRedirectTarget?: (this: Battle, target: Pokemon, source: Pokemon, source2: Effect, move: ActiveMove) => Pokemon | void
+	onRedirectTarget?: (this: Battle, target: Pokemon, source: Pokemon, source2: Effect, move: ActiveMove) => Pokemon | Pokemon[] | void
```
Zarel: The relayVar is of type `Pokemon`, so this return type will actively cause crashes if someone ever adds a lower-priority `RedirectTarget` handler.

data/statuses.js
```diff
@@ -462,6 +462,31 @@ let BattleStatuses = {
 			return this.chainModify([0x14CD, 0x1000]);
 		},
 	},
+	smarttarget: {
+		name: 'smarttarget',
+		id: 'smarttarget',
+		num: 0,
+		duration: 1,
+		onRedirectTargetPriority: -1,
+		onRedirectTarget(target, source, source2, move) {
```
Zarel: I'm actually really confused how this handler isn't applying for redirected moves.

PR #6285, opened 2020-01-14
Add a GroupSymbol type

TheJetOU:
I tried dynamically generating the type but it turned out to be
too much of a headache, but it doesn't seem worth it to block on.

### DIFF
--- a/server/global-types.ts
+++ b/server/global-types.ts
@@ -2,6 +2,21 @@ type ChildProcess = import('child_process').ChildProcess;
 
 type Config = typeof import('../config/config-example') & AnyObject;
 
+type ServerGroup =
+	| '~'
+	| '&'
+	| '#'
+	| '★'
+	| '*'
+	| '@'
+	| '%'
+	| '☆'
+	| '+'
+	| ' '
+	| '‽'
+	| '!'
+
+
 // Chat
 type CommandContext = Chat.CommandContext;
 type PageContext = Chat.PageContext;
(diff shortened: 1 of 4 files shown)


server/users.ts
```diff
@@ -642,7 +642,7 @@ export class User extends Chat.MessageContext {
 			return true;
 		}
 
-		let group: string = '';
+		let group: ServerGroup = ' ';
```
Zarel: This is a mechanical change and I'm not sure I like it.

server/global-types.ts
```diff
@@ -2,6 +2,21 @@ type ChildProcess = import('child_process').ChildProcess;
 
 type Config = typeof import('../config/config-example') & AnyObject;
 
+type ServerGroup =
```
Zarel: I'd call this `GroupSymbol`.

server/global-types.ts
```diff
@@ -2,6 +2,21 @@ type ChildProcess = import('child_process').ChildProcess;
...
+	| '~'
```
Zarel: Why are these all on their own lines, anyway? Seems unnecessarily verbose.

Zarel: Also, even if TypeScript allows a leading `|`, I don't think it's very readable.

server/users.ts
TheJetOU: The initial value doesn't matter here, because it's assigned in an if-else so it's guaranteed to exist. Giving it a closer look though, not initializing it with any value (undefined) seems shorter.

server/global-types.ts
TheJetOU: I thought collapsing them into one line would make it long, but it turned out to be 90 col so it's OK.

Zarel:
Looks good, thanks!

PR #6287, opened 2020-01-16
Make Steadfast process flinches after Dynamax

urkerab:
Reported [here](https://www.smogon.com/forums/posts/8340882).

I changed Steadfast rather than Dynamax because I wanted Inner Focus to have the same priority as Dynamax, so it was easier to change the other event user.


Marty-D:
Now that this came up I see that it's actually Inner Focus and Dynamax that are implemented wrong. They shouldn't prevent flinching when it comes time to flinch, but rather prevent the flinch flag from being set on the Pokemon at all when the flinching effect happened. Inner Focus doesn't remove the flag either; you can't get hit by Fake Out, gain Inner Focus later that turn and then move freely. Once it's set you're flinching for the rest of the turn no matter how many times Dancer and Instruct try to make you move again.

urkerab:
- Steadfast now activates after the Flinch message to match the cart
- Fake Out vs. Inner Focus still allows the target to move
- Mold Breaker Fake Out vs. Inner Focus still flinches the target
- Mold Breaker Fake Out vs. Dynamax Inner Focus still allows the target to move

Marty-D:
Neat, thanks!

PR #6289, opened 2020-01-16
Propagate promises to the highest level

TheJetOU:
~`Roomlog#setupRoomlogStream` has been changed to only be async,
because it was using async functions so it was never really
"sync" even if the sync param was true.~

See: https://github.com/smogon/pokemon-showdown/pull/6289#pullrequestreview-344125695.

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -34,10 +34,45 @@ interface PendingUpdate {
 	pendingDataFetcher: (() => string | Buffer) | null;
 	pendingOptions: object | null;
 	throttleTime: number; // throttling until time (0 for no throttle)
-	throttleTimer: NodeJS.Timer | null;
+	throttleTimer: CancellableTimerPromise<void> | null;
+	finish: Promise<void>;
 }
 
 const pendingUpdates = new Map<string, PendingUpdate>();
+
+/**
+ * Givn an async function and ms, this will create a promise that resolves
+ * when the timer and async function ends
+ *
+ * It also has a `cancel` function that will clear the timer
+ * and resolve the promise
+ */
+class CancellableTimerPromise<T, U extends T | PromiseLike<T> | undefined = undefined> {
+	promise: Promise<T>;
+	private readonly resolveVal: U | undefined;
+	private timeout: NodeJS.Timeout;
+	private resolve: (value?: any) => void;
+	constructor(fn: () => Promise<T>, ms: number, resolveVal?: U) {
+		this.timeout = null!;
+		this.resolve = null!;
+		this.resolveVal = resolveVal;
+		this.promise = new Promise<T>(resolve => {
+			const timeout = setTimeout(async () => {
+				await fn();
+				resolve(resolveVal);
+			}, ms);
+			this.timeout = timeout;
+			this.resolve = resolve;
+		});
+	}
+	/**
+	 * Clears the timer and resolves the promise
+	 */
+	cancel() {
+		clearTimeout(this.timeout);
+		this.resolve(this.resolveVal);
+	}
+}
 
 class FSPath {
 	path: string;
@@ -158,7 +193,7 @@ class FSPath {
 	 * No synchronous version because there's no risk of race conditions
 	 * with synchronous code; just use `safeWriteSync`.
 	 */
-	writeUpdate(dataFetcher: () => string | Buffer, options: object = {}) {
+	async writeUpdate(dataFetcher: () => string | Buffer, options: object = {}) {
 		if (Config.nofswriting) return;
 		const pendingUpdate: PendingUpdate | undefined = pendingUpdates.get(this.path);
 
@@ -170,13 +205,15 @@ class FSPath {
 			pendingUpdate.pendingOptions = options;
 			if (pendingUpdate.throttleTimer && throttleTime < pendingUpdate.throttleTime) {
 				pendingUpdate.throttleTime = throttleTime;
-				clearTimeout(pendingUpdate.throttleTimer);
-				pendingUpdate.throttleTimer = setTimeout(() => this.checkNextUpdate(), throttleTime - Date.now());
+				pendingUpdate.throttleTimer.cancel();
+				pendingUpdate.throttleTimer = new CancellableTimerPromise(async () => {
+					await this.checkNextUpdate();
+				}, throttleTime - Date.now());
 			}
-			return;
+			return pendingUpdate.finish;
 		}
 
-		this.writeUpdateNow(dataFetcher, options);
+		return this.writeUpdateNow(dataFetcher, options);
 	}
 
 	writeUpdateNow(dataFetcher: () => string | Buffer, options: object) {
@@ -188,9 +225,10 @@ class FSPath {
 			pendingOptions: null,
 			throttleTime,
 			throttleTimer: null,
+			finish: this.safeWrite(dataFetcher(), options).then(() => this.finishUpdate()),
 		};
 		pendingUpdates.set(this.path, update);
-		void this.safeWrite(dataFetcher(), options).then(() => this.finishUpdate());
+		return update.finish;
 	}
 	checkNextUpdate() {
 		const pendingUpdate = pendingUpdates.get(this.path);
@@ -204,9 +242,9 @@ class FSPath {
 			return;
 		}
 
-		this.writeUpdateNow(dataFetcher, options);
+		return this.writeUpdateNow(dataFetcher, options);
 	}
-	finishUpdate() {
+	async finishUpdate() {
 		const pendingUpdate = pendingUpdates.get(this.path);
 		if (!pendingUpdate) throw new Error(`FS: Pending update not found`);
 		if (!pendingUpdate.isWriting) throw new Error(`FS: Conflicting update`);
@@ -214,11 +252,13 @@ class FSPath {
 		pendingUpdate.isWriting = false;
 		const throttleTime = pendingUpdate.throttleTime;
 		if (!throttleTime || throttleTime < Date.now()) {
-			this.checkNextUpdate();
+			await this.checkNextUpdate();
 			return;
 		}
-
-		pendingUpdate.throttleTimer = setTimeout(() => this.checkNextUpdate(), throttleTime - Date.now());
+		pendingUpdate.throttleTimer = new CancellableTimerPromise(async () => {
+			await this.checkNextUpdate();
+		}, throttleTime - Date.now());
+		return pendingUpdate.throttleTimer.promise;
 	}
 
 	append(data: string | Buffer, options: object = {}) {
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -304,10 +304,8 @@ export const commands: ChatCommands = {
 				global.Dex = require('../../sim/dex').Dex;
 				// rebuild the formats list
 				delete Rooms.global.formatList;
-				// respawn validator processes
-				void TeamValidatorAsync.PM.respawn();
-				// respawn simulator processes
-				void Rooms.PM.respawn();
+				// respawn validator and simulator processes
+				await Promise.all([TeamValidatorAsync.PM.respawn(), Rooms.PM.respawn()]);
 				// broadcast the new formats list to clients
 				Rooms.global.send(Rooms.global.formatListText);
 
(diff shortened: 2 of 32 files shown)


scheibo:
Not to say "I told you so", but propagating promises to the top level is what I have been preaching since day 1. Thank you for taking this on.

server/index.ts
```diff
@@ -105,12 +106,16 @@ import {Ladders} from './ladders';
 global.Ladders = Ladders;
 
 import {Chat} from './chat';
+// tslint:disable-next-line: no-floating-promises - cannot be propogated any higher
+Chat.loadTranslations();
```
monsanto: Remove tslint comment and 
```suggestion
void Chat.loadTranslations();
```

TheJetOU: I'm not sure about this. I like the clarity / explicitness of the TSLint comment. Maybe voiding it + comment is a good compromise?

lib/fs.ts
```diff
@@ -164,7 +164,7 @@ class FSPath {
 			return;
 		}
 
-		this.writeUpdateNow(dataFetcher, options);
+		return this.writeUpdateNow(dataFetcher, options);
```
Zarel: This current return value is meaningless which is why it's not propagated.

I think if we wanted to return a Promise from `writeUpdate`, it should be a meaningful Promise.

Zarel: A meaningful response might be a Promise that resolves the next time the update was written. This would require the `if (pendingUpdate)` branch to return a Promise.

server/index.ts
Zarel: I'm not sure what the comment adds that the `void` doesn't?

(Also it's spelled "propagated".)

I think if it only happened once ever, it would make sense, but if you want to comment `void` every time, I mean, just explain what `void` means in `CONTRIBUTING.md`. The purpose of a comment is definitely _not_ to explain what a keyword means.

server/punishments.ts
```diff
@@ -172,13 +172,14 @@ export const Punishments = new class {
...
 	]);
-	constructor() {
-		setImmediate(() => {
-			void Punishments.loadPunishments();
-			void Punishments.loadRoomPunishments();
-			void Punishments.loadBanlist();
-			void Punishments.loadSharedIps();
-		});
+
+	load() {
```
Zarel: I still don't like the requirement that you need to call `load` after requiring this file, or the file will crash. I really don't like sacrificing ergonomics for some random purity requirement that has zero practical benefit.

Calling `load` at the bottom of this file will give nearly the same benefits, anyway. I'm pretty sure.

server/roomlogs.ts
```diff
@@ -117,7 +118,7 @@ export class Roomlog {
 		this.modlogStream = stream;
 		this.sharedModlog = true;
 	}
-	async setupRoomlogStream(sync = false) {
+	async setupRoomlogStream() {
```
Zarel: > `Roomlog#setupRoomlogStream` has been changed to only be async,
> because it was using async functions so it was never really
> "sync" even if the sync param was true.

I don't care; the `sync` parameter gets around a race condition. Call it `partialSync` if you think the name is misleading, but please don't introduce bugs for purity reasons.

Zarel:
> Not to say "I told you so", but propagating promises to the top level is what I have been preaching since day 1. Thank you for taking this on.

@scheibo This would only be an "I told you so" if we got a bug that this fixes. Was this PR made in response to a bug?

server/index.ts
TheJetOU: > I think if it only happened once ever, it would make sense, but if you want to comment void every time, I mean, just explain what void means in CONTRIBUTING.md. The purpose of a comment is definitely not to explain what a keyword means.

There are other cases that are not just "cannot be propagated higher" where I'd like to keep a comment, like https://github.com/TheJetOU/Pokemon-Showdown/blob/propogate-promises/server/room-battle.ts#L561.

So, I'll just use ``void`` with no comment if it simply cannot be propagated higher, and ``void`` + comment in those special cases.

server/punishments.ts
TheJetOU: > I still don't like the requirement that you need to call `load` after requiring this file, or the file will crash. I really don't like sacrificing ergonomics for some random purity requirement that has zero practical benefit.

Needing to call `load` after importing definitely isn't useless, it's nice to be able to be sure that state was loaded before using `Punishments.userids` or whatever.

> Calling `load` at the bottom of this file will give nearly the same benefits, anyway. I'm pretty sure.

It's currently ran in `index.ts`, which is where PS itself imports Punishments. I'm not sure if it's a good idea to make synchronous loading an inherent part of Punishments.

server/roomlogs.ts
TheJetOU: > I don't care; the sync parameter gets around a race condition. Call it partialSync if you think the name is misleading, but please don't introduce bugs for purity reasons.

+1

lib/fs.ts
TheJetOU: I'm not sure how it's meaningless? It returns `writeUpdateNow` which returns `safeWrite` who only resolves when the file is written to.

server/punishments.ts
Zarel: Hm, yeah, I can see the merits of that approach. I suppose I'll leave it up to you.

TheJetOU: I'll choose to keep the current status quo of this PR.

lib/fs.ts
```diff
@@ -176,9 +184,14 @@ class FSPath {
 			pendingOptions: null,
 			throttleTime,
 			throttleTimer: null,
+			finish: new Promise(async resolve => {
+				await this.safeWrite(dataFetcher(), options);
+				await this.finishUpdate();
+				resolve();
+			}) as Promise<void>,
```
urkerab: Would it be worth while writing this without involving an explicit Promise? I think you could do it with an immediately invoked async arrow function, although something like this might possibly be less confusing:

          finish: this.writeAndFinish(dataFetcher, options),

    private async writeAndFinish(dataFetcher: () => string | Buffer, options: object) {
       await this.safeWrite(dataFetcher(), options);
       await this.finishUpdate();
    }

monsanto: Either is preferable to using an explicit promise, I'd lean towards async IIFE

server/chat-commands/admin.ts
```diff
@@ -299,9 +299,9 @@ export const commands: ChatCommands = {
 				// rebuild the formats list
 				delete Rooms.global.formatList;
 				// respawn validator processes
-				void TeamValidatorAsync.PM.respawn();
+				await TeamValidatorAsync.PM.respawn();
 				// respawn simulator processes
-				void Rooms.PM.respawn();
+				await Rooms.PM.respawn();
 				// broadcast the new formats list to clients
```
monsanto: Should these run in parallel?

Zarel: I agree, these should probably be in parallel.

Zarel: `Promise.all` will let you parallelize them.

TheJetOU:
I'll put this on hold until #6342 gets merged because typescript-eslint's `no-floating-promises` returns an error on `Promise<void> | void` whereas TSLint didn't.

TheJetOU:
Actually, disregard my last post. I'd hate for this to die in a backlog.

Bumping this, @Zarel.

lib/fs.ts
```diff
@@ -176,9 +184,13 @@ class FSPath {
 			pendingOptions: null,
 			throttleTime,
 			throttleTimer: null,
+			finish: (async () => {
```
Zarel: This seems like a slower way to write

`finish: this.safeWrite(dataFetcher(), options).then(() => this.finishUpdate()),`

lib/fs.ts
```diff
@@ -192,21 +204,29 @@ class FSPath {
...
 		const throttleTime = pendingUpdate.throttleTime;
 		if (!throttleTime || throttleTime < Date.now()) {
-			this.checkNextUpdate();
+			await this.checkNextUpdate();
 			return;
 		}
 
-		pendingUpdate.throttleTimer = setTimeout(() => this.checkNextUpdate(), throttleTime - Date.now());
+		let timeout: NodeJS.Timeout;
+		const promise: Promise<void> = new Promise(resolve => {
```
Zarel: I don't think this makes sense as a Promise? The timer is cancelable.

Zarel: A Promise must either resolve or reject, but this Promise isn't resolving if the timer is canceled.

I would think this should be designed in such a way that it either resolves or rejects if the timer is canceled.

Zarel: In other words, a cancelable timer is fine, but the entire cancellation process needs to be handled inside the Promise.

lib/fs.ts
```diff
@@ -34,12 +34,46 @@ interface PendingUpdate {
...
 const pendingUpdates = new Map<string, PendingUpdate>();
 
+/**
+ * Givn an async function and ms, this will create a promise that resolves
+ * when the timer and async function ends
+ *
+ * It also has a `cancel` function that will clear the timer
+ * and resolve the promise
+ */
+export class CancellableTimerPromise<T, U extends T | PromiseLike<T> | undefined = undefined> {
```
Zarel: I wonder if this could extend `Promise`.

TheJetOU: It should be able to, but I don't see _why_ it should extend it.

lib/fs.ts
```diff
@@ -34,11 +34,46 @@ interface PendingUpdate {
...
+ */
+class CancellableTimerPromise<T, U extends T | PromiseLike<T> | undefined = undefined> {
+	promise: Promise<T>;
+	private resolveVal: U | undefined;
+	// @ts-ignore - definitely assigned in the constructor
+	private timeout: NodeJS.Timeout;
+	// @ts-ignore - see above
+	private resolve: (value?: any) => void;
+	constructor(fn: () => Promise<T>, ms: number, resolveVal?: U) {
+		this.resolveVal = resolveVal;
```
Zarel: I would use

```suggestion
		this.resolveVal = resolveVal;
		this.timeout = null!;
		this.resolve = null!;
```
instead of your ignores.

Zarel:
I still don't _really_ get the point of this, but I guess it's fine.

Zarel:
This is pending a conflict resolution and the constructor change, after which it's safe to merge.

Zarel:
merge conflict :(

Zarel:
This has skipped enough restarts that I'm actually going to manually merge this for the restart.

Zarel:
Sadly, manually merging has highlighted some issues, so I'm going to punt on this for another restart.

server/users.ts
```diff
@@ -942,14 +940,14 @@ export class User extends Chat.MessageContext {
 			if (this.named) user.prevNames[this.id] = this.name;
 			this.destroy();
 
-			Punishments.checkName(user, userid, registered);
+			await Punishments.checkName(user, userid, registered);
```
Zarel: There's no reason to wait for this before finishing a rename.

server/users.ts
```diff
@@ -942,14 +940,14 @@ export class User extends Chat.MessageContext {
...
 
 			Rooms.global.checkAutojoin(user);
 			Chat.loginfilter(user, this, userType);
 			return true;
 		}
 
-		Punishments.checkName(this, userid, registered);
+		await Punishments.checkName(this, userid, registered);
```
Zarel: There's no reason to wait for this before finishing a rename.

Zarel:
So my new problem with this is that it turns a bunch of "write updates to the FS" from spawning the write to awaiting the write, which slows down the rest of the sim massively, making something sequential when it should have been parallel.

I think probably every call to `FS(...).writeUpdate` should be spawned (with `void`) rather than awaited, because every single one should be parallelized.

monsanto:
If every call to `writeUpdate` should be `void`ed it probably shouldn't return a promise at all

Zarel:
That's what I've been saying! But others think that error propagation is important. And not everyone has a global error handler like we do.

FS is in `lib/` and intended for general use; I'm fine with just `void`ing every call to it.

TheJetOU:
>  I think probably every call to FS(...).writeUpdate should be spawned (with void) rather than awaited, because every single one should be parallelized.

I still think the error should at least be handled for the people who don't have `Config.crashguard` turned on.

Zarel:
Why? That's literally exactly what `Config.crashguard` controls: how you want to handle errors.

Zarel:
I think this at least needs `writeUpdate` calls to be async. There are other performance issues in here, but the `writeUpdate` calls are the biggest.

AnnikaCodes:
This is a year-old PR from a contributor who has since left.

PR #6291, opened 2020-01-17
Add 'hide' property to Pokemon

KrisXV:
This will mostly be used for the smogdex and for the teambuilder tables.

### DIFF
--- a/data/pokedex.js
+++ b/data/pokedex.js
@@ -144,6 +144,7 @@ let BattlePokedex = {
 		color: "Red",
 		eggGroups: ["Monster", "Dragon"],
 		inheritsFrom: "charizard",
+		hide: "SpriteOnly",
 	},
 	squirtle: {
 		num: 7,
@@ -255,6 +256,7 @@ let BattlePokedex = {
 		color: "White",
 		eggGroups: ["Bug"],
 		inheritsFrom: "butterfree",
+		hide: "SpriteOnly",
 	},
 	weedle: {
 		num: 13,
@@ -430,6 +432,7 @@ let BattlePokedex = {
 		weightkg: 105,
 		color: "Black",
 		eggGroups: ["Field"],
+		hide: "SpriteOnly",
 	},
 	spearow: {
 		num: 21,
@@ -509,6 +512,7 @@ let BattlePokedex = {
 		weightkg: 6,
 		color: "Yellow",
 		eggGroups: ["Undiscovered"],
+		hide: "Minute",
 	},
 	pikachurockstar: {
 		num: 25,
@@ -523,6 +527,7 @@ let BattlePokedex = {
 		weightkg: 6,
 		color: "Yellow",
 		eggGroups: ["Undiscovered"],
+		hide: "Minute",
 	},
 	pikachubelle: {
 		num: 25,
@@ -537,6 +542,7 @@ let BattlePokedex = {
 		weightkg: 6,
 		color: "Yellow",
 		eggGroups: ["Undiscovered"],
+		hide: "Minute",
 	},
 	pikachupopstar: {
 		num: 25,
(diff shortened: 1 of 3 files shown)


sim/dex-data.ts
```diff
@@ -606,7 +606,9 @@ export class Template extends BasicEffect implements Readonly<BasicEffect & Temp
 	 */
 	readonly learnset?: {[moveid: string]: MoveSource[]};
 	/** Source of learnsets for Pokemon that lack their own */
-	readonly inheritsFrom: string | string[];
+	readonly inheritsFrom?: string | string[];
+	/** Whether or not a Pokemon should be hidden in the teambuilder */
+	readonly hide?: 'SpriteOnly' | 'SpriteMinute' | 'Minute';
```
Zarel: Neither the word `hide` nor these possible values are very clear to me.

`SpriteOnly` appears to be a cosmetic form? I don't understand `SpriteMinute` and `Minute` at all.

data/pokedex.js
```diff
@@ -430,6 +432,7 @@ let BattlePokedex = {
 		weightkg: 105,
 		color: "Black",
 		eggGroups: ["Field"],
+		hide: "SpriteOnly",
```
monsanto: I believe @Marty-D said the totems do have competitive differences beyond their sprite, such as weight (which matters for doubles?)

data/pokedex.js
```diff
@@ -9346,6 +9376,7 @@ let BattlePokedex = {
 		evoLevel: 35,
 		eggGroups: ["Field"],
 		inheritsFrom: "darmanitan",
+		hide: "SpriteMinute",
```
monsanto: Not sure what property you're going here (further elaboration from myself and TDP on Discord)

data/pokedex.js
```diff
@@ -10749,6 +10784,7 @@ let BattlePokedex = {
 		color: "Purple",
 		eggGroups: ["Undiscovered"],
 		inheritsFrom: "genesect",
+		hide: "Minute",
```
monsanto: What differences does this have beyond sprite?

data/pokedex.js
```diff
@@ -12451,6 +12497,7 @@ let BattlePokedex = {
 		weightkg: 78.6,
 		color: "Blue",
 		eggGroups: ["Water 2"],
+		hide: "Minute",
```
monsanto: I would argue there is a huge difference here

data/pokedex.js
```diff
@@ -14848,6 +14921,7 @@ let BattlePokedex = {
 		weightkg: 89,
 		color: "Blue",
 		eggGroups: ["Water 1", "Field"],
+		hide: "Minute",
```
monsanto: Looks big to me

sim/dex-data.ts
monsanto: `SpriteOnly` refers to formes that only differ from their base in appearance. The difference with cosmetic formes is that their appearance can convey competitively-relevant information. For example if you see Mimikyu-Busted you know that Disguise is no longer in play.

`Minute` (as in, the adjective meaning small/insignificant, and not the unit of time) refers to formes that do technically have distinctions from their base, but not very many. It makes sense to tier these mons together. Any strategy we write would cover both of them. We'd display them on the same Dex page, noting the few differences. Basculin-Blue-Striped is a good example.

Per Discord discussion I'm not sure what `SpriteMinute` is supposed to be.

monsanto:
I'd call this property something like `differences` rather than `hide`.

data/pokedex.js
KrisXV: i dont think any of the totems are used competitively in smogon formats at least (except for a small period in time where totem araquanid was viable in RU)

sim/dex-data.ts
Zarel: I don't like `Minute`, maybe something more like `Minor`.

I also don't like `hide`, I think I'd prefer `unimportantForme`.

Zarel: I also want the "can convey competitively-relevant information" clarification in the doc comment.

PR #6295, opened 2020-01-19
Support renaming rooms

TheJetOU:
Client PR: smogon/pokemon-showdown-client#1447.

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -80,7 +80,7 @@ class FSPath {
 		return fs.readFileSync(this.path, options) as Buffer;
 	}
 
-	checkIfExists(): Promise<boolean> {
+	exists(): Promise<boolean> {
 		return new Promise(resolve => {
 			fs.exists(this.path, exists => {
 				resolve(exists);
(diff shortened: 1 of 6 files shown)


lib/fs.ts
```diff
@@ -80,6 +80,18 @@ class FSPath {
 		return fs.readFileSync(this.path, options) as Buffer;
 	}
 
+	checkIfExists(): Promise<boolean> {
```
Zarel: No other function prepends `checkIf` to its name.

```suggestion
	exists(): Promise<boolean> {
```

Zarel: Also do we need this? In theory, `isFile`, `isFolder`, etc should cover this use-case.

TheJetOU: > Also do we need this? In theory, isFile, isFolder, etc should cover this use-case.

I guess I missed them when skimming through it, but `FS#exists` maybe be worth keeping in other cases where we cannot infer if the file is a file or directory.

Zarel: This hasn't been changed yet...

TheJetOU:
Bump.

Zarel:
Looks good, I think.

Zarel:
Shit, forgot to merge this for this restart...

PR #6296, opened 2020-01-19
Fix Healing Wish for Gen 8

QuiteQuiet:
It's been years since I did this last so idk if I did this right.

Reference https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-39#post-8341844

- Healing Wish is not consumed if the Pokemon switching in
  has full health and no status condition
- There is no timeout for this effect, it lasts until consumed
- Being statused with full health will consume the Healing Wish
- It does not trigger on damage after switching in
- The effect will be consumed if an injured Pokemon is swapped
  into a slot with an active Healing Wish by Ally Switch

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -8373,8 +8373,8 @@ let BattleMovedex = {
 		accuracy: true,
 		basePower: 0,
 		category: "Status",
-		desc: "The user faints and the Pokemon brought out to replace it has its HP fully restored along with having any major status condition cured. The new Pokemon is sent out at the end of the turn, and the healing happens before hazards take effect. Fails if the user is the last unfainted Pokemon in its party.",
-		shortDesc: "User faints. Replacement is fully healed.",
+		desc: "The user faints and the next injured or statused Pokemon brought in has its HP fully restored along with having any major status condition cured. The healing happens before hazards take effect. Is not consumed if the Pokemon sent out is not injured or statused. Fails if the user is the last unfainted Pokemon in its party.",
+		shortDesc: "User faints. Next hurt Pokemon is fully healed.",
 		id: "healingwish",
 		isViable: true,
 		name: "Healing Wish",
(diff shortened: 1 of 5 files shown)


QuiteQuiet:
bump @Marty-D

data/moves.js
```diff
@@ -8390,16 +8390,31 @@ let BattleMovedex = {
...
 			onSwitchIn(target) {
-				if (!target.fainted) {
+				if (!target.fainted && (target.hp < target.maxhp || target.status)) {
+					target.heal(target.maxhp);
+					target.setStatus('');
+					this.add('-heal', target, target.getHealth, '[from] move: Healing Wish');
+					target.side.removeSlotCondition(target, 'healingwish');
+				}
+			},
+			onSwap(target) {
```
Zarel: We might consider running `Swap` after (or before, as applicable) `SwitchIn`, so this doesn't need to be duplicated.

data/moves.js
```diff
@@ -8390,16 +8390,31 @@ let BattleMovedex = {
...
+				if (!target.fainted && (target.hp < target.maxhp || target.status)) {
 					target.heal(target.maxhp);
 					target.setStatus('');
 					this.add('-heal', target, target.getHealth, '[from] move: Healing Wish');
 					target.side.removeSlotCondition(target, 'healingwish');
 				}
 			},
+			onSourceSwap(target, source) {
```
Zarel: I would rather run Swap on both pokemon, than listen for `SourceSwap` here.

data/moves.js
QuiteQuiet: Running it after won't work, then it would trigger after hazards and other SwitchIn events that it should go before. Before should work, unless I've missed some switch in event that healing wish goes after.

data/moves.js
QuiteQuiet: Sure

QuiteQuiet:
Side thing, but don't know if this has changed: Do you rather want amended commits, or multiple commits you squash when merging?

Zarel:
Either works for me.

Zarel:
Now that squashmerging exists, I don't really care how you implement commit history.

Zarel:
I personally still amend because I think it makes it easier to code review specific commits, but I definitely don't care enough to enforce that on other people.

Zarel:
Looks great, thanks!

PR #6298, opened 2020-01-21
Add EBC to Gens 1 and 2

scheibo:
Also add test for 1000 turn termination behavior.

### DIFF
--- a/test/sim/misc/endlessbattleclause.js
+++ b/test/sim/misc/endlessbattleclause.js
@@ -98,4 +98,23 @@ describe('Endless Battle Clause (slow)', () => {
 		assert(battle.ended);
 		assert(battle.winner === 'Player 2');
 	});
+
+	it('should allow for a maximum of 1000 turns', function () {
+		this.timeout(0);
+		battle = common.createBattle({endlessBattleClause: true});
+		battle.setPlayer('p1', {team: [
+			{species: "Gengar", moves: ['splash']},
+			{species: "Clefable", moves: ['splash']},
+		]});
+		battle.setPlayer('p2', {team: [
+			{species: "Blissey", moves: ['splash']},
+			{species: "Vaporeon", moves: ['splash']},
+		]});
+		for (let i = 0; i < 999; i++) {
+			battle.makeChoices('switch 2', 'switch 2');
+		}
+		assert(!battle.ended);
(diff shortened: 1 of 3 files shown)


test/sim/misc/endlessbattleclause.js
```diff
@@ -98,4 +98,23 @@ describe('Endless Battle Clause (slow)', () => {
 		assert(battle.ended);
 		assert(battle.winner === 'Player 2');
 	});
+
+	it('should allow for a maximum of 1000 turns', function () {
+		this.timeout(0);
+		battle = common.createBattle({endlessBattleClause: true});
```
scheibo: Switching this to `common.gen(2).createBattle` causes the test to fail - I believe this reveals there is a bug in how custom rule parsing is handled (ie. `"gen2customgame@@@!Team Preview,!Cancel Mod,Endless Battle Clause"`), but after an hour I gave up trying to track it down since its out of scope.

Zarel:
Isn't this controversial?

Zarel:
I suppose we can just add this now and let Smogon sort it out?

Zarel:
https://www.smogon.com/forums/threads/endless-battle-clause-in-gsc-and-rby-rse.3659131/

Found the thread.

It's inconclusive, but various conversations lead me to believe most people want it in. Given that it was only out by mistake, this is mergeable.

scheibo:
Yeah, I didn't just push immediately because I wanted a second person to confirm my thoughts - I agree its somewhat controversial in that there's still a debate around it, but the community was under the impression that the clause *should* be in effect and it was left out as an oversight. I'll be more than happy to make any changes to the EBC that come out of the discussion, but give we have the TDs creating posts like https://www.smogon.com/forums/threads/spl-xi-administrative-decisions.3656323/post-8344552 it seems important to implement in the meantime.

---

Did you have any thoughts on the [comment](https://github.com/smogon/pokemon-showdown/pull/6298#discussion_r369289011) on the code? I think theres a bug with Custom Rule processing

test/sim/misc/endlessbattleclause.js
Zarel: Weird. I feel like I've worked on this code recently enough that I have a pretty good handle on how it works, and I'm very surprised this kind of bug is still here. I'll have to investigate when I have time.

PR #6301, opened 2020-01-22
Refactor trustuser checks

whalemer:
Most notably, allow the commands to be used without needing the exact user to be online.


Zarel:
(For future reference, "refactor" means no major functionality changes; you should generally name a commit after the functionality change if there is one.)

Zarel:
I wouldn't call this a refactor at all, there doesn't seem to be any change in here that _isn't_ directly adding functionality.

PR #6302, opened 2020-01-22
Sockets: shutdown any open connections when workers exit gracefully

Kaiepi:
When exiting a process, the kernel will call close(3) on any open
sockets it created, but there's more to closing a TCP connection than
just closing its socket; unless the TCP FIN handshake is completed by
calling shutdown(3) beforehand, connections can wind up being forced to
wait for a timeout before they can be closed, potentially preventing
users from being notified that they need to reconnect immediately after
the server exits depending on what platform the server is being run on.


Kaiepi:
Closing this since I can't find any cases where the changes made in this PR actually help anything. The way we close connections now probably violates WebSocket protocol though, but that's easier to fix in a separate PR.

PR #6303, opened 2020-01-22
Fix Pickpocket Eject Button Interaction

MrEconomical:
Fix of mechanic bug: https://github.com/smogon/pokemon-showdown/projects/3#card-25870242


urkerab:
This probably belongs in an `onTakeItem` event handler on the Eject Button itself. The fourth parameter to `onTakeItem` might actually be an Effect rather than the Move that `global-types.js` claims, which thus allows you to check that it's Pickpocket that's trying to take the item.

PR #6309, opened 2020-01-24
Update Emergency Exit test

TheImmortal:


TheImmortal:
@Zarel this should pass.

Zarel:
I haven't forgotten...

Zarel:
#6351 will also massively change Emergency Exit, so I guess this means I'll do both at once.

Zarel:
Also, should both tests exist? Why did you replace it instead of adding a separate test?

TheImmortal:
No particular reason. I can re-add it.

TheImmortal:
But also, presumably you’d just copy this test locally and push it yourself when you fix it. Either works.

Zarel:
Yeah, that's what I'm going to do.

PR #6313, opened 2020-01-25
Fix Pickpocket Eject Button Interaction

MrEconomical:
Fix of mechanic bug: https://github.com/smogon/pokemon-showdown/projects/3#card-25870242


Marty-D:
I think it makes more sense to do this in Pickpocket itself, so you can also defend against other forced switches from Dragon Tail and such (a situation where Pickpocket shouldn't steal the item either, and is probably also bugged on PS). Just check for `forceSwitchFlag` and `switchFlag` there, like Emergency Exit already does.

PR #6314, opened 2020-01-26
Fix pickpocket eject button interaction

MrEconomical:
Fix of mechanic bug https://github.com/smogon/pokemon-showdown/projects/3#card-25870242

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -2697,7 +2697,7 @@ let BattleAbilities = {
 					return;
 				}
 				let yourItem = source.takeItem(target);
-				if (!yourItem) {
+				if (!yourItem || yourItem.id === 'ejectbutton') {
 					return;
 				}
 				if (!target.setItem(yourItem)) {


HoeenCoder:
Hey, please dont close and re-open your pull request if you need to make an edit. Instead apply the changes to your branch. Theres no need to rebase either since we squash merge changes.

MrEconomical:
-_-

Zarel:
This is certainly one way to do it, but I wonder if it would be possible to change the order they resolve.

Zarel:
Also:

- there's no check for a contact move
- it still removes Eject Button, probably preventing it from activating?

How does Thief handle this interaction?

Marty-D:
The thing I suggested to you in your last PR is already done 4 lines above this, so this specific bug should already be fixed.

Marty-D:
So, yes, this was fixed in https://github.com/smogon/pokemon-showdown/pull/6254.

PR #6315, opened 2020-01-26
Allow safer and exact type retrieval of Room#game

TheJetOU:
By adding a `getGame` function of type:

```typescript
// null is returned if the gameids don't match
// or the game doesn't exist
getGame<T extends RoomGame>(constructor: new (...args: any[]) => T) => T | null
```
(Credits @urkerab and @whalemer for the function signature.)

It allows refactoring previous code of:

```typescript
if (room.game && room.game.gameid !== 'hangman') return;
const game = room.game as Hangman;
```

to:

```typescript
const game = room.getGame<Hangman>(Hangman);
if (!game) return;
```

This has a couple of advantages:
- TypeScript will throw an error if the if condition is not present.

- In the new code, the template must extends `RoomGame` and be assignable to the same ID, so it's 100% typesafe

### DIFF
--- a/server/chat-plugins/hangman.ts
+++ b/server/chat-plugins/hangman.ts
@@ -7,7 +7,7 @@
 
 const maxMistakes = 6;
 
-class Hangman extends Rooms.RoomGame {
+export class Hangman extends Rooms.RoomGame {
 	gameNumber: number;
 	creator: ID;
 	word: string;
--- a/server/chat-plugins/hangman.ts
+++ b/server/chat-plugins/hangman.ts
@@ -240,7 +240,7 @@ export const commands: ChatCommands = {
 
 		guess(target, room, user) {
 			if (!target) return this.parse('/help guess');
-			const game = room.getGame<Hangman>(Hangman);
+			const game = room.getGame(Hangman);
 			if (!game) return this.errorReply("There is no game of hangman running in this room.");
 			if (!this.canTalk()) return;
 
@@ -255,7 +255,7 @@ export const commands: ChatCommands = {
 		end(target, room, user) {
 			if (!this.can('minigame', null, room)) return false;
 			if (!this.canTalk()) return;
-			const game = room.getGame<Hangman>(Hangman);
+			const game = room.getGame(Hangman);
 			if (!game) return this.errorReply("There is no game of hangman running in this room.");
 			game.end();
 			this.modlog('ENDHANGMAN');
@@ -290,7 +290,7 @@ export const commands: ChatCommands = {
 		},
 
 		display(target, room, user) {
-			const game = room.getGame<Hangman>(Hangman);
+			const game = room.getGame(Hangman);
 			if (!game) return this.errorReply("There is no game of hangman running in this room.");
 			if (!this.runBroadcast()) return;
 			room.update();
@@ -315,7 +315,7 @@ export const commands: ChatCommands = {
 	],
 
 	guess(target, room, user) {
-		const game = room.getGame<Hangman>(Hangman);
+		const game = room.getGame(Hangman);
 		if (!game) return this.errorReply("There is no game of hangman running in this room.");
 		if (!this.canTalk()) return;
 
--- a/server/chat-plugins/hangman.ts
+++ b/server/chat-plugins/hangman.ts
@@ -240,10 +240,10 @@ export const commands: ChatCommands = {
 
 		guess(target, room, user) {
 			if (!target) return this.parse('/help guess');
-			if (!room.game || room.game.gameid !== 'hangman') return this.errorReply("There is no game of hangman running in this room.");
+			const game = room.getGame<Hangman>(Hangman);
+			if (!game) return this.errorReply("There is no game of hangman running in this room.");
 			if (!this.canTalk()) return;
 
-			const game = room.game as Hangman;
 			game.guess(target, user);
 		},
 		guesshelp: [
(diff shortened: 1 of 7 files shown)


Zarel:
I still want to get this working with `getGame(Hangman)` rather than `getGame<Hangman>('hangman')`.

Zarel:
But yes, as discussed, I do like this more than the current code.

TheJetOU:
> I still want to get this working with `getGame(Hangman)` rather than `getGame<Hangman>('hangman')`.

`getGame(Hangman)` will be able to have the same run-time affect (only return if it's a hangman game), but it won't be typesafe because you will need generics for that, unless you hardcode every game, which I think generics is more preferable than.

Zarel:
Like I said on Discord, you just need `class Hangman { static readonly gameid = 'hangman'; ...`

Zarel:
I'm going to see if I can make the generics work.

Zarel:
After some struggling, apparently the answer is no, I can't.

Zarel:
I'm still sort of partial to `getGame<Hangman>(Hangman)` though.

TheJetOU:
> I'm still sort of partial to `getGame<Hangman>(Hangman)` though.

I'm guessing you definitely won't like `getGame<Hangman, 'hangman'>(Hangman)` for 100% type safety-ness then :p.

Zarel:
I don't see how that's safer.

TheJetOU:
> I don't see how that's safer.

If we make `RoomGame` a generic class with `T` being `gameid`, we can make `getGame` have the type `getGame<T extends RoomGame<U>, U extends string>(gameid: U) => null`. This will make TypeScript complain about the use of `getGame<Hangman, 'mafia'>(mafia)`, because `Hangman` implements `RoomGame<'hangman'>` and not `RoomGame<'mafia'>`.

Zarel:
As far as I can tell, that's no safer than `static readonly gameid`.

Zarel:
What about `getGame<T extends RoomGame<U>>(gameid: U) => null`?

TheJetOU:
> What about `getGame<T extends RoomGame<U>>(gameid: U) => null`?

TypeScript doesn't like that, unfortunately.

```typescript
server/rooms.ts:308:29 - error TS2304: Cannot find name 'U'.

308  getGame<T extends RoomGame<U>>(gameid: U) {
                                ~

server/rooms.ts:308:41 - error TS2304: Cannot find name 'U'.

308  getGame<T extends RoomGame<U>>(gameid: U) {
```

:(

server/chat-plugins/hangman.ts
```diff
@@ -240,10 +240,10 @@ export const commands: ChatCommands = {
 
 		guess(target, room, user) {
 			if (!target) return this.parse('/help guess');
-			if (!room.game || room.game.gameid !== 'hangman') return this.errorReply("There is no game of hangman running in this room.");
+			const game = room.getGame<Hangman>(Hangman);
```
whalemer: Can typescript not infer the template type?

server/chat-plugins/mafia.ts
```diff
@@ -2800,7 +2802,7 @@ export const commands: ChatCommands = {
 			}
 
 			// hack to let hosts broadcast
-			const game = room && room.game && room.game.gameid === 'mafia' ? room.game as MafiaTracker : null;
+			const game = room && room.game && room.game.gameid === 'mafia' ? room.getGame<MafiaTracker>(MafiaTracker) : null;
```
whalemer: This can just be `room && room.getGame(MafiaTracker)`

server/chat-plugins/uno.ts
```diff
@@ -810,7 +810,7 @@ export const commands: ChatCommands = {
 		},
 
 		color(target, room, user) {
-			const game = room.game as UnoGame;
+			const game = room.getGame<UnoGame>(UnoGame);
 			if (!game || game.gameid !== 'uno') return false;
```
whalemer: don't need the `gameid` checks anymore, same with a few more places in this file.

Zarel:
Just need to fix that merge conflict...

TheJetOU:
I wish GitHub would send users notifications/emails when a merge conflict occurs.

Anyway, it should be fixed now.

server/chat-plugins/hangman.ts
```diff
@@ -255,9 +255,8 @@ export const commands: ChatCommands = {
 		end(target, room, user) {
 			if (!this.can('minigame', null, room)) return false;
 			if (!this.canTalk()) return;
-			if (!room.game || room.game.gameid !== 'hangman') return this.errorReply("There is no game of hangman running in this room.");
-
-			const game = room.game as Hangman;
+			const game = room.getGame(Hangman);
+			if (!game) return this.errorReply("There is no game of hangman running in this room.");
```
Zarel:
```suggestion
			if (!game) return this.errorReply("There is no game of hangman running in this room.");

```

server/chat-plugins/hangman.ts
```diff
@@ -291,11 +290,11 @@ export const commands: ChatCommands = {
 		},
 
 		display(target, room, user) {
-			if (!room.game || room.game.title !== 'Hangman') return this.errorReply("There is no game of hangman running in this room.");
+			const game = room.getGame(Hangman);
+			if (!game) return this.errorReply("There is no game of hangman running in this room.");
 			if (!this.runBroadcast()) return;
 			room.update();
```
Zarel: ??? I don't think this is necessary at all...

Zarel: I guess this is an unrelated PR for fixing this sort of thing, though.

server/rooms.ts
```diff
@@ -305,6 +305,10 @@ export abstract class BasicRoom {
 		}
 		if (this.parent) return this.parent.getMuteTime(user);
 	}
+	getGame<T extends RoomGame>(constructor: new (...args: any[]) => T): T | null {
+		if (this.game && this.game.constructor.prototype.gameid === constructor.prototype.gameid) return this.game as T;
```
Zarel: Um, `constructor.prototype.gameid` should be `undefined` right now, I don't think you've actually defined it yet?

Zarel:
```
>> room.game.constructor.prototype.gameid
<< undefined
```

Your changes don't seem to add a prototype anywhere.

Anyway, like I said, I think `static readonly gameid` is the best approach.

BUT if you wanted to be lazy, `this.game.constructor.name === constructor.name` _is_ good enough.

TheJetOU: I'll go with the 'lazy' route since I don't want to deal with breaking the untyped files.

PR #6317, opened 2020-01-28
Add TRs

KrisXV:
Thanks to @YaBoiJD for helping with this

https://www.smogon.com/forums/threads/adding-trs-as-held-items.3659367/


Zarel:
I'm kind of apprehensive about having a bunch of TRs in the teambuilder, but I guess if we section them off, it's not a problem.

PR #6319, opened 2020-01-28
(gen 7) Properly resolve speed ties on the turn of Mega Evolution

jamesqo:
Original bug reported here: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/page-107#post-8345248

Currently, in gen 7, any Pokemon that speed ties with the opponent but has to mega evolve on that turn always goes last. This PR fixes that by adding logic to check/account for speed ties.

Example replay demonstrating the fix working https://pastebin.com/7P1B1wGs

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2531,7 +2531,8 @@ export class Battle {
 			chosenAction.pokemon.updateSpeed();
 		}
 		const action = this.resolveAction(chosenAction, midTurn);
-		let tieStart = -1, tieEnd = this.queue.length;
+		let tieStart = -1;
+		let tieEnd = this.queue.length;
 		for (const [i, curAction] of this.queue.entries()) {
 			const delta = this.comparePriority(action, curAction);
 			if (delta === 0 && tieStart === -1) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
The deterministic Speed tie bug is much more widespread than just Mega Evolution; I can't tell how much of the PR is actually addressing Megas specifically, but deterministic Speed ties have been replicated with:
- Perish Song: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8201686
- Primal Reversion: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8201694
- Entrance Abilities: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8203577
- Confirmation of Mega Evolution: https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8201736

There's clearly a pattern here with Speed ties being bugged in _some_ way.

Zarel:
I don't think screwing with the sorting algorithm is the right call here. `subOrder` is probably the right one?

AnnikaCodes:
Has this bug since been fixed?

DaWoblefet:
> Has this bug since been fixed?

No, still broken in the places I referenced.

PR #6321, opened 2020-01-30
Fix Clangorous Soul

ZardMX:
It should not fail if one stat is +6 stages, rather when all stats are +6 stages.
Proof: https://www.youtube.com/watch?v=-LLt-bnoQss (Thanks KennedyLFC)

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -2490,7 +2490,7 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {snatch: 1, sound: 1, dance: 1},
 		onTryHit(pokemon, target, move) {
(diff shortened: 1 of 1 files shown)


urkerab:
You should probably add parentheses for readability.

ZardMX:
Actually it doesn't fail at when all stats +6 stages either, rather says "Stat won't go any higher!" but doesn't reduce the HP.

urkerab:
So, this depends on what happens when Contrary is combined, but one approach is to remove the `boosts` from the move data and add it in the hit function, something like this:

    if (this.boost{atk: 1, def: 1, spa: 1, spd: 1, spe: 1}) {
       this.directDamage(pokemon.maxhp / 3);
    }

ZardMX:
@urkerab then Clangorous Soul won't show up while using /ms boosts

DaWoblefet:
Clangorous Soul will lower stats and reduce HP if it is at +6 all with Contrary. Clangorous Soul will say "< Pokemon >'s < stat > won't go any lower!" and not reduce HP if it is at -6 all with Contrary.

PR #6324, opened 2020-01-31
Fix Lottery Help

tmagicturtle:
There is no lottery editmarkup command. The proper use is to use /lottery edit, which allows you to edit the number of max winners, the name of the lottery, and the HTML, but also requires all three parameters be included.

### DIFF
--- a/server/chat-plugins/lottery.ts
+++ b/server/chat-plugins/lottery.ts
@@ -248,7 +248,7 @@ export const commands: ChatCommands = {
 		`/lottery create max winners, name, html - creates a new lottery with [name] as the header and [html] as body. Max winners is the amount of people that will win the lottery. Requires # & ~`,
 		`/lottery delete - deletes the current lottery without declaring a winner. Requires # & ~`,
 		`/lottery end - ends the current lottery, declaring a random participant as the winner. Requires # & ~`,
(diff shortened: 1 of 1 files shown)


server/chat-plugins/lottery.ts
```diff
@@ -248,7 +248,7 @@ export const commands: ChatCommands = {
 		`/lottery create max winners, name, html - creates a new lottery with [name] as the header and [html] as body. Max winners is the amount of people that will win the lottery. Requires # & ~`,
 		`/lottery delete - deletes the current lottery without declaring a winner. Requires # & ~`,
 		`/lottery end - ends the current lottery, declaring a random participant as the winner. Requires # & ~`,
-		`/lottery editmarkup html - edits the lottery markup with the provided HTML. Requires # & ~`,
+		`/lottery edit max winners, name, html - edits the lottery markup with the provided HTML. Requires # & ~`,
```
TheJetOU:
```suggestion
		`/lottery edit max winners, name, html - edits the lottery with the provided parameters. Requires # & ~`,
```

Zarel: bump ^

Zarel: I still can't merge this suggestion for some reason.

PR #6329, opened 2020-02-02
Add modlog button to appeal tickets

asgdf:
When taking appeals, you're probably wanting to check their modlog roughly 100% of the time, so this is useful for people who like having something to click instead of having to type the command.
(It does hide the button under "other options" but I guess having to click another button first isn't that bad)


Zarel:
Should we also have room modlog buttons for all rooms they're punished in ?

asgdf:
Hm. Generally, appeal tickets are just for locks, not room punishments, so that'd only be relevant for room punishment autolocks, most of which already involve roombans, which are in the global modlog.
So I guess it would be useful in niche cases but I don't know if it'd be worth the clutter.

Zarel:
Oh, yeah, good point.

PR #6335, opened 2020-02-04
Tournaments: Send name and format to room separately

KrisXV:


Zarel:
Does this require a client change?

KrisXV:
No

PR #6337, opened 2020-02-04
Rooms: Fix modjoin message for staff rooms

AustinXII:


AustinXII:
Hm maybe it should just say “must be room staff or higher” in all circumstances considering room staff overrides modjoin In every room

server/rooms.ts
```diff
@@ -1238,7 +1238,11 @@ export class BasicChatRoom extends BasicRoom {
 		}
 		if (this.modjoin) {
 			const modjoin = this.modjoin === true ? this.modchat : this.modjoin;
-			message += ` [${modjoin} or higher to join]`;
+			if (this.staffRoom) {
+				message += ` [room staff or higher to join]`;
```
Zarel: ????

This seems clearly wrong to me?

Zarel: Really, any room accepts either the modjoin rank in what it's inheriting from, or any local rank.

Zarel:
This was after some conversation in Discord.

My main points were:

- Yes, this is slightly inaccurate; a more accurate message might be something like `[${modjoin} or higher, or any roomauth, to join]`

- But the message is intended to be a short notice that modjoin is in effect, it's not supposed to be a full description of how modjoin works

- I think it makes intuitive sense for room ranks to bypass the modjoin setting: if you give someone a room rank, obviously they're supposed to be in the room

PR #6338, opened 2020-02-04
Improve handling of mobile IPs with permalocks

asgdf:
I believe this is something @Asheviere had wanted to do in the past, but I think it's necessary to do at least a step in the right direction as early as possible.
Currently permalocks can cause 30 day locks on mobile IPs. This is bad and causes a lot of IP-Appeal tickets that could have been avoided. As such, permalocks will only lock for 6 hours on mobile IPs now.
I mostly merged user types 5 and 6 here except for the fact that the latter disconnects. The only practical difference should be that it always says "permalocked" now but I don't mind because it's accurate anyway. A permaban is a permalock with a forced disconnect, after all.

This is difficult to test locally without a bunch of setup, so I wrote a unit test instead. I hope I didn't forget anything.


Zarel:
I thought I already did other changes to mobile stuff.

Why does this only apply to permalocks? It should probably apply to all locks and bans.

server/punishments.ts
```diff
@@ -1060,7 +1060,7 @@ export const Punishments = new class {
 			if (!punishment) punishment = ['LOCK', user.locked, 0, ''];
 		}
 
-		const ticket = Chat.pages.help
+		const ticket = Chat.pages && Chat.pages.help
```
Zarel:
```suggestion
		const ticket = Chat.pages?.help
```

server/users.ts
```diff
@@ -902,12 +903,14 @@ export class User extends Chat.MessageContext {
...
 				this.autoconfirmed = userid;
-			} else if (userType === '5') {
+			} else if (userType === '5' || userType === '6') {
 				this.permalocked = userid;
-				Punishments.lock(this, Date.now() + PERMALOCK_CACHE_TIME, userid, true, `Permalocked as ${name}`);
-			} else if (userType === '6') {
-				Punishments.lock(this, Date.now() + PERMALOCK_CACHE_TIME, userid, true, `Permabanned as ${name}`);
-				this.disconnectAll();
+				if (this.latestHostType === 'mobile') {
+					Punishments.lock(this, Date.now() + PERMALOCK_MOBILE_CACHE_TIME, userid, true, `Permalocked as ${name}`);
```
Zarel: This should very likely be handled inside `Punishments.lock` rather than outside.

Zarel: Especiallly since the userid and other IP locks should be for the full time; only the mobile IP lock should be short.

And all mobile IP locks should be short, not just permalocks.

asgdf:
I did it just for permalocks because that's where 99% of the problems come from.
Doing separate handling for every IP requires an async IPtools lookup for each of them, I didn't like the complications that would bring, but I guess it'd have to?
I'm not 100% sure what would happen if there are punishments for the same userid with different durations. I don't think the punishment system is made to handle that?

Zarel:
IPtools does caching, and every IP is audited upon connection, so the async lookup isn't actually very slow in practice. It's not like locks are _that_ common.

Hm, you're right. They might have to be implemented as separate punishments, which is probably possible...

server/punishments.ts
Zarel: This can _definitely_ be done now.

Zarel:
This is waiting on moving the mobile IP check inside `Punishments.lock`.

You don't need to separate by IP if you _really_ don't want to, but it still needs to happen for autolocks and weeklocks as well as permalocks.

asgdf:
Yeah sorry got caught up with exams, I'll get back to it towards the end of the month

asgdf:
Okay this is attempt #2. I tried to await in all situations where it would be needed, and not in situations where it wasn't.
I decided to not persist the locks on mobile IPs and instead to just keep them in memory. It shouldn't really matter for those in particular (the non-mobile IPs/username are still persisted ofc).

Re: the tslint complaints, what am I supposed to do with these? I didn't await in these cases because the functions don't really need to wait for the async stuff to finish, they just trigger it. Should I put a dummy .then that catches errors?

Zarel:
Wrap the call in `void`, like `void callAsyncFunction()`

asgdf:
Okay. Did I forget anything?

server/punishments.ts
```diff
@@ -385,10 +387,17 @@ export const Punishments = new class {
 			reason,
 			rest,
 		}, id, PUNISHMENT_FILE);
+
+		const mobileExpireTime = Date.now() + MOBILE_PUNISHMENT_DURATIION;
+		const mobilePunishment = [punishType, id, mobileExpireTime, reason, ...rest] as Punishment;
+		for (const mobileIp of mobileIps) {
+			Punishments.ips.set(mobileIp, mobilePunishment);
+		}
```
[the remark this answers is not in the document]
TheJetOU: `Set#size` is a variable and not a function.

Zarel:
Can you fix the merge conflict?

server/punishments.ts
```diff
@@ -1063,7 +1079,7 @@ export const Punishments = new class {
 			if (!punishment) punishment = ['LOCK', user.locked, 0, ''];
 		}
 
-		const ticket = Chat.pages.help
+		const ticket = Chat.pages?.help
```
Zarel: ?

asgdf: https://github.com/smogon/pokemon-showdown/pull/6338#discussion_r375246859

Zarel: Why can `Chat.pages` be null now?

Zarel: Oh, apparently it can. That sucks. :/

server/punishments.ts
```diff
@@ -1063,9 +1079,9 @@ export const Punishments = new class {
 			if (!punishment) punishment = ['LOCK', user.locked, 0, ''];
 		}
 
-		const ticket = Chat.pages.help ?
-			`<a href="view-help-request--appeal"><button class="button"><strong>Appeal your punishment</strong></button></a>` :
-			'';
+		const ticket = Chat.pages?.help ?
+			`<a href="view-help-request--appeal"><button class="button"><strong>Appeal your punishment</strong></button></a>` : '';
+
```
Zarel:
```suggestion
```

server/punishments.ts
```diff
@@ -1110,7 +1126,7 @@ export const Punishments = new class {
 		if (registered && id === 'BAN') {
 			user.popup(`Your username (${user.name}) is banned${bannedUnder}. Your ban will expire in a few days.${reason}${Config.appealurl ? `||||Or you can appeal at: ${Config.appealurl}` : ``}`);
```
Zarel:
```suggestion
			user.popup(
				`Your username (${user.name}) is banned${bannedUnder}. Your ban will expire in a few days.${reason}` +
				`${Config.appealurl ? `||||Or you can appeal at: ${Config.appealurl}` : ``}`
			);
```

server/chat-plugins/chat-monitor.ts
```diff
@@ -108,7 +108,7 @@ Chat.registerMonitor('autolock', {
 			message = message.replace(/(https?):\/\//g, '$1__:__//');
 			message = message.replace(/\./g, '__.__');
 			if (room) {
-				Punishments.autolock(
+				void Punishments.autolock(
 					user, room,
 					'ChatMonitor', `Filtered phrase: ${word}`, `<${room.roomid}> ${user.name}: ${message}${reason ? ` __(${reason})__` : ''}`, true
```
Zarel:
```suggestion
					'ChatMonitor', `Filtered phrase: ${word}`,
					`<${room.roomid}> ${user.name}: ${message}${reason ? ` __(${reason})__` : ''}`, true
```

server/chat-plugins/chat-monitor.ts
```diff
@@ -178,7 +178,7 @@ Chat.registerMonitor('evasion', {
 			if (isStaff) return `${message} __[would be locked for filter evading: ${match[0]} (${word})]__`;
 			message = message.replace(/(https?):\/\//g, '$1__:__//');
 			if (room) {
-				Punishments.autolock(
+				void Punishments.autolock(
 					user, room,
 					'FilterEvasionMonitor', `Evading filter: ${message} (${match[0]} => ${word})`, `<${room.roomid}> ${user.name}: SPOILER: \`\`${message}\`\` __(${match[0]} => ${word})__`
```
Zarel:
```suggestion
					'FilterEvasionMonitor', `Evading filter: ${message} (${match[0]} => ${word})`,
					`<${room.roomid}> ${user.name}: SPOILER: \`\`${message}\`\` __(${match[0]} => ${word})__`
```

server/chat-plugins/chat-monitor.ts
Zarel:
```suggestion
					user, room, 'ChatMonitor', `Filtered phrase: ${word}`,
					`<${room.roomid}> ${user.name}: ${message}${reason ? ` __(${reason})__` : ''}`, true
```

server/chat-plugins/chat-monitor.ts
Zarel:
```suggestion
					user, room, 'FilterEvasionMonitor', `Evading filter: ${message} (${match[0]} => ${word})`,
					`<${room.roomid}> ${user.name}: SPOILER: \`\`${message}\`\` __(${match[0]} => ${word})__`
```

server/chat-plugins/chat-monitor.ts
```diff
@@ -365,7 +365,7 @@ export const namefilter: NameFilter = (name, user) => {
 
 			if (regex.test(lcName)) {
 				if (Chat.monitors[list].punishment === 'AUTOLOCK') {
-					Punishments.autolock(
+					void Punishments.autolock(
 						user, 'staff', `NameMonitor`,
 						`inappropriate name: ${name}`, `using an inappropriate name: ${name} (from ${user.name})`, false, name
```
Zarel:
```suggestion
						user, 'staff', `NameMonitor`, `inappropriate name: ${name}`,
						`using an inappropriate name: ${name} (from ${user.name})`, false, name
```

server/chat-plugins/chat-monitor.ts
```diff
@@ -418,7 +418,7 @@ export const nicknamefilter: NameFilter = (name, user) => {
 
 			if (regex.test(lcName)) {
 				if (Chat.monitors[list].punishment === 'AUTOLOCK') {
-					Punishments.autolock(
+					void Punishments.autolock(
 						user, 'staff',
 						`NameMonitor`, `inappropriate Pokémon nickname: ${name}`, `${user.name} - using an inappropriate Pokémon nickname: ${name}`, true
```
Zarel:
```suggestion
						user, 'staff', `NameMonitor`, `inappropriate Pokémon nickname: ${name}`,
						`${user.name} - using an inappropriate Pokémon nickname: ${name}`, true
```

server/chat-plugins/chat-monitor.ts
```diff
@@ -457,7 +457,7 @@ export const statusfilter: StatusFilter = (status, user) => {
 
 			if (regex.test(lcStatus)) {
 				if (Chat.monitors[list].punishment === 'AUTOLOCK') {
-					Punishments.autolock(
+					void Punishments.autolock(
 						user, 'staff', `NameMonitor`,
 						`inappropriate status message: ${status}`, `${user.name} - using an inappropriate status: ${status}`, true
```
Zarel:
```suggestion
						user, 'staff', `NameMonitor`, `inappropriate status message: ${status}`,
						`${user.name} - using an inappropriate status: ${status}`, true
```

Zarel:
I hope you don't mind me sneaking some formatting fixes into your PR. This is otherwise ready to merge.

asgdf:
Did the formatting stuff as well

PR #6343, opened 2020-02-06
Toxtricity-Low-Key can Gigantamax

KrisXV:


data/moves.js
```diff
@@ -7316,7 +7316,7 @@ let BattleMovedex = {
 		pp: 10,
 		priority: 0,
 		flags: {},
-		isMax: "Toxtricity",
+		isMax: ["Toxtricity", "Toxtricity-Low-Key"],
```
Zarel: I don't see support for `isMax` being an array.

I also don't think this needs to be an array. Are there situations where only one forme is Gmaxable?

KrisXV: I already added support for `isMax` being an array in this PR. And I'm not sure what you mean, so I'm going to say no.

Zarel: Um, you added type information for `isMax` being an array in this PR. I see support for it in `/dt` and `/ds`, but I see no support in the sim itself.

Zarel: https://github.com/smogon/pokemon-showdown/blob/b30758a3e174c8be5348df687e24dc945c1f75f0/data/scripts.js#L1206

Oh, I see, no support code is needed in the sim itself.

Zarel: I don't see why this can't just be `isMax: "Toxtricity",`

Zarel:
```suggestion
		isMax: "Toxtricity",
```

KrisXV: Hmm. I guess.

KrisXV: Should I remove the array support or leave it in in the case that other GMax moves are shared?

PR #6344, opened 2020-02-06
SSB: Prevent salamander's ability from crashing in edge cases

Salaaa:


data/mods/ssb/abilities.js
```diff
@@ -998,7 +998,7 @@ let BattleAbilities = {
 		desc: "This Pokemon is immune to volatile statuses.",
 		shortDesc: "This Pokemon is immune to volatile statuses.",
 		onTryAddVolatile(status, target) {
-			if (toID(target.name).includes(status.id)) return;
+			if ([toID(target.name), 'furycutter', 'stockpile'].includes(status.id)) return;
```
Zarel: Why isn't `toID(target.name)` hardcoded? It's just `'salamander'`, right?

Salaaa: Fair enough

HoeenCoder: It was not hardcoded to handle any case where this ability could be applied to another pokemon. I don't think anything could cause that now but it prevents bugs in the case something like skill swap is introduced in the future.

Salaaa: the entire purpose of this pull request was to handle a bug from a skill swap, but like, if you want I can just revert the previous commit

HoeenCoder: Oh right we have metronome-like move(s). Yeah this would be needed.

PR #6349, opened 2020-02-07
Add Pokemon Home support

KrisXV:
This isn't 100% accurate yet but will make implementing Pokemon Home a lot easier when it does come out.


KrisXV:
It's out and this should be mostly accurate barring some testing

KrisXV:
@Zarel @TheImmortal all the stuff in here is confirmed 100% accurate

TheImmortal:
OU/DOU haven't really announced what's banned but I guess this'll do for now.

KrisXV:
I spoke with Finch with what to tier as what

PR #6350, opened 2020-02-07
Glitch

Ampha972:


TheImmortal:
This is not the repository for the Glitch server.

Zarel:
We're blocking you so you don't accidentally PR us again. This isn't a punishment, just to protect against mistakes - if you ever need to comment or PR to our repository, just email us at `staff@pokemonshowdown.com` and we'll unblock you.

PR #6351, opened 2020-02-07
Refactor `AfterDamage` to `DamagingHit`

### LINKED ISSUE #6346: Move secondaries should run before the AfterDamage event
Currently: the event triggers before secondaries (https://replay.pokemonshowdown.com/gen8customgame-1058947435)

What should happen: move secondaries should occur first (https://twitter.com/Sibuna_Switch/status/1225472059393769472)

It's not just this one ability but all abilities that currently use ``onAfterDamage``.

Zarel:
Fixes #6346

#6346 reports that all existing `AfterDamage` handlers should happen after secondaries.

This does that:

- The `AfterDamage` event has been replaced with `DamagingHit`, which which happens for damaging moves after secondaries.

- The `AfterHit` event has also been moved after `DamagingHit`, to make sure Knock Off still procs after Rocky Helmet.

- `AfterHit` is no longer a valid event on `secondary` and `self` blocks, because it's meaningless in those blocks, anyway. All `self.onAfterHit` and `secondary.onAfterHit` handlers have been moved to `onHit`, which should have the same timing in practice.

### DIFF
--- a/data/mods/gen4/items.js
+++ b/data/mods/gen4/items.js
@@ -125,17 +125,6 @@ let BattleItems = {
 		onEffectiveness() {},
 		desc: "Holder's Speed is halved and it becomes grounded.",
 	},
-	"jabocaberry": {
-		inherit: true,
-		onAfterDamage() {},
-		onAfterMoveSecondary(target, source, move) {
-			if (source && source !== target && move && move.category === 'Physical') {
-				if (target.eatItem()) {
-					this.damage(source.baseMaxhp / 8, source, target, null, true);
-				}
-			}
-		},
-	},
 	"kingsrock": {
 		inherit: true,
 		onModifyMove(move) {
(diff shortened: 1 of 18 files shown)


Zarel:
Draft because Emergency Exit and Wimp Out use `AfterDamage`, but I haven't been told how they're supposed to work yet.

sim/battle.ts
```diff
@@ -1750,8 +1753,6 @@ export class Battle {
 			}
 		}
 
-		// @ts-ignore - FIXME AfterDamage passes an Effect, not an ActiveMove
```
Zarel: (yay we fixed a FIXME)

data/mods/gen4/items.js
```diff
@@ -125,17 +125,6 @@ let BattleItems = {
 		onEffectiveness() {},
 		desc: "Holder's Speed is halved and it becomes grounded.",
 	},
-	"jabocaberry": {
```
Zarel: @Marty-D you fixed these in edc2b8f98a08477c858

But according to #6346, all damaging-hit handlers are timed like these, so this is no longer necessary, right?

Zarel:
WAIT

Can we merge `AfterMoveSecondary` and `DamagingHit` now?

Zarel:
After more closely inspecting `AfterMoveSecondary`, it seems clearly different from `DamagingHit` - it happens after all hits (and is suppressed by a LO with a secondary), while `DamagingHit` happens on each hit.

I'm still concerned about the _number_ of `AfterMoveSecondary` effects, though. Have all these been tested to be suppressed by Life Orb?

PR #6353, opened 2020-02-08
Update LGPE OU Tiers

CasualPokePlayer:
https://www.smogon.com/stats/2019-11/gen7letsgoou-1630.txt

LGPE OU hasn't had its tiers updated on the server a long while, so here's the changes from the latest usage stats.


Zarel:
Wait, you're not Kris and that's just a stats link...

PR #6356, opened 2020-02-10
Support polls allowing multiple responses

whalemer:
image

This makes polls store their options and require a separate submission step, and also adds a new type of poll for selecting any number of options.

A lot of the caching for the _italicized selected_ answer on the results screen has been removed, but in my testing the function's fast enough to have practically 0 impact.

### DIFF
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -124,7 +124,7 @@ export class Poll {
 				const selected = pendingVotes.includes(num);
 				output += `<div style="margin-top: 5px"><button style="text-align: left; border: none; background: none; color: inherit;${selected ? 'font-weight:bold' : ''}" value="/poll ${selected ? 'de' : ''}select ${num}" name="send" title="${selected ? "Deselect" : "Select"} ${num}. ${Chat.escapeHTML(option.name)}">${selected ? chosen : empty} ${num}. ${this.getOptionMarkup(option)}</button></div>`;
 			}
-			output += `<div style="margin-top: 7px; padding-left: 12px"><button class="button" value="/poll submit" name="send" title="${pendingVotes.length ? "Submit your vote" : "View results - you will not be able to vote after viewing results"}">${pendingVotes.length ? "Submit" : "View results"}</button></div>`;
+			output += `<div style="margin-top: 7px; padding-left: 12px"><button class="button" style="font-weight: bold" value="/poll submit" name="send" title="${pendingVotes.length ? "Submit your vote" : "View results - you will not be able to vote after viewing results"}">${pendingVotes.length ? "Submit" : "View results"}</button></div>`;
 			output += `</div>`;
 		} else {
 			for (const [num, option] of this.options) {
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -54,13 +54,16 @@ export class Poll {
 
 	select(user: User, option: number) {
 		const userid = user.id;
-		if (!this.pendingVotes[userid]) {
-			this.pendingVotes[userid] = [];
-		} else if (!this.multiPoll) {
-			// can only have one option selected, automatically deselect the previous one
-			this.pendingVotes[userid] = [];
+		if (!this.multiPoll) {
+			// vote immediately
+			this.pendingVotes[userid] = [option];
+			this.submit(user);
+			return;
 		}
 
+		if (!this.pendingVotes[userid]) {
+			this.pendingVotes[userid] = [];
+		}
 		this.pendingVotes[userid].push(option);
 		this.updateFor(user);
 	}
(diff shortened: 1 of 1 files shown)


Asheviere:
Good song, code also looks good from what I can glance. Not entirely sure how I feel about the visibility of the indication that an option is selected. Just the checkbox alone might not be enough I think. I also think changing single option polls to require a submit is a double edged sword, but with the amount of complaints I've gotten about people misclicking the poll, the extra required click might not be bad at all.

whalemer:
Always requiring a confirmation click was something I did mostly intentionally to help prevent misclicks, but if people really object to it it's simple enough to remove.

I'm not sure what else we can do to highlight that an option is selected, I feel like it's already reasonably clear with the checkboxes, since you get feedback about your selection when you click.

The main concern I had with the design was making it clear if the poll supported single or multiple responses. Right now, the only distinguishing feature is that one uses radio buttons and one uses checkboxes, which is a _reasonably_ common pattern, but I don't know if that's clear enough to the user.

DaWoblefet:
I'm in the camp of single option polls being one click only. I'm not sure misclicks due to chat scrolling are worth the extra click on every other poll. I also think that single option polls would be significantly more common than multi-option polls (at least that would be the case from personal experience).

Zarel:
I don't think Submit needs parentheses, and I actually think it shouldn't have parens, and should also be in bold.

Zarel:
I feel like polls need to be even more different, honestly. I think a lot of users will be unaware that these need to be submitted.

I would probably make the poll options not look like buttons, and make it so only the Submit button looks like a button.

whalemer:
>I would probably make the poll options not look like buttons, and make it so only the Submit button looks like a button.

Not quite sure what you mean by this, everything I've tried just makes it look more confusing.

Zarel:
So, you know what the individual poll option buttons currently look like? Make the Submit button look like that.

Then, for the new individual poll option buttons, remove the border and background.

I think at minimum you would need that, for people to realize you're supposed to press Submit.

whalemer:
image
I'm still not sure how to make it clear what you have selected, but I think this isn't completely awful.

server/chat-plugins/poll.ts
```diff
@@ -71,42 +101,48 @@ export class Poll {
...
+			const selected = pendingVotes.includes(num);
+			output += `<div style="margin-top: 5px"><button style="text-align: left; border: none; background: none; color: inherit;${selected ? 'font-weight:bold' : ''}" value="/poll ${selected ? 'de' : ''}select ${num}" name="send" title="${selected ? "Deselect" : "Select"} ${num}. ${Chat.escapeHTML(option.name)}">${selected ? chosen : empty} ${num}. ${this.getOptionMarkup(option)}</button></div>`;
 		}
-		output += `<div style="margin-top: 7px; padding-left: 12px"><button value="/poll results" name="send" title="View results - you will not be able to vote after viewing results"><small>(View results)</small></button></div>`;
+		output += `<div style="margin-top: 7px; padding-left: 12px"><button class="button" value="/poll submit" name="send" title="${pendingVotes.length ? "Submit your vote" : "View results - you will not be able to vote after viewing results"}">${pendingVotes.length ? "Submit" : "View results"}</button></div>`;
```
Zarel: This should be non-button-class and small, for "view results".

It should also be bold, for "Submit". I'd just put them in different `if` blocks, tbh.

server/chat-plugins/poll.ts
```diff
@@ -71,42 +101,48 @@ export class Poll {
...
-	generateVotes() {
+	generateVotes(user: User | null) {
+		const empty = this.multiPoll ? `<i class="fa fa-square-o" aria-hidden="true"></i>` : `<i class="fa fa-circle-o" aria-hidden="true"></i>`;
+		const chosen = this.multiPoll ? `<i class="fa fa-check-square-o" aria-hidden="true"></i>` : `<i class="fa fa-dot-circle-o" aria-hidden="true"></i>`;
+
+		const pendingVotes = (user && this.pendingVotes[user.id]) || [];
 		const iconText = this.isQuiz ? '<i class="fa fa-question"></i> Quiz' : '<i class="fa fa-bar-chart"></i> Poll';
 		let output = `<div class="infobox"><p style="margin: 2px 0 5px 0"><span style="border:1px solid #6A6;color:#484;border-radius:4px;padding:0 3px">${iconText}</span> <strong style="font-size:11pt">${this.getQuestionMarkup()}</strong></p>`;
 		for (const [num, option] of this.options) {
-			output += `<div style="margin-top: 5px"><button class="button" style="text-align: left" value="/poll vote ${num}" name="send" title="Vote for ${num}. ${Chat.escapeHTML(option.name)}">${num}. <strong>${this.getOptionMarkup(option)}</strong></button></div>`;
```
Zarel: We should still use the old style if it's _not_ a multi-poll.

whalemer: Do we want to keep one-click submits for single polls?

Zarel: Yeah, I prefer the "chat doesn't scroll when hovering over a button" option.

whalemer: Sure.

server/chat-plugins/poll.ts
```diff
@@ -3,8 +3,6 @@
  * By bumbadadabum and Zarel.
```
Zarel: Feel free to add yourself to this; I think this is a pretty major contribution.

server/chat-plugins/poll.ts
```diff
@@ -71,42 +104,57 @@ export class Poll {
...
+		if (this.multiPoll) {
+			const empty = `<i class="fa fa-square-o" aria-hidden="true"></i>`;
+			const chosen = `<i class="fa fa-check-square-o" aria-hidden="true"></i>`;
+
+			const pendingVotes = (user && this.pendingVotes[user.id]) || [];
+			for (const [num, option] of this.options) {
+				const selected = pendingVotes.includes(num);
+				output += `<div style="margin-top: 5px"><button style="text-align: left; border: none; background: none; color: inherit;${selected ? 'font-weight:bold' : ''}" value="/poll ${selected ? 'de' : ''}select ${num}" name="send" title="${selected ? "Deselect" : "Select"} ${num}. ${Chat.escapeHTML(option.name)}">${selected ? chosen : empty} ${num}. ${this.getOptionMarkup(option)}</button></div>`;
+			}
+			output += `<div style="margin-top: 7px; padding-left: 12px"><button class="button" style="font-weight: bold" value="/poll submit" name="send" title="${pendingVotes.length ? "Submit your vote" : "View results - you will not be able to vote after viewing results"}">${pendingVotes.length ? "Submit" : "View results"}</button></div>`;
```
Zarel:
```suggestion
			output += `<div style="margin-top: 7px; padding-left: 12px"><button class="button" value="/poll submit" name="send" title="${pendingVotes.length ? "Submit your vote" : "View results - you will not be able to vote after viewing results"}">${pendingVotes.length ? "<strong>Submit</strong>" : "View results"}</button></div>`;
```

whalemer: Should we strong tags for the selected option too?

Zarel: Yes, I would, at least.

PR #6358, opened 2020-02-11
Refactor queue to new BattleQueue class

Zarel:
Previously, battle queue stuff was just strewn around `battle.ts`. This gives it a new home: `battle-queue.ts`.

This was intended to make `battle.ts` slightly more tractable, although the difference is so small that maybe I shouldn't bother. Oh, well, every little bit helps.

### DIFF
--- a/sim/battle-queue.ts
+++ b/sim/battle-queue.ts
@@ -1,0 +1,367 @@
+/**
+ * Simulator Battle Action Queue
+ * Pokemon Showdown - http://pokemonshowdown.com/
+ *
+ * The action queue is the core of the battle simulation. A rough overview of
+ * the core battle loop:
+ *
+ * - chosen moves/switches are added to the action queue
+ * - the action queue is sorted in speed/priority order
+ * - we go through the action queue
+ * - repeat
+ *
+ * @license MIT
+ */
+
+import {Battle} from './battle';
+
+/** A move action */
+export interface MoveAction {
+	/** action type */
+	choice: 'move' | 'beforeTurnMove';
+	order: 3 | 5 | 200 | 201 | 199;
+	/** priority of the action (lower first) */
+	priority: number;
+	/** fractional priority of the action (lower first) */
+	fractionalPriority: number;
+	/** speed of pokemon using move (higher first if priority tie) */
+	speed: number;
+	/** the pokemon doing the move */
+	pokemon: Pokemon;
+	/** location of the target, relative to pokemon's side */
+	targetLoc: number;
+	/** original target pokemon, for target-tracking moves */
+	originalTarget: Pokemon;
+	/** a move to use (move action only) */
+	moveid: ID;
+	/** a move to use (move action only) */
+	move: Move;
+	/** true if megaing or ultra bursting */
+	mega: boolean | 'done';
+	/** if zmoving, the name of the zmove */
+	zmove?: string;
+	/** if dynamaxed, the name of the max move */
+	maxMove?: string;
+	/** effect that called the move (eg Instruct) if any */
+	sourceEffect?: Effect | null;
+}
+
+/** A switch action */
+export interface SwitchAction {
+	/** action type */
+	choice: 'switch' | 'instaswitch';
+	order: 3 | 103;
+	/** priority of the action (lower first) */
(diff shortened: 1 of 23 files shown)


sim/battle-queue.ts
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	debug() {
+		return this.map(action =>
+			// @ts-ignore
+			`${action.order || ''}:${action.priority || ''}:${action.speed || ''}:${action.subOrder || ''} - ${action.choice}${action.pokemon ? ' ' + action.pokemon : ''}${action.move ? ' ' + action.move : ''}`
+		).join('\n') + '\n';
+	}
+
+	sort(DO_NOT_USE?: unknown) {
```
HoeenCoder: Whats this argument for? I'm not aware of `battle.sortQueue` previously taking an argument.

Zarel: It's for TypeScript. Without it, TypeScript complains about `sort` not matching the signature of `Array#sort`

Zarel: It's nice to be able to throw an error if someone tries to pass a comparator, though.

whalemer: I think if you type it as `DO_NOT_USE?: never` then TS will complain if you pass it anything.

Zarel: I'd imagine that if I did that, TypeScript would complain (the same way it complained about having the signature be `sort()`), but apparently it's not complaining about that anymore. I wonder why the inconsistency, but I'm not complaining.

sim/battle-queue.ts
```diff
@@ -0,0 +1,367 @@
...
+	choice: string;
+	[k: string]: any;
+}
+
+/**
+ * Kind of like a priority queue, although not sorted mid-turn in Gen 1-7.
+ *
+ * Sort order is documented in `BattleQueue.comparePriority`.
+ */
+export class BattleQueue extends Array<Action> {
```
Zarel: I'm still not super convinced about `extends Array` instead of having an array as a property (which among other things means we can't overwrite the queue with `= []`.

Certain awkwardnesses:

`queue = []` becomes `queue.splice(0)`

`queue = array` becomes `queue.splice(0); queue.push(...array)`

`queue.slice()` becomes `[...queue]`

`queue = queue.filter` becomes "rewrite it as a for-loop"

Still, though, it's kinda slick, and I don't mind the workarounds. It's easy enough to back out of. In theory.

Zarel: (I'm merging now, but please do still give feedback on this.)

sim/battle-queue.ts
```diff
@@ -0,0 +1,367 @@
...
+	battle: Battle;
+	constructor(battle: Battle) {
+		super();
+		this.battle = battle;
```
urkerab, line 127: The cute thing here is that whenever you call `queue.slice()` or `this.splice()` the return value is a new Queue that is constructed with a single parameter being the number of elements that will be returned. This makes trying to add stuff to the constructor interesting to say the least...

Zarel: Yeah, it was definitely a mistake to extend Array rather than wrap it. Feel free to refactor it to a wrapper instead.

Zarel: Oh the other hand, I would probably override `slice` to return a normal array, and `splice` to return itself.

urkerab: I looked into that but then you've got the ugliness of either referring to `battle.queue.queue` all the time or forwarding all of the Array methods like this:
```typescript
// Delegated array methods
[Symbol.iterator]() {
	return this.queue[Symbol.iterator]();
}
get [0]() {
	return this.queue[0];
}
entries() {
	return this.queue.entries();
}
item(index: number) {
	return this.queue[index];
}
get length() {
	return this.queue.length;
}
push(...actions: Action[]) {
	return this.queue.push(...actions);
}
shift() {
	return this.queue.shift();
}
slice() {
	return this.queue.slice();
}
splice(start: Number, end: Number, ...actions: Action[]) {
	this.splice(start, end, ...actions);
}
unshift(action: Action) {
	return this.queue.unshift(action);
}
```

PR #6359, opened 2020-02-12
Make User#update more typesafe

whalemer:
Would let us catch people mistyping `user.update('blockpms')` and the like, if any of those files were typescripted.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -1021,11 +1021,10 @@ export class User extends Chat.MessageContext {
 	 */
 	getUpdateuserText(...updated: UserSetting[]) {
 		const named = this.named ? 1 : 0;
-		const diff: Partial<Pick<User, UserSetting>> = {};
+		const diff: AnyObject = {};
 		const settings = updated.length ? updated : SETTINGS;
 		for (const setting of settings) {
-			// typescript doesn't seem to understand the setting <- setting assignment
-			diff[setting] = this[setting] as any;
+			diff[setting] = this[setting];
 		}
 		return `|updateuser|${this.getIdentityWithStatus()}|${named}|${this.avatar}|${JSON.stringify(diff)}`;
 	}
--- a/server/users.ts
+++ b/server/users.ts
@@ -412,11 +412,12 @@ export class Connection {
 
 type ChatQueueEntry = [string, RoomID, Connection];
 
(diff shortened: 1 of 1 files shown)


TheJetOU:
Is it actually `keyof User` though? Because `SETTINGS` doesn't contain all of the keys in `User`. Something such as `user.update('connections')` also doesn't look correct.

I would think it would be something like:
```typescript
const SETTINGS = ['isSysop', ...] as const;

function getUpdateuserText(...updated: typeof SETTINGS) { ... }
```

server/users.ts
```diff
@@ -1018,20 +1018,22 @@ export class User extends Chat.MessageContext {
...
-			// @ts-ignore - dynamic lookup
 			diff[setting] = this[setting];
 		}
 		return `|updateuser|${this.getIdentityWithStatus()}|${named}|${this.avatar}|${JSON.stringify(diff)}`;
 	}
 	/**
 	 * @param updated the settings which have been updated or none for all settings.
 	 */
-	update(...updated: string[]) {
+	update(...updated: (keyof User)[]) {
```
Zarel: Wait, why does `update` support partial updates in the first place? Did I approve this?

Zarel: apparently it was #5438

Zarel: `keyof User` seems way too broad, anyway; it should be a subset of `SETTINGS`.

I would do `SETTINGS = [...] as const`, and then `typeof SETTINGS[0]` or something?

whalemer:
Had no idea about `as const`, looks like a nice feature.

server/users.ts
```diff
@@ -1018,20 +1019,20 @@ export class User extends Chat.MessageContext {
 	/**
 	 * @param updated the settings which have been updated or none for all settings.
 	 */
-	getUpdateuserText(...updated: string[]) {
+	getUpdateuserText(...updated: UserSetting[]) {
 		const named = this.named ? 1 : 0;
-		const diff = {};
+		const diff: Partial<Pick<User, UserSetting>> = {};
```
Zarel: tbh I would just let `diff` be `any`. It's about to become JSON anyway, and building JSON usually requires things TS doesn't understand.

Zarel:
Cool, thanks!

PR #6361, opened 2020-02-12
Update VGC banlist

KrisXV:


DaWoblefet:
Looks good at a glance, VGC is debating legality of stuff like Alolan Ninetales / Hoenn Linoone, but most judges appear to be leaning towards allowing (which this PR does).

PR #6367, opened 2020-02-13
Add Let's Go transfer support

KrisXV:
@Zarel


sim/global-types.ts
```diff
@@ -73,6 +73,7 @@ type PokemonSet = {
  * - S = event, 3rd char+ is the index in .eventPokemon
  * - D = Dream World, only 5D is valid
  * - V = Virtual Console transfer, only 7V is valid
+ * - G = LGPE transfer, only 7G is valid
```
Zarel: I would use 8V for this. For the same reasons VC transfer uses 7V rather than 1V and 2V.

Zarel: I guess using 8G is fine. We're not _super_ low on letters. But if we are one day, we might want to consolidate them.

KrisXV: Yeah it was a typo

KrisXV:
I didn't mean to edit the SSB file; I just did a file-wide find and replace for transfered into transferred because I assumed my typo was the only one.

Zarel:
Whoops, should probably have waited for unit tests.

KrisXV:
It passed locally on my end with fully up-to-date dependencies, so I think you're good

PR #6368, opened 2020-02-13
VGC2020: Ban pre-Home regional forms

DaWoblefet:
TPCi has put out a statement to several players via support tickets that Regional forms that were otherwise only obtainable pre-Home are banned.

Here's two judges getting the same thing:

1. https://twitter.com/YureeVGC/status/1228010525512040450
2. https://twitter.com/BillaVGC/status/1228011410887790593

We likely are going to be reversing this after Australian Internationals happen later this month, but for now this should be correct. I also removed Gmax Melmetal since TI added it as part of the standard banlist.


Zarel:
That makes sense and matches how the in-game ladder works.

PR #6372, opened 2020-02-18
Add /shuffle command 

maxalexandderpi:
from https://www.smogon.com/forums/threads/shuffle-command.3660152/

### DIFF
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -2344,7 +2344,7 @@ const commands = {
 		if (!target) return false;
 		if (!target.includes(',')) return this.parse('/help shuffle');
 		if (!this.runBroadcast(true)) return false;
-		let args = target.split(',');
+		const args = target.split(',');
 		if (args.length < 3) return this.errorReply('Give at least 3 arguments to shuffle.');
 		const results = Dex.shuffle(args);
 		return this.sendReplyBox(Chat.html`<em>Shuffled:</em><br> ${results.join(', ')}`);
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -2341,7 +2341,6 @@ const commands = {
 
 	'!shuffle': true,
 	shuffle(target, room, user) {
-		if (!target) return false;
 		if (!target || !target.includes(',')) return this.parse('/help shuffle');
 		const args = target.split(',');
 		if (args.length < 3) return this.parse(`/pick ${target}`);
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -2345,6 +2345,7 @@ const commands = {
 		if (!target.includes(',')) return this.parse('/help shuffle');
 		if (!this.runBroadcast(true)) return false;
 		let args = target.split(',');
+		if (args.length < 3) return this.errorReply('Give at least 3 arguments to shuffle.');
 		const results = Dex.shuffle(args);
 		return this.sendReplyBox(Chat.html`<em>Shuffled:</em><br> ${results.join(', ')}`);
 	},
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -2339,6 +2339,20 @@ const commands = {
 	},
 	pickrandomhelp: [`/pick [option], [option], ... - Randomly selects an item from a list containing 2 or more elements.`],
 
+	'!shuffle': true,
+	shuffle(target, room, user) {
+		if (!target) return false;
+		if (!target.includes(',')) return this.parse('/help pick');
+		if (!this.runBroadcast(true)) return false;
+		if (this.broadcasting) {
+			[, target] = Chat.splitFirst(this.message, ' ');
+		}
+		let args = target.split(',');
+		const results = Dex.shuffle(args);
+		return this.sendReplyBox(Chat.html`<em>Shuffled:</em> ${results.join(', ')}`);
+	},
+	shufflehelp: [`/shuffle [option], [option], [option] ... - Randomly shuffles a list of 3 or more elements.`],
+
 	showimage(target, room, user) {
 		if (!target) return this.parse('/help showimage');
 		if (!this.can('declare', null, room)) return false;
--- a/server/chat-commands/info.js
+++ b/server/chat-commands/info.js
@@ -2345,7 +2345,7 @@ const commands = {
 		const args = target.split(',');
 		if (!this.runBroadcast(true)) return false;
 		if (args.length < 3) return this.parse(`${this.cmdToken}pick ${target}`);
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.js
```diff
@@ -2339,6 +2339,18 @@ const commands = {
...
+	shuffle(target, room, user) {
+		if (!target) return false;
+		if (!target.includes(',')) return this.parse('/help shuffle');
+		if (!this.runBroadcast(true)) return false;
+		let args = target.split(',');
+		if (args.length < 3) return this.errorReply('Give at least 3 arguments to shuffle.');
+		const results = Dex.shuffle(args);
+		return this.sendReplyBox(Chat.html`<em>Shuffled:</em><br> ${results.join(', ')}`);
+	},
+	shufflehelp: [`/shuffle [option], [option], [option] ... - Randomly shuffles a list of 3 or more elements.`],
```
KrisXV:
```suggestion
	shufflehelp: [
		`/shuffle [option], [option], [option], ... - Randomly shuffles a list of 3 or more elements.`,
	],
```

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,18 @@ const commands = {
 	},
 	pickrandomhelp: [`/pick [option], [option], ... - Randomly selects an item from a list containing 2 or more elements.`],
 
+	'!shuffle': true,
+	shuffle(target, room, user) {
+		if (!target) return false;
+		if (!target.includes(',')) return this.parse('/help shuffle');
```
KrisXV: These can be merged into one line of `if (!target || !target.includes(',')) return this.parse('/help shuffle');` imo

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,18 @@ const commands = {
...
+		if (!this.runBroadcast(true)) return false;
+		let args = target.split(',');
```
KrisXV:
```suggestion
		const args = target.split(',');
```

it's never reassigned

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,18 @@ const commands = {
...
+		if (args.length < 3) return this.errorReply('Give at least 3 arguments to shuffle.');
```
KrisXV: Alternatively, you could make it that if the arguments length is less than 3, parse /pick, idk

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,18 @@ const commands = {
...
+		const results = Dex.shuffle(args);
```
whalemer: hyper nit- `Dex.shuffle` works in place, using the return value here might just be misleading.

server/chat-commands/info.js
scheibo: Why is the a length requirement at all? It seems possible to me to be able to shuffle 0, 1 or 2 elements

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,19 @@ const commands = {
 	},
 	pickrandomhelp: [`/pick [option], [option], ... - Randomly selects an item from a list containing 2 or more elements.`],
 
+	'!shuffle': true,
+	shuffle(target, room, user) {
+		if (!target) return false;
+		if (!target || !target.includes(',')) return this.parse('/help shuffle');
+		const args = target.split(',');
+		if (args.length < 3) return this.parse(`/pick ${args}`);
```
KrisXV:
```suggestion
		if (args.length < 3) return this.parse(`/pick ${target}`);
```

if you left it as args it would parse as "/pick [option1, option2]" and show the brackets in the output

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,19 @@ const commands = {
 	},
 	pickrandomhelp: [`/pick [option], [option], ... - Randomly selects an item from a list containing 2 or more elements.`],
 
+	'!shuffle': true,
+	shuffle(target, room, user) {
+		if (!target) return false;
```
whalemer: This check is counterproductive with the line after.

maxalexandderpi: mb, fixed

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,19 @@ const commands = {
...
+		if (!target || !target.includes(',')) return this.parse('/help shuffle');
+		const args = target.split(',');
+		if (args.length < 3) return this.parse(`/pick ${target}`);
```
whalemer: This should probably parse `${this.cmdToken}pick`, so it'll broadcast the pick if you try to broadcast shuffle

server/chat-commands/info.js
Zarel:
```suggestion
		const results = Dex.shuffle(args.map(arg => arg.trim()));
```

This is irrelevant because the args need to be trimmed anyway.

server/chat-commands/info.js
TheJetOU: I can understand 2 elements, but how can you shuffle a list of n < 2 elements? Or, what use case would it fulfill?

Zarel: I support shuffling 2 elements. 0 or 1 should be an error message. It's not like this is a programming API; zero bots should be using it.

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,19 @@ const commands = {
 	},
 	pickrandomhelp: [`/pick [option], [option], ... - Randomly selects an item from a list containing 2 or more elements.`],
 
+	'!shuffle': true,
+	shuffle(target, room, user) {
+		if (!target || !target.includes(',')) return this.parse('/help shuffle');
+		const args = target.split(',');
+		if (!this.runBroadcast(true)) return false;
+		if (args.length < 3) return this.parse(`${this.cmdToken}pick ${target}`);
```
```suggestion
```

This is completely unnecessary. What advantage does `/pick` give for lists of length 2, besides confusing users?

server/chat-commands/info.js
```diff
@@ -2339,6 +2339,19 @@ const commands = {
...
+		const results = Dex.shuffle(args.map(arg => arg.trim()));
+		return this.sendReplyBox(Chat.html`<em>Shuffled:</em><br> ${results.join(', ')}`);
+	},
+	shufflehelp: [
+		`/shuffle [option], [option], [option], ... - Randomly shuffles a list of 3 or more elements.`,
```
```suggestion
		`/shuffle [option], [option], [option], ... - Randomly shuffles a list of 2 or more elements.`,
```

server/chat-commands/info.js
scheibo: 0 and 1 are noops. why write a partial function when you dont need to?

Zarel:
Cool, thanks!

PR #6376, opened 2020-02-19
Typescript Datasearch

KrisXV:


server/chat-plugins/datasearch.ts
```diff
@@ -1709,48 +1874,47 @@ function runLearn(target, cmd) {
 		return {error: "You must specify at least one move."};
 	}
 
-	let lsetProblem;
-	let moveNames = [];
+	let lsetProblem: {[k: string]: any, type: string, moveName: string} = Object.create(null);
```
KrisXV: There was a crash surrounding this in my initial commit; I fixed it, but now running /learn just always results in saying the Pokemon cant learn a move, even if they actually can. I'm honestly not sure how to fix this.

KrisXV: I've managed to fix this but the code could probably be optimized

server/chat-plugins/datasearch.ts
```diff
@@ -10,15 +10,59 @@
...
+	lower: {[k: string]: boolean};
+	zboost: {[k: string]: boolean};
+	status: {[k: string]: boolean};
+	volatileStatus: {[k: string]: boolean};
+	recoil: boolean;
+	skip: boolean;
+}
+
+type Direction = 'less' | 'greater' | 'equal';
+type StatAliases = 'attack' | 'defense' | 'specialattack' | 'spc' | 'special' | 'spatk' | 'specialdefense' | 'spdef' | 'speed' | 'wt' | 'ht' | 'generation';
```
Zarel: Is this necessary? As far as I can tell, this provides zero type safety, since you just cast straight past this.

server/chat-plugins/datasearch.ts
```diff
@@ -10,15 +10,59 @@
 
 'use strict';
 
+import {QueryProcessManager} from '../../lib/process-manager';
+import {Repl} from '../../lib/repl';
+import {Chat} from '../chat';
+import * as CL from '../config-loader';
+import {Dex} from '../../sim/dex';
```
Zarel: Where were these imports previously? Isn't this file supposed to just use the global one?

KrisXV: Child processes manually imported all of them

Zarel: I think the child processes should continue doing that, to allow for monkeypatching and hotpatching.

Zarel: You should `ts-ignore` or whatever as necessary for that.

KrisXV: Tried doing this with tslint dsabling and ts-ignoring where necessary and got this error
```
CRASH: TypeError: Dex.includeData is not a function
    at Object.<anonymous> (C:\Users\feeds\Desktop\Pokemon-Showdown\.server-dist\chat-plugins\datasearch.js:2049:6)
    at Module._compile (internal/modules/cjs/loader.js:777:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:788:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:840:10)
    at internal/main/run_main_module.js:17:11
```

Zarel: You're importing the wrong thing as Dex. You're probably doing `require('..dex').includeData()` instead of `require('..dex').Dex.includeData()`

KrisXV: yep, fixed

server/chat-plugins/datasearch.ts
```diff
@@ -86,6 +79,7 @@ export const commands: ChatCommands = {
 			cmd: 'dexsearch',
 			canAll: (!this.broadcastMessage || (room && room.isPersonal)),
 			message: (this.broadcastMessage ? "" : message),
+		// @ts-ignore
```
Zarel: Why does this need to be ts-ignored now?

Zarel: `runSearch`'s return signature should still be defined in-file, right?

KrisXV: because it has no designated type since not manually importing QueryProcessManager like i did before

Zarel: Shouldn't `QueryProcessManager` be a global? Otherwise how did it access it in the past?

Zarel: Ah, looks like it's imported inside this file. `QueryProcessManager` should still be imported at the top; it's just the other files that shouldn't.

KrisXV: did

server/chat-plugins/datasearch.ts
Zarel: bump

KrisXV: nope, removed

server/chat-plugins/datasearch.ts
```diff
@@ -10,15 +10,53 @@
...
+	boost: {[k: string]: boolean};
+	lower: {[k: string]: boolean};
+	zboost: {[k: string]: boolean};
+	status: {[k: string]: boolean};
+	volatileStatus: {[k: string]: boolean};
+	recoil: boolean;
+	skip: boolean;
+}
+
+type Direction = 'less' | 'greater' | 'equal';
```
Zarel: I'm not _particularly_ happy with this either, but I guess it's used for some minor safety in some places, we can keep it.

scheibo: I realised its not what you're complaining about, but I wonder why we went with the english words instead of the more concise `'<' | '=' | '>'` instead

PR #6378, opened 2020-02-20
Refactor Dragon Darts

urkerab:
This makes `move.smartTarget` a three-state. I was considering turning it into an integer though.

If it's `true`, then this means that there are two targets, so that Wonder Guard, Protect, semi-invulnerability, type immunity and evasion should be silent.

However, as soon as one of these effects applies, it then goes and sets it to `false`, which means that further failure messages for the other target will appear normally.

Thus if Dragon Darts misses both targets due to evasion, only the neighbouring ally's miss message will appear, as desired.

Note that `false` still counts as smart targeting from the point of view of the hit step loop.

I also tweaked the early exit condition from the hit step loop so that Dragon Darts can hit both foes even if the first faints.

### DIFF
--- a/data/scripts.js
+++ b/data/scripts.js
@@ -631,7 +631,13 @@ let BattleScripts = {
 				targetsCopy = targets.slice(0);
 			}
 			let target = targetsCopy[0]; // some relevant-to-single-target-moves-only things are hardcoded
-			if (target && typeof move.smartTarget === 'boolean') this.addMove('-anim', pokemon, move.name, target);
+			if (target && typeof move.smartTarget === 'boolean') {
+				if (hit > 1) {
+					this.addMove('-anim', pokemon, move.name, target);
+				} else {
+					this.retargetLastMove(target);
+				}
+			}
 
 			// like this (Triple Kick)
 			if (target && move.multiaccuracy && hit > 1) {
--- a/data/scripts.js
+++ b/data/scripts.js
@@ -378,8 +378,12 @@ let BattleScripts = {
 		const hitResults = this.runEvent('Invulnerability', targets, pokemon, move);
 		for (const [i, target] of targets.entries()) {
 			if (hitResults[i] === false) {
-				if (!move.spreadHit) this.attrLastMove('[miss]');
-				this.add('-miss', pokemon, target);
+				if (move.smartTarget) {
+					move.smartTarget = false;
+				} else {
+					if (!move.spreadHit) this.attrLastMove('[miss]');
+					this.add('-miss', pokemon, target);
+				}
 			}
 		}
 		return hitResults;
@@ -402,7 +406,8 @@ let BattleScripts = {
 
 		const hitResults = [];
 		for (let i = 0; i < targets.length; i++) {
-			hitResults[i] = (move.ignoreImmunity && (move.ignoreImmunity === true || move.ignoreImmunity[move.type])) || targets[i].runImmunity(move.type, true);
+			hitResults[i] = (move.ignoreImmunity && (move.ignoreImmunity === true || move.ignoreImmunity[move.type])) || targets[i].runImmunity(move.type, !move.smartTarget);
+			if (move.smartTarget && !hitResults[i]) move.smartTarget = false;
 		}
 
 		return hitResults;
@@ -480,9 +485,12 @@ let BattleScripts = {
 				accuracy = this.runEvent('Accuracy', target, pokemon, move, accuracy);
 			}
 			if (accuracy !== true && !this.randomChance(accuracy, 100)) {
-				if (!move.spreadHit) this.attrLastMove('[miss]');
-				const hideFailure = move.smartTarget && (i < targets.length - 1 || hitResults.some(Boolean));
-				if (!hideFailure) this.add('-miss', pokemon, target);
+				if (move.smartTarget) {
+					move.smartTarget = false;
+				} else {
+					if (!move.spreadHit) this.attrLastMove('[miss]');
+					this.add('-miss', pokemon, target);
+				}
 				if (pokemon.hasItem('blunderpolicy') && pokemon.useItem()) this.boost({spe: 2}, pokemon);
 				hitResults[i] = false;
 				continue;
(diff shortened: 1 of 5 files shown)


Zarel:
Ugh, I already fixed this, but I like your fix better.

Zarel:
I'm going to push my fix, and you can rebase on top of it. Your implementation does seem much closer to how Dragon Darts is implemented in-game.

Zarel:
`move.smartTarget !== undefined` should probably not be used (PS convention is to treat `null` and `undefined` identically). I would use `typeof move.smartTarget === 'boolean'`.

data/scripts.js
```diff
@@ -378,8 +378,12 @@ let BattleScripts = {
 		const hitResults = this.runEvent('Invulnerability', targets, pokemon, move);
 		for (const [i, target] of targets.entries()) {
 			if (hitResults[i] === false) {
-				if (!move.spreadHit) this.attrLastMove('[miss]');
-				this.add('-miss', pokemon, target);
+				if (move.smartTarget) {
+					move.smartTarget = false;
+				} else {
+					if (!move.spreadHit) this.attrLastMove('[miss]');
```
Zarel: I actually like my implementation since it can also be used to `attrLastMove('[miss]')` if all hits of a spread hit miss.

sim/pokemon.ts
```diff
@@ -602,33 +602,21 @@ export class Pokemon {
 	}
 
 	/** Get targets for Dragon Darts */
-	getSmartTargets(target: Pokemon, move: Move) {
+	getSmartTargets(target: Pokemon, move: ActiveMove) {
```
Zarel: Do we need this function?

sim/pokemon.ts
```diff
@@ -602,33 +602,21 @@ export class Pokemon {
...
+		this.battle.attrLastMove('[still]');
```
Zarel: Why? I prefer to special-case it for nicer-looking logs.

Zarel: Also `attrLastMove` is very not-functional side-effect-wazoo, so I'd prefer to avoid it where possible.

sim/pokemon.ts
Zarel: This seems short enough that we might want to inline it in `getMoveTargets`... but then again, maybe not.

data/scripts.js
```diff
@@ -614,15 +623,15 @@ let BattleScripts = {
 		for (hit = 1; hit <= targetHits; hit++) {
 			if (damage.includes(false)) break;
 			if (hit > 1 && pokemon.status === 'slp' && !isSleepUsable) break;
-			if (targets.some(target => target && !target.hp)) break;
+			if (targets.every(target => target && !target.hp)) break;
 			move.hit = hit;
 			if (move.smartTarget && targets.length > hit - 1) {
```
Zarel: Oh, this is ugly and not good.

sim/pokemon.ts
urkerab: Actually it's the intended target being excluded and both hits going to the other target that was the special case I was trying to solve here.

sim/pokemon.ts
urkerab: I think we might still need it to compute the initial list of targets depending on Ally Switch etc.

data/scripts.js
urkerab: I guess this used to work because it used to be the case that at least one of `targetHits` and `targets.length` was 1, so that it made no difference whether you used `some` or `every`.

sim/pokemon.ts
Zarel: Ohh, that makes sense.

`retargetLastMove` _does_ exist. I think we should just always call it on the first hit.

data/scripts.js
Zarel: Oh, no, I was specifically referring to `targets.length > hit - 1`, which I just fixed in 1d09dd11748176 but was confused how it worked in the past.

sim/pokemon.ts
Zarel: Understandable. Let's keep this in for now.

data/scripts.js
urkerab: Oh, are you talking about the `&&` to `||` change? I did wonder about that...

data/scripts.js
urkerab: What's correct behaviour for a mixture of move failures e.g. a Protect, a Flying type immunity and an evasion miss while attempting an Earthquake?

urkerab:
> move.smartTarget !== undefined should probably not be used (PS convention is to treat null and undefined identically). I would use typeof move.smartTarget === 'boolean'.

I knew there was something else I meant to ask you about, rather than just guessing... I've updated my other use to match.

data/scripts.js
Zarel: If it's not Dragon Darts, all the messages should appear, I think.

data/scripts.js
```diff
@@ -619,11 +627,11 @@ let BattleScripts = {
 			move.hit = hit;
 			if (move.smartTarget && targets.length > 1) {
 				targetsCopy = [targets[hit - 1]];
-				if (targetsCopy[0] && hit > 1) this.addMove('-anim', pokemon, move.name, targetsCopy[0]);
 			} else {
 				targetsCopy = targets.slice(0);
 			}
 			let target = targetsCopy[0]; // some relevant-to-single-target-moves-only things are hardcoded
+			if (target && typeof move.smartTarget === 'boolean') this.addMove('-anim', pokemon, move.name, target);
```
Zarel: This should still use `retargetLastMove` for the first hit, instead of the `[still]` hack...

Zarel:
You should probably add a test for Electrify Dragon Darts vs Motor Drive. This refactor fixes that implementation.

urkerab:
The test checks that the Motor Drive user does not get damaged but gets their speed boosted, while their ally does get damaged. Is that enough?

data/scripts.js
Zarel: bump

Zarel:
Follow my test format: Use Stamina to detect how many hits you take.

urkerab:
Ah right, I had copypasta'd the wrong test to base my test on...

Zarel:
I can't tell if you're missing this bump:

https://github.com/smogon/pokemon-showdown/pull/6378/commits/a7e2253a5367f2df424598b9f95fb6c3ff761cb8#r382343166

urkerab:
I thought that the (2) was due to the title change rather than a second comment, so I missed that. Sorry about that. I've added a commit for that.

PR #6380, opened 2020-02-21
Automatically upload replays when reporting

whalemer:


server/rooms.ts
```diff
@@ -27,6 +27,7 @@ import {GTSGiveaway, LotteryGiveaway, QuestionGiveaway} from './chat-plugins/wif
 import {PM as RoomBattlePM, RoomBattle, RoomBattlePlayer, RoomBattleTimer} from "./room-battle";
 import {RoomGame, RoomGamePlayer} from './room-game';
 import {Roomlogs} from './roomlogs';
+import * as crypto from 'crypto';
```
urkerab: You forgot to remove the `require` from `core.js` (this is just an eslint warning so you may have overlooked it).

PR #6382, opened 2020-02-22
Add Config.startuphook

monsanto:
This is called when the server is fully initialized and ready to serve requests. Useful for notifying a process manager.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -415,6 +415,12 @@ exports.disablehotpatchall = false;
 exports.forcedpublicprefixes = [];
 
 /**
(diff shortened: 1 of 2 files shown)


monsanto:
For anyone that Googles "pokemon showdown systemd", the following can be used to tell systemd the service is ready:

```js
exports.startuphook = function () {
	if (process.env["INVOCATION_ID"]) {
		require('child_process').spawnSync("systemd-notify", ["--ready", `--pid=${process.pid}`]);
	}
};
```

Use `Type=notify` in the `[Service]` section of your unit.

PR #6386, opened 2020-02-22
Add ADV UU NFE Clause

KrisXV:
Even though there is discussion about its existence, I was asked by hogg to implement it for the time being so that ADV UU is being accurately simulated.

I originally did this with ``onValidateSet`` and ``template.nfe``, but the ``template.nfe`` check isnt supported in previous generations.

### DIFF
--- a/data/rulesets.js
+++ b/data/rulesets.js
@@ -835,7 +835,13 @@ let BattleFormats = {
 		effectType: 'ValidatorRule',
 		name: 'Standard Pet Mod',
 		desc: "Holds all custom Pet Mod ruleset validation",
-		// Implemented in mods/petmod/rulesets.js
+		// Implemented in mods/[petmod]/rulesets.js
+	},
+	uunfeclause: {
+		effectType: 'ValidatorRule',
+		name: 'UU NFE Clause',
+		desc: "Bans all NFE Pokemon, except Scyther, from [Gen 3] UU.",
+		// Implemented in mods/gen3/rulesets.js
 	},
 };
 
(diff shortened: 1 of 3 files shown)


data/rulesets.js
```diff
@@ -835,7 +835,13 @@ let BattleFormats = {
...
 		name: 'Standard Pet Mod',
 		desc: "Holds all custom Pet Mod ruleset validation",
-		// Implemented in mods/petmod/rulesets.js
+		// Implemented in mods/[petmod]/rulesets.js
+	},
+	uunfeclause: {
+		effectType: 'ValidatorRule',
+		name: 'UU NFE Clause',
+		desc: "Bans all NFE Pokemon, except Scyther, from [Gen 3] UU.",
+		// Implemented in mods/gen3/rulesets.js
```
Zarel: Why not just implement it here?

Zarel: Also why not do it programmatically?

KrisXV: It doesn't support past gen Pokemon that were formerly not NFE, like I said. For example, in generation 3, it marks tangela as NFE even though it didnt become an NFE Pokemon until generation 4. If I did the banlist in the main rulesets.js, it would be inaccurate because it would have the same issue.

Zarel: That's annoying.

Zarel: But having no bans in the main rulesets.js is kind of also wrong. We'd probably need a special "past gen only" flag/error or something.

PR #6391, opened 2020-02-24
[WIP] Make FS safer to use concurrently

Kaiepi:
Some of `FSPath`'s methods are known to be racey. This is currently
"fixed" by piling more behaviour on top of them and hoping for the best.
Instead, make the methods safe to use concurrently in the first place by
implementing a simulation of file locks (which Node.js does not support
natively), and use them to protect the syscalls performed by its
methods.

Known issues:
- This is a naive implementation and will be inefficient if multiple method calls are used in succession.
- `FSPath` method calls aren't guaranteed to be completed before the process exits, so file locks can still exist when the server restarts. This appears to already be a problem with existing FS code, but becomes more of an issue with this.

Todo:
- Ensure sync and async `FSPath` methods cooperate WRT file locking.
- Ensure file locking works properly with `FSPath`'s stream methods.
- Remove `FSPath#writeUpdate` or replace it with something like `FSPath#throttledWrite` if its throttling is worth keeping.

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -41,69 +41,132 @@ const pendingUpdates = new Map<string, PendingUpdate>();
 
 class FSPath {
 	path: string;
+	queue: Promise<any>;
 
 	constructor(path: string) {
 		this.path = pathModule.resolve(ROOT_PATH, path);
+		this.queue = Promise.resolve();
 	}
 
 	parentDir() {
 		return new FSPath(pathModule.dirname(this.path));
 	}
 
+	lock(): Promise<number> {
+		return new Promise((resolve, reject) => {
+			fs.open(`${this.path}.lock`, 'ax', 0o000, (err, fd) => {
+				err ? reject(err) : resolve(fd);
+			});
+		});
+	}
+
+	unlock(fd: number): Promise<true> {
+		return new Promise((resolve, reject) => {
+			fs.close(fd, closeErr => {
+				if (closeErr) return reject(closeErr);
+				fs.unlink(`${this.path}.lock`, unlinkErr => {
+					unlinkErr ? reject(unlinkErr) : resolve(true);
+				});
+			});
+		});
+	}
+
+	async protect<T>(syscall: () => Promise<T>): Promise<T> {
+		return this.queue = this.queue.then(() => this.doProtect(syscall));
+	}
+	private async doProtect<T>(syscall: () => Promise<T>): Promise<T> {
+		const lock = await this.lock();
+		const result = await syscall();
+		await this.unlock(lock);
+		return result;
+	}
+
+	lockSync(): number {
+		return fs.openSync(`${this.path}.lock`, 'ax', 0o000);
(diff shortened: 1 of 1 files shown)


Zarel:
I don't see the point of this.

Zarel:
You want to add locks so we can't do things concurrently? Being able to do things concurrently is the whole point, though.

Zarel:
> This is a naive implementation and will be inefficient if multiple method calls are used in succession.

Yes, your locks are global rather than file-level.

But honestly, this is overengineering and I feel like we have no need for it. We already have `writeUpdate` which adds _more_ safety than this PR provides, and `createWriteStream`/`createAppendStream` for those use-cases. The other functions are for the general use case of "conflicts are not a problem".

As far as I can tell, this PR adds overhead for no benefit.

Kaiepi:
You're right about this going way too far for the problem it sets out to solve and being a bad implementation, but I made this PR because I don't think `FSPath#writeUpdate` is the right solution either. It adds a lot of complexity to writing, which as far as I can tell is only done because `FSPath#write` doesn't behave well when used concurrently and so people can throttle writes, but more importantly it allows the process to exit while waiting for any writes scheduled with it. If that ever happens, you probably wouldn't lose the entire file, but you'll lose any changes scheduled to be written.

I think `FSPath#write` should just be safe to use concurrently in the first place. This doesn't need to keep a lock file in the filesystem to be possible to do when only one instance of PS is being run at a time anyway; state for whether or not there is an ongoing write and its promise (if any) could be kept instead. This would allow people to await writes that would be handled by `FSPath#writeUpdate` currently and would make it possible to wait for any ongoing writes to finish before the process exits (unless node coredumps or something).

scheibo:
> state for whether or not there is an ongoing write and its promise (if any) could be kept instead. This would allow people to await writes that would be handled by `FSPath#writeUpdate` currently and would make it possible to wait for any ongoing writes to finish before the process exits (unless node segfaults or something).

This would fix the main problem I have with the current FS API.

Zarel:
> It adds a lot of complexity to writing, which as far as I can tell is only done because `FSPath#write` doesn't behave well when used concurrently and so people can throttle writes

I feel like it's a relatively small amount of complexity: You just pass a callback rather than writing the contents directly.

I also think it's the correct level of abstraction: If you call `FS.write()` ten times in a row, only the last call should count, and the first nine calls should never be made in the first place. Passing a callback allows the middle eight callbacks to not be called at all. Your approach doesn't have this property.

> but more importantly it allows the process to exit while waiting for any writes scheduled with it.

Yes, it does that too, which yours doesn't. Yet another reason why I prefer my implementation.

> I think `FSPath#write` should just be safe to use concurrently in the first place.

I feel like we have a safe write function and a low-overhead write function, and it's good to have both. I'm not sure what the middle-ground option offers in addition to this.

AnnikaCodes:
This is a WIP PR from a contributor who has left.

PR #6392, opened 2020-02-25
Fix bugs with renaming rooms

whalemer:
Recreate the logs properly, fix aliases

Incidentally, should the rename command modlog, or log to the staff room?

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1373,7 +1373,14 @@ export class BasicChatRoom extends BasicRoom {
 				Rooms.aliases.set(alias, newID);
 			}
 		}
+		// add an alias from the old id
 		Rooms.aliases.set(oldID, newID);
+		if (!this.aliases) this.aliases = [];
+		this.aliases.push(oldID);
+		if (this.chatRoomData) {
+			this.chatRoomData.aliases = this.aliases;
+			Rooms.global.writeChatRoomData();
+		}
 
 		for (const user of Object.values(this.users)) {
 			user.inRooms.delete(oldID);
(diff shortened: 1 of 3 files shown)


server/rooms.ts
```diff
@@ -1368,6 +1368,12 @@ export class BasicChatRoom extends BasicRoom {
 		Rooms.rooms.delete(oldID);
 		Rooms.rooms.set(newID, this as ChatRoom);
 
+		for (const [alias, roomid] of Rooms.aliases.entries()) {
+			if (roomid === oldID) {
+				Rooms.aliases.set(alias, newID);
```
TheJetOU: You should delete the old aliases now that it's useless.

whalemer: It's an alias -> roomid map, I'm just updating it from `alias -> oldid` to `alias -> newid`

TheJetOU: Ah, right.

Zarel: Unrelatedly, do we set an alias from the old room to the new room? That should probably be done.

Zarel:
Notifying Staff is weird because secret rooms usually only notify Upper Staff. I'm fine with the exception but we might want to think about something more consistent.

whalemer:
>Notifying Staff is weird because secret rooms usually only notify Upper Staff. 

Isn't that what my code does? see https://github.com/whalemer/Pokemon-Showdown/blob/c3cf7d4e721d77a5da3630eb6d1caa1972b5d2cc/server/chat-commands/room-settings.ts#L833-L834

server/rooms.ts
whalemer: Sure.

Zarel:
Ah, I missed that while reading.

PR #6393, opened 2020-02-25
Add some common rooms to the roomid type

whalemer:


Zarel:
Why did the first one fail unit tests? I couldn't figure out the error message.

whalemer:
I accidentally changed some things from `function(param = 'thing' as RoomID)` to `function(param: thing)`, which changes the param type from RoomID to string

Zarel:
Hmm.

Zarel:
I'm not _sure_ if I like this...

PR #6394, opened 2020-02-25
Update CAP learnsets for SwSh

KrisXV:


KrisXV:
The rest of the learnsets will be updated within the next week

PR #6395, opened 2020-02-25
Correct validation for UU NFE Clause/Little Cup

KrisXV:


data/rulesets.js
```diff
@@ -292,6 +292,10 @@ let BattleFormats = {
 			if (!template.nfe || futureGenEvo) {
 				return [set.species + " doesn't have an evolution family."];
 			}
+			// Temporary hack for LC past-gen formats and other mashups
+			if (set.level !== 5) {
+				return [`${set.species} needs to be level 5.`];
```
Zarel: LC requires level 5, it's not just a cap?

KrisXV: True, fixed

data/rulesets.js
```diff
@@ -840,8 +844,15 @@ let BattleFormats = {
 	uunfeclause: {
 		effectType: 'ValidatorRule',
 		name: 'UU NFE Clause',
-		desc: "Bans all NFE Pokemon, except Scyther, from [Gen 3] UU.",
-		// Implemented in mods/gen3/rulesets.js
+		desc: "Bans all NFE Pokemon, barring some exceptions",
```
Zarel:
```suggestion
		desc: "Bans all NFE Pokemon, except Scyther in Gen 3",
```

data/rulesets.js
```diff
@@ -292,6 +292,10 @@ let BattleFormats = {
 			if (!template.nfe || futureGenEvo) {
 				return [set.species + " doesn't have an evolution family."];
 			}
+			// Temporary hack for LC past-gen formats and other mashups
+			if (set.level > 5) {
+				return [`${set.species} needs to be level 5 or lower.`];
```
Zarel:
```suggestion
				return [`${set.species} can't be above level 5 in Little Cup formats.`];
```

PR #6401, opened 2020-02-28
Fix /tour autodq off

tmagicturtle:
The most recent change to this file prevents "/tour autodq off" from working and sends a "That isn't a valid timeout value." error to the client. This is because it checks for a condition where timeout is greater than 1 hour, but when turning autodq off, the timeout variable is set to Infinity.

This proposed change adds a check in the condition for if the timeout value is Infinity and does not send the error if that is the case (but if timeout is a numeric value greater than 1 hour, it still will fail).


Zarel:
Thanks!

PR #6405, opened 2020-02-29
1v1: Simplify banlist

KrisXV:


TheImmortal:
Is it really simplifying though? It's unclear that Dusk-Mane and Dawn-Wings are banned from looking at the banlist in ``/om 1v1``. A regular user would see that formes are specifically banned. I would revert this.

TheImmortal:
Megas, I can understand. If Mewtwo is banned, obviously you can't use Mega Mewtwo X/Y as you have to start the battle as Mewtwo. But Necrozma are different Pokemon, not in-battle forme changes.

Zarel:
I feel like if you don't specify a specific forme, you mean "all its formes". This has always been PS's default expectation.

PR #6407, opened 2020-02-29
Add ENVy to Pokemon Showdown

gbateman:
Hello!
​
We (@magmastonealex, @omstrumpf, @gbateman, @Tim-Willard) have been working on a tool called ENVy.
​
ENVy is an environment manager to make working on projects easier. Setup with ENVy is as simple as typing `envy up` for any project you want to work on. Running `envy nuke` removes all traces of that environment. With ENVy, you don't need to hunt down a lint command or how you run the tests. `envy -h` shows you all the commands you can run - or just jump into a shell with `envy shell`.
​
You can read more about ENVy [here](http://envy-project.github.io/), but the short version is that we think it'll make it easier for contributors to help out. 
​
This PR adds a `.envyfile`, the file that describes a development environment and contains the commands to run. We used pokemon-showdown while developing in order to test ENVy, and we hoped you'd consider commiting this file to help out casual contributors.

While pokemon-showdown does not require any particularly difficult-to-set-up native packages, ENVy can ensure consistency in the environment, as the project will be run in isolation from individual user environments. This helps to remove issues with old versions of node, or first-time contributor environments.

We hope you consider using ENVy :)

### DIFF
--- a/.envyfile
+++ b/.envyfile
@@ -1,0 +1,37 @@
+environment:
+  base:
+    image: node:latest
+  setup-steps:
+    - name: npm install
+      label: 'Installing NPM packages'
+      type: script
+      run:
+        - 'npm install --dev'
+      triggers:
+        files:
+          - package.json
+          - package-lock.json
+
+actions:
+  - name: 'pokemon-showdown'
+    script: './pokemon-showdown'
+    help: 'Runs the pokemon-showdown script'
+  - name: start
+    script: 'npm run start 8000'
+    help: 'Starts a Pokemon Showdown server on port 8000'
+  - name: lint
+    script: 'npm run lint && echo "Linting Succeeded"'
+    help: 'Lint the project'
+  - name: build
+    script: 'npm run build && echo "Build Succeeded"'
+    help: 'Build the project'
+  - name: test
+    script: 'npm run test'
+    help: 'Run mocha tests'
+  - name: npm
+    script: 'npm run'
+    help: 'Run an npm script'
+
+network:
+  ports: 
+    - '8000:8000'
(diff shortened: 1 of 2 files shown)


Zarel:
Interesting. I don't think we need an environment manager, but "showing all the commands you can run" is interesting; there's definitely not a _good_ way to do that right now.

Zarel:
Looks like it requires Docker? We don't really use Docker, either; it's overkill for our situation. #6184 is currently on hold because we still don't really see a good reason for it.

Zarel:
I'm not opposed to steps that would make PS easier to deploy, for others, but I do need a clearer idea of what the benefit is. So far it seems like this is still more complicated than PS's current "clone and run `./pokemon-showdown`" deployment.

gbateman:
> Looks like it requires Docker? We don't really use Docker, either; it's overkill for our situation. #6184 is currently on hold because we still don't really see a good reason for it.

That's true, using ENVy does require installing Docker and ENVy itself.

> I'm not opposed to steps that would make PS easier to deploy, for others, but I do need a clearer idea of what the benefit is. So far it seems like this is still more complicated than PS's current "clone and run ./pokemon-showdown" deployment.

ENVy wouldn't make deployment easier - given that ENVy helps with creating a container using sane defaults, there could be a world where deployment tools were added to ENVy, however that's not currently planned.

I think you're probably right that this makes setup more complicated than it needs to be given the lack of native dependencies and other complicated setup steps required for development on Pokemon Showdown.

Thanks for your feedback :) - closing this PR as I agree the upside isn't worth the complication for Pokemon Showdown.

PR #6412, opened 2020-03-01
Scalemons: Ban Huge Power

KrisXV:


KrisXV:
No link, I am the metagame leader, it was supposed to be banned from the get-go

Zarel:
You could still link to the original announcement, which presumably would mention that it should be banned.

Zarel:
Unless this is a weird metagame with secret bans, which I doubt?

PR #6413, opened 2020-03-02
Inheritance: Add basic move validation support

KrisXV:
This fixes some some legal sets not being able to validate, such as a normal Abomasnow set not validating because it thinks the set is Vanillish.


Zarel:
I assume this fixes https://twitter.com/PolarhumeDuWeb/status/1234574713394851841

KrisXV:
Partially yes. the validator is very stingy for reasons i dont know how to fix

PR #6416, opened 2020-03-04
Add a skipSets parameter to validateTeam

scheibo:
This option allows for decoupling intra-set validation from the inter-set team validations which is useful for the team validator to be used incrementally more efficiently in a generative context (i.e. each time a new set is added to a team the team can be validated and then *just* the new set, ie. N set validations instead of N!)

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -202,19 +202,23 @@ export class TeamValidator {
 
 	validateTeam(
 		team: PokemonSet[] | null,
-		removeNicknames = false,
-		skipSets: Record<string, Record<string, boolean>> = {}
+		options: {
+			removeNicknames?: boolean,
+			skipSets?: {[name: string]: {[key: string]: boolean}},
+		} = {}
 	): string[] | null {
 		if (team && this.format.validateTeam) {
-			return this.format.validateTeam.call(this, team, removeNicknames, skipSets) || null;
+			return this.format.validateTeam.call(this, team, options) || null;
 		}
-		return this.baseValidateTeam(team, removeNicknames, skipSets);
+		return this.baseValidateTeam(team, options);
 	}
 
 	baseValidateTeam(
 		team: PokemonSet[] | null,
-		removeNicknames = false,
-		skipSets: Record<string, Record<string, boolean>> = {}
+		options: {
+			removeNicknames?: boolean,
+			skipSets?: {[name: string]: {[key: string]: boolean}},
+		} = {}
 	): string[] | null {
 		const format = this.format;
 		const dex = this.dex;
@@ -253,8 +257,8 @@ export class TeamValidator {
 			if (!set) return [`You sent invalid team data. If you're not using a custom client, please report this as a bug.`];
 
 			let setProblems: string[] | null = null;
-			if (skipSets[set.name]) {
-				for (const i in skipSets[set.name]) {
+			if (options.skipSets && options.skipSets[set.name]) {
+				for (const i in options.skipSets[set.name]) {
 					teamHas[i] = (teamHas[i] || 0) + 1;
 				}
 			} else {
@@ -270,7 +274,7 @@ export class TeamValidator {
 			if (setProblems) {
 				problems = problems.concat(setProblems);
 			}
-			if (removeNicknames) {
+			if (options.removeNicknames) {
 				let crossTemplate: Template;
 				if (format.name === '[Gen 7] Cross Evolution' && (crossTemplate = dex.getTemplate(set.name)).exists) {
 					set.name = crossTemplate.species;
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -200,14 +200,14 @@ export class TeamValidator {
 			this.ruleTable.minSourceGen[0] : 1;
 	}
 
-	validateTeam(team: PokemonSet[] | null, removeNicknames = false): string[] | null {
+	validateTeam(team: PokemonSet[] | null, removeNicknames = false, skipSets = false): string[] | null {
 		if (team && this.format.validateTeam) {
-			return this.format.validateTeam.call(this, team, removeNicknames) || null;
+			return this.format.validateTeam.call(this, team, removeNicknames, skipSets) || null;
 		}
-		return this.baseValidateTeam(team, removeNicknames);
+		return this.baseValidateTeam(team, removeNicknames, skipSets);
 	}
 
-	baseValidateTeam(team: PokemonSet[] | null, removeNicknames = false): string[] | null {
+	baseValidateTeam(team: PokemonSet[] | null, removeNicknames = false, skipSets = false): string[] | null {
 		const format = this.format;
 		const dex = this.dex;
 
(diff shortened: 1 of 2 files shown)


Zarel:
I'm not opposed, but I wonder if there's a better architecture here.

Zarel:
Also, on second thought, I think this is invalid for complex bans (like `! Drizzle ++ Swift Swim`), you'd need to refactor `setHas` and `teamHas`.

Do you still want to go forward with this?

scheibo:
> I'm not opposed, but I wonder if there's a better architecture here.

Originally I was interested in being able to perform validations on **partial** sets (ie. validate the details that are present, but don't complain about required details missing), **fast failures** (aborting on the first error instead of accumulating all problems) and ideally **incremental** validation (not redoing work from previous partial validations), but what I started looking into trying to refactor things to work that way I lost hope (partly because of how it would mean changing all of the format specific methods as well, but mostly just because validation is already quite complex and I worry about adding to the complexity with further options). I also decided validation is going to be too slow for the my most important use case, not necessarily by virtue of the code but because of how involved it is and how often it will actually be relevant for what I need. 

This `skipSets` param (or some version of it) covers like 90% of what I care about in a simple way (hopefully) without adding too much complexity. `skipSets` unfortunately requires knowledge of the inner workings of the validator to use correctly (see below), but given the use case is niche and the optimization is large I'm OK with that from my end.

If you have thoughts on a better architecture for what I'm after I'm all for it, I just dont think I have enough context to take along a large rework of what from the outside seems to be a somewhat fragile and critical piece of the app.

> Also, on second thought, I think this is invalid for complex bans (like ! Drizzle ++ Swift Swim), you'd need to refactor setHas and teamHas.
> Do you still want to go forward with this?

Yes. I will refactor to account for `setHas`/`teamHas` - thanks for catching! I am already forced to call `checkSpecies` to get tier legality (though that only needs to be done once per species in the lifetime of the application) and thats where the majority of `setHas` appears to be set (item/ability/move seem fairly trivial fill in)? I think this can maybe make the `skipSets` param be a map of `name => setHas`, and then do like

```ts
let setProblems: string[] | null = null;
if (skipSets[set.name]) {
	for (const i in skipSets[set.name]) {
		teamHas[i] = (teamHas[i] || 0) + 1;
	}
} else {
	setProblems = (format.validateSet || this.validateSet).call(this, set, teamHas);
}
```

The problem is once again overrides by formats - from what I can tell, outside of the Inheritance OM nothing mutations `setHas`/`teamHas` outside of validator (currently!) and Inheritance would actually still work with my proposed code above given whats its currently doing if im not mistaken. The risk is that anything *could* in the future - making `teamHas` /`setHas` `DeepReadonly` in the sub calls would help with this, though maybe we want them to be able to mutate these

sim/team-validator.ts
```diff
@@ -200,14 +200,22 @@ export class TeamValidator {
 			this.ruleTable.minSourceGen[0] : 1;
 	}
 
-	validateTeam(team: PokemonSet[] | null, removeNicknames: boolean = false): string[] | null {
+	validateTeam(
+		team: PokemonSet[] | null,
+		removeNicknames = false,
```
scheibo: With 2 optionals its less obvious whether its necessary or not, but this could potentially be changed into a `{removeNicknames?: boolean, skipSets?: Record<string, Record<string, boolean>>}` option type if we saw more optionals being added. I think its OK for now, but would be happy to change it

scheibo:
The effort:benefit tradeoff doesnt appear to be worth it yet for me at the point

Zarel:
Incremental validation in sets would require a decent amount of work, but I think it would also be possible.

I think the biggest question is: How would you architect an incremental validator? I suppose you'd need to store validation state somewhere, but how would you keep track of which step you're on? Especially if the steps can be in any order? If I have an idea of how you would designed it, I'd have a better idea of how easy my current code would be to convert.

If I were to design an incremental validator myself: Validation state is currently stored in `setHas`, `teamHas`, and `setSources`. As far as I'm aware, we have no other validation state. I think we could stuff `setHas` inside `setSources`, and allow `setSources` to be passed into `validateSet` to incrementally validate parts.

After that, the main things I think are:

- species can't be validated incrementally; it must be validated first before the rest of the set can be validated
- `setSources` would need to track whether or not any required part (such as Ability, and at least one Move) were missing

In hindsight, I don't think there are major barriers to incremental/partial validation.

scheibo:
> Incremental validation in sets would require a **decent amount of work**, but I think it would also be possible.

Everything is possible! Isn't coding fun?

I don't disagree with your design for an incremental validator, I just don't have cycles to take it on. Realistically, I would be more likely to fork with this current code if its not going to be accepted than attempt to handle incremental validation at this time.

Zarel:
I know, in case it was unclear from my previous comments, I was volunteering.

sim/team-validator.ts
Zarel: What's a Record?

Zarel: I think this signature should be refactored to an options object.

scheibo: `Record<string, number>` => `{[key: string]: number}`

scheibo: > I think this signature should be refactored to an options object.

Done (but leaving this thread unresolved in case you were curious about `Record`. I switched away from it, I don't think it adds value here it was just easier to type originally)

Zarel: Okay, I understand what a Record is now, but `Record<string, boolean>` seems less readable than `{[k: string]: boolean}` to me.

Zarel:
I think it might make sense to put `state` in `options`, especially since I can think of plenty of use cases for e.g. `removeNickname` without passing `state`.

sim/team-validator.ts
scheibo: Agreed, which is why I switched back

Zarel: I think it's even less readable several levels deep, and an `interface` with proper indentation would be better in that case.

scheibo:
> I think it might make sense to put `state` in `options`, especially since I can think of plenty of use cases for e.g. `removeNickname` without passing `state`.

I agree. That method signature was just a first pass, if you continue reading I discovered it wasn't sufficient for the more generic use case of arbitrary incremental validation which doesnt occur in a fixed order.

Anyway, if incremental validation is something you're interested in working on in the near future then this PR doesnt need to be merged, but if its something you wouldnt plan on taking on immediately merging this `skipSets` solution (and pulling it out later when better support is added) would be helpful :)

Zarel:
I'm not sure how much structured errors would help, anyway. Basically every error boils down to "There are no known ways of legally obtaining a Pokémon with that combination of moves, ability, and IVs." – we have some hardcoded messages for specific situations, but for nearly any error, there are multiple possible ways to make changes so the set passes validation, and PS doesn't exactly have awareness of all of them (nor would it be easy to).

Zarel:
And yes, I plan to merge this before working on incremental validation.

Zarel:
(After you fix that merge conflict.)

Incidentally, I realized another problem for incremental/partial validation: We need to know level before moves. That shouldn't be a problem for your in-battle checker, but you might want to keep that in mind for other use-cases.

scheibo:
> (After you fix that merge conflict.)

Sigh. Done. (also, new lint is fucking slow! hopefully with things cached it shouldnt be, but damn)

> Incidentally, I realized another problem for incremental/partial validation: We need to know level before moves. That shouldn't be a problem for your in-battle checker, but you might want to keep that in mind for other use-cases.

Noted.

Zarel:
Okay, this explains a lot. You didn't update `team-validator-async`'s signature accordingly, when you made this change.

Zarel:
I added a new lint rule which caught that bug, and I was wondering why we'd never gotten reports that it stopped working, turns out, because it was only pushed three days ago.

scheibo:
> Okay, this explains a lot. You didn't update team-validator-async's signature accordingly, when you made this change.

Gah, sorry! Bad oversight on my part

Zarel:
Don't worry, it'll never happened again! Now that `process-manager` is TypeScripted, it supports generics, so there aren't any more `any`s in that part of the codebase for something like this to slip through.

PR #6418, opened 2020-03-05
Force timer if blitz mod is in effect

mia-pi-git:
This should work for both the format (when it's back) and tours (when the mod is applied). 
Let me know if there's a way to do this better, but this seemed optimal, since it gets `@@@blitz` too.
from: https://www.smogon.com/forums/threads/enforce-timer-when-blitz-is-added.3655991/

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -567,11 +567,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 			this.addPlayer(options.p4, options.p4team || '', options.p4rating);
 		}
 		this.timer = new RoomBattleTimer(this);
-		if (Config.forcetimer) {
-			this.timer.start();
-		} else if (this.format.includes('blitz')) {
-			this.timer.start();
-		}
+		if (Config.forcetimer || this.format.includes('blitz')) this.timer.start();
 		this.start();
 	}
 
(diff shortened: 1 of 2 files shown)


server/room-battle.ts
```diff
@@ -568,6 +568,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 		}
 		this.timer = new RoomBattleTimer(this);
 		if (Config.forcetimer) this.timer.start();
+		if (this.format.includes('blitz')) this.timer.start();
```
Zarel: Why isn't this an `else`?

Zarel: Heck, why isn't this an `||`?

mia-pi-git: because i'm dense!

mia-pi-git: that is... much better and i'll set it to that now, not sure why i didn't.. remember that exists

Zarel:
How much warning does Blitz give on the first turn? We should probably give significantly more than subsequent turns, just so people don't autolose because they didn't respond to the notification fast enough.

mia-pi-git:
from the blitz mod rules: 
`timer: {starting: 15, addPerTurn: 5, maxPerTurn: 15, maxFirstTurn: 30, grace: 30},`

> How much warning does Blitz give on the first turn? We should probably give significantly more than subsequent turns, just so people don't autolose because they didn't respond to the notification fast enough.
from the mod rules: 
`timer: {starting: 15, addPerTurn: 5, maxPerTurn: 15, maxFirstTurn: 30, grace: 30}`

PR #6420, opened 2020-03-05
Trivia: implement case-insensitive search

xfix:


Zarel:
Shouldn't case-insensitive be the default?

PR #6421, opened 2020-03-05
Fix more doubles tiers

KrisXV:


KrisXV:
Regarding Cutiefly and Vulpix-Alola:

The LC teambuilder uses the doubles tier table to display its Pokemon (this is done because less LC/NFE Pokemon are used in higher tiers in doubles); therefore, if they're displayed as LC Uber in doubles, they display as banned in the LC teambuilder, which is not true.

PR #6342, opened 2020-02-05
Migrate to typescript-eslint

TheJetOU:
TSLint has officially been deprecated and are only accepting
security patches currently.

So, it's time to move to typescript-eslint.

For the uninitiated, typescript-eslint is essentially an eslint
plugin that allows TypeScript files to be linted. This means that
our linters will be merged to one.

This PR is a rough draft of the migration. I was extremely liberal
in turning some rules off, as to not bloat up the diff and
prioritize getting it to a workable state.

I plan to create future PRs with fixes such as `prefer-optional-chain`,
`prefer-nullish-coalescing`, etc that I assume will be extremely
large diffs.

TODO:

- [x] Figure out the best solution to the hundreds of `max-el` warnings

- [x] Separate `.eslintrc` to `.eslintrc-syntax` and `.eslintrc-full` [0]

- [x] Make sure all the `tslint` rules are converted

[0] - typescript-eslint can lint with type information. It leverages
TypeScript to achieve this, as you can imagine in a large codebase
like this, it will be slow. To circumvent this, `.eslintrc-syntax`
will only contain rules that won't require type checking, and
`.eslintrc-full` will contain all of them. `npm run lint` will use
the former as its config and will be utilized in `npm run test`,
while `npm run fulllint` will use the latter as its config file
and will be used in `npm run fulltest`.

PR #6422, opened 2020-03-06
Use optional chaining and linter improvements

TheJetOU:
This is mostly just a follow up to #6342.

`prefer-optional-chaining` was turned on and fixed in every location it
complained in. The transformed function [0] looks expensive from a
glance but from skimming through the replaced sites it doesn't appear
to be ran in any important place, so it should be OK.

The linter improvements are:
- Increase linter performance
	- Make `full-lint` and `lint` write to different caches so we
	  avoid overwriting their caches since they're different configs
	- Change husky's hook to `npm run lint` so as to write to the
	  same cache
	- Remove `@typescript-eslint/eslint-plugin-tslint` which is
	  essentially a wrapper to TSLint because the rules aren't worth
	  running another linter
- Convert `.eslintrc.json` and `.eslintrc-syntax.json` to two spaces
  rather than four tabs to respect PS' `.editorconfig`
- Rename `fulllint` to `full-lint` to ease spelling it

[0] - https://pastie.io/mmtxpf.js (prettified)


server/ladders-remote.ts
```diff
@@ -26,7 +26,7 @@ export class LadderStore {
 	 * client.
 	 */
 	async getTop(prefix?: string): Promise<[string, string] | null> {
-		return new Promise(resolve => resolve(null));
+		return Promise.resolve(null);
```
TheJetOU: @Zarel, Removing `async` make `typescript-eslint` complain because of the `require-await` since it's an async function with no Promise or `await`. In our case this doesn't make much sense because it needs to conform to the `LadderStore` interface.

Nevertheless, I think this is a reasonable rule as it's usually indicative of a unintended outcome refactor, so turning it off completely might not be the best idea.

Of course, you can just turn of the rule for the next line here and explain why. If you prefer it that way I'll change it, I don't really care.

scheibo:
> Rename fulllint to full-lint to ease spelling it

I like this, can we do it for full-test as well (possibly in another PR)?

server/ladders-remote.ts
Zarel: I would 100% suppress it, either with an eslint-ignore directive or by removing the require-await rule.

Zarel: I actually think the require-await rule is mostly pointless and we should just remove it.

server/chat-plugins/datasearch.ts
```diff
@@ -146,7 +146,7 @@ export const commands: ChatCommands = {
 		return runSearch({
 			tar: targetsBuffer.join(","),
 			cmd: 'randmove',
-			canAll: (!this.broadcastMessage || (room && room.isPersonal)),
+			canAll: (!this.broadcastMessage || (room?.isPersonal)),
```
whalemer: Probably take out the extra brackets here

server/chat.ts
```diff
@@ -974,7 +974,7 @@ export class CommandContext extends MessageContext {
 			user.lastMessageTime = Date.now();
 		}
 
-		if (room && room.highTraffic &&
+		if (room?.highTraffic &&
 			toID(message).replace(/[^a-z]+/, '').length < 2
 			&& !user.can('broadcast', null, room)) {
```
whalemer: just while i'm here, what's the standard for where to put the &&?

server/chat-plugins/datasearch.ts
Zarel:
```suggestion
			canAll: (!this.broadcastMessage || room?.isPersonal),
```

server/chat.ts
Zarel: End of line.

Zarel: `.` is the only infix operator that should be at the beginning of a line.

Zarel: https://eslint.org/docs/rules/operator-linebreak

Oh, good, it's eslint-enforceable.

server/ladders-remote.ts
Zarel: > Of course, you can just turn off the rule for the next line here and explain why. If you prefer it that way I'll change it, I don't really care.

I basically _always_ prefer turning off the rule, to writing worse code because the rule forces you to. This has come up many times in the past, and my position is always the same: add an eslint-ignore directive.

TheJetOU: > I actually think the require-await rule is mostly pointless and we should just remove it.

> I think this is a reasonable rule as it's usually indicative of a unintended outcome refactor, so turning it off completely might not be the best idea.

Zarel: I feel like it's usually indicative of implementing an interface that requires it, but fine, we can leave it if you want. Just do the ignore.

sim/dex-data.ts
```diff
@@ -45,9 +47,11 @@ export class Tools {
 		} else if (text && text.roomid) {
 			text = text.roomid;
 		}
+		/* eslint-enable @typescript-eslint/prefer-optional-chain */
```
Zarel:
```suggestion
```

Zarel: (You put the real one in the right place at the end of this function.)

TheJetOU:
> > Rename fulllint to full-lint to ease spelling it
> 
> I like this, can we do it for full-test as well (possibly in another PR)?

Sure, consistency is nice.

PR #6429, opened 2020-03-07
Properly implement Sinistea and Polteageist formes

KrisXV:


KrisXV:
The egg groups should be Mineral and Amorphous instead of Undiscovered, but the Antique forme cannot be bred and i dont know where in the validator that goes

data/items.js
```diff
@@ -922,7 +922,7 @@ let BattleItems = {
 		},
 		num: 1254,
 		gen: 8,
-		desc: "Evolves Sinistea into Polteageist when used.",
+		desc: "Evolves Sinistea-Antique into Polteageist-Antique when used.",
```
Zarel: Does it not do both?

KrisXV: Nope, the items are forme specific

PR #6431, opened 2020-03-08
Fix Wi-Fi GAs

TheJetOU:


server/chat-commands/admin.ts
```diff
@@ -41,12 +41,12 @@ export const commands: ChatCommands = {
 		target = this.canHTML(target)!;
 		if (!target) return;
 		if (!this.can('addhtml', null, room)) return;
-		target = Chat.collapseLineBreaksHTML(target);
 		if (!user.can('addhtml')) {
 			target += Chat.html`
```
Zarel: never ever ever ever add linebreaks to a string just to shut up a linter

Zarel: This is a special case of my general rule: Never make your code worse to shut up a linter.

PR #6440, opened 2020-03-10
Clarify Punk Rock description

KrisXV:


data/abilities.js
```diff
@@ -2981,8 +2981,8 @@ let BattleAbilities = {
 		num: 227,
 	},
 	"punkrock": {
-		desc: "Boosts the power of sound-based moves. The Pokémon also takes half the damage from these kinds of moves.",
-		shortDesc: "Boosts sound move power, 0.5× damage from sound moves.",
+		desc: "This Pokemon's sound-based moves have their power multiplied by 1.3x. This Pokemon also receives 1/2 damage from sound-based attacks.",
+		shortDesc: "Sound moves: Self: 1.3x boost, recieve 1/2 damage.",
```
Marty-D:
```suggestion
		shortDesc: "This Pokemon receives 1/2 damage from sound moves. Its own have 1.3x power.",
```
Ability shortDescs have a lot of available space!

KrisXV: Oh for some reason I was restricting it in a move desc length my bad

Zarel:
I think that build failure was fixed in TI's latest push, so it should be fine once you fix the `shortDesc`.

PR #6441, opened 2020-03-11
Add plugin to view roomlogs in-sim

mia-pi-git:
ts is as-of-yet unfinished (as far as i know), JS is working at the moment, though.
currently looking like https://imgur.com/a/OiCb9Ie
3/12 edit: https://imgur.com/a/IXSx17p

### DIFF
--- a/server/chat-plugins/logviewer.ts
+++ b/server/chat-plugins/logviewer.ts
@@ -63,7 +63,7 @@ class RoomlogViewer<T extends RoomlogReader> {
 			timestamp = timestamp.trim();
 			if (cp.includes(protocol)) {
 				const [,, user, message] = line.split('|');
-				line = `[${timestamp}] <b>${user}</b>: ${message.replace('/log', '')}`;
+				line = `[${timestamp}] <b>${user}</b>: ${Chat.formatText(message.replace('/log', ''))}`;
 				cleaned.push(`&nbsp;${line}`);
 			} else if (protocol === 'html' || protocol === 'raw') {
 				line = `[${timestamp}]: ${rest}`;
--- a/server/chat-plugins/logviewer.ts
+++ b/server/chat-plugins/logviewer.ts
@@ -52,7 +52,11 @@ export const commands: ChatCommands = {
 	viewlogs(target, room, user) {
 		if (!room.chatRoomData || room.roomid.includes('-')) return this.errorReply("This room has no logs.");
 		target = target.trim();
-		if (target) room = Rooms.get(target) as ChatRoom | GameRoom;
+		if (target) { 
+			room = Rooms.get(target) as ChatRoom | GameRoom;
+		} else { 
+			target = room;
+		}
 		if (!Rooms.search(target)) return this.errorReply(`Room ${target} does not exist.`);
 		if (!this.can('mute', null, room) || !this.can('lock')) return false;
 		const today = Chat.toTimestamp(new Date()).split(' ')[0];
--- a/server/chat-plugins/logviewer.ts
+++ b/server/chat-plugins/logviewer.ts
@@ -2,7 +2,7 @@ import {FS} from '../../lib/fs';
 
 interface RoomlogReader {
 	readLogs: (roomid: RoomID, year: string, month: string, date: string) => string;
-	directory: (room: Room) => string;
+	months: (room: Room) => string;
 }
 class RoomlogReaderFS implements RoomlogReader {
 	readLogs(roomid: RoomID, year: string, month: string, date: string) {
@@ -14,21 +14,20 @@ class RoomlogReaderFS implements RoomlogReader {
 		}
 		return buf;
 	}
-	months(room:  Room) { 
+	months(room: Room) {
 		const months = [];
 		const days = [];
-		for (const m of FS(`logs/chat/${room}/`).readdirSync()) { 
+		for (const m of FS(`logs/chat/${room}/`).readdirSync()) {
 			if (!FS(`logs/chat/${room}/${m}/`).isDirectorySync()) continue;
 			months.push(m);
 		}
 		for (const month of months) {
-			for (const day of FS(`logs/chat/${room}/${month}`).readdirSync()) { 
+			for (const day of FS(`logs/chat/${room}/${month}`).readdirSync()) {
 				if (day.endsWith('.txt')) days.push(day.slice(0, -4));
 			}
 		}
 		return [months, days];
 	}
-
 }
 
 class RoomlogViewer<T extends RoomlogReader> {
@@ -44,7 +43,7 @@ class RoomlogViewer<T extends RoomlogReader> {
 		}
 	}
 	structure(room: Room) {
-		let buf = `<b>Logs for ${room}</b><br>`
+		let buf = `<b>Logs for ${room}</b><br>`;
 		const [months, days] = this.reader.months(room);
 		for (const month of months) {
 			buf += `<br><br><b>${month}</b><br>`;	
--- a/server/chat-plugins/logviewer.ts
+++ b/server/chat-plugins/logviewer.ts
@@ -46,7 +46,7 @@ class RoomlogViewer<T extends RoomlogReader> {
 		let buf = `<b>Logs for ${room}</b><br>`;
 		const [months, days] = this.reader.months(room);
 		for (const month of months) {
-			buf += `<br><br><b>${month}</b><br>`;	
+			buf += `<br><br><b>${month}</b><br>`;
 			for (const day of days) {
 				if (!day.includes(month)) continue;
 				buf += this.button(day, room);
--- a/server/chat-plugins/logviewer.ts
+++ b/server/chat-plugins/logviewer.ts
@@ -100,32 +100,36 @@ export const commands: ChatCommands = {
 export const pages: PageTable = {
 	logs(query, user, connection) {
 		let [, room, date, all] = this.pageid.split('--');
-		this.extractRoom();
 		if (!date) date = '';
-		room = Rooms.get(room) as ChatRoom | GameRoom;
+		this.room = Rooms.get(room) as ChatRoom | GameRoom; 
+		if (!this.room) {
+ 			return `<h2>Invalid Room</h2>`;
+			//extractRoom() fails on -- id's 
+			//manually do what this.extractRoom() would if there's no room
+		}
 		const [Y, M, D] = date.split('-');
 		this.title = `[Logs] [${room.title}] ${date}`;
 		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
 		let content;
 		if (D && M && Y) {
-			content = viewer.reader.readLogs(room as RoomID, Y, M, D);
+			content = viewer.reader.readLogs(this.room as RoomID, Y, M, D);
 		} else {
 			content = null;
 		}
-		let buf = `<b><h2>Logs for ${Y}-${M}-${D} on ${room.title}:</h2></b>`;
+		let buf = `<b><h2>Logs for ${Y}-${M}-${D} on ${this.room.title}:</h2></b>`;
 		if (!user.can('mute', null, room) || !user.can('lock')) return;
 		if (!content) {
-			buf = `<h2>All logs for ${room.title}. <br> ${viewer.structure(room)}</h2>`;
-			this.title = `[Logs] [${room.title}] Main Directory`;
+			buf = `<h2>All logs for ${this.room.title}. <br> ${viewer.structure(this.room)}</h2>`;
+			this.title = `[Logs] [${this.room.title}] Main Directory`;
 		} else {
 			if (all) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,107 @@
...
+		this.extractRoom();
+		this.title = `[Logs] ${datestring}`;
+		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
+		let content;
+		let buf = `<b><h2>Logs for ${datestring} on ${this.room}:</h2></b>`;
+		try {
+			if (parts[3] === 'today') {
+				this.title = `[Logs] Today`;
+				buf = `<b><h2>Logs for today on ${this.room}:</h2></b>`;
+				content = FS(`logs/chat/${this.room}/today.txt`).readSync();
```
Zarel: I would not take advantage of this symlink. It's easier just to manually calculate the current date.

Zarel:
```
let dateString = Chat.toTimestamp(new Date()).split(' ')[0];
```

mia-pi-git: that exists.
yea that’s better, will add and push with fixing build errors

mia-pi-git:
typescript is currently functioning with the rest of the plugin; not sure what else i need to do

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,96 @@
...
+		const today = Chat.toTimestamp(new Date()).split(' ')[0];
+		const msg = `${viewer.button(today, room)}${viewer.button(`Main Directory for ${room.title}`, room, true)}`;
+		return this.sendReplyBox(msg);
+	},
+	viewlogshelp: [`/viewlogs [room] - view logs of the specified room. Requires: % @ & ~`],
+};
+
+export const pages: PageTable = {
+	logs(query, user, connection) {
+		const parts = this.pageid.split('-');
```
Zarel: It'd probably be better to name the parts:

```suggestion
		const [_view, _logs, _roomid, Y, M, D, all] = this.pageid.split('-');
```

Zarel: This won't let you view the logs of a help room, though.

Zarel: Because the help room contains `-`.

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,96 @@
...
+		const datestring = `${parts[3]}-${parts[4]}-${parts[5]}`;
+		const monthstring = `${parts[3]}-${parts[4]}`;
+		this.extractRoom();
+		this.title = `[Logs] ${datestring}`;
+		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
+		const content = FS(`logs/chat/${this.room}/${monthstring}/${datestring}.txt`).readIfExistsSync();
```
Zarel: Don't rely on auto-casting; actually `this.room` shouldn't be looked at at all. Grab the roomid from the pageid.

server/chat-plugins/logviewer.ts
mia-pi-git: for a ticket; `ticket` or something could be attached as an identifier before `true` - i'll test that now

mia-pi-git: it might also be easier to identify the `roomid` as something like `ticket|miapi` then if it contains `|`, go to tickets

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,96 @@
...
+		}
+		if (!Rooms.search(target)) return this.errorReply(`Room ${target} does not exist.`);
+		if (!this.can('mute', null, room) || !this.can('lock')) return false;
+		const today = Chat.toTimestamp(new Date()).split(' ')[0];
+		const msg = `${viewer.button(today, room)}${viewer.button(`Main Directory for ${room.title}`, room, true)}`;
+		return this.sendReplyBox(msg);
+	},
+	viewlogshelp: [`/viewlogs [room] - view logs of the specified room. Requires: % @ & ~`],
+};
+
```
Zarel:
```suggestion

function splitPageId(pageid: string) {
	const dashIndex = pageid.lastIndexOf('-');
	const D = pageid.slice(dashIndex + 1);
	const dashIndex2 = pageid.lastIndexOf('-', dashIndex - 1);
	const M = pageid.slice(dashIndex2 + 1, dashIndex);
	const dashIndex3 = pageid.lastIndexOf('-', dashIndex2 - 1);
	const Y = pageid.slice(dashIndex3 + 1, dashIndex2);
	const START_LENGTH = 10; // 'view-logs-'.length
	if (dashIndex3 <= START_LENGTH) throw new Error(`invalid pageid ${pageid}`);
	const roomid = pageid.slice(START_LENGTH, dashIndex3);
	return [roomid, Y, M, D];
}

```

server/chat-plugins/logviewer.ts
Zarel:
```suggestion
		const [roomid, Y, M, D] = splitPageId(this.pageid);
```

server/chat-plugins/logviewer.ts
TheJetOU: You should probably abstract and move the reading logic (see below). That will enable use to easily migrate the viewer too whenever we support SQLite/Postgres databases.

```typescript
interface RoomlogReader {
    readLogs: (roomid: RoomID, month: string, date: string) => string;
    directory: (room: Room) => string
}
class RoomlogFSReader implements RoomlogReader {
    readLogs(roomid: RoomID, month: string, date: string) {}
    directory(room: Room) {}
}
class RoomlogViewer<T extends RoomlogReader> {
    reader: T;
    constructor(reader: T) {
        this.reader = reader;
    }
   // other functions
}

const viewer = new RoomlogViewer(new RoomlogFSReader());
```

TheJetOU:
Updating `Config.modlogLink` (https://github.com/smogon/pokemon-showdown/blob/master/server/chat-plugins/modlog.ts#L266) sounds like a good idea.

It appears to be in private code though, so you'll need to re-write it from scratch (doesn't look like a hard function to write). Making it public now sounds reasonable since it isn't main-only anymore.

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,138 @@
...
+		const days = FS(`logs/chat/${room}/${month}`).readdirSync();
+		for (const file of days) {
+			if (!file.endsWith('.txt')) continue;
+			buf += `${this.button(file.slice(0, -4), room)}`;
+		}
+
+		return buf;
+	}
+	
+	button(datestring: string, room: Room, nodate = false) {
```
TheJetOU: This has no relation to reading, so it should be in `RoomlogViewer`.

mia-pi-git: directory() needs it, so i shifted it because i wasn't sure if it was possible to call from there

TheJetOU: I think a better solution is to make `RoomlogReader#directory` return an array of roomids and then `RoomlogViewer#directory` will turn it to HTML, so then the button can be moved back to `RoomlogViewer`.

TheJetOU: Maybe month should have the same refactor done on it to have a clean `RoomlogReader` API (ie. not intermingled with HTML).

mia-pi-git: yea i'm working on that right now

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,136 @@
...
+		} else {
+			content = null;
+		}
+		let buf = `<b><h2>Logs for ${Y}-${M}-${D} on ${room}:</h2></b>`;
+		if (!user.can('mute', null, Rooms.get(room)) || !user.can('lock')) return;
+		if (!content) {
+			buf = `<h2>All logs for ${room}. <br> ${viewer.structure(room)}</h2>`;
+			this.title = `[Logs] Main Directory`;
+		} else if (!content && M && Y && !D) {
+			buf += viewer.reader.month(`${Y}-${M}`, room.roomid);
```
KrisXV:
```suggestion
			buf += viewer.reader.months(`${Y}-${M}`, room.roomid);
```

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,136 @@
...
+		let [, room, date, all] = this.pageid.split('--');
+		this.extractRoom();
+		if (!date) date = '';
+		if (!room) room = this.room.roomid;
+		const [Y, M, D] = date.split('-');
+		this.title = `[Logs] ${date}`;
+		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
+		let content;
+		if (D && M && Y) {
+			content = viewer.reader.readLogs(room, Y, M, D);
```
KrisXV:
```suggestion
			content = viewer.reader.readLogs(room as RoomID, Y, M, D);
```

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,132 @@
...
+		try {
+			buf = FS(`logs/chat/${roomid}/${year}-${month}/${year}-${month}-${date}.txt`).readIfExistsSync();
+		} catch (e) {
+			buf = '';
+		}
+		return buf;
+	}
+	months(room: Room) {
+		const months = [];
+		const days = [];
```
KrisXV:
```suggestion
		const months: string[] = [];
		const days: string[] = [];
```

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,132 @@
+import {FS} from '../../lib/fs';
+
+interface RoomlogReader {
+	readLogs: (roomid: RoomID, year: string, month: string, date: string) => string;
+	months: (room: Room) => any[];
```
KrisXV:
```suggestion
	months: (room: Room) => [string[], string[]];
```

Zarel:
Do you not have the ability to edit on your own computer and run `npm test`?

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,137 @@
...
+		}
+		return cleaned;
+	}
+}
+
+const viewer = new RoomlogViewer(new RoomlogReaderFS());
+
+export const commands: ChatCommands = {
+	viewlogs(target, room, user) {
+		if (!room.chatRoomData || room.roomid.includes('-')) return this.errorReply("This room has no logs.");
```
TheJetOU: Help tickets contain hyphens and they do have logs. TBH, I don't really see the use of this check, if the room doesn't have logs they can see that it doesn't in the HTML page.

server/chat-plugins/logviewer.ts
```diff
@@ -0,0 +1,137 @@
...
+		this.title = `[Logs] [${room.title}] ${date}`;
+		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
+		let content;
+		if (D && M && Y) {
+			content = viewer.reader.readLogs(this.room as RoomID, Y, M, D);
+		} else {
+			content = null;
+		}
+		let buf = `<b><h2>Logs for ${Y}-${M}-${D} on ${this.room.title}:</h2></b>`;
+		if (!user.can('mute', null, room) || !user.can('lock')) return;
```
TheJetOU: `PageHandler` has that function too, with the added bonus it returns a "Permission Denied." string.
```suggestion
		if (!this.can('mute', null, room)) return;
```

Nitpick: Move this line above the reader so we don't read 2MB files for no reason.

mia-pi-git:
rest are addressed and have fixes pushed for, but wrt this 
> The timestamp should ideally be sharable (ie. clicking the timestamp should change the URL and highlight the message) to easily share specific logs.
i worked this out but it didn't really seem to have that much gain to it - especially in longer rooms, the highlighted message takes a while to find and the url is clunky either way, so it seems as though one could just ctrl-f / link it as <<view-logs--room--date>> already.

TheJetOU:
> rest are addressed and have fixes pushed for

The latest screenshot you sent me didn't seem to have a back button.

> i worked this out but it didn't really seem to have that much gain to it - especially in longer rooms, > the highlighted message takes a while to find and the url is clunky either way, so it seems as though one could just ctrl-f / link it as <> already.

I should have specified that it should scroll to the highlighted message, but I'm pretty sure the only way to do it with pure HTML is to use `<a name="foo" href="#foo">bar</a>` which I'm assuming won't work in PS HTML pages, so you'd need to support JavaScript in HTML pages which is probably more suited for a followup rather than the initial PR.

mia-pi-git:
main directory (back) buttons, as well as buttons to either show or hide hidden protocol messages have been implemented at top and bottom. 
current shots: https://imgur.com/a/IXSx17p

mia-pi-git:
superseded by #6454

PR #6444, opened 2020-03-11
Attribute addhtml and code commands

whalemer:
Serverside half of https://github.com/smogon/pokemon-showdown-client/pull/1475

Requires the client change, because of the /uhtml fix.


Zarel:
Merged 9 days ago; I'm going to call this safe.

PR #6445, opened 2020-03-11
Block hotpatches while updating/rebuilding

whalemer:


server/chat-commands/admin.ts
```diff
@@ -889,10 +892,12 @@ export const commands: ChatCommands = {
 		if (!user.can('hotpatch')) {
 			return this.errorReply(`/updateserver - Access denied.`);
 		}
+		Monitor.updateServerLock = true;
 		const [, , stderr] = await exec('node ../../build');
 		if (stderr) {
 			return this.errorReply(`Crash while rebuilding: ${stderr}`);
 		}
+		Monitor.updateServerLock = false;
```
Zarel: Does this need to be after the error return?

whalemer: It's copying the current behaviour of updateserver (leaving the server locked if the update crashes), which sounds sensible to me.

Zarel:
`/rebuild` will remove the lock, but I think that's desirable?

PR #6447, opened 2020-03-12
SSB: Update martha's rank

KrisXV:


Zarel:
Should we like rewrite SSB to auto-update ranks and/or draw ranks from a private file, so we don't have to clutter the commit log with rank updates?

TheImmortal:
Just get it from ``Users.usergroups``. Side-servers wouldn't have it, which I don't think is a big issue. It can default to no rank.

PR #6462, opened 2020-03-16
Remove restrictions on concurrent battle searches

scheibo:
https://www.smogon.com/forums/threads/improve-searching-for-concurrent-ladder-battles.3660989/

### DIFF
--- a/server/ladders.ts
+++ b/server/ladders.ts
@@ -406,17 +406,6 @@ class Ladder extends LadderStore {
 			return;
 		}
 
-		const roomid = this.needsToMove(user);
-		if (roomid) {
-			connection.popup(`Error: You need to make a move in <<${roomid}>> before you can look for another battle.\n\n(This restriction doesn't apply in the first five turns of a battle.)`);
-			return;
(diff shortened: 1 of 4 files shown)


Zarel:
I wonder if there's a compromise solution here?

Zarel:
We talked about this in Staff.

The only thing I want to fix is the case of "player is losing, player starts a new game but leaves the old game open (which is bad manners), instead of just forfeiting".

Most other games straight-up don't allow concurrent games, which is their solution. I do agree that our current solution is too harsh; I just wish we had a better one.

Zarel:
For everyone at home watching: I _did_ propose removing restrictions completely, and I'm fine with it as a last resort, but I do wish there was something better than nothing we could do about it.

scheibo:
I don't really care to police bad manners nor do I have an elegant solution for doing so. Its 3 minutes and the person on the other end can open a new game themselves and be notified if their rude original opponent comes back.

Zarel:
I guess that's valid enough.

PR #6463, opened 2020-03-16
Add Dex#getOutOfBattleSpecies

KrisXV:


tools/set-import/importer.ts
```diff
@@ -359,7 +359,7 @@ function toPokemonSet(dex: ModdedDex, format: Format, pokemon: string, set: Deep
 	const template = dex.getTemplate(pokemon);
 	const mega = ['Mega', 'Primal', 'Ultra'].some(f => template.forme.startsWith(f));
 	if (template.battleOnly || (mega && !format.id.includes('balancedhackmons'))) {
-		copy.species = template.inheritsFrom || template.baseSpecies;
```
scheibo: Hmm, I supposed this was a bug then before. Thanks for fixing!

PR #6466, opened 2020-03-17
Refactor learnsets.js

KrisXV:
I'm aware some of the tests are failing, but I'm struggling to figure out how to fix them right now. Feel free to let me know.


Zarel:
I think the error messages are pretty clear? The variables aren't used at all, so you can just remove the line that declares them and it won't change what the code does.

data/formats-data.js
```diff
@@ -3,14 +3,6 @@
...
 	bulbasaur: {
-		eventPokemon: [
-			{"generation": 3, "level": 70, "moves": ["sweetscent", "growth", "solarbeam", "synthesis"], "pokeball": "pokeball"},
-			{"generation": 3, "level": 10, "gender": "M", "moves": ["tackle", "growl", "leechseed", "vinewhip"], "pokeball": "pokeball"},
-			{"generation": 5, "level": 10, "gender": "M", "isHidden": true, "moves": ["tackle", "growl", "leechseed", "vinewhip"]},
-			{"generation": 5, "level": 1, "shiny": 1, "ivs": {"def": 31}, "isHidden": false, "moves": ["falseswipe", "block", "frenzyplant", "weatherball"], "pokeball": "pokeball"},
-			{"generation": 6, "level": 5, "isHidden": false, "moves": ["growl", "leechseed", "vinewhip", "poisonpowder"], "pokeball": "cherishball"},
-			{"generation": 6, "level": 5, "isHidden": true, "moves": ["tackle", "growl", "celebrate"], "pokeball": "cherishball"},
-		],
 		encounters: [
```
Zarel: `encounters` definitely belongs in learnsets rather than formats-data; it's essentially "levels at which you can get them in the wild".

data/formats-data.js
```diff
@@ -21,9 +13,6 @@ let BattleFormatsData = {
 	},
 	venusaur: {
 		randomBattleMoves: ["gigadrain", "leechseed", "sleeppowder", "sludgebomb", "substitute"],
-		eventPokemon: [
-			{"generation": 6, "level": 100, "isHidden": true, "moves": ["solarbeam", "frenzyplant", "synthesis", "grasspledge"], "pokeball": "cherishball"},
-		],
 		tier: "UU",
```
Zarel: In the future, I want `tier` to be an array of tiers, instead of a single tier, so we don't have the LC Uber issue.

Zarel: (But that's out of scope for this PR.)

data/formats-data.js
```diff
@@ -115,15 +71,15 @@ let BattleFormatsData = {
 	},
 	caterpie: {
 		encounters: [
-			{"generation": 1, "level": 3, "shiny": false},
+			{"generation": 1, "level": 3},
```
Zarel: ?

I forget, are events non-shiny by default, and `"shiny": false` means nothing?

Zarel: Looks like yes. I'll update the documentation.

KrisXV: Sorry for late reply; yeah, it's either `true` or `1`, `false` means never

KrisXV:
@Zarel those weren't the errors I was referring to, some of the tests are failing, specifically regarding rockruff and some event validation. but it might just be on my end if they're not showing up here? It's kind of weird that GH Actions prioritized the warnings over the errors

sim/team-validator.ts
```diff
@@ -926,11 +930,12 @@ export class TeamValidator {
 	) {
 		let eventData: EventInfo | null = null;
 		let eventTemplate = template;
+		const eventLsetData = this.dex.getLearnsetData(eventTemplate.speciesid);
```
Zarel: This is the wrong place for this, `eventTemplate` changes on line 937; that's one of your bugs.

PR #6467, opened 2020-03-18
Fix formatting links for /randbats

mia-pi-git:
currently takes you to play.pokemonshowdown.com/dex.pokemonshowdown.com, which is obviously not right.


Zarel:
Does `Config.routes.dex` not include `https://`? Should it be changed there rather than here?

Zarel:
Looks like routes just never include `https://`.

PR #6471, opened 2020-03-19
Support /hidetext [line count]

tmagicturtle:
This patch (coupled with its accompanying client-side patch) amends the /hidetext command to accept an integer that represents the number of lines that should be hidden.

Sometimes not all lines need to be cleared and several staff (room and global) have suggested that the option to hide just some lines would be useful.

### DIFF
--- a/server/chat-commands/moderation.js
+++ b/server/chat-commands/moderation.js
@@ -1413,7 +1413,7 @@ exports.commands = {
 				userid,
 				...Object.keys(targetUser.prevNames),
 				...targetUser.getAltUsers(true).map(user => user.getLastId()),
-			], lineCount);
+			]);
 		} else {
 			room.send(`|c|~|${name}'s messages were cleared from ${room.title} by ${user.name}.`);
 			this.modlog('HIDETEXT', targetUser || userid, null, {noip: 1, noalts: 1});
--- a/server/chat-commands/moderation.js
+++ b/server/chat-commands/moderation.js
@@ -1393,7 +1393,7 @@ exports.commands = {
 	hidetext(target, room, user, connection, cmd) {
 		if (!target) return this.parse(`/help hidetext`);
 
-		this.splitTarget(target);
+		const lineCount = parseInt(this.splitTarget(target)) || 0;
 		let targetUser = this.targetUser;
 		let name = this.targetUsername;
 		if (!targetUser && !room.log.hasUsername(name)) {
--- a/server/chat-commands/moderation.js
+++ b/server/chat-commands/moderation.js
@@ -1415,7 +1415,11 @@ exports.commands = {
 				...targetUser.getAltUsers(true).map(user => user.getLastId()),
 			]);
 		} else {
-			room.send(`|c|~|${name}'s messages were cleared from ${room.title} by ${user.name}.`);
+			if (lineCount > 0) {
+				room.send(`|c|~|${lineCount} of ${name}'s messages were cleared from ${room.title} by ${user.name}.`);
+			} else {
+				room.send(`|c|~|${name}'s messages were cleared from ${room.title} by ${user.name}.`);
+			}
 			this.modlog('HIDETEXT', targetUser || userid, null, {noip: 1, noalts: 1});
 			room.hideText([userid], lineCount);
 		}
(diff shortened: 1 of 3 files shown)


server/rooms.ts
```diff
@@ -1159,10 +1159,10 @@ export class BasicChatRoom extends BasicRoom {
 		this.log.modlog(message);
 		return this;
 	}
-	hideText(userids: ID[]) {
+	hideText(userids: ID[], lineCount = 0) {
 		const cleared = this.log.clearText(userids);
```
whalemer: This should probably also be modified.

tmagicturtle: In what way? To log the amount of lines cleared?

whalemer: Sorry, I meant in the Roomlog#clearText function (to only remove the first n lines from the scrollback log)

tmagicturtle: Aha, I see what the function does now. Will be working on that, thanks for the heads up.

tmagicturtle:
Working on roomlogs.ts now, making decent progress

server/chat-commands/moderation.js
```diff
@@ -1396,7 +1396,9 @@ exports.commands = {
 		this.splitTarget(target);
```
Zarel: Grab the line count from here.

```suggestion
		const lineCount = parseInt(this.splitTarget(target)) || 0;
```

server/roomlogs.ts
```diff
@@ -178,22 +178,29 @@ export class Roomlog {
...
+					if (lineCount > 0) {
+						lineCount--;
+						return false;
+					}
+					return true;
 				}
 			}
 			return true;
 		});
+		this.log = this.log.reverse();
```
Zarel:
```suggestion
		}).reverse();
```

server/chat-commands/moderation.js
```diff
@@ -1396,7 +1396,9 @@ exports.commands = {
...
 		let targetUser = this.targetUser;
 		let name = this.targetUsername;
-		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${target} not found or has no roomlogs.`);
+		let lineCount = 0;
+		if (target.includes(',')) lineCount = Math.abs(parseInt(target.split(',')[1]));
```
Zarel:
```suggestion
```

server/chat-commands/moderation.js
```diff
@@ -1396,7 +1396,9 @@ exports.commands = {
...
+		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${name} not found or has no roomlogs.`);
```
Zarel:
```suggestion
		if (!targetUser && !room.log.hasUsername(target)) {
			return this.errorReply(`User ${name} not found or has no roomlogs.`);
		}
```

server/chat-commands/moderation.js
```diff
@@ -1411,11 +1413,11 @@ exports.commands = {
 				userid,
 				...Object.keys(targetUser.prevNames),
 				...targetUser.getAltUsers(true).map(user => user.getLastId()),
-			]);
+			], lineCount);
```
whalemer: imo this should probably not work with hidealtstext.

tmagicturtle: That's fair.

server/chat-commands/moderation.js
```diff
@@ -1415,7 +1417,7 @@ exports.commands = {
 		} else {
 			room.send(`|c|~|${name}'s messages were cleared from ${room.title} by ${user.name}.`);
```
whalemer: hyper nit - maybe this should say if they're only clearing the last n lines?

tmagicturtle: Sure, that's easy enough.

server/chat-commands/moderation.js
```diff
@@ -1393,10 +1393,12 @@ exports.commands = {
...
 		if (!target) return this.parse(`/help hidetext`);
 
-		this.splitTarget(target);
+		const lineCount = parseInt(this.splitTarget(target)) || 0;
 		let targetUser = this.targetUser;
 		let name = this.targetUsername;
-		if (!targetUser && !room.log.hasUsername(target)) return this.errorReply(`User ${target} not found or has no roomlogs.`);
+		if (!targetUser && !room.log.hasUsername(name)) {
+			return this.errorReply(`User ${name} not found or has no roomlogs.`);
+		}
```
Zarel:
```suggestion
		}
		if (lineCount && cmd.includes('alt')) {
			return this.errorReply(`You can't specify a line count when using /hidealtstext.`);
		}
```

Silent failure is bad! Add an error message!

Zarel:
I haven't heard anything from Ashe so I guess I'm merging now.

PR #6472, opened 2020-03-20
Doubles OU: Add suspect notice

KrisXV:
https://www.smogon.com/forums/threads/np-ss-dou-stage-1-when-i-grow-up-dynamax-suspect.3661422/

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -280,6 +280,9 @@ let Formats = [
 		gameType: 'doubles',
 		ruleset: ['Standard Doubles'],
 		banlist: ['DUber', 'Beat Up'],
+		onBegin() {
+			if (this.rated && this.format.id === 'gen8doublesou') this.add('html', `<div class="broadcast-red"><strong>Doubles OU is currently suspecting Dynamax! For information on how to participate check out the <a href="https://www.smogon.com/forums/threads/3661422/">suspect thread</a>.</strong></div>`);
+		},
 	},
 	{
 		name: "[Gen 8] Doubles Ubers",


config/formats.js
```diff
@@ -280,6 +280,9 @@ let Formats = [
 		gameType: 'doubles',
 		ruleset: ['Standard Doubles'],
 		banlist: ['DUber', 'Beat Up'],
+		onBegin() {
+			if (this.rated && this.format.id === 'gen8doublesou') this.add('html', `<div class="broadcast-red"><strong>Doubles OU is currently suspecting Dynamax! For information on how to participate check out the <a href="https://www.smogon.com/forums/threads/3661422/">suspect thread</a>.</strong></div>`);
```
Zarel:
```suggestion
			if (this.rated && this.format.id === 'gen8doublesou') {
				this.add('html', `<div class="broadcast-red"><strong>Doubles OU is currently suspecting Dynamax! For information on how to participate check out the <a href="https://www.smogon.com/forums/threads/3661422/">suspect thread</a>.</strong></div>`);
			}
```

Does the linter not check formats? This should be over line length limit...

TheImmortal: Apparently not because this line is pretty much copy/pasted for every suspect. There's a couple in formats currently in fact, so I'm just gonna merge the original for consistency.

Zarel: Just fix them all :[

KrisXV: I already told you zarel, the linter doesnt check line length for JS files, so there's no need for this to be fixed

Zarel: The linter doesn't check line length _yet_, I want to turn it on eventually, and the more of this sort of thing we can fix ahead of time, the better.

PR #6473, opened 2020-03-20
Add Snow Cloak to gen4ou bans

Alumn2000:
Bans for Mamoswine familly. Glaceon and Frostlass have only 1 ability.
[https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8395762](url)

### DIFF
--- a/config/formats.js
+++ b/config/formats.js
@@ -1213,7 +1213,7 @@ let Formats = [
 
 		mod: 'gen4',
 		ruleset: ['Standard', 'Baton Pass Clause'],
-		banlist: ['Uber', 'Sand Veil', 'Soul Dew'],
+		banlist: ['Uber', 'Mamoswine ++ Snow Cloak', 'Piloswine ++ Snow Cloak', 'Sand Veil', 'Soul Dew', 'Swinub ++ Snow Cloak'],
 	},
 	{
 		name: "[Gen 3] OU",
--- a/config/formats.js
+++ b/config/formats.js
@@ -1213,7 +1213,7 @@ let Formats = [
 
 		mod: 'gen4',
 		ruleset: ['Standard', 'Baton Pass Clause'],
-		banlist: ['Uber', 'Mamoswine ++ Snow Cloak', 'Piloswine ++ Snow Cloak', 'Sand Veil', 'Soul Dew', 'Swinub ++ Snow Cloak'],
+		banlist: ['Uber', 'Sand Veil', 'Soul Dew', 'Swinub ++ Snow Cloak', 'Piloswine ++ Snow Cloak', 'Mamoswine ++ Snow Cloak'],
 	},
 	{
 		name: "[Gen 3] OU",


Zarel:
@Alumn2000 that's not how you link to threads. If you're just linking the URL, you don't need to do anything special:

`https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8395762`

what you did was this:

`[https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8395762](url)`

which makes it a link to `https://github.com/smogon/pokemon-showdown/pull/url`

If you wanted to put text in a link, it should go like this:

`[click here](https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8395762)`

config/formats.js
```diff
@@ -1213,7 +1213,7 @@ let Formats = [
 
 		mod: 'gen4',
 		ruleset: ['Standard', 'Baton Pass Clause'],
-		banlist: ['Uber', 'Sand Veil', 'Soul Dew'],
+		banlist: ['Uber', 'Sand Veil', 'Soul Dew', 'Swinub ++ Snow Cloak', 'Piloswine ++ Snow Cloak', 'Mamoswine ++ Snow Cloak'],
```
urkerab, line 1216: Should these have been one `+` instead of two? As it stands, this bans any of those three users if there is another Snow Cloak user on the team.

Zarel: A lot of people seem to mix up `+` and `++`. We should probably replace `++` with something clearer?

Zarel: Especially since the current syntax seems to imply you can mix `+` and `++` when you really can't.

Zarel: I wonder if a keyword meaning "in the same team" would help

`Drizzle + Swift Swim in the same team`

But some way to make that more concise...

Zarel: But who knows, maybe it _should_ be that long? It has very few uses, mostly just a Gen 5 complex ban.

TheImmortal:
Replace + with @. “Mamoswine @ Snow Cloak” for example makes perfect sense to me. And then probably change ++ to +.

config/formats.js
TheImmortal: Replace + with @. “Mamoswine @ Snow Cloak” for example makes perfect sense to me. And then probably change ++ to +.

Zarel: "Shell Smash @ Baton Pass" doesn't really make sense to me, though.

KrisXV: why not replace `++` with `*`? i dont have any reasoning that makes sense, but its another mathematical method

Zarel: `Drizzle + Swift Swim in the same team` is my preference, because the point is that it's very rarely used and would benefit from the additional clarity.

KrisXV: I'm not too bothered by it I guess

TheImmortal: What about ``&`` instead of ``++``?

Zarel: I don't think that'll fix the "get people to realize they should be using `+` and not `&`" problem.

PR #6474, opened 2020-03-20
Typescript TCG & Tabletop's plugin

mia-pi-git:
i'm a bit concerned about the `search: any` specification on line 16 but that was the only way to fix a specific error (and two deriving from it)  - lmk if there's a better way to handle that.
Passes tests, though, but was advised against `any` unless there was nothing better (not sure if there is or not, since Object and AnyObject also disagree with it).

### DIFF
--- a/server/chat-plugins/tcgtabletop.ts
+++ b/server/chat-plugins/tcgtabletop.ts
@@ -13,7 +13,7 @@ import * as querystring from 'querystring';
 const SEARCH_PATH = '/api/v1/Search/List/';
 const DETAILS_PATH = '/api/v1/Articles/Details/';
 
-async function getFandom(site: string, pathName: string, search: string) {
+async function getFandom(site: string, pathName: string, search: any) {
 	const reqOpts = {
 		hostname: `${site}.fandom.com`,
 		method: 'GET',
@@ -26,10 +26,10 @@ async function getFandom(site: string, pathName: string, search: string) {
 	const body: any = await new Promise((resolve, reject) => {
 		https.request(reqOpts, res => {
 			if (!res.statusCode || !(res.statusCode >= 200 && res.statusCode < 300)) return reject(new Error(`Not found.`));
-			const body: string[] = [];
+			const data: string[] = [];
 			res.setEncoding('utf8');
-			res.on('data', chunk => body.push(chunk));
-			res.on('end', () => resolve(body.join('')));
+			res.on('data', chunk => data.push(chunk));
+			res.on('end', () => resolve(data.join('')));
 		}).on('error', reject).setTimeout(5000).end();
 	});
 
@@ -39,14 +39,14 @@ async function getFandom(site: string, pathName: string, search: string) {
 	return json;
 }
 
-async function searchFandom(site: string, query: Object) {
+async function searchFandom(site: string, query: any) {
 	const result = await getFandom(site, SEARCH_PATH, {query, limit: 1});
 	if (!Array.isArray(result.items) || !result.items.length) throw new Error(`Malformed data`);
 	if (!result.items[0] || typeof result.items[0] !== 'object') throw new Error(`Malformed data`);
 	return result.items[0];
 }
 
-async function getCardDetails(site: string, id: Object) {
+async function getCardDetails(site: string, id: string) {
 	const specifications = {
 		ids: id,
 		abstract: 0,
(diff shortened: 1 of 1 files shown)


Zarel:
What's wrong with `AnyObject`?

mia-pi-git:
> What's wrong with `AnyObject`?

Last time I attempted it, it threw tsc errors. However, I seem to have fixed whatever was causing the `cannot be used as an index type` error, as it's passing checks now. Fixing one other thing that's running off, then I'll commit.

server/chat-plugins/tcgtabletop.ts
```diff
@@ -38,39 +39,46 @@ async function getFandom(site, pathName, search) {
...
 		if (!this.canBroadcast()) return;
 		if (room.roomid !== 'tcgtabletop') return this.errorReply("This command can only be used in the TCG & Tabletop room.");
-		let subdomain = 'yugioh';
-		let query = target.trim();
+		const subdomain = 'yugioh';
+		const query = target.trim();
 		if (!query) return this.parse('/help yugioh');
 
-		return searchFandom(subdomain, query).then(data => {
+		return searchFandom(subdomain, query).then((data: { url: any, title: any, id: any}) => {
```
Zarel: I'm surprised this isn't a style error, but there should be no space in front of `url`.

Also, these can all be `unknown`.

```suggestion
		return searchFandom(subdomain, query).then((data: {url: unknown, title: unknown, id: unknown}) => {
```

server/chat-plugins/tcgtabletop.ts
```diff
@@ -38,39 +39,46 @@ async function getFandom(site, pathName, search) {
...
-			let entryTitle = Dex.getString(data.title);
-			let id = Dex.getString(data.id);
+			const entryUrl = Dex.getString(data.url);
+			const entryTitle = Dex.getString(data.title);
+			const id = Dex.getString(data.id);
 			let htmlReply = Chat.html`<strong>Best result for ${query}:</strong><br /><a href="${entryUrl}">${entryTitle}</a>`;
 			if (id) {
-				getCardDetails(subdomain, id).then(card => {
-					let thumb = Dex.getString(card.thumbnail);
+				getCardDetails(subdomain, id).then((card: { thumbnail: any }) => {
```
Zarel:
```suggestion
				getCardDetails(subdomain, id).then((card: {thumbnail: unknown}) => {
```

server/chat-plugins/tcgtabletop.ts
```diff
@@ -1,34 +1,35 @@
...
 const SEARCH_PATH = '/api/v1/Search/List/';
 const DETAILS_PATH = '/api/v1/Articles/Details/';
 
-async function getFandom(site, pathName, search) {
+async function getFandom(site: string, pathName: string, search: AnyObject) {
 	const reqOpts = {
 		hostname: `${site}.fandom.com`,
 		method: 'GET',
-		path: `${pathName}?${querystring.stringify(search)}`,
+		path: pathName + '?' + querystring.stringify(search),
```
Zarel:
```suggestion
		path: `${pathName}?${querystring.stringify(search)}`,
```

Can you revert this? I think I prefer how it was before.

Zarel:
@Asheviere tagging Astrid in case she has some opinions. This looks good besides my last suggestion.

server/chat-plugins/tcgtabletop.ts
```diff
@@ -84,7 +92,7 @@ exports.commands = {
 				if (!this.broadcasting) return this.sendReply(`|raw|<div class="infobox">${htmlReply}</div>`);
 				room.addRaw(`<div class="infobox">${htmlReply}</div>`).update();
 			}
-		}, err => {
+		}, (err: AnyObject) => {
```
Zarel:
```suggestion
		}, (err: Error) => {
```

Does this work?

mia-pi-git: > does this work

Looks like no: 
> Property 'code' does not exist on type 'Error'.

Zarel: `NodeJS.Error`?

Zarel: looks like also no.

Zarel:
```suggestion
		}, (err: Error & {code: number}) => {
```

Zarel:
I've heard no response, and this seems good, so I guess I'm merging this.

PR #6477, opened 2020-03-21
Update Custom Rules documentation and implementation

KrisXV:


TheImmortal:
"No NFE Clause"... what. That is very inconsistent. Literally none of the clauses have "No" in front of them. It's not called "No OHKO Clause" but you're not allowed to bring OHKO moves. Strong objection here, change it back.

TheImmortal:
Not Fully Evolved is fine tho.

KrisXV:
OK did

PR #6482, opened 2020-03-22
Fix crash

KrisXV:


TheImmortal:
This is not the fix. It results in Struggle Pokemon.

PR #6484, opened 2020-03-23
Typescript chat-commands/core

mia-pi-git:
draft because there's an issue with /cmd that seems rather pertinent.


server/chat-commands/core.ts
```diff
@@ -1324,28 +1357,30 @@ exports.commands = {
...
 				}
 				if (targetRoom.auth && targetUser.id in targetRoom.auth) {
-					roomid = targetRoom.auth[targetUser.id] + roomid;
+					(roomid as string) = targetRoom.auth[targetUser.id] + roomid;
 				}
 				roomList[roomid] = roomData;
 			}
 			if (!targetUser.connected) roomList = false;
-			let userdetails = {
+			const userdetails: Record<string, any> = {
```
Zarel:
```suggestion
			const userdetails: AnyObject = {
```

I don't know who taught you `Record`, but PS writes this as `{[k: string]: any}`. Or, we would, but we already have a type alias for that: `AnyObject`.

server/chat-commands/core.ts
```diff
@@ -1324,28 +1357,30 @@ exports.commands = {
...
 				}
 				if (targetRoom.battle) {
-					let battle = targetRoom.battle;
+					const battle = targetRoom.battle;
 					roomData.p1 = battle.p1 ? ' ' + battle.p1.name : '';
 					roomData.p2 = battle.p2 ? ' ' + battle.p2.name : '';
 				}
 				if (targetRoom.auth && targetUser.id in targetRoom.auth) {
-					roomid = targetRoom.auth[targetUser.id] + roomid;
+					(roomid as string) = targetRoom.auth[targetUser.id] + roomid;
```
Zarel:
```suggestion
					roomid = (targetRoom.auth[targetUser.id] + roomid) as RoomID;
```

server/chat-commands/core.ts
mia-pi-git: oh, strange, npm test was complaining to use that

server/chat-commands/core.ts
```diff
@@ -349,29 +348,30 @@ exports.commands = {
...
+			const availableRoom = targetRoom?.checkModjoin(user);
 			if (targetRoom && availableRoom) return this.parse(`/roomauth1 ${target}`);
 			return this.parse(`/userauth ${target}`);
 		}
 		/** @type {{[k: string]: string[]}} */
-		let rankLists = {};
-		let ranks = Object.keys(Config.groups);
-		for (let u in Users.usergroups) {
-			let rank = Users.usergroups[u].charAt(0);
+		const rankLists: AnyObject = {};
```
Zarel: Did you miss the type annotation above this?

```suggestion
		const rankLists: {[k: string]: string[]} = {};
```

server/chat-commands/core.ts
```diff
@@ -349,29 +348,30 @@ exports.commands = {
...
 	authority(target, room, user, connection) {
 		if (target) {
-			let targetRoom = Rooms.search(target);
-			let availableRoom = targetRoom && targetRoom.checkModjoin(user);
+			const targetRoom = Rooms.search(target);
+			const availableRoom = targetRoom?.checkModjoin(user);
 			if (targetRoom && availableRoom) return this.parse(`/roomauth1 ${target}`);
 			return this.parse(`/userauth ${target}`);
 		}
 		/** @type {{[k: string]: string[]}} */
```
Zarel:
```suggestion
```

mia-pi-git: oops, i can't see apparently - will batch this in a sec

server/chat-commands/core.ts
```diff
@@ -349,29 +348,30 @@ exports.commands = {
...
+		const ranks = Object.keys(Config.groups);
+		for (const u in Users.usergroups) {
+			const rank = Users.usergroups[u].charAt(0);
 			if (rank === ' ' || rank === '+') continue;
 			// In case the usergroups.csv file is not proper, we check for the server ranks.
 			if (ranks.includes(rank)) {
-				let name = Users.usergroups[u].substr(1);
-				if (!rankLists[rank]) rankLists[rank] = [];
+				const name: string = Users.usergroups[u].substr(1);
+				const place: string[] = rankLists[rank];
```
Zarel:
```suggestion
				const place = rankLists[rank];
```

Your mistake was using `AnyObject` earlier.

server/chat-commands/core.ts
```diff
@@ -349,29 +348,30 @@ exports.commands = {
...
+		const rankLists: AnyObject = {};
+		const ranks = Object.keys(Config.groups);
+		for (const u in Users.usergroups) {
+			const rank = Users.usergroups[u].charAt(0);
 			if (rank === ' ' || rank === '+') continue;
 			// In case the usergroups.csv file is not proper, we check for the server ranks.
 			if (ranks.includes(rank)) {
-				let name = Users.usergroups[u].substr(1);
-				if (!rankLists[rank]) rankLists[rank] = [];
+				const name: string = Users.usergroups[u].substr(1);
```
Zarel:
```suggestion
				const name = Users.usergroups[u].substr(1);
```

TS has type inference for this.

server/chat-commands/core.ts
```diff
@@ -349,29 +348,30 @@ exports.commands = {
...
 				if (name) rankLists[rank].push(name);
 			}
 		}
 
-		let buffer = Object.keys(rankLists).sort(
+		const buffer = Object.keys(rankLists).sort(
 			(a, b) => (Config.groups[b] || {rank: 0}).rank - (Config.groups[a] || {rank: 0}).rank
 		).map(
-			r => `${(Config.groups[r] ? `**${Config.groups[r].name}s** (${r})` : r)}:\n${rankLists[r].sort((a, b) => toID(a).localeCompare(toID(b))).join(", ")}`
+			r => `${(Config.groups[r] ? `**${Config.groups[r].name}s** (${r})` : r)}:\n${rankLists[r].sort((a: string, b: string) => toID(a).localeCompare(toID(b))).join(", ")}`
```
Zarel:
```suggestion
			r => `${(Config.groups[r] ? `**${Config.groups[r].name}s** (${r})` : r)}:\n${rankLists[r].sort((a, b) => toID(a).localeCompare(toID(b))).join(", ")}`
```

TS has type inference for this. If you need to add types to an arrow function, you're doing something wrong (in this case, the wrong thing you're doing is failing to correctly type `rankLists`.

server/chat-commands/core.ts
```diff
@@ -384,10 +384,10 @@ exports.commands = {
 	],
 
 	userlist(target, room, user) {
-		let userList = [];
+		const userList = [];
 
-		for (let i in room.users) {
-			let curUser = Users.get(room.users[i]);
+		for (const i in room.users) {
+			const curUser = Users.get(room.users[i]);
```
Zarel:
```suggestion
		for (const id in room.users) {
			const curUser = Users.get(room.users[id]);
```

server/chat-commands/core.ts
```diff
@@ -549,7 +551,7 @@ exports.commands = {
 			return;
 		} else {
 			this.pmTarget = targetUser;
-			this.room = null;
+			(this.room as ChatRoom | GameRoom | null) = null;
```
Zarel:
```suggestion
			this.room = null!;
```

server/chat-commands/core.ts
```diff
@@ -640,37 +645,45 @@ exports.commands = {
...
+		}
 		target = this.statusfilter(target);
 		if (!target) return this.errorReply(this.tr("Your status contains a banned word."));
 
 		user.setUserMessage(target);
 		this.sendReply(this.tr `Your status has been set to: ${target}.`);
 	},
-	statushelp: [`/status [note] - Sets a short note as your status, visible when users click your username. Use /clearstatus to clear your status message.`],
+	statushelp: [`/status [note] - Sets a short note as your status, visible when users click your username.`,
+	 `Use /clearstatus to clear your status message.`],
```
Zarel:
```suggestion
	statushelp: [
		`/status [note] - Sets a short note as your status, visible when users click your username.`,
		 `Use /clearstatus to clear your status message.`,
	],
```

server/chat-commands/core.ts
```diff
@@ -640,37 +645,45 @@ exports.commands = {
...
+			this.errorReply("Setting status messages in /busy is no longer supported. Set a status using /status.");
+		}
 		user.setStatusType('busy');
 		this.parse('/blockpms +');
 		this.parse('/blockchallenges');
 		this.sendReply(this.tr("You are now marked as busy."));
 	},
-	busyhelp: [`/busy - Marks you as busy, blocking private messages and challenges. Use /back to mark yourself as back.`],
+	busyhelp: [`/busy - Marks you as busy, blocking private messages and challenges.`,
+	 `Use /back to mark yourself as back.`],
```
Zarel:
```suggestion
	busyhelp: [
		`/busy - Marks you as busy, blocking private messages and challenges.`,
	 	`Use /back to mark yourself as back.`,
	 ],
```

server/chat-commands/core.ts
```diff
@@ -728,15 +741,16 @@ exports.commands = {
...
 			buffer += Chat.html`<tr><td colspan="8">User: <strong>${target}</strong></td></tr>`;
 
-			let ratings = values.join(``);
+			const ratings = values.join(``);
 			if (!ratings) {
 				buffer += `<tr><td colspan="8"><em>This user has not played any ladder games yet.</em></td></tr>`;
 			} else {
-				buffer += `<tr><th>Format</th><th><abbr title="Elo rating">Elo</abbr></th><th>W</th><th>L</th><th>Total</th>`;
+				const fail = `<tr><th>Format</th><th><abbr title="Elo rating">Elo</abbr></th><th>W</th><th>L</th><th>Total</th>`;
+				buffer += fail;
```
Zarel:
```suggestion
				buffer += `<tr><th>Format</th><th><abbr title="Elo rating">Elo</abbr></th><th>W</th><th>L</th><th>Total</th>`;
```

I'm pretty sure our length limit regex handles this correctly; if it doesn't, we'll need to fix it in the regex, not here.

server/chat-commands/core.ts
```diff
@@ -769,10 +783,11 @@ exports.commands = {
 		}
 		battle.allowExtraction[targetUser.id].add(user.id);
 		this.addModAction(`${user.name} consents to sharing battle team and choices with ${targetUser.name}.`);
-		if (Object.keys(battle.playerTable).length === battle.allowExtraction[targetUser.id].size) {
+		if (battle.inputLog && Object.keys(battle.playerTable).length === battle.allowExtraction[targetUser.id].size) {
 			this.addModAction(`${targetUser.name} has extracted the battle input log.`);
 			const inputLog = battle.inputLog.map(Chat.escapeHTML).join(`<br />`);
-			targetUser.sendTo(room, `|html|<div class="chat"><code style="white-space: pre-wrap; overflow-wrap: break-word; display: block">${inputLog}</code></div>`);
+			const formatted = `<div class="chat"><code style="white-space: pre-wrap; overflow-wrap: break-word; display: block">${inputLog}</code></div>`;
+			targetUser.sendTo(room, `|html|${formatted}`);
```
Zarel:
```suggestion
			targetUser.sendTo(
				room,
				`|html|<div class="chat"><code style="white-space: pre-wrap; overflow-wrap: break-word; display: block">${inputLog}</code></div>`,
			);
```

server/chat-commands/core.ts
```diff
@@ -769,10 +783,11 @@ exports.commands = {
 		}
 		battle.allowExtraction[targetUser.id].add(user.id);
 		this.addModAction(`${user.name} consents to sharing battle team and choices with ${targetUser.name}.`);
-		if (Object.keys(battle.playerTable).length === battle.allowExtraction[targetUser.id].size) {
+		if (battle.inputLog && Object.keys(battle.playerTable).length === battle.allowExtraction[targetUser.id].size) {
```
Zarel: ?

Zarel: If it's possible for `battle.inputLog` to be falsy, add an error message for that case (like the others). Don't silently fail.

mia-pi-git: checking for the log, since it can be possibly null - wasn't sure if there was a better way to handle that?

Zarel: You could also use `battle.inputLog!` to loudly fail.

But the point is, when you're writing code, think about what will happen when it gets run.

What will happen if `battle.inputLog` is falsy? No message at all, nothing to tell you what succeeded or failed. What a horrible user experience.

server/chat-commands/core.ts
```diff
@@ -796,7 +811,8 @@ exports.commands = {
 				if (playerUser.id === user.id) {
 					battle.allowExtraction[user.id].add(user.id);
 				} else {
-					playerUser.sendTo(room, Chat.html`|html|${user.name} wants to extract the battle input log. <button name="send" value="/allowexportinputlog ${user.id}">Share your team and choices with "${user.name}"</button>`);
+					const msg = `${user.name} wants to extract the battle input log. <button name="send" value="/allowexportinputlog ${user.id}">Share your team and choices with "${user.name}"</button>`;
+					playerUser.sendTo(room, Chat.html`|html|${msg}`);
```
Zarel: This will completely fail to render the button. Try it out and you'll see what's wrong.

In the future, just do the multiline thing like I suggested for `allowexportinputlog`.

server/chat-commands/core.ts
```diff
@@ -807,7 +823,8 @@ exports.commands = {
 				const playerUser = player.getUser();
 				if (!playerUser || battle.allowExtraction[user.id].has(playerUser.id)) continue;
 				logExported = false;
-				playerUser.sendTo(room, Chat.html`|html|${user.name} wants to extract the battle input log. <button name="send" value="/allowexportinputlog ${user.id}">Share your team and choices with "${user.name}"</button>`);
+				const share = `<button name="send" value="/allowexportinputlog ${user.id}">Share your team and choices with "${user.name}"</button>`;
+				playerUser.sendTo(room, Chat.html`|html|${user.name} wants to extract the battle input log. ${share}`);
```
Zarel: Multiline thing like I suggested for `allowexportinputlog`.

server/chat-commands/core.ts
```diff
@@ -832,18 +849,19 @@ exports.commands = {
...
 			battle.p2.name = target.slice(nameIndex2 + 8, nameNextQuoteIndex2);
 		}
-
-		battleRoom.auth[user.id] = Users.HOST_SYMBOL;
-		this.parse(`/join ${battleRoom.roomid}`);
-		setTimeout(() => {
-			// timer to make sure this goes under the battle
-			battleRoom.add(`|html|<div class="broadcast broadcast-blue"><strong>This is an imported replay</strong><br />Players will need to be manually added with <code>/addplayer</code> or <code>/restoreplayers</code></div>`);
-		}, 500);
+		if (battleRoom) {
```
Zarel: Silent failure

server/chat-commands/core.ts
```diff
@@ -1089,10 +1117,11 @@ exports.commands = {
 
 	timer(target, room, user) {
 		target = toID(target);
-		if (!room.game || !room.game.timer) {
+		// const game = room.getGame(require('../../.server-dist/room-battle').RoomBattle)
+		if (!room.battle || !room.battle.timer) {
```
Zarel: This was intentionally written to be generic for any game with a similar timer API.

This should be fixed in the `RoomGame` definition, not here.

server/chat-commands/core.ts
```diff
@@ -1103,13 +1132,15 @@ exports.commands = {
...
-		if (!force && !room.game.playerTable[user]) {
+		if (!force && !room.game.playerTable[user.id]) {
 			return this.errorReply(`Access denied.`);
 		}
 		if (this.meansNo(target) || target === 'stop') {
 			if (timer.timerRequesters.size) {
-				timer.stop(force ? undefined : user);
-				if (force) room.send(`|inactiveoff|Timer was turned off by staff. Please do not turn it back on until our staff say it's okay.`);
+				const staffForced: any = force ? undefined : user;
+				timer.stop(staffForced);
```
Zarel: Who taught you to write code like this?

server/chat-commands/core.ts
```diff
@@ -1103,13 +1132,15 @@ exports.commands = {
...
+				const msg = `Timer was turned off by staff. Please do not turn it back on until our staff say it's okay.`;
+				if (force) room.send(`|inactiveoff|${msg}`);
```
Zarel:
```suggestion
				if (force) {
					room.send(`|inactiveoff|Timer was turned off by staff. Please do not turn it back on until our staff say it's okay.`);
				}
```

server/chat-commands/core.ts
```diff
@@ -1168,14 +1199,14 @@ exports.commands = {
 	search(target, room, user, connection) {
 		if (target) {
 			if (Config.laddermodchat) {
-				let userGroup = user.group;
+				const userGroup = user.group;
 				if (Config.groupsranking.indexOf(userGroup) < Config.groupsranking.indexOf(Config.laddermodchat)) {
-					let groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
+					const groupName = Config.groups[Config.laddermodchat as any].name || Config.laddermodchat;
```
Zarel:
```suggestion
					const groupName = Config.groups[Config.laddermodchat as string].name || Config.laddermodchat;
```

server/chat-commands/core.ts
mia-pi-git: me, which would be the problem - wasn't sure how to handle it and this worked :s

server/chat-commands/core.ts
```diff
@@ -1275,15 +1308,15 @@ exports.commands = {
 			return;
 		}
 		if (!target) return this.errorReply("Provide a valid format.");
-		let originalFormat = Dex.getFormat(target);
+		const originalFormat = Dex.getFormat(target);
 		// Note: The default here of [Gen 7] Pokebank Anything Goes isn't normally hit; since the web client will send a default format
-		let format = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 7] Pokebank Anything Goes');
-		if (format.effectType !== 'Format') return this.popupReply("Please provide a valid format.");
+		const tier = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 7] Pokebank Anything Goes');
+		if (tier.effectType !== 'Format') return this.popupReply("Please provide a valid format.");
```
Zarel: Why did you rename `format` to `tier`?

Inside PS, `tier` always refers to a banlist, while `format` refers to a ruleset for a playable game.

Zarel: (This should be renamed back.)

Zarel: Like, UUBL is a tier, Random Battle is a format.

server/chat-commands/core.ts
```diff
@@ -1486,8 +1521,8 @@ exports.commands = {
 		} else {
 			let altCommandHelp;
 			let helpCmd;
-			let targets = target.split(' ');
-			let allCommands = Chat.commands;
+			const targets = target.split(' ');
+			const allCommands: AnyObject = Chat.commands;
```
Zarel:
```suggestion
			const allCommands = Chat.commands;
```

This does not need a cast, just fix it in the implementation.

Zarel: Feel free to push the errors, I'll know how to fix them once GitHub Actions tells me what they are.

server/chat-commands/core.ts
```diff
@@ -1275,15 +1310,15 @@ exports.commands = {
 			return;
 		}
 		if (!target) return this.errorReply("Provide a valid format.");
-		let originalFormat = Dex.getFormat(target);
+		const originalFormat = Dex.getFormat(target);
 		// Note: The default here of [Gen 7] Pokebank Anything Goes isn't normally hit; since the web client will send a default format
-		let format = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 7] Pokebank Anything Goes');
-		if (format.effectType !== 'Format') return this.popupReply("Please provide a valid format.");
+		const tier = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 7] Pokebank Anything Goes');
```
KrisXV:
```suggestion
		const tier = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 8] Anything Goes');
```

gen 7 hasnt existed for years

server/chat-commands/core.ts
```diff
@@ -1252,7 +1295,7 @@ exports.commands = {
 		if (target) return this.popupReply(`This command does not support specifying multiple users`);
 		const targetUser = this.targetUser || this.pmTarget;
 		if (!targetUser) return this.popupReply(`User "${this.targetUsername}" not found.`);
-		Ladders.acceptChallenge(connection, targetUser);
+		void Ladders.acceptChallenge(connection, targetUser);
```
TheJetOU:
```suggestion
		return Ladders.acceptChallenge(connection, targetUser);
```

server/chat-commands/core.ts
```diff
@@ -1275,13 +1318,15 @@ exports.commands = {
...
+		const originalFormat = Dex.getFormat(target);
+		// Note: The default here of [Gen 8] Anything Goes isn't normally hit; since the web client will send a default format
+		const format = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat(
+			'[Gen 7] Pokebank Anything Goes'
+		);
 		if (format.effectType !== 'Format') return this.popupReply("Please provide a valid format.");
 
-		TeamValidatorAsync.get(format.id).validateTeam(user.team).then(result => {
-			let matchMessage = (originalFormat === format ? "" : `The format '${originalFormat.name}' was not found.`);
+		void TeamValidatorAsync.get(format.id).validateTeam(user.team).then(result => {
```
TheJetOU:
```suggestion
		return TeamValidatorAsync.get(format.id).validateTeam(user.team).then(result => {
```

server/chat-commands/core.ts
```diff
@@ -1324,28 +1369,30 @@ exports.commands = {
 				}));
 				return false;
 			}
-			let roomList = {};
+			let roomList: any = {};
```
TheJetOU:
```suggestion
			let roomList: {[roomid: string]: {p1: string, p2: string, isPrivate: boolean}} = {};
```

server/chat-commands/core.ts
```diff
@@ -1324,28 +1369,30 @@ exports.commands = {
...
+			// has to be disabled because roomid is reassigned, but only in an if statement.
+			// eslint-disable-next-line prefer-const
 			for (let roomid of targetUser.inRooms) {
 				if (roomid === 'global') continue;
-				let targetRoom = Rooms.get(roomid);
+				const targetRoom = Rooms.get(roomid);
 				if (!targetRoom) continue; // shouldn't happen
-				let roomData = {};
+				const roomData: AnyObject = {};
```
TheJetOU:
```suggestion
				const roomData: Partial<{p1: string, p2: string, isPrivate: string}> = {};
```

server/chat-commands/core.ts
```diff
@@ -1324,28 +1369,30 @@ exports.commands = {
...
 				}
 				if (targetRoom.battle) {
-					let battle = targetRoom.battle;
+					const battle = targetRoom.battle;
 					roomData.p1 = battle.p1 ? ' ' + battle.p1.name : '';
 					roomData.p2 = battle.p2 ? ' ' + battle.p2.name : '';
 				}
 				if (targetRoom.auth && targetUser.id in targetRoom.auth) {
-					roomid = targetRoom.auth[targetUser.id] + roomid;
+					roomid = (targetRoom.auth[targetUser.id] + roomid) as RoomID;
```
TheJetOU: This is not a `RoomID`. You might want to define a `roomname` variable and assign this value to. That will also enable you to remove the ESLint diable.

server/chat-commands/core.ts
```diff
@@ -1324,28 +1369,30 @@ exports.commands = {
...
 				}
 				roomList[roomid] = roomData;
 			}
 			if (!targetUser.connected) roomList = false;
-			let userdetails = {
+			const userdetails: AnyObject = {
```
TheJetOU: Why does this need to be an `AnyObject`?

server/chat-commands/core.ts
```diff
@@ -1370,18 +1417,18 @@ exports.commands = {
 		} else if (cmd === 'laddertop') {
 			if (!trustable) return false;
 			const [format, prefix] = target.split(',').map(x => x.trim());
-			Ladders(toID(format)).getTop(prefix).then(result => {
+			void Ladders(toID(format)).getTop(prefix).then(result => {
```
TheJetOU:
```suggestion
			return Ladders(toID(format)).getTop(prefix).then(result => {
```

server/chat-commands/core.ts
```diff
@@ -1397,7 +1444,7 @@ exports.commands = {
 				visibility = 'public';
 			}
 
-			let roominfo = {
+			const roominfo: AnyObject = {
```
TheJetOU: Ditto.

server/chat-commands/core.ts
```diff
@@ -1437,7 +1484,7 @@ exports.commands = {
 
 		let commaIndex = target.indexOf(',');
 		let targetName = target;
-		let targetRegistered = false;
+		let targetRegistered: any = false;
```
TheJetOU: Again, why are you using `any` here? There's no real point in migrating a file to TypeScript if you're throwing out `any` or `AnyObject` at any slightly difficult type.

Your goal is not to shut up the compiler; it's to look at it's concerns and fix the issue.

server/chat-commands/core.ts
```diff
@@ -1449,7 +1496,7 @@ exports.commands = {
 				targetToken = target.substr(commaIndex + 1);
 			}
 		}
-		user.rename(targetName, targetToken, targetRegistered, connection);
+		void user.rename(targetName, targetToken, targetRegistered, connection);
```
TheJetOU:
```suggestion
		return user.rename(targetName, targetToken, targetRegistered, connection);
```

server/chat-commands/core.ts
```diff
@@ -1498,10 +1545,12 @@ exports.commands = {
 				}
 			} else if (targets.length > 1 && typeof allCommands[targets[0]] === 'object') {
 				// Handle internal namespace commands
-				let helpCmd = `${targets.pop()}help`;
-				let namespace = allCommands[targets.shift()];
+				helpCmd = `${targets.pop()}help`;
+				let namespace = allCommands[targets.shift() as string];
```
TheJetOU:
```suggestion
				let namespace = allCommands[targets.shift()!];
```

server/chat-commands/core.ts
Zarel: "Need" is a strong word, but I don't like going out of my way to type protocol messages.

server/chat-commands/core.ts
Zarel: You never responded to this! I told you this wouldn't work!

server/chat-commands/core.ts
```diff
@@ -1275,13 +1309,15 @@ exports.commands = {
 			return;
 		}
 		if (!target) return this.errorReply("Provide a valid format.");
-		let originalFormat = Dex.getFormat(target);
-		// Note: The default here of [Gen 7] Pokebank Anything Goes isn't normally hit; since the web client will send a default format
-		let format = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat('[Gen 7] Pokebank Anything Goes');
+		const originalFormat = Dex.getFormat(target);
+		// Note: The default here of [Gen 8] Anything Goes isn't normally hit; since the web client will send a default format
+		const format = originalFormat.effectType === 'Format' ? originalFormat : Dex.getFormat(
+			'[Gen 7] Pokebank Anything Goes'
```
urkerab, line 1315: You changed the comment to [Gen 8] Anything Goes but the actual code still refers to [Gen 7] Pokebank Anything Goes... it's a good thing that this code path is never taken!

PR #6493, opened 2020-03-26
Fix MissingNo.'s learnset

KrisXV:


Zarel:
Why `5`? This should probably be `3` for all of them?

PR #6494, opened 2020-03-27
Support searching roomlog entries

mia-pi-git:
[original comment is not useful anymore]


Zarel:
Okay, I definitely do want to add a text box feature for pages, for this sort of use case.

server/chat-plugins/chatlog.ts
```diff
@@ -366,7 +383,38 @@ export const pages: PageTable = {
 };
 
 export const commands: ChatCommands = {
-	chatlog(target, room, user) {
-		this.parse(`/join view-chatlog-${room.roomid}--today`);
+	chatlog: {
+		search(target, room, user) {
+			const [search, date] = target.split(',');
+			if (!date && search) {
+				this.parse(`/join view-chatlog-${room.roomid}--today--search-${target}`);
```
Zarel: `target` is not a roomid. You need to convert it with Dashycode.

server/chat-plugins/chatlog.ts
```diff
@@ -366,7 +383,38 @@ export const pages: PageTable = {
...
+			if (hide && !date) {
+				return this.parse(`/join view-chatlog-${room.roomid}--today--hide-${hide}`);
+			} else if (hide && date) {
+				return this.parse(`/join view-chatlog-${room.roomid}--${date}--hide-${hide}`);
+			} else {
+				return this.parse(`/help chatlog`);
+			}
+		},
+
+		''(target, room, user) {
```
Zarel: d-does this actually work?

server/chat-plugins/chatlog.ts
```diff
@@ -366,7 +383,38 @@ export const pages: PageTable = {
 };
 
 export const commands: ChatCommands = {
-	chatlog(target, room, user) {
-		this.parse(`/join view-chatlog-${room.roomid}--today`);
+	chatlog: {
+		search(target, room, user) {
```
Zarel: This will fail if you have a room named `search`.

I'm not sure how to best work around this; maybe have a separate `/chatlogroom [roomid], [search]` command, and then `/chatlog [search]`?

server/chat-plugins/chatlog.ts
mia-pi-git: it indeed does.

server/chat-plugins/chatlog.ts
mia-pi-git: i could just rename the '' command to 'view'?

mia-pi-git: renamed it, removed '' altogether.

server/chat-plugins/chatlog.ts
mia-pi-git: not certain why this is necessary, seeing as `Rooms.search` handles getting the roomid.

Zarel: Okay, let me put it this way.

I type `/chatlog search 日本語`

This does `/join view-chatlog-staff--today--search-日本語`

That's not a roomid.

server/chat-plugins/chatlog.ts
Zarel: ...I think that's worse, but whatever.

`/chatlog` by itself still needs to show the current room's log.

Zarel: Also how are you going to search deleted room logs?

server/chat-plugins/chatlog.ts
Zarel: Also why `today`? Is this only a search of a single day? That's pointless, you can just use ctrl+f. The only way this is remotely useful is if it searches multiple days.

mia-pi-git:
going to refactor this after my other major refactor is merged

mia-pi-git:
current UI: 
https://prnt.sc/rsf99u
https://prnt.sc/rsf9ei
open to suggestions.

Zarel:
ok I cleaned this one up for you too

Kaiepi:
Searching logs like this might take a long time for rooms with a lot of them. How does this benchmark compared to a regular old `grep` or `ripgrep`?

Zarel:
Delegating to ripgrep is probably a pretty good idea, actually.

mia-pi-git:
I'll look into that.

mia-pi-git:
This should probably be good.

TheJetOU:
> e: yea, discussed, ripgrep might not be necessary.

Why? Considering how slow reading one file is now, I won't be surprised if reading _and_ filtering a month of logs takes upwards of 5 minutes in active rooms. Ripgrep should in theory make it faster by using better searching strategies.

Counterpoint: I do plan to move chatlogs to a database with hopes of making it faster, so it might be wasted effort to optimize it, but I think shipping a feature which in theory is that slow isn't ideal.

mia-pi-git:
> Why?

Zarel just said it might not be necessary; lacking in the ability to adequately test lag, and knowing less about server operations, i assumed it was fine. I can refactor it to use ripgrep once i get it installed... which may take some time.

TheJetOU:
> Later edit: Also possible to just grep it, then every server could use it pretty easily. Downside is, no context. A combination, perhaps, or offering the ability to search either the pretty way or the fast way.

Grep has the command-line flag `--count N` to get N lines below and above the searched string.

mia-pi-git:
Did. 
https://prnt.sc/s036zg (FS search) 
https://prnt.sc/s037fa (grep search)
Not sure if there's still a point to FS, now, but i'll see if anyone else wants to weigh in.

mia-pi-git:
Ripgrep is operational.
 Unfortunately, I really don't know what to do about this build error..?

server/chat-plugins/chatlog.ts
```diff
@@ -319,6 +440,121 @@ const LogViewer = new class {
...
+			);
+		}
+		const matches = [];
+		for (const result of output.stdout.split('--').reverse()) {
+			matches.push(result);
+		}
+		return this.render(matches, roomid, search, cap);
+	}
+
+	async grepSearch(roomid: RoomID, search: string, cap?: number): Promise<string> {
```
TheJetOU, line 545: What's the point of still supporting grep?

server/chat-plugins/chatlog.ts
```diff
@@ -353,24 +591,39 @@ export const pages: PageTable = {
...
-			if (isNaN(parsedDate.getTime())) return LogViewer.error(`Invalid date.`);
-
-			if (date.split('-').length === 3) {
+			} else if (date.split('-').length === 3) {
 				return LogViewer.day(roomid, parsedDate.toISOString().slice(0, 10), opts);
 			} else {
 				return LogViewer.month(roomid, parsedDate.toISOString().slice(0, 7));
 			}
+		} else if (date && hasSearch && search) {
+			if (context) {
```
TheJetOU, line 616: Ripgrep does support context..

What I'd do is: Add a `Config.chatlogsreader` where it can be `fs` or `ripgrep`, and then run `fsSearch` or `ripgrepSearch` corresponding to the config option.

server/chat-plugins/chatlog.ts
mia-pi-git: I figured i'd leave it for side servers that wanted a faster option.

TheJetOU: Ripgrep is the faster option not grep.

mia-pi-git: Fair. I was thinking for people who don't want to / can't install it, but that's a decent point.

server/chat-plugins/chatlog.ts
```diff
@@ -380,4 +618,26 @@ export const commands: ChatCommands = {
...
+		const curRoom = tarRoom ? Rooms.search(tarRoom) : room;
+		const limit = cap ? `--${cap}` : `--500`;
+		return this.parse(`/join view-chatlog-${curRoom}--${date ? date : currentMonth}--search-${input}${limit}`);
+	},
+
+	searchlogshelp: [
+		"/searchlogs [search], [room], [cap], [date] - searches logs in the current room for [search].",
+		"A comma can be used to search for multiple words in a single line - in the format arg1, arg2, etc.",
+		"If a [cap] is given, limits it to only that many lines. Defaults to 500.",
+		"Requires: % @ & ~",
```
TheJetOU, line 641:
```suggestion
		"Requires: % @ # & ~",
```

mia-pi-git:
@Zarel this should be ready for review again now.

mia-pi-git:
Current UI for ripgrep: https://prnt.sc/sc5czv
FS UI is the same as above posts. 
extra-ready for review now.

mia-pi-git:
@Zarel bump? 
this is blocking a few things i want to do on chatlog (don't want to cause conflicts) and is definitely ready for review.

server/chat-plugins/chatlog.ts
```diff
@@ -6,6 +6,13 @@
  */
 
 import {FS} from "../../lib/fs";
+import * as child_process from 'child_process';
+import * as util from 'util';
+import * as path from 'path';
+
+
```
Zarel, line 13:
```suggestion

```

server/chat-plugins/chatlog.ts
```diff
@@ -380,4 +601,26 @@ export const commands: ChatCommands = {
...
+		target = target.trim();
+		const [search, tarRoom, cap, date] = target.split(',') as [string, string, number, number];
+		if (!target) return this.parse('/help searchlogs');
+		if (!search) return this.errorReply('Specify a query to search the logs for.');
+		if (cap && isNaN(cap) && toID(cap) !== 'all') return this.errorReply(`Cap must be a number or [all].`);
+		const input = search.includes('|') ? search.split('|').map(item => item.trim()).join('-') : search;
+		const currentMonth = Chat.toTimestamp(new Date()).split(' ')[0].slice(0, -3);
+		const curRoom = tarRoom ? Rooms.search(tarRoom) : room;
+		const limit = cap ? `--${cap}` : `--500`;
+		return this.parse(`/join view-chatlog-${curRoom}--${date ? date : currentMonth}--search-${input}${limit}`);
```
Zarel, line 617: This still has the problem I mentioned a long time about not supporting searching for things that contain things other than numbers or lowercase letters.

Zarel: Modlog uses Dashycode to solve this problem; I continue to recommend it here.

server/chat-plugins/chatlog.ts
```diff
@@ -353,24 +557,41 @@ export const pages: PageTable = {
...
+			if (Config.chatlogreader === 'fs') {
+				return LogSearcher.fsSearch(roomid, search, date, toID(cap));
+			} else if (Config.chatlogreader === 'ripgrep') {
+				return LogSearcher.ripgrepSearch(roomid, search, cap);
+			} else {
+				return LogViewer.error(
+					`<strong>Log searching has been configured incorrectly.<br>` +
+					`Please set Config.chatlogreader to 'fs' or 'ripgrep'.</strong>	`
+				);
+			}
```
Zarel, line 591: Maybe just default it to fs?

I'd also recommend throwing errors instead of using a more local error report. This ensures that the server owners are notified about the problem (rather than just the person using the command, who probably doesn't have the necessary access to fix it).

```suggestion
			if (Config.chatlogreader === 'fs' || !Config.chatlogreader) {
				return LogSearcher.fsSearch(roomid, search, date, toID(cap));
			} else if (Config.chatlogreader === 'ripgrep') {
				return LogSearcher.ripgrepSearch(roomid, search, cap);
			} else {
				throw new Error(`Config.chatlogreader must be 'fs' or 'ripgrep'.`);
			}
```

server/chat-plugins/chatlog.ts
mia-pi-git: My problem with dashycode is that it can create ids with '--', which breaks the chatlog opts.
Not really sure what to do about that.

mia-pi-git: I guess push it to '---' as the delimiter? That seems rather ridiculous, though.

Zarel: Yeah, I can understand how that would be a problem. You can use `splitFirst` (I think it's available as `Chat.splitFirst`) which partially solves it, and then manually use a regex to see if it ends with `--page-[number]`.

Zarel: I know, it's ugly, but I can't think of a better solution, and I think it's worth it.

server/chat-plugins/chatlog.ts
```diff
@@ -380,4 +601,28 @@ export const commands: ChatCommands = {
...
+	searchlogs(target, room) {
+		target = target.trim();
+		const [search, tarRoom, cap, date] = target.split(',') as [string, string, number, number];
+		if (!target) return this.parse('/help searchlogs');
+		if (!search) return this.errorReply('Specify a query to search the logs for.');
+		if (cap && isNaN(cap) && toID(cap) !== 'all') return this.errorReply(`Cap must be a number or [all].`);
+		const input = search.includes('|') ? search.split('|').map(item => item.trim()).join('-') : search;
+		const currentMonth = Chat.toTimestamp(new Date()).split(' ')[0].slice(0, -3);
+		const curRoom = tarRoom ? Rooms.search(tarRoom) : room;
+		const limit = cap ? `&${cap}` : `&500`;
```
Zarel, line 616:
```suggestion
		const limit = cap ? `&${cap}` : `&500`;
```

`&` isn't a valid character in roomids, either :(

server/chat-plugins/chatlog.ts
```diff
@@ -340,7 +543,7 @@ export const pages: PageTable = {
 		if (roomid.startsWith('spl') && roomid !== 'splatoon' && !user.can('rangeban')) {
 			return LogViewer.error("SPL team discussions are super secret.");
 		}
-		if (roomid.startsWith('wcop') && !user.can('rangeban')) {
+		if (roomid.startsWith('wcop') &&	!user.can('rangeban')) {
```
Zarel, line 546: Um, the fact that GitHub thinks this was changed worries me.

Weirdly enough, it still passes lint, so I'm confused about what was changed here.

Zarel: lol, apparently it's more obvious here: it's a tab instead of a space.

I'm surprised the linter isn't complaining.

server/chat-plugins/chatlog.ts
```diff
@@ -554,13 +550,13 @@ export const pages: PageTable = {
...
 		void accessLog.writeLine(`${user.id}: <${roomid}> ${date}`);
 		this.title = '[Logs] ' + roomid;
-		let cap = 'all';
+		let limit = null;
 		let search;
-		const [, term] = opts?.includes('search-') ? Chat.splitFirst(opts, 'search-') : [];
-		if (term) {
-			const [input, num] = term.split('&');
+		if (opts?.startsWith('search-')) {
+			const [input, limitString] = opts.split('--limit-');
```
Zarel, line 556: Oh whoops, I forgot to add special detection for `--all`, can you add that?

mia-pi-git: Gotcha.

server/chat-plugins/chatlog.ts
```diff
@@ -383,4 +602,35 @@ export const commands: ChatCommands = {
...
+		const [search, tarRoom, limit, date] = target.split(',');
+		if (!target) return this.parse('/help searchlogs');
+		if (!search) return this.errorReply('Specify a query to search the logs for.');
+		let limitString = '';
+		if (/^[0-9]+$/.test(limit)) {
+			limitString = `--limit-${limit}`;
+		} else if (toID(limit) === 'all') {
+			limitString = `--limit-all`;
+		} else if (!limit) {
+			limitString = `--limit-500`;
```
Zarel, line 619:
```suggestion
		let limitString;
		if (/^[0-9]+$/.test(limit)) {
			limitString = `--limit-${limit}`;
		} else if (toID(limit) === 'all') {
			limitString = `--limit-all`;
		} else if (!limit) {
			limitString = ``;
```

This just makes the URL be shorter if you don't already have a limit.

server/chat-plugins/chatlog.ts
```diff
@@ -356,24 +551,48 @@ export const pages: PageTable = {
...
+		let limit = null;
+		let search;
+		if (opts?.startsWith('search-')) {
+			let [input, limitString] = opts.split('--limit-');
+			input = input.slice(7);
+			search = Dashycode.decode(input);
+			if (limitString) {
+				limit = parseInt(limitString);
+			} else {
+				limit = null;
```
Zarel, line 564: Wait a minute, I don't see handling for the `--limit-all` case.

```suggestion
		/** null = no limit */
		let limit: number | null = null;
		let search;
		if (opts?.startsWith('search-')) {
			let [input, limitString] = opts.split('--limit-');
			input = input.slice(7);
			search = Dashycode.decode(input);
			if (limitString) {
				limit = parseInt(limitString) || null;
			} else {
				limit = 500;
```

Zarel:
There's some repetition in `searchDay`/`searchMonth`/`searchYear` that I should fix one day when I have free time, but there's no reason for that to block the merge. LGTM!

Zarel:
I ended up doing that fix now. I did want to get rid of the "manually text-search for `capped` to display the 'more' button" hack.

There's still some ways you could better arrange which code is in `LogViewer` and which is in `LogSearch` but I'm tired of working on this for now, so that can wait for a future PR.

PR #6497, opened 2020-03-28
Add Pokemon#hasBeenActive

KrisXV:
Currently not used for anything, but it will be used for OMs in the future at the very least. Also potentially useful for future mechanics.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -9824,7 +9824,6 @@ let BattleMovedex = {
 			}
 			// Instructed Fake Out etc. should fail
 			target.activeTurns++;
-			target.totalActiveTurns++;
 			this.add('-singleturn', target, 'move: Instruct', '[of] ' + source);
 			this.runMove(target.lastMove.id, target, /** @type {number} */(target.lastMoveTargetLoc));
 		},
--- a/data/moves.js
+++ b/data/moves.js
@@ -9824,6 +9824,7 @@ let BattleMovedex = {
 			}
 			// Instructed Fake Out etc. should fail
 			target.activeTurns++;
+			target.totalActiveTurns++;
 			this.add('-singleturn', target, 'move: Instruct', '[of] ' + source);
 			this.runMove(target.lastMove.id, target, /** @type {number} */(target.lastMoveTargetLoc));
 		},
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1465,6 +1465,7 @@ export class Battle {
 					sideStaleness = sideStaleness === 'external' ? sideStaleness : pokemon.staleness;
 				}
 				pokemon.activeTurns++;
+				pokemon.totalActiveTurns++;
 			}
 			trappedBySide.push(sideTrapped);
 			stalenessBySide.push(sideStaleness);
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1465,7 +1465,7 @@ export class Battle {
 					sideStaleness = sideStaleness === 'external' ? sideStaleness : pokemon.staleness;
 				}
 				pokemon.activeTurns++;
-				pokemon.totalActiveTurns++;
+				pokemon.hasBeenActive = true;
 			}
 			trappedBySide.push(sideTrapped);
 			stalenessBySide.push(sideStaleness);
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -191,7 +191,7 @@ export class Pokemon {
 
 	isActive: boolean;
 	activeTurns: number;
-	totalActiveTurns: number;
+	hasBeenActive: boolean;
 	truantTurn: boolean;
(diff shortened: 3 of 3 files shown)


Zarel:
Will it remain `0` if a Pokémon switches in and out the same turn? (e.g. an intentional switch followed by phazing?) Is that what you want?

KrisXV:
Not sure if it does that, but yes that would be preferable because thats not truly an active turn

data/moves.js
```diff
@@ -9824,6 +9824,7 @@ let BattleMovedex = {
 			}
 			// Instructed Fake Out etc. should fail
 			target.activeTurns++;
+			target.totalActiveTurns++;
```
Zarel: This is a horrible hack I think?

KrisXV: Why? if it counts towards increasing the active Pokemon's turn count, it should increase the overall turn count.

Zarel: Yes, and increasing the active turns is a horrible hack

Zarel: Piling hacks upon hacks is a bad idea, I think.

KrisXV: I'll remove it for now then.

Zarel:
I'm going to leave this open while I figure out how bad of a hack this actually is.

KrisXV:
can you figure it out within the next 4 days? The next projected OMotM will likely need this to exist.

Zarel:
Is a boolean `hasBeenActive` enough for the next OMotM?

KrisXV:
Yes

sim/battle.ts
```diff
@@ -1465,6 +1465,7 @@ export class Battle {
 					sideStaleness = sideStaleness === 'external' ? sideStaleness : pokemon.staleness;
 				}
 				pokemon.activeTurns++;
+				pokemon.hasBeenActive = true;
```
Zarel: This is the wrong place to set a previouslyActive flag...

KrisXV: well where would it be then?

Zarel: The point of switch-out seems like a sane place.

Zarel: Although depending on how the OM is designed, the point of move use might also work? Depends on the use-case.

Zarel: The variable name should reflect which one it tracks, though.

Zarel: `previouslySwitchedIn` vs `previouslyUsedMove`

scheibo: Should this not be a field on `m: PokemonModData;`?

Zarel: It's sufficiently general-purpose that I'm fine with it here.

KrisXV: It could be, but it seems like something that could have actual use in the future

Zarel:
After looking into this, it looks like it's a mess.

Anyway, `activeTurns` should probably be replaced with either one or two things, or at least renamed something to better reflect its status as a weird hack.

KrisXV:
I did not make this pull request with the intent of fixing a hack and I do not plan on doing that in regards to this anywhere in the near future.

Zarel:
You know you don't have to; you could have just ignored it and set `hasBeenActive` on switch-out...

KrisXV:
The way it was handled initially ie totalActiveTurns was the most preferable option so i will just resort to using PokemonModData until a non-hacky way is found to handle this

Zarel:
Can you stop being vague and just tell me how the mod works? Considering that it also works as a boolean, I'd be surprised if that was the best solution.

Zarel:
@KrisXV I never expected you to fix the hack. I fixed it myself in 2d78044604f47284e5.

Feel free to base your new implementation on either the new `activeTurns` or the new `activeMoveActions`, or even tell me if you want a `previouslySwitchedIn` or `previouslyUsedMove` boolean - I can implement those, too.

KrisXV:
I can't tell you how it works because there's no point in trying to quickly implement this if it doesn't win OMotM. If you really want to know, there is a thread for the OM, which is led by TI.

https://www.smogon.com/forums/threads/shared-power.3660877/

Zarel:
I don't understand why this needs a "hasn't switched in before" counter?

KrisXV:
The abilities don't apply to all Pokemon unless the Pokemon that originally has said ability has been active during the battle.

Zarel:
This definitely sounds like `previouslySwitchedIn` is the correct metric here.

KrisXV:
I'm still unsure on where to implement that; I'm unfamiliar with pokemon.ts

Zarel:
It's relatively easy to implement if you don't get distracted by making multiple refactors to the battle engine.

Zarel:
ba4fab47d2259f77daf6f279baf

There, I did this for you.

It will be `0` before your first switch-in, `1` after, `2` after your second switch in, etc.

PR #6498, opened 2020-03-28
Fix /weak not showing Pokemon names

KrisXV:


Zarel:
`pokemon` should probably be renamed `species` in this function?

server/chat-commands/info.js
```diff
@@ -801,19 +801,19 @@ const commands = {
...
+			species = {types: types};
 			target = types.join("/");
 		}
 
 		let weaknesses = [];
 		let resistances = [];
 		let immunities = [];
 		for (let type in mod.data.TypeChart) {
-			let notImmune = mod.getImmunity(type, pokemon);
+			let notImmune = mod.getImmunity(type, species);
```
```suggestion
			const notImmune = mod.getImmunity(type, species);
```

server/chat-commands/info.js
```diff
@@ -801,19 +801,19 @@ const commands = {
...
 			if (notImmune) {
-				let typeMod = mod.getEffectiveness(type, pokemon);
+				let typeMod = mod.getEffectiveness(type, species);
```
```suggestion
				const typeMod = mod.getEffectiveness(type, species);
```

PR #6501, opened 2020-03-29
Unmuddle code by using if statements.

pi1024e:
Having only one case in the default statement that won't be expanded upon is better off as an if statement, so the code is more readable.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1234,10 +1234,8 @@ export class Pokemon {
 			for (const typeid of type) {
 				if (this.hasType(typeid)) return true;
 			}
-		} else {
-			if (this.getTypes().includes(type)) {
-				return true;
-			}
+		} else if (this.getTypes().includes(type)) {
+			return true;
 		}
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


Zarel:
I appreciate the attempt, but this sort of thing is really a matter of personal taste, and definitely not the sort of change you should be making as a first contribution.

The advantage of the `switch` is to convey that this code is intended to be extended with more cases.

sim/pokemon.ts
```diff
@@ -1234,10 +1234,8 @@ export class Pokemon {
 			for (const typeid of type) {
 				if (this.hasType(typeid)) return true;
 			}
-		} else {
-			if (this.getTypes().includes(type)) {
-				return true;
-			}
+		} else if (this.getTypes().includes(type)) {
+			return true;
```
Zarel: This is also a situation where the original code is clearer: There's one check for the case where the types are an array, and another check for the case where the types aren't an array. It makes sense for them to be at the same level of indentation.

Zarel:
I'm not strongly opinionated on this style, though. Regular contributors to this repository should comment on how they feel about each of the two changes - if I'm in the minority, I'd be happy to accept this PR. Otherwise, I'll be closing this PR within at most a week.

sim/pokemon.ts
scheibo: make the else `return false`, then `return this.getTypes().includes(type)` outside of the block

scheibo:
> The advantage of the switch is to convey that this code is intended to be extended with more cases.

I agree with this - `switch` is definitely awkward in the situations where theres a single case, but I think if you look more at the rest of the file/project the `switch` statement is more consistent than 'use `if` if theres only a few number of cases'

sim/pokemon.ts
Zarel: If I were writing this myself, I'd do something like:

```ts
	hasType(type: string | string[]) {
		if (!type) return false;
		if (typeof type === 'string')) {
			return this.getTypes().includes(type);
		}

		const types = this.getTypes();
		for (const typeName of type) {
			if (types.includes(typeName)) return true;
		}
		return false;
	}
```

scheibo: Sure, go with that

Zarel:
When I say "not the sort of change you should be making as a first contribution", what I mean is: You shouldn't have your first pull request to a project be a personal-preference change. That's just not a good way to leave a good first impression.

If you were a long-time contributor, I might be willing to make accommodations to make the code look more like how you wanted it to look. But you don't even work on this codebase.

PR #6502, opened 2020-03-30
Add test for when Dynamax should revert

DaWoblefet:
Currently, Dynamax reversion is treated as an end-of-turn effect (and a very early one, I might add). It should be the last thing to occur before the start of the 4th turn, after switches have resolved. Because of the improper timing, things like Wishiwashi's transformation, various chip damage (e.g. G-Max Wildfire, Sticky Barb), etc. can all result in slightly different amounts of HP. This commits a skipped failing test for the interaction that gets as close to it as I can think of.

Cartridge interaction: https://youtu.be/5C6JfhhXn2w?t=1017
Sample replay showcasing why this matters, e.g. with G-Max Wildfire: https://replay.pokemonshowdown.com/gen8doublescustomgame-1087718887

A bandaid fix would be to just set Dynamax reversion as the last effect to happen in the end-turn resolution order.

### DIFF
--- a/test/sim/misc/dynamax.js
+++ b/test/sim/misc/dynamax.js
@@ -85,7 +85,7 @@ describe("Dynamax", function () {
 		assert.equal(pokemon.maxhp - pokemon.hp, expectedDamage, `${pokemon.name} should take ${expectedPercent * 100}%`);
 	});
 
-	it('should revert before the start of the 4th turn, not as an end-of-turn effect on the 3rd turn', function () {
+	it.skip('should revert before the start of the 4th turn, not as an end-of-turn effect on the 3rd turn', function () {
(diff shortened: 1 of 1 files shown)


Marty-D:
Technically, Dynamax ending _is_ part of end-of-turn effects, it just happens to take place after replacing fainted Pokemon, which is also an end-of-turn effect.

Like you mentioned, I'm thinking we can use a `ResidualOrder` of 30 for Dynamax, pending a real fix to replacing fainted Pokemon. I can't come up with any situation where it actually matters which happens first, since off the top of my head there are no existing switch-in Abilities that attempt to start an effect on the opponent that Dynamax Pokemon would be immune to.

PR #6507, opened 2020-04-01
Rework special scavenger modes

sparkychildcharlie:
- apply them as "formats"/"mods" instead of a large parent game with another room game nested inside
- the parent game is attached to ``room.scavgame`` as an independent object instead.
- add a command to set the twist (/scav settwist) and reset the twist (/scav resettwist) for scripted formats
     - this means that scavenger-games.js might be updated once per month as new twists are created.
- add a command to start a regular hunt with twists

More than one twist can be loaded into a single scavenger hunt (from scavenger game mode + /scav createtwist)

And as usual, this has Meicoo's seal of approval.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,4 @@
-server/chat-plugins/scavenger*.js @xfix
+server/chat-plugins/scavenger*.js @xfix @sparkychildcharlie
 server/chat-plugins/trivia.js @xfix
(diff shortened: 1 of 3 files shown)


Zarel:
You still haven't fixed #5912.

Zarel:
I will merge this because I don't want to get in the way of April Fool's fun, I guess, but I am serious, I will delete the scavs plugin if you don't fix that soon.

PR #6509, opened 2020-04-01
Add test for priority of shift action in Triples battles

urkerab:
This is a regression from changeset 28be861 where the test passes.


Zarel:
I'm surprised you didn't just fix it. It should be relatively straightforward.

urkerab:
I have never looked into how shifting is implemented, so I didn't realise that the fix (ff0b80f) would be so simple.

PR #6521, opened 2020-04-03
Fix `Battle#switchIn`

TheJetOU:
Regression caused by: https://github.com/smogon/pokemon-showdown/commit/47b55f96bc5c62e51255012e154ea81a46a2bb90

`Battle#dragIn` used to run the `SwitchOut` event among other functions
but were accidentally left off in the refactor. This commit adds them back.

The main bug caused by this is Natural Cure not curing status aliments when
phased out by Roar.

@Zarel: **please review/merge ASAP** as a major ADV tournament is going on
currently, so Natural Cure being broken is a big deal.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1256,9 +1256,11 @@ export class Battle {
 		pokemon.isActive = true;
 		this.runEvent('BeforeSwitchIn', pokemon);
 		if (oldActive) {
-			this.runEvent('SwitchOut', oldActive);
-			oldActive.illusion = null;
-			this.singleEvent('End', oldActive.getAbility(), oldActive.abilityData, oldActive);
+			if (isDrag) {
+				this.runEvent('SwitchOut', oldActive);
+				oldActive.illusion = null;
+				this.singleEvent('End', oldActive.getAbility(), oldActive.abilityData, oldActive);
+			}
 			oldActive.isActive = false;
 			oldActive.isStarted = false;
 			oldActive.usedItemThisTurn = false;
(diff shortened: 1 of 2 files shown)


Zarel:
How did the old switch-in code work, then? I was going to look into that but then it got lost in the hustle/bustle of April Fool's.

Zarel:
Apparently it's in the `switch` action code.

Zarel:
This is why I did the refactor in the first place: _one_ of those timings is nearly definitely wrong, and it's our job to figure out which one, instead of just running this at both timings.

Zarel:
Also I thought Marty already pushed a skipped test for this?

Marty-D:
You never asked me to? I asked you to fix your mistake three days ago, because I was busy with other stuff. I don't see why you can't just merge this and figure out whatever you want to later.

Zarel:
"Two switchouts" seems likely to be a regression source. I'd rather fix it properly.

sim/battle.ts
```diff
@@ -1256,6 +1256,9 @@ export class Battle {
 		pokemon.isActive = true;
 		this.runEvent('BeforeSwitchIn', pokemon);
 		if (oldActive) {
+			this.runEvent('SwitchOut', oldActive);
+			oldActive.illusion = null;
+			this.singleEvent('End', oldActive.getAbility(), oldActive.abilityData, oldActive);
```
Zarel, line 1261:
```suggestion
			if (isDrag) {
				this.runEvent('SwitchOut', oldActive);
				oldActive.illusion = null;
				this.singleEvent('End', oldActive.getAbility(), oldActive.abilityData, oldActive);
			}
```

Zarel:
I reviewed it basically the moment it was PRed, and I would have merged except that "two switchouts" is likely to cause as major bugs as it solves.

Zarel:
And I was right! This would have caused Regenerator to heal 2/3 instead of 1/3. That's why I didn't merge when it was PRed.

PR #6522, opened 2020-04-04
Add Shared Power

KrisXV:
thanks to urkerab for helping me fix the crashes

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -537,7 +537,8 @@ export class ModdedDex {
 		} else if (name.startsWith('item:')) {
 			effect = this.getItem(name.slice(5));
 		} else if (name.startsWith('ability:')) {
-			effect = this.getAbility(name.slice(8));
+			const ability = this.getAbility(name.slice(8));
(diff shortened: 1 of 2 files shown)


sim/dex.ts
```diff
@@ -537,7 +537,8 @@ export class ModdedDex {
 		} else if (name.startsWith('item:')) {
 			effect = this.getItem(name.slice(5));
 		} else if (name.startsWith('ability:')) {
-			effect = this.getAbility(name.slice(8));
+			const ability = this.getAbility(name.slice(8));
+			effect = Object.assign(Object.create(ability), {id: 'ability:' + ability.id});
```
scheibo, line 541: Surely this is a horrible idea? @Zarel 

Doesn't this overwrite an id field which should be type `ID`?

KrisXV: luckily getEffect is unused to call abilities and this only lasts for a month; you can ask @urkerab for more info though.

scheibo: I strongly disagree with adding bugs, even temporarily, for niche OM usecases.

PR #6524, opened 2020-04-05
Mark Plates (except Pixie Plate) as past gen items

DaWoblefet:
You can't obtain Plates, save Pixie Plate, in Sword and Shield.


Zarel:
What does "can't obtain" mean? PS uses "Unreleased", not "Past", for items that exist in the game code but are unobtainable.

Zarel:
("Past" means "doesn't exist in the game code at all", like PSNCureBerry)

PR #6528, opened 2020-04-06
Typescript info.js

KrisXV:


Zarel:
I don't like the custom pokedex in Gen 3. I'd much rather do it in `getSpecies`.

Zarel:
The validator is just hardcoded to consider Gen 4 abilities invalid in Gen 3. I've given up on a lot of that premature optimization these days but I still think updating `getSpecies` is a lot better than individually updating the dex entries.

data/mods/gen3/scripts.js
```diff
@@ -6,7 +6,8 @@ let BattleScripts = {
 	gen: 3,
 	init() {
 		for (let i in this.data.Pokedex) {
-			delete this.data.Pokedex[i].abilities['H'];
```
Zarel, line 9: I assume this means you tested and found this unnecessary?

KrisXV: yes

KrisXV:
I accidentally pushed to this pull request, but the code was completely reset to master before I started typescripting so this can just be repurposed

PR #6529, opened 2020-04-06
Additional fix for NatDex validation

tmagicturtle:
This fix re-introduces Arceus plates.


Zarel:
Didn't we agree to add `itemUser` to Plates instead?

Zarel:
I guess I understand "any pokemon can get a benefit from Plates" so it's not exactly the same thing... I guess this is fine.

PR #6530, opened 2020-04-06
Make it easier to use abilities as volatiles

urkerab:
Also making `Dex#getEffect` typesafe by moving the hack into the new `Battle#getVolatile`, which is allowed to return an `Effect` anyway.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -282,6 +282,13 @@ export class Battle {
 		this.prng = new PRNG(this.prng.startingSeed);
 	}
 
+	getVolatile(volatile: string): Effect {
+		if (volatile.startsWith('move:')) return this.dex.getActiveMove(volatile.slice(5));
+		if (volatile.startsWith('item:')) return this.dex.getItem(volatile.slice(5));
+		if (volatile.startsWith('ability:')) return this.dex.getAbility(volatile.slice(8));
+		return this.dex.getEffect(volatile);
+	}
+
 	suppressingAttackEvents(target?: Pokemon) {
 		return this.activePokemon && this.activePokemon.isActive && this.activePokemon !== target &&
 			this.activeMove && this.activeMove.ignoreAbility;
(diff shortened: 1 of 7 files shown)


KrisXV:
Could you also utilize this in the Shared Power code?

Zarel:
This doesn't actually solve the problem...

sim/battle.ts
```diff
@@ -282,6 +282,13 @@ export class Battle {
 		this.prng = new PRNG(this.prng.startingSeed);
 	}
 
+	getVolatile(id: string): Effect {
+		if (id.startsWith('move:')) return Object.assign(Object.create(this.dex.getMove(id.slice(5))), {id});
```
scheibo, line 286: This also still seems wrong, `id` for Effect should be an `ID`.

urkerab:
The volatile system was laxly allowing you to use effect names as volatiles names. In order to support OMs I've had to switch everyone to use effect ids or prefixed move, item or ability ids. I've also removed support for using effects as volatile names, which was basically only linked volatiles at this point, so `volatiles[volatile].linkedStatus` is now a string rather than a PureEffect.

sim/battle.ts
urkerab: This was just bad naming.

Zarel:
The correct thing, I think, is to replace `getRelevantEffects`?

urkerab:
You mean `findEventHandlers`? How would that help? `addVolatile` and `removeVolatile` still need to know the real effect.

Zarel:
You wouldn't use `addVolatile` and `removeVolatile` to add/remove items/abilities, of course.

urkerab:
In that case, this has suffered scope creep...

Zarel:
perhaps we _could_ do something, though

Zarel:
I think it makes sense to support having non-IDs as keys in the volatiles table.

PR #6538, opened 2020-04-08
Add rulesets directly in Roulettemons formats.js

charizard8888:
For Standard Pet Mods Clause which was earlier in mods/csm

https://github.com/smogon/pokemon-showdown/blob/master/data/rulesets.js#L851
```
standardpetmod: {
		effectType: 'ValidatorRule',
		name: 'Standard Pet Mod',
		desc: "Holds all custom Pet Mod ruleset validation",
		// Implemented in mods/[petmod]/rulesets.js
	},
```

Rulesets weren't in there so everything is legal in battles at the moment https://replay.pokemonshowdown.com/gen8petmodroulettemons-1092931925


Zarel:
I don't think it makes sense to have a "Standard Pet Mod" at all.

You should either inherit "Standard" and add/remove clauses as desired, or just put the clauses directly into your format. There's no reason for a Standard anything if it's only one format.

charizard8888:
Removed
By the way, the remaining sprites from here https://github.com/smogon/pokemon-showdown/pull/6508#issuecomment-608958349 will be ready in the next week.

config/formats.js
```diff
@@ -798,7 +798,7 @@ let Formats = [
 		],
 
 		mod: 'roulettemons',
-		ruleset: ['Standard Pet Mod'],
+		ruleset: ['Standard NatDex', 'Dynamax Clause', 'Sleep Clause Mod', 'Species Clause', 'Moody Clause', 'Evasion Moves Clause', 'Swagger Clause', 'Baton Pass Clause', 'OHKO Clause'],
```
Zarel, line 801: I'm surprised Standard NatDex doesn't have these clauses.

charizard8888: Wait was there a change or I just hit resolve, my bad

charizard8888: National Dex formats have these clauses added separately instead of that for some reason
https://github.com/smogon/pokemon-showdown/blob/master/config/formats.js#L422
```
ruleset: ['Standard NatDex', '+Unobtainable', '+Past', 'OHKO Clause', 'Evasion Moves Clause', 'Species Clause', 'Dynamax Clause', 'Sleep Clause Mod'],
```

PR #6539, opened 2020-04-08
Roulettemons final pokemon

scoopapa:
On a different note, we are having an issue with Z moves and Dexited moves now being illegal. We have Standard Natdex in the rulesets, which I would have thought would handle that. What needs to be done to fix that?


Zarel:
TI broke Standard NatDex here:

https://github.com/smogon/pokemon-showdown/commit/1e07e2aa1c8773cf6d770a4dc2d0670d7f08ed22

Weirdly enough, I told him the right way to do it, the day before:

https://github.com/smogon/pokemon-showdown/commit/387dd7ce2d0b834580f1b9b8c3a8f95abb544df4#r38327719

And then I told him it was broken two days ago.

So I'm surprised he still hasn't fixed it.

Zarel:
By the way, your PR looks to be very poorly based: out of the 10 commits in this PR, only two of them are actually the change you wanted to make. Do you need help cleaning your fork?

scoopapa:
Yeah I'm very bad at GitHub, I thought I would have handles that because I deleted my fork and re-forked, but apparently not. How do I prevent that in the future?

scoopapa:
Also should I leave standard natdex in and wait for TI to fix it? Or should I try to get a working set of rules right now?

Zarel:
You never need to delete your fork and re-fork. You can just hard-reset and force-push.

If you haven't set up upstream yet (GitHub for Desktop automatically does that for you), you first need to do this:

```
git remote add upstream git@github.com:smogon/pokemon-showdown.git
```

Once upstream is set up, you can force your master branch back to PS's master branch with this:

```
git checkout master
git reset --hard upstream/master
git push -f
```

Zarel:
(I fixed Standard NatDex for you.)

scoopapa:
Oh awesome, thanks!

Zarel:
In the future, you should never put your own code into master: you should always create a new branch to do work on. Let your master stay clean so you can update it and rebase on top of it where necessary.

PR #6541, opened 2020-04-08
Support custom ladders

Zarel:
This allows formats to specify a different ladder than their own, for their rated battles to go on.

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -457,6 +457,8 @@ export class RoomBattle extends RoomGames.RoomGame {
 	readonly title: string;
 	readonly allowRenames: boolean;
 	readonly format: string;
+	/** Will exist even if the game is unrated, in case it's later forced to be rated */
+	readonly ladder: string;
 	readonly gameType: string | undefined;
 	readonly challengeType: ChallengeType;
 	/**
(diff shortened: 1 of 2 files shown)


TheImmortal:
What's happening with this?

Zarel:
hm, we should probably use this solution. I didn't hear any objections, and we don't really have a better solution to our problems.

Zarel:
Would you prefer this or like `rulesetNoInherit`?

TheImmortal:
Tbh, I think the awareness ``(suspect test)`` brings is worth it. There isn't really a way for users to know which tiers are running suspect tests anymore unless they play on that ladder or visit the tier's forum. From the Random Battle suspect test, I can tell you that PS has a huge community that don't even use Smogon. So yeah. I think being able to see which tiers are currently running suspect tests just by looking at the formats list is a big plus.

Zarel:
Neat, we'll do this next restart.

Zarel:
@TheImmortal sorry for the wait, but now this is finally happening!

Zarel:
There's no longer an example in the diff, so example here:

```
	{
		name: "[Gen 8] OU (suspect test)",

		mod: 'gen8',
		rated: "[Gen 8] OU",
		ruleset: ['[Gen 8] OU'],
		unbanlist: ['Melmetal'],
		onBegin() {
			if (this.rated && this.format.id === 'gen8ou') this.add('html', `<div class="broadcast-red"><strong>OU is currently suspecting Melmetal! For information on how to participate check out the <a href="https://www.smogon.com/forums/threads/3662190/">suspect thread</a>.</strong></div>`);
		},
	},
```

PR #6546, opened 2020-04-09
Fix crash in roomlogs

whalemer:


server/chat-plugins/chatlog.ts
```diff
@@ -357,10 +357,18 @@ export const pages: PageTable = {
 
 		this.title = '[Logs] ' + roomid;
 		if (date && date.length === 10 || date === 'today') {
-			return LogViewer.day(roomid, date, opts);
+			if (date === 'today') {
+				return LogViewer.day(roomid, 'today', opts);
+			}
```
urkerab, line 362: Wouldn't this be better off as its own top-level condition at this point?

whalemer: Looking at it again, I think I'd prefer to put the ``if (day === 'today') day = LogReader.today();`` in the chatpage function.

PR #6548, opened 2020-04-09
Universal moves for the last four pokemon of rulettemons. 

scoopapa:
Should be the last necessary element, sorry I didn't catch these sooner.


KrisXV:
Why are all of the learnsets in gen 7? isnt this a gen 8 meta?

scoopapa:
Well, 7L1 is what my coding tool spits out for creating fakemons' movesets, but this mod did start in gen 7 which is why we have Z moves still. We just decided to add gen 8 moves and abilities into the randomized pool when that came around.

On another note, we just noticed that mega pokemon are still legal, even though all of the other canon pokemon were succesfully made illegal by adding "All Pokemon" to the banlist. What would be the most effectient way to make sure mega pokemon aren't legal?

Zarel:
You have a coding tool for fakemons' movesets? Can you show me?

scoopapa:
Download - https://www.dropbox.com/s/70yqpzbj31fggxu/SS2PD.zip?dl=0
Tutorial - https://www.smogon.com/forums/threads/pet-mods-general-discussion-thread.3621931/post-7963043

scoopapa:
It takes data from a spreadsheet ( which most pet mods have to catalogue all of their changes / data ) and create code. It's out of date for gen 8 though. And it's coded in autohotkey which i regret but at the time that was what i knew how to use

Zarel:
I see the problem for megas. It's because Standard NatDex whitelists past-generation megas and unobtainable Pokémon, bypassing the `-All Pokemon` rule. You need to manually ban `-Past` and `-Unobtainable`.

scoopapa:
banlist: ['All Pokemon', '-Past', '-Unobtainable'], 
? with the minus as part of the string?

Zarel:
```
banlist: ["asdfjkl"]
```
and
```
ruleset: ["-asdfjkl"]
```
mean the same thing.

I just call it `-asdfjkl` no matter how it's written, to make it clear that it's a ban rule.

scoopapa:
oh, I see, thanks, in that case I will put them in the banlist so ruleset is shorter

Zarel:
I could probably integrate this into Standard NatDex, though.

scoopapa:
Integrate -Unobtainable into standard natdex?
To be honest I don't think roulettemons need Standard Natdex in the ruleset. It worked on dragon heaven without that ( only with +Past ), and I think it was only added to save space.

scoopapa:
Okay I there is an update: the only pokemon currently banned are those that are in Gen 8. I mistakenly thought it was allowing Ubers before, but it seems it's all non-galar pokemon that are currently erroneously allowed in roulettemons. I would guess +Past is the culprit, however without that, Z-moves and old gen moves are not legal. Do you know an easy way to fix this? I was able to get the megas banned by putting each of their names in hte banlist, but that doesn't seem feasible for hundreds of pokemon.

Zarel:
I'm trying to figure this out because it's not behaving as expected. `Gen 8 National Dex@@@-All Pokemon` does ban Megazard Y, but `Gen 8 Roulettemons` doesn't, so I think something subtler is going on here.

Zarel:
Turns out, the format name is `Gen 8 Pet Mod Roulettemons`. It seems to be working now. Can you give me an example set that should be banned but isn't?

scoopapa:
Tyrantrum @ Assault Vest  
Ability: Strong Jaw  
EVs: 248 HP / 252 Atk / 8 SpD  
Adamant Nature  
- Brick Break  
- Dragon Tail  
- Earthquake  
- Fire Fang

scoopapa:
In the meantime I have come up with a workaround that works on Dragon Heaven
- In formats-data.js, list all the new fakemons' tier as "Roulettemons"
- add a custom ruleset "Roulettemons Clause" that rejects your team if all pokemon don't have the tier Roulettemons, and add it to the ruleset for the format

PR #6549, opened 2020-04-10
Chatlog: fix a bug with dates

mia-pi-git:
peach's fix for the crash was causing an error; `parsedDate` would return a date one day behind the given date, so going backwards / forwards would go to the wrong date (eg: 04-04 log's back a day button to 04-03 would skip you to 04-02, and forwards a day would take you from 04-04 to 04-06.) 
So if date's not valid, return error, else just take it as normal.


mia-pi-git:
closing since peach has a better fix

PR #6550, opened 2020-04-10
Fix bugs in chatlog date parsing

whalemer:
fuck the js date api

mia's doesn't work because `${date}-15` isn't guaranteed to be valid, ie `2010-04-5` is valid while `2010-04-5-15` isn't.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -130,7 +130,6 @@ const LogReader = new class {
 
 const LogViewer = new class {
 	async day(roomid: RoomID, day: string, opts?: string) {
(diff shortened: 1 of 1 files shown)


Zarel:
Yes, I know, JavaScript's date API is super fucked up.

But, um, I was pretty sure our app doesn't need any advanced date parsing.

Zarel:
What was wrong with the old code? I can't tell what this is trying to fix.

whalemer:
`new Date(whatever)` parses `whatever` as UCT, `<date>.getDate` and friends read it back as local time.

Zarel:
Oh! Yeah, that makes sense.

PR #6553, opened 2020-04-10
Typescript data

KrisXV:
This'll take a little bit


KrisXV:
@Zarel @scheibo (i know i already spoke to you about this pre but im gonna tag again) I'm not sure how to fix the errors that im getting. Any ideas?

urkerab:
Is it necessary to remove all of those quotes? If so, is it possible to split those changes into a separate commit so that they don't pollute the real changes?

KrisXV:
it is necessary to do it because typescript files typically throw errors for unnecessary quotes.

scheibo:
> it is necessary to do it because typescript files typically throw errors for unnecessary quotes.

It is not necessary, just disable that lint rule for these files

Zarel:
I'd prefer removing quotes. Why keep them?

Zarel:
I'll be honest, the only reason we have them is because the script I wrote to generate the data from an Excel spreadsheet had them, and I was too lazy to remove them.

That, and quoting only "return" was kind of dumb, but ES5 exists now.

Zarel:
I'm fine with putting the quotes change in a separate commit, though?

scheibo:
> I'm fine with putting the quotes change in a separate commit, though?

I agree quotes are probably not worth retaining, but I think Kris on chat says hed have trouble undoing the quote changes. 

If its really important than the quote formatting change be saved for a later commit instead of just committed as part of the typescripting I imagine this is also something he can just locally edit the eslint config to require quotes, then run `--fix` on the files he changed, then revert the config.

Zarel:
I don't have a strong preference for separating the quotes change. urkerab asked for one to make the change easier to review, but I don't personally know whether or not that's worth the effort. I'll let Kris decide.

Zarel:
> @Zarel @scheibo (i know i already spoke to you about this pre but im gonna tag again) I'm not sure how to fix the errors that im getting. Any ideas?

Add `./data/**/*.ts` to `overrides.files` in `.eslintrc-*.json`.

KrisXV:
Theres type errors that i'm confused by right now; I probably just missed some instances. some assistance would be appreciated

Zarel:
There's too much `max-len` in the CI output. Can you raise the max-len like I asked?

KrisXV:
I did i thought

data/random-teams.ts
```diff
@@ -361,43 +317,46 @@ class RandomTeams {
...
 
 			// Random IVs
-			let ivs = {hp: this.random(32), atk: this.random(32), def: this.random(32), spa: this.random(32), spd: this.random(32), spe: this.random(32)};
+			const ivs: StatsTable = {
+				hp: this.random(32),
+				atk: this.random(32),
+				def: this.random(32),
+				spa: this.random(32),
+				spd: this.random(32),
+				spe: this.random(32)};
```
TheImmortal, line 352: } on a new line.

data/mods/gen2/scripts.ts
```diff
@@ -258,19 +243,18 @@ let BattleScripts = {
...
+			let i: number;
 			for (i = 0; i < hits && target.hp && pokemon.hp; i++) {
 				if (pokemon.status === 'slp' && !isSleepUsable) break;
 				move.hit = i + 1;
 				if (move.hit === hits) move.lastHit = true;
 				moveDamage = this.moveHit(target, pokemon, move);
 				if (moveDamage === false) break;
 				if (nullDamage && (moveDamage || moveDamage === 0 || moveDamage === undefined)) nullDamage = false;
-				damage = (moveDamage || 0);
+				damage = ((moveDamage as number | undefined) || 0);
```
Zarel, line 257:
```suggestion
				damage = (moveDamage || 0);
```

data/mods/gen6/random-teams.ts
```diff
@@ -774,8 +758,8 @@ class RandomGen6Teams extends RandomGen7Teams {
 				item = 'Sitrus Berry';
 			} else {
 				item = 'Red Card';
-				for (let m in moves) {
-					let move = this.dex.getMove(moves[m]);
+				for (const m in moves) {
+					const move = this.dex.getMove(moves[m]);
```
Zarel, line 762:
```suggestion
				for (const moveid of moves) {
					const move = this.dex.getMove(moveid);
```

data/mods/gennext/moves.ts
```diff
@@ -1315,8 +1305,8 @@ let BattleMovedex = {
 				this.add('-end', user, 'Leech Seed', '[from] move: Rapid Spin', '[of] ' + user);
 				doubled = true;
 			}
-			let sideConditions = ['spikes', 'toxicspikes', 'stealthrock'];
-			for (let condition in sideConditions) {
+			const sideConditions = ['spikes', 'toxicspikes', 'stealthrock'];
+			for (const condition in sideConditions) {
```
Zarel, line 1309:
```suggestion
			for (const condition of sideConditions) {
```

data/scripts.ts
```diff
@@ -787,39 +783,42 @@ let BattleScripts = {
...
+			if (!damage[i]) t = false;
 		}
 		// 1. call to this.getDamage
 		damage = this.getSpreadDamage(damage, targets, pokemon, move, moveData, isSecondary, isSelf);
 
-		for (let i = 0; i < targets.length; i++) {
-			if (damage[i] === false) targets[i] = false;
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		for (let [i, t] of targets.entries()) {
+			if (damage[i] === false) t = false;
```
Zarel, line 802: I'm just impressed that eslint caught a real bug and you just disabled it instead of trying to figure out what the bug was. ;)

data/mods/gen6/random-teams.ts
```diff
@@ -1,46 +1,33 @@
...
-	 * @param {?PRNG | [number, number, number, number]} [prng]
-	 */
-	constructor(format, prng) {
+export class RandomGen6Teams extends RandomGen7Teams {
+	constructor(format: Format | string, prng: PRNG | PRNGSeed | null) {
 		super(format, prng);
-		/**@type {AnyObject} */
-		// @ts-ignore
-		this.randomFactorySets = require('./factory-sets.json');
+		this.randomFactorySets = JSON.parse(FS('./factory-sets.json').readIfExistsSync() || "{}");
```
Zarel, line 9: why though????????

Zarel: First: `FS` is relative to PS root.

Second: This code silently fails if the file doesn't exist (causing line 981 to crash), instead of giving an error message like `require` would.

This is a really important point: **Error messages are good because they tell you what's wrong.** Suppressing errors is bad unless the error condition is something that's actually intended to be supported.

Zarel:
By the way, you still have some `.js` files in `data/` which haven't been deleted yet.

KrisXV:
I got rid of the extra js files @Zarel

Zarel:
Where did the new errors come from? o_o

Zarel:
Oh, I guess a more recent rebase.

Zarel:
Took me like four hours to lower the line length limit back down to 140. :(

scheibo:
> Took me like four hours to lower the line length limit back down to 140. :(

I the future, i would suggest trying `prettier` on the file with the desired line length set and then unfucking the stuff prettier makes ugly. I personally find it easier, YMMV

data/mods/ssb/statuses.ts
```diff
@@ -1318,16 +1335,16 @@ let BattleStatuses = {
 			this.add(`c|+SamJo|Oof, gotta get thiccer……. ;(`);
 		},
 	},
-	salamander: {
+	instruct: {
```
Zarel, line 1338: ?

PR #6554, opened 2020-04-10
Allow ROs to set what rank can edit /roomevents

mia-pi-git:
as per this approved suggestion https://www.smogon.com/forums/threads/change-events-permission-to-driver.3660370/
I thought there was probably a reason it was set for mods already, so i just figured i'd go for the second option and let ROs choose.  i also heard from some ROs it'd be preferred to let them choose.

### DIFF
--- a/data/formats-data.ts
+++ b/data/formats-data.ts
@@ -7,6 +7,7 @@ export const BattleFormatsData: {[k: string]: SpeciesFormatsData} = {
 	},
 	venusaur: {
 		randomBattleMoves: ["gigadrain", "leechseed", "sleeppowder", "sludgebomb", "substitute"],
+		randomBattleLevel: 82,
 		randomDoubleBattleMoves: ["earthpower", "gigadrain", "leechseed", "protect", "sleeppowder", "sludgebomb"],
 		tier: "UUBL",
 		doublesTier: "DOU",
@@ -23,6 +24,7 @@ export const BattleFormatsData: {[k: string]: SpeciesFormatsData} = {
 	},
 	charizard: {
 		randomBattleMoves: ["airslash", "earthquake", "fireblast", "focusblast", "roost"],
+		randomBattleLevel: 84,
 		randomDoubleBattleMoves: ["airslash", "dragonpulse", "heatwave", "overheat", "protect", "tailwind"],
 		tier: "RU",
 		doublesTier: "DOU",
@@ -48,6 +50,7 @@ export const BattleFormatsData: {[k: string]: SpeciesFormatsData} = {
 	},
 	blastoise: {
 		randomBattleMoves: ["earthquake", "hydropump", "icebeam", "rapidspin", "scald", "shellsmash"],
+		randomBattleLevel: 82,
 		randomDoubleBattleMoves: ["fakeout", "followme", "icywind", "lifedew", "muddywater", "protect", "scald"],
 		tier: "UU",
 		doublesTier: "DOU",
@@ -64,12 +67,14 @@ export const BattleFormatsData: {[k: string]: SpeciesFormatsData} = {
 	},
 	butterfree: {
 		randomBattleMoves: ["energyball", "hurricane", "quiverdance", "sleeppowder"],
+		randomBattleLevel: 86,
 		randomDoubleBattleMoves: ["hurricane", "pollenpuff", "protect", "quiverdance", "ragepowder", "sleeppowder", "tailwind"],
 		tier: "NU",
 		doublesTier: "(DUU)",
 	},
 	butterfreegmax: {
 		randomBattleMoves: ["airslash", "bugbuzz", "quiverdance", "sleeppowder"],
+		randomBattleLevel: 86,
 		randomDoubleBattleMoves: ["hurricane", "pollenpuff", "protect", "quiverdance", "ragepowder", "sleeppowder", "tailwind"],
 		tier: "(Uber)",
 		doublesTier: "(DUber)",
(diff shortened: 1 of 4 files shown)


server/chat-plugins/room-events.ts
```diff
@@ -47,7 +47,13 @@ export const commands: ChatCommands = {
 		edit: 'add',
 		add(target, room, user) {
 			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
-			if (!this.can('ban', null, room)) return false;
+			if (room.eventsRank === '%') {
+				if (!this.can('mute', null, room)) return false;
+			} else if (room.eventsRank === '@') {
+				if (!this.can('ban', null, room)) return false;
+			} else if (room.eventsRank === '#') {
+				if (!this.can('declare')) return false;
```
whalemer, line 55: This probably shouldn't be global.

server/chat-plugins/room-events.ts
```diff
@@ -222,6 +245,22 @@ export const commands: ChatCommands = {
...
+			if (!this.can('gamemanagement')) return false;
+			const ranks = ['%', '@', '#'];
+			const rank = target as GroupSymbol;
+			if (!ranks.includes(target)) {
+				return this.errorReply(`Invalid rank for this setting. Valid options are [%|@|#].`);
+			}
+			if (room.eventsRank === rank) return this.errorReply(`Events management is already set to ${rank}.`);
+			if (room.eventsRank !== rank && rank) {
+				room.eventsRank = rank;
+				room.chatRoomData!.eventsRank = rank;
```
whalemer, line 258: this assert isn't right, using this in a groupchat will probably crash

server/chat-plugins/room-events.ts
```diff
@@ -52,7 +52,7 @@ export const commands: ChatCommands = {
 			} else if (room.eventsRank === '@') {
 				if (!this.can('ban', null, room)) return false;
 			} else if (room.eventsRank === '#') {
-				if (!this.can('declare')) return false;
+				if (!this.can('gamemanagement')) return false;
```
whalemer, line 55: again, why is this global?

whalemer: also did you mean to change it in the other places?

server/chat-plugins/room-events.ts
mia-pi-git: does crash, resolved by blocking it in groupchats

server/chat-plugins/room-events.ts
mia-pi-git: yes, I did.  Dumb mistake.

Slayer95:
You may leverage ``User#authAtLeast(group, room)``

server/chat-plugins/room-events.ts
```diff
@@ -47,7 +47,7 @@ export const commands: ChatCommands = {
 		edit: 'add',
 		add(target, room, user) {
 			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
-			if (!this.can('ban', null, room)) return false;
+			if (!user.authAtLeast(room.eventsRank, room)) return false;
```
whalemer, line 50: this probably shouldn't fail silently.

mia-pi-git: I have no idea why this is failing silently. It executes a proper fail when i give it something like 
> if (!user.authAtLeast(room.eventsRank)) {
				return this.errorReply(`Must be rank ${room.eventsRank} or higher to use this command.`);
			}

whalemer: Yes, that's what I'm saying it should be. It's failing silently because `authAtLeast` doesn't send any kind of message on failure, unlike `can`

server/chat-plugins/room-events.ts
```diff
@@ -222,6 +221,25 @@ export const commands: ChatCommands = {
...
 			return this.sendReply(resultString);
 		},
+		enable(target, room, user, cmd) {
+			if (!this.can('gamemanagement')) return false;
+			const ranks = ['%', '@', '#'];
+			const rank = target as GroupSymbol;
+			if (!ranks.includes(target)) {
+				return this.errorReply(`Invalid rank for this setting. Valid options are [%|@|#].`);
+			}
+			if (!room.chatRoomData || room.title.includes('-')) {
```
whalemer, line 231: The `title` check doesn't really achieve anything useful, if anything you'd want a `roomid` check but even that's unnecessary when you have the `chatRoomData` check.

server/chat-plugins/room-events.ts
```diff
@@ -47,7 +47,9 @@ export const commands: ChatCommands = {
 		edit: 'add',
 		add(target, room, user) {
 			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
-			if (!this.can('ban', null, room)) return false;
+			if (!user.authAtLeast(room.eventsRank)) {
```
whalemer, line 50: also, you need to pass the room here, otherwise it'll default to global auth.

server/chat-plugins/room-events.ts
```diff
@@ -222,6 +229,25 @@ export const commands: ChatCommands = {
 			this.modlog('ROOMEVENT', null, resultString);
 			return this.sendReply(resultString);
 		},
+		enable(target, room, user, cmd) {
+			if (!this.can('gamemanagement')) return false;
```
whalemer, line 233: again, why are you not passing a room to this? this currently requires global auth.

whalemer:
hyper nitpick - using the command without an argument should probably show you the current setting, a la mochcat.

server/chat-plugins/room-events.ts
mia-pi-git: dense mistake

Zarel:
idk if I like that much configuration? As long as the changes are in the modlog/roomlog, a driver messing things up can be fixed, anyway.

Zarel:
There was some discussion amongst the staff about this PR.

My position is that I think I might be fine with, like, an overarching "per-room configurable permissions" system, but I definitely don't want to merge this much code every time someone wants to configure every command's permissions.

This, on top of "I don't actually see a problem with just lowering the permissions for `/roomevents` to Driver for everyone", means that I will not be accepting this PR at this time. I'll leave this open for a few days, in case someone wants to convince me otherwise.

mia-pi-git:
Ugh. I suck at git terminal. Please ignore this.

PR #6557, opened 2020-04-10
Moxie and Beast Boost should have one combined boost for all fainted Pokémon.

urkerab:
In addition, they should not activate for a spread move that ends the game.
(This contrasts with Soul-Heart, which will activate until the last target.)

This behaviour is on one of @DaWoblefet's lists somewhere.

(Given my recent luck, this is probably not the best way to implement this.)

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -264,7 +264,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 	battlebond: {
 		desc: "If this Pokemon is a Greninja, it transforms into Ash-Greninja after knocking out a Pokemon. As Ash-Greninja, its Water Shuriken has 20 base power and always hits 3 times.",
 		shortDesc: "After KOing a Pokemon: becomes Ash-Greninja, Water Shuriken: 20 power, hits 3x.",
-		onSourceFaint(target, source, effect) {
+		onSourceAfterFaint(length, target, source, effect) {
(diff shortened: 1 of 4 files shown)


DaWoblefet:
Verifying that tests look good!

Zarel:
Is `onFaint` used for anything after this, or should it be removed?

Zarel:
@urkerab bump?

urkerab:
Variations of `onFaint` are used by Battle Bond, Destiny Bond, Grudge, Power of Alchemy, Receiver, Sky Drop, and Soul-Heart.

It's also used to provide the faint messages for Super Staff Bros Brawl.

And it turns out that SSBB also has a "Beast Boost 2" Ability. Should I update that as well?

Zarel:
Can/should those also be moved to `onAfterFaint`?

urkerab:
Destiny Bond, Grudge, Power of Alchemy, Receiver and Sky Drop need to know the victim, so they can only use `onFaint`.

Soul-Heart definitely activates as each victim faints, so it also should keep using `onFaint`.

I don't know the exact mechanics of Battle Bond so I've asked the research team for clarification.

urkerab:
Battle Bond should activate at the same time Moxie and Beast Boost would, so I should fix that too.

Zarel:
Neat, thanks!

PR #6563, opened 2020-04-12
Fix Minior-Meteor forme not being allowed in Gen 7

scheibo:


Zarel:
I would much rather fix this in the engine, but I'll need some time to investigate what exactly is going wrong in the engine.

Zarel:
I'd merge-now-fix-later, but it doesn't seem like this is causing any problems at the moment (the only thing I think this affects is Minior-Meteor in Gen 7 Hackmons which isn't a thing anyone actually uses).

TheImmortal:
All the other formes do this, or similar, so not merging it doesn’t make much sense.

PR #6565, opened 2020-04-12
Fix residual order for G-Max Volcalith / Wildfire

DaWoblefet:
Currently, G-Max Volcalith / G-Max Wildfire / Sea of Fire do not have a proper residual order variable, so it runs _after_ the end-of-turn effects which I am sick of dealing with in VGC haha. This fix isn't perfect, though. Sea of Fire / G-Max Volcalith / G-Max Wildfire all definitely have a onResidualOrder of 5 and a suborder of 1, but that suborder of 1 is queue-based; it happens in whatever order the effects were set. So currently, if two of these 3 effects are active, it will be random which one happens first, which can matter but is more niche. My hacky solution is to set the G-Max moves to be suborder 1.1, and then write a skipped test for it, except there's _another_ mechanics bug that doesn't let healing Berries activate in between Sea of Fire and G-Max Volcalith, so my test is passing when it shouldn't be so yeah. I can adjust that however we want to do it. I'll see if I can think of a better test in the meantime.

### DIFF
--- a/test/sim/moves/gmaxvolcalith.js
+++ b/test/sim/moves/gmaxvolcalith.js
@@ -40,4 +40,34 @@ describe('G-Max Volcalith', function () {
 		battle.makeChoices('move sleeptalk, move sleeptalk', 'move sleeptalk, move sleeptalk');
 		assert.strictEqual(battle.p2.active[0].hp, battle.p2.active[0].maxhp - (Math.floor(battle.p2.active[0].maxhp / 6) * 4));
 	});
+
+	it.skip('should deal damage alongside Sea of Fire or G-Max Wildfire in the order those field effects were set', function () {
+		battle = common.createBattle({gameType: 'doubles'});
+		battle.setPlayer('p1', {team: [
+			{species: 'Coalossal-Gmax', item: 'Eject Button', moves: ['rockthrow', 'sleeptalk']},
+			{species: 'Charmander', moves: ['firepledge', 'sleeptalk']},
+			{species: 'Wynaut', level: 72, moves: ['grasspledge', 'nightshade', 'sleeptalk']},
+		]});
+		battle.setPlayer('p2', {team: [
+			{species: 'Roggenrola', ability: 'Receiver', item: 'Focus Sash', level: 1, moves: ['sleeptalk', 'tackle']},
+			{species: 'Boldore', ability: 'Receiver', item: 'Sitrus Berry', evs: {hp: 28}, moves: ['sleeptalk']},
+			{species: 'Torracat', ability: 'Cheek Pouch', item: 'Focus Sash', level: 1, moves: ['sleeptalk']},
+		]});
+		//Set up Volcalith first, then Sea of Fire
+		battle.makeChoices('move rockthrow 1 dynamax, move sleeptalk', 'move tackle 1, move sleeptalk');
+		battle.makeChoices('switch 3');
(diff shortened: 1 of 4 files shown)


test/sim/moves/gmaxvolcalith.js
```diff
@@ -40,4 +40,34 @@ describe('G-Max Volcalith', function () {
...
+		battle.makeChoices('switch 3');
+		battle.makeChoices('move grasspledge 1, move firepledge 1', 'move sleeptalk, move sleeptalk');
+		battle.makeChoices('', 'switch 3');
+
+		//Bring Cheek Pouch down to Sash so it will faint and set Receiver's HP to exactly 5/8 of max
+		battle.makeChoices('move nightshade 2, move firepledge 1', 'move sleeptalk, move sleeptalk');
+		let expectedHP = battle.p2.active[1].maxhp;
+
+		/* TODO: this passes when it shouldn't because Sitrus Berry isn't triggering between Sea of Fire and G-Max Volcalith like it should;
+		 * Sitrus Berry should be able to activate in between those damaging effects. */
```
Zarel, line 67: So does this mean you want more time to work on this, or are you okay committing as-is?

test/sim/moves/gmaxvolcalith.js
```diff
@@ -40,4 +40,34 @@ describe('G-Max Volcalith', function () {
...
+			{species: 'Coalossal-Gmax', item: 'Eject Button', moves: ['rockthrow', 'sleeptalk']},
+			{species: 'Charmander', moves: ['firepledge', 'sleeptalk']},
+			{species: 'Wynaut', level: 72, moves: ['grasspledge', 'nightshade', 'sleeptalk']},
+		]});
+		battle.setPlayer('p2', {team: [
+			{species: 'Roggenrola', ability: 'Receiver', item: 'Focus Sash', level: 1, moves: ['sleeptalk', 'tackle']},
+			{species: 'Boldore', ability: 'Receiver', item: 'Sitrus Berry', evs: {hp: 28}, moves: ['sleeptalk']},
+			{species: 'Torracat', ability: 'Cheek Pouch', item: 'Focus Sash', level: 1, moves: ['sleeptalk']},
+		]});
+		//Set up Volcalith first, then Sea of Fire
```
Zarel, line 56:
```suggestion
		// Set up Volcalith first, then Sea of Fire
```

PS convention is to have spaces after `//`, if you could also make those changes to the other places.

test/sim/moves/gmaxvolcalith.js
DaWoblefet: I just thought of a better test that gets around the issue. I'll toss it up in a bit.

DaWoblefet: Never mind, my other test doesn't function how I thought it would in-game.

DaWoblefet: Ok, came up with a new test that works much better!

DaWoblefet:
The new test is much better, as it actually fails instead of being a false positive! The new test sets Volcalith, then Sea of Fire, and checks whether or not a Golisopod takes only Volcalith damage. Currently on sim, Emergency Exit is bugged not to switch out at all there in this situation, so it's kind of two bugs in one. But since Sitrus Berry can't trigger in between Volcalith and Sea of Fire, I can't do my original elaborate test of Receiver Cheek Pouch + Sitrus Berry. Should be ready to go once Marty or whomever says the mechanics are good for onResidual.

Marty-D:
Looks good to me, nice work! 👍

PR #6566, opened 2020-04-12
Fix doubled effectiveness for victims of Tar Shot

urkerab:
When a Fire type move is used against a dual typed victim of Tar Shot, the resulting effectiveness is applied twice. This is because the Effectiveness event gets called for each type of the target. (Freeze Dry relies on this for instance.)

- For (dual) types that normally dual resist Fire, this means that they still dual resist Fire.
- Types that normally resist Fire are unaffected as they are now neutral to fire anyway.
- For dual types that were neutral to Fire, this means that they are now quad weak to Fire.
- For dual types that were weak to Fire, this means that they are now 16× weak to Fire.
- For (dual) types that were quad weak to Fire, this means that they are now 64× weak to Fire. (The sim clamps the weakness to this level, so you can't go higher using Forest's Curse.)

The test compares the damage by Coalossal's Flame Charge against Ferrothorn with the damage when Ferrothorn is hit by Tar Shot (doubling damage) but holding an Occa Berry (halving damage). Currently the latter cleanly one-shots Ferrothorn when it should only do about the same amount of damage.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -18457,10 +18457,10 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 				this.add('-start', pokemon, 'Tar Shot');
 			},
 			onEffectiveness(typeMod, target, type, move) {
+				if (move.type !== 'Fire') return;
 				if (!target) return;
-				if (move.type === 'Fire') {
-					return this.dex.getEffectiveness('Fire', target) + 1;
-				}
+				if (type !== target.getTypes()[0]) return;
+				return typeMod + 1;
 			},
 		},
 		boosts: {
--- a/test/sim/moves/tarshot.js
+++ b/test/sim/moves/tarshot.js
@@ -49,4 +49,25 @@ describe('Tar Shot', function () {
 		const damage = torn.maxhp - torn.hp;
 		assert.bounded(damage, [62, 74]);
 	});
+
+	it('should make the target weaker to fire', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Coalossal', ability: 'steamengine', moves: ['tarshot', 'flamecharge']}]});
+		battle.setPlayer('p2', {team: [{species: 'Snorlax', ability: 'thickfat', item: 'occaberry', moves: ['rest']}]});
+		battle.makeChoices('move tarshot', 'move rest');
+		battle.makeChoices('move flamecharge', 'move rest');
+		assert.equal(battle.p2.active[0].item, '');
+	});
+
+	it('should not make the target over 2x weaker to fire', function () {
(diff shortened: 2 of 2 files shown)


Zarel:
> This is because the Effectiveness event gets called for each type of the target. (Freeze Dry relies on this for instance.)

Wait, why? And why would Freeze Dry rely on this? It seems to me like it should just be one event. Freeze Dry can just loop through the target's types.

urkerab:
I guess we could do that. I think we'd need to do that to Inverse Mod too; Inverse Mod doesn't want to accidentally invert Freeze Dry's super-effectiveness against water, but it wants to invert Freeze Dry's effectiveness against any other type(s) the target might have. (Freeze Dry's event runs first because it's a `singleEvent` rather than a `runEvent`.) (Note that if you use Electrify on Freeze-Dry with Inverse Mod active then it's still super-effective against water.)

Zarel:
Hm. I think my way around that would be to hardcode Freeze-Dry in Inverse Mod.

urkerab:
The problem then arises that the attacks Delta Stream protects you against are of course different in Inverse Battles. In the current code this is easily handled as the Inverse Mod has already changed the effectiveness of the type of the attack against the Flying subtype of the defender, so that Delta Stream doesn't actually need to know whether the Inverse Mod is active or not, just whether this particular effectiveness calculation is against the Flying subtype of the defender.

I'm not sure where to go here; maybe a new event for Tar Shot to handle its unique effectiveness modification?

Zarel:
I would hardcode Delta Stream and Inverse Mod.

urkerab:
Sorry, I don't understand what you mean about hardcoding Delta Stream here. How do hardcode Delta Stream to know whether Inverse Mod is active or not?

Zarel:
`this.ruleTable.has('inversemod')`

urkerab:
So I tried to code this up, and I had to special-case Inverse Mod yet again for Flying Press, but now I can't call the `onEffectiveness` handlers from `server/chat-commands/info.js`, because they all want a `Pokemon` object in order to calculate the correct effectiveness, which means that they will no longer display the correct effectiveness for Flying Press or Freeze Dry.

Zarel:
Wow, that sounds more complicated than necessary. Let's just merge this PR as-is and deal with that later. Are there already tests for Delta Stream and Inverse Mod? If not, can you write them, so when I get around to trying to code that up, it's easy to tell if I broke something?

urkerab:
There are tests for Delta Stream and Inverse Mod on their own, but not one for the effect of Inverse Mod on Delta Stream or Flying Press.
I also saw a bug reported relating to Tar Shot and Delta Stream, which I don't think this PR completely fixes. I'll write a test for that too.

PR #6571, opened 2020-04-13
Fix Pastel Veil mechanics

urkerab:
Makes Pastel Veil pass the tests from #6568

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -2651,6 +2651,20 @@ let BattleAbilities = {
 	},
 	"pastelveil": {
 		shortDesc: "This Pokemon and its allies cannot be poisoned. On switch-in, cures poisoned allies.",
+		onStart(pokemon) {
+			for (const ally of pokemon.allies()) {
+				if (['psn', 'tox'].includes(ally.status)) {
+					this.add('-activate', pokemon, 'ability: Pastel Veil');
+					ally.cureStatus();
+				}
+			}
+		},
+		onUpdate(pokemon) {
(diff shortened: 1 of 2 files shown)


data/abilities.js
```diff
@@ -2651,6 +2651,20 @@ let BattleAbilities = {
...
+				}
+			}
+		},
+		onUpdate(pokemon) {
+			if (['psn', 'tox'].includes(pokemon.status)) {
+				this.add('-activate', pokemon, 'ability: Pastel Veil');
+				pokemon.cureStatus();
+			}
+		},
 		onAllySwitchIn(pokemon) {
```
TheImmortal, line 2668: Is this necessary? I never read anything that mentioned it in @DaWoblefet's research.

DaWoblefet: Not entirely sure what it's doing in particular, but if I'm reading this right, Pastel Veil should announce itself when blocking poison (except for switching in on Toxic Spikes), indicate its source, and indicate the target being cured or being made immune to poison.

TheImmortal: It cures the ally's status when the ally switches in. Not the Pastel Veil user, they're already active.

DaWoblefet: That behavior is correct too, it should cure an ally's poison on switchin if the other active Pokemon is Pastel Veil.

Zarel: Isn't `onUpdate` enough to cover the others?

Zarel: Oh, `onUpdate` only affects the user.

PR #6573, opened 2020-04-13
Prevent Crafty Shield from blocking Perish Song 

timbosu:
Fixes the behavior from this replay:
https://replay.pokemonshowdown.com/gen8doublescustomgame-1096032925.log

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -3120,6 +3120,7 @@ let BattleMovedex = {
 			onTryHitPriority: 3,
 			onTryHit(target, source, move) {
 				if (move && (move.target === 'self' || move.category !== 'Status')) return;
+				if (move.id === 'perishsong') return;
 				this.add('-activate', target, 'move: Crafty Shield');
 				return this.NOT_FAIL;
 			},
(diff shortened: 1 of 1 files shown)


Zarel:
Does Crafty Shield normally block moves that bypass Protect? I don't see the bypass-Protect detection...

timbosu:
Crafty Shield is working as intended, but the message "Crafty Shield protected the opposing [Pokemon]!" comes up twice when it shouldn't - those Pokemon are still being affected by Perish Song.

Zarel:
Does this affect other all-targeting moves? I don't think this should be hardcoded...

data/moves.js
```diff
@@ -3120,6 +3120,7 @@ let BattleMovedex = {
 			onTryHitPriority: 3,
 			onTryHit(target, source, move) {
 				if (move && (move.target === 'self' || move.category !== 'Status')) return;
+				if (move.id === 'perishsong') return;
```
Zarel, line 3123:
```suggestion
				if (['self', 'all'].includes(move.target) || move.category !== 'Status') return;
```

(`move.target === 'all'` covers Perish Song.)

Zarel:
Neat, thanks!

PR #6577, opened 2020-04-14
Improve hosttype detection for special case

asgdf:
Hostfiltering the whole range might cause casualties, but based on experience I am almost 100% sure that these (rather small) ranges specifically are not residential connections.
This is kinda bypassing the datacenters list so I'm open to better implementation ideas


Zarel:
I don't have ideas, either, sadly. :(

PR #6579, opened 2020-04-15
Add support for PMOTS

petuuuhhh:


sim/global-types.ts
```diff
@@ -24,7 +24,7 @@ type SparseStatsTable = Partial<StatsTable>;
 type BoostName = StatNameExceptHP | 'accuracy' | 'evasion';
 type BoostsTable = {[boost in BoostName]: number };
 type SparseBoostsTable = Partial<BoostsTable>;
-type Nonstandard = 'Past' | 'Future' | 'Unobtainable' | 'CAP' | 'LGPE' | 'G-Max' | 'Custom';
+type Nonstandard = 'Past' | 'Future' | 'Unobtainable' | 'CAP' | 'LGPE' | 'G-Max' | 'Custom' | 'PM';
```
scheibo, line 27: I do not know what a `'PM'` is, please use a more descriptive name. Also, why does `'Custom'` not apply here?

petuuuhhh: 'PM' is for Pet Mod

scheibo: Then please use 'Pet Mod' instead. But I still think 'Custom' applies here and 'Pet Mod' is unnecessary

petuuuhhh:
the problem is, now that I removed 'PM', the PMOTS fakemon aren't showing up

sim/global-types.ts
petuuuhhh: It only works with 'PM'

scheibo: Yes, please remove `G-Max`. 

If you add `'Pet Mod'` you will run into he same problem youre having with Pokestars showing up  if there are more than one Pet Mod included in the codebase. You need another way to distinguish Roulettemons specifically outside of the `Nonstandard` type (tier? a new field?).

Fundamentally, Pet Mods Pokemon are all a type of custom Pokemon

Zarel: PokeStars should be `Past`, not `Custom`, in theory. Should we fix that?

Zarel: Also pet mods should just whitelist like Roulettemons does.

scheibo: > PokeStars should be Past,
 Hard to say, because theyre effectively both? If only one is possible, Id say `Custom` better represents them / is more accurate. if `isNonstandard` was a `Nonstandard[]` then yeah, adding `Past` in there would also be more accurate

petuuuhhh: Yeah, but we all know we're not going to have multiple Pet Mods at the same time. and Zarel's Gen 6 mod isn't able to be played/doesn't change TeamBuilder's Pokemon displaying. I honestly don't see why we can't just have a PM tier

Zarel: PokeStars in Gen 5 are Unobtainable, and in future gens are Past.

I would treat them the same way as Floette-Eternal, which does work as I described.

Zarel: _During_ Gen 5, it was sensible to treat Floette-Eternal and Pokéstars differently, because people thought it was likely that Floette-Eternal would be distributed as an event, but since it never happened, there's no difference between them. We don't currently have a need to distinguish "Unobtainable" from "Unreleased".

Zarel: I'm fine with a PetMod tier. `Nonstandard` isn't a tier, but a separate flag, though.

scheibo: > PokeStars in Gen 5 are Unobtainable, and in future gens are Past.

Interesting. What uses `Custom` if not Pokestars? Isn't `Past` for Pokestars somewhat confusing - how are they filtered out of National Dex for instance (or are they)?

> I'm fine with a PetMod tier. 

Well, `Pet Mod` or something. I have no idea why something we call Pet Mods in regular speak needs to be turned into `PM` or `PetMod` in the codebase. If its actually a tier maybe `PM` makes sense for symmetry with `OU`, but I dont know that its as commonly as used an acronym that is as immediately obvious to readers as regular tiers.

> Nonstandard isn't a tier, but a separate flag, though.

^

petuuuhhh: didn't I make it a tier :/

Zarel: > Interesting. What uses `Custom` if not Pokestars?

Magikarp's Revenge and nothing else, as far as I'm aware.

Maybe we should just give CAP `isNonstandard: "Custom"`, and just let `tier: "CAP"` do the job for unbanning (although we would need to manually unban Mountaineer, Shadow Strike, etc, or allow giving tiers to CAP moves/abilities - this would require a tagging system which _is_ on the to-do list but would likely need to be post-Preact).

> Isn't `Past` for Pokestars somewhat confusing - how are they filtered out of National Dex for instance (or are they)?

Floette-E is currently manually blacklisted. Pokéstars would need to be blacklisted the same way, although in hindsight it might be better to refactor that check to "legal in either USUM or SwSh".

> Well, `Pet Mod` or something. I have no idea why something we call Pet Mods in regular speak needs to be turned into `PM` or `PetMod` in the codebase. If its actually a tier maybe `PM` makes sense for symmetry with `OU`, but I dont know that its as commonly as used an acronym that is as immediately obvious to readers as regular tiers.

The tier column in teambuilder tables is pretty narrow; `Pet Mod` doesn't fit. Actually, `PetMod` doesn't fit, either. They're truncated to `Pet M` and `PetMo`. I suppose I'm fine with either.

scheibo: > Maybe we should just give CAP `isNonstandard: "Custom"`,

As opposed to `isNonstandard: CAP`? Im down with that, streamlines things IMO

> and just let `tier: "CAP"` do the job for unbanning (although we would need to manually unban Mountaineer, Shadow Strike, etc, or allow giving tiers to CAP moves/abilities - this would require a tagging system which _is_ on the to-do list but would likely need to be post-Preact).


Ah, I see. OK, I understand better now why CAP has its own thing. That seems to imply Pet Mods with their own tiers would want them too?

What is SSB doing? Its a pretty elaborate mod with its own mons/abilities/moves and doesnt have these issues? Is this all about having client teambuilder support (which SSB doesnt require?)

> 
> > Isn't `Past` for Pokestars somewhat confusing - how are they filtered out of National Dex for instance (or are they)?
> 
> Floette-E is currently manually blacklisted. Pokéstars would need to be blacklisted the same way, although in hindsight it might be better to refactor that check to "legal in either USUM or SwSh".

I see. It would be nice to have a way to filter out things that have never been legal in games (like Floette-E and Pokestars) easily, which is kind of what I was hoping isNonstandard to be.


> The tier column in teambuilder tables is pretty narrow; `Pet Mod` doesn't fit. Actually, `PetMod` doesn't fit, either. They're truncated to `Pet M` and `PetMo`. I suppose I'm fine with either.

I'd prefer `Mod`, `Pet` and `PM` to `PetMo` lol

petuuuhhh: ok so I made it Pet M. it's showing the mons 4 times now (in the client)

Zarel: > Ah, I see. OK, I understand better now why CAP has its own thing. That seems to imply Pet Mods with their own tiers would want them too?

Pet mods who use custom moves/abilities can just manually whitelist them for now. That _would_ be the plan for CAP going forward.

> What is SSB doing? Its a pretty elaborate mod with its own mons/abilities/moves and doesnt have these issues? Is this all about having client teambuilder support (which SSB doesnt require?)

Close. It's about validator support, not teambuilder support. SSB uses a random team generator, which bypasses the validator.

> I see. It would be nice to have a way to filter out things that have never been legal in games (like Floette-E and Pokestars) easily, which is kind of what I was hoping isNonstandard to be.

I'm not opposed to something like `Past-Unobtainable` as a Nonstandard reason.

Zarel: @petuuuhhh please change it to `Pet Mod`. The client teambuilder will automatically shorten it to `Pet M`.

Zarel:
NO, STOP, GO BACK.

Your `tag2` commit was completely unnecessary; the bug in `tag` was a GitHub Actions issue that I fixed in a049d85a987ddc8c

sim/dex.ts
```diff
@@ -1017,7 +1017,7 @@ export class ModdedDex {
 				// valid pokemontags
 				const validTags = [
 					// singles tiers
-					'uber', 'ou', 'uubl', 'uu', 'rubl', 'ru', 'nubl', 'nu', 'publ', 'pu', 'zu', 'nfe', 'lcuber', 'lc', 'cap', 'caplc', 'capnfe', 'ag', 'pm',
+					'uber', 'ou', 'uubl', 'uu', 'rubl', 'ru', 'nubl', 'nu', 'publ', 'pu', 'zu', 'nfe', 'lcuber', 'lc', 'cap', 'caplc', 'capnfe', 'ag', 'pet mod',
```
KrisXV:
```suggestion
					'uber', 'ou', 'uubl', 'uu', 'rubl', 'ru', 'nubl', 'nu', 'publ', 'pu', 'zu', 'nfe', 'lcuber', 'lc', 'cap', 'caplc', 'capnfe', 'ag', 'petmod',
```

PR #6580, opened 2020-04-16
Add G-Max move validator illegality test

DaWoblefet:
Complements https://github.com/smogon/pokemon-showdown/commit/f194d31aacfcdd2edc33b95dd68738f84436c580

### DIFF
--- a/test/sim/team-validator.js
+++ b/test/sim/team-validator.js
@@ -492,6 +492,8 @@ describe('Team Validator', function () {
 		];
 		let illegal = TeamValidator.get('gen8anythinggoes').validateTeam(team);
 		assert(illegal);
+		illegal = TeamValidator.get('gen8customgame@@@-nonexistent').validateTeam(team);
+		assert(illegal);
 	});
 
 	/*********************************************************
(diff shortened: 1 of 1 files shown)


Zarel:
This is specifically about "g-max moves are invalid in Hackmons", in case that was unclear.

Zarel:
We don't need this for AG because AG already checks learnsets, which don't contain G-max moves.

test/sim/team-validator.js
```diff
@@ -486,6 +486,14 @@ describe('Team Validator', function () {
 		assert(illegal);
 	});
 
+	it('should reject exclusive G-Max moves added directly to a Pokemon\'s moveset', function () {
+		let team = [
+			{species: 'charizard-gmax', ability: 'blaze', moves: ['gmaxwildfire'], evs: {hp: 1}},
+		];
+		let illegal = TeamValidator.get('gen8anythinggoes').validateTeam(team);
+		assert(illegal);
```
Zarel, line 494:
```suggestion
		let illegal = TeamValidator.get('gen8anythinggoes').validateTeam(team);
		assert(illegal);
		illegal = TeamValidator.get('gen8balancedhackmons').validateTeam(team);
		assert(illegal);
```

TheImmortal: I would probably future-proof it and use ``gen8customgame@@@-nonexistent`` because gen 8 balanced hackmons might not exist on PS in some generation in the distant future.

PR #6584, opened 2020-04-16
Refactor Sockets to ProcessManager+TypeScript

Zarel:
This is fully working except for crash recovery.

Closes #5795


Zarel:
There's also a bunch of sockets tests that need to be rewritten now...

Zarel:
I spent a lot of time thinking about how ProcessManager should abstract something like this. In the end, I went for `RawProcessManager` - a process manager that specifically puts a single stream into a subprocess.

Kaiepi:
I wouldn't worry too much about the sockets tests. There's more work that needs to be done with sockets in order to make them testable in a more efficient way anyway.

Zarel:
The sockets tests are skipped right now, anyway. It turns out I just needed to fix `users-utils`. Now the only thing left is crash recovery.

Zarel:
@scheibo @Kaiepi I think you're the only ones with the right technical knowledge to review this.

Personally the part I'm most indecisive about are class names; I'm not a huge fan of `StreamWorker`.

Zarel:
(For the record, I'd still like a review, whenever you have time.)

PR #6587, opened 2020-04-18
Fix Technician interaction with Battery/Auras/PowerSpot

yghdj:
As pointed by @DaWoblefet, the interaction of Technician with some abilities boosting move BPs was faulty; Technician would apply even though the move was boosted over 60 BP. This should now be fixed. The concerned abilities are Fairy Aura, Dark Aura, Aura Break, Battery and Power Spot.

Also, according to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Technician_(Ability)), this behaviour doesn't apply to Steely Spirit. I'm not sure if I can trust Bulbapedia on this, and seeing as the only concerned move would be Metal Claw, which can only be learned by Scizor and is therefore unavailable in gen 8, this is not really checkable anyway. I coded it so that Technician would apply first.

https://github.com/smogon/pokemon-showdown/projects/3#card-25870312

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -65,7 +65,7 @@ let BattleAbilities = {
 				move.aerilateBoosted = true;
 			}
 		},
-		onBasePowerPriority: 8,
+		onBasePowerPriority: 7,
 		onBasePower(basePower, pokemon, target, move) {
 			if (move.aerilateBoosted) return this.chainModify([0x1333, 0x1000]);
 		},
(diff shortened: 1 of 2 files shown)


DaWoblefet:
We actually haven't had any official research done on the relative locations of the Gen 8 BP modifiers (nor any confirmation on their relative locations, though it's probably (tm) the same). I'll check out if the new Gen 8 BP multipliers apply before or after Technician today, doubt we'll get it extremely precise though.

DaWoblefet:
Also, and I assume this is the case, but this change doesn't affect the damage formula in generations prior to 7 right?

Marty-D:
It looks like every single Ability prior to this change was using a `BasePowerPriority` of 7 or 8, which is completely pointless. Other than that this looks good, so I'll merge now and deal with the rest of the problems surrounding this once DaWoblefet has results. Thanks!

PR #6556, opened 2020-04-10
SSB: Update several ranks

Nolnis:
Update ranks of SSB fighters.

PR #6588, opened 2020-04-18
SSBB Ranks: Update

DeltaCoderr:
I think there are some more users who are recently promoted. I will update it soon.


Zarel:
Please do not make PRs for SSB updates anymore. They are now batched in #6556.

PR #3432, opened 2017-04-05
Fix Weather Ball Z-Move

QuiteQuiet:
This fixes the Z-Weather Ball interaction with weather. As far as I'm aware, every other type-changing move retains their base typing with Z-Moves, so Weather Ball should be the only exception here?

There might be a better way than to rely on Weather Ball's `onModifyMove` to resolve the typing, but without reimplementing it again, I don't know how.

PR #6592, opened 2020-04-19
Fix Max Move base move source from leaking

DaWoblefet:
Currently, Max Moves are sending `sourceEffect` of their base move when the original move type isn't identical to source move type. This results in the server leaking which base move was used. For example:

|move|p2a: Nine lives|Max Flare|p1a: Smeargle|[from]Weather Ball
|move|p1a: Silvally|Max Flutterby|p2a: Necrozma|[from]Multi-Attack

It's more apparent with Weather Ball because it has a custom string attached to its Z-move transformation, but it's getting sent with any Max Move that dynamically changes typing, e.g. https://replay.pokemonshowdown.com/gen8doublescustomgame-1099210961.

### DIFF
--- a/data/scripts.js
+++ b/data/scripts.js
@@ -161,11 +161,9 @@ let BattleScripts = {
 			move = this.getActiveZMove(move, pokemon);
 		}
 		if (maxMove && move.category !== 'Status') {
-			let moveType = move.type;
 			// Max move outcome is dependent on the move type after type modifications from ability and the move itself
 			this.singleEvent('ModifyType', move, null, pokemon, target, move, move);
 			this.runEvent('ModifyType', pokemon, target, move, move);
-			if (move.type !== moveType) sourceEffect = move;
 		}
 		if (maxMove || (move.category !== 'Status' && sourceEffect && /** @type {ActiveMove} */(sourceEffect).isMax)) {
 			move = this.getActiveMaxMove(move, pokemon);


DaWoblefet:
Just so it's clear, I don't think there's any Max Moves on cartridge with a custom string attached to it. See examples for comparison:
https://www.youtube.com/watch?v=5C6JfhhXn2w&t=1487s
https://youtu.be/TalIrfSpPCk?t=131

Zarel:
Looks like this was copied from Z-Weather Ball:

https://github.com/smogon/pokemon-showdown/commit/143ea606fda93686dac12d5772830ccd1f526d96#diff-3b4e4b552bbef3758b2f3596fb94ae2fR159

It's probably safe to remove.

Zarel:
Looks like the Z-Weather Ball code was added here:

#3432

It also looks like it was a mistake. Can you try removing that (`if (move.type !== 'Normal') sourceEffect = move;`) and seeing what happens?

DaWoblefet:
Dropping that line prevents the client from recognizing the proper `sourceEffect` to display Weather Ball's custom message. I tried setting `sourceEffect = this.dex.getActiveMove('Breakneck Blitz');` but the client didn't recognize it because it didn't know it was Weather Ball anymore. The string associated with Z-Weather Ball is unique in Gen 7 so it's not an information leak there. Breakneck Blitz turned into [Z-move] due to the weather!

Zarel:
Oh! Maybe it should have a comment explaining that's what it's fore.

PR #6593, opened 2020-04-19
Fix Sky Drop Wonder Guard interaction

yghdj:
https://github.com/smogon/pokemon-showdown/projects/3#card-25870318

Sky Drop should affect every Wonder Guard Pokemon on the first turn, effectively taking them in the sky. Then Wonder Guard would activate on the second turn, when the attack actually happens. This should now be the case. Additionally, Sky Drop still affects Flying-weak Wonder Guard Pokemon.

I noticed the message "[Pokemon] is freed from the Sky Drop" still displays even if the Pokemon faints. I don't know if it should be the case or if it should be omitted.

### DIFF
--- a/data/abilities.js
+++ b/data/abilities.js
@@ -4659,6 +4659,7 @@ let BattleAbilities = {
 		shortDesc: "This Pokemon can only be damaged by supereffective moves and indirect damage.",
 		onTryHit(target, source, move) {
 			if (target === source || move.category === 'Status' || move.type === '???' || move.id === 'struggle') return;
+			if (move.id === 'skydrop' && !source.volatiles['skydrop']) return;
 			this.debug('Wonder Guard immunity: ' + move.id);
 			if (target.runEffectiveness(move) <= 0) {
 				if (move.smartTarget) {
--- a/data/moves.js
+++ b/data/moves.js
@@ -17060,9 +17060,7 @@ let BattleMovedex = {
 			}
 		},
 		onHit(target, source) {
-			if (target.hp > 0) {
-				this.add('-end', target, 'Sky Drop');
-			}
+			if (target.hp) this.add('-end', target, 'Sky Drop');
 		},
 		effect: {
 			duration: 2,
--- a/data/moves.js
+++ b/data/moves.js
@@ -17060,7 +17060,9 @@ let BattleMovedex = {
 			}
 		},
 		onHit(target, source) {
-			this.add('-end', target, 'Sky Drop');
+			if (target.hp > 0) {
+				this.add('-end', target, 'Sky Drop');
+			}
 		},
 		effect: {
 			duration: 2,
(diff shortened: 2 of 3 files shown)


data/abilities.js
```diff
@@ -4658,7 +4658,7 @@ let BattleAbilities = {
 	wonderguard: {
 		shortDesc: "This Pokemon can only be damaged by supereffective moves and indirect damage.",
 		onTryHit(target, source, move) {
-			if (target === source || move.category === 'Status' || move.type === '???' || move.id === 'struggle') return;
+			if (target === source || move.category === 'Status' || move.type === '???' || move.id === 'struggle' || move.id === 'skydrop') return;
```
Zarel, line 4661:
```suggestion
			if (target === source || move.category === 'Status' || move.type === '???' || move.id === 'struggle') return;
			if (move.id === 'skydrop' && !target.volatiles['skydrop']) return;
```

data/moves.js
```diff
@@ -17045,6 +17045,11 @@ let BattleMovedex = {
 					this.add('-immune', target);
 					return null;
 				}
+
+				if (target.hasAbility('wonderguard') && target.runEffectiveness(move) <= 0) {
+					this.add('-immune', target, '[from] ability: Wonder Guard');
+					return null;
+				}
```
Zarel, line 17052:
```suggestion
```

Zarel:
I'm not sure if my suggestion works, but if it does, I think it would be more elegant.

Either way, thanks for the contribution!

yghdj:
Yeah, it seems way easier that what I've originally done. I'll update this asap.

I'm grateful I can help with this, it helps me pass the time and practice :)

yghdj:
As a matter of fact, your suggestion doesn't work, because the 'skydrop' status is removed in onHit(), and the Wonder Guard check happens in onTryHit(), so I suppose it's not removed yet when the program comes to this point.

DaWoblefet:
In Gen 7, "[Pokemon] was freed from the Sky Drop" doesn't send if the Pokemon faints in the middle of Sky Drop (e.g. a No Guard combo attack) or if it faints from the Sky Drop itself.

Zarel:
Interesting. In general we probably need better detection for "prepare phase of move".

data/abilities.js
Marty-D: If you check `!source.volatiles['skydrop']` rather than `target`, this suggestion works.

yghdj:
I modified the code to match @Marty-D 's solution, and I added code to remove the message "[pokemon] freed from Sky Drop", when the target faints due to the move. I noticed the behaviour if the target fainted during the first turn of the move, is that the user would stay in the sky until the completion of the move, which I'm not sure is the right behaviour. Either way, here's the solution.

DaWoblefet:
> I noticed the behaviour if the target fainted during the first turn of the move, is that the user would stay in the sky until the completion of the move, which I'm not sure is the right behaviour.

This is correct. https://youtu.be/SdfhhZ1cQyM?t=807

Marty-D:
Looks good, thanks! 👍

PR #6594, opened 2020-04-19
SSB: Batch Update

Nolnis:
Name change that I missed in my batch update

### DIFF
--- a/data/mods/ssb/statuses.ts
+++ b/data/mods/ssb/statuses.ts
@@ -221,13 +221,13 @@ export const BattleStatuses: {[k: string]: ModdedPureEffectData} = {
 	asheviere: {
 		noCopy: true,
 		onStart() {
-			this.add(`c|~Asheviere|Blessings to you.`);
+			this.add(`c|+Asheviere|Blessings to you.`);
 		},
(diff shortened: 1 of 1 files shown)


Zarel:
This will have to wait a month for another batch, unfortunately.

Instructing:
Hi Instruct here!
I'm heavily against doing this, or else I would've done it myself, because of the reason of why I namechanged is why I haven't transferred this from the previous username to this one. PM me on PS for more information.

Instructing:
Hi nol! Instruct here!!!!

Just so you know, this is about half done in regards to a namechange (sorry that I didn't see your previous pr) for ssb, but you'll also have to update the name in ssb/random-teams.js (or however its formatted) or else the quotes will not show up. Thanks!

Zarel:
> Just so you know, this is about half done in regards to a namechange (sorry that I didn't see your previous pr) for ssb, but you'll also have to update the name in ssb/random-teams.js (or however its formatted) or else the quotes will not show up. Thanks!

I think the best solution to this sort of problem is to group things together in their own file, rather than leaving them spread out over multiple files.

Zarel:
Thanks!

PR #6598, opened 2020-04-21
Fix Snatch interation with Throat Chop and Heal Bell

yghdj:
https://github.com/smogon/pokemon-showdown/projects/3#card-25870296
https://github.com/smogon/pokemon-showdown/projects/3#card-25870288

Snatched sound-based moves should now be disabled when the Snatch user is under the effects of Throat Chop. Additionally, the message "A bell chimed!" is now displayed before the message "[Pokemon] snatched the opponent's move" when Heal Bell is snatched. Tbh there is probably a better way to do that but I wasn't sure how to use the event handlers correctly.

### DIFF
--- a/data/moves.js
+++ b/data/moves.js
@@ -8278,7 +8278,9 @@ let BattleMovedex = {
 		priority: 0,
 		flags: {snatch: 1, sound: 1, distance: 1, authentic: 1},
 		onHit(pokemon, source) {
-			this.add('-activate', source, 'move: Heal Bell');
+			if (!source.volatiles['snatch']) {
+				this.add('-activate', source, 'move: Heal Bell');
+			}
 			let side = pokemon.side;
 			let success = false;
 			for (const ally of side.pokemon) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
There are actually more than just this; those two cards happened to be from before SadisticMystic did his research on the checks done before move success in Generation 7. In [this list](https://www.smogon.com/forums/threads/ultra-sun-ultra-moon-battle-mechanics-research-read-post-2.3620030/page-10#post-8096570), anything between steps 12-37 are basically run again for Snatch. Basically, it might be preferable to do this the other way around; always let Snatch attempt to do all checks _except_ 1-11. It's hard because the simulator doesn't attempt to order move success like it should, which is a bit beyond the scope of this PR, but nothing you can really do about that.

yghdj:
Indeed, that looks like a lot of work... I can try to order the checks in the Snatch onTryHit() callback, but it would be much easier to have the simulator recheck everything in the useMove() method in the correct order.

DaWoblefet:
This was superseded by https://github.com/smogon/pokemon-showdown/pull/7758

PR #6602, opened 2020-04-22
Remove Scald from Raikou in randbats

tmagicturtle:


DaWoblefet:
This was part of the datamine from Pokemon Home, same with things like Aura Sphere Latios. I'm guessing @TheImmortal still wants it there?

tmagicturtle:
Oh, was it? That's my bad then. I saw it in a randbats match and did /learn raikou, scald but I guess if it's unreleased then it would still appear as unlearnable. I'll close this, then.

Zarel:
Honestly I still don't like using the datamine data, but I guess TI does what he wants :/

PR #6603, opened 2020-04-23
Typescript Thing of the Day

mia-pi-git:


Zarel:
Neat, thanks!

mia-pi-git:
That was... easier than i expected, lmao.

Zarel:
Yeah, it was already .js-TypeScript so none of the conversions here were very dangerous.

server/chat-plugins/thing-of-the-day.ts
```diff
@@ -79,25 +73,24 @@ class OtdHandler {
 
 		this.lastPrenom = 0;
 
-		/** @type {AnyObject[]} */
 		this.winners = [];
 
-		this.file.read().then(content => {
+		this.file.read().then((content: string) => {
```
Zarel, line 78: Actually this is weird and is symptomatic of something worth worrying about.

server/chat-plugins/thing-of-the-day.ts
```diff
@@ -682,7 +688,7 @@ exports.pages = {};
 /** @type {ChatCommands} */
```
Zarel, line 688: Also wait, you forgot this.

PR #6604, opened 2020-04-24
Unify Move's recoil fields

scheibo:
This commit is not purely a refactoring, behavior has been changed!

- Gen Next Reckless and datasearch.ts's recoil checks now also apply
to Mind Blown/Steelbeam/Struggle recoil as well, which seems to have
been their intent in addition to regular recoil and crash damage.

calcRecoilDamage's type signature has changed to make it typesafe -
the compiler can verify that we checked Array.isArray and that we're
passing the correct kind of recoil damage to calcRecoilDamage.

The client will also require a change to handle this new type
(recoil damage fields are used for the Reckless tooltip)

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3170,7 +3170,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		shortDesc: "This Pokemon's attacks with recoil or crash damage have 1.2x power; not Struggle.",
 		onBasePowerPriority: 23,
 		onBasePower(basePower, attacker, defender, move) {
-			if (move.recoil || move.hasCustomRecoil) {
+			if (Array.isArray(move.recoil) || move.recoil === 'crash') {
 				this.debug('Reckless boost');
 				return this.chainModify([0x1333, 0x1000]);
 			}
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2666,7 +2666,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		desc: "This Pokemon's damaging moves become multi-hit moves that hit twice. The second hit has its damage quartered. Does not affect multi-hit moves or moves that have multiple targets.",
 		shortDesc: "This Pokemon's damaging moves hit twice. The second hit has its damage quartered.",
 		onPrepareHit(source, target, move) {
-			if (move.category === 'Status' || move.selfdestruct || move.multihit) return;
+			if (move.category === 'Status' || move.selfDamage?.startsWith("KO") || move.multihit) return;
 			if (['iceball', 'rollout'].includes(move.id)) return;
 			if (!move.flags['charge'] && !move.spreadHit && !move.isZ && !move.isMax) {
 				move.multihit = 2;
@@ -3170,7 +3170,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		shortDesc: "This Pokemon's attacks with recoil or crash damage have 1.2x power; not Struggle.",
 		onBasePowerPriority: 23,
 		onBasePower(basePower, attacker, defender, move) {
-			if (Array.isArray(move.recoil) || move.recoil === 'crash') {
+			if ((move.recoil || move.selfDamage === "Crash")) {
 				this.debug('Reckless boost');
 				return this.chainModify([0x1333, 0x1000]);
 			}
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -881,6 +881,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		desc: "Deals Normal-type damage. If this move was successful, the user takes damage equal to 1/2 the HP lost by the target, rounded down, but not less than 1 HP. This move is automatically used if none of the user's known moves can be selected.",
 		shortDesc: "User loses 1/2 the HP lost by the target.",
+		selfDamage: "Recoil",
 		recoil: [1, 2],
 		onModifyMove() {},
 	},
@@ -904,6 +905,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		name: "Substitute",
 		pp: 10,
 		priority: 0,
+		selfDamage: "25% (can't KO)",
 		volatileStatus: 'substitute',
 		onTryHit(target) {
 			if (target.volatiles['substitute']) {
@@ -963,7 +965,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 				}
 				// Drain/recoil does not happen if the substitute breaks
 				if (target.volatiles['substitute']) {
-					if (Array.isArray(move.recoil)) {
+					if (move.recoil) {
 						this.damage(Math.round(uncappedDamage * move.recoil[0] / move.recoil[1]), source, target, 'recoil');
 					}
 					if (move.drain) {
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -963,7 +963,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 				}
 				// Drain/recoil does not happen if the substitute breaks
 				if (target.volatiles['substitute']) {
-					if (move.recoil) {
+					if (Array.isArray(move.recoil)) {
 						this.damage(Math.round(uncappedDamage * move.recoil[0] / move.recoil[1]), source, target, 'recoil');
 					}
 					if (move.drain) {
--- a/data/mods/gen3/scripts.ts
+++ b/data/mods/gen3/scripts.ts
@@ -84,7 +84,6 @@ export const BattleScripts: ModdedBattleScriptsData = {
 
 		if (!this.singleEvent('TryMove', move, null, pokemon, target, move) ||
 			!this.runEvent('TryMove', pokemon, target, move)) {
-			if (move.recoil === 'mindblown') move.recoil = undefined; // TODO: ???
 			return false;
 		}
 
@@ -94,7 +93,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 			move.ignoreImmunity = (move.category === 'Status');
 		}
 
-		if (move.selfdestruct === 'always') {
+		if (move.selfDamage === "KO") {
 			this.faint(pokemon, pokemon, move);
 		}
 
@@ -355,8 +354,8 @@ export const BattleScripts: ModdedBattleScriptsData = {
 			move.totalDamage = damage;
 		}
 
-		if (Array.isArray(move.recoil) && move.totalDamage) {
-			this.damage(this.calcRecoilDamage(move.totalDamage, move.recoil), pokemon, target, 'recoil');
+		if (move.recoil && move.totalDamage) {
+			this.damage(this.calcRecoilDamage(move.totalDamage, move), pokemon, target, 'recoil');
 		}
 
 		if (target && pokemon !== target) target.gotAttacked(move, damage, pokemon);
@@ -375,7 +374,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 		return damage;
 	},
 
-	calcRecoilDamage(damageDealt, recoil) {
-		return this.dex.clampIntRange(Math.floor(damageDealt * recoil[0] / recoil[1]), 1);
+	calcRecoilDamage(damageDealt, move) {
+		return this.dex.clampIntRange(Math.floor(damageDealt * move.recoil![0] / move.recoil![1]), 1);
 	},
 };
--- a/data/mods/gen3/scripts.ts
+++ b/data/mods/gen3/scripts.ts
@@ -84,7 +84,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 
 		if (!this.singleEvent('TryMove', move, null, pokemon, target, move) ||
 			!this.runEvent('TryMove', pokemon, target, move)) {
-			move.mindBlownRecoil = false;
+			if (move.recoil === 'mindblown') move.recoil = undefined; // TODO: ???
 			return false;
 		}
 
@@ -355,8 +355,8 @@ export const BattleScripts: ModdedBattleScriptsData = {
 			move.totalDamage = damage;
 		}
 
-		if (move.recoil && move.totalDamage) {
-			this.damage(this.calcRecoilDamage(move.totalDamage, move), pokemon, target, 'recoil');
+		if (Array.isArray(move.recoil) && move.totalDamage) {
+			this.damage(this.calcRecoilDamage(move.totalDamage, move.recoil), pokemon, target, 'recoil');
 		}
 
 		if (target && pokemon !== target) target.gotAttacked(move, damage, pokemon);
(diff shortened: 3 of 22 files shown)


data/mods/gen3/scripts.ts
```diff
@@ -84,7 +84,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 
 		if (!this.singleEvent('TryMove', move, null, pokemon, target, move) ||
 			!this.runEvent('TryMove', pokemon, target, move)) {
-			move.mindBlownRecoil = false;
+			if (move.recoil === 'mindblown') move.recoil = undefined; // TODO: ???
```
scheibo, line 87: why in Gen 3 concerned with mindblown recoil moves?

data/mods/gennext/abilities.ts
```diff
@@ -280,7 +280,7 @@ export const BattleAbilities: {[k: string]: ModdedAbilityData} = {
 	reckless: {
 		inherit: true,
 		onBasePower(basePower, attacker, defender, move) {
-			if (move.recoil || move.hasCustomRecoil || attacker.item === 'lifeorb') {
+			if (move.recoil || attacker.item === 'lifeorb') {
```
scheibo, line 283: This is a change in behavior, I dont know if its desirable. I dont think mindblown moves exist in Gen Next but this would provide a boost to Struggle which is perhaps not desired? The fact that life orb was also included meant me think this was meant to be more inclusive of all types of recoil moves

sim/global-types.ts
```diff
@@ -779,6 +779,7 @@ interface EventMethods {
 	onTypePriority?: number;
 }
 
+// FIXME: half of these fields do not belong here
```
scheibo, line 782: I will address this in a followup change as we discussed on Discord a while back

server/chat-plugins/datasearch.ts
```diff
@@ -1509,7 +1509,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			}
 			if (matched) continue;
 			if (alts.recoil) {
-				if (dex[move].recoil || dex[move].hasCustomRecoil) matched = true;
+				if (dex[move].recoil) matched = true;
```
scheibo, line 1512: This will cause dex search to also return Struggle/Mindblown/Steelbeam, which I would imagine is the intention, but I could be wrong

data/abilities.ts
```diff
@@ -3170,7 +3170,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		shortDesc: "This Pokemon's attacks with recoil or crash damage have 1.2x power; not Struggle.",
 		onBasePowerPriority: 23,
 		onBasePower(basePower, attacker, defender, move) {
-			if (move.recoil || move.hasCustomRecoil) {
+			if (Array.isArray(move.recoil) || move.recoil === 'crash') {
```
scheibo, line 3173: `Array.isArray` checks are not super pretty, this could be turned into a more pleasantly named helper type guard that does the check and return `is [number, number]`, but I couldnt think of a good name for 'normal' recoil

Zarel: `typeof move.recoil !== 'string'` would do the same thing. But I'm fine with `Array.isArray` too.

server/chat-plugins/datasearch.ts
Zarel: Don't worry, it's definitely the intention.

data/mods/gen3/scripts.ts
Zarel: `git blame`?

Zarel: 6c486409f456e7cddc385324bc3cb915fef399c3

Looks like it was copy/pasted from modern code.

data/mods/gennext/abilities.ts
Zarel: It's minor enough that you can just implement whatever makes it easier to code.

Zarel:
I think properly reviewing this would involve documentation about exactly how Struggle, Crash, and Mind Blown recoil differ. I'll ask the researchers for some.

Zarel:
From what I've been able to gather from @Marty-D and confirmed by @DaWoblefet:

- Regular recoil: boosted by Reckless, prevented by Rock Head and Magic Guard, can KO self
- High Jump Kick and Jump Kick: not prevented by Rock Head
- Mind Blown, Steel Beam, Life Orb: not boosted by Reckless
- Ghost-type Curse's self-damage, Struggle: not prevented by Magic Guard
- Belly Drum, Substitute: can't KO self

Each level of this is slightly less like recoil than the previous level.

I think it makes sense to use Reckless's definition: that only HJK, JK, and regular recoil counts as "recoil" and deserves to be in the `recoil` property. The others should be considered "recoil-like"; maybe call it "self-damage".

Zarel:
It also make sense not to consider HJK/JK damage to be recoil. PS's descriptions take this approach (Reckless's description says "recoil or crash damage"), and it's consistent with Rock Head not protecting against it.

server/chat-plugins/datasearch.ts
scheibo: "its definitely the intention" is why i think it makes sense to have these as one property, because thats what people *intend* when they think about recoil.  but see below about this, id want to come up with something that also makes sense to researchers

scheibo:
Does the game have a concept called 'recoil'? Is there an in game reckless description that differentiates between it and self damage from other means?

I'm OK with recoil and selfDamage properties for everything else (3 booleans can easily be turned into an enum), but my concern is that i do think when people think of recoil they would lump in crash/struggle/life orb/mind blown etc (but perhaps not bellydrum/sub, which i agree are definitely on the same spectrum but i think because they themselves dont cause damage it isnt necessary 'seen' as 'recoil'? though arguably mindblown has 'recoil' even if it only sometimes inflicts damage!)

My main point is - having all these concepts in separate fields (or even 2) tends to make things more error prone (eg. data search probably intended to return more than just strict percentage recoil moves?) and that the complex type forces callers to be explicity and handle all of the cases when making the call?

I agree this is nuanced - I'm not sure. If we want to go with 2 fields thats OK (though as I mentioned, i still thing thats easy to forget?), or a different name than recoil for the one field. I think we're probably both on the same page that 4 fields is too much?

Zarel:
> Does the game have a concept called 'recoil'? Is there an in game reckless description that differentiates between it and self damage from other means?

The in-game Reckless description is "Powers up moves that have recoil damage." It's the same for Gen 4-8, at least according to Bulbapedia. I can't find the Japanese so I can't tell you if it makes it clearer there.

> I'm OK with recoil and selfDamage properties for everything else (3 booleans can easily be turned into an enum), but my concern is that i do think when people think of recoil they would lump in crash/struggle/life orb/mind blown etc (but perhaps not bellydrum/sub, which i agree are definitely on the same spectrum but i think because they themselves dont cause damage it isnt necessary 'seen' as 'recoil'? though arguably mindblown has 'recoil' even if it only sometimes inflicts damage!)

I don't think data representation should be affected by how the community uses the word "recoil". Perhaps it should affect dexsearch, whose "recoil" keyword should probably include all self-damage moves as well.

And anyway, it's not like the community is consistent about it. Bulbapedia's standard is for "recoil" to be as strict as possible, not even including HJK:

https://bulbapedia.bulbagarden.net/wiki/Recoil

> My main point is - having all these concepts in separate fields (or even 2) tends to make things more error prone (eg. data search probably intended to return more than just strict percentage recoil moves?) and that the complex type forces callers to be explicity and handle all of the cases when making the call?

I'm not sure it would be an _error_ to classify non-recoil moves as non-recoil. "Hey, this concept is pretty complicated" isn't exactly an easy thing to embed in a type system, anyway. I think it makes more sense to just put "see also: selfDamage" in the recoil field's jsdoc.

> I agree this is nuanced - I'm not sure. If we want to go with 2 fields thats OK (though as I mentioned, i still thing thats easy to forget?), or a different name than recoil for the one field. I think we're probably both on the same page that 4 fields is too much?

As a sim dev, I don't think it's easy to forget that in addition to recoil, there are a lot of recoil-like effects not specifically named "recoil".

Also, I think if Bulbapedia and Marty agree that crash damage should not be considered recoil, I would also consider it `selfDamage`.

So I think this structure would make a lot of sense:

```
selfDamage?: 'recoil' | 'crash' | 'struggle' | 'halfResidual' | 'half' | 'halfNoKO' | 'selfdestruct' | 'selfdestructUnlessFailed';
recoil?: [number, number]; // only exists if `selfDamage === 'recoil'
```

If I'm understanding correctly, you would prefer

```
selfDamage?: [number, number] | 'crash' | 'struggle' | 'halfResidual' | 'half' | 'halfNoKO' | 'selfdestruct' | 'selfdestructUnlessFailed';
```

(which would also help avoid the situation of inconsistent state between `recoil` and `selfDamage`)

I'm... not a huge fan of this mainly because the word "Recoil" doesn't appear at all. Sadly we don't have checked enums so we could name the `[number, number]` pair. Although I guess we could do:

```
selfDamage?: ['recoil', number, number] | 'crash' | 'struggle' | 'halfResidual' | 'half' | 'halfNoKO' | 'selfDestruct' | 'selfDestructUnlessFailed';
```

I think I'd be reasonably happy with that one?

Zarel:
I'm also considering not using arrays to represent fractions. I think strings might make more sense: `'1/2' | '1/3' | '1/4'`.

Zarel:
(This would also be changed for healing and for multihit ranges, I think.)

scheibo:
> And anyway, it's not like the community is consistent about it. Bulbapedia's standard is for "recoil" to be as strict as possible, not even including HJK:

But Bulbapediadoes seem to include Struggle as recoil???

> ```
> selfDamage?: 'recoil' | 'crash' | 'struggle' | 'halfResidual' | 'half' | 'halfNoKO' | 'selfdestruct' | 'selfdestructUnlessFailed';
> recoil?: [number, number]; // only exists if `selfDamage === 'recoil'
> ```

I like this representation. This is my least favorite enum string presentation (`'Display Case'` > `'TitleCase'` > `'lisp-case'` > `'mixedCase'`, especially with Self-Destruct)  but w/e

scheibo:
> I'm also considering not using arrays to represent fractions. I think strings might make more sense: `'1/2' | '1/3' | '1/4'`.

Not sure about this. Youd have us need to do string parsing? :S Or a table lookup? :S

Zarel:
> But Bulbapedia does seem to include Struggle as recoil???

Shit. Yeah that's complicated. I think I would just add a note in the jsdoc about Struggle.

> I like this representation. This is my least favorite enum string presentation (`'Display Case'` > `'TitleCase'` > `'lisp-case'` > `'mixedCase'`, especially with Self-Destruct)  but w/e

Feel free to use display case.

> Not sure about this. Youd have us need to do string parsing? :S Or a table lookup? :S

Table lookup.

We could also just store the floats `1/2 | 1/3 | 1/4`.

server/chat-plugins/datasearch.ts
scheibo: This is now *just* showing recoil move by their most pedantic definition, not all self damage moves.

data/moves.ts
```diff
@@ -3283,6 +3285,7 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		},
 		secondary: null,
 		target: "normal",
+		selfDamage: "50% (if Ghost)",
```
scheibo, line 3288: you originally had this called "half" / "50%", but that seems a little misleading because its only when Ghost so I expanded the desc

Zarel: oh, I would just remove it completely, just set `move.selfDamage = "50%"` if the user is a Ghost.

scheibo: then its not reflected in the data, only an ActiveMove thing

Zarel: It's not data in the first place, it's implementation detail...

Zarel: If this is for the benefit of search engines, you could set it to `"50%"` and remove it if the user isn't a Ghost, instead.

data/scripts.ts
```diff
@@ -1013,7 +1013,7 @@ export const BattleScripts: BattleScriptsData = {
 		}
 
 
-		if (!didAnything && didAnything !== 0 && !moveData.self && !moveData.selfdestruct) {
+		if (!didAnything && didAnything !== 0 && !moveData.self && !moveData.selfDamage?.startsWith('ko')) {
```
Zarel, line 1016: They all start with `'KO'` not `'ko'`, right? Is this a bug?

Zarel: I do think that if you're doing substring searches, this should probably be its own field instead of rolled into `selfDamage`, though.

data/mods/gen1/moves.ts
```diff
@@ -904,6 +905,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		name: "Substitute",
 		pp: 10,
 		priority: 0,
+		selfDamage: "25% (can't KO)",
```
Zarel, line 908: It's definitely weird that some of these are implemented by the engine, and some of these are scripted.

I guess it's fine to just do this for now and add engine support for them in the future.

data/abilities.ts
```diff
@@ -3170,7 +3170,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		shortDesc: "This Pokemon's attacks with recoil or crash damage have 1.2x power; not Struggle.",
 		onBasePowerPriority: 23,
 		onBasePower(basePower, attacker, defender, move) {
-			if (move.recoil || move.hasCustomRecoil) {
+			if ((move.recoil || move.selfDamage === "Crash")) {
```
Zarel, line 3173:
```suggestion
			if (move.recoil || move.selfDamage === "Crash") {
```

sim/global-types.ts
```diff
@@ -1232,7 +1232,7 @@ interface BattleScriptsData {
 	zMoveTable?: {[k: string]: string};
 	maxMoveTable?: {[k: string]: string};
 	afterMoveSecondaryEvent?: (this: Battle, targets: Pokemon[], pokemon: Pokemon, move: ActiveMove) => undefined;
-	calcRecoilDamage?: (this: Battle, damageDealt: number, move: Move) => number;
+	calcRecoilDamage?: (this: Battle, damageDealt: number, mvoe: Move) => number;
```
Zarel, line 1235:
```suggestion
	calcRecoilDamage?: (this: Battle, damageDealt: number, move: Move) => number;
```

data/mods/gen1/moves.ts
```diff
@@ -881,6 +881,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		desc: "Deals Normal-type damage. If this move was successful, the user takes damage equal to 1/2 the HP lost by the target, rounded down, but not less than 1 HP. This move is automatically used if none of the user's known moves can be selected.",
 		shortDesc: "User loses 1/2 the HP lost by the target.",
+		selfDamage: "Recoil",
```
Zarel, line 884:
```suggestion
```

This should be filled in by the Move constructor, probably. Otherwise it's a lot of redundant data.

data/abilities.ts
```diff
@@ -2666,7 +2666,7 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		desc: "This Pokemon's damaging moves become multi-hit moves that hit twice. The second hit has its damage quartered. Does not affect multi-hit moves or moves that have multiple targets.",
 		shortDesc: "This Pokemon's damaging moves hit twice. The second hit has its damage quartered.",
 		onPrepareHit(source, target, move) {
-			if (move.category === 'Status' || move.selfdestruct || move.multihit) return;
+			if (move.category === 'Status' || move.selfDamage?.startsWith("KO") || move.multihit) return;
```
Zarel, line 2669: Yeah, I've seen `move.selfDamage?.startsWith("KO")` in enough places to think we shouldn't have merged `selfdestruct` and `selfDamage`. Can we split them back out?

scheibo:
I agree, this is probably more trouble than its worth

PR #6611, opened 2020-04-25
Refactor inheritsFrom to changesFrom

KrisXV:
I renamed it from `inheritsFrom` to `changesFrom`, and it is no longer an `ID` to be consistent with upcoming changes to `otherFormes` and `cosmeticFormes`

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -862,11 +862,8 @@ export const BattleFormats: {[k: string]: FormatsData} = {
 				if (baseSpecies.otherFormes) {
 					for (const formeid of baseSpecies.otherFormes) {
 						const forme = dex.getSpecies(formeid);
-						if (!forme.battleOnly) {
-							if (!forme.forme.includes('Alola') && forme.forme !== 'Galar' && forme.baseSpecies !== 'Wormadam') {
-								types = types.concat(forme.types).concat(baseSpecies.types);
-							}
-						}
+						if (!forme.changesFrom) continue;
+						types = types.concat(forme.types).concat(baseSpecies.types);
 					}
 				}
 				if (types.includes(move.type)) return null;
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -849,22 +849,30 @@ export const BattleFormats: {[k: string]: FormatsData} = {
 		desc: "Allows Pok&eacute;mon to use any move that they or a previous evolution/out-of-battle forme share a type with",
 		checkLearnset(move, species, setSources, set) {
 			const restrictedMoves = this.format.restricted || [];
-			if (!restrictedMoves.includes(move.name) && !move.isNonstandard && !move.isMax) {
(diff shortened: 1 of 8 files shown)


Zarel:
(Don't merge this yet, let me review it)

TheImmortal:
Looks good to me.

sim/global-types.ts
```diff
@@ -1088,7 +1088,7 @@ interface SpeciesData {
 	requiredMove?: string;
 	battleOnly?: string | string[];
 	isGigantamax?: string;
-	inheritsFrom?: string;
+	changesFrom?: string;
```
Zarel, line 1091:
```suggestion
	/**
	 * How to obtain this Pokémon? This is like `prevo` but for forme changes.
	 * For in-battle forme changes, use `battleOnly`. If a Pokémon needs to
	 * HOLD an item to keep the forme it will be in `requiredItem`, but if it only
	 * needs a one-time USE of an item, it will be in `evoItem`.
	 */
	changesFrom?: string;
```

(Can you fill out `evoItem` for me?)

Zarel: Also we should update `FORMES.md` with this change.

Zarel: /**
	 * How to obtain this Pokémon? This is like `prevo` but for forme changes.
	 * If a Pokémon needs to HOLD an item to keep the forme, it will be in
	 * `requiredItem`.
	 */
	changesFrom?: string;

data/rulesets.ts
```diff
@@ -862,11 +862,8 @@ export const BattleFormats: {[k: string]: FormatsData} = {
...
 						const forme = dex.getSpecies(formeid);
-						if (!forme.battleOnly) {
-							if (!forme.forme.includes('Alola') && forme.forme !== 'Galar' && forme.baseSpecies !== 'Wormadam') {
-								types = types.concat(forme.types).concat(baseSpecies.types);
-							}
-						}
+						if (!forme.changesFrom) continue;
+						types = types.concat(forme.types).concat(baseSpecies.types);
 					}
 				}
```
Zarel, line 868: This whole block (up until `const dex =`) is implemented very messily and only works because no Pokémon has `changesFrom` and also other formes.

I'd strongly recommend this instead:

```ts
				const dex = this.dex;
				let types;
				if (species.forme || species.otherFormes) {
					const baseSpecies = dex.getSpecies(species.baseSpecies);
					const originalForme = dex.getSpecies(species.changesFrom) || species;
					types = originalForme.types;
					for (const formeid of baseSpecies.otherFormes) {
						const forme = dex.getSpecies(formeid);
						if (forme.changesFrom === originalForme.name) {
							types = types.concat(forme.types);
						}
					}
				} else {
					types = species.types;
				}

				let prevo = species.prevo;
				while (prevo) {
					const prevoSpecies = dex.getSpecies(prevo);
					types = types.concat(prevoSpecies.types);
					prevo = prevoSpecies.prevo;
				}
```

Zarel:
Neat, thanks!

PR #6620, opened 2020-04-26
Add Dynamax Struggle tests

DaWoblefet:
DO NOT MERGE

OK so these tests are failing but presumably not for the right reasons. On sim right now Taunting into a Pokemon with _only_ status moves should prevent that Pokemon from Dynamaxing. My test fails for that case.

The situation I care more about is Dynamax being able to bypass Imprison, and Taunt not stopping Max Guard. The former can cause really ugly and wrong error messages when the target should Struggle (see https://replay.pokemonshowdown.com/gen8customgame-1103346271). The latter can lead to virtual softlocks in Metronome battles. It's partially the client's fault too, I think, since Max Guard Taunt isn't actually disabled in the client.
https://replay.pokemonshowdown.com/gen8metronomebattle-1102183415 (turn 6)

### DIFF
--- a/test/sim/misc/dynamax.js
+++ b/test/sim/misc/dynamax.js
@@ -120,11 +120,11 @@ describe("Dynamax", function () {
 
 	it('Max Guard should be disallowed by Taunt', function () {
 		battle = common.createBattle([[
-			{species: "Feebas", moves: ['splash']},
+			{species: "Feebas", moves: ['splash', 'tackle']},
 		], [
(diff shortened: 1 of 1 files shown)


DaWoblefet:
New test for Max Guard Taunt still has the same problem as before but it at least is based on a better idea now (with Tackle and Splash, Taunted Dynamaxed Feebas should be unable to select Max Guard via Splash, because Max Guard should be stopped by Taunt).

PR #6621, opened 2020-04-26
Match in-game forme names

TheImmortal:
This removes the dash and replaces it with a space to match in-game forme names for the following Pokemon:

Pikachu-Rock Star
Pikachu-Pop Star
Necrozma-Dusk Mane
Necrozma-Dawn Wings
Toxtricity-Low Key


data/pokedex.ts
```diff
@@ -14488,7 +14488,7 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 	},
 	toxtricitylowkeygmax: {
 		num: 849,
-		name: "Toxtricity-Low-Key-Gmax",
+		name: "Toxtricity-Low Key-Gmax",
 		baseSpecies: "Toxtricity",
 		forme: "Gmax",
```
TheImmortal, line 14493: This should be Low Key-Gmax?

KrisXV: Yes imo

Zarel:
I went over this in staff.

In English, "A-B C" is grouped "(A B) C" not "A (B C)". For instance: `low-end car` means `"low end" car`. So `Toxtricity-Low Key-Gmax` would mean `"Toxtricity Low" "Key Gmax"`.

You would need to write "A - B C" to group them "A (B C)". Which would look like: `Toxtricity – Low Key – Gmax`.

Which would still look ugly, I think. The only reason PS uses dashes is because of Smogon. Otherwise, I'd probably rather have it be `Toxtricity (Low Key) (Gmax)`.

We could switch, if enough people are in favor.

The advantage of the dash is that it's easier to type. I think a lot of users actively type "Rotom-Wash" or "Rotom-W", but basically no one is going to type "Rotom (Wash)".

Another option is to put the forme first: "Dusk Mane Necrozma", "Mega Kangaskhan", "Gmax Low Key Toxtricity". I don't like that option because I think the species is most important and should go first, for sorting but also for glancing or skimming through lists to find a Pokémon.

Zarel:
In conclusion, acceptable options:

- `Toxtricity (Low Key) (Gmax)`
- `Toxtricity - Low Key - Gmax`
- `Gmax Low Key Toxtricity`

Unacceptable options:

- `Toxtricity-Low Key-Gmax` ← grammatically incorrect

Zarel:
Also: In English, it's normal and grammatically correct to replace spaces with dashes in dash constructs. For instance:

"This car is high end."

"This is a high-end car."

That's why I don't have a problem with `Toxtricity-Low-Key-Gmax`.

TheImmortal:
What about Toxtricity-Low Key (Gmax)? I think that makes more sense.

TheImmortal:
In the same way, we can make Raticate-Alola (Totem).

Zarel:
Let's discuss in Staff. I want PS and Smogon to match, anyway.

(But the short answer is no, "Toxtricity-Low" and "Key" aren't meaningful words by themselves.)

Zarel:
The Staff discussion with most people preferring `Toxtricity-Low-Key-Gmax` to the alternatives. We did agree that `forme: "Low Key Gmax"` made sense, though, although that'll need more investigation to see if it breaks anything.

PR #6622, opened 2020-04-26
Shell Trap should not power up if hit by a Z- or Max move

urkerab:
Error introduced as part of the fix for PR #6231.

Reported here: https://www.smogon.com/forums/posts/8438387

Relevant replay in the post: https://replay.pokemonshowdown.com/gen8hackmonscup-1101486710 turn 47

### DIFF
--- a/test/sim/moves/shelltrap.js
+++ b/test/sim/moves/shelltrap.js
@@ -32,4 +32,22 @@ describe('Shell Trap', function () {
 		battle.makeChoices('move shelltrap, move splash', 'move tackle, move splash');
 		assert.equal(move.pp, move.maxpp - 2);
 	});
+
+	it('should not Z-power if hit by a Z-move', function () {
+		battle = common.createBattle({}, [
+			[{species: 'Turtonator', moves: ['shelltrap']}],
+			[{species: 'Magikarp', item: 'normaliumz', moves: ['flail']}],
+		]);
+		battle.makeChoices('move shelltrap', 'move flail zmove');
+		assert(battle.log.some(line => line.includes('|Shell Trap|')));
+	});
+
+	it('should not Max if hit by a Max move', function () {
+		battle = common.createBattle({}, [
+			[{species: 'Turtonator', moves: ['shelltrap']}],
+			[{species: 'Magikarp', moves: ['flail']}],
+		]);
+		battle.makeChoices('move shelltrap', 'move flail dynamax');
+		assert(battle.log.some(line => line.includes('|Shell Trap|')));
+	});
 });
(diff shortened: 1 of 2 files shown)


test/sim/moves/shelltrap.js
```diff
@@ -32,4 +32,13 @@ describe('Shell Trap', function () {
...
 		assert.equal(move.pp, move.maxpp - 2);
 	});
+
+	it('should not power up if hit by a Z- or Max move', function () {
+		battle = common.createBattle({}, [
+			[{species: 'Turtonator', moves: ['shelltrap']}],
+			[{species: 'Magikarp', moves: ['flail']}],
+		]);
+		battle.makeChoices('move shelltrap', 'move flail dynamax');
+		assert(battle.log.some(line => line.includes('|Shell Trap|')));
```
Zarel, line 42: ?

`|cant|...|Shell Trap|` and `|move|...|Shell Trap|` both contain `|Shell Trap|`.

Is this because `|move|...|Shell Trap` doesn't? I don't really want to rely on that.

Zarel: It looks like Shell Trap either succeeds or does nothing. Why not just test to see if Magikarp's HP equals its max HP?

urkerab: There's no `cant` in this test?

urkerab: Bah, Git only showed me your first comment.

urkerab: I'm checking whether Shell Trap succeeds or whether it turns in to a successful Max Flare. (But because Max Flare thinks you're using a hacked move, it doesn't set sun, so I've got to test it this way instead.)

Zarel: Oh! I think "power up" is too ambiguous there.

test/sim/moves/shelltrap.js
```diff
@@ -32,4 +32,13 @@ describe('Shell Trap', function () {
 		battle.makeChoices('move shelltrap, move splash', 'move tackle, move splash');
 		assert.equal(move.pp, move.maxpp - 2);
 	});
+
+	it('should not power up if hit by a Z- or Max move', function () {
```
Zarel, line 36:
```suggestion
	it('should not Max if hit by a Max move', function () {
```

urkerab: The bug also affects Z-moves, I was just too lazy to include both tests...

Zarel: I know, I just think clarifying this is more important than mentioning Z moves.

You can mention Z-moves in a comment if you really care? But I don't think it's necessary.

urkerab: ... I've added the other test now, so that should make it clearer.

Zarel:
Neat, thanks

PR #6624, opened 2020-04-26
Update Defog's description

KrisXV:


data/mods/gen7/moves.ts
```diff
@@ -156,6 +156,8 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 	},
 	defog: {
 		inherit: true,
+		desc: "Lowers the target's evasiveness by 1 stage. If this move is successful and whether or not the target's evasiveness was affected, the effects of Reflect, Light Screen, Aurora Veil, Safeguard, Mist, Spikes, Toxic Spikes, Stealth Rock, and Sticky Web end for the target's side, and the effects of Spikes, Toxic Spikes, Stealth Rock, and Sticky Web end for the user's side. Ignores a target's substitute, although a substitute will still block the lowering of evasiveness.",
+		shortDesc: "-1 evasion; clears user and target side's hazards.",
```
Zarel, line 160:
```suggestion
		shortDesc: "-1 evasion; clears hazards on both sides.",
```
I honestly think the new phrasing is clearer, and worth using even when we're not strapped for space.

Marty-D: This is false information since it only clears from both sides if you hit an opponent. If you absolutely must do this, just say "clears hazards" without being specific. shortDescs are already bad enough without adding intentionally misleading stuff.

Zarel: oh ok nvm then

PR #6629, opened 2020-04-29
Remove redundant code from mods/gen1/pokedex.ts

KrisXV:


Zarel:
Wow, so it really was just base stats?

KrisXV:
And magnemite/magneton's types. We could include more minute things like listing Eevee's evos as `Jolteon, Vaporeon, Flareon` if we really wanted, but no other past gen mod file does that for anything else so i dont think it is necessary.

Zarel:
Neat, thanks!

PR #6630, opened 2020-04-29
Correct Self-Destruct and Explosion base power

scheibo:
As confirmed by SadisticMystic, base power cannot be above 255, so
this move data is clearly incorrect and results in downstream users
of the data files such as the Smogon or PS dex or the tooltips
displaying misleading information.


Zarel:
Gen 2's base power overriding is no longer necessary and should be removed. Also the short description should mention it now that the base power doesn't.

PR #6631, opened 2020-04-30
Add a few G-Max Chi Strike tests

DaWoblefet:
G-Max Chi Strike currently sets Focus Energy, when it should actually set an independent stat boost. G-Max Chi Strike boosts crit rate by 1 and is stackable, unlike Focus Energy which boosts by 2 and is not stackable. See research:

https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8440929

My tests here are weird. For one thing, crits are random; normally it'd be easy to test because you'd just give the mon Scope Lens and Super Luck and check if the move crit after a single use since the mon is +3, but since Focus Energy is being set you can't really use that as a test because the old implementation would just reach +4 and you'd crit anyway.

The 1st test checks that it's different from Focus Energy. Some subsequent tests pass when they should not because I'm setting up guaranteed crits. They also might not be accurate at all because these are RNG-based tests and I'm not confident in how to rig the RNG.

### DIFF
--- a/test/sim/moves/gmaxchistrike.js
+++ b/test/sim/moves/gmaxchistrike.js
@@ -43,6 +43,7 @@ describe('G-Max Chi Strike', function () {
 	});
 
 	it('should be copied by Psych Up', function () {
+		// hardcoded RNG seed ensures Magikarp will not crit at +0
 		battle = common.createBattle({gameType: 'doubles', seed: [1, 2, 3, 4]}, [[
 			{species: 'Machamp-Gmax', moves: ['rocksmash', 'sleeptalk']},
 			{species: 'Wynaut', moves: ['sleeptalk']},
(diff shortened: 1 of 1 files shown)


Zarel:
How to rig the RNG: try seeds until you find one that does what you want, then add a comment about what the seed is intended to achieve.

Zarel:
Although I wonder if "always-max" and "always-min" RNGs might make this easier to achieve without needing constant fiddling.

DaWoblefet:
@Zarel I rebased with urkerab's changes and adjusted the tests to account for RNG properly. I ended up only needing to hardcode in seeds for two of the tests.

Zarel:
I still want to make changes to how the RNG works.

Are you hardcoding all cases where the seed matters? You should probably do that.

DaWoblefet:
I added a hardcoded seed to the other tests as well.
> Although I wonder if "always-max" and "always-min" RNGs might make this easier to achieve without needing constant fiddling.

Not entirely sure if this is what you're referring to, but the first 3 tests here check for guaranteed crits; it's only Baton Pass that doesn't and relies on not generating a critical hit "naturally" for the test to pass. The RNG seeds here mostly are just to hopefully ensure that if someone messed up the implementation and regressed back to Focus Energy or something this would do its best to catch that mistake.

DaWoblefet:
Bump @Zarel, was there anything wrong with this after last commit that you'd like me to fix?

Zarel:
There's nothing that needs to be fixed, I just don't like that this sort of test will need to be updated every time the RNG is updated. I'll merge now and deal with that later.

PR #6633, opened 2020-04-30
Rulesets: support 1v1 and 2v2 mods

mia-pi-git:
from: https://www.smogon.com/forums/threads/add-a-1v1-mod-for-room-tours.3655296/
seemed like a pretty reasonable suggestion.


Zarel:
Has this been tested to work?

TheImmortal:
> Has this been tested to work?

Look at the names of the rules. That will answer your question.

But yeah, I'm pretty sure ``teamLength`` isn't inherited so this doesn't work regardless.

mia-pi-git:
will reopen when i make this work properly again.

PR #6634, opened 2020-05-01
Fix default Max/Z power for isMax/isZ moves

scheibo:
Disclaimer: I don't know whether or not this is correct or desirable. I noticed "Let's Snuggle Forever" getting assigned a `zMovePower` of 200 which seemed very odd to me given its base power is `190`. It seems confusing to set these attributes (though I imagine the engine is not looking at them for these moves)

### DIFF
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -828,6 +828,8 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 	readonly isZ: boolean | string;
 	/** How many times does this move hit? */
 	readonly multihit?: number | number[];
+	/** Is this move a Max move? */
+	readonly isMax: boolean | string;
 	/** Max/G-Max move power */
 	readonly gmaxPower?: number;
 	/** Z-move power */
@@ -890,6 +892,7 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 		this.pp = Number(data.pp!);
 		this.noPPBoosts = !!data.noPPBoosts;
 		this.isZ = data.isZ || false;
+		this.isMax = data.isMax || false;
 		this.flags = data.flags || {};
 		this.selfSwitch = (typeof data.selfSwitch === 'string' ? (data.selfSwitch as ID) : data.selfSwitch) || undefined;
 		this.pressureTarget = data.pressureTarget || '';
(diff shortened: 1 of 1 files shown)


Zarel:
I believe if you hack a Z move onto a set and tried to Z-power it up, it would just fail, so this is correct.

Tagging @DaWoblefet to confirm, though.

scheibo:
Would you look at that, the `full-test` I didnt run locally seems to have found an issue:

```
 Error: G-Max Steelsurge doesn't have a gmaxPower
```

So perhaps this isnt desirable for max moves at least (i was less confident about those anyway)?

Zarel:
Do you still have the stack trace for that error? G-Max Steelsurge should definitely not have a `gmaxPower`.

Zarel:
As far as I'm aware, G-Max moves are like Z-status moves; they don't actually exist in the move table.

sim/dex-data.ts
```diff
@@ -939,10 +944,12 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 				}
 			}
 		}
-		if (this.category !== 'Status' && !this.zMovePower) {
+		if (this.category !== 'Status' && !this.zMovePower && !this.isZ) {
```
scheibo, line 947: Still TBD here, should we be giving isZ moves a zMovePower equal to 1, their base power, or something else? When is zMovePower used vs. the moves basePower?

Zarel: Apparently they can't be used at all, so their `zMovePower` should be 0, and this should be detectable by `canZMove`.

Zarel:
```suggestion
		if (this.category !== 'Status' && !this.zMovePower) {
```

sim/dex-data.ts
```diff
@@ -939,10 +944,12 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
 			let basePower = this.basePower;
 			if (Array.isArray(this.multihit)) basePower *= 3;
-			if (!basePower) {
+			if (this.isMax) {
+				this.zMovePower = 1;
```
Zarel, line 951:
```suggestion
			if (this.isMax || this.isZ) {
				this.zMovePower = 0;
```

sim/dex-data.ts
scheibo: Why should the zMovePower be 0 and not just absent?

Zarel: `0` means absent, here.

I suppose we could use `null` instead. I'm not opposed. Although note that the in-game representation is probably `0` since the games don't really have nullable ints.

scheibo: the type is `?`, undefined means absent....

scheibo: Like status moves are already `undefined` here

PR #6640, opened 2020-05-02
Fix metronome calling moves it shouldn't

Dawnofares:


Zarel:
What moves was it calling that it shouldn't? Do you have a replay?

Dawnofares:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/post-8449955 struggle, and I heard at least v-create as well.

Zarel:
Aha! This was the problem:

6c1992b067db6fe86f6

Thanks for bringing it to my attention.

PR #6641, opened 2020-05-02
Prevent user identity updates from sending leave messages for guests

Kaiepi:
Guests usually don't appear in join/leave/rename messages, but it's
possible for a bot to /trn to a guest nick with an assertion from the
login server and trigger a leave message for a guest nick to get sent.
This shouldn't have any consequences for the official client, but can
crash bots that keep state for users.


Zarel:
I think the real bug here is that the login server signs assertions for guest nicks? That should probably be fixed in `session.lib.php`.

Kaiepi:
When I was testing, I was giving the login server a real nick, but giving a guest nick to `/trn`. IIRC the nick given to the login server's included somewhere in the assertion, so maybe that should be verified against the one given to `/trn` instead?

PR #6644, opened 2020-05-02
escapeHTML when notifications blocked

tmagicturtle:
https://www.smogon.com/forums/threads/bug-reports-v3-read-original-post-before-posting.3634749/page-128#post-8451861


Zarel:
You're escaping the wrong thing.

Zarel:
Lines 1914/1919 are the ones needing to be escaped.

tmagicturtle:
Ah woops. Yeah, I just noticed that and was going in to fix. My bad.

PR #6645, opened 2020-05-03
Fix sprite formes for Wi-Fi lotto giveaways

Lusamine:
Gmax and Galarian/Alolan sprites were broken after recent changes.  Thanks for the help, @scheibo!

### DIFF
--- a/server/chat-plugins/wifi.ts
+++ b/server/chat-plugins/wifi.ts
@@ -152,7 +152,7 @@ class Giveaway {
 				if (value.cosmeticFormes) {
 					for (const forme of value.cosmeticFormes.map(toID)) {
 						if (text.includes(forme)) {
-							spriteid += '-' + forme.substr(key.length);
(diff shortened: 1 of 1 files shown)


server/chat-plugins/wifi.ts
```diff
@@ -150,15 +150,15 @@ class Giveaway {
 			for (const [key, value] of mons) {
 				let spriteid = value.spriteid;
 				if (value.cosmeticFormes) {
-					for (const form of value.cosmeticFormes) {
-						if (text.includes(form)) {
-							spriteid += '-' + form.substr(key.length);
+					for (const forme of value.cosmeticFormes.map(toID)) {
+						if (text.includes(forme)) {
+							spriteid += '-' + forme.substr(key.length);
```
Zarel, line 155:
```suggestion
							spriteid += '-' + forme.slice(key.length);
```

(We're deprecating `substr`)

PR #6652, opened 2020-05-04
Group zMove and maxMove attributes

scheibo:
This mostly serves to provide cleaner and more consistent field
naming. maxMove currently doesn't have boosts or effects to group
together but who knows what will be thrown at us via DLC, and being
symmetrical with zMoves is a nice.

### DIFF
--- a/data/mods/ssb/moves.ts
+++ b/data/mods/ssb/moves.ts
@@ -3456,7 +3456,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 				return null;
 			},
 			onHit(target, source, move) {
-				if (move.isZPowered && move.flags['contact']) {
+				if (move.isZOrMaxPowered && move.flags['contact']) {
 					this.heal(target.baseMaxhp / 4, target, target);
 				}
 			},
--- a/data/mods/ssb/moves.ts
+++ b/data/mods/ssb/moves.ts
@@ -25,8 +25,8 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		// Other useful things
 		noPPBoosts: true, // add this to not boost the PP of a move, not needed for Z moves, dont include it otherwise
 		isZ: "crystalname", // marks a move as a z move, list the crystal name inside
-		zMoveEffect: '', // for status moves, what happens when this is used as a Z move? check data/moves.js for examples
-		zMoveBoost: {atk: 2}, // for status moves, stat boost given when used as a z move
+		zMove: {effect: '', // for status moves}, what happens when this is used as a Z move? check data/moves.js for examples
+		zMove: {boost: {atk: 2}, // for status moves}, stat boost given when used as a z move
 		critRatio: 2, // The higher the number (above 1) the higher the ratio, lowering it lowers the crit ratio
 		drain: [1, 2], // recover first num / second num % of the damage dealt
 		heal: [1, 2], // recover first num / second num % of the target's HP
@@ -132,7 +132,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		},
 		target: "normal",
 		type: "Electric",
-		zMovePower: 200,
+		zMove: {basePower: 200},
 	},
 	// Aeonic
 	shitpost: {
@@ -1372,7 +1372,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "self",
 		type: "Ground",
-		zMoveEffect: 'healhalf',
+		zMove: {effect: 'healhalf'},
 	},
 	// E4 Flint
 	fangofthefireking: {
@@ -1471,7 +1471,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "normal",
 		type: "Fighting",
-		zMovePower: 175,
+		zMove: {basePower: 175},
 	},
 	// eternally
 	quack: {
@@ -1952,7 +1952,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "foeSide",
 		type: "Normal",
-		zMoveBoost: {def: 1},
+		zMove: {boost: {def: 1}},
 	},
 	// HoeenHero
 	scriptedterrain: {
@@ -2641,7 +2641,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		},
 		target: "normal",
 		type: "Fire",
-		zMovePower: 160,
+		zMove: {basePower: 160},
 	},
 	// MacChaeger
 	naptime: {
@@ -2697,7 +2697,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		},
 		target: "self",
 		type: "Fairy",
-		zMoveEffect: 'clearnegativeboosts',
+		zMove: {effect: 'clearnegativeboosts'},
 	},
 	// Mad Monty ¾°
 	llamacide: {
@@ -2894,7 +2894,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "normal",
 		type: "Normal",
-		zMoveEffect: 'heal',
+		zMove: {effect: 'heal'},
 	},
 	// Meicoo
 	scavengesu: {
@@ -3126,7 +3126,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "normal",
 		type: "Water",
-		zMoveEffect: "boostreplacement",
+		zMove: {effect: "boostreplacement"},
 	},
 	// OM
 	omboom: {
@@ -4256,7 +4256,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		secondary: null,
 		target: "normal",
 		type: "Steel",
-		zMovePower: 160,
+		zMove: {basePower: 160},
 		contestType: "Cool",
 	},
 	// Tony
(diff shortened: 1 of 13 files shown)


data/mods/ssb/moves.ts
```diff
@@ -25,8 +25,8 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 		// Other useful things
 		noPPBoosts: true, // add this to not boost the PP of a move, not needed for Z moves, dont include it otherwise
 		isZ: "crystalname", // marks a move as a z move, list the crystal name inside
-		zMoveEffect: '', // for status moves, what happens when this is used as a Z move? check data/moves.js for examples
-		zMoveBoost: {atk: 2}, // for status moves, stat boost given when used as a z move
+		zMove: {effect: '', // for status moves}, what happens when this is used as a Z move? check data/moves.js for examples
+		zMove: {boost: {atk: 2}, // for status moves}, stat boost given when used as a z move
```
Zarel, line 29:
```suggestion
		zMove: {effect: ''}, // for status moves, what happens when this is used as a Z move? check data/moves.js for examples
		zMove: {boost: {atk: 2}}, // for status moves, stat boost given when used as a z move
```

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
-					this.gmaxPower = 100;
+					this.maxMove!.basePower = 100;
 				} else {
-					this.gmaxPower = 90;
+					this.maxMove!.basePower = 90;
 				}
 			}
 		}
-		if (this.category !== 'Status' && !this.zMovePower && !this.isZ && !this.isMax) {
+		if (this.category !== 'Status' && !this.zMove?.basePower && !this.isZ && !this.isMax) {
```
Zarel, line 954:
```suggestion
		if (this.category !== 'Status' && !this.zMove && !this.isZ && !this.isMax) {
```

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
 			let basePower = this.basePower;
+			this.zMove = this.zMove || {} as MoveData['zMove'];
```
Zarel, line 956:
```suggestion
			this.zMove = {};
```

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
 			if (Array.isArray(this.multihit)) basePower *= 3;
 			if (!basePower) {
-				this.zMovePower = 100;
+				this.zMove!.basePower = 100;
```
Zarel, line 959:
```suggestion
				this.zMove.basePower = 100;
```

Zarel: (etc for the others)

(also do the same thing for the max move fill-in)

data/mods/ssb/scripts.ts
```diff
@@ -203,7 +203,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 		}
 		zMove = this.dex.getActiveMove(this.zMoveTable[move.type]);
 		// @ts-ignore
```
Zarel, line 205:
```suggestion
```

data/mods/ssb/scripts.ts
```diff
@@ -203,7 +203,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
...
-		zMove.basePower = move.zMovePower;
+		zMove.basePower = move.zMove?.basePower;
```
Zarel, line 206:
```suggestion
		zMove.basePower = move.zMove!.basePower;
```

data/mods/ssb/moves.ts
scheibo: Lmao how did this compile? Good ol regex find and replace

sim/dex-data.ts
scheibo: Isn't this incorrect? I guess I dont know how zmoves work, but do not some have both a boost/effect and base power?

sim/dex-data.ts
scheibo: it doesnt work for the maxMove fill in because maxMove's basePower is required.

data/mods/ssb/moves.ts
Zarel: This entire block is in `/* ... */` which is how it compiled.

sim/dex-data.ts
Zarel: There are unique z-moves that have both, but they're unique z-moves with their own entries in the Moves table. `move.zMove` is about generic z-moves, which for status moves are a `boost`/`effect` on top of their existing effect, and for physical/special moves are a high-damage (defined by `basePower`) Protect-breaking move with no other effect.

If a move already has generic Z-move data defined, this function should not overwrite or supplement it.

sim/dex-data.ts
Zarel: I should probably have been more specific. I'll directly comment what needs to be changed about `maxMove`s.

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 		this.stab = data.stab || undefined;
 		this.volatileStatus = typeof data.volatileStatus === 'string' ? (data.volatileStatus as ID) : undefined;
 
-		if (this.category !== 'Status' && !this.gmaxPower) {
+		if (this.category !== 'Status' && !this.maxMove?.basePower) {
```
Zarel, line 914:
```suggestion
		if (this.category !== 'Status' && !this.maxMove && this.id !== 'struggle') {
```

The only difference between `!this.maxMove` and `!this.maxMove?.basePower` is that the latter will be truthy if `maxMove.basePower` is `0`. This is undesirable; if a move explicitly specifies a `basePower` of `0` for whatever reason, we should not overwrite it.

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
+			this.maxMove = this.maxMove || {} as MoveData['maxMove'];
```
Zarel, line 915:
```suggestion
			this.maxMove = {};
```

With the `!this.maxMove` change, there's no need for this to be this complicated.

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
...
 			if (this.isMax || this.isZ) {
-				this.gmaxPower = 1;
+				this.maxMove!.basePower = 1;
```
Zarel, line 917:
```suggestion
				this.maxMove.basePower = 1;
```

(Also do this for all the other `maxMove!.basePower` assignments.)

I don't know if this was necessary before, but it won't be necessary after the above changes.

sim/dex-data.ts
Zarel: (With the `!this.zMove` change, there's no need for this to be more complicated.)

sim/dex-data.ts
Zarel: (If `maxMove.basePower` should never be `0`, that should be enforced in `test/sim/data`, not here.)

sim/dex-data.ts
scheibo: This cast is still needed because `basePower` is not optional.

data/mods/ssb/moves.ts
scheibo: Gotcha

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 		this.stab = data.stab || undefined;
 		this.volatileStatus = typeof data.volatileStatus === 'string' ? (data.volatileStatus as ID) : undefined;
 
-		if (this.category !== 'Status' && !this.gmaxPower) {
+		if (this.category !== 'Status' && !this.maxMove && this.id !== 'struggle') {
+			this.maxMove = {} as {basePower: number};
```
Zarel, line 915:
```suggestion
			this.maxMove = {basePower: null!};
```

sim/dex-data.ts
```diff
@@ -905,70 +911,72 @@ export class Move extends BasicEffect implements Readonly<BasicEffect & MoveData
 		this.stab = data.stab || undefined;
 		this.volatileStatus = typeof data.volatileStatus === 'string' ? (data.volatileStatus as ID) : undefined;
 
-		if (this.category !== 'Status' && !this.gmaxPower) {
+		if (this.category !== 'Status' && !this.maxMove && this.id !== 'struggle') {
+			this.maxMove = {basePower: 1};
 			if (this.isMax || this.isZ) {
-				this.gmaxPower = 1;
+				this.maxMove.basePower = 1;
```
Zarel, line 917:
```suggestion
				// do nothing
```

Zarel:
Do we need separate flags for `isZPowered` and `isMaxPowered`? It looks like they're mostly identical, and any differences can be treated as inter-generational differences anyway.

scheibo:
I do not know. What would you call the flag? `isPoweredUp` `isSpecialPowered` ick. I dont know

Zarel:
`isZOrMaxPowered`

Zarel:
(If you don't want do to that, you can merge this and I'll just do it in a separate commit.)

sim/battle.ts
```diff
@@ -2206,7 +2206,7 @@ export class Battle {
 		// Final modifier. Modifiers that modify damage after min damage check, such as Life Orb.
 		baseDamage = this.runEvent('ModifyDamage', pokemon, target, move, baseDamage);
 
-		if ((move.isZPowered || move.maxPowered) && target.getMoveHitData(move).zBrokeProtect) {
+		if ((move.isZOrMaxPowered || move.isZOrMaxPowered) && target.getMoveHitData(move).zBrokeProtect) {
```
scheibo, line 2209: lol will fix

urkerab:
Copycat and Grudge (not yet implemented) work on Dynamax powered moves, but they don't work on Z-powered moves. So what's the correct test here, `if (move.isMax && move.isZOrMaxPowered) move = this.dex.getActiveMove(move.baseMove);`?

PR #6655, opened 2020-05-05
Fix Glitzer Popping bug

tmagicturtle:


Instructing:
Can you make sure that Glitch Out doesnt fall prey to the same issue?

PR #6661, opened 2020-05-07
Add room categories

KrisXV:
The categories themselves and how to sort them on the rooms list is still in discussion


KrisXV:
I'm taking this out of a draft state since the code already works fully, there just needs to be discussion on the category names/how we decide to sort them.

PR #6662, opened 2020-05-07
Unskip passing tests

DaWoblefet:
This removes the skip from two passing tests (Shell Trap was failing due to a syntax error that I corrected, and Flash Fire's oldgen mechanic was actually working). I also simplified a failing Disguise critical hit test.

It also adds notes that the socket tests are failing due to a syntax error rather than actually failing tests, and that dex is failing due to the relative paths of the files it's trying to find being wrong. I don't feel confident enough to fix those later two, but I did add comments for them.

### DIFF
--- a/test/server/sockets.js
+++ b/test/server/sockets.js
@@ -3,6 +3,7 @@
 const assert = require('assert').strict;
 const cluster = require('cluster');
 
+// These tests are failing due to syntax errors, not necessarily failing tests
 describe.skip('Sockets', function () {
 	const spawnWorker = () => (
 		new Promise(resolve => {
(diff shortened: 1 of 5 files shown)


test/server/sockets.js
```diff
@@ -3,6 +3,7 @@
 const assert = require('assert').strict;
 const cluster = require('cluster');
 
+// These tests are failing due to syntax errors, not necessarily failing tests
```
Zarel, line 6: This test has never worked, and was not updated to the latest version of Sockets anyway. That's probably what should be noted.

Zarel:
I don't like the mod loader comment, either, but I'll deal with that soonish.

PR #6664, opened 2020-05-07
update formats.js

DarkPhoenix911:
Added code for National Dex Monotype Tier.  This will be version 1.0, to be updated when the council votes on bans and other changes.


TheImmortal:
It's a side meta. Side metas don't get on PS unless some special monthly format.

PR #6666, opened 2020-05-08
Chat plugins: support a way to set custom avatars

mia-pi-git:
Continuation of #5515, but i deleted my old fork, so i can't just use that PR.
Note for those looking that this is just for side servers.

### DIFF
--- a/server/chat-plugins/customavatars.ts
+++ b/server/chat-plugins/customavatars.ts
@@ -30,8 +30,7 @@ export const avatarManager = new class {
 					}
 					// weird bug with FSPath that doesn't like this, so normal fs is required.
 					const stream = fs.createWriteStream(`${this.dir}${name}.png`);
-					response.pipe(stream);
-					stream.on('finish', () => {
+					response.pipe(stream).on('finish', () => {
 						resolve(true);
 					});
 				});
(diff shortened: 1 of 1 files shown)


Zarel:
I thought this was drafted, but it's possible that it was finished sometime between our incredibly busy phase where I didn't have time to look at it. In any case, it could at least use an update to `Net`.

server/chat-plugins/customavatars.ts
```diff
@@ -0,0 +1,152 @@
...
+
+import {FS} from '../../lib/fs';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as https from 'https';
+import * as http from 'http';
+import * as url from 'url';
+
+
+export const avatarManager = new class {
```
Zarel, line 14:
```suggestion
export const AvatarManager = new class {
```

DaWoblefet:
Superseded by https://github.com/smogon/pokemon-showdown/pull/8563 which is live.

Zarel:
Not completely superseded. There's one more feature I'd expect side servers to want, which is a way to grab an avatar from a URL and put it in `config/avatars/`.

Zarel:
Which this code _does_ provide!

PR #6667, opened 2020-05-09
Fix movesearch with max flag

DaWoblefet:
Returns this now:
image

instead of this:
image

I didn't make it return G-Max moves since there's a separate flag already for them.


KrisXV:
the gmax flag doesnt work btw

DaWoblefet:
? gmax flag works just fine both on main and after this change

Main:
image

This branch:
image

KrisXV:
That’s peculiar. It wasn’t working for me. Carry on then.

PR #6668, opened 2020-05-09
Allow leaders+ to exportinputlog directly

DaWoblefet:
@asgdf  brought this up recently to me. This is purely a convenience command so it doesn't have to be done via replays and can instead grab the inputlog natively within Showdown. I'm not sure if this needs policy hashed out or anything, but the only difference I'm seeing is that it's being done through Showdown and not through replays. Leaders still have access either way. Players still have to explicitly request permission to obtain the inputlog from a battle, even if a leader+ grabbed the inputlog first. As a sanity check, it also still logs to the battle that the leader+ exported the input log.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -815,7 +815,15 @@ export const commands: ChatCommands = {
 			return;
 		}
 		if (!this.can('exportinputlog', null, room)) return;
-		if (!battle.allowExtraction[user.id]) {
+		if (user.can('forcewin')) {
+			if (!battle.inputLog) return this.errorReply('No input log found.');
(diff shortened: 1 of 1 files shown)


Zarel:
Do leaders have inputlog access in replays? It's probable that they shouldn't.

scheibo:
Leaders can ?manage a replay

Zarel:
Yeah, I figured that from the PR description.

I think we might want to re-evaluate whether leaders should be able to manage a replay. I could see it going either way, tbh.

Zarel:
Well, I haven't heard of them being abused and I haven't heard complaints. I suppose this is okay for now.

PR #6669, opened 2020-05-09
Add alt form index data to dex

scheibo:
Numbers obtained from kwsch/PKHeX's FormConverter.cs (thanks Kurt).
Thanks to @Lusamine (Anubis) for help with the peculiarities.

### DIFF
--- a/data/mods/gen6/pokedex.ts
+++ b/data/mods/gen6/pokedex.ts
@@ -7,6 +7,10 @@ export const BattlePokedex: {[k: string]: ModdedSpeciesData} = {
 		inherit: true,
 		baseStats: {hp: 60, atk: 85, def: 69, spa: 65, spd: 79, spe: 80},
 	},
+	pikachu: {
+		inherit: true,
+		formeOrder: ["Pikachu", "Pikachu-Rock-Star", "Pikachu-Belle", "Pikachu-Pop-Star", "Pikachu-PhD", "Pikachu-Libre", "Pikachu-Cosplay"],
+	},
 	dugtrio: {
 		inherit: true,
 		baseStats: {hp: 35, atk: 80, def: 50, spa: 50, spd: 70, spe: 120},
--- a/data/pokedex.ts
+++ b/data/pokedex.ts
@@ -41,6 +41,7 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 		evoLevel: 32,
 		eggGroups: ["Monster", "Grass"],
 		otherFormes: ["Venusaur-Mega"],
+		formeOrder: ["Venusaur", "Venusaur-Mega"],
 	},
 	venusaurmega: {
 		num: 3,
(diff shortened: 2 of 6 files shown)


scheibo:
I need to add tests (confirm no gaps or dupes in forme numbers for each gen), but I don't know how you would like me to handle numbering the 'cosmetic' formes. Please advise

Zarel:
I think I would rather just have the `otherFormes` array in index number order, for the very niche use-cases where it's needed at all.

scheibo:
@monsanto felt this was too fragile, and its also not possible given how PS does not map formes correctly to the formes on the cartridge. Natdex/keeping past and future gen formes around, FORMES.md, arbitrarily not supporting certain forms (Vivillon, Minior-Meteor colors) etc makes indexes into `otherFormes` not work

Zarel:
That _would_ pose a problem for Pokémon with mixed cosmetic/non-cosmetic formes, most notably Vivillon and Floette.

I would probably merge cosmetic and non-cosmetic formes into the `otherFormes` array.

scheibo:
Still impossible, please see the numbers on the indexes in the pokedex.ts and the comments Ive left

Zarel:
PS can put nonexistent formes last, so the formes that actually exist are correctly indexed.

As far as I'm aware, PS currently only diverges from cartridge forme lists in two ways: we have Gmax formes, and we don't have formes for event abilities. We are planning to add formes for event abilities, and Gmax formes can be after all other formes.

scheibo:
- natdex means Pikachu formes are incorrect (gen 6 pikachus are currently in the same other formes array as later gens, there are 2 different pikachu-forme0 depending on the gen)
- you need to add in minior meteor colored  formes and make minior meteor the base because minior meteor red is forme 0, minior red is forme 7
- you need to reorder vivillon to have its baseForme be Icy Snow, not Meadow which is forme 6
- pokestar formes are not formes

Zarel:
Anyway I don't need to block on major data rearrangement. This is fine for now.

scheibo:
> Anyway I don't need to block on major data rearrangement. This is fine for now

OK thanks, Im still not sure how to assign forme numbers to cosmetic formes before things get rearranged though?

Zarel:
Alternatively, we could have a `formeOrder` array, which just lists all the formes by speciesName in in-game order?

scheibo:
> Alternatively, we could have a `formeOrder` array (on the base species, next to `otherFormes`), which just lists all the formes by speciesName in in-game order? With duplicates for Minior-Meteor and event formes, for now?

I will do this in the morning. I believe i will need to change the gen 6 pokedex as well for pikachu. otherwise i will leave in formes that dont impact actual forme numbers (leaving LGPE's starter pikachu affects Pikachu-Gmax's forme number, but gmax isnt a forme so its unimportant)

monsanto:
I was planning to tackle this soonish, pleasant surprise to wake up early and see it mostly done. Thanks all.

`formeOrder` sounds fine for now as it does not require cosmetics to have pokedex entries. Good idea

Zarel:
Do you want this merged now? It looks good to me.

PR #6670, opened 2020-05-09
Typescript info

KrisXV:


server/chat-commands/room-settings.ts
```diff
@@ -1299,6 +1299,52 @@ export const commands: ChatCommands = {
 	removeroomaliashelp: [
 		`/removeroomalias [alias] - removes the given room alias of the room the command was entered in. Requires: & ~`,
 	],
+
+	resettierdisplay: 'roomtierdisplay',
+	roomtierdisplay(target, room, user, connection, cmd) {
```
Zarel, line 1304: Was this a mistake? It should probably go in a different PR than a TS refactor.

KrisXV: I asked you about this in #sim-dev and you said it was fine, unless I misinterpreted something.

server/chat-commands/info.ts
```diff
@@ -190,21 +190,21 @@ const commands = {
 		}
 
 		const gameRooms = [];
-		for (const room of Rooms.rooms.values()) {
-			if (!room.game) continue;
-			if ((targetUser.id in room.game.playerTable && !targetUser.inRooms.has(room.roomid)) ||
-				(room.auth && room.auth[targetUser.id] === Users.PLAYER_SYMBOL)) {
-				if (room.isPrivate && !canViewAlts) {
+		for (const r of Rooms.rooms.values()) {
```
Zarel, line 193: ?

Why was this renamed?

Zarel: Oh, variable shadowing.

How about `curRoom`?

server/chat-commands/info.ts
```diff
@@ -213,8 +213,8 @@ const commands = {
 			if (punishments.length) {
 				buf += `<br />Room punishments: `;
 
-				buf += punishments.map(([room, punishment]) => {
-					const [punishType, punishUserid, expireTime, reason] = punishment;
+				buf += punishments.map(([r, p]) => {
```
Zarel, line 216: ditto here (`curRoom`, `punishment`)

KrisXV: punishment is a shadowed variable too, any ideas for that

server/chat-commands/info.ts
```diff
@@ -483,7 +489,7 @@ const commands = {
 		const targetId = toID(target);
 		if (!targetId) return this.parse('/help data');
 		const targetNum = parseInt(target);
-		if (!isNaN(targetNum) && '' + targetNum === target) {
+		if (!isNaN(targetNum) && String(targetNum) === target) {
```
Zarel, line 492:
```suggestion
		if (!isNaN(targetNum) &&`${targetNum}` === target) {
```

server/chat-commands/info.ts
```diff
@@ -517,30 +522,36 @@ const commands = {
...
+			let details: {[k: string]: string} = {};
 			switch (newTarget.searchType) {
 			case 'nature':
 				const nature = Dex.getNature(newTarget.name);
-				buffer += "" + nature.name + " nature: ";
+				buffer += `${nature.name} nature: `;
 				if (nature.plus) {
-					const statNames = {'atk': "Attack", 'def': "Defense", 'spa': "Special Attack", 'spd': "Special Defense", 'spe': "Speed"};
-					buffer += "+10% " + statNames[nature.plus] + ", -10% " + statNames[nature.minus] + ".";
+					const statNames: {[k in StatNameExceptHP]: string} = {
```
Zarel, line 531:
```suggestion
					const statNames = {
```

This annotation doesn't actually add anything...

KrisXV: Have you tested this?

server/chat-commands/room-settings.ts
Zarel: I don't remember this at all. Can you move it to a separate PR now, though?

server/chat-commands/info.ts
Zarel: `curPunishment`

server/chat-commands/info.ts
Zarel: No. What error comes up?

KrisXV: Nothing, I had it confused with another line

KrisXV:
Done @Zarel

server/chat-commands/info.ts
```diff
@@ -2396,7 +2496,7 @@ const commands = {
 		// XXX: target is trimmed by Chat#splitMessage. Let's not add another
 		// awful hack like ! or help command keys for whether or not the target
 		// is raw for now.
-		target = this.message.substr(this.cmdToken.length + this.cmd.length + +this.message.includes(' ')).trimEnd();
+		target = this.message.substr(this.cmdToken.length + this.cmd.length + +this.message.includes(' ')).trimRight();
```
Zarel, line 2499: Why was this changed? According to MDN, the standard name is `trimEnd`.

server/chat-commands/info.ts
```diff
@@ -2198,12 +2288,16 @@ const commands = {
 
 		Config.potd = target;
 		// TODO: support eval in new PM
-		Rooms.PM.eval('Config.potd = \'' + toID(target) + '\'');
+		(Rooms.PM as any).eval('Config.potd = \'' + toID(target) + '\'');
```
Zarel, line 2291: Um, does this work at all? I would just comment it out if it doesn't...

server/chat-commands/info.ts
KrisXV: according to VS code with an up-to-date package, trimEnd doesnt exist

server/chat-commands/info.ts
KrisXV: i dont know, we havent used POTD in literal ages

KrisXV: Just tested, can confirm it just causes crashes, so i commented it out

server/chat-commands/info.ts
```diff
@@ -2198,12 +2288,16 @@ const commands = {
 
 		Config.potd = target;
```
Zarel, line 2289:
```suggestion
		return this.errorReply(`Sorry, PotD is currently unsupported.`);

		Config.potd = target;
```

Zarel: I hope there's no lint rule against this. We might need to comment out _everything_ after the error.

KrisXV: A lint rule against.... putting inaccessible code in? I dont think this has never NOT been a lint rule

KrisXV: I commented out the rest of the code and added the early return.

Zarel:
Why did you end up deleting it entirely?

Zarel:
I think it's useful to at least have an error message representing:

1. one day I will make it work again

2. if anyone tries to use it, they'll realize what happened to it

Zarel:
Whatever, it's not really that big of a deal to me and I don't like leaving refactors uncommitted for so long.

KrisXV:
I deleted it because it's still in the git history for when we decide to look into fixing it.

Zarel:
Digging for it in the Git history (when there's not even a line to blame) is a _lot_ harder than looking at a comment.

Zarel:
That's the advantage of the error message - it's a line that you can use Git Blame on to find the old commit.

PR #6671, opened 2020-05-10
Support commands that let you operate on a room without joining it

scheibo:
Full disclosure: I don't know what the implications of this are, I just know @monsanto and I are frustrated at not being able to operate on a room without joining it, and line 492 appears to be that check. Is this safe? Is this the best way to do it? /shrug

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -494,7 +494,7 @@ export class CommandContext extends MessageContext {
 
 		if (typeof commandHandler === 'function' && requireGlobalCommand) {
 			const baseCmd = typeof curCommands[cmd] === 'string' ? curCommands[cmd] : cmd;
-			if (!curCommands['!' + baseCmd]) {
+			if (!curCommands['!' + baseCmd] && !this.user.can('bypassall')) {
 				return '!';
 			}
 		}


server/chat.ts
```diff
@@ -494,7 +494,7 @@ export class CommandContext extends MessageContext {
 
 		if (typeof commandHandler === 'function' && requireGlobalCommand) {
 			const baseCmd = typeof curCommands[cmd] === 'string' ? curCommands[cmd] : cmd;
-			if (!curCommands['!' + baseCmd]) {
+			if (!curCommands['!' + baseCmd] && !this.user.can('bypassall')) {
```
Zarel, line 497: This seems like the wrong line to do it in.

`"!cmdname": true,` is an opt-in sigil meaning "this command won't crash if you use it outside of a chatroom". If there's a specific command you want to use but can't, just add that sigil.

scheibo:
Does this PR not allow admins to bypass the check for using commands in certain places? Because if so, it does exactly what it was intended to do.

I think the main problem is that being able to run without a user is opt in for commands, and its easy to forgot. While I agree that all the commands that can be run without the user in the room should have a `!` in front of them or whatever to opt-in, the problem is that the majority of them do not, and I should not be blocked on needing to fix that command first. 

I'm also not sure why *any* command actually needs the user to be in said room? Can you give me some examples of commands where this is important? I feel it is important in the minority of cases?

Zarel:
A lot of commands would crash if `room` were `null`. The opt-in is just a confirmation "I promise I won't crash".

We could probably do a pass over all our commands to accept a `null` room argument. That wouldn't take _too_ long.

scheibo:
> A lot of commands would crash if room were null. The opt-in is just a confirmation "I promise I won't crash"

But why would room be null? If I do a `Chat.parse('/foo', room, me, me.connections[0])` there is a room, but it wont let me because `me` is not in the `room` i passed in. Also, `this.room` is very much not null, which is what is being checked when the error message is being decided on.

> We could probably do a pass over all our commands to accept a null room argument. That wouldn't take too long.

I dont deny that this is a good idea, just that it doesnt seem to solve the problem of future commands inevitably also not opting in

Zarel:
> But why would room be null? If I do a `Chat.parse('/foo', room, me, me.connections[0])` there is a room, but it wont let me because `me` is not in the `room` i passed in. Also, `this.room` is very much not null, which is what is being checked when the error message is being decided on.

I still have a vague inkling of "I don't think this code does what you think it does".

To be clear, your change lets admins bypass the "this command can't be used in PMs" check, allowing commands like `/privateroom` to be used in PMs, where `room` is `null`. This will make those commands crash.

It's sounding more and more like you're trying to do something else?

> (re: changing all commands to accept `null` room) I dont deny that this is a good idea, just that it doesnt seem to solve the problem of future commands inevitably also not opting in

The idea is that we would change the function signature from `room: Room` to `room: Room | null` and then make all commands support being used in PMs going forward, removing the opt-in.

scheibo:
> I still have a vague inkling of "I don't think this code does what you think it does".
> 
> To be clear, your change lets admins bypass the "this command can't be used in PMs" check, allowing commands like `/privateroom` to be used in PMs, where `room` is `null`. This will make those commands crash.

I mean, I admins can already eval in anything they want and potentially cause crashes, and because the commands wouldnt let me do what I wanted I had to manually do it for absolutely no reason. `bypassall` should literally let you bypass all, IMO. I would support bypasses being opt-in, like a `/force cmd arg1 arg2 etc` command that only *then* triggers the bypass because that feels safer,. In almost all cases (other than the recent crash that shows up in the dev logs and the time i turned room.auth into a boat) I test out all my evals on a local server anyway where I dont care if it crashes, but having to inline the body of two commands that I should just have been able to call because "I know what I'm doing" is what prompted this.

> The idea is that we would change the function signature from `room: Room` to `room: Room | null` and then make all commands support being used in PMs going forward, removing the opt-in.

I did not read your original response as this, I absolutely think this is a good idea

Zarel:
> I would support bypasses being opt-in, like a `/force cmd arg1 arg2 etc` command that only _then_ triggers the bypass because that feels safer.

Works for me. This commit doesn't make it opt-in, though.

Zarel:
Also, I stand by my change request of "PR does not do what it claims to do". This PR is not about commands that operate on a room at all, it's about allowing admins to use commands that don't operate on rooms in PMs.

scheibo:
>  it's about allowing admins to use commands that don't operate on rooms in PMs.

I mean, I agree it is poorly titled. But I dont think my intention was to allow admins "to use commands that don't operate on rooms in PMs." I'm just trying to run a command in room A (staff or upper staff) that operates on room B. 

---

I will close this PR though - and to be clear, its not because I'm salty or anything, I was originally stating in the PR description I wasn't sure about this approach. PRs to me are the exact same thing as bug reports (they even use the same ID namespace), the fact that theres token code attached is unimportant. I think we have had a productive discussion with two followups:

- change the type to `Room | null` and audit commands to make them work
- create an opt-in `/force` command or similar for admins to use to bypass gating logic that they have ruled to be unnecessary for their use case

I think both of these are substantially better approaches than the code in this PR. 

Thank you

Zarel:
> I'm just trying to run a command in room A (staff or upper staff) that operates on room B.

See, that's what I was confused about. I don't understand how this PR would let you do that. Do you have a specific example?

PR #6675, opened 2020-05-11
Support changing /data's tier in /roomsettings

KrisXV:


server/rooms.ts
```diff
@@ -111,6 +111,7 @@ export abstract class BasicRoom {
 	gtsga: GTSGiveaway | null;
 	toursEnabled: '%' | boolean;
 	tourAnnouncements: boolean;
+	tierDisplay: string;
```
Zarel, line 114:
```suggestion
	dataCommandTierDisplay: string;
```

I've started preferring longer names because I'd rather not have people see this definition and not know what it means.

Zarel: (A jsdoc comment would work too, but considering this variable is rarely ever used, I'd rather just rename the variable.)

KrisXV: I do not know what you mean by jsdoc. You keep using it when you comment on my pull requests and I've never known what that means.

scheibo: /** foo */
https://jsdoc.app/about-getting-started.html

KrisXV:
bump @zarel

server/chat-commands/info.ts
```diff
@@ -541,7 +541,9 @@ export const commands: ChatCommands = {
 				if (format?.onModifySpecies) {
 					pokemon = format.onModifySpecies.call({dex} as Battle, pokemon) || pokemon;
 				}
-				let displayedTier = pokemon.tier;
+				let displayedTier = room?.dataCommandTierDisplay === 'tiers' ? pokemon.tier :
+					room.dataCommandTierDisplay === 'doubles tiers' ? pokemon.doublesTier :
```
Zarel, line 545: Is it possible for `room` to be `null`? If so, `room.dataCommandTierDisplay` is wrong, if not, `room?.dataCommandTierDisplay` is wrong.

Zarel: Looks like yes.

```suggestion
				let displayedTier = room?.dataCommandTierDisplay === 'tiers' ? pokemon.tier :
					room?.dataCommandTierDisplay === 'doubles tiers' ? pokemon.doublesTier :
```

server/rooms.ts
```diff
@@ -111,6 +111,7 @@ export abstract class BasicRoom {
 	gtsga: GTSGiveaway | null;
 	toursEnabled: '%' | boolean;
 	tourAnnouncements: boolean;
+	dataCommandTierDisplay: string;
```
Zarel, line 114:
```suggestion
	dataCommandTierDisplay: 'tiers' | 'doubles tiers' | 'numbers';
```

KrisXV:
done @Zarel

Zarel:
TODO: `room.settings` to organize these things.

PR #6680, opened 2020-05-12
Require forcing for bans and blacklists on trusted users

mia-pi-git:
re https://www.smogon.com/forums/threads/add-forceroomban-command.3664175/#post-8464728


Zarel:
These should also notify CrisisMonitor.

KrisXV:
It already does.

Zarel:
Oh! I see it now.

PR #6681, opened 2020-05-12
Fix the Not Fully Evolved validator rule for past gens

KrisXV:


KrisXV:
bump @TheImmortal

data/rulesets.ts
```diff
@@ -649,7 +649,8 @@ export const BattleFormats: {[k: string]: FormatsData} = {
 		desc: "Bans Pok&eacute;mon that are fully evolved or can't evolve",
 		onValidateSet(set) {
 			const species = this.dex.getSpecies(set.species);
-			if (!species.nfe) {
+			const feInCurrentGen = species.evos && this.dex.getSpecies(species.evos[0]).gen > this.gen;
+			if (!species.nfe || feInCurrentGen) {
```
Zarel, line 653: It seems to me that the correct answer here is to set `.nfe` differently depending on the generation.

Zarel: (i.e. this implementation goes in `dex-data.ts` or `dex.ts` rather than here)

KrisXV: @Zarel how would you approach this? you can't currently access any of the functions in `dex.ts` in `dex-data.ts` and it makes more sense to handle `this.nfe` in `dex-data.ts`

Zarel: I would just give up and do it in `getSpecies`.

KrisXV:
I'll reopen this later

PR #6682, opened 2020-05-13
Correct the Farfetch’d's line's names

scheibo:
From Anubis and Marty's research, Farfetch’d should be using U+2019.
With the exception of Gen 5 and Pokemon Home it seems like it was
always meant to be the 'smart quote' symbol (though using a custom
encoding scheme in earlier gens).


scheibo:
(failing check appears unrelated to the code in question)

data/mods/gen5/pokedex.ts
```diff
@@ -67,6 +67,10 @@ export const BattlePokedex: {[k: string]: ModdedSpeciesData} = {
 		inherit: true,
 		baseStats: {hp: 80, atk: 110, def: 130, spa: 55, spd: 65, spe: 45},
 	},
+	farfetchd: {
+		inherit: true,
+		name: "Farfetch'd",
+	},
```
Zarel, line 73: PS has never accurately simulated past-gen names, so it's somewhat weird to do it here.

Several other name changes also involve ID changes, which further complicates things.

But if we were to be consistent with non-ID name changes like `SolarBeam` -> `Solar Beam` and `Compoundeyes` -> `Compound Eyes`, we would leave this off, too.

Zarel: (Name legality changes, if we do implement them, can be handled elsewhere.)

Zarel:
I'm not sure how I feel about this. Which Unicode character is used seems to me like an implementation detail, rather than relevant to simulation accuracy. You might as well complain that PS doesn't use the same font the games use, or that we don't use the same charset used in the earliest gens.

Still, though, the Unicode does look better to typography purists (even if it makes the code uglier). I'd accept it on that basis (assuming no one else has strong opinions), but that also makes the Gen 5 change unnecessary.

data/mods/gen5/pokedex.ts
Zarel:
```suggestion
```

In hindsight, I don't think there's any sense in which the Gen 5 difference is remotely meaningful to PS.

scheibo:
> I'm not sure how I feel about this. Which Unicode character is used seems to me like an implementation detail, rather than relevant to simulation accuracy. 

Why do we support `Flabébé`'s name then? Surely its more convenient to just use `Flabebe`.

It costs us pretty much nothing to be correct here

Zarel:
> Why do we support `Flabébé`'s name then? Surely its more convenient to just use `Flabebe`.
>
> It costs us pretty much nothing to be correct here

Because `Flabebe` is incorrect. `e` and `é` are different characters. It's a correctness difference.

Meanwhile, `'` and `’` are different representations of the same character. They're the same "character", just in a different font. PS already doesn't use the same font as the games, because I see no real reason to simulate the in-game fonts.

You're essentially insisting on using the same encoding as the games, which we already don't do: PS uses UTF-8, and the games have switched encodings multiple times and haven't even used Unicode until relatively recently.

Ironically, PS uses a different encoding for Flabébé in modern games, anyway: PS uses `e\u0301` rather than `\u00C9` for `é` (as a workaround to make `.id === 'flabebe'` rather than `'flabb'`).

PR #6683, opened 2020-05-13
Correct Xerneas formeOrder oversight

scheibo:
This was missed in #6669

### DIFF
--- a/data/pokedex.ts
+++ b/data/pokedex.ts
@@ -12102,6 +12102,23 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 	xerneas: {
 		num: 716,
 		name: "Xerneas",
+		baseForme: "Active", // Neutral is technically the base, but it can't be used in battle
+		types: ["Fairy"],
+		gender: "N",
+		baseStats: {hp: 126, atk: 131, def: 95, spa: 131, spd: 98, spe: 99},
+		abilities: {0: "Fairy Aura"},
+		heightm: 3,
+		weightkg: 215,
+		color: "Blue",
+		eggGroups: ["Undiscovered"],
+		otherFormes: ["Xerneas-Neutral"],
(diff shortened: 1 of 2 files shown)


data/pokedex.ts
```diff
@@ -12110,6 +12110,8 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 215,
 		color: "Blue",
 		eggGroups: ["Undiscovered"],
+		cosmeticFormes: ["Xerneas-Neutral"],
```
Zarel, line 12113: Xerneas-Neutral is not a cosmetic forme, because PS defines cosmetic formes as having no legality differences, while Xerneas-Neutral is illegal in battles.

The easiest solution would be just to have the forme named in `formeOrder` but not otherwise mentioned, like this:

```suggestion
```

But a more thorough solution might be to make it an illegal forme in some way, possibly marking it as `Unobtainable`?

Zarel: While we're at it, we probably want to set `baseForme: "Active",`

scheibo: I think the latter option is preferable, the test I wrote (which can be modified of course, but I feel its fairly useful for testing the constraints and avoid typos) expects that the forme is formeOrder is a real species

Zarel:
http://play.pokemonshowdown.com/sprites/gen5/xerneas-neutral.png

(and corresponding sprites) need to exist before I can merge this.

Zarel:
(If you don't want to wait for sprites, the test you wrote could hardcode an exception for Xerneas-Neutral.)

scheibo:
> (and corresponding sprites) need to exist before I can merge this.

LMAO, there are so many broken fucking sprites, as if *this* is the showstopper.
Anyway, Xerneas-Neutral -> Xerneas now in `sprites/`, so I'll merge this

PR #6684, opened 2020-05-13
Mafia: Make CS indication more obvious for hosts

ZestOfLife:
Qol for hosts

### DIFF
--- a/server/chat-plugins/mafia.ts
+++ b/server/chat-plugins/mafia.ts
@@ -1773,9 +1773,9 @@ export const pages: PageTable = {
 					buf += `<p>${game.theme.desc}</p>`;
 				}
 				if (game.noReveal) {
-					buf += `<p><span style="font-weight:bold; color:${game.closedSetup ? '#ffff00' : 'initial'}">Original Rolelist</span>: ${game.originalRoleString}</p>`;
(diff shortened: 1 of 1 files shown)


HoeenCoder:
Have seen, am communicating in mafia staff discord.

Zarel:
@HoeenCoder bump?

HoeenCoder:
Suggested this in #code-nerds but it got buried and I forgot about this sorry: What about something other than a color indicator since that seems unituitive. Maybe just add a text indicator to the end (eg: `Rolelist (CS):`

server/chat-plugins/mafia.ts
```diff
@@ -1773,9 +1773,9 @@ export const pages: PageTable = {
 					buf += `<p>${game.theme.desc}</p>`;
 				}
 				if (game.noReveal) {
-					buf += `<p><span style="font-weight:bold;">Original Rolelist</span>: ${game.originalRoleString}</p>`;
+					buf += `<p><span style="font-weight:bold;">Original Rolelist ${game.closedSetup ? '(CS)' : ''}</span>: ${game.originalRoleString}</p>`;
```
whalemer, line 1776: Won't this render as `Original rolelist : whatever, whatever` if it's not a cs?

whalemer:
thanks!

Zarel:
After two months... this certainly deserves congratulations!

🎉

PR #1097, opened 2014-07-19
Add tournament announcements II

sirDonovan:
Tournaments that are started in official rooms are announced in the Lobby.
e: <a>http://i.imgur.com/4b6oxMI.png</a>

PR #6685, opened 2020-05-13
Add test for unmatched < or >

PartMan7:
This change prevents using an unmatched < to escape UHTML name tags, but also prevents direct nesting of tags (of the form ``<button value="<..>">`` or similar). However, most of these uses (if they exist) can be covered by escaping them.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1076,30 +1076,25 @@ export class CommandContext extends MessageContext {
 			this.errorReply('Do not use "click here"');
 			return null;
 		}
-		if (!/^[^<>]*(?:<[^<>]*?>[^<>]*)*$/.test(htmlContent)) {
-			this.errorReply('Found unmatched < or >. If you wish to use them, please use `&lt;` or `&gt;`.');
-			return null;
-		}
 
 		// check for mismatched tags
 		const tags = htmlContent
-			.toLowerCase()
 			// eslint-ignore-next-line @typescript-eslint/prefer-regexp-exec
-			.match(/<\/?(?:div|a|button|b|strong|em|i|u|center|font|marquee|blink|details|summary|code|table|td|tr|style|script)\b/g);
+			.match(/<\/?[a-z0-9]*/gi);
 		if (tags) {
 			const stack = [];
 			for (const tag of tags) {
 				if (tag.charAt(1) === '/') {
 					if (!stack.length) {
-						this.errorReply("Extraneous </" + tag.substr(2) + "> without an opening tag.");
+						this.errorReply(`Extraneous </${tag.substr(2)}> without an opening tag.`);
 						return null;
 					}
 					if (tag.substr(2) !== stack.pop()) {
-						this.errorReply("Missing </" + tag.substr(2) + "> or it's in the wrong place.");
+						this.errorReply(`Missing </${tag.substr(2)}> or it's in the wrong place.`);
 						return null;
 					}
-				} else {
-					stack.push(tag.substr(1));
+				} else if (/<(?:div|a|button|b|strong|em|i|u|center|font|marquee|blink|details|summary|code|table|td|tr|th|style|script|h\d)\b/.test(tag.toLowerCase())) {
+					this.errorReply(`Unrecognized tag \`${match.slice(1)}\` (often caused by using an unmatched <). If you meant to use the character, use a \`&lt;\` instead.`);
 				}
 			}
 			if (stack.length) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1094,7 +1094,9 @@ export class CommandContext extends MessageContext {
 						return null;
 					}
 				} else if (/<(?:div|a|button|b|strong|em|i|u|center|font|marquee|blink|details|summary|code|table|td|tr|th|style|script|h\d)\b/.test(tag.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


PartMan7:
The code works properly on a sideserver, but I'm unsure of whether this is the appropriate place to use it. Is there a different location for validating UHTML?

monsanto:
If you want to validate HTML, why not use an off-the-shelf HTML parser, instead of an ad-hoc regular expression?

PartMan7:
Wasn't the stance to avoid unnecessary modules?

monsanto:
Someone else will have to speak to that, but I don't see the point in half-solving a problem.

PartMan7:
There was a certain... exploit that this PR was meant to counter, but I'll look into a parser.

PartMan7:
This is in the wrong place, anyways; it should be in chat.ts

monsanto:
If you spoke to someone beforehand and they said not to use a parser then you don't have to change that for me; I'm not the maintainer of this bit of code. (It doesn't change my opinion, though)

If PS is currently vulnerable to exploit then that's something that should be raised privately with technical staff, not mentioned as an aside in a PR that needs further work before merge

PartMan7:
I did contact technical staff; and they said I could send a PR on the same. Whether or not to use a parser wasn't discussed, though.

Zarel:
The validator is in `canHTML`, and this should go there, too.

Anyway, I don't mind this sort of solution. I don't like unnecessary dependencies.

Zarel:
I assume this is a minor exploit (hiding a bot credit, which can be found in the roomlog anyway).

PartMan7:
Yeah, this allowed using a certain tag to hide the username.

PartMan7:
Tested, works on the samples I tried.

Zarel:
Anyway, this is not the regex I'd use.

Zarel:
It's actually only `<` that's a problem. `>` is completely fine.

Zarel:
We already have a tag mismatch checker. I'd just generalize that one into something that additionally handles `<` and `>` in general.

PartMan7:
Alright, renaming this to WIP until I combine 'em.

Zarel:
There's a "draft" button for that, now.

PartMan7:
Ah, thanks.

PartMan7:
Wrote a new one that combines the checks, validated on: https://pastie.io/raw/znxrgt

PartMan7:
Passed the tests on a sideserver, so the code should work, now.

Though now I have no idea how to reduce #1097 by 20 characters.

server/chat.ts
```diff
@@ -1081,25 +1081,28 @@ export class CommandContext extends MessageContext {
 		const tags = htmlContent
 			.toLowerCase()
 			// eslint-ignore-next-line @typescript-eslint/prefer-regexp-exec
-			.match(/<\/?(?:div|a|button|b|strong|em|i|u|center|font|marquee|blink|details|summary|code|table|td|tr|style|script)\b/g);
+			.match(/<\/?[a-z0-9]*/g);
 		if (tags) {
 			const stack = [];
```
Zarel, line 1086:
```suggestion
			const LEGAL_TAGS = ['div', 'a', 'button', 'b', 'strong', 'em', 'i', 'u', 'center', 'font', 'marquee', 'blink', 'details', 'summary', 'code', 'table', 'td', 'tr', 'th', 'style', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
			const LEGAL_AUTOCLOSE_TAGS = ['br', 'area', 'embed', 'hr', 'img', 'link', 'source', 'track'];
```

server/chat.ts
```diff
@@ -1081,25 +1081,28 @@ export class CommandContext extends MessageContext {
...
 			for (const tag of tags) {
 				if (tag.charAt(1) === '/') {
```
Zarel, line 1088:
```suggestion
				const isCloseTag = tag.charAt(1) === '/';
				const tagName = tag.slice(isCloseTag ? 2 : 1);
				if (isCloseTag) {
```

server/chat.ts
```diff
@@ -1081,25 +1081,28 @@ export class CommandContext extends MessageContext {
...
 					if (!stack.length) {
-						this.errorReply("Extraneous </" + tag.substr(2) + "> without an opening tag.");
+						this.errorReply(`Extraneous </${tag.substr(2)}> without an opening tag.`);
```
Zarel, line 1090:
```suggestion
						this.errorReply(`Extraneous </${tagName}> without an opening tag.`);
```

server/chat.ts
```diff
@@ -1081,25 +1081,28 @@ export class CommandContext extends MessageContext {
...
 						return null;
 					}
 					if (tag.substr(2) !== stack.pop()) {
```
Zarel, line 1093:
```suggestion
					if (tagName !== stack.pop()) {
```

Zarel:
In hindsight, I'm basically rewriting this entire thing at this point, let me just push this myself.

Zarel:
So the problem with your whitelist is that you're missing a lot of tags. I added `<small>`, `<sup>`, `<sub>`, `<kbd>` and like ten others before I realized I didn't really want a huge unwieldy unnecessary solution. We can just tokenize tags the same way actual HTML engines do.

PR #6686, opened 2020-05-14
Support different output number formats in /math

asgdf:
I felt this might be useful for doing math with memory addresses in hex (which I have seen some people do occasionally) or maybe doing some bit shuffling for whatever reason. Since JS supports binary, octal, decimal and hex literals, that's the ones I allowed as outputs.

### DIFF
--- a/server/chat-plugins/calculator.ts
+++ b/server/chat-plugins/calculator.ts
@@ -139,18 +139,43 @@ export const commands: ChatCommands = {
 	math: "calculate",
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
-		const targets = target.split(',');
-		const expression = targets[0];
-		let base = parseInt(targets[1]);
-		if (isNaN(base)) base = 10;
-		if (targets.length > 1 && !(base in BASE_PREFIXES)) {
-			return this.errorReply("Invalid base. Valid options are: 2 8 10 16");
+		let base = 10;
+		if (target.includes('0x')) {
+			base = 16;
+		} else if (target.includes('0o')) {
+			base = 8;
+		} else if (target.includes('0b')) {
+			base = 2;
 		}
+
+		const baseMatchResult = (/\b(?:in|to)\s+([a-zA-Z]+)\b/).exec(target);
+		if (baseMatchResult) {
+			switch (toID(baseMatchResult[1])) {
+			case 'decimal': case 'dec': base = 10; break;
+			case 'hexadecimal': case 'hex': base = 16; break;
+			case 'octal': case 'oct': base = 8; break;
+			case 'binary': case 'bin': base = 2; break;
+			default:
+				return this.errorReply(`Unrecognized base "${baseMatchResult[1]}". Valid options are binary or bin, octal or oct, decimal or dec, and hexadecimal or hex.`);
+			}
+		}
+		const expression = target.replace(/\b(in|to)\s+([a-zA-Z]+)\b/g, '').trim();
 
 		if (!this.runBroadcast()) return;
 		try {
 			const result = solveRPN(parseMathematicalExpression(expression));
-			this.sendReplyBox(Chat.html`${expression}<br />= <strong>${BASE_PREFIXES[base] + result?.toString(base)}</strong>`);
+			let baseResult = '';
+			if (result && base !== 10) {
+				baseResult = `${BASE_PREFIXES[base]}${result.toString(base).toUpperCase()}`;
+				if (baseResult === expression) baseResult = '';
+			}
+			let resultStr = '';
+			if (baseResult) {
+				resultStr = `<strong>${baseResult}</strong> = ${result}`;
+			} else {
+				resultStr = `<strong>${result}</strong>`;
+			}
+			this.sendReplyBox(`${expression}<br />= ${resultStr}`);
 		} catch (e) {
 			this.sendReplyBox(
 				Chat.html`${expression}<br />= <span class="message-error"><strong>Invalid input:</strong> ${e.message}</span>`
@@ -158,7 +183,7 @@ export const commands: ChatCommands = {
 		}
 	},
 	calculatehelp: [
-		`/calculate [arithmetical question], [result base] - Calculates an arithmetical question. Supports PEMDAS (Parenthesis, Exponents, Multiplication, Division, Addition and Subtraction), pi and e. Possible bases for the result are 2, 8, 10 and 16.
-		`,
+		`/calculate [arithmetic question] - Calculates an arithmetical question. Supports PEMDAS (Parenthesis, Exponents, Multiplication, Division, Addition and Subtraction), pi and e.`,
+		`/calculate [arithmetic question] in [base] - Returns the result in a specific base. [base] can be bin, oct, dec or hex.`,
 	],
 };
--- a/server/chat-plugins/calculator.ts
+++ b/server/chat-plugins/calculator.ts
@@ -31,6 +31,13 @@ const OPERATORS: {[k in Operator]: Operators} = {
 	},
 };
 
+const BASE_PREFIXES: {[base: number]: string} = {
+	2: "0b",
+	8: "0o",
+	10: "",
+	16: "0x",
+};
+
 function parseMathematicalExpression(infix: string) {
 	// Shunting-yard Algorithm -- https://en.wikipedia.org/wiki/Shunting-yard_algorithm
 	const outputQueue: string[] = [];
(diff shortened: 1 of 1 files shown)


Slayer95:
Probably the main legitimate use case for this in PS would be teaching how much ``0x14CD/0x1000`` and friends exactly are, which would be obscured by hex output. Fix: always display result in decimal format (too).

asgdf:
By default, `0x14CD/0x1000` will still display in decimal. This just gives the additional option of making it display in hex when you wish to do so. So I'm not sure I understand the issue.

Slayer95:
So that has already been taken care of. Please disregard the comment then.

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
 	math: "calculate",
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
+		const targets = target.split(',');
```
Zarel, line 142: idk how to feel about using comma as a separator.

asgdf: I guess it risks being misleading for people who are used to , are the decimal separator. What other options are there, though? Using | ? Putting the base in the command name, like math16?

Zarel: I would probably want to do it Google-like, where you'd postfix "to hex", "to binary", "to octal".

Also I would automatically show both bases.

So it'd display:

`0x8 + 0x8 = 0x10 = 16`

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
...
+		const expression = targets[0];
```
Zarel, line 143:
```suggestion
		const expression = targets[0].trim();
```

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
...
+		let base = parseInt(targets[1]);
+		if (isNaN(base)) base = 10;
```
Zarel, line 145:
```suggestion
		if (isNaN(base)) {
			if (expression.includes('0x') {
				base = 16;
			} else if (expression.includes('0o') {
				base = 8;
			} else if (expression.includes('0b') {
				base = 2;
			} else {
				base = 10;
			}
		}
```

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
...
+		if (targets.length > 1 && !(base in BASE_PREFIXES)) {
+			return this.errorReply("Invalid base. Valid options are: 2 8 10 16");
+		}
+
 		if (!this.runBroadcast()) return;
 		try {
-			const result = solveRPN(parseMathematicalExpression(target));
-			this.sendReplyBox(Chat.html`${target}<br />= <strong>${Chat.stringify(result)}</strong>`);
+			const result = solveRPN(parseMathematicalExpression(expression));
+			this.sendReplyBox(Chat.html`${expression}<br />= <strong>${BASE_PREFIXES[base] + result?.toString(base)}</strong>`);
```
Zarel, line 153:
```suggestion
			let baseResult = '';
			if (result) {
				if (base in BASE_PREFIXES) {
					baseResult = `${BASE_PREFIXES[base]}${result.toString(base)}`;
					if (baseResult === expression) baseResult = '';
				} else {
					baseResult = `(${result.toString(base)})<sub>${base}</sub>`;
				}
			}
			if (baseResult) {
				result = `<strong>${baseResult}</strong> = ${result}`;
			} else {
				result = `<strong>${result}</strong>`;
			}
			this.sendReplyBox(Chat.html`${expression}<br />= ${result}`);
```

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
 	math: "calculate",
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
+		const targets = target.split(',');
+		const expression = targets[0];
+		let base = parseInt(targets[1]);
+		if (isNaN(base)) base = 10;
+		if (targets.length > 1 && !(base in BASE_PREFIXES)) {
+			return this.errorReply("Invalid base. Valid options are: 2 8 10 16");
+		}
```
Zarel, line 148: Oh! You only support bases 2, 8, 10, and 16. All the more reason to use the text version, so it's clearer what bases are supported.

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,26 @@ export const commands: ChatCommands = {
 	math: "calculate",
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
+		const targets = target.split(',');
+		const expression = targets[0];
+		let base = parseInt(targets[1]);
+		if (isNaN(base)) base = 10;
+		if (targets.length > 1 && !(base in BASE_PREFIXES)) {
```
Zarel, line 146:
```suggestion
		if (!(base in BASE_PREFIXES)) {
```

`base` should be `10` if `targets.length <= 1`

server/chat-plugins/calculator.ts
asgdf: I'm not sure how obvious it would be to people that they can do "in hex" or something like that. But I guess it would be a niche feature anyway.

Zarel: It can still be documented. I don't see how the comma is any more obvious?

asgdf: Because comma-separated arguments is what just about every other PS command does.

Zarel: Yes, but it's not immediately clear what the second argument _means_.

Zarel:
@asgdf what happened? I'm going to merge my change soon because I haven't heard from you.

asgdf:
Ah sorry, it was just "oh dear that's a lot of additional code for this small change, I'll get to it at some point" and then I got busy.

Zarel:
I assume you won't mind if I just take this over?

server/chat-plugins/calculator.ts
Zarel: (Like it's not immediately obvious to me that `/calc 1, 2`  means "show me 1 in base 2")

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,51 @@ export const commands: ChatCommands = {
...
+
 		if (!this.runBroadcast()) return;
 		try {
-			const result = solveRPN(parseMathematicalExpression(target));
-			this.sendReplyBox(Chat.html`${target}<br />= <strong>${Chat.stringify(result)}</strong>`);
+			const result = solveRPN(parseMathematicalExpression(expression));
+			let baseResult = ``;
+			if (result) {
+				baseResult = `${BASE_PREFIXES[base]}${result.toString(base).toUpperCase()}`;
+				if (baseResult === result.toString()) baseResult = '';
```
Zarel, line 170:
```suggestion
				if (baseResult === result.toString() || baseResult === expression) {
					baseResult = '';
				}
```

There's no need to display `0x10 = 0x10 = 16`

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,51 @@ export const commands: ChatCommands = {
...
+			switch (toID(baseMatchResult[1])) {
+			case 'decimal': case 'dec': base = 10; break;
+			case 'hexadecimal': case 'hex': base = 16; break;
+			case 'octal': case 'oct': base = 8; break;
+			case 'binary': case 'bin': base = 2; break;
+			default:
+				return this.errorReply(`Unrecognized base "${baseMatchResult[1]}". Valid options are binary or bin, octal or oct, decimal or dec, and hexadecimal or hex.`);
+			}
+		}
+		const expression = target.replace(/\bin\s+([a-zA-Z]+)\b/g, '').trim();
```
Zarel, line 162:
```suggestion
		const expression = target.replace(/\b(?:in|to)\s+([a-zA-Z]+)\b/g, '').trim();
```

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,51 @@ export const commands: ChatCommands = {
...
 		} catch (e) {
 			this.sendReplyBox(
-				Chat.html`${target}<br />= <span class="message-error"><strong>Invalid input:</strong> ${e.message}</span>`
+				Chat.html`${expression}<br />= <span class="message-error"><strong>Invalid input:</strong> ${e.message}</span>`
 			);
 		}
 	},
 	calculatehelp: [
-		`/calculate [arithmetical question] - Calculates an arithmetical question. Supports PEMDAS (Parenthesis, Exponents, Multiplication, Division, Addition and Subtraction), pi and e.`,
+		`/calculate [arithmetical question] [in [base]] - Calculates an arithmetical question. Supports PEMDAS (Parenthesis, Exponents, Multiplication, Division, Addition and Subtraction), pi and e.
```
Zarel, line 186:
```suggestion
		`/calculate [arithmetical question] [in [base]] - Calculates an arithmetical question. Supports PEMDAS (Parenthesis, Exponents, Multiplication, Division, Addition and Subtraction), pi and e.
		`/calculate [arithmetical question] in [base] - Returns the result in a specific base. [base] can be bin, oct, or hex.
```

asgdf:
Is the prefer-exec rule thing new? I didn't get that error locally

server/chat-plugins/calculator.ts
asgdf: I'm not a native speaker, but "to" feels weird to me in "10 + 12 to hex". I guess it implies a verb like "change base to" or something? But since with singular numbers I'd always say "this is in hex" or "this is in binary", that feels more natural to me.

Zarel:
Yeah, it's relatively recent, but it's still like a few months old I think?

server/chat-plugins/calculator.ts
```diff
@@ -132,17 +139,51 @@ export const commands: ChatCommands = {
...
+		let base = 10;
+		if (target.includes('0x')) {
+			base = 16;
+		} else if (target.includes('0o')) {
+			base = 8;
+		} else if (target.includes('0b')) {
+			base = 2;
+		}
+
+		const baseMatchResult = target.match(/\bin\s+([a-zA-Z]+)\b/);
```
Zarel, line 151:
```suggestion
		const baseMatchResult = /\bin\s+([a-zA-Z]+)\b/.exec(target);
```

server/chat-plugins/calculator.ts
Zarel: To me, `to` feels more natural; the implied verb is `converted to`. They're both grammatically correct, but `in` feels slightly more awkward. I don't mind either one, if you feel strongly.

Zarel:
Thanks!

PR #6693, opened 2020-05-15
Replace attributed uhtml on change

xfix:
Not doing so wastes a lot of bandwidth.

This fixes https://www.smogon.com/forums/threads/dont-send-all-uhtml-messages.3664370/

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -90,8 +90,11 @@ export const commands: ChatCommands = {
 			html += Chat.html`<div style="float:right;color:#888;font-size:8pt">[${user.name}]</div><div style="clear:both"></div>`;
 		}
 
-		html = `/uhtml${(cmd === 'changeuhtml' ? 'change' : '')} ${name},${html}`;
-		return html;
+		if (cmd === 'changeuhtml') {
+			room.attributedUhtmlchange(user, name, html);
+		} else {
+			return `/uhtml ${name},${html}`;
+		}
 	},
 	adduhtmlhelp: [
 		`/adduhtml [name], [message] - Shows everyone a message that can change, parsing HTML code contained.  Requires: * & ~`,
--- a/server/roomlogs.ts
+++ b/server/roomlogs.ts
@@ -156,14 +156,22 @@ export class Roomlog {
 		if (!Roomlogs.rollLogTimer) void Roomlogs.rollLogs();
 	}
 	add(message: string) {
-		if (message.startsWith('|uhtmlchange|')) return this.uhtmlchange(message);
-		this.roomlog(message);
-		if (this.logTimes && message.startsWith('|c|')) {
-			message = '|c:|' + (~~(Date.now() / 1000)) + '|' + message.substr(3);
+		if (message.startsWith('|uhtmlchange|')) {
+			this.uhtmlchange(message);
+		} else {
+			this.roomlog(message);
+			message = this.withTimestamp(message);
+			this.log.push(message);
+			this.broadcastBuffer += message + '\n';
 		}
-		this.log.push(message);
-		this.broadcastBuffer += message + '\n';
 		return this;
+	}
+	private withTimestamp(message: string) {
+		if (this.logTimes && message.startsWith('|c|')) {
+			return `|c:|${Math.trunc(Date.now() / 1000)}|${message.slice(3)}`;
+		} else {
+			return message;
+		}
 	}
 	hasUsername(username: string) {
 		const userid = toID(username);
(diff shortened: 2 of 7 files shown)


Zarel:
Why did we start attributing uhtml changes in this way in the first place?

xfix:
https://github.com/smogon/pokemon-showdown/pull/6444

My guess is so that /hidetext would work.

server/chat-commands/admin.ts
```diff
@@ -90,8 +90,11 @@ export const commands: ChatCommands = {
 			html += Chat.html`<div style="float:right;color:#888;font-size:8pt">[${user.name}]</div><div style="clear:both"></div>`;
 		}
 
-		html = `/uhtml${(cmd === 'changeuhtml' ? 'change' : '')} ${name},${html}`;
-		return html;
+		if (cmd === 'changeuhtml') {
+			room.addUserUhtmlchange(user, name, html);
```
Zarel, line 94: Our previous implementation did it by detecting `|uhtmlchange|`; the new one should probably do it by detecting `|/uhtmlchange`. I'm aware that that'd be a lot harder since you'd have to parse, so I guess I'm tentatively fine with this, but we should probably rip out the old implementation completely.

Assuming there was something wrong with the old implementation that makes the attributed version preferable; I'm going to need to hunt that down.

Zarel:
Aha, that makes sense although it's annoying. :/

server/chat-commands/admin.ts
xfix: `|uhtmlchange|` is being used by dynamic HTMLs from chat plugins (chat plugins don't really have an user). Those probably could be changed to use `|c:|` with `~` user however.

server/roomlogs.ts
```diff
@@ -211,8 +217,25 @@ export class Roomlog {
...
+		const fullMessage = this.addTimestamp(`|c|${user.getIdentity()}|${start}${message}`);
+		for (const [i, line] of this.log.entries()) {
+			if (this.parseChatLine(line)?.message.startsWith(start)) {
+				this.log[i] = fullMessage;
+				break;
+			}
+		}
+		this.broadcastBuffer += fullMessage + '\n';
+	}
+	private parseChatLine(line: string) {
```
Zarel, line 232: Considering it's only used for the username at the moment, I would make it do only that. No need to make this more complicated than it needs to be.

server/roomlogs.ts
```diff
@@ -211,8 +217,25 @@ export class Roomlog {
 				break;
 			}
 		}
-		this.broadcastBuffer += message + '\n';
-		return this;
+	}
+	addUserUhtmlchange(user: User, name: string, message: string) {
```
Zarel, line 221: This should just be named `uhtmlchange` - the old `uhtmlchange` isn't used anywhere, is it?

If it is, I would call this `uhtmlchangeByUser` or something, so they make sense.

server/roomlogs.ts
```diff
@@ -156,15 +156,23 @@ export class Roomlog {
 		if (!Roomlogs.rollLogTimer) void Roomlogs.rollLogs();
 	}
 	add(message: string) {
-		if (message.startsWith('|uhtmlchange|')) return this.uhtmlchange(message);
-		this.roomlog(message);
-		if (this.logTimes && message.startsWith('|c|')) {
-			message = '|c:|' + (~~(Date.now() / 1000)) + '|' + message.substr(3);
+		if (message.startsWith('|uhtmlchange|')) {
```
Zarel, line 159: I would probably remove this magic, if one uses a function call and the other doesn't. I'd just update them so they all both function calls.

server/roomlogs.ts
xfix: I wanted to avoid duplication of code from `clearText` and `addUserUhtmlchange`, considering it's a rather complicated code. Do you have a better idea than to do this?

server/roomlogs.ts
```diff
@@ -156,15 +156,23 @@ export class Roomlog {
...
+			message = this.addTimestamp(message);
+			this.log.push(message);
 		}
-		this.log.push(message);
 		this.broadcastBuffer += message + '\n';
 		return this;
 	}
+	private addTimestamp(message: string) {
+		if (this.logTimes && message.startsWith('|c|')) {
+			return '|c:|' + (~~(Date.now() / 1000)) + '|' + message.substr(3);
```
Zarel, line 171:
```suggestion
			return `|c:|${Math.trunc(Date.now() / 1000)}|${message.slice(3)}`;
```

server/roomlogs.ts
xfix: The `uhtmlchange` method is used in `add` method.

I don't think `uhtmlchangeByUser` is a good name for method, considering it's not working by user, rather it works on attributed `uhtmlchange`. Maybe I should name the method `addAttributedUhtmlchange` instead.

Note that `uhtmlchange` is distinct from this method, as it doesn't actually `add`.

server/roomlogs.ts
xfix: There are plenty of `this.room.add` calls. I want to limit the changes to minimum, considering current implementation causes issues when connecting to Showdown (you have to refresh the page multiple times sometimes).

server/roomlogs.ts
Zarel: Like I said, I would personally replace this with a `getUsernameFromChatLine` function (because that's its only use-case), but I suppose it's unlikely to have a noticeable performance impact. Do what you like.

server/roomlogs.ts
Zarel: This is really just a bandwidth optimization; I don't think third party server code being slightly suboptimal is worth having this unintuitive API in place. If you _really_ care about diff size, I can do this change in a separate commit, but it seems like the wrong thing to optimize for.

xfix: Fixed.

xfix:
Updated.

PR #6697, opened 2020-05-16
Make /addline require console access

HoeenCoder:
Honestly why isn't this limited to those with console access. Its extremely powerful still, being the (literal) equivalent of `/eval room.add(target);`.

This becomes more important with the proposed flattening.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -126,8 +126,10 @@ export const commands: ChatCommands = {
 		`/changerankuhtml [rank], [name], [message] - Changes the message previously shown with /addrankuhtml [rank], [name]. Requires: * & ~`,
 	],
 
-	addline(target, room, user) {
-		if (!this.can('rawpacket')) return false;
+	addline(target, room, user, connection) {
+		if (!user.hasConsoleAccess(connection)) {
+			return this.errorReply("/addline - Access denied.");
+		}
 		// secret sysop command
 		room.add(target);
 	},


Zarel:
It's powerful but it doesn't seem like it needs to be denied from admins? It's significantly less powerful than `/eval`. It can be used to screw around and trick people, but at admin-only I don't really see abuse potential being problematic at its permission level.

Zarel:
Like, sure, it can be abused, but only slightly more than `/adduhtml` which we give to every global bot. I guess you can spoof user joins/leaves and tournaments and stuff, but that doesn't seem _dangerous_ to me, just pointless?

Zarel:
Like this user has the ability to demote and ban whoever they want, but you're worried about them spoofing another user joining?

Like the main issue with spoofing is on side servers where people spoof me talking in them as a joke, but this change won't solve that problem at all.

Zarel:
I'll leave this open (and feel free to send me security concerns privately on Discord) but unless I hear something very different, this is a no.

xfix:
I think this change is more likely to pass by making the change remove the command entirely :).

Zarel:
It's a shortcut for `/eval room.add(TARGET);`

Honestly, I think a shortcut for `/eval this.sendReply(TARGET)` might also be useful. The latter is useful for testing code, and the former is useful as an alias of `/addhtml` without adding the box and with some more functionality like fine-grained notification control, etc.

PR #6702, opened 2020-05-16
Rename data tables from Battle[foo] to Dex[foo]

Zarel:
Files in `data/` now export `Dex...` instead of `Battle...`.

- `BattlePokedex` -> `DexPokedex`, etc

In addition, this is a good time to rename `Movedex` to `Moves` and `TypeChart` to `Types`.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -32,7 +32,7 @@ Ratings and how they work:
 
 */
 
-export const BattleAbilities: {[abilityid: string]: AbilityData} = {
+export const DexAbilities: {[abilityid: string]: AbilityData} = {
 	noability: {
 		shortDesc: "Does nothing.",
 		isNonstandard: "Past",
--- a/data/aliases.ts
+++ b/data/aliases.ts
@@ -1,4 +1,4 @@
-export const BattleAliases: {[alias: string]: string} = {
+export const DexAliases: {[alias: string]: string} = {
 	// formats
 	randbats: "[Gen 8] Random Battle",
 	uber: "[Gen 8] Ubers",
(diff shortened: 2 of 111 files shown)


Zarel:
Other related changes I might add to this PR, please thumbs-up/down (or comment, commenting is always allowed):

Zarel:
- doing something about `rulesets`; probably rename `Formats` to `Rulesets` - it makes more sense that a format is a type of rule than that a rule is a type of format. I'm mulling `rulesets` vs `rules`.

Zarel:
- doing something about `Types`, it's currently a `name:value` table instead of an `id:value` table like all the others, even `DexNatures` is an `id:value`.

monsanto:
Why not just `module.exports = ...` instead `exports.<name> = ...`? Doesn't change what you name things on `require()`, and avoids backwards compatibility breaks of this kind in the future.

> doing something about rulesets; probably rename Formats to Rulesets - it makes more sense that a format is a type of rule than that a rule is a type of format.

No opinion as long as this does not change UI, it's been a long process to get folks to standardize on "format" vs slang usage of "tier" vs "metagame" and we're still not really there yet

Zarel:
> Why not just `module.exports = ...` instead `exports.<name> = ...`?

This isn't compatible with ES6 modules, and so has poor compatibility with TypeScript. It also has doesn't work with the ES6 module workaround we're using in client.

> No opinion as long as this does not change UI, it's been a long process to get folks to standardize on "format" vs slang usage of "tier" vs "metagame" and we're still not really there yet

This is specifically about PS's combined format/ruleset table. Formats themselves would still be called formats.

Zarel:
I've given up on this because of pre's emoji react, anyway. Damned if you do, damned if you don't, so I guess I'll go with the direction of less effort.

PR #6710, opened 2020-05-17
Fix the rename command to log the new room & require a target

Volco:
If someone were to type ``/rename `` it would then glitch any server that has the autojoin for lobby enabled. Also for whatever reason it just didn't log the newly named room in chat logs.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -822,8 +822,6 @@ export const commands: ChatCommands = {
 			return this.errorReply(`An error occured while renaming the room.`);
 		}
 		this.modlog(`RENAMEROOM`, null, `from ${oldTitle}`);
-		room.log.roomlogFilename = "";
-		await room.log.setupRoomlogStream(true);
 		const privacy = room.isPrivate === true ? "Private" :
 			room.isPrivate === false ? "Public" :
 			`${room.isPrivate.charAt(0).toUpperCase()}${room.isPrivate.slice(1)}`;
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -809,6 +809,7 @@ export const commands: ChatCommands = {
 		const oldTitle = room.title;
 		const roomid = toID(target) as RoomID;
 		const roomtitle = target;
+		if (!roomid.length) return this.errorReply("The new room needs a title.");
 		// `,` is a delimiter used by a lot of /commands
 		// `|` and `[` are delimiters used by the protocol
 		// `-` has special meaning in roomids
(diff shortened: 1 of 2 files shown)


server/chat-commands/room-settings.ts
```diff
@@ -821,6 +822,8 @@ export const commands: ChatCommands = {
 			return this.errorReply(`An error occured while renaming the room.`);
 		}
 		this.modlog(`RENAMEROOM`, null, `from ${oldTitle}`);
+		room.log.roomlogFilename = "";
```
Zarel, line 825: I was confused by what this line was doing, and after further investigation, this doesn't seem to be doing anything useful. It looks like all it does is force `setupRoomlogStream` to re-create the roomlog stream even if it hasn't changed filenames, which isn't useful.

```suggestion
```

Zarel:
The rest of this commit seems good, though, although I regret approving the original command name as `/rename` in #6295. I might change the name to `/renameroom`.

Zarel:
Thanks for finding and fixing this bug!

server/chat-commands/room-settings.ts
Zarel: I might be wrong, though. Am I missing something?

server/chat-commands/room-settings.ts
```diff
@@ -821,6 +822,8 @@ export const commands: ChatCommands = {
...
+		await room.log.setupRoomlogStream(true);
```
Zarel, line 826: Actually, in hindsight, this should also be unnecessary since `room.log.rename` (which is called by `room.rename`) should already handle it.

```suggestion
```

Is there a problem with its current implication?

mia-pi-git: Tested this myself, and it seems to function fine currently, other than the fact you can ``>> room.rename()`` and it doesn't fail (which it probably should.)
I may be missing something, though.

mia-pi-git: ``room.log.rename`` and ``room.rename`` should both be throwing errors if there's no ID and title given.

server/chat-commands/room-settings.ts
Volco: it's necessary due to this line: https://github.com/smogon/pokemon-showdown/blob/master/server/roomlogs.ts#L136 
Honestly i would prefer to exclude that line because to me it just looks gross but as a result of the filename already being set to the current date after the rename it just causes it to fail despite not actually writing any logs.

server/chat-commands/room-settings.ts
Volco: it does not log any messages in the chat after renaming it which is why i put the ``await room.log.setupRoomlogStream(true);`` which not necessarily a big issue but could be also be an issue if say a room was renamed and there hadn't been a restart and staff needed to see the logs with ``/chatlog`` but found none since the rename

Volco: https://github.com/smogon/pokemon-showdown/blob/master/server/rooms.ts#L1379
https://github.com/smogon/pokemon-showdown/blob/master/server/roomlogs.ts#L247
@mia-pi-git seems no one bothered to add something to make it throw errors if there's no IDs given so that's fun lol

Volco:
Okay, I fixed the placement of the ``roomlogFileName`` reassignment after I looked at the called functions and saw it would be better there rather than thrown hackily into the command itself.

server/chat-commands/room-settings.ts
Volco: this should show the issue better than how i'm explaining it

Zarel:
Thanks! I like that solution a lot more.

PR #6711, opened 2020-05-17
Add case insensitivity to Scripted Buttons

PartMan7:
Please don't exceed the line length this time. ;;


Zarel:
Wow, whoops, I forgot about that part, thanks.

Don't worry if line length is an issue; I can suggest ways to fix it if you're having trouble thinking of a good way to deal with it. You should always feel free to send PRs with errors in them if you want to ask me about the best way to fix them.

PR #6712, opened 2020-05-18
Fix Trace's interaction with Neutralizing Gas

DaWoblefet:
Missed Trace when I was doing research on NGas for https://github.com/smogon/pokemon-showdown/pull/6581


Zarel:
Thanks!

PR #6714, opened 2020-05-18
Add various mechanics tests

DaWoblefet:
Specifically, Fling, Tar Shot, Symbiosis, Steely Spirit, the item Metronome, Max Guard, and a bizarre interaction with moves that call other moves + moves that have a charge turn + Ally Switch all have additional tests. There are 8 skipped tests in here to account for the mechanics bugs reported so far in the new bug reports thread (up to post 61) as well as 11 passing tests to complement behavior and hopefully ensure behavior doesn't regress if any of the failing tests are tackled.

### DIFF
--- a/test/sim/abilities/steelyspirit.js
+++ b/test/sim/abilities/steelyspirit.js
@@ -1,0 +1,50 @@
+'use strict';
+
+const assert = require('./../../assert');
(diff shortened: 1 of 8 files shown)


Zarel:
Thanks!

PR #6721, opened 2020-05-20
Fix Farfetch'd / Sirfetch'd with new apostrophe

DaWoblefet:
Several checks (the one I care about being Leek) were not running correctly because we changed the name of the Pokemon entirely. I used toID() for comparison tests and copied the \u2019 for things like description strings.


data/aliases.ts
```diff
@@ -1032,7 +1032,7 @@ export const BattleAliases: {[alias: string]: string} = {
 	yadoran: "Slowbro",
 	koiru: "Magnemite",
 	reakoiru: "Magneton",
-	kamonegi: "Farfetch'd",
+	kamonegi: "Farfetch\u2019d",
```
scheibo, line 1035: I purposefully skipped this because gets toID-ed anyway (https://github.com/smogon/pokemon-showdown/blob/master/sim/dex.ts#L354)

data/items.ts
```diff
@@ -3005,14 +3005,14 @@ export const BattleItems: {[itemid: string]: ItemData} = {
 		},
 		spritenum: 475,
 		onModifyCritRatio(critRatio, user) {
-			if (["Farfetch'd", "Sirfetch'd"].includes(user.baseSpecies.baseSpecies)) {
+			if (["farfetchd", "sirfetchd"].includes(toID(user.baseSpecies.baseSpecies))) {
 				return critRatio + 2;
 			}
 		},
-		itemUser: ["Farfetch'd", "Sirfetch'd"],
+		itemUser: ["Farfetch\u2019d", "Sirfetch\u2019d"],
```
scheibo, line 3012: I missed this, thanks!

data/mods/gen2/items.ts
```diff
@@ -87,10 +87,10 @@ export const BattleItems: {[k: string]: ModdedItemData} = {
 	},
 	stick: {
 		inherit: true,
-		desc: "If held by a Farfetch'd, its critical hit ratio is always at stage 2. (25% crit rate)",
+		desc: "If held by a Farfetch\u2019d, its critical hit ratio is always at stage 2. (25% crit rate)",
 		onModifyCritRatioPriority: -1,
 		onModifyCritRatio(critRatio, user) {
-			if (user.species.name === 'Farfetch\'d') {
+			if (toID(user.species.name) === 'farfetchd') {
```
scheibo, line 93:
```suggestion
			if (user.species.id === 'farfetchd') {
```

I think?

data/mods/gen3/random-teams.ts
```diff
@@ -395,7 +395,7 @@ export class RandomGen3Teams extends RandomGen4Teams {
 		// First, the high-priority items
 		if (species.name === 'Ditto') {
 			item = this.sample(['Metal Powder', 'Quick Claw']);
-		} else if (species.name === 'Farfetch\'d') {
+		} else if (toID(species.name) === 'farfetchd') {
```
scheibo, line 398: species.id here as well, but the rest of the code seems to be checking names and it would look out of place to do IDs, so think actually might be a better place to use the \u

data/mods/gen4/items.ts
```diff
@@ -253,7 +253,7 @@ export const BattleItems: {[k: string]: ModdedItemData} = {
 	stick: {
 		inherit: true,
 		onModifyCritRatio(critRatio, user) {
-			if (user.species.name === 'Farfetch\'d') {
+			if (toID(user.species.name) === 'farfetchd') {
```
scheibo, line 256:
```suggestion
			if (user.species.id === 'farfetchd') {
```

data/mods/gen2/items.ts
DaWoblefet: I think the `name` check is specifically for Transform purposes. It might be equivalent here but I'm not sure.

data/mods/gen4/random-teams.ts
```diff
@@ -547,7 +547,7 @@ export class RandomGen4Teams extends RandomGen5Teams {
 			item = species.requiredItem;
 
 		// First, the extra high-priority items
-		} else if (species.name === 'Farfetch\'d') {
+		} else if (toID(species.name) === 'farfetchd') {
```
scheibo, line 550: see comment above, probably to stay checking name

data/aliases.ts
DaWoblefet: I swapped back.

data/mods/gen3/random-teams.ts
DaWoblefet: Swapped to \u2019 for the random-teams cases

data/mods/gen2/items.ts
scheibo: `Species` objects are immutable, so name and id should not be changed. I check - Ditto's baseSpecies stays the same (Ditto) after it transforms, but the species changes, so species.id should work

DaWoblefet: Done.

scheibo:
Thanks! Sorry I broke this!

Issue #6722, opened 2020-05-20
/om chat command: wrong URL


spixi:
The ```/om``` command still links to the Gen 7 OM board instead of the Gen 8 OM board.

PR #6723, opened 2020-05-20
Fix /om chat command: wrong URL (#6722)

spixi:


Zarel:
Oh, looks like the URL was changed for Sw/Sh. Thanks for the catch!

PR #6724, opened 2020-05-20
Add msg button

PartMan7:
(No longer needs the client change)

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1112,7 +1112,7 @@ export class CommandContext extends MessageContext {
 						(this.room.isPersonal || this.room.isPrivate === true) && !this.user.can('lock')) {
 						if ((tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '')
 							.replace(/\s*\bname\s*=\s*"\s*send\s*"/i, '')
-							.replace(new RegExp(`\\s*\\bvalue\\s*=\\s*"/(?:msg|pm|w|whisper)\\s+${by}\\s*,[^"]*"\s*`, 'i'), '')) !== 'button') {
+							.replace(new RegExp(`\\s*\\bvalue\\s*=\\s*"/(?:msg|pm|w|whisper)\\s+${this.user.id}\\s*,[^"]*"\s*`, 'i'), '')) !== 'button') {
 							this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
 							this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
 							return null;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1108,11 +1108,10 @@ export class CommandContext extends MessageContext {
 					}
 				}
 				if (tagName === 'button') {
-					if (tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '').trim() !== "button" && 
+					if (tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '').trim() !== "button" &&
 						(this.room.isPersonal || this.room.isPrivate === true) && !this.user.can('lock')) {
-						if ((tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '')
-							.replace(/\s*\bname\s*=\s*"\s*send\s*"/i, '')
-							.replace(new RegExp(`\\s*\\bvalue\\s*=\\s*"/(?:msg|pm|w|whisper)\\s+${this.user.id}\\s*,[^"]*"\s*`, 'i'), '')) !== 'button') {
+						if ((tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '').replace(/\s*\bname\s*=\s*"\s*send\s*"/i, '')
+							.replace(new RegExp(`\\s*\\bvalue\\s*=\\s*"/(?:msg|pm|w|whisper)\\s+${this.user.id}\\s*,[^"]*"\\s*`, 'i'), '')) !== 'button') {
 							this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
 							this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
 							return null;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1038,6 +1038,11 @@ export class CommandContext extends MessageContext {
 		// unknown URI, allow HTTP to be safe
 		return uri;
 	}
+	/**
+	 * This is a quick and dirty first-pass "is this good HTML" check. The full
+	 * sanitization is done on the client by Caja in `src/battle-log.ts`
+	 * `BattleLog.sanitizeHTML`.
+	 */
 	canHTML(htmlContent: string | null) {
 		htmlContent = ('' + (htmlContent || '')).trim();
 		if (!htmlContent) return '';
@@ -1064,9 +1069,9 @@ export class CommandContext extends MessageContext {
 			const stack = [];
 			for (const tag of tags) {
 				const isClosingTag = tag.charAt(1) === '/';
-				const tagContent = tag.slice(isClosingTag ? 2 : 1).toLowerCase();
-				const tagNameEndIndex = tagContent.search(/[\s/]/);
-				const tagName = tagContent.slice(0, tagNameEndIndex >= 0 ? tagNameEndIndex : undefined);
+				const tagContent = tag.slice(isClosingTag ? 2 : 1).replace(/\s+/, ' ').trim();
+				const tagNameEndIndex = tagContent.indexOf(' ');
+				const tagName = tagContent.slice(0, tagNameEndIndex >= 0 ? tagNameEndIndex : undefined).toLowerCase();
 				if (isClosingTag) {
 					if (LEGAL_AUTOCLOSE_TAGS.includes(tagName)) continue;
 					if (!stack.length) {
@@ -1094,17 +1099,22 @@ export class CommandContext extends MessageContext {
 						this.errorReply(`Images are not allowed in personal rooms.`);
 						return null;
 					}
-					if (!/width=([0-9]+|"[0-9]+")/i.test(tagContent) || !/height=([0-9]+|"[0-9]+")/i.test(tagContent)) {
+					if (!/width ?= ?(?:[0-9]+|"[0-9]+")/i.test(tagContent) || !/height ?= ?(?:[0-9]+|"[0-9]+")/i.test(tagContent)) {
 						// Width and height are required because most browsers insert the
 						// <img> element before width and height are known, and when the
 						// image is loaded, this changes the height of the chat area, which
 						// messes up autoscrolling.
 						this.errorReply(`All images must have a width and height attribute`);
+						this.errorReply(`This image doesn't: <${tagContent}>`);
 						return null;
 					}
-					const srcMatch = /src\s*=\s*"?([^ "]+)(\s*")?/i.exec(tagContent);
+					const srcMatch = / src ?= ?"?([^ "]+)(?: ?")?/i.exec(tagContent);
 					if (srcMatch) {
 						if (!this.canEmbedURI(srcMatch[1])) return null;
+					} else {
+						this.errorReply(`All images must have a src attribute with no spaces in the URL`);
+						this.errorReply(`This image doesn't: <${tagContent}>`);
+						return null;
 					}
 				}
 				if (tagName === 'button') {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1114,7 +1114,8 @@ export class CommandContext extends MessageContext {
 							.replace(new RegExp(
 								("\\s*\\bvalue\\s*=\\s*\"/(?:msg|pm|w|whisper)\\s+" +
 								this.user.id.split('').join('[^a-z0-9]*') +
-								"\\s*,[^\"]*\"\\s*"), 'i')) !== 'button') {
+								"\\s*,[^\"]*\"\\s*"), 'i')
+							) !== 'button') {
 							this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
 							this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
 							return null;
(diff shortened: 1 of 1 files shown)


PartMan7:
(Oh, and does the name 'whisper' work?)

Zarel:
It looks like the official name of the command is `/msg`. I'd probably prefer that one.

PartMan7:
Noted.

Zarel:
Hm... This wouldn't need a client change if this was just `<button name="send" value="/msg <username>, <contents>">`...

I think that'd make more sense than dedicated client-side support.

server/chat.ts
```diff
@@ -1108,13 +1108,19 @@ export class CommandContext extends MessageContext {
...
 				}
 				if (tagName === 'button') {
-					if (
-						(this.room.isPersonal || this.room.isPrivate === true) &&
-						!this.user.can('lock') && /<button[^>]/.test(htmlContent.toLowerCase().replace(/\s*style\s*=\s*"?[^"]*"\s*>/g, '>'))
-					) {
-						this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
-						this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
-						return null;
+					if (tagContent.replace(/\s*(?:name\s*=\s*"\s*msg\s*"|(?:value|style)\s*=\s*"[^"]*")/gi, '') !== "button") {
```
Zarel, line 1111: I don't think this is accounting for word boundaries correctly.

```suggestion
					if (tagContent.replace(/\b(?:name\s*=\s*"msg"|(?:value|style)\s*=\s*"[^"]*")/gi, '').trim() !== "button") {
```

server/chat.ts
```diff
@@ -1124,6 +1130,10 @@ export class CommandContext extends MessageContext {
 			}
 		}
 
+		htmlContent = htmlContent.replace(/button[^>]*?name\s*=\s*"\s*msg\s*"[^>]*>/gi, match => {
+			return match.replace(/name\s*=\s*"msg"/i, `name="send"`).replace(/value\s*=\s*"\s*/i, `value="/msg ${this.user.id}, `);
+		});
```
Zarel, line 1135:
```suggestion
		htmlContent = htmlContent.replace(/<button\s[^>]*\bname\s*=\s*"msg"[^>]*>/gi, match => (
			match.replace(/\bname\s*=\s*"msg"/i, `name="send"`).replace(/\bvalue\s*=\s*"\s*/i, `value="/msg ${this.user.id}, `)
		));
```

PartMan7:
Sorry about those.

Is there any particular reason for `match => (` over `match => {`?

Zarel:
> Is there any particular reason for `match => (` over `match => {`?

I find `a => b` to be significantly more readable than `a => { return b; }`, and there's no reason for the latter as far as I can tell.

Zarel:
On reflection, I think this should probably specifically specify the bot that should be PMed.

At which point maybe it should just be literally a validation that it starts with `<button name="send" value="/msg <username>,` and that the `<username>` matches the username doing the sending.

Sorry to make you change this so much, by the way. It just took me a while to realize the best design here.

PartMan7:
I mean, I wrote it so badly that you ended up writing more of it than I did. Sorry about that.

Anyways, I wrote a validator that yields:
``1) As Permitted: 
	Input : "button name=\"send\" value=\"normal\""
	Pass: Yep.

2) As Permitted: 
	Input : "button name=\"send\" value=\"/normal\""
	Pass: Yep.

3) As Permitted: 
	Input : "button name=\"send\" value=\"!normal\""
	Pass: Yep.

4) As Permitted: 
	Input : "button name=\"send\" value=\" /normal\""
	Pass: Yep.

5) As Permitted: 
	Input : "button name=\"send\" value=\"/msg PartBot, normal\""
	Pass: Yep.

6) As Permitted: 
	Input : "button name=\"send\" value=\"/msg normal\""
	Pass: Yep.

7) As Permitted: 
	Input : "button name=\"send\" value=\"/msg NotPartBot, normal\""
	Pass: Yep.

8) As Permitted: 
	Input : "button name=\"send\" value=\"/ms normal\""
	Pass: Yep.

9) As Permitted: 
	Input : "button bname=\"send\" value=\"/msg normal\""
	Pass: Yep.

10) As Permitted: 
	Input : "button name=\"receive\" value=\"normal\""
	Pass: Yep.

11) As non-Permitted: 
	Input : "button name=\"send\" value=\"normal\""
	Pass: Nope.

12) As non-Permitted: 
	Input : "button name=\"send\" value=\"/normal\""
	Pass: Nope.

13) As non-Permitted: 
	Input : "button name=\"send\" value=\"!normal\""
	Pass: Nope.

14) As non-Permitted: 
	Input : "button name=\"send\" value=\" /normal\""
	Pass: Nope.

15) As non-Permitted: 
	Input : "button name=\"send\" value=\"/msg PartBot, normal\""
	Pass: Yep.

16) As non-Permitted: 
	Input : "button name=\"send\" value=\"/msg normal\""
	Pass: Nope.

17) As non-Permitted: 
	Input : "button name=\"send\" value=\"/msg NotPartBot, normal\""
	Pass: Nope.

18) As non-Permitted: 
	Input : "button name=\"send\" value=\"/ms normal\""
	Pass: Nope.

19) As non-Permitted: 
	Input : "button bname=\"send\" value=\"/msg normal\""
	Pass: Nope.

20) As non-Permitted: 
	Input : "button name=\"receive\" value=\"normal\""
	Pass: Nope.
``

I'll edit the PR with it.

Zarel:
Wow, that's great! Do you think you could add those as unit tests? If not, I can do it myself.

PartMan7:
I don't really know how to add unit tests, sorry.
Also, the sample size I used doesn't cover every use case - should I compile a list of those?

Zarel:
You could add some more if you wanted, yeah.

PartMan7:
One issue with the current validator is that it doesn't use userID - so ``<button name="send" value="/msg PartBot, Test">`` would be denied if my username were Part Bot and not PartBot. I'll address this; changing this to a draft until then.

PartMan7:
This version is a bit hacky - it tests for ``this.user.id.split('').join('[^a-z0-9]*')`` instead of ``this.id``.

PartMan7:
Also, dunno if this is where I should be asking, but how does PS! normally format multi-line statements?

server/chat.ts
```diff
@@ -1108,13 +1108,18 @@ export class CommandContext extends MessageContext {
...
+						if (tagContent.replace(/\s*\bstyle\s*=\s*"[^"]*"\s*/i, '').replace(/\s*\bname\s*=\s*"\s*send\s*"/i, '')
+							.replace(new RegExp(
+								("\\s*\\bvalue\\s*=\\s*\"/(?:msg|pm|w|whisper)\\s+" +
+								this.user.id.split('').join('[^a-z0-9]*') +
+								"\\s*,[^\"]*\"\\s*"), 'i')
+							) !== 'button') {
+							this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
+							this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
+							return null;
+						}
```
Zarel, line 1122: Honestly, the `\s*` stuff is annoying me.

I would update the `tagContent` definition like this:

```ts
				const tagContent = tag.slice(isClosingTag ? 2 : 1).replace(/\s+/, ' ').trim();
				const tagNameEndIndex = tagContent.indexOf(' ');
				const tagName = tagContent.slice(0, tagNameEndIndex >= 0 ? tagNameEndIndex : undefined).toLowerCase();
```

(This also removes the lowercase-conversion of `tagContent`, which is important later.

Then all runs of whitespace are guaranteed to be a single space, and then you can do:

```suggestion
					let attrs = tagContent.slice(6).replace(/ style ?= ?"[^"]*"/i, '');
					if (attrs && (this.room.isPersonal || this.room.isPrivate === true) && !this.user.can('lock')) {
						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
						let legalButton = !buttonName;
						if (buttonName === 'send' && buttonValue.startsWith('/msg ')) {
							const [pmTarget] = buttonValue.slice(5).split(',');
							if (room.auth[toID(pmTarget)] === '*') {
								legalButton = true;
							}
						}
						if (!legalButton) {
							this.errorReply(`You do not have permission to use most buttons. Here are the two types you can use:`);
							this.errorReply(`1. Linking to a room: <a href="/roomid"><button>go to a place</button></a>`);
							this.errorReply(`2. Sending a message to a bot: <button name="send" value="/msg, BOT_USERNAME, MESSAGE">send the thing</button>`);
							return null;
						}
```

PartMan7:
And in the end, I still ended up making you write the whole thing, sorry. ;-;

Zarel:
Don't worry about making me write the whole thing! Your starting point is actually really useful for helping me realize what needs to be fixed and how it can all be improved.

Zarel:
Anyway, you still need the updated `tagContent` definition to make this work:

```ts
				const tagContent = tag.slice(isClosingTag ? 2 : 1).replace(/\s+/, ' ').trim();
				const tagNameEndIndex = tagContent.indexOf(' ');
				const tagName = tagContent.slice(0, tagNameEndIndex >= 0 ? tagNameEndIndex : undefined).toLowerCase();
```

server/chat.ts
```diff
@@ -1108,13 +1108,22 @@ export class CommandContext extends MessageContext {
...
-						this.errorReply(`You do not have permission to use scripted buttons (buttons with attributes other than style="...") in HTML.`);
-						this.errorReply(`If you just want to link to a room, you can do this: <a href="/roomid"><button>button contents</button></a>`);
-						return null;
+					if ((this.room.isPersonal || this.room.isPrivate === true) && !this.user.can('lock')) {
+						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
+						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
+						let legalButton = !buttonName;
+						if (buttonName === 'send' && buttonValue?.startsWith('/msg ')) {
+							const [pmTarget] = buttonValue.slice(5).split(',');
+							if (room.auth[toID(pmTarget)] === '*') {
```
Zarel, line 1117:
```suggestion
							if (this.room.auth?.[toID(pmTarget)] === '*') {
```

Zarel:
@PartMan7 I think you missed the part I said about the updated `tagContent` definition. I'll add it myself, I guess.

PartMan7:
Oh, snap, I must've missed that. Sorry about that. 'o.o

PartMan7:
What is the current CI error about?

Zarel:
GitHub Actions went down. I think retrying it should fix it.

server/chat.ts
```diff
@@ -1094,27 +1099,41 @@ export class CommandContext extends MessageContext {
...
+							if (this.room.auth?.[toID(pmTarget)] === '*') {
+								legalButton = true;
+							}
+						}
+						if (!legalButton) {
+							this.errorReply(`You do not have permission to use most buttons. Here are the two types you're allowed can use:`);
+							this.errorReply(`1. Linking to a room: <a href="/roomid"><button>go to a place</button></a>`);
+							this.errorReply(`2. Sending a message to a Bot: <button name="send" value="/msg, BOT_USERNAME, MESSAGE">send the thing</button>`);
+							return null;
+						}
```
Zarel, line 1136: I never liked the `legalButton` variable, and better error messages _would_ preclude it. Therefore:

```suggestion
						if (buttonName === 'send' && buttonValue?.startsWith('/msg ')) {
							const [pmTarget] = buttonValue.slice(5).split(',');
							if (this.room.auth?.[toID(pmTarget)] !== '*') {
								this.errorReply(`This button is not allowed: <${tagContent}>`);
								this.errorReply(`Your scripted button can't send PMs to ${pmTarget}, because that user is not a roombot.`);
								return null;
							}
						} else if (buttonName) {
							this.errorReply(`This button is not allowed: <${tagContent}>`);
							this.errorReply(`You do not have permission to use most buttons. Here are the two types you're allowed can use:`);
							this.errorReply(`1. Linking to a room: <a href="/roomid"><button>go to a place</button></a>`);
							this.errorReply(`2. Sending a message to a Bot: <button name="send" value="/msg, BOT_USERNAME, MESSAGE">send the thing</button>`);
							return null;
						}
```

server/chat.ts
```diff
@@ -1094,27 +1099,41 @@ export class CommandContext extends MessageContext {
...
 						return null;
 					}
-					if (!/width=([0-9]+|"[0-9]+")/i.test(tagContent) || !/height=([0-9]+|"[0-9]+")/i.test(tagContent)) {
+					if (!/width ?= ?(?:[0-9]+|"[0-9]+")/i.test(tagContent) || !/height ?= ?(?:[0-9]+|"[0-9]+")/i.test(tagContent)) {
 						// Width and height are required because most browsers insert the
 						// <img> element before width and height are known, and when the
 						// image is loaded, this changes the height of the chat area, which
 						// messes up autoscrolling.
 						this.errorReply(`All images must have a width and height attribute`);
+						this.errorReply(`This image doesn't: <${tagContent}>`);
```
Zarel, line 1108:
```suggestion
						this.errorReply(`This image is missing a width/height attribute: <${tagContent}>`);
						this.errorReply(`Images without predefined width/height cause problems with scrolling because loading them changes their height.`);
```

server/chat.ts
```diff
@@ -1094,27 +1099,41 @@ export class CommandContext extends MessageContext {
...
 						return null;
 					}
-					const srcMatch = /src\s*=\s*"?([^ "]+)(\s*")?/i.exec(tagContent);
+					const srcMatch = / src ?= ?"?([^ "]+)(?: ?")?/i.exec(tagContent);
 					if (srcMatch) {
 						if (!this.canEmbedURI(srcMatch[1])) return null;
+					} else {
+						this.errorReply(`All images must have a src attribute with no spaces in the URL`);
+						this.errorReply(`This image doesn't: <${tagContent}>`);
```
Zarel, line 1116:
```suggestion
						this.errorReply(`This image has a broken src attribute: <${tagContent}>`);
						this.errorReply(`The src attribute must exist and have no spaces in the URL`);
```

server/chat.ts
```diff
@@ -1094,27 +1099,41 @@ export class CommandContext extends MessageContext {
 						this.errorReply(`Images are not allowed in personal rooms.`);
```
Zarel, line 1099:
```suggestion
						this.errorReply(`This tag is not allowed: <${tagContent}>`);
						this.errorReply(`Images are not allowed in personal rooms.`);
```

PartMan7:
>sees 7 notifications
:scared_noises:

Zarel:
Heh, I just thought it would be a good time to improve the error messages on this.

Zarel:
Thanks!

PR #6726, opened 2020-05-21
Scavengers: Typescript

sparkychildcharlie:
痛い _にゃの_ です！タイプスクリプトが大嫌い！

Convert scavengers to typescript!  Thanks to Zarel-sensei for his guidance - I hope I've done a good job! Please let me know if I missed anything!

<sub>PS. scavengers > trivia</sub>


server/chat-plugins/scavenger-games.ts
```diff
@@ -58,20 +78,20 @@ class Leaderboard {
...
+			this.visualize('points').then(d => {
+				const data = d as AnyObject[];
+				const display = `<div class="ladder" style="overflow-y: scroll; max-height: 170px;"><table style="width: 100%"><tr><th>Rank</th><th>Name</th><th>Points</th></tr>${data.map(line =>
+					`<tr><td>${line.rank}</td><td>${line.name}</td><td>${line.points}</td></tr>`).join('')
+				}</table></div>`;
 				resolve(display);
 			});
 		});
 	}
 }
```
Zarel, line 92: I would make this an async function:

```ts
	async htmlLadder(): Promise<string> {
		const data = await this.visualize('points') as AnyObject[];
		return `<div class="ladder" style="overflow-y: scroll; max-height: 170px;"><table style="width: 100%"><tr><th>Rank</th><th>Name</th><th>Points</th></tr>${data.map(line =>
			`<tr><td>${line.rank}</td><td>${line.name}</td><td>${line.points}</td></tr>`).join('')
		}</table></div>`;
	}
```

server/chat-plugins/scavengers.ts
```diff
@@ -426,16 +539,16 @@ class ScavengerHunt extends Rooms.RoomGame {
 		this.room.add(message).update();
 	}
 
-	runEvent(event_id, ...args) {
+	runEvent(event_id: string, ...args: any[]) {
 		let events = this.mods['on' + event_id];
 		if (!events) return;
 
-		events = events.sort((a, b) => b.priority - a.priority);
+		events = events.sort((a: any, b: any) => b.priority - a.priority);
```
Zarel, line 546: The `: any` should be unnecessary here. Can you revert this and show me the error? Is `events` not being inferred correctly?

```suggestion
		events = events.sort((a, b) => b.priority - a.priority);
```

server/chat-plugins/scavengers.ts
```diff
@@ -763,54 +907,74 @@ class ScavengerHunt extends Rooms.RoomGame {
 		});
 
 		if (filtered) return "Please do not leak the answer. Use /scavenge [guess] to submit your guess instead.";
```
Zarel, line 909: I think we should have a `/guess` function which aliases to `/choose`, for this.

server/chat-plugins/scavengers.ts
sparkychildcharlie: oh lol there is none now - there was at some point when i didnt set everything up at the start

server/chat-plugins/scavengers.ts
sparkychildcharlie: traditionally scavengers has always used /scavenge [answer] for guessing before the advent of the roomgame system, which is why we're keeping the same commands.

server/chat-plugins/scavengers.ts
```diff
@@ -27,30 +47,22 @@ const HOST_DATA_FILE = 'config/chat-plugins/scavhostdata.json';
 const PLAYER_DATA_FILE = 'config/chat-plugins/scavplayerdata.json';
 const DATABASE_FILE = 'config/chat-plugins/scavhunts.json';
 
-const SCAVENGE_REGEX = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
+const SCAV_REGEX_STRING = "^((?:\\s)?(?:[^\\w]+)|\\s\\/)?(?:\\s)?" +
+	"(?:s\\W?cavenge|s\\W?cav(?:engers)? guess|d\\W?t|d\\W?ata|d\\W?etails|g\\W?(?:uess)?|v)\\b";
+const SCAVENGE_REGEX = new RegExp(SCAV_REGEX_STRING, 'i');
```
Zarel, line 52: Why this? The old regex should definitely already bypass the line length limit.

```suggestion
// a regex of some of all the possible slips for leaks.
const ACCIDENTAL_LEAKS = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i;
```

Also I would probably rename this `ACCIDENTAL_LEAKS` to make it clearer what it means.

server/chat-plugins/scavengers.ts
```diff
@@ -1686,21 +1972,25 @@ const commands = {
 
 		if (!sortingFields.includes(sortMethod)) sortMethod = 'finish'; // default sort method
 
-		PlayerLeaderboard.visualize(sortMethod).then(raw => {
+		PlayerLeaderboard.visualize(sortMethod).then(data => {
```
Zarel, line 1975: This should also be refactored into an async function. Can you do that for me? I can guide you through it if you have trouble.

server/chat-plugins/scavengers.ts
```diff
@@ -1468,88 +1731,104 @@ const commands = {
...
+			room.chatRoomData.scavSettings = room.scavSettings;
 			Rooms.global.writeChatRoomData();
 		}
 		this.privateModAction(`(${user.name} has set the points awarded for hosting regular scavenger hunts to ${points})`);
-		this.modlog('SCAV SETHOSTPOINTS', null, points);
+		this.modlog('SCAV SETHOSTPOINTS', null, `${points}`);
 
 		// double modnote in scavs room if it is a subroomgroupchat
-		if (room.parent && !room.chatRoomData) {
+		if (room.parent && !room.chatRoomData && scavsRoom) {
```
Zarel, line 1784: Don't you already have a `"This command can only be used in the scavengers room."` error up above? I would maybe combine those? Make a `getScavsRoom` function that errors if it doesn't exist.

server/chat-plugins/scavengers.ts
```diff
@@ -1468,88 +1731,104 @@ const commands = {
...
+			scavsRoom.sendMods(`(${user.name} has set the points awarded for blitz for ${gameType} hunts to ${blitzPoints} in <<${room.roomid}>>.)`);
+			scavsRoom.roomlog(`(${user.name} has set the points awarded for blitz for ${gameType} hunts to ${blitzPoints} in <<${room.roomid}>>.)`);
 		}
 	},
 
 	sethostpoints(target, room, user) {
-		if (room.roomid !== 'scavengers' && !(room.parent && room.parent.roomid === 'scavengers')) return this.errorReply("This command can only be used in the scavengers room.");
+		if (room.roomid !== 'scavengers' && !(room.parent && room.parent.roomid === 'scavengers')) {
+			return this.errorReply("This command can only be used in the scavengers room.");
+		}
```
Zarel, line 1763: Maybe make it like this:

```ts
function getScavsRoom(room?: Room | null) {
	if (!room) return Rooms.getRoom('scavengers');
	if (room.roomid === 'scavengers') return room;
	if (room.parent?.roomid === 'scavengers') return room.parent;
	return null;
}
```

And then this would be:

```suggestion
		const scavsRoom = getScavsRoom(room);
		if (!scavsRoom) {
			return this.errorReply("This command can only be used in the scavengers room.");
		}
```

server/chat-plugins/scavengers.ts
```diff
@@ -747,14 +891,14 @@ class ScavengerHunt extends Rooms.RoomGame {
 
 	onUpdateConnection() {}
 
-	onChatMessage(msg) {
-		let msgId = toID(msg);
+	onChatMessage(msg: string) {
+		let msgId = toID(msg) as string;
 
 		const commandMatch = msg.match(SCAVENGE_REGEX);
```
Zarel, line 897: And then this:

```suggestion
		const commandMatch = ACCIDENTAL_LEAKS.exec(msg);
```

server/chat-plugins/scavengers.ts
```diff
@@ -747,14 +891,14 @@ class ScavengerHunt extends Rooms.RoomGame {
...
 		if (commandMatch) msgId = msgId.slice(toID(commandMatch[0]).length);
```
Zarel, line 898: If you're not going to use the return value of `match`/`exec`, you should be using `test` for performance:

```suggestion
		if (ACCIDENTAL_LEAKS.test(msg)) {
			msgId = msgId.slice(toID(commandMatch[0]).length);
		}
```

server/chat-plugins/scavengers.ts
```diff
@@ -747,14 +891,14 @@ class ScavengerHunt extends Rooms.RoomGame {
...
 
 		const filtered = this.questions.some(q => {
-			return q.answer.some(a => {
+			return q.answer.some((a: string) => {
```
Zarel, line 901: Apparently this is because `this.questions` is an array of `AnyObject`. Can you add the type definition now?

Zarel: (It's not a big deal if you don't want to.)

server/chat-plugins/scavengers.ts
sparkychildcharlie: This was only for typescript to shut up since scavsRoom (Rooms.get()) could be null, and typescript complained.

sparkychildcharlie:
async function lookingForHeadpats() {
    await Zarels_approval();
}

(I kept scavsRoom as is unless you have another reason to do it since the variable is used outside of commands as well)

sparkychildcharlie:
I suppose i could also do 
```const scavsRoom = Rooms.get('scavengers') as Room;``` 
(remove the null possibility from typescript checks) and remove all the extra ``&& scavsRoom`` checks elsewhere

File: server/chat-plugins/scavengers.ts (hunk 6 above)
Comment (Zarel) [reply server/chat-plugins/scavengers.ts:1782]: I know. I think it should check for it to be `null` up in the same place as the "can only be used in scavengers" error message, and it should be a combined check. I posted some code for that earlier.

**Comment by sparkychildcharlie** [comment]
> ```
> async function lookingForHeadpats() {
>     await Zarels_approval();
> }
> ```
> 
> (I kept scavsRoom as is unless you have another reason to do it since the variable is used outside of commands as well)

i think all the linting errors should be gone this time

File: server/chat-plugins/scavenger-games.ts [hunk 11]
```diff
@@ -7,29 +7,49 @@
...
-		const userid = toID(name);
+	addPoints(name: string, aspect: string, points: number, noUpdate?: boolean) {
+		const userid: string = toID(name);
 
 		if (!userid || userid === 'constructor' || !points) return this;
 		if (!this.data[userid]) this.data[userid] = {name: name};
 
 		if (!this.data[userid][aspect]) this.data[userid][aspect] = 0;
-		this.data[userid][aspect] += points;
+		this.data[userid][aspect] = this.data[userid][aspect] = points;
```
Zarel, line 45: This seems like it would do something completely different from the earlier code (and also looks like it's clearly a bug). What was your intent here? Did the old code not work?

sparkychildcharlie: That was a mistake - I dont even know how that happened.  I'm not enough of a furbrain to do something like this.

server/chat-plugins/scavenger-games.ts
```diff
@@ -199,14 +227,15 @@ const MODES = {
 				this.allowRenames = false; // don't let people change their name in the middle of the hunt.
 			},
 
-			onJoin(user) {
-				if (this.room.scavgame.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
+			onJoin(user: User) {
+				if (this.room.scavgame && this.room.scavgame.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
```
Zarel, line 231:
```suggestion
				if (this.room.scavgame?.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
```

sparkychildcharlie: oh right optional chaining....

server/chat-plugins/scavengers.ts
```diff
@@ -27,30 +47,23 @@ const HOST_DATA_FILE = 'config/chat-plugins/scavhostdata.json';
 const PLAYER_DATA_FILE = 'config/chat-plugins/scavplayerdata.json';
 const DATABASE_FILE = 'config/chat-plugins/scavhunts.json';
 
-const SCAVENGE_REGEX = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
+/* eslint-disable max-len */
+const ACCIDENTAL_LEAKS = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
+/* eslint-enable max-len */
```
Zarel, line 52:
```suggestion
const ACCIDENTAL_LEAKS = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
```

The max-len regex should allow this. Let me see what's wrong.

server/chat-plugins/scavenger-games.ts
```diff
@@ -125,7 +143,12 @@ const TWISTS = {
 			const now = Date.now();
 			const time = Chat.toDurationString(now - this.startTime, {hhmmss: true});
 
-			const blitz = (((this.room.blitzPoints && this.room.blitzPoints[this.gameType]) || this.gameType === 'official') && now - this.startTime <= 60000);
+			const blitz = (
+				(
+					(this.room.scavSettings.hostPoints && this.room.scavSettings.hostPoints[this.gameType]) ||
+					this.gameType === 'official'
+				) && now - this.startTime <= 60000
+			);
```
Zarel, line 151:
```suggestion
			const blitz = now - this.startTime <= 60000 &&
				(this.room.scavSettings.hostPoints?.[this.gameType] || this.gameType === 'official');
```

server/chat-plugins/scavenger-games.ts
```diff
@@ -269,24 +298,25 @@ const MODES = {
 			},
 
 			onJoin(user) {
-				if (this.room.scavgame.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
+				if (this.room.scavgame && this.room.scavgame.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
```
Zarel, line 301:
```suggestion
				if (this.room.scavgame?.playerlist && !this.room.scavgame.playerlist.includes(user.id)) {
```

server/chat-plugins/scavenger-games.ts
```diff
@@ -269,24 +298,25 @@ const MODES = {
...
 					user.sendTo(this.room, 'You are not allowed to join this scavenger hunt.');
 					return true;
 				}
 			},
 
 			onAfterEnd() {
+				if (!this.room.scavgame) return; // should never happen, but typescript is happy
```
Zarel, line 308: If it should never happen, consider `throw`ing instead:

```suggestion
				if (!this.room.scavgame) throw new Error("should never happen");
```

It might also make sense to make a temporary variable to access, instead:

```ts
				const game = this.room.scavgame!;
```

server/chat-plugins/scavengers.ts
```diff
@@ -1775,27 +2062,33 @@ const commands = {
...
 		let cmd;
 		[cmd, target] = Chat.splitFirst(target, ' ');
 		cmd = toID(cmd);
 
-		if (cmd === '' || cmd === 'help' || !['addhunt', 'list', 'removehunt', 'addhint', 'removehint', 'autostart'].includes(cmd)) {
+		if (
+			cmd === '' ||
+			cmd === 'help' ||
+			!['addhunt', 'list', 'removehunt', 'addhint', 'removehint', 'autostart'].includes(cmd)
+		) {
```
Zarel, line 2077:
```suggestion
		if (!['addhunt', 'list', 'removehunt', 'addhint', 'removehint', 'autostart'].includes(cmd)) {
```

The tests for `''` or `'help'` seem useless here.

server/chat-plugins/scavengers.ts
```diff
@@ -1853,12 +2159,16 @@ const commands = {
 			return this.privateModAction(`Hint #${hintNumber} was removed from Recycled hunt #${huntNumber} question #${questionNumber}.`);
 		} else if (cmd === 'autostart') {
 			if (params[0] !== 'on' && params[0] !== 'off') return this.errorReply(usageMessages[cmd]);
-			if (params[0] === 'on' && room.addRecycledHuntsToQueueAutomatically || params[0] === 'off' && !room.addRecycledHuntsToQueueAutomatically) return this.errorReply(`Autostarting recycled hunts is already ${room.addRecycledHuntsToQueueAutomatically ? 'on' : 'off'}.`);
-			room.addRecycledHuntsToQueueAutomatically = !room.addRecycledHuntsToQueueAutomatically;
+			if (params[0] === 'on' && room.scavSettings.addRecycledHuntsToQueueAutomatically || params[0] === 'off' &&
+				!room.scavSettings.addRecycledHuntsToQueueAutomatically
+			) {
```
Zarel, line 2164:
```suggestion
			if ((params[0] === 'on') === !!room.scavSettings.addRecycledHuntsToQueueAutomatically) {
```

server/chat-plugins/scavengers.ts
```diff
@@ -1833,12 +2126,25 @@ const commands = {
 
 		const cmdsNeedingQuestionNumber = ['addhint', 'removehint'];
 		if (cmdsNeedingQuestionNumber.includes(cmd)) {
-			if (isNaN(questionNumber) || questionNumber <= 0 || questionNumber > scavengersData.recycledHunts[huntNumber - 1].questions.length) return this.errorReply("You specified an invalid question number.");
+			if (
+				isNaN(questionNumber) ||
+				questionNumber <= 0 ||
+				questionNumber > scavengersData.recycledHunts[huntNumber - 1].questions.length
+			) {
```
Zarel, line 2133:
```suggestion
			const numQuestions = scavengersData.recycledHunts[huntNumber - 1].questions.length;
			if (isNaN(questionNumber) || questionNumber <= 0 || questionNumber > numQuestions) {
```

Zarel:
Okay, I've skimmed everything else and it looks fine. It should be fine to merge after you take care of everything I've pointed out so far.

server/chat-plugins/scavengers.ts
Zarel: This is my secret trick for "concisely see if two settings match" - just see if they're equal to each other. :3

server/chat-plugins/scavengers.ts
```diff
@@ -27,30 +47,27 @@ const HOST_DATA_FILE = 'config/chat-plugins/scavhostdata.json';
 const PLAYER_DATA_FILE = 'config/chat-plugins/scavplayerdata.json';
 const DATABASE_FILE = 'config/chat-plugins/scavhunts.json';
 
-const SCAVENGE_REGEX = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
+const ACCIDENTAL_LEAKS = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i; // a regex of some of all the possible slips for leaks.
```
Zarel, line 50:
```suggestion
const ACCIDENTAL_LEAKS = /^((?:\s)?(?:\/{2,}|[^\w/]+)|\s\/)?(?:\s)?(?:s\W?cavenge|s\W?cav(?:engers)? guess|d\W?t|d\W?ata|d\W?etails|g\W?(?:uess)?|v)\b/i;
```

Zarel:
Looks great! Thanks!

PR #6727, opened 2020-05-21
Add Sound flag to Howl

PartMan7:
(In Gen8, according to both Bulbapedia and Serebii, Howl is a Sound move)


Zarel:
@Marty-D or @DaWoblefet? Was this just an oversight?

Marty-D:
Yes, Howl is now the only sound move that doesn't also bypass substitutes, besides Clangorous Soul (which is irrelevant to its functioning). This just has to be undone in the Gen 7 mod.

Marty-D:
Thanks!

PR #6728, opened 2020-05-21
Increase autojoin limit to 15

tmagicturtle:
Per https://www.smogon.com/forums/threads/increase-chatroom-autojoin-limit.3657450 and mc56556's conversations with dev staff

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -290,7 +290,7 @@ export const commands: ChatCommands = {
 	autojoin(target, room, user, connection) {
 		const targets = target.split(',');
 		if (targets.length > 16 || connection.inRooms.size > 1) {
-			connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
+			return connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
 		}
 		Rooms.global.autojoinRooms(user, connection);
 		const autojoins: string[] = [];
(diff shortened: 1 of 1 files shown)


Zarel:
Yep, thanks!

Keep in mind that this will also need to be updated in client.

server/chat-commands/moderation.ts
```diff
@@ -289,7 +289,7 @@ export const commands: ChatCommands = {
 	'!autojoin': true,
 	autojoin(target, room, user, connection) {
 		const targets = target.split(',');
-		if (targets.length > 11 || connection.inRooms.size > 1) return;
+		if (targets.length > 16 || connection.inRooms.size > 1) return;
```
Zarel, line 292: This should probably popup an error message instead of silently failing? I'm sure bots would appreciate that.

Zarel:
```suggestion
		if (targets.length > 16 || connection.inRooms.size > 1) {
			connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
		}
```

server/chat-commands/moderation.ts
```diff
@@ -289,7 +289,9 @@ export const commands: ChatCommands = {
 	'!autojoin': true,
 	autojoin(target, room, user, connection) {
 		const targets = target.split(',');
-		if (targets.length > 11 || connection.inRooms.size > 1) return;
+		if (targets.length > 16 || connection.inRooms.size > 1) {
+			connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
```
Zarel, line 293:
```suggestion
			return connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
```

Zarel:
I think you missed the request to add an error message. No problem, I can do that myself.

PR #6729, opened 2020-05-21
Fix Visual Lottery Bug

PartMan7:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8474970


Zarel:
Thanks!

PR #6733, opened 2020-05-23
Fix Effectiveness showing undefined

PartMan7:
(change .species => .name)


Zarel:
Thanks for finding this!

PR #6734, opened 2020-05-23
Scavengers: initialize all settings as null

sparkychildcharlie:
Make scavs hotpatchable

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -2026,7 +2026,7 @@ const ScavengerCommands: ChatCommands = {
 		'': 'update',
 		'update'(target, room, user) {
 			if (!this.can('declare', null, room) || !getScavsRoom(room)) return false;
-			const settings = room.scavSettings.scavmod || {};
+			const settings = room.scavSettings?.scavmod || {};
 
 			this.sendReply(`|uhtml${this.cmd === 'update' ? 'change' : ''}|scav-modsettings|<div class=infobox><strong>Scavenger Moderation Settings:</strong><br /><br />` +
 				`<button name=send value='/scav modsettings ipcheck toggle'><i class="fa fa-power-off"></i></button> Multiple connection verification: ${settings.ipcheck ? 'ON' : 'OFF'}` +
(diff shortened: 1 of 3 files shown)


server/chat-plugins/scavengers.ts
```diff
@@ -800,16 +795,16 @@ export class ScavengerHunt extends Rooms.RoomGame {
...
+		if (this.room.scavgame || this.room.scavSettings?.scavQueueDisabled) return; // don't run the queue for child games.
 		// prepare the next queue'd game
 		if (this.room.scavQueue && this.room.scavQueue.length) {
 			setTimeout(() => {
 				const room = Rooms.get(roomid) as ChatRoom;
-				if (!room || room.game || !room.scavQueue.length) return;
+				if (!room || room.game || !room.scavQueue?.length) return;
 
-				const next = room.scavQueue.shift();
+				const next = room.scavQueue?.shift();
```
xfix, line 805: Is `?` necessary here, considering you just have checked `room.scavQueue.length`?

sparkychildcharlie: it complained when i wrote it locally.. but probably no

server/chat-plugins/scavengers.ts
```diff
@@ -800,16 +795,16 @@ export class ScavengerHunt extends Rooms.RoomGame {
...
+		if (this.room.scavgame || this.room.scavSettings?.scavQueueDisabled) return; // don't run the queue for child games.
 		// prepare the next queue'd game
 		if (this.room.scavQueue && this.room.scavQueue.length) {
 			setTimeout(() => {
 				const room = Rooms.get(roomid) as ChatRoom;
-				if (!room || room.game || !room.scavQueue.length) return;
+				if (!room || room.game || !room.scavQueue?.length) return;
 
 				const next = room.scavQueue.shift();
 				if (!next) return; // shouldn't happen, but it makes typescript happy.
```
Zarel, line 806:
```suggestion
				const next = room.scavQueue.shift()!;
```

(this is the correct way to make TS happy)

PR #6736, opened 2020-05-23
Add 1v1 samples link

TheMezStrikes:


Zarel:
This seems correct but I don't know which links belong there. @TheImmortal?

KrisXV:
This is correct

Zarel:
I suppose that's good enough for me. Thanks for the fix!

PR #6738, opened 2020-05-24
Life Orb + Emergency Exit Interaction Fix

tmagicturtle:
Life Orb should trigger Emergency Exit.
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-2#post-8465511

I apologize if this isn't the cleanest way to do it. I initially had placed the Emergency Exit runEvent inside of Life Orb's code but I know that's not an acceptable way to accomplish this. This passes all tests and I believe should be okay, but if there's tweaks I should make to it, please let me know.


Zarel:
I thought I originally wrote it this way because I heard reports that LO didn't trigger Emergency Exit? Ugh, I'll need to revisit the mechanic at some point.

Zarel:
Thanks, though!

PR #6739, opened 2020-05-24
Scavgames KO Games: Allow finishers to continue and announce correctly

cheese-testing:
Previously, the code only allowed for those who were not on the playerlist to join hunts, when it should be only those who are on the playerlist.
Also, the first end-of-round announcement announced those who were in the game, rather than only those who had completed it.


xfix:
Oh yeah, this broke during TypeScript migration.

PR #6740, opened 2020-05-24
Youtube: add plugin for viewing videos/channels

mia-pi-git:
This is mostly for the youtube room. However, /requestapproval and /link are meant for all rooms, since they're very commonly used through Kid A and very useful, and can be streamlined a bit here (they're in this plugin so they can make use of the Youtube interface.) 
Specifics are here https://www.smogon.com/forums/threads/youtube-channel-database.3664726/.
This has been approved by both Kie and A Random Duck.
As for the Config setting for the api key, i have a key that can be used, and can share privately for someone who can set that.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -20,7 +20,7 @@ try {
 }
 
 export class YoutubeInterface {
-	interval: NodeJS.Timeout | null;
+	interval: NodeJS.Timer | null;
 	constructor() {
 		this.interval = null;
 	}
@@ -157,9 +157,10 @@ export class YoutubeInterface {
 		}
 		return rearranged;
 	}
-};
+}
 
 const YouTube = new YoutubeInterface();
+
 export const commands: ChatCommands = {
 	randchannel(target, room, user) {
 		if (room.roomid !== 'youtube') return this.errorReply(`This command can only be used in the YouTube room.`);
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -200,6 +200,7 @@ export const commands: ChatCommands = {
 			const id = YouTube.channelSearch(target);
 			if (!id) return this.errorReply(`Channel with ID or name ${target} not found.`);
 			delete channelData[id];
+			FS(STORAGE_PATH).writeUpdate(() => JSON.stringify(channelData));
 			this.privateModAction(`(${user.name} deleted channel with ID or name ${target}.)`);
 			return this.modlog(`REMOVECHANNEL`, null, id);
 		},
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -240,22 +240,9 @@ export const commands: ChatCommands = {
 			return this.parse(`/j view-channels${all ? '-all' : ''}`);
 		},
 		help(target, room, user) {
-			const buf = (
-				`YouTube plugin commands:<br>` +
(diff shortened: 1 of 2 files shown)


server/chat-plugins/youtube.ts
```diff
@@ -0,0 +1,422 @@
...
+import * as https from 'https';
+import {FS} from '../../lib/fs';
+const ROOT = 'https://www.googleapis.com/youtube/v3/';
+const CHANNEL = `${ROOT}channels`;
+const STORAGE_PATH = 'config/chat-plugins/youtube.json';
+
+let channelData: AnyObject;
+
+try {
+	channelData = JSON.parse(FS(STORAGE_PATH).readSync()) || {};
```
KrisXV, line 17:
```suggestion
	channelData = JSON.parse(FS(STORAGE_PATH).readIfExistsSync() || "{}");
```

just to keep consistent with the other files

server/chat-plugins/youtube.ts
```diff
@@ -0,0 +1,422 @@
...
+} catch (e) {
+	channelData = {};
+}
+
+export const YouTube = new class {
+	interval: NodeJS.Timeout | null;
```
KrisXV, line 23: wouldnt this be better suited as `NodeJS.Timer`?

server/chat-plugins/youtube.ts
```diff
@@ -0,0 +1,422 @@
...
+
+let channelData: AnyObject;
+
+try {
+	channelData = JSON.parse(FS(STORAGE_PATH).readSync()) || {};
+} catch (e) {
+	channelData = {};
+}
+
+export const YouTube = new class {
```
KrisXV, line 22:
```suggestion
export class Youtube {
```

server/chat-plugins/youtube.ts
```diff
@@ -0,0 +1,422 @@
...
+		if (username) {
+			buf += `<p style="text-align:left;font-style:italic;color:white;">PS username: ${username}</p></td></tr></table></div>`;
+		} else {
+			buf += '</td></tr></table></div>';
+		}
+		return buf;
+	}
+	randChannel() {
+		const keys = Object.keys(channelData);
+		const id = keys[Math.floor(Math.random() * keys.length)].trim();
```
KrisXV, line 79:
```suggestion
		const id = Dex.shuffle(keys)[0].trim();
```

(this is optional, i just think that if we have a function, why not use it. though then it brings the question of why `sample` is on the `PRNG` class and not on `Dex` as well)

server/chat-plugins/youtube.ts
mia-pi-git: I don't think so?
VSCode says setInterval is `NodeJS.Timeout`.

mia-pi-git:
> I personally think the requestapproval/link function that is ported from Kid A should be in a separate commit.

I felt it fit here, since it uses `YoutubeInterface` functions, but if it's desired, I can move it out of this PR.

server/chat-plugins/youtube.ts
xfix: It doesn't matter. Checking `@types/node`'s source code, `Timeout` is a concrete class that implements `Timer` interface.

mia-pi-git: Alright.

mia-pi-git:
(requesting since they ran the plugin on the bot, so i thought input from them might be helpful, lmk if I shouldn't have done that)

mia-pi-git:
Removed approvals/link for now.

Zarel:
Looks good to me!

PR #6743, opened 2020-05-25
Trivia: remove Weakest Link

AnnikaCodes:
Weakest Link is no longer played in the Trivia room and the ROs agreed that it'd be ok to remove it.

### DIFF
--- a/server/chat-plugins/trivia.js
+++ b/server/chat-plugins/trivia.js
@@ -36,13 +36,6 @@ const MODES = {
 	triumvirate: 'Triumvirate',
 	tri: 'Triumvirate',
 	triforce: 'Triumvirate',
-	wl: "Weakest Link",
-	weakestlink: "Weakest Link",
-};
-
-const TYPES = {
-	trivia: 'Trivia',
-	weakestlink: 'Weakest Link',
 };
 
 const LENGTHS = {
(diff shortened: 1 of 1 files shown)


Zarel:
Is there a need to remove something just because no one uses it? Maybe someone wants to run their own Weakest Link game in a groupchat or side server? Is it a maintenance burden?

AnnikaCodes:
> Is there a need to remove something just because no one uses it? Maybe someone wants to run their own Weakest Link game in a groupchat or side server? Is it a maintenance burden?

If someone were to typescript trivia or change anything that Weakest Link uses then yes, it would be a maintenance burden. I suppose side servers could host it, that's a fair point — I don't know how much importance we place on the activities of side servers but if that's an issue then I can close this pull request. (Weakest Link and Trivia only work in the Trivia room, so it couldn't be hosted in a groupchat.)

Zarel:
It really depends on how hard it is to maintain, and why people don't play it anymore. Do most people consider it not very fun? How hard would it be to continue maintaining? You've worked on Trivia more than me, so I'll let it be your decision.

mia-pi-git:
Can confirm it's a burden - I'm working on typescripting trivia right now and this is an obnoxious barrier towards completion.

AnnikaCodes:
> It really depends on how hard it is to maintain, and why people don't play it anymore. Do most people consider it not very fun? How hard would it be to continue maintaining? You've worked on Trivia more than me, so I'll let it be your decision.

Weakest Link hasn't been hosted at all in the time I've been in Trivia; I believe this is because it was unbalanced as a game and thus not very fun to play - I've never played myself but apparently when it was hosted a lot of regs and auth were unhappy with it. Knowing that and what @mia-pi-git said about it being a maintenance issue, I would think that keeping Weakest Link is more work than it's worth.

Zarel:
Sounds good, thanks for explaining it to me!

xfix:
Can confirm Weakest Link sucks, and nobody cares in Trivia room cares about it.

PR #6744, opened 2020-05-25
Add a library for making HTTP/S requests 

### LINKED ISSUE #6691: Flag to disable all outgoing connections
Over the years I've privately reported this issue and met zero interest on it, so I'm just overtly opening this as an issue now.

User-initiated commands which cause the server to contact third-party sites (``/showimage``, ``*-otd``, ``/yugioh``, etc) will [nullify all the protection that CloudFlare and other reverse proxies normally grant to PS servers](https://community.cloudflare.com/t/outgoing-requests/25212), so there ought to be a simple way, enabled by default, to:

* Disable any and all of these commands.
* Disable outgoing connections - note that, for completeness, this applies to ``login-server.ts`` as well.

mia-pi-git:
Config setting name is tentative.
Loginserver error handling might be done a little bit better, so if anyone has a better idea let me know ig. 
It functions normally, however.

Fixes #6691

### DIFF
--- a/lib/net.ts
+++ b/lib/net.ts
@@ -40,6 +40,8 @@ export class NetRequest {
 		}
 
 		if (body && typeof body !== 'string') {
+			if (!opts.headers) opts.headers = {};
+			opts.headers['Content-Type'] = 'application/x-www-form-urlencoded';
 			let out = '';
 			for (const key in body) {
 				if (out) out += `&`;
@@ -51,7 +53,6 @@ export class NetRequest {
 		const postBody = body;
 		if (postBody) {
 			if (!opts.headers) opts.headers = {};
-			opts.headers['Content-Type'] = 'application/x-www-form-urlencoded';
 			opts.headers['Content-Length'] = postBody.length;
 		}
 
--- a/lib/net.ts
+++ b/lib/net.ts
@@ -9,23 +9,47 @@ import * as https from 'https';
 import * as http from 'http';
 import * as url from 'url';
 import * as Streams from './streams';
+import { resolve } from 'dns';
 
-export interface PostData {
-	[key: string]: string | number;
-}
 export interface NetRequestOptions extends https.RequestOptions {
-	body?: string | PostData;
-	query?: PostData;
+	body?: string;
+	query?: {[k: string]: string | number};
 }
 
-export class NetRequest {
-	uri: string;
+export class NetStream extends Streams.ReadWriteStream {
 	statusCode?: number;
-	constructor(uri: string) {
-		this.uri = uri;
-		this.statusCode = undefined;
+	headers?: http.IncomingHttpHeaders;
+	opts?: NetRequestOptions;
+	uri: string;
+	request: http.ClientRequest | null;
+	responseStream: Streams.ReadStream | null;
+	constructor(uri: string, opts?: NetRequestOptions | null) {
+		super();
+		this.statusCode = 0;
+		this.headers = {};
+		this.opts = opts ? opts : undefined;
+		this.uri = uri
+		this.responseStream = null;
+		// make request
+		this.request = null;
+		this.makeRequest();
 	}
-	encodeQuery(data: PostData) {
+	makeRequest() {
+		const protocol = url.parse(this.uri).protocol as string;
+		const net = protocol.includes('https:') ? https : http;
+		return new Promise(resolve => {
+			 this.request = net.get(this.opts ? this.opts : this.uri, response => {
+				  response.setEncoding('utf-8');
+				  const stream = new Streams.ReadStream({nodeStream: response});
+				  resolve({
+						stream: stream,
+						headers: response.headers,
+						statusCode: response.statusCode
+				  });
+			 });
+		});
+  }
+	encodeQuery(data: AnyObject) {
 		let out = '';
 		for (const key in data) {
 			if (out) out += `&`;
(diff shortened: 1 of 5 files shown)


lib/net.ts
```diff
@@ -0,0 +1,95 @@
...
+			const req = net.get(opts ? opts : this.uri, res => {
+				res.setEncoding('utf-8');
+				this.statusCode = res.statusCode;
+				void Streams.readAll(res).then(buffer => resolve(buffer));
+			});
+			req.on('error', (err) => {
+				throw err;
+			});
+			req.end();
+		});
```
Zarel, line 39: something something sync/async throws

```suggestion
		return new Promise((resolve, reject) => {
			const req = net.get(opts ? opts : this.uri, res => {
				res.setEncoding('utf-8');
				this.statusCode = res.statusCode;
				void Streams.readAll(res).then(buffer => resolve(buffer));
			});
			req.on('error', err => {
				reject(err);
			});
			req.end();
		});
```

lib/net.ts
```diff
@@ -0,0 +1,95 @@
...
+ * - easier acquiring of data
+ * - mass disabling of outgoing requests via Config.
+ */
+
+import * as https from 'https';
+import * as http from 'http';
+import * as url from 'url';
+import * as Streams from './streams';
+
+export class URIRequest {
```
Zarel, line 13: I would call this `NetRequest`.

lib/net.ts
```diff
@@ -0,0 +1,95 @@
...
+		this.uri = uri;
+		this.statusCode = undefined;
+	}
+	/**
+	 * Makes a basic http/https request to the URI.
+	 * Returns the response data.
+	 * @param opts request opts - headers, etc.
+	 */
+	async get(opts?: AnyObject): Promise<string | null> {
+		if (Config.noURIRequests) return null;
```
Zarel, line 26: I would also call this `noNetRequests`.

Maybe this should throw instead of returning `null`? Would make it easier to use in practice.

server/loginserver.ts
```diff
@@ -76,23 +74,25 @@ class LoginServerInstance {
...
-
-		return new Promise((resolve, reject) => {
-			const req = http.get(actionUrl, (res: IncomingMessage) => {
-				void Streams.readAll(res).then((buffer: string) => {
-					const result = parseJSON(buffer).json || null;
-					resolve([result, res.statusCode || 0, null]);
-					this.openRequests--;
+			'&nocache=' + new Date().getTime() + dataString).href;
+
+		return new Promise((resolve) => {
```
Zarel, line 79: Do you think you can refactor this into async/await for me?

mia-pi-git: I can try, sure.

lib/net.ts
```diff
@@ -0,0 +1,96 @@
...
+			const req = net.get(opts ? opts : this.uri, res => {
+				res.setEncoding('utf-8');
+				this.statusCode = res.statusCode;
+				void Streams.readAll(res).then(buffer => resolve(buffer));
+			});
+			req.on('error', err => {
+				reject(err);
+			});
+			req.end();
+		});
```
Zarel, line 39: Is this necessary? It seems like it's probably easiest just to defer to `request`, right?

Heck, I would make it all defer to `getFullResponse`.

lib/net.ts
```diff
@@ -0,0 +1,96 @@
...
+	 */
+	async request(opts?: AnyObject, chunk?: string, timeout?: number): Promise<string> {
+		if (Config.noNetRequests) throw new Error(`Net requests are disabled.`);
+		return new Promise(resolve => {
+			const protocol = url.parse(this.uri).protocol as string;
+			const net = protocol.includes('https:') ? https : http;
+			const req = net.request(opts ? opts : this.uri, res => {
+				res.setEncoding('utf-8');
+				this.statusCode = res.statusCode;
+				void Streams.readAll(res).then(buffer => resolve(buffer));
```
urkerab, line 83: `resolve` accepts a `Promise`, so this can just be
```suggestion
			resolve(Streams.readAll(res));
```
(This may apply in other places too.)

In fact, this will automatically pass any rejection from `readAll` to the caller of `request`, which your code fails to do. (Always assuming that it's possible for `readAll` to reject, of course.)

lib/net.ts
```diff
@@ -0,0 +1,96 @@
...
+			});
+			req.on('err', err => {
+				throw err;
+			});
```
urkerab, line 87: This still needs to be fixed to use `reject`. (This may apply ion other places too.)

(I notice that server/chat-plugins/tcgtabletop.ts used to use the style `req.on('err', reject);` rather than explicitly calling `reject`).

lib/net.ts
```diff
@@ -0,0 +1,96 @@
...
+	}
+	/**
+	 * Makes a basic http/https request to the URI.
+	 * Returns the response data.
+	 * @param opts request opts - headers, etc.
+	 */
+	async get(opts?: AnyObject): Promise<string | null> {
+		if (Config.noNetRequests) throw new Error(`Net requests are disabled.`);
+		const protocol = url.parse(this.uri).protocol as string;
+		const net = protocol.includes('https:') ? https : http;
```
Kaiepi, line 28: This should probably be `protocol.startsWith('https://')`

server/chat-plugins/tcgtabletop.ts
```diff
@@ -6,33 +6,15 @@
...
-		https.request(reqOpts, res => {
-			if (!res.statusCode || !(res.statusCode >= 200 && res.statusCode < 300)) return reject(new Error(`Not found.`));
-			const data: string[] = [];
-			res.setEncoding('utf8');
-			res.on('data', chunk => data.push(chunk));
-			res.on('end', () => resolve(data.join('')));
-		}).on('error', reject).setTimeout(5000).end();
-	});
-
+	const body = await Net(`https://${site}.fandom.com/${pathName}?${querystring.stringify(search)}`).get();
```
Zarel, line 17: this is missing a `Content-Type` `application/json` header, which might be relevant.

Zarel: Looks like it's not. @Asheviere, do you know why you originally sent it? `Content-Type` is a response header, not a request header; the corresponding request header is `Accept` to list which content types you support.

lib/net.ts
```diff
@@ -10,49 +10,57 @@ import * as http from 'http';
...
+				out += `${key}=${encodeURIComponent('' + body[key])}`;
+			}
+			body = out;
+		}
+
+		const postBody = body;
+		if (postBody) {
+			if (!opts.headers) opts.headers = {};
+			opts.headers['Content-Type'] = 'application/x-www-form-urlencoded';
+			opts.headers['Content-Length'] = postBody.length;
```
Kaiepi, line 56: This will not handle Unicode properly:

```
bastille% node -e 'console.log("\u{1F914}".length)'
2
```

There are 4 bytes in this string, which is what the `Content-Length` header should include as a length if this were the request's body. You can get the proper length with `Buffer.byteLength(postBody)`.

server/loginserver.ts
```diff
@@ -73,27 +71,26 @@ class LoginServerInstance {
 			}
 		}
 
-		const actionUrl = url.parse(this.uri + 'action.php' +
+		const actionUrl = url.parse(
```
Zarel, line 74: `url.parse` shouldn't be necessary, actually. It's only used to parse the URI for use by `http.get`/`https.get`, which is already done inside `Net`.

lib/net.ts
Zarel: Thanks!

Zarel: I actually did think of that problem right after pushing this, but I didn't know `Buffer.byteLength` existed.

lib/net.ts
```diff
@@ -0,0 +1,145 @@
...
+	 * Makes a http/https get request to the given link and returns a stream.
+	 * The request data itself can be read with ReadStream#read().
+	 * The NetStream class also holds headers and statusCode as a property.
+	 * @param opts request opts - headers, etc.
+	 * @param body POST body
+	 */
+	getStream(opts: NetRequestOptions = {}) {
+		if (Config.noNetRequests) throw new Error(`Net requests are disabled.`);
+		const stream = new NetStream(this.uri);
+		this.statusCode = stream.statusCode;
```
Zarel, line 116: This is pointless because `stream.statusCode` is `undefined` here. It won't be defined until after we get a response from the server.

lib/net.ts
```diff
@@ -0,0 +1,145 @@
...
+		this.request = null;
+		this.makeRequest();
+	}
+	makeRequest() {
+		const protocol = url.parse(this.uri).protocol as string;
+		const net = protocol.includes('https:') ? https : http;
+		return new Promise(resolve => {
+			 this.request = net.get(this.opts ? this.opts : this.uri, response => {
+				  response.setEncoding('utf-8');
+				  const stream = new Streams.ReadStream({nodeStream: response});
```
Zarel, line 43: My idea was to make the new stream _be_ the stream, not _contain_ the stream.

mia-pi-git: Alright, i'll rewrite it.

lib/net.ts
```diff
@@ -9,107 +9,129 @@ import * as https from 'https';
...
-			const net = protocol.includes('https:') ? https : http;
-			const req = net.request(opts ? opts : this.uri, response => {
-				const stream = new Streams.ReadStream({nodeStream: response});
-				resolve({
-					statusCode: response.statusCode,
-					statusMessage: response.statusMessage,
-					headers: response.headers,
-					stream: stream,
-				});
-			});
```
Zarel: You deleted this old good code. :(

Your new code can't handle options or POST requests.

lib/net.ts
```diff
@@ -9,107 +9,129 @@ import * as https from 'https';
...
 
-export interface PostData {
-	[key: string]: string | number;
-}
 export interface NetRequestOptions extends https.RequestOptions {
-	body?: string | PostData;
-	query?: PostData;
+	body?: string;
+	query?: {[k: string]: string | number};
 }
```
Zarel: What's this about?

PR #6747, opened 2020-05-26
Fix Battle Substitutions

PartMan7:
Right now, it just doesn't detect the slot, and then doesn't sub properly. (The old code, I mean.)

**NOTE:** 
I've tested these changes on a sideserver, where they worked during normal challenge battles. I have not tested on rated battles or tournament battles, but those shouldn't be affected. 
There have been no crashes on renames or multi-substituting until the original players are switched; but I'm not completely sure if that's all that this was meant to handle - are there any cases I'm missing?

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -1048,7 +1048,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 			};
 			void this.stream.write(`>player ${slot} ` + JSON.stringify(options));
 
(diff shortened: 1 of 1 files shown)


monsanto:
Failed check discussed on discord

Zarel:
Thanks for helping me figure out this bug! I didn't end up going with your solution, but it helped me figure out the correct solution.

PR #6672, opened 2020-05-10
Support renaming events

TheMezStrikes:
as mentioned in approved suggestion https://www.smogon.com/forums/threads/support-renaming-events-in-roomevents.3663574/#post-8453567

PR #6748, opened 2020-05-27
Room events: Support aliases and renaming

AnnikaCodes:
As per [this suggestion](https://www.smogon.com/forums/threads/support-renaming-events-in-roomevents.3663574/).

### DIFF
--- a/server/chat-plugins/room-events.ts
+++ b/server/chat-plugins/room-events.ts
@@ -136,7 +136,7 @@ export const commands: ChatCommands = {
 			if (!target) return this.errorReply("Usage: /roomevents start [event name]");
 			target = toID(target);
 			const event = room.events[getEventID(target, room)];
-			if (!event) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
+			if (!event) return this.errorReply(`There is no event titled '${target}'. Check spelling?`);
 			if (event.started) {
 				return this.errorReply(`The event ${event.eventName} has already started.`);
 			}
@@ -169,7 +169,7 @@ export const commands: ChatCommands = {
 			if (!target) return this.errorReply("Usage: /roomevents remove [event name]");
 			target = toID(target);
 			if (getAllAliases(room).includes(target)) return this.errorReply("To delete aliases, use /roomevents removealias.");
-			if (!room.events[target]) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
+			if (!room.events[target]) return this.errorReply(`There is no event titled '${target}'. Check spelling?`);
 			delete room.events[target];
 			this.privateModAction(`(${user.name} removed a roomevent titled "${target}".)`);
 			this.modlog('ROOMEVENT', null, `removed "${target}"`);
@@ -187,7 +187,7 @@ export const commands: ChatCommands = {
 			target = toID(target);
 
 			const event = room.events[getEventID(target, room)];
-			if (!event) return this.errorReply(`There is no such event named '${target}'. Check spelling?`);
+			if (!event) return this.errorReply(`There is no event titled '${target}'. Check spelling?`);
 			if (!this.runBroadcast()) return;
 			const buff = `<table border="1" cellspacing="0" cellpadding="3">${formatEvent(event)}</table>`;
 			this.sendReply(`|raw|<div class="infobox-limited">${buff}</div>`);
--- a/server/chat-plugins/room-events.ts
+++ b/server/chat-plugins/room-events.ts
@@ -7,7 +7,7 @@
  * @license MIT license
  */
 
-function formatEvent(event: {eventName: string, date: string, desc: string, started: boolean}) {
+function formatEvent(event: {eventName: string, date: string, desc: string, started: boolean, aliases: string[]}) {
 	const timeRemaining = new Date(event.date).getTime() - new Date().getTime();
 	let explanation = timeRemaining.toString();
 	if (!timeRemaining) explanation = "The time remaining for this event is not available";
@@ -18,9 +18,26 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
 	}
 	let ret = `<tr title="${explanation}">`;
 	ret += Chat.html`<td>${event.eventName}</td>`;
+	ret += Chat.html`<td>${(event.aliases ? event.aliases.join(", ") : "")}</td>`;
 	ret += `<td>${Chat.formatText(event.desc, true)}</td>`;
 	ret += Chat.html`<td><time>${event.date}</time></td></tr>`;
 	return ret;
+}
+
+function getAllAliases(room: Room) {
+	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+}
+
+function getEventIDByNameOrAlias(nameOrAlias: string, room: Room): ID {
+	let id = toID(nameOrAlias);
+	if (!room.events[id]) {
+		for (const possibleEvent in room.events) {
+			if (room.events[possibleEvent].aliases.includes(id)) {
+				id = toID(possibleEvent);
+			}
+		}
+	}
+	return id;
 }
 
 export const commands: ChatCommands = {
@@ -34,7 +51,7 @@ export const commands: ChatCommands = {
 			}
 			if (!this.runBroadcast()) return;
 			let buff = '<table border="1" cellspacing="0" cellpadding="3">';
-			buff += '<th>Event Name:</th><th>Event Description:</th><th>Event Date:</th>';
+			buff += '<th>Event Name:</th><th>Event Aliases:</th><th>Event Description:</th><th>Event Date:</th>';
 			for (const i in room.events) {
 				buff += formatEvent(room.events[i]);
 			}
@@ -55,7 +72,7 @@ export const commands: ChatCommands = {
 				return this.errorReply("You're missing a command parameter - see /help roomevents for this command's syntax.");
 			}
 
-			const eventNameActual = eventName.trim();
+			let eventNameActual = eventName.trim();
 			const dateActual = date.trim();
 			const descString = desc.join(target.includes('|') ? '|' : ',').trim();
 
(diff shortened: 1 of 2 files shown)


KrisXV:
This was actually partially already done (renaming, not aliases) in #6672, but was never reviewed by Asheviere due to other happenings. I will say that I like the look of this PR better because it has the aliasing functionality as well, though.

AnnikaCodes:
> This was actually partially already done (renaming, not aliases) in #6672, but was never reviewed by Asheviere due to other happenings. I will say that I like the look of this PR better because it has the aliasing functionality as well, though.

oh I wasn't aware that was being worked on, my bad. If you want I can use Mez' code instead for renaming.

Zarel:
I don't think we have anyone available to code review old chat plugins. Do any of you want to take on that job?

KrisXV:
Sure @Zarel

mia-pi-git:
Would also be willing and able to help out with review, if it's desired.

server/chat-plugins/room-events.ts
```diff
@@ -18,11 +18,28 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
...
 	ret += `<td>${Chat.formatText(event.desc, true)}</td>`;
 	ret += Chat.html`<td><time>${event.date}</time></td></tr>`;
 	return ret;
 }
 
+function getAllAliases(room: Room) {
+	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+}
+
+function getEventIDByNameOrAlias(nameOrAlias: string, room: Room): ID {
```
KrisXV, line 31: I think this can just be called `getEventID`, its obviously grabbing the ID via alias or name with the name of the first argument and how it's being used throughout the code.

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
...
 			if (eventNameActual.length > 50) return this.errorReply("Event names should not exceed 50 characters.");
 			if (dateActual.length > 150) return this.errorReply("Event dates should not exceed 150 characters.");
 			if (descString.length > 1000) return this.errorReply("Event descriptions should not exceed 1000 characters.");
 
-			const eventId = toID(eventName);
+			const eventId = getEventIDByNameOrAlias(eventNameActual, room);
 			if (!eventId) return this.errorReply("Event names must contain at least one alphanumerical character.");
 
-			this.privateModAction(`(${user.name} ${room.events[eventId] ? "edited the" : "added a"} roomevent titled "${eventNameActual}".)`);
+			eventNameActual = (room.events[eventId] ? room.events[eventId].eventName : eventNameActual);
```
KrisXV, line 86:
```suggestion
			const eventNameActual = (room.events[eventId] ? room.events[eventId].eventName : eventNameActual);
```

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
 				return this.errorReply("You're missing a command parameter - see /help roomevents for this command's syntax.");
 			}
 
-			const eventNameActual = eventName.trim();
+			let eventNameActual = eventName.trim();
 			const dateActual = date.trim();
 			const descString = desc.join(target.includes('|') ? '|' : ',').trim();
 
 			if (eventNameActual.length > 50) return this.errorReply("Event names should not exceed 50 characters.");
```
KrisXV, line 79:
```suggestion
			if (eventName.trim().length > 50) return this.errorReply("Event names should not exceed 50 characters.");
```

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
 				return this.errorReply("You're missing a command parameter - see /help roomevents for this command's syntax.");
 			}
 
-			const eventNameActual = eventName.trim();
+			let eventNameActual = eventName.trim();
```
KrisXV, line 75: I would just remove this line, you can see what i did in the other suggestions I made

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
...
 			if (dateActual.length > 150) return this.errorReply("Event dates should not exceed 150 characters.");
 			if (descString.length > 1000) return this.errorReply("Event descriptions should not exceed 1000 characters.");
 
-			const eventId = toID(eventName);
+			const eventId = getEventIDByNameOrAlias(eventNameActual, room);
 			if (!eventId) return this.errorReply("Event names must contain at least one alphanumerical character.");
 
-			this.privateModAction(`(${user.name} ${room.events[eventId] ? "edited the" : "added a"} roomevent titled "${eventNameActual}".)`);
+			eventNameActual = (room.events[eventId] ? room.events[eventId].eventName : eventNameActual);
+			this.privateModAction(`(${user.name} ${room.events[eventId] ? "edited the" : "added a"} roomevent "${eventNameActual}".)`);
```
KrisXV, line 87:
```suggestion
			this.privateModAction(`(${user.name} ${room.events[eventId] ? "edited the" : "added a"} roomevent titled "${eventNameActual}".)`);
```
It seems like you accidentally removed this?

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
...
+			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+			if (!this.can('ban', null, room)) return false;
+			let [oldName, newName] = target.split(target.includes('|') ? '|' : ',');
+			if (!(oldName && newName)) return this.errorReply("Usage: /roomevents rename [old name], [new name]");
+
+			newName = newName.trim();
+			const newID = toID(newName);
+			const oldID = (getAllAliases(room).includes(toID(oldName)) ? getEventIDByNameOrAlias(oldName, room) : toID(oldName));
+			if (newID === oldID) return this.errorReply("The new name must be different from the old one.");
+			if (!newID) return this.errorReply("Event names must contain at least one alphanumerical character.");
```
KrisXV, line 111:
```suggestion
			if (!newID) return this.errorReply("Event names must contain at least one alphanumeric character.");
```

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
...
+			if (newName.length > 50) return this.errorReply("Event names should not exceed 50 characters.");
+
+			const eventData = room.events[oldID];
+			if (!eventData) return this.errorReply(`There is no event named '${oldName}'.`);
```
KrisXV, line 115:
```suggestion
			if (!eventData) return this.errorReply(`There is no event titled "${oldName}".`);
```

Just to keep consistent with the quotation usage elsewhere

server/chat-plugins/room-events.ts
```diff
@@ -55,30 +72,61 @@ export const commands: ChatCommands = {
...
+			if (room.events[newID] || getAllAliases(room).includes(newID)) {
+				return this.errorReply(`${newName} is already an event or alias.`);
```
KrisXV, line 117:
```suggestion
				return this.errorReply(`"${newName}" is already an event or alias.`);
```

server/chat-plugins/room-events.ts
```diff
@@ -88,25 +136,26 @@ export const commands: ChatCommands = {
...
 					);
 				}
 			}
 			this.add(
-				Chat.html`|raw|<div class="broadcast-blue"><b>The "${room.events[target].eventName}" roomevent has started!</b></div>`
+				Chat.html`|raw|<div class="broadcast-blue"><b>The "${event.eventName}" roomevent has started!</b></div>`
 			);
-			this.modlog('ROOMEVENT', null, `started "${target}"`);
-			room.events[target].started = true;
+			this.modlog('ROOMEVENT', null, `started "${toID(event.eventName)}"`);
```
KrisXV, line 157: I don't think this needs to be changed to an ID, can I ask your reasoning? I might be missing something.

server/chat-plugins/room-events.ts
```diff
@@ -120,6 +169,7 @@ export const commands: ChatCommands = {
 			}
 			if (!target) return this.errorReply("Usage: /roomevents remove [event name]");
 			target = toID(target);
+			if (getAllAliases(room).includes(target)) return this.errorReply("Use /roomevents removealias to delete aliases.");
```
KrisXV, line 172:
```suggestion
			if (getAllAliases(room).includes(target)) return this.errorReply("To delete aliases, use /roomevents removealias.");
```
JUST in case someone actually thinks you need to type `/roomevents removealias to delete aliases`. This is somewhat of a nitpick though

server/chat-plugins/room-events.ts
```diff
@@ -136,18 +186,57 @@ export const commands: ChatCommands = {
...
 		},
+		alias: 'addalias',
+		addalias(target, room, user) {
+			if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+			if (!this.can('ban', null, room)) return false;
+			const [alias, eventId] = target.split(target.includes('|') ? '|' : ',').map(argument => toID(argument));
+			if (!(alias && eventId)) {
+				return this.errorReply("Usage: /roomevents addalias [alias], [event name]. Aliases must contain at least one alphanumeric character.");
+			}
+			if (!room.events?.[eventId]) return this.errorReply(`There is no event named "${eventId}".`);
```
KrisXV, line 210: I would make a separate error message if the room has no events listed; something like `There are currently no scheduled events.`

KrisXV: Also I would change "named" to "titled" to keep consistent

server/chat-plugins/room-events.ts
```diff
@@ -18,11 +18,28 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
 	}
 	let ret = `<tr title="${explanation}">`;
 	ret += Chat.html`<td>${event.eventName}</td>`;
+	ret += Chat.html`<td>${(event.aliases ? event.aliases.join(", ") : "")}</td>`;
```
KrisXV, line 21:
```suggestion
	ret += Chat.html`<td>${event.aliases ? event.aliases.join(", ") : ""}</td>`;
```

Missed this; the parentheses are unnecessary with the brackets.

server/chat-plugins/room-events.ts
AnnikaCodes: the only reason I made it an ID is because the original code used an ID there

KrisXV: oh ok

server/chat-plugins/room-events.ts
AnnikaCodes: Pretty sure the code I added in b1b7533 does that

AnnikaCodes: turns out I didn't fix the named vs titled thing, thats done now

server/chat-plugins/room-events.ts
```diff
@@ -18,11 +18,28 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
...
 	ret += Chat.html`<td>${event.eventName}</td>`;
+	ret += Chat.html`<td>${event.aliases ? event.aliases.join(", ") : ""}</td>`;
 	ret += `<td>${Chat.formatText(event.desc, true)}</td>`;
 	ret += Chat.html`<td><time>${event.date}</time></td></tr>`;
 	return ret;
 }
 
+function getAllAliases(room: Room) {
+	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+}
```
Zarel, line 29: This... returns a string.

That's nearly definitely not what you wanted.

For one thing, if your only alias is "asdfjkl", `getAllAliases(...).includes("asdf")` is `true`, and `getAllAliases(...).includes("undefined")` is `true`, and your code doesn't actually seem to assume they wouldn't be.

AnnikaCodes: Good catch! I'm not sure how that happened but it's definitely on me. Anyway, #6805 should fix the issue.

PR #6750, opened 2020-05-27
Chatlog: fix memory crashes, improve UI

mia-pi-git:
Fixes a bug with the html where highlighted lines would show up despite being null. 
Fixes a bug where spaces would break date/cap.


server/chat-plugins/chatlog.ts
```diff
@@ -640,7 +646,7 @@ export const commands: ChatCommands = {
 		if (!target) return this.parse('/help searchlogs');
 		if (!search) return this.errorReply('Specify a query to search the logs for.');
 		let limitString;
-		if (/^[0-9]+$/.test(limit)) {
+		if (/^[ 0-9]+$/.test(limit)) {
```
Zarel, line 649: Room IDs can't contain spaces; you should probably trim `limit` in its definition if this is an issue.

server/chat-plugins/chatlog.ts
```diff
@@ -660,6 +665,8 @@ export const commands: ChatCommands = {
 		"/searchlogs [search], [room], [cap], [date] - searches logs in the current room for [search].",
 		"A comma can be used to search for multiple words in a single line - in the format arg1, arg2, etc.",
 		"If a [cap] is given, limits it to only that many lines. Defaults to 500.",
+		"The delimiter | can be used to space searching for multiple terms.",
+		"Date formatting is ISO formatting (YYYY-MMMM-DDDD.) E.g 2020-05, 2020, or `all`.",
```
KrisXV, line 669:
```suggestion
		"Date formatting is ISO formatting (YYYY-MM-DD.) E.g 2020-05, 2020, or `all`.",
```

Months and days only have 2 digits :P

server/chat-plugins/chatlog.ts
```diff
@@ -180,20 +180,25 @@ export const LogViewer = new class {
...
-		const renderResult = (match: SearchMatch) => (
-			this.renderLine(match[0]) +
-			this.renderLine(match[1]) +
-			`<div class="chat chatmessage highlighted">${this.renderLine(match[2])}</div>` +
-			this.renderLine(match[3]) +
-			this.renderLine(match[4])
-		);
+	renderDayResults(results: {[day: string]: SearchMatch[]}, roomid: RoomID) {
+		const renderResult = (match: SearchMatch) => {
+			if (!match[2]) return null;
```
Zarel, line 185: This should probably be rejected before being considered a `SearchMatch` in the first place.

server/chat-plugins/chatlog.ts
```diff
@@ -208,13 +221,15 @@ export const LogViewer = new class {
...
+		buf += this.renderDayResults(results, roomid);
+		if (limit && total > limit || total > MAX_RESULTS) {
 			// cap is met & is not being used in a year read
-			buf += `<br><strong>Max results reached, capped at ${limit}</strong>`;
+			buf += `<br><strong>Max results reached, capped at ${limit && total > limit ? limit : MAX_RESULTS}</strong>`;
 			buf += `<br><div style="text-align:center">`;
-			buf += `<button class="button" name="send" value="/sl ${search}|${roomid}|${month}|${limit + 100}">View 100 more<br />&#x25bc;</button>`;
-			buf += `<button class="button" name="send" value="/sl ${search}|${roomid}|${month}|all">View all<br />&#x25bc;</button></div>`;
+			if (total < MAX_RESULTS) {
+				buf += `<button class="button" name="send" value="/sl ${search}|${roomid}|${month}|${limit! + 100}">View 100 more<br />&#x25bc;</button>`;
```
HoeenCoder, line 230: Whats with `limit!`?

server/chat-plugins/chatlog.ts
```diff
@@ -492,13 +511,19 @@ const LogSearcher = new class {
 				'-e', `[^a-zA-Z0-9]${search.split('').join('[^a-zA-Z0-9]*')}([^a-zA-Z0-9]|\\z)`,
 				`${__dirname}/../../logs/chat/${roomid}`,
 				'-C', '3',
+				'-m', `${MAX_RESULTS}`,
```
HoeenCoder, line 514: You could use limit here too.

server/chat-plugins/chatlog.ts
```diff
@@ -492,13 +511,19 @@ const LogSearcher = new class {
...
 			];
 			output = await execFile('rg', options, {maxBuffer: Infinity, cwd: path.normalize(`${__dirname}/../`)});
 		} catch (error) {
 			if (error.message.includes('Command failed')) return LogViewer.error(`No results found.`);
 			return LogViewer.error(`${error.message}`);
 		}
-		return this.render(output.stdout.split('--'), roomid, search, limit);
+		return this.render(
+			output.stdout.split('--').reverse().slice(0, MAX_RESULTS),
```
HoeenCoder, line 522: Do you need to slice to `MAX_RESULTS` if your already limiting it in your ripgrep query?

HoeenCoder:
Very much appreciate your work on this!

server/chat-plugins/chatlog.ts
mia-pi-git: Nope. forgot to remove.

monsanto:
For future reference, can you not combine bug fixes and refactors/UI improvements into the same PR? HH asked me to quickly review this code before it was deployed, but I can't easily see what changes are relevant to fixing the OOM errors.

mia-pi-git:
> For future reference, can you not combine bug fixes and refactors/UI improvements into the same PR? HH asked me to quickly review this code before it was deployed, but I can't easily see what changes are relevant to fixing the OOM errors.

Yeah, sorry. This was originally just one minor bug and some UI stuff, but it was already open. 
The relevant changes are to `MAX_RESULTS`, `LogReader#read()`, and the `limit` changes, if it helps.

server/chat-plugins/chatlog.ts
```diff
@@ -109,12 +110,25 @@ const LogReader = new class {
...
+		const stream = await roomLog!.getLog(day);
+		let buf = '';
+		let i = LogViewer.results || 0;
+		if (!stream) {
+			buf += `<p class="message-error">Room "${roomid}" doesn't have logs for ${day}</p>`;
+		} else {
+			let line;
+			while ((line = await stream.readLine()) !== null) {
+				const rendered = LogViewer.renderLine(line);
+				if (i++ > limit) return buf;
```
HoeenCoder, line 124: You need to use `++i` here or else it increments after comparing. Also why are you incrementing here and on line 127?

A wiser order might be render -> add to buf if rendered & increment i ->if i >= limit break -> next loop

HoeenCoder:
Ran some before and after tests with this locally using 3 months of lobby logs, server froze and OOMed before but did not after, so thats great!

server/chat-plugins/chatlog.ts
```diff
@@ -636,8 +678,9 @@ export const commands: ChatCommands = {
 	searchlog: 'searchlogs',
 	searchlogs(target, room) {
 		target = target.trim();
-		const [search, tarRoom, limit, date] = target.split(',');
+		const [search, tarRoom, limit, date] = target.split(',').map(str => str.trim());
 		if (!target) return this.parse('/help searchlogs');
+		if (search.length < 3) return this.errorReply(`Too short of a search query.`);
```
Zarel, line 683: This should fall back to showing command help, incidentally.

Zarel: Limit should also be optional; why is it suddenly required?

Zarel: room should also be optional; it should default to the current room...

mia-pi-git: limit is required since it will always be passed, since there's now a max cap of 3000 to prevent OOM

Zarel: Oh, this is the length of the search text! My bad.

Zarel: This should definitely specify what the minimum is. This is a very confusing error message.

Zarel: I don't think this is necessary, anyway. Short search queries should probably be exact-word matches, but there's nothing wrong with that. As long as the limit exists, this doesn't need to also exist.

server/chat-plugins/chatlog.ts
```diff
@@ -660,6 +703,8 @@ export const commands: ChatCommands = {
 		"/searchlogs [search], [room], [cap], [date] - searches logs in the current room for [search].",
 		"A comma can be used to search for multiple words in a single line - in the format arg1, arg2, etc.",
 		"If a [cap] is given, limits it to only that many lines. Defaults to 500.",
+		"The delimiter | can be used to space searching for multiple terms.",
```
Zarel, line 706: Weird to use `|` when you're searching for all the terms, not any of the terms.

Zarel: Ideally, it should support `,`.

mia-pi-git: I decided on | because `,` is more commonly used for different params, so i decided it'd make more sense to have `/sl [queries], [room]...etc` as opposed to some complicated way to sort through the `,`s to find the data.

mia-pi-git: I'll work on improving the system for this, though.

server/chat-plugins/chatlog.ts
mia-pi-git: I benchmarked FS search with this limit several times. It strains it a lot - in the zone of like, 20 second searches, which worries me when you consider that likely people will be like "it's not working? better try again." Additionally, I don't see a benefit to being able to search one or two letters. It would just lag, and you'd get a _lot_ of false positives. 
It seems like, since this is already here, why not keep it and keep the extra lag protection?

Zarel: FS search should definitely show a loading screen if it takes that long; as should everything else that takes that long.

As I mentioned on Discord:

I think there are certain letter combinations like `XD` where you won't get many false positives; it seems way worse to refuse to give good results than to sometimes not give bad results.

I'm also confused why it would strain over this search but not other searches. The default limit is 500, right? It should be _faster_ than a search with 0 results, because it wouldn't need to search anywhere near as much of the history.

server/chat-plugins/chatlog.ts
Zarel: Yeah, I know, we're already using `,` to separate parameters.

I don't really like `|` to separate parameters but we already have other commands that use it that way. :/ I think I'd much rather do it the other way around (using `|` to give room IDs and search limits and date ranges), because searching for multiple things is probably generally more useful.

KrisXV: (fwiw, I suggested using `+` to separate the first argument, since it more easily communicates a combined search)

mia-pi-git: I'll go with that.

Zarel: I believe dexsearch uses `,`? It might be better to be consistent with that.

I just don't think searching for rooms (rather than just `/join`ing the room) and having custom limits/dates is worth taking up the usual `,` symbol.

Zarel: I've reflected, and I think I like using `limit:` etc as prefixes.

Something like:

`/sl hello, zarel, limit: 30, date: 2020-03`

That would also help you not need to memorize the argument order.

mia-pi-git: Will do.

PR #6752, opened 2020-05-27
Add Emergency Exit tests for new bugged behavior

DaWoblefet:
fart's change in https://github.com/smogon/pokemon-showdown/pull/6738 unfortunately broke Emergency Exit in another way; it's requesting switchout if its HP is already below 50%, see bugged behavior turn 17: https://replay.pokemonshowdown.com/gen8randombattle-1122012552

Adds skipped test (what's failing) and passing test for if Emergency Exit heals HP above 50% and gets brought back down again (currently passing)


HoeenCoder:
Thanks!

PR #6756, opened 2020-05-28
Add more move target types to /details

AnnikaCodes:
This is a pretty minor issue but I stumbled upon it while adding searching by target to `/movesearch` so I figured I'd just open a PR fixing it.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -738,6 +738,9 @@ export const commands: ChatCommands = {
 						allAdjacent: "All Adjacent Pok\u00e9mon",
 						any: "Any Pok\u00e9mon",
 						all: "All Pok\u00e9mon",
+						scripted: "Opposing Pok\u00e9mon that last dealt damage to the user",
+						randomNormal: "Random Opposing Pok\u00e9mon",
+						allies: "User and Allies",
 					};
 					details["Target"] = targetTypes[move.target] || "Unknown";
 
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.ts
```diff
@@ -738,6 +738,9 @@ export const commands: ChatCommands = {
 						allAdjacent: "All Adjacent Pok\u00e9mon",
 						any: "Any Pok\u00e9mon",
 						all: "All Pok\u00e9mon",
+						scripted: "Opposing Pok\u00e9mon that last dealt damage to the user",
```
Zarel, line 741: I'm not sure if we want to use this phrasing? I chose the name "scripted" because my plan was to use it for any target that the game chose for you.

server/chat-commands/info.ts
```diff
@@ -738,6 +738,9 @@ export const commands: ChatCommands = {
...
+						randomNormal: "Random Opposing Pok\u00e9mon",
```
Zarel, line 742:
```suggestion
						randomNormal: "Random Adjacent Opposing Pok\u00e9mon",
```

server/chat-commands/info.ts
AnnikaCodes: so would something like "Chosen Automatically" be better?

Zarel: Yeah, I like that.

AnnikaCodes: Done.

Zarel:
I wonder if "One Ally" and "Ally or User" should be edited to add "adjacent".

We might also consider doing a `.replace("Adjacent ", "")` in generations without Triples.

Zarel:
Those changes can be in a separate PR, though. This one looks good, thanks!

PR #6758, opened 2020-05-28
Fix and enhance /editbattle

mathfreak231:
- Allow whitespace around the party slot number
- Fix finding a Pokemon by species name
- Fix adjusting PP
- Make the help message clearer in that a number for a pokemon represents the position in the party

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -1082,6 +1082,6 @@ export const commands: ChatCommands = {
 		`/editbattle weather [weather]`,
 		`/editbattle terrain [terrain]`,
 		`Short forms: /ebat h OR s OR pp OR b OR v OR sc OR fc OR w OR t`,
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -1015,10 +1015,10 @@ export const commands: ChatCommands = {
 			return 'p3';
 		}
 		function getPokemon(input: string) {
-			if (/^[0-9]+$/.test(input)) {
+			if (!isNaN(parseInt(input))) {
```
Zarel, line 1018:
```suggestion
			if (/^[0-9]+$/.test(input.trim())) {
```

I'd allow whitespace this way, because the way you have it implemented, it will match anything starting with a digit `0-9`, rather than just numbers.

Zarel:
This is great, thanks!

server/chat-commands/admin.ts
Zarel: (That's why I had the regex in the first place.)

PR #6759, opened 2020-05-28
Datasearch: add /abilitysearch command

AnnikaCodes:
As per [this suggestion](https://www.smogon.com/forums/threads/ability-search-command.3665079/). I didn't implement Pokémon searching because, as said in the thread, one can just `/dt` the Pokémon to see its abilities.


Zarel:
Seems unlikely to break, so sure

PR #6760, opened 2020-05-29
Typescript config/

KrisXV:


config/formats.ts
```diff
@@ -957,18 +949,18 @@ const Formats = [
...
 		onSwitchInPriority: 100,
 		onSwitchIn(pokemon) {
-			let name = toID(pokemon.illusion ? pokemon.illusion.name : pokemon.name);
+			let name: string = toID(pokemon.illusion ? pokemon.illusion.name : pokemon.name);
 			if (this.dex.getSpecies(name).exists || name === 'rage') {
 				// Certain pokemon have volatiles named after their id
 				// To prevent overwriting those, and to prevent accidentaly leaking
 				// that a pokemon is on a team through the onStart even triggering
 				// at the start of a match, users with pokemon names will need their
 				// statuse's to end in "user".
```
Zarel, line 958:
```suggestion
				// statuses to end in "user".
```

> statuse's

config/formats.ts
```diff
@@ -199,7 +196,7 @@ const Formats = [
 		],
 
 		mod: 'gen8',
-		ruleset: ['[Gen 8] OU', '-Melmetal', '+CAP'],
+		ruleset: ['[Gen 8] OU', '+CAP'],
```
Zarel, line 199: This seems like a weirdly consequential change to do in a TS refactor. Can it (and similar changes) be pulled out into a separate PR?

config/formats.ts
```diff
@@ -565,8 +561,10 @@ const Formats = [
 				if (restrictedPokemon.includes(species.name)) {
 					return [`${species.name} is not allowed to hold ${item.name}.`];
 				}
-				if (itemTable[item.id]) return ["You are limited to one of each mega stone.", "(You have more than one " + item.name + ")"];
-				itemTable[item.id] = true;
+				if (itemTable.has(item.id)) {
+					return ["You are limited to one of each mega stone.", "(You have more than one " + item.name + ")"];
```
Zarel, line 565:
```suggestion
					return [
						`You are limited to one of each mega stone.`,
						`(You have more than one ${item.name})`
					];
```

(I don't go out of my way to change old strings into template strings, but if you're going to be modifying them anyway, I think that's a great time to do it.)

config/formats.ts
```diff
@@ -675,11 +673,8 @@ const Formats = [
 			'Regenerator ++ Wimp Out', 'Sand Rush', 'Sand Veil', 'Shadow Tag', 'Simple', 'Slush Rush', 'Snow Cloak',
 			'Speed Boost', 'Steelworker ++ Steely Spirit', 'Tinted Lens', 'Trace', 'Unaware', 'Unburden', 'Water Bubble',
 		],
-		/** @param {Pokemon} pokemon */
-		// @ts-ignore
 		getSharedPower(pokemon) {
-			/** @type {Set<string>} */
-			const sharedPower = new Set();
+			const sharedPower: Set<string> = new Set();
```
Zarel, line 677:
```suggestion
			const sharedPower = new Set<string>();
```

config/formats.ts
KrisXV: I mean, I guess? They are remnants of the OU Melmetal suspect test that happened ages ago, so it seems very unproblematic to do this here.

PR #6762, opened 2020-05-29
Youtube: Accredit videos and improve interval

mia-pi-git:
Also re-adds a limiter that got removed somehow?
Should be the last change that needs to be made for a while.
e; rebased wrong apparently but the rest of the changes are the same as master so only the final commit should matter.


server/chat-plugins/youtube.ts
```diff
@@ -217,11 +221,15 @@ export const commands: ChatCommands = {
 			this.runBroadcast();
 			if (this.broadcasting) {
 				return YouTube.generateChannelDisplay(channel).then(res => {
+					if (!res) return this.errorReply(`Error in getting channel data.`);
 					this.addBox(res);
 					room.update();
 				});
 			} else {
-				return YouTube.generateChannelDisplay(channel).then(res => this.sendReplyBox(res));
+				return YouTube.generateChannelDisplay(channel).then(res => {
```
Zarel, line 229: Personally I'd refactor this into an async function.

server/chat-plugins/youtube.ts
```diff
@@ -267,22 +262,27 @@ export const commands: ChatCommands = {
...
+			if (isNaN(parseInt(target))) return this.errorReply(`Specify a number (in minutes) for the interval.`);
 			let interval = Number(target);
 			if (interval < 10) return this.errorReply(`${interval} is too low - set it above 10 minutes.`);
 			interval = interval * 60 * 1000;
+			YouTube.timer = interval;
 			if (YouTube.interval) clearInterval(YouTube.interval);
-			YouTube.interval = setInterval(
-				() => void YouTube.randChannel().then(res => {
+			YouTube.interval =
+			setInterval(() => {
```
Zarel, line 278:
```suggestion
			YouTube.interval = setInterval(() => {
```

server/chat-plugins/youtube.ts
```diff
@@ -267,22 +262,27 @@ export const commands: ChatCommands = {
...
+				void (async () => {
+					const res = await YouTube.randChannel();
 					if (!res) return this.errorReply(`Error in getting channel data.`);
```
Zarel, line 280: Every 10 minutes from now until infinity, it might send an error message to the original user who used the `/repeat` command... that's probably not what you want.

Also, did you seriously add a command called `/repeat` at the top level? That's going to need to be removed.

server/chat-plugins/youtube.ts
Zarel: Also, shouldn't you clear the original interval, if it exists?

server/chat-plugins/youtube.ts
```diff
@@ -267,22 +262,27 @@ export const commands: ChatCommands = {
 			this.privateModAction(`(${user.name} updated channel ${id}'s username to ${name}.)`);
 			return FS(STORAGE_PATH).writeUpdate(() => JSON.stringify(channelData));
 		},
+		interval: 'repeat',
 		repeat(target, room, user) {
```
Zarel, line 266:
```suggestion
		videointerval: 'videorepeat',
		videorepeat(target, room, user) {
```

Zarel:
The `/update` command also needs to be renamed `/videoupdate` or something.

server/chat-plugins/youtube.ts
Zarel: oh, nevermind, this isn't at the top level

server/chat-plugins/youtube.ts
```diff
@@ -267,22 +262,28 @@ export const commands: ChatCommands = {
...
+			if (!target) return this.sendReply(`Interval is currently set to ${Chat.toDurationString(YouTube.timer)}.`);
+			if (Object.keys(channelData).length < 1) return this.errorReply(`No channels in the database.`);
+			if (isNaN(parseInt(target))) return this.errorReply(`Specify a number (in minutes) for the interval.`);
 			let interval = Number(target);
 			if (interval < 10) return this.errorReply(`${interval} is too low - set it above 10 minutes.`);
 			interval = interval * 60 * 1000;
+			const res = await YouTube.randChannel();
+			// no channels 
+			if (!res) return this.errorReply(`Error in getting channel data.`);
+			YouTube.timer = interval;
```
Zarel, line 278: What happens if `YouTube.timer` already exists?

mia-pi-git: it should be overwritten, i believe.

Zarel: Oh, wait, `YouTube.timer` and `YouTube.interval` are just confusingly named.

Maybe `YouTube.interval` and `YouTube.intervalTime` would be clearer?

server/chat-plugins/youtube.ts
```diff
@@ -267,22 +262,28 @@ export const commands: ChatCommands = {
 			this.privateModAction(`(${user.name} updated channel ${id}'s username to ${name}.)`);
 			return FS(STORAGE_PATH).writeUpdate(() => JSON.stringify(channelData));
 		},
+		interval: 'repeat',
 		repeat(target, room, user) {
```
Zarel, line 266:
```suggestion
		async repeat(target, room, user) {
```

Zarel:
This is going to cause a global crash if a network request fails... But I guess we can fix that in another PR.

PR #6765, opened 2020-05-29
Fix bugs caused by not specifying effectiveness priority

urkerab:
Inverse Mod needs to go first, to calculate the negated effectiveness.
Disguise goes second, to suppress effectiveness.
Delta Stream goes third, to weaken moves super-effective against Flying types.
Tar Shot goes last, to make its victim weak to Fire type moves.

This allows the existing test for Delta Stream with Tar Shot to pass.
Additionally a new test for Delta Stream with Inverse Mod now passes.
A test for Flying Press with Inverse Mod is also included.


Zarel:
Thanks!

PR #6766, opened 2020-05-29
Disguise breaking prevents critical hits

urkerab:
Basically in the same way that effectiveness is ignored.

Also enabling the test, of course.


Zarel:
Thanks!

PR #6768, opened 2020-05-30
Typescript Trivia

KrisXV:


server/chat-plugins/trivia.ts
```diff
@@ -71,24 +68,45 @@ const INTERMISSION_INTERVAL = 20 * 1000;
...
+interface TriviaQuestion {
+	question: string;
+	category: string;
+	answers: string[];
+	user?: string;
+	type?: string;
+}
+
+// Should be [number, number, number]
+type TriviaRank = [number, number, number];
```
Zarel: ...is that comment still necessary?

```suggestion
type TriviaRank = [number, number, number];
```

server/chat-plugins/trivia.ts
```diff
@@ -1659,14 +1604,14 @@ const commands = {
...
 		if (!allTimeScore) return this.sendReplyBox(`User '${name}' has not played any trivia games yet.`);
-		const score = triviaData.altLeaderboard[userid] || [0, 0, 0];
+		const score = triviaData.altLeaderboard![userid] || [0, 0, 0];
 
 		const ranks = cachedAltLadder.get().ranks[userid];
 		const allTimeRanks = cachedLadder.get().ranks[userid];
-		const row = i => `<strong>${score[i]}</strong>${ranks ? ` (#${ranks[i]})` : ""}, ` +
-			`all time: <strong>${allTimeScore[i]}</strong> (#${allTimeRanks ? allTimeRanks[i] : allTimeRanks.length})<br />`;
+		const row = (i: number) => `<strong>${score[i]}</strong>${ranks ? ` (#${ranks[i]})` : ""}, ` +
+			`all time: <strong>${allTimeScore[i]}</strong> (#${allTimeRanks ? allTimeRanks[i] : allTimeRanks!.length})<br />`;
```
Zarel, line 1613: What, what does this mean? If `allTimeRanks` is falsy, `allTimeRanks.length` should be meaningless...

KrisXV: yeah, it was a remnant from awhile ago; fixed

server/chat-plugins/trivia.ts
```diff
@@ -1659,14 +1603,14 @@ const commands = {
...
+		const allTimeScore = triviaData.leaderboard![userid];
 		if (!allTimeScore) return this.sendReplyBox(`User '${name}' has not played any trivia games yet.`);
-		const score = triviaData.altLeaderboard[userid] || [0, 0, 0];
+		const score = triviaData.altLeaderboard![userid] || [0, 0, 0];
 
 		const ranks = cachedAltLadder.get().ranks[userid];
-		const allTimeRanks = cachedLadder.get().ranks[userid];
-		const row = i => `<strong>${score[i]}</strong>${ranks ? ` (#${ranks[i]})` : ""}, ` +
-			`all time: <strong>${allTimeScore[i]}</strong> (#${allTimeRanks ? allTimeRanks[i] : allTimeRanks.length})<br />`;
+		const allTimeRanks: TriviaRank = cachedLadder.get().ranks[userid];
```
Zarel, line 1611: I can't figure out a reason this can't be inferred.

```suggestion
		const allTimeRanks = cachedLadder.get().ranks[userid];
```

server/chat-plugins/trivia.ts
```diff
@@ -344,17 +348,12 @@ class Trivia extends Rooms.RoomGame {
...
 	getRoundLength() {
 		return 12 * 1000 + 500;
 	}
 
-	/**
-	 * @param {User} user
-	 * @return {string | undefined}
-	 */
-	addPlayer(user) {
+	addTriviaPlayer(user: User) {
```
Zarel, line 356: Why was this renamed?

KrisXV: `addPlayer` is already a function in `Room.RoomGame`, and this function was attempting to mutate the return of the base function. This modifies it to allow for the proper return statements required for error checking when a user tries to enter a Trivia game and still adds the player to the game; renaming was largely unproblematic because It already called `super.addPlayer` if the inputted user passed all error checking.

Zarel: I suppose this isn't a bad solution, but I wonder if exceptions would be better.

PR #6770, opened 2020-05-30
Trivia: Fix bug in getting the game

mia-pi-git:
room.getGame always returned null, since trivia games have different constructor names, so no one could join and no games could be ended.

### DIFF
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -122,6 +122,22 @@ function isTriviaRoom(room: Room) {
 	return false;
 }
 
+export function getTriviaGame(context: CommandContext) {
+	const room = Rooms.get('trivia');
+	if (!room) return false;
+	if (!room.game) {
+		context.errorReply(`There is no game in progress.`);
+		return false;
+	}
+	if (room.game && !room.game.gameid.includes('trivia')) {
+		context.errorReply(`There is already a game of ${room?.game.title} running.`);
+		return false;
+	}
+	if (room.game.gameid.includes('trivia')) {
+		return room.game as Trivia;
+	}
+}
+
 function writeTriviaData() {
 	FS(PATH).writeUpdate(() => (
 		JSON.stringify(triviaData, null, 2)
@@ -1099,11 +1115,8 @@ const commands: ChatCommands = {
 
 	join(target, room, user) {
 		if (!isTriviaRoom(room)) return this.errorReply("This command can only be used in Trivia.");
-		const game = room.getGame(Trivia);
-		if (!game) return this.errorReply("There is no game of trivia in progress.");
-		if (game.gameid !== 'trivia') {
-			return this.errorReply(`There is already a game of ${game.title} in progress.`);
-		}
+		const game = getTriviaGame(this);
+		if (!game) return;
 		const res = game.addTriviaPlayer(user);
 		if (res) return this.errorReply(res);
 		this.sendReply('You are now signed up for this game!');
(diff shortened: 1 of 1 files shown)


server/chat-plugins/trivia.ts
```diff
@@ -122,6 +122,22 @@ function isTriviaRoom(room: Room) {
...
+		context.errorReply(`There is no game in progress.`);
+		return false;
+	}
+	if (room.game && !room.game.gameid.includes('trivia')) {
+		context.errorReply(`There is already a game of ${room?.game.title} running.`);
+		return false;
+	}
+	if (room.game.gameid.includes('trivia')) {
+		return room.game as Trivia;
+	}
```
Zarel, line 138: This should probably be modeled way more like `getGame`? Just make sure the constructor ends with `Trivia`.

server/chat-plugins/trivia.ts
```diff
@@ -122,6 +122,22 @@ function isTriviaRoom(room: Room) {
 	return false;
 }
 
+function getTriviaGame(context) {
+	const room = Rooms.get('trivia');
+	if (!room) return false;
```
Zarel, line 127:
```suggestion
	const room = context.room;
	if (!room) return false;
```

Zarel: (The original command wasn't scoped to the Trivia room, so this shouldn't be, either.)

server/chat-plugins/trivia.ts
```diff
@@ -122,6 +122,22 @@ function isTriviaRoom(room: Room) {
...
+		context.errorReply(`There is no game in progress.`);
+		return false;
+	}
+	const name = room.game.constructor.name;
+	if (!name.endsWith('Trivia')) {
+		context.errorReply(`There is already a game of ${room.game.title} running.`);
+		return false;
+	} else {
+		return room.game as Trivia;
+	}
```
Zarel, line 138:
```suggestion
	const game = room.game;
	if (!game) {
		context.errorReply(`There is no game in progress.`);
		return false;
	}
	if (!game.constructor.name.endsWith('Trivia')) {
		context.errorReply(`The currently running game is not Trivia, it's ${room.game.title}.`);
		return false;
	} else {
		return game as Trivia;
	}
```

server/chat-plugins/trivia.ts
```diff
@@ -122,6 +122,22 @@ function isTriviaRoom(room: Room) {
...
+	if (!game) {
+		context.errorReply(`There is no game in progress.`);
+		return false;
+	}
+	if (!game.constructor.name.endsWith('Trivia')) {
+		context.errorReply(`The currently running game is not Trivia, it's ${room.game.title}.`);
+		return false;
+	} else {
+		return game as Trivia;
+	}
```
Zarel, line 138:
```suggestion
	}
	return game as Trivia;
```

server/chat-plugins/trivia.ts
```diff
@@ -122,6 +122,21 @@ function isTriviaRoom(room: Room) {
...
+function getTriviaGame(context: CommandContext) {
+	const room = context.room;
+	if (!room) return false;
+	const game = room.game;
+	if (!game) {
+		context.errorReply(`There is no game in progress.`);
+		return false;
+	}
+	if (!game.constructor.name.endsWith('Trivia')) {
+		context.errorReply(`The currently running game is not Trivia, it's ${room.game.title}.`);
```
Zarel, line 134:
```suggestion
		context.errorReply(`The currently running game is not Trivia, it's ${game.title}.`);
```

PR #6771, opened 2020-05-30
Update moves.ts

SSlayer9:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8482944

I looked it up, and Soundproof Pokémon are now affected by Heal Bell in Generation 8.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -7881,7 +7881,7 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		accuracy: true,
 		basePower: 0,
 		category: "Status",
-		desc: "Every Pokemon in the user's party is cured of its major status condition. Allies with Soundproof are not healed.",
+		desc: "Every Pokemon in the user's party is cured of its major status condition. Active Pokemon with the Soundproof Ability are not cured, unless they are the user.",
 		shortDesc: "Cures the user's party of all status conditions.",
 		name: "Heal Bell",
 		pp: 5,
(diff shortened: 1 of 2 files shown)


data/moves.ts
```diff
@@ -7881,7 +7881,7 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		accuracy: true,
 		basePower: 0,
 		category: "Status",
-		desc: "Every Pokemon in the user's party is cured of its major status condition. Active Pokemon with the Soundproof Ability are not cured.",
```
HoeenCoder, line 7884: Don't forget to place the old description in gen 7's version of the move so it displays correct information for gen 7.

SSlayer9: I believe the description was updated with my last update. Why did the check fail?

PartMan7: You indented with spaces instead of a tab; does your device have a tab key?

SSlayer9: It does, but I'm not generally one for neat code so I've never used it before. Please do edit it to make it cleaner if necessary.

HoeenCoder: RE: description, github didn't update for me or indicate I was out-of-date then sorry!

RE: Check failed, `Expected indentation of 2 tabs but found 6 spaces` means that spaces were used for indentation rather than tab characters, PS enforces the use of tab characters on its repositories. 

Inserting tab characters rather than spaces is as easy as hitting the tab key ofc, but depending on your code editor the tab key might be inserting spaces instead of tabs. Thats usually configurable, how exactly to do that depends on your editor though.

Marty-D: This isn't true; Heal Bell still doesn't affect allies with Soundproof in Gen 8. The description and implementation need changing, and you can check Aromatherapy for how to do this appropriately.

SSlayer9: Do you mind editing in the tabs instead of spaces? I don't really understand the difference it makes, and I don't really know how to change it.

SSlayer9: @Marty-D I've fixed that now, so this should work now. (Please edit the spaces to tabs though, I haven't done that)

Marty-D:
Neat, thanks!

PR #6774, opened 2020-05-30
Fix old generation dexsearching by move.

SSlayer9:
Potential fix to https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8469484.

Untested and will probably fail.

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -540,8 +540,8 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 				if (invalid) return {error: invalid};
 				if (!isNaN(parseInt(cmd.replace('ds', '')))) {
 					 const gen = parseInt(cmd.replace('ds', ''));
-					if (gen < targetMove.gen) return {error: `${targetMove.name} is not available in this generation.`}
-				}
+					if (gen < targetMove.gen) return {error: `${targetMove.name} is not available in this generation.`};
+				};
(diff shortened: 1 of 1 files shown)


server/chat-plugins/datasearch.ts
```diff
@@ -538,6 +538,10 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 			if (targetMove.exists) {
 				const invalid = validParameter("moves", targetMove.id, isNotSearch, target);
 				if (invalid) return {error: invalid};
+				if (!isNaN(parseInt(cmd.replace('ds', '')))) {
+					 let gen = partseInt(cmd.replace('ds', ''));
```
mia-pi-git, line 542:
```suggestion
					 let gen = parseInt(cmd.replace('ds', ''));
```

KrisXV:
Why not just move the maxgen check above the other checks?

mia-pi-git:
This code doesn't work yet - you should probably change this to a draft PR for now.

AnnikaCodes:
> Why not just move the maxgen check above the other checks? That way, you can just do `if (move.gen > dex.gen)`

This only works if the `maxgen` parameter is parsed before the move, which is not the case when using `/ds(gen #)`.
The way I'd handle this is to compare the move's generation and `maxGen` later on, when all the parameters have been parsed and the search is actually being run. I have tested this and confirmed it works.

SSlayer9:
Left to annika

PR #6775, opened 2020-05-31
Datasearch: Support searching for fully evolved Pokemon

AnnikaCodes:
I also redesigned dexsearchhelp.
(suggestion [here](https://www.smogon.com/forums/threads/more-clear-ds-paramater-to-exclude-nfe-and-lc-mons.3665165/#post-8484543))


Zarel:
Thanks!

PR #6777, opened 2020-05-31
Refactor auth and room.settings

mia-pi-git:
This PR implements two big changes:

- All settings shared between `room.chatRoomData` and `room`  have been merged into `room.settings` (so, for instance, `room.slowchat` is now only `room.settings.slowchat`).

  This makes it so we never have to worry about them getting "out of sync".

  - Checking to see if a room is persistent is now `if (room.persist)` instead of `if (room.chatRoomData)`

  - `Rooms.global.writeChatRoomData()` is now rarely called directly; there's a new `room.saveSettings()` which will handle it for you.

  - All properties of `room.settings` are now optional (except `title`).

- There's a new file `user-groups.ts` which handles authority.

  - `room.auth` and `Users.globalAuth` are now `Auth extends Map<ID, GroupSymbol>` objects.

  - `room.auth` is now always defined, removing the need for `room.auth?.[userid]` workarounds.

  - A lot of code relating to usergroups and permission checks have been refactored.


server/rooms.ts
```diff
@@ -63,6 +63,46 @@ interface UserTable {
 	[userid: string]: User;
 }
 
+interface Settings {
```
Zarel, line 66:
```suggestion
interface RoomSettings {
```

Zarel: (remember to rename it everywhere else, too)

Zarel:
I hate to do this after you've already done so much work, but I think we should probably refactor authority into its own file (like I was talking about in sim-dev) before doing this. Your current approach to auth is a bit too messy for me.

Zarel:
Hm, on second thought, I don't think there's anything _too_ wrong with doing this before the auth refactor. My main objection is to `.settings?.auth!`. `settings` and `auth` should both be always-defined so you don't have to do that.

mia-pi-git:
> 

> Hm, on second thought, I don't think there's anything _too_ wrong with doing this before the auth refactor. My main objection is to `.settings?.auth!`. `settings` and `auth` should both be always-defined so you don't have to do that.

Settings should already be always defined. I'll look at that. 
As for auth, I'll change that.

server/chat-commands/info.ts
```diff
@@ -42,7 +42,9 @@ export const commands: ChatCommands = {
 		}
 		if (!targetUser.connected) buf += ` <em style="color:gray">(offline)</em>`;
 		let roomauth = '';
-		if (usedRoom.auth && targetUser.id in usedRoom.auth) roomauth = usedRoom.auth[targetUser.id];
+		if (usedRoom.settings.auth && targetUser.id in usedRoom.settings.auth) {
```
Zarel, line 45:
```suggestion
		if (targetUser.id in usedRoom.settings.auth) {
```

server/chat-commands/info.ts
```diff
@@ -62,14 +64,15 @@ export const commands: ChatCommands = {
 			if (roomid === 'global') continue;
 			const targetRoom = Rooms.get(roomid)!;
 
-			const authSymbol = (targetRoom.auth && targetRoom.auth[targetUser.id] ? targetRoom.auth[targetUser.id] : '');
+			const authSymbol = (targetRoom.settings.auth && targetRoom.settings.auth[targetUser.id] ?
+				targetRoom.settings.auth[targetUser.id] : '');
```
Zarel, line 68:
```suggestion
			const authSymbol = targetRoom.settings.auth[targetUser.id] || '';
```

Zarel: Making `auth` guaranteed to exist actually simplifies a lot of code along these lines.

server/chat-commands/info.ts
```diff
@@ -248,7 +251,7 @@ export const commands: ChatCommands = {
 		if (!targetUser || !targetUser.connected) buf += ` <em style="color:gray">(offline)</em>`;
 
 		let roomauth = '';
-		if (room?.auth && userid in room.auth) roomauth = room.auth[userid];
+		if (room?.settings.auth && userid in room.settings.auth) roomauth = room.settings.auth[userid];
```
Zarel, line 254:
```suggestion
		if (room && userid in room.settings.auth) roomauth = room.settings.auth[userid];
```

server/chat-commands/info.ts
```diff
@@ -317,8 +320,8 @@ export const commands: ChatCommands = {
 		const battles = [];
 		for (const curRoom of Rooms.rooms.values()) {
 			if (!curRoom.battle) continue;
-			if ((user1?.inRooms.has(curRoom.roomid) || (curRoom.auth && curRoom.auth[userID1])) &&
-				(user2?.inRooms.has(curRoom.roomid) || (curRoom.auth && curRoom.auth[userID2]))) {
+			if ((user1?.inRooms.has(curRoom.roomid) || (curRoom.settings.auth && curRoom.settings.auth[userID1])) &&
+				(user2?.inRooms.has(curRoom.roomid) || (curRoom.settings.auth && curRoom.settings.auth[userID2]))) {
```
Zarel, line 324:
```suggestion
			if (
				(user1?.inRooms.has(curRoom.roomid) || userID1 in curRoom.settings.auth) &&
				(user2?.inRooms.has(curRoom.roomid) || userID2 in curRoom.settings.auth)
			) {
```

server/chat-commands/info.ts
```diff
@@ -1886,7 +1889,7 @@ export const commands: ChatCommands = {
 				`- /rules <em>rules link</em>: set the room rules link seen when using /rules`,
 				`- /roommod, /roomdriver <em>username</em>: appoint a room moderator/driver`,
 				`- /roomdemod, /roomdedriver <em>username</em>: remove a room moderator/driver`,
-				`- /roomdeauth <em>username</em>: remove all room auth from a user`,
+				`- /roomdeauth <em>username</em>: remove all room.settings.auth from a user`,
```
Zarel, line 1892:
```suggestion
				`- /roomdeauth <em>username</em>: remove all room auth from a user`,
```

Let me guess, you used a regex so searching for `room.auth` meant "any character between `room` and `auth`".

mia-pi-git: correct! i thought i got a lot of these reverted but apparently i missed this one

server/chat-commands/info.ts
```diff
@@ -334,7 +337,7 @@ export const commands: ChatCommands = {
 
 	sp: 'showpunishments',
 	showpunishments(target, room, user) {
-		if (!room.chatRoomData || room.roomid.includes('-')) {
+		if (!room.settings || room.roomid.includes('-')) {
```
Zarel, line 340:
```suggestion
		if (!room.settings.persistSettings) {
```

server/chat-commands/info.ts
```diff
@@ -1949,18 +1952,17 @@ export const commands: ChatCommands = {
...
+			room.settings.rulesLink = target;
 			this.privateModAction(`(${user.name} changed the room rules link to: ${target})`);
 			this.modlog('RULES', null, `changed link to: ${target}`);
 		}
 
-		if (room.chatRoomData) {
-			room.chatRoomData.rulesLink = room.rulesLink;
+		if (room.settings.persistSettings) {
 			Rooms.global.writeChatRoomData();
 		}
```
Zarel, line 1967: I think we should provide a `room.saveSettings()` that does this internally.

server/rooms.ts
```diff
@@ -63,6 +63,46 @@ interface UserTable {
 	[userid: string]: User;
 }
 
+interface RoomSettings {
+	aliases: string[];
+	auth: {[userid: string]: GroupSymbol} | null;
```
Zarel, line 68:
```suggestion
	auth: {[userid: string]: GroupSymbol};
```

I thought you said you made this non-null?

server/rooms.ts
```diff
@@ -177,44 +184,54 @@ export abstract class BasicRoom {
 
 		// room settings
 
-		this.chatRoomData = null;
-		this.isPrivate = false;
+		this.settings = {
+			aliases: [],
+			auth: null,
```
Zarel, line 189:
```suggestion
			auth: {},
```

server/rooms.ts
```diff
@@ -396,14 +415,16 @@ export abstract class BasicRoom {
...
-		if (this.staffRoom && !user.isStaff && (!this.auth || (this.auth[user.id] || ' ') === ' ')) return false;
+		if (this.settings.staffRoom && !user.isStaff &&
+			 (!this.settings.auth || (this.settings.auth[user.id] || ' ') === ' ')
+		) return false;
 		if (user.id in this.users) return true;
-		if (!this.modjoin) return true;
+		if (!this.settings.modjoin) return true;
 		// users with a room rank can always join
-		if (this.auth && user.id in this.auth) return true;
+		if (this.settings.auth && user.id in this.settings.auth) return true;
```
Zarel, line 424:
```suggestion
		if (user.id in this.settings.auth) return true;
```

server/rooms.ts
mia-pi-git: Huh, i thought i had committed that. I know i did it locally. Apparently not?

server/rooms.ts
```diff
@@ -396,14 +415,16 @@ export abstract class BasicRoom {
 		return globalGroup;
 	}
 	checkModjoin(user: User) {
-		if (this.staffRoom && !user.isStaff && (!this.auth || (this.auth[user.id] || ' ') === ' ')) return false;
+		if (this.settings.staffRoom && !user.isStaff &&
+			 (!this.settings.auth || (this.settings.auth[user.id] || ' ') === ' ')
+		) return false;
```
Zarel, line 420:
```suggestion
		if (this.settings.staffRoom && !user.isStaff && !(user.id in this.settings.auth)) {
			return false;
		}
```

Zarel: In general, I'd strongly prefer

```js
if (a) {
  b;
}
```

instead of

```js
if (
  a
) b;
```

server/chat-commands/info.ts
```diff
@@ -193,8 +195,8 @@ export const commands: ChatCommands = {
 		for (const curRoom of Rooms.rooms.values()) {
 			if (!curRoom.game) continue;
 			if ((targetUser.id in curRoom.game.playerTable && !targetUser.inRooms.has(curRoom.roomid)) ||
-				(curRoom.auth && curRoom.auth[targetUser.id] === Users.PLAYER_SYMBOL)) {
-				if (curRoom.isPrivate && !canViewAlts) {
+				(curRoom.settings.auth && curRoom.settings.auth[targetUser.id] === Users.PLAYER_SYMBOL)) {
```
Zarel, line 198:
```suggestion
				curRoom.settings.auth[targetUser.id] === Users.PLAYER_SYMBOL)) {
```

server/chat-commands/info.ts
```diff
@@ -2522,7 +2531,7 @@ export const pages: PageTable = {
 		let buf = "";
 		this.extractRoom();
 		if (!user.named) return Rooms.RETRY_AFTER_LOGIN;
-		if (!this.room.chatRoomData) return;
+		if (!this.room.settings) return;
```
Zarel, line 2534:
```suggestion
		if (!this.room.settings.persistSettings) return;
```

server/chat-commands/moderation.ts
```diff
@@ -21,7 +21,7 @@ const REQUIRE_REASONS = true;
 export const commands: ChatCommands = {
 
 	roomowner(target, room, user) {
-		if (!room.chatRoomData) {
+		if (!room.settings) {
```
Zarel, line 24:
```suggestion
		if (!room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -37,9 +37,9 @@ export const commands: ChatCommands = {
 
 		if (!this.can('makeroom')) return false;
 
-		if (!room.auth) room.auth = room.chatRoomData.auth = {};
+		if (!room.settings.auth) room.settings.auth = room.settings.auth = {};
```
Zarel, line 40:
```suggestion
```

server/chat-commands/moderation.ts
```diff
@@ -62,7 +62,7 @@ export const commands: ChatCommands = {
 			// this command isn't marked as room-only because it's usable in PMs through /invite
 			return this.errorReply("This command is only available in rooms");
 		}
-		if (!room.auth) {
+		if (!room.settings.auth) {
 			this.sendReply("/roompromote - This room isn't designed for per-room moderation.");
 			return this.sendReply("Before setting room staff, you need to set a room owner with /roomowner");
 		}
```
Zarel, line 68:
```suggestion
```

server/chat-commands/moderation.ts
```diff
@@ -163,7 +163,7 @@ export const commands: ChatCommands = {
 				}
 			}
 		}
-		if (room.chatRoomData) Rooms.global.writeChatRoomData();
+		if (room.settings) room.saveSettings();
```
Zarel, line 166:
```suggestion
		room.saveSettings();
```

server/chat-commands/moderation.ts
```diff
@@ -220,12 +220,12 @@ export const commands: ChatCommands = {
...
 		}
-		if (!curRoom.isPrivate) {
+		if (!curRoom.settings.isPrivate) {
 			buffer.push(`${curRoom.title} is a public room, so global auth with no relevant roomauth will have authority in this room.`);
-		} else if (curRoom.isPrivate === 'hidden' || curRoom.isPrivate === 'voice') {
+		} else if (curRoom.settings.isPrivate === 'hidden' || curRoom.settings.isPrivate === 'voice') {
 			buffer.push(`${curRoom.title} is a hidden room, so global auth with no relevant roomauth will have authority in this room.`);
 		}
-		if (targetRoom !== room) buffer.unshift(`${targetRoom.title} room auth:`);
+		if (targetRoom !== room) buffer.unshift(`${targetRoom.title} room.settings.auth:`);
```
Zarel, line 228:
```suggestion
		if (targetRoom !== room) buffer.unshift(`${targetRoom.title} room auth:`);
```

server/chat-commands/moderation.ts
```diff
@@ -244,8 +244,8 @@ export const commands: ChatCommands = {
 			buffer.push(`Global auth: ${group}`);
 		}
 		for (const curRoom of Rooms.rooms.values()) {
-			if (!curRoom.auth || curRoom.isPrivate) continue;
-			group = curRoom.auth[targetId];
+			if (!curRoom.settings.auth || curRoom.settings.isPrivate) continue;
```
Zarel, line 247:
```suggestion
			if (curRoom.settings.isPrivate) continue;
```

server/chat-commands/moderation.ts
```diff
@@ -255,31 +255,31 @@ export const commands: ChatCommands = {
 		if (targetId === user.id || user.can('lock')) {
 			innerBuffer = [];
 			for (const curRoom of Rooms.rooms.values()) {
-				if (!curRoom.auth || !curRoom.isPrivate) continue;
-				if (curRoom.isPrivate === true) continue;
-				const auth = curRoom.auth[targetId];
+				if (!curRoom.settings.auth || !curRoom.settings.isPrivate) continue;
```
Zarel, line 258:
```suggestion
				if (!curRoom.settings.isPrivate) continue;
```

server/chat-commands/moderation.ts
```diff
@@ -255,31 +255,31 @@ export const commands: ChatCommands = {
...
+				if (curRoom.settings.isPrivate === true) continue;
+				const auth = curRoom.settings.auth[targetId];
 				if (!auth) continue;
 				innerBuffer.push(auth + curRoom.roomid);
 			}
 			if (innerBuffer.length) {
-				buffer.push(`Hidden room auth: ${innerBuffer.join(', ')}`);
+				buffer.push(`Hidden room.settings.auth: ${innerBuffer.join(', ')}`);
```
Zarel, line 265:
```suggestion
				buffer.push(`Hidden room auth: ${innerBuffer.join(', ')}`);
```

server/chat-commands/moderation.ts
```diff
@@ -255,31 +255,31 @@ export const commands: ChatCommands = {
...
 			}
 		}
 		if (targetId === user.id || user.can('makeroom')) {
 			innerBuffer = [];
 			for (const chatRoom of Rooms.global.chatRooms) {
-				if (!chatRoom.auth || !chatRoom.isPrivate) continue;
-				if (chatRoom.isPrivate !== true) continue;
-				const auth = chatRoom.auth[targetId];
+				if (!chatRoom.settings.auth || !chatRoom.settings.isPrivate) continue;
```
Zarel, line 271:
```suggestion
				if (!chatRoom.settings.isPrivate) continue;
```

server/chat-commands/moderation.ts
```diff
@@ -255,31 +255,31 @@ export const commands: ChatCommands = {
...
+				if (chatRoom.settings.isPrivate !== true) continue;
+				const auth = chatRoom.settings.auth[targetId];
 				if (!auth) continue;
 				innerBuffer.push(auth + chatRoom.roomid);
 			}
 			if (innerBuffer.length) {
-				buffer.push(`Private room auth: ${innerBuffer.join(', ')}`);
+				buffer.push(`Private room.settings.auth: ${innerBuffer.join(', ')}`);
```
Zarel, line 278:
```suggestion
				buffer.push(`Private room auth: ${innerBuffer.join(', ')}`);
```

server/chat-commands/moderation.ts
```diff
@@ -255,31 +255,31 @@ export const commands: ChatCommands = {
...
 			}
 		}
 		if (!buffer.length) {
-			buffer.push("No global or room auth.");
+			buffer.push("No global or room.settings.auth.");
```
Zarel, line 282:
```suggestion
			buffer.push("No global or room auth.");
```

server/chat-commands/moderation.ts
```diff
@@ -553,13 +553,13 @@ export const commands: ChatCommands = {
...
 			Monitor.log(`[CrisisMonitor] Trusted user ${targetUser.name} ${(targetUser.trusted !== targetUser.id ? ` (${targetUser.trusted})` : ``)} was roombanned from ${room.roomid} by ${user.name}, and should probably be demoted.`);
 		}
 
 		if (targetUser.id in room.users || user.can('lock')) {
 			targetUser.popup(
-				`|modal||html|<p>${Chat.escapeHTML(user.name)} has banned you from the room ${room.roomid} ${(room.subRooms ? ` and its subrooms` : ``)}.</p>${(target ? `<p>Reason: ${Chat.escapeHTML(target)}</p>` : ``)}<p>To appeal the ban, PM the staff member that banned you${(!room.battle && room.auth ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
+				`|modal||html|<p>${Chat.escapeHTML(user.name)} has banned you from the room ${room.roomid} ${(room.subRooms ? ` and its subrooms` : ``)}.</p>${(target ? `<p>Reason: ${Chat.escapeHTML(target)}</p>` : ``)}<p>To appeal the ban, PM the staff member that banned you${(!room.battle && room.settings.auth ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
```
Zarel, line 562:
```suggestion
				`|modal||html|<p>${Chat.escapeHTML(user.name)} has banned you from the room ${room.roomid} ${(room.subRooms ? ` and its subrooms` : ``)}.</p>${(target ? `<p>Reason: ${Chat.escapeHTML(target)}</p>` : ``)}<p>To appeal the ban, PM the staff member that banned you${(!room.battle ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
```

server/chat-commands/moderation.ts
```diff
@@ -568,7 +568,7 @@ export const commands: ChatCommands = {
 
 		const affected = Punishments.roomBan(room, targetUser, null, null, target);
 
-		if (!room.isPrivate && room.chatRoomData) {
+		if (!room.settings.isPrivate && room.settings) {
```
Zarel, line 571:
```suggestion
		if (!room.settings.isPrivate && room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -581,7 +581,7 @@ export const commands: ChatCommands = {
 		}
 		room.hideText([userid, toID(this.inputUsername)]);
 
-		if (room.isPrivate !== true && room.chatRoomData) {
+		if (room.settings.isPrivate !== true && room.settings) {
```
Zarel, line 584:
```suggestion
		if (room.settings.isPrivate !== true && room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -600,7 +600,7 @@ export const commands: ChatCommands = {
 
 		if (name) {
 			this.addModAction(`${name} was unbanned from ${room.title} by ${user.name}.`);
-			if (room.isPrivate !== true && room.chatRoomData) {
+			if (room.settings.isPrivate !== true && room.settings) {
```
Zarel, line 603:
```suggestion
			if (room.settings.isPrivate !== true && room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -971,18 +971,18 @@ export const commands: ChatCommands = {
...
 			}
 		}
 		if (!count) {
 			return this.sendReply("(This room has zero roomvoices)");
 		}
-		if (room.chatRoomData) {
-			Rooms.global.writeChatRoomData();
+		if (room.settings) {
+			room.saveSettings();
 		}
```
Zarel, line 986:
```suggestion
		room.saveSettings();
```

server/chat-commands/moderation.ts
```diff
@@ -1517,7 +1517,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(`If you want to blacklist an offline account by name (not IP), consider /blacklistname`);
 		}
 		if (!this.can('editroom', targetUser, room)) return false;
-		if (!room.chatRoomData) {
+		if (!room.settings) {
```
Zarel, line 1520:
```suggestion
		if (!room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -1543,22 +1543,22 @@ export const commands: ChatCommands = {
...
-		if (targetUser.trusted && room.isPrivate !== true) {
+		if (targetUser.trusted && room.settings.isPrivate !== true) {
 			Monitor.log(`[CrisisMonitor] Trusted user ${targetUser.name}${(targetUser.trusted !== targetUser.id ? ` (${targetUser.trusted})` : '')} was blacklisted from ${room.roomid} by ${user.name}, and should probably be demoted.`);
 		}
 
 		if (targetUser.id in room.users || user.can('lock')) {
 			targetUser.popup(
 				`|modal||html|<p>${Chat.escapeHTML(user.name)} has blacklisted you from the room ${room.roomid}${(room.subRooms ? ` and its subrooms` : '')}. Reason: ${Chat.escapeHTML(target)}</p>` +
-				`<p>To appeal the ban, PM the staff member that blacklisted you${(!room.battle && room.auth ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
+				`<p>To appeal the ban, PM the staff member that blacklisted you${(!room.battle && room.settings.auth ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
```
Zarel, line 1553:
```suggestion
				`<p>To appeal the ban, PM the staff member that blacklisted you${(!room.battle ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
```

server/chat-commands/moderation.ts
```diff
@@ -1543,22 +1543,22 @@ export const commands: ChatCommands = {
...
 			);
 		}
 
 		this.privateModAction(`(${name} was blacklisted from ${room.title} by ${user.name}. ${(target ? ` (${target})` : '')})`);
```
Zarel, line 1557:
```suggestion
		this.privateModAction(`(${name} was blacklisted from ${room.title} by ${user.name}. ${target ? ` (${target})` : ''})`);
```

server/chat-commands/moderation.ts
```diff
@@ -1543,22 +1543,22 @@ export const commands: ChatCommands = {
 		const name = targetUser.getLastName();
 		const userid = targetUser.getLastId();
 
-		if (targetUser.trusted && room.isPrivate !== true) {
+		if (targetUser.trusted && room.settings.isPrivate !== true) {
 			Monitor.log(`[CrisisMonitor] Trusted user ${targetUser.name}${(targetUser.trusted !== targetUser.id ? ` (${targetUser.trusted})` : '')} was blacklisted from ${room.roomid} by ${user.name}, and should probably be demoted.`);
```
Zarel, line 1547:
```suggestion
			Monitor.log(`[CrisisMonitor] Trusted user ${targetUser.name}${targetUser.trusted !== targetUser.id ? ` (${targetUser.trusted})` : ''} was blacklisted from ${room.roomid} by ${user.name}, and should probably be demoted.`);
```

server/chat-commands/moderation.ts
```diff
@@ -1543,22 +1543,22 @@ export const commands: ChatCommands = {
...
+				`<p>To appeal the ban, PM the staff member that blacklisted you${(!room.battle && room.settings.auth ? ` or a room owner. </p><p><button name="send" value="/roomauth ${room.roomid}">List Room Staff</button></p>` : `.</p>`)}`
 			);
 		}
 
 		this.privateModAction(`(${name} was blacklisted from ${room.title} by ${user.name}. ${(target ? ` (${target})` : '')})`);
 
 		const affected = Punishments.roomBlacklist(room, targetUser, null, null, target);
 
-		if (!room.isPrivate && room.chatRoomData) {
+		if (!room.settings.isPrivate && room.settings) {
```
Zarel, line 1561:
```suggestion
		if (!room.settings.isPrivate && room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -1570,7 +1570,7 @@ export const commands: ChatCommands = {
 			}
 		}
 
-		if (!room.isPrivate && room.chatRoomData) {
+		if (!room.settings.isPrivate && room.settings) {
```
Zarel, line 1573:
```suggestion
		if (!room.settings.isPrivate && room.settings.persistSettings) {
```

server/chat-commands/moderation.ts
```diff
@@ -1570,7 +1570,7 @@ export const commands: ChatCommands = {
...
 			this.globalModlog("BLACKLIST", targetUser, ` by ${user.id}${(target ? `: ${target}` : '')}`);
```
Zarel, line 1574:
```suggestion
			this.globalModlog("BLACKLIST", targetUser, ` by ${user.id}${target ? `: ${target}` : ''}`);
```

server/chat-commands/moderation.ts
```diff
@@ -1663,7 +1663,7 @@ export const commands: ChatCommands = {
 		if (!target) return this.parse('/help blacklistname');
 		if (!this.canTalk()) return;
 		if (!this.can('editroom', null, room)) return false;
-		if (!room.chatRoomData) {
+		if (!room.settings) {
```
Zarel, line 1666:
```suggestion
		if (!room.settings.persistSettings) {
```

mia-pi-git:
Drafting this since there's an error I can't locate within `Users#socketReceive` that i'm having difficulty finding. 
```CRASH: TypeError: room.auth.get is not a function
    at Object.socketReceive (/Users/mia/Documents/GitHub/pokemon-showdown-dev/.server-dist/users.js:1634:14)```
leaving it as draft until i can find it.

File: server/user-groups.ts [hunk 39]
```diff
@@ -0,0 +1,91 @@
...
+	constructor() {
+		super();
+		this.save = () => null;
+	}
+	setGroup(user: User, group: GroupSymbol) {
+		if (!user) throw new Error(`No user passed to setGroup, use forceGroup instead.`);
+		super.set(user.id, group);
+		return this.save;
+	}
+	forceGroup(userid: ID, group: GroupSymbol) {
```
Zarel, line 16: Why did you rename this from `setOfflineGroup` which I personally think was clearer?

Zarel: I actually think overriding `set` is probably better, anyway. Let me fiddle with this.

server/chat-commands/moderation.ts
```diff
@@ -36,9 +36,9 @@ export const commands: ChatCommands = {
 		}
 
 		if (!this.can('makeroom')) return false;
+		if (room.auth.has(toID(name))) return this.errorReply(`${name} is already a room owner.`);
```
Zarel, line 39: ...this means the user has some auth, not that the user is an RO.

mia-pi-git: oops i'm dense

server/chat-commands/moderation.ts
```diff
@@ -36,9 +36,9 @@ export const commands: ChatCommands = {
...
 
-
-		room.auth.setGroup(user, '#');
+		room.auth.set(toID(name), '#');
```
Zarel, line 41: You use `toID(name)` twice; it should be set to a variable `userid` instead.

Zarel: Or possibly `set` could return whether or not it was already that value; combining the check.

server/chat-commands/moderation.ts
```diff
@@ -116,11 +116,7 @@ export const commands: ChatCommands = {
 		if (nextGroup === Config.groupsranking[0]) {
 			room.auth.delete(userid);
 		} else {
-			if (!targetUser) {
-				room.auth.forceGroup(userid, nextGroup);
-			} else {
-				room.auth.setGroup(targetUser, nextGroup);
-			}
+			room.auth.set(toID(userid), nextGroup);
```
Zarel, line 119: Why `toID`? `userid` is already an ID.

server/chat-commands/moderation.ts
mia-pi-git: if it doesn't return `this`, typescript disagrees, since the base type `set` in `Map` returns the map object

server/chat-commands/core.ts
```diff
@@ -1453,18 +1451,15 @@ export const commands: ChatCommands = {
...
 				users: [],
 			};
 
-			if (targetRoom.auth) {
-				for (const userid in targetRoom.auth) {
-					const rank = targetRoom.auth[userid];
-					if (!roominfo.auth[rank]) roominfo.auth[rank] = [];
-					roominfo.auth[rank].push(userid);
-				}
+			for (const [k, v] of targetRoom.auth) {
```
Zarel, line 1460: `k, v` should not be used because it's unclear what they mean; I'd use `userid, groupSymbol` instead, here and everywhere else `auth` is iterated.

server/chat-commands/info.ts
```diff
@@ -42,7 +42,9 @@ export const commands: ChatCommands = {
 		}
 		if (!targetUser.connected) buf += ` <em style="color:gray">(offline)</em>`;
 		let roomauth = '';
-		if (usedRoom.auth && targetUser.id in usedRoom.auth) roomauth = usedRoom.auth[targetUser.id];
+		if (usedRoom.auth.has(targetUser.id)) {
+			roomauth = usedRoom.auth.get(targetUser.id) as string;
```
Zarel, line 46: Why is this a cast?

server/chat-commands/info.ts
```diff
@@ -62,14 +64,14 @@ export const commands: ChatCommands = {
 			if (roomid === 'global') continue;
 			const targetRoom = Rooms.get(roomid)!;
 
-			const authSymbol = (targetRoom.auth && targetRoom.auth[targetUser.id] ? targetRoom.auth[targetUser.id] : '');
+			const authSymbol = targetRoom.auth.get(targetUser.id) || '';
```
Zarel, line 67: We could make `get` default to returning `''` or something.

server/chat-commands/info.ts
Zarel: Oh, it returns `string | undefined`. It could just be `get(...)!`, but I would probably recommend making it default to returning `''`.

server/chat-commands/moderation.ts
Zarel: Unfortunate.

server/chat-commands/info.ts
Zarel: Which would simplify this (and also line 253) to `const roomauth = userRoom.auth.get(targetUser.id)`

server/chat-commands/info.ts
```diff
@@ -248,11 +250,11 @@ export const commands: ChatCommands = {
 		if (!targetUser || !targetUser.connected) buf += ` <em style="color:gray">(offline)</em>`;
 
 		let roomauth = '';
-		if (room?.auth && userid in room.auth) roomauth = room.auth[userid];
+		if (room?.auth.has(userid)) roomauth = room.auth.get(userid) as string;
```
Zarel, line 253: should just be

```
const roomauth = room?.auth.get(userid) || '';
```

server/chat-commands/info.ts
```diff
@@ -248,11 +250,11 @@ export const commands: ChatCommands = {
...
 		if (Config.groups[roomauth] && Config.groups[roomauth].name) {
 			buf += `<br />${Config.groups[roomauth].name} (${roomauth})`;
 		}
-		const group = (Users.usergroups[userid] || '').charAt(0);
+		const group = (Users.groups.get(userid) || '').charAt(0);
```
Zarel, line 257: `Users.groups.get` should also default to `''`, which would simplify this part, too.

server/chat-commands/moderation.ts
Zarel: This isn't resolved, you're still using `toID(name)`. The old code even correctly uses `userid`!

server/chat-commands/moderation.ts
```diff
@@ -86,7 +81,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(`User '${name}' is unregistered, and so can't be promoted.`);
 		}
 
-		const currentGroup = room.auth[userid] || Config.groupsranking[0];
+		const currentGroup = room.auth.get(userid) || Config.groupsranking[0];
```
Zarel, line 84: Should probably support `room.auth.get(userid, true)` to fall back to the default group.

server/chat-commands/room-settings.ts
```diff
@@ -1177,9 +1169,9 @@ export const commands: ChatCommands = {
 		this.roomlog(target);
 
 		delete room.introMessage;
-		if (room.chatRoomData) {
-			delete room.chatRoomData.introMessage;
-			Rooms.global.writeChatRoomData();
+		if (room.settings.persistSettings) {
+			delete room.settings.introMessage;
+			room.saveSettings();
 		}
```
Zarel, line 1175: `introMessage` should be in `settings`, so this can just be:

```ts
delete room.settings.introMessage;
room.saveSettings();
```

server/chat-commands/room-settings.ts
```diff
@@ -1310,11 +1301,13 @@ export const commands: ChatCommands = {
...
 			if (!this.runBroadcast()) return;
-			return this.sendReplyBox(`This room is currently displaying ${room.dataCommandTierDisplay} as the tier when using /data.`);
+			return this.sendReplyBox(
+				`This room is currently displaying ${room.settings.dataCommandTierDisplay} as the tier when using /data.`
+			);
 		}
 		if (!this.can('declare', null, room)) return false;
 
-		const displayIDToName: {[k: string]: typeof room.dataCommandTierDisplay} = {
+		const displayIDToName: {[k: string]: "tiers" | "doubles tiers" | "numbers"} = {
```
Zarel, line 1310:
```suggestion
		const displayIDToName: {[k: string]: typeof room.settings.dataCommandTierDisplay} = {
```

The `typeof` is important! It makes the code DRYer: DRY stands for Don't Repeat Yourself, and making the code DRY means that if we ever add more settings, we only have to change it in one place instead of multiple places.

server/chat-commands/room-settings.ts
```diff
@@ -1362,13 +1354,16 @@ export const roomSettings: SettingsHandler[] = [
...
 			[
 				'off',
 				'autoconfirmed',
 				'trusted',
 				...RANKS.slice(1, threshold! + 1),
-			].map(rank => [rank, (rank === 'off' ? !room.modchat : rank === room.modchat) || `modchat ${rank || 'off'}`]);
+			].map(
+				rank =>
+					[rank, (rank === 'off' ? !room.settings.modchat : rank === room.settings.modchat) || `modchat ${rank || 'off'}`]
+			);
```
Zarel, line 1366:
```suggestion
			].map(
				rank =>
					[rank, rank === (room.settings.modchat || 'off') || `modchat ${rank || 'off'}`]
			);
```

server/chat-commands/room-settings.ts
```diff
@@ -1386,54 +1381,57 @@ export const roomSettings: SettingsHandler[] = [
 			// first rank is for modjoin off
 			...RANKS.slice(1, room.isPersonal && !user.can('makeroom') ? 2 : undefined),
 		].map(
-			rank => [rank, (rank === 'off' ? !room.modjoin : rank === room.modjoin) || `modjoin ${rank || 'off'}`]
+			rank =>
+				[rank, (rank === 'off' ? !room.settings.modjoin : rank === room.settings.modjoin) || `modjoin ${rank || 'off'}`]
 		),
```
Zarel, line 1386: (Do what I did to modchat.)

server/chat-commands/room-settings.ts
```diff
@@ -1386,54 +1381,57 @@ export const roomSettings: SettingsHandler[] = [
...
 	}),
 	room => ({
 		label: "Language",
 		permission: 'editroom',
 		options: [...Chat.languages].map(
-			([id, name]) => [name, (id === 'english' ? !room.language : id === room.language) || `roomlanguage ${id}`]
+			([id, name]) =>
+				[name, (id === 'english' ? !room.settings.language : id === room.settings.language) || `roomlanguage ${id}`]
 		),
```
Zarel, line 1394: (Do what I did to modchat.)

server/chat-commands/room-settings.ts
```diff
@@ -1386,54 +1381,57 @@ export const roomSettings: SettingsHandler[] = [
...
 		],
 	}),
 	room => ({
 		label: "Slowchat",
 		permission: room.userCount < SLOWCHAT_USER_REQUIREMENT ? 'bypassall' : 'editroom',
 		options: ['off', 5, 10, 20, 30, 60].map(
-			time => [`${time}`, (time === 'off' ? !room.slowchat : time === room.slowchat) || `slowchat ${time || 'false'}`]
+			time => [`${time}`, (time === 'off' ?
+				!room.settings.slowchat : time === room.settings.slowchat) || `slowchat ${time || 'false'}`]
 		),
```
Zarel, line 1426: (Do what I did to modchat.)

server/chat-plugins/chat-monitor.ts
```diff
@@ -305,15 +305,15 @@ export const chatfilter: ChatFilter = function (message, user, room) {
 		.replace(/\u039d/g, 'e');
 	lcMessage = lcMessage.replace(/__|\*\*|``|\[\[|\]\]/g, '');
 
-	const isStaffRoom = room && ((room.chatRoomData && room.roomid.endsWith('staff')) || room.roomid.startsWith('help-'));
+	const isStaffRoom = room && ((room.settings && room.roomid.endsWith('staff')) || room.roomid.startsWith('help-'));
```
Zarel, line 308:
```suggestion
	const isStaffRoom = room && ((room.settings.persistSettings && room.roomid.endsWith('staff')) || room.roomid.startsWith('help-'));
```

What was your regex? You should be grepping for `\.settings[^.]` because it's nearly definitely a bug.

server/chat-plugins/chatlog.ts
```diff
@@ -76,15 +76,15 @@ const LogReader = new class {
 			const room = Rooms.get(roomid);
 			const forceShow = room && (
 				// you are authed in the room
-				(room.auth && user.id in room.auth && user.can('mute', null, room)) ||
+				(room.auth.get(user.id) && user.can('mute', null, room)) ||
```
Zarel, line 79:
```suggestion
				(room.auth.has(user.id) && user.can('mute', null, room)) ||
```

server/chat-plugins/daily-spotlight.ts
```diff
@@ -72,7 +72,7 @@ export const pages: PageTable = {
 
 export const commands: ChatCommands = {
 	removedaily(target, room, user) {
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 75:
```suggestion
		if (!room.settings.persistSettings) return this.errorReply("This command is unavailable in temporary rooms.");
```

...didn't you say you've fixed all these already?

server/chat-plugins/hangman.ts
```diff
@@ -268,26 +268,26 @@ export const commands: ChatCommands = {
...
 			}
-			room.hangmanDisabled = true;
-			if (room.chatRoomData) {
-				room.chatRoomData.hangmanDisabled = true;
-				Rooms.global.writeChatRoomData();
+			room.settings.hangmanDisabled = true;
+			if (room.settings.persistSettings) {
+				room.settings.hangmanDisabled = true;
+				room.saveSettings();
 			}
```
Zarel, line 278:
```suggestion
			room.settings.hangmanDisabled = true;
			room.saveSettings();
```
You have `room.settings.hangmanDisabled = true` in there twice...

server/chat-plugins/hangman.ts
```diff
@@ -268,26 +268,26 @@ export const commands: ChatCommands = {
...
-			delete room.hangmanDisabled;
-			if (room.chatRoomData) {
-				delete room.chatRoomData.hangmanDisabled;
-				Rooms.global.writeChatRoomData();
+			delete room.settings.hangmanDisabled;
+			if (room.settings.persistSettings) {
+				delete room.settings.hangmanDisabled;
+				room.saveSettings();
 			}
```
Zarel, line 291: .

server/chat-plugins/helptickets.ts
```diff
@@ -578,10 +578,7 @@ export const pages: PageTable = {
 					tickets[ticket.userid].open = false;
 					writeTickets();
 				} else {
-					if (!helpRoom.auth) {
-						helpRoom.auth = {};
-					}
-					if (!helpRoom.auth[user.id]) helpRoom.auth[user.id] = '+';
+					if (!helpRoom.auth.get(user.id)) helpRoom.auth.set(user.id, '+');
```
Zarel, line 581:
```suggestion
					if (!helpRoom.auth.has(user.id)) helpRoom.auth.set(user.id, '+');
```

server/chat-plugins/helptickets.ts
```diff
@@ -1095,10 +1092,7 @@ export const commands: ChatCommands = {
...
```
Zarel, line 1095: .

server/chat-plugins/jeopardy.ts
```diff
@@ -567,27 +567,27 @@ export const commands: ChatCommands = {
...
 			}
-			room.jeopardyDisabled = true;
-			if (room.chatRoomData) {
-				room.chatRoomData.jeopardyDisabled = true;
-				Rooms.global.writeChatRoomData();
+			room.settings.jeopardyDisabled = true;
+			if (room.settings.persistSettings) {
+				room.settings.jeopardyDisabled = true;
+				room.saveSettings();
 			}
```
Zarel, line 577: .

server/chat-plugins/jeopardy.ts
```diff
@@ -567,27 +567,27 @@ export const commands: ChatCommands = {
...
-			delete room.jeopardyDisabled;
-			if (room.chatRoomData) {
-				delete room.chatRoomData.jeopardyDisabled;
-				Rooms.global.writeChatRoomData();
+			delete room.settings.jeopardyDisabled;
+			if (room.settings.persistSettings) {
+				delete room.settings.jeopardyDisabled;
+				room.saveSettings();
 			}
```
Zarel, line 591: .

server/chat-plugins/mafia.ts
```diff
@@ -1596,7 +1596,7 @@ class MafiaTracker extends Rooms.RoomGame {
 
 		if (!player) {
 			if (user.isStaff ||
-				(this.room.auth && user.id in this.room.auth && this.room.auth[user.id] !== '+')
+				(this.room.auth.get(user.id) && this.room.auth.get(user.id) !== '+')
```
Zarel, line 1599: I would probably have an `auth.above('+')` function, for this which should be a relatively common use-case.

server/chat-plugins/mafia.ts
```diff
@@ -3682,13 +3682,13 @@ export const commands: ChatCommands = {
...
 			}
-			room.mafiaDisabled = false;
-			if (room.chatRoomData) {
-				room.chatRoomData.mafiaDisabled = false;
-				Rooms.global.writeChatRoomData();
+			room.settings.mafiaDisabled = false;
+			if (room.settings.persistSettings) {
+				room.settings.mafiaDisabled = false;
+				room.saveSettings();
 			}
```
Zarel, line 3692: .

server/chat-plugins/mafia.ts
```diff
@@ -3667,13 +3667,13 @@ export const commands: ChatCommands = {
...
-			room.mafiaDisabled = true;
-			if (room.chatRoomData) {
-				room.chatRoomData.mafiaDisabled = true;
-				Rooms.global.writeChatRoomData();
+			room.settings.mafiaDisabled = true;
+			if (room.settings.persistSettings) {
+				room.settings.mafiaDisabled = true;
+				room.saveSettings();
 			}
```
Zarel, line 3677: .

server/chat-plugins/room-events.ts
```diff
@@ -28,15 +28,14 @@ function formatEvent(
 }
 
 function getAllAliases(room: Room) {
-	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+	return Object.keys(room.settings.events).map(event => room.settings.events[event].aliases).join();
```
Zarel, line 31:
```suggestion
	return Object.values(room.settings.events).map(event => event.aliases).join();
```

server/chat-commands/core.ts
```diff
@@ -361,15 +361,13 @@ export const commands: ChatCommands = {
 		}
 		const rankLists: {[k: string]: string[]} = {};
 		const ranks = Object.keys(Config.groups);
-		for (const u in Users.usergroups) {
-			const rank = Users.usergroups[u].charAt(0);
-			if (rank === ' ' || (rank === '+' && !target)) continue;
+		for (const [k, v] of Users.groups.entries()) {
```
Zarel, line 364: .

server/chat-plugins/daily-spotlight.ts
```diff
@@ -102,7 +102,7 @@ export const commands: ChatCommands = {
 	},
 	queuedaily: 'setdaily',
 	async setdaily(target, room, user, connection, cmd) {
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 105: .

server/chat-plugins/daily-spotlight.ts
```diff
@@ -140,7 +140,7 @@ export const commands: ChatCommands = {
 		saveSpotlights();
 	},
 	async daily(target, room, user) {
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 143: .

server/chat-plugins/daily-spotlight.ts
```diff
@@ -157,7 +157,7 @@ export const commands: ChatCommands = {
 		room.update();
 	},
 	viewspotlights(target, room, user) {
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 160: .

server/chat-plugins/room-events.ts
```diff
@@ -28,15 +28,14 @@ function formatEvent(
...
-		for (const possibleEvent in room.events) {
-			if (room.events[possibleEvent].aliases) {
-				if (room.events[possibleEvent].aliases.includes(id)) {
+	if (!room.settings.events[id]) {
+		for (const possibleEvent in room.settings.events) {
+			if (room.settings.events[possibleEvent].aliases) {
+				if (room.settings.events[possibleEvent].aliases.includes(id)) {
 					id = toID(possibleEvent);
 				}
 			}
```
Zarel, line 41:
```suggestion
	if (!room.settings.events[id]) {
		for (const eventid of room.settings.events) {
			if (room.settings.events[eventid].aliases?.includes(id)) {
				return eventid as ID;
			}
```

server/chat-plugins/room-faqs.ts
```diff
@@ -30,7 +30,7 @@ function getAlias(roomid: RoomID, key: string) {
 export const commands: ChatCommands = {
 	addfaq(target, room, user, connection) {
 		if (!this.can('ban', null, room)) return false;
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 33: .

server/chat-plugins/room-faqs.ts
```diff
@@ -58,7 +58,7 @@ export const commands: ChatCommands = {
 	removefaq(target, room, user) {
 		if (!this.canTalk()) return this.errorReply("You cannot do this while unable to talk.");
 		if (!this.can('ban', null, room)) return false;
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 61: .

server/chat-plugins/room-faqs.ts
```diff
@@ -77,7 +77,7 @@ export const commands: ChatCommands = {
 	addalias(target, room, user) {
 		if (!this.canTalk()) return this.errorReply("You cannot do this while unable to talk.");
 		if (!this.can('ban', null, room)) return false;
-		if (!room.chatRoomData) return this.errorReply("This command is unavailable in temporary rooms.");
+		if (!room.settings) return this.errorReply("This command is unavailable in temporary rooms.");
```
Zarel, line 80: .

server/chat-plugins/scavengers.ts
```diff
@@ -869,9 +869,10 @@ export class ScavengerHunt extends Rooms.RoomGame {
...
 
 				// update the saved queue.
-				if (room.chatRoomData) {
-					room.chatRoomData.scavQueue = room.scavQueue;
-					Rooms.global.writeChatRoomData();
+				if (room.settings.persistSettings) {
+					// @ts-ignore
+					room.settings.scavQueue = room.scavQueue;
+					room.saveSettings();
 				}
```
Zarel, line 876: .

server/chat-plugins/scavengers.ts
```diff
@@ -1720,9 +1721,10 @@ const ScavengerCommands: ChatCommands = {
...
 		this.privateModAction(`(${user.name} has added a scavenger hunt to the queue.)`);
 
-		if (room.chatRoomData) {
-			room.chatRoomData.scavQueue = room.scavQueue;
-			Rooms.global.writeChatRoomData();
+		if (room.settings.persistSettings) {
+			// @ts-ignore
+			room.settings.scavQueue = room.scavQueue;
+			room.saveSettings();
 		}
```
Zarel, line 1728: .

server/chat-plugins/scavengers.ts
```diff
@@ -1740,9 +1742,10 @@ const ScavengerCommands: ChatCommands = {
...
 		this.sendReply(`|uhtmlchange|scav-queue|${formatQueue(room.scavQueue, user, room)}`);
 
-		if (room.chatRoomData) {
-			room.chatRoomData.scavQueue = room.scavQueue;
-			Rooms.global.writeChatRoomData();
+		if (room.settings.persistSettings) {
+			// @ts-ignore
+			room.settings.scavQueue = room.scavQueue;
+			room.saveSettings();
 		}
```
Zarel, line 1749: .

Zarel: (and so on for every other scav saving thing)

server/chat-plugins/scavengers.ts
```diff
@@ -1959,15 +1965,16 @@ const ScavengerCommands: ChatCommands = {
...
+			// @ts-ignore
+			room.settings.scavSettings = room.scavSettings;
+			room.saveSettings();
 		}
 		this.privateModAction(`(${user.name} has set the points awarded for blitz for ${gameType} hunts to ${blitzPoints}.)`);
 		this.modlog('SCAV BLITZ', null, `${gameType}: ${blitzPoints}`);
 
 		// double modnote in scavs room if it is a subroomgroupchat
-		if (room.parent && !room.chatRoomData && scavsRoom) {
+		if (room.parent && !room.settings && scavsRoom) {
```
Zarel, line 1977: .

Zarel:
Please un-draft this once it's ready.

server/chat-commands/info.ts
mia-pi-git: Typescript also won't allow this. 
```ts
	get(id: ID) {
		if (!super.has(id)) return '';
		return super.get(id);
	}````


```Property 'get' in type 'Auth' is not assignable to the same property in base type 'Map<ID, GroupSymbol>'.
  Type '(id: ID) => "" | "~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "!" | undefined' is not assignable to type '(key: ID) => "~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "!" | undefined'.
    Type '"" | "~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "!" | undefined' is not assignable to type '"~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "!" | undefined'.
      Type '""' is not assignable to type '"~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "!" | undefined'.```

Comment (mia-pi-git) [reply server/chat-commands/info.ts:46]: I can change it to `get(...)!` though.

File: server/chat-commands/moderation.ts [hunk 80]
```diff
@@ -959,7 +953,7 @@ export const commands: ChatCommands = {
 
 	deroomvoiceall(target, room, user) {
 		if (!this.can('editroom', null, room)) return false;
-		if (!room.auth) return this.errorReply("Room does not have roomauth.");
+		if (![...room.auth.entries()].length) return this.errorReply("Room does not have roomauth.");
```
Zarel, line 956:
```suggestion
		if (!room.auth.size) return this.errorReply("Room does not have roomauth.");
```

lol, `[...map.entries()].length` is certainly one way to write `map.size`, but I do want to point out that `map.size` does exist:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size

server/chat-commands/info.ts
```diff
@@ -2110,10 +2112,16 @@ export const commands: ChatCommands = {
...
 				// Limited support for translated analysis
 				// Translated analysis do not support automatic redirects from a id to the proper page
-				this.sendReplyBox(Chat.html`<a href="https://www.smogon.com/dex/${generation}/pokemon/${id}/${formatId}/?lang=${supportedLanguages[room.language]}">${generation.toUpperCase()} ${formatName} ${pokemon.name} analysis</a>, brought to you by <a href="https://www.smogon.com">Smogon University</a>`);
+				this.sendReplyBox(
+					Chat.html`
+					<a href="https://www.smogon.com/dex/${generation}/pokemon/${id}/${formatId}/?lang=
+					${supportedLanguages[room.settings.language]}">
+					${generation.toUpperCase()} ${formatName} ${pokemon.name} analysis</a>,
+					brought to you by <a href="https://www.smogon.com">Smogon University</a>`
+				);
```
Zarel, line 2124: Um, does `Chat.html` filter out newlines? If not, this might not work the way it was intended. Either way, we're generally not fond of multiline strings here at PS.

```suggestion
				this.sendReplyBox(
					Chat.html`<a href="https://www.smogon.com/dex/${generation}/pokemon/${id}/${formatId}/?lang=${supportedLanguages[room.language]}">${generation.toUpperCase()} ${formatName} ${pokemon.name} analysis</a>, brought to you by <a href="https://www.smogon.com">Smogon University</a>`
				);
```

server/chat-commands/room-settings.ts
```diff
@@ -50,13 +50,13 @@ export const commands: ChatCommands = {
 			output += `<br />`;
 		}
 		output += '</div>';
-		user.sendTo(room, `|${uhtml}|roomsettings|${output}`);
+		user.sendTo(room, `|${uhtml}|room.settings|${output}`);
```
Zarel, line 53: .

Zarel:
```suggestion
		user.sendTo(room, `|${uhtml}|roomsettings|${output}`);
```

server/chat-commands/room-settings.ts
```diff
@@ -103,25 +106,24 @@ export const commands: ChatCommands = {
...
+		if ((room as GameRoom).requestModchat && !room.settings.modchat) (room as GameRoom).requestModchat(null);
+		this.privateModAction(`(${user.name} set modchat to ${room.settings.modchat || "off"})`);
+		this.modlog('MODCHAT', null, `to ${room.settings.modchat || "false"}`);
 
-		if (room.chatRoomData) {
-			room.chatRoomData.modchat = room.modchat;
-			Rooms.global.writeChatRoomData();
+		if (room.settings.persistSettings) {
+			room.saveSettings();
 		}
```
Zarel, line 127:
```suggestion
		room.saveSettings();
```

server/chat-commands/room-settings.ts
```diff
@@ -243,18 +244,17 @@ export const commands: ChatCommands = {
...
 
-		room.language = targetLanguage === 'english' ? false : targetLanguage;
+		room.settings.language = targetLanguage === 'english' ? false : targetLanguage;
 
-		if (room.chatRoomData) {
-			room.chatRoomData.language = room.language;
-			Rooms.global.writeChatRoomData();
+		if (room.settings.persistSettings) {
+			room.saveSettings();
 		}
```
Zarel, line 258:
```suggestion
		room.saveSettings();
```

server/chat-plugins/scavengers.ts
```diff
@@ -2205,9 +2196,9 @@ const ScavengerCommands: ChatCommands = {
 			formattedData.map(entry => {
 				const userid = toID(entry.name);
 
-				const auth = room.auth && room.auth[userid] ? room.auth[userid] :
-					Users.usergroups[userid] ? Users.usergroups[userid].charAt(0) : '&nbsp;';
-				const color = room.auth && userid in room.auth ? 'inherit' : 'gray';
+				const auth = room.auth.get(user.id) ? room.auth.get(user.id) :
+					Users.groups.get(userid).trim() ? Users.groups.get(userid) : '&nbsp;';
+				const color = room.auth.get(user.id) ? 'inherit' : 'gray';
```
Zarel, line 2201: lol, I knew this was you and not sparky.

You changed this from `userid` to `user.id`, so all the table rows would be colored like the user who used the command, rather than the user whose row it was about.

Zarel:
I'm still not fully satisfied with this API: `get(id)` vs `get(id, true)` vs `getDirect(id)`, but I think it's a step in the right direction.

Zarel:
My latest update does several things:

1. Fixes scav settings to work as specified (only have one `scavSettings`, rather than tracking `room.scavSettings` and `room.settings.scavSettings`)

2. Makes every attribute in `room.settings` optional (except title and auth), so that it correctly matches `chatrooms.json`

3. Fixes everything still using `room.settings` so they support things being optional

4. Renames `room.settings.persistSettings` to `room.persist`, because it shouldn't be saved in `chatrooms.json`

5. Standardizes how `room.gameNumber` works (it was previously very inconsistent in terms of was it expected to exist, was it incremented before/after, was it cached, etc)

6. Fixes a bug where `room.auth.load()` was called before auth was loaded, rather than after.

Zarel:
Okay, at this point, I think this is actually ready.

Zarel:
I regret using "it'll require a restart anyway" as a reason to put both changes in one PR, but we're sadly too far in to back out of that.

PR #6779, opened 2020-05-31
Fix misspelled instances of facade in learnsets

KrisXV:
Can whoever merges this also hotpatch formats and rebuild the client upon merge? The CAP Pokemon learnsets with `faade` are crashing.


Zarel:
Looks like we should be validating learnsets...

PR #6780, opened 2020-05-31
Improve UNO error message

arifin6686:


server/chat-plugins/uno.ts
```diff
@@ -763,7 +763,7 @@ export const commands: ChatCommands = {
 			const game = room.getGame(UnoGame);
 			if (!game) return this.errorReply("There is no UNO game going on in this room right now.");
 			if (!this.canTalk()) return false;
-			if (!game.joinGame(user)) return this.errorReply("Unable to join the game.");
+			if (!game.joinGame(user)) return this.errorReply("You have already joined the game of UNO.");
```
Zarel, line 766: Is this the only case `joinGame` can fail?

SSlayer9: I don't believe so, no.

Zarel:
If you want to do this, the error messages will need to be fixed so it only gives the "You have already joined the game of UNO" message when you've actually already joined, and not for other join errors. Once you've done that, click "Ready for review" to un-draft this PR.

PR #6788, opened 2020-06-02
Update Gourgeist / Type: Null Doubles Sets

PartMan7:
(i) Gourgeist can't learn Shadow Claw.
(ii) I was also asked by A Cake Wearing A Hat to remove Type: Null's entry from Doubles.


Zarel:
I assume this is @TheImmortal's job to review? But correct me if I'm wrong.

PR #6789, opened 2020-06-02
Give 2v2 Doubles a ladder

KrisXV:


config/formats.ts
```diff
@@ -468,7 +468,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'gen8',
 		gameType: 'doubles',
-		searchShow: false,
+		// searchShow: false,
```
Zarel, line 471: Why comment out rather than remove?

KrisXV: It will become a challenge-only format again when the month is over.

PR #6790, opened 2020-06-02
Hidden Ability Starters

PartMan7:
Libero is gonna be fun.

Edit: source: https://twitter.com/SerebiiNet/status/1267810302713671680?s=19
 (Thanks, xJoelituh)


DaWoblefet:
Beat me to it! Verifying that they are Battle Stadium legal as well, so VGC etc. can use them.

PR #6791, opened 2020-06-02
Remove event aliases if none exist.

SSlayer9:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8488586 Might not work but it's worth a shot.

I literally just looked at all the html bits and edited it, i still don't know how events are stored or anything, please feel free to edit if I'm being stupid, UwU.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2463,7 +2463,6 @@ export const commands: ChatCommands = {
 	},
 	showimagehelp: [`/showimage [url], [width], [height] - Show an image. Any CSS units may be used for the width or height (default: px). If width and height aren't provided, automatically scale the image to fit in chat. Requires: # & ~`],
 
-	'!pi': true,
 	pi(target, room, user) {
 		return this.sendReplyBox(
 			'Did you mean: 1. 3.1415926535897932384626... (Decimal)<br />' +
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2462,8 +2462,8 @@ export const commands: ChatCommands = {
 		});
 	},
 	showimagehelp: [`/showimage [url], [width], [height] - Show an image. Any CSS units may be used for the width or height (default: px). If width and height aren't provided, automatically scale the image to fit in chat. Requires: # & ~`],
-	
-	"!pi": true,
+
+	'!pi': true,
 	pi(target, room, user) {
 		return this.sendReplyBox(
 			'Did you mean: 1. 3.1415926535897932384626... (Decimal)<br />' +
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2462,7 +2462,8 @@ export const commands: ChatCommands = {
 		});
 	},
 	showimagehelp: [`/showimage [url], [width], [height] - Show an image. Any CSS units may be used for the width or height (default: px). If width and height aren't provided, automatically scale the image to fit in chat. Requires: # & ~`],
-
+	
+	"!pi": true,
 	pi(target, room, user) {
 		return this.sendReplyBox(
 			'Did you mean: 1. 3.1415926535897932384626... (Decimal)<br />' +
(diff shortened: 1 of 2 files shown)


server/chat-plugins/room-events.ts
```diff
@@ -18,7 +18,7 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
 	}
 	let ret = `<tr title="${explanation}">`;
 	ret += Chat.html`<td>${event.eventName}</td>`;
-	ret += Chat.html`<td>${event.aliases ? event.aliases.join(", ") : ""}</td>`;
+	ret += event.aliases ? Chat.html`<td>${event.aliases.join(", ")}</td>` : ""
```
AnnikaCodes, line 21:
```suggestion
	ret += event.aliases.length ? Chat.html`<td>${event.aliases.join(", ")}</td>` : "";
```

AnnikaCodes: Also, you need to figure out a way to include empty `<td>` tags when displaying all events if one event has alias(es) and others do not - I'll leave that for you to figure out (maybe pass an argument to `formatEvent()`?). Currently, your code doesn't work when showing all aliases.

server/chat-commands/info.ts
```diff
@@ -2463,7 +2463,6 @@ export const commands: ChatCommands = {
 	},
 	showimagehelp: [`/showimage [url], [width], [height] - Show an image. Any CSS units may be used for the width or height (default: px). If width and height aren't provided, automatically scale the image to fit in chat. Requires: # & ~`],
 
-	'!pi': true,
```
AnnikaCodes, line 2466: This should be in a separate pull request. `!pi` has nothing to do with room events.

server/chat-commands/info.ts
```diff
@@ -2462,8 +2462,8 @@ export const commands: ChatCommands = {
 		});
 	},
 	showimagehelp: [`/showimage [url], [width], [height] - Show an image. Any CSS units may be used for the width or height (default: px). If width and height aren't provided, automatically scale the image to fit in chat. Requires: # & ~`],
-
-	'!pi': true,
+	
+	"!pi": true,
```
AnnikaCodes, line 2466:
```suggestion

	'!pi': true,
```

server/chat-plugins/room-events.ts
```diff
@@ -53,7 +53,7 @@ export const commands: ChatCommands = {
 			}
 			if (!this.runBroadcast()) return;
 			let buff = '<table border="1" cellspacing="0" cellpadding="3">';
-			buff += '<th>Event Name:</th><th>Event Aliases:</th><th>Event Description:</th><th>Event Date:</th>';
+			buff += '<th>Event Name:</th>' + getAllAliases(room).length ? '<th>Event Aliases:</th>' : "" + '<th>Event Description:</th><th>Event Date:</th>';
```
Zarel, line 56:
```suggestion
			buff += `<th>Event Name:</th>${getAllAliases(room).length ? `<th>Event Aliases:</th>` : ``}<th>Event Description:</th><th>Event Date:</th>`;
```

server/chat-plugins/room-events.ts
```diff
@@ -18,7 +18,7 @@ function formatEvent(event: {eventName: string, date: string, desc: string, star
 	}
 	let ret = `<tr title="${explanation}">`;
 	ret += Chat.html`<td>${event.eventName}</td>`;
-	ret += Chat.html`<td>${event.aliases ? event.aliases.join(", ") : ""}</td>`;
+	ret += event.aliases.length ? Chat.html`<td>${event.aliases.join(", ")}</td>` : "";
```
Zarel, line 21:
```suggestion
	ret += event.aliases.length ? Chat.html`<td>${event.aliases.join(", ")}</td>` : ``;
```

Zarel:
Thanks!

PR #6793, opened 2020-06-02
Datasearch: Support gen and maxgen flags in /is and /as

AnnikaCodes:
Also the usual conversion of help messages to HTML reply boxes.


AnnikaCodes:
If it's ok for me to hijack this with a bug fix for /is, I'll do so (commit b3e45ec), but if not let me know and I can make that its own PR.

PR #6794, opened 2020-06-03
Scavengers: new game mode

sparkychildcharlie:
Add Team Scavs as well as the necessary supporting commands.

This has Meicoo-chan's seal of approval!

kawaiiiii

### DIFF
--- a/server/chat-plugins/scavenger-games.ts
+++ b/server/chat-plugins/scavenger-games.ts
@@ -25,9 +25,9 @@ interface GameMode {
 	mod: Twist;
 	round?: number;
 	leaderboard?: true;
-	teamAnnounce?: GameModeFunction,
-	getPlayerTeam?: GameModeFunction,
-	advanceTeam?: GameModeFunction,
+	teamAnnounce?: GameModeFunction;
+	getPlayerTeam?: GameModeFunction;
+	advanceTeam?: GameModeFunction;
 	[k: string]: any;
 }
 
@@ -346,6 +346,7 @@ const MODES: {[k: string]: GameMode | string} = {
 		playerlist: null,
 	},
 
+	pr: 'pointrally',
 	pointrally: {
 		name: 'Point Rally',
 		id: 'pointrally',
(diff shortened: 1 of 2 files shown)


server/chat-plugins/scavenger-games.ts
```diff
@@ -25,9 +25,17 @@ interface GameMode {
...
+	teamAnnounce?: GameModeFunction;
+	getPlayerTeam?: GameModeFunction;
+	advanceTeam?: GameModeFunction;
 	[k: string]: any;
 }
 
+type GameModeFunction = (
+	this: ScavengerGameTemplate,
+	...args: any[]
+) => void;
```
KrisXV, line 37:
```suggestion
type GameModeFunction = (this: ScavengerGameTemplate, ...args: any[]) => void;
```

server/chat-plugins/scavengers.ts
```diff
@@ -663,7 +686,10 @@ export class ScavengerHunt extends Rooms.RoomGame {
 					}</details>` :
 					``)
 			}</td>` +
-			`</tr></table></div>`
+			`</tr></table></div>`;
+
+		player.sendRoom(
+			questionDisplay
 		);
```
KrisXV, line 693:
```suggestion
		player.sendRoom(questionDisplay);
```

server/chat-plugins/scavengers.ts
```diff
@@ -652,8 +676,7 @@ export class ScavengerHunt extends Rooms.RoomGame {
 
 		const finalHint = current.number === this.questions.length ? "final " : "";
 
-		player.sendRoom(
-			`|raw|<div class="ladder"><table><tr>` +
+		const questionDisplay = `|raw|<div class="ladder"><table><tr>` +
```
KrisXV, line 679: There's a ton of unnecessary new lines in this

sparkychildcharlie: @Zarel wrote this

KrisXV: Yes, but that doesn't mean the new lines are necessary? It's a variable that's separately declared now so the line length shouldn't affect it I'm pretty sure.

KrisXV: The new lines were necessary when it was still wrapped in the function.

Zarel: I definitely prefer the newlines for readability.

Zarel: I don't like the addition of the `questionDisplay` variable, though. What's the point?

PR #6795, opened 2020-06-03
Fix the remaining learnsets

KrisXV:


data/learnsets.ts
```diff
@@ -63711,7 +63717,18 @@ export const BattleLearnsets: {[speciesid: string]: LearnsetData} = {
 		eventOnly: true,
 	},
 	zygarde10: {
-		// FIXME: figure out what's going on here
+		learnset: {
```
Zarel, line 63720: I think Zygarde 10% events should be considered Zygarde events, because they can switch freely, so Zygarde has access to these moves.

Zarel: I guess I'm okay with leaving this here if it doesn't cause any relevant legality problems, for other non-sim use-cases.

Zarel:
Thanks!

data/learnsets.ts
KrisXV: zygarde-10 gets zygarde-50 as a learnset parent, so it gets all of its moves anyways

Zarel: Yes, but Zygarde-50 doesn't get all of Zygarde-10's moves.

PR #6796, opened 2020-06-03
Make /pi command Broadcastable

DeltaCoderr:


server/chat-commands/info.ts
```diff
@@ -2465,6 +2465,7 @@ export const commands: ChatCommands = {
 
 	'!pi': true,
 	pi(target, room, user) {
+		if (!this.runBroadcast(true)) return false;
```
xfix, line 2468: Why is it ignoring cooldown?

DeltaCoderr: Oh, sorry forgot about that.

server/chat-commands/info.ts
```diff
@@ -2465,6 +2465,7 @@ export const commands: ChatCommands = {
 
 	'!pi': true,
 	pi(target, room, user) {
+		if (!this.runBroadcast()) return false;
```
xfix, line 2468: I'm honestly not sure, should commands return `false` or not? This is a question to @Zarel.

Zarel: It's complicated. I don't mind them returning `false`.

PR #6797, opened 2020-06-03
SSB: Batch Update 

DeltaCoderr:


Zarel:
Okay, as usual, this will be a batch update. Put all the rank updates in here, and we'll merge this in a month.

Zarel:
I'd normally wait a month, but the admin rank changes is probably a good time for an exception. Thanks for doing this!

PR #6798, opened 2020-06-03
Roomevents: Fix bug with not displaying aliases

AnnikaCodes:
Fixes the bug seen in the screenshot, where aliases are displayed for some events but not others, and thus descriptions are shown as aliases (and dates as descriptions).
Skärmavbild 2020-06-03 kl  9 14 15 fm


Zarel:
Thanks! I didn't think of this problem when I was reviewing the other PR>

PR #6799, opened 2020-06-03
Back out erroneous change from 1fa9e14 to fix After You regression

urkerab:
Commmit 1fa9e14 erroneously added the source effect to all the calls to `prioritizeAction`, not just those for the Pledge moves and Round. As well as breaking After You, this also confused me when I created PR #6231 and I had to fix that separately in PR #6622.

No, I don't have a test handy.


Zarel:
Fair enough.

PR #6804, opened 2020-06-05
Support specifying a reason for /hidetext

AnnikaCodes:


server/chat-commands/moderation.ts
```diff
@@ -1459,16 +1459,32 @@ export const commands: ChatCommands = {
...
+	hlines: 'hidetext',
 	hidetext(target, room, user, connection, cmd) {
 		if (!target) return this.parse(`/help hidetext`);
 
-		const lineCount = parseInt(this.splitTarget(target)) || 0;
+		const isPartial = cmd.includes('lines');
+		const showAlts = cmd.includes('alt');
+		const lineCount = isPartial ? parseInt(this.splitTarget(target)) || 0 : 0;
+		const reason = lineCount ? target.split(',').slice(2).join(',').trim() : this.splitTarget(target);
+		if (parseInt(reason) && !cmd.includes('force')) {
```
Zarel, line 1472:
```suggestion
		if (parseInt(reason) && !cmd.includes('force')) {
```

Zarel:
It's a common misconception that `parseInt` will only return a non-`NaN` if the input is an integer. Unfortunately, what it actually does is that it returns the first integer in the text, so `parseInt("123.45abc 678")` is `123`.

server/chat-commands/moderation.ts
```diff
@@ -1459,16 +1459,32 @@ export const commands: ChatCommands = {
...
+			return this.errorReply(`Your reason was a number; use /forcehidetext if you didn't mean to clear ${reason} messages.`);
+		}
+		if (!lineCount && isPartial) {
+			return this.errorReply(`You must specify a number of messages to clear. To clear all messages, use /hidetext.`);
+		}
```
Zarel, line 1477:
```suggestion
		const hasLineCount = cmd.includes('lines');
		target = this.splitTarget(target);
		let lineCount = 0;
		if (/^[0-9]+\s*(,|$)/.test(target)) {
			if (hasLineCount) {
				let lineCountString;
				[lineCountString, target] = Chat.splitFirst(target, ',');
				lineCount = parseInt(lineCountString);
			} else if (!cmd.includes('force')) {
				return this.errorReply(`Your reason was a number; use /hidelines if you wanted to clear a specific number of lines, or /forcehidelines if you really wanted your reason to be a number.`);
			}
		}
		const showAlts = cmd.includes('alt');
		const reason = target.trim();
		if (!lineCount && hasLineCount) {
			return this.errorReply(`You must specify a number of messages to clear. To clear all messages, use /hidetext.`);
		}
```

AnnikaCodes:
> It's a common misconception that `parseInt` will only return a non-`NaN` if the input is an integer. Unfortunately, what it actually does is that it returns the first integer in the text, so `parseInt("123.45abc 678")` is `123`.

That's odd, but good to know.

PR #6805, opened 2020-06-05
Roomevents: Fix bug in getAllAliases()

AnnikaCodes:


server/chat-plugins/room-events.ts
```diff
@@ -33,7 +33,11 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
 }
 
 function getAllAliases(room: Room) {
-	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+	let aliases: string[] = [];
+	for (const aliasList of Object.keys(room.events).map(event => room.events[event].aliases)) {
+		if (aliasList) aliases = aliases.concat(aliasList);
+	}
```
Zarel, line 39: I feel like this is the point at which `Object.keys`/`map` shenanigans just makes it less readable.

```suggestion
	for (const eventid in room.events) {
		const event = room.events[eventid];
		if (event.aliases) aliases.push(...event.aliases);
	}
```

server/chat-plugins/room-events.ts
```diff
@@ -33,7 +33,11 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
 }
 
 function getAllAliases(room: Room) {
-	return Object.keys(room.events).map(event => room.events[event].aliases).join();
+	let aliases: string[] = [];
```
AnnikaCodes, line 36:
```suggestion
	const aliases: string[] = [];
```

PR #6808, opened 2020-06-06
Support Crit in Movesearch

PartMan7:
A lot of people had been asking about whether it was possible to search for moves with a high critical-hit ratio (relevant for building with things like Sniper or Focus Energy).

Note: This does not include directly searching for moves that have a critRatio of 3 (only one move has that) or 4 (not many have those - if you guys feel I should support searching for crit levels like priority does, I'll add that in.)

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -1214,7 +1214,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			if (target === 'crit' || toID(target) === 'highcrit') target = 'highcrit';
 			if (allFlags.includes(target)) {
 				if ((orGroup.flags[target] && isNotSearch) || (orGroup.flags[target] === false && !isNotSearch)) {
-					return {error: 'A search cannot both exclude and include \'' + target + '\'.'};
+					return {error: `A search cannot both exclude and include '${target}'.`};
 				}
 				orGroup.flags[target] = !isNotSearch;
 				continue;
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -1211,7 +1211,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			if (target === 'z') target = 'zmove';
 			if (target === 'max') target = 'maxmove';
 			if (target === 'gmax') target = 'gmaxmove';
-      if (target === 'crit' || toID(target) === 'highcrit') target = 'highcrit';
+			if (target === 'crit' || toID(target) === 'highcrit') target = 'highcrit';
 			if (allFlags.includes(target)) {
 				if ((orGroup.flags[target] && isNotSearch) || (orGroup.flags[target] === false && !isNotSearch)) {
 					return {error: 'A search cannot both exclude and include \'' + target + '\'.'};
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -1578,7 +1578,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 
 			for (const flag in alts.flags) {
 				if (flag === 'secondary') {
-					if ((!move.secondary && !move.secondaries) === !alts.flags[flag]) {
+					if (!(move.secondary || move.secondaries) === !alts.flags[flag]) {
 						matched = true;
 						break;
 					}
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -1100,7 +1100,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 	const allContestTypes = ['beautiful', 'clever', 'cool', 'cute', 'tough'];
 	const allProperties = ['basePower', 'accuracy', 'priority', 'pp'];
 	const allFlags = [
-		'authentic', 'bite', 'bullet', 'charge', 'contact', 'crit', 'dance', 'defrost', 'gravity', 'mirror',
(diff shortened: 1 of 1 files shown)


Zarel:
Just calling it "crit" seems weird to me, what exactly is it searching? Moves that always crit, like Frost Breath?

PartMan7:
Moves with an increased critical-hit rate, like Night Slash or Leaf Blade.

Zarel:
I'd call it "high crit rate", then.

But I think if you do, it should search any move with a higher crit rate than normal. I don't really like the arbitrary cutoff thing.

PartMan7:
I'd toyed with `high crit rate`, but it seems unnecessarily verbose - given how people literally use G7 instead of Gen7, `crit` sounded both short and apparent enough.

Also, the arbitrary cutoff is for all moves - all moves, by default, have a critRatio of 1.

Zarel:
Alias it to `hcrit` or something, I guess. I just want simple words to mean what they mean.

Zarel:
> Also, the arbitrary cutoff is for all moves - all moves, by default, have a critRatio of 1.

I don't know what that means, but I was responding to this:

> Note: This does not include moves that have a critRatio of 3 (only one move has that) or 4 (not many have those - if you guys feel I should support searching for crit levels like priority does, I'll add that in.)

Looking at your actual code, it looks like it does include all moves with high crit ratio as expected. Was this just a typo?

PartMan7:
Oh, I mis-framed my sentence.

My code doesn't allow explicitly searching for moves with a specific critical-hit ratio (so you can't directly search for all moves that are guaranteed to crit).

Zarel:
Hm. I think I'm fine calling this "crit" if it includes Frost Breath etc.

PartMan7:
Oh, just noticed the reply. Should I revert that?

Zarel:
Up to you.

PartMan7:
I think this implementation works, then.

(Also, any idea why this is crashing? The code doesn't deviate from the others above / below.)

Zarel:
Your problem is that TypeScript can't handle type narrowing three levels deep.

`typeof dex[move].isMax === 'boolean' && dex[move].isMax` doesn't crash only because `&& undefined` is a valid check, unlike `&& undefined > 1`.

I would cache `dex[move]`; it doesn't need to be that deep, anyway.

Zarel:
(I'll do that for you.)

Zarel:
You still need to include Frost Breath and Storm Throw.

server/chat-plugins/datasearch.ts
```diff
@@ -1549,81 +1550,90 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
...
-					if (!dex[move].isZ === !alts.flags[flag]) {
+					if (!move.isZ === !alts.flags[flag]) {
+						matched = true;
+						break;
+					}
+				} else if (flag === 'highcrit') {
+					if (
+						!((typeof move.critRatio === 'number' && move.critRatio > 1) ||
+						  (typeof move.willCrit === 'boolean' && move.willCrit)
+						 ) === !alts.flags[flag]) {
```
Zarel, line 1594:
```suggestion
					const crit = move.willCrit || (move.critRatio && move.critRatio > 1);
					if (!crit === !alts.flags[flag]) {
```

Zarel: The others only test `typeof` because they're necessary to distinguish max moves from g-max moves. You don't actually need it here.

server/chat-plugins/datasearch.ts
```diff
@@ -1549,81 +1550,90 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
...
-				if (alts.targets[dex[move].target]) continue;
-				if (Object.values(alts.targets).includes(false) && alts.targets[dex[move].target] !== false) continue;
+				if (alts.targets[move.target]) continue;
+				if (Object.values(alts.targets).includes(false) && alts.targets[move.target] !== false) continue;
 			}
 
 			for (const flag in alts.flags) {
 				if (flag === 'secondary') {
-					if ((!dex[move].secondary && !dex[move].secondaries) === !alts.flags[flag]) {
+					if ((!move.secondary && !move.secondaries) === !alts.flags[flag]) {
```
Zarel, line 1581:
```suggestion
					if (!(move.secondary || move.secondaries) === !alts.flags[flag]) {
```
Might as well refactor this while we're changing everything else.

server/chat-plugins/datasearch.ts
```diff
@@ -1211,6 +1211,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			if (target === 'z') target = 'zmove';
 			if (target === 'max') target = 'maxmove';
 			if (target === 'gmax') target = 'gmaxmove';
+			if (target === 'crit' || toID(target) === 'highcrit') target = 'highcrit';
 			if (allFlags.includes(target)) {
 				if ((orGroup.flags[target] && isNotSearch) || (orGroup.flags[target] === false && !isNotSearch)) {
 					return {error: 'A search cannot both exclude and include \'' + target + '\'.'};
```
AnnikaCodes, line 1217:
```suggestion
					return {error: `A search cannot both exclude and include '${target}'.`};
```
While you're refactoring stuff, this is more readable imo.

server/chat-plugins/datasearch.ts
Zarel: @PartMan7 I think you missed this?

PartMan7: Oh, I did, sorry. 'o.o 
Apologies for the unnecessary hassle.

PartMan7:
Wait, misclick.

PR #6809, opened 2020-06-07
Tier shift: Ban Eviolite

sparkychildcharlie:
as per https://www.smogon.com/forums/threads/tier-shift-om-of-the-month.3662165/page-2#post-8490625

<sub>(salty i lost like 5 minutes on a scavenger hunt that asked about the banlists but i referred to the server code instead of smogon thread)</sub>


Zarel:
Thanks!

PR #6810, opened 2020-06-07
Gen 7 Random Doubles Battle Updates

ACakeWearingAHat:
A Big'Un. To do: reject Kommo-o altogether in Gen 7 Random Doubles if teamdetails zmove.


data/mods/gen7/random-teams.ts
```diff
@@ -200,15 +201,23 @@ export class RandomGen7Teams extends RandomTeams {
...
 					if (counter.setupType || !!counter['speedsetup'] || hasMove['dragontail']) rejected = true;
+					if (isDoubles && movePool.includes('protect') || movePool.includes('spikyshield')) rejected = true;
 					break;
 				case 'nightshade': case 'seismictoss': case 'superfang':
 					if (!hasAbility['Parental Bond'] && (counter.damagingMoves.length > 1 || counter.setupType)) rejected = true;
 					break;
 				case 'protect':
 					if (counter.setupType && !hasMove['wish']) rejected = true;
 					if (hasMove['rest'] || hasMove['lightscreen'] && hasMove['reflect']) rejected = true;
+					if (isDoubles && (hasMove['tailwind'] && hasMove['roost']) || hasMove['fakeout'] || movePool.includes('bellydrum') || movePool.includes('shellsmash')) rejected = true;
```
TheImmortal, line 220: What? If you're rejecting Protect basically all the time then just remove it from the move pools...

ACakeWearingAHat: we still want protect; we just don't want mons to not have Belly Drum/Shell Smash when they otherwise could, and we want to avoid Roost + Tailwind + Protect specifically

PR #6811, opened 2020-06-07
Fix Custom Rules Error

PartMan7:
`/challenge [Gen 8] OU @@@ +Reshiram` failed earlier, since the rule wasn't trimmed.

Also going to add the fix for chat only including the first rule.

### DIFF
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -581,7 +581,7 @@ export class ModdedDex {
 		if (!customRulesString) return format.id;
 		const ruleTable = this.getRuleTable(format);
 		const customRules = customRulesString.split(',').map(rule => {
-			const ruleSpec = this.validateRule(rule);
+			const ruleSpec = this.validateRule(rule.trim());
 			if (typeof ruleSpec === 'string' && ruleTable.has(ruleSpec)) return null;
 			return rule.replace(/[\r\n|]*/g, '').trim();
 		}).filter(rule => rule);
--- a/sim/dex.ts
+++ b/sim/dex.ts
@@ -581,7 +581,7 @@ export class ModdedDex {
 		if (!customRulesString) return format.id;
 		const ruleTable = this.getRuleTable(format);
 		const customRules = customRulesString.split(',').map(rule => {
-			const ruleSpec = this.validateRule(rule.trim());
+			const ruleSpec = this.validateRule(rule);
 			if (typeof ruleSpec === 'string' && ruleTable.has(ruleSpec)) return null;
 			return rule.replace(/[\r\n|]*/g, '').trim();
 		}).filter(rule => rule);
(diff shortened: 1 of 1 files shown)


PartMan7:
Okay, so the issue is that this.splitTarget disregards all arguments past the second comma, since the split cap returns `[user, format @@@ firstRule]` instead of `[user, format @@@ firstRule, nextRule, extraRuleForNoReason]`. I can try editing it to use all rules, but this.splitTarget is a widely used function - is there anything that changing it might break?

sim/dex.ts
```diff
@@ -581,7 +581,7 @@ export class ModdedDex {
 		if (!customRulesString) return format.id;
 		const ruleTable = this.getRuleTable(format);
 		const customRules = customRulesString.split(',').map(rule => {
-			const ruleSpec = this.validateRule(rule);
+			const ruleSpec = this.validateRule(rule.trim());
```
Zarel, line 584: Maybe this should be inside `validateRule`?

Zarel:
We should probably split by `@@@` before using `splitTarget`.

Zarel:
Also, `splitTarget` shouldn't matter as long as the first `,` is guaranteed to be before the first `@@@`, because `splitTarget` only splits on the first `,` instead of all of them.

Zarel:
I've been wanting to fix this but haven't had time, so thanks for tackling this, by the way!

PartMan7:
Oh, didn't notice the further comments. I'll get on them.

sim/dex.ts
```diff
@@ -953,6 +953,7 @@ export class ModdedDex {
 		switch (rule.charAt(0)) {
 		case '-':
 		case '+':
+			rule = rule.trim();
```
Zarel, line 956:
```suggestion
		rule = rule.trim();
		switch (rule.charAt(0)) {
		case '-':
		case '+':
```

(This is necessary if the rule starts with a space.)

Zarel:
Apparently this entire thing is kind of convoluted. I might need to take a look myself to figure out exactly which part is missing trimming.

PartMan7:
When you say `part`, which are you talking about? I thought the rules were the only things missing trimming. 'o.o

Zarel:
The validation code is pretty complicated, and I would need to look at every level to figure out which step the trimming should be happening in.

Zarel:
It's plausible to me that `validateRule` should simply fail if the input is untrimmed; the code that calls it looks like it would bug out with untrimmed input anyway.

KrisXV:
I would like to bump this because this issue is still affecting `/chall`

KrisXV:
bumping again @PartMan7 @Zarel

PartMan7:
I thought Zarel wanted to suggest some changes; should I mark this as ready?

PartMan7:
@Zarel Hiya?

Zarel:
Oh sorry I'll prioritize this.

PartMan7:
At this point, I'm setting it as Ready for Review since there's been no comments at all.

Zarel:
I'm still working on this, although I sympathize with your impatience. Rest assured this is high priority to me.

Zarel:
Okay, this is definitely a good step. But I'm still confused about the `splitTarget` issue?

Zarel:
I tested `/challenge [Gen 8] OU @@@ +Reshiram, +Zekrom` and it seems to be working?

PartMan7:
When I tried it earlier, the challenge code you used wouldn't allow Zekrom. The first one would work as normal.

PR #6813, opened 2020-06-07
Allow formats to reorder all six Pokémon again

urkerab:
PR #4281 again, but with a comment this time, so that it won't get accidentally reverted like it did in PR #5427.


Zarel:
Sorry about that!

PR #6814, opened 2020-06-07
Groups: Edit after the merger

theodelhay:
Editing of the /groups after the merger of the administrator and leader ranks


sparkychildcharlie:
This isn't reflected in config-example.js tho

Zarel:
I appreciate the enthusiasm, but yes, this should probably simultaneously change `config-example`.

AnnikaCodes:
it'd also probably be a good idea to modify command help prompts that say `Requires: @ # & ~` and the like

theodelhay:
I just saw the problem, but Annika managed to do it all https://github.com/smogon/pokemon-showdown/pull/6815.

PR #6817, opened 2020-06-08
Add a library for utility functions

mia-pi-git:
Moves `splitFirst`, `html`, `escapeHTML`, and `stringify` out of `Chat` (renames `stringify` to `visualize`, as well) and into a new `lib/utils` library. 
Also adds an `escapeRegex` method, and borrows `sortBy` and `compare` from the client's `PSUtils`.


lib/utils.ts
```diff
@@ -0,0 +1,175 @@
+type Comparable = number | string | boolean | Comparable[] | {reverse: Comparable};
```
Zarel, line 1:
```suggestion
/**
 * Utils library
 *
 * Miscellaneous utility functions that don't really have a better place.
 *
 * It'll always be a judgment call whether or not a function goes into a
 * "catch-all" library like this, so here are some guidelines:
 *
 * - It must not have any dependencies
 *
 * - It must conceivably have a use in a wide variety of projects, not just
 *   Pokémon (if it's Pokémon-specific, Dex is probably a good place for it)
 *
 * - A lot of Chat functions are kind of iffy, but I'm going to say for now
 *   that if it's English-specific, it should be left out of here.
 */

type Comparable = number | string | boolean | Comparable[] | {reverse: Comparable};
```

Zarel:
I believe `Dex.shuffle` and `Dex.clampIntRange` probably go here.

lib/utils.ts
```diff
@@ -0,0 +1,192 @@
...
+				splitStr.push(str);
+				str = '';
+			}
+		}
+		splitStr.push(str);
+		return splitStr;
+	}
+
+	/**
+	* Species string tag function for escaping HTML
```
Zarel, line 181:
```suggestion
	* Template string tag function for escaping HTML
```

lol, this apparently got typoed in the Template -> Species refactor.

Zarel:
On Discord we discussed that `Dex.shuffle` should be removed completely, and `Dex.clampIntRange` should probably be dealt with the way `this.trunc` currently works. I'll draft this PR until those changes are in.

sim/battle.ts
```diff
@@ -148,6 +149,7 @@ export class Battle {
 		this.zMoveTable = {};
 		this.maxMoveTable = {};
 		this.trunc = this.dex.trunc;
+		this.clampIntRange = this.dex.clampIntRange;
```
Zarel, line 152:
```suggestion
		this.clampIntRange = Utils.clampIntRange;
```

You'll need to import `Utils`, of course.

mia-pi-git: Isn't sim supposed to be self-contained?

Zarel: You should be deleting `clampIntRange` in Dex, too.

mia-pi-git: I can do this but i was under the impression outside imports were bad.

Zarel:
This is definitely going to cause problems with private chat plugins, incidentally.

Zarel:
One thing we could do is anonymously attach them to Chat, like this:

```
(Chat as any).escapeHTML = Utils.escapeHTML;
(Chat as any).html = Utils.html;
(Chat as any).splitFirst = Utils.splitFirst;
```

Which would let us do a staggered deploy, while still letting TypeScript catch errors.

Zarel:
I'm surprised Dexsearch doesn't use levenshtein anymore. That's why it's in the codebase in the first place.

tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+
```
Zarel, line 2:
```suggestion
```

server/chat.ts
```diff
@@ -2013,6 +1896,10 @@ export const Chat = new class {
 	}
 };
 
+(Chat as any).escapeHTML = Utils.escapeHTML;
+(Chat as any).html = Utils.html;
+(Chat as any).splitFirst = Utils.splitFirst;
```
Zarel, line 1901:
```suggestion
// backwards compatibility; don't actually use these
// they're just there so forks have time to slowly transition
(Chat as any).escapeHTML = Utils.escapeHTML;
(Chat as any).html = Utils.html;
(Chat as any).splitFirst = Utils.splitFirst;
```

Zarel:
Does this actually require a restart? afaict it shouldn't.

PR #6820, opened 2020-06-08
Roomevents: Support event categories

AnnikaCodes:


Zarel:
Oh, hey, I remember that suggestion.

Zarel:
Oh no, more conflicts with Settings refactor. :(

Zarel:
Hm. I think if categories and aliases are going to continue sharing the same namespace, they should probably be in the same table.

So `room.events` would be changed from:

    events: {[k: string]: RoomEvent};

to

    events: {[k: string]: RoomEvent | RoomEventAlias | RoomEventCategory};

As-is, I think your code is a bit too brittle to namespace conflicts.

Zarel:
There are other ways to avoid namespace conflicts. Your current approach definitely saves storage space, which is probably more precious than performance (making the table contain all three is O(1) rather than O(n), which can be important but is probably unnoticeable in this specific case).

You could separate the tables: an `events: {[k: string]: RoomEvent}` table, and then a cached `events: Map<string, RoomEvent | RoomEventAlias | RoomEventCategory>` somewhere else.

If you wanted to keep using your current approach you'd probably at least a search function to check if a given name is already an event/alias/category (and if so, which it is). Possibly also a command to list all categories (your current implementation has duplicates).

Either way, this problem needs to be solved _some_ way, so I'm going to draft this until that's solved.

AnnikaCodes:
> Hm. I think if categories and aliases are going to continue sharing the same namespace, they should probably be in the same table.
> 
> So `room.events` would be changed from:
> 
> ```
> events: {[k: string]: RoomEvent};
> ```
> 
> to
> 
> ```
> events: {[k: string]: RoomEvent | RoomEventAlias | RoomEventCategory};
> ```
> 
> As-is, I think your code is a bit too brittle to namespace conflicts (i.e. what happens if you try to set an alias and a category to the same thing?)

This seems sensible and I'm working on it now, but wouldn't this be incompatible with preexisting aliases stored on the server? (i.e. they wouldn't show up since `room.events['someEvent'].aliases` would never be accessed)

Zarel:
> This seems sensible and I'm working on it now, but wouldn't this be incompatible with preexisting aliases stored on the server? (i.e. they wouldn't show up since `room.events['someEvent'].aliases` would never be accessed)

There are several ways to deal with this:

1. **Don't actually store things that way in the server.**

So you would have, stored:

`room.settings.events?: {[k: string]: RoomEvent};`

And then you would load them into:

`room.events?: Map<string, RoomEvent | RoomEventAlias | RoomEventCategory>`

2. **Make everyone re-enter their aliases after the change**

3. **Do a one-time conversion on the server from the old format to the new format**

I'd probably recommend #1.

Zarel:
You might consider adding categories as separate "things", though? I think it's probably useful to manually add and remove categories as a thing, so when you add an event to a category, it will say "that category doesn't exist" if you typo.

AnnikaCodes:
> You might consider adding categories as separate "things", though? I think it's probably useful to manually add and remove categories as a thing, so when you add an event to a category, it will say "that category doesn't exist" if you typo.

I agree, but it seems odd to me to load aliases out of `room.settings.events?` into a separate Map but not do the same for categories. Option 2 seems simplest for that, but can we actually do that, or will people be too angry at having to reenter their aliases?

Zarel:
A one-time conversion is about as straightforward, and amounts to the same thing. Heck, I can do the conversion for you when it happens. Shouldn't be a problem.

Zarel:
I could also just write the one-time conversion into this PR, actually. Just write the rest of it, and I'll write it in as a suggestion if you can't figure it out.

Zarel:
The main reason I'd want to treat aliases and categories differently is that it makes sense to delete an alias when you delete its corresponding event. I don't think you should delete categories when you delete all events in that category.

AnnikaCodes:
Alright, that should be good (let me know if I missed anything).

Nested `for..of` loops seemed the best way to do the one-time conversion, but I'm not really familiar with how room settings are stored, so feel free to correct me.

server/chat-plugins/room-events.ts
```diff
@@ -78,7 +115,7 @@ export const commands: ChatCommands = {
 			const [eventName, date, ...desc] = target.split(target.includes('|') ? '|' : ',');
 
 			if (!(eventName && date && desc)) {
-				return this.errorReply("You're missing a command parameter - see /help roomevents for this command's syntax.");
+				return this.errorReply("You're missing a command parameter - to see this .");
```
AnnikaCodes, line 118:
```suggestion
				return this.errorReply("You're missing a command parameter - to see this command's syntax, use /help roomevents.");
```

server/rooms.ts
```diff
@@ -533,6 +536,18 @@ export class GlobalRoom extends BasicRoom {
 				Monitor.warn(`ERROR: Room number ${i} has no data and could not be loaded.`);
 				continue;
 			}
+
+			// convert from old RoomEvent.aliases format to RoomEventAlias format
+			if (settings.events) {
+				for (const event of Object.values(settings.events).map(e => e as AnyObject)) {
```
Zarel, line 542:
```suggestion
				for (const event of Object.values(settings.events) as AnyObject[]) {
```

server/chat-plugins/room-events.ts
```diff
@@ -19,30 +34,47 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
 	}
 	return aliases;
 }
+
+function getAllCategories(room: Room) {
+	const categories: string[] = [];
+	for (const category of Object.values(room.settings.events!)
+		.filter(cat => 'events' in cat)
+		.map(cat => cat as RoomEventCategory)
+	) {
```
Zarel, line 68:
```suggestion
	if (!room.settings.events) return [];
	const categories: string[] = [];
	for (const category of Object.values(room.settings.events).filter(cat => 'events' in cat)) {
```

If this isn't enough for TS, I'd probably define `events?: undefined` on `RoomEvent` and `RoomEventAlias` and then switch this to `cat => cat.events`.

server/chat-plugins/room-events.ts
```diff
@@ -19,30 +34,47 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
-	ret += showAliases ? Utils.html`<td>${event.aliases?.join(", ")}</td>` : ``;
+	if (showAliases) ret += Utils.html`<td>${aliases.join(", ")}</td>`;
+	if (showCategories) ret += Utils.html`<td>${categories.join(", ")}</td>`;
 	ret += `<td>${Chat.formatText(event.desc, true)}</td>`;
 	ret += Utils.html`<td><time>${event.date}</time></td></tr>`;
 	return ret;
 }
 
 function getAllAliases(room: Room) {
-	if (!room.settings.events) return [];
```
Zarel, line 31: ?

This should be restored; `room.settings.events` should not be taken as guaranteed to exist.

server/chat-plugins/room-events.ts
```diff
@@ -19,30 +34,47 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
 		explanation = `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`;
 	}
+
+	const eventID = toID(event.eventName);
+	const aliases = getAllAliases(room).filter(
+		alias => (room.settings.events?.[alias] as RoomEventAlias).eventID === eventID
+	);
+	const categories = getAllCategories(room).filter(
+		category => (room.settings.events?.[category] as RoomEventCategory).events.includes(eventID)
+	);
```
Zarel, line 44:
```suggestion
	const aliases = getAllAliases(room).filter(
		alias => (room.settings.events![alias] as RoomEventAlias).eventID === eventID
	);
	const categories = getAllCategories(room).filter(
		category => (room.settings.events![category] as RoomEventCategory).events.includes(eventID)
	);
```

server/chat-plugins/room-events.ts
```diff
@@ -8,9 +8,24 @@
...
-type RoomEvent = NonNullable<import('../rooms').RoomSettings['events']>[''];
+export interface RoomEvent {
+	eventName: string;
+	date: string;
+	desc: string;
+	started: boolean;
+}
+
+export interface RoomEventAlias {
+	id: ID;
```
Zarel, line 19: Hm, I think it's confusing for aliases/categories to have `id` properties, but not events. I'd probably omit `id` from all of them - that prevents `id` from getting out of sync with the `room.settings.events` key, after all.

AnnikaCodes: Alright.

server/rooms.ts
AnnikaCodes: This makes the checks fail, though.

server/chat-plugins/room-events.ts
```diff
@@ -30,19 +53,25 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
 function getAllAliases(room: Room) {
 	if (!room.settings.events) return [];
 	const aliases: string[] = [];
-	for (const event of Object.values(room.settings.events)) {
-		if (event.aliases) aliases.push(...event.aliases);
+	for (const aliasID of Object.keys(room.settings.events).filter(id => 'eventID' in room.settings.events![id])) {
+		aliases.push(aliasID);
 	}
```
Zarel, line 58: tbh this is a weird mix of functional and regular code, and it makes more sense to pick one or the other.

```suggestion
	for (const aliasID in room.settings.events) {
		if ('eventID' in room.settings.events[aliasID]) aliases.push(aliasID);
	}
```

server/chat-plugins/room-events.ts
```diff
@@ -30,19 +53,25 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
 	return aliases;
 }
+
+function getAllCategories(room: Room) {
+	if (!room.settings.events) return [];
+	const categories: string[] = [];
+	for (const categoryID of Object.keys(room.settings.events).filter(id => 'events' in room.settings.events![id])) {
+		categories.push(categoryID);
+	}
```
Zarel, line 67:
```suggestion
	for (const categoryID in room.settings.events) {
		if ('events' in room.settings.events[id]) categories.push(categoryID);
	}
```

server/chat-plugins/room-events.ts
Zarel: If you instead wanted to make it purely functional, it would look like this:

```ts
return Object.entries(room.settings.events).filter(
  ([id, event]) => 'eventID' in event
).map(
  ([id]) => id
);
```

I don't have a strong opinion of which style to prefer; it's up to you.

server/chat-plugins/room-events.ts
```diff
@@ -19,9 +32,19 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
 	if (!isNaN(timeRemaining)) {
 		explanation = `This event will start in: ${Chat.toDurationString(timeRemaining, {precision: 2})}`;
 	}
+
+	const eventID = toID(event.eventName);
+	const aliases = getAllAliases(room).filter(
+		alias => (room.settings.events![alias] as RoomEventAlias).eventID === eventID
+	);
```
Zarel, line 39: This suggests to me that `getAllAliases` isn't the right level of abstraction. You might perhaps want to pass an `eventID` parameter, so you could do `getAliases(room)` but also `getAliases(room, eventID)` to only get aliases for a specific event.

server/chat-plugins/room-events.ts
```diff
@@ -30,19 +53,25 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
 function getEventID(nameOrAlias: string, room: Room): ID {
 	let id = toID(nameOrAlias);
-	if (room.settings.events && !room.settings.events[id]) {
-		for (const possibleEvent in room.settings.events) {
-			if (room.settings.events[possibleEvent].aliases?.includes(id)) {
-				id = toID(possibleEvent);
-			}
-		}
+	if (room.settings.events?.[id] && 'eventID' in room.settings.events[id]) {
+		id = (room.settings.events[id] as RoomEventAlias).eventID;
```
Zarel, line 74:
```suggestion
	const event = room.settings.events?.[id];
	if (event && 'eventID' in event) {
		id = event.eventID;
```

(It's generally at three repetitions that I start using intermediate variables.)

server/chat-plugins/room-events.ts
```diff
@@ -58,11 +87,16 @@ export const commands: ChatCommands = {
...
-			for (const i in room.settings.events) {
-				buff += formatEvent(room.settings.events[i], hasAliases);
+			buff += '<th>Event Name:</th>';
+			if (hasAliases) buff += '<th>Event Aliases:</th>';
+			if (hasCategories) buff += '<th>Event Categories:</th>';
+			buff += '<th>Event Description:</th><th>Event Date:</th>';
+
+			for (const event of Object.values(room.settings.events).filter(e => 'eventName' in e).map(e => e as RoomEvent)) {
+				buff += formatEvent(room, event, hasAliases, hasCategories);
 			}
```
Zarel, line 100:
```suggestion
			for (const event of Object.values(room.settings.events)) {
				if ('eventName' in event) {
					buff += formatEvent(room, event, hasAliases, hasCategories);
				}
			}
```

I suspect you might want a `getAllEvents` to go with your other functions...

server/chat-plugins/room-events.ts
```diff
@@ -319,15 +503,23 @@ export const commands: ChatCommands = {
...
-		`/roomevents remove [event name] - Deletes an event. Requires: @ # &`,
-		`/roomevents rename [old event name] | [new name] - Renames an event. Requires: @ # &`,
-		`/roomevents addalias [alias] | [event name] - Adds an alias for the event. Requires: @ # &`,
-		`/roomevents removealias [alias] - Removes an event alias. Requires: @ # &`,
-		`/roomevents sortby [column name] | [asc/desc (optional)] - Sorts events table by column name and an optional argument to ascending or descending order. Ascending order is default. Requires: @ # &`,
-		`/roomevents view [event name] - Displays information about a specific event.`,
-	],
+	roomeventshelp() {
+		this.sendReply(
+			`|html|<div class="infobox infobox-limited">` +
```
Zarel, line 508: Why not a readmore? They're generally more convenient than a scrollybox.

server/chat-plugins/room-events.ts
Zarel: (Both styles support filtering by event ID roughly equally well.)

Zarel:
Do you know what the build error is about? It's confusing to me.

AnnikaCodes:
> Do you know what the build error is about? It's confusing to me.

It's confusing me, too - it's about the assertion `Object.values(settings.events) as AnyObject[]`. That's why I originally used `Object.values(settings.events).map(e => e as AnyObject)`.

Zarel:
Hm, let me try screwing with it myself.

Zarel:
I added some strong typing to `Rooms.global.settingsList`, which works around your bug. It seems to have exposed a different bug in your code, which I'll let you fix yourself.

I strongly recommend putting your conversion script inside your chat plugin. I want as little chat plugin code in `rooms.ts` as possible.

Zarel:
(There's another reason why your conversion script should be inside your chat plugin: Putting it in `rooms.ts` will break hotpatching. I've added a flag to note that, but of course as mentioned earlier, the preferred solution would be for you to move it inside the chat plugin.)

server/chat-plugins/room-events.ts
```diff
@@ -19,30 +43,57 @@ function formatEvent(event: RoomEvent, showAliases?: boolean) {
...
-function getAllAliases(room: Room) {
+function getAliases(room: Room, eventID?: ID) {
 	if (!room.settings.events) return [];
 	const aliases: string[] = [];
-	for (const event of Object.values(room.settings.events)) {
-		if (event.aliases) aliases.push(...event.aliases);
+	for (const aliasID in room.settings.events) {
+		if (
+			'eventID' in room.settings.events[aliasID] &&
+			(!eventID || (room.settings.events[aliasID] as RoomEventAlias).eventID === eventID)
```
Zarel, line 68:
```suggestion
			(!eventID || room.settings.events[aliasID].eventID === eventID)
```

I'm surprised this isn't recognized as a useless cast.

Zarel:
Honestly, this works fine and you've done a lot of work on this already. Let's save the optimization pass for a different PR.

PR #6823, opened 2020-06-09
Scavengers: fix /scavtop error due to auth/settings refactor

sparkychildcharlie:
meow


Zarel:
lol, I think I actually told Mia about this bug when I caught it in one place, apparently I didn't catch it in all the places.

server/chat-plugins/scavengers.ts
```diff
@@ -1860,7 +1860,7 @@ const ScavengerCommands: ChatCommands = {
 		const ladder = await Leaderboard.visualize('points') as AnyObject[];
 		this.sendReply(
 			`|uhtml${isChange ? 'change' : ''}|scavladder|<div class="ladder" style="overflow-y: scroll; max-height: 300px;"><table style="width: 100%"><tr><th>Rank</th><th>Name</th><th>Points</th></tr>${ladder.map(entry => {
-				const isStaff = room.auth.has(user.id);
+				const isStaff = room.auth.has(toId(entry.name));
```
Zarel, line 1863:
```suggestion
				const isStaff = room.auth.has(toID(entry.name));
```

sparkychildcharlie: yep, this is what i get for "polyfilling" in toId on Dawn

PR #6827, opened 2020-06-11
Change |pm|~ to |pm|&

AustinXII:


Zarel:
I think it's fine to leave `~` for server messages.

AustinXII:
@scheibo said it makes more sense as & so I did it mostly for him. Idrc either way personally but I do agree it makes more sense

Zarel:
I guess I don't really care, so sure.

PR #6828, opened 2020-06-11
Don't default to showing 'trusted' in /userauth

AnnikaCodes:


server/chat-commands/moderation.ts
```diff
@@ -286,9 +286,9 @@ export const commands: ChatCommands = {
 		const buffer = [];
 		let innerBuffer = [];
 		const group = Users.globalAuth.get(targetId);
-		if (group) {
-			buffer.push(`Global auth: ${group === ' ' ? 'trusted' : group}`);
-		}
+		let showTrusted = false;
```
scheibo, line 289:
```suggestion
const showTrusted = Users.isTrusted(targetID) && group === ' ';
```

server/chat-commands/moderation.ts
```diff
@@ -286,9 +286,9 @@ export const commands: ChatCommands = {
...
+		if (group === ' ' && Users.isTrusted(targetId)) showTrusted = true;
```
scheibo, line 290:
```suggestion
const showTrusted = group === ' ' && Users.isTrusted(targetID);
```

server/chat-commands/moderation.ts
```diff
@@ -286,9 +286,8 @@ export const commands: ChatCommands = {
 		const buffer = [];
 		let innerBuffer = [];
 		const group = Users.globalAuth.get(targetId);
-		if (group) {
-			buffer.push(`Global auth: ${group === ' ' ? 'trusted' : group}`);
-		}
+		const showTrusted = group === ' ' && Users.isTrusted(targetID);
```
scheibo, line 289:
```suggestion
		const showTrusted = group === ' ' && Users.isTrusted(targetId);
```

server/chat-commands/moderation.ts
```diff
@@ -286,9 +286,8 @@ export const commands: ChatCommands = {
 		const buffer = [];
 		let innerBuffer = [];
 		const group = Users.globalAuth.get(targetId);
-		if (group) {
-			buffer.push(`Global auth: ${group === ' ' ? 'trusted' : group}`);
-		}
+		const showTrusted = group === ' ' && Users.isTrusted(targetId);
+		if (group !== ' ' || showTrusted) buffer.push(`Global auth: ${showTrusted ? 'trusted' : group}`);
```
Zarel, line 290:
```suggestion
```suggestion
		if (group !== ' ' || Users.isTrusted(targetId)) {
			buffer.push(`Global auth: ${group === ' ' ? 'trusted' : group}`);
		}
``````

File: server/chat-commands/moderation.ts [hunk 5]
```diff
@@ -286,7 +286,7 @@ export const commands: ChatCommands = {
 		const buffer = [];
 		let innerBuffer = [];
 		const group = Users.globalAuth.get(targetId);
-		if (group) {
+		if (group !== ' ' || Users.isTrusted(targetId)) {
```
Zarel, line 289:
```suggestion
		if (Users.globalAuth.has(targetId)) {
```

Zarel: n-nothing wrong with your code; I'm just golfing

AnnikaCodes: yeah, it looks like trusted users who aren't global auth aren't in `Users.globalAuth`

PR #6829, opened 2020-06-12
CAP: Volkraken and Volkritter are gen 6

TheMezStrikes:
Evidence, if needed:
- https://www.smogon.com/forums/threads/cap-18-final-product.3507516/post-5471155 final product was 2014, well after gen 6 release
- https://www.smogon.com/cap/pokemon/ has it listed under gen 6
- learnsets.js doesn't contain gen 5 data


Zarel:
Good catch!

PR #6830, opened 2020-06-13
Fix uncommon evolution methods

KrisXV:


data/pokedex.ts
```diff
@@ -11650,6 +11654,7 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 		color: "Blue",
 		prevo: "Inkay",
 		evoLevel: 30,
+		evoCondition: "with an console turned upside-down",
```
urkerab, line 11657:
```suggestion
		evoCondition: "with the console turned upside-down",
```

KrisXV: I don’t like the wording of “the”. I’ll fix it to “a” when I’m on a computer.

Zarel: "the" is definitely better English.

Zarel:
(Specifically, "the console" means "a specific console, you know which one I'm talking about, the one you're trying to evolve the Inkay on", while "a console" means "any of multiple consoles will work, it doesn't matter which one". So using "a console" here is incorrect English.)

PR #6832, opened 2020-06-14
Fix event validator for RBY Tradebacks

Enigami:
If you attempt to challenge an opponent to Tradebacks, it will reject Pokemon with NYPC moves even though they should be legal. For example, attempting to use Lovely Kiss Snorlax in RBY Tradebacks:

"Your team was rejected for the following reasons:

- Snorlax has an event-exclusive move that it doesn't qualify for (only one of several ways to get the move will be listed):
- This format is in gen 1 and Snorlax is from gen 2 because it has a move only available from an event."

By changing dex.gen to this.maxSourceGen in the validator for event moves, NYPC moves work as they should in Tradebacks.

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1457,6 +1457,7 @@ export class TeamValidator {
 		const dex = this.dex;
 		let name = set.species;
 		const species = dex.getSpecies(set.species);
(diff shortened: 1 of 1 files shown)


Enigami:
Odd, it says maxSourceGen does not exist. This did not occur on my test server. Investigating...

Zarel:
Good catch!

PR #6834, opened 2020-06-14
Implement Stadium Sleep Clause

Enigami:
Adds Stadium's Sleep Clause to [Gen 1] Stadium OU.

In Pokemon Stadium, Sleep Clause only checks whether any Pokemon on the opposing side is asleep to activate, and does not check the source of sleep.

This means if a player's Pokemon uses Rest, Sleep Clause is activated for their side and is protected from Sleep moves, even though the sleep was self-inflicted from Rest.

Proof:
https://www.smogon.com/forums/threads/gen-1-and-tradebacks-dev-post-bugs-here.3524844/page-14#post-8502315

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -2273,7 +2273,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'stadium',
 		searchShow: false,
-		ruleset: ['Obtainable', 'Team Preview', 'Stadium Sleep Clause Mod', 'Species Clause', 'Nickname Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Endless Battle Clause', 'HP Percentage Mod', 'Cancel Mod'],
+		ruleset: ['Standard', 'Team Preview', '!Sleep Clause Mod', 'Stadium Sleep Clause Mod'],
 		banlist: ['Uber',
 			'Nidoking + Fury Attack + Thrash', 'Exeggutor + Poison Powder + Stomp', 'Exeggutor + Sleep Powder + Stomp',
 			'Exeggutor + Stun Spore + Stomp', 'Jolteon + Focus Energy + Thunder Shock', 'Flareon + Focus Energy + Ember',
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -2273,7 +2273,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'stadium',
 		searchShow: false,
-		ruleset: ['Obtainable', 'Team Preview', 'Stadium Sleep Clause Mod', 'Species Clause', 'Nickname Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Endless Battle Clause', 'HP Percentage Mod', 'Cancel Mod',],
+		ruleset: ['Obtainable', 'Team Preview', 'Stadium Sleep Clause Mod', 'Species Clause', 'Nickname Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Endless Battle Clause', 'HP Percentage Mod', 'Cancel Mod'],
 		banlist: ['Uber',
 			'Nidoking + Fury Attack + Thrash', 'Exeggutor + Poison Powder + Stomp', 'Exeggutor + Sleep Powder + Stomp',
 			'Exeggutor + Stun Spore + Stomp', 'Jolteon + Focus Energy + Thunder Shock', 'Flareon + Focus Energy + Ember',
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -2273,7 +2273,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'stadium',
 		searchShow: false,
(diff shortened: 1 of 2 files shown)


config/formats.ts
```diff
@@ -2273,7 +2273,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'stadium',
 		searchShow: false,
-		ruleset: ['Standard', 'Team Preview'],
+		ruleset: ['Obtainable', 'Team Preview', 'Stadium Sleep Clause Mod', 'Species Clause', 'Nickname Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Endless Battle Clause', 'HP Percentage Mod', 'Cancel Mod'],
```
Zarel, line 2276: Are all the other rules supposed to be the same? if so, it might be easier to do this:

```suggestion
		ruleset: ['Standard', 'Team Preview', '!Sleep Clause Mod', 'Stadium Sleep Clause Mod'],
```

data/rulesets.ts
```diff
@@ -735,6 +735,27 @@ export const BattleFormats: {[k: string]: FormatsData} = {
...
+			if (status.id === 'slp') {
+				for (const pokemon of target.side.pokemon) {
+					if (pokemon.hp && pokemon.status === 'slp') {
+						this.add('-message', "Sleep Clause Mod activated. (In Stadium, Sleep Clause activates if any of the opponent's Pokemon are asleep, even if self-inflicted from Rest)");
+						return false;
+					}
+				}
+			}
+		},
+	},
```
Zarel, line 758:
```suggestion
	stadiumsleepclause: {
		effectType: 'Rule',
		name: 'Stadium Sleep Clause',
		desc: "Prevents players from putting one of their opponent's Pok&eacute;mon to sleep if any of the opponent's other Pok&eacute;mon are asleep",
		onBegin() {
			this.add('rule', 'Stadium Sleep Clause: Limit one foe put to sleep');
		},
		onSetStatus(status, target, source) {
			if (source && source.side === target.side) {
				return;
			}
			if (status.id === 'slp') {
				for (const pokemon of target.side.pokemon) {
					if (pokemon.hp && pokemon.status === 'slp') {
						this.add('-message', "Sleep Clause activated. (In Stadium, Sleep Clause activates if any of the opponent's Pokemon are asleep, even if self-inflicted from Rest)");
						return false;
					}
				}
			}
		},
	},
```

It's not a mod if it's how it actually works in Stadium.

data/rulesets.ts
```diff
@@ -735,6 +735,27 @@ export const BattleFormats: {[k: string]: FormatsData} = {
...
+		effectType: 'Rule',
+		name: 'Stadium Sleep Clause Mod',
+		desc: "Prevents players from putting one of their opponent's Pok&eacute;mon to sleep if any of the opponent's other Pok&eacute;mon are asleep",
+		onBegin() {
+			this.add('rule', 'Stadium Sleep Clause Mod: Limit one foe put to sleep');
+		},
+		onSetStatus(status, target, source) {
+			if (source && source.side === target.side) {
+				return;
+			}
```
Zarel, line 748: Wait, I thought you said it activates on Rest? Why do you still have the part of the code that stops it from activating from Rest?

config/formats.ts
Enigami: Yes, they should be the same. I didn't know you could do that "!clause" thing, that's convenient!

data/rulesets.ts
Enigami: Because Rest also calls onSetStatus, and if I don't have that check to see if it's coming from your own side, Sleep Clause  would prevent the user from using Rest if any other Pokemon on the same side is asleep, which as far as I am aware should not occur.

data/rulesets.ts
Enigami: Right, removed "Mod" from the clause's name.

data/rulesets.ts
Zarel: Wait, that's the opposite of what you said in the intro:

> This means if a player's Pokemon uses Rest, Sleep Clause is activated for their side and is protected from Sleep moves, even though the sleep was self-inflicted from Rest.

Zarel: Oh, I see, the difference is that you can still use Rest on yourself after you've already Rested, but your opponent can't use sleep moves on you.

I think that should probably be documented clearly in the mod description.

data/rulesets.ts
```diff
@@ -735,6 +735,27 @@ export const BattleFormats: {[k: string]: FormatsData} = {
 			}
 		},
 	},
+	stadiumsleepclause: {
+		effectType: 'Rule',
+		name: 'Stadium Sleep Clause',
+		desc: "Prevents players from putting one of their opponent's Pok&eacute;mon to sleep if any of the opponent's other Pok&eacute;mon are asleep",
```
Zarel, line 741:
```suggestion
		desc: "Prevents players from putting one of their opponent's Pokémon to sleep if any of the opponent's other Pokémon are asleep (different from Sleep Clause Mod because putting your own Pokémon to sleep is enough to prevent opponents from putting your others to sleep).",
```

PR #6838, opened 2020-06-15
Fix various Stadium issues

Enigami:
Fixed the following issues with Stadium:
-Must Recharge did not display after using Hyper Beam
-Hyper Beam did not require recharging after missing
-Self-destruct and Explosion did not cause the user to faint if it broke a Substitute
-All moves ignored the semi-invulnerability from Dig/Fly
-Dream Eater hit Pokemon that were not asleep
-Recoil moves inflicted recoil when the opponent was KO'd

Based fixes on research done by Plague von Karma and Beelzemon 2003
https://www.smogon.com/forums/threads/stadium-format-is-now-available-on-ps.3526616/page-2#post-8502629

### DIFF
--- a/data/mods/stadium/moves.ts
+++ b/data/mods/stadium/moves.ts
@@ -21,6 +21,12 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 			}
 		},
 	},
+	hyperbeam: {
+		inherit: true,
+		onMoveFail(target, source, move) {
+			source.addVolatile('mustrecharge', target, move, 'trapper');
+		},
+	},
 	jumpkick: {
 		inherit: true,
 		desc: "If this attack misses the target, the user 1HP of damage.",
--- a/data/mods/stadium/moves.ts
+++ b/data/mods/stadium/moves.ts
@@ -24,7 +24,7 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 	hyperbeam: {
 		inherit: true,
 		onMoveFail(target, source, move) {
-			source.addVolatile('mustrecharge', target, move, 'trapper');
+			source.addVolatile('mustrecharge');
 		},
 	},
 	jumpkick: {
--- a/data/mods/stadium/scripts.ts
+++ b/data/mods/stadium/scripts.ts
@@ -132,7 +132,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 			return false;
 		}
 
-		// Then, check if the PokÃ©mon is immune to this move.
+		// Then, check if the Pokemon is immune to this move.
 		if (
 			(!move.ignoreImmunity || (move.ignoreImmunity !== true && !move.ignoreImmunity[move.type])) &&
 			!target.runImmunity(move.type, true)
--- a/data/mods/stadium/scripts.ts
+++ b/data/mods/stadium/scripts.ts
@@ -97,7 +97,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 		this.useMove(move, pokemon, target, sourceEffect);
 		this.singleEvent('AfterMove', move, null, pokemon, target, move);
 
-		// If rival fainted
+		// If target fainted
 		if (target && target.hp <= 0) {
 			// We remove screens
 			target.side.removeSideCondition('reflect');
@@ -105,6 +105,7 @@ export const BattleScripts: ModdedBattleScriptsData = {
 		} else {
 			this.runEvent('AfterMoveSelf', pokemon, target, move);
 		}
+		if (pokemon.volatiles['mustrecharge']) this.add('-mustrecharge', pokemon);
 
 		// For partial trapping moves, we are saving the target.
 		if (move.volatileStatus === 'partiallytrapped' && target && target.hp > 0) {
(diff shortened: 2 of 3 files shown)


data/mods/stadium/moves.ts
```diff
@@ -21,6 +21,12 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 			}
 		},
 	},
+	hyperbeam: {
+		inherit: true,
+		onMoveFail(target, source, move) {
+			source.addVolatile('mustrecharge', target, move, 'trapper');
```
Zarel, line 27: Why 'trapper'? Was this copy/pasted from a trapping move? Please don't tell me RBY is also like this...

Zarel: (`trapper` is a flag for modern trapping moves, which means "this effect goes away if the opponent switches out", which obviously doesn't apply to Stadium Hyper Beam.)

Zarel: It looks like no, it wasn't copy/pasted from RBY. Whew, I was afraid RBY needed to be fixed. Where did you get this from?

```suggestion
			source.addVolatile('mustrecharge');
```

data/mods/stadium/scripts.ts
```diff
@@ -121,10 +122,17 @@ export const BattleScripts: ModdedBattleScriptsData = {
...
-		// First, check if the Pokémon is immune to this move.
+		// First, check if the target is semi-invulnerable
+		let hitResult = this.runEvent('Invulnerability', target, pokemon, move);
+		if (hitResult === false) {
+			if (!move.spreadHit) this.attrLastMove('[miss]');
+			this.add('-miss', pokemon);
+			return false;
+		}
+
+		// Then, check if the PokÃ©mon is immune to this move.
```
Zarel, line 135:
```suggestion
		// Then, check if the Pokémon is immune to this move.
```

This is why I prefer to leave Unicode out of source code.

But also wow, what are you doing with a Unicode-incompatible text editor???

Zarel:
Thanks! To you and also to Plague von Karma and Beelzemon 2003 and everyone else who helped find and document these issues.

I don't suppose I could get you to write unit tests for this? Completely optional, but it would help a lot to make sure Stadium isn't broken by future changes in the code.

DisasterAreaPS:
> 
> 
> Thanks! To you and also to Plague von Karma and Beelzemon 2003 and everyone else who helped find and document these issues.
> 
> I don't suppose I could get you to write unit tests for this? Completely optional, but it would help a lot to make sure Stadium isn't broken by future changes in the code.

What are unit tests?

Zarel:
> What are unit tests?

A unit test is some code that describes what should happen if you do something, and crashes if it doesn't do what it's supposed to do.

Here's an example:

https://github.com/smogon/pokemon-showdown/blob/master/test/sim/abilities/arenatrap.js

PS runs unit tests for every commit and pull request, and the green check will turn into a red X if one of them crashes.

DisasterAreaPS:
Btw, one thing stadium needs is self-KO clause (unavoidable when playing on the actual system). I presume there's a really easy way to do that at the format level, like adding 'Self-KO Clause' to the list of clauses in formats.js? Can you do it or show me how to do it or is it more complicated?

Zarel:
Are you asking me or Enigami? I don't personally have free time right now, it's basically all used up fixing Mia's PRs, and if I have free time beyond that I'm working on getting the Preact client to a shippable state, and overhauling the databases.

Self-KO Clause is standard in modern generations, so it shouldn't be too hard for Enigami to bring that code to Stadium. Does Stadium/RBY have ties at all?

DisasterAreaPS:
RBY has ties, from self-KO at the end of the game - on cartridge, it says both players lose. Stadium however has Self-KO clause (also, worth noting that recoil moves such as Double-Edge don't have recoil when they are used to KO the last Pokemon, so the pokemon not taking recoil would win in that circumstance).

Question was more aimed at you, I would ask Enigami through different channels. Not sure if Enigami knows either, and I figured it was probably a simple question.

data/mods/stadium/moves.ts
Enigami: I don't remember where I copied it from, but somehow I didn't pay attention to the unnecessary extra arguments.

data/mods/stadium/scripts.ts
Enigami: The terminal I use with my test server has a built in text editor that I used for quick changes which seems to lack Unicode, forgot to fix the unicode afterwards

Enigami:
I could not find Self-KO clause for formats or challenges, and doing a search for Self-KO (and a few variants of that text) yields nothing. If it does currently exist in Showdown, it must be unmarked and buried somewhere...

Marty-D:
Self-KO Clause isn't really a thing. It's either part of the game or it isn't; you can't turn it on or off. So it's just done in the faint checks when there are no remaining Pokemon:
https://github.com/smogon/pokemon-showdown/blob/6839a1f4dbf0a716bf80bf16d13b97a6e031cbc5/sim/battle.ts#L2419-L2422
(To clarify, Gens 5-8 pick the side that fainted last as the winner and Gens 1-4 rule this as a tie. You can probably also check for the Stadium mod here.)

DisasterAreaPS:
> 
> 
> Self-KO Clause isn't really a thing. It's either part of the game or it isn't; you can't turn it on or off. So it's just done in the faint checks when there are no remaining Pokemon:
> 
> https://github.com/smogon/pokemon-showdown/blob/6839a1f4dbf0a716bf80bf16d13b97a6e031cbc5/sim/battle.ts#L2419-L2422
> 
> (To clarify, Gens 5-8 pick the side that fainted last as the winner and Gens 1-4 rule this as a tie. You can probably also check for the Stadium mod here.)

Yeah would probably make sense to include stadium as one that effectively has Self-KO clause. Anyone know about stadium 2/XD/PBR/etc. - probably good to update to include as many of those non-cartridge battling formats as possible when updating this? :]

Zarel:
I'm merging because this is an improvement on existing code – feel free to keep discussing future RBY/Stadium improvements here, though.

PlagueVonKarma:
To clear up info on Stadium Self-KO Clause; 

**Regardless of who faints first**, the user of Explosion or Self-Destruct will lose. 

This is also enforced in Stadium 2, with Perish Song and Destiny Bond failing. I believe on Cartridge for GSC, these cases would be enforced as a game loss.

PR #6840, opened 2020-06-16
Isle of Armor: Abilities

KrisXV:


data/abilities.ts
```diff
@@ -4284,6 +4284,18 @@ export const BattleAbilities: {[abilityid: string]: AbilityData} = {
 		rating: 1.5,
 		num: 127,
 	},
+	unseenfist: {
+		desc: "All of this Pokemon's moves that make contact bypass protection.",
+		shortDesc: "All contact moves hit through protection.",
+		onModifyMove(move) {
+			if (!move.flags['contact']) return;
+			if (move.flags['protect']) delete move.flags['protect'];
```
urkerab, line 4292:
```suggestion
			if (move.flags['contact']) delete move.flags['protect'];
```

KrisXV: I don’t like this. What if there is a move with a contact flag but not a protect flag?

Marty-D: Then it won't matter since there will be nothing to delete. Long Reach does a similar thing, just deleting the contact flag off any move regardless of whether it actually is contact.

PR #6841, opened 2020-06-16
Isle of Armor: Moves

KrisXV:


data/moves.ts
```diff
@@ -609,7 +609,7 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 					const moveid = moveSlot.id;
 					if (noAssist.includes(moveid)) continue;
 					const move = this.dex.getMove(moveid);
-					if (move.isZ || move.isMax) {
+					if (move.isZ || (move.isMax && !['gmaxoneblow', 'gmaxrapidflow'].includes(move.id))) {
```
urkerab, line 612: It looks as if you've been too eager with your find and replace... this (and similar occurrences) probably shouldn't have been changed.

data/moves.ts
```diff
@@ -991,7 +991,9 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 			onTryHitPriority: 3,
 			onTryHit(target, source, move) {
 				if (!move.flags['protect']) {
-					if (move.isZ || move.isMax) target.getMoveHitData(move).zBrokeProtect = true;
+					if (move.isZ || (move.isMax && !['gmaxoneblow', 'gmaxrapidflow'].includes(move.id))) {
+						target.getMoveHitData(move).zBrokeProtect = true;
```
urkerab, line 995: The description for those moves says that they break protection. Assuming that's accurate, this suggests that the protection should be removed before the move hits, and then this change (and similar occurrences) won't be necessary.

data/moves.ts
```diff
@@ -7305,6 +7450,35 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
...
+		// NOT CONFIRMED
+		pp: 5,
+		priority: 0,
+		// NOT CONFIRMED
+		flags: {contact: 1, protect: 1},
+		onModifyMove(move, pokemon) {
+			if (this.field.getTerrain().name === "Grassy Terrain") {
+				move.priority = 2;
+			}
+		},
```
urkerab, line 7474: Priority is calculated in `getActionSpeed` which happens before `onModifyMove`.

(I assume here that you can switch in a Grassy Surge user and get priority on your ally's Grassy Glide.)

data/moves.ts
```diff
@@ -15785,6 +15969,32 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
...
+		category: "Special",
+		desc: "This move becomes a physical attack if the target's Defense is lower than its Special Defense, including stat stage changes. This move has a 10% chance to poison the target.",
+		shortDesc: "Physical if target's Def > Sp. Def. 10% poison chance.",
+		name: "Shell Side Arm",
+		// NOT CONFIRMED
+		pp: 5,
+		priority: 0,
+		flags: {},
+		onModifyMove(move, pokemon, target) {
+			if (target.getStat('def', false, true) > pokemon.getStat('spd', false, true)) move.category = 'Physical';
```
urkerab, line 15988:
```suggestion
 			if (target.getStat('def', false, true) > target.getStat('spd', false, true)) move.category = 'Physical';
```

data/moves.ts
KrisXV: We don’t know until we can test. I’ll make a change when we have more info.

data/moves.ts
KrisXV: Nah, just an instance of me forgetting assist (and company) exist. I’ll fix when I’m in a computer

data/moves.ts
KrisXV: I believe i did it correctly, do you think I should move the move check inside of the `if` statement but return before `zBrokeProtect`?

urkerab: I'd better wait until I know the exact mechanics.

data/moves.ts
```diff
@@ -1992,6 +1993,29 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
...
+		flags: {protect: 1, mirror: 1, mystery: 1},
+		secondary: {
+			chance: 100,
+			onHit(target, source, move) {
+				if (target?.statsRaisedThisTurn) {
+					target.trySetStatus('brn', source, move);
+				}
+			},
+		},
+		target: "allAdjacent",
```
urkerab, line 2015:
```suggestion
		target: "allAdjacentFoes",
```

data/moves.ts
```diff
@@ -2709,6 +2752,31 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		type: "Steel",
 		contestType: "Cool",
 	},
+	corrosivegas: {
+		num: 810,
+		accuracy: 100,
+		basePower: 0,
+		category: "Status",
+		// Needs research for Sticky Hold interaction
+		desc: "If the target is holding an item that can be removed from it, ignoring the Sticky Hold Ability. If the user has not fainted, the target loses its held item. This move cannot cause Pokemon with the Sticky Hold Ability to lose their held item or cause a Kyogre, a Groudon, a Giratina, an Arceus, a Genesect, a Silvally, a Zacian, or a Zamazenta to lose their Blue Orb, Red Orb, Griseous Orb, Plate, Drive, Memory, Rusted Sword, or Rusted Shield respectively. Items lost to this move cannot be regained with Recycle or the Harvest Ability.",
```
urkerab, line 2761: Note that you've got the remnants of the Knock Off power boost message here.

data/moves.ts
```diff
@@ -4671,6 +4756,32 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
...
+		basePower: 80,
+		category: "Special",
+		desc: "If the current active terrain is Psychic Terrain, this move's base power is boosted, and this move hits all opposing Pokemon.",
+		shortDesc: "Psychic Terrain:Boosted power+hits all foes.",
+		name: "Expanding Force",
+		pp: 20,
+		priority: 0,
+		flags: {protect: 1, mirror: 1},
+		onBasePower(basePower) {
+			if (this.field.getTerrain().id === 'psychicterrain') {
```
urkerab, line 4771:
```suggestion
			if (this.field.isTerrain('psychicterrain')) {
```

(similarly in other places where you check for a single terrain)

data/moves.ts
```diff
@@ -17445,6 +17964,25 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
...
+		category: "Physical",
+		desc: "Ends the effects of Electric Terrain, Grassy Terrain, Misty Terrain, and Psychic Terrain. This move fails if there is not a terrain active.",
+		shortDesc: "Fails if no terrain. Ends terrain.",
+		name: "Steel Roller",
+		pp: 5,
+		priority: 0,
+		flags: {contact: 1, protect: 1, mirror: 1},
+		onPrepareHit() {
+			if (!this.field.getTerrain()) return false;
+			this.field.clearTerrain();
```
urkerab, line 17980: Should this still clear the terrain if it subsequently fails to execute for some reason?

KrisXV: I'm not sure I understand what you mean

urkerab: `onPrepareHit` runs very early in the lifetime of a move, before such things as checking for Protect. It's possible that the terrain doesn't get cleared if the move doesn't hit its target.

KrisXV: oh true

data/moves.ts
urkerab: I've changed my mind on this.

I now think the protection breaking code should reset the broke through protect flag.

data/moves.ts
urkerab: Your latest code is now what I was originally expecting, so that's all good.

PR #6843, opened 2020-06-16
Trivia: Fix bug with computing user ranking

AnnikaCodes:
This is the bug: 
image


Zarel:
I'll be honest, I don't understand this code at all, but I'll trust it does what you say it does. Thanks for the fix!

PR #6844, opened 2020-06-16
Modlog: prevent non-gstaff from searching for IPs

mia-pi-git:
Currently, if you /modlog [ip] (as a non-gstaff user) it will show actions with that IP,  even though the IP is removed. EG: https://prnt.sc/t10k1r
This prevents anyone who isn't gstaff from searching _just_ numbers, since without it it will include IP results.
I'm not sure if there's a better way to do this with regex but this blocks all cases i can think of and doesn't hurt functionality.

### DIFF
--- a/server/chat-plugins/modlog.ts
+++ b/server/chat-plugins/modlog.ts
@@ -309,7 +309,7 @@ async function getModlog(
 	const addModlogLinks = !!(
 		Config.modloglink && (user.group !== ' ' || (targetRoom && targetRoom.settings.isPrivate !== true))
 	);
-	if (hideIps && /^[0-9.*]+$/.test(searchString)) {
+	if (hideIps && /^\[?[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\]?$/.test(searchString)) {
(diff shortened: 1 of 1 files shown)


Zarel:
Is this something we necessarily want to prevent?

server/chat-plugins/modlog.ts
```diff
@@ -309,7 +309,10 @@ async function getModlog(
 	const addModlogLinks = !!(
 		Config.modloglink && (user.group !== ' ' || (targetRoom && targetRoom.settings.isPrivate !== true))
 	);
-
+	if (hideIps && /^[0-9.*]+$/.test(searchString)) {
```
Zarel, line 312:
```suggestion
	if (hideIps && /^\[?[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\]?$/.test(searchString)) {
```

mia-pi-git:
> Is this something we necessarily want to prevent?

I would say yes, as this rather defeats the point of hiding IPs.

Zarel:
I disagree; "getting an IP" is different from "already having an IP and using it to get other information".

PR #6845, opened 2020-06-16
Rooms: save creation time for all rooms

mia-pi-git:
Creation time is set on creation of the room - eg restart / command creation. 
All rooms have uptime regardless now, and uptime is saved on settings for perm rooms.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -656,7 +656,8 @@ export const commands: ChatCommands = {
 	roomuptime(target, room, user, connection, cmd) {
 		if (!this.runBroadcast()) return;
 		if (!room) return this.errorReply(`Can only be used in a room.`);
-		// legacy handling
+
+		// for hotpatching
 		if (!room.settings.creationTime) room.settings.creationTime = Date.now();
 		const uptime = Chat.toDurationString(Date.now() - room.settings.creationTime);
 		this.sendReplyBox(`Room uptime: <b>${uptime}</b>`);
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -655,6 +655,7 @@ export const commands: ChatCommands = {
 	groupchatuptime: 'roomuptime',
 	roomuptime(target, room, user, connection, cmd) {
 		if (!this.runBroadcast()) return;
+		if (!room) return this.errorReply(`Can only be used in a room.`);
 		// legacy handling
 		if (!room.settings.creationTime) room.settings.creationTime = Date.now();
 		const uptime = Chat.toDurationString(Date.now() - room.settings.creationTime);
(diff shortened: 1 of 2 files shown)


server/chat-commands/room-settings.ts
```diff
@@ -704,14 +704,16 @@ export const commands: ChatCommands = {
...
-	groupchatuptime(target, room, user) {
-		if (!room || !room.creationTime) return this.errorReply("Can only be used in a groupchat.");
+	'!roomuptime': true,
+	groupchatuptime: 'roomuptime',
+	roomuptime(target, room, user, connection, cmd) {
 		if (!this.runBroadcast()) return;
-		const uptime = Chat.toDurationString(Date.now() - room.creationTime);
-		this.sendReplyBox(`Groupchat uptime: <b>${uptime}</b>`);
+		// legacy handling
+		if (!room.settings.creationTime) room.settings.creationTime = Date.now();
```
Zarel, line 712: This will crash if `room` is `null`.

server/rooms.ts
```diff
@@ -100,6 +100,7 @@ export interface RoomSettings {
 	staffMessage?: string | null;
 	rulesLink?: string | null;
 	dataCommandTierDisplay?: 'tiers' | 'doubles tiers' | 'numbers';
+	creationTime?: number;
```
Zarel, line 103: This isn't optional.

server/rooms.ts
```diff
@@ -1686,6 +1687,7 @@ export const Rooms = {
 	},
 	createChatRoom(roomid: RoomID, title: string, options: AnyObject) {
 		if (Rooms.rooms.has(roomid)) throw new Error(`Room ${roomid} already exists`);
+		if (!options.creationTime) options.creationTime = Date.now();
```
Zarel, line 1690: This should be filled in the same place other options are filled in (inside the `BasicChatRoom` constructor).

server/rooms.ts
```diff
@@ -102,6 +102,7 @@ export interface RoomSettings {
 	dataCommandTierDisplay?: 'tiers' | 'doubles tiers' | 'numbers';
 	requestShowEnabled?: boolean | null;
 	showEnabled?: GroupSymbol | true;
+	creationTime?: number | null;
```
Zarel, line 105: ...this is still not optional. It should also go at the top with the other non-optional properties.

server/chat-commands/room-settings.ts
```diff
@@ -704,14 +704,17 @@ export const commands: ChatCommands = {
...
-		if (!room || !room.creationTime) return this.errorReply("Can only be used in a groupchat.");
+	'!roomuptime': true,
+	groupchatuptime: 'roomuptime',
+	roomuptime(target, room, user, connection, cmd) {
 		if (!this.runBroadcast()) return;
-		const uptime = Chat.toDurationString(Date.now() - room.creationTime);
-		this.sendReplyBox(`Groupchat uptime: <b>${uptime}</b>`);
+		if (!room) return this.errorReply(`Can only be used in a room.`);
+		// legacy handling
+		if (!room.settings.creationTime) room.settings.creationTime = Date.now();
```
Zarel, line 713:
```suggestion

		// for hotpatching
		if (!room.settings.creationTime) room.settings.creationTime = Date.now();

```

PR #6846, opened 2020-06-17
Fix a massive exploit in /requestshow

sparkychildcharlie:
bad kitten

This would be really really horrible to have on the server because it would allow people to actually target staff.

1. they could inject something malicious there and make your computer download malicious files
2. they can grab the IPs of all the staff on within a room by using an image hosted on a server they own. (sounds like the perfect revenge story)

Also fixes the "access denied" message that would show when you use this.can instead of user.can whenever the user doesn't have the permission, as well as approve showing the approver's name, instead of the requester's name


Zarel:
Whoops, I didn't realize you PRed this when I pushed 32b5b43244c1c8b7

PR #6847, opened 2020-06-17
Fix server crash from overlength posts

sparkychildcharlie:
- server will crash and be unable to handle any further inputs from users.   Got this error when i accidentally posted a super long HTML post that exploited /requestshow

Byproduct of https://github.com/smogon/pokemon-showdown/pull/6745


Zarel:
Good catch, I guess we probably have an `any` somewhere in TS that made this non-automatic.

PR #6848, opened 2020-06-17
Youtube: Don't crash if key isn't configured

mia-pi-git:


Zarel:
Honestly this does feel like the perfect use-case for `throw UserError`

PR #6858, opened 2020-06-18
Youtube: fix ID crash 

mia-pi-git:
This ensures ID exists when checking if it includes. It already handles if this returns undefined, so no other changes should need to be made.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -114,8 +114,8 @@ export class YoutubeInterface {
 		} else {
 			id = link.split('channel/')[1];
 		}
-		if (id.includes('&')) id = id.split('&')[0];
-		if (id.includes('?')) id = id.split('?')[0];
+		if (id?.includes('&')) id = id.split('&')[0];
+		if (id?.includes('?')) id = id.split('?')[0];
 		return id;
 	}
 	async generateVideoDisplay(link: string) {
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -114,8 +114,8 @@ export class YoutubeInterface {
 		} else {
 			id = link.split('channel/')[1] || '';
 		}
-		if (id.includes('&')) id = id.split('&')[0] || '';
-		if (id.includes('?')) id = id.split('?')[0] || '';
+		if (id.includes('&')) id = id.split('&')[0];
+		if (id.includes('?')) id = id.split('?')[0];
 		return id;
 	}
 	async generateVideoDisplay(link: string) {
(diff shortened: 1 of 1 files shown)


Zarel:
Ah, I see, the problem is in the `[1]` `[3]` stuff. It seems like those should be replaced with `[1] || ''`.

Zarel:
It's best not to have TypeScript think `id` is type `string` if it isn't.

server/chat-plugins/youtube.ts
```diff
@@ -105,17 +105,17 @@ export class YoutubeInterface {
 		if (channelData[link]) return link;
 		if (!link.includes('channel')) {
 			if (link.includes('youtube')) {
-				id = link.split('v=')[1];
+				id = (link.split('v=')[1] || '');
```
Zarel, line 108:
```suggestion
				id = link.split('v=')[1] || '';
```

server/chat-plugins/youtube.ts
```diff
@@ -105,17 +105,17 @@ export class YoutubeInterface {
...
 			} else if (link.includes('youtu.be')) {
-				id = link.split('/')[3];
+				id = (link.split('/')[3] || '');
```
Zarel, line 110:
```suggestion
				id = link.split('/')[3] || '';
```

server/chat-plugins/youtube.ts
```diff
@@ -105,17 +105,17 @@ export class YoutubeInterface {
...
 			} else {
 				return null;
 			}
 		} else {
-			id = link.split('channel/')[1];
+			id = (link.split('channel/')[1] || '');
```
Zarel, line 115:
```suggestion
			id = link.split('channel/')[1] || '';
```

server/chat-plugins/youtube.ts
```diff
@@ -105,17 +105,17 @@ export class YoutubeInterface {
...
 		}
-		if (id.includes('&')) id = id.split('&')[0];
-		if (id.includes('?')) id = id.split('?')[0];
+		if (id?.includes('&')) id = (id.split('&')[0] || '');
+		if (id?.includes('?')) id = (id.split('?')[0] || '');
```
Zarel, line 118:
```suggestion
		if (id.includes('&')) id = id.split('&')[0] || '';
		if (id.includes('?')) id = id.split('?')[0] || '';
```

server/chat-plugins/youtube.ts
```diff
@@ -105,17 +105,17 @@ export class YoutubeInterface {
...
 				return null;
 			}
 		} else {
-			id = link.split('channel/')[1];
+			id = link.split('channel/')[1] || '';
 		}
-		if (id.includes('&')) id = id.split('&')[0];
-		if (id.includes('?')) id = id.split('?')[0];
+		if (id.includes('&')) id = id.split('&')[0] || '';
+		if (id.includes('?')) id = id.split('?')[0] || '';
```
Zarel, line 118:
```suggestion
		if (id.includes('&')) id = id.split('&')[0];
		if (id.includes('?')) id = id.split('?')[0];
```

Zarel:
Fun fact: `.split(...)[0]` is guaranteed to be a string.

PR #6860, opened 2020-06-18
Fix setting modchat in battles

mia-pi-git:
``Config.groupsranking[1]`` is now whitelist, not voice, so this needs to be changed.


mia-pi-git:
Although, maybe this should be `Config.groupsranking.indexOf('+')``?
Seems easier for the future but i may be missing a usecase.

Zarel:
I'd worry about what happens if you don't have a `+` group in your config.

Zarel:
(That's the genius `Math.max` trick I thought of!)

Zarel:
Good catch, by the way!

PR #6864, opened 2020-06-18
Fix Cosmog's learnset

Sawsbuck:


data/learnsets.ts
```diff
@@ -80367,4 +80293,3 @@ export const BattleLearnsets: {[speciesid: string]: LearnsetData} = {
 		],
 	},
 };
```
KrisXV, line 80295:
```suggestion
};

```

Sawsbuck: For some reason, GitHub is just acting flat out strange in terms of just not allowing me to commit that. I wonder why. I'll get back to this when I'm fully awake since at the time of writing its 3am my time.

Zarel: I don't think two trailing lines is super necessary, so I'd say this is nbd.

KrisXV: Oh, I didn't realize there were two trailing lines initially

PR #6881, opened 2020-06-20
Allow the unova sword trio in gen 8 VGC

HoeenCoder:
@DaWoblefet said these should be VGC legal even though they aren't obtainable in gen 8. Not sure if this is the best way to implement this.

### DIFF
--- a/data/learnsets.ts
+++ b/data/learnsets.ts
@@ -58535,6 +58535,8 @@ export const BattleLearnsets: {[speciesid: string]: LearnsetData} = {
 			{generation: 5, level: 65, shiny: 1, moves: ["sacredsword", "swordsdance", "quickguard", "workup"]},
 			{generation: 6, level: 50, shiny: 1, moves: ["retaliate", "ironhead", "sacredsword", "swordsdance"]},
 			{generation: 7, level: 60, shiny: 1, moves: ["sacredsword", "swordsdance", "quickattack", "ironhead"]},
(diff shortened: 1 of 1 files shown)


TheImmortal:
How are they legal if they aren’t obtainable? We don’t support hacks on standard ladders...

Zarel:
It sounds like it's possible but we'll need more details from Wob about exactly how it's possible.

DaWoblefet:
OK so there's a new mechanic in game where an NPC is able to add a so-called "battle-ready" symbol to the Pokemon. What it does is completely wipes the learnset (egg moves, past gen tutor moves, event moves, everything) and replaces it with a copy of moves that it could learn in its natural level-up movepool at the Pokemon's current level. What this effectively does is allow any past-gen Pokemon to participate in VGC, and the rules document clarifies that this applies to the Unova trio specifically: https://assets.pokemon.com//assets/cms2/pdf/play-pokemon/rules/play-pokemon-vg-rules-formats-and-penalty-guidelines-06172020-en.pdf

Basically, it's a way to pretend you caught your pastgen Pokemon in SwSh in the wild somewhere, or received it as an event or something. I asked HoeenHero to add an event because this is the easiest way to get it in the system. The proper refactor would be "all past-generation Pokemon are allowed, so long as they're obtainable in SwSh by any means and not on the banlist, but only Gen 8 learnsets are allowed".

PR #6883, opened 2020-06-20
Support customizing permissions per-room

mia-pi-git:
Edit: Fuck, i opened this too early.
Edit 2: what i meant to have this say if i hadn't been dumb:
Permissions are now checked within a class inside of user-groups. Each room can set custom permissions (held in a new JSON file) for each command.
`CommandContext#can` will pass in the command name if it's within the approved editable commands - all of which are room-only. 
Also fixes an issue within `Auth.atLeast` that wouldn't check superseding global auth, since that's needed to optimize this.
Internal checking names (`canTarget`, `canRoom`, and `canGlobal`) are tentative. 
(personally, I think `canTargetUser`, `allowedInRoom`, and `allowedGlobally` are probably clearer but i thought I'd leave it open to others.)


server/chat-commands/room-settings.ts
```diff
@@ -290,7 +292,54 @@ export const commands: ChatCommands = {
...
+
+		view(target, room, user) {
+			return this.parse(`/join view-permissions-${room.roomid}`);
+		},
 
+		help: '',
+		''(target, room, user) {
+			const configGroups = ["showmedia", "ban", "mute", "alts", "modlog", "broadcast", "declare",
+				"announce", "modchat", "tournaments", "gamemoderation", "gamemanagement", "minigame", "game",
+			];
```
KrisXV, line 333:
```suggestion
			const configGroups = [
				"ban", "mute", "alts", "modlog", "broadcast", "declare", "announce", "modchat",
				"tournaments", "gamemoderation", "gamemanagement", "minigame", "game",
			];
```

server/chat-commands/room-settings.ts
```diff
@@ -290,7 +292,54 @@ export const commands: ChatCommands = {
...
+			let buffer = `<strong>Room permissions help:</strong><hr> `;
+			buffer += `<strong>Usage:</strong><code> /permissions set [permission], [rank]</code><br>`;
+			buffer += `<strong>Usable permissions:</strong><br>`;
+			buffer += Chat.getReadmoreCodeBlock(Permissions.approvedPermissions(room).join(', '));
+			buffer += `<br><strong>Command groups: </strong>${Chat.getReadmoreCodeBlock(configGroups.join(', '))}<br>`;
```
KrisXV, line 338:
```suggestion
			let buffer = `<strong>Room permissions help:</strong><hr />`;
			buffer += `<strong>Usage:</strong><code> /permissions set [permission], [rank]</code><br />`;
			buffer += `<strong>Usable permissions:</strong><br />`;
			buffer += Chat.getReadmoreCodeBlock(Permissions.approvedPermissions(room).join(', '));
			buffer += `<br /><strong>Command groups: </strong>${Chat.getReadmoreCodeBlock(configGroups.join(', '))}<br />`;
```

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,49 @@ export const roomSettings: SettingsHandler[] = [
...
+		if (!rank) {
+			let buf = `<div class="pad"><center><strong>Groups on ${this.room.title}:</strong><hr/ >`;
+			for (const r of Config.groupsranking.slice(2)) {
+				if (Config.groups[r]?.globalonly) continue;
+				const name = (Config.groups[r] ? Config.groups[r].name : r).replace(/ /, '');
+				buf += `<a href="/view-permissions-${room}-${Dashycode.encode(name)}" `;
+				buf += `target="_blank" rel="noopener"> <button style="border: 1px solid black ;`;
+				buf += ` border-radius: 30px ; background-color: #eeeeee ; height: 35px">${name}</button> </a><hr>`;
+			}
+			buf += `<strong>All possible permissions for this room:</strong><br>`;
```
KrisXV, line 1528:
```suggestion
				buf += ` border-radius: 30px ; background-color: #eeeeee ; height: 35px">${name}</button> </a><hr />`;
			}
			buf += `<strong>All possible permissions for this room:</strong><br />`;
```

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,49 @@ export const roomSettings: SettingsHandler[] = [
...
+			buf += `${Permissions.approvedPermissions(this.room).join(', ')}`;
+			return buf.replace(/<a roomid="/g, `<a target="replace" href="/`);
+		}
+		rank = Dashycode.decode(rank).toLowerCase();
+		let buf = `<div class="pad"><p>` +
+		`<a roomid="view-permissions-${room}">◂ All ranks</a> / ` +
+		`<strong>${u(rank)}</strong></p><hr />`;
```
KrisXV, line 1535:
```suggestion
		let buf = `<div class="pad">`;
		buf += `<p><a roomid="view-permissions-${room}">◂ All ranks</a> / <strong>${u(rank)}</strong></p><hr />`;
```

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,49 @@ export const roomSettings: SettingsHandler[] = [
...
+		let symbol = '';
+		for (const s of Config.groupsranking) {
+			if (toID(Config.groups[s]?.name) === rank) symbol = s;
+		}
+		buf += `Custom room permissions: `;
+		buf += `<div class="ladder pad"><table><tr><th>Permission</th><th>Rank</th></tr><tr>`;
```
KrisXV, line 1541:
```suggestion
		buf += `<div class="ladder pad"><table><tr><th>Permission</th><th>Rank</th></tr>`;
```

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,49 @@ export const roomSettings: SettingsHandler[] = [
...
+		const perms = Permissions.getPermissions(this.room);
+		for (const key in perms) {
+			if (Config.groups[symbol]?.rank >= (Config.groups[perms[key]].rank || -1)) {
+				buf += `<tr><td>${u(key)}</td> <td>${perms[key]}</small></td></tr>`;
```
KrisXV, line 1545:
```suggestion
				buf += `<tr><td>${u(key)}</td><td><small>${perms[key]}</small></td></tr>`;
```

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,49 @@ export const roomSettings: SettingsHandler[] = [
...
+			}
+		}
+		buf += `</td></tr></table></div>`;
```
KrisXV, line 1548:
```suggestion
		buf += `</table></div>`;
```

server/chat.ts
```diff
@@ -1619,6 +1622,15 @@ export const Chat = new class {
 		return `${num}${space}${num > 1 ? pluralSuffix : singular}`;
 	}
 
+	/**
+	 * Takes the name of a command and gets the base command, if there is one.
+	 * @param cmd string
```
KrisXV, line 1627:
```suggestion
```

server/chat.ts
```diff
@@ -1619,6 +1622,15 @@ export const Chat = new class {
...
+	 */
+	baseCommand(cmd: string) {
+		if (typeof this.commands[cmd] === 'string') return this.commands[cmd] as string;
```
KrisXV, line 1630:
```suggestion
		if (typeof this.commands[cmd] === 'string') return this.commands[cmd];
```
This should be implied by the if statement immediately beforehand

mia-pi-git: Not implied, I presume since the typing of chat commands allows for it to be several things, so although the actual code ensures it will always be a string, typescript doesn't acknowledge that `Chat.commands[alias]` will be a string.

server/users.ts
```diff
@@ -522,22 +522,8 @@ export class User extends Chat.MessageContext {
...
-		const targetGroup = target ? auth.get(target) : undefined;
-
-		const roomIsTemporary = room && !room.persist;
-		if (roomIsTemporary && group === this.group) {
-			const replaceGroup = Auth.getGroup(group).globalGroupInPersonalRoom;
-			if (replaceGroup) group = replaceGroup;
-		}
-
-		return Auth.hasPermission(group, permission, targetGroup, target === this);
+		if (!room) room = Rooms.global;
```
Zarel, line 525: :(

tbh I wrote #6885 just because I really didn't like this line.

Why did you write it, anyway? What is this intended to accomplish?

server/chat-commands/room-settings.ts
```diff
@@ -290,7 +292,55 @@ export const commands: ChatCommands = {
...
+			`If no [rank] is given, resets the needed permission to the default. Requires: # &`,
+		],
+
+		view(target, room, user) {
+			return this.parse(`/join view-permissions-${room.roomid}`);
+		},
 
+		help: '',
+		''(target, room, user) {
+			const configGroups = [
```
Zarel, line 331: This should probably be a global at the top of the file, or even exported from `user-groups`.

Zarel: Actually, `Permissions.approvedPermissions` already exists. Why is this separated from that?

mia-pi-git: this is separated so it can be made clear that those permissions serve to group permissions, as opposed to configuring just one command.

server/users.ts
mia-pi-git: This, as i recall, was intended to support changing global perms in the future.

server/chat-commands/room-settings.ts
mia-pi-git: I wanted it to be clear that say, `gamemanagement` will change multiple things, as opposed to just, say, `notifyrank`.

server/users.ts
mia-pi-git: I'm also fairly certain it was also partially because I needed a room to be passed and didn't want to have it pass a specific room if it wasn't a room-only command, so I decided; two birds one stone.

Zarel: Global permissions should be associated with `Users.globalAuth` (support passing an `auth` object), and `room` should be nullable. That would be a better solution to your problems.

server/chat-commands/room-settings.ts
Zarel: Even if they're different in that way, they should still be in the same part of the code and named similarly, because they serve similar functions.

Zarel: It's bad to have two drastically different implementations for what are conceptually the same thing.

mia-pi-git: I mean, the config groups _are_ in `Permissions.approvedPermssions` already, which is where they're needed, this is just using an extra array so people know which permissions do that. I don't see how they're drastically different.

server/users.ts
Zarel: This is not resolved. You should not consider `Rooms.global` a room.

Zarel:
This code seems to have tons of bugs around sysop permissions, which makes it really hard for me to test. I've found a lot of random bugs and the code is just really messy and impossible to read, but I get the general sense that it duplicates a lot of PS's current features but makes it much harder to understand how the code works than it currently does...

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,48 @@ export const roomSettings: SettingsHandler[] = [
...
+			return char + str;
+		};
+		if (!room) return `<h2>This room does not exist or does not support permissions.</h2>`;
+		if (!user.authAtLeast('%', this.room)) return `<h2>Access denied.</h2>`;
+		if (!rank) {
+			let buf = `<div class="pad"><center><strong>Groups on ${this.room.title}:</strong><hr/ >`;
+			for (const r of Config.groupsranking.slice(2)) {
+				if (Config.groups[r]?.globalonly) continue;
+				const name = (Config.groups[r] ? Config.groups[r].name : r).replace(/ /, '');
+				buf += `<a href="/view-permissions-${room}-${Dashycode.encode(name)}" `;
```
Zarel, line 1524: These buttons are pointless and just make it harder to see what permissions exist. They should all be on one page.

server/chat-commands/room-settings.ts
```diff
@@ -1454,3 +1503,48 @@ export const roomSettings: SettingsHandler[] = [
...
+				buf += `target="_blank" rel="noopener"> <button style="border: 1px solid black ;`;
+				buf += ` border-radius: 30px ; background-color: #eeeeee ; height: 35px">${name}</button> </a><hr />`;
```
Zarel, line 1526: I don't understand this design, either. What's wrong with `class="button"`?

package.json
```diff
@@ -5,6 +5,7 @@
   "version": "0.11.2",
   "dependencies": {
     "probe-image-size": "^5.0.0",
+    "prompt-sync": "^4.2.0",
```
Zarel, line 8: Was this a mistake?

Zarel: You don't really need this, btw. `await Streams.stdin.readLine()` will do the same thing; just wrap your script in an async function.

server/user-groups.ts
```diff
@@ -246,3 +287,42 @@ export class GlobalAuth extends Auth {
...
+		if (!rank) {
+			delete permissions[permission];
+		} else {
+			const ALLOWED_COMMANDS = Permissions.approvedPermissions(room);
+			permission = Chat.baseCommand(permission);
+			if (!ALLOWED_COMMANDS.includes(toID(permission))) return false;
+			permissions[permission] = rank;
+		}
+		room.settings.permissions = permissions;
+		room.saveSettings()
```
Zarel, line 322:
```suggestion
		room.saveSettings();
```

Zarel:
OK, so this is nearly complete (although it'll need one more merge).

The main problem left is probably its detection of which commands can be overridden.

server/user-groups.ts
```diff
@@ -246,3 +287,42 @@ export class GlobalAuth extends Auth {
...
+		const ALLOWED_COMMANDS = [
+			"showmedia", "broadcast", "tournaments", "gamemoderation", "gamemanagement", "minigame", "game",
+		];
+		for (const cmd in Chat.commands) {
+			const entry = Chat.commands[cmd];
+			if (typeof entry !== 'function') continue;
+			// remove room-plugin commands unless it is that room
+			if (entry.roomSpecific && entry.roomSpecific !== room?.roomid) continue;
+			// Assume if it passes a room into CommandContext.can that it's roomonly
+			if (Chat.commands[cmd].toString().includes('room)) return false;')) {
```
Zarel, line 304: `room)) return false;` is a pretty awkward check tbh...

mia-pi-git: Perhaps ``` /(this|user).can\((.+), (.+), (.+)\)/.test()```?

Zarel: It would still be weird, but assuming `.` doesn't include newlines, it might be good enough.

I would remove the `(this|user)` part, though; and also your first `.` needs to be escaped.

server/user-groups.ts
```diff
@@ -246,3 +287,42 @@ export class GlobalAuth extends Auth {
 		return true;
 	}
 }
+
+export const Permissions = new class {
+	approvedPermissions(room: Room | null = null) {
+		// support some config permission groups that aren't already a command name - these need to be hardcoded
+		// since there's no good way to automate this
+		const ALLOWED_COMMANDS = [
+			"showmedia", "broadcast", "tournaments", "gamemoderation", "gamemanagement", "minigame", "game",
```
Zarel, line 296: This part is still wrong; I still need to show you how to automate this.

server/user-groups.ts
```diff
@@ -246,3 +287,42 @@ export class GlobalAuth extends Auth {
...
+		];
+		for (const cmd in Chat.commands) {
+			const entry = Chat.commands[cmd];
+			if (typeof entry !== 'function') continue;
+			// remove room-plugin commands unless it is that room
+			if (entry.roomSpecific && entry.roomSpecific !== room?.roomid) continue;
+			// Assume if it passes a room into CommandContext.can that it's roomonly
+			if (/\.can\((.+), (.+), (.+)\)/.test(Chat.commands[cmd].toString())) {
+				ALLOWED_COMMANDS.push(cmd);
+			}
```
Zarel, line 306: Also wasn't this supposed to be moved into `annotateCommands`?

Zarel:
Can I get you to merge master? I backported some of the changes I made here out there, because there was a permission bug that one of them fixed. Also, it should make the diff here smaller, and thus easier to review.

server/users.ts
```diff
@@ -533,11 +535,11 @@ export class User extends Chat.MessageContext {
 		const auth = (room && !this.can('makeroom') ? room.auth.get(this.id) : this.group);
 		return auth in Config.groups && Config.groups[auth].rank >= Config.groups[minAuth].rank;
 	}
-	can(permission: RoomPermission, target: User | null, room: BasicRoom): boolean;
+	can(permission: RoomPermission, target: User | null, room: Room | BasicChatRoom, cmd?: string): boolean;
 	can(permission: GlobalPermission, target?: User | null): boolean;
 	can(permission: RoomPermission & GlobalPermission, target: User | null, room?: BasicRoom | null): boolean;
 	can(permission: string, target: User | null = null, room: BasicRoom | null = null): boolean {
```
Zarel, line 541:
```suggestion
	can(permission: string, target: User | null = null, room: BasicRoom | null = null, cmd?: string): boolean {
```

server/users.ts
```diff
@@ -533,11 +535,11 @@ export class User extends Chat.MessageContext {
 		const auth = (room && !this.can('makeroom') ? room.auth.get(this.id) : this.group);
 		return auth in Config.groups && Config.groups[auth].rank >= Config.groups[minAuth].rank;
 	}
-	can(permission: RoomPermission, target: User | null, room: BasicRoom): boolean;
+	can(permission: RoomPermission, target: User | null, room: Room | BasicChatRoom, cmd?: string): boolean;
```
Zarel, line 538:
```suggestion
	can(permission: RoomPermission, target: User | null, room: Room | BasicRoom, cmd?: string): boolean;
```

I'm surprised why this isn't an error, because `BasicChatRoom` doesn't exist anymore.

server/chat.ts
```diff
@@ -814,7 +814,7 @@ export class CommandContext extends MessageContext {
 	can(permission: RoomPermission, target: User | null, room: Room): boolean;
 	can(permission: GlobalPermission, target?: User | null): boolean;
 	can(permission: string, target: User | null = null, room: Room | null = null) {
-		if (!this.user.can(permission as any, target, room as any)) {
+		if (!this.user.can(permission as any, target, room!, this.cmd)) {
```
Zarel, line 817:
```suggestion
		if (!Users.Auth.hasPermission(this.user, permission, target, room, cmd)) {
```

I wonder...

server/chat-commands/room-settings.ts
```diff
@@ -294,7 +295,57 @@ export const commands: ChatCommands = {
...
+			this.modlog(`SETPERMISSION`, null, `${perm}: ${rank}`);
+			return this.privateModAction(`(${user.name} set the required rank for ${perm} to ${rank}.)`);
+		},
+		sethelp: [
+			`/setpermission [command, rank] - sets the required permission to use the command [command] to [rank].`,
+			`If no [rank] is given, resets the needed permission to the default. Requires: # &`,
+		],
+
+		view(target, room, user) {
+			return this.parse(`/join view-permissions-${room.roomid}`);
```
Zarel, line 330:
```suggestion
			if (!room) return this.requiresRoom();
			return this.parse(`/join view-permissions-${room.roomid}`);
```

server/user-groups.ts
```diff
@@ -108,9 +109,32 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
...
+				if (!auth.atLeast(user, roomPermissions[permission])) return false;
+				jurisdiction = 'su';
+			}
+		}
 
 		return Auth.hasJurisdiction(symbol, jurisdiction, targetSymbol, target === user);
 	}
+	static supportedRoomPermissions(room: Room | null = null) {
+		// support some config permission groups that aren't already a command name - these need to be hardcoded
+		// since there's no good way to automate this
```
Zarel, line 127:
```suggestion
```

(These are no longer hardcoded.)

server/user-groups.ts
```diff
@@ -92,7 +92,8 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 		user: User,
 		permission: string,
 		target: User | GroupSymbol | null,
-		room?: Room | BasicRoom | null
+		room?: Room | BasicRoom | null,
```
Zarel, line 95:
```suggestion
		room?: BasicRoom | null,
```

(`Room` is a subtype of `BasicRoom`, so you don't need both)

PR #6886, opened 2020-06-20
Rising Voltage boost applies if target is grounded

tmagicturtle:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-12#post-8510904

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -15174,8 +15174,8 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		pp: 20,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
-		onBasePower(basePower) {
-			if (this.field.isTerrain('electricterrain')) {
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

data/moves.ts
```diff
@@ -15174,8 +15174,8 @@ export const BattleMovedex: {[moveid: string]: MoveData} = {
 		pp: 20,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
-		onBasePower(basePower) {
-			if (this.field.isTerrain('electricterrain')) {
+		onBasePower(basePower, pokemon, target) {
+			if (this.field.isTerrain('electricterrain') && !target.runImmunity('Ground')) {
```
KrisXV, line 15178: Shouldn't you just use `target.isGrounded()`?

PR #6887, opened 2020-06-20
Prevent empty string from being added to banwords

Nixola:
Filter out empty strings from the `banword add` and `banword delete`
arguments. Added a check in `banword add` for an empty array of
arguments.

Thought it could be useful after today's Video Games room issue.


Zarel:
Thanks!

PR #6889, opened 2020-06-20
Show status immunities in /weak output

urkerab:
I saw this feature request somewhere although I forget where now.


Zarel:
Sounds useful.

PR #6891, opened 2020-06-21
Youtube: ensure video data exists in the response

mia-pi-git:
Not sure how this hasn't been tripped before. The first item holds the video data, so without that you get an issue.


server/chat-plugins/youtube.ts
```diff
@@ -35,7 +35,7 @@ export class YoutubeInterface {
 		const queryUrl = `${CHANNEL}?part=snippet%2Cstatistics&id=${encodeURIComponent(id)}&key=${Config.youtubeKey}`;
 		const raw = await Net(queryUrl).get();
 		const res = JSON.parse(raw);
-		if (!res || !res.items) return null;
+		if (!res || !res.items || res.items.length > 1) return;
```
Zarel, line 38: Do you mean `>= 1`? Or is the first item not video data?

PR #6892, opened 2020-06-21
Monotype: Ban Magearna

KrisXV:
https://www.smogon.com/forums/threads/magearna-is-now-banned-from-monotype.3666296/


Zarel:
Thanks!

PR #6895, opened 2020-06-21
Fix Rising Voltage

tmagicturtle:
Sorry, the most recent change I did to this (which I weirdly based off of Rototiller) actually made it so that the terrain boost would never occur. As Kris suggested, this is the correct way to handle it.


urkerab:
The typical use case of `runImmunity` is to find out whether a move affects a Pokémon, in which case its return value of `true` to mean "yes, the move hits" makes sense. In this case, `runImmunity('Ground')` actually returns `true` if the target is grounded, so you didn't want to negate the result. (But Kris's `isGrounded` is obviously better anyway.)

PR #6901, opened 2020-06-22
Update tiering

KrisXV:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/#post-8513440


Zarel:
tyty

PR #6902, opened 2020-06-22
Ban Indeedee-M from RU

Sawsbuck:
https://www.smogon.com/forums/threads/np-ru-stage-1-i-lost-something-in-the-hills-indeedee-m-banned.3659533/page-6#post-8513573


Zarel:
tyty

PR #6911, opened 2020-06-24
Add st as servertime alias

TheMezStrikes:
suggested at https://www.smogon.com/forums/threads/abbreviate-servertime.3666028/ seems simple and harmless


Zarel:
Looks good, thanks!

PR #6913, opened 2020-06-24
Fix multiple bugs in events

TheMezStrikes:
Pretty self-explanatory, `/events remove Test Event` didn't remove the event because Test Event isn't an ID. Now it uses eventID

Edit: found a second bug, added a commit to fix.

### DIFF
--- a/server/chat-plugins/room-events.ts
+++ b/server/chat-plugins/room-events.ts
@@ -245,7 +245,7 @@ export const commands: ChatCommands = {
 				return this.errorReply(`There is no event titled '${target}'. Check spelling?`);
 			}
 
-			delete room.settings.events[target];
+			delete room.settings.events[eventID];
 			for (const alias of getAliases(room, eventID)) {
 				delete room.settings.events[alias];
 			}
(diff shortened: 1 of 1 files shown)


server/chat-plugins/room-events.ts
```diff
@@ -522,6 +522,7 @@ export const commands: ChatCommands = {
 			}
 
 			// rebuild the room.settings.events object
+			room.settings.events = {};
```
Zarel, line 525: This looks like it would clear all aliases and categories out of the event table, which would be bad.

TheMezStrikes: You're right, I forgot that table contained more than just the raw events. 
Would you recommend adding `delete room.settings.events[eventId]` on the line before `room.settings.events[eventId] = sortedObj;` instead? That'd work but I don't know enough about optimization to know if that's good enough

Zarel: Yeah, I think that's a reasonably good solution.

Zarel:
Thanks!

PR #6918, opened 2020-06-25
Fix STABmons Move Legality

KrisXV:


data/rulesets.ts
```diff
@@ -927,6 +928,12 @@ export const BattleFormats: {[k: string]: FormatsData} = {
...
 						for (const formeid of baseSpecies.otherFormes) {
+							let skipForme = false;
+							if (baseSpecies.prevo) {
+								const prevo = dex.getSpecies(baseSpecies.prevo);
+								skipForme = prevo.evos.map(toID).includes(toID(formeid));
+							}
+							if (skipForme) continue;
 							const forme = dex.getSpecies(formeid);
 							if (forme.changesFrom === originalForme.name) {
 								types = types.concat(forme.types);
```
Zarel, line 939:
```suggestion
						for (const formeName of baseSpecies.otherFormes) {
							if (baseSpecies.prevo) {
								const prevo = dex.getSpecies(baseSpecies.prevo);
								if (prevo.evos.includes(formeName)) continue;
							}
							const forme = dex.getSpecies(formeName);
							if (forme.changesFrom === originalForme.name) {
								types = types.concat(forme.types);
```

The `toID` mapping seems excessive...

KrisXV: I disagree. It's there to ensure that typos (such as a miscapitalized O) in the data file don't get in the way of validation.

KrisXV: In any case, I did this (locally) and pushed because the type of `Species#evos` was incorrect.

sim/dex-data.ts
```diff
@@ -587,9 +587,9 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 	/** Added type (used in OMs). */
 	readonly addedType?: string;
 	/** Pre-evolution. '' if nothing evolves into this Pokemon. */
-	readonly prevo: ID;
+	readonly prevo: string;
```
Zarel, line 590: Actually wait, did we switch these over?

Zarel: Looks like yes; whew.

data/rulesets.ts
Zarel: > It's there to ensure that typos (such as a miscapitalized O) in the data file don't get in the way of validation.

This is why I added data tests, so we wouldn't have to worry about stuff like this...

Zarel:
Thanks!

data/rulesets.ts
KrisXV: Oh yeah

PR #6922, opened 2020-06-28
Datasearch: support 'monotype' parameter in /ds

AnnikaCodes:
also piggybacked a fix for /as in here


server/chat-plugins/datasearch.ts
```diff
@@ -829,6 +837,12 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 			error: "No search parameters other than 'all' were found. Try '/help dexsearch' for more information on this command.",
 		};
 	}
+
+	if (singleTypeSearch && Object.values(searches)
+		.map(search => Object.keys(search.types).filter(type => search.types[type])[0])
+		.filter(type => type).length !== 1
+	) return {error: "A monotype search must include only one type."};
```
urkerab, line 844: I can't help thinking that there might be some combination that this excludes but is strictly legal, although I don't know how useful being able to search for e.g. monotypes that aren't normal or water is.

Zarel:
Thanks!

PR #6928, opened 2020-06-29
Adjust fractional priority for latest research

urkerab:
The main change is due to Anubis's research into Quick Draw and Quick Claw; the latter only gets a chance to activate if the former does not.

While I was there I noticed that the fractional priority is always `-0.1`, `0`, or `0.1`, so that the `Math.round()` is pointless which simplifies the code.

I also assumed that Custap Berry, should it ever be released, will activate in preference to Quick Draw.


Zarel:
Thanks!

PR #6929, opened 2020-06-29
Update german.json

xMarcell:
I updated some translated lines, in order to make it more cohesive and grammatically correct; also added missing parts of sentences and punctuations.


Zarel:
Thanks!

PR #6934, opened 2020-06-30
Add a command for bots to send HTML pages

mia-pi-git:
https://www.smogon.com/forums/threads/add-a-command-for-bots-to-send-users-html-rooms.3666630/#post-8521882

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -1127,7 +1127,7 @@ export const pages: PageTable = {
 		const [bot, pageid] = args;
 		const targetBot = Users.get(bot);
 		if (!targetBot) {
-			return `<div class="pad"><h2>User ${bot} not found, and so cannot be requested for this page.</h2></div>`;
+			return `<div class="pad"><h2>The bot "${bot}" is not available.</h2></div>`;
 		}
 		targetBot.send(`|requestpage|${user.name}|${pageid}`);
 	},
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -210,6 +210,26 @@ export const commands: ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
+	sendhtmlpage(target, room, user, connection) {
+		if (!this.can('addhtml', null, room)) return false;
+		let [targetID, pageid, content] = target.split(',') as [string, string, string | null];
+		if (!target || !pageid || !content) return this.parse(`/help sendhtmlpage`);
+		const targetUser = Users.get(targetID);
+		if (!targetUser) return this.errorReply(`User not found.`);
+		content = this.canHTML(content);
+		if (!content) return;
+		const context = new Chat.PageContext({
+			user: targetUser,
+			connection: targetUser.connections[0],
+			pageid: pageid,
+		});
+		context.title = `[${user.name}] ${pageid}`;
+		return context.send(content);
+	},
+	sendhtmlpagehelp: [
+		`/sendhtmlpage: [target], [page id], [html] - sends the [target] a HTML room with the HTML [content] and the [pageid]. Requires: s* # &`,
+	],
+
 	nick() {
 		this.sendReply(`||New to the Pokémon Showdown protocol? Your client needs to get a signed assertion from the login server and send /trn`);
 		this.sendReply(`||https://github.com/smogon/pokemon-showdown/blob/master/PROTOCOL.md#global-messages`);
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -15,7 +15,6 @@ import {FS} from '../../lib/fs';
 import {Utils} from '../../lib/utils';
 
 import * as ProcessManager from '../../lib/process-manager';
-import { brotliCompress } from 'zlib';
 
 export const commands: ChatCommands = {
 
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -15,6 +15,7 @@ import {FS} from '../../lib/fs';
 import {Utils} from '../../lib/utils';
 
 import * as ProcessManager from '../../lib/process-manager';
+import { brotliCompress } from 'zlib';
 
 export const commands: ChatCommands = {
 
@@ -212,7 +213,7 @@ export const commands: ChatCommands = {
 
 	sendhtmlpage(target, room, user, connection) {
 		if (!this.can('addhtml', null, room)) return false;
-		let [targetID, pageid, content] = target.split(',') as [string, string, string | null];
+		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2) as [string, string, string | null];
 		if (!target || !pageid || !content) return this.parse(`/help sendhtmlpage`);
 		const targetUser = Users.get(targetID);
 		if (!targetUser) return this.errorReply(`User not found.`);
@@ -229,7 +230,6 @@ export const commands: ChatCommands = {
 	sendhtmlpagehelp: [
 		`/sendhtmlpage: [target], [page id], [html] - sends the [target] a HTML room with the HTML [content] and the [pageid]. Requires: s* # &`,
 	],
-
 	nick() {
 		this.sendReply(`||New to the Pokémon Showdown protocol? Your client needs to get a signed assertion from the login server and send /trn`);
 		this.sendReply(`||https://github.com/smogon/pokemon-showdown/blob/master/PROTOCOL.md#global-messages`);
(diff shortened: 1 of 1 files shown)


Zarel:
Could you also implement `/join view-bot-userid-inputpageid`?

mia-pi-git:
> Could you also implement `/join view-bot-userid-inputpageid`?

I have some questions about how to approach this, I'll dm you on discord.

server/chat-commands/admin.ts
```diff
@@ -210,6 +210,26 @@ export const commands: ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
+	sendhtmlpage(target, room, user, connection) {
+		if (!this.can('addhtml', null, room)) return false;
+		let [targetID, pageid, content] = target.split(',') as [string, string, string | null];
```
Zarel, line 215:
```suggestion
		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2);
```

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,14 @@ export const commands: ChatCommands = {
...
 };
+
+export const pages: PageTable = {
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
+		if (!targetBot) {
+			return `<div class="pad"><h2>User ${bot} not found, and so cannot be requested for this page.</h2></div>`;
+		}
+		return targetBot.send(`|requestpage|${user.name}|${pageid}`);
```
Zarel, line 1132:
```suggestion
		targetBot.send(`|requestpage|${user.name}|${pageid}`);
```

The `return` is super misleading here.

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,14 @@ export const commands: ChatCommands = {
 		`[player] must be a username or number, [pokemon] must be species name or party slot number (not nickname), [move] must be move name.`,
 	],
 };
+
+export const pages: PageTable = {
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
+		if (!targetBot) {
+			return `<div class="pad"><h2>User ${bot} not found, and so cannot be requested for this page.</h2></div>`;
```
Zarel, line 1130:
```suggestion
			return `<div class="pad"><h2>The bot "${bot}" is not available.</h2></div>`;
```

(Maybe have a bot check so it can't be used to spam regular users?)

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,14 @@ export const commands: ChatCommands = {
...
 };
+
+export const pages: PageTable = {
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
+		if (!targetBot) {
+			return `<div class="pad"><h2>The bot "${bot}" is not available.</h2></div>`;
+		}
+		targetBot.send(`|requestpage|${user.name}|${pageid}`);
```
Zarel, line 1132: How is this going to affect regular users? That's probably worth a check?

Zarel: I think it adds some red text to Lobby, which is non-ideal.

mia-pi-git: I'll write a check in.

mia-pi-git: (looks like it just adds some ``code`` text to lobby)

Zarel: It should maybe be a PM?

Zarel: Yeah, a PM from the user; that would probably make it least confusing if a user accidentally gets it.

mia-pi-git: A pm seems more confusing? It's a lot more noticeable, not to mention that if you're not in lobby you won't see the `code` message, so it'd be harder to spam that. 
(also |requestpage| seems easier for bots to parse)

Zarel: You can PM `|requestpage|` to the bot.

I don't want to spam users in lobby either way. PMs can be ignored.

mia-pi-git: Do you want both of these to happen, or just one? 
I've written a check in to ensure that the user is a bot in at least one room (which i'm waiting to commit) -  would you prefer it also be a pm'd message? 
(personally i don't think it should be, since it'd be easier to parse for said bot, but that's just me)

Zarel: Checking that the user is a bot in at least one room is a good start, but I think it should also be a PM just in case.

It's still approximately as easy to parse for the bot, but it'd be a lot easier to figure out if someone makes a mistake somewhere.

Zarel:
I'll draft until the current problem with PMing is fixed.

I'm putting my foot down because I don't want to give people the power to bypass the `/ignore` command.

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,28 @@ export const commands: ChatCommands = {
...
+				canSend = true;
+				room = curRoom;
+			}
+		}
+		if (!canSend) {
+			return `<div class="pad"><h2>"${bot}" is not a bot.</h2></div>`;
+		}
+		targetBot.sendTo(
+			room ? room.roomid : 'lobby',
+			`|pm|${user.name}|${targetBot.name}|requestpage|${user.name}|${pageid}`
```
Zarel, line 1145:
```suggestion
			`|pm|${user.name}|${targetBot.name}||requestpage|${user.name}|${pageid}`
```

server/chat-commands/admin.ts
```diff
@@ -210,6 +210,25 @@ export const commands: ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
+	sendhtmlpage(target, room, user, connection) {
+		if (!this.can('addhtml', null, room)) return false;
+		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2) as [string, string, string | null];
```
Zarel, line 215:
```suggestion
		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2);
```

server/chat-commands/admin.ts
```diff
@@ -210,6 +210,25 @@ export const commands: ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
+	sendhtmlpage(target, room, user, connection) {
+		if (!this.can('addhtml', null, room)) return false;
+		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2);
+		if (!target || !pageid || !content) return this.parse(`/help sendhtmlpage`);
+		const targetUser = Users.get(targetID);
+		if (!targetUser) return this.errorReply(`User not found.`);
+		content = this.canHTML(content);
```
Zarel, line 219:
```suggestion
		content = this.canHTML(content)!;
```

whatever, we can deal with this in other ways.

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,28 @@ export const commands: ChatCommands = {
...
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
+		if (!targetBot) {
+			return `<div class="pad"><h2>The bot "${bot}" is not available.</h2></div>`;
+		}
+		let canSend = false;
+		let room;
+		for (const curRoom of Rooms.global.chatRooms) {
+			if (targetBot.authAtLeast('*', curRoom)) {
```
Zarel, line 1135: `authAtLeast` would let you spam ROs with this command... probably undesirable.

Zarel:
```suggestion
			if (curRoom.auth.get(targetBot) === '*') {
```

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,28 @@ export const commands: ChatCommands = {
...
 };
+
+export const pages: PageTable = {
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
+		if (!targetBot) {
+			return `<div class="pad"><h2>The bot "${bot}" is not available.</h2></div>`;
+		}
+		let canSend = false;
```
Zarel, line 1132:
```suggestion
		let canSend = Users.globalAuth.get(targetBot) === '*';
```

server/chat-commands/admin.ts
```diff
@@ -1102,3 +1121,28 @@ export const commands: ChatCommands = {
 		`[player] must be a username or number, [pokemon] must be species name or party slot number (not nickname), [move] must be move name.`,
 	],
 };
+
+export const pages: PageTable = {
+	bot(args, user) {
+		const [bot, pageid] = args;
+		const targetBot = Users.get(bot);
```
Zarel, line 1128:
```suggestion
		const [botid, pageid] = args;
		const bot = Users.get(botid);
```

That's not what `target` means...

Zarel:
Looks reasonable to me, at this point.

PR #6943, opened 2020-07-02
Refactor Pokemon tiers into arrays

KrisXV:
I dont know how to fix my commit history

also this is a draft since i havent done client-side changes yet


Zarel:
I still think this is a good idea, for the record.

But I suppose actually naming which category the tier is in would also be nice, but it would make some use-cases more complicated.

PR #6945, opened 2020-07-02
Don't sort player array in-place when scoring round robin tournaments

urkerab:
Fixes a regression from PR #5492 which removed a `map` from the relevant expression, thus causing the player array itself to become sorted, which means that players are no longer associated with their correct battle results (the scores are still correct though).


Zarel:
Good catch!

urkerab:
Well, it was obvious only because I had run a 21-player double round robin tournament on my server (0v0, so it was over quickly), and the player lists on the results suddenly became sorted, thus disconnecting them from the battle results.

PR #6946, opened 2020-07-03
Support throwing error messages

Zarel:
This has been discussed extensively on Discord, but in case any of you don't keep up there or want more details...

This change adds a new class `Chat.ErrorMessage extends Error`, which can be thrown. If thrown inside a chat command, PS won't crash, but will instead send its message to the user as an error message. Otherwise, PS will crash normally.

The purpose here is to eliminate some boilerplate. This would otherwise have been written:

```ts
try {
  doAThing();
} catch (e) {
  return this.errorReply(e.message);
}
```

But by standardizing an error type that's intended to represent a user error, we can simplify this to:

```ts
doAThing();
```

The main disadvantage here is that it's no longer obvious that `doAThing()` can throw, exiting the scope and preventing anything more from happening. I've mulled a lot of possibilities and think there's no real good solution to this problem, but also that it isn't _that_ important (`throw` already exists). I think we would do well to prefix these sorts of functions with `check` to make it clearer what's going on, where relevant.

### DIFF
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -130,17 +130,16 @@ function isTriviaRoom(room: Room) {
 	return false;
 }
 
-function getTriviaGame(context: CommandContext) {
-	const room = context.room;
-	if (!room) return false;
+function getTriviaGame(room: Room | null) {
+	if (!room) {
+		throw new Chat.ErrorMessage(`This command can only be used in the Trivia room.`);
+	}
 	const game = room.game;
 	if (!game) {
-		context.errorReply(`There is no game in progress.`);
-		return false;
+		throw new Chat.ErrorMessage(`There is no game in progress.`);
 	}
 	if (!game.constructor.name.endsWith('Trivia')) {
-		context.errorReply(`The currently running game is not Trivia, it's ${game.title}.`);
-		return false;
+		throw new Chat.ErrorMessage(`The currently running game is not Trivia, it's ${game.title}.`);
 	}
 	return game as Trivia;
 }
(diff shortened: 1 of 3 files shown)


Zarel:
PS currently uses a common pattern of:

```ts
if (!doAThing(this)) return false;
```

Where `doAThing` takes the passed context and handles sending error messages itself. This has the advantage that it's immediately obvious which parts of the function can return, but the flaw that the actual error output code now has to be done in `doAThing`, making it impure when it's often conceptually a pure function, while also causing problems for writing unit tests (this is a big reason why `CommandContext#canHTML` doesn't have tests).

Zarel:
I've written c2748f1e55428152b as an example of how this can simplify the codebase.

server/chat.ts
```diff
@@ -548,6 +568,10 @@ export class CommandContext extends MessageContext {
 		try {
 			result = commandHandler.call(this, this.target, this.room, this.user, this.connection, this.cmd, this.message);
 		} catch (err) {
+			if (err.name?.endsWith('ErrorMessage')) {
```
scheibo, line 571: A comment about why we have to do this (ie hotpatching) would be helpful

server/chat-plugins/trivia.ts
```diff
@@ -130,17 +130,16 @@ function isTriviaRoom(room: Room) {
 	return false;
 }
 
-function getTriviaGame(context: CommandContext) {
-	const room = context.room;
-	if (!room) return false;
+function getTriviaGame(room: Room) {
+	if (!room) {
```
scheibo, line 134: I'd probably remove the braces on these first two `if`s now that theyre fairly short

Zarel: I'd rather have them all match.

server/chat.ts
Zarel: I think that'd be appropriate if it was `name === 'ErrorMessage'`, but this seems clearly intended to match all errors whose class names end with `ErrorMessage`. A comment would just do the same thing this code does.

PR #6947, opened 2020-07-03
Server side changes to support persistent user settings and privacy changes

scheibo:
Server side portion of https://github.com/smogon/pokemon-showdown-client/issues/1271, in two commits which can be rebased (instead of squashed). Pretty sure this requires a restart because I dont think we can hotpatch the first change. The change to `SETTINGS` and `updateuser` should be safe, this was its original state before I randomly added the other crap by mistake in a previous commit.

I've confirmed hackily with

```ts
var json = JSON.stringify({inviteOnlyNextBattle: true});
app.send('/updatesettings ' + json  + '\n/trn ' + name + ',0,' + assertion);
```

on the client that the setting seems to accurately be updated:

Screen Shot 2020-07-02 at 20 03 55

I'll open a client side PR tomorrow.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -16,6 +16,7 @@
 
 /* eslint no-else-return: "error" */
 import {Utils} from '../../lib/utils';
+type UserSettings = import('../users').UserSettings;
 
 const avatarTable = new Set([
 	'aaron',
@@ -567,7 +568,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(this.tr("You are already blocking private messages! To unblock, use /unblockpms"));
 		}
 		if (target in Config.groups) {
-			user.settings.blockPMs = target;
+			user.settings.blockPMs = target as GroupSymbol;
 			this.sendReply(this.tr `You are now blocking private messages, except from staff and ${target}.`);
 		} else if (target === 'autoconfirmed' || target === 'trusted' || target === 'unlocked') {
 			user.settings.blockPMs = target;
@@ -577,7 +578,7 @@ export const commands: ChatCommands = {
 			user.settings.blockPMs = true;
 			this.sendReply(this.tr("You are now blocking private messages, except from staff."));
 		}
-		user.update('blockPMs');
+		user.update();
 		return true;
 	},
 	blockpmshelp: [
@@ -593,7 +594,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(this.tr("You are not blocking private messages! To block, use /blockpms"));
 		}
 		user.settings.blockPMs = false;
-		user.update('blockPMs');
+		user.update();
 		return this.sendReply(this.tr("You are no longer blocking private messages."));
 	},
 	unblockpmshelp: [`/unblockpms - Unblocks private messages. Block them with /blockpms.`],
@@ -710,6 +711,29 @@ export const commands: ChatCommands = {
 
 			this.sendReply(`|raw|${buffer}`);
 		});
+	},
+
+	updatesettings(target, room, user) {
+		const settings: Partial<UserSettings> = {};
+		try {
+			const raw = JSON.parse(target);
+			if (typeof raw !== 'object' || Array.isArray(raw) || !raw) {
+				this.errorReply("/updatesettings expects JSON encoded object.");
+			}
+			for (const setting in user.settings) {
+				if (setting in raw) {
+					if (setting === 'blockPMs' &&
+						(raw[setting] in Config.groups || ['autoconfirmed', 'trusted', 'unlocked'].includes(raw[setting]))) {
+						settings[setting] = raw[setting];
+					} else {
+						settings[setting as keyof UserSettings] = !!raw[setting];
+					}
+				}
+			}
+			user.updateSettings(settings);
+		} catch {
+			this.errorReply("Unable to parse settings in /updatesettings!");
+		}
 	},
 
 	/*********************************************************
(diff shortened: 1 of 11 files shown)


server/chat-commands/core.ts
```diff
@@ -726,6 +729,23 @@ export const commands: ChatCommands = {
...
+		const settings: Partial<UserSettings> = {};
+		try {
+			const raw = JSON.parse(target);
+			if (typeof raw !== 'object' || Array.isArray(raw) || !raw) {
+				this.errorReply("/updatesettings expects JSON encoded object.");
+			}
+			for (const setting of User.SETTINGS) {
+				if (setting in raw) settings[setting] = settings !== 'blockPMs' ? !!raw[setting] : raw[setting];
+			}
+			user.updateSettings(JSON.parse(target));
```
whalemer, line 743: should this be `settings`?

server/chat-commands/core.ts
```diff
@@ -726,6 +729,23 @@ export const commands: ChatCommands = {
...
+	'!updatesettings': true,
+	updatesettings(target, room, user) {
+		const settings: Partial<UserSettings> = {};
+		try {
+			const raw = JSON.parse(target);
+			if (typeof raw !== 'object' || Array.isArray(raw) || !raw) {
+				this.errorReply("/updatesettings expects JSON encoded object.");
+			}
+			for (const setting of User.SETTINGS) {
+				if (setting in raw) settings[setting] = settings !== 'blockPMs' ? !!raw[setting] : raw[setting];
```
whalemer, line 741: also while i'm nitpicking, i think it's possible to cause a crash in `authAtLeast` if you pass something like `{blockPMs: {toString: null}}`

whalemer: also i think this should be `setting !== 'blockPMs'`? i'm not sure why typescript isn't complaining about that.

server/chat-commands/core.ts
scheibo: Yes!

server/chat-commands/core.ts
scheibo: "nitpicking" aka fixing my broken code! thanks. And yeah, i was worried about blockPMs because it takes a string here. Possibly need to include a set with all the possible values it can take (and give blockPMs a stricter type than `boolean | string` while we're at it?)

scheibo: OK, I've actually typed `blockPMs` properly (I think?) and filter it correctly now!

scheibo:
This isn't actually going to be what we want to use for `/inviteNextBattle` ('persistent `/ionext` isn't going to be a thing, a sticky client toggle for a new and improved `/hidenext` command we be added instead), but its still necessary for the other settings.

server/chat-commands/core.ts
```diff
@@ -16,6 +16,7 @@
 
 /* eslint no-else-return: "error" */
 import {Utils} from '../../lib/utils';
+import {User, UserSettings} from '../users';
```
Zarel, line 19:
```suggestion
type UserSettings = import('../users').UserSettings;
```

At some point we should investigate making sure sucrase is recent enough for `import type`. Until then, this, I guess.

server/chat-commands/core.ts
```diff
@@ -726,6 +729,30 @@ export const commands: ChatCommands = {
...
 
+	'!updatesettings': true,
+	updatesettings(target, room, user) {
+		const settings: Partial<UserSettings> = {};
+		try {
+			const raw = JSON.parse(target);
+			if (typeof raw !== 'object' || Array.isArray(raw) || !raw) {
+				this.errorReply("/updatesettings expects JSON encoded object.");
+			}
+			for (const setting of User.SETTINGS) {
```
Zarel, line 740:
```suggestion
			for (const setting of Users.User.SETTINGS) {
```

Personally I would just `setting in user.settings` but if you insist on the TS-safe way...

scheibo: `user.settings` works!

scheibo:
If we restart soon this can be merged as is (before the restart). I don't know that I'm going to get the password + renaming functionality added to this PR as opposed to a followup

scheibo:
This is going to break private code:

#  The persistent settings hooks are going to need to be modified before restarting

(this is fine, because they should probably be altered to remove blockPMs/blockChallenges from the loginhooks anyway, as this should be merged along with the client changes that make those two settings self-service anyway)

Zarel:
Can you PR the private code and I can just accept both simultaneously on restart?

server/chat-commands/core.ts
```diff
@@ -726,6 +729,31 @@ export const commands: ChatCommands = {
 		});
 	},
 
+	'!updatesettings': true,
```
Zarel, line 732: I was wondering if this would be merged before or after nullable rooms. Turns out, after!

```suggestion
```

server/chat-commands/room-settings.ts
```diff
@@ -969,6 +969,28 @@ export const commands: ChatCommands = {
 		`/publicroom - Makes a room public. Requires: \u2606 &`,
 	],
 
+	'!hidenext': true,
```
Zarel, line 972:
```suggestion
```

scheibo:
Updated (I've been doing the work to keep these commits all rebase-able, for the record (as opposed to being squashed). Please consider rebasing instead of squashing if you are the one that merges.

> Can you PR the private code and I can just accept both simultaneously on restart?

Yup, will do.

PR #6952, opened 2020-07-03
Fix bug in /renameroom

AnnikaCodes:


server/rooms.ts
```diff
@@ -693,7 +693,7 @@ export abstract class BasicRoom {
 		// add an alias from the old id
 		Rooms.aliases.set(oldID, newID);
 		if (!this.settings.aliases) this.settings.aliases = [];
-		this.settings.aliases.push(oldID);
+		if (!this.settings.aliases.includes(oldID)) this.settings.aliases.push(oldID);
```
Zarel, line 696: In theory, this should never happen. Did you manage to get it to happen somehow?

server/rooms.ts
```diff
@@ -719,10 +719,8 @@ export abstract class BasicRoom {
 			}
 		}
 
-		if (this.persist) {
-			this.title = newTitle;
-			this.saveSettings();
-		}
+		this.settings.title = newTitle;
+		this.saveSettings();
```
Zarel, line 723: This is probably a good change, but I'm curious how it constitutes a bug. We shouldn't be reading `this.settings.title` except on room creation, which a non-persistent room shouldn't be doing, right?

Zarel: > While there is a banner stating the room has been renamed, it's not a link -- it seems that all user connections are moved to the new room, as is the modlog. There is, however, a bug in `/renameroom`, which I opened a PR to fix (#6952): the chatroom's new name is never saved to `chatrooms.json`, and thus the rename partially reverts after restarting.

Oh, I understand now! I think this implies a worse underlying bug: that `room.persist` isn't being set correctly? This will need more investigation.

AnnikaCodes: The bug I found was that it was just setting `this.title`, which never actually gets saved.

AnnikaCodes: (and since `saveSettings()` checks `room.persist`, I found it silly to keep the conditional in there)

server/rooms.ts
AnnikaCodes: yes - if `/renameroom` was used while the title-saving bug was in effect, an alias would be added for the "old" name, but then the old name is restored on restart.

AnnikaCodes: We can probably remove the line if this hasn't happened to any rooms on main, though.

server/rooms.ts
Zarel: Oh! I didn't even notice that until now.

server/rooms.ts
Zarel:
```suggestion
		// resolve an old (fixed) bug in /renameroom
		if (!this.settings.aliases.includes(oldID)) this.settings.aliases.push(oldID);
```

Zarel: Oh, that's fine, although it could use comment documentation.

Zarel:
Thanks!

PR #6953, opened 2020-07-03
Add /nofeedback command

AnnikaCodes:
Note that this only suppresses `sendReply()`, so something like `/nofeedback /hourmute SomeTroll` will still display the mute message in chat (which I think is important for transparency). However, this does mean that things like `/nofeedback /hideroom` will still display in chat for the opponent to see (note that `/nofeedback /ionext`, which seems to be the main use-case, is completely suppressed).

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -484,9 +484,8 @@ export const commands: ChatCommands = {
 		user.resetName();
 	},
 
-	nfb: 'nofeedback',
 	nofeedback(target, room, user) {
-		if (!target?.startsWith('/')) return this.parse('/help nofeedback');
+		if (!target.startsWith('/')) return this.parse('/help nofeedback');
 		return this.parse(target, true);
 	},
 	nofeedbackhelp: [`/nofeedback [command] - Runs the command without displaying the response.`],
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -484,11 +484,11 @@ export const commands: ChatCommands = {
 		user.resetName();
 	},
 
-	nofeedback(target, room, user) {
-		if (!target.startsWith('/')) return this.parse('/help nofeedback');
+	noreply(target, room, user) {
+		if (!target.startsWith('/')) return this.parse('/help noreply');
 		return this.parse(target, true);
 	},
-	nofeedbackhelp: [`/nofeedback [command] - Runs the command without displaying the response.`],
+	noreplyhelp: [`/noreply [command] - Runs the command without displaying the response.`],
 
 	r: 'reply',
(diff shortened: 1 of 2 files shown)


DaWoblefet:
I know Zarel brought this up for ionext, but what exactly is the point of this command? Why would I ever bother to type extra characters to hide the response from the server?

Zarel:
@DaWoblefet you would never type it yourself. This is for scripting, like buttons, like `<button name="send" value="/nofeedback /ionext">` so they don't spam your PMs.

server/chat-commands/core.ts
```diff
@@ -484,6 +484,13 @@ export const commands: ChatCommands = {
 		user.resetName();
 	},
 
+	nfb: 'nofeedback',
```
Zarel, line 487:
```suggestion
```

I would not give an alias because this is not actually intended to be used by users.

server/chat-commands/core.ts
```diff
@@ -484,6 +484,13 @@ export const commands: ChatCommands = {
...
+	nofeedback(target, room, user) {
+		if (!target?.startsWith('/')) return this.parse('/help nofeedback');
```
Zarel, line 489:
```suggestion
		if (!target.startsWith('/')) return this.parse('/help nofeedback');
```

(`target` isn't nullable)

Zarel:
I'm still thinking about command naming, though. `/nofeedback` isn't maximally clear. Maybe `/silentcmd`?

Zarel:
`/noreply`?

DaWoblefet:
That makes more sense now, thanks.

scheibo:
Preference for `/noreply` over `/nofeedback`, but just `/silent` would have been my choice.

Can we please merge? I need this for the UI now :)

Zarel:
I think `/silent` might be confusing to users who stumble upon it; it's a bit jargony. Let's go with `/noreply`.

server/chat-commands/core.ts
```diff
@@ -484,6 +484,12 @@ export const commands: ChatCommands = {
 		user.resetName();
 	},
 
+	nofeedback(target, room, user) {
+		if (!target.startsWith('/')) return this.parse('/help nofeedback');
+		return this.parse(target, true);
+	},
+	nofeedbackhelp: [`/nofeedback [command] - Runs the command without displaying the response.`],
```
Zarel, line 491:
```suggestion
	noreply(target, room, user) {
		if (!target.startsWith('/')) return this.parse('/help noreply');
		return this.parse(target, true);
	},
	noreplyhelp: [`/noreply [command] - Runs the command without displaying the response.`],
```

PR #6955, opened 2020-07-04
Chatlog: Improve search

mia-pi-git:
Changes: 
- ripgrep now uses childprocess#exec, for more precise error handling and to prevent an unidentified bug that made execFile fail. 
- ripgrep supports searching for multiple terms 
- ripgrep supports searching a specific month.
- argument handling is now handled based on regex.exec, using `room: targetRoom`, `limit: number` etc to retrieve the arguments.
- also doesn't reverse the results like it previously did.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -528,14 +528,8 @@ export const LogSearcher = new class {
 		}
 		return {results, total};
 	}
-
-	async ripgrepSearch(
-		roomid: RoomID,
-		search: string,
-		limit?: number | null,
-		date?: string | null
-	) {
-		if (!limit || limit > MAX_RESULTS) limit = MAX_RESULTS;
+	async ripgrepSearchMonth(roomid: RoomID, search: string, limit: number, date: string) {
+		let results;
 		try {
 			const {stdout} = await execFile('rg', [
 				'-e', this.constructRegex(search),
@@ -547,11 +541,37 @@ export const LogSearcher = new class {
 				maxBuffer: Infinity,
 				cwd: path.normalize(`${__dirname}/../../`),
 			});
-			return this.render(stdout.split('--'), roomid, search, limit, date);
+			results = stdout.split('--');
 		} catch (e) {
-			if (e.stdout) return this.render(e.stdout.split('--'), roomid, search, limit, date);
+			if (e.stdout) results = e.stdout.split('--');
 			throw e;
 		}
+		const count = results.length;
+		return {results, count};
+	}
+	async ripgrepSearch(
+		roomid: RoomID,
+		search: string,
+		limit?: number | null,
+		date?: string | null
+	) {
+		const months = date ? [date] : await new LogReaderRoom(roomid).listMonths();
+		let count = 0;
+		let results: string[] = [];
+		if (!limit || limit > MAX_RESULTS) limit = MAX_RESULTS;
+		if (!date) date = 'all';
+		while (count < MAX_RESULTS) {
+			const month = months.shift();
+			if (!month) break;
+			const output = await this.ripgrepSearchMonth(roomid, search, limit, month);
+			results = results.concat(output.results);
+			count += output.count;
+		}
+		if (count > MAX_RESULTS) {
+			const diff = count - MAX_RESULTS;
+			results = results.slice(0, -diff);
+		}
+		return this.render(results, roomid, search, limit, date);
 	}
 
 	render(results: string[], roomid: RoomID, search: string, limit: number, month?: string | null) {
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -14,6 +14,7 @@ import * as Dashycode from '../../lib/dashycode';
 
 const DAY = 24 * 60 * 60 * 1000;
 const MAX_RESULTS = 3000;
+const MAX_MEMORY = 67108864; // 64MB
 const execFile = util.promisify(child_process.execFile);
 
 export class LogReaderRoom {
@@ -530,6 +531,7 @@ export const LogSearcher = new class {
 	}
 	async ripgrepSearchMonth(roomid: RoomID, search: string, limit: number, month: string) {
 		let results;
+		let count = 0;
 		try {
 			const {stdout} = await execFile('rg', [
 				'-e', this.constructRegex(search),
@@ -538,7 +540,7 @@ export const LogSearcher = new class {
 				'-m', `${limit}`,
 				'-P',
 			], {
-				maxBuffer: Infinity,
+				maxBuffer: MAX_MEMORY,
 				cwd: path.normalize(`${__dirname}/../../`),
 			});
 			results = stdout.split('--');
@@ -546,7 +548,7 @@ export const LogSearcher = new class {
 			if (e.stdout) results = e.stdout.split('--');
 			throw e;
 		}
-		const count = results.length;
+		count += results.length;
 		return {results, count};
 	}
 	async ripgrepSearch(
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -545,7 +545,11 @@ export const LogSearcher = new class {
 			});
 			results = stdout.split('--');
 		} catch (e) {
-			if (e.stdout) results = e.stdout.split('--');
+			if (e.stdout || e.message.includes('stdout maxBuffer')) {
+				results = e.stdout.split('--');
+				count += results.length;
+				return {results, count};
+			}
 			throw e;
 		}
 		count += results.length;
@@ -557,7 +561,7 @@ export const LogSearcher = new class {
 		limit?: number | null,
 		date?: string | null
 	) {
-		const months = (date ? [date] : await new LogReaderRoom(roomid).listMonths()).reverse();
+		const months = (date && toID(date) !== 'all' ? [date] : await new LogReaderRoom(roomid).listMonths()).reverse();
 		let count = 0;
 		let results: string[] = [];
 		if (!limit || limit > MAX_RESULTS) limit = MAX_RESULTS;
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -8,15 +8,15 @@
 import {FS} from "../../lib/fs";
 import {Utils} from '../../lib/utils';
 import * as child_process from 'child_process';
-import * as util from 'util';
+import * as Util from 'util';
 import * as path from 'path';
 import * as Dashycode from '../../lib/dashycode';
 
-const execFile = util.promisify(child_process.execFile);
 const DAY = 24 * 60 * 60 * 1000;
 const MAX_RESULTS = 3000;
+const execFile = Util.promisify(child_process.execFile);
 
-class LogReaderRoom {
+export class LogReaderRoom {
 	roomid: RoomID;
 	constructor(roomid: RoomID) {
 		this.roomid = roomid;
@@ -429,10 +429,20 @@ export const LogViewer = new class {
 	}
 };
 
-/** match with two lines of context in either direction */
+/** Match with two lines of context in either direction */
 type SearchMatch = readonly [string, string, string, string, string];
 
-const LogSearcher = new class {
+export const LogSearcher = new class {
+	constructRegex(str: string) {
+		const searches = str.split('+').map(term => Utils.escapeRegex(term));
+		if (searches.length <= 1) {
+			if (str.length <= 3) return `\b${str}`;
+			return str;
+		}
+
+		return `^` + searches.map(term => `(?=.*${term})`).join('');
+	}
+
 	fsSearch(roomid: RoomID, search: string, date: string, limit: number | null) {
 		const isAll = (date === 'all');
 		const isYear = (date.length === 4);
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -429,10 +427,26 @@ export const LogViewer = new class {
...
+		} else if (searches.length > 1) {
+			regex = '^';
+			for (const term of searches) {
+				regex += `(?=.*${term})`;
+			}
+		} else {
+			regex = str;
+		}
+		return regex;
+	}
```
AnnikaCodes, line 448:
```suggestion
	constructRegex(str: string) {
		const searches = str.split('+');
		if (str.length <= 3 && searches.length < 1) return `\\b${str}`;
		if (searches.length > 1) {
			let regex = '^';
			for (const term of searches) {
				regex += `(?=.*${term})`;
			}
			return regex;
		}
		return str;
	}
```

server/chat-plugins/chatlog.ts
```diff
@@ -429,10 +427,23 @@ export const LogViewer = new class {
...
+		if (str.length <= 3 && searches.length < 1) return `\b${str}`;
+		if (searches.length > 1) {
+			let regex = '^';
+			for (const term of searches) {
+				regex += `(?=.*${term})`;
+			}
+			return regex;
+		}
+		return str;
+	}
```
Zarel, line 445:
```suggestion
	constructRegex(str: string) {
		const searches = str.split('+');
		if (searches.length <= 1) {
			if (str.length <= 3) return `\b${str}`;
			return str;
		}

		return `^` + searches.map(term => `(?=.*${term})`).join('');
	}
```

`searches.length < 1` will always be false; `''.split('+')` is still `['']`, not `[]`. You meant `searches.length <= 1`. The funny thing is that Annika didn't notice, either. :p

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +527,53 @@ const LogSearcher = new class {
...
+			`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
+			'-C', '3',
+			'-m', `${limit}`,
+			'-P',
+		].join(' ');
+		const execOptions = {
+			maxBuffer: Infinity,
+			cwd: path.normalize(`${__dirname}/../../`),
+		};
+		child_process.exec(`rg ${searchOptions}`, execOptions, (err, stdout, stderr) => {
```
Zarel, line 549:
```suggestion
		child_process.execFile('rg', searchOptions, execOptions, (err, stdout, stderr) => {
```

Would this still work?

Zarel, line 549: The first test is this: Does this still work?

```suggestion
		const searchOptions = [
			'-e', `"${this.constructRegex(search)}"`,
			`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
			'-C', '3',
			'-m', `${limit}`,
			'-P',
		];
		const execOptions = {
			maxBuffer: Infinity,
			cwd: path.normalize(`${__dirname}/../../`),
		};
		child_process.execFile('rg', searchOptions, execOptions, (err, stdout, stderr) => {
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,53 @@ const LogSearcher = new class {
...
-			if (error.message.includes('Command failed')) return LogViewer.error(`No results found.`);
-			return LogViewer.error(`${error.message}`);
-		}
-		return this.render(
-			output.stdout.split('--').reverse(),
-			roomid,
-			search,
-			limit
-		);
+			output = await execFile('rg', searchOptions, execOptions);
```
Zarel, line 552:
```suggestion
			output = await execFile('rg', [
				'-e', `${this.constructRegex(search)}`,
				`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
				'-C', '3',
				'-m', `${limit}`,
				'-P',
			], {
				maxBuffer: Infinity,
				cwd: path.normalize(`${__dirname}/../../`),
			});
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,53 @@ const LogSearcher = new class {
...
+		} catch (e) {
+			if (output?.stdout) return this.render(output.stdout.split('--'), roomid, search, limit, date);
```
Zarel, line 554:
```suggestion
			if (e.stdout) return this.render(e.stdout.split('--'), roomid, search, limit, date);
```

Zarel: (If this crashes, it will crash before `output` is set, so reading `output` is meaningless here.

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,62 @@ const LogSearcher = new class {
...
+			'-e', `${this.constructRegex(search)}`,
+			`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
+			'-C', '3',
+			'-m', `${limit}`,
+			'-P',
+		];
+		const execOptions = {
+			maxBuffer: Infinity,
+			cwd: path.normalize(`${__dirname}/../../`),
+		};
```
Zarel, line 549:
```suggestion
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,62 @@ const LogSearcher = new class {
...
+		let output;
 		try {
-			const options = [
-				'-e', `[^a-zA-Z0-9]${search.split('').join('[^a-zA-Z0-9]*')}([^a-zA-Z0-9]|\\z)`,
-				`${__dirname}/../../logs/chat/${roomid}`,
+			output = await execFile('rg', [
```
Zarel, line 552:
```suggestion
			const {stdout} = await execFile('rg', [
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,62 @@ const LogSearcher = new class {
...
-			output.stdout.split('--').reverse(),
-			roomid,
-			search,
-			limit
-		);
+				'-P',
+			], {
+				maxBuffer: Infinity,
+				cwd: path.normalize(`${__dirname}/../../`),
+			});
```
Zarel, line 561:
```suggestion
			});
			return this.render(stdout.split('--'), roomid, search, limit, date);
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,62 @@ const LogSearcher = new class {
...
+		} catch (e) {
+			if (e.stdout) return this.render(e.stdout.split('--'), roomid, search, limit, date);
+			throw e;
+		}
+		return this.render(output.stdout.split('--'), roomid, search, limit, date);
```
Zarel, line 566:
```suggestion
```

server/chat-plugins/chatlog.ts
```diff
@@ -519,48 +529,62 @@ const LogSearcher = new class {
...
+			`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
+			'-C', '3',
+			'-m', `${limit}`,
+			'-P',
+		];
+		const execOptions = {
+			maxBuffer: Infinity,
+			cwd: path.normalize(`${__dirname}/../../`),
+		};
+		let output;
```
Zarel, line 550:
```suggestion
```

Zarel:
Why did you revert the output change? I would have shown you how to fix it if you hadn't reverted it. :T

server/chat-plugins/chatlog.ts
```diff
@@ -678,34 +690,41 @@ export const commands: ChatCommands = {
...
-			limitString = ``;
-		} else {
-			return this.errorReply(`Cap must be a number or [all].`);
+		let date = 'all';
+		const searches: string[] = [];
+		let limit = '500';
+		let tarRoom = room.roomid;
+		for (const arg of args) {
+			if (/room:(.+)/.exec(arg)) {
+				const id = /room:(.+)/.exec(arg)![1];
```
Zarel, line 703: You definitely never want to run a regex twice. Just save the output in a variable.

In this case, though, why are you even using a regex? Just use `arg.startsWith('room:')` and `arg.slice(5)`.

server/chat-plugins/chatlog.ts
```diff
@@ -678,34 +710,41 @@ export const commands: ChatCommands = {
...
-			limitString = ``;
-		} else {
-			return this.errorReply(`Cap must be a number or [all].`);
+		let date = 'all';
+		const searches: string[] = [];
+		let limit = '500';
+		let tarRoom = room.roomid;
+		for (const arg of args) {
+			if (arg.startsWith('room:')) {
+				const id = arg.slice(arg.indexOf('room:') + 5);
```
Zarel, line 723: `arg.indexOf('room:')` is guaranteed to be `0`. That's what `startsWith` is for (and why we used `trim` earlier).

server/chat-plugins/chatlog.ts
```diff
@@ -678,34 +710,41 @@ export const commands: ChatCommands = {
...
+				tarRoom = id as RoomID;
```
Zarel, line 724: It's also weird to have an intermediate variable `id` when you don't for the others.

server/chat-plugins/chatlog.ts
```diff
@@ -678,34 +710,41 @@ export const commands: ChatCommands = {
...
+		let date = 'all';
+		const searches: string[] = [];
+		let limit = '500';
+		let tarRoom = room.roomid;
+		for (const arg of args) {
+			if (arg.startsWith('room:')) {
+				const id = arg.slice(5);
+				tarRoom = id as RoomID;
+			} else if (arg.startsWith('limit:')) {
+				limit = arg.slice(6);
```
Zarel, line 726: But also, these should probably be trimmed, in case they put a space after `limit:`

server/chat-plugins/chatlog.ts
```diff
@@ -8,15 +8,15 @@
 import {FS} from "../../lib/fs";
 import {Utils} from '../../lib/utils';
 import * as child_process from 'child_process';
-import * as util from 'util';
+import * as Util from 'util';
```
Zarel, line 11: ???? no this should be lowercase

server/chat-plugins/chatlog.ts
```diff
@@ -518,49 +528,71 @@ const LogSearcher = new class {
...
-		if (!limit || limit > MAX_RESULTS) limit = MAX_RESULTS;
+	async ripgrepSearchMonth(roomid: RoomID, search: string, limit: number, date: string) {
+		let results;
 		try {
-			const options = [
-				'-e', `[^a-zA-Z0-9]${search.split('').join('[^a-zA-Z0-9]*')}([^a-zA-Z0-9]|\\z)`,
-				`${__dirname}/../../logs/chat/${roomid}`,
+			const {stdout} = await execFile('rg', [
+				'-e', this.constructRegex(search),
+				`logs/chat/${roomid}${toID(date) === 'all' ? '' : `/${date}`}`,
```
Zarel, line 536: This will break unless `date` is a month. If `date` is a month, it should be named `month`.

server/chat-plugins/chatlog.ts
```diff
@@ -518,49 +528,71 @@ const LogSearcher = new class {
...
-		}
-		return this.render(
-			output.stdout.split('--').reverse(),
-			roomid,
-			search,
-			limit
-		);
+				'-P',
+			], {
+				maxBuffer: Infinity,
```
Zarel, line 541: `maxBuffer: Infinity` sounds like you _want_ out-of-memory errors?

mia-pi-git: Wasn't sure how this should be handled, i don't know enough about this particular setting in order to figure out the best way to set it.

mia-pi-git: Should this just be using ``process.memoryUsage().heapTotal``?

Zarel: No.

Zarel: I would probably set it to some const.

mia-pi-git: Is there a good way to determine said const, though?

Zarel: Apparently `maxBuffer` defaults to 1MB. Feel free to raise that to, like, 64MB if you need it. I'd be wary of raising it higher than that.

Zarel: Also, I think limiting results to 100-200 a day should be plenty.

Zarel: Maybe tell people to redo the search in a specific day if they want more than 200 results in a day? Then cap it at 201.

mia-pi-git: That seems annoying for users, tbh. 
Why not just cap it at 64 and then lower ``MAX_RESULTS`` a bit?

mia-pi-git: 200 seems inconvenient for active rooms, so i capped it at 64mb per month and made it catch and return if it's a maxBuffer exceeded error.

Zarel: Crashing the entire server seems much more annoying for users.

I suspect if you think it's annoying, it's because you don't understand what I'm recommending.

mia-pi-git: well, yes, lol, crashing would be worse.
From what i understand, you're recommending 200 results per day, with more only if it's specifically for that day?

Zarel:
I haven't fully reviewed this, but there are some improvements here and it'd be easier to fix the problems than to try to review every individual change. Thanks!

PR #6958, opened 2020-07-04
Allow custom formats (e.g. for tours) to modify format restrictions

urkerab:
This puts the restrictions in the rule table, thus allowing them to be inherited and modified, but also avoids having to guess where to find them in the code that implements them.

I just picked `*` because it's the next mathematical symbol after `+` and `-`...

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -455,7 +455,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 		],
 
 		mod: 'gen8',
-		ruleset: ['[Gen 8] Ubers', 'Dynamax Clause'],
+		ruleset: ['Standard', 'Dynamax Clause'],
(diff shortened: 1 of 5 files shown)


Zarel:
I'm confused. What's a format restriction?

Zarel:
Oh, I see! It's for applying custom rules to restricted things. That's an interesting idea.

Zarel:
This is a half-measure but I think it's still valuable.

It looks like this only supports adding to restrictions, not removing them. What do you think of `!*` for removing them? I think we might also want to support e.g. `!-` for "unban without whitelisting".

urkerab:
What's the problem with whitelisting to unrestrict something?

Zarel:
Oh! I didn't think of that. I suppose it works.

Zarel:
Good change, btw.

PR #6965, opened 2020-07-06
Support renaming groupchats

mia-pi-git:
useful.


server/chat-commands/room-settings.ts
```diff
@@ -726,6 +726,41 @@ export const commands: ChatCommands = {
 		`/subroomgroupchat [roomname] - Creates a subroom groupchat of the current room. Can only be used in a public room you have staff in.`,
 	],
 
+	async renamegroupchat(target, room, user) {
+		if (!room) return this.requiresRoom();
+		if (!user.authAtLeast(Users.HOST_SYMBOL, room) || !user.can('lock')) return false;
```
whalemer, line 731: this is checking for "host or higher AND gdriver or higher", which seems wrong.

whalemer, line 731: also it probably shouldn't fail completely silently.

whalemer:
there's currently nothing that stops you from using this in a normal room.

KrisXV:
I'm sorry but I'm lost. What's the point of this command? Couldn't someone just delete the old groupchat and make a new one? I'm not trying to be rhetorical or anything btw, I just don't know if I missed a conversation or something.

PartMan7:
One reason I can think of is the groupchat limit.

Zarel:
There's some marginal utility in preserving the history, I think.

Zarel:
(And also in automatically moving everyone over.)

Zarel:
I think these checks can all be combined into a function like `setTitle` though. Another use-case for `Chat.ErrorMessage`!

mia-pi-git:
> I think these checks can all be combined into a function like `setTitle` though. Another use-case for `Chat.ErrorMessage`!

Why not just do the checks in `BasicRoom#rename` then?

Zarel:
> > I think these checks can all be combined into a function like `setTitle` though. Another use-case for `Chat.ErrorMessage`!
> 
> Why not just do the checks in `BasicRoom#rename` then?

Not a bad idea, either.

mia-pi-git:
Ugh.

Zarel:
I think the biggest issue with doing the checks in `rename` is that we can't reuse those checks for `/makechatroom`.

I think a `validateTitle` method might actually be the right place for it? Which would then be used in the `BasicRoom` constructor as well as the `rename` function.

server/chat-commands/room-settings.ts
```diff
@@ -730,6 +717,41 @@ export const commands: ChatCommands = {
 		`/subroomgroupchat [roomname] - Creates a subroom groupchat of the current room. Can only be used in a public room you have staff in.`,
 	],
 
+	async renamegroupchat(target, room, user) {
```
Zarel, line 720: Why is this a different command than `renameroom`, anyway?

mia-pi-git: i figured it'd be clearer for some people.

Zarel: An alias would achieve that with less code duplication.

mia-pi-git: Okay, then

server/rooms.ts
```diff
@@ -679,6 +692,11 @@ export abstract class BasicRoom {
 	 */
 	async rename(newTitle: string, newID?: RoomID, noAlias?: boolean) {
 		if (!newID) newID = toID(newTitle) as RoomID;
+		try {
+			this.validateTitle(newTitle, newID);
+		} catch (e) {
+			return;
```
Zarel, line 698: ?

server/rooms.ts
```diff
@@ -665,6 +666,18 @@ export abstract class BasicRoom {
...
+		if (!newID) newID = toID(newTitle);
+		// `,` is a delimiter used by a lot of /commands
+		// `|` and `[` are delimiters used by the protocol
+		// `-` has special meaning in roomids
+		if (newTitle.includes(',') || newTitle.includes('|') || newTitle.includes('[') || newTitle.includes('-')) {
+			throw new Chat.ErrorMessage("Room titles can't contain any of: ,|[-");
+		}
+		if (newID.length > 225) throw new Chat.ErrorMessage("The given room title is too long.");
+		if (Rooms.search(newTitle)) throw new Chat.ErrorMessage(`The room '${newTitle}' already exists.`);
+		return true;
```
Zarel, line 679:
```suggestion
```

The return value doesn't seem to be used anywhere.

server/chat-commands/room-settings.ts
```diff
@@ -850,24 +872,15 @@ export const commands: ChatCommands = {
...
-		// `,` is a delimiter used by a lot of /commands
-		// `|` and `[` are delimiters used by the protocol
-		// `-` has special meaning in roomids
-		if (roomtitle.includes(',') || roomtitle.includes('|') || roomtitle.includes('[') || roomtitle.includes('-')) {
-			return this.errorReply("Room titles can't contain any of: ,|[-");
-		}
-		if (roomid.length > MAX_CHATROOM_ID_LENGTH) return this.errorReply("The given room title is too long.");
-		if (Rooms.search(roomtitle)) return this.errorReply(`The room '${roomtitle}' already exists.`);
+		const oldTitle = room.title;
+		room.validateTitle(roomtitle);
```
Zarel, line 883:
```suggestion
```

`rename` should already call `validateTitle`

server/rooms.ts
Zarel: Oh, this is because of the async `Chat.ErrorMessage` problem. That's fixed now, so you can just make this make sense again.

server/rooms.ts
```diff
@@ -679,6 +692,11 @@ export abstract class BasicRoom {
 	 */
 	async rename(newTitle: string, newID?: RoomID, noAlias?: boolean) {
 		if (!newID) newID = toID(newTitle) as RoomID;
+		try {
+			this.validateTitle(newTitle, newID);
+		} catch (e) {
+			return;
+		}
```
Zarel, line 698:
```suggestion
		this.validateTitle(newTitle, newID);
```

server/rooms.ts
```diff
@@ -665,6 +666,17 @@ export abstract class BasicRoom {
...
 	}
+	validateTitle(newTitle: string, newID?: string) {
+		if (!newID) newID = toID(newTitle);
+		// `,` is a delimiter used by a lot of /commands
+		// `|` and `[` are delimiters used by the protocol
+		// `-` has special meaning in roomids
+		if (newTitle.includes(',') || newTitle.includes('|') || newTitle.includes('[') || newTitle.includes('-')) {
+			throw new Chat.ErrorMessage("Room titles can't contain any of: ,|[-");
+		}
+		if (newID.length > 225) throw new Chat.ErrorMessage("The given room title is too long.");
```
Zarel, line 677: Rather than hardcoding, I think `const MAX_CHATROOM_ID_LENGTH = 225;` should be defined on top of this file.

server/chat-commands/room-settings.ts
```diff
@@ -846,32 +832,36 @@ export const commands: ChatCommands = {
...
 	},
-
-	async renameroom(target, room) {
+	renamegroupchat: 'renameroom',
+	async renameroom(target, room, user, connection, cmd) {
 		if (!this.can('makeroom')) return;
 		if (!room) return this.requiresRoom();
-		if (room.minorActivity || room.game || room.tour) {
-			return this.errorReply("Cannot rename room when there's a tour/game/poll/announcement running.");
+		if (room.game || room.minorActivity || room.tour) {
```
Zarel, line 839:
```suggestion
		if (room.game || room.minorActivity) {
```

server/chat-commands/room-settings.ts
```diff
@@ -881,7 +871,7 @@ export const commands: ChatCommands = {
 
 		const toNotify: RoomID[] = ['upperstaff'];
 		if (room.settings.isPrivate !== true) toNotify.push('staff');
```
Zarel, line 873:
```suggestion
		const toNotify: RoomID[] = room.settings.isPrivate === true ? ['upperstaff'] : ['upperstaff', 'staff'];
```

server/chat-commands/room-settings.ts
```diff
@@ -881,7 +871,7 @@ export const commands: ChatCommands = {
...
-		Rooms.global.notifyRooms(toNotify, message);
+		if (!isGroupchat) Rooms.global.notifyRooms(toNotify, message);
```
Zarel, line 873:
```suggestion
		if (!isGroupchat) {
			const toNotify: RoomID[] = room.settings.isPrivate === true ? ['upperstaff'] : ['upperstaff', 'staff'];
			Rooms.global.notifyRooms(toNotify, message);
		}
```

server/chat-commands/room-settings.ts
```diff
@@ -846,42 +832,45 @@ export const commands: ChatCommands = {
...
 			!room.settings.isPrivate ? "Public" :
 			`${room.settings.isPrivate.charAt(0).toUpperCase()}${room.settings.isPrivate.slice(1)}`;
 		const message = Utils.html`
 			|raw|<div class="broadcast-green">${privacy} chat room <b>${oldTitle}</b> renamed to <b>${target}</b></div>
 		`;
 
-		const toNotify: RoomID[] = ['upperstaff'];
-		if (room.settings.isPrivate !== true) toNotify.push('staff');
-		Rooms.global.notifyRooms(toNotify, message);
+		const toNotify: RoomID[] = room.settings.isPrivate === true ? ['upperstaff'] : ['upperstaff', 'staff'];
```
Zarel, line 872:
```suggestion
```

server/chat-commands/room-settings.ts
```diff
@@ -846,42 +832,48 @@ export const commands: ChatCommands = {
...
 			return this.errorReply(`An error occured while renaming the room.`);
 		}
-		this.modlog(`RENAMEROOM`, null, `from ${oldTitle}`);
+		this.modlog(`RENAME${isGroupchat ? 'GROUPCHAT' : 'ROOM'}`, null, `from ${oldTitle}`);
 		const privacy = room.settings.isPrivate === true ? "Private" :
 			!room.settings.isPrivate ? "Public" :
 			`${room.settings.isPrivate.charAt(0).toUpperCase()}${room.settings.isPrivate.slice(1)}`;
 		const message = Utils.html`
 			|raw|<div class="broadcast-green">${privacy} chat room <b>${oldTitle}</b> renamed to <b>${target}</b></div>
 		`;
```
Zarel, line 870: wtf wait what wtf why wtf

Zarel: how long has this been like this? ugh

PR #6967, opened 2020-07-06
Fix bugs from recent restrictions update

KrisXV:


scheibo:
Its a little confusing to identify the fix when you mix in in a bunch of formatting changes. 

Could you also consider adding a test if its not too difficult? Thank you for fixing this!

config/formats.ts
```diff
@@ -676,7 +676,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 				if (!item || !item.megaStone) continue;
 				const species = this.dex.getSpecies(set.species);
 				if (species.isNonstandard) return [`${species.baseSpecies} does not exist in gen 8.`];
-				if (this.ruleTable.isRestricted('pokemon:' + species.id)) {
+				if (this.ruleTable.isRestricted(`pokemon:${species.id}`) || this.ruleTable.isRestricted(`basepokemon:${species.id}`)) {
```
KrisXV, line 679: this was the fix @scheibo

data/rulesets.ts
```diff
@@ -861,7 +861,10 @@ export const BattleFormats: {[k: string]: FormatsData} = {
 		desc: "Prevents Pok&eacute;mon on the Ubers dynamax banlist from dynamaxing",
 		onBegin() {
 			for (const pokemon of this.getAllPokemon()) {
-				if (this.ruleTable.isRestricted('pokemon:' + pokemon.species.id)) pokemon.canDynamax = false;
+				if (this.ruleTable.isRestricted(`pokemon:${pokemon.species.id}`) ||
+					this.ruleTable.isRestricted(`basepokemon:${pokemon.species.id}`)) {
+					pokemon.canDynamax = false;
```
scheibo, line 866: are you sure its not this @Kris? this seems changed as well?

KrisXV: This was also a fix (and my @ is not kris), i was pointing out that the fix was adding `basepokemon` checks in general

Zarel:
@urkerab can you review?

config/formats.ts
Zarel: I would personally make a function that runs through the entirety of:

https://github.com/smogon/pokemon-showdown/blob/c124720885419d05f214a63ca20616390abd9299/sim/team-validator.ts#L1239-L1338

but for `*` rather than `-`.

urkerab:
> @urkerab can you review?

If you need to check `basepokemon`, shouldn't you use `toID(species.baseSpecies)`?

I used the `+` style because that's how it's done in `sim/team-validator.ts` (e.g. the `ruleTable.check` invocations).

I thought I had checked to ensure that the restrictions worked but obviously I had checked the wrong species...

Zarel:
@urkerab I think the long-term correct fix is to copy/paste `checkSpecies` into a new function but for `*` instead of `-`, as I mentioned in the comment. Is that something you can do?

urkerab:
Something like this in `sim/team-validator.ts`?
```typescript
        isRestrictedSpecies(template: Template) {
                if (ruleTable.has('*pokemon:' + template.id)) return true;
                if (ruleTable.has('+pokemon:' + template.id)) return false;
                return ruleTable.has('*basepokemon' + toID(template.baseSpecies));
        }
```

urkerab:
... I meant `sim/dex-data.ts` of course. It's no use in `sim/team-validator.ts`, since battles need it.

Zarel:
I think `sim/dex` might be more appropriate, but yes.

PR #6971, opened 2020-07-07
Trivia: Changes to the random category selector

AnnikaCodes:
This will make it so that creating a Trivia game in a random category won't select the same category as the last game. (And if the last game was in the All "category", it's free to select any category.)


Zarel:
Looks cool!

PR #6973, opened 2020-07-07
Centralise the check for a restricted species

urkerab:
I put all the tier stuff in there too, just in case someone wants to be able to restrict a tier (using Mix and Mega as an example, it would be able to set restricted: Uber, Gengar, Zeraora; unbanlist: Darmanitan, Dracovish, Lunala, Zamazenta).

### DIFF
--- a/sim/dex-data.ts
+++ b/sim/dex-data.ts
@@ -187,6 +187,20 @@ export class RuleTable extends Map<string, string> {
 		return this.has(`*${thing}`);
 	}
 
+	isRestrictedSpecies(species: Species) {
+		if (this.has(`+pokemon:${species.id}`)) return false;
+		if (this.has(`*pokemon:${species.id}`)) return true;
+		if (this.has(`+basepokemon:${toID(species.baseSpecies)}`)) return false;
+		if (this.has(`*basepokemon:${toID(species.baseSpecies)}`)) return true;
+		const tier = species.tier === '(PU)' ? 'ZU' : species.tier === '(NU)' ? 'PU' : species.tier;
+		if (this.has(`+pokemontag:${toID(tier)}`)) return false;
+		if (this.has(`*pokemontag:${toID(tier)}`)) return true;
+		const doublesTier = species.doublesTier === '(DUU)' ? 'DNU' : species.doublesTier;
+		if (this.has(`+pokemontag:${toID(doublesTier)}`)) return false;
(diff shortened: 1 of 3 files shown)


sim/dex-data.ts
```diff
@@ -187,6 +187,20 @@ export class RuleTable extends Map<string, string> {
...
+		if (this.has('*pokemon:' + species.id)) return true;
+		if (this.has('+basepokemon:' + toID(species.baseSpecies))) return false;
+		if (this.has('*basepokemon:' + toID(species.baseSpecies))) return true;
+		const tier = species.tier === '(PU)' ? 'ZU' : species.tier === '(NU)' ? 'PU' : species.tier;
+		if (this.has('+pokemontag:' + toID(tier))) return false;
+		if (this.has('*pokemontag:' + toID(tier))) return true;
+		const doublesTier = species.doublesTier === '(DUU)' ? 'DNU' : species.doublesTier;
+		if (this.has('+pokemontag:' + toID(doublesTier))) return false;
+		if (this.has('*pokemontag:' + toID(doublesTier))) return true;
+		return this.has('*pokemontag:allpokemon');
```
KrisXV, line 201: nitpick but shouldnt these all be template strings to remain consistent with other parts of this file?

urkerab: Oops; as I mentioned in #6967 I originally thought of putting this in a different file where the style was different.

sim/dex-data.ts
```diff
@@ -187,6 +187,20 @@ export class RuleTable extends Map<string, string> {
 		return this.has(`*${thing}`);
 	}
 
+	isRestrictedSpecies(species: Species) {
+		if (this.has(`*pokemon:${species.id}`)) return true;
+		if (this.has(`+pokemon:${species.id}`)) return false;
```
Zarel, line 192: Aren't these all backwards? Shouldn't the `+` check come first?

urkerab: I can swap them over to make more sense to you, but it makes no difference because you can't have both versions in the same rule table at once.

Zarel:
Thanks!

sim/dex-data.ts
Zarel: I think there might be complicated interactions with subrules that let it happen. Either way, there's nothing in the type system disallowing it.

PR #6974, opened 2020-07-07
Daily spotlights: Add queue management commands

AnnikaCodes:
As per the suggestion [here](https://www.smogon.com/forums/threads/allow-one-queued-daily-to-be-swapped-out-for-another.3666432/).


Zarel:
I wonder if we should have generalized list management helpers somewhere, so this stuff doesn't need to be repeated every time someone wants to reorder a list...

PR #6978, opened 2020-07-07
Move Orbeetle and Rotom-Frost to PUBL

KrisXV:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/#post-8530565


Zarel:
Thanks!

PR #6981, opened 2020-07-08
Datasearch: Support /ds monotype without a type

AnnikaCodes:
When I originally coded this, I assumed no one wanted to see all monotype Pokémon, but apparently that's a useful thing so I changed the implementation to permit it.


Zarel:
Neat! I like how it also simplifies the code a little.

server/chat-plugins/datasearch.ts
```diff
@@ -1043,6 +1035,7 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 	}
 	let results: string[] = [];
 	for (const mon of Object.keys(dex).sort()) {
+		if (singleTypeSearch && Object.keys(dex[mon].types).length !== 1) continue;
```
Zarel, line 1038: Oh, wait, you shouldn't be using this.

```suggestion
		if (singleTypeSearch && dex[mon].types.length !== 1) continue;
```

AnnikaCodes: Alright, I'll open another PR for that.

PR #6983, opened 2020-07-09
Add type and category support for <psicon>

KrisXV:


KrisXV:
@Zarel this needs merged alongside the client side portion

Zarel:
I know.

This will require either a week or a server restart, though.

PR #6985, opened 2020-07-09
Daily spotlights: Fix regex to allow whitespace

AnnikaCodes:


Zarel:
I'd probably use `.trim()` rather than a regex for this sort of thing?

PR #6986, opened 2020-07-09
Fix pollqueue page crash

mia-pi-git:
/j view-pollqueue without a room crashes, since there's no ``args[0]`` to run Rooms.get on.

### DIFF
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -542,7 +542,8 @@ export const commands: ChatCommands = {
 
 export const pages: PageTable = {
 	pollqueue(args, user) {
-		const room = Rooms.get(args[0]) as ChatRoom | GameRoom;
+		this.extractRoom();
+		const room = this.room;
 		if (!room) return `<div class="pad"><h2>Specify a room.</h2></div>`;
 		let buf = `<div class="pad"><strong>Queued polls:</strong>`;
 		buf += `<button class="button" name="send" value="/join view-pollqueue-${room.roomid}" style="float: right">`;
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -542,9 +542,9 @@ export const commands: ChatCommands = {
 
 export const pages: PageTable = {
 	pollqueue(args, user) {
-		this.extractRoom();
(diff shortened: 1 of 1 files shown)


server/chat-plugins/poll.ts
```diff
@@ -542,9 +542,9 @@ export const commands: ChatCommands = {
 
 export const pages: PageTable = {
 	pollqueue(args, user) {
-		this.extractRoom();
 		const room = Rooms.get(args[0]) as ChatRoom | GameRoom;
```
Zarel, line 545: Shouldn't this just be `this.extractRoom(); const room = this.room;`?

server/chat-plugins/poll.ts
```diff
@@ -543,8 +543,9 @@ export const commands: ChatCommands = {
 export const pages: PageTable = {
 	pollqueue(args, user) {
 		this.extractRoom();
-		const room = Rooms.get(args[0]) as ChatRoom | GameRoom;
-		let buf = `<div class = "pad"><strong>Queued polls:</strong>`;
+		const room = this.room;
+		if (!room) return `<div class="pad"><h2>Specify a room.</h2></div>`;
```
Zarel, line 547: We should probably change it so `extractRoom` can throw an error message for requiring a room.

Zarel:
The fix is good, though.

whalemer:
`extractRoom` already returns an error message;

https://github.com/smogon/pokemon-showdown/blob/6690cfb037d2730e46318877947a85fae63c7554/server/chat.ts#L252-L254
```
>> |/j view-pollqueue-notaroom 
<< >view-pollqueue-notaroom
|init|html
|title|Page
|pagehtml|<h2>Invalid room.</h2> 
<< >view-pollqueue-notaroom
|title|Page
|pagehtml|<div class="pad"><h2>Specify a room.</h2></div>
```

Zarel:
@whalemer Another bug that wouldn't have existed if we'd used `Chat.ErrorMessage`. ;)

whalemer:
>Another bug that wouldn't have existed if we'd used Chat.ErrorMessage. ;)

loath as I am to admit it, yeah, this is a pretty perfect use for it.

PR #6987, opened 2020-07-09
Trim !show image comment

TheMezStrikes:


Zarel:
Trimming and the lack thereof seems to be a frequent issue. I wonder if there's something that can make it easier.

PR #6988, opened 2020-07-09
Fix trustuser commands

asgdf:
Currently, untrustuser straight up doesn't work, presumably because things are set up slightly differently now.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -467,8 +467,8 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 
 		mod: 'gen8',
 		ruleset: ['Standard', 'Dynamax Clause'],
-		banlist: ['Baton Pass', 'King\'s Rock'],
-		restricted: ['Shedinja', 'Lunala', 'Solgaleo', 'Type: Null', 'Huge Power', 'Pure Power', 'Gorilla Tactics', 'Shadow Tag'],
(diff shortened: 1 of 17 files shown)


Zarel:
Looks good, although I wonder if these should just alias to `/globalwhitelist` and `/globaldeauth`.

Zarel:
(conflict)

asgdf:
In terms of conflict resolution this looks about right I think?

I haven't been paying attention to the whitelist stuff, does "whitelist" supplant the "manual trustuser" concept?

asgdf:
o shoot I forgot to push my updated version that fixes those warnings

Zarel:
> I haven't been paying attention to the whitelist stuff, does "whitelist" supplant the "manual trustuser" concept?

"Whitelist" is a new rank between regular user and voice, which does not have a symbol (they look exactly the same as regular users).

It's implemented in the code as having an explicit rank equal to the implicit default rank - basically exactly the same way explicitly trusted users were implemented in the past. I don't know if "supplant" is the right word; it was implemented for room ranks, but the implementation is generic enough that it should work fine for global ranks as well.

asgdf:
It seems theoretically possible to refactor trustuser to use that system, though I imagine it would involve somewhat more effort than just setting up some aliases

Zarel:
Fair. This is a good incremental step.

PR #6990, opened 2020-07-10
Linked: Fix priority of linked moves

Slayer95:
Second move in a chain ignores Gen 8 dynamic turn-order mechanic.

### DIFF
--- a/data/mods/linked/scripts.ts
+++ b/data/mods/linked/scripts.ts
@@ -394,7 +394,8 @@ export const BattleScripts: ModdedBattleScriptsData = {
 
 		if (this.gen < 5) this.eachEvent('Update');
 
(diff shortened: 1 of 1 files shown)


KrisXV:
Does this also fix the issue with the first move in a link that’s boosted by triage has +6 priority and Snipe shot hitting the user when used in a link?

Slayer95:
From reading the description of ``Snipe Shot``, it sounds like something that might happen if the foe is KO-ed by the first move, which *could be* a carry-over bug from unmodded game.

No clue what the deal with Triage is, but from its desc it sounds like the same sort of ability Gale Wings is, which is fixed.

KrisXV:
Snipe Shot has the user hit itself in a link if the first move is Focus Energy as well

PR #6991, opened 2020-07-10
Fix Fusion Bolt + Fusion Flare interaction when the first move fails.

jmclemo6:
Fixes https://github.com/smogon/pokemon-showdown/projects/3#card-25870231

### DIFF
--- a/data/scripts.ts
+++ b/data/scripts.ts
@@ -96,7 +96,7 @@ export const BattleScripts: BattleScriptsData = {
 			pokemon.side.zMoveUsed = true;
 		}
 		const moveDidSomething = this.useMove(baseMove, pokemon, target, sourceEffect, zMove, maxMove);
-		this.lastSuccessfulMoveThisTurn = moveDidSomething ? this.activeMove : null;
+		this.lastSuccessfulMoveThisTurn = moveDidSomething ? this.activeMove && this.activeMove.id : null;
 		if (this.activeMove) move = this.activeMove;
 		this.singleEvent('AfterMove', move, null, pokemon, target, move);
 		this.runEvent('AfterMove', pokemon, target, move);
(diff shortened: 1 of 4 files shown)


Zarel:
Nice, thanks!

data/scripts.ts
```diff
@@ -96,6 +96,7 @@ export const BattleScripts: BattleScriptsData = {
 			pokemon.side.zMoveUsed = true;
 		}
 		const moveDidSomething = this.useMove(baseMove, pokemon, target, sourceEffect, zMove, maxMove);
+		this.lastSuccessfulMoveThisTurn = moveDidSomething ? this.activeMove && this.activeMove.id : null;
```
urkerab, line 99: Wouldn't `this.activeMove?.id` have worked here?

jmclemo6: `this.activeMove?.id` short circuits with a value of `undefined` if `activeMove` is nullish. Admittedly, this is my first contribution to PS in awhile, but I seemed to remember the convention being `T | null` for optional properties like this. I could go back and update the type to `T | null | undefined` and be able to use optional chaining here.

Zarel: Clearly that means it should be `(moveDidSomething && this.activeMove?.activeMove.id) || null`

jmclemo6: Eh, I personally wouldn't lean towards that as I think that's less readable then just skipping the optional chaining here, but I can make that update if people agree on it.

urkerab: Oh, it does? That's annoying. Never mind then.

Zarel: I was half-shitposting. I agree that `&&` being used for its return value is less readable than `||` being used for its return value.

PR #6992, opened 2020-07-11
Move Gallade to NUBL

KrisXV:
https://www.smogon.com/forums/threads/np-nu-stage-2-dont-start-now-gallade-banned.3666717/#post-8533561


Zarel:
Thanks!

PR #6993, opened 2020-07-11
Set global bot permissions in groupchats to bot

TheMezStrikes:
Currently, because global bots inherit from %, their groupchat jurisdiction is actually lowered to @, denying them the (valuable) addhtml permission


Zarel:
Good catch!

scheibo:
I've updated `config/config.js` on the the sim server.

PR #6999, opened 2020-07-13
Move Butterfree and Kangaskhan to PUBL

KrisXV:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/#post-8535959


Zarel:
Thanks!

PR #7001, opened 2020-07-13
Handle Gigantamaxing properly

KrisXV:


sim/team-validator.ts
```diff
@@ -569,7 +573,7 @@ export class TeamValidator {
 		const lsetProblems = this.reconcileLearnset(outOfBattleSpecies, setSources, lsetProblem, name);
 		if (lsetProblems) problems.push(...lsetProblems);
 
-		if (ruleTable.has('obtainablemisc') && outOfBattleSpecies.forme?.includes('Gmax')) {
+		if (ruleTable.has('obtainablemisc') && set.gigantamax) {
 			if (!setSources.sourcesBefore) {
 				problems.push(`${name} has an exclusive move that it doesn't qualify for (because Gmax Pokemon can only be obtained from a Max Raid).`);
```
urkerab, line 578: Well this isn't true any more is it?

KrisXV: Probably not, but I wanted to touch validation as little as possible

Zarel: I would add a flag for this so we could simulate the old behavior, for the record. But yeah, probably a separate PR.

TheImmortal:
Hmm. I'd probably rename it to ``gmaxFactor``. You can have Charmeleon and such with the Gigantamax Factor in-game which, if transformed into a Pokemon with a Gigantamax, Dynamaxes into the Gigatamax forme. Whereas if it does not have the Gigantamax Factor, it would change into the regular Dynamax.

We should be able to set Gigantamax Factor for Pokemon in the teambuilder not only for those pre-evolution cases, but because it is hackable and thus affects Hackmons.

Lusamine:
> Hmm. I'd probably rename it to `gmaxFactor`. You can have Charmeleon and such with the Gigantamax Factor in-game which, if transformed into a Pokemon with a Gigantamax, Dynamaxes into the Gigatamax forme. Whereas if it does not have the Gigantamax Factor, it would change into the regular Dynamax.

I haven't tested post-DLC but [initially when I did this test](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8339360), GMax factor Charmander and Charmeleon transformed into a mon with GMax factor was no different from the regular Dynamax version of that mon.

TheImmortal:
> > Hmm. I'd probably rename it to `gmaxFactor`. You can have Charmeleon and such with the Gigantamax Factor in-game which, if transformed into a Pokemon with a Gigantamax, Dynamaxes into the Gigatamax forme. Whereas if it does not have the Gigantamax Factor, it would change into the regular Dynamax.
> 
> I haven't tested post-DLC but [initially when I did this test](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8339360), GMax factor Charmander and Charmeleon transformed into a mon with GMax factor was no different from the regular Dynamax version of that mon.

Oops, yeah. I mixed up what you tested.

Zarel:
I'd consider `canGigantamax` clearer than `gigantamax`, though.

data/pokedex.ts
```diff
@@ -40,8 +40,9 @@ export const BattlePokedex: {[speciesid: string]: SpeciesData} = {
 		prevo: "Ivysaur",
 		evoLevel: 32,
 		eggGroups: ["Monster", "Grass"],
-		otherFormes: ["Venusaur-Mega", "Venusaur-Gmax"],
+		otherFormes: ["Venusaur-Mega"],
 		formeOrder: ["Venusaur", "Venusaur-Mega", "Venusaur-Gmax"],
```
Zarel, line 44: tbh I think it makes more sense to remove Gmaxes from `formeOrder` than from `otherFormes`.

Zarel: Ideally both. Maybe mark Gmax formes with `isNonstandard: "Gigantamax"`?

test/sim/team-validator.js
```diff
@@ -476,42 +476,34 @@ describe('Team Validator', function () {
...
-			{species: 'charizard-gmax', ability: 'blaze', moves: ['gmaxwildfire'], evs: {hp: 1}},
+			{species: 'charizard', ability: 'blaze', moves: ['gmaxwildfire'], evs: {hp: 1}, gigantamax: true},
 		];
 		let illegal = TeamValidator.get('gen8anythinggoes').validateTeam(team);
 		assert(illegal);
 		illegal = TeamValidator.get('gen8customgame@@@-nonexistent').validateTeam(team);
 		assert(illegal);
 	});
 
-	it('should reject Gmax Pokemon from formats with Dynamax Clause', function () {
```
Zarel, line 507: I see no reason to remove this test; I think it should be revised to `gigantamax: true` like the others.

Zarel: We should also keep some tests using the `species: '...-gmax'` format, to make sure the validator is correctly converting those to `gigantamax: true`.

KrisXV: This test is no longer necessary since the clause doesn’t have any validation on it anymore

data/pokedex.ts
KrisXV: They are literally not formes in any sense of the word, I'd rather not remove them from formeOrder because I believe the sprite project needs them (@scheibo can you confirm/deny?). The only reason they have entries in `pokedex.ts` and `formats-data.ts` is for backwards compatibility and convenience for the teambuilder and random battles.

KrisXV: I will add a `isNonstandard: "Gigantamax"` option though.

scheibo: I thought they were only added into `formeOrder` because Zarel had wanted them there (https://github.com/smogon/pokemon-showdown/pull/6669#issuecomment-626109797). Not entirely sure though, the discussion might be in #art? Shinies are not formes either and the sprite project handles them. I dont think non-formes belong in `formeOrder`.

scheibo: > I will add a isNonstandard: "Gigantamax" option though.

wait why. I specifically removed this. GIGANTAMAXES ARE NOT NONSTANDARD. please do not mark them as such

Zarel: I wanted them in `otherFormes`. My comment you linked to was specifically about a strategy to use `otherFormes` rather than having a separate `formeOrder` array. If we do have a separate `formeOrder` array, it makes sense to leave off non-formes like Gigantamaxes from it.

Zarel: Also, with this PR comes a new strategy for handling gigantamaxes that I no longer think it's necessary to have gigantamaxes in `otherFormes`. Well, you could make arguments either way, it's just not a slam-dunk anymore.

Zarel: I'm not opposed to just using `isGigantamax` as a stopgap without adding the `isNonstandard` flag.

scheibo:
To elaborate: my use case for `isNonstandard` is that I desire an easy way to filter out data which is impossible on the cartridge in a certain generation - things which are PS quirks. The `Nonstandard` type enum is deliberately vaguely named and I agree it makes no guarantees about what it is to be used for, but up until now, simply removing `isNonstandard` data is sufficient for getting rid of all the crap I don't want.

Gigantamax Moves are not handled in the standard way moves are in a game (ie not 'indexable'), so it might make sense to call them 'Nonstandard', and it seems like this is much ado about nothing because I would just need to change my filter to `isNonstandard && isNonstandard !== 'Gigantamax'`, only I imagine that is going to expand when we make up new ways things can be 'nonstandard' in the future.

I don't know if Nonstandard is the correct field for what I want to acheive, maybe it makes more sense to be able to repurpose for all sorts of non-standard behavior like gmax moves/species (though I feel its important to chose a general name for that like 'Nonindexable' or 'Nonhackable' etc, though both aren't amazing, for the reasons I espoused on discord), however, I strongly believe **there should be a trivial way to filter out data which is impossible on the cartridge in a certain generation that isn't going to be changed depending on implementation details.** Maybe a new field needs to be added for my purpose, but I think it is a very valid purpose, not just for my projects, but for any project that uses PS data and just wants to be able to work with the data Game Freak provided without having to figure out a bunch of gotchas.

KrisXV:
okay @Zarel @scheibo @TheImmortal this should be good, I'd rather tackle the moving gmax moves inside of max moves and making a new message to send to the client in separate pull requests

Zarel:
I'm confused why this wasn't `isNonstandard: "Gigantamax"`? `Custom` seems unnecessarily generic for this use-case.

Zarel:
<3

scheibo:
I approved this when it was `'Custom'`, not `'Gigantamax'`, which as I explained [above](https://github.com/smogon/pokemon-showdown/pull/7001#issuecomment-660375104) feels *way* too specific. I agree `'Custom'` is perhaps too generic, but something like `'Virtual'` or one of the other options that outlines that it is not a real concept is more desirable

Zarel:
It really shouldn't be blocked on; this PR was blocking a lot more data refactors. This merge should not be interpreted as a "final answer" for anything like that.

Your linked post doesn't say anything about "Gigantamax" being way too specific. Did you link to the wrong post?

`"Custom"` is clearly the wrong name because it's the `Nonstandard` label we use for SSB moves. I would be fine with `"Virtual"`.

Zarel:
To be more precise: I would hate `Virtual` (it's less clear/informative than `Gigantamax` and doesn't solve any of the underlying problems) but I'm willing to compromise.

TheImmortal:
I think Custom is the right label. It is data that exists in the game but isn’t Hackable. Aka data that PS only has for simplicity / ease of use.

PR #7003, opened 2020-07-14
Cross Evolution: Fix validating with the set's species

KrisXV:
Currently, a set with a banned Pokemon as the species that is named after an unbanned evolution allows the Pokemon to be used (i.e a Type: Null named "Urshifu" validates), and it shouldn't.


Zarel:
Good catch!

urkerab:
Out of interest, why doesn't the `this.validateSet` catch that case?

Zarel:
Actually, that's a good point...

PR #7005, opened 2020-07-14
Support persisting battles through restarts

mia-pi-git:
All of this is complete except for the rejoining battles function. the `getPlayerIDs` function requires having the input log from pre-restart to be parsed, and unfortunately Streams requires the acquisition of the input log to be async. So, i'm not sure if there's a better way to handle that, or not. That being said, the actual writing of state / loading state / handling chat messages in input logs / getting input logs is done. Will leave this as draft until that gets sorted out.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -369,6 +369,11 @@ exports.logchallenges = false;
 exports.loguserstats = 1000 * 60 * 10; // 10 minutes
 
 /**
+ * persistbattles - Whether or not to persist battles through server restarts.
+*/
+exports.persistbattles = true;
+
+/**
  * validatorprocesses - the number of processes to use for validating teams
  * simulatorprocesses - the number of processes to use for handling battles
  * You should leave both of these at 1 unless your server has a very large
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -697,17 +697,17 @@ export const commands: ChatCommands = {
 
 		const logRoom = Rooms.get('staff') || Rooms.lobby || room;
 
-		if (!logRoom?.log.roomlogStream) return process.exit();
+		if (!logRoom?.log.roomlogStream) return Rooms.global.kill();
 
 		logRoom.roomlog(`${user.name} used /kill`);
 
 		void logRoom.log.roomlogStream.writeEnd().then(() => {
-			process.exit();
+			void Rooms.global.kill();
 		});
 
 		// In the case the above never terminates
 		setTimeout(() => {
-			process.exit();
+			void Rooms.global.kill();
 		}, 10000);
 	},
 	killhelp: [`/kill - kills the server. Can't be done unless the server is in lockdown state. Requires: &`],
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -684,7 +684,7 @@ export const commands: ChatCommands = {
 		this.stafflog(`${user.name} used /endemergency.`);
 	},
 
-	async kill(target, room, user) {
+	kill(target, room, user) {
 		if (!this.can('lockdown')) return false;
 
 		if (Rooms.global.lockdown !== true) {
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -684,7 +684,7 @@ export const commands: ChatCommands = {
 		this.stafflog(`${user.name} used /endemergency.`);
 	},
 
-	kill(target, room, user) {
+	async kill(target, room, user) {
 		if (!this.can('lockdown')) return false;
 
 		if (Rooms.global.lockdown !== true) {
@@ -701,9 +701,10 @@ export const commands: ChatCommands = {
 
 		logRoom.roomlog(`${user.name} used /kill`);
 
-		void logRoom.log.roomlogStream.writeEnd().then(() => {
-			void Rooms.global.kill();
-		});
+		void Promise.all([
+			logRoom.log.roomlogStream.writeEnd(),
+			Rooms.global.kill(),
+		]);
 
 		// In the case the above never terminates
 		setTimeout(() => {
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1139,7 +1139,12 @@ export const commands: ChatCommands = {
 	restoreplayershelp: [
 		`/restoreplayers - Restore previous players in an imported input log.`,
 	],
-
+	rejoinbattle(target, room, user) {
+		if (!user.rejoinBattle()) return this.errorReply(`You do not have a battle you can rejoin.`);
+	},
+	rejoinbattlehelp: [
+		`/rejoinbattle - rejoins a battle that has been restored after a server restart.`,
+	],
 	joinbattle: 'joingame',
 	joingame(target, room, user) {
 		if (!room) return this.requiresRoom();
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1139,6 +1139,7 @@ export const commands: ChatCommands = {
 	restoreplayershelp: [
 		`/restoreplayers - Restore previous players in an imported input log.`,
 	],
+
 	joinbattle: 'joingame',
 	joingame(target, room, user) {
 		if (!room) return this.requiresRoom();
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1139,12 +1139,6 @@ export const commands: ChatCommands = {
 	restoreplayershelp: [
 		`/restoreplayers - Restore previous players in an imported input log.`,
 	],
-	rejoinbattle(target, room, user) {
-		if (!user.rejoinBattle()) return this.errorReply(`You do not have a battle you can rejoin.`);
-	},
-	rejoinbattlehelp: [
-		`/rejoinbattle - rejoins a battle that has been restored after a server restart.`,
-	],
 	joinbattle: 'joingame',
 	joingame(target, room, user) {
 		if (!room) return this.requiresRoom();
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1154,6 +1154,7 @@ export class CommandContext extends MessageContext {
 		if (Chat.filters.length) {
 			return Chat.filter(this, message, user, room, connection, targetUser);
 		}
+
 		return message;
 	}
 	canPMHTML(targetUser: User | null) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1119,8 +1119,13 @@ export class CommandContext extends MessageContext {
 			return null;
 		}
 		const gameFilter = this.checkGameFilter();
+<<<<<<< HEAD
 		if (typeof gameFilter === 'string') {
 			if (gameFilter) this.errorReply(gameFilter);
+=======
+		if (gameFilter && !user.can('bypassall')) {
+			this.errorReply(gameFilter);
+>>>>>>> stuff
 			return null;
 		}
 
@@ -1130,7 +1135,7 @@ export class CommandContext extends MessageContext {
 				!user.can('bypassall') && (['help', 'lobby'].includes(room.roomid)) && (normalized === user.lastMessage) &&
 				((Date.now() - user.lastMessageTime) < MESSAGE_COOLDOWN)
 			) {
-				this.errorReply(this.tr("You can't send the same message again so soon."));
+				this.errorReply(this.tr("You can't send sthe same message again so soon."));
 				return null;
 			}
 			user.lastMessage = message;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1118,6 +1118,7 @@ export class CommandContext extends MessageContext {
 			this.errorReply(this.tr("Your message contained banned words in this room."));
 			return null;
 		}
+
 		const gameFilter = this.checkGameFilter();
 		if (typeof gameFilter === 'string') {
 			if (gameFilter) this.errorReply(gameFilter);
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1135,7 +1135,7 @@ export class CommandContext extends MessageContext {
 				!user.can('bypassall') && (['help', 'lobby'].includes(room.roomid)) && (normalized === user.lastMessage) &&
 				((Date.now() - user.lastMessageTime) < MESSAGE_COOLDOWN)
 			) {
-				this.errorReply(this.tr("You can't send sthe same message again so soon."));
+				this.errorReply(this.tr("You can't send the same message again so soon."));
 				return null;
 			}
 			user.lastMessage = message;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1149,6 +1149,7 @@ export class CommandContext extends MessageContext {
 		if (Chat.filters.length) {
 			return Chat.filter(this, message, user, room, connection, targetUser);
 		}
+
 		return message;
 	}
 	canPMHTML(targetUser: User | null) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1118,7 +1118,6 @@ export class CommandContext extends MessageContext {
 			this.errorReply(this.tr("Your message contained banned words in this room."));
 			return null;
 		}
-
 		const gameFilter = this.checkGameFilter();
 		if (typeof gameFilter === 'string') {
 			if (gameFilter) this.errorReply(gameFilter);
@@ -1150,7 +1149,6 @@ export class CommandContext extends MessageContext {
 		if (Chat.filters.length) {
 			return Chat.filter(this, message, user, room, connection, targetUser);
 		}
-
 		return message;
 	}
 	canPMHTML(targetUser: User | null) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1149,7 +1149,6 @@ export class CommandContext extends MessageContext {
 		if (Chat.filters.length) {
 			return Chat.filter(this, message, user, room, connection, targetUser);
 		}
-
 		return message;
 	}
 	canPMHTML(targetUser: User | null) {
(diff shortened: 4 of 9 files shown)


mia-pi-git:
On reflection, that doesn't really need to be a separate function at all, so this can be simplified a lot and un-drafted.

server/room-battle.ts
```diff
@@ -1111,6 +1114,16 @@ export class RoomBattle extends RoomGames.RoomGame {
 		// @ts-ignore
 		this.room = null;
 	}
+	onChatMessage(message: string, user: User) {
+		void this.stream.write(`>chat ${user.getIdentity(this.room.roomid)}|${message}`);
+		return ' ';
+	}
+	getLog(): Promise<string[]> {
```
Zarel, line 1121: This would be more readable as async.

server/room-battle.ts
```diff
@@ -1111,6 +1114,16 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+		void this.stream.write('>requestlog');
+		return this.stream.read().then(() => {
```
Zarel, line 1123: This is super brittle and a race condition just waiting to happen.

Zarel: I'm actually surprised this works at all, this should nearly definitely result in data loss: You're reading from the stream and discarding its data.

server/room-battle.ts
```diff
@@ -1111,6 +1114,16 @@ export class RoomBattle extends RoomGames.RoomGame {
 		// @ts-ignore
 		this.room = null;
 	}
+	onChatMessage(message: string, user: User) {
+		void this.stream.write(`>chat ${user.getIdentity(this.room.roomid)}|${message}`);
+		return ' ';
+	}
```
Zarel, line 1120: Instead of making the entire chat message roundtrip from battles, I would probably just ignore chat messages coming from the subprocess.

Zarel: Might as well make chat messages include timestamps, too.

server/room-battle.ts
```diff
@@ -726,6 +726,10 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
+		case 'log':
```
Zarel, line 729: I suspect `inputlog` would be a better name than `log`.

server/chat-commands/admin.ts
```diff
@@ -749,17 +749,17 @@ export const commands: ChatCommands = {
...
 		const logRoom = Rooms.get('staff') || Rooms.lobby || room;
 
-		if (!logRoom?.log.roomlogStream) return process.exit();
+		if (!logRoom?.log.roomlogStream) return Rooms.global.kill();
 
 		logRoom.roomlog(`${user.name} used /kill`);
 
 		void logRoom.log.roomlogStream.writeEnd().then(() => {
-			process.exit();
+			void Rooms.global.kill();
```
Zarel, line 757: This should not be done serially, but in parallel. Do both at the same time, and then wait for both with `Promise.all`.

server/chat-commands/core.ts
```diff
@@ -1068,7 +1068,12 @@ export const commands: ChatCommands = {
 	restoreplayershelp: [
 		`/restoreplayers - Restore previous players in an imported input log.`,
 	],
-
+	rejoinbattle(target, room, user) {
+		if (!user.rejoinBattle()) return this.errorReply(`You do not have a battle you can rejoin.`);
```
Zarel, line 1072: This probably shouldn't be a command at all; it should just happen automatically, right? Maybe have it as part of `/autojoin`.

server/chat.ts
```diff
@@ -1049,10 +1049,9 @@ export class CommandContext extends MessageContext {
 			this.errorReply(this.tr("Your message contained banned words in this room."));
 			return null;
 		}
-
 		const gameFilter = this.checkGameFilter();
-		if (gameFilter && !user.can('bypassall')) {
-			this.errorReply(gameFilter);
+		if (gameFilter && (this.room?.battle || room?.battle || !user.can('bypassall'))) {
+			this.errorReply(gameFilter.trim());
```
Zarel, line 1054: Oh, I didn't even realize how weird this workaround is. For future reference, you should use a sentinel value like `true` rather than this weirdness (which will send an empty error for no reason).

And anyway, all this is moot because for lag reasons it's much better to just let this one through, and ignore the one that comes from the battle process.

server/room-battle.ts
```diff
@@ -528,7 +528,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 		this.logData = null;
 		this.endType = 'normal';
 		this.score = null;
-		this.inputLog = null;
+		this.inputLog = options.inputLog ? options.inputLog.split('\n') : null;
```
Zarel, line 531: I'm not sure if this actually affects anything...

server/room-battle.ts
```diff
@@ -726,6 +726,10 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
+		case 'log':
+			if (!this.logData) this.logData = {};
+			this.logData.inputLog = lines.slice(1);
```
Zarel, line 731: So we have a `battle.inputLog` and a `battle.logData.inputLog` which are usually out of sync? That seems pretty ugly tbh...

server/rooms.ts
```diff
@@ -1363,17 +1366,53 @@ export class GlobalRoomState {
...
+	async kill() {
+		if (Config.persistbattles) await this.writeBattleState();
+		return process.exit();
+	}
+	async writeBattleState() {
+		const buffer: AnyObject = {};
+		for (const [id, room] of Rooms.rooms) {
+			if (!room.battle) continue;
+			const formatid = room.battle.format;
+			const logData = await room.battle.getLog();
```
Zarel, line 1389: Wow, it does all this in serial? That would make a lockdown take forever. These should probably be in parallel.

server/rooms.ts
```diff
@@ -1363,17 +1366,53 @@ export class GlobalRoomState {
...
+			if (!logData) continue;
+			if (!buffer[formatid]) buffer[formatid] = {};
+			buffer[formatid][id] = {
+				inputLog: logData.map(item => item.replace(/\r/g, '')).join('\n'),
+				title: room.title,
+				num: room.roomid.split('-')[2],
```
Zarel, line 1395: Why not just log the entire roomid?

server/rooms.ts
```diff
@@ -1363,17 +1366,53 @@ export class GlobalRoomState {
...
+		const buffer: AnyObject = {};
+		for (const [id, room] of Rooms.rooms) {
+			if (!room.battle) continue;
+			const formatid = room.battle.format;
+			const logData = await room.battle.getLog();
+			if (!logData) continue;
+			if (!buffer[formatid]) buffer[formatid] = {};
+			buffer[formatid][id] = {
+				inputLog: logData.map(item => item.replace(/\r/g, '')).join('\n'),
+				title: room.title,
```
Zarel, line 1394: I wonder if the title should just be reconstructed from the inputlog. Maybe this is easier, though.

server/rooms.ts
```diff
@@ -1363,17 +1366,53 @@ export class GlobalRoomState {
...
+				num: room.roomid.split('-')[2],
+			};
+		}
+		return FS(`logs/battles.json`).writeUpdate(() => JSON.stringify(buffer));
+	}
+	loadBattleState() {
+		if (!Config.persistbattles) return;
+		const battleData = JSON.parse(FS(`logs/battles.json`).readIfExistsSync() || "{}");
```
Zarel, line 1402: We should probably delete battle state after reading it, or it could cause crash loops and other bad crash behavior.

server/rooms.ts
```diff
@@ -1363,17 +1366,53 @@ export class GlobalRoomState {
...
+		for (const formatid in battleData) {
+			for (const battle in battleData[formatid]) {
+				const entry = battleData[formatid][battle];
+				const battleRoom = Rooms.createBattle(formatid, entry);
+				// ensure log is at the end of the battle so that users rejoining see it
+				setTimeout(() => {
+					battleRoom?.add(
+						`|html|<div class="broadcast-red"><b>This battle has been restored.</b><br />Use /rejoinbattle to rejoin.`
```
Zarel, line 1410: It should definitely be designed so they don't need to use `/rejoinbattle`.

I also think this should probably be part of the inputlog, which would remove the need for ten thousand timers.

server/rooms.ts
```diff
@@ -1764,3 +1805,5 @@ export const Rooms = {
 // initialize
 
 Rooms.global = new GlobalRoomState();
+// ensure battle loading happens after the rest of the server is intialized
+setTimeout(() => Rooms.global.loadBattleState(), 1 * 1000);
```
Zarel, line 1809: What sort of things here rely on the rest of the server needing to be initialized?

server/chat.ts
```diff
@@ -1062,7 +1061,7 @@ export class CommandContext extends MessageContext {
 				!user.can('bypassall') && (['help', 'lobby'].includes(room.roomid)) && (normalized === user.lastMessage) &&
 				((Date.now() - user.lastMessageTime) < MESSAGE_COOLDOWN)
 			) {
-				this.errorReply(this.tr("You can't send the same message again so soon."));
+				this.errorReply(this.tr("You can't send sthe same message again so soon."));
```
Zarel, line 1064: .

server/chat.ts
```diff
@@ -1049,7 +1049,6 @@ export class CommandContext extends MessageContext {
 			this.errorReply(this.tr("Your message contained banned words in this room."));
 			return null;
 		}
-
```
Zarel, line 1052: .

server/chat.ts
```diff
@@ -1081,7 +1080,6 @@ export class CommandContext extends MessageContext {
 		if (Chat.filters.length) {
 			return Chat.filter(this, message, user, room, connection, targetUser);
 		}
-
```
Zarel, line 1084: .

server/room-battle.ts
```diff
@@ -740,7 +744,6 @@ export class RoomBattle extends RoomGames.RoomGame {
 			if (!this.ended) this.timer.nextRequest();
 			this.checkActive();
 			break;
-
```
Zarel, line 743: .

server/room-battle.ts
```diff
@@ -1150,9 +1162,18 @@ export class RoomBattleStream extends BattleStream {
 			console.log(`[slow battle] ${deltaTime}ms - ${chunk}`);
 		}
 	}
-
```
Zarel, line 1153: .

server/room-battle.ts
```diff
@@ -1150,9 +1162,18 @@ export class RoomBattleStream extends BattleStream {
...
 	_writeLine(type: string, message: string) {
 		switch (type) {
+		case 'c':
+			this.battle.inputLog.push(`>chat ${message}`);
+			break;
+		case 'chat':
+			this.battle.inputLog.push(`>chat ${message}`);
+			this.battle.add('chat', `${message}`);
+			break;
```
Zarel, line 1173: Genius.

Zarel: I might call it `chat-inputlogonly` for clarity, though.

server/rooms.ts
```diff
@@ -1363,17 +1366,51 @@ export class GlobalRoomState {
...
+		const promises = [];
+		for (const [id, room] of Rooms.rooms) {
+			if (!room.battle) continue;
+			const formatid = room.battle.format;
+			if (!buffer[formatid]) buffer[formatid] = {};
+			const promise = room.battle.getLog().then((log) => {
+				buffer[formatid][id] = {
+					inputLog: log.map(item => item.replace(/\r/g, '')).join('\n'),
+					title: room.title,
+					num: room.roomid.split('-')[2],
```
Zarel, line 1395: Reminder that storing the `roomid` would simplify a lot of this code and doesn't seem to have disadvantages.

server/rooms.ts
```diff
@@ -1363,17 +1366,51 @@ export class GlobalRoomState {
...
+	async writeBattleState() {
+		const buffer: AnyObject = {};
+		const promises = [];
+		for (const [id, room] of Rooms.rooms) {
+			if (!room.battle) continue;
+			const formatid = room.battle.format;
+			if (!buffer[formatid]) buffer[formatid] = {};
+			const promise = room.battle.getLog().then((log) => {
+				buffer[formatid][id] = {
+					inputLog: log.map(item => item.replace(/\r/g, '')).join('\n'),
```
Zarel, line 1393: Why replace `\r`?

server/rooms.ts
```diff
@@ -1363,17 +1366,51 @@ export class GlobalRoomState {
...
+					title: room.title,
```
Zarel, line 1394: Hm, you'll need to extract the players from the roomlog, anyway. The title is probably superfluous, given this?

Zarel: How does `/importinputlog` work? Maybe we should work on that first.

mia-pi-git: Importinputlog requires you to re-add players in the new battle with /addplayer.
The title is also, likewise `Player 1 vs Player 2.` so storing the title, i felt, would be useful. 
I can, however, remove it.

server/rooms.ts
mia-pi-git: i used some of the sanitizing code /importinputlog did - this is one of those things.

Zarel: Sanitizing is for external input. This is internal data that has never gotten dirty in the first place; there's no need to sanitize it.

server/rooms.ts
Zarel: Hm, I think it would be useful to actually extract username and format from the inputlog.

TheImmortal:
If I'm understanding this correctly, you could technically have two different sets of mechanics in the same battle? Not sure if that is the best implementation.

mia-pi-git:
> If I'm understanding this correctly, you could technically have two different sets of mechanics in the same battle? Not sure if that is the best implementation.

What do you mean by this exactly? It reads all the data from the battle stream, including the seed, so everything should be handled the same. Or am I misunderstanding what you're saying?

Zarel:
I think TI is saying that the mechanics could have been updated, which is true. In practice, I don't consider this too big of a deal; mostly it just means that we need to make sure to hotpatch battles before restarting.

mia-pi-git:
Oh, i see. Should i try to figure out something around that....? 
I'm not sure how that would work.

Zarel:
Like I said, I don't think it's worth worrying about.

server/chat-commands/core.ts
```diff
@@ -1112,7 +1112,6 @@ export const commands: ChatCommands = {
 	restoreplayershelp: [
 		`/restoreplayers - Restore previous players in an imported input log.`,
 	],
-
 	joinbattle: 'joingame',
```
Zarel, line 1115:
```suggestion

	joinbattle: 'joingame',
```

server/rooms.ts
```diff
@@ -1216,7 +1216,6 @@ export class GlobalRoomState {
 		do {
 			roomid = `${roomPrefix}${++battleNum}` as RoomID;
 		} while (Rooms.rooms.has(roomid));
-
 		this.lastBattle = battleNum;
```
Zarel, line 1219:
```suggestion

		this.lastBattle = battleNum;
```

Zarel:
For the record, this PR is currently on hold: I'm not sure if the server can handle 2000 battle log queries simultaneously, every restart.

It _probably_ can, but I want to try it on a restart where we aren't also implementing a whole bunch of other brittle things.

config/config-example.js
```diff
@@ -368,6 +368,11 @@ exports.logchallenges = false;
  */
 exports.loguserstats = 1000 * 60 * 10; // 10 minutes
 
+/**
+ * persistbattles - Whether or not to persist battles through server restarts.
+*/
+exports.persistbattles = true;
```
MacChaeger, line 374: I don’t know if this should be true by default in the example file. I can’t speak for anyone else, but I know I would have disabled this had it existed when I started doing testing on a local server because I often restart or kill the server without actually ending the battle I was testing with.

config/config-example.js
```diff
@@ -368,6 +368,11 @@ exports.logchallenges = false;
  */
 exports.loguserstats = 1000 * 60 * 10; // 10 minutes
 
+/**
+ * persistbattles - Whether or not to persist battles through server restarts.
+*/
```
MacChaeger, line 373: Is it possible to make some kind of linter rule that catches these asterisks not lining up? Not in this PR, of course, but just in general.

server/rooms.ts
```diff
@@ -1376,17 +1376,49 @@ export class GlobalRoomState {
...
+				inputLog: log.join('\n'),
+				title: room.title,
+				roomid: room.roomid,
+			};
+		}
+		return FS(`logs/battles.json`).writeUpdate(() => JSON.stringify(buffer));
+	}
+	loadBattleState() {
+		if (!Config.persistbattles) return;
+		const battleData = JSON.parse(FS(`logs/battles.json`).readIfExistsSync() || "{}");
```
MacChaeger, line 1412: Could I ask you to cast this somehow so it’s not `any` or whatever generic type `JSON.parse` returns? That would make it a lot easier for other people to debug this code later.

server/rooms.ts
```diff
@@ -1709,6 +1755,8 @@ export const Rooms = {
 		let roomTitle;
 		if (gameType === 'multi') {
 			roomTitle = `Team ${p1name} vs. Team ${p2name}`;
+		} else if (options.title) {
```
MacChaeger, line 1758: This should either be first or last in this if-else chain, right? I’m thinking first.

MacChaeger (commented):
Some general concerns of mine:
- what happens if a restart, for better or worse, changes battle mechanics in a way that would have changed the outcome of a battle that was still active? I imagine it would desync and make a garbage battle state, but I don’t know what happens if you create a battle with a bad input log so it might just crash (or worse, hang forever on an awaited request that’s never coming).
- The longer the server’s been down, the more users will stop trying to reconnect before it restarts, leading to lots of timer wins/ties. I think if the server’s been offline for long enough the saved battles should expire, and maybe there should be some kind of timer forgiveness where timers are always off when a battle is reloaded and both players have a full minute or more to reconnect.

mia-pi-git:
Superseded by #8442. (new PR since that code is ultimately extremely different).

PR #7006, opened 2020-07-15
Youtube: support viewing trends in data

mia-pi-git:
This PR adds a few things. 
- Youtube JSON data is now automatically updated every 24 hours. 
- The important parts of the updated data (view count / sub count) are added into an array in the data.
- /youtube viewtrend allows users to view changes in the data.
- deprecates /youtube video, as /show is better.


server/chat-plugins/youtube.ts
```diff
@@ -23,9 +23,11 @@ try {
 export class YoutubeInterface {
 	interval: NodeJS.Timer | null;
 	intervalTime: number;
+	updateTimer?: NodeJS.Timer;
```
Zarel, line 26: Like I said, there's no need for a timer. :[

Zarel: Well, I suppose there is, if you want to store historical data.

But also, no, blah, why are we storing historical data? Memory doesn't grow on trees!

mia-pi-git: some historical data would be nice - maybe just havig it go back the last ~20 days would be nice?

mia-pi-git: Capped it at 15, is that a decent solution?

mia-pi-git:
@Zarel?

server/chat-plugins/youtube.ts
```diff
@@ -41,19 +43,36 @@ export class YoutubeInterface {
...
+				name: data.snippet.title,
+				description: data.snippet.description,
+				url: data.snippet.customUrl,
+				icon: data.snippet.thumbnails.medium.url,
+				videos: Number(data.statistics.videoCount),
+				subs: Number(data.statistics.subscriberCount),
+				views: Number(data.statistics.viewCount),
+				username: username,
+			};
+			this.data[id] = {...cache};
```
Zarel, line 57:
```suggestion
			this.data[id] = {
				name: data.snippet.title,
				description: data.snippet.description,
				url: data.snippet.customUrl,
				icon: data.snippet.thumbnails.medium.url,
				videos: Number(data.statistics.videoCount),
				subs: Number(data.statistics.subscriberCount),
				views: Number(data.statistics.viewCount),
				username: username,
			};
```

server/chat-plugins/youtube.ts
Zarel: I wouldn't mind historical data in an SQLite database, but in-memory storage should be limited. A monthish could be fine in theory but it depends on just how much data there is.

server/chat-plugins/youtube.ts
```diff
@@ -41,19 +43,36 @@ export class YoutubeInterface {
...
+			// add new data to the beginning
+			cached.subs.unshift(Number(data.statistics.subscriberCount));
+			cached.views.unshift(Number(data.statistics.viewCount));
+			cached.subs.slice(0, 15);
+			cached.views.slice(0, 15);
+			// update all the rest
+			for (const key in data) {
+				if (Array.isArray(cached[key])) continue;
+				cached[key] = data[key];
+			}
```
Zarel, line 71: These should be done one at a time, for safety.

server/chat-plugins/youtube.ts
```diff
@@ -24,10 +24,12 @@ export class YoutubeInterface {
 	interval: NodeJS.Timer | null;
 	intervalTime: number;
 	data: AnyObject;
```
Zarel, line 26: I'd probably want this to be strongly typed if you're going to do something like this.

server/chat-plugins/youtube.ts
Zarel: i.e. Don't use `for...in` on an AnyObject or otherwise bypass TypeScript's safety checks.

server/chat-plugins/youtube.ts
mia-pi-git: I'm gonna call it two weeks for now.

server/chat-plugins/youtube.ts
```diff
@@ -166,9 +185,29 @@ export class YoutubeInterface {
 	save() {
 		return FS(STORAGE_PATH).writeUpdate(() => JSON.stringify(this.data));
 	}
+	setUpdateTimer() {
+		if (this.updateTimer) clearTimeout(this.updateTimer);
+		const time = Date.now();
+		const nextMidnight = new Date(time + 24 * 60 * 60 * 1000);
+		nextMidnight.setHours(0, 0, 1);
+		this.updateTimer = setTimeout(() => this.updateData(), nextMidnight.getTime() - time);
+		return this.updateTimer;
```
Zarel, line 193: tbh the return is unnecessary, if you're already setting `this.updateTimer`. Just update the definition to `updateTimer!: NodeJS.Timer`.

mia-pi-git:
This should wait until i've migrated the plugin to SQLite, which i will do after #7277. In the meantime, i've split the PR into two; this one, and one that can be merged now - #7355.

mia-pi-git:
Closing this for now. Takes a bunch of space, ram usage is high, yada yada. I might look again at a later date.

PR #7008, opened 2020-07-15
Copycat and Encore should fail against a Z move

urkerab:
PR #6652 accidentally changed the logic for Copycat and Encore which previously only special-cased Max Moves (the previous condition was just `move.maxPowered`). In order to appease TypeScript I just check `move.baseMove` direcfly, however I still need to check `move.isMax` because of a hack SSB needs for Snaquaza's Z-move, and someone might try to Encore it, which would be bad.


Zarel:
Thanks!

PR #7010, opened 2020-07-16
Added another example of client implementation

MajeurAndroid:
Added Android-Unofficial-Showdown-Client as an example of ps client implementation. This might help people to understand Pokémon Showdown's protocol easily.


Zarel:
Sure, I can see this being useful.

PR #7011, opened 2020-07-16
Minor cleanup of global demotion messages

Lusamine:
Removes the extra parentheses in global demote message and adds a missing space in modlog.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -714,7 +714,7 @@ export class CommandContext extends MessageContext {
 
 	/** like privateModAction, but also notify Staff room */
 	privateGlobalModAction(msg: string) {
-		this.privateModAction(`${msg}`);
+		this.privateModAction(msg);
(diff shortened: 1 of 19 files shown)


Zarel:
Wait, no! The parens aren't supposed to be in the announcement in Staff. That's why it's added there. We should probably add parens inside `privateModAction` and remove them from all the calls to those functions.

server/chat.ts
```diff
@@ -714,7 +714,7 @@ export class CommandContext extends MessageContext {
 
 	/** like privateModAction, but also notify Staff room */
 	privateGlobalModAction(msg: string) {
-		this.privateModAction(`(${msg})`);
+		this.privateModAction(`${msg}`);
```
Zarel, line 717:
```suggestion
		this.privateModAction(msg);
```

Zarel:
Wow, thanks for doing all this! Just one more tiny change and that should be good.

PR #7012, opened 2020-07-17
Support throttling /requestshow

mia-pi-git:
This was requested by the staff of Arts & Culture, since apparently people often /requestshow too much.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2440,6 +2440,12 @@ export const commands: ChatCommands = {
 		if (!/^https?:\/\//.test(link)) link = `https://${link}`;
 		link = encodeURI(link);
 		if (!room.pendingApprovals) room.pendingApprovals = new Map();
+		if (!room.settings.showThrottle) room.settings.showThrottle = 0;
+		if (!room.lastShowRequest) room.lastShowRequest = {};
+		if (room.lastShowRequest[user.id] && Date.now() - room.lastShowRequest[user.id] <= room.settings.showThrottle) {
+			return this.errorReply(`Your request was not submitted because you've been sending them too quickly.`);
+		}
+		room.lastShowRequest[user.id] = Date.now();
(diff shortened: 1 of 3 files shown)


mia-pi-git:
Bad forcepush.

server/chat-commands/info.ts
```diff
@@ -2440,6 +2440,12 @@ export const commands: ChatCommands = {
 		if (!/^https?:\/\//.test(link)) link = `https://${link}`;
 		link = encodeURI(link);
 		if (!room.pendingApprovals) room.pendingApprovals = new Map();
+		if (!room.settings.showThrottle) room.settings.showThrottle = 0;
+		if (!room.lastShowRequest) room.lastShowRequest = {};
+		if (room.lastShowRequest[user.id] && Date.now() - room.lastShowRequest[user.id] <= room.settings.showThrottle) {
+			return this.errorReply(`Your request was not submitted because you've been sending them too quickly.`);
```
Zarel, line 2446: "You" is not really accurate. Is it supposed to be a global timer and not a per-user timer?

Zarel:
I'd like to speak to the staff directly.

server/chat-commands/info.ts
mia-pi-git: It's supposed to be per-user, yes.

Zarel: Your current implementation is a global timer, in case that was unclear.

mia-pi-git:
Discussed, not worth it. Closing.

Zarel:
Specifically: Most room staff mentioned they don't really need it. Some pointed out that there are a few situations where a user has multiple related images and should be able to post them all at once.

We also noted:

- `/mute` can already prevent people from using `/requestshow`, if they're spamming it
- You can only `/requestshow` again after your previous `/requestshow` is accepted/denied, so there's no risk of a single user creating a huge backlog or otherwise spamming multiple requests per second

Issue #6950, opened 2020-07-03
User privacy enhancements


scheibo:
### Add a `/nofeedback` command

> I've been thinking about some sort of command like
> /nofeedback /hidenext so it doesn't print a response
> hence /nofeedback as a command that runs the next command, but hides all sendReply responses

### Add sticky checkboxes to the Find Battle/Challenge UIs which will cause the battle to be hidden

`<abbr title="You can still invite spectators by giving them the URL or using the /invite command" class="checkbox">Ban spectators</abbr>`

> I would put it between the team selector and "Battle!" button & above challenge/cancel and below the team selector

The state of the Ban Spectators checkbox can be stored in local storage to ensure it is sticky.

### Support renaming rooms

> apparently the rename functionality just adds a broadcast linking to the new room
> but we do also have protocol infrastructure for actually renaming a room
> it's used by room aliases to redirect to the actual roomid

### Passwords

When a room is made invite only (either from the beginning or mid battle), kick out all guests and randomize the URL

- the new room id (and thus URL) generated by the sim server should follow the same pattern as replays which were made private: https://replay.pokemonshowdown.com/gen7randombattle-956127160-fz0wmd2n1wu7gorbay9jtvqxpc3tkl4pw. ie/ normal room id plus a random character string that ends in 'pw' (for 'password')
- if a replay is saved of the invite only room, the password generated by the sim server is to be used instead of the password generated by the replay server
- the password generation on the replay server is still needed in case a replay is made private after the fact, though the password generated may be different than the password generated by the sim server (though will follow the same format)

Zarel:
I did some edits to remove things that were a combination of my typos and some miscommunications.

I have no intention of removing the distinction between `/hideroom` and `/inviteonly`. Both should exist, but the UI should only surface `/hideroom` because the difference is too minor to have both.

AnnikaCodes:
> apparently the rename functionality just adds a broadcast linking to the new room

While there is a banner stating the room has been renamed, it's not a link -- it seems that all user connections are moved to the new room, as is the modlog. There is, however, a large bug in `/renameroom`: the chatroom's new name is never saved to `chatrooms.json`, and thus the rename partially reverts after restarting.

Zarel:
The client part of manual password specification is now done:

https://github.com/smogon/pokemon-showdown-client/commit/fdca8fe5dce755dd8390bc781dc5e5a3d9708cb5

[...]

PR #7015, opened 2020-07-18
Support renaming battles

AnnikaCodes:
This will pave the way for @scheibo's code in #6950 to be implemented.
If you want to actually _use_ `renameBattle()`, smogon/pokemon-showdown-client#1559 will need to be merged (otherwise the client is unable to display chat messages in a renamed battle until you reload it).

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -878,8 +878,16 @@ export class RoomBattle extends RoomGames.RoomGame {
 		const logfolder = logsubfolder.split('-', 2).join('-');
 		const tier = this.room.format.toLowerCase().replace(/[^a-z0-9]+/g, '');
 		const logpath = `logs/${logfolder}/${tier}/${logsubfolder}/`;
+
+		let logID = this.roomid;
+		if (logID.endsWith('pw')) {
+			const split = logID.split('-');
+			split.pop();
+			logID = split.join('-') as RoomID;
+		}
+
 		await FS(logpath).mkdirp();
-		await FS(logpath + this.room.roomid + '.log.json').write(JSON.stringify(logData));
+		await FS(`${logpath}${logID}.log.json`).write(JSON.stringify(logData));
 		// console.log(JSON.stringify(logData));
 	}
 	onConnect(user: User, connection: Connection | null = null) {
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -879,15 +879,8 @@ export class RoomBattle extends RoomGames.RoomGame {
 		const tier = this.room.format.toLowerCase().replace(/[^a-z0-9]+/g, '');
 		const logpath = `logs/${logfolder}/${tier}/${logsubfolder}/`;
 
-		let logID = this.roomid;
-		if (logID.endsWith('pw')) {
-			const split = logID.split('-');
-			split.pop();
-			logID = split.join('-') as RoomID;
-		}
-
 		await FS(logpath).mkdirp();
-		await FS(`${logpath}${logID}.log.json`).write(JSON.stringify(logData));
+		await FS(`${logpath}${this.room.getReplayData().id}.log.json`).write(JSON.stringify(logData));
 		// console.log(JSON.stringify(logData));
 	}
 	onConnect(user: User, connection: Connection | null = null) {
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -684,7 +684,7 @@ export abstract class BasicRoom {
 		if (oldID === 'lobby') {
 			Rooms.lobby = null;
 		} else if (newID === 'lobby') {
-			Rooms.lobby = this as ChatRoom;
+			Rooms.lobby = this;
 		}
 
 		if (keepAliases) {
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -668,44 +668,44 @@ export abstract class BasicRoom {
 
 	/**
 	 * @param newID Add this param if the roomid is different from `toID(newTitle)`
+	 * @param moveLogs Whether or not to rename the log file. Defaults to true.
+	 * @param keepAliases Whether or not to point the room's aliases and name to its new name. Defaults to true.
 	 */
-	async rename(newTitle: string, newID?: RoomID) {
+	async rename(newTitle: string, newID?: RoomID, moveLogs = true, keepAliases = true) {
 		if (!newID) newID = toID(newTitle) as RoomID;
-		if (this.game || this.tour) return;
+		if ((this.game && moveLogs) || this.tour) return;
 
 		const oldID = this.roomid;
 		this.roomid = newID;
 		this.title = newTitle;
 		Rooms.rooms.delete(oldID);
-		Rooms.rooms.set(newID, this as ChatRoom);
+		Rooms.rooms.set(newID, (this.game ? this as unknown as GameRoom : this as ChatRoom));
 
 		if (oldID === 'lobby') {
 			Rooms.lobby = null;
 		} else if (newID === 'lobby') {
-			Rooms.lobby = this as ChatRoom;
+			Rooms.lobby = this as ChatRoom; // lobby shouldn't be a GameRoom
 		}
 
-		for (const [alias, roomid] of Rooms.aliases.entries()) {
-			if (roomid === oldID) {
-				Rooms.aliases.set(alias, newID);
+		if (keepAliases) {
+			for (const [alias, roomid] of Rooms.aliases.entries()) {
+				if (roomid === oldID) {
+					Rooms.aliases.set(alias, newID);
+				}
+			}
+
+			// add an alias from the old id
+			Rooms.aliases.set(oldID, newID);
+			if (this.settings) {
+				if (!this.settings.aliases) this.settings.aliases = [];
+				// resolve an old (fixed) bug in /renameroom
+				if (!this.settings.aliases.includes(oldID)) this.settings.aliases.push(oldID);
+				this.saveSettings();
 			}
 		}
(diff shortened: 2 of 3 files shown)


server/rooms.ts
```diff
@@ -1559,6 +1552,24 @@ export class GameRoom extends BasicRoom {
...
+		Rooms.rooms.delete(oldID);
+		Rooms.rooms.set(newID, this as GameRoom);
+
+		for (const user of Object.values(this.users)) {
+			user.moveConnections(oldID, newID);
+			user.send(`>${oldID}\n|noinit|rename|${newID}|${this.title}`);
+		}
+
+		return true;
+	}
```
Zarel, line 1572: It worries me that this is a different function from the one for renaming rooms. That seems wrong to me.

Zarel: Is there a reason why this functionality can't be part of `rename`?

AnnikaCodes: it was discussed on discord in #sim-dev and iirc the consensus was that rensning without moving logs was different enough to warrant its own function.

Zarel: I disagree.

Even if there wasn't as much shared code as there is, it should at least be a function called by `rename`, rather than an entirely separate function.

But because there's so much shared code, I think it'd be better for it to be one function.

Zarel:
I think having aliases is useful, but it obviously defeats the purpose in this particular case of the rename being used to grant privacy.

I would probably still use an alias for the private-to-public rename, though. I don't think it's likely to break much.

server/rooms.ts
AnnikaCodes: Alright, fair.

server/rooms.ts
```diff
@@ -668,44 +668,42 @@ export abstract class BasicRoom {
...
 		if (!newID) newID = toID(newTitle) as RoomID;
-		if (this.game || this.tour) return;
+		if ((this.game && moveLogs) || this.tour) return;
 
 		const oldID = this.roomid;
 		this.roomid = newID;
 		this.title = newTitle;
 		Rooms.rooms.delete(oldID);
-		Rooms.rooms.set(newID, this as ChatRoom);
+		Rooms.rooms.set(newID, (this.game ? this as unknown as GameRoom : this as ChatRoom));
```
Zarel, line 682:
```suggestion
		Rooms.rooms.set(newID, this as Room);
```

(You can double-cast `this as any as Room` if it yells at you, but it shouldn't.)

server/rooms.ts
```diff
@@ -668,44 +668,42 @@ export abstract class BasicRoom {
...
 	/**
 	 * @param newID Add this param if the roomid is different from `toID(newTitle)`
+	 * @param moveLogs Whether or not to rename the log file. Defaults to true.
+	 * @param keepAliases Whether or not to point the room's aliases and name to its new name. Defaults to true.
 	 */
-	async rename(newTitle: string, newID?: RoomID) {
+	async rename(newTitle: string, newID?: RoomID, moveLogs = true, keepAliases = true) {
 		if (!newID) newID = toID(newTitle) as RoomID;
-		if (this.game || this.tour) return;
+		if ((this.game && moveLogs) || this.tour) return;
```
Zarel, line 676: n.b. ChatRooms can also have games.

Possibly `(this.type === 'chat' && this.game)`.

Also possibly `throw Chat.ErrorMessage` instead of just returning.

AnnikaCodes: The intent of the `moveLogs` check is to disallow renaming the logs of battles, since that would disrupt usage stats (or so I'm told), so that needs to stay, but I should check for being a chatroom with an ongoing game.

server/rooms.ts
```diff
@@ -668,44 +668,42 @@ export abstract class BasicRoom {
...
 		this.title = newTitle;
 		Rooms.rooms.delete(oldID);
-		Rooms.rooms.set(newID, this as ChatRoom);
+		Rooms.rooms.set(newID, this as Room);
 
 		if (oldID === 'lobby') {
 			Rooms.lobby = null;
 		} else if (newID === 'lobby') {
-			Rooms.lobby = this as ChatRoom;
+			Rooms.lobby = this as ChatRoom; // lobby shouldn't be a GameRoom
```
Zarel, line 687: I don't think this needs a comment?

```suggestion
			Rooms.lobby = this as ChatRoom;
```

A common rule of thumb is that a comment should say more than what the code already says. Like this is just repeating what `as ChatRoom` means.

Zarel: Heck, forget the cast, this might benefit from throwing if you try to rename a GameRoom to `lobby`.

AnnikaCodes: That's a good idea.

server/rooms.ts
Zarel: That check could probably happen inside the logs, and it probably needs to, because battles that start hidden also shouldn't have the password in their logs.

server/rooms.ts
```diff
@@ -668,44 +668,42 @@ export abstract class BasicRoom {
...
 	/**
 	 * @param newID Add this param if the roomid is different from `toID(newTitle)`
+	 * @param moveLogs Whether or not to rename the log file. Defaults to true.
+	 * @param keepAliases Whether or not to point the room's aliases and name to its new name. Defaults to true.
 	 */
-	async rename(newTitle: string, newID?: RoomID) {
+	async rename(newTitle: string, newID?: RoomID, moveLogs = true, keepAliases = true) {
 		if (!newID) newID = toID(newTitle) as RoomID;
-		if (this.game || this.tour) return;
+		if ((this.type === 'battle' && moveLogs) || (this.type === 'chat' && this.game) || this.tour) return;
```
Zarel: `Chat.ErrorMessage`, right?

```suggestion
		if (this.type === 'chat' && this.game) {
			throw new Chat.ErrorMessage(`Please finish your game (${this.game.title}) before renaming ${this.roomid}.`);
		}
		if (this.type === 'battle' && moveLogs) {
			idk, think of a suitable error message
		}
```

Zarel:
(In other words, `moveLogs` shouldn't be a variable, all that complication probably needs to happen inside `roomlogs.ts`)

server/rooms.ts
scheibo: Where is 'inside the logs'?

Zarel: Inside `server/roomlogs.ts`.

Zarel: Or whatever code handles log writing.

AnnikaCodes: Battle logs (which is what we're talking about) seem to be in `server/room-battle.ts`.

server/room-battle.ts
```diff
@@ -878,8 +878,16 @@ export class RoomBattle extends RoomGames.RoomGame {
 		const logfolder = logsubfolder.split('-', 2).join('-');
 		const tier = this.room.format.toLowerCase().replace(/[^a-z0-9]+/g, '');
 		const logpath = `logs/${logfolder}/${tier}/${logsubfolder}/`;
+
+		let logID = this.roomid;
+		if (logID.endsWith('pw')) {
```
scheibo, line 883: I think you might want to use the `getReplayData` function from https://github.com/smogon/pokemon-showdown/issues/6950#issuecomment-653822792? In general Zarel and I discussed wanting a single method for parsing the pw-ed ids

server/rooms.ts
Zarel: I wonder if those should be moved... either way, it doesn't seem to need to be handled in `room.rename`. The `moveLogs` parameter appears superfluous.

server/room-battle.ts
Zarel: Agree with pre on this one.

Zarel:
Looks good now. Requires a restart but doesn't break hotpatch, so it can be merged immediately. Thanks!

server/rooms.ts
```diff
@@ -666,46 +666,52 @@ export abstract class BasicRoom {
...
+		return {id: this.roomid.slice(7, lastHyphen), password: this.roomid.slice(lastHyphen, end)};
+	}
+
 	/**
 	 * @param newID Add this param if the roomid is different from `toID(newTitle)`
+	 * @param moveLogs Whether or not to rename the log file. Defaults to true.
+	 * @param keepAliases Whether or not to point the room's aliases and name to its new name. Defaults to true.
 	 */
-	async rename(newTitle: string, newID?: RoomID) {
+	async rename(newTitle: string, newID?: RoomID, keepAliases = true) {
```
Zarel, line 681: Although I do prefer parameters that default to `false`. I'd invert this parameter and call it `noAlias`.

server/rooms.ts
```diff
@@ -666,46 +666,51 @@ export abstract class BasicRoom {
...
+				}
 			}
+
+			// add an alias from the old id
+			Rooms.aliases.set(oldID, newID);
+			if (!this.settings.aliases) this.settings.aliases = [];
+			// resolve an old (fixed) bug in /renameroom
+			if (!this.settings.aliases.includes(oldID)) this.settings.aliases.push(oldID);
+			this.saveSettings();
 		}
```
Zarel, line 710:
```suggestion
		} else {
			// clear aliases
			for (const [alias, roomid] of Rooms.aliases.entries()) {
				if (roomid === oldID) {
					Rooms.aliases.delete(alias);
				}
			}
			this.settings.aliases = undefined;
		}
		this.saveSettings();
```

Zarel: There are definitely situations this could come up – taking a battle public and then private again, for instance. But even if it didn't, this is just a generally good thing to do – you don't know how the function might be used in the future, and it's best to play it safe.

PR #7016, opened 2020-07-18
Cross Evolution: Simplify validation

KrisXV:
All NFE Pokemon are going to be banned instead of restricted from now on, so the second check is useless.


Zarel:
Thanks!

Do you have a link to the announcement?

KrisXV:
https://www.smogon.com/forums/threads/cross-evolution-om-of-the-month.3657562/page-5#post-8535338

Zarel:
Thanks!

PR #7018, opened 2020-07-18
Support Chat.ErrorMessage in async functions

mia-pi-git:
Currently, it just crashes as a normal crash might, even though that's not the intent.
edit: will draft for now, i noticed an error that i'll fix asap, missed it because I tested this wrong.


Zarel:
I thought I already did this?

Zarel:
Did I forget to push?

PR #7024, opened 2020-07-20
Add Armor Beginnings Online Competition

LegoFigure11:
Adds Armor Beginnings: https://www.smogon.com/forums/threads/3667440/


cantsay:
this format is annoyingly time-sensitive. It was only just announced in-game and the actual competition is in 4 days, so people are clamouring for practice

PR #7025, opened 2020-07-20
Damp Negates Misty Explosion

PartMan7:
`Damp prevents all Pokémon from using Self-Destruct, Explosion, Mind Blown, and Misty Explosion;` - Bulbapedia
SmogDex says so, too.


Zarel:
@Marty-D?

Marty-D:
Thanks, don't forget to remove mentions of things that don't exist in past gens next time! (The implementation is fine to inherit; descriptions are not)

PartMan7:
Right, sorry.

PR #7031, opened 2020-07-22
Fix bug in destroyPersonalRooms()

AnnikaCodes:
This will actually destroy groupchats when a user is locked/globally banned, which to my understanding is the intended behavior.
`continue`ing when a room isn't persistent makes no sense, since groupchats are by definition not persistent.


Zarel:
Good catch!

PR #7032, opened 2020-07-22
Add cooldown for unregistered namechanges

AnnikaCodes:
If it would be more useful with a different cooldown length (I just grabbed 2 minutes off the top of my head) or enforcing a cooldown on namechanges to registered alts as well, I'm happy to change it.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -592,11 +592,8 @@ export abstract class BasicRoom {
 		this.destroy();
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
-		let reportJoins = this.reportJoins;
-		if (reportJoins && this.settings.modchat && !user.authAtLeast(this.settings.modchat, this)) {
-			reportJoins = false;
-		}
-		if (reportJoins) {
+		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
+		if (this.reportJoins && canTalk) {
 			this.add(`|${type}|${entry}`).update();
 			return;
 		}
--- a/server/users.ts
+++ b/server/users.ts
@@ -37,6 +37,7 @@ const THROTTLE_MULTILINE_WARN = 3;
 const THROTTLE_MULTILINE_WARN_STAFF = 6;
 
 const NAMECHANGE_THROTTLE = 2 * 60 * 1000; // 2 minutes
+const NAMES_PER_THROTTLE = 3;
 
 const PERMALOCK_CACHE_TIME = 30 * 24 * 60 * 60 * 1000; // 30 days
 
@@ -370,6 +371,7 @@ export class User extends Chat.MessageContext {
 	lastMessageTime: number;
 	lastReportTime: number;
 	lastNewNameTime = 0;
+	newNames = 0;
 	s1: string;
 	s2: string;
 	s3: string;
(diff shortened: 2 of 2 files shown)


Zarel:
If it's not done yet, the first step is to make semilocked/locked user namechanges not show up in the log at all.

Zarel:
Probably also modchatted and muted users, tool

Zarel:
I don't know how I feel about a throttle. Maybe just combine successive renames, like we do joins/leaves. Or not show renames at all, for users who haven't talked recently?

scheibo:
Agree theres probably a couple things we want here to cut down on spam. Combining alone still seems like it may be annoying? Unless you collapse all the renames into 1? Agree to not showing renames at all for people not talking 'recently' (whats the constant there?), but some sort of throttling mechanism beyond that still seems desirable - these are done by bots who could pretty easily break collapsing i think by just also interspersing text in between their name changes. Throttling is heavy handed, but what we're trying to target here is malicious spam bots

Zarel:
> Unless you collapse all the renames into 1?

Yes, that's what I meant.

> Throttling is heavy handed, but what we're trying to target here is malicious spam bots

A 1 per 2 minute throttle just feels like it could affect some legitimate use-cases (like actual typoes), which is why I'm trying to think of alternate solutions. If it was like 3 per 2 minutes, I'd probably feel better.

AnnikaCodes:
> locked



> If it's not done yet, I think an important first step is to make semilocked/locked user namechanges not show up in the log at all.

This is currently implemented, but only when modchat is set — I added a fix so that it works regardless.

> Yes, that's what I meant. If there's no chat between "A renamed to B" and "B renamed to C", collapse that into "A renamed to C".

This is a client-side problem — see https://github.com/smogon/pokemon-showdown-client/pull/1565.

> A 1 per 2 minute throttle just feels like it could affect some legitimate use-cases (like actual typoes), which is why I'm trying to think of alternate solutions. If it was like 3 per 2 minutes, I'd probably feel better.

Sure.

server/rooms.ts
```diff
@@ -592,11 +592,8 @@ export abstract class BasicRoom {
 		this.destroy();
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
-		let reportJoins = this.reportJoins;
-		if (reportJoins && this.settings.modchat && !user.authAtLeast(this.settings.modchat, this)) {
-			reportJoins = false;
-		}
-		if (reportJoins) {
+		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
+		if (this.reportJoins && canTalk) {
```
Zarel, line 596: Oh, this code is a lot better.

Does this cover semilocks?

Zarel: Looks like no, can you add that condition? Maybe with an exception for players (i.e. any roomauth)

server/users.ts
```diff
@@ -709,6 +715,23 @@ export class User extends Chat.MessageContext {
 		this.s2 = tokenDataSplit[6];
 		this.s3 = tokenDataSplit[7];
 
+		if (!this.trusted && userType === '1') { // userType '1' means unregistered
+			const elapsed = Date.now() - this.lastNewNameTime;
+			if (elapsed < NAMECHANGE_THROTTLE) {
+				if (this.newNames >= NAMES_PER_THROTTLE) {
+					this.send(
+						`|nametaken|${name}|You must wait ${Math.round((NAMECHANGE_THROTTLE - elapsed) / 1000)} more
```
mia-pi-git, line 723: ``
```suggestion
						`|nametaken|${name}|You must wait ${Chat.toDurationString(NAMECHANGE_THROTTLE - elapsed)} more
```

server/rooms.ts
AnnikaCodes:
```ts
authAtLeast(minAuth: string, room: BasicRoom | null = null) {
		if (this.locked || this.semilocked) return false;
...
```
`User#authAtLeast` will return false if a user is semilocked. I can still add an exception for players / voices in the battle, if you would like.

server/users.ts
Zarel: Oh, good suggestion!

server/rooms.ts
Zarel: Ohhh, yeah, that works.

I do think semilocked players should be able to talk in their own battles, yes.

Zarel: They wouldn't be able to rename spam anyway, because after the first rename they're no longer players.

Zarel:
A good improvement as-is, in any case.

PR #7033, opened 2020-07-22
Remove some unnecessary references to the global room

whalemer:


server/users.ts
```diff
@@ -1232,11 +1232,6 @@ export class User extends Chat.MessageContext {
 		force = false
```
Zarel, line 1232: Can we also remove this parameter?

Zarel:
Nice, thanks!

server/users.ts
whalemer: oh yeah, looks like it

half surprised the linter isn't complaining, honestly.

Zarel: The linter allows unused parameters by default because it's sometimes necessary for compatibility reasons. There's probably a setting to change it.

PR #7034, opened 2020-07-22
Random Doubles updates/upkeep

xInstruct:


TheImmortal:
Removed some unnecessary stuff. They should already get the appropriate ability.

PR #7036, opened 2020-07-23
Put item/move/ability descriptions in data/text/

Zarel:
This adds some new `data/text/` files, which are the new home for item/move/ability descriptions, as well as in-game messages, in one place for ease of translation.


data/mods/letsgo/moves.ts
```diff
@@ -83,8 +83,6 @@ export const BattleMovedex: {[k: string]: ModdedMoveData} = {
 	},
 	teleport: {
 		inherit: true,
-		desc: "If this move is successful and the user has not fainted, the user switches out even if it is trapped and is replaced immediately by a selected party member. The user does not switch out if there are no unfainted party members.",
-		shortDesc: "User switches out.",
```
KrisXV, line 87: You seem to be removing these and then not doing anything with them later.

sim/dex.ts
```diff
@@ -471,6 +479,35 @@ export class ModdedDex {
 		return learnsetData;
 	}
 
+	getDescs(table: keyof TextTableData, id: ID, dataEntry: AnyObject) {
```
scheibo, line 482: These APIs should really be `async`

sim/dex.ts
```diff
@@ -102,6 +102,12 @@ interface DexTableData {
 	Statuses: DexTable<EffectData>;
 	TypeChart: DexTable<TypeData>;
 }
+interface TextTableData {
+	Abilities: AnyObject;
```
scheibo, line 106: I think you can give these a better type than AnyObject?

data/text/abilities.ts
```diff
@@ -0,0 +1,1480 @@
...
+		name: "Adaptability",
+		desc: "This Pokemon's moves that match one of its types have a same-type attack bonus (STAB) of 2 instead of 1.5.",
+		shortDesc: "This Pokemon's same-type attack bonus (STAB) is 2 instead of 1.5.",
+	},
+	aerilate: {
+		name: "Aerilate",
+		desc: "This Pokemon's Normal-type moves become Flying-type moves and have their power multiplied by 1.2. This effect comes after other effects that change a move's type, but before Ion Deluge and Electrify's effects.",
+		shortDesc: "This Pokemon's Normal-type moves become Flying type and have 1.2x power.",
+		descGen6: "This Pokemon's Normal-type moves become Flying-type moves and have their power multiplied by 1.3. This effect comes after other effects that change a move's type, but before Ion Deluge and Electrify's effects.",
+		shortDescGen6: "This Pokemon's Normal-type moves become Flying type and have 1.3x power.",
```
scheibo, line 16: How does the `fieldN` approach scale to other mods (like Stadium or Lets Go even)?

data/text/abilities.ts
```diff
@@ -0,0 +1,1480 @@
+export const AbilitiesText = {
```
scheibo, line 1: Should these files be `data/text/en/abilities.ts`? Or i suppose like with `data/`, only modifications to the default live in subdirs?

data/mods/letsgo/moves.ts
Zarel: Yeah, these should be added back in.

sim/dex.ts
Zarel: It's temporary. monkfish didn't want to change the API yet and this turned out to be reasonably straightforward.

sim/dex.ts
Zarel: I could but laziness.

data/text/abilities.ts
Zarel: It doesn't.

data/text/abilities.ts
Zarel: Yeah, that's probably how I'd end up doing it.

data/text/abilities.ts
Zarel: You could do e.g. `descGen1Stadium`, `descGen7LetsGo`.

Zarel: Underscore followed by formatid is certainly an option, though. It would break existing naming conventions but it would be easy to parse and be compatible with other formats.

scheibo:
```ts
type Text = {
  name?: string;
  desc?: string;
  shortDesc?: string;
};
type Texts = Text & {[mod]: Text}
```

I dont think nesting is prohibitive

Zarel: Do you want to write a conversion script and/or find someone to manually do the conversion?

Zarel:
I would probably rather merge now and deal with tweaks later, in case of conflicts. Sorry @KrisXV for the conflicts with your Gmax PR, btw.

data/text/abilities.ts
scheibo: Did you not write this with a script already? I would have imagined it could just be edited? If not, its not hard, I can write one. Appending `6` or `Gen6` or `_Gen6` is just much harder to make typesafe:

```js
const after = {};
for (let key in before) {
  const b = before[key];
  const a = after[key] = {};
  for (let k in b) {
    if (['name', 'desc', 'shortDesc'].includes(k)) {
      a[k] = b[k];    
    } else {
     const g = // TODO get last char
     k = // remove last characer - k.slice(0, -1) or something?
     const mod = `gen${g}`;
     a[mod] = a[mod] || {};
     a[mod][k] = b[k];
    }
  }
} 
```

Zarel: That's a value converter, not a file converter; it would wreck the comments...

Zarel: (I wrote a script to fill the current-gen data, and then had Flare manually fill in the past-gen data.)

Marty-D:
Obligatory bump

Zarel:
Since this was merged, I should note that the formatting isn't final. This just makes it easier to transition to the final version later, without having to worry about merge conflicts.

(@Marty-D feel free to deploy your text changes at any time!)

Marty-D:
Thanks! This will make multi-gen edits much less painful.

sim/dex.ts
```diff
@@ -472,6 +480,35 @@ export class ModdedDex {
 		return learnsetData;
 	}
 
+	getDescs(table: keyof TextTableData, id: ID, dataEntry: AnyObject) {
+		if (dataEntry.shortDesc) {
```
urkerab, line 484: Why just `shortDesc`? It's confusing particularly for items since many of them only have a `desc`, so many OMs with custom items (e.g. SSB) only specify an updated `desc`. (Or should `shortDesc` be always present, and `desc` only for things that need it?)

Zarel: I've never actually standardized what it means for only one of `desc` or `shortDesc` to exist. I think most of our accesses backfill in both directions. If you'd like to standardize, please do, and document that in our TS types.

sim/dex.ts
```diff
@@ -472,6 +480,35 @@ export class ModdedDex {
...
+			return {
+				desc: dataEntry.desc,
+				shortDesc: dataEntry.shortDesc,
+			};
+		}
+		const entry = this.loadTextData()[table][id];
+		if (!entry) return null;
+		const descs = {
+			desc: '',
+			shortDesc: '',
```
Zarel, line 494: Replacing these with the `dataEntry`'s current values should fix this, right, @urkerab?

sim/dex.ts
urkerab: Actually what I was actually concerned about is about how mods override descriptions. Hopefully none of them want to override just one of `desc` and `shortDesc`, so it should be enough to change this to test both.

urkerab: Ugh, it took me 20 minutes to compose that and I still made a fool of myself...

sim/dex.ts
urkerab: That depends on what you want the inheritance behaviour to be... given that you added an early return for mod-supplied descriptions, I wouldn't like to guess.

Zarel: As mentioned earlier, I have no clue what the inheritance behavior should be and I'm open to suggestions. But I think we should probably at least want to support overriding `desc` while inheriting `shortDesc`.

PR #7039, opened 2020-07-23
Remove "Battle" prefix from data exports

Zarel:
`BattlePokedex` is now `Pokedex`, `BattleItems` is now `Items`, etc.

I also renamed `Movedex` to `Moves`.

`TypeChart` isn't `Types` yet, because unlike the others, it's not indexed by ID. That should probably be fixed one day.

### DIFF
--- a/data/mods/gen1/scripts.ts
+++ b/data/mods/gen1/scripts.ts
@@ -4,7 +4,7 @@
  * This generation inherits all the changes from older generations, that must be taken into account when editing code.
  */
 
-export const BattleScripts: ModdedBattleScriptsData = {
+export const Scripts: ModdedBattleScriptsData = {
 	inherit: 'gen2',
 	gen: 1,
 	init() {
@@ -865,4 +865,4 @@ export const BattleScripts: ModdedBattleScriptsData = {
 	},
 };
 
-exports.BattleScripts = BattleScripts;
+exports.Scripts = Scripts;
--- a/data/mods/gen1/scripts.ts
+++ b/data/mods/gen1/scripts.ts
@@ -864,5 +864,3 @@ export const Scripts: ModdedBattleScriptsData = {
 		return Math.floor(damage);
 	},
 };
-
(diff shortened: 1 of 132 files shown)


Zarel:
This will also require a client change, which I'll just push directly once this is merged.

Zarel:
For some reason, I had weird test errors until I did `./build --force`. I'm still not clear on why that was necessary, though.

Zarel:
> For some reason, I had weird test errors until I did `./build --force`. I'm still not clear on why that was necessary, though.

The error was `Error: Cannot find module '../sim/dex'`

Which presumably means that sucrase ran, but our text-replacement didn't, which should be impossible given the architecture of our build script...

data/mods/gen1/statuses.ts
```diff
@@ -8,7 +8,7 @@
  * under certain conditions and re-applied under other conditions.
  */
 
-export const BattleStatuses: {[k: string]: ModdedPureEffectData} = {
+export const Statuses: {[k: string]: ModdedPureEffectData} = {
```
scheibo, line 11: Consider `Conditions`/`conditions.ts`

data/mods/gen1/scripts.ts
```diff
@@ -865,4 +865,4 @@ export const BattleScripts: ModdedBattleScriptsData = {
 	},
 };
 
-exports.BattleScripts = BattleScripts;
+exports.Scripts = Scripts;
```
KrisXV, line 868: is this line necessary?

data/mods/gen1/statuses.ts
Zarel: Good idea.

data/mods/gen1/scripts.ts
Zarel: No.

Zarel:
Client updated, everything still works, whew.

PR #7040, opened 2020-07-23
Support hiding battles from /cmd userdetails

AnnikaCodes:
I've heard that some global staff want this feature as a way to prevent people from joining battles by clicking their usercard, without full-on hiding all battles they play in.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1321,15 +1321,17 @@ export const commands: ChatCommands = {
 		});
 	},
 
-	nodisplaybattles: 'displaybattles',
-	displaybattles(target, room, user, connection, cmd) {
-		const shouldHide = cmd.includes('no');
-		user.settings.hideBattles = shouldHide;
-		this.sendReply(`Battles are now ${shouldHide ? "hidden (except to staff)" : "visible"} in your usercard.`);
+	showbattlesinusercard: 'hidebattlesfromtrainercard',
+	hidebattlesfromusercard: 'hidebattlesfromtrainercard',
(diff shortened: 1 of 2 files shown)


Zarel:
:/ We have to be pretty careful with the name for this function so it doesn't get confused with other similar functionality.

AnnikaCodes:
> :/ We have to be pretty careful with the name for this function so it doesn't get confused with other similar functionality.

Yeah, /nodisplaybattles was the best one I could think of, but I'm open to suggestions.

Zarel:
Generally I default to "really long name" if it's something that can easily be misunderstood.

Something like `/hidebattlesfromtrainercard`

Zarel:
Probably alias `/hidebattlesfromusercard`

Zarel:
I would also call the variable `settings.hideBattlesFromTrainerCard`.

Zarel:
I like this! @scheibo do you have opinions? Otherwise I'm merging.

server/chat-commands/core.ts
```diff
@@ -1321,6 +1321,19 @@ export const commands: ChatCommands = {
 		});
 	},
 
+	showbattlesinusercard: 'hidebattlesfromtrainercard',
+	hidebattlesfromusercard: 'hidebattlesfromtrainercard',
+	showbattlesintrainercard: 'hidebattlesfromtrainercard',
+	hidebattlesfromtrainercard(target, room, user, connection, cmd) {
+		const shouldHide = cmd.includes('hide');
+		user.settings.hideBattlesFromTrainerCard = shouldHide;
```
scheibo, line 1329: You need to call user.update() or whatever to propagate the setting via |updateuser| to the client

PR #7041, opened 2020-07-24
Send group for punished users in /cmd userdetails

AnnikaCodes:
The server-side portion of https://github.com/smogon/pokemon-showdown-client/pull/1568. This might be better restricted to trusted users (since they can use `/checkpunishment`) — personally, I think it's fine since a locked user in a room will have their status publicly visible, but there is an argument otherwise.


Zarel:
Looks god!

PR #7043, opened 2020-07-24
Don't override a higher rank in personal rooms

AnnikaCodes:
Currently if you set a lower rank's `globalGroupInPersonalRoom` attribute, the higher ranks will inherit that, potentially causing them to have permission issues in personal rooms.

### DIFF
--- a/server/user-groups.ts
+++ b/server/user-groups.ts
@@ -182,7 +182,7 @@ export class RoomAuth extends Auth {
 		const symbol = super.getEffectiveSymbol(user);
 		if (!this.room.persist && symbol === user.group) {
 			const replaceGroup = Auth.getGroup(symbol).globalGroupInPersonalRoom;
-			if (replaceGroup) return replaceGroup;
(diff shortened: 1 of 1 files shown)


Zarel:
I think this should probably be the part we consider a bug:

\>\> `Users.Auth.getGroup('#').globalGroupInPersonalRoom`  
<< `"@"`

I would rather not keep hardcoding symptoms rather than address this bug.

AnnikaCodes:
> I think this should probably be the part we consider a bug:
> 
> >> `Users.Auth.getGroup('#').globalGroupInPersonalRoom`
> << `"@"`
> 
> I would rather not keep hardcoding symptoms rather than address this bug.

The issue here is that `globalGroupInPersonalRoom` is inherited from lower groups. The best solution I see is to have a way to mark some permissions as non-inheritable, probably with another Config setting or a const in config-loader.

Zarel:
To be clear, I do really appreciate you looking into permissions bugs. I just don't like hardcoding symptom-fixes because I can relatively quickly fix the underlying bug if it's causing trouble.

PR #7044, opened 2020-07-24
Make sendhtmlpage perms more strict, refactor perm checks

whalemer:


server/chat.ts
```diff
@@ -1107,6 +1107,29 @@ export class CommandContext extends MessageContext {
 
 		return message;
 	}
+	canPMHTML(targetUser: User | null) {
+		if (!targetUser || !targetUser.connected) {
+			this.errorReply(`User ${this.targetUsername} is not currently online.`);
```
whalemer, line 1112: (editors note: this is written like this so it's theoretically easier to refactor all the perm checks to use `ErrorMessage`s)

server/chat.ts
```diff
@@ -1107,6 +1107,29 @@ export class CommandContext extends MessageContext {
...
+			return false;
+		}
+		if (!(this.room && (targetUser.id in this.room.users)) && !this.user.can('addhtml')) {
+			this.errorReply("You do not have permission to use this command to users who are not in this room.");
```
Zarel, line 1116: The problem with this is: what if a user tries to join `view-bot-scrappie-test` without sharing any rooms with Scrappie?

Zarel:
I really like unifying the permissions checks, though.

server/chat.ts
whalemer: i mean, the obvious solution is just to stop you from using that unless the bot can PM you back.

alternately, i could see some kind of "which connection last requested which page" property being nice to have (for this and for #7051

Zarel: Yeah, I think that might be the best way to d oit.

Zarel: This part doesn't seem fixed?

whalemer: look at the sendhtmlpage function, this only gets called if there's no relevant connection to send the page to.

Zarel: Oh!

Zarel:
This still has the (admittedly rarer) race condition if two pages are requested at once. Might I trouble you to make `lastRequestedPage` an Array? (at n<10 or so Array is faster than Set)

Zarel:
Actually, since this is already much better than before, I'll just accept it now. But if you could send another PR for the race condition, that'd be great.

PR #7045, opened 2020-07-24
Add a throttle for pending challenges

mia-pi-git:
This is to resolve an issue where a user would be spammed by challenges from unregged bot accounts until their browser crashed. 
This PR:
- prevents unnamed users from challenging 
- Prevents a user from being challenged by a non-autoconfirmed user if they already have 3 challenges pending.
In short, should stop users from spambot-challenging.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1237,6 +1237,9 @@ export const commands: ChatCommands = {
 		if (Punishments.isBattleBanned(user)) {
 			return this.popupReply(`You are banned from battling and cannot challenge users.`);
 		}
+		if (!user.named || user.namelocked) {
+			return this.popupReply(`Unnamed users cannot challenge.`);
+		}
 		if (Config.pmmodchat) {
 			const userGroup = user.group;
 			if (Config.groupsranking.indexOf(userGroup) < Config.groupsranking.indexOf(Config.pmmodchat as GroupSymbol)) {
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1237,8 +1237,8 @@ export const commands: ChatCommands = {
 		if (Punishments.isBattleBanned(user)) {
 			return this.popupReply(`You are banned from battling and cannot challenge users.`);
 		}
-		if (!user.named || user.namelocked) {
-			return this.popupReply(`Unnamed users cannot challenge.`);
+		if (!user.named) {
+			return this.popupReply(`You must choose a username before you challenge someone.`);
 		}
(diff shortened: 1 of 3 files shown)


server/users.ts
```diff
@@ -382,6 +382,7 @@ export class User extends Chat.MessageContext {
 	statusType: StatusType;
 	userMessage: string;
 	lastWarnedAt: number;
+	currentChallengeCount: number;
```
Zarel, line 385: I don't think this needs more state; the challenge count should be able to be calculated from existing state (`Ladders.challenges`?)

server/ladders.ts
```diff
@@ -296,7 +304,9 @@ class Ladder extends LadderStore {
 			const fromUser = Users.get(challenge.from);
 			if (fromUser) Ladder.updateChallenges(fromUser);
 			const toUser = Users.get(challenge.to);
-			if (toUser) Ladder.updateChallenges(toUser);
+			if (toUser) {
+				Ladder.updateChallenges(toUser);
+			}
```
Zarel, line 309:
```suggestion
			if (toUser) Ladder.updateChallenges(toUser);
```

server/ladders.ts
```diff
@@ -314,7 +324,9 @@ class Ladder extends LadderStore {
...
```
Zarel, line 329:
```suggestion
			if (toUser) Ladder.updateChallenges(toUser);
```

server/chat-commands/core.ts
```diff
@@ -1237,6 +1237,9 @@ export const commands: ChatCommands = {
 		if (Punishments.isBattleBanned(user)) {
 			return this.popupReply(`You are banned from battling and cannot challenge users.`);
 		}
+		if (!user.named || user.namelocked) {
+			return this.popupReply(`Unnamed users cannot challenge.`);
+		}
```
Zarel, line 1242:
```suggestion
		if (!user.named) {
			return this.popupReply(`You must choose a username before you challenge someone.`);
		}
```

Namelocked users should be able to challenge other locked/namelocked users but no one else; this was discussed elsewhere.

Zarel:
Nice! `Ladders.challenges.get(...)?.length` makes it a lot nicer!

server/chat-commands/core.ts
mia-pi-git: Should i add that here, as well?

Zarel: New PR also works.

mia-pi-git: Okay, then!
I had that committed so i guess i'll just push and PR.

PR #7046, opened 2020-07-24
Remove User#authAtLeast and add more sysop bypasses

AnnikaCodes:
This ~~allows global staff to set modchat up to % in battles, and~~ allows sysops to bypass a few permissions checks that they previously weren't able to.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -288,12 +288,12 @@ exports.restrictLinks = false;
 exports.chatmodchat = false;
 /**
  * battle modchat - default minimum group for speaking in battles; changeable with /modchat
- * @type {false | string}
+ * @type {false | AuthLevel}
  */
 exports.battlemodchat = false;
 /**
- * pm modchat - minimum group for PMing other users, challenging other users
- * @type {false | string}
+ * PM modchat - minimum group for sending private messages or challenges to other users
+ * @type {false | AuthLevel}
  */
 exports.pmmodchat = false;
 /**
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1223,7 +1223,7 @@ export const commands: ChatCommands = {
 	async search(target, room, user, connection) {
 		if (target) {
 			if (Config.laddermodchat) {
-				if (!Users.globalAuth.atLeast(user, Config.laddermodchat)) {
+				if (!user.isSysop && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
 					const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
 					this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
 					return false;
@@ -1266,7 +1266,7 @@ export const commands: ChatCommands = {
 			return this.popupReply(`You must choose a username before you challenge someone.`);
 		}
 		if (Config.pmmodchat) {
-			if (Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
+			if (!user.isSysop && !Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
 				const groupName = Config.groups[Config.pmmodchat].name || Config.pmmodchat;
 				this.popupReply(`Because moderated chat is set, you must be of rank ${groupName} or higher to challenge users.`);
 				return false;
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1222,12 +1222,10 @@ export const commands: ChatCommands = {
 
 	async search(target, room, user, connection) {
 		if (target) {
-			if (Config.laddermodchat) {
-				if (!user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
-					const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
-					this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
-					return false;
-				}
+			if (Config.laddermodchat && !user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
+				const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
+				this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
+				return false;
 			}
 			const ladder = Ladders(target);
 			if (!user.registered && Config.forceregisterelo && await ladder.getRating(user.id) >= Config.forceregisterelo) {
@@ -1265,12 +1263,10 @@ export const commands: ChatCommands = {
 		if (!user.named) {
 			return this.popupReply(`You must choose a username before you challenge someone.`);
 		}
-		if (Config.pmmodchat) {
-			if (!user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
-				const groupName = Config.groups[Config.pmmodchat].name || Config.pmmodchat;
-				this.popupReply(`Because moderated chat is set, you must be of rank ${groupName} or higher to challenge users.`);
-				return false;
-			}
+		if (Config.pmmodchat &&!user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
+			const groupName = Config.groups[Config.pmmodchat].name || Config.pmmodchat;
+			this.popupReply(`Because moderated chat is set, you must be of rank ${groupName} or higher to challenge users.`);
+			return false;
 		}
 		return Ladders(target).makeChallenge(connection, targetUser);
 	},
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1224,7 +1224,7 @@ export const commands: ChatCommands = {
 		if (target) {
 			if (Config.laddermodchat && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
 				const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
-				this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
+				this.popupReply(`This server requires you to be rank ${groupName} or higher to search for a battle.`);
 				return false;
 			}
 			const ladder = Ladders(target);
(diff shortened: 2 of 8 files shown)


server/users.ts
```diff
@@ -527,7 +527,7 @@ export class User extends Chat.MessageContext {
 		return status;
 	}
 	authAtLeast(minAuth: string, room: BasicRoom | null = null) {
-		if (!minAuth || minAuth === ' ') return true;
+		if (!minAuth || minAuth === ' ' || this.isSysop) return true;
```
scheibo:
```suggestion
		if (this.isSysop) return true;
		if (!minAuth || minAuth === ' ') return true;
```

server/rooms.ts
```diff
@@ -485,7 +485,7 @@ export abstract class BasicRoom {
 		}
 		if (!(userGroup in Config.groups)) return false;
 		if (!(modjoinGroup in Config.groups)) throw new Error(`Invalid modjoin setting in ${this.roomid}: ${modjoinGroup}`);
-		return Config.groups[userGroup].rank >= Config.groups[modjoinGroup].rank;
+		return Config.groups[userGroup].rank >= Config.groups[modjoinGroup].rank || user.isSysop;
```
scheibo, line 488: I'd add `if (user.isSysop) return true` to the top - sysop checks up bail early, they dont really get checkout alone with other things

AnnikaCodes: > they dont really get checkout alone with other things

?

Bailing out early is good though, I'll change that.

scheibo: LOL this not entirely sure where I was going with that sentence either

Zarel:
This is a hardcode of a lot of permissions that should be fixed at the permission engine level... I'll need to look more closely at the engine first.

Zarel:
I've pushed 64b12ae58eb1c6d and c08a1a31998dc02704b. How much of this is still necessary/useful?

AnnikaCodes:
Just the stuff about sysops.

server/rooms.ts
Zarel: "they don't really need to get checked along with the other things"?

server/chat-commands/core.ts
```diff
@@ -1203,13 +1203,14 @@ export const commands: ChatCommands = {
...
-				const userGroup = user.group;
-				if (Config.groupsranking.indexOf(userGroup) < Config.groupsranking.indexOf(Config.laddermodchat)) {
-					const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
-					this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
-					return false;
-				}
+			if (
+				!user.hasSysopAccess() &&
+				Config.laddermodchat &&
+				Config.groupsranking.indexOf(user.group) < Config.groupsranking.indexOf(Config.laddermodchat)
```
Zarel, line 1209: How much of these comparisons can be replaced with `authAtLeast`?

Should we maybe look into consolidating all our "auth at least this high" code? iirc in addition to the function in `user-groups` we also have one in `rooms` or something.

server/chat-commands/core.ts
```diff
@@ -1202,12 +1202,10 @@ export const commands: ChatCommands = {
 
 	async search(target, room, user, connection) {
 		if (target) {
-			if (Config.laddermodchat) {
-				if (!Users.globalAuth.atLeast(user, Config.laddermodchat)) {
-					const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
-					this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
-					return false;
-				}
+			if (Config.laddermodchat && !user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
```
Zarel, line 1205: Maybe put the sysop check inside `Auth#atLeast`?

server/rooms.ts
```diff
@@ -469,6 +469,7 @@ export abstract class BasicRoom {
 		Rooms.global.writeChatRoomData();
 	}
 	checkModjoin(user: User) {
+		if (user.hasSysopAccess()) return true;
```
Zarel, line 472: In fact, imagine if the entirety of `checkModjoin` were moved inside `Auth#atLeast` (so it would support `trusted` and `autoconfirmed` as possible values that could be passed to `atLeast`).

server/users.ts
```diff
@@ -530,6 +530,7 @@ export class User extends Chat.MessageContext {
 		return status;
 	}
 	authAtLeast(minAuth: string, room: BasicRoom | null = null) {
```
Zarel, line 532: This function, too: it's probably redundant with `Auth#atLeast`, right?

server/rooms.ts
AnnikaCodes: Implemented.

server/users.ts
AnnikaCodes: It is.

server/chat-commands/core.ts
```diff
@@ -1265,12 +1263,10 @@ export const commands: ChatCommands = {
...
 		}
-		if (Config.pmmodchat) {
-			if (Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
-				const groupName = Config.groups[Config.pmmodchat].name || Config.pmmodchat;
-				this.popupReply(`Because moderated chat is set, you must be of rank ${groupName} or higher to challenge users.`);
-				return false;
-			}
+		if (Config.pmmodchat && !user.hasSysopAccess() && !Users.globalAuth.atLeast(user, Config.pmmodchat as GroupSymbol)) {
+			const groupName = Config.groups[Config.pmmodchat].name || Config.pmmodchat;
+			this.popupReply(`Because moderated chat is set, you must be of rank ${groupName} or higher to challenge users.`);
```
Zarel, line 1268:
```suggestion
			this.popupReply(`This server requires you to be rank ${groupName} or higher to challenge users.`);
```

server/chat-commands/core.ts
```diff
@@ -1222,12 +1222,10 @@ export const commands: ChatCommands = {
...
 		if (target) {
-			if (Config.laddermodchat) {
-				if (!Users.globalAuth.atLeast(user, Config.laddermodchat)) {
-					const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
-					this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
-					return false;
-				}
+			if (Config.laddermodchat && !Users.globalAuth.atLeast(user, Config.laddermodchat)) {
+				const groupName = Config.groups[Config.laddermodchat].name || Config.laddermodchat;
+				this.popupReply(`On this server, you must be of rank ${groupName} or higher to search for a battle.`);
```
Zarel, line 1227:
```suggestion
				this.popupReply(`This server requires you to be rank ${groupName} or higher to search for a battle.`);
```

server/chat-commands/core.ts
Zarel, line 1268: "Moderated chat" is a term from IRC and never actually made sense to me; we might want to think of a better word for it in general.

Zarel: "Authfilter"?

server/chat.ts
```diff
@@ -108,6 +108,7 @@ import {formatText, linkRegex, stripFormatting} from './chat-formatter';
 
 // @ts-ignore no typedef available
 import ProbeModule = require('probe-image-size');
+import {ExtendedGroup} from './user-groups';
```
Zarel, line 111: I would call this `AuthLevel` and also define `room.settings.modchat` and `Config.whatevermodchat` as it, so they don't need to be cast.

server/rooms.ts
```diff
@@ -473,23 +473,12 @@ export abstract class BasicRoom {
 		if (!this.settings.modjoin) return true;
 		// users with a room rank can always join
 		if (this.auth.has(user.id)) return true;
-		const userGroup = user.can('makeroom') ? user.group : this.auth.get(user.id);
 
 		const modjoinSetting = this.settings.modjoin !== true ? this.settings.modjoin : this.settings.modchat;
 		if (!modjoinSetting) return true;
-		let modjoinGroup = modjoinSetting;
+		if (!(modjoinSetting in Config.groups)) throw new Error(`Invalid modjoin setting in ${this.roomid}: ${modjoinSetting}`);
```
Zarel, line 479: This prevents `trusted` and `autoconfirmed` to be used. I might get rid of this `throw` entirely (maybe replace it with `Monitor.error`); it's probably bad form to crash the server because you removed a group from the config file.

Zarel: Maybe with a `isAuthLevel` function in `Users.Auth`?

server/chat-commands/core.ts
AnnikaCodes: Considering the command is called `/modchat`, I think just calling it modchat (as in only **mod**s [or anyone with rank > X] can **chat**) would make sense. I think having a `/modchat` command to enable authfilter would just be confusing, although we could rename the command, I suppose. (Also, the namechange should probably go in a different commit.)

server/user-groups.ts
```diff
@@ -2,6 +2,7 @@ import {FS} from '../lib/fs';
 
 export type GroupSymbol = '~' | '&' | '#' | '★' | '*' | '@' | '%' | '☆' | '+' | ' ' | '‽' | '!';
```
scheibo, line 3: Should ~ still be included here?

AnnikaCodes: I feel like a lot of side servers still use it; keeping it in would save them the trouble of having to patch it back in. On the other hand, PS main doesnt use it at all, so keeping it in as a legacy rank does feel a bit odd.

server/chat-commands/core.ts
Zarel: Yeah, I know. I think for now we can stick with `modchat`, but I've been leaning towards not calling it that in error messages, as you've seen here.

config/config-example.js
```diff
@@ -288,12 +288,12 @@ exports.restrictLinks = false;
 exports.chatmodchat = false;
 /**
  * battle modchat - default minimum group for speaking in battles; changeable with /modchat
```
Zarel, line 290: In hindsight it would make more sense to call these `modbattle` and `modpm`, but I guess the ship has sailed. Or maybe we could support them as the new names for these?

server/user-groups.ts
Zarel: I think this is fine as-is.

Zarel:
Sadly, the chat command changes aren't hotpatchable.

PR #7047, opened 2020-07-24
Remove mention of chat.commands.js from /hotpatch's help

mia-pi-git:
suuuuuper minor but inaccurate thing.


Zarel:
Yep, that's definitely outdated.

PR #7048, opened 2020-07-24
Link IPs in modlog

whalemer:


server/chat-plugins/modlog.ts
```diff
@@ -44,6 +44,7 @@ const LONG_QUERY_DURATION = 2000;
 const LINES_SEPARATOR = 'lines=';
 const MAX_RESULTS_LENGTH = MORE_BUTTON_INCREMENTS[MORE_BUTTON_INCREMENTS.length - 1];
 const LOG_PATH = 'logs/modlog/';
+const IPS_REGEX = /[([]([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)[)\]]/g;
```
scheibo, line 47: `\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}` seems more desirable, or at the very least `\d+` instead of `[0-9]+` (though obviously you didnt write this regex)

scheibo: from chat, maybe `[0-9]{1,3}` is more desirable

Zarel:
Oh, this will be useful, thanks!

PR #7049, opened 2020-07-24
Don't modnote twice on joining a helpticket

whalemer:


server/chat-plugins/helptickets.ts
```diff
@@ -207,7 +207,7 @@ export class HelpTicket extends Rooms.RoomGame {
 	}
 
 	modnote(user: User, text: string) {
-		this.room.addByUser(user, text);
+		this.room.addByUser(user, text).update();
```
Zarel, line 210: ?

How would this change the number of modnotes?

Also, `this.room` should automatically be updated at the end of a chat command; do you know why it's not?

whalemer: when you join a helpticket, the message gets added to the scrollback, so you see it in the backlog.

then, when you speak, the room updates, and the message gets broadcast, so you see a second copy.

joining updates the room after sending the `|j|` message, but that happens before the `onJoin` function gets called.

Zarel: It sounds like the joining code is what needs to be fixed?

whalemer: the joining code only actually updates the room if the user is named and can talk, so we'd either have to update the room unconditionally in it (bad), or have the condition that the room will update after your plugin's join handler _if_ the user was named (bad, ugly). 

as far as I know, helptickets are unique in that they can send a message in their join handler, so the current behaviour (needing to manually update the room if you send a message in your join handler) seems sensible enough to me.

Zarel:
I'll have to take a look at the joining code later, but in the meantime, thanks for this fix!

PR #7050, opened 2020-07-24
Moderation: Support private reasons in /warn

AnnikaCodes:


Zarel:
I'm not sure `[proof]` is the right term; I would go with `[private]`.

Also, is there any reason to use "spoiler"? Is that already used by other commands?

mia-pi-git:
> I'm not sure `[proof]` is the right term; I would go with `[private]`.
> 
> Also, is there any reason to use "spoiler"? Is that already used by other commands?

Spoiler is used by other commands with this functionality, yes.

Zarel:
On reflection, I would call it `[staff-only reason]` or `[private reason]`. I think that's just much clearer about its purpose.

AnnikaCodes:
Rebased this! @Zarel, ETA on merging this?

Zarel:
I don't remember why I didn't merge earlier, but I think it might have been the conflict? I certainly don't have anything else waiting on this.

PR #7051, opened 2020-07-24
Chat commands: Make /sendhtmlpage send to the last active connection

mia-pi-git:
Minor QOL thing, but apparently it'd be nice to have this.
See: Felucia in dev talking about it being sent to the first conn, not the current.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -217,7 +217,7 @@ export const commands: ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
-	sendhtmlpage(target, room, user, connection) {
+	sendhtmlpage(target, room, user) {
 		if (!room) return this.requiresRoom();
 		if (!this.can('addhtml', null, room)) return false;
 		let [targetID, pageid, content] = Utils.splitFirst(target, ',', 2);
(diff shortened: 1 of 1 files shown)


Zarel:
This is a really great improvement, thanks!

Even better would be some way of tracking which connection the request came from, which I think we should also do. I think that's very doable, actually; just include the connection index with the request message.

TheMezStrikes:
If you track which connection the request came from, you'd also need to let the `/sendhtmlpage` command select which connection to send data to, which is likely to end up being unintuitive

mia-pi-git:
It'd be uninutive, sure, but doable. 
just something like `|pm|sender|receiver||requestpage|pageid|connIndex` and then providing an option for /sendhtmlpage shouldn't be that hard. I find that the only issue would be, what would be the proper way to specify an index? Break all bots currently using the command or risk the provided HTML tripping up on it?

Zarel:
I'm not opposed to breaking all bots currently using it, tbh. I'd rather than than use an unwieldy syntax for it.

mia-pi-git:
Okay. I'm still kind of confused why specifying a connection is better than just going for the last active connection, but i suppose that can be discussed in the follow-up PR. I'll write it now.

Zarel:
> why specifying a connection is better than just going for the last active connection

Race conditions, mostly. Imagine I click `view-bot-scrappie-whatever` on my phone right before I click "search" on my computer.

mia-pi-git:
Ah, fair.

PR #7053, opened 2020-07-24
Support requiring registration on certain IPs

AnnikaCodes:


Zarel:
Was this a requested feature? I remember opposition to general forced-registration on the grounds that it's just going to force bots to register and clutter up the username database but not otherwise change things.

AnnikaCodes:
> Was this a requested feature? I remember opposition to general forced-registration on the grounds that it's just going to force bots to register and clutter up the username database but not otherwise change things.

Yes, I've been told it was requested by global staff, since registered users are easier to moderate.

scheibo:
@Zarel https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/post-8526566. Though perhaps the answer there is being able to permalock unregistered users? For Annika's benefit:

> Lately there has been a few users that should be permalocked yet, since they are unregistered users, we cannot nominate them, namely serson and kokirisfdp. Only thing we can do is monthlock, which currently is only available to admins.

> What we would like to suggest is that there be a way to force a given IP to register the next time they log on, so that IP can get permalocked with an account linked to it.

> An alternative would be to enable monthlocks to mods to better deal with these types of users. This can be enabled once a user has already been monthlocked by an admin, if necessary.

AnnikaCodes:
Oh, given the full context, it seems like allowing mods to monthlock or storing IPs for unregistered users so they can be permalocked would be a better solution (though I'm fine going through with this as well; I'm not really supposed to have a voice in this sort of thing).

Zarel:
> What we would like to suggest is that there be a way to force a given IP to register the next time they log on, so that IP can get permalocked with an account linked to it.

Permalocked IPs are also locked for one month; it's the exact same. :[

scheibo:
> Permalocked IPs are also locked for one month; it's the exact same. :[

Are permalocks IPs not locked permanently? Seems like very poor naming otherwise

Is that answer here to just change our `Config.groups` to give mods `monthlock`?

Zarel:
> Are permalocks IPs not locked permanently? Seems like very poor naming otherwise

Permalock will lock the account permanently and the IP for a month. We don't support locking IPs for longer than a month on the grounds that false positives just get too high past that point.

> Is that answer here to just change our `Config.groups` to give mods `monthlock`?

Mods can't permalock, either, though. And unless something is wrong, mods can still weeklock and get an admin to upgrade it to a monthlock.

scheibo:
From chatting with Annika and Anubis, I think people agree that this is perhaps not the correct approach to take here (though Annika implemented it well!) - sorry, thats on me for not gathering specifications better. I'm closing this, @Zarel, please see https://discordapp.com/channels/534298043176976395/565617827298541572/736420124902555780 and the logs linked there for the discussion.

Thanks Annika!

Zarel:
I do agree that Annika implemented it well!

PR #7056, opened 2020-07-24
Add a command to enable hotpatching

HoeenCoder:
Allows admins to easily enable hotpatching after disabling it. This allows us to use `/nohotpatch` for more than just restart blocks (eg: temporary issues where we should not patch like when we use eval to disable commands that have very serious bugs).

A reason is required so some information on why the lock was removed is available.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -428,7 +428,8 @@ export const commands: ChatCommands = {
 	],
 
 	hotpatchlock: 'nohotpatch',
-	nohotpatch(target, room, user) {
+	allowhotpatch: 'nohotpatch',
+	nohotpatch(target, room, user, connection, cmd) {
 		if (!this.can('gdeclare')) return;
 		if (!target) return this.parse('/help nohotpatch');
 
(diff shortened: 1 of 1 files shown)


Zarel:
I previously just evaled the re-enable, but I suppose a command makes it easier.

server/chat-commands/admin.ts
```diff
@@ -428,7 +428,8 @@ export const commands: ChatCommands = {
 	],
 
 	hotpatchlock: 'nohotpatch',
-	nohotpatch(target, room, user) {
+	allowhotpatch: 'nohotpatch',
```
Zarel, line 431: I'd alias `yeshotpatch` tbh

Zarel:
```suggestion
	yeshotpatch: 'nohotpatch',
	allowhotpatch: 'nohotpatch',
```

server/chat-commands/admin.ts
```diff
@@ -441,7 +442,15 @@ export const commands: ChatCommands = {
 		const lock = Monitor.hotpatchLock;
 		const validDisable = ['chat', 'battles', 'formats', 'validator', 'tournaments', 'punishments', 'all'];
 
-		if (validDisable.includes(hotpatch)) {
+		if (!validDisable.includes(hotpatch)) return this.errorReply(`This hot-patch is not an option to disable`);
+		const enable = cmd === 'allowhotpatch';
```
Zarel, line 446:
```suggestion
		const enable = ['allowhotpatch', 'yeshotpatch'].includes(cmd);
```

server/chat-commands/admin.ts
```diff
@@ -441,7 +442,15 @@ export const commands: ChatCommands = {
 		const lock = Monitor.hotpatchLock;
 		const validDisable = ['chat', 'battles', 'formats', 'validator', 'tournaments', 'punishments', 'all'];
 
-		if (validDisable.includes(hotpatch)) {
+		if (!validDisable.includes(hotpatch)) return this.errorReply(`This hot-patch is not an option to disable`);
```
Zarel, line 445:
```suggestion
		if (!validDisable.includes(hotpatch)) return this.errorReply(`Disabling hotpatching "${hotpatch}" is not supported.`);
```

PR #7057, opened 2020-07-25
Chat commands: Support specifying a target connection in /sendhtmlpage

mia-pi-git:
see #7051 .


Zarel:
Unfortunately, peach just came up with an even better idea: Just store which pages a connection is requesting:

https://github.com/smogon/pokemon-showdown/pull/7044#discussion_r460354529

Zarel:
We still need to do the fix mentioned in #7044 though; can you get on that?

PR #7058, opened 2020-07-25
Add a /hiderank command

AnnikaCodes:


server/users.ts
```diff
@@ -540,11 +542,21 @@ export class User extends Chat.MessageContext {
...
+		target: User | null,
+		room?: BasicRoom | null,
+		useVisualGroup?: boolean
+	): boolean;
+	can(
+		permission: string,
+		target: User | null = null,
+		room: BasicRoom | null = null,
+		useVisualGroup?: boolean
+	): boolean {
```
Zarel, line 558: Instead of overriding `user.can`, I think `this.can` should just call `Auth.hasPermission` directly.

server/user-groups.ts
```diff
@@ -92,13 +99,14 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 		user: User,
 		permission: string,
 		target: User | GroupSymbol | null,
-		room?: Room | BasicRoom | null
+		room?: Room | BasicRoom | null,
+		useVisualGroup?: boolean
```
Zarel, line 103: I'm not sure the `useVisualGroup` parameters everywhere are necessary. One thing we could do to get around this is to just always unhiderank+retry on permission failure. And then all permissions would be based on visual rank.

AnnikaCodes: The issue I see with that is that (from my understanding; I'm probably missing something) non-essential checks would be based on visual group. For example, if an admin hid their rank as a Voice, then tried to use `/alts` on another user, they wouldn't be able to see that user's alts or IP address without unhiding their rank. Is that what you want?

Zarel:
I swear, a lot of permission complexity definitely has to do with `ip` and `ipself` being the same permission. They really should be separate permissions; it would simplify a lot of the other code that's forced to use `gban` for IP checking because there's no specific user to target.

AnnikaCodes:
> I swear, a lot of permission complexity definitely has to do with `ip` and `ipself` being the same permission. They really should be separate permissions; it would simplify a lot of the other code that's forced to use `gban` for IP checking because there's no specific user to target.

Fair enough, but even with a separate `ipself` permission, a user who was hiding their rank would have to reset it / have it reset in order to perform "invisible" actions that require a permission check (viewing modlogs is another common example).

Zarel:
To be clear, I think the code is getting more complex here in a way I don't particularly like, and I think that refactor would help simplify parts so that it's not too complex overall.

But yeah, there's no need to do that before merging your changes, it's easy enough to do after.

Zarel:
Sadly, this conflicts with #6883.

Zarel:
I suppose I'll have to manually merge them next restart.

AnnikaCodes:
> I suppose I'll have to manually merge them next restart.

I take it you don't want me to rebase this then?

Zarel:
I was going to rebase it for the last restart, but after manually rebasing the other two I realized this might need more workshopping.

Zarel:
Yes, please do rebase this.

Zarel:
I think something about the API complexity is still bothering me.

I think it might be the defaulting to "using the underlying group". I think the visual group is a better default.

We might also want to remove `user.group` completely. `Users.globalAuth` already has it anyway, and then only `user.visualGroup` would be a thing. Although maybe that belongs in a separate PR, too.

AnnikaCodes:
Alrighty, I rebased it. I think that removing `user.group`, while a good idea, would probably be a bit beyond the scope of adding /hiderank.

Zarel:
Sadly, an older version of this PR has already hit the codebase. It might be better just to iterate on that one or something?

Zarel:
Drafting because _something_ needs to be done here even if I'm not sure what.

AnnikaCodes:
OK, I rebased this and it turns out the only thing left to do is add the command.

AnnikaCodes:
@Zarel is this good for next restart?

Zarel:
I still need to review the implementation. Shouldn't take long.

PR #7060, opened 2020-07-26
Disallow global bots from bypassing blockpms

SSlayer9:
https://www.smogon.com/forums/threads/disallow-global-bots-from-bypassing-blocked-pms-or.3666899/
this ^. If one of the alternative solutions is better lmk.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -989,7 +989,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !user.authAtLeast(targetUser.settings.blockPMs)) &&
-					(!user.can('lock' || user.auth === '*'))) {
+					(!user.can('lock' || user.group === '*'))) {
 					Chat.maybeNotifyBlocked('pm', targetUser, user);
 					if (!targetUser.can('lock')) {
 						this.errorReply(`This user is blocking private messages right now.`);
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -989,7 +989,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !user.authAtLeast(targetUser.settings.blockPMs)) &&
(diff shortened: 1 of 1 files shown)


SSlayer9:
check failed which is odd, i tested this and it worked

server/chat.ts
```diff
@@ -989,7 +989,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !user.authAtLeast(targetUser.settings.blockPMs)) &&
-					!user.can('lock')) {
+					(!user.can('lock' || user.group === '*'))) {
```
AnnikaCodes, line 992:
```suggestion
					(!user.can('lock') || user.group === '*') {
```

AnnikaCodes: Personally, I'd implement this by adding a `bypassBlockPMs` permission rather than hardcoding `*` rank.

AnnikaCodes: (Also, you should test your changes locally before opening a pull request -- that way, you could have caught the error I fixed in my suggestion.)

server/chat.ts
```diff
@@ -989,7 +989,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !user.authAtLeast(targetUser.settings.blockPMs)) &&
-					!user.can('lock')) {
+					(!user.can('lock') || user.group === '*') {
```
AnnikaCodes, line 992:
```suggestion
					(!user.can('lock') || user.group === '*')) {
```
that's my fault, I'm bad at reading :)

Zarel: Global bots can't lock, though...

Zarel: Is this actually the check? If so, in theory, nothing else needs to be done, bot PMs are already blockable.

AnnikaCodes: I can confirm that global bots don't bypass blocked PMs on main. This PR can probably be closed along with the suggestion thread.

Zarel:
Thanks for the attempt, at least.

PR #7061, opened 2020-07-26
Refactor IPTools and add a host management chat plugin

AnnikaCodes:
This PR:
 - Refactors datacenter handling, giving them their own type and proper methods to manage them (instead of manually accessing the file in `admin.ts`)
 - Refactors proxy and host storage, moving them into their own `.csv` file (instead of hardcoding them in `ip-tools.ts`). (Eventually we should probably move all the various files used for data storage on the server into some kind of proper database, but that's out of scope of this PR.)
 - Adds a new chat plugin, `hosts.ts`, that contains commands to view and manage proxies, datacenters, and hosts. This also includes the `/adddatacenters` command, which has been migrated from `admin` and now uses the proper access methods (rather than writing straight to the file).

This should greatly reduce the need for PRs to update the actual `ip-tools.ts` file. (There are still a few things that are hardcoded, but not nearly so many.)


server/chat-plugins/hosts.ts
```diff
@@ -0,0 +1,352 @@
+/**
+ * Chat plugin to help manage hosts, proxies, and datacenters
+ * Written by Annika
+ * Original /adddatacenters command written by Zarel
+ */
+
+import {Utils} from "../../lib/utils";
+import {Datacenter} from "../ip-tools";
+
+const IP_REGEX = /^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/;
```
scheibo, line 10: I think I've seen this regex in a couple CLs now. Would it make sense for it to be a constant on Utils if its used in multiple places?

server/chat-plugins/hosts.ts
```diff
@@ -0,0 +1,352 @@
...
+			html += Utils.html`<td>${datacenter.name}</td>`;
+			html += Utils.html`<td>${datacenter.host}</td>`;
+			html += `</tr>`;
+		}
+		html += `</table></div>`;
+		return html;
+	},
+
+	hosts(query) {
+		function ipSort(a: string, b: string) {
```
scheibo, line 34: declaring a `function` here is slightly awkward - would it make sense to move this to the top level?

server/chat-plugins/hosts.ts
```diff
@@ -0,0 +1,352 @@
...
+		// Originally by Zarel
+		widen: 'add',
+		async add(target, room, user, connection, cmd) {
+			if (!this.can('lockdown')) return false;
+			if (!target) return this.parse('/help datacenters add');
+			// should be in the format: IP, IP, name, URL
+			const widen = (cmd === 'widendatacenters');
+
+			const datacentersToAdd: Datacenter[] = [];
+			for (const row of String(target).split("\n")) {
```
scheibo, line 107: I believe our guidelines on stringles would have you use `'\n'` here?

Zarel: Why the string cast? I also, I would use `` `\n` ``.

server/chat-plugins/hosts.ts
AnnikaCodes: Based on my searching, it's only really used here -- there's an `IPS_REGEX` in modlog, but that searches for IPs specifically in the context of modlogs (i.e. with `[a.b.c.d]` format), so I'm not sure it should be in Utils.

server/chat-plugins/hosts.ts
AnnikaCodes: (I'm not sure - this code comes from `admin.ts` and was ostensibly Zarel's work, so I figured there was a reason)

server/chat-plugins/hosts.ts
AnnikaCodes: (I might be missing a place, though)

Zarel:
Is this hotpatchable?

Zarel:
Looks like `/hotpatch iptools` should do it, right? But I'm asking just in case.

AnnikaCodes:
It's hotpatchable, yes, but you'd need to `/hotpatch chat` after hotpatching iptools.

Also, I'd like to add a whitelisting command to this so that @Lusamine can use the command without being Admin rank. (I wasn't aware permissions would be an issue when I opened the PR.) If you don't mind waiting a day to merge this, I can code that up. (If you do want to merge I can always make a separate PR.)

AnnikaCodes:
OK @Zarel I added the whitelist; this should be good whenever.

Zarel:
I would rather hardcode an Anubis exception than add a whole entire whitelist system just for her, tbh.

AnnikaCodes:
> I would rather hardcode an Anubis exception than add a whole entire whitelist system just for her, tbh.

Alright, that works too

Zarel:
Looks good, thanks!

Zarel:
I notice `config/hosts.csv` don't have any IP ranges. I could have sworn we supported those. Are those still around somewhere?

AnnikaCodes:
> I notice `config/hosts.csv` don't have any IP ranges. I could have sworn we supported those. Are those still around somewhere?

It seems like config/ipbans.txt contains some range bans, but that's handled in `punishments.ts`, not the IPTools.

Zarel:
Maybe they were moved to datacenters? Is that where e.g. the T-Mobile ranges are?

AnnikaCodes:
> Maybe they were moved to datacenters? Is that where e.g. the T-Mobile ranges are?

Maybe the hardcoded ranges in `getHost()`?

```ts
if (IPTools.checkPattern(rangeTmobile, ipNumber) || ip.startsWith('149.254.')) {
	resolve('tmobile.mobile-nohost');
	return;
}
```

Zarel:
Oh, yeah, are those still there? If you could push them into `hosts.csv` too that'd be even nicer.

AnnikaCodes:
> Oh, yeah, are those still there? If you could push them into `hosts.csv` too that'd be even nicer.

Yeah, that's definitely something I want to do, probably in a future PR.

PR #7065, opened 2020-07-26
Implement a /cleartext command

AnnikaCodes:
This command will allow staff to clear messages while telling the client not to show a reveal button. 
Requires client support in the form of https://github.com/smogon/pokemon-showdown-client/pull/1572.


Zarel:
Hm, this does not have special casing for roomauth yet. Do you want to add that before merge? This will be another few days before merge, at least; we generally wait one week or one server restart to make sure everyone has a new client.

AnnikaCodes:
What do you mean by "special casing for roomauth"?

I already added a button to view hidden lines for room/global staff:
```ts
const showRevealButton = !hideRevealButton || Users.globalAuth.isStaff(userid) || this.auth.isStaff(userid);
```

Zarel:
Oh, I didn't see that. Yeah, this looks good, then.

PR #7066, opened 2020-07-26
Display join/leave messages for locked players

AnnikaCodes:
This should fix the bugs about locked users not showing up in the userlist of battle rooms.
(Not letting the join/leave/rename messages of users who can't talk be displayed was originally done to prevent spamming — I assume that the spammers don't actually _play_ the battles? If they do, I can look for another solution.)

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -594,7 +594,7 @@ export abstract class BasicRoom {
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
 		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
-		if (this.reportJoins && (canTalk || this.auth.get(user) === Users.PLAYER_SYMBOL)) {
+		if (this.reportJoins && (canTalk || this.auth.has(user))) {
 			this.add(`|${type}|${entry}`).update();
 			return;
 		}
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -35,7 +35,7 @@ import {PM as RoomBattlePM, RoomBattle, RoomBattlePlayer, RoomBattleTimer} from 
 import {RoomGame, RoomGamePlayer} from './room-game';
 import {Roomlogs} from './roomlogs';
 import * as crypto from 'crypto';
-import {RoomAuth} from './user-groups';
+import {RoomAuth, PLAYER_SYMBOL} from './user-groups';
 
 /*********************************************************
  * the Room object.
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -594,7 +594,7 @@ export abstract class BasicRoom {
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
 		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
-		if (this.reportJoins && canTalk) {
+		if (this.reportJoins && (canTalk || this.auth.get(user) === PLAYER_SYMBOL)) {
```
Zarel, line 597:
```suggestion
		if (this.reportJoins && (canTalk || this.auth.get(user) === Users.PLAYER_SYMBOL)) {
```

server/rooms.ts
```diff
@@ -35,7 +35,7 @@ import {PM as RoomBattlePM, RoomBattle, RoomBattlePlayer, RoomBattleTimer} from
 import {RoomGame, RoomGamePlayer} from './room-game';
 import {Roomlogs} from './roomlogs';
 import * as crypto from 'crypto';
-import {RoomAuth} from './user-groups';
+import {RoomAuth, PLAYER_SYMBOL} from './user-groups';
```
Zarel, line 38:
```suggestion
import {RoomAuth} from './user-groups';
```

Zarel: tbh should we even be importing `RoomAuth` at all?

AnnikaCodes: It's used in the `BasicRoom` constructor to provide the auth handler:
```ts
this.auth = new RoomAuth(this);
```

server/rooms.ts
```diff
@@ -594,7 +594,7 @@ export abstract class BasicRoom {
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
 		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
-		if (this.reportJoins && canTalk) {
+		if (this.reportJoins && (canTalk || this.auth.get(user) === Users.PLAYER_SYMBOL)) {
```
Zarel, line 597:
```suggestion
		if (this.reportJoins && (canTalk || this.auth.has(user))) {
```

In hindsight, I don't think players need to be hardcoded here.

server/rooms.ts
Zarel: Hm, this should plausibly be `new Users.RoomAuth` (which should be exported from Users if it isn't currently).

server/rooms.ts
```diff
@@ -594,7 +594,7 @@ export abstract class BasicRoom {
 	}
 	reportJoin(type: 'j' | 'l' | 'n', entry: string, user: User) {
 		const canTalk = user.authAtLeast(this.settings.modchat ?? 'unlocked', this) && !this.isMuted(user);
-		if (this.reportJoins && canTalk) {
+		if (this.reportJoins && (canTalk || this.auth.has(user))) {
```
Zarel, line 597:
```suggestion
		if (this.reportJoins && (canTalk || this.auth.has(user.id))) {
```

Hrm, apparently we never updated `auth.has` to take users?

PR #7067, opened 2020-07-26
Trivia: Show creator in /trivia history

AnnikaCodes:
Also, now that we track the user who created a game, we can conform better to modlog conventions!


Zarel:
Thanks!

PR #7068, opened 2020-07-27
Add a command to show the user's team in battles

mia-pi-git:
See: https://www.smogon.com/forums/threads/make-a-showset-command-for-battle-rooms.3657680/

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -891,7 +891,7 @@ export const commands: ChatCommands = {
 				teamStrings = [indexedSet];
 			}
 		}
-		let resultString = Chat.stringifyTeam(teamStrings);
+		let resultString = Dex.stringifyTeam(teamStrings);
 		if (showAll) {
 			resultString = `<details><summary>View team</summary><${resultString}</details>`;
 		}
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -923,8 +923,8 @@ export const commands: ChatCommands = {
 		return this.sendReplyBox(resultString);
 	},
 	showsethelp: [
-		`/showteam - show the team you're using in the current battle (must be used in a battle you're a player in).`,
-		`/showset [number] - shows the set of the pokemon corresponding to that number (in original Team Preview order, not necessarily current order)`,
+		`!showteam - show the team you're using in the current battle (must be used in a battle you're a player in).`,
+		`!showset [number] - shows the set of the pokemon corresponding to that number (in original Team Preview order, not necessarily current order)`,
 	],
 
 	acceptdraw: 'offertie',
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -872,14 +872,22 @@ export const commands: ChatCommands = {
 		const battle = room.battle;
 		if (!showAll && !target) return this.parse(`/help showset`);
 		if (!battle) return this.errorReply("This command can only be used in a battle.");
-		let setNum;
-		if (target) {
-			setNum = parseInt(target);
-			if (isNaN(setNum) || setNum > 6) return this.errorReply(`Use a number between 1-6 to view a specific set.`);
-			setNum -= 1;
-		}
-		const teamStrings = await battle.getTeam(user, setNum);
+		let teamStrings = await battle.getTeam(user);
 		if (!teamStrings) return this.errorReply("Only players can extract their team.");
+		if (target) {
+			let setIndex;
+			const parsed = parseInt(target);
+			if (parsed > 6) return this.errorReply(`Use a number between 1-6 to view a specific set.`);
+			if (isNaN(parsed)) {
+				for (const set of teamStrings) {
+					if (toID(set.name) === toID(target)) setIndex = teamStrings.indexOf(set);
+				}
+				if (!setIndex) return this.errorReply(`Pokemon "${target}" is not on your team.`);
+			} else {
+				setIndex = parsed - 1;
+			}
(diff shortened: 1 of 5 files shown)


server/room-battle.ts
```diff
@@ -1190,6 +1207,16 @@ export class RoomBattleStream extends BattleStream {
...
 			}
 			break;
+		case 'requestteam':
+			message = message.trim();
+			const slotNum = parseInt(message.slice(1)) - 1;
+			if (isNaN(slotNum) || slotNum < 0) {
+				throw new Error(`Team requested for slot ${message}, but that slot does not exist.`);
+			}
+			const side = this.battle.sides[slotNum];
+			const team = side.team.map(set => JSON.stringify(set)).join('\n');
```
Zarel, line 1217: Instead of JSON, it's probably easier to pack and unpack the team.

server/chat-commands/core.ts
```diff
@@ -865,6 +865,58 @@ export const commands: ChatCommands = {
...
+			}).join(' / ');
+			const moveString = set.moves.map((moveID: string) => {
+				const move = Dex.getMove(moveID);
+				return `- ${move.name ? move.name : moveID}`;
+			}).join('<br />');
+
+			let buf = `${set.name} ${set.name === set.species ? '' : `(${set.species}) `}`;
+			buf += `${set.item ? `@ ${set.item}` : ''}<br />`;
+			buf += `Ability: ${Dex.getAbility(set.ability).name}<br />`;
+			buf += `Shiny: ${set.shiny === true ? 'Yes' : 'No'}<br />`;
```
Zarel, line 904: I suppose we don't have team exports currently available. I would probably grab them from the client rather than reimplementing them.

server/chat-commands/core.ts
```diff
@@ -865,6 +865,58 @@ export const commands: ChatCommands = {
...
+		if (!this.canTalk()) return false;
+		const showAll = cmd === 'showteam';
+		if (!room) return this.requiresRoom();
+		const battle = room.battle;
+		if (!showAll && !target) return this.parse(`/help showset`);
+		if (!battle) return this.errorReply("This command can only be used in a battle.");
+		let setNum;
+		if (target) {
+			setNum = parseInt(target);
+			if (isNaN(setNum) || setNum > 6) return this.errorReply(`Use a number between 1-6 to view a specific set.`);
```
Zarel, line 879: You should definitely be able to specify sets by nickname or species name.

server/chat-commands/core.ts
```diff
@@ -865,6 +865,58 @@ export const commands: ChatCommands = {
...
+		}).join('<br /><br />');
+		this.runBroadcast();
+		if (showAll) {
+			resultString = `<details><summary>View team</summary><${resultString}</details>`;
+		}
+		return this.sendReplyBox(resultString);
+	},
+	showsethelp: [
+		`/showset [number] OR /showteam - show part, or all, of your team in battle.`,
+		`If a [number] is provided, shows the set of the pokemon corresponding to that number.`,
```
Zarel, line 917:
```suggestion
		`/showteam - show the team you're using in the current battle (must be used in a battle you're a player in).`,
		`/showset [number] - shows the set of the pokemon corresponding to that number (in original Team Preview order, not necessarily current order)`,
```

server/room-battle.ts
```diff
@@ -1112,6 +1116,18 @@ export class RoomBattle extends RoomGames.RoomGame {
...
 	}
+	async getTeam(user: User) {
+		const id = user.id;
+		const player = this.playerTable[id];
+		if (!player) return;
+		void this.stream.write(`>requestteam ${player.slot}`);
+		const teamDataPromise = new Promise<string[]>(resolve => {
+			this.playerTeamResolver = resolve;
+		});
+		const resultStrings = await teamDataPromise;
```
Zarel, line 1127: Nice! You figured out how to grab data from subprocesses! I remember you were struggling with this in the battle persistence PR.

mia-pi-git: yeah, lol, I was actually looking at that for this

server/chat-commands/core.ts
```diff
@@ -865,6 +865,69 @@ export const commands: ChatCommands = {
...
+			}).join(' / ');
+			const moveString = set.moves.map((moveID: string) => {
+				const move = Dex.getMove(moveID);
+				return `- ${move.name ? move.name : moveID}`;
+			}).join('<br />');
+
+			let buf = `${set.name} ${set.name === set.species ? '' : `(${set.species}) `}`;
+			buf += `${set.item ? `@ ${set.item}` : ''}<br />`;
+			buf += `Ability: ${Dex.getAbility(set.ability).name}<br />`;
+			buf += `Shiny: ${set.shiny === true ? 'Yes' : 'No'}<br />`;
```
KrisXV, line 915:
```suggestion
			if (set.shiny) buf += `Shiny: Yes<br />`;
```

Zarel:
Drafting because of the team packing thing (which I can help with if you need help), and also we might want to unify team export code.

mia-pi-git:
Oh, I wasn't sure if that was a solid thing or not. I'll get on that soon, then. 
Dex can already pack teams, so it might be worth putting an unpacker in there? (can't find one easily but i'll look harder later)
Is there one that can be borrowed from client?

mia-pi-git:
Oh, no, I'm wrong, there is one.

Zarel:
We're gonna have to iterate more on `Dex.stringifyTeam` but I guess that can be done later. For now, this is good, thanks!

PR #7070, opened 2020-07-27
Helptickets: Stop modlogging to Help room

mia-pi-git:
The issue that caused them to log to help room was that when checking shared modlogs, it splits the ID by ``-``. As a result, it got the id `help`, and would write to `modlog_help.txt`. This PR renames shared modlog files to ``[id]-rooms``, so that it doesn't conflict with existing rooms.
Likewise, i've also written a conversion script (https://gist.github.com/mia-pi-git/20b4f501b99445ad592f521776fbf718) to move old ticket entries from the help modlog to the new modlog file. This can also be hotpatchable, someone just needs to force helpticket rooms to set up their modlogStreams again (i can also add that into the script, if that's easier?).

### DIFF
--- a/server/chat-plugins/modlog-viewer.ts
+++ b/server/chat-plugins/modlog-viewer.ts
@@ -61,6 +61,20 @@ function getMoreButton(
 	}
 }
 
+function getRoomID(id: ModlogID) {
+	switch (id) {
+	case 'helpticket':
+		return 'help-rooms' as ModlogID;
+	case 'groupchat':
+		return 'groupchat-rooms' as ModlogID;
+	case 'battle':
+		return 'battle-rooms' as ModlogID;
+	// more aliases can be added here
+	default:
+		return id;
+	}
+}
+
 function prettifyResults(
 	resultArray: string[], roomid: ModlogID, searchString: string, exactSearch: boolean,
 	addModlogLinks: boolean, hideIps: boolean, maxLines: number, onlyPunishments: boolean
@@ -145,6 +159,7 @@ async function getModlog(
 ) {
 	const targetRoom = Rooms.search(roomid);
 	const user = connection.user;
+	roomid = getRoomID(roomid);
 
 	// permission checking
 	if (roomid === 'all' || roomid === 'public') {
--- a/server/modlog.ts
+++ b/server/modlog.ts
@@ -131,7 +131,7 @@ export class Modlog {
 	}
 
 	getSharedID(roomid: ModlogID): ID | false {
-		return roomid.includes('-') ? toID(roomid.split('-')[0]) : false;
+		return roomid.includes('-') ? `${toID(roomid.split('-')[0])}-rooms` as ID : false;
 	}
 
 	/**


Zarel:
I always thought it wasn't a bug but was intentional?

mia-pi-git:
It leaks a lot of information that help staff shouldn't have.

AnnikaCodes:
> I always thought it wasn't a bug but was intentional?

If help ticket modlogs are supposed to be intermixed with a room's modlog (which seems like a bad idea in and of itself to me), it would make more sense for them to be in Staff's modlog. The Help room staff don't really have anything to do with help tickets.

Zarel:
> If help ticket modlogs are supposed to be intermixed with a room's modlog (which seems like a bad idea in and of itself to me), it would make more sense for them to be in Staff's modlog. The Help room staff don't really have anything to do with help tickets.

I think that's a weird way to think about it. Help tickets were originally thought of as Help subrooms, and were designed in an age where Help did not have room staff that weren't also global staff.

There's no reason to attach them to staff, other than wanting to continue intermixing them with a room's modlog. But I don't understand why doing it in the past means we have to keep doing it now.

Zarel:
Did you interpret my comment as objecting to the change? I'm not, I'm just objecting to the part where Mia calls the old behavior a bug.

mia-pi-git:
Oh, my bad then - seemed like a bug from my PoV.

Zarel:
There are a few changes in here that might warrant a closer look, though. Like, do we want to rename `help-USERID` to `helpticket-USERID`? It would certainly be easier than the other workaround I discussed with you, @mia-pi-git.

scheibo:
https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/post-8324863 is where a Help RO asked for help tickets to be separated from Help room logs (supported by another Help RO). It may certainly have been intentional, but I don't know that Help room staff or owners find the help tickets being intermingled with their logs.

If Help room staff were actually able to handle Help tickets themselves it might make more sense.

mia-pi-git:
it's pretty annoying. Unless we were to be able to handle tickets as well, for whatever reason, I see no reason our modlog should be spammed. For context, out of 3200 results, only 73 were related to help, on one search.

Zarel:
It's unfortunate that Annika's modlog refactor broke this. :(

mia-pi-git:
The script should still work. I'll update this to handle her modlog refactor in a second.

TheImmortal:
@Zarel?

AnnikaCodes:
For the record, #7277 will also accomplish that (although this is probably important for the interim).

Zarel:
Sorry, we keep getting restarts with huge changes and this keeps getting pushed back.

Zarel:
This requires a pretty complicated migration, right? Can you coordinate with me on Discord, @mia-pi-git?

mia-pi-git:
> This requires a pretty complicated migration, right? Can you coordinate with me on Discord, @mia-pi-git?

The migration should be fairly simple. I'll detail more on discord.

PR #7071, opened 2020-07-27
/hidetext: Don't highlight the user

mia-pi-git:
Minor QOL thing that's slightly annoying.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1552,9 +1552,9 @@ export const commands: ChatCommands = {
 		}
 
 		if (targetUser && showAlts) {
-			room.send(
-				`|c|${user.getIdentity()}|/log ${name}'s alts messages ` +
-				`were cleared from ${room.title} by ${user.name}.${(reason ? ` (${reason})` : ``)}`
+			room.sendByUser(
+				user,
+				`${name}'s alts messages were cleared from ${room.title} by ${user.name}.${(reason ? ` (${reason})` : ``)}`
 			);
 			this.modlog('HIDEALTSTEXT', targetUser, reason, {noip: 1});
 			room.hideText([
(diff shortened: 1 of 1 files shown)


server/chat-commands/moderation.ts
```diff
@@ -1552,7 +1552,7 @@ export const commands: ChatCommands = {
 		}
 
 		if (targetUser && showAlts) {
-			room.send(`|c|~|${name}'s alts messages were cleared from ${room.title} by ${user.name}.${(reason ? ` (${reason})` : ``)}`);
+			room.send(`|c|${user.name}|${name}'s alts messages were cleared from ${room.title} by ${user.name}.${(reason ? ` (${reason})` : ``)}`);
```
Zarel, line 1555:
```suggestion
			room.sendByUser(user, `${name}'s alts messages were cleared from ${room.title} by ${user.name}.${(reason ? ` (${reason})` : ``)}`);
```

Zarel: You don't want it to be a chat message, after all.

Zarel:
Why aren't you using `sendByUser`?

Zarel:
Looks good now, thanks!

Zarel:
For general reference, I really don't like cutting off a string in the middle of a sentence. We intentionally make string-only lines an exception to the line length limits, and if you do prefer to split a string into multiple strings, I generally want them to be done at obvious break points, like after each `<br />` or something.

whalemer:
the reason this was implemented this way is so that hidetexting yourself won't hide the log message (and the attached reveal button).

(sending the log message after the hidetext doesn't work, because then the reveal button has nothing to attach to)

(yes, I realize now I probably should've put that info in a comment somewhere)

Zarel:
Oh, I guess we can hardcode that one?

whalemer:
could also just make the message a non-attributed non-hling message, like `||`, since I'm not sure if anyone cares about being able to HL on it.

Zarel:
I pushed a partial workaround in add3a5d9f0a.

But on reflection, I think the correct fix is client-side: make it skip the most recent line (that's the one the reveal button needs to be attached to, anyway).

PR #7073, opened 2020-07-27
Support requiring high-rated users to register

AnnikaCodes:
This would, for example, let PS require users to register their accounts to continue laddering past 1500 ELO.
The Register button in the popup will not work without client support (https://github.com/smogon/pokemon-showdown-client/pull/1574).

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -310,6 +310,13 @@ exports.laddermodchat = false;
 exports.forcetimer = false;
 
 /**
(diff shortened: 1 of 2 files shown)


scheibo:
For context - this is the staff suggestion that has the highest support out of pretty much any of them: https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/post-8387549

PR #7074, opened 2020-07-28
Punishments: Make autolocks last a week on users with bad records

AnnikaCodes:
In the future I should probably split modlog into two files so we're not importing from a chat plugin.


server/chat-plugins/modlog.ts
```diff
@@ -46,6 +46,12 @@ const MAX_RESULTS_LENGTH = MORE_BUTTON_INCREMENTS[MORE_BUTTON_INCREMENTS.length
 const LOG_PATH = 'logs/modlog/';
 const IPS_REGEX = /[([]([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})[)\]]/g;
 
+const GLOBAL_PUNISHMENTS = [
+	'WEEKLOCK', 'LOCK', 'BAN', 'RANGEBAN', 'RANGELOCK', 'FORCERENAME',
+	'TICKETBAN', 'AUTOLOCK', 'AUTONAMELOCK', 'NAMELOCK', 'AUTOBAN', 'MONTHLOCK',
+];
+const GLOBAL_PUNISHMENTS_REGEX_STRING = `\\b(${GLOBAL_PUNISHMENTS.join('|')}):.*`;
+
 const PUNISHMENTS = [
```
scheibo, line 55: this should probably `..GLOBAL_PUNISHMENTS,`

AnnikaCodes: yea it probably should

server/punishments.ts
```diff
@@ -13,6 +13,7 @@
 
 import {FS} from '../lib/fs';
 import {Utils} from '../lib/utils';
+import {getGlobalPunishments} from './chat-plugins/modlog';
```
Zarel, line 16: This looks like it would break if you tried to hotpatch only one of chat and punishments.

If you don't want to do a full refactor, you might want to put `getGlobalPunishments` on a global that the plugin can hotpatch.

AnnikaCodes: tbh I could just draft this until I refactor it; I don't think the functionality is so vital that it can't wait a few weeks.

AnnikaCodes:
OK, this now uses the refactored modlog.

AnnikaCodes:
@Zarel this is ready for review.

Zarel:
Sure, I'm fine with this.

PR #7076, opened 2020-07-29
Refactor ip and modchat permissions

Zarel:
- `ip` and `ipself` are now separate permissions. This means that `ip`  is now a generalized permission for viewing IPs (no more need to use `globalban` because `can('ip')` is true for all users)

- `alts` and `altsself` are now also separate (all users used to be able to `altsself`).

- `modchat`, `modchatall`, and `manageroom` are now just one `modchat` permission whose jurisdiction controls how high you can set modchat.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -600,7 +600,7 @@ exports.grouplist = [
 
 		globalban: true,
 		ban: true,
-		modchatall: true,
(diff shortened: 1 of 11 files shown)


Zarel:
This will, of course, conflict with #7058 and #6883, which will be fun.

PR #7077, opened 2020-07-29
Fix Inheritance and Shared Power

KrisXV:
This fixes an issue with Inheritance's validation and removes a crash that occurs when Shared Power is added to randomized formats.


Zarel:
Thanks!

PR #7078, opened 2020-07-29
Hosts & IPTools: Refactor to remove hardcoded ranges

AnnikaCodes:
- Refactors `config/datacenters.csv`, `config/hosts.csv`, and the hardcoded ranges in `ip-tools.ts` into two files: `config/hosts.csv`, which has info about non-proxy hosts (such as ISPs), and `config/proxies.csv`, which contains data on proxies. 
- Makes it possible to view and manage ISP ranges through the Hosts plugin.
- Adds tests for IPTools and the Hosts plugin


server/ip-tools.ts
```diff
@@ -512,14 +370,14 @@ export const IPTools = new class {
 				if (!hosts || !hosts[0]) {
 					if (ip.startsWith('50.')) {
 						resolve('comcast.net.res-nohost');
-					} else if (IPTools.checkPattern(rangeTelstra, ipNumber)) {
-						resolve('telstra.net.res-nohost');
+					} else if (ipNumber >= telstraRange.minIP && ipNumber <= telstraRange.maxIP) {
+						resolve(telstraRange.host);
```
AnnikaCodes, line 374: Is it really necessary to check the Telstra range separately (i.e. are there IPs on 101.160.0.0 - 101.191.255.255 that are not Telstra IPs and would be resolved by `dns.reverse()`)?

Zarel:
There's some uncommitted changes; how should we merge them?

server/ip-tools.ts
Zarel: I don't know what you mean?

AnnikaCodes: This is checked after all the other ranges.

AnnikaCodes:
> There's some uncommitted changes; how should we merge them?

They'd just need to be put in the new format and added to the appropriate file. If they're not too sensitive for me to look at, you can DM them to me on Discord and I'll do it for you. Otherwise, the format should be pretty apparent by looking at the CSV files.

server/ip-tools.ts
Zarel: Oh, yes, I understand now. Yes, I do believe that's why it's done this way.

Zarel:
Hm, I can just push them. That'd probably be the easiest solution there.

server/ip-tools.ts
AnnikaCodes: OK, cool, I'll keep `telstraRange` as a const.

AnnikaCodes:
> Hm, I can just push them. That'd probably be the easiest solution there.

Thanks! I added the changes to proxies.csv and rebased.

test/server/chat-plugins/hosts.js
```diff
@@ -0,0 +1,55 @@
+/**
+ * Tests for the Hosts chat plugin
```
Zarel, line 2: I wonder if any functionality that needs testing should just be moved into `ip-tools`...

Zarel: All of this is stuff that could plausibly go in `ip-tools`, to me.

AnnikaCodes: I think that the visualization function, at least, makes more sense in the chat plugin. Sorting and checking for conflicts can be moved if you want though.

Zarel:
It seems like Anubis is too hardworking for you. :(

PR #7079, opened 2020-07-29
Don't hardcrash when loading config

whalemer:
Monitor depends on config, so we can't call any of its functions while loading config.


Zarel:
Annoying.

Actually can we just put it in a `setTimeout`? I mostly just want it to appear in the server's scrollback so it's incredibly obvious that it needs to be fixed.

PR #7080, opened 2020-07-29
Fix Config error reporting

whalemer:


server/config-loader.ts
```diff
@@ -123,4 +123,11 @@ export function cacheGroupData(config: ConfigType) {
 	}
 }
 
+function reportError(msg: string) {
+	// This module generally loads before Monitor, so we put this in a setImmediate to wait for it to load.
+	// Most child processes don't have Monitor.error, but the main process should always have them, and Config
+	// errors should always be the same across processes, so this is a neat way to avoid unnecessary logging.
+	// @ts-ignore typescript doesn't like us accessing Monitor like this
+	setImmediate(() => global.Monitor?.error?.(msg));
```
Zarel, line 131: Nice! Is the second `?.` necessary, though?

```suggestion
	setImmediate(() => global.Monitor?.error(msg));
```

Would this be better?

whalemer: yeah, some child processes define `Monitor.crashlogger` but not `error` or any other functions.

Zarel:
Thanks!

PR #7081, opened 2020-07-29
Youtube: Add tests

mia-pi-git:
caught a few bugs with this so, that was nice.

### DIFF
--- a/test/server/chat-plugins/youtube.js
+++ b/test/server/chat-plugins/youtube.js
@@ -6,6 +6,7 @@
 const YoutubeInterface = require('../../../.server-dist/chat-plugins/youtube').YoutubeInterface;
 const assert = require('../../assert').strict;
 const Config = require('../../../.server-dist/config-loader').load();
+Config.nofswriting = true;
 
 describe(`Youtube features`, function () {
 	it.skip(`should correctly add channels to the database`, async function () {
--- a/test/server/chat-plugins/youtube.js
+++ b/test/server/chat-plugins/youtube.js
@@ -4,7 +4,7 @@
  */
 'use strict';
 const YoutubeInterface = require('../../../.server-dist/chat-plugins/youtube').YoutubeInterface;
-const assert = require('../../assert').strict;
+const assert = require('../../assert');
 
 describe(`Youtube features`, function () {
 	it.skip(`should correctly add channels to the database`, async function () {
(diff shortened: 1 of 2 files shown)


mia-pi-git:
Is there a better way to handle the lack of API key beyond just returning if it doesn't exist? 
I didn't consider that.

Zarel:
idk if I want active network-dependent tests. They should probably at least be skipped by default (with `it.skip` or something?)

Zarel:
The alternative is mocking, but that's probably excessive.

Zarel:
That CI failure is very weird to me. Do you know why it's trying to get `modlog_r/relationshipadvice.txt`?

mia-pi-git:
I don't - nothing i did should relate to that. 
I don't understand why it's never failed before, if this happens when creating a room for the test.

AnnikaCodes:
> I don't - nothing i did should relate to that.
> I don't understand why it's never failed before, if this happens when creating a room for the test.

I would expect `Config.nofswriting` to prevent this?

test/server/chat-plugins/youtube.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Tests for the Youtube room plugin.
+ * Written by mia-pi.
+ */
+'use strict';
+const YoutubeInterface = require('../../../.server-dist/chat-plugins/youtube').YoutubeInterface;
+const assert = require('../../assert').strict;
+const Config = require('../../../.server-dist/config-loader').load();
```
AnnikaCodes, line 8:
```suggestion
const assert = require('../../assert').strict;
```
Config is already set up in `test/main.js`, so requiring it again is unnecessary, and moreover resets `Config.nofswriting`, which caused the tests to fail.

test/server/chat-plugins/youtube.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * Tests for the Youtube room plugin.
+ * Written by mia-pi.
+ */
+'use strict';
+const YoutubeInterface = require('../../../.server-dist/chat-plugins/youtube').YoutubeInterface;
+const assert = require('../../assert').strict;
```
Zarel, line 7:
```suggestion
const assert = require('../../assert');
```

I'm surprised `.strict` exists at all, but it clearly shouldn't be used here. Our assert library is already strict.

Zarel:
Thanks!

PR #7083, opened 2020-07-29
Room-Battle: Improve API

mia-pi-git:
Merges stuff from #7005 and #7068 by @Zarel's request.

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -21,6 +21,7 @@ import * as RoomGames from "./room-game";
 
 type ChannelIndex = 0 | 1 | 2 | 3 | 4;
 type PlayerIndex = 1 | 2 | 3 | 4;
+type DataResolver = (args: string[]) => void;
 export type ChallengeType = 'rated' | 'unrated' | 'challenge' | 'tour';
 
 interface BattleRequestTracker {
@@ -491,6 +492,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 	turn: number;
 	rqid: number;
 	requestCount: number;
+	dataResolvers?: DataResolver[];
 	constructor(room: GameRoom, formatid: string, options: AnyObject) {
 		super(room);
 		const format = Dex.getFormat(formatid, true);
@@ -726,6 +728,14 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
+		case 'requestlog':
+			lines = lines.slice(1);
+			if (this.dataResolvers && this.dataResolvers.length > 0) {
+				const resolver = this.dataResolvers.shift();
+				if (resolver) resolver(lines);
+			}
+			break;
+
 		case 'update':
 			for (const line of lines.slice(1)) {
 				if (line.startsWith('|turn|')) {
@@ -741,6 +751,13 @@ export class RoomBattle extends RoomGames.RoomGame {
 			this.checkActive();
 			break;
 
+		case 'requestedteam':
+			lines = lines.slice(1);
+			if (this.dataResolvers && this.dataResolvers.length > 0) {
+				const resolver = this.dataResolvers.shift();
+				if (resolver) resolver(lines);
+			}
+			break;
 		case 'sideupdate': {
 			const slot = lines[1] as SideID;
 			const player = this[slot];
@@ -1112,6 +1129,33 @@ export class RoomBattle extends RoomGames.RoomGame {
 		// @ts-ignore
 		this.room = null;
 	}
+	async getTeam(user: User) {
+		const id = user.id;
+		const player = this.playerTable[id];
+		if (!player) return;
+		void this.stream.write(`>requestteam ${player.slot}`);
+		const teamDataPromise = new Promise<string[]>(resolve => {
+			if (!this.dataResolvers) this.dataResolvers = [];
+			this.dataResolvers.push(resolve);
+		});
+		const resultStrings = await teamDataPromise;
+		const result = resultStrings.map(item => Dex.fastUnpackTeam(item))[0];
+		return result;
+	}
+	onChatMessage(message: string, user: User) {
+		void this.stream.write(`>chat-inputlogonly ${user.getIdentity(this.room.roomid)}|${message}`);
+		return;
+	}
+	async getLog(): Promise<string[]> {
+		if (!this.logData) this.logData = {};
+		void this.stream.write('>requestlog');
+		const logPromise = new Promise<string[]>(resolve => {
+			if (!this.dataResolvers) this.dataResolvers = [];
+			this.dataResolvers.push(resolve);
+		});
+		const result = await logPromise;
+		return result;
+	}
 }
 
 export class RoomBattleStream extends BattleStream {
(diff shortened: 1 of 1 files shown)


server/room-battle.ts
```diff
@@ -1112,6 +1121,31 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+		const teamDataPromise = new Promise<string[]>(resolve => {
+			this.playerTeamResolver = resolve;
+		});
+		const resultStrings = await teamDataPromise;
+		const result = resultStrings.map(item => Dex.fastUnpackTeam(item))[0];
+		return result;
+	}
+	onChatMessage(message: string, user: User): string | false {
+		void this.stream.write(`>chat-inputlogonly ${user.getIdentity(this.room.roomid)}|${message}`);
+		return false;
```
Zarel, line 1138: tbh this API should probably be the one where "no return value" means "don't do anything".

Returning `false` definitely makes me nervous.

mia-pi-git: Oh, can i rewrite this part of the API then? I don't like how it handles anything other than strings.

server/room-battle.ts
```diff
@@ -491,6 +491,8 @@ export class RoomBattle extends RoomGames.RoomGame {
 	turn: number;
 	rqid: number;
 	requestCount: number;
+	playerTeamResolver?: (args: string[]) => void;
+	resolveLogData?: (args: string[]) => void;
```
Zarel, line 495:
```suggestion
	playerTeamResolver?: (args: string[]) => void;
	resolveLogData?: (args: string[]) => void;
```

To protect against race conditions, this should probably be a single array of requests.

server/room-battle.ts
Zarel: Hm, it depends on how hotpatchable an API change to `onChatMessage` would be. I might need to handle this myself.

mia-pi-git: Mostly should be hotpatchable? 
All the games are hotpatchable, and so is CommandContext.canTalk

Zarel: I handled it.

Zarel: (You'll need to rebase, though.)

server/room-battle.ts
```diff
@@ -726,6 +728,14 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
+		case 'requestlog':
```
Zarel, line 731: `requestlog` and `requestedteam` can be the same part of this now. Maybe call it `requesteddata`?

server/room-battle.ts
```diff
@@ -491,6 +492,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 	turn: number;
 	rqid: number;
 	requestCount: number;
+	dataResolvers?: DataResolver[];
```
Zarel, line 495:
```suggestion
	dataResolvers?: ((args: string[]) => void)[];
```

I think it's more readable to put the type definition directly here, instead of forcing people to look up the definition for `DataResolver`. Especially since the definition is only used once, and is reasonably concise.

server/room-battle.ts
```diff
@@ -726,6 +728,14 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
+		case 'requesteddata':
+			lines = lines.slice(1);
+			if (this.dataResolvers && this.dataResolvers.length > 0) {
+				const resolver = this.dataResolvers.shift();
+				if (resolver) resolver(lines);
+			}
```
Zarel, line 736:
```suggestion
			const resolver = this.dataResolvers!.shift()!;
			resolver(lines);
```

I suspect that if `resolver` doesn't exist, that's a bug that should invoke the crashlogger.

Zarel: A minor issue is: what if the battle is destroyed while a resolver is waiting? Some Promises would hang forever.

This is extremely obscure: there are only two possibilities here: a battle is destroyed after 10 minutes of activity (unlikely for a resolver to wait that long), or a battle process hard-crashes (bypassing crashlogger; this can happen from OOMs, from externally killing them, or possibly from stack overflows).

Of course I'd accept this PR without the fix, but if you have time, you might want to support rejecting the Promise with "battle destroyed" in this case.

server/room-battle.ts
```diff
@@ -1112,6 +1122,33 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+			this.dataResolvers.push(resolve);
+		});
+		const resultStrings = await teamDataPromise;
+		const result = resultStrings.map(item => Dex.fastUnpackTeam(item))[0];
+		return result;
+	}
+	onChatMessage(message: string, user: User) {
+		void this.stream.write(`>chat-inputlogonly ${user.getIdentity(this.room.roomid)}|${message}`);
+		return;
+	}
```
Zarel, line 1141:
```suggestion
	onChatMessage(message: string, user: User) {
		void this.stream.write(`>chat-inputlogonly ${user.getIdentity(this.room.roomid)}|${message}`);
	}
```

server/room-battle.ts
mia-pi-git: Will do.

server/room-battle.ts
```diff
@@ -730,7 +729,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 		switch (lines[0]) {
 		case 'requesteddata':
 			lines = lines.slice(1);
-			const resolver = this.dataResolvers!.shift()!;
+			const resolver = this.dataResolvers!.shift()![0];
```
Zarel, line 732:
```suggestion
			const [resolver] = this.dataResolvers!.shift()!;
```

server/room-battle.ts
```diff
@@ -1119,29 +1118,40 @@ export class RoomBattle extends RoomGames.RoomGame {
 
 		// @ts-ignore
 		this.room = null;
+		if (this.dataResolvers) {
+			for (const [, reject] of this.dataResolvers) {
+				// reject the promise, make whatever function called it return undefined
+				reject('Battle was destroyed.');
```
Zarel, line 1124:
```suggestion
				reject(new Error('Battle was destroyed.'));
```

server/room-battle.ts
```diff
@@ -492,7 +491,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 	turn: number;
 	rqid: number;
 	requestCount: number;
-	dataResolvers?: ((args: string[]) => void)[];
+	dataResolvers?: [((args: string[]) => void), ((args?: string) => void)][];
```
Zarel, line 494:
```suggestion
	dataResolvers?: [((args: string[]) => void), ((error: Error) => void)][];
```

server/room-battle.ts
```diff
@@ -1119,29 +1118,40 @@ export class RoomBattle extends RoomGames.RoomGame {
...
 		const player = this.playerTable[id];
 		if (!player) return;
 		void this.stream.write(`>requestteam ${player.slot}`);
-		const teamDataPromise = new Promise<string[]>(resolve => {
+		const teamDataPromise = new Promise<string[]>((resolve, reject) => {
 			if (!this.dataResolvers) this.dataResolvers = [];
-			this.dataResolvers.push(resolve);
+			this.dataResolvers.push([resolve, reject]);
+		}).catch(e => {
+			if (!e.message.includes("Battle was destroyed.")) throw e;
```
Zarel, line 1137:
```suggestion
```

I'm surprised this doesn't crash, considering you tried to reject with a string.

I suspect you might not have tested it.

Zarel: But anyway you shouldn't catch errors here at all, I think.

server/room-battle.ts
```diff
@@ -1119,29 +1118,40 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+	async getLog(): Promise<string[] | void> {
 		if (!this.logData) this.logData = {};
 		void this.stream.write('>requestlog');
-		const logPromise = new Promise<string[]>(resolve => {
+		const logPromise = new Promise<string[]>((resolve, reject) => {
 			if (!this.dataResolvers) this.dataResolvers = [];
-			this.dataResolvers.push(resolve);
+			this.dataResolvers.push([resolve, reject]);
+		}).catch(e => {
+			if (!e.message.includes("Battle was destroyed.")) throw e;
```
Zarel, line 1154:
```suggestion
```

Zarel:
Thanks!

PR #7085, opened 2020-07-29
Add short alias for /hidebattlesfromtrainercard

AnnikaCodes:
This was requested and makes sense to me, since it's quite a long command to type out each time.


Zarel:
Looks good!

PR #7088, opened 2020-07-30
Update CODEOWNERS

KrisXV:


KrisXV:
(Added @TheImmortal as a reviewer because I know he doesn't have very many opinions on chat-plugins typically)

PR #7089, opened 2020-07-30
Disallow namelocked users to challenge

asgdf:


mia-pi-git:
> _No description provided._

I looked into this a while ago along with a challenge throttle - user.locked exists on namelocked users, so this shouldn't be necessary, i don't think.

asgdf:
...it does? Appears so, but I remember a very specific instance where namelocked users were able to challenge when they shouldn't have been. I don't know what was up with that. Huh.

Zarel:
They can challenge other locked/namelocked users. I wouldn't be _too_ surprised at a bug that namelocked but didn't lock users, though...

PR #7090, opened 2020-07-30
Fix RO rank display on side servers

### LINKED ISSUE #7007: Room Owner group problem
Hello, when I install the server privately the roomowner group is below the Driver and Moderator grade. I think there is an error but I don't know where. (I haven't changed anything on my side).

Thank !

mia-pi-git:
ROs can addhtml, so it returned normal automatically, when it should have returned `leadership`.
Fixes #7007


Zarel:
Mia and I discussed several ways this might have been done. Using the `*` symbol would make it hard for servers to choose their own symbols, while using the `bot` ID might interfere with servers in other languages.

PR #7093, opened 2020-07-30
Update links to vr and sample teams for gen7ou

Alumn2000:


Zarel:
Thanks for the update!

@TheImmortal or @KrisXV, can you confirm that this is correct?

PR #7097, opened 2020-07-31
Trivia: Display question numbers

AnnikaCodes:


server/chat-plugins/trivia.ts
```diff
@@ -612,7 +612,7 @@ class Trivia extends Rooms.RoomGame {
 	 */
 	sendQuestion(question: TriviaQuestion) {
 		this.broadcast(
-			`Question: ${question.question}`,
+			`Question ${this.questionNumber}: ${question.question}`,
```
KrisXV, line 615:
```suggestion
			`Question #${this.questionNumber}: ${question.question}`,
```

AnnikaCodes: Personally, I think Question 1, Question 2, etc, is cleaner. I don't really care either way, though.

KrisXV: That's fair, i also dont really care either way

Zarel: On the other hand there's an argument for preferring "#1" over "Question 1".

PR #7099, opened 2020-07-31
Fix hidden messages

sparkychildcharlie:
Whenever there is a silent join in room, previously any other silent joins following will not display.  You'll end up with ``|J|muted|J|locked`` which not long doesn't display user "locked" as joining the room, nor will "locked" appear in the userlist.  

However, the big issue with this is: when a chat message ``room.log.add()`` is sent within 100ms of a silent join, the message will disappear. The resultant message ends up as ``|J|troll|c:|696969696969|sparkychild|I can evade filters all i want and none of you can do anything about it``.

I'm not sure if there was originally a purpose for not having \n between entries (without doing .update() in between makes a giant mess!), but this should fix any missing messages.

Shoutout to @Volco  for discovering and ranting about this mysterious phenomenon in Dawn admin chat.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -609,7 +609,7 @@ export abstract class BasicRoom {
 		}
 		entry = `|${ucType}|${entry}`;
 		if (this.batchJoins) {
-			this.log.broadcastBuffer += entry;
+			this.log.broadcastBuffer += entry + '\n';
 
 			if (!this.reportJoinsInterval) {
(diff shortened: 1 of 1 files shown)


Zarel:
`broadcastBuffer` should already have a trailing `\n`. This will need more investigation. Thanks for finding this, though!

Zarel:
LOL I misread your PR and ended up doing the exact same thing.

Sorry, that was my mistake.

PR #3859, opened 2017-08-06
Help room: Create a basic AI to answer questions

panpawn:
WIP

PR #7100, opened 2020-07-31
Add a filter for auto-answering questions in the Help room

mia-pi-git:
Approved by all of Help staff.
_Very_ loosely based on the work panpawn did in #3859, but I've since rewritten everything about it, so I'm unsure if that needs to be credited or not. Let me know, I guess.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,6 +2,7 @@ server/chat-plugins/othermetas.ts @KrisXV @TheImmortal
 server/chat-plugins/random-battles.ts @KrisXV @TheImmortal
 server/chat-plugins/scavenger*.ts @xfix @sparkychildcharlie
 server/chat-plugins/trivia.ts @AnnikaCodes
+server/chat-plugins/help.ts @mia-pi-git
 server/chat-plugins/hosts.ts @AnnikaCodes
 server/chat-plugins/mafia.ts @HoeenCoder
 data/mods/ssb/ @HoeenCoder
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -217,7 +217,8 @@ export class HelpAnswerer {
 export const Answerer = new HelpAnswerer(helpData);
 
 export const chatfilter: ChatFilter = (message, user, room) => {
-	if (room?.roomid === 'help' && room.auth.get(user.id) === ' ' && !Answerer.disabled) {
+	if (!helpRoom) return message;
+	if (room?.roomid === helpRoom.roomid && helpRoom.auth.get(user.id) === ' ' && !Answerer.disabled) {
 		const reply = Answerer.visualize(message, false, user);
 		if (message.startsWith('/') || message.startsWith('!')) return message;
 		if (!reply) {
@@ -239,6 +240,7 @@ export const chatfilter: ChatFilter = (message, user, room) => {
 
 export const commands: ChatCommands = {
 	question(target, room, user) {
+		if (!helpRoom) return this.errorReply(`There is no room configured for use of the Help filter.`);
 		if (!target) return this.parse("/help question");
 		const reply = Answerer.visualize(target, true);
 		if (!reply) return this.sendReplyBox(`No answer found.`);
@@ -250,6 +252,7 @@ export const commands: ChatCommands = {
 	hf: 'helpfilter',
 	helpfilter: {
 		''(target) {
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of the Help filter.`);
 			if (!target) {
 				this.parse('/help helpfilter');
 				return this.sendReply(`The Help auto-response filter is currently set to: ${Answerer.disabled ? 'OFF' : "ON"}`);
@@ -261,7 +264,8 @@ export const commands: ChatCommands = {
 		},
 		toggle(target, room, user) {
 			if (!room) return this.requiresRoom();
-			if (room.roomid !== 'help') return this.errorReply(`This command is only available in the Help room.`);
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
+			if (room.roomid !== helpRoom.roomid) return this.errorReply(`This command is only available in the Help room.`);
 			if (!target) {
 				return this.sendReply(`The Help auto-response filter is currently set to: ${Answerer.disabled ? 'OFF' : "ON"}`);
 			}
@@ -280,8 +284,9 @@ export const commands: ChatCommands = {
 		},
 		forceadd: 'add',
 		add(target, room, user, connection, cmd) {
-			if (room?.roomid !== 'help') return this.errorReply(`This command is only available in the Help room.`);
-			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
+			if (!room) return this.requiresRoom();
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
+			if (room.roomid !== helpRoom.roomid) return this.errorReply(`This command is only available in the Help room.`);
 			const force = cmd === 'forceadd';
 			if (force && (!REGEX_WHITELIST.includes(user.id) && !user.can('rangeban'))) {
 				return this.errorReply(`You cannot use raw regex - use /helpfilter add instead.`);
@@ -292,7 +297,7 @@ export const commands: ChatCommands = {
 			this.modlog(`HELPFILTER ADD`, null, target);
 		},
 		remove(target, room, user) {
-			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
 			if (!this.can('ban', null, helpRoom)) return false;
 			const [faq, index] = target.split(',');
 			// intended for use mainly within the page, so supports being used in all rooms
@@ -304,8 +309,9 @@ export const commands: ChatCommands = {
 			this.modlog('HELPFILTER REMOVE', null, index);
 		},
 		queue(target, room, user) {
-			if (room?.roomid !== 'help') return this.errorReply(`This command is only available in the Help room.`);
-			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
+			if (!room) return this.requiresRoom();
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
+			if (room.roomid !== helpRoom.roomid) return this.errorReply(`This command is only available in the Help room.`);
 			if (!this.can('show', null, helpRoom)) return false;
 			if (!room.auth.has(user.id)) return this.errorReply(`Only roomauth can submit regexes to the filter.`);
 			if (!target) return this.errorReply(`Specify regex.`);
@@ -321,7 +327,7 @@ export const commands: ChatCommands = {
 			return this.sendReply(`Added "${target}" to the regex suggestion queue.`);
 		},
 		approve(target, room, user) {
-			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
 			if (!this.can('ban', null, helpRoom)) return false;
 			// intended for use mainly within the page, so supports being used in all rooms
 			this.room = helpRoom;
@@ -342,7 +348,7 @@ export const commands: ChatCommands = {
 			this.modlog(`HELPFILTER APPROVE`, null, `${target}: ${str}`);
 		},
 		deny(target, room, user) {
-			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
 			if (!this.can('ban', null, helpRoom)) return false;
 			// intended for use mainly within the page, so supports being used in all rooms
 			this.room = helpRoom;
@@ -375,7 +381,7 @@ export const commands: ChatCommands = {
 
 export const pages: PageTable = {
 	helpfilter(args, user) {
-		if (!helpRoom) return `<h2>No Help room to view data for.</h2>`;
+		if (!helpRoom) return `<h2>There is no room configured to use the help filter.</h2>`;
 		const canChange = helpRoom.auth.atLeast(user, '@');
 		let buf = '';
 		const refresh = (type: string, extra?: string[]) => {
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -339,7 +339,7 @@ export const commands: ChatCommands = {
 			Answerer.writeState();
 
 			this.privateModAction(`${user.name} approved regex for use with queue number ${target}`);
-			this.modlog(`HELPFILTER APPROVE`, null, `${target}`);
+			this.modlog(`HELPFILTER APPROVE`, null, `${target}: ${str}`);
 		},
 		deny(target, room, user) {
 			if (!helpRoom) return this.errorReply(`No Help room to manage the filter for.`);
@@ -376,7 +376,7 @@ export const commands: ChatCommands = {
 export const pages: PageTable = {
 	helpfilter(args, user) {
 		if (!helpRoom) return `<h2>No Help room to view data for.</h2>`;
-		const canChange = user.authAtLeast(`@`, helpRoom);
+		const canChange = helpRoom.auth.atLeast(user, '@');
 		let buf = '';
 		const refresh = (type: string, extra?: string[]) => {
 			let button = `<button class="button" name="send" value="/join view-helpfilter-${type}`;
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -14,7 +14,6 @@ import {ROOMFAQ_FILE} from './room-faqs';
 const PATH = 'config/chat-plugins/help.json';
 // 6: filters out conveniently short aliases
 const MINIMUM_LENGTH = 6;
-const HELP_ROOM = Config.helpFilterRoom ? Rooms.get(Config.helpFilterRoom) : Rooms.get('help');
 const REGEX_WHITELIST = ['miapi', 'annika', 'kris'];
 
 export let helpData: PluginData;
@@ -79,6 +78,9 @@ export class HelpResponder {
 			// refresh on modifications to keep it up to date
 			this.roomFaqs = this.loadFaqs();
 		});
+	}
+	getRoom() {
+		return Config.helpFilterRoom ? Rooms.get(Config.helpFilterRoom) : Rooms.get('help');
 	}
 	find(question: string, user?: User) {
 		const faqs = Object.keys((this.roomFaqs || '{}'))
@@ -207,8 +209,9 @@ export class HelpResponder {
 const Answerer = new HelpResponder(helpData);
 
 export const chatfilter: ChatFilter = (message, user, room) => {
-	if (!HELP_ROOM) return message;
-	if (room?.roomid === HELP_ROOM.roomid && HELP_ROOM.auth.get(user.id) === ' ' && !Answerer.disabled) {
+	const helpRoom = Answerer.getRoom();
+	if (!helpRoom) return message;
+	if (room?.roomid === helpRoom.roomid && helpRoom.auth.get(user.id) === ' ' && !Answerer.disabled) {
 		const reply = Answerer.visualize(message, false, user);
 		if (message.startsWith('/') || message.startsWith('!')) return message;
 		if (!reply) {
@@ -230,7 +233,7 @@ export const chatfilter: ChatFilter = (message, user, room) => {
 
 export const commands: ChatCommands = {
 	question(target, room, user) {
-		if (!HELP_ROOM) return this.errorReply(`There is no room configured for use of the Help filter.`);
+		if (!Answerer.getRoom()) return this.errorReply(`There is no room configured for use of the Help filter.`);
 		if (!target) return this.parse("/help question");
 		const reply = Answerer.visualize(target, true);
 		if (!reply) return this.sendReplyBox(`No answer found.`);
@@ -242,7 +245,7 @@ export const commands: ChatCommands = {
 	hf: 'helpfilter',
 	helpfilter: {
 		''(target) {
-			if (!HELP_ROOM) return this.errorReply(`There is no room configured for use of the Help filter.`);
+			if (!Answerer.getRoom()) return this.errorReply(`There is no room configured for use of the Help filter.`);
 			if (!target) {
 				this.parse('/help helpfilter');
 				return this.sendReply(`The Help auto-response filter is currently set to: ${Answerer.disabled ? 'OFF' : "ON"}`);
@@ -254,8 +257,9 @@ export const commands: ChatCommands = {
 		},
 		toggle(target, room, user) {
 			if (!room) return this.requiresRoom();
-			if (!HELP_ROOM) return this.errorReply(`There is no room configured for use of this filter.`);
-			if (room.roomid !== HELP_ROOM.roomid) return this.errorReply(`This command is only available in the Help room.`);
+			const helpRoom = Answerer.getRoom();
+			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
+			if (room.roomid !== helpRoom.roomid) return this.errorReply(`This command is only available in the Help room.`);
 			if (!target) {
 				return this.sendReply(`The Help auto-response filter is currently set to: ${Answerer.disabled ? 'OFF' : "ON"}`);
 			}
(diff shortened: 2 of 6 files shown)


CODEOWNERS
```diff
@@ -2,6 +2,7 @@ server/chat-plugins/othermetas.ts @KrisXV @TheImmortal
 server/chat-plugins/random-battles.ts @KrisXV @TheImmortal
 server/chat-plugins/scavenger*.ts @xfix @sparkychildcharlie
 server/chat-plugins/trivia.ts @AnnikaCodes
+server/chat-plugins/help.ts @mia-pi-git
 server/chat-plugins/hosts.ts @AnnikaCodes
```
AnnikaCodes, line 6: CODEOWNERS should probably be alphabetized by file name?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,469 @@
...
+	}
+	visualize(question: string, hideButton?: boolean, user?: User) {
+		const response = this.find(question, user);
+		if (response) {
+			let buf = '';
+			buf += Utils.html`<strong>You said:</strong> ${question}<br />`;
+			buf += `<strong>Our automated reply:</strong> ${Chat.formatText(response)}`;
+			question = Utils.escapeHTML(question);
+			if (!hideButton) {
+				buf += `<hr /><button class="button" name="send" value="A: ${question}">`;
```
AnnikaCodes, line 108:
```suggestion
			if (!hideButton) {
				buf += Utils.html`<hr /><button class="button" name="send" value="A: ${question}">`;
```
You should be able to use `Utils.html` here.

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,469 @@
...
+		if (!alias) return;
+		// ignore short aliases, they cause too many false positives
+		if (faq.length < MINIMUM_LENGTH || alias.length < MINIMUM_LENGTH) return;
+		if (alias.charAt(0) !== '>') return faq; // not an alias
+		return alias.replace('>', '');
+	}
+	stringRegex(str: string, raw?: boolean) {
+		str = str.split('=>')[0];
+		const args = str.split(',').map(item => item.trim());
+		// if (args.length < 2) throw new Chat.ErrorMessage("You need more than 2 arguments.");
```
AnnikaCodes, line 127:
```suggestion
```

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,469 @@
...
+		const day = Chat.toTimestamp(new Date).split(' ')[0];
+		if (!this.data.stats[day]) this.data.stats[day] = {};
+		const today = this.data.stats[day];
+		const log: LoggedMessage = {
+			message: entry,
+			faqName: faq,
+			regex: expression,
+		};
+		const stats = {
+			matches: today.matches ? today.matches : [],
```
AnnikaCodes, line 172:
```suggestion
			matches: today.matches || [],
```

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,469 @@
...
+			const regex = Answerer.stringRegex(str);
+			// validated on submission
+			const faq = Answerer.getFaqID(str.split('=>')[1].trim());
+			if (!faq) return this.errorReply(`Invalid FAQ.`);
+			if (!Answerer.data.pairs[faq]) helpData.pairs[faq] = [];
+			Answerer.data.pairs[faq].push(regex);
+			Answerer.queue.splice(index, 1);
+			Answerer.writeState();
+
+			this.privateModAction(`${user.name} approved regex for use with queue number ${target}`);
```
AnnikaCodes, line 341: I feel like logging the regex itself might also be worthwhile.

mia-pi-git: I mean, it's modlogged already

AnnikaCodes: Oh, i misread that

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,467 @@
...
+			`Requires: @ # &`,
+		];
+		return this.sendReplyBox(help.join('<br/ >'));
+	},
+};
+
+export const pages: PageTable = {
+	helpfilter(args, user) {
+		if (!helpRoom) return `<h2>No Help room to view data for.</h2>`;
+		const canChange = user.authAtLeast(`@`, helpRoom);
```
AnnikaCodes, line 379:
```suggestion
		const canChange = helpRoom.auth.atLeast(user, `@`);
```
User#authAtLeast is being removed on next restart in https://github.com/smogon/pokemon-showdown/pull/7046.

test/server/chat-plugins/help.js
```diff
@@ -0,0 +1,52 @@
...
+	stats: {},
+	pairs: {},
+	disabled: false,
+	queue: [],
+};
+
+describe('Help', function () {
+	it('should only return true on added regexes', function () {
+		const Answerer = require('../../../.server-dist/chat-plugins/help').HelpAnswerer;
+		const Help = new Answerer(defaultData);
```
AnnikaCodes, line 20: It's probably better to do `let Help;` at the top and then set this up in a `before()`.

mia-pi-git: Honestly, they all use effectivley the same instance, it could probably just be one constant at the top.

AnnikaCodes: Yes, exactly. (I assumed you didn't do that because something needed `Config`, which is what the let; workaround is for -- if it works to just put it in a const, that'd be best.)

Zarel:
I don't see anything obviously wrong with this, and it seems reasonably self-contained. I'll give it a day for others to review, though.

CODEOWNERS
AnnikaCodes: Alternatively, we could alphabetize by GitHub username, but this does neither.

scheibo:
Isn't a concern here that this is relatively main-specific? 

If thats not a concern, can we merge the private Help room related chat filter code in here as well?

Zarel:
Oh, sure. I don't have a problem with putting Main-specific code in the regular repo. We can have a Config setting to turn it off by default if side servers actually complain.

scheibo:
> Oh, sure. I don't have a problem with putting Main-specific code in the regular repo. We can have a Config setting to turn it off by default if side servers actually complain.

Should `/notdev` be in the main codebase? Chatfilters for the chinese/language rooms? Lagtest filtering?

I'm all about minimizing as much private code as possible, this just seems like a departure from what has previously been the case? We've been reluctant in the past to check things like this in (or else why would they been in private code to begin with), and often make sure to avoid encoding too much of main's conventions into the public code (like how our global auth structure works)

mia-pi-git:
This is also fairly easy to disable if you turn the ``helpRoom`` file-wide const to `undefined`, or a different room, so it could potentially have use on other servers, if that helps?
I'll push something to ensure the rest obeys the ``helpRoom`` const instead of the ``room.roomid !== 'help'`` check.

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+ */
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {LogViewer} from './chatlog';
+import {ROOMFAQ_FILE} from './room-faqs';
+const PATH = 'config/chat-plugins/help.json';
+const MINIMUM_LENGTH = 7;
+const helpRoom = Rooms.get('help');
+const REGEX_WHITELIST = ['miapi', 'annika', 'kris'];
```
scheibo, line 16: @Zarel 

This is the main thing I would like to talk about. This is ugly AF in code we're checking in, in the same way it was ugly AF to check in the anubis hardcode in the datacenters code. I agree that adding a parallel whitelist command (and the necessarily persistence of the whitelist etc) is overkill, but adding a *general* whitelist that all commands can use is something I think is valuable:

`/whitelist helpfilter miapi` or `/whitelist datacenters anubis` etc. You then only need to set up the whitelisting infrastructure one and be able to apply it to all commands. Yes, this does further decouple the auth symbol from the powers that comes with it, but is clearly something that does come up quite often.

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+ * Uses Regex to match room frequently asked question (RFAQ) entries,
+ * and replies if a match is found.
+ * Supports configuration.
+ * Written by mia-pi.
+ */
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {LogViewer} from './chatlog';
+import {ROOMFAQ_FILE} from './room-faqs';
```
scheibo, line 12:
```suggestion
import {ROOMFAQ_FILE} from './room-faqs';

```

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+const PATH = 'config/chat-plugins/help.json';
+const MINIMUM_LENGTH = 7;
+const helpRoom = Rooms.get('help');
```
scheibo, line 15: this constant not being CONSTANT_CASE while the others are is odd to me. This should also be configurable?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+		stats: {},
+		pairs: {},
+		disabled: false,
+		queue: [],
+	};
+}
+/**
+ * Terms commonly used in helping that should be ignored
+ * within question parsing (the Help#find function).
+ * */
```
scheibo, line 33:
```suggestion
 */
```

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+	stats?: {[k: string]: DayStats};
+	/** Word pairs that have been marked as a match for a specific FAQ. */
+	pairs: {[k: string]: string[]};
+	/** Whether or not the filter is disabled. */
+	disabled?: boolean;
+	/** Queue of suggested regex. */
+	queue?: string[];
+}
+
+export class HelpAnswerer {
```
scheibo, line 66: `HelpResponder` seems like it would be desirable compared to `HelpAnswerer`, though see my note above about this just being `AutoResponder` and not specific to `Help`

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
+/**
+ * PS Help room auto-response plugin.
```
scheibo, line 2: At a higher level, why is this a 'Help' room plugin instead of being a generic `AutoResponder` plugin that can be used in any room? Obviously it is most desirable in main's PS main's Help room, but other rooms (for example PS main's Development room) might want similar features, and coding it generically as an `AutoResponder` removes my concerns about main-specific code being checked in?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+		this.queue = data.queue ? data.queue : [];
+
+		FS(ROOMFAQ_FILE).onModify(() => {
+			// refresh on modifications to keep it up to date
+			this.roomFaqs = this.loadFaqs();
+		});
+	}
+	find(question: string, user?: User) {
+		const faqs: string[] = Object.keys((this.roomFaqs || '{}'))
+			.filter(item => item.length >= 6 && !this.roomFaqs[item].startsWith('>'));
```
scheibo, line 83: where does the number 6 come from?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+
+export class HelpAnswerer {
+	roomFaqs: AnyObject;
+	disabled?: boolean;
+	queue: string[];
+	data: PluginData;
+	constructor(data: PluginData) {
+		this.data = data;
+		this.roomFaqs = this.loadFaqs();
+		this.queue = data.queue ? data.queue : [];
```
scheibo, line 74:
```suggestion
		this.queue = data.queue || [];
```

(or `?? []`, but PS doesnt like `??`)

server/chat-plugins/help.ts
scheibo: is the `string[]` actually required, i would assume it is inferrable from `Object.keys`?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+		if (!this.data.pairs) this.data.pairs = {};
+		if (!this.data.pairs[faq]) throw new Chat.ErrorMessage(`There are no regexes for ${faq}.`);
+		if (!this.data.pairs[faq][index]) throw new Chat.ErrorMessage("Your provided index is invalid.");
+		this.data.pairs[faq].splice(index, 1);
+		this.writeState();
+		return true;
+	}
+}
+
+export const Answerer = new HelpAnswerer(helpData);
```
scheibo, line 217: Why is this exported?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+			.filter(item => item.length >= 6 && !this.roomFaqs[item].startsWith('>'));
+		for (const term of COMMON_TERMS) {
+			if (term.test(question)) return null;
+		}
+		for (const faq of faqs) {
+			const match = this.match(question, faq);
+			if (match) {
+				if (user) {
+					const timestamp = Chat.toTimestamp(new Date()).split(' ')[1];
+					const log = `${timestamp} |c| ${user.name}|${question}`;
```
scheibo, line 92: are these spaces correct? also, why is the users rank elided?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+			// refresh on modifications to keep it up to date
+			this.roomFaqs = this.loadFaqs();
+		});
+	}
+	find(question: string, user?: User) {
+		const faqs: string[] = Object.keys((this.roomFaqs || '{}'))
+			.filter(item => item.length >= 6 && !this.roomFaqs[item].startsWith('>'));
+		for (const term of COMMON_TERMS) {
+			if (term.test(question)) return null;
+		}
```
scheibo, line 86: Optionally

```suggestion
		if (COMMON_TERMS.some(t => t.test(question)) return null;
```

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+ * Supports configuration.
+ * Written by mia-pi.
+ */
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {LogViewer} from './chatlog';
+import {ROOMFAQ_FILE} from './room-faqs';
+const PATH = 'config/chat-plugins/help.json';
+const MINIMUM_LENGTH = 7;
```
scheibo, line 14: Please add a comment as to why 7 was chosen

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+		if (!faq) return;
+		const alias: string = this.roomFaqs[faq];
+		if (!alias) return;
+		// ignore short aliases, they cause too many false positives
+		if (faq.length < MINIMUM_LENGTH || alias.length < MINIMUM_LENGTH) return;
+		if (alias.charAt(0) !== '>') return faq; // not an alias
+		return alias.replace('>', '');
+	}
+	stringRegex(str: string, raw?: boolean) {
+		str = str.split('=>')[0];
```
scheibo, line 125: seems like `splitFirst` was literally designed for this?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+				// escape otherwise
+				return string.replace(/[\\^$.*+?()[\]{}]/g, '\\$&').trim();
+			});
+			return split.map(term => {
+				if (term.startsWith('!')) {
+					return `^(?!.*${term.slice(1)})`;
+				}
+				if (!term.trim()) return null;
+				return `(?=.*?(${term.trim()}))`;
+			}).filter(Boolean).join('');
```
scheibo, line 140: clever, ive been using `x => x`, which churns closures

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+		}).filter(Boolean).join('');
+	}
+	validateRegex(regex: string) {
```
scheibo, line 143: does similar code already exist in other commands which accept regexes? (`/filter`? `/banwords`?). can it be consolidated?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+			if (e.message.includes("regular expression")) throw new Chat.ErrorMessage(e.message);
+			throw e;
+		}
+		return true;
+	}
+	match(question: string, faq: string) {
+		const regexes = this.data.pairs[faq];
+		if (!regexes) return;
+		for (const regex of regexes) {
+			const regexString = new RegExp(regex, "i");
```
scheibo, line 156: huh? what makes this a "string"? Also, why churn `RegExp` objects here, why cant `new RegExp(regex, "i")` simply be sorted in `regexes` to begin with? I think below its so that they can be compared? I feel `regex.toString()` is faster than building RegExps each time you match?

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,473 @@
...
+			if (regexString.test(Chat.normalize(question))) return {faq, regex};
+		}
+		return;
+	}
+	log(entry: string, faq: string, expression: string) {
+		if (!this.data.stats) this.data.stats = {};
+		const day = Chat.toTimestamp(new Date).split(' ')[0];
```
scheibo, line 163: `splitFirst`?

test/server/chat-plugins/help.js
```diff
@@ -0,0 +1,48 @@
...
+
+describe('Help', function () {
+	before(function () {
+		const defaultData = {
+			stats: {},
+			pairs: {},
+			disabled: false,
+			queue: [],
+		};
+		const Answerer = require('../../../.server-dist/chat-plugins/help').HelpAnswerer;
```
scheibo, line 18: why is this not a top level require? you should not need to require this before each test?

server/chat-plugins/help.ts
mia-pi-git: I mean it's fairly easy to configure, you just change the Rooms.get(); 
I can make it an actual config setting but that seems.. annoying.

server/chat-plugins/help.ts
mia-pi-git: I mean, no other room has use for it. I can make it generic, i guess, but that frankly doesn't seem useful.
Also, memory note, storing debug data has use for the help room but doing it for every room would be too much, i think.

server/chat-plugins/help.ts
mia-pi-git: Spaces are intentional. This only will activate on regular users, so i saw no point in adding the rank.

server/chat-plugins/help.ts
mia-pi-git: It does. I'll push that into a Chat function.

test/server/chat-plugins/help.js
AnnikaCodes: It requires `Config`, which isn't defined at the top level -- other tests for chatplugins that require `Config` use similar workarounds. (Also, it's not required before *each* test, just before the test set for Help is run.)

server/chat-plugins/help.ts
mia-pi-git: Eval purposes, mostly, didn't remember to remove it.

test/server/chat-plugins/help.js
AnnikaCodes: OK, that's actually quite easy to fix.

AnnikaCodes: This workaround will be unnecessary when #7109 is merged.

server/chat-plugins/help.ts
```diff
@@ -0,0 +1,463 @@
...
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {LogViewer} from './chatlog';
+import {ROOMFAQ_FILE} from './room-faqs';
+
+const PATH = 'config/chat-plugins/help.json';
+// 6: filters out conveniently short aliases
+const MINIMUM_LENGTH = 6;
+const HELP_ROOM = Config.helpFilterRoom ? Rooms.get(Config.helpFilterRoom) : Rooms.get('help');
```
Slayer95, line 17: This is a persistent reference to the ``Room`` instance that would prevent it from GC if the room is deleted.
This is also the reason why this plugin cannot be required from tests.

It'd Just reference the ``Config`` global from the relevant functions.

test/server/chat-plugins/help.js
Slayer95: I've just checked: in fact, there aren't any other chat plugins which assume a global ``Config`` exists. So I'd change that...

server/chat-plugins/help.ts
Zarel: I really really don't like whitelisting in general, but I guess if it's going to be done anyway...

In this specific case: what if we just used "roomdriver in Help and Dev" as the whitelist? Maybe the Annika whitelist could also just be "roommod in Dev".

server/chat-plugins/help.ts
Zarel: I don't consider it a constant. It's certainly not immutable.

Zarel: Like I could see:

```ts
const HELP_ROOMID = 'help';
const helpRoom = Rooms.get(HELP_ROOMID);
```

server/chat-plugins/help.ts
AnnikaCodes: Roommod in Dev seems rather broad to be granting access to IP range and host management.

server/chat-plugins/help.ts
Zarel: I don't think it's wrong for things like this to be exported. It is, in fact, useful for monkeypatching.

server/chat-plugins/help.ts
Zarel: No, `splitFirst` was designed for if you needed to split a string like `"foo,bar,baz"` into `"foo"` and `"bar,baz"`. For this sort of use-case, you could go with `str.split('=>', 1)[0]` if you cared about performance, but I think this is fine as-is.

server/chat-plugins/help.ts
Zarel: PS has been using `filter(Boolean)` for a while; I guess you've stopped reading our diffs recently. :p

image

server/chat-plugins/help.ts
Zarel: I might prefer

```ts
[str] = str.split('=>');
```

for readability, though. But I don't know if that's just me.

server/chat-plugins/help.ts
Zarel: no :(

server/chat-plugins/help.ts
mia-pi-git:
```ts
	[str] = Utils.splitFirst(str, '=>');```
is what i ended up doing, fwiw.

File: server/chat.ts [hunk 27]
```diff
@@ -1703,6 +1703,21 @@ export const Chat = new class {
 		if (!htmlContent) return '';
 		return htmlContent.replace(/<[^>]*>/g, '');
 	}
+	/**
+	 * Validates input regex and ensures it won't crash.
+	 */
+	validateRegex(word: string) {
```
Zarel, line 1709: I wonder if we want to also check if the regex starts/ends with `|`, because that's definitely a mistake.

mia-pi-git: Seems useful. Did it.

server/chat-plugins/help.ts
scheibo: I think there are very real reasons we might want to consider given Anubis access to more commands then just datacenters as well...

server/chat-plugins/help.ts
scheibo: I don't feel this is resolved :)

@Zarel PTAL

server/chat-plugins/help.ts
scheibo: I almost never read anything in chat-plugins, this should not be surprising

server/chat-plugins/help.ts
scheibo: Exports define your API. It doesnt make sense to me that this would be part of an API? Furthermore, why are you monkeypatching chat plugins when theyre already hotpatchable, am I missing something?

server/chat-plugins/help.ts
Zarel: I feel like anyone who wants to make it generic is free to do so, but I'm not going to make Mia do it, and especially not blocking it from being useful in Help.

mia-pi-git: I feel making it generic, due to the nature of some of the constants, would either make it risker and less useful in Help, or less useful in other rooms. I'm personally not a fan.

mia-pi-git: I'd be open to giving it some weeks of live testing and then seeing how things go, and then coming back to making it general, but for now, i feel it'd probably inhibit it in help to make it generalized right now.

server/chat-plugins/help.ts
Zarel: Gmod + roommod in dev, then? I don't think it's too broad, I'd definitely trust anyone with that specific permission combination.

Zarel: I think if Anubis wants to be a special snowflake I will grant it based on all she's contributed to PS, but the hardcode also reflects my extreme distaste for the practice in general and is at least partially intended to discourage others from trying to get at that.

scheibo: Can you elaborate on your concerns over whitelisting? You're literally already allowing whitelists in at least 5 places that I'm aware of with hardcoded lists of users ids (datacenters, here, pmlogs, sysops, console)

CODEOWNERS
Zarel: Filename makes more sense to me.

server/chat-plugins/help.ts
Zarel: I suppose I don't see a problem with whitelisting specific admin users. I don't like whitelists below that, though, because the rank system should ideally already cover it.

PR #7101, opened 2020-07-31
Fix /renamegroupchat's permission checking

mia-pi-git:
Currently it just checks if they can rename the room. This intentionally allows users with mute perms to rename so that both the host and gstaff can, in case the user names it something that also bypasses the filter.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -914,7 +914,7 @@ export const commands: ChatCommands = {
 		if (room.persist && isGroupchat) return this.errorReply(`This isn't a groupchat.`);
 		if (!room.persist && !isGroupchat) return this.errorReply(`Use /renamegroupchat instead.`);
 		if (isGroupchat) {
-			if (!user.can('mute) && !this.can('editroom', null, room)) return false;
+			if (!user.can('mute') && !this.can('editroom', null, room)) return false;
 			const existingRoom = Rooms.search(toID(target));
 			if (existingRoom && !existingRoom.settings.modjoin) {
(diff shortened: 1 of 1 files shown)


server/chat-commands/room-settings.ts
```diff
@@ -915,6 +914,7 @@ export const commands: ChatCommands = {
 		if (room.persist && isGroupchat) return this.errorReply(`This isn't a groupchat.`);
 		if (!room.persist && !isGroupchat) return this.errorReply(`Use /renamegroupchat instead.`);
 		if (isGroupchat) {
+			if (!this.can('modchat', null, room)) return false;
```
Zarel, line 917: I mean, this is a very elegant way of making sure both Hosts and global Drivers can rename rooms, but I'm not sure I like this.

```suggestion
			if (!user.can('mute) && !this.can('editroom', null, room)) return false;
```

server/chat-commands/room-settings.ts
```diff
@@ -915,6 +914,7 @@ export const commands: ChatCommands = {
 		if (room.persist && isGroupchat) return this.errorReply(`This isn't a groupchat.`);
 		if (!room.persist && !isGroupchat) return this.errorReply(`Use /renamegroupchat instead.`);
 		if (isGroupchat) {
+			if (!user.can('mute', null, room) && !this.can('editroom', null, room)) return false;
```
Zarel, line 917:
```suggestion
			if (!user.can('lock') && !this.can('editroom', null, room)) return false;
```

Probably don't give roomdrivers the ability, if it comes up?

Zarel:
Thanks!

PR #7102, opened 2020-07-31
Announcements: Make strings translatable

AnnikaCodes:
Unfortunately, I don't speak any of the languages PS! is translated into well enough to actually do any translation, but I did handle the technical side.

One minor snag (which I want to discuss either here or on Discord) is that `Chat#plural` only correctly creates a plural form for English, and not other languages. Personally, I think if we are to be truly multilingual, assuming that words are always plural is probably simpler (albeit less clean-looking) than trying to programmatically generate plural forms for every language PS is translated into or gets a translation for in the future.

### DIFF
--- a/translations/dutch.json
+++ b/translations/dutch.json
@@ -131,6 +131,19 @@
 		"You are now blocking all incoming challenge requests.": "Je blokkeert nu alle inkomende uitdagingen.",
 		"You are already blocking challenges!": "Je blokkeert al uitdagingen!",
 		"You are already available for challenges!": "Je bent al beschikbaar voor uitdagingen!",
-		"You are available for challenges from now on.": "Je bent vanaf nu beschikbaar voor uitdagingen."
+		"You are available for challenges from now on.": "Je bent vanaf nu beschikbaar voor uitdagingen.",
+
+		"The announcement has ended.": "",
+		"Battles do not support announcements.": "",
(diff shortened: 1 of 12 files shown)


translations/dutch.json
```diff
@@ -131,6 +131,19 @@
...
 		"You are already blocking challenges!": "Je blokkeert al uitdagingen!",
 		"You are already available for challenges!": "Je bent al beschikbaar voor uitdagingen!",
-		"You are available for challenges from now on.": "Je bent vanaf nu beschikbaar voor uitdagingen."
+		"You are available for challenges from now on.": "Je bent vanaf nu beschikbaar voor uitdagingen.",
+
+		"The announcement has ended.": "",
+		"Battles do not support announcements.": "",
+		"You are not allowed to use filtered words in announcements.": "",
+		"There is already a poll or announcement in progress in this room.": "",
+		"An announcement was started by ${user.name}.": "",
```
Zarel, line 140: Is this how translations are usually implemented?

Zarel: Looks like yes. Interesting.

Zarel:
Looks like a straight improvement, so sure.

translations/dutch.json
AnnikaCodes: Yeah, I try to make any variables in template strings I'm going to translate descriptive enough that translators can understand with minimal technical knowledge.

PR #7104, opened 2020-08-01
Fix target leak of two-turn charge moves in Doubles

DaWoblefet:
Because -prepare protocol currently sends target with two-turn moves, the server currently leaks the target of two-turn moves like Phantom Force, etc. in doubles play in the console. This removes all instances of -prepare sending the target, except for Sky Drop, where the target should be known on the first turn of the move.

Thanks to PlatypusVGC for telling me about the info leak and @Marty-D for pointing me in the right direction to fix it! Marty asked me to check to make sure nothing breaks on client from this change and nothing does.


Zarel:
Thanks!

PR #7105, opened 2020-08-01
Scavengers: general improvements

sparkychildcharlie:
- Add new command to directly stuff a hunt back into the queue
- Better queue response
- Fix extra spaces in teamscav team names
- Get rid of some old code tidbits that were neccessary when roomlogs were still bork


Zarel:
tyty

PR #7107, opened 2020-08-01
Ensure committed csv files have \n line endings

whalemer:
we split these on `\n`, so if git changes the endings to `\r\n` every line has a trailing `\r`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+config/*.csv eol=lf


Zarel:
I really wish we could split by line without caring what line endings were. :(

AnnikaCodes:
https://github.com/smogon/pokemon-showdown/pull/7108 removes any leftover `\r`s when the hosts and proxies CSV files are read.

Zarel:
Should we close this because it's obsoleted by #7108? I think I'd prefer that approach.

whalemer:
I like having this, but sure, it's not totally necessary with that.

Zarel:
I think you can globally set Git to use non-horrible line endings in Windows? But we probably shouldn't manually dictate that if it's not necessary.

AnnikaCodes:
Is this an issue for other CSV files? If so, we should probably add logic to strip carriage returns when reading those, as well.

Zarel:
PS should probably support both somehow. Maybe a `Utils.splitByLine` function?

PR #7108, opened 2020-08-01
IPTools: Remove '-nohost' suffix and make minor improvements

AnnikaCodes:


scheibo:
I think this needs to be paired with a private code change that removes the hardcoded `-nohost` strings there as well

AnnikaCodes:
> I think this needs to be paired with a private code change that removes the hardcoded `-nohost` strings there as well

Yes, it does - I believe I mentioned that in your DMs on Discord but forgot to put it in the PR description.

Zarel:
Maybe the delimiter should be something that isn't a valid domain? Something like `:` or `,` or `|` or `/`.

Zarel:
The `nohost` suffix was previously used to mean "we filled this in from `datacenters.csv` instead of it being the real rDNS". Do you think that's worth preserving in some way?

AnnikaCodes:
> Maybe the delimiter should be something that isn't a valid domain? Something like `:` or `,` or `|` or `/`.

Personally I think `/` would work best, but any of those would work (though I dont want to use `,` since it's already the delimiter we use for CSV files),.

> The `nohost` suffix was previously used to mean "we filled this in from `datacenters.csv` instead of it being the real rDNS". Do you think that's worth preserving in some way?

It'd be easy enough to preserve, but I personally don't see a reason to (of course, I can't IP check people so I'd defer to those who actually use IP tools).

Zarel:
I think it might be worth one character, so `/proxy` for real rdns and `?/proxy` for "from hosts/proxies.csv". That's definitely useful to me just for debugging purposes.

AnnikaCodes:
> Well, I like this. Did you guys want to do something with the private code stuff, though?

Yes, if there are references to the `(hostname).(tld).(type)-nohost` format in private code they will need to be updated to the new `?/proxy` format.

AnnikaCodes:
For some reason GitHub doesn't let me fork private repositories, so instead of a PR to fix the private code, I have a diff I can send you - want me to PM it to you on Discord?

PR #7109, opened 2020-08-01
Tests: Support requiring Config at the top level

AnnikaCodes:
This solves issues like https://github.com/smogon/pokemon-showdown/pull/7100#discussion_r463922425 and makes it easier to test chat plugins.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -5,6 +5,50 @@ const fs = require('fs');
 
 const noop = () => {};
 
+// Load and override configuration before starting the server
+let config;
+try {
+	require.resolve('../config/config');
+} catch (err) {
+	if (err.code !== 'MODULE_NOT_FOUND' && err.code !== 'ENOENT') throw err; // Should never happen
+
+	console.log("config.js doesn't exist - creating one with default settings...");
+	fs.writeFileSync(
+		path.resolve(__dirname, '../config/config.js'),
+		fs.readFileSync(path.resolve(__dirname, '../config/config-example.js'))
+	);
+} finally {
+	config = require('../config/config');
+}
+Object.assign(config, require('../config/config-example'));
+
+// Actually crash if we crash
+config.crashguard = false;
+// Don't allow config to be overridden while test is running
+config.watchconfig = false;
+// Don't try to write to file system
+config.nofswriting = true;
+// Test a normal ladder
+config.fakeladder = false;
+
+require('./../.lib-dist/process-manager').disabled = true;
+
+// stop chatrooms from loading through modifying the require cache
+try {
+	const chatrooms = require('../config/chatrooms.json');
+	chatrooms.splice(0, chatrooms.length);
+} catch (e) {}
+
+// Don't create a REPL
+require('../.lib-dist/repl').Repl.start = noop;
+
+// Start the server.
+// NOTE: This used "server" before when we needed ".server-dist"
+require('../.server-dist');
+
+LoginServer.disabled = true;
+Ladders.disabled = true;
+
 before('initialization', function () {
 	this.timeout(0); // Remove timeout limitation
 	process.on('unhandledRejection', err => {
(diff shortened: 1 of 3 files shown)


test/server/chat-plugins/trivia.js
```diff
@@ -28,16 +28,6 @@ function destroyUser(user) {
 
 describe('Trivia', function () {
 	before(function () {
-		// The trivia module cannot be loaded outside of this scope because
-		// it makes reference to global.Config in the modules outermost scope,
-		// which makes the module fail to be loaded. Within the scope of thess
-		// unit test blocks however, Config is defined.
```
Slayer95, line 34: This comment is outdated, and therefore the changes to this file don't require changes to ``test/main.js``

test/server/chat-plugins/hosts.js
```diff
@@ -6,13 +6,9 @@
 'use strict';
 
 const assert = require('assert').strict;
-let hosts;
+const hosts = require('../../../.server-dist/chat-plugins/hosts');
```
Slayer95, line 9: I'm fairly sure this change can also be done without polluting tests global scope with ``Config``.

test/server/chat-plugins/trivia.js
AnnikaCodes: It isn't outdated - classes in Trivia that extend Rooms.RoomGame and Rooms.RoomGamePlayer cause it to crash if loaded at the top level.

test/server/chat-plugins/hosts.js
AnnikaCodes: Yes, a try/except makes that simple:
```ts
process.nextTick(() => {
	// Wrap this in a try/except because Chat isn't defined in unit tests
	try {
		Chat.multiLinePattern.register('/(datacenters|datacenter|dc|iprange|ipranges) (add|widen|remove)');
	} catch (e) {
		if (!(e instanceof ReferenceError)) throw e;
	}
});
```

Slayer95:
So the crux of this PR is ``require('../.server-dist');`` which lets ``hosts`` have ``Chat`` and ``trivia`` have ``Rooms``.
I didn't remember we did that when loading tests, but now I understand. And yea, when loading the full server we get ``Config`` as a global too.

Thanks for the explanation. I am afraid I am getting more and more rusty >,<

test/server/chat-plugins/trivia.js
Slayer95: It's outdated because ``it makes reference to global.Config in the modules outermost scope`` is no longer true, but my comment above is a non-sequitur.

Zarel:
I'm still confused about exactly what the problem is here.

_Is_ `Config` being defined at the global scope? As far as I can tell, this PR doesn't change that.

I would presume it was already defined at the global scope, though, because the server expects it, and we have tests of some server functionality.

Zarel:
I would assume that `Config` should be required and already available globally at the time chat plugins and tests are loaded. Is that not the case?

AnnikaCodes:
Globals are defined within the scope of tests (i.e. in the functions supplied to `describe()`), but not at the top level of a test file (necessitating `before()` workarounds for things like Trivia, which can't be `require`d without the global `Rooms`).

Zarel:
What does this PR change about that?

test/main.js
```diff
@@ -5,55 +5,55 @@ const fs = require('fs');
...
+	const chatrooms = require('../config/chatrooms.json');
+	chatrooms.splice(0, chatrooms.length);
+} catch (e) {}
+
+// Don't create a REPL
+require('../.lib-dist/repl').Repl.start = noop;
+
+// Start the server.
+// NOTE: This used "server" before when we needed ".server-dist"
+require('../.server-dist');
```
AnnikaCodes, line 47: This line, for example, sets up the `Rooms` global so it's accessible at the top level of test files.

Zarel: As far as I can tell, you just moved this up from line 54.

AnnikaCodes: Yes; prior to my change, it was in a `before()` function, so `Rooms` wouldn't be accessible at the top level of test files (only within `before`/`after`/`it`/other Mocha functions).

Zarel: Oh! That makes sense.

PR #7110, opened 2020-08-01
Polls: Make strings translatable

AnnikaCodes:
Same idea as #7102. 
<sub><sup><sub>~~free Swedish translation so I don't have a bunch of empty JSON~~</sub></sup></sub>


Zarel:
tyty

PR #7112, opened 2020-08-01
Detach 2v2 banlist from doubles tiering

TheMezStrikes:
as per https://www.smogon.com/forums/threads/2v2-doubles.3656321/post-8553113


TheMezStrikes:
https://www.smogon.com/forums/threads/2v2-doubles.3656321/post-8285330

According to OP, Beat Up was supposed to be on the banlist but wasn't

TheMezStrikes:
2v2 TL Mubs informed me that it wasn't supposed to be banned but was erroneously noted as banned on 2v2 OP. He'll make a post clarifying soonish that I'll link

PR #7114, opened 2020-08-01
Trivia: Add Infinite length

AnnikaCodes:


Zarel:
Are you set to CODEOWNERS for Trivia? At this point, you probably should be.

AnnikaCodes:
> Are you set to CODEOWNERS for Trivia? At this point, you probably should be.

Yes, I am (although it doesn't really have a point other than marking me as its maintainer, since I don't get automatically requested for review).

PR #7116, opened 2020-08-02
Fix /lottery participants from escaping container

DaWoblefet:
Before [from this bug report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8474970)
image

After:
image


Zarel:
Thanks!

PR #7117, opened 2020-08-02
Display G-Max moves properly in /details

AnnikaCodes:
Fixes this https://www.smogon.com/forums/threads/show-g-max-pokemons-signature-moves-in-their-dt-descriptions.3668136/.

### DIFF
--- a/data/pokedex.ts
+++ b/data/pokedex.ts
@@ -162,6 +162,7 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 0,
 		color: "Red",
 		eggGroups: ["Monster", "Dragon"],
+		changesFrom: "Charizard",
 	},
 	squirtle: {
 		num: 7,
@@ -236,6 +237,7 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 0,
 		color: "Blue",
 		eggGroups: ["Monster", "Water 1"],
+		changesFrom: "Blastoise",
 	},
 	caterpie: {
 		num: 10,
@@ -289,6 +291,7 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 0,
 		color: "White",
 		eggGroups: ["Bug"],
+		changesFrom: "Butterfree",
 	},
 	weedle: {
 		num: 13,
@@ -769,6 +772,7 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 0,
 		color: "Yellow",
 		eggGroups: ["Field", "Fairy"],
+		changesFrom: "Pikachu",
 	},
 	pikachuworld: {
 		num: 25,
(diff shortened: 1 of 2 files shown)


server/chat-commands/info.ts
```diff
@@ -582,7 +582,10 @@ export const commands: ChatCommands = {
 						Height: `${pokemon.heightm} m`,
 					};
 					details["Weight"] = `${pokemon.weighthg / 10} kg <em>(${weighthit} BP)</em>`;
-					if (pokemon.canGigantamax) details["G-Max Move"] = pokemon.canGigantamax;
+					if (pokemon.forme === 'Gmax') {
+						const gmaxMove = dex.getSpecies(pokemon.baseSpecies).canGigantamax;
+						if (gmaxMove) details["G-Max Move"] = gmaxMove;
+					}
```
KrisXV, line 588: I think they should still display on the base form as well, and this also doesn't work for formes like Urshifu-Rapid-Strike-Gmax.

AnnikaCodes:
```ts
if (pokemon.forme === 'Gmax') {
	const gmaxMove = dex.getSpecies(pokemon.name.replace('-Gmax', '')).canGigantamax;
	if (gmaxMove) details["G-Max Move"] = gmaxMove;
}``` This is sort of hacky (but does what you want); is there a way to programmatically get the specific form a Gmax mon transformed from?

Comment (TheImmortal) [reply server/chat-commands/info.ts:588]: I don't see why they should be on the base.

``changesFrom`` is what you're looking for.

Comment (AnnikaCodes) [reply server/chat-commands/info.ts:588]: I thought `changesFrom` was what I wanted originally, but it turns out it's undefined for Gigantamaxes. (I originally used `baseSpecies` as per this comment https://github.com/smogon/pokemon-showdown/blob/master/sim/dex-data.ts#L699.)

Comment (Zarel) [reply server/chat-commands/info.ts:588]: `baseSpecies` should basically never be used. The linked comment is outdated; I believe `changesFrom` now works correctly for megas/primals.

Comment (AnnikaCodes) [reply server/chat-commands/info.ts:588]: It doesn't work for Gigantamax formes, though. 
image

Comment (Zarel) [reply server/chat-commands/info.ts:588]: We might want to fix that.

**Comment by Zarel** [comment]
Drafting. I agree with Kris that this should appear on both the base and gmax formes.

**Comment by AnnikaCodes** [comment]
OK, this seems to implement the desired functionality.

File: data/pokedex.ts [hunk 2]
```diff
@@ -162,6 +162,7 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 0,
 		color: "Red",
 		eggGroups: ["Monster", "Dragon"],
+		changesFrom: "Charizard",
```
Zarel, line 165: These are supposed to be filled in automatically from `getSpecies`, right? Or did we change that?

KrisXV: We changed it

Zarel:
I don't know how to feel about the new `changesFrom` stuff, but I guess that's no reason to block this.

PR #7121, opened 2020-08-02
Searchlog: fix crash

mia-pi-git:
Ripgrep doesn't debug well, given the only error it gives is `` Error: Command failed``, which it can throw for legit reasons like no results. Given that, the best solution is to handle it like modlog and just return an empty result array. 
It also checks `e.stdout` since sometimes it errors but still stores the results there. 
If anyone has any better ideas, i'm open, but given the awful nature of child_process.execFile's errors, i think this is the best handling.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -545,7 +545,7 @@ export const LogSearcher = new class {
 			});
 			results = stdout.split('--');
 		} catch (e) {
-			if (e.code === 2) throw e; // 2 means an error in ripgrep
+			if (e.code !== 1) throw e; // 2 means an error in ripgrep
 			if (e.stdout) {
 				results = e.stdout.split('--');
 			} else {
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -545,12 +545,12 @@ export const LogSearcher = new class {
 			});
 			results = stdout.split('--');
 		} catch (e) {
-			if (e.stdout || e.message.includes('stdout maxBuffer')) {
+			if (e.code === 2) throw e; // 2 means an error in ripgrep
+			if (e.stdout) {
 				results = e.stdout.split('--');
-				count += results.length;
-				return {results, count};
+			} else {
+				results = [];
 			}
-			throw e;
 		}
 		count += results.length;
 		return {results, count};


Zarel:
In theory you can find the specific result code ripgrep uses for "no results found" and handle that separately.

server/chat-plugins/chatlog.ts
```diff
@@ -545,12 +545,12 @@ export const LogSearcher = new class {
 			});
 			results = stdout.split('--');
 		} catch (e) {
-			if (e.stdout || e.message.includes('stdout maxBuffer')) {
+			if (e.code === 2) throw e; // 2 means an error in ripgrep
```
Zarel, line 548: Is 2 the "no results found" error code? Shouldn't it be the other way around?

mia-pi-git: 2 is "there is an error with ripgrep that _isn't_ no results", which should be thrown.
Otherwise, e.stdout might exist, which is handled, or there will be no results - also handled.

Zarel: What's the error code if there _is_ no result? That's the one that should be special-cased.

mia-pi-git: no result is `1`, but it seems like it would be easier to just handle the actual error and then check stdout, since that also covers things like the weird issues with execFile that we had last time?

Zarel: `if (e.code !== 1)` seems safer.

As usual: It's much better to crash if you don't know what's going on, than to not crash and have a bug that's very difficult to track down.

scheibo:
I think I'd like to see a command which can be used to trivially disable another command and which lasts through restarts and hotpatching. We've had to disable searchlogs numerous times at this point and have had multiple crashes due to it not being properly disabled on occasion.

mia-pi-git:
> I think I'd like to see a command which can be used to trivially disable another command and which lasts through restarts and hotpatching. We've had to disable searchlogs numerous times at this point and have had multiple crashes due to it not being properly disabled on occasion.

That would be feasible. I presume I should open another PR for that? Or would it be worth pushing in here, since it's kind of a "fix" for /sl?

scheibo:
Yeah sorry, that was me just commenting on a further change I came up with that I'd like to see before I'd be ready to enable `/searchlogs` again given its proven troublesome. I think it should go in its own PR :)

Zarel:
I don't see a reason to disable the command.

Zarel:
As mentioned earlier: Better that a crash is known than that it silently fails.

scheibo:
Um, we've disabled the command 3-4 times already, with various consequences (leaks, constant OOMs). You dont need to see the same crash multiple times to debug it. If anything, commands which were involved in a crash should be auto disabled.

Zarel:
Disabling automatically would definitely cause some major chaos if something like `/join` crashes.

I think it's easy enough to quickly fix a command if it's causing trouble. The repeated crashes are also a signal that the fix should be prioritized.

scheibo:
lol

PR #7122, opened 2020-08-02
Support forcing a crash to be private with configurable terms

mia-pi-git:
due to my recent, um, mishap, pre thought this would be a good idea and it was relatively easy to do.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1423,7 +1423,7 @@ export class GlobalRoomState {
 		let hasPrivateTerm = stack.includes('private');
 
 		for (const term of (Config.privatecrashterms || [])) {
-			if (typeof term === 'string' ? err.stack.includes(term) : term.test(err.stack)) {
+			if (typeof term === 'string' ? stack.includes(term) : term.test(err.stack)) {
 				hasPrivateTerm = true;
 				break;
 			}
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1423,7 +1423,6 @@ export class GlobalRoomState {
 		let hasPrivateTerm = stack.includes('private');
 
 		const privateTerms = (Config.privateCrashTerms ? [...Config.privateCrashTerms] : []).map(item => new RegExp(item, "i"));
-		console.log(privateTerms);
 		for (const term of privateTerms) {
 			if (term.test(stack) || term.test(err.message)) hasPrivateTerm = true;
 		}
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1412,9 +1412,8 @@ export class GlobalRoomState {
 			return;
 		}
 		this.lastReportedCrash = time;
-		// @ts-ignore
-		const stackLines = (err ? Utils.escapeHTML(err.stack).split(`\n`) : []);
-		const stack = stackLines.slice(1).join(`<br />`);
+		const stack = err.stack || err.message || err.name || '';
+		const stackLines = (err ? Utils.escapeHTML(stack).split(`\n`) : []);
 
 		let crashMessage = `|html|<div class="broadcast-red"><details class="readmore"><summary><b>${crasher} crashed:</b> ${stackLines[0]}</summary>${stack}</details></div>`;
 		let privateCrashMessage = null;
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1422,9 +1422,11 @@ export class GlobalRoomState {
 		const upperStaffRoom = Rooms.get('upperstaff');
 		let hasPrivateTerm = stack.includes('private');
 
-		const privateTerms = (Config.privatecrashterms ? [...Config.privatecrashterms] : []).map(item => new RegExp(item, "i"));
-		for (const term of privateTerms) {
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -1420,7 +1420,14 @@ export class GlobalRoomState {
 		let privateCrashMessage = null;
 
 		const upperStaffRoom = Rooms.get('upperstaff');
-		if (stack.includes("private")) {
+		let hasPrivateTerm = stack.includes('private');
+
+		const privateTerms = (Config.privateCrashTerms ? [...Config.privateCrashTerms] : []).map(item => new RegExp(item, "i"));
+		console.log(privateTerms);
```
AnnikaCodes, line 1426:
```suggestion
```

Zarel:
I don't know what this is for. What sort of crash are you trying to force to be private?

server/rooms.ts
```diff
@@ -1420,7 +1420,13 @@ export class GlobalRoomState {
 		let privateCrashMessage = null;
 
 		const upperStaffRoom = Rooms.get('upperstaff');
-		if (stack.includes("private")) {
+		let hasPrivateTerm = stack.includes('private');
+
+		const privateTerms = (Config.privateCrashTerms ? [...Config.privateCrashTerms] : []).map(item => new RegExp(item, "i"));
```
KrisXV, line 1425: dont we typically have all letters for config variables lowercased?

server/rooms.ts
```diff
@@ -1420,7 +1420,13 @@ export class GlobalRoomState {
 		let privateCrashMessage = null;
 
 		const upperStaffRoom = Rooms.get('upperstaff');
-		if (stack.includes("private")) {
+		let hasPrivateTerm = stack.includes('private');
+
+		const privateTerms = (Config.privatecrashterms ? [...Config.privatecrashterms] : []).map(item => new RegExp(item, "i"));
+		for (const term of privateTerms) {
```
Zarel, line 1426:
```suggestion
		for (const term of (Config.privatecrashterms || [])) {
```

I'm not sure why you're converting `Config.privatecrashterms` to an array. What are you anticipating it being?

server/rooms.ts
```diff
@@ -1420,7 +1420,13 @@ export class GlobalRoomState {
...
+			if (term.test(stack) || term.test(err.message)) hasPrivateTerm = true;
```
Zarel, line 1427:
```suggestion
			if (typeof term === 'string' ? err.stack.includes(term) : term.test(err.stack)) {
				hasPrivateTerm = true;
				break;
			}
```

This will support both string matching and regex matching.

server/rooms.ts
```diff
@@ -1422,9 +1422,11 @@ export class GlobalRoomState {
 		const upperStaffRoom = Rooms.get('upperstaff');
 		let hasPrivateTerm = stack.includes('private');
 
-		const privateTerms = (Config.privatecrashterms ? [...Config.privatecrashterms] : []).map(item => new RegExp(item, "i"));
-		for (const term of privateTerms) {
-			if (term.test(stack) || term.test(err.message)) hasPrivateTerm = true;
+		for (const term of (Config.privatecrashterms || [])) {
+			if (typeof term === 'string' ? err.stack.includes(term) : term.test(err.stack)) {
```
Zarel, line 1426: I don't know why GH Actions is having trouble displaying the correct error message, but I managed to get it by running tests on my own computer:

```
server/rooms.ts:1426:35 - error TS2532: Object is possibly 'undefined'.

1426  		if (typeof term === 'string' ? err.stack.includes(term) : term.test(err.stack)) {
      		                               ~~~~~~~~~
```

```suggestion
		for (const term of (Config.privatecrashterms || [])) {
			if (typeof term === 'string' ? stack.includes(term) : term.test(stack)) {
```

server/rooms.ts
```diff
@@ -1422,9 +1422,11 @@ export class GlobalRoomState {
 		const upperStaffRoom = Rooms.get('upperstaff');
```
Zarel, line 1422: I sadly can't suggest further up, but line 1416 should be:

```
const stack = err.stack || err.message || err.name || '';
const stackLines = (err ? Utils.escapeHTML(stack).split(`\n`) : []);
```

Also `@ts-ignore` is no longer necessary.

server/rooms.ts
```diff
@@ -1412,15 +1412,22 @@ export class GlobalRoomState {
...
-		if (stack.includes("private")) {
+		let hasPrivateTerm = stack.includes('private');
+
+		for (const term of (Config.privatecrashterms || [])) {
+			if (typeof term === 'string' ? stack.includes(term) : term.test(stack)) {
+				hasPrivateTerm = true;
+				break;
+			}
+		}
```
Zarel, line 1429:
```suggestion

		let hasPrivateTerm = stack.includes('private');
		for (const term of (Config.privatecrashterms || [])) {
			if (typeof term === 'string' ? stack.includes(term) : term.test(stack)) {
				hasPrivateTerm = true;
				break;
			}
		}

```

PR #7123, opened 2020-08-02
Fix renaming groupchats and groupchat creation

Volco:
This allows you to create groupchats again which #6965 entirely broke, @mia-pi-git i suggest you test your code thoroughly in the future because it's apparent you do not, and also patches renaming rooms to no name, for the second time since I already fixed this prior. Again mia, please test your code thoroughly.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -741,7 +741,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(`Your group chat name is too similar to existing chat room '${title}'.`);
 		}
 		// Room IDs for groupchats are groupchat-TITLEID
-		let titleid = toID(title);
+		const titleid = toID(title);
 		if (!titleid) {
 			title = `${Math.floor(Math.random() * 100000000)}` as ID;
 		}
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -743,7 +743,7 @@ export const commands: ChatCommands = {
 		// Room IDs for groupchats are groupchat-TITLEID
 		let titleid = toID(title);
 		if (!titleid) {
-			titleid = `${Math.floor(Math.random() * 100000000)}` as ID;
+			title = `${Math.floor(Math.random() * 100000000)}` as ID;
 		}
 		const roomid = `groupchat-${parent || user.id}-${titleid}` as RoomID;
 		// Titles must be unique.
@@ -911,6 +911,7 @@ export const commands: ChatCommands = {
 		}
 		const oldTitle = room.title;
 		const isGroupchat = cmd === 'renamegroupchat';
+		if (!toID(target)) return this.errorReply("Rooms need a title.");
 		if (room.persist && isGroupchat) return this.errorReply(`This isn't a groupchat.`);
 		if (!room.persist && !isGroupchat) return this.errorReply(`Use /renamegroupchat instead.`);
 		if (isGroupchat) {
@@ -921,6 +922,12 @@ export const commands: ChatCommands = {
 			}
 			if (this.filter(target) !== target) {
 				return this.errorReply("Invalid title.");
+			}
+			// `,` is a delimiter used by a lot of /commands
+			// `|` and `[` are delimiters used by the protocol
+			// `-` has special meaning in roomids
+			if (target.includes(',') || target.includes('|') || target.includes('[') || target.includes('-')) {
+				return this.errorReply("Room titles can't contain any of: ,|[-");
 			}
 			target = `[G] ${target}`;
 		} else {
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -741,9 +741,9 @@ export const commands: ChatCommands = {
 			return this.errorReply(`Your group chat name is too similar to existing chat room '${title}'.`);
 		}
 		// Room IDs for groupchats are groupchat-TITLEID
-		const titleid = toID(title);
+		let titleid = toID(title);
 		if (!titleid) {
-			title = `${Math.floor(Math.random() * 100000000)}` as ID;
+			titleid = `${Math.floor(Math.random() * 100000000)}` as ID;
 		}
 		const roomid = `groupchat-${parent || user.id}-${titleid}` as RoomID;
 		// Titles must be unique.
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -672,7 +672,8 @@ export abstract class BasicRoom {
 		// `|` and `[` are delimiters used by the protocol
 		// `-` has special meaning in roomids
 		// groupchats use `[` so allow it to bypass this and run the checks in the commands themselves
-		if (this.persist && (newTitle.includes(',') || newTitle.includes('|') || newTitle.includes('[') || newTitle.includes('-'))) {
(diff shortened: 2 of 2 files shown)


KrisXV:
Can you not be an asshole?

mia-pi-git:
Seconded, also wasn't this already fixed on the main server? @Zarel?

AnnikaCodes:
> Can you not be an asshole?

Thirded, please be respectful while developing for PS! :)

KrisXV:
@mia-pi-git i believe Zarel only quickpatched this through eval and didnt actually commit the code, so the most recent chat hotpatch reverted it, though I could be mistaken

Volco:
it breaks on restart, only reason you aren't having that issue right now is no restart, i updated less than an hour ago then did this

Volco:
i do apologize for being rude. i'm just a bit annoyed at the fact of having fixed a bug only for it to be unfixed, at that point it's like "why bother fixing bugs"?

KrisXV:
You can simply re-fix the bug and not be condescending towards other contributors btw.

Zarel:
I fixed it in uncommitted code on the main server. I haven't committed it yet because it requires rethinking how to validate titles.

Zarel:
Anyway, I don't believe this is the same bug as happened earlier. This is a new bug that happened because I decided to move some title validation code inside `rooms.ts` so it could be shared; and in the process revealed that what I thought were title requirements don't actually apply to all titles, a thing I wasn't aware of at the time.

So in the end, this is my mistake; no need to yell at Mia for it.

Volco:
fair enough, given how many bugs i've seen from prs committed by mia i thought it was mia who caused it

server/chat-commands/room-settings.ts
```diff
@@ -741,9 +741,9 @@ export const commands: ChatCommands = {
 			return this.errorReply(`Your group chat name is too similar to existing chat room '${title}'.`);
 		}
 		// Room IDs for groupchats are groupchat-TITLEID
-		let titleid = toID(title);
+		const titleid = toID(title);
 		if (!titleid) {
-			titleid = `${Math.floor(Math.random() * 100000000)}` as ID;
+			title = `${Math.floor(Math.random() * 100000000)}` as ID;
```
Zarel, line 746: This change seems wrong to me. If the title is "こんにちは", the title should be unchanged and a random-number title should be generated.

```suggestion
			titleid = `${Math.floor(Math.random() * 100000000)}` as ID;
			if (!title) title = titleid;
```

Volco: i didn't realize that was intended so i'll undo that then

Zarel:
> fair enough, given how many bugs i've seen from prs committed by mia i thought it was mia who caused it

A good attitude to have is that a bug is usually not any specific person's fault. All changes should be reviewed by a number of people, and if no one caught it, that means everyone is equally culpable, and what should be blamed is probably more the process as a whole, than any specific person's fault.

I do get a bit more annoyed at people when they push controversial changes _without_ making a pull request out of them, because it makes those bugs harder to catch. But that obviously isn't what happened here: Mia doesn't have the ability to do direct pushes.

server/rooms.ts
```diff
@@ -671,7 +671,9 @@ export abstract class BasicRoom {
 		// `,` is a delimiter used by a lot of /commands
 		// `|` and `[` are delimiters used by the protocol
 		// `-` has special meaning in roomids
-		if (newTitle.includes(',') || newTitle.includes('|') || newTitle.includes('[') || newTitle.includes('-')) {
+		// groupchats use `[` so allow it to bypass this and run the checks in the commands themselves
+		if (this.persist &&
+			(newTitle.includes(',') || newTitle.includes('|') || newTitle.includes('[') || newTitle.includes('-'))) {
```
Zarel, line 676: This sorta subverts the thing I was going for earlier. I'll try to find the currently running version on Main for this.

Zarel:
```ts
function(newTitle, newID) {
	if (!newID) newID = toID(newTitle);
	// `,` is a delimiter used by a lot of /commands
	// `|` and `[` are delimiters used by the protocol
	// `-` has special meaning in roomids
	if (newTitle.includes(',') || newTitle.includes('|')) {
		throw new Chat.ErrorMessage(`Room title "${newTitle}" can't contain any of: ,|`);
	}
	if ((!newID.includes('-') || newID.startsWith('groupchat-')) && newTitle.includes('-')) {
		throw new Chat.ErrorMessage(`Room title "${newTitle}" can't contain -`);
	}
	if (newID.length > 225) throw new Chat.ErrorMessage("The given room title is too long.");
	if (Rooms.search(newTitle)) throw new Chat.ErrorMessage(`The room '${newTitle}' already exists.`);
}
```

is what's currently running in Main.

Volco: why not add that to the repo then while running it on main at the same time? main has a lot of forks and if any of them update the errors then affect them too.

Zarel: I was going to, but it was late and it needed to be cleaned up better. Feel free to add it to your PR, or I can just push it if you won't.

Volco: okay, i just need 15 minutes then i'll edit it

server/chat-commands/room-settings.ts
```diff
@@ -922,6 +923,12 @@ export const commands: ChatCommands = {
 			if (this.filter(target) !== target) {
 				return this.errorReply("Invalid title.");
 			}
+			// `,` is a delimiter used by a lot of /commands
+			// `|` and `[` are delimiters used by the protocol
+			// `-` has special meaning in roomids
+			if (target.includes(',') || target.includes('|') || target.includes('[') || target.includes('-')) {
+				return this.errorReply("Room titles can't contain any of: ,|[-");
+			}
```
Zarel, line 931:
```suggestion
```

This restriction isn't enforced in `/creategroupchat` anyway. I don't think it's particularly a big deal.

AnnikaCodes: Apologies if I'm misreading this, but it does enforce that? I can't currently create groupchats with those special characters in their titles on main.

scheibo:
> > fair enough, given how many bugs i've seen from prs committed by mia i thought it was mia who caused it
> 
> A good attitude to have is that a bug is usually not any specific person's fault. All changes should be reviewed by a number of people, and if no one caught it, that means everyone is equally culpable, and what should be blamed is probably more the process as a whole, than any specific person's fault.
> 
> I do get a bit more annoyed at people when they push controversial changes _without_ making a pull request out of them, because it makes those bugs harder to catch. But that obviously isn't what happened here: Mia doesn't have the ability to do direct pushes.

What Zarel says here is exactly correct, but I'd also like to add on that the developers pushing the most code at any point in time are also likely to be pushing the most bugs. When I was submitting a flurry of PRs last year I was similarly causing a lot of crashes. We can work to improve testing infrastructure, make crashes less costly, etc, but ultimately bugs are always going to happen.

server/chat-commands/room-settings.ts
Zarel: Most of the checks here are redundant with `validateTitle`. The main restriction that's been lifted is that `[` is no longer banned.

Zarel: Wait, apparently that's still there. Ugh, I guess this will need a separate fix.

PR #7124, opened 2020-08-02
Inheritance: Fix validation issue

KrisXV:
There's currently an issue in validation where sets will not validate unless the Pokemon has a nickname, and since locked users can't nickname Pokemon, they're unable to play the format.

nicknames get reassigned later in the validation process, so this is just to make the validator be quiet.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1049,6 +1049,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 				if (validSources.includes(evoFamily)) continue;
 
 				set.species = donorSpecies.name;
+				set.name = donorSpecies.baseSpecies;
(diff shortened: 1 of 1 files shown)


config/formats.ts
```diff
@@ -1049,6 +1049,7 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 				if (validSources.includes(evoFamily)) continue;
 
 				set.species = donorSpecies.name;
+				set.name = donorSpecies.baseSpecies;
```
Zarel, line 1052:
```suggestion
				if (!set.name) set.name = donorSpecies.baseSpecies;
```

This seems safer to me?

Zarel:
tyty

PR #7126, opened 2020-08-02
Announcements: Fix crash

AnnikaCodes:


Zarel:
Why would this crash?

Zarel:
It looks like something is seriously wrong with `Chat.tr` if that causes a crash...

Zarel:
Oh, I see, `Chat.tr` is completely different from `this.tr`. That's pretty bad function naming tbh.

Issue #6469, opened 2020-03-18
Migrating Modlog to Postgres


thejetou:
## Context

The current Modlog setup of using text files and ripgrep to meet the search criteria is sluggish and has been the result of numerous complaints.

### Solution

Databases! It has shown to work in PM logs, from taking ~5 minutes to being instantaneous. It is obviously not a 'silver bullet' though, like everything it requires optimizations and monitoring to ensure it's meeting requirements.

### Why Postgres?

While SQLite is easier to set up (it's just a file), I think Postgres is a better choice for a couple reasons

- The server will  use the same database engine as the client (see: https://github.com/smogon/pokemon-showdown-client/issues/1423)
- SQLite is meant to be a 'lightweight' database, intended for use in embedded devices (or mobile phones), so scaling Postgres higher should in theory be easier.

## Proposed Schema

```sql
CREATE TABLE modlog (
    -- UNIX timestamp
    -- Integer technically works here, but would break on 2038
    timestamp bigint NOT NULL
    room_id TEXT NOT NULL
    -- ROOMBAN, NOTE etc.
    action TEXT NOT NULL
    user_id TEXT
    autoconfirmed TEXT
    alts TEXT[],
    ip TEXT,
    logged_by TEXT,
    note TEXT
);

-- We use `WHERE roomid = $1` _every_ single time, so an index seems worth it
CREATE INDEX ml_index ON modlog(room_id)
```

## Proposed Query

```sql
SELECT *
FROM modlog
WHERE
    room_id = $1
    AND
    concat_ws(" ", time_stamp, action, user_id, autoconfirmed, array_to_string(autoconfirmed, " "), alts, ip, logged_by, note) LIKE $2
ORDER BY timestamp DESC
LIMIT $3
```

Of course, concatenation could in theory be expensive and thus make this query slow, but I don't think making premature optimization is worth it. If it _does_ turn out to be too slow, we could rewrite the modlog command such that it takes a filter for each query and we don't have to concatenate all of them.

## Conversion

Create a `tools/database-converters` where we support converting modlog from the old format (with the script @sparkychildcharlie wrote) to the new format, converting the new one to Postgres, and vice versa **only for the latter**.

---

\cc @Zarel @scheibo

Zarel:
[...]

PR #7127, opened 2020-08-02
Refactor Modlog into one file

AnnikaCodes:
This will make it easier to implement changes to modlog, such as #7074 and #6469.

I'm not sure if battlesearch should stay with the modlog. I find it pretty illogical to group them together in the same file, since they do fairly different things, but it would reduce the number of subprocesses (and they do both search logs, although by that logic you could argue for putting `/searchlog` in modlog.ts as well).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /config/*
 !/config/formats.ts
 /logs/*
+/test/modlogs
 /node_modules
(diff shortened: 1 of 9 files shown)


Zarel:
Moment of truth.

I... think this is a step in the right direction, I have nitpicks but I don't think they're urgent.

PR #7128, opened 2020-08-02
Remove /offlinewhois

AnnikaCodes:
Per my discussion with @DaWoblefet on PS, those commands were deprecated and aren't needed by staff anymore.


Zarel:
What's wrong with it? I still use it and don't want it to be gone.

AnnikaCodes:
After discussing this in the Development room, it seems this PR a misunderstanding -- `/offlinewhois` and `/chp` are fallbacks, and users are intended to use `/alts`, so there isn't a need to remove `/offlinewhois`.

PR #7129, opened 2020-08-02
Room Settings: Don't display ' ' as a modchat option

AnnikaCodes:
We already have an off button that's hardcoded.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -1475,7 +1475,7 @@ export const roomSettings: SettingsHandler[] = [
 			'off',
 			'autoconfirmed',
 			'trusted',
-			...RANKS.filter(symbol => Users.Auth.hasPermission(user, 'modchat', symbol, room)),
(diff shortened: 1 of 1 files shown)


server/chat-commands/room-settings.ts
```diff
@@ -1475,7 +1475,7 @@ export const roomSettings: SettingsHandler[] = [
 			'off',
 			'autoconfirmed',
 			'trusted',
-			...RANKS.filter(symbol => Users.Auth.hasPermission(user, 'modchat', symbol, room)),
+			...RANKS.filter(symbol => symbol !== ' ' && Users.Auth.hasPermission(user, 'modchat', symbol, room)),
```
Zarel, line 1478:
```suggestion
			...RANKS.filter(symbol => symbol !== Users.Auth.defaultSymbol() && Users.Auth.hasPermission(user, 'modchat', symbol, room)),
```

Zarel:
(The `slice` technique also solves the issue where the default symbol isn't a space. I originally considered replacing `' '` with `Auth.defaultSymbol()` but then I realized `Auth.defaultSymbol` just returns `RANKS[0]` anyway.)

PR #7130, opened 2020-08-02
Gen 3-4: Fix Camouflage to change to Normal-type

DaWoblefet:
Custom behavior for Gen 5 was being inherited by 3-4. I also bothered to test to make sure Camouflage really does change the user to be Ground-type in Gen 5 Link Battles on cart and it does, which is really stupid. It's Normal-type in Gen 4 and 6-7.

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -231,7 +231,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		onHit(target) {
 			// includes() is used here over hasType() because for whatever reason Arceus-Water
 			// always returns Normal-type from hasType() which is stupid
-			if (target.types.includes('Normal') || !target.setType('Normal')) return false;
+			if (target.hasType('Normal') || !target.setType('Normal')) return false;
 			this.add('-start', target, 'typechange', 'Normal');
 		},
 	},
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -229,8 +229,6 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		desc: "The user's type changes based on the battle terrain. Normal type on the regular Wi-Fi terrain. Fails if the user has the Multitype Ability or if the type is one of the user's current types.",
 		shortDesc: "Changes user's type based on terrain. (Normal)",
 		onHit(target) {
-			// includes() is used here over hasType() because for whatever reason Arceus-Water
-			// always returns Normal-type from hasType() which is stupid
 			if (target.hasType('Normal') || !target.setType('Normal')) return false;
 			this.add('-start', target, 'typechange', 'Normal');
 		},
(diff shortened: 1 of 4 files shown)


Marty-D:
This will be easier to notice when all the descriptions are next to each other, but an important difference in Gens 3-4 is it fails if even _one_ of your types is already that type (like Color Change). Gen 4 also has the difference that Multitype is the thing that prevents type changes, and not Arceus itself as in Gen 5+.

Zarel:
Wait, should that be changed, then?

Zarel:
Like `if (target.hasType('Normal')) return false;`

Zarel:
Or is that already implemented in `setType`?

DaWoblefet:
Yeah, I'll just add the fix to this PR

DaWoblefet:
@Marty-D I've updated this PR with the mechanics you mentioned (they have associated passing tests which you can review). Of note, I am assuming that all affects which would call setType() in Gen 4 should fail on Multitype rather than Arceus (it seems that way from other effects like Conversion 2 but if this is wrong, stop me now :P)

data/mods/gen4/moves.ts
```diff
@@ -229,7 +229,9 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		desc: "The user's type changes based on the battle terrain. Normal type on the regular Wi-Fi terrain. Fails if the user has the Multitype Ability or if the type is one of the user's current types.",
 		shortDesc: "Changes user's type based on terrain. (Normal)",
 		onHit(target) {
-			if (!target.setType('Normal')) return false;
+			// includes() is used here over hasType() because for whatever reason Arceus-Water
```
Zarel, line 232: It seems like that should be fixed there probably :S

DaWoblefet: Yeah that'd be fine with me, it's just that I don't know where it's originating the bug from - hasType() calls another function getTypes and it is not obvious at all to me where anything Arceus-related is in here

``getTypes(excludeAdded?: boolean): string[] {
		const types = this.battle.runEvent('Type', this, null, null, this.types);
		if (!excludeAdded && this.addedType) return types.concat(this.addedType);
		if (types.length) return types;
		return [this.battle.gen >= 5 ? 'Normal' : '???'];
	}``

So I had given up originally since I didn't want to start making changes in more broadly called functions without knowing what was going wrong.

Marty-D:
I was going to deal with setType() later but you figured it out already, nice!

Re: getTypes(), even in Gen 4 Arceus is always Normal type if it doesn't also have Multitype and a Plate, the only difference is you just can't bluff say, Water but actually be Fire. In Gen 4 if Arceus has Multitype and a Plate it instantly transforms to the correct sprite on switch-in (plus, if it's the Arceus-??? sprite with not-Multitype, it transforms to Normal anyway and regains Multitype).

So if you wanted to, the way you could add a test for Camouflage Arceus in Gen 4 is to have Conversion/Camouflage/some non-Normal move, use Conversion first, then use Camouflage.

DaWoblefet:
Conversion didn't help. Here's the order of what's happening:

1. Conversion sets type to exclusively Fire-type (this is verified by checking the Pokemon's typing directly)
2. Camouflage checks if the Pokemon has Normal-type and returns true

So _something_ here is causing hasType('Normal'), as well as .getTypes().includes('Normal') to return true.

Marty-D:
Okay _now_ your test should work, sorry. I forgot the type override in conditions.ts was super aggressive and not appropriate for Gen 4 at all. Changed it to match the "Transforming into Arceus but holding a different Plate" behaviour, but on every switch-in like the game does it.

Marty-D:
Woo! Thanks for the tests 👍

Zarel:
Good work, team!

PR #7134, opened 2020-08-03
Move Flapple/Thievul to PUBL, Gourgeist-XL to PU

KrisXV:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/#post-8555593


Zarel:
tyty

PR #7136, opened 2020-08-03
Improve help message formatting for /searchlogs

AnnikaCodes:


Zarel:
Whoops, I think I missed this when it was PRed. Thanks!

PR #7137, opened 2020-08-03
Chat: Make broadcasted sendReplyBoxes hidetextable

mia-pi-git:
See https://www.smogon.com/forums/threads/allow-images-displayed-with-show-to-be-hidden-by-hidetext-and-similar-commands.3668239/. Should also be useful to make some other commands (see: /ds, etc) that are really spammy, hidetextable.


Zarel:
Does this work in PMs?

mia-pi-git:
> Does this work in PMs?

It does.

Zarel:
Neat, thanks!

We should probably look into how we send messages to users, but this should work for now.

PR #7138, opened 2020-08-03
Add a 'bst' property to Species

KrisXV:


sim/dex-data.ts
```diff
@@ -756,6 +758,7 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 		this.requiredItem = data.requiredItem || undefined;
 		this.requiredItems = this.requiredItems || (this.requiredItem ? [this.requiredItem] : undefined);
 		this.baseStats = data.baseStats || {hp: 0, atk: 0, def: 0, spa: 0, spd: 0, spe: 0};
+		this.bst = Object.values(this.baseStats).reduce((x, y) => x + y);
```
Zarel, line 761: This is an inner loop, so I'd rewrite this for performance.

```suggestion
		this.bst = this.baseStats.hp + this.baseStats.atk + this.baseStats.def + this.baseStats.spa + this.baseStats.spd + this.baseStats.spe;
```

KrisXV: Should I do this for the other instances where I reassign?

Zarel: Reassign? I didn't see any other instances, but yes, sure.

KrisXV: I might've used the wrong word. In some of the chat commands I assign a new BST to the cloned Species.

server/chat.ts
```diff
@@ -1908,8 +1908,8 @@ export const Chat = new class {
 		}
 	}
 
-	getDataPokemonHTML(species: Species, gen = 7, tier = '') {
-		if (typeof species === 'string') species = Dex.deepClone(Dex.getSpecies(species));
+	getDataPokemonHTML(species: Species, gen = 8, tier = '') {
+		species = Dex.deepClone(species);
```
Zarel, line 1912: Wait, why deepclone?

server/chat.ts
```diff
@@ -1940,10 +1940,7 @@ export const Chat = new class {
 			}
 			buf += '</span>';
 		}
-		let bst = 0;
-		for (const baseStat of Object.values(species.baseStats)) {
-			bst += baseStat;
-		}
+		let bst = species.bst;
```
Zarel, line 1943: You can remove this entirely, this is a kind of silly alias. I'm also surprised that `prefer-const` isn't yelling at you.

KrisXV: It isn't yelling at me since it gets modified later into the code

KrisXV: https://github.com/smogon/pokemon-showdown/pull/7138/files#diff-14924f4ee018df5f23596233f35205d6R1949

Zarel: That should probably also be implemented in dex-data instead?

KrisXV: There's no way to check what the dex's gen is in dex-data

data/mods/gen7/random-teams.ts
```diff
@@ -985,7 +985,8 @@ export class RandomGen7Teams extends RandomTeams {
 			// If Wishiwashi, use the school-forme's much higher stats
 			if (species.baseSpecies === 'Wishiwashi') baseStats = this.dex.getSpecies('wishiwashischool').baseStats;
 
-			let bst = baseStats.hp + baseStats.atk + baseStats.def + baseStats.spa + baseStats.spd + baseStats.spe;
+			let bst = species.bst;
+			if (species.baseSpecies === 'Wishiwashi') bst = this.dex.getSpecies('wishiwashischool').bst;
```
urkerab, line 989: Do we need both Wishiwashi checks?

server/chat-plugins/othermetas.ts
```diff
@@ -210,10 +214,12 @@ export const commands: ChatCommands = {
 		const deltas: StoneDeltas = {
 			baseStats: Object.create(null),
 			weighthg: megaSpecies.weighthg - baseSpecies.weighthg,
+			bst: 0,
```
urkerab, line 217:
```suggestion
			bst: megaSpecies.bst - baseSpecies.bst,
```
(Along with the removal of line 222 of course.)

KrisXV: I'm unsure why I didnt think of this in the first place lol

data/mods/gen7/random-teams.ts
KrisXV: id say up to @TheImmortal if he wants them consolidated

server/chat.ts
KrisXV: Unsure, it was like that presumably before this file was even typescripted, given the species === string check

data/mods/gen7/random-teams.ts
TheImmortal: Yeah we don't need the first one anymore.

server/chat.ts
Zarel: I suppose it should be implemented in dex, then?

server/chat.ts
Zarel: I think that's probably safe to remove since this function doesn't do mutation.

server/chat.ts
KrisXV: I don't see why we would do that

Zarel: I don't see why you would rather have an incorrect `species.bst` in Gen 1. That seems obviously worse to me.

Zarel:
I think this looks good now. I'm still not entirely sure how I feel about having it be precomputed (and thus vulnerable to getting desynced) but if you find it more convenient I suppose that's worth it.

PR #7141, opened 2020-08-04
Chat: Hide private commands better

mia-pi-git:
``CommandContext.secretCommand()`` is per Zarel, for any server's use. 
The other two are information leaks.


server/chat-commands/core.ts
```diff
@@ -1583,6 +1583,11 @@ export const commands: ChatCommands = {
 			this.errorReply(`Could not find help for '/${target}' - displaying help for '/${currentBestHelp.for.join(' ')}' instead`);
 		}
 
+		const curNamespace = this.splitCommand(`/${currentBestHelp.for[0]}`) as Chat.AnnotatedChatHandler;
```
Zarel, line 1586: This is not what "namespace" means...

server/chat-commands/core.ts
```diff
@@ -1583,6 +1583,11 @@ export const commands: ChatCommands = {
 			this.errorReply(`Could not find help for '/${target}' - displaying help for '/${currentBestHelp.for.join(' ')}' instead`);
```
Zarel, line 1583: Oh, could you remove `typeof nextNamespace === 'boolean'` on line 1563? It no longer exists after one of our older permission refactors.

Zarel:
This code all seems very clunky. I suspect the API of stuff like `splitCommand` just isn't designed for your use-case. I'll look into it.

server/chat.ts
```diff
@@ -1767,6 +1782,24 @@ export const Chat = new class {
...
+				if (typeof handler !== 'object') continue;
+				if (cmd in handler) cmd = command;
+			}
+		}
+		if (typeof this.commands[cmd] === 'string') return this.commands[cmd] as string;
+		return cmd;
+	}
+
+	/**
+>>>>>>> Chat: Add a method for marking commands as private
```
Zarel, line 1802:
```suggestion
```

LOL your merge conflict markers didn't get found because they were inside comments

server/chat.ts
```diff
@@ -1767,6 +1782,24 @@ export const Chat = new class {
 	}
 
 	/**
+<<<<<<< HEAD
+=======
```
Zarel, line 1786:
```suggestion
```

server/chat.ts
```diff
@@ -1767,6 +1782,24 @@ export const Chat = new class {
...
+				const handler = this.commands[command];
+				if (typeof handler !== 'object') continue;
+				if (cmd in handler) cmd = command;
+			}
+		}
+		if (typeof this.commands[cmd] === 'string') return this.commands[cmd] as string;
+		return cmd;
+	}
+
+	/**
```
Zarel, line 1801:
```suggestion
```

Actually `baseCommand` doesn't seem to be used at all.

server/chat-commands/core.ts
```diff
@@ -1583,6 +1583,11 @@ export const commands: ChatCommands = {
...
 		}
 
+		const curHandler = this.parseCommand(`/${currentBestHelp.for[0]}`)?.handler;
```
Zarel, line 1586: This will not work correctly for subcommands.

server/chat-commands/core.ts
```diff
@@ -1583,6 +1583,11 @@ export const commands: ChatCommands = {
...
+		if (curHandler?.isPrivate && !user.can('lock')) {
```
Zarel, line 1587: I wonder if putting it on the `lock` permission is bad here. Maybe the command should specify permission.

server/chat.ts
```diff
@@ -1339,6 +1340,20 @@ export class CommandContext extends MessageContext {
 	requiresRoom() {
 		this.errorReply(`/${this.cmd} - must be used in a chat room, not a ${this.pmTarget ? "PM" : "console"}`);
 	}
+	secretCommand() {
+		const handler = this.handler;
+		handler!.isPrivate = true;
```
Zarel, line 1345: This needs to be set by `annotateCommands` or whatever it's called, not here.

server/chat.ts
```diff
@@ -1339,6 +1340,20 @@ export class CommandContext extends MessageContext {
...
+		const full = this.fullCmd;
+		const token = this.cmdToken;
+		if (token === '!') {
+			return this.errorReply(`The command "${token}${this.cmd}" does not exist.`);
+		}
+		this.errorReply(
+			`The command "${token}${this.cmd}" does not exist. ` +
+			`To send a message starting with "${token}${full}", ` +
+			`type "${token.repeat(2)}${full}".`
```
Zarel, line 1354:
```suggestion
			`type "${token}${token}${full}".`
```
I'm surprised `repeat` exists now, but I don't think it actually adds to the readability here.

Zarel: Also, these should be `throw new Chat.ErrorMessage` right?

mia-pi-git:
Honestly, unless i'm missing something, couldn't ``CommandContext.can`` be modified to check if the handler is private, and if it is, and if the `user.can` returns false, to return the secretCommand method? 
Seems it'd be easier but i also wouldn't be the one who'd modify private code otherwise.

Zarel:
> Honestly, unless i'm missing something, couldn't `CommandContext.can` be modified to check if the handler is private, and if it is, and if the `user.can` returns false, to return the secretCommand method?
> Seems it'd be easier but i also wouldn't be the one who'd modify private code otherwise.

Actually that's a genius idea. Also you can do it without modifying private code.

mia-pi-git:
> Also you can do it without modifying private code.

Well, yeah, that's what i'm saying, lol - this way i don't have to, the other way someone else would.

Zarel:
Well, maybe a new `context.privatelyCan` function. You still want to mark it as private some way.

server/chat.ts
```diff
@@ -839,15 +840,32 @@ export class CommandContext extends MessageContext {
...
 		this.errorReply(`${this.cmdToken}${this.fullCmd} - Access denied.`);
 		return false;
 	}
+	privatelyCan(permission: RoomPermission, target: User | null, room: Room): boolean;
+	privatelyCan(permission: GlobalPermission, target?: User | null): boolean;
+	privatelyCan(permission: string, target: User | null = null, room: Room | null = null) {
+		if (!this.handler!.isPrivate) {
+			// @ts-ignore (can be several things, depending)
+			return this.can(permission, target, room);
+		}
```
Zarel, line 859:
```suggestion
		if (!this.handler!.isPrivate) {
			// @ts-ignore (can be several things, depending)
			return this.can(permission, target, room);
		}
```

This shouldn't happen, because `privatelyCan` should set `isPrivate`.

mia-pi-git: Oh, okay

server/chat.ts
```diff
@@ -839,15 +840,32 @@ export class CommandContext extends MessageContext {
 	can(permission: RoomPermission, target: User | null, room: Room): boolean;
 	can(permission: GlobalPermission, target?: User | null): boolean;
 	can(permission: string, target: User | null = null, room: Room | null = null) {
-		if (Users.Auth.hasPermission(this.user, permission, target, room, this.cmd, true)) return true;
+		if (Users.Auth.hasPermission(this.user, permission, target, room, this.fullCmd, true)) return true;
 		if (Users.Auth.hasPermission(this.user, permission, target, room, this.fullCmd, false)) {
```
Zarel, line 844: wtf did we get another botched merge somewhere?

server/chat.ts
```diff
@@ -1657,7 +1684,7 @@ export const Chat = new class {
 	}
 	loadPluginData(plugin: AnyObject) {
 		if (plugin.commands) {
-			Object.assign(Chat.commands, this.annotateCommands(plugin.commands));
+			Object.assign(Chat.commands, this.annotateCommands(plugin.commands, ''));
```
Zarel, line 1687: ಠ_ಠ

server/chat.ts
```diff
@@ -1647,6 +1661,7 @@ export const Chat = new class {
 			entry.requiresRoom = /\bthis\.requiresRoom\(/.test(handlerCode);
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
+			entry.isPrivate = /\bthis\.(?:privatelyCan|commandDoesntExist)\(/.test(handlerCode);
```
Zarel, line 1664: You forgot the most important part: actually annotating this command. Just so you know. :p

server/chat.ts
```diff
@@ -1358,6 +1364,14 @@ export class CommandContext extends MessageContext {
 	requiresRoom() {
 		this.errorReply(`/${this.cmd} - must be used in a chat room, not a ${this.pmTarget ? "PM" : "console"}`);
 	}
+	commandDoesntExist(): never {
```
KrisXV, line 1367:
```suggestion
	commandDoesNotExist(): never {
```

I don't like using combined words as parts of function names

Zarel: <_< All the same to me, I guess.

PR #7143, opened 2020-08-04
Add a BasicRoom#tr method

AnnikaCodes:
This makes it possible to translate things outside of a MessageContext, such as poll objects. (This could eventually replace `MessageContext#tr`, but I don't see a reason to convert old uses [and it is shorter!]).


Zarel:
It looks like this has the same API as `context.tr`, so it makes more sense as `room.tr`. If anything should be renamed to `translate`, it should probably be `Chat.tr`.

server/chat.ts
```diff
@@ -219,7 +219,6 @@ export abstract class MessageContext {
 		}
 		return false;
 	}
-
 	tr(strings: TemplateStringsArray | string, ...keys: any[]) {
```
Zarel, line 222:
```suggestion

	tr(strings: TemplateStringsArray | string, ...keys: any[]) {
```
(you seem to have accidentally deleted a linebreak)

server/chat-plugins/poll.ts
```diff
@@ -126,23 +128,25 @@ export class Poll {
...
 			output += `</div>`;
 		} else {
 			for (const [num, option] of this.options) {
-				output += `<div style="margin-top: 5px"><button class="button" style="text-align: left" value="/poll vote ${num}" name="send" title="Vote for ${num}. ${Utils.escapeHTML(option.name)}">${num}. <strong>${this.getOptionMarkup(option)}</strong></button></div>`;
+				output += `<div style="margin-top: 5px"><button class="button" style="text-align: left" value="/poll vote ${num}" name="send" title="${this.room.tr`Vote for ${num}`}. ${Utils.escapeHTML(option.name)}">${num}. <strong>${this.getOptionMarkup(option)}</strong></button></div>`;
 			}
-			output += `<div style="margin-top: 7px; padding-left: 12px"><button value="/poll results" name="send" title="View results - you will not be able to vote after viewing results"><small>(View results)</small></button></div>`;
+			output += `<div style="margin-top: 7px; padding-left: 12px"><button value="/poll results" name="send" title="${this.room.tr`View results - you will not be able to vote after viewing results`}"><small>(${this.room.tr`View results`})</small></button></div>`;
```
Zarel, line 138: I suspect `View results` and `you will not be able to vote after viewing results` should probably be separate translation options.

AnnikaCodes: Ya, fair.

Zarel:
Thanks!

PR #7145, opened 2020-08-04
[es] Announcement and poll message translations

FlareonBlitz:
This is for the messages shown when writing the different announcements and poll commands to Spanish. Translated by me, reviewed by Iyarito and other Spanish staff members.


Zarel:
Thanks!

PR #7146, opened 2020-08-04
Redesign /permissions help

mia-pi-git:


Zarel:
I personally don't think this looks better. But before we start investigating what we _do_ want, I think we should first split the code apart from the design.

Let's do the implementation first:

```ts
const subPermissions = allPermissions.filter(perm => perm.startsWith('/') && perm.includes(' '));
const subPermissionsByNamespace = {};
for (const perm of subPermissions) {
	const [namespace] = perm.split(' ', 1);
	if (!subPermissionsByNamespace[namespace]) subPermissionsByNamespace[namespace] = [];
	subPermissionsByNamespace[namespace].push(perm);
}
```

Zarel:
After that, the way I'd probably want to render it:

```ts
buffer += `<p><details class="readmore"><summary><strong>Sub-commands:</strong> (will affect one sub-command, like /roomevents view)</summary>`;
for (const subPerms of Object.values(subPermissionsByNamespace) {
	buffer += `<br /><code>` + subPerms.join(`</code> <code>`) + `</code><br />`;
}
buffer += `</details></p>`;
```

Zarel:
What does this look like, btw? Do you have a screenshot and also opinions?

mia-pi-git:
https://prnt.sc/tuwv4n
Personally, i think it's very wide, but you already nixed putting them in their own readmores, so I think this is good.

server/chat-commands/room-settings.ts
```diff
@@ -360,7 +366,11 @@ export const commands: ChatCommands = {
 			buffer += `<p><strong>Single-command permissions:</strong> (will affect one command)<br />`;
 			buffer += `<code>` + permissions.join(`</code> <code>`) + `</code></p>`;
 			buffer += `<p><details class="readmore"><summary><strong>Sub-commands:</strong> (will affect one sub-command, like /roomevents view)</summary>`;
```
Zarel, line 368:
```suggestion
```

I like how you didn't even notice that you repeated this line. :p

mia-pi-git: Hey, i noticed when I did it locally :c

Zarel: I mean, it's repeated in your screenshot too.

mia-pi-git: yes but i thought i'd prioritize screenshot over removing a line

Zarel:
I'm a fan.

PR #7149, opened 2020-08-04
Make the !faq command translatable

AnnikaCodes:


Zarel:
How useful is this if the individual links aren't translated, anyway?

Zarel:
Maybe these should be pages?

https://github.com/smogon/pokemon-showdown-client/tree/master/website/pages

The website CMS is public now.

AnnikaCodes:
> How useful is this if the individual links aren't translated, anyway?

I mostly made this PR for the `!faq ac` and `!faq tours` commands, but I felt silly not translating them all. I think the pages on the Smogon CMS (i.e. badge faq and tiering faq) can already be translated, but I can move those Smogon threads to website pages.

server/chat-commands/info.ts
```diff
@@ -1991,36 +1991,39 @@ export const commands: ChatCommands = {
...
+			buffer.push(`<a href="https://www.smogon.com/forums/posts/6774482/">${this.tr`Staff FAQ`}</a>`);
 		}
 		if (showAll || target === 'autoconfirmed' || target === 'ac') {
-			buffer.push(`A user is autoconfirmed when they have won at least one rated battle and have been registered for one week or longer. In order to prevent spamming and trolling, most chatrooms only allow autoconfirmed users to chat. If you are not autoconfirmed, you can politely PM a staff member (staff have %, @, or # in front of their username) in the room you would like to chat and ask them to disable modchat. However, staff are not obligated to disable modchat.`);
+			buffer.push(this.tr(autoconfirmedFAQ));
 		}
 		if (showAll || target === 'coil') {
-			buffer.push(`<a href="https://www.smogon.com/forums/threads/3508013/">What is COIL?</a>`);
+			buffer.push(`<a href="https://www.smogon.com/forums/threads/3508013/">${this.tr`What is COIL?`}</a>`);
 		}
```
Zarel, line 2009: Should be removed; COIL no longer exists.

server/chat-commands/info.ts
```diff
@@ -1991,36 +1991,39 @@ export const commands: ChatCommands = {
...
-			buffer.push(`<a href="https://www.smogon.com/ingame/battle/tiering-faq">Tiering FAQ</a>`);
+			buffer.push(`<a href="https://www.smogon.com/ingame/battle/tiering-faq">${this.tr`Tiering FAQ`}</a>`);
 		}
 		if (showAll || target === 'badge' || target === 'badges') {
-			buffer.push(`<a href="https://www.smogon.com/badge_faq">Badge FAQ</a>`);
+			buffer.push(`<a href="https://www.smogon.com/badge_faq">${this.tr`Badge FAQ`}</a>`);
 		}
 		if (showAll || target === 'rng') {
-			buffer.push(`<a href="https://${Config.routes.root}/pages/rng">How Pokémon Showdown's RNG works</a>`);
+			buffer.push(`<a href="https://${Config.routes.root}/pages/rng">${this.tr`How Pokémon Showdown's RNG works`}</a>`);
```
Zarel, line 2020: "Common misconceptions about our RNG"?

server/chat-commands/info.ts
```diff
@@ -1991,36 +1991,39 @@ export const commands: ChatCommands = {
 		target = target.toLowerCase().trim();
 		const showAll = target === 'all';
 		if (showAll && this.broadcasting) {
-			return this.sendReplyBox("You cannot broadcast all FAQs at once.");
+			return this.sendReplyBox(this.tr`You cannot broadcast all FAQs at once.`);
 		}
 
+		const autoconfirmedFAQ = `A user is autoconfirmed when they have won at least one rated battle and have been registered for one week or longer. In order to prevent spamming and trolling, most chatrooms only allow autoconfirmed users to chat. If you are not autoconfirmed, you can politely PM a staff member (staff have %, @, or # in front of their username) in the room you would like to chat and ask them to disable modchat. However, staff are not obligated to disable modchat.`;
```
Zarel, line 1997: Is there a reason this was pulled out? I think it makes more sense where it is.

AnnikaCodes: It was pulled out because of eslint's line length rules -- #7175 will make it possible to put it back.

Zarel: I mean, it can still be in the line above where it's used, instead of way up here. I think it's generally more readable that way.

For this specifically, yes, I'll merge #7175 so you don't need to worry about this. But for future reference, this sort of constant should just be declared the line above where it's used, for best readability.

AnnikaCodes: Alright, I'll keep that in mid.

Zarel:
Drafting until the merge conflict is fixed. Everything else in this PR is fine and looks good to me.

AnnikaCodes:
I fixed the merge conflict, and also updated it to use the pages from the pokemonshowdown.com CMS.

Zarel:
Thanks!

PR #7150, opened 2020-08-05
Add short alias for /sharedbattles

AnnikaCodes:
This was requested in Staff room a few days back but I forgot about it until now.


Zarel:
Thanks!

PR #7151, opened 2020-08-05
Allow locked users to be demoted to regular user

AnnikaCodes:


Zarel:
Wait, what's rank 1 then?

AnnikaCodes:
There is no rank 1. ` ` was rank 2 both before and after your refactor, so I assume it's intentional.

Zarel:
Huh, I guess punishments are below that? That's probably a bug.

PR #7152, opened 2020-08-05
Gen 3 Tier changes for month of August

petuuuhhh:
Breloom, Venusaur, and Smeargle to OU
Gyarados to UUBL


KrisXV:
This is not how smogon's tiering works.

PR #7156, opened 2020-08-06
Trivia: Make strings translatable

AnnikaCodes:
Trivia is a great example of why `Room#tr` is so useful.


AnnikaCodes:
(Also this PR will not be viable until the server is restarted for #7143.)

Zarel:
tyty

PR #7158, opened 2020-08-06
Help: fix crash

mia-pi-git:


server/chat-plugins/help.ts
```diff
@@ -69,7 +67,7 @@ export class HelpResponder {
 	queue: string[];
 	data: PluginData;
 	constructor(data: PluginData) {
-		this.data = data;
+		this.data = data || {};
```
Zarel, line 70: It does not seem like `|| {}` should be happening at all.

mia-pi-git: No, that should be throwing at this point, you're right.

server/chat-plugins/help.ts
```diff
@@ -15,18 +15,16 @@ const PATH = 'config/chat-plugins/help.json';
...
-} catch (e) {
-	helpData = {
-		stats: {},
-		pairs: {},
-		disabled: false,
-		queue: [],
-	};
-}
+	helpData = JSON.parse(FS(PATH).readIfExistsSync());
+} catch (e) {}
```
Zarel, line 27: I'm not sure the catch is correct, either. The only error we want to suppress is "doesn't exist". We should actually just use a regular `read`, and re-throw if we get any error other than "file not found".

Zarel: We can even initialize to your initial state there.

mia-pi-git:
(npm fulltest is passing fine on my own machine, fwiw)

PR #7159, opened 2020-08-06
Inheritance: Actually fix validator

KrisXV:


KrisXV:
The build failure is completely unrelated to this btw

Zarel:
It does seem like the ts-eslint repo is having trouble.

PR #7160, opened 2020-08-06
Fix crash message formatting

whalemer:


whalemer:
(again, test failure unrelated to my change)

server/rooms.ts
```diff
@@ -1408,10 +1408,11 @@ export class GlobalRoomState {
 			return;
 		}
 		this.lastReportedCrash = time;
+
 		const stack = err.stack || err.message || err.name || '';
-		const stackLines = (err ? Utils.escapeHTML(stack).split(`\n`) : []);
+		const stackLines = Utils.escapeHTML(stack).split(`\n`);
 
-		let crashMessage = `|html|<div class="broadcast-red"><details class="readmore"><summary><b>${crasher} crashed:</b> ${stackLines[0]}</summary>${stack}</details></div>`;
+		let crashMessage = `|html|<div class="broadcast-red"><details class="readmore"><summary><b>${crasher} crashed:</b> ${stackLines[0]}</summary>${stackLines.slice(1).join('<br>')}</details></div>`;
```
Zarel, line 1415:
```suggestion
		let crashMessage = `|html|<div class="broadcast-red"><details class="readmore"><summary><b>${crasher} crashed:</b> ${stackLines[0]}</summary>${stackLines.slice(1).join('<br />')}</details></div>`;
```

server/rooms.ts
```diff
@@ -1408,10 +1408,11 @@ export class GlobalRoomState {
 			return;
 		}
 		this.lastReportedCrash = time;
+
 		const stack = err.stack || err.message || err.name || '';
```
Zarel, line 1412:
```suggestion
		const stack = (err && (err.stack || err.message || err.name)) || '';
```

whalemer: when does that happen? ts says that it's not nullable, and I'm inclined to agree based on every obvious call I can see?

Zarel: image

Zarel: This is clearly a bug and no one should ever throw `null` but... this _is_ a crashlogger for bugs.

Zarel: Also TS doesn't say it's not nullable. The only reason TS doesn't complain is because `err` has type `any`, which is currently hugely controversial anyway.

whalemer: fun!

PR #7164, opened 2020-08-06
Permissions: Support setting permissions for all sub-commands at once

mia-pi-git:
This makes it so you can do ``/permissions set /roomevents,+`` and, as a result, all /roomevents commands will be usable by voices (unless you, say, /permissions set /roomevents add,%, which will make it so voices can't use that command, etc)


Zarel:
Looks good, sorry I missed it.

PR #7171, opened 2020-08-07
Show OHKO message if OHKO move breaks substitute

Dawnofares:
Based on https://github.com/smogon/pokemon-showdown/pull/5637. I'm not sure what gens specifically have this interaction, but maybe a researcher can talk about that?


Marty-D:
Since DaWoblefet said the OHKO message happens before the sub faded message, you can probably just check for an OHKO move right above each `target.removeVolatile('substitute');`, add the message there, and forget about editing scripts at all.

DaWoblefet:
Let me verify oldgen behavior before merging this - I don't think it plays the OHKO message into Sub in every generation.

Dawnofares:
Sounds good.

DaWoblefet:
This behavior happens from Gen 4-8 at least. I don't have a great way of testing before 4.

Marty-D:
Checked Gens 1-3, happens there too except the message comes after the sub breaking.

Marty-D:
And I just checked Gen 4 because I find it more likely the engine change between 4 and 5 caused the difference, and sure enough Gen 4 also has the OHKO message come up after the sub faded message.

Gens 1-4: fade -> OHKO
Gens 5-8: OHKO -> fade

DaWoblefet:
Good catch! I was only checking for the message at all, and didn't pay attention to the order.

Dawnofares:
What's the status on this?

Marty-D:
I don't understand the change you made to Gen 3; Gens 1-4 behave the same way in this situation.

DaWoblefet:
@Marty bump, I think this was ready to go a while ago.

Marty-D:
Oh, I don't get notifications for commits. Now I see what the Gen 3 change was meant to do, but the latest change just removes the message entirely when a substitute doesn't exist.

PR #7173, opened 2020-08-07
Fix modlog formatting

AnnikaCodes:
Fixes this.
image

### DIFF
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -208,7 +208,7 @@ export class HelpTicket extends Rooms.RoomGame {
 
 	modnote(user: User, text: string) {
 		this.room.addByUser(user, text).update();
-		this.room.modlog(`${text}`);
+		this.room.modlog(text);
 	}
 
 	getPreview() {
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -208,7 +208,7 @@ export class HelpTicket extends Rooms.RoomGame {
 
 	modnote(user: User, text: string) {
 		this.room.addByUser(user, text).update();
-		this.room.modlog(`(${this.room.roomid}) ${text}`);
+		this.room.modlog(`${text}`);
 	}
 
 	getPreview() {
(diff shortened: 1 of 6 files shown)


server/chat-plugins/helptickets.ts
```diff
@@ -208,7 +208,7 @@ export class HelpTicket extends Rooms.RoomGame {
 
 	modnote(user: User, text: string) {
 		this.room.addByUser(user, text).update();
-		this.room.modlog(`(${this.room.roomid}) ${text}`);
+		this.room.modlog(`${text}`);
```
Zarel, line 211:
```suggestion
		this.room.modlog(text);
```

server/chat-plugins/helptickets.ts
```diff
@@ -208,7 +208,7 @@ export class HelpTicket extends Rooms.RoomGame {
 
 	modnote(user: User, text: string) {
 		this.room.addByUser(user, text).update();
```
Zarel, line 210: Wait, is this supposed to be public?

Zarel: I thought modnotes were mod-only.

server/chat-plugins/scavengers.ts
```diff
@@ -898,7 +898,7 @@ export class ScavengerHunt extends Rooms.RoomGame {
 
 			this.room.sendMods(staffMsg);
 			this.room.roomlog(staffMsg);
-			this.room.modlog(`(${this.room.roomid}) ${logMsg}`);
+			this.room.modlog(`${logMsg}`);
```
Zarel, line 901:
```suggestion
			this.room.modlog(logMsg);
```

server/chat-plugins/scavengers.ts
```diff
@@ -881,7 +881,7 @@ export class ScavengerHunt extends Rooms.RoomGame {
 			const logMsg = `([${player.id}] has been caught trying to do their own hunt.)`;
 			this.room.sendMods(staffMsg);
 			this.room.roomlog(staffMsg);
-			this.room.modlog(`(${this.room.roomid}) ${logMsg}`);
+			this.room.modlog(`${logMsg}`);
```
Zarel, line 884:
```suggestion
			this.room.modlog(logMsg);
```

server/chat-plugins/scavengers.ts
AnnikaCodes: oh, looks like I missed some lol

server/chat-plugins/helptickets.ts
AnnikaCodes: Perhaps it's different in help tickets, since only the global staff handling the ticket and the person making the ticket would be in the room?

AnnikaCodes: Modnotes in regular rooms are definitely mod-only.

Zarel: This function should probably not be called "modnote", so it's not confused with the mod-only modnotes.

Zarel: But that can be a separate PR>

Zarel:
Was this from your modlog refactor, or was this an older/newer issue?

PR #7176, opened 2020-08-07
Groupchats: Inherit parent settings if it's a subroom

mia-pi-git:
https://www.smogon.com/forums/threads/make-subroomgroupchats-inherit-parent-rooms-roomsettings-by-default.3660035/#post-8560349

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -766,12 +766,12 @@ export const commands: ChatCommands = {
 			isPrivate: 'hidden',
 			creationTime: Date.now(),
 			modjoin: parent ? null : '+',
-			auth: {},
 		};
 		const settings = parent ? room.settings : defaultSettings;
 		const targetRoom = Rooms.createChatRoom(roomid, `[G] ${title}`, {
 			...settings,
 			parentid: parent,
+			auth: {},
 			introMessage: `` +
 				`<div style="text-align: center"><table style="margin:auto;"><tr><td><img src="//${Config.routes.client}/fx/groupchat.png" width=120 height=100></td><td><h2>${titleMsg}</h2><p>Follow the <a href="/rules">Pokémon Showdown Global Rules</a>!<br>Don't be disruptive to the rest of the site.</p></td></tr></table></div>`,
 			staffMessage: `` +
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -761,13 +761,18 @@ export const commands: ChatCommands = {
 
 		const titleMsg = Utils.html`Welcome to ${parent ? room.title : user.name}'s` +
 			`${!/^[0-9]+$/.test(title) ? ` ${title}` : ''}${parent ? ' subroom' : ''} groupchat!`;
-		const targetRoom = Rooms.createChatRoom(roomid, `[G] ${title}`, {
+		const defaultSettings = {
 			isPersonal: true,
 			isPrivate: 'hidden',
 			creationTime: Date.now(),
 			modjoin: parent ? null : '+',
 			parentid: parent,
 			auth: {},
+		};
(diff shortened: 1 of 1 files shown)


Zarel:
Which settings do you want to inherit? I get the impression a lot of settings (maybe even most) are dangerous to inherit.

Zarel:
You're also shallow-inheriting auth, which will cause bugs because auth isn't designed to be synced like that.

Zarel:
I actually can't think of any settings you'd _want_ to inherit, besides auth which is already inherited.

mia-pi-git:
> I actually can't think of any settings you'd _want_ to inherit, besides auth which is already inherited.

I presume things like languages, tournament settings, filters (emoji, slowchat, etc), and so on.
Which ones would be dangerous? Those could probably just be manually overridden.
```ts
isPersonal?: boolean;
	isHelp?: boolean;
	noLogTimes?: boolean;
	noAutoTruncate?: boolean;
	isMultichannel?: boolean;
       staffRoom?: boolean;
      readonly staffAutojoin?: string | boolean;
      readonly autojoin?: boolean;
```

Beyond these, it seems like a lot of the settings would be desirable to inherit, at least imo.

Zarel:
You're right, language is good to inherit.

The others don't really make sense to me. Like people often make subroom groupchats for tournaments, and subrooms are smaller so they don't need as strict filters as their parent rooms, usually.

mia-pi-git:
> You're right, language is good to inherit.
> 
> The others don't really make sense to me. Like people often make subroom groupchats for tournaments, and subrooms are smaller so they don't need as strict filters as their parent rooms, usually.

True, but given the use-case for subrooms in public rooms (events and the like) it seems like you'd wanna still regulate.
Looking at the interface, banwords, game settings, rules, data command tier, /show and /requestshow permissions, room permissions, rules link, possibly scavs, and slowchat also seem good to inherit.

Zarel:
Most of those seem good, except game settings which should probably all stay at default (enabled), and slowchat, which as mentioned doesn't seem like a good default for a subroom that has fewer people.

Zarel:
My thoughts are:

- language, data command tier - obvious yes

- rules, banwords, permissions - sure, I suppose that's convenient

- game permissions, slowchat, everything else - I think resetting these to default makes more sense

mia-pi-git:
> My thoughts are:
> 
> * language, data command tier - obvious yes
> * rules, banwords, permissions - sure, I suppose that's convenient
> * game permissions, slowchat, everything else - I think resetting these to default makes more sense

So, with regards to [[this thread<https://www.smogon.com/forums/threads/make-subroomgroupchats-inherit-parent-rooms-roomsettings-by-default.3660035/>]], should i still make it just the few settings, or continue with the current approach?

Zarel:
From TI's post, it sounds to me like the problem is that the defaults are bad.

Zarel:
I suppose yeah, games default to off, but you definitely want them playable in groupchats.

Maybe groupchats should just default all games to on?

server/chat-commands/room-settings.ts
```diff
@@ -761,11 +761,15 @@ export const commands: ChatCommands = {
 
 		const titleMsg = Utils.html`Welcome to ${parent ? room.title : user.name}'s` +
 			`${!/^[0-9]+$/.test(title) ? ` ${title}` : ''}${parent ? ' subroom' : ''} groupchat!`;
-		const targetRoom = Rooms.createChatRoom(roomid, `[G] ${title}`, {
+		const defaultSettings = {
 			isPersonal: true,
 			isPrivate: 'hidden',
 			creationTime: Date.now(),
 			modjoin: parent ? null : '+',
```
Zarel, line 768: Anyway, do these all seem like things that you _don't_ want to be true for subroom groupchats? Because literally none of them seem like that to me.

PR #7177, opened 2020-08-07
Random Doubles: fix Lopunny

ACakeWearingAHat:
also removed a protect rejection for doubles


data/random-teams.ts
```diff
@@ -1230,7 +1230,7 @@ export class RandomTeams {
 		} else if (species.name === 'Farfetch\u2019d') {
 			item = 'Leek';
 		} else if (species.name === 'Lopunny') {
-			item = 'Toxic Orb';
+			item = (isDoubles) ? 'Iron Ball' : 'Toxic Orb';
```
AnnikaCodes, line 1233: You don't need parentheses around `isDoubles`.

PR #7178, opened 2020-08-07
Add German translation for poll text

xMarcell:
New german lines for polls, GPed by asgdf

[ I was a bit confused by the trivia lines, since trivia games can only be started in the trivia room and the room does not have its language setting set to "german" , due to that it would not be possible for the server to translate them, so I left them open; if it's still possible and I am wrong, please let me know, then I will add the remaining lines, thanks :) ]


AnnikaCodes:
> [ I was a bit confused by the trivia lines, since trivia games can only be started in the trivia room and the room does not have its language setting set to "german" , due to that it would not be possible for the server to translate them, so I left them open; if it's still possible and I am wrong, please let me know, then I will add the remaining lines, thanks :) ]

Trivia is translatable in case side servers that are in other languages want to play Trivia. It's not a big priority to translate it.

translations/german.json
```diff
@@ -124,21 +124,21 @@
...
-		"Submit your vote": "",
-		"Quiz": "",
-		"Poll": "",
-		"Submit": "",
-		"ended": "",
-		"votes": "",
-		"delete": "",
+		"That option is not selected.": "Diese Wahloption ist nicht ausgew&auml;hlt.",
+		"You have already voted for this poll.": "Du hast bereits f&uuml;r diese Umfrage abgestimmt.",
+	  "No options selected.": "Keine Wahloption wurde ausgew&auml;hlt.",
```
Zarel, line 129:
```suggestion
		"No options selected.": "Keine Wahloption wurde ausgew&auml;hlt.",
```

I believe you messed up the indentation for this one.

Zarel:
Also, as mentioned in the other one, please don't use HTML entities. You can use ü or whatever in the translations directly.

xMarcell:
Alright, I will avoid HTML entities in the future, thank you both

xMarcell:
I commited your suggestion regarding the indentation, Zarel, is there anything else to do, or can this be merged now? Thank you

Zarel:
You haven't changed the HTML entities yet.

Zarel:
Specifically, you should be using `ä` instead of `&auml;`, and `ü` instead of `&uuml;`

xMarcell:
Hey, there are other HTML entities from previous pull requests, shall I replace them aswell or leave them and just edit the ones from the present request? Sorry for asking too much questions, I am relatively new to this

Zarel:
Yes, please do, thank you!

xMarcell:
Removed any HTML entities from the file, should be good now

Zarel:
Thanks!

PR #7180, opened 2020-08-08
Support ignoring room invites

mia-pi-git:
Requested by various auth members.


Zarel:
Should we also block room links with no surrounding context?

mia-pi-git:
> Should we also block room links with no surrounding context?

That feels riskier - there might be actual reasons for that. I've seen people who, say, try to report a user by just sending a link.  
(not saying that's the right thing, just don't wanna discourage possibly needlessly.)
There are other legitimate uses but i can't really think of any off the top of my head.

mia-pi-git:
Said example might not be _bad_ to block per se, but I do think there might be legitimate use-cases for just a room link.
The question is if the number of use-cases outweigh the cases where it'd be better to block, i guess.

Zarel:
Presumably users might also want to report by sending an invite.

I get invites and context-free room links about equally often, and they're usually used exactly the same way.

mia-pi-git:
Should now block links that have no context.

server/chat.ts
```diff
@@ -2054,6 +2064,11 @@ export const Chat = new class {
 				targetUser.send(`${prefix}The user '${Utils.escapeHTML(user.name)}' attempted to challenge you to a battle but was blocked. To enable challenges, use /unblockchallenges ${options}`);
 				targetUser.notified.blockChallenges = true;
 			}
+		} else if (blocked === 'invite') {
+			if (!targetUser.notified.blockChallenges) {
```
Zarel, line 2068: Did you plan to use the same notification system as `blockChallenges` or a different one?

If "same", you should remove `notified.blockInvites` from `users.ts`. If different, this line should be `notified.blockInvites`.

server/users.ts
```diff
@@ -44,7 +44,7 @@ const PERMALOCK_CACHE_TIME = 30 * 24 * 60 * 60 * 1000; // 30 days
 const DEFAULT_TRAINER_SPRITES = [1, 2, 101, 102, 169, 170, 265, 266];
 
 import {FS} from '../lib/fs';
-import {Auth, GlobalAuth, PLAYER_SYMBOL, HOST_SYMBOL, RoomPermission, GlobalPermission} from './user-groups';
+import {Auth, GlobalAuth, PLAYER_SYMBOL, HOST_SYMBOL, RoomPermission, GlobalPermission, AuthLevel} from './user-groups';
```
Zarel, line 47: It looks like `AuthLevel` is already a global type?

mia-pi-git:
@Zarel bump

Zarel:
Thanks!

PR #7181, opened 2020-08-08
Roomsettings: Fix bug with single-command permissions

AnnikaCodes:
We shouldn't be checking `Auth#hasPermission` if the permission isn't a room or global permission, which I didn't understand in my previous bugfix.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -7,7 +7,7 @@
  * @license MIT
  */
 import {Utils} from '../../lib/utils';
-import {ROOM_PERMISSIONS, GLOBAL_PERMISSIONS, GlobalPermission, RoomPermission} from '../user-groups';
+import type {RoomPermission} from '../user-groups';
 
 const RANKS = Config.groupsranking;
 
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -301,13 +301,14 @@ export const commands: ChatCommands = {
 			if (!Users.Auth.supportedRoomPermissions(room).includes(perm)) {
 				return this.errorReply(`${perm} is not a valid room permission.`);
 			}
-			if (!room.auth.atLeast(user, '#') ||
-				(
-					[...ROOM_PERMISSIONS, ...GLOBAL_PERMISSIONS].includes(perm as GlobalPermission | RoomPermission) &&
-					Users.Auth.hasPermission(user, perm, null, room)
-				)
+			if (!room.auth.atLeast(user, '#')) {
+				return this.errorReply(`/permissions set - You must be at least a Room Owner to set permissions.`);
+			}
+			if (
+				Users.Auth.ROOM_PERMISSIONS.includes(perm as RoomPermission) &&
+				!Users.Auth.hasPermission(user, perm, null, room)
 			) {
-				return this.errorReply(`/permissions set - Access denied.`);
+				return this.errorReply(`/permissions set - You can't set the permission "${perm}" because you don't have it.`);
 			}
 
 			const currentPermissions = room.settings.permissions || {};
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -7,6 +7,7 @@
  * @license MIT
  */
 import {Utils} from '../../lib/utils';
+import {ROOM_PERMISSIONS, GLOBAL_PERMISSIONS, GlobalPermission, RoomPermission} from '../user-groups';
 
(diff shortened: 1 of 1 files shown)


server/chat-commands/room-settings.ts
```diff
@@ -300,7 +301,12 @@ export const commands: ChatCommands = {
...
 			}
-			if (!(room.auth.atLeast(user, '#') && Users.Auth.hasPermission(user, perm, null, room))) {
+			if (!room.auth.atLeast(user, '#') ||
+				(
+					[...ROOM_PERMISSIONS, ...GLOBAL_PERMISSIONS].includes(perm as GlobalPermission | RoomPermission) &&
+					Users.Auth.hasPermission(user, perm, null, room)
+				)
+			) {
 				return this.errorReply(`/permissions set - Access denied.`);
 			}
```
Zarel, line 311: These would almost be better as two separate tests.

```suggestion
			if (!room.auth.atLeast(user, '#')) {
				return this.errorReply(`/permissions set - You must be at least a Room Owner to set permissions.`);
			}
			if (
				Users.Auth.ROOM_PERMISSIONS.includes(perm as RoomPermission) &&
				Users.Auth.hasPermission(user, perm, null, room)
			) {
				return this.errorReply(`/permissions set - You can't set the permission "${perm}" because you already have it.`);
			}
```

Actually this test looks weird. What is this actually supposed to be guarding?

server/chat-commands/room-settings.ts
```diff
@@ -7,6 +7,7 @@
  * @license MIT
  */
 import {Utils} from '../../lib/utils';
+import {ROOM_PERMISSIONS, GLOBAL_PERMISSIONS, GlobalPermission, RoomPermission} from '../user-groups';
```
Zarel, line 10:
```suggestion
import type {RoomPermission} from '../user-groups';
```

Zarel: tbh we might want to export this type from `Users` so we don't have to import it all the time.

server/chat-commands/room-settings.ts
AnnikaCodes: It's supposed to be preventing users from setting permissions for things they can't do; for example, a Room Owner who is a Global Voice shouldn't be able to set the permissions for appointing new ROs to # and then demote/promote ROs.

Zarel: Shouldn't it be `!Users.Auth.hasPermission` rather than `Users.Auth.hasPermission`, then?

Zarel:
```suggestion
			if (!room.auth.atLeast(user, '#')) {
				return this.errorReply(`/permissions set - You must be at least a Room Owner to set permissions.`);
			}
			if (
				Users.Auth.ROOM_PERMISSIONS.includes(perm as RoomPermission) &&
				!Users.Auth.hasPermission(user, perm, null, room)
			) {
				return this.errorReply(`/permissions set - You can't set the permission "${perm}" because you don't have it.`);
			}
```

Zarel: Just to make sure: Was the previous code just bugged?

server/chat-commands/room-settings.ts
```diff
@@ -7,6 +7,7 @@
  * @license MIT
  */
 import {Utils} from '../../lib/utils';
+import type {RoomPermission} from '../user-groups';
```
Zarel, line 10: Moment of truth: how many copies of PS crash with this code in?

This requires a relatively more recent version of sucrase than before, and not everyone remembers that compile errors mean you should rerun `npm install`.

Zarel: Normally we use feature-detection and auto-reinstall, but I'm not sure if there's an easy way to do that for sucrase.

Zarel:
```
> require('sucrase').getVersion()
'3.15.0'
```

AHA

Can you add this check to the reinstall checker?

whalemer: like https://github.com/smogon/pokemon-showdown/blob/master/build#L66 ?

Zarel: Actually, I will.

Zarel: Yeah, don't worry about that.

server/chat-commands/room-settings.ts
AnnikaCodes: Yes, it prevented ROs from setting permissions for things like blacklisting.

Zarel:
Thanks!

PR #7183, opened 2020-08-08
Support updating private code in /updateserver

AnnikaCodes:
Requested by pre.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -738,11 +738,11 @@ export const commands: ChatCommands = {
 	async updateserver(target, room, user, connection) {
 		if (!this.canUseConsole()) return false;
 		const isPrivate = toID(target) === 'private';
-		if (Monitor.updateServerLock === (isPrivate ? 'private' : 'public')) {
+		if (Monitor.updateServerLock) {
 			return this.errorReply(`/updateserver - Another update is already in progress (or a previous update crashed).`);
 		}
 
-		Monitor.updateServerLock = isPrivate ? 'private' : 'public';
+		Monitor.updateServerLock = true;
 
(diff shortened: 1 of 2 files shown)


AnnikaCodes:
ok i cant type

server/chat-commands/admin.ts
```diff
@@ -735,18 +737,18 @@ export const commands: ChatCommands = {
 
 	async updateserver(target, room, user, connection) {
 		if (!this.canUseConsole()) return false;
-
-		if (Monitor.updateServerLock) {
+		const isPrivate = toID(target) === 'private';
+		if (Monitor.updateServerLock === (isPrivate ? 'private' : 'public')) {
```
Zarel, line 741: I think having separate private/public locks is overkill, but even if it's not, this is wrong since updating private would override the public lock and vice versa.

AnnikaCodes: That's a... large mistake on my part. Fixed (and yeah thinking about how often /updateserver is used, we probably don't need to support updating private and public code concurrently, although it wouldn't be hard).

server/chat-commands/admin.ts
```diff
@@ -16,6 +16,8 @@ import {Utils} from '../../lib/utils';
 
 import * as ProcessManager from '../../lib/process-manager';
 
+const PRIVATE_CODE_PATH = '../pokemon-showdown-private/';
```
scheibo, line 19: not sure about committing this constant (why is this not on `Config`?) especially given its actually incorrect (its called `main-private` on the box, not `pokemon-showdown-private`)...

AnnikaCodes: Sorry, I didn't know that. I'll move it to Config.

Zarel:
Looks good now, thanks!

PR #7187, opened 2020-08-09
Fix modlog tests on Windows

AnnikaCodes:
While it's certainly possible to implement a `FS#rmdir` method, removing directories on Node versions <12 is rather clunky given the minute idiosyncrasies of different operating systems and Node versions. Since PS! doesn't really need to remove directories, I think simply removing all the files within the directory is a reasonable solution for this usecase.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x]
+        node-version: [12.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ If your distro package manager has an old Node.js version, the simplest way to u
 Detailed installation instructions
 ------------------------------------------------------------------------
 
-Pokémon Showdown requires you to have [Node.js][6] installed, v10.x or later.
+Pokémon Showdown requires you to have [Node.js][6] installed, v12.x or later.
 
 Next, obtain a copy of Pokémon Showdown. If you're reading this outside of GitHub, you've probably already done this. If you're reading this in GitHub, there's a "Clone or download" button near the top right (it's green). I recommend the "Open in Desktop" method - you need to install GitHub Desktop which is more work than "Download ZIP", but it makes it much easier to update in the long run (it lets you use the `/updateserver` command).
 
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -331,7 +331,7 @@ export class FSPath {
 			// In Node versions before 12, fs.rmdir() didn't support recursion,
 			// so we have to implement recursion ourselves to support Node 10 and 11...
 			for (const file of await this.readdir()) {
-				const subpath = FS(`${this.path}/${file}`);
+				const subpath = FS(`${this.path}${pathModule.sep}${file}`);
 				if (await subpath.isDirectory()) {
 					await subpath.rmdir(true);
 				} else {
@@ -350,7 +350,7 @@ export class FSPath {
 		if (Config.nofswriting) return;
 		if (recursive) {
 			for (const file of this.readdirSync()) {
-				const subpath = FS(`${this.path}/${file}`);
+				const subpath = FS(`${this.path}${pathModule.sep}${file}`);
 				if (subpath.isDirectorySync()) {
 					subpath.rmdirSync(true);
 				} else {
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -325,40 +325,20 @@ export class FSPath {
 		}
 	}
 
-	async rmdir(recursive?: boolean) {
+	async removeFilesFromDirectory() {
 		if (Config.nofswriting) return Promise.resolve();
-		if (recursive) {
-			// In Node versions before 12, fs.rmdir() didn't support recursion,
-			// so we have to implement recursion ourselves to support Node 10 and 11...
-			for (const file of await this.readdir()) {
-				const subpath = FS(`${this.path}${pathModule.sep}${file}`);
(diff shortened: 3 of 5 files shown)


DaWoblefet:
Verifying as a Windows user that this works around the issue me / Kris / HoeenHero / any other Windows user was having with tests failing. Would another solution be just defaulting to Node 12? (not sure if Zarel would like it, and leaving an empty directory is fine imo).

AnnikaCodes:
> Verifying as a Windows user that this works around the issue me / Kris / HoeenHero / any other Windows user was having with tests failing. Would another solution be just requiring Node 12? (not sure if Zarel would like it, and leaving an empty directory is fine imo).

I would love just requiring Node 12, since it has a native recursive `rmdir` method. OTOH, this works just fine for the purpose of testing modlogs, and we could delay requiring Node 12 until we need to "properly" remove a directory (idk how many servers are running on Node 10/11).

Zarel:
I remember this being a weird Windows issue, but is this really the only way? That's really surprising.

Zarel:
Why not just drop support for Node versions before 12?

Zarel:
I'm fine with PS bumping the Node requirement up to 12.

Zarel:
I'm confused what specific idiosyncrasy is affecting the Node pre-12 code, though. It looks like it should work fine...

AnnikaCodes:
According to [Node's own source](https://github.com/nodejs/node/blob/master/lib/internal/fs/rimraf.js#L236-L239), "Windows has a habit of not closing handles promptly when files are deleted, resulting in spurious ENOTEMPTY failures."

DaWoblefet:
Welp I was wrong, bumping to Node 12 with these changes doesn't seem to help the issue. For Zarel's sake, this is what's currently happening on Windows:

 1) Modlog (with FS writes)
       "after all" hook for "should support filtering out non-punishment-related logs":
     Error: ENOTEMPTY: directory not empty, rmdir 'C:\Users\dawob\JavaScript Projects\Pokemon-Showdown\test\modlogs'


  2) Rooms features
       BasicRoom
         getGame
           should return the game only when the gameids match:
     Uncaught Error: ENOENT: no such file or directory, open 'C:\Users\dawob\JavaScript Projects\Pokemon-Showdown\logs\modlog\modlog_r\relationshipadvice.txt'


It's claiming that the directory isn't empty when it definitely is after removing the other files.

TheJetOU:
I suspect that you'd need to add a `maxRetries` in the `options` object of `fs#rmdir` until Windows actually closes the handlers.

Zarel:
I'll merge this for now. But is Wob saying that this doesn't actually fix the problem? That's frustrating.

DaWoblefet:
Yeah it didn't actually change anything unfortunately :(

Annika's original fix was to just not actually delete the directory, but I think we'll have to do something like jetou's approach if we want to get anywhere

AnnikaCodes:
> Yeah it didn't actually change anything unfortunately :(
> 
> Annika's original fix was to just not actually delete the directory, but I think we'll have to do something like jetou's approach if we want to get anywhere

Does specifying `maxRetries` help? I'm kind of flying blind here because I don't have access to a Windows testing environment.

DaWoblefet:
Don't think so; this is what happens when I set ``maxRetries`` to 10:

``async rmdir(recursive?: boolean) {
		if (Config.nofswriting) return Promise.resolve();
			fs.rmdir(this.path, {recursive, maxRetries: 10}, err => {
				err ? reject(err) : resolve();
			});
		});
	}``
  ................................................,,,...................................!....!

  87 passing (5s)
  3 pending
  2 failing

  1) Modlog (with FS writes)
       "after all" hook for "should support filtering out non-punishment-related logs":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (C:\Users\dawob\JavaScript Projects\Pokemon-Showdown\test\server\modlog.js)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)

  2) Rooms features
       BasicRoom
         getGame
           should return the game only when the gameids match:
     Uncaught Error: ENOENT: no such file or directory, open 'C:\Users\dawob\JavaScript Projects\Pokemon-Showdown\logs\modlog\modlog_r\relationshipadvice.txt'

DaWoblefet:
Ok some more testing:

Since the previously failing test was due to a timeout and not a "real" crash, I increased the mocha timeout locally, set maxRetries to a high value and observed that the file "modlog_battle.txt" wasn't being successfully destroyed (all other text files were destroyed). It seems to lack permissions to destroy or view this file in particular:
image

image

image

Doesn't make much sense to me, as I have full admin rights on my machine (hopping on an admin command prompt and attempting to delete the file doesn't work either). Immediately after I kill the process Showdown is running on (Ctrl+C), the file disappears.

AnnikaCodes:
> Ok some more testing:
> 
> Since the previously failing test was due to a timeout and not a "real" crash, I increased the mocha timeout locally, set maxRetries to a high value and observed that the file "modlog_battle.txt" wasn't being successfully destroyed (all other text files were destroyed). It seems to lack permissions to destroy or view this file in particular:
> image
> 
> image
> 
> image
> 
> Doesn't make much sense to me, as I have full admin rights on my machine (hopping on an admin command prompt and attempting to delete the file doesn't work either). Immediately after I kill the process Showdown is running on (Ctrl+C), the file disappears.

Maybe it's something going on with the modlog stream? Try calling `modlog.destroy('battle')` before deleting the directory? (this all works fine on Mac and Linux)

DaWoblefet:
Nope, no luck :(

PR #7191, opened 2020-08-10
Support a per-user PM blacklist

mia-pi-git:
https://www.smogon.com/forums/threads/add-a-pm-blacklist.3658140/


server/chat-commands/core.ts
```diff
@@ -732,12 +788,7 @@ export const commands: ChatCommands = {
 			}
 			for (const setting in user.settings) {
 				if (setting in raw) {
-					if (setting === 'blockPMs' &&
-						Users.Auth.isAuthLevel(raw[setting])) {
-						settings[setting] = raw[setting];
-					} else {
-						settings[setting as keyof UserSettings] = !!raw[setting];
-					}
+					settings[setting as keyof UserSettings] = typeof raw[setting] === 'object' ? raw[setting] : !!raw[setting];
```
scheibo, line 791: you still need to check if `Users.Auth.isAuthLevel`... the whole point of this code is to not blindly expect potentially malicious input...

server/chat.ts
```diff
@@ -1141,6 +1116,55 @@ export class CommandContext extends MessageContext {
...
 	}
+	canPM(targetUser: User) {
+		// hotpatching
+		const user = this.user;
+		if (typeof user.settings.blockPMs === 'boolean') {
+			user.settings.blockPMs = {all: user.settings.blockPMs};
+		}
+		if (typeof targetUser.settings.blockPMs === 'boolean') {
+			targetUser.settings.blockPMs = {all: targetUser.settings.blockPMs};
+		}
```
Zarel, line 1127: You have this code in five different places and I can't help but think there's a better way.

At least make it a `Chat.upgradeSettings` function?

server/chat.ts
```diff
@@ -1141,6 +1116,55 @@ export class CommandContext extends MessageContext {
...
+				tarBlocks.specific?.includes(curUser.id)
+			) && !curUser.can('lock');
+		}
+
+		if (targetUser.locked && !user.can('lock')) {
+			this.errorReply(`The user "${targetUser.name}" is locked and cannot be PMed.`);
+			return null;
+		}
+		if (Config.pmmodchat && !Users.globalAuth.atLeast(user, Config.pmmodchat) &&
+			!Users.Auth.hasPermission(targetUser, 'promote', Config.pmmodchat as GroupSymbol)
```
Zarel, line 1144:
```suggestion
		if (
			Config.pmmodchat && !Users.globalAuth.atLeast(user, Config.pmmodchat) &&
			!Users.Auth.hasPermission(targetUser, 'promote', Config.pmmodchat as GroupSymbol)
```

Zarel: Still really sad eslint can't enforce this.

Zarel: There's `array-bracket-newline`, `function-paren-newline`, and `object-curly-newline`, but no generalized paren newline. :[

server/users.ts
```diff
@@ -1063,7 +1069,7 @@ export class User extends Chat.MessageContext {
 				this.semilocked = '#dnsbl.' as PunishType;
 			}
 		}
-		if (this.settings.blockPMs && this.can('lock') && !this.can('bypassall')) this.settings.blockPMs = false;
+		if (this.settings.blockPMs && this.can('lock') && !this.can('bypassall')) this.settings.blockPMs = {all: false};
```
Zarel, line 1072:
```suggestion
		if (this.settings.blockPMs && this.can('lock') && !this.can('bypassall')) {
			this.settings.blockPMs = {all: false};
		}
```

server/chat-commands/core.ts
```diff
@@ -733,10 +784,10 @@ export const commands: ChatCommands = {
 			for (const setting in user.settings) {
 				if (setting in raw) {
 					if (setting === 'blockPMs' &&
-						Users.Auth.isAuthLevel(raw[setting])) {
+						Users.Auth.isAuthLevel(raw[setting].all)) {
 						settings[setting] = raw[setting];
 					} else {
-						settings[setting as keyof UserSettings] = !!raw[setting];
+						settings[setting as keyof UserSettings] = typeof raw[setting] === 'object' ? raw[setting] : !!raw[setting];
```
scheibo, line 790: this does not work. this is the exact same problem. You are not guarding against bad input here.

scheibo: `{blockPMs: {foo: 'bar'}}` will pass through here.

mia-pi-git: the question then, though, is how do you compare it? i'm planning to make a few more things an object and hardcoding just isn't practical.

mia-pi-git: Actually, on testing, bad input doesn't crash or anything - maybe it'd be better to validate in their respective setting checkers?

mia-pi-git: like, in canPM, if blockPMs.all isn't in Config.groupsranking and it's not a boolean, why not just clear it?

Zarel: I would rather validate on input than on use. You would have to set the type to `unknown` if you want to validate on use, and TypeScript basically makes it as unergonomic as possible. (TypeScript doesn't even let you use `a?.b` on things with type `unknown`!)

mia-pi-git: So.... hardcoding it is, then?

scheibo: The alternative to "hardcoding" is creating some sort of generic typesafe JSON schema based validator which is the equivalent of using a sledgehammer to open a walnut (and specifying such schema for a basic type like settings would likely be both longer and more complicated for people to understand).

server/chat-commands/core.ts
```diff
@@ -732,10 +783,14 @@ export const commands: ChatCommands = {
...
-						Users.Auth.isAuthLevel(raw[setting])) {
-						settings[setting] = raw[setting];
+					if (setting === 'blockPMs') {
+						const {all, specific} = raw[setting];
+						settings.blockPMs = {
+							all: typeof all === 'boolean' || Users.Auth.isAuthLevel(all) ? all : false,
+							specific: specific && Array.isArray(specific) ? specific : [],
+						};
 					} else {
+						// @ts-ignore (the only exception is parsed above, so this can be ignored)
```
scheibo, line 793: What is the error here? You should not need to ignore anything

mia-pi-git: > Type 'boolean' is not assignable to type '(false & { all: boolean | "!" | "~" | "&" | "#" | "★" | "*" | "@" | "%" | "☆" | "+" | " " | "‽" | "whitelist" | "unlocked" | "trusted" | "autoconfirmed"; specific?: string[] | undefined; }) | (true & { ...; }) | undefined'.

mia-pi-git: it was previously fine with the typeof .... === object fix, but that's off the table

scheibo: use `as keyof Omit<UserSettings, 'blockPMs'>`: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

server/chat-commands/core.ts
mia-pi-git: Yeah, fair. 
This way at least we get to tailor it more specifically to each setting.

server/chat.ts
```diff
@@ -1141,6 +1116,53 @@ export class CommandContext extends MessageContext {
...
 		return message;
 	}
+	canPM(targetUser: User) {
+		const user = this.user;
+
+		// for hotpatching, remove later.
+		Chat.updateSettings(user);
+		Chat.updateSettings(targetUser);
+
+		function checkBlocked(curUser: User, tarUser: User) {
```
scheibo, line 1126: i do not like this nested function - this looks like it can be lifted to the top level (and then `tarUser` can be renamed to `targetUser` etc when its no longer shadowing). if this has to be inline, prefer lambda notation, `function` lifts names to the upper scope

server/users.ts
```diff
@@ -307,7 +307,10 @@ type ChatQueueEntry = [string, RoomID, Connection];
 
 export interface UserSettings {
 	blockChallenges: boolean;
-	blockPMs: boolean | AuthLevel;
+	blockPMs: {
```
scheibo, line 310: is this the correct type? can someone specify `all` and `specific` (as is currently the case with this type?) is
`{all: boolean | AuthLevel} | {specific?: string[]}` more correct?

Alternatively, `boolean | AuthLevel | string[]` seems like it would work? (ie/ things are specific when theyre an array, otherwise theyre all)

mia-pi-git: Currently this should allow you to specify all (block pms from all users by default not that rank) and allow you to also block specific users - for example, if i did /blockpms autoconfirmed then also added another autoconfirmed user to my blocklist. I feel it'd be nicer to have both at once, but if you can prefer i can remove the hotpatching hack.

scheibo: Oh, I didnt realize it was intended to let you do both. That makes sense, and thus this type signature makes sense. Sorry! This can be resolved :)

server/chat.ts
mia-pi-git: I went with inline, so it's easier to edit.

server/chat.ts
```diff
@@ -2022,6 +2035,16 @@ export const Chat = new class {
 		return ' ' + message.toLowerCase() + ' ';
 	}
 
+	/**
+	 * JUST FOR HOTPATCHING. To be removed later.
```
scheibo, line 2039: I thought about this - I don't think this actually can be removed later? `blockPMs: true` is stored in client's localStorage now, and old clients are going to still be sending this? I don't think its reasonable to remove this without a sufficiently long sunset period?

server/users.ts
```diff
@@ -896,6 +901,7 @@ export class User extends Chat.MessageContext {
 			...this.settings,
 			// Battle privacy state needs to be propagated in addition to regular settings so that the
 			// 'Ban spectators' checkbox on the client can be kept in sync (and disable privacy correctly)
+			// blockPMs needs to be made into a string
```
scheibo, line 904: is... this the right place for this comment?

server/chat.ts
```diff
@@ -1141,6 +1116,53 @@ export class CommandContext extends MessageContext {
 
 		return message;
 	}
+	canPM(targetUser: User) {
+		const user = this.user;
+
+		// for hotpatching, remove later.
+		const curSettings = Chat.updateSettings(user); // will just be user.settings
+		const tarSettings = Chat.updateSettings(targetUser); // will just be targetUser.settings
```
scheibo, line 1124:
```suggestion
		const userSettings = Chat.updateSettings(user);
		const targetUserSettings = Chat.updateSettings(targetUser);
```

server/chat.ts
```diff
@@ -1141,6 +1116,53 @@ export class CommandContext extends MessageContext {
...
+
+		if (targetUser.locked && !user.can('lock')) {
+			this.errorReply(`The user "${targetUser.name}" is locked and cannot be PMed.`);
+			return null;
+		}
+		if (
```
scheibo, line 1130: I seem to recall @Zarel not liking braces to be styled this way (and below - though I could very well be misremembering, if so, I apologize)?

Zarel: No, this is fine. What I don't like is if they don't match (i.e. there should be a newline after `(` if and only if there's a newline before the matching `)`)

Zarel:
At this point, I would rather this made requires-restart so it doesn't need the super ugly conversion process.

You'll still need a conversion process for client messages, but let's forget about the hotpatching craziness.

server/chat.ts
```diff
@@ -1141,6 +1116,53 @@ export class CommandContext extends MessageContext {
...
+				return null;
+			}
+		}
+		if (
+			(curSettings.all === true ||
+			!Users.globalAuth.atLeast(user, curSettings.all as GroupSymbol) && typeof tarSettings.all === 'string' ||
+			curSettings.specific?.includes(user.id)) &&
+			!targetUser.can('lock')
+		) {
+			const isBlocked = user.settings.blockPMs.specific?.includes(targetUser.id);
```
scheibo, line 1160: Oh, this might be completely out of scope for this change (though I remembered while reading this), but one feature request I would super super super love is that if I as a global administrator/sysop who bypassesall/skips can etc PMs someone but my PMs are blocked so they can't respond, could you please at least give me a friendly warning that the user wont be able to respond? This would be huge for me

server/chat-commands/core.ts
```diff
@@ -642,7 +642,58 @@ export const commands: ChatCommands = {
...
+			if (index < 0) return this.errorReply(`The user "${target}" is not on your PM blocklist.`);
+			specific.splice(index, 1);
+			user.update();
+			this.sendReply(`You have removed the user "${target}" from your PM blocklist.`);
+			return this.parse(`/pmblocklist view`);
+		},
+		view(target, room, user) {
+			Chat.updateSettings(user);
+
+			if (!user.settings.blockPMs.specific) user.settings.blockPMs.specific = [];
```
scheibo, line 683: Pretty sure this will mean everyones type looks like `{all: ..., specific: []}` anyway as things converge - nothing removes `specific: []` AFAICT, so once this is ever assigned on a user object it will pretty much last forever (because the client will persist it and propagate it across the wire.

The point being - might as well just stop making `specific` optional because in practice it will always exist eventually anyway

server/chat.ts
Zarel: I vaguely remember recommending making the check a method so it can be symmetric. I would much rather `bypassall` just not bypass PM permission checks asymmetrically.

scheibo:
I'm pretty confident this requires restart (or gross monkeypatching) because it changes `User` in a non trivial way?

Also, dont we need to land the client side change here first?

mia-pi-git:
Does this necessarily need a client change, unless i'm missing something? 
Can't /updatesettings just handle the case of it not being an object? (in fact, doesn't it atm?)

scheibo:
> Does this necessarily need a client change, unless i'm missing something?
> Can't /updatesettings just handle the case of it not being an object?

You are sending an object to the client now where it expects a boolean (and will be sending back a boolean). You have handled the case in the server code where the client sends a boolean (though your comment implied it was temporary), have you confirmed the client side code can handle an object?

server/chat.ts
mia-pi-git: So should i not let admins bypass the "you are blocking pms and so can't pm x user check" then? 
(If i'm understanding correctly)

Zarel: Yes, you should not let admins bypass it at all.

mia-pi-git: Okay.

mia-pi-git:
@Zarel bump?

Zarel:
Sorry, I'm going to need to look into the usersettings thing first. I'll be able to handle this after that.

mia-pi-git:
Bump again, @Zarel?

Zarel:
I'm going to punt on this for a little on the grounds that we should clean out `room.settings` before adding more state to RAM. Bump again in a few weeks?

mia-pi-git:
Closing this since it's no longer necessary (/ignore now persists and is effectively the same as this.)

PR #7195, opened 2020-08-10
Introduce async generator methods for readable streams

Kaiepi:
In some parts of PS, while loops combined with await are used to read
chunks of data from a stream sequentially. Async generators offer `for
await ... of` syntax, which can be used instead of this pattern. This
allows these loops to exist without the need to keep mutable state in
most cases and doesn't annoy TSLint.

### DIFF
--- a/lib/STREAMS.md
+++ b/lib/STREAMS.md
@@ -132,3 +132,19 @@ Can return synchronously. Use `await` or wrap the return value in `Promise.resol
 ### readStream.peekBuffer(byteCount)
 
 Like `readStream.read` and `readStream.peek`, but returns a Buffer instead of a string.
+
+### readStream.peeker([encoding])
+### readStream.reader([encoding])
+### readStream.delimitedReader(delimiter, [encoding])
+### readStream.lineReader([encoding])
+
+Returns an async generator that yields values returned by
+`readStream.peek`/`readStream.read`/`readStream.readDelimitedBy`/`readStream.readLine`
+until they return null or undefined. This can be iterated over with `for await
+... of` syntax:
+
+```js
+for await (const line of stream.lineReader()) {
+	console.log(line);
+}
+```
--- a/lib/streams.ts
+++ b/lib/streams.ts
@@ -307,6 +307,27 @@ export class ReadStream {
 		return line;
 	}
 
+	peeker(encoding: BufferEncoding = this.encoding) {
+		return this.sequence<string>('peek', encoding);
+	}
+	reader(encoding: BufferEncoding = this.encoding) {
+		return this.sequence<string>('read', encoding);
+	}
+	delimitedReader(symbol: string, encoding: BufferEncoding = this.encoding) {
+		return this.sequence<string>('readDelimitedBy', symbol, encoding);
+	}
+	lineReader(encoding: BufferEncoding = this.encoding) {
+		return this.sequence<string>('readLine', encoding);
+	}
+	protected async *sequence<T>(name: keyof ReadStream, ...args: any[]): AsyncGenerator<T, void, void> {
+		const method = this[name] as (...a: typeof args) => Promise<T | null>;
+		while (true) {
+			const result = await method.call(this, ...args);
+			if (result === null) return;
+			yield result;
+		}
+	}
+
 	destroy() {
 		this.atEOF = true;
 		this.bufStart = 0;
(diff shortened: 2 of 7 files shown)


AnnikaCodes:
This can also be used in `Modlog#readRoomModlog`.

lib/STREAMS.md
```diff
@@ -132,3 +132,19 @@ Can return synchronously. Use `await` or wrap the return value in `Promise.resol
 ### readStream.peekBuffer(byteCount)
 
 Like `readStream.read` and `readStream.peek`, but returns a Buffer instead of a string.
+
+### readStream.peeker([encoding])
```
Zarel, line 136: Is `peeker` ever necessary? I can't imagine any use-case where you'd want `peek` to be a stream.

lib/STREAMS.md
```diff
@@ -132,3 +132,19 @@ Can return synchronously. Use `await` or wrap the return value in `Promise.resol
...
+### readStream.reader([encoding])
```
Zarel, line 137: I've been considering renaming the regular read methods to `readChunk` etc, so it's clearer that if you're not doing a delimited read, the boundaries are meaningless.

Zarel: On the other hand, for `ObjectReadStream`s, you can probably just do `for await (const chunk of stream)` directly, skipping the `.reader()`.

lib/STREAMS.md
```diff
@@ -132,3 +132,19 @@ Can return synchronously. Use `await` or wrap the return value in `Promise.resol
...
+### readStream.delimitedReader(delimiter, [encoding])
```
Zarel, line 138: Something that might be extra nice would be to call this `split` and have it return an `ObjectReadStream<string>`...

lib/STREAMS.md
Kaiepi: If one of the `read` methods were to be used within the loop, I think it could be useful for things like binary streams. I can't think of a case offhand where it would be useful for text, though.

lib/STREAMS.md
Kaiepi: I think `readChunk` would be clearer, yeah. I'm not sure how I feel about `for await (const chunk of stream)` for `ObjectReadStream` only though...it would make the API less consistent, but at the same time `reader` doesn't share a type between the two anyway.

Kaiepi:
I'm wondering if this could be expanded a bit. I'm thinking something like:

- Add unit tests for streams
- Type the options objects passed to constructors
- Moving `_read`/`_write`/etc. from `*Stream.prototype` to an `config` property and make them always return a `Promise<void>` so they can be awaited from the methods that use them
- Add back the `peeker` and `reader` methods, replacing `delimitedReader`/`lineReader` with `split`

lib/STREAMS.md
Zarel: Hm, I guess I can envision something like that. I still think using a stream for that kind of use-case is really unreadable, especially since it's not just streaming the entire data, but only the parts you don't read. I think a while-loop would make it much more readable so I'd rather enforce that by not supporting a peek stream.

lib/STREAMS.md
Zarel: I would also call it `chunkReader` to be in line with `readChunk`. I think reading an arbitrary (or fixed-size) chunk from a ReadStream is just an entirely different operation from reading exactly one element from an ObjectReadStream.

Zarel:
Anyway, since I haven't mentioned it yet, I'm a huge fan of `for await` support in general, and I think most of these changes are pushing in the right direction!

Zarel:
I thought PS streams were already async iterable, though. Did I implement the spec wrong?

Zarel:
Turns out I implemented it right. It just needs:

```
	[Symbol.asyncIterator]() {
		return this;
	}
```

The generator is unnecessary.

lib/streams.ts
```diff
@@ -643,6 +664,21 @@ export class ObjectReadStream<T> {
...
+		return this.sequence<string>('read');
+	}
+	protected async *sequence<U>(name: keyof ObjectReadStream<T>, ...args: any[]): AsyncGenerator<U, void, void> {
+		const method = this[name] as (...a: typeof args) => Promise<U | null>;
+		while (true) {
+			const result = await method.call(this, ...args);
+			if (result === null) return;
+			yield result;
+		}
+	}
```
Zarel, line 680:
```suggestion
	[Symbol.asyncIterator]() {
		return this;
	}
```

PR #7196, opened 2020-08-11
Improvements to the /dnd command

AnnikaCodes:
I don't want my challenges to be unblocked when I use `/back` if I was just on regular `/busy`, not dnd (I have challenges autoblocked on login). Also, I made `dnd` its own status type.


Zarel:
Hm, idk how I feel about this being a `statusType`. Maybe just another user setting?

mia-pi-git:
Why does it need to be either? the intent is pretty much "i want this command to/not to block these settings" - busy as a statusType already says "i'm busy."

AnnikaCodes:
I'd rather have a user setting than an alias to `/back`; in my mind, `/back` is supposed to mean "undo the effects of the last away/dnd/busy, and I think having a `/nodnd` or whatever is a bit convoluted. I can do it if you want (I have no qualms personally), but it seems rather complex for the end user (as well as making the concept of /back more complicated).

Zarel:
I agree that Annika's system more closely maps to how users expect it to work.

Zarel:
My issue is that Annika's system doesn't affect the protocol which is what `statusType` is designed for, which makes me suspect that `statusType` is the wrong place for it.

Zarel:
For the record, I still would rather it be a user setting than a status type.

AnnikaCodes:
Okay; I kinda forgot about this. I'll try to get to it soon.

Zarel:
Thanks!

PR #7198, opened 2020-08-11
Add support for inverse in /weak

Spandamn:
image

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -822,12 +822,14 @@ export const commands: ChatCommands = {
 		const modName = target.split(',');
 		let mod = Dex;
 		let format: Format | null = null;
+		let isInverse = false;
 		if (modName[modName.length - 1] && toID(modName[modName.length - 1]) in Dex.dexes) {
 			mod = Dex.mod(toID(modName[modName.length - 1]));
 		} else if (room?.battle) {
 			format = Dex.getFormat(room.battle.format);
 			mod = Dex.mod(format.mod);
 		}
+		if (modName[modName.length - 1] && toID(modName[modName.length - 1]) === 'inverse') isInverse = true;
 		const targets = target.split(/ ?[,/] ?/);
 		let species: {types: string[], [k: string]: any} = mod.getSpecies(targets[0]);
 		const type1 = mod.getType(targets[0]);
@@ -860,8 +862,9 @@ export const commands: ChatCommands = {
 		const immunities = [];
 		for (const type in mod.data.TypeChart) {
 			const notImmune = mod.getImmunity(type, species);
-			if (notImmune) {
-				const typeMod = mod.getEffectiveness(type, species);
+			if (notImmune || isInverse) {
+				let typeMod = !notImmune && isInverse ? 1 : 0;
+				typeMod += (isInverse ? -1 : 1) * mod.getEffectiveness(type, species);
 				switch (typeMod) {
 				case 1:
 					weaknesses.push(type);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -819,18 +819,22 @@ export const commands: ChatCommands = {
 		if (!target) return this.parse('/help weakness');
 		if (!this.runBroadcast()) return;
 		target = target.trim();
-		const modName = target.split(',').length > 1 ? target.split(',').pop() : null;
+		const targets = target.split(',').map(toID);
+		const maybeMod = targets[targets.length - 1];
 		let mod = Dex;
 		let format: Format | null = null;
 		let isInverse = false;
-		if (modName && toID(modName) in Dex.dexes) {
-			mod = Dex.mod(toID(modName));
+		if (maybeMod && maybeMod in Dex.dexes) {
+			mod = Dex.mod(maybeMod);
+			targets.pop();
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.ts
```diff
@@ -822,12 +822,14 @@ export const commands: ChatCommands = {
...
 		let mod = Dex;
 		let format: Format | null = null;
+		let isInverse = false;
 		if (modName[modName.length - 1] && toID(modName[modName.length - 1]) in Dex.dexes) {
 			mod = Dex.mod(toID(modName[modName.length - 1]));
 		} else if (room?.battle) {
 			format = Dex.getFormat(room.battle.format);
 			mod = Dex.mod(format.mod);
 		}
+		if (modName[modName.length - 1] && toID(modName[modName.length - 1]) === 'inverse') isInverse = true;
```
AnnikaCodes, line 832: I'd use an intermediate variable for `modName[modName.length - 1]`.

server/chat-commands/info.ts
```diff
@@ -822,12 +822,14 @@ export const commands: ChatCommands = {
 		const modName = target.split(',');
```
Zarel, line 822: The `modName` and `targets` stuff looks pretty complicated. I suspect they can be combined into one thing, though.

Spandamn:
nah not yet

server/chat-commands/info.ts
```diff
@@ -819,15 +819,17 @@ export const commands: ChatCommands = {
 		if (!target) return this.parse('/help weakness');
 		if (!this.runBroadcast()) return;
 		target = target.trim();
-		const modName = target.split(',');
+		const modName = target.split(',').pop();
```
Zarel, line 822: This will cause the command to fail if you don't specify a modname.

Spandamn:
Done @Zarel

server/chat-commands/info.ts
```diff
@@ -819,15 +819,17 @@ export const commands: ChatCommands = {
 		if (!target) return this.parse('/help weakness');
 		if (!this.runBroadcast()) return;
 		target = target.trim();
-		const modName = target.split(',');
+		const modName = target.split(',').length > 1 ? target.split(',').pop() : null;
```
Zarel, line 822: This will _still_ cause the command to fail if you specify two things that don't include a modname.

For instance, `/weak fire, fighting`

Zarel:
Drafting because the modname support is still broken.

server/chat-commands/info.ts
Spandamn: okay ill try another way

server/chat-commands/info.ts
```diff
@@ -819,15 +819,17 @@ export const commands: ChatCommands = {
...
-			mod = Dex.mod(toID(modName[modName.length - 1]));
+		let isInverse = false;
+		if (modName && toID(modName) in Dex.dexes) {
+			mod = Dex.mod(toID(modName));
 		} else if (room?.battle) {
 			format = Dex.getFormat(room.battle.format);
 			mod = Dex.mod(format.mod);
 		}
+		if (modName && toID(modName) === 'inverse') isInverse = true;
 		const targets = target.split(/ ?[,/] ?/);
```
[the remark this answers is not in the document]
Zarel: Oh. I think the command is still super unreadable, so let me just tidy it up myself, I guess.

server/chat-commands/info.ts
Zarel, line 822:
```suggestion
		const targets = target.split(',').map(t => t.trim());
		const maybeModName = targets[targets.length - 1];
```

server/chat-commands/info.ts
```diff
@@ -819,15 +819,17 @@ export const commands: ChatCommands = {
...
 		target = target.trim();
-		const modName = target.split(',');
+		const modName = target.split(',').length > 1 ? target.split(',').pop() : null;
 		let mod = Dex;
 		let format: Format | null = null;
-		if (modName[modName.length - 1] && toID(modName[modName.length - 1]) in Dex.dexes) {
-			mod = Dex.mod(toID(modName[modName.length - 1]));
+		let isInverse = false;
+		if (modName && toID(modName) in Dex.dexes) {
+			mod = Dex.mod(toID(modName));
```
Zarel, line 827:
```suggestion
		if (maybeMod && maybeMod in Dex.dexes) {
			mod = Dex.mod(maybeMod);
			targets.pop();
```

server/chat-commands/info.ts
```diff
@@ -819,15 +819,17 @@ export const commands: ChatCommands = {
...
 		} else if (room?.battle) {
 			format = Dex.getFormat(room.battle.format);
 			mod = Dex.mod(format.mod);
 		}
+		if (modName && toID(modName) === 'inverse') isInverse = true;
```
Zarel, line 832:
```suggestion
		if (maybeMod === 'inverse') {
			isInverse = true;
			targets.pop();
		}
```

server/chat-commands/info.ts
Zarel, line 833:
```suggestion
		if (maybeMod === 'inverse') {
			isInverse = true;
			targets.pop();
		}
```

Zarel:
Thanks!

PR #7200, opened 2020-08-12
Fix modlog crash

AnnikaCodes:


server/chat-plugins/chat-monitor.ts
```diff
@@ -20,7 +21,30 @@ interface Monitor {
...
+	 * Each key in fields represents an attribute of a User object.
+	 * The attribute's value must be in the provided array for the monitor to activate.
+	 */
+	fields?: {[k: string]: any[]};
+	/** Custom function that must return true if specified to activate the monitor. */
+	checker?: (user: User) => boolean;
+}
+
+const EVADER_MONITOR_FILE = '../../config/monitors.js';
+const CHAT_FILTER_FILE = 'config/chat-plugins/chat-monitor.tsv';
```
monsanto, line 47: I take it you renamed to `CHAT_FILTER_FILE` to reduce confusion with the evader monitor. Is there a plan to rename the tsv, too? Otherwise, it might make sense to stay consistent with the established naming.

AnnikaCodes: I can rename the tsv too; it'd just make things a bit more complex since someone would need to rename it on the server before hotpatching.

PR #7202, opened 2020-08-12
Hosts: Fix bug with /ipranges widen

AnnikaCodes:
We need to actually remove ranges that widen is replacing.
(This requires an IPTools hotpatch as well as chat.)


Zarel:
Thanks!

PR #7203, opened 2020-08-12
Support persisting statuses

mia-pi-git:
Widely suggested. 
Still unsure about the validation in /updatesettings, since hardcoding is annoying, but from the discussion in #7191, that seems to be the way to go.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -28,7 +28,7 @@
  * @license MIT
  */
 
-type StatusType = 'online' | 'busy' | 'idle';
+export type StatusType = 'online' | 'busy' | 'idle';
 
 const THROTTLE_DELAY = 600;
 const THROTTLE_DELAY_TRUSTED = 100;
@@ -308,6 +308,10 @@ type ChatQueueEntry = [string, RoomID, Connection];
 export interface UserSettings {
 	blockChallenges: boolean;
 	blockPMs: boolean | AuthLevel;
+	status: {
+		type: StatusType,
+		message: string,
+	};
 	ignoreTickets: boolean;
 	hideBattlesFromTrainerCard: boolean;
 }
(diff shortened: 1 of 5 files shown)


server/users.ts
```diff
@@ -308,6 +308,10 @@ type ChatQueueEntry = [string, RoomID, Connection];
 export interface UserSettings {
 	blockChallenges: boolean;
 	blockPMs: boolean | AuthLevel;
+	status: {
+		type: StatusType,
+		message: string,
+	};
```
Zarel, line 314: It seems to me like you would have less code if this was `statusType` and `statusMessage`.

mia-pi-git: thought it would be cleaner to handle it similar to #7191, but I can move it to just statusType and statusMessage if you'd prefer.

mia-pi-git: (also imo it's slightly cleaner to parse in /updatesettings, since it requires one "else if" as an object, not two, but that's just me)

mia-pi-git: Actually, on reflection, i agree - i'll do this now

translations/portuguese.json
```diff
@@ -117,7 +117,7 @@
 		"This room has slow-chat enabled. You can only talk once every ${time} seconds.":
 		"O slow-chat desta sala está ativado. Você apenas pode falar a cada ${time} segundos.",
 		"Your username contains a phrase banned by this room.": "Seu nome de usuário contém uma frase banida por esta sala.",
-		"Your status message contains a phrase banned by this room.": "A mensagem do seu status contém uma frase banida por esta sala.",
+		"Your statusMessage contains a phrase banned by this room.": "A mensagem do seu status contém uma frase banida por esta sala.",
```
Zarel, line 120:
```suggestion
		"Your statusMessage contains a phrase banned by this room.": "A mensagem do seu status contém uma frase banida por esta sala.",
```

I'm guessing you did a regex search/replace `status.message`? You should have just turned it to plaintext whole-word mode...

mia-pi-git: Ugh.

mia-pi-git: Fixed. My bad.

AnnikaCodes:
Please fix checks and merge conflicts.

PR #7204, opened 2020-08-12
Fix getBattles skip behavior wrt private rooms

scheibo:
Rooms.global.battleCount considers both public and private battles,
but getBattles already passes over isPrivate battles, meaning if a
non-trivial number of battles are private, battleCount is much
larger than the number of public battles and all battles will be
skipped resulting in an empty "Watch Battles" tab for "(all
formats)".

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1111,14 +1111,10 @@ export class GlobalRoomState {
 		return Config.rankList;
 	}
 
-	getBattles(/** "formatfilter, elofilter, usernamefilter */ filter: string) {
-		const rooms: GameRoom[] = [];
-		let skipCount = 0;
+	getBattles(/** formatfilter, elofilter, usernamefilter */ filter: string) {
+		let rooms: GameRoom[] = [];
 		const [formatFilter, eloFilterString, usernameFilter] = filter.split(',');
 		const eloFilter = +eloFilterString;
-		if (this.battleCount > 150 && !formatFilter && !eloFilter && !usernameFilter) {
-			skipCount = this.battleCount - 150;
-		}
 		for (const room of Rooms.rooms.values()) {
 			if (!room || !room.active || room.settings.isPrivate) continue;
 			if (room.type !== 'battle') continue;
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -1130,11 +1126,13 @@ export class GlobalRoomState {
...
 			}
-			if (skipCount && skipCount--) continue;
-
 			rooms.push(room);
 		}
 
+		if (rooms.length > 150 && !formatFilter && !eloFilter && !usernameFilter) {
+			rooms = rooms.slice(-150);
+		}
+
```
Zarel, line 1135:
```suggestion
```

The original cap of 150 was for performance reasons. This line doesn't actually impact performance.

server/rooms.ts
```diff
@@ -1111,14 +1111,10 @@ export class GlobalRoomState {
 		return Config.rankList;
 	}
 
-	getBattles(/** "formatfilter, elofilter, usernamefilter */ filter: string) {
-		const rooms: GameRoom[] = [];
-		let skipCount = 0;
+	getBattles(/** formatfilter, elofilter, usernamefilter */ filter: string) {
+		let rooms: GameRoom[] = [];
```
Zarel, line 1115:
```suggestion
		const rooms: GameRoom[] = [];
```

server/rooms.ts
```diff
@@ -1130,11 +1126,13 @@ export class GlobalRoomState {
...
-
 			rooms.push(room);
 		}
 
+		if (rooms.length > 150 && !formatFilter && !eloFilter && !usernameFilter) {
+			rooms = rooms.slice(-150);
+		}
+
 		const roomTable: {[roomid: string]: BattleRoomTable} = {};
 		for (let i = rooms.length - 1; i >= rooms.length - 100 && i >= 0; i--) {
```
Zarel, line 1137: Specifically, what the code was doing was searching the most recent 150 battles, under the assumption that it will contain at least 100 public battles. The actual code to grab 100 battles is here.

Zarel:
Good catch, though!

server/rooms.ts
scheibo: Yeah, I understood there would be performance implications but didnt notice there was a second level of filtering below . Do we want to just increase the 150 to like 1500 or something more conservative (and leave a comment explaining where the number came from...)? On main I was noticing like ~6000 rooms, what motivated this additional skip optimization?

Zarel: I used to do a lot of premature optimization. It's probably unnecessary.

PR #7206, opened 2020-08-12
/showteam: Fix bugs

mia-pi-git:
There was an extraneous ``<`` that cut off the first line, when using !showteam.
Additionally, this now formats natures better.


Zarel:
tyty

PR #7207, opened 2020-08-13
Help: Only write state when relevant to the room

mia-pi-git:
Previously this would write state whenever a faq was changed anywhere - an oversight on my part. 
now this checks to ensure that the roomfaq object has the faq - if not, it deletes the data for that faq and writes state. 
This DOES complicate testing, but the way i went with is asserting that nofswriting is off, since that (mostly) eliminates the point of the step with it on,

### DIFF
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -180,15 +180,19 @@ export class HelpResponder {
 			return this.roomFaqs;
 		}
 		const roomid = room.roomid;
+		let hasDeletion = false;
 		this.roomFaqs = JSON.parse(FS(ROOMFAQ_FILE).readIfExistsSync() || `{"${roomid}":{}}`)[roomid];
 		for (const key in this.data.pairs) {
-			if (!this.roomFaqs[key]) delete this.data.pairs[key];
+			if (!this.roomFaqs[key]) {
+				delete this.data.pairs[key];
+				hasDeletion = true;
+			}
 		}
-		this.writeState();
+		if (hasDeletion) this.writeState();
 		return this.roomFaqs;
 	}
 	tryAddRegex(inputString: string, raw?: boolean) {
-		let [args, faq] = inputString.split('=>');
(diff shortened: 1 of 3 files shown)


server/chat-plugins/help.ts
```diff
@@ -144,6 +139,14 @@ export class HelpResponder {
 	}
 	match(question: string, faq: string) {
 		if (!this.data.pairs[faq]) this.data.pairs[faq] = [];
+		const room = this.getRoom();
+		if (!room) return;
+		// mostly for the purpose of tests
```
Zarel, line 144: ?

mia-pi-git: this is because Peach suggested a better way to handle deleting data tied to RFAQs, but the problem is tests fail since I can't edit the roomfaq object in tests in the way it'd need to be edited, nor can I rely on just being able to make the data, so I had to do some fuckery to try to get around it.

Zarel: ...I have no clue how this stuff all works so I guess I'll just accept this once the tests pass.

test/server/chat-plugins/help.js
```diff
@@ -15,10 +15,17 @@ const Help = new Responder({
 
 describe('Help', function () {
 	it('should only return true on added regexes', function () {
+		const roomFaqs = require('../../../.server-dist/chat-plugins/room-faqs').roomFaqs;
+		const room = Help.getRoom();
+		if (room) { // if there's no room, this will fail, but if there's no room this test doesn't matter anyway
```
whalemer, line 20: unit tests don't use the normal room data (or config data), they have their own clean copies of lobby and staff. this room will never exist.

mia-pi-git:
I still think an FS().onModify watch would be a better way to handle this, but in lieu of that, this should work fine, given that the data for deleted faqs becomes inaccessible when the faq is deleted, with the new implementation of looking directly at the roomfaq object. 
So, this should be good then, @Zarel.

Zarel:
@_@

PR #7209, opened 2020-08-14
Show error when using ! on a non-broadcastable command

mia-pi-git:
https://www.smogon.com/forums/threads/rank-not-showing-on-both-sides-and-in-replays.3668496/#post-8565216

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1653,7 +1653,7 @@ export const Chat = new class {
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
 			// assign SOME properties from the base command if the current command uses CommandContext.run.
-			const runsCommand = /this.run\(('|"|`)(.*)('|"|`)\)/.exec(handlerCode);
+			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
 			if (runsCommand) {
 				const [, , baseCommand] = runsCommand;
 				const baseEntry = commandTable[baseCommand];
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1652,7 +1652,7 @@ export const Chat = new class {
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
-			// assign SOME properties from the base command if the current command uses CommandContext.run.
+			// assign properties from the base command if the current command uses CommandContext.run.
 			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
 			if (runsCommand) {
 				const [, baseCommand] = runsCommand;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1658,9 +1658,9 @@ export const Chat = new class {
 				const [, baseCommand] = runsCommand;
 				const baseEntry = commandTable[baseCommand];
 				if (baseEntry) {
-					entry.requiresRoom = baseEntry.requiresRoom;
-					entry.hasRoomPermissions = baseEntry.hasRoomPermissions;
-					entry.broadcastable = baseEntry.broadcastable;
+					if (baseEntry.requiresRoom) entry.requiresRoom = baseEntry.requiresRoom;
+					if (baseEntry.hasRoomPermissions) entry.hasRoomPermissions = baseEntry.hasRoomPermissions;
+					if (baseEntry.broadcastable) entry.broadcastable = baseEntry.broadcastable;
 				}
 			}
 
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1655,7 +1655,7 @@ export const Chat = new class {
 			// assign SOME properties from the base command if the current command uses CommandContext.run.
 			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
 			if (runsCommand) {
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -1647,6 +1652,13 @@ export const Chat = new class {
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
+			// assign SOME properties from the base command if the current command uses CommandContext.run.
+			const runsCommand = /this.run\(('|"|`)(.*)('|"|`)\)/.exec(handlerCode);
+			if (runsCommand) {
+				const [, , baseCommand] = runsCommand;
+				if (commandTable[baseCommand]) Object.assign(entry, commandTable[baseCommand]);
```
Zarel, line 1659: That's not just "some", that's all of them. I'm surprised this doesn't crash, actually; I guess the read-only ones are enumerable. But you should copy them one by one instead of using `Object.assign`.

server/chat.ts
```diff
@@ -1647,6 +1652,13 @@ export const Chat = new class {
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
+			// assign SOME properties from the base command if the current command uses CommandContext.run.
+			const runsCommand = /this.run\(('|"|`)(.*)('|"|`)\)/.exec(handlerCode);
+			if (runsCommand) {
+				const [, , baseCommand] = runsCommand;
```
Zarel, line 1658:
```suggestion
			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
			if (runsCommand) {
				const [, baseCommand] = runsCommand;
```

server/chat.ts
```diff
@@ -1647,6 +1652,18 @@ export const Chat = new class {
...
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
+			// assign SOME properties from the base command if the current command uses CommandContext.run.
+			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
+			if (runsCommand) {
+				const [, , baseCommand] = runsCommand;
+				const baseEntry = commandTable[baseCommand];
+				if (baseEntry) {
+					entry.requiresRoom = baseEntry.requiresRoom;
+					entry.hasRoompermissions = baseEntry.hasRoomPermissions;
```
Zarel, line 1662:
```suggestion
					entry.hasRoomPermissions = baseEntry.hasRoomPermissions;
```

`entry` having type `any` means you need to be extra careful here.

server/chat.ts
```diff
@@ -1647,6 +1652,18 @@ export const Chat = new class {
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
+			// assign SOME properties from the base command if the current command uses CommandContext.run.
+			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
+			if (runsCommand) {
+				const [, , baseCommand] = runsCommand;
```
Zarel, line 1658: Have you tested this? Shouldn't this need to be one comma and not two?

mia-pi-git: it's two, since the check for " or ', etc also returns a value in exec

Zarel: No it doesn't. That's what the `?:` part prevents. You _definitely_ didn't test this because I tested it just now.

Zarel:
```
> [.../this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(`this.run('test')`)]
["this.run('test')", "test"]
```

mia-pi-git: oh, okay then - my bad, sorry

server/chat.ts
```diff
@@ -1647,6 +1652,18 @@ export const Chat = new class {
 			entry.hasRoomPermissions = /\bthis\.can\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = /\bthis\.(?:canBroadcast|runBroadcast)\(/.test(handlerCode);
 
+			// assign SOME properties from the base command if the current command uses CommandContext.run.
```
Zarel, line 1655:
```suggestion
			// assign properties from the base command if the current command uses CommandContext.run.
```

server/chat.ts
```diff
@@ -1647,6 +1652,18 @@ export const Chat = new class {
...
 
+			// assign properties from the base command if the current command uses CommandContext.run.
+			const runsCommand = /this.run\((?:'|"|`)(.*?)(?:'|"|`)\)/.exec(handlerCode);
+			if (runsCommand) {
+				const [, baseCommand] = runsCommand;
+				const baseEntry = commandTable[baseCommand];
+				if (baseEntry) {
+					entry.requiresRoom = baseEntry.requiresRoom;
+					entry.hasRoomPermissions = baseEntry.hasRoomPermissions;
+					entry.broadcastable = baseEntry.broadcastable;
```
Zarel, line 1663:
```suggestion
					entry.requiresRoom = baseEntry.requiresRoom;
					entry.hasRoomPermissions = baseEntry.hasRoomPermissions;
					entry.broadcastable = baseEntry.broadcastable;
```

These should all be `||=` but sadly sucrase can't compile it yet.

Zarel:
Thanks!

PR #7211, opened 2020-08-14
Modlog: Fix exact searches and improve hotpatching

AnnikaCodes:
Hotpatching modlog "properly" (i.e. reinitializing each stream) turns out to be far too slow in practice, causing the main server to lock up for >30 seconds. This will hopefully be quicker, albeit hackier.


Zarel:
Thanks!

scheibo:
What makes it hackier? Stream code potentially changing and causing issues?

AnnikaCodes:
> What makes it hackier? Stream code potentially changing and causing issues?

Yeah, if the backend was changed, the hotpatching code would also need to be changed.

PR #7212, opened 2020-08-15
Allow users locked for room punishments to PM ROs

AnnikaCodes:
As per [this suggestion](https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/page-15#post-8568758).

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -940,8 +940,8 @@ export class CommandContext extends MessageContext {
 		return true;
 	}
 	canPMLocked(user: User, lockedUser: User) {
-		const isPunishmentsAutolock = Punishments.userids.get(user.id)![3].includes('Autolocked for having punishments');
-		return user.can('lock') || !(
+		const isPunishmentsAutolock = Punishments.userids.get(lockedUser.id)![3].includes('Autolocked for having punishments');
+		return user.can('lock') || (
 			isPunishmentsAutolock &&
 			[...Rooms.rooms.values()].some(curRoom => !curRoom.settings.isPersonal && user?.can('editroom', null, curRoom))
 		);
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -940,7 +940,7 @@ export class CommandContext extends MessageContext {
 		return true;
 	}
 	canPMLocked(user: User, lockedUser: User) {
-		const isPunishmentsAutolock = Punishments.userids.get(lockedUser.id)![3].includes('Autolocked for having punishments');
+		const isPunishmentsAutolock = Punishments.userids.get(lockedUser.id)?.[3].includes('Autolocked for having punishments');
 		return user.can('lock') || (
 			isPunishmentsAutolock &&
 			[...Rooms.rooms.values()].some(curRoom => !curRoom.settings.isPersonal && user?.can('editroom', null, curRoom))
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -1006,8 +1007,15 @@ export class CommandContext extends MessageContext {
...
 			if (targetUser) {
-				if (lockType && !targetUser.can('lock')) {
-					this.errorReply(`You are ${lockType} and can only private message members of the global moderation team. ${lockExpiration}`);
+				if (
+					lockType &&
+					!targetUser.can('lock') &&
+					!(canPMRoomOwners && [...Rooms.rooms.values()].some(chatroom => {
+						if (chatroom.battle || chatroom.settings.isPersonal) return false;
+						return targetUser?.can('editroom', null, chatroom);
+					}))
```
Zarel, line 1016:
```suggestion
					!(canPMRoomOwners && [...Rooms.rooms.values()].some(curRoom => (
						!curRoom.settings.isPersonal &&  targetUser?.can('editroom', null, chatroom)
					)))
```

server/chat.ts
```diff
@@ -1006,8 +1007,14 @@ export class CommandContext extends MessageContext {
...
 			// own language these messages should also be translated. - Asheviere
 			if (targetUser) {
-				if (lockType && !targetUser.can('lock')) {
-					this.errorReply(`You are ${lockType} and can only private message members of the global moderation team. ${lockExpiration}`);
+				if (
+					lockType &&
+					!targetUser.can('lock') &&
+					!(canPMRoomOwners && [...Rooms.rooms.values()].some(curRoom => (
+						!curRoom.settings.isPersonal && targetUser?.can('editroom', null, curRoom)
+					)))
```
Zarel, line 1015: Oh, wait, this is one direction. The RO won't be able to reply since you haven't also edited:

image

I think you probably want a function for "can PM locked users" so this isn't duplicated.

AnnikaCodes: Oh. I'm not sure how I missed that. Anyway, it's fixed now.

server/chat.ts
```diff
@@ -939,6 +939,13 @@ export class CommandContext extends MessageContext {
 
 		return true;
 	}
+	canPMLocked(user: User, lockedUser: User) {
+		const isPunishmentsAutolock = Punishments.userids.get(lockedUser.id)![3].includes('Autolocked for having punishments');
```
Zarel, line 943:
```suggestion
		const isPunishmentsAutolock = Punishments.userids.get(lockedUser.id)?.[3].includes('Autolocked for having punishments');
```

Locked users are not necessarily in the Punishments table, from what I'm aware.

server/chat.ts
```diff
@@ -1006,12 +1013,13 @@ export class CommandContext extends MessageContext {
 			// and these PM-related messages aren't attached to any rooms. If we ever get to letting users set their
 			// own language these messages should also be translated. - Asheviere
 			if (targetUser) {
-				if (lockType && !targetUser.can('lock')) {
-					this.errorReply(`You are ${lockType} and can only private message members of the global moderation team. ${lockExpiration}`);
+				if (lockType && !this.canPMLocked(targetUser, user)) {
+					const canPMRoomOwners = Punishments.userids.get(user.id)![3].includes('Autolocked for having punishments');
```
Zarel, line 1017:
```suggestion
					const canPMRoomOwners = Punishments.userids.get(user.id)?.[3].includes('Autolocked for having punishments');
```

server/chat.ts
AnnikaCodes: huh

Zarel:
Unfortunately, Upper Staff has decided this is a bad idea.

The problem is that ROs don't have as many tools as globals do to defend themselves against PM harassment. In addition, ROs aren't empowered to unlock users, even if they caused the lock. A global staff member will have to be involved regardless.

PR #7214, opened 2020-08-15
Datasearch: Support excluding monotype Pokemon

AnnikaCodes:
Apparently this is something people want.


Zarel:
Apparently this is something people want. :p

PR #7215, opened 2020-08-15
Introduce chunk readers for ReadStream

Zarel:
This adds new functions `stream.byChunk(bytes)`, `stream.byLine()` etc which parse a `ReadStream` into an `ObjectReadStream<string>` which can then be consumed with for-await.

Fixes #7195


Zarel:
No reviews, so I guess I'll just merge and see if anything breaks.

PR #7216, opened 2020-08-16
Only allow the player who set modjoin to change it

AnnikaCodes:


server/chat-commands/room-settings.ts
```diff
@@ -175,6 +175,10 @@ export const commands: ChatCommands = {
 			if (prefix) {
 				return this.errorReply(`This battle is required to be public due to a player having a name prefixed by '${prefix}'.`);
 			}
+			if (room.battle.inviteOnlySetter && !user.can('mute', null, room) && room.battle.inviteOnlySetter !== user.id) {
+				return this.errorReply(`Only the person who set this battle to be invite-only can turn it off.`);
+			}
+			room.battle.inviteOnlySetter = user.can('mute', null, room) ? 'staff' : user.id;
```
Zarel, line 181:
```suggestion
			room.battle.inviteOnlySetter = user.id;
```

Is the `staff` key necessary here? `user.id` should be sufficient, right?

server/room-battle.ts
```diff
@@ -483,6 +483,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 	p2: RoomBattlePlayer;
 	p3: RoomBattlePlayer;
 	p4: RoomBattlePlayer;
+	inviteOnlySetter: ID | 'staff' | null;
```
Zarel, line 486:
```suggestion
	inviteOnlySetter: ID | null;
```

(assuming I wasn't wrong)

server/chat-commands/room-settings.ts
AnnikaCodes: I added it because battle timers use it, but I can't think of anything that removing it would break. Let's try without it?

Zarel:
Oh, actually, that's a good point; battle timers use it so it's clear that it's set by staff and unchangeable.

PR #7217, opened 2020-08-16
Slight HTML Button Improvements

PartMan7:
Changes this makes:
a) Lets people use `/pm` instead of `/msg` in buttons, which is a lot more intuitive.
b) Permits the author of the HTML to direct messages towards themselves via buttons (useful for when Bots create subroomgroupchats and become #, or if they suddenly want hugs from random people).

(Didn't add /w since line limits are worse than horizontally sliced bread)

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1293,17 +1293,17 @@ export class CommandContext extends MessageContext {
 					if ((!this.room || this.room.settings.isPersonal || this.room.settings.isPrivate === true) && !this.user.can('lock')) {
 						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
 						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
-						if (buttonName === 'send' && buttonValue?.startsWith('/msg ')) {
-							const [pmTarget] = buttonValue.slice(5).split(',');
+						if (buttonName === 'send' && (buttonValue?.startsWith('/msg ') || buttonValue?.startsWith('/pm '))) {
+							const [pmTarget] = buttonValue.replace(/^\/(?:msg|pm) /, '').split(',');
 							const auth = this.room ? this.room.auth : Users.globalAuth;
-							if (auth.get(toID(pmTarget)) !== '*') {
+							if (auth.get(toID(pmTarget)) !== '*' && toID(pmTarget) !== this.user.id) {
 								this.errorReply(`This button is not allowed: <${tagContent}>`);
 								this.errorReply(`Your scripted button can't send PMs to ${pmTarget}, because that user is not a Room Bot.`);
 								return null;
 							}
 						} else if (buttonName) {
 							this.errorReply(`This button is not allowed: <${tagContent}>`);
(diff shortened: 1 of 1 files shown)


Zarel:
Just make an intermediate variable. It's more readable that way, anyway. :|

server/chat.ts
```diff
@@ -1293,17 +1293,17 @@ export class CommandContext extends MessageContext {
 					if ((!this.room || this.room.settings.isPersonal || this.room.settings.isPrivate === true) && !this.user.can('lock')) {
 						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
 						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
-						if (buttonName === 'send' && buttonValue?.startsWith('/msg ')) {
-							const [pmTarget] = buttonValue.slice(5).split(',');
+						if (buttonName === 'send' && (buttonValue?.startsWith('/msg ') || buttonValue?.startsWith('/pm '))) {
+							const [pmTarget] = buttonValue.replace(/^\/(?:msg|pm) /, '').split(',');
```
Zarel, line 1297:
```suggestion
						const msgCommandRegex = /^\/(?:msg|pm|w) /;
						if (buttonName === 'send' && msgCommandRegex.test(buttonValue || '')) {
							const [pmTarget] = buttonValue.replace(msgCommandRegex, '').split(',');
```

Zarel:
Good change, though!

PartMan7:
Come to think of it, this should probably be case-insensitive.

server/chat.ts
PartMan7: Right, I didn't think of that. Thanks. n.n

server/chat.ts
```diff
@@ -1293,17 +1293,18 @@ export class CommandContext extends MessageContext {
 					if ((!this.room || this.room.settings.isPersonal || this.room.settings.isPrivate === true) && !this.user.can('lock')) {
 						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
 						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
-						if (buttonName === 'send' && buttonValue?.startsWith('/msg ')) {
-							const [pmTarget] = buttonValue.slice(5).split(',');
+						const msgCommandRegex = /^\/(?:msg|pm|w|whisper) /i;
+						if (buttonName === 'send' && msgCommandRegex.test(buttonValue || '')) {
+							const [pmTarget] = buttonValue.replace(msgCommandRegex, '').split(',');
```
mia-pi-git, line 1298:
```suggestion
						const msgCommandRegex = /^\/(?:msg|pm|w|whisper) /i;
						if (buttonName === 'send' && buttonValue && msgCommandRegex.test(buttonValue)) {
							const [pmTarget] = buttonValue.replace(msgCommandRegex, '').split(',');
```

Not sure why it's so low, but your check failure is coming from here. RegExp.exec can return null, so you need to assert that it exists - the ``|| ''`` isn't what you want to do there.

Zarel: Yeah, this was my mistake thinking through the logic wrong.

PR #7222, opened 2020-08-17
Change Poltergeist's activation method

KrisXV:
Server-side part of https://github.com/smogon/pokemon-showdown-client/pull/1598

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -13714,7 +13714,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 			}
 		},
 		onTryHit(target, source, move) {
-			this.add('-activate', target, 'move: Poltergeist', this.dex.getItem(target.item).name);
+			this.add('-item', target, this.dex.getItem(target.item), '[from] move: Poltergeist', '[of] ' + source);
 		},
 		secondary: null,
 		target: "normal",


Zarel:
No, this should be an `-activate`. We used to be inconsistent about this, but in the new protocol, `-item` is only for item changes, `-activate` is for item reveals.

data/moves.ts
```diff
@@ -13714,7 +13714,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 			}
 		},
 		onTryHit(target, source, move) {
-			this.add('-activate', target, 'move: Poltergeist', this.dex.getItem(target.item).name);
+			this.add('-item', target, this.dex.getItem(target.item), '[from] move: Poltergeist', '[of] ' + source);
```
Zarel, line 13717:
```suggestion
			this.add('-activate', target, this.dex.getItem(target.item).fullName, '[from] move: Poltergeist', '[of] ' + source);
```

Something like _this_ I think might be closer to what you want, though.

Zarel: Although the old version should in theory also work. Is there a reason you wanted to change it?

KrisXV: no the server-side is useless, my bad

PR #7223, opened 2020-08-17
IPTools: Support wildcards in string ranges

AnnikaCodes:
This is needed for private code.

### DIFF
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -151,6 +151,13 @@ export const IPTools = new class {
 	}
 	stringToRange(range: string): AddressRange | null {
 		if (!range) return null;
+		if (range.endsWith('*')) {
+			const [a, b, c] = range.replace('*', '').split('.');
+			return {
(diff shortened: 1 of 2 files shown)


server/ip-tools.ts
```diff
@@ -151,6 +151,13 @@ export const IPTools = new class {
 	}
 	stringToRange(range: string): AddressRange | null {
 		if (!range) return null;
+		if (range.endsWith('*')) {
+			const [a, b, c] = range.replace('*', '').split('.');
+			return {
+				minIP: IPTools.ipToNumber(`${a || '0'}.${b || '0'}.${c || '0'}.0`),
+				maxIP: IPTools.ipToNumber(`${a || '255'}.${b || '255'}.${c || '255'}.255`),
+			};
+		}
```
Zarel, line 160: This is pretty genius.

AnnikaCodes: Thanks!

PR #7224, opened 2020-08-17
Fix ipsearch error message when no results found

asgdf:
Ipsearch takes singular IPs, IP ranges using asterisks, and hostname suffixes. Having a strict regex for all of these doesn't seem very straightforward, and I consider the ability to catch malformed IPs less important than the accuracy of the output messages of a tool I use regularly.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -419,7 +419,7 @@ export const commands: ChatCommands = {
 			}
 		}
 		if (!results.length) {
-			if (!IPTools.ipRegex.test(ip)) return this.errorReply(`${ip} is not a valid IP or host.`);
+			if (!ip.includes('.') && !ip.includes('?/')) return this.errorReply(`${ip} is not a valid IP or host.`);
 			return this.sendReply(`No results found.`);
 		}
 		return this.sendReply(results.join('; '));


scheibo:
> Having a strict regex for all of these doesn't seem very straightforward

Still seems pretty doable...? Unless I'm missing something

asgdf:
You can maybe have a straightforward one for IPs, but hostnames can be pretty arbitrary. And at that point, it doesn't strike me as worth it.

Zarel:
Oh, it should also match the part before `/`. Hostnames also don't have to contain `?/`.

asgdf:
Hm. Honestly, is it even worth having the check at all? The hostname can also be a hostname suffix, which means it doesn't have to contain anything in particular at all, that is true. At that point, it feels easier to just have it be the responsibility of the user to double check what they're passing in if nothing is found.

Zarel:
Yeah I think ipsearch should also search hosts with an `includes` filter, and it doesn't need a spellcheck.

AnnikaCodes:
The issue with ranges has since been fixed. I'm fine with the idea of not checking it at all, though, since we can't really determine whether it's a valuid hostname search.

Zarel:
I'm also fine with it not checking at all. It will say "Hostnames with X: none found" which is clear enough, anyway.

DaWoblefet:
Closing as this was fixed a while ago and there haven't been any gripes about how /ipsearch has worked since then.

PR #7225, opened 2020-08-18
Add more Simplified + Traditional Chinese translations

KrisXV:
Provided by bobochan and neycwang


Zarel:
Thanks bobochan and neycwang!

PR #7227, opened 2020-08-18
Fix HTML error with pending media requests

tmagicturtle:
This was an extraneous `</div>` tag which resulted in only the first pending media request to show inside of the `<details>` tag.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -643,7 +643,7 @@ export abstract class BasicRoom {
 				message += `<strong>Comment:</strong> ${entry.comment ? entry.comment : 'None.'}<br />`;
 				message += `<button class="button" name="send" value="/approveshow ${userid}">Approve</button>` +
 				`<button class="button" name="send" value="/denyshow ${userid}">Deny</button></div>`;
-				message += `</div><hr />`;
+				message += `<hr />`;
 			}
 			message += `</details></div>`;
 		}


Zarel:
Good catch!

Zarel:
...why is there a details tag in the first place? @mia-pi-git you always seem to overuse it...

Zarel:
Making the approval thingy a link is also weird, it should really just be inline.

mia-pi-git:
> ...why is there a details tag in the first place? @mia-pi-git you always seem to overuse it...

I felt it was useful so it wouldn't get super spammy on join (these are large sometimes!)



> Making the approval thingy a link is also weird, it should really just be inline.

Easier for staff to see, imo. Just click, check, approve/deny.

Zarel:
> I felt it was useful so it wouldn't get super spammy on join (these are large sometimes!)

Media requests should not be ignored; they should be approved or denied ASAP. If they're spammy, that's good, it means they're less likely to be ignored.

> Easier for staff to see, imo. Just click, check, approve/deny.

Do you mean "harder"? It'd be easier if it showed the image directly, then you wouldn't need to click.

mia-pi-git:
> Media requests should not be ignored; they should be approved or denied ASAP. If they're spammy, that's good, it means they're less likely to be ignored.

Fair, but also what if you don't want to see that? For example, i imagine a gstaff member trying to deal with an issue wouldn't want to see that spam upon join. 
> Do you mean "harder"? It'd be easier if it showed the image directly, then you wouldn't need to click.

Fair - I mostly thought it was an issue about, once again, too much content.

Zarel:
> Fair, but also what if you don't want to see that? For example, i imagine a gstaff member trying to deal with an issue wouldn't want to see that spam upon join.

Maybe it should start expanded but be collapsible? I think I would like that the most.

mia-pi-git:
> > Fair, but also what if you don't want to see that? For example, i imagine a gstaff member trying to deal with an issue wouldn't want to see that spam upon join.
> 
> Maybe it should start expanded but be collapsible? I think I would like that the most.

<details open> it is! i'll PR that + direct showing of image later, then?

Zarel:
Yeah, sure, that'd be great, thanks!

PR #7228, opened 2020-08-18
Show staff intro + pending requests on autojoin

tmagicturtle:
At the moment, the staff intro and pending requests only appear on manual joins, not on autojoin. This is because `user.can('mute', null, this)` results false when it is run since the user is not yet logged in. This change creates a new method for generating the staff intro that is run after the user is fully logged in. It also modifies the original `getIntroMessage` so that the staff intro and pending requests still appear on manual joins.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -768,12 +768,8 @@ export abstract class BasicRoom {
 			this.reportJoin('j', user.getIdentityWithStatus(this.roomid), user);
 		}
 
-		if (user.can('mute', null, this)) {
-			this.sendUser(
-				user,
-				this.getStaffIntroMessage(user)
-			);
-		}
+		const staffIntro = this.getStaffIntroMessage(user);
+		if (staffIntro) this.sendUser(user, staffIntro);
 
 		this.users[user.id] = user;
 		this.userCount++;
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -655,6 +655,7 @@ export abstract class BasicRoom {
 			}
 			message += `</details></div>`;
 		}
+		return message ? `|raw|${message}` : ``;
 		return message;
 	}
 	getSubRooms(includeSecret = false) {
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -634,7 +634,8 @@ export abstract class BasicRoom {
 		return message;
 	}
 	getStaffIntroMessage(user: User) {
-		let message = Utils.html`\n|raw|`;
+		if (!user.can('mute', null, this)) return ``;
+		let message = ``;
 		if (this.settings.staffMessage) {
 			message += `\n|raw|<div class="infobox">(Staff intro:)<br /><div>` +
 				this.settings.staffMessage.replace(/\n/g, '') +
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -628,9 +628,8 @@ export abstract class BasicRoom {
 				this.settings.introMessage.replace(/\n/g, '') +
 				`</div></div>`;
 		}
-		if (user.can('mute', null, this)) {
-			message += this.getStaffIntroMessage(user);
-		}
+		const staffIntro = this.getStaffIntroMessage(user);
+		if (staffIntro) message += `\n${staffIntro}`;
 		return message;
 	}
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -782,10 +796,10 @@ export abstract class BasicRoom {
 		this.users[user.id] = user;
 		if (joining) {
 			this.reportJoin('j', user.getIdentityWithStatus(this.roomid), user);
-			if (this.settings.staffMessage && user.can('mute', null, this)) {
+			if (user.can('mute', null, this)) {
 				this.sendUser(
 					user,
-					`|raw|<div class="infobox">(Staff intro:)<br /><div>${this.settings.staffMessage.replace(/\n/g, '')}</div></div>`
+					this.getStaffIntroMessage(user)
```
Zarel, line 802: This is good, but will send an empty message if a room has neither a staff message nor pending approvals.

I think the `user.can('mute', null, this)` check should be moved inside `getStaffIntroMessage`, and this line should have a check that if it's not there, it's not sent.

server/rooms.ts
```diff
@@ -628,12 +628,19 @@ export abstract class BasicRoom {
...
 				`</div></div>`;
 		}
-		if (this.settings.staffMessage && user.can('mute', null, this)) {
+		if (user.can('mute', null, this)) {
+			message += this.getStaffIntroMessage(user);
+		}
+		return message;
+	}
+	getStaffIntroMessage(user: User) {
+		let message = Utils.html`\n|raw|`;
```
Zarel, line 637:
```suggestion
		if (!user.can('mute', null, this)) return ``;
		let message = ``;
```

server/rooms.ts
```diff
@@ -643,7 +650,7 @@ export abstract class BasicRoom {
 				message += `<strong>Comment:</strong> ${entry.comment ? entry.comment : 'None.'}<br />`;
 				message += `<button class="button" name="send" value="/approveshow ${userid}">Approve</button>` +
 				`<button class="button" name="send" value="/denyshow ${userid}">Deny</button></div>`;
-				message += `</div><hr />`;
+				message += `<hr />`;
 			}
 			message += `</details></div>`;
 		}
```
Zarel, line 656:
```suggestion
		}
		return message ? `|raw|${message}` : ``;
```

server/rooms.ts
```diff
@@ -628,12 +628,19 @@ export abstract class BasicRoom {
 				this.settings.introMessage.replace(/\n/g, '') +
 				`</div></div>`;
 		}
-		if (this.settings.staffMessage && user.can('mute', null, this)) {
+		if (user.can('mute', null, this)) {
+			message += this.getStaffIntroMessage(user);
+		}
```
Zarel, line 633:
```suggestion
		const staffIntro = this.getStaffIntroMessage(user);
		if (staffIntro) message += `\n${staffIntro}`;
```

server/rooms.ts
```diff
@@ -761,6 +768,13 @@ export abstract class BasicRoom {
 			this.reportJoin('j', user.getIdentityWithStatus(this.roomid), user);
 		}
 
+		if (user.can('mute', null, this)) {
+			this.sendUser(
+				user,
+				this.getStaffIntroMessage(user)
+			);
+		}
```
Zarel, line 776:
```suggestion
		const staffIntro = this.getStaffIntroMessage(user);
		if (staffIntro) this.sendUser(user, staffIntro);
```

(And the same for `onRename`)

Is `onJoin` the right place for this, though? It seems wrong to me.

tmagicturtle: I'm not entirely sure. It was originally in `onConnect` which made certain things not appear on autojoin since the user wasn't logged in yet. I moved it to `onJoin` because it sounded correct and ended up working when I tried it there... where do you suggest it go?

Zarel: `onConnect` happens when you open the room in any connection. `onJoin` happens whenever the "user joined" message appears - when you go from a guest to logged in, or when you join a room while logged in (so only the _first_ time you open the room).

Zarel:
I believe this will not show the staff intro when `/nick`ing from a non-staff to a staff account, but that's comparatively minor. This is a good improvement, thanks!

PR #7231, opened 2020-08-19
Add missing lines to German translation

AnnikaCodes:


Zarel:
Thanks!

Zarel:
Are you doing this manually? I wonder if we should have a way to automate this.

AnnikaCodes:
> Are you doing this manually? I wonder if we should have a way to automate this.

The missing lines I added manually, but generally I add the lines to `english.json` and then use VSCode's regex find and replace tool to add them to all the other translation files. (I also have a Python script that helps me translate chat plugins,  although a human eye is _very_ necessary.)

Zarel:
I think we might benefit from a "remove lines that are removed from `english.json`, add lines that are added" script.

Zarel:
I originally thought about preserving comments, but I guess these are JSON files that don't support comments. So simply preserving `english.json`'s whitespace should be enough.

PR #7232, opened 2020-08-20
Scavengers: fix display bugs

sparkychildcharlie:
I guess this has been in since the multiple hosts update... which was a long time ago... oops i guess?

Nyaaa~ :3

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -709,8 +709,9 @@ export class ScavengerHunt extends Rooms.RoomGame {
 
 		const now = Date.now();
 		const time = Chat.toDurationString(now - this.startTime, {hhmmss: true});
+		const canBlitz = this.completed.length < 3;
 
-		const blitz = now - this.startTime <= 60000 &&
+		const blitz = now - this.startTime <= 60000 && canBlitz &&
 			(this.room.settings.scavSettings?.blitzPoints?.[this.gameType] || DEFAULT_BLITZ_POINTS[this.gameType]);
 
 		player.completed = true;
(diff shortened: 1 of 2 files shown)


server/chat-plugins/scavengers.ts
```diff
@@ -1424,7 +1424,7 @@ const ScavengerCommands: ChatCommands = {
 			room,
 			gameType === 'official' || gameType === 'recycled'
 		);
-		if (!hosts) return this.errorReply("The user(s) you specified as the host is not online, or is not in the room.");
+		if (!hosts.length) return this.errorReply("The user(s) you specified as the host is not online, or is not in the room.");
```
AnnikaCodes, line 1427:
```suggestion
		if (!hosts.length) {
				return this.errorReply("One or more of the user(s) you specified as the host is offline or not in the room.");
		}
```

server/chat-plugins/scavengers.ts
```diff
@@ -1733,7 +1733,7 @@ const ScavengerCommands: ChatCommands = {
 		} else {
 			const [hostsArray, ...params] = target.split('|');
 			const hosts = ScavengerHunt.parseHosts(hostsArray.split(/[,;]/), room);
-			if (!hosts) return this.errorReply("The user(s) you specified as the host is not online, or is not in the room.");
+			if (!hosts.length) return this.errorReply("The user(s) you specified as the host is not online, or is not in the room.");
```
AnnikaCodes, line 1736:
```suggestion
			if (!hosts.length) {
					return this.errorReply("One or more of the user(s) you specified as the host(s) is offline or not in the room.");
			}
```

Zarel:
@sparkychildcharlie you've marked those as resolved but it's a day later and your fix isn't in yet...

Zarel:
I'm drafting until it's fixed.

sparkychildcharlie:
Took me like 4 days to realize there was a ready for review button

Zarel:
Ah, it's fine to take a while. I was just afraid you marking things as resolved was a misclick, since Mia often does it and then forgets to actually fix the thing.

PR #7233, opened 2020-08-20
Rooms: Clean formatting for pending approval display on join

mia-pi-git:
see #7227.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2427,7 +2427,7 @@ export const commands: ChatCommands = {
 		return this.errorReply(`/showimage has been deprecated - use /show instead.`);
 	},
 
-	requestshow(target, room, user) {
+	async requestshow(target, room, user) {
 		if (!room) return this.requiresRoom();
 		if (!this.canTalk()) return false;
 		if (!room.settings.requestShowEnabled) {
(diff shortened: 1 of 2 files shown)


server/rooms.ts
```diff
@@ -56,6 +56,13 @@ interface ChatRoomTable {
 	subRooms?: string[];
 }
 
+interface ShowRequest {
+	name: string;
+	link: string;
+	comment: string;
+	dimensions?: [number, number, boolean];
```
Zarel, line 63: Why isn't this just three separate properties `width`, `height`, `resized`?

mia-pi-git: Since that's what Chat.fitImage puts out, and i felt it would unnecessarily complicate things to move it into a new object / parse it later.

mia-pi-git: In the end, you're still going to have to destructure it, and this makes it a bit cleaner on the /requestshow side.

Zarel: Makes sense.

server/chat-commands/info.ts
```diff
@@ -2440,11 +2440,20 @@ export const commands: ChatCommands = {
 		let [link, comment] = target.split(',');
 		if (!/^https?:\/\//.test(link)) link = `https://${link}`;
 		link = encodeURI(link);
+		let dimensions;
+		if (!/^(https?:\/\/)?(www\.)?(youtube\.com|youtu\.be)(\/|$)/i.test(link)) {
+			try {
+				dimensions = await Chat.fitImage(link);
+			} catch (e) {
+				throw new Chat.ErrorMessage('Invalid link.');
```
Zarel, line 2448: Does `Chat.fitImage` not throw `Chat.ErrorMessage`? Maybe it should...

Zarel: Ah, I see how it works now. I think this is fine for me.

PR #7234, opened 2020-08-20
Chat: Refactor CommandContext and PageContext to use ErrorMessage more

mia-pi-git:
This makes `CommandContext.can` and `PageContext.can` into functions that use `Chat.ErrorMessage`, so that instead of 
```ts
if (!this.can('lock')) return false;
``` 
you can just add, say, 
```ts
this.checkPermission('lock');
```
to the command, and it will throw the access denied message if they can't use it. It also still returns true, so you can still use it as part of an if statement where it checks something else in addition to the permission, but this will make it easier to use.
Likewise, `CommandContext.privatelyCan` was also renamed to `CommandContext.privatelyCheckPermission`, and `CommandContext.canUseConsole` was also edited to use `ErrorMessage`.
(I also removed the `Utils.html` in `PageContext.send`, so that HTML can be used in errors - if that's something people disagree with, i'll remove it, but i feel like the pages should be sanitizing the errors on their own, if necessary.)


AnnikaCodes (changes requested):
PMd you on Discord like you asked.

server/chat.ts
```diff
@@ -2116,6 +2097,16 @@ export const Chat = new class {
...
+// @ts-ignore
+CommandContext.prototype.can = CommandContext.prototype.checkCan;
+// @ts-ignore
+CommandContext.prototype.canTalk = CommandContext.prototype.checkChat;
+// @ts-ignore
+CommandContext.prototype.canBroadcast = CommandContext.prototype.checkBroadcast;
+// @ts-ignore
+CommandContext.prototype.canHTML = CommandContext.prototype.checkHTML;
+// @ts-ignore
+CommandContext.prototype.canEmbedURI = CommandContext.prototype.checkEmbedURI;
```
Zarel, line 2109: I think you should also add `requiresRoom` to these.

Zarel: And `checkPMHTML` and `privatelyCheck`... actually can you just make sure you alias all the ones that make sense to alias?

Zarel:
Thanks!

I'll wait until tomorrow to hotpatch since I need to sleep now, but it looks good.

PR #7236, opened 2020-08-20
Fix crash in /updateserver private

AnnikaCodes:


monsanto:
I don't know the context here, but any path that relies on the working directory of the process is suspect to me.

monsanto:
Does any server other than main even use the private code functionality? Would they find a `../main-private` hardcoded path useful? I mean, that wasn't even the path we used on the previous server. Maybe we should just require setting `Config.privatecodepath`

AnnikaCodes:
> Does any server other than main even use the private code functionality? Would they find a `../main-private` hardcoded path useful? I mean, that wasn't even the path we used on the previous machine. Maybe we should just require setting `Config.privatecodepath`

I don't know if side servers want this command but I'd be all for requiring `privatecodepath`.

Zarel:
Thanks!

PR #7238, opened 2020-08-20
Display whitelist correctly in demotion messages

AnnikaCodes:
Fixes [this bug](https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/page-15#post-8572439).


Zarel:
Thanks!

PR #7239, opened 2020-08-21
Don't show error message on leaving a deleted room

AnnikaCodes:
This will require either a restart or a monkeypatch to add `Rooms.deletedRooms`. It fixes [this bug](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-26#post-8574270).


HoeenCoder:
Would the size of that set possibly get too large over time? Would it be wise to purge it every once in a while if so?

Zarel:
New client is designed not to have this sort of problem.

For the old client, I suppose `/nofeedback /leave [roomid]` would fix it. That would probably be a better solution than this.

AnnikaCodes:
Hmm, I suppose this would be a good usecase for `/noreply`.

PR #7251, opened 2020-08-23
Fix extraneous error message in /back

AnnikaCodes:
As per [this bug report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8517371), the help message directs you to type `/back` when marked as idle/away, when really, `/back` is only necessary for the Busy status.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -438,7 +438,9 @@ export class CommandContext extends MessageContext {
 			}
 		}
 
-		if (this.user.statusType === 'idle') this.user.setStatusType('online');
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
> setting themselves as back without sending a message

`/back` _is_ sending a message, though. I could suppress the message for checking if a user is online, I suppose, but tracking a user's most recent statusType change seems overly cumbersome.

Zarel:
The easiest way might be to hardcode `/back` not to automatically set yourself back.

AnnikaCodes:
Sure that works too

AnnikaCodes:
I didn't mean to close this.

Zarel:
Thanks!

PR #7252, opened 2020-08-23
Fix /unlockip's interaction with namelocks

AnnikaCodes:
https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/page-15#post-8576576


Zarel:
Thanks!

PR #7256, opened 2020-08-24
Add a Room Prize Winner rank

AnnikaCodes:
As per [this suggestion](https://www.smogon.com/forums/threads/allow-room-owners-to-promote-to-%E2%9C%A9.3669018/#post-8577989). This will obviously necessitate a change to main's `config.js` as well.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -652,6 +652,13 @@ exports.grouplist = [
 		nooverride: true,
 	},
 	{
+		symbol: '\u2729',
+		id: "star",
+		name: "Star",
+		inherit: '+',
+		roomonly: true,
+	},
+	{
 		symbol: '+',
 		id: "voice",
 		name: "Voice",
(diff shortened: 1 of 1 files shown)


Zarel:
This looks and sounds a bit too much like the Player rank. I don't like the symbol overlap tbh.

Zarel:
I'll comment on the suggestion with more thoughts.

AnnikaCodes:
Hmm, how about the black chess queen as a prize symbol (♛)

Zarel:
Looks kinda unrecognizable at normal resolutions, I think. You should discuss it in the Suggestions thread, though.

AnnikaCodes:
I'm going to draft this until we reach a consensus on what symbol the ROs want.

AnnikaCodes:
Alrighty, it looks like `^` is a reasonable choice.

Zarel:
Yep, sounds good to me, too.

PR #7257, opened 2020-08-24
Implement Battle Stadium Singles/Doubles Series 6

DaWoblefet:
Series 6 bans 16 different Pokemon (17 if you count Indeedee as two mons).

The Immortal and I discussed this on Discord. VGC without bans still needs to exist due to several online tournaments using Series 5. Series 6 VGC will be implemented separately as Battle Stadium Doubles, which is still accurate while giving players the option to ladder in either format (Series 6 will only last 2 months). Battle Stadium Singles is also updated but will just use the Series 6 banlist and not have a separate ladder.

I'm not sure if it would be useful to clarify in Battle Stadium Doubles that this is Series 6 somehow. iirc we don't typically include the series information.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -212,6 +212,10 @@ export const Formats: (FormatsData | {section: string, column?: number})[] = [
 			battle: 3,
 		},
 		ruleset: ['Standard GBU'],
+		banlist: [
+			'Cinderace', 'Dragapult', 'Excadrill', 'Gyarados', 'Hippowdon', 'Incineroar', 'Indeedee', 'Magnezone', 'Mimikyu',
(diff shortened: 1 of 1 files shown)


LegoFigure11:
Any chance we could have Series 5 BSS still present (as a challenge-only format)? It's still relevant for some community tours that BSS are running.

TheImmortal:
> Any chance we could have Series 5 BSS still present (as a challenge-only format)? It's still relevant for some community tours that BSS are running.

That can easily be played via custom challenge (or groupchat tour) unbanning the Pokemon. Just inform the relevant tournament host to specify how. I don't believe it is necessary to keep a challenge format.

PR #7260, opened 2020-08-25
Help: Allow regular users to suggest regexes

mia-pi-git:
Approved by Help staff.

### DIFF
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -15,6 +15,18 @@ const PATH = 'config/chat-plugins/help.json';
 // 4: filters out conveniently short aliases
 const MINIMUM_LENGTH = 4;
 
+const BAN_DURATION = 2 * 24 * 60 * 60 * 1000;
+/**
+ * Terms commonly used in helping that should be ignored
+ * within question parsing (the Help#find function).
+ */
+const COMMON_TERMS = [
+	'<<(.+)>>', '(do|use|type)( |)(``|)(/|!|//)(.+)(``|)', 'go to', '/rfaq (.+)', 'you can', Chat.linkRegex, 'click',
+	'need to',
+].map(item => new RegExp(item, "i"));
+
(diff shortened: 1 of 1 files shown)


mia-pi-git:
> I PMd you on Discord with review changes as you requested

(resolved)

DaWoblefet:
I've been out of the loop with Help room filters, but why would we ever want regular users to be able to suggest regexes in some sort of formal manner like this? Are there really so many _regular users_ who are 1) technically skilled enough to write a proper regular expression and 2) can be trusted enough, to warrant a full-blown command? This seems very niche and if you trust someone enough to do those things, they should just be Help staff.

mia-pi-git:
> I've been out of the loop with Help room filters, but why would we ever want regular users to be able to suggest regexes in some sort of formal manner like this? Are there really so many _regular users_ who are 1) technically skilled enough to write a proper regular expression and 2) can be trusted enough, to warrant a full-blown command? This seems very niche and if you trust someone enough to do those things, they should just be Help staff.

It doesn't actually allow regexes. It allows use of some symbols that are transformed into very simple regexes. You can only use actual regex when you're both dev staff and help mod. 
It comes down to like, &, !, and |, which i feel are rather self-explanatory.

server/chat-plugins/help.ts
```diff
@@ -322,22 +383,36 @@ export const commands: ChatCommands = {
...
+				return this.errorReply(`You are banned from making suggestions to the Help filter.`);
+			}
 			if (!faq) return this.errorReply(`Invalid FAQ.`);
 			const regex = Answerer.stringRegex(target);
-			if (Answerer.queue.includes(target)) {
+			const entry = {
+				regexString: target,
+				userid: user.id,
+			};
+			if (Answerer.queue.includes(entry)) {
```
whalemer, line 411: this is an object, so it's checked by identity and this won't ever be true

Zarel:
I guess if they want it...

PR #7261, opened 2020-08-25
Chat plugins: Use the room language to translate mod actions

AnnikaCodes:
Now that we actually use user languages we shouldn't be translating moderator messages. We could translate them with the room language, I suppose, but that seems rather overkill for a rather narrow case (moderators of language rooms who don't know enough English to understand these short strings).


Zarel:
I'd prefer translating them with the room language. Do we have a `room.tr` function for this use?

AnnikaCodes:
Good plan

Zarel:
Thanks!

PR #7262, opened 2020-08-25
Properly inform client of statusType changes

AnnikaCodes:
This should fix the bug where usernames remain grey on the client after reloading.


Zarel:
ty

PR #7263, opened 2020-08-26
Chat: Make translations not crash after hotpatching

AnnikaCodes:
This will make things that are translated before the translations JSONs are loaded just be English (untranslated) instead of crashing PS!


Zarel:
Thanks!

PR #7264, opened 2020-08-26
Improve handling of locks when users are merged

AnnikaCodes:


Zarel:
Is there any reason `prevNames` is an Object?

Zarel:
Thanks, by the way.

Zarel:
But I wonder if `prevNames` should be refactored into an Array.

AnnikaCodes:
> But I wonder if `prevNames` should be refactored into an Array.

Yeah I don't believe there's a need for keeping IDs _and_ names; just IDs should be fine. I'll probably get around to that later this week.

PR #7265, opened 2020-08-26
Searchlogs: Fix crash

mia-pi-git:
Currently, the escaping isn't actually taken into account properly by either ripgrep or child process (not sure which), but this escapes them for a proper fix.
(also fixes a error with date searching)

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -435,7 +435,9 @@ type SearchMatch = readonly [string, string, string, string, string];
 
 export const LogSearcher = new class {
 	constructRegex(str: string) {
-		const searches = str.split('+').map(term => Utils.escapeRegex(term));
+		// modified regex replace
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -435,7 +435,9 @@ type SearchMatch = readonly [string, string, string, string, string];
 
 export const LogSearcher = new class {
 	constructRegex(str: string) {
-		const searches = str.split('+').map(term => Utils.escapeRegex(term));
+		// modified regex replace
+		str = str.replace(/[\\^$.*?()[\]{}|]/g, '\\$&');
+		const searches = str.split('+');
```
Zarel, line 440: I don't understand what's wrong with `escapeRegex`? What's going on here?

Zarel:
I suppose I can merge now and understand later, but the removal of `Utils.escapeRegex` bothers me a lot.

Zarel:
Annika's fix makes a lot more sense to me.

PR #7266, opened 2020-08-26
Fix /unrangelock

AnnikaCodes:
This allows `/unrangelock` to actually undo rangelocks, makes it unlock all users on the IP who are _not_ locked by name (per Anubis in the staff room), as well as supporting IP ranges in `Users#findUsers`.


Zarel:
Thanks! I assume this isn't hotpatchable because of the `Users#findUsers` changes?

AnnikaCodes:
> Thanks! I assume this isn't hotpatchable because of the `Users#findUsers` changes?

`/hotpatch iptools` followed by `/hotpatch chat` will allow staff to use `/unrangelock` just fine. The change that makes users on the unlocked ranges get unlocked with the range won't be live until restart, though.

Zarel:
Good to know, thanks!

PR #7267, opened 2020-08-26
Refactor User#prevNames and User#ips to arrays

AnnikaCodes:


server/chat-plugins/thing-of-the-day.ts
```diff
@@ -156,7 +156,7 @@ class OtdHandler {
 		const nomObj = {
 			nomination: nomination,
 			name: user.name,
-			userids: Object.assign(obj, user.prevNames),
+			userids: [...user.previousIDs],
```
Zarel, line 159: iirc this is still slower than `user.previousIDs.slice()` so we might want to just use that.

Zarel:
```suggestion
			userids: user.previousIDs.slice(),
```

server/chat-plugins/trivia.ts
```diff
@@ -386,29 +386,26 @@ class Trivia extends Rooms.RoomGame {
...
-
-				const tarIps = Object.keys(tarUser.ips);
-				const ipMatch = tarIps.some(ip => (ip in user.ips));
-				if (ipMatch) return this.room.tr('You have already signed up for this game.');
+			const targetUser = Users.get(id);
+			if (targetUser) {
+				const isSameUser = (
+					targetUser.previousIDs.includes(user.id) ||
+					targetUser.previousIDs.some(tarId => user.previousIDs.includes(tarId)) ||
+					Object.keys(targetUser.ips).some(ip => (ip in user.ips))
```
Zarel, line 406: Should `.ips` also be an array?

Zarel:
Looks good other than those points!

server/chat-plugins/trivia.ts
AnnikaCodes: `User#ips` also tracks the number of connections on each IP, which _seems_ useful, but isn't actually used anywhere in public or private code. I think making it an array should be fine?

PR #7272, opened 2020-08-28
Update Instruct's longdesc

Ocean-ey:
Instruct fails for all Max moves (similarly to Z-Moves), though its current description does not reflect this.


AnnikaCodes:
@Marty-D, I believe this is your area of expertise?

Marty-D:
I was waiting on https://github.com/smogon/pokemon-showdown/pull/7036, but I guess what's another conflict for the pile, right?

Thanks @Ocean-ey, sorry for the delay. I also added the fact that it also fails on Dynamax Pokemon regardless of last move.

PR #7273, opened 2020-08-28
Make the /faq autoconfirmed more accurate

AnnikaCodes:
Per [the loginserver code](https://github.com/smogon/pokemon-showdown-client/blob/master/lib/ntbb-session.lib.php#L498), one must only _play_ a rated battle and be registered for a week to become autoconfirmed (losing or tying is OK).


Zarel:
I'd rather fix the loginserver code than this.

`WHERE userid = ?` -> `WHERE userid = ?, w != 0`?

AnnikaCodes:
OK, I opened a PR there.

PR #7276, opened 2020-08-28
Fix /unlockip

AnnikaCodes:
#7266 fixed the behavior of `/unrangelock`, but introduced another bug. As far as I can tell, `/unlockip` should unlock all  users locked by name on the IP, while `/unrangelock` should unlock all users locked by a rangelock on the IP range.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -863,6 +863,7 @@ export const commands: ChatCommands = {
 
 		for (const curUser of Users.findUsers([], [target])) {
 			if (
+				curUser &&
 				(range ? curUser.locked === '#rangelock' : !curUser.locked.startsWith('#')) &&
 				!Punishments.getPunishType(curUser.id)
 			) {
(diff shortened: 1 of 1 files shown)


server/chat-commands/moderation.ts
```diff
@@ -862,7 +862,11 @@ export const commands: ChatCommands = {
 		Punishments.savePunishments();
 
 		for (const curUser of Users.findUsers([], [target])) {
-			if (curUser.locked?.startsWith('#') && !Punishments.getPunishType(curUser.id)) {
+			if (
+				curUser &&
+				(range ? curUser.locked === '#rangelock' : !curUser.locked.startsWith('#')) &&
```
AnnikaCodes, line 867:
```suggestion
				(range ? curUser.locked === '#rangelock' : !curUser.locked?.startsWith('#')) &&
```

AnnikaCodes, line 867:
```suggestion
				(range ? curUser.locked === '#rangelock' : !curUser.locked?.startsWith('#')) &&
```

Zarel:
Thanks!

PR #7271, opened 2020-08-27
Support pre-loading SQLite schemas on initialization

mia-pi-git:
This PR adds framework to initialize all .sql files before chat-plugins are loaded, so that plugins don't need to do it themselves in weird ways. Along with this, it adds a Config setting for configuring what uses SQLite and what doesn't.
It also adds a ``/evalsql databasefile, query``  (or ``>>sql`` for evaluating SQL on the server, looks like https://prnt.sc/u766yc), attaches a database accessor to Chat (for plugins), and adds a README to the databases/ file laying out how the initializer works (credit to Annika for writing the original in her layout, before i rewrote it (with her permission and by her suggestion) to use as a README in the `./databases/` README.)
As well, it sets up the `/databases/schemas/` folder, with a small readme on usage, and gitignores /databases/*.db* correctly.
(If anyone else can suggest things that would be useful for this, please comment.)

This comes from the ideas Annika and I came up with in https://gist.github.com/AnnikaCodes/afa36fc8b17791be812eebbb22182426.

PR #7277, opened 2020-08-28
Convert modlogs to SQLite

### LINKED ISSUE #6469: Migrating Modlog to Postgres

AnnikaCodes:
Fixes #6469.

This will require either restarting the server or hotpatching chat and modlog and monkeypatching Rooms.

There is a conversion script (shoutout to @TheJetOU for writing most of it) in `tools/`. It should be run to convert our existing modlogs to SQLite. It also converts old-style modlogs, so there will be no more of `([thepoll] was ended)` type modlog entries.

### DIFF
--- a/databases/schemas/modlog-schema.sql
+++ b/databases/schemas/modlog-schema.sql
@@ -1,0 +1,26 @@
+ CREATE TABLE IF NOT EXISTS modlog (
+    modlog_id INTEGER NOT NULL PRIMARY KEY,
+    -- UNIX timestamp
+    timestamp INTEGER NOT NULL,
+    --- roomid OR global-roomid
+    roomid TEXT NOT NULL,
+    action TEXT NOT NULL,
+    visual_roomid TEXT,
+    action_taker TEXT,
+    userid TEXT,
+    autoconfirmed_userid TEXT,
+    ip TEXT,
+    note TEXT
+);
+
+CREATE TABLE IF NOT EXISTS alts (
+    alts_id INTEGER NOT NULL,
+    userid TEXT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS alts_index_1 ON alts(alts_id);
+CREATE INDEX IF NOT EXISTS ml_index_1 ON modlog(timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_2 ON modlog(roomid, userid, autoconfirmed_userid, timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_3 ON modlog(roomid, ip, timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_4 ON modlog(roomid, note, timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_5 ON modlog(roomid, action_taker, timestamp);
--- a/databases/schemas/modlog.sql
+++ b/databases/schemas/modlog.sql
@@ -6,7 +6,7 @@
     roomid TEXT NOT NULL,
     action TEXT NOT NULL,
     visual_roomid TEXT,
-    action_taker TEXT,
+    action_taker_userid TEXT,
     userid TEXT,
     autoconfirmed_userid TEXT,
     ip TEXT,
@@ -14,13 +14,16 @@
 );
 
 CREATE TABLE IF NOT EXISTS alts (
-    alts_id INTEGER NOT NULL,
-    userid TEXT NOT NULL
+    modlog_id INTEGER NOT NULL,
+    userid TEXT NOT NULL,
+    PRIMARY KEY (modlog_id, userid),
+    FOREIGN KEY (modlog_id) REFERENCES modlog(modlog_id)
 );
 
-CREATE INDEX IF NOT EXISTS alts_index_1 ON alts(alts_id);
 CREATE INDEX IF NOT EXISTS ml_index_1 ON modlog(timestamp);
 CREATE INDEX IF NOT EXISTS ml_index_2 ON modlog(roomid, userid, autoconfirmed_userid, timestamp);
 CREATE INDEX IF NOT EXISTS ml_index_3 ON modlog(roomid, ip, timestamp);
 CREATE INDEX IF NOT EXISTS ml_index_4 ON modlog(roomid, note, timestamp);
-CREATE INDEX IF NOT EXISTS ml_index_5 ON modlog(roomid, action_taker, timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_5 ON modlog(roomid, action_taker_userid, timestamp);
+
+PRAGMA journal_mode=WAL;
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -422,7 +422,7 @@ export const commands: ChatCommands = {
 			this.checkCan('warn', null, room);
 
 			this.addModAction(`${targetUser.name} would be warned by ${user.name} but is offline.${(publicReason ? ` (${publicReason})` : ``)}`);
-			this.globalModlog('WARN OFFLINE', targetUser, ` by ${user.id}${(target ? `: ${publicReason} ${privateReason}` : ``)}`);
+			this.globalModlog('WARN OFFLINE', targetUser, target ? `${publicReason} ${privateReason}` : ``);
 			return;
 		}
 		if (!(targetUser.id in room.users) && !globalWarn) {
@@ -443,7 +443,7 @@ export const commands: ChatCommands = {
 
 		this.addModAction(`${targetUser.name} was warned by ${user.name}.${(publicReason ? ` (${publicReason})` : ``)}`);
 		if (globalWarn) {
-			this.globalModlog('WARN', targetUser, ` by ${user.id}${(target ? `: ${publicReason} ${privateReason}` : ``)}`);
+			this.globalModlog('WARN', targetUser, target ? `${publicReason} ${privateReason}` : ``);
 		} else {
 			this.modlog('WARN', targetUser, target ? `${publicReason} ${privateReason}` : ``, {noalts: 1});
 		}
@@ -662,9 +662,9 @@ export const commands: ChatCommands = {
 		room.hideText([userid, toID(this.inputUsername)]);
 
 		if (room.settings.isPrivate !== true && room.persist) {
-			this.globalModlog(`${week ? 'WEEK' : ''}ROOMBAN`, targetUser, ` by ${user.id}${(target ? `: ${target}` : ``)}`);
+			this.globalModlog(`${week ? 'WEEK' : ''}ROOMBAN`, targetUser, target);
 		} else {
-			this.modlog(`${week ? 'WEEK' : ''}ROOMBAN`, targetUser, ` by ${user.id}${(target ? `: ${target}` : ``)}`);
+			this.modlog(`${week ? 'WEEK' : ''}ROOMBAN`, targetUser, target);
 		}
 		return true;
 	},
@@ -685,7 +685,7 @@ export const commands: ChatCommands = {
 		if (name) {
 			this.addModAction(`${name} was unbanned from ${room.title} by ${user.name}.`);
 			if (room.settings.isPrivate !== true && room.persist) {
-				this.globalModlog("UNROOMBAN", name, ` by ${user.id}`);
+				this.globalModlog("UNROOMBAN", name);
 			}
 		} else {
 			this.errorReply(`User '${target}' is not banned from this room.`);
@@ -770,7 +770,7 @@ export const commands: ChatCommands = {
 
 		const globalReason = (target ? `: ${userReason} ${proof}` : '');
 		this.globalModlog(
-			(week ? "WEEKLOCK" : (month ? "MONTHLOCK" : "LOCK")), targetUser || userid, ` by ${user.id}${globalReason}`
+			(week ? "WEEKLOCK" : (month ? "MONTHLOCK" : "LOCK")), targetUser || userid, globalReason
 		);
 
 		const durationMsg = week ? ' for a week' : (month ? ' for a month' : '');
@@ -835,7 +835,7 @@ export const commands: ChatCommands = {
 
 		if (unlocked) {
 			this.addGlobalModAction(`${unlocked.join(", ")} ${((unlocked.length > 1) ? "were" : "was")} unlocked by ${user.name}.${reason}`);
-			if (!reason) this.globalModlog("UNLOCK", toID(target), ` by ${user.id}`);
+			if (!reason) this.globalModlog("UNLOCK", toID(target));
 			if (targetUser) targetUser.popup(`${user.name} has unlocked you.`);
 		} else {
 			this.errorReply(`User '${target}' is not locked.`);
@@ -863,7 +863,7 @@ export const commands: ChatCommands = {
 		}
 
 		this.addGlobalModAction(`The name '${target}' was unlocked by ${user.name}.`);
-		this.globalModlog("UNLOCKNAME", userid, ` by ${user.name}`);
+		this.globalModlog("UNLOCKNAME", userid);
 	},
 	unrangelock: 'unlockip',
 	rangeunlock: 'unlockip',
@@ -899,7 +899,12 @@ export const commands: ChatCommands = {
 		}
 
 		this.privateGlobalModAction(`${user.name} unlocked the ${range ? "IP range" : "IP"}: ${target}`);
-		this.globalModlog(`UNLOCK${range ? 'RANGE' : 'IP'}`, target, ` by ${user.name}`);
+		(room || Rooms.global).modlog({
+			action: `UNLOCK${range ? 'RANGE' : 'IP'}`,
+			ip: target,
+			isGlobal: true,
+			loggedBy: user.id,
+		});
 	},
 	unlockiphelp: [`/unlockip [ip] - Unlocks a punished ip while leaving the original punishment intact. Requires: @ &`],
 	unlocknamehelp: [`/unlockname [name] - Unlocks a punished alt, leaving the original lock intact. Requires: % @ &`],
@@ -979,7 +984,7 @@ export const commands: ChatCommands = {
 		room?.hideText([userid, toID(this.inputUsername)]);
 
 		const globalReason = (target ? `: ${userReason} ${proof}` : '');
-		this.globalModlog("BAN", targetUser, ` by ${user.id}${globalReason}`);
+		this.globalModlog("BAN", targetUser, globalReason);
 		return true;
 	},
 	globalbanhelp: [
@@ -999,7 +1004,7 @@ export const commands: ChatCommands = {
 		}
 
 		this.addGlobalModAction(`${name} was globally unbanned by ${user.name}.`);
-		this.globalModlog("UNBAN", target, ` by ${user.id}`);
+		this.globalModlog("UNBAN", target);
 	},
 	unglobalbanhelp: [`/unglobalban [username] - Unban a user. Requires: @ &`],
 
@@ -1143,8 +1148,9 @@ export const commands: ChatCommands = {
 		}
 		this.checkCan('receiveauthmessages', null, room);
 		target = target.replace(/\n/g, "; ");
-		if (room.roomid === 'staff' || room.roomid === 'upperstaff') {
-			this.globalModlog('NOTE', null, ` by ${user.id}: ${target}`);
+		if (room.roomid === 'staff' || room.roomid === 'upperstaff' || room.roomid.startsWith('help-')) {
+			// Staff rooms and help tickets should have global modnotes
+			this.globalModlog('NOTE', null, target);
 		} else {
 			this.modlog('NOTE', null, target);
 		}
@@ -1211,7 +1217,7 @@ export const commands: ChatCommands = {
 		}
 		if (Users.Auth.getGroup(nextGroup).rank < Users.Auth.getGroup(currentGroup).rank) {
 			this.privateGlobalModAction(`${name} was demoted to ${groupName} by ${user.name}.`);
-			this.globalModlog(`GLOBAL ${groupName.toUpperCase()}`, userid, ` by ${user.id}: (demote)`);
+			this.globalModlog(`GLOBAL ${groupName.toUpperCase()}`, userid, `(demote)`);
 			if (targetUser) targetUser.popup(`You were demoted to ${groupName} by ${user.name}.`);
 		} else {
 			this.addGlobalModAction(`${name} was promoted to ${groupName} by ${user.name}.`);
@@ -1463,13 +1469,13 @@ export const commands: ChatCommands = {
 		let forceRenameMessage;
 		if (targetUser.connected) {
 			forceRenameMessage = `was forced to choose a new name by ${user.name}${(reason ? `: ${reason}` : ``)}`;
-			this.globalModlog('FORCERENAME', targetUser, ` by ${user.name}${(reason ? `: ${reason}` : ``)}`);
+			this.globalModlog('FORCERENAME', targetUser, reason);
 			Chat.forceRenames.set(targetUser.id, (Chat.forceRenames.get(targetUser.id) || 0) + 1);
 			Ladders.cancelSearches(targetUser);
 			targetUser.send(`|nametaken||${user.name} considers your name inappropriate${(reason ? `: ${reason}` : ".")}`);
 		} else {
 			forceRenameMessage = `would be forced to choose a new name by ${user.name} but is offline${(reason ? `: ${reason}` : ``)}`;
-			this.globalModlog('FORCERENAME OFFLINE', targetUser, ` by ${user.name}${(reason ? `: ${reason}` : ``)}`);
+			this.globalModlog('FORCERENAME OFFLINE', targetUser, reason);
 			if (!Chat.forceRenames.has(targetUser.id)) Chat.forceRenames.set(targetUser.id, 0);
 		}
 
@@ -1510,7 +1516,7 @@ export const commands: ChatCommands = {
 
 		const reasonText = reason ? ` (${reason})` : `.`;
 		this.privateGlobalModAction(`${targetUser.name} was namelocked by ${user.name}${reasonText}`);
-		this.globalModlog("NAMELOCK", targetUser, ` by ${user.id}${reasonText}`);
+		this.globalModlog("NAMELOCK", targetUser, reasonText);
 
 		const roomauth = Rooms.global.destroyPersonalRooms(targetUser.id);
 		if (roomauth.length) {
@@ -1544,7 +1550,7 @@ export const commands: ChatCommands = {
 		}
 
 		this.addGlobalModAction(`${unlocked} was unnamelocked by ${user.name}.${reason}`);
-		if (!reason) this.globalModlog("UNNAMELOCK", toID(target), ` by ${user.id}`);
+		if (!reason) this.globalModlog("UNNAMELOCK", toID(target));
 		if (targetUser) targetUser.popup(`${user.name} has unnamelocked you.`);
 	},
 	unnamelockhelp: [`/unnamelock [username] - Unnamelocks the user. Requires: % @ &`],
@@ -1706,7 +1712,7 @@ export const commands: ChatCommands = {
 		}
 
 		if (!room.settings.isPrivate && room.persist) {
-			this.globalModlog("BLACKLIST", targetUser, ` by ${user.id}${target ? `: ${target}` : ''}`);
+			this.globalModlog("BLACKLIST", targetUser, target);
 		} else {
 			// Room modlog only
 			this.modlog("BLACKLIST", targetUser, target);
@@ -1758,7 +1764,7 @@ export const commands: ChatCommands = {
 			Monitor.log(`[CrisisMonitor] Trusted user ${targetUser.name} was banned from battling by ${user.name}, and should probably be demoted.`);
 		}
 
-		this.globalModlog("BATTLEBAN", targetUser, ` by ${user.id}${reasonText}`);
+		this.globalModlog("BATTLEBAN", targetUser, reasonText);
 		Ladders.cancelSearches(targetUser);
 		Punishments.battleban(targetUser, null, null, reason);
 		targetUser.popup(`|modal|${user.name} has prevented you from starting new battles for 2 days${reasonText}`);
@@ -1781,7 +1787,7 @@ export const commands: ChatCommands = {
 
 		if (unbanned) {
 			this.addModAction(`${unbanned} was allowed to battle again by ${user.name}.`);
-			this.globalModlog("UNBATTLEBAN", toID(target), ` by ${user.name}`);
+			this.globalModlog("UNBATTLEBAN", toID(target));
 			if (targetUser) targetUser.popup(`${user.name} has allowed you to battle again.`);
 		} else {
 			this.errorReply(`User ${target} is not banned from battling.`);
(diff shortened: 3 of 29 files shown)


monsanto:
Have you tested the conversion script? I remember jetou saying he had trouble converting things

server/chat-plugins/modlog-viewer.ts
```diff
@@ -197,6 +198,21 @@ async function getModlog(
  * Battle Search Functions
  *********************************************************/
 
+function checkRipgrepAvailability() {
+	if (Config.ripgrepmodlog === undefined) {
+		Config.ripgrepmodlog = (async () => {
```
monsanto, line 203: This value is always truthy (note: it is a Promise<boolean>, not a boolean)

AnnikaCodes:
I tested them, yes, but not on a terribly large scale.

test/tools/modlog/converter.js
```diff
@@ -0,0 +1,518 @@
...
+				`[2020-08-23T19:50:49.947Z] (development) GLOBAL UNITTEST: [annika] ac: [heartofetheria] alts: [googlegoddess], [princessentrapta] [127.0.0.1] by yourmom: Global test\n`
+			);
+
+			assert.strictEqual(
+				mlConverter.isTesting.files.get('/modlog_global.txt'),
+				`[2020-08-23T19:50:49.947Z] (development) GLOBAL UNITTEST: [annika] ac: [heartofetheria] alts: [googlegoddess], [princessentrapta] [127.0.0.1] by yourmom: Global test\n`
+			);
+		});
+
+		it('should convert from text to SQLite, including old-format lines and nonsense lines', async () => {
```
AnnikaCodes, line 476: Here's where I test it in the unit tests; I also ran the script on some autogenerated files (~10,000 entries) and it worked fine in both directions.

monsanto:
I'll try to give this a more thorough review tomorrow. Stray thought for today: we should probably run this code on main for a bit before we merge it into master. Once this is merged all side servers will have to convert their existing modlogs before they can get other fixes. It may be that we need at least one round of iteration with the staff before we can set the schema in stone (which we need to do at least temporarily, until the migration framework is merged)

databases/schemas/modlog-schema.sql
```diff
@@ -0,0 +1,26 @@
...
+    visual_roomid TEXT,
+    action_taker TEXT,
+    userid TEXT,
+    autoconfirmed_userid TEXT,
+    ip TEXT,
+    note TEXT
+);
+
+CREATE TABLE IF NOT EXISTS alts (
+    alts_id INTEGER NOT NULL,
```
monsanto, line 17: `alts_id` should be called `modlog_id`, you'd name a column `alts_id` if it was the integer primary key for `alts`

monsanto: `(alts_id, userid)` is the primary key for this table

databases/schemas/modlog-schema.sql
```diff
@@ -0,0 +1,26 @@
+ CREATE TABLE IF NOT EXISTS modlog (
+    modlog_id INTEGER NOT NULL PRIMARY KEY,
+    -- UNIX timestamp
+    timestamp INTEGER NOT NULL,
+    --- roomid OR global-roomid
+    roomid TEXT NOT NULL,
+    action TEXT NOT NULL,
+    visual_roomid TEXT,
+    action_taker TEXT,
```
monsanto, line 9: if this is a userid i'd suffix the column with that much like you did with `visual_roomid`

databases/schemas/modlog-schema.sql
```diff
@@ -0,0 +1,26 @@
...
+    ip TEXT,
+    note TEXT
+);
+
+CREATE TABLE IF NOT EXISTS alts (
+    alts_id INTEGER NOT NULL,
+    userid TEXT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS alts_index_1 ON alts(alts_id);
```
monsanto, line 21: Implied by primary key suggested

server/chat-commands/moderation.ts
```diff
@@ -873,7 +873,12 @@ export const commands: ChatCommands = {
...
 
 		this.privateGlobalModAction(`${user.name} unlocked the ${range ? "IP range" : "IP"}: ${target}`);
-		this.globalModlog(`UNLOCK${range ? 'RANGE' : 'IP'}`, target, ` by ${user.name}`);
+		(room || Rooms.global).modlog({
+			action: `UNLOCK${range ? 'RANGE' : 'IP'}`,
+			ip: target,
+			isGlobal: true,
+			loggedBy: user.id,
+		});
 	},
```
monsanto, line 882: Why does this have such a different function signature than `globalModlog`?

server/chat-plugins/modlog-viewer.ts
```diff
@@ -197,6 +198,21 @@ async function getModlog(
  * Battle Search Functions
  *********************************************************/
 
+async function checkRipgrepAvailability() {
+	if (Config.ripgrepmodlog === undefined) {
+		Config.ripgrepmodlog = await (async () => {
```
monsanto, line 203: redundant IIFE

server/chat-plugins/scavengers.ts
```diff
@@ -895,11 +898,14 @@ export class ScavengerHunt extends Rooms.RoomGame {
 
 			// notify staff
 			const staffMsg = `(${player.name} has been caught attempting a hunt with ${uniqueConnections} connections on the account. The user has also been given 1 infraction point on the player leaderboard.)`;
-			const logMsg = `([${player.id}] has been caught attempting a hunt with ${uniqueConnections} connections on the account. The user has also been given 1 infraction point on the player leaderboard.)`;
 
 			this.room.sendMods(staffMsg);
 			this.room.roomlog(staffMsg);
-			this.room.modlog(logMsg);
+			this.room.modlog({
+				action: 'SCAVCHEATER',
```
monsanto, line 905: why is this `SCAVCHEATER` but the others have a space like `SCAV BLITZ`? are there any guidelines on what these should be called?

server/modlog.ts
```diff
@@ -36,284 +31,241 @@ const PUNISHMENTS = [
...
-		if (this.list.length > this.maxSize) {
-			this.list.pop();
-		}
+export class Modlog {
+	readonly database: Database.Database;
+	readonly usePM: boolean;
+
+	constructor(path: string, noProcessManager = false) {
+		this.database = new Database(path);
+		this.database.exec(FS(`${__dirname}/../databases/schemas/modlog-schema.sql`).readIfExistsSync()); // Set up tables, etc.
```
monsanto, line 76: seems inconsistent to have `MODLOG_DB_PATH` for the database and this as a hardcoded string

server/modlog.ts
```diff
@@ -36,284 +31,241 @@ const PUNISHMENTS = [
...
+			`VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+			args: [
+				entry.time || Date.now(), entry.isGlobal ? `global-${roomid}` : roomid, entry.visualRoomID,
+				entry.action, entry.userid, entry.autoconfirmedID, entry.ip, entry.loggedBy, entry.note,
+			],
+		}) as Database.RunResult;
+		return Promise.all((entry.alts || []).map(alt => {
+			return this.runSQL({
+				query: `INSERT INTO alts (alts_id, userid) VALUES (?, ?)`,
+				args: [result.lastInsertRowid, alt],
```
monsanto, line 116: Needs to execute within a transaction

server/modlog.ts
```diff
@@ -36,284 +31,241 @@ const PUNISHMENTS = [
...
-}
 
-export function checkRipgrepAvailability() {
-	if (Config.ripgrepmodlog === undefined) {
-		Config.ripgrepmodlog = (async () => {
+	async runSQL(query: ModlogQuery, forceNoPM = false): Promise<Database.RunResult | any[]> {
+		if (!forceNoPM && this.usePM) {
+			return PM.query(query);
+		} else {
+			const statement = this.database.prepare(query.query).bind(...query.args);
```
monsanto, line 87: Slow to prepare a statement every time you query. Queries should be prepared once and reused.

databases/schemas/modlog-schema.sql
```diff
@@ -0,0 +1,26 @@
+ CREATE TABLE IF NOT EXISTS modlog (
```
monsanto, line 1: Need to make this a WAL mode db with `PRAGMA journal_mode=WAL;`

databases/schemas/modlog-schema.sql
monsanto: should establish foreign key between what is now called `alts_id` and `modlog_id`

server/chat-commands/moderation.ts
AnnikaCodes: because `globalModlog` doesn't support logging a random IP address not associated with the user and since we only need to do that in one place I didn't think a special case in `globalModlog` was worth it

server/chat-plugins/scavengers.ts
AnnikaCodes: there really isn't - SCAV CHEATER works fine too

server/modlog.ts
AnnikaCodes: I can prepare most of the queries on initialization, but since search queries are dynamically generated based on which fields are searched for, I think that preparing all possible combinations would be rather cumbersome. Maybe have a cache in the form of a string:statement map?

server/roomlogs.ts
```diff
@@ -223,13 +222,13 @@ export class Roomlog {
...
-	modlog(message: string, overrideID?: string) {
-		void Rooms.Modlog.write(this.roomid, message, overrideID);
+	modlog(entry: ModlogEntry) {
+		void Rooms.Modlog.write(this.roomid, entry);
 	}
 	async rename(newID: RoomID): Promise<true> {
 		const roomlogPath = `logs/chat`;
 		const roomlogStreamExisted = this.roomlogStream !== null;
-		await this.destroy(false); // don't destroy modlog, since it's renamed later
+		await this.destroy(); // don't destroy modlog, since it's renamed later
```
monsanto, line 231: What does this comment mean? Is it still relevant?

server/modlog.ts
```diff
@@ -36,284 +32,255 @@ const PUNISHMENTS = [
...
-	 * Writes to the modlog
-	 * @param overrideID Specify this parameter for when the room ID to be displayed
-	 * is different from the ID for the modlog stream
-	 * (The primary use case of this is tournament battles.)
-	 */
-	write(roomid: ModlogID, message: string, overrideID?: string) {
-		const stream = this.streams.get(roomid);
-		if (!stream) throw new Error(`Attempted to write to an uninitialized modlog stream for the room '${roomid}'`);
-		void stream.write(`[${new Date().toJSON()}] (${overrideID || roomid}) ${message}\n`);
+	formatArray(arr: any[], args: any[]) {
```
monsanto, line 157:
```suggestion
	formatArray(arr: unknown[], args: unknown[]) {
```

server/modlog.ts
```diff
@@ -36,284 +32,255 @@ const PUNISHMENTS = [
...
-		return this.list.slice();
-	}
+export class Modlog {
+	readonly database: Database.Database;
+	readonly usePM: boolean;
+	readonly modlogInsertionQuery: Database.Statement;
+	readonly altsInsertionQuery: Database.Statement;
+	readonly renameQuery: Database.Statement;
+	readonly globalPunishmentsSearchQuery: Database.Statement;
+	readonly insertionTransaction: Database.Transaction;
```
monsanto, line 78: Should be able to type the parameters these queries expect by providing an argument to `Statement`/`Transaction`

server/tournaments/index.ts
```diff
@@ -603,7 +603,7 @@ export class Tournament extends Rooms.RoomGame {
 			return false;
 		}
 
-		if (this.players.length < 2) {
+		if (this.players.length < 0) {
```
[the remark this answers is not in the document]
AnnikaCodes: it shouldn't be in there whatever it is

server/roomlogs.ts
AnnikaCodes: It's not

server/modlog.ts
```diff
@@ -36,284 +32,279 @@ const PUNISHMENTS = [
...
+			if (search.note.searches.length === 1) {
+				query += ` AND regex(?, note)`;
+				args.push(this.generateRegex(search.note.searches[0], search.note.isExact));
+			} else {
+				const parts = [];
+				for (const noteSearch of search.note.searches) {
+					parts.push(`regex(?, note)`);
+					args.push(this.generateRegex(noteSearch, search.note.isExact));
+				}
+				query += ` AND (${parts.join(' OR ')})`;
```
monsanto, line 229: These branches generate the same query

server/modlog.ts
```diff
@@ -36,284 +32,279 @@ const PUNISHMENTS = [
...
+		this.insertionTransaction({
+			action: entry.action,
+			roomID: roomid,
+			visualRoomID: overrideID || entry.visualRoomID,
+			userid: entry.userid,
+			autoconfirmedID: entry.autoconfirmedID,
+			ip: entry.ip,
+			loggedBy: entry.loggedBy,
+			note: entry.note,
+			time: entry.time || Date.now(),
```
monsanto, line 152: You insert with `Date.now()` but query with `Date.now() / 1000`

server/modlog.ts
```diff
@@ -36,284 +32,279 @@ const PUNISHMENTS = [
...
+
+		let roomChecker = `roomid IN (${this.formatArray(rooms, args)})`;
+		if (rooms.includes('global')) roomChecker = `(roomid LIKE 'global-%' OR ${roomChecker})`;
+
+		let query = `SELECT *, (SELECT group_concat(userid, ',') FROM alts WHERE alts.modlog_id = modlog.modlog_id) as alts FROM modlog`;
+		query += ` WHERE ${roomChecker}`;
+
+		if (search.action) {
+			query += ` AND action LIKE '%' || ? || '%'`;
+			args.push(search.action.toUpperCase());
```
monsanto, line 198: Would rather make it an invariant of `search.action` to be uppercase (iow, when you create a `ModlogSearch`, ensure it is already uppercase)

server/modlog.ts
```diff
@@ -36,284 +32,279 @@ const PUNISHMENTS = [
...
+					return !room.settings.isPrivate && !room.settings.isPersonal;
+				})
+				.map(room => room.roomid as ModlogID) :
+			[roomid]
+		);
+
+		const query = this.prepareSearch(rooms, maxLines, onlyPunishments, search);
+		const start = Date.now();
+		const rows = await this.runSQLWithResults(query);
+		const results: ModlogEntry[] = rows.map((row: any) => {
```
monsanto, line 269: chained `map` + `filter` is inefficient if you plan to return a non-trivial # of results

server/chat-plugins/modlog-viewer.ts
```diff
@@ -451,17 +486,29 @@ export const commands: ChatCommands = {
 		void getModlog(
 			connection,
 			roomid,
-			target,
+			search,
+			target.replace(/,?\s*(room|lines)\s*=[^,]*,?/g, ''),
```
monsanto, line 490: What is this?

server/punishments.ts
```diff
@@ -748,9 +748,14 @@ export const Punishments = new class {
 			Monitor.log(`[CrisisMonitor] Autolocked user ${name} has public roomauth (${roomauth.join(', ')}), and should probably be demoted.`);
 		}
 
-		const ipStr = typeof user !== 'string' ? ` [${(user as User).latestIp}]` : '';
-		const roomid = typeof room !== 'string' ? (room as Room).roomid : room;
-		Rooms.global.modlog(`AUTO${punishment.replace('ED', '')}: [${userid}]${ipStr}: ${reason}`, roomid);
+		const logEntry: ModlogEntry = {
+			action: `AUTO${punishment.replace('ED', '')}`,
```
monsanto, line 752: The `.replace` here feels error-prone.

server/modlog.ts
```diff
@@ -36,284 +32,279 @@ const PUNISHMENTS = [
...
+			note: string,
+			time: number,
+			alts: ID[],
+		}) => {
+			const result = this.modlogInsertionQuery.run(entry);
+			for (const alt of entry.alts || []) {
+				this.altsInsertionQuery.run(result.lastInsertRowid as number, alt);
+			}
+		});
+		this.usePM = !noProcessManager && !Config.nofswriting;
```
monsanto, line 118: I don't know what `nofswriting` controls here, if it means we don't write to a DB then I don't know why we are opening a database at all or creating queries

monsanto (commented):
I'm wondering if it makes sense to have a "group" column, instead of prefixing everything with "MAFIA", "TICKET", etc. You also have places in the code where you want to operate uniformly on punishments, that could also be a group?

server/chat-plugins/modlog-viewer.ts
AnnikaCodes: this is because the `getModlog()` function re-adds room= and line= qualifiers, so we remove them from the query to avoid duplicating them

server/modlog.ts
AnnikaCodes: `nofswriting` indicates that we are in a test environment; we don't want to use a multi-process modlog setup in a test environment but we can still query an in-memory database.

server/punishments.ts
AnnikaCodes: The punishment is hardcoded to be one of `NAMELOCKED`, `WEEKLOCKED`, `NAMEWEEKLOCKED`, or `LOCKED` above. I suppose I could manually add the ED in the monitor.

AnnikaCodes:
> I'm wondering if it makes sense to have a "group" column, instead of prefixing everything with "MAFIA", "TICKET", etc. You also have places in the code where you want to operate uniformly on punishments, that could also be a group?

I feel like a group column would make the code just as prefixed (since it'd be `{actionPrefix: 'MAFIA', action: 'HOSTBAN'}` vs `{action: 'MAFIA HOSTBAN'}`) and would increase query complexity when searching by action.

AnnikaCodes:
Also, I'm having a weird bug where the ProcessManager crashes because `statement.all` is undefined. I'm not sure what's causing this — I've been struggling with it for a few hours and I've tried reverting numerous changes to figure it out. Do you have any ideas?

server/punishments.ts
monsanto: I see, not as urgent then, but I think it would be clearer what is going on

Zarel:
Drafting because it doesn't _sound_ like this is ready yet? Feel free to undraft if I'm misinterpreting what's going on.

Zarel:
As for deployment: Either chaos or I can handle it. If you want me to handle a deployment, though, I would generally prefer a flatfile→SQLite conversion to happen in multiple steps:

1. continue to write to and read from flatfiles, but also write to SQLite - also publish some sort of migration script that can be run at this time, with instructions on when to run it, and also a test script for reading from flatfiles to check that the migration was successful
2. remove the functionality for writing to and reading from flatfiles, finish up the migration

AnnikaCodes:
>  it doesn't sound like this is ready yet?

This is correct - we're still discussing the multiprocess architecture in #sim-dev afaik.

And sure, I was planning to figure out deployment once we're settled on the design and code.

monsanto:
Yeah we handled the PM log transition by dual logging for a time. I think this PR has removed flat file support, I'm not sure how easy that is to add back in.

monsanto:
Tagging @mia-pi-git 

On multiprocess:

- Unless there is a way to avoid regex + limit or streaming rows, the modlog is locked into better-sqlite3, and thus locked into multiprocess architecture.

- With a combination of WAL mode, keeping the database in cache, a low durability level, and carefully tuning queries, we could maybe get away with running stuff synchronously. Even in this case I would be worried about cases in which `fsync` is unavoidable (and therefore hitting disk is unavoidable), which could introduce unacceptable latency.

- We could use both better-sqlite3 and node-sqlite3, depending on which makes more sense. I think using two different libraries is undesirable.

- The PM log currently uses node-sqlite3. Moving it to better-sqlite3 would introduce some overhead, as every PM would need to be serialized, and copied to a different process, and deserialized. This overhead is probably acceptable for modlogs which are used significantly less often. 

- I think a decent next step is looking into [worker threads](https://nodejs.org/api/worker_threads.html), which are isolated like a separate process, but have faster message passing. I'm envisioning a small database library that uses these & better-sqlite3 and handles things like cross-thread prepared statements (the issue you were running into with `statement.all`) that consumers of PS would use instead of `PM.query`.

AnnikaCodes:
We could add flat file support in if that was necessary for the migration.

> small database library that uses these

What are "these"?

monsanto:
> > small database library that uses these
> 
> What are "these"?
> 
> A library seems like a good idea though.

Oh just a typo from rearranging the paragraph, it refers to worker threads

mia-pi-git:
I can write the library, if that's the decision that's made (and if no one else wants to.) I have drafts of something similar that I wrote a while back, but dropped due to lack of necessity at that time

AnnikaCodes:
> if no one else wants to

If you or chaos wants to, you're welcome to. If not, I can work on it.

monsanto:
I'll pass on writing the code, instead I'll give some notes on what I imagine will be the interesting parts:

- SQL stuff in a separate file, should not have dependencies other than `better-sqlite3` (and maybe `sql-template-strings`)
- Opening a database takes a path to a sqlite database and a path to a extension file.
- The extension file exports an object with name -> function. For modlogs, this would include `regex`
- A [new Worker](https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options) is created with the SQL stuff filename, passing the sqlite db path & extension file as `options.workerData`
- SQL worker opens database, `require`s extension file, adds all of the functions. Could just go ahead and `PRAGMA foreign_keys=on` here as I believe every database will want this on. (Can skip `PRAGMA journal_mode=WAL`, this is a persistent database-level pragma and should be part of the schema)
- Preparing a statement [sends the string to the worker](https://nodejs.org/api/worker_threads.html#worker_threads_worker_postmessage_value_transferlist), the worker allocates a new integer and messages it back to `parentPort`. That integer is given a type like `Statement<Params, RowType>` (more expressive than what better-sqlite3 does, I think)
- Querying follows a similar pattern, we send the statement integer, the worker maps the integer to the prepared statement, and messages back the rows.

I imagine the rest will be straightforward but just ping me if not.

It is probably easier if this is added as part of this PR (which I intend to be the start of SQLite usage in PS) as the design may need to evolve in response to feedback on the modlog. If mia is writing it maybe push access could be given to this branch. Alternatively a separate PR could be created and this PR could every now and then merge it in.

mia-pi-git:
@monsanto, does it work if i add it to #7271? Seems to fit better there.

monsanto:
That PR includes a lot of stuff unrelated to this change, and this is a huge transition for PS. I don't want to have to worry about chat plugins until the modlog is stable. It could have gone the other way too, but people were working on the modlog first.

mia-pi-git:
> That PR includes a lot of stuff unrelated to this change, and this is a huge transition for PS. I don't want to have to worry about chat plugins until the modlog is stable. It could have gone the other way too, but people were working on the modlog first.

Alright. I'll open a separate PR then.

monsanto:
While we work around the details of the library, a good way to unblock this PR would be to rip out the process logic, no matter what that is going to be hidden behind the async interface of that library. You can just have `runSQL` be synchronous for now.

We should minimize `generateRegex` usage as much as possible. Can the uses of `regex(r, s2)` where `r = generateRegex(s1, true)` just be replaced with `s1 = s2`? Or am I misunderstanding what `isExact` does?

AnnikaCodes:
> While we work around the details of the library, a good way to unblock this PR would be to rip out the process logic, no matter what that is going to be hidden behind the async interface of that library. You can just have `runSQL` be synchronous for now. (EDIT: actually, will this interfere with eventually restoring the text-based functionality, if we were to do that? Either way, `runSQL` shouldn't make use of it, which should be enough to make this PR work again)
> 
> We should minimize `generateRegex` usage as much as possible. Can the uses of `regex(r, s2)` where `r` is bound to `generateRegex(s1, true)` just be replaced with `s1 = s2`? Or am I misunderstanding what `isExact` does?

I think that the `isExact` regexes can be replaced with `LIKE` operators. I'll do that in a bit.

AnnikaCodes:
Also, I don't think `runSQL`'s synchronicity or lack thereof would affect our ability to log to flat files and a database concurrently?

AnnikaCodes:
(done)

databases/schemas/modlog.sql
```diff
@@ -0,0 +1,29 @@
...
+
+CREATE TABLE IF NOT EXISTS alts (
+    modlog_id INTEGER NOT NULL,
+    userid TEXT NOT NULL,
+    PRIMARY KEY (modlog_id, userid),
+    FOREIGN KEY (modlog_id) REFERENCES modlog(modlog_id)
+);
+
+CREATE INDEX IF NOT EXISTS ml_index_1 ON modlog(timestamp);
+CREATE INDEX IF NOT EXISTS ml_index_2 ON modlog(roomid, userid, autoconfirmed_userid, timestamp);
```
xfix, line 24: This index seems unnecessary and I don't think any query in the code would be able to use that (a query to make use of `autoconfirmed_userid` would have to match `userid` in the same row, for example `WHERE roomid = ? AND userid = ? AND autoconfirmed_userid = ?`). You have a query that does `userid = ? OR autoconfirmed_userid = ?`, but that's won't be able to use this index.

tools/modlog/converter.ts
```diff
@@ -0,0 +1,492 @@
...
+			const roomid = file.slice(7, -4);
+			const entries = raw.split('\n')
+				.map((line, index) => {
+					const entry = parseModlog(line, raw[index + 1], roomid === 'global');
+					if (roomid === 'global' && entry?.roomID === 'global') return null;
+					return entry;
+				});
+			for (const entry of entries) {
+				if (entry) logs.push(entry);
+			}
```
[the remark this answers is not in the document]
AnnikaCodes: Yup, done.

monsanto:
Performance is good now, nice work.

Does not properly roundtrip `OLD MODLOG` entries, examples:

```diff
-[2014-11-20T13:46:00.288Z] (lobby) OLD MODLOG: by unknown: [punchoface] would be muted by [thecaptain] but was already muted.)
+[2014-11-20T13:46:00.288Z] (lobby) OLD MODLOG:: [punchoface] would be muted 
```

```diff
-[2017-04-20T18:20:40.487Z] (1v1) OLD MODLOG: by unknown: The tournament auto start timer was set to 7 by TIBot
-[2017-04-20T18:20:42.408Z] (1v1) OLD MODLOG: by unknown: The tournament auto disqualify timer was set to 2 by Scrappie
+[2017-04-20T18:20:40.487Z] (1v1) OLD MODLOG:: The tournament auto start timer was set to 7 
+[2017-04-20T18:20:42.408Z] (1v1) OLD MODLOG:: The tournament auto disqualify timer was set to 2 
```

```diff
-[2020-09-19T23:24:13.310Z] (lobby) HANGMAN::  by archastl
-[2020-09-19T23:24:33.686Z] (lobby) HANGMAN::  by fart
-[2020-09-19T23:25:24.908Z] (lobby) HANGMAN::  by archastl
-[2020-09-19T23:25:56.341Z] (lobby) HANGMAN::  by finland
-[2020-09-19T23:27:21.323Z] (lobby) HANGMAN::  by fart
-[2020-09-19T23:28:49.309Z] (lobby) HANGMAN::  by archastl
+[2020-09-19T23:24:13.310Z] (lobby) HANGMAN: by archastl:   
+[2020-09-19T23:24:33.686Z] (lobby) HANGMAN: by fart:   
+[2020-09-19T23:25:24.908Z] (lobby) HANGMAN: by archastl:   
+[2020-09-19T23:25:56.341Z] (lobby) HANGMAN: by finland:   
+[2020-09-19T23:27:21.323Z] (lobby) HANGMAN: by fart:   
+[2020-09-19T23:28:49.309Z] (lobby) HANGMAN: by archastl:   
```

```diff
-[2020-09-20T22:57:27.263Z] (lobby) NOTIFYRANK::  by officerjenny: %, You are the last staff left in lobby.
+[2020-09-20T22:57:27.263Z] (lobby) NOTIFYRANK: by officerjenny: %, You are the last staff left in lobby.
```

AnnikaCodes:
I fixed the OLD MODLOG thing. The others seem to be tripping up because of having two `:`s, but I was unable to reproduce the issue of two colons getting added in the first place from `parseModlog` and `rawifyLog`. Would it be possible for me to get the original line that caused those `HANGMAN::  by archastl` issues?

monsanto:
```
[2020-09-19T23:24:13.310Z] (lobby) HANGMAN:  by archastl
```

AnnikaCodes:
Hmm, the double space is weird and might be worth normalizing for its own sake, but it's not getting mangled locally for me (using `node tools/modlog/convert --outputlogdir logs/modlog2  --logdir logs/modlog --from txt --to txt`).

monsanto:
Can you move this stuff to your other PR?

monsanto:
original:
```
[2014-11-20T16:30:17.661Z] (lobby) LOCK: [violight] (spamming) by joim
[2014-11-20T16:30:17.673Z] (lobby) (violight's ac account: violight)
```

roundtrip (note space at the end):
```diff
-[2014-11-20T16:30:17.661Z] (lobby) LOCK: [violight] by joim: (spamming) 
+[2014-11-20T16:30:17.661Z] (lobby) LOCK: [violight] by joim: (spamming)
```

------------------------

original:

```
[2020-08-12T00:23:49.183Z] (lobby) MUTE: [hypergigahax] [<ip>] by epicnikolai: please only speak english. Checking your modlog you should have understood the rules by now.
```

roundtrip:

```diff
-[2020-08-12T00:23:49.183Z] (lobby) MUTE: [hypergigahax] [<ip>]: please only speak english. Checking your modlog you should have understood the rules 
+[2020-08-12T00:23:49.183Z] (lobby) MUTE: [hypergigahax] [<ip>] by easeonlyspeakenglishcheckingyourmodlogyoushouldhaveunderstoodtherules: please only speak english. Checking your modlog you should have understood the rules
```

------------

original:
```
[2020-09-07T02:17:22.132Z] (lobby) ROOMBAN: [gamingisawesome] [<ip>] by tenshi: wow this is the most written by a 10 year old sentence I've read in a long time
```

roundtrip

```diff
-[2020-09-07T02:17:22.132Z] (lobby) ROOMBAN: [gamingisawesome] [<ip>]: wow this is the most written 
+[2020-09-07T02:17:22.132Z] (lobby) ROOMBAN: [gamingisawesome] [<ip>] by wthisisthemostwritten: wow this is the most written
```

AnnikaCodes:
This is done.

PR #7278, opened 2020-08-29
Translate missing ht line to german/continuity fix

xMarcell:
I translated the missing help ticket line into german, also replaced "Username" with "Nutzername" for continuity.


Zarel:
Thanks!

PR #7281, opened 2020-08-30
Improve privacy for hidden battles

### LINKED ISSUE #6950: User privacy enhancements

AnnikaCodes:
Incorporates [these changes](https://github.com/smogon/pokemon-showdown/issues/6950#issuecomment-653822792) and also kicks guests out of battles while hiding them.

Also, I'm pretty sure this fixes #6950.

### DIFF
--- a/server/modlog.ts
+++ b/server/modlog.ts
@@ -171,7 +171,9 @@ export class Modlog {
 	async rename(oldID: ModlogID, newID: ModlogID) {
 		const streamExists = this.streams.has(oldID);
 		if (streamExists) await this.destroy(oldID);
-		await FS(`${this.logPath}/modlog_${oldID}.txt`).rename(`${this.logPath}/modlog_${newID}.txt`);
+		if (!this.getSharedID(oldID)) {
+			await FS(`${this.logPath}/modlog_${oldID}.txt`).rename(`${this.logPath}/modlog_${newID}.txt`);
+		}
 		if (streamExists) this.initialize(newID);
 	}
 
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1658,13 +1658,13 @@ export class GameRoom extends BasicRoom {
 				user.popup(`The battle <<${this.roomid}>> has been made private; you must log in to watch private battles.`);
 			}
 		}
-		if (this.roomid.endsWith('pw')) return Promise.resolve(true);
+		if (this.roomid.endsWith('pw')) return true;
 		this.rename(this.title, `${this.roomid}-${password}pw` as RoomID, true);
 	}
 
 	makePublic() {
 		this.settings.isPrivate = false;
-		if (!this.roomid.endsWith('pw')) return Promise.resolve(true);
+		if (!this.roomid.endsWith('pw')) return true;
 		this.rename(this.title, this.getReplayData().id as RoomID);
 	}
 }
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -696,7 +696,7 @@ export abstract class BasicRoom {
 	 * @param newID Add this param if the roomid is different from `toID(newTitle)`
 	 * @param noAlias Set this param to true to not redirect aliases and the room's old name to its new name.
 	 */
-	async rename(newTitle: string, newID?: RoomID, noAlias?: boolean) {
+	rename(newTitle: string, newID?: RoomID, noAlias?: boolean) {
 		if (!newID) newID = toID(newTitle) as RoomID;
 		this.validateTitle(newTitle, newID);
 		if (this.type === 'chat' && this.game) {
@@ -756,7 +756,7 @@ export abstract class BasicRoom {
 		this.settings.title = newTitle;
 		this.saveSettings();
 
-		return this.log.rename(newID);
+		void this.log.rename(newID);
 	}
 
 	onConnect(user: User, connection: Connection) {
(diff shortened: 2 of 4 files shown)


server/rooms.ts
```diff
@@ -1756,11 +1783,11 @@ export const Rooms = {
 
 		if (privacySetter.size) {
 			if (battle.forcePublic) {
-				room.settings.isPrivate = false;
+				void room.makePublic();
```
scheibo, line 1786: why are we not propagating the promise here and below?

server/modlog.ts
```diff
@@ -171,7 +171,9 @@ export class Modlog {
 	async rename(oldID: ModlogID, newID: ModlogID) {
 		const streamExists = this.streams.has(oldID);
 		if (streamExists) await this.destroy(oldID);
-		await FS(`${this.logPath}/modlog_${oldID}.txt`).rename(`${this.logPath}/modlog_${newID}.txt`);
+		if (!this.getSharedID(oldID)) {
```
scheibo, line 174: is this change related? should it be in a separate PR?

server/rooms.ts
AnnikaCodes: Because propagating clutters up a lot of unrelated code with `async`s (including a `setTimeout` in ladder that would look a lot less readable). It's definitely possible to propagate it but I don't forsee it causing problems.

server/modlog.ts
AnnikaCodes: No, it's necessary; otherwise, trying to rename the battle crashes PS!

server/rooms.ts
scheibo: I strongly disagree with ever `void`-ing promises (at the very least they need to be propagated to the top level and cases of ignoring it at the top level should be accompanied with a comment justifying it) and I really which PS didn't normalize this and teach developers it is OK to sacrifice correctness at the alter of 'avoiding clutter', but I've lost that battle.

Zarel: Unnecessarily propagating promises also makes the code less parallel and ultimately slower. It looks like the async-ness here comes from `rename`, which I'm pretty sure doesn't need to be propagated outside of the relevant roomlog function at all (the relevant roomlog functions needs to be able to handle the rename without race conditions whether or not the Promise is propagated). But I suppose you're not interested in having this argument.

scheibo: If you believe changing the argument in favor from 'avoiding clutter' to 'performance reasons' is going to convince me that making something untestable and incorrect is going to satisfy me then yes, I'm not interested in having that argument.

Feel free to ignore this, of course. I am never going to approve/merge code that doesn't propagate promises, but you do you.

AnnikaCodes: (I'm happy to propagate it or not once you two figure this out - I was just going off PS' existing code style choices)

Zarel: It doesn't seem like I'm going to understand pre's perspective; he's explained it to me a few times and it's never clicked for me. Let's just put the `void` inside `this.rename` (change the last `return` to `void`).

server/modlog.ts
```diff
@@ -171,7 +171,9 @@ export class Modlog {
...
+			await FS(`${this.logPath}/modlog_${oldID}.txt`).rename(`${this.logPath}/modlog_${newID}.txt`);
+		}
 		if (streamExists) this.initialize(newID);
```
Zarel, line 177: So, the problem is that this function races with `this.write`.

Look at `this.write`:

```
	write(roomid: ModlogID, message: string, overrideID?: string) {
		const stream = this.streams.get(roomid);
		if (!stream) throw new Error(`Attempted to write to an uninitialized modlog stream for the room '${roomid}'`);
		void stream.write(`[${new Date().toJSON()}] (${overrideID || roomid}) ${message}\n`);
	}
```

It will crash if it's called during either of the two `await`s in this rename function.

To be fair, this isn't a problem your PR introduces, and propagating the Promise will not fix this problem at all... but it does need to be fixed at some point.

Zarel: If renames are atomic (which I believe they nearly always are), we could initialize the new stream _immediately_ instead of waiting. So `void` instead of `await` in every call here. I believe that's sufficient to solve the race condition (except in obscure situations where the rename isn't atomic).

If we can't assume that the rename is atomic or want to do this the "right" way, you would want to store a promise for the stream that `write` can use to wait until it can write for real.

AnnikaCodes: > it does need to be fixed at some point.

We're migrating modlogs to SQLite, which should solve the race condition.

Zarel:
Does this require a server restart?

AnnikaCodes:
Oh, yeah, it does. I always forget to assign the label.

Zarel (approved):
This is great! I think it's ready for the next restart.

Zarel:
oh no conflict

Zarel:
(I'll deal with it)

PR #7282, opened 2020-08-30
Remove trusted user check in /hidetext

AnnikaCodes:
https://www.smogon.com/forums/threads/remove-forcehidelines.3669531/


Zarel:
Thanks!

PR #7283, opened 2020-08-30
Filter Pokémon nicknames after team validation

AnnikaCodes:
This will prevent the staff room from getting spammed when people do things like nickname their Pokémon 10,000 n-slurs in a row.

### DIFF
--- a/server/ladders.ts
+++ b/server/ladders.ts
@@ -113,6 +113,36 @@ class Ladder extends LadderStore {
 			return null;
 		}
 
+		let rating = 0;
+		let valResult;
+		if (isRated && !Ladders.disabled) {
+			const uid = user.id;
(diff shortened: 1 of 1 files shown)


Zarel:
It looks like this just moves the Unown validation after the other checks? I don't understand how this fixes the problem?

AnnikaCodes:
```ts
if (nickname) {
	nickname = Chat.nicknamefilter(nickname, user);
	if (!nickname || nickname !== match[1]) {
		connection.popup(
			`Your team was rejected for the following reason:\n\n` +
			`- Your Pokémon has a banned nickname: ${match[1]}`
		);
		return null;
	}
}
```

It's not just the Unown validation, but also the nickname filter.

Zarel:
Oh, I don't know how I missed that. Thanks!

PR #7284, opened 2020-08-30
Log searching: Escape inputs properly

AnnikaCodes:


AnnikaCodes:
#7265 already fixes this.

PR #7286, opened 2020-08-31
Chat: Show the full command in nonexistent errors

mia-pi-git:
small QOL thing, but just erroring with `this.cmd` can be slightly misleading sometimes. 
(not sure if this was intentional, but it seemed off)


Zarel:
Hm, this might leak private namespaces, but I don't think we have any protection for those anyway.

PR #7287, opened 2020-08-31
Help: Disable if a room is not set in Config

mia-pi-git:
This causes issues for other servers to default to help if it doesn't exist. Not worth providing. If you want it, it can just be configured.

### DIFF
--- a/server/chat-plugins/help.ts
+++ b/server/chat-plugins/help.ts
@@ -95,14 +95,19 @@ export class HelpResponder {
 		this.settings = data.settings || {queueDisabled: false, filterDisabled: false};
 	}
 	getRoom() {
-		const room = Config.helpFilterRoom ? Rooms.get(Config.helpFilterRoom) : Rooms.get('help');
+		if (!Config.helpFilterRoom) return null;
+		const room = Rooms.get(Config.helpFilterRoom);
 		if (!room) {
-			throw new Chat.ErrorMessage(`A room for the Help filter has not been specified. Set Config.helpFilterRoom to enable it`);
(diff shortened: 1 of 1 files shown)


server/chat-plugins/help.ts
```diff
@@ -463,6 +473,7 @@ export const commands: ChatCommands = {
 		unban: 'ban',
 		ban(target, room, user, connection, cmd) {
 			this.room = Answerer.getRoom();
+			if (!this.room) return;
```
HoeenCoder, line 476: For a command, it would be wise to give the user an error message rather than silently failing.

server/chat-plugins/help.ts
```diff
@@ -335,7 +345,7 @@ export const commands: ChatCommands = {
 		toggle(target, room, user) {
 			if (!room) return this.requiresRoom();
 			const helpRoom = Answerer.getRoom();
-			if (!helpRoom) return this.errorReply(`There is no room configured for use of this filter.`);
+			if (!helpRoom) HelpResponder.roomNotFound();
```
HoeenCoder, line 348: Correct me if I'm wrong, but since `roomNotFound` is `throw`ing, this results in a crash message in dev and an email, which really seems like a bad idea for something simply triggerable via a command. If thats the case a plain error message to the command's user would be better.

mia-pi-git: it throws a Chat.ErrorMessage, which is transformed into an `errorReply` by the command parser.

mia-pi-git: I leveraged a throw since the `roomNotFound` can be used in other functions that are called within commands, in that case, and it's easier to just use one thing.

PR #7290, opened 2020-09-01
Implement 3v3, 4v4, 5v5 and apply to formats

DaWoblefet:
Factoring out the length of battle in teamLength and applying it in a ruleset allows more freeflow customization of battle formats for things like tournaments. It is fairly trivial to implement VGC Little Cup by adding "Four vs Four" to it, for example, or make a 3v3 BSS-style Ubers or something of that nature. I added 5v5 because it was trivial to do while we're here. 

I just followed The Immortal's hack of modifying the read-only teamLength at the start of battle. It's probably preferable to just drop the readOnly since we're not treating it that way with an implementation like this. I don't know how to do so the correct way, however.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2536,11 +2536,9 @@ export const commands: ChatCommands = {
 		this.runBroadcast();
 		const {quote, date, userid} = room.settings.quotes[Math.floor(Math.random() * room.settings.quotes.length)];
 		const formatted = quote.split('\n').map(item => Chat.formatText(item)).join('<br />');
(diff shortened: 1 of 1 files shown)


TheImmortal:
Eh, the hack was meant to be for the niche 1v1/2v2 usage. Not sure if it should be used overall like this.

Ideally, we should be able to adjust ``teamLength`` in tournaments.

DaWoblefet:
That makes more sense on reflection. Being able to configure things like teamLength or forcedLevel directly would be a better solution, in a similar way to how ruleset mods are added/removed. I'll rework this later.

PR #7292, opened 2020-09-01
Trivia: Reformat help and don't use module.exports

AnnikaCodes:
This uses `<details>` and `<li>` rather than a scroll bar: 
image


Zarel:
Thanks!

PR #7295, opened 2020-09-01
Chatfilters: Spoiler autolocking banwords

AnnikaCodes:
Currently, this is only done for slurs, which has led to globals unhappy about porn sites being mentioned in Staff room.


AnnikaCodes:
@Zarel how does this look?

Zarel:
Looks like a good improvement for people! I think this got lost in some other stuff.

PR #7300, opened 2020-09-03
Help: Fix vulnerabilities

mia-pi-git:


HoeenCoder:
This is a good start, not sure what else needs to be done but that can be discussed later.

PR #7304, opened 2020-09-03
Add a command to highlight users in bot html pages

mia-pi-git:
https://www.smogon.com/forums/threads/add-highlights-to-bot-pages.3669880/
Command name is tentative. Didn't like it, but couldn't think of a better one.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -232,6 +232,7 @@ export const commands: ChatCommands = {
 
 	highlighthtmlpage(target, room, user) {
 		target = target.trim();
+		if (!room) return this.requiresRoom();
 		if (!this.can('addhtml', null, room)) return false;
 		let [userid, pageid, highlight] = Utils.splitFirst(target, ',', 2);
 		pageid = `${user.id}-${toID(pageid)}`;
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -243,7 +243,7 @@ export const commands: ChatCommands = {
 			throw new Chat.ErrorMessage("This user is currently locked, so you cannot send them highlights.");
 		}
 
-		const buf = `|tempnotify|bot-${pageid}|A bot has notified you!|${highlight ? highlight : ''}`;
+		const buf = `|tempnotify|bot-${pageid}|${title} [from ${user.name}]|${highlight ? highlight : ''}`;
 		let targetConnections = [];
 		this.checkPMHTML(targetUser);
 		// try to locate connections that have requested the page recently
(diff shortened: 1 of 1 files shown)


Zarel:
"requested the page recently, otherwise the first connection in the list" is rather different from "only if they have the page open", isn't it? Am I missing something?

mia-pi-git:
Leaving this here too.

> "requested the page recently, otherwise the first connection in the list" is rather different from "only if they have the page open", isn't it? Am I missing something?

> [10:31:32] %mia pi: the last request page thing is only a way to figure out what connection to send it to
> [10:31:55] %mia pi: the highlight will only trigger if they have the page open since it's being sent to "view-bot-pageid-userid"
> [10:32:55] %mia pi: it's the same as how /ht list works

sparkychildcharlie:
Is it just me, or is there no permission check for this command.  As in users can abuse it to highlight spam other people once they figure out who has what page open? (or just randomly send ``|tempnotify|`` messages to pretty much everyone on the server - idk if that would still hl if room is nonexistent, but yeah...)

Zarel:
Hm. We might want to worry about notification spam.

If the client has separate "notification" and "bot notification" settings, that might help.

mia-pi-git:
I mean, if you don't want to be spammed, just close the page? I would also assume most bot owners who own a bot with the power to do this, would know not to.

Zarel:
I think it's pretty common to want to use a page but also be annoyed by notification spam.

...I might even make bot notifications opt-in. That's probably a good idea, actually. Put it in user settings?

server/chat-commands/admin.ts
```diff
@@ -235,6 +235,44 @@ export const commands: ChatCommands = {
...
+		if (!userid || !pageid || !target) return this.parse(`/help highlighthtmlpage`);
+		const targetUser = Users.get(userid);
+		if (!targetUser || !targetUser.connected) {
+			throw new Chat.ErrorMessage(`User ${this.targetUsername} is not currently online.`);
+		}
+		if (targetUser.locked && !this.user.can('lock')) {
+			throw new Chat.ErrorMessage("This user is currently locked, so you cannot send them highlights.");
+		}
+
+		const buf = `|tempnotify|bot-${pageid}|A bot has notified you!|${highlight ? highlight : ''}`;
```
Zarel, line 254:
```suggestion
		const buf = `|tempnotify|bot-${pageid}|A bot has notified you!|${highlight ? highlight : ''}`;
```

I think "A bot has notified you" is a pretty ugly notification message. Why not let the bot set it? Maybe add `[from ${botName}]` or something before or after it to be clearer.

Zarel:
This commit log is hella messy; did you mess up a merge?

mia-pi-git:
> This commit log is hella messy; did you mess up a merge?

Yes, apparently. I'll handle it when I'm back on a computer.

mia-pi-git:
Fixed, @Zarel.

Zarel:
I would still prefer if this was opt-in, to be clear. But I won't block on that.

PR #7305, opened 2020-09-03
Users: Fix bug with IP handling

AnnikaCodes:
This fixes the issue where global moderators/admins can't view the IPs of offline users with `/ip`. It should probably be monkeypatched (unless we want to restart).

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -1135,9 +1135,6 @@ export class User extends Chat.MessageContext {
 				for (const roomid of connection.inRooms) {
 					this.leaveRoom(Rooms.get(roomid)!, connection);
 				}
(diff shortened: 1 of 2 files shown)


Zarel:
I think there's some confusion with the "requires restart" tag; it's supposed to be for PRs that will crash a hotpatch, not PRs that don't take effect until after a restart. I'll rename it to be clearer.

Zarel:
Also thanks!

AnnikaCodes:
> I think there's some confusion with the "requires restart" tag; it's supposed to be for PRs that will crash a hotpatch, not PRs that don't take effect until after a restart. I'll rename it to be clearer.

Oh. Thanks for clearing it up!

PR #7306, opened 2020-09-03
Yeet myself

Lusamine:


server/chat-plugins/hosts.ts
```diff
@@ -10,7 +10,7 @@ import {GlobalPermission} from "../user-groups";
 
 const HOST_SUFFIXES = ['res', 'proxy', 'mobile'];
 
-const WHITELISTED_USERIDS = ['anubis'];
+const WHITELISTED_USERIDS = [''];
```
AnnikaCodes, line 13:
```suggestion
const WHITELISTED_USERIDS = [];
```

server/chat-plugins/hosts.ts
```diff
@@ -10,7 +10,7 @@ import {GlobalPermission} from "../user-groups";
 
 const HOST_SUFFIXES = ['res', 'proxy', 'mobile'];
 
-const WHITELISTED_USERIDS = ['anubis'];
+const WHITELISTED_USERIDS = [];
```
AnnikaCodes, line 13:
```suggestion
const WHITELISTED_USERIDS: ID[] = [];
```

Zarel:
For what it's worth, I'm sorry.

PR #7307, opened 2020-09-04
Searchlogs: Fix crash in date parsing

mia-pi-git:
Apparently this wasn't trimmed, so when the date was sliced later on, it fucked with the handling.
Also now has a safety against when you search logs for a month that doesn't have logs.


Zarel:
Thanks.

PR #7309, opened 2020-09-04
Log viewer: Escape usernames

AnnikaCodes:
People with HTML tags in their usernames can mess with chatlog formatting.
image


Zarel:
Thanks!

PR #7311, opened 2020-09-04
Support week-long roombans

mia-pi-git:
https://www.smogon.com/forums/threads/implement-a-weekban-command.3669889/#post-8588113

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -576,7 +576,10 @@ export const commands: ChatCommands = {
 	unmutehelp: [`/unmute [username] - Removes mute from user. Requires: % @ # &`],
 
 	rb: 'ban',
+	weekban: 'ban',
+	wrb: 'ban',
 	forceroomban: 'ban',
+	forceweekban: 'ban',
 	forcerb: 'ban',
 	roomban: 'ban',
 	b: 'ban',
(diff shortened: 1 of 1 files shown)


AnnikaCodes (commented):
I PMd you my thoughts on Discord.

mia-pi-git:
> Is there a reason there's no `weekban` alias?

I felt it'd be slightly out of pattern - i can add that, though.

Zarel:
We've gradually, over time, transitioned commands over from defaulting to global (back when we didn't have chatrooms) to defaulting to room.

At this point, I believe all commands either have no default (promote/demote) or default to room (banning). Weekban is a ban command, so I think if `/ban` defaults to roombans, there's no ambiguity for `/weekban`.

Zarel:
I would also add `/forceweekban`, and make `/forceweekban` and `/weekban` the defaults.

`/help ban` shows `/ban` and not `/roomban`, after all.

Zarel:
Speaking of `/help ban`, do you want to add this to `/help ban` while you're at it?

mia-pi-git:
> Speaking of `/help ban`, do you want to add this to `/help ban` while you're at it?
> I would also add `/forceweekban`, and make `/forceweekban` and `/weekban` the defaults.

Done. Removed /forceweekroomban and /weekroomban.

mia-pi-git:
Shit, mistake in push.

Zarel:
Thanks. I think maybe we should still have `/weekroomban` and maybe `/forceweekroomban` as aliases, though?

Also, how do you feel about `/wb` as an alias? "wb" often means "welcome back" so it might be weird, but I can't think of an actual conflict.

PR #7316, opened 2020-09-06
Trivia: Support specifying case sensitivity in searches

AnnikaCodes:
This is all xfix, I just TypeScripted it, implemented Zarel's suggestions from #6420, and fixed the merge conflicts.

Fixes #6420.


Zarel:
Thanks!

PR #7318, opened 2020-09-06
Trademarked: Update bans

KrisXV:
https://www.smogon.com/forums/threads/trademarked.3656980/page-4#post-8589098

complex bans arent supported for restrictions currently, so this is the best alternative


Zarel:
Thanks.

PR #7319, opened 2020-09-06
Weekban: Fix error message and add /wb alias

AnnikaCodes:
`/forceweekroomban` does not exist so having it in an error is not ideal. I'm adding `/wb` for consistency since `/roomban` is abbreviated to `/rb`


KrisXV:
Can you update `banhelp` as well to fix the bracket inconsistency (https://github.com/smogon/pokemon-showdown/pull/7319/files#diff-4a12d95620a16b64c1252efbbb897ea9R663)?

Zarel:
Thanks!

PR #7323, opened 2020-09-06
Auth: Fix changing a group's permission for commands

whalemer:
as far as I can tell, all uses of `checkCan` will provide `cmd`, so setting a group permission to change perms across lots of commands doesn't work at all.


Zarel:
Thanks.

PR #7325, opened 2020-09-06
Trivia: Add short alias for casesensitivesearch

AnnikaCodes:
Requested by Question Workshop staff.


Zarel:
Interesting. I'm guessing they use case-sensitivity more than expected. What do they use them for, out of curiosity?

PR #7326, opened 2020-09-07
Fix renaming battle rooms

AnnikaCodes:
This prevents "while searching, room ... expired" errors.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -974,7 +974,7 @@ export const commands: ChatCommands = {
 	hiddenroom: 'privateroom',
 	secretroom: 'privateroom',
 	publicroom: 'privateroom',
-	privateroom(target, room, user, connection, cmd) {
+	async privateroom(target, room, user, connection, cmd) {
 		room = this.requireRoom();
 		if (room.battle) {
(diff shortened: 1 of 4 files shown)


server/chat-commands/core.ts
```diff
@@ -370,7 +370,7 @@ export const commands: ChatCommands = {
 		if (this.cmd === 'mee' && /[A-Z-a-z0-9/]/.test(target.charAt(0))) {
 			return this.errorReply(this.tr`/mee - must not start with a letter or number`);
 		}
-		this.checkChat(`/${this.cmd} ${target || ''}`);
+		target = this.checkChat(`/${this.cmd} ${target || ''}`);
```
Zarel, line 373: This is probably a mistake.

Zarel:
The rest looks good and is ready for the next restart.

server/chat-commands/core.ts
AnnikaCodes: I don't think so? Without this, `/me` doesn't work.

mia-pi-git: I'm fixing this in #7327. You can remove it.

Zarel (approved):

PR #7329, opened 2020-09-07
Move Silvally-Water to PUBL

KrisXV:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/#post-8590185


Zarel:
Thanks!

PR #7330, opened 2020-09-07
Add "η" to the evasion filter

AnnikaCodes:
See https://replay.pokemonshowdown.com/gen8nationaldex-1182878993 for an example of how this can be used to evade.


Zarel:
Thanks!

PR #7331, opened 2020-09-07
Remove outdated getAwakeningValues function

urkerab:
`getAwakeningValues` used to be used by Let's Go's `rulesets.ts` to check that AVs were capped at 200. This was moved to `sim/team-validator.ts` by PR #5840 so now all that remains is a thick wrapper around `set.evs`, which can simply be removed, conveniently additionally increasing type safety.


Zarel:
Thanks.

PR #7333, opened 2020-09-08
Support adding friends

### LINKED ISSUE #1736: Friend list
Self-explanatory. An old implementation can be viewed at #607 and Zarel/Pokemon-Showdown-Client#126.

mia-pi-git:
Suggestions: https://www.smogon.com/forums/threads/friends-list.3650632/#post-8589151
Most of this is pretty straightforward, but the reason friend getting is handled in the loginfilter is that otherwise it would be done in effectively the same place, and this adds modularity so that the plugin can be removed. 
Max friends will probably be something we can increase, but i want to start it low to mitigate risk of load.
Closes #1736


Zarel:
You might want to `npm i @types/better-sqlite3`

AnnikaCodes (commented):
PMd you on Discord as requested

databases/schemas/friends.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS friends (
+	userid TEXT NOT NULL,
+	friend TEXT NOT NULL,
+	previous_name TEXT,
+	change_date INTEGER,
+	last_login INTEGER,
+	PRIMARY KEY (userid, friend) ON CONFLICT IGNORE
```
monsanto, line 7: Keep the schema as flexible as possible. Move `ON CONFLICT IGNORE` to inserts/updates where you specifically want that functionality

databases/schemas/friends.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS friends (
```
monsanto, line 1: No need for `IF NOT EXISTS`. Don't run this code every connection. Migrations will require more than this.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,623 @@
...
+		return {sent, received};
+	}
+	getRenames() {
+		// this is only run once, on initialization, so it doesn't need to be in statements
+		const results = this.database.prepare(
+			`SELECT friend, previous_name, change_date FROM friends WHERE previous_name IS NOT NULL`
+		).all();
+		for (const result of results) {
+			const {friend, previous_name, change_date} = result;
+			// delete old data after one week
```
monsanto, line 94: Why?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,623 @@
...
+ * Allows for adding and removing friends, as well as seeing their activity.
+ * Written by Mia-pi.
+ * @author mia-pi-git
+ */
+import * as Sqlite from 'better-sqlite3';
+import {Utils} from '../../lib/utils';
+import {FS} from '../../lib/fs';
+
+/** Max friends per user */
+const MAX_FRIENDS = 50;
```
monsanto, line 12: Seems low, if you're worried about DOS-ability I'd bump the limit for voices+ at least

server/chat-plugins/friends.ts
mia-pi-git: Because why keep around the name-change data forever?

server/users.ts
```diff
@@ -356,6 +361,9 @@ export class User extends Chat.MessageContext {
 	lastDisconnected: number;
 	lastConnected: number;
 	foodfight?: {generatedTeam: string[], dish: string, ingredients: string[], timestamp: number};
+	// We want this to be null on creation so that the database can be queried only once per user object.
+	friends!: Set<string>;
+	friendRequests!: {sent: Set<string>, received: Set<string>} | null;
```
Zarel, line 366: You should really only do this if these variables are set in the constructor, which they don't appear to be.

server/users.ts
```diff
@@ -226,6 +226,8 @@ export class Connection {
 	autojoins: string;
 	/** The last bot html page this connection requested, formatted as `${bot.id}-${pageid}` */
 	lastRequestedPage: string | null;
+	/** The current chat page the connection has open */
+	currentPage: string | null;
```
Zarel, line 230: This should probably be refactored out into its own PR.

mia-pi-git: Oh, yeah, i meant to do that. Forgot. Will do right now.

server/users.ts
mia-pi-git: No, but they're set immediately after login in the loginfilter.

Zarel: That's not good enough; it doesn't apply to guests, and the login process can take hundreds of milliseconds anyway. The `!` assertion is only safe if you set it in the same tick (i.e. you don't wait for any Promises or callbacks before setting it).

Zarel: I would recommend `friends: Set<string> | null`, and setting it to `null` in the constructor.

mia-pi-git: Alright.

mia-pi-git: Done.

mia-pi-git: Added bonus, this is hotpatchable with no monkeypatch now!

server/users.ts
mia-pi-git: Removed.

DaWoblefet:
Worked quite a bit with Mia concerning UI and behavior fixes. Has a couple kinks to work out that will depend on currentPage, but otherwise I really, really like this.

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
     "editor.insertSpaces": false,
-    "typescript.tsdk": "node_modules/typescript/lib",
+    "typescript.tsdk": "/Users/mia/Documents/GitHub/pokemon-showdown-dev/node_modules/typescript/lib",
```
monsanto, line 3: ?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,753 @@
...
+	private readonly statements: {[k: string]: Sqlite.Statement};
+	constructor() {
+		this.database = new Sqlite(`${__dirname}/../../databases/friends.db`);
+		this.renames = new Map();
+		// remove this when sqlite tables are automatically loaded
+		// check if the tables exist - if not; create them
+		try {
+			this.getRenames();
+		} catch (e) {
+			if (e.message.includes('no such table:')) {
```
monsanto, line 44: Instead of doing this I would check if `friends.db` exists, and if not, create it with the current schema. Eventually this will have to be generalized to "if friends.db exists, but its schema version is different than the current, run transitions"

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,753 @@
...
+			idle: [],
+			busy: [],
+			offline: [],
+		};
+		for (const friendID of [...user.friends].sort()) {
+			const friend = Users.get(friendID);
+			if (friend?.connected) {
+				friends[friend.statusType].push(friend.id);
+			} else {
+				friends.offline.push(toID(friendID));
```
monsanto, line 229: Either `toID(friendID)` is redundant or `friendID` is misnamed

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,753 @@
...
+		for (const userid of user.friends!) {
+			const curUser = Users.get(userid);
+			if (curUser?.settings.allowFriendNotifications) {
+				curUser.send(`|pm| &|${curUser.getIdentity()}|${message}`);
+			}
+		}
+	}
+	hasFriendship(userid1: string, userid2: string) {
+		userid1 = toID(userid1);
+		userid2 = toID(userid2);
```
monsanto, line 353: Why not just take IDs as parameters, instead of coercing them? Again, another situation where parameters are either misnamed or redundant

server/users.ts
```diff
@@ -356,6 +358,9 @@ export class User extends Chat.MessageContext {
 	lastDisconnected: number;
 	lastConnected: number;
 	foodfight?: {generatedTeam: string[], dish: string, ingredients: string[], timestamp: number};
+	// We want this to be null on creation so that the database can be queried only once per user object.
+	friends: Set<string> | null;
+	friendRequests: {sent: Set<string>, received: Set<string>} | null;
```
monsanto, line 363: Can you elaborate on why these have to be null? The code is littered with `.friends?`/etc

server/chat-plugins/friends.ts
monsanto: Can you elaborate on the purpose of name change data? And, why only 1 previous name?

.vscode/settings.json
mia-pi-git: Accident, forgot to uncommit that.

server/users.ts
mia-pi-git: friendRequests is null because it's deleted if you're disabling friendRequests. 
friends is initialized as null as a way to slow how often the db is queried (the loginfilter returns undefined immediately if user.friends exists already( but is set after login.

.vscode/settings.json
monsanto: Was there a problem that required changing this file, or just some tool acting up?

server/chat-plugins/friends.ts
mia-pi-git: This doesn't work because `better-sqlite3` creates the db file on creation of the class (eg `new require('better-sqlite3')('test.db')`` creates the file `test.db`) even if it's not there already.

server/chat-plugins/friends.ts
mia-pi-git: Redid.

server/chat-plugins/friends.ts
mia-pi-git: I suppose you could initialize the table before the rest of the database, but that seems messier than what's there now.

monsanto: [See `fileMustExist`](https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/api.md#new-databasepath-options) (also, even if this doesn't work for you, just check before creating the database object?)

mia-pi-git: Done.

data/learnsets.ts
```diff
@@ -2812,7 +2812,7 @@ export const Learnsets: {[speciesid: string]: LearnsetData} = {
 			echoedvoice: ["7M"],
 			electricterrain: ["8M"],
 			electroball: ["8M", "8L12", "7L13"],
-			electroweb: ["8M", "7T"],
+			electroweb: ["8M", "8S1", "7T"],
```
monsanto, line 2815: Please remove from this PR

server/chat-plugins/friends.ts
mia-pi-git: So that you don't see a new name on your friends list and get confused who it might be. It's only one previous name because you should only need one, most likely. If not, your friends will probably find out anyway.

data/learnsets.ts
mia-pi-git: Ugh.

mia-pi-git: Fixed.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,748 @@
...
+		const prefix = `|pm|&|${user.getIdentity()}|`;
+		user.send(`${prefix}You have ${pendingCount} friend requests pending!`);
+		user.send(`${prefix}/raw <button class="button" name="send" value="/j view-friends-received">View</button></div>`);
+	}
+	notifyConnection(user: User) {
+		const message = `/nonotify Your friend ${Utils.escapeHTML(user.name)} has just connected!`;
+		for (const userid of user.friends!) {
+			const curUser = Users.get(userid);
+			if (curUser?.settings.allowFriendNotifications) {
+				curUser.send(`|pm| &|${curUser.getIdentity()}|${message}`);
```
monsanto, line 342:
```suggestion
				curUser.send(`|pm|&|${curUser.getIdentity()}|${message}`);
```

KrisXV:
Why do I keep getting shadow review requests on this?

monsanto:
I will be ready to look at this more closely soon. Can you please add a `Config` setting that disables this feature by default (Annika is using something like `Config.usesqlitemodlog`), and can you disable this feature initially for non-global staff? I'd like to be able to test this a bit in production & potentially blast the database, and I don't want regular users to lose their data.

mia-pi-git:
> I will be ready to look at this more closely soon. Can you please add a `Config` setting that disables this feature by default (Annika is using something like `Config.usesqlitemodlog`), and can you disable this feature initially for non-global staff? I'd like to be able to test this a bit in production & potentially blast the database, and I don't want regular users to lose their data.

Done. I said this on discord but i'm leaving it here too.
> set it to Config.usesqlitefriends, which can be set to an auth level to prevent users below that auth level from using it

(or just left empty to disable it.)

mia-pi-git:
(test failure is in test/sim/, so unrelated to this PR)

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,910 @@
...
+ * Written by Mia.
+ * @author mia-pi-git
+ */
+import * as Sqlite from 'better-sqlite3';
+import {Utils} from '../../lib/utils';
+import {FS} from '../../lib/fs';
+import {QueryProcessManager} from '../../lib/process-manager';
+import {Repl} from '../../lib/repl';
+import {Chat} from '../chat';
+import {Dex} from '../../sim';
```
monsanto, line 13: Why would a friends list need this import?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,910 @@
...
+			sender.friends.add(receiverID);
+		}
+		if (receiver?.friendRequests) {
+			if (!receiver.friends) receiver.friends = new Set();
+			receiver.friends.add(senderID);
+		}
+
+		return Promise.all([
+			PM.query({statement: 'add', data: {userid: senderID, friend: receiverID, login: Date.now()}, type: 'run'}),
+			PM.query({statement: 'add', data: {userid: receiverID, friend: senderID, login: Date.now()}, type: 'run'}),
```
monsanto, line 333: Multiple statements need to be done in transaction

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,910 @@
...
+			}
+		}
+		newUser.friends = oldUser.friends;
+		oldUser.friends?.clear();
+
+		this.renames.set(oldID, newID);
+		return Promise.all([
+			PM.query({statement: 'renameFriend', data: {oldID, newID}, type: 'run'}),
+			PM.query({statement: 'renameUserid', data: {oldID, newID}, type: 'run'}),
+			PM.query({statement: 'rename', data: [oldID, newID, Date.now()], type: 'run'}),
```
monsanto, line 437: Multiple statements need to be done in transaction

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,910 @@
...
+		if (!receiverID) throw new Chat.ErrorMessage(`Invalid name.`);
+		if (receiverID === user.id) throw new Chat.ErrorMessage(`You cannot friend yourself.`);
+		const sentRequests = user.friendRequests?.sent;
+		if (await this.hasFriendship(user.id, receiverID)) {
+			throw new Chat.ErrorMessage(`You are already friends with '${receiverID}'.`);
+		}
+		if (sentRequests?.has(receiverID)) {
+			throw new Chat.ErrorMessage(`You have already sent a friend request to '${receiverID}'.`);
+		}
+		if (sentRequests!.size >= MAX_REQUESTS) {
```
monsanto, line 158: Don't bypass the typechecker with `!` unless you absolutely have to

monsanto (commented):
I should have some time to look at this more tomorrow. Here is some quick feedback.

Also:
- Need to gate on `usesqlite` in addition to `usesqlitefriends`
- I'm probably going to complain about using `?` on everything, it is a sign that the types need to be tightened up

server/chat-plugins/friends.ts
mia-pi-git: It's needed for the QueryProcessManager.

mia-pi-git:
> * Need to gate on `usesqlite` in addition to `usesqlitefriends`

This is done.
> * I'm probably going to complain about using `?` on everything, it is a sign that the types need to be tightened up

I'll look into removing some of the `?.` uses - though it should be noted some of it is necessary due to the fact that both `User#friendRequests` and `User#friends` can be null. In fact, they're initialized as null so that the loginfilter can tell "this is a new user object and we should be querying their friends / friend requests", which we want to do as little as possible.

server/chat-plugins/friends.ts
monsanto: (Generally the reviewer resolves these when they accept the explanation)

server/users.ts
```diff
@@ -359,6 +362,9 @@ export class User extends Chat.MessageContext {
 	lastDisconnected: number;
 	lastConnected: number;
 	foodfight?: {generatedTeam: string[], dish: string, ingredients: string[], timestamp: number};
+	// We want this to be null on creation so that the database can be queried only once per user object.
+	friends: Set<string> | null;
+	friendRequests: {sent: Set<string>, received: Set<string> | null} | null;
```
monsanto, line 367: I don't think these are good attributes to have on the `User` object.

These are caches, not the source of truth. The database is the source of truth. Ideally, you would query the database every time you wanted to know a friend related fact. In this situation you would never use these attributes.

You have decided that it would be too expensive to query the database every single time. I don't know if that is true or not: SQLite keeps its own cache, the query planning is done beforehand with `prepare`, and there are no IPC costs. But assuming it is, that optimization should not bleed into the interface--nothing that uses the friends list should be able to tell whether you are using this optimization or not.

If you need a cache, use a `Map<User, ...>` local to the chat plugin. And don't have the rest of the code interface with the map directly; when you ask for friends list data, it should check if the data is already present in the cache, and if not, load from the database and store it in the cache.

I suspect this will remove a lot of the `?` problems, and will make it easier to remove the caching/have cache expiration policies if necessary.

monsanto: In the interest of merging in a timely manner, I'd recommend just doing this first without any caching, especially since a limited # of people can use this to start with. Please reach out on Discord if you have any questions.

mia-pi-git: Done. It now checks the database exclusively.

monsanto:
This is currently blocking on properly using transactions. Further elaboration in PM

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,914 @@
...
+			statements.rename.run(oldID, newID, Date.now());
+		}
+		return {result: []};
+	},
+	add: requests => {
+		const addQuery = statements.add;
+		for (const request of requests) {
+			const [to, from] = request;
+			addQuery.run({userid: to, friend: from, login: Date.now()});
+			addQuery.run({userid: from, friend: to, login: Date.now()});
```
monsanto, line 836: It might not matter, but `Date.now()` isn't guaranteed the return the same value both times, I'd assign this to a variable once so all the rows have the same date

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,915 @@
...
+		}
+		buf += `</h3> `;
+
+		for (const key in friends) {
+			const friendArray = categorized[key].sort();
+			if (friendArray.length === 0) continue;
+			buf += `<h4>${STATUS_TITLES[key]} (${friendArray.length})</h4>`;
+			for (const friend of friendArray) {
+				buf += `<div class="pad"><div>`;
+				buf += await this.displayFriend(friend);
```
monsanto, line 229: `displayFriend` calls `getLastLogin`, so this does `O(n)` queries. Surely you already have the last login handy from `getFriends`?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,915 @@
...
+		if (!friendID || !userid) throw new Chat.ErrorMessage(`Invalid usernames supplied.`);
+
+		const user = Users.getExact(userid);
+		if (user) {
+			const userFriends = await this.getFriends(user);
+			if (!userFriends.has(friendID)) throw new Chat.ErrorMessage(`You do not have '${friendID}' friended.`);
+		}
+		return this.query({
+			statement: 'delete', type: 'run', data: {userid},
+		});
```
monsanto, line 289: Delete depends on `getFriends` so, if this was the right thing to do, it would need to be in a transaction.

I don't think `delete` is doing the right thing though? Its definition is `DELETE FROM friends WHERE userid = $userid OR friend = $userid` which will delete the other persons entire friend list.

What you want to do is remove where `userid = $userid AND friend = $friend`, skip the `getFriends` check, and check if it actually removed a friend with [`changes`](https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/api.md#runbindparameters---object) property to do the error message

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,915 @@
...
+		// and when the user the request is sent to / sent from logs in.
+		const results = this.database.prepare(`SELECT * FROM friend_requests`).all();
+		const removed = [];
+		for (const request of results) {
+			if ((Date.now() - request.sent_at) > REQUEST_EXPIRY_TIME) {
+				// expires after a month
+				await this.removeRequest(request.receiver, request.sender);
+				removed.push(request);
+			}
+		}
```
monsanto, line 82: Correlated queries must be in transaction, and you're making `O(n)` queries here. Just do a `DELETE from friend_requests WHERE <expiry condition>`

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,915 @@
...
+			});
+			return {sent, received: null};
+		}
+		const sentResults = await this.query({
+			statement: 'getSent', type: 'all', data: [user.id],
+		});
+		for (const request of sentResults) {
+			if ((Date.now() - request.sent_at) > (30 * 24 * 60 * 60 * 1000)) {
+				// expires after a month
+				await this.removeRequest(request.receiver, request.sender);
```
monsanto, line 108: `O(n)` queries, correlation, etc.

Honestly, you might just want some sort of cron service in PS to do tasks like this.

mia-pi-git:
Done, @monsanto. Chose to go with a sync action done on startup for clearing old request - the rest of the time, if it's lasted too long, getRequest will ignore it.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+		const senderID = toID(senderName);
+		const receiverID = toID(receiverName);
+		if (!senderID) throw new Chat.ErrorMessage(`Invalid sender username.`);
+		if (!receiverID) throw new Chat.ErrorMessage(`Invalid receiver username.`);
+
+		return this.query({
+			statement: 'deleteRequest', data: [senderID, receiverID], type: 'run',
+		});
+	}
+	async approveRequest(receiver: User, senderName: string) {
```
monsanto, line 184: What is the point of taking a `User` object here over an ID?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+		const receiverID = toID(receiverName);
+		const senderID = toID(senderName);
+		if (!receiverID) throw new Chat.ErrorMessage(`Invalid user.`);
+		return this.query({
+			type: 'transaction', statement: 'add', data: [receiverID, senderID],
+		});
+	}
+	async removeFriend(userName: string, friendName: string) {
+		const userid = toID(userName);
+		const friendID = toID(friendName);
```
monsanto, line 281: I think I left a similar comment on this PR before, but maybe I didn't properly explain.

Internally, we work entirely with IDs, right? The only time you can encounter something that isn't an ID is when interfacing with the user. So do `toID` once, when parsing human input (i.e. only @ in the chat commands). Any library functions should take `ID` as their type instead of `string`, and not redundantly call `toID` on their inputs.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+			Utils.html`<i>${name}</i> <small>(${statusType})</small>`;
+		buf += `<br />`;
+
+		const oldName = this.renames.get(userid);
+		if (oldName) {
+			buf += Utils.html`<small>(recently renamed from ${oldName})</small><br />`;
+		}
+
+		const curUser = Users.get(userid); // might be an alt
+		if (user?.connected) {
```
monsanto, line 257: You assign `user?.connected` to the variable `connected` earlier, pick one or the other

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+			});
+			return {sent, received: null};
+		}
+		const sentResults = await this.query({
+			statement: 'getSent', type: 'all', data: [user.id],
+		});
+		for (const request of sentResults) {
+			if ((Date.now() - request.sent_at) > (30 * 24 * 60 * 60 * 1000)) {
+				continue; // ignore this for now, it'll be caught by the expiry function later
+			}
```
monsanto, line 111: Ditto

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+		}
+
+		return buf;
+	}
+	displayFriend(userid: string, login?: number) {
+		const user = Users.getExact(userid); // we want this to be exact
+		const connected = user?.connected;
+		const name = Utils.escapeHTML(user ? user.name : userid);
+		const statusType = connected ?
+			`<strong style="color:${STATUS_COLORS[user!.statusType]}">\u25C9 ${STATUS_TITLES[user!.statusType]}</strong>` :
```
monsanto, line 244: Don't use the `!` operator.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+const STATUS_TITLES: {[k: string]: string} = {
+	online: 'Online',
+	idle: 'Idle',
+	busy: 'Busy',
+	offline: 'Offline',
+};
+
+export const Friends = new class {
+	/** `Map<oldID, newID> */
+	renames: Map<string, string>;
```
monsanto, line 48: Appears to be a redundant cache for the database `friend_renames` table, my previous comment applies

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,911 @@
...
+			sendingUser.send(`|pm|${fromIdentity}|${toIdentity}|${message}`);
+		}
+		if (typeof receivingUser === 'object' && receivingUser) {
+			receivingUser.send(`|pm|${fromIdentity}|${toIdentity}|${message}`);
+		}
+	}
+	async notifyPending(user: User) {
+		if (user.settings.blockFriendRequests) return;
+		const friendRequests = await this.getRequests(user);
+		const pendingCount = friendRequests.received?.size;
```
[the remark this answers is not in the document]
mia-pi-git: I had previously made it `null` if the user was blocking friend requests, but realized that was a bit redundant. Removed.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+			statements.deleteRequest.run(to, from);
+		}
+		return {result: []};
+	},
+	visualData: requests => {
+		for (const request of requests) {
+			const [userid, limit] = request;
+			let data = statements.get.all(userid, limit);
+			data = data.map(f => {
+				const isAllowingLogin = statements.hidingLogin.get(f.friend);
```
monsanto, line 867: A straightforward left join in `get` removes the extra queries, and probably the need for `visualData` at all

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+			if (newRequests.length || sentRequests.length) {
+				throw new Chat.ErrorMessage(`You have pending friend requests, and so cannot transfer friends.`);
+			}
+			if (!oldFriends.length) {
+				throw new Chat.ErrorMessage(`You have no friends to transfer.`);
+			}
+			if (oldFriends.includes(newID)) {
+				// unfriend the two
+				statements.delete.run({userid: newID});
+			}
```
monsanto, line 800: There is no need to fetch all these rows. You can do `SELECT count(*)` for `newRequests`/`sentRequests`, but it is unclear to me why you can't transfer with stuff pending? For `oldFriends`, just do the delete unconditionally

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+			statements.insertRequest.run(senderID, receiverID, Date.now());
+		}
+		return {result: []};
+	},
+	add: requests => {
+		for (const request of requests) {
+			const [senderID, receiverID] = request;
+			const last_login = Date.now();
+			statements.add.run({userid: senderID, friend: receiverID, last_login});
+			statements.add.run({friend: senderID, userid: receiverID, last_login});
```
monsanto, line 839: What is the point of storing this relationship twice? Can we adjust the schema/queries so only storing it once is necessary?

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+		return {result: []};
+	},
+	accept: requests => {
+		for (const request of requests) {
+			const [senderID, receiverID] = request;
+			const receivedRequests = statements.getReceived.all(receiverID).map(item => item.sender);
+			if (!receivedRequests.includes(senderID)) {
+				throw new Chat.ErrorMessage(`You have not received a friend request from '${senderID}'.`);
+			}
+			TRANSACTIONS.removeRequest([request]);
```
monsanto, line 850: Don't fetch all rows to check if something is present. In this case, you can just check the # of rows deleted by `removeRequest` to report if it was a success

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+				throw new Chat.ErrorMessage(`You have already sent a friend request to '${receiverID}'.`);
+			}
+			if (sentRequests.length >= MAX_REQUESTS) {
+				throw new Chat.ErrorMessage(
+					`You already have ${MAX_REQUESTS} outgoing friend requests. Use "/friends view sent" to see your outgoing requests.`
+				);
+			}
+			if (friends.includes(receiverID)) {
+				throw new Chat.ErrorMessage(`You have already friended '${receiverID}'.`);
+			}
```
monsanto, line 829: Don't fetch rows to check length/set membership

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,905 @@
...
+			const receivedRequests = statements.getReceived.all(receiverID).map(item => item.sender);
+			if (!receivedRequests.includes(senderID)) {
+				throw new Chat.ErrorMessage(`You have not received a friend request from '${senderID}'.`);
+			}
+			TRANSACTIONS.removeRequest([request]);
+			TRANSACTIONS.add([request]);
+		}
+		return {result: []};
+	},
+	removeRequest: requests => {
```
monsanto, line 855: This doesn't appear to be used anywhere with more than 1 request. If you need this functionality, how about a generic message to run a # of queries inside a transaction?

databases/schemas/friends.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE friends (
+	user1 TEXT NOT NULL,
+	user2 TEXT NOT NULL,
+	PRIMARY KEY (userid, friend)
```
monsanto, line 4: Can you please write a test that exercises the schema?

mia-pi-git:
@Zarel bump :D?

mia-pi-git:
Bump?

lib/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+	readonly cache: {[k: string]: {data: T, lastCache: number}};
+	expiryTime: number;
+	dataFetcher: (key: string) => T | Promise<T>;
+	constructor(fetcher: (key: string) => T | Promise<T>, invalidateTime = CACHE_EXPIRY_TIME) {
+		this.cache = {};
+		this.expiryTime = invalidateTime;
+		this.dataFetcher = fetcher;
+	}
+	// todo make this only return T
+	// <T>roubling it doesn't, since that was the original intent, but for now this will do
```
AnnikaCodes, line 19: Shouldn't this return `Promise<T>` (but the promise will resolve immediately if it's pulled from cache)?

package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@swc/core": "^1.2.61",
     "@types/better-sqlite3": "^5.4.0",
+    "node-oom-heapdump": "^2.2.0",
```
AnnikaCodes, line 9:
```suggestion
```

server/chat-commands/core.ts
```diff
@@ -716,7 +716,11 @@ export const commands: Chat.ChatCommands = {
 		} else if (target === 'autoconfirmed' || target === 'trusted' || target === 'unlocked') {
 			user.settings.blockPMs = target;
 			target = this.tr(target);
-			this.sendReply(this.tr`You are now blocking private messages, except from staff and ${target} users.`);
+			this.sendReply(this.tr `You are now blocking private messages, except from staff and ${target} users.`);
+		} else if (target === 'friends') {
+			user.settings.blockPMs = target;
+			target = this.tr(target);
```
AnnikaCodes, line 722:
```suggestion
```

server/chat-commands/core.ts
```diff
@@ -716,7 +716,11 @@ export const commands: Chat.ChatCommands = {
...
+			this.sendReply(this.tr `You are now blocking private messages, except from staff and friended users.`);
```
AnnikaCodes, line 723:
```suggestion
			this.sendReply(this.tr`You are now blocking private messages, except from staff and friends.`);
```

lib/cache.ts
mia-pi-git: No, it needs to be explicitly synchronous. The whole point of this is to not have to propagate awaits/Promises.

AnnikaCodes: Okay...
```ts
while (!variable) {
    variable = cache.get(key);
}
```
seems like a lot of jank for retrieving from the cache compared to just propagating a promise (which will resolve immediately some of the time anyway).

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,539 @@
...
+		const friends = await Chat.Friends.getFriends(userid);
+		const categorized: {[k: string]: string[]} = {
+			online: [],
+			idle: [],
+			busy: [],
+			offline: [],
+		};
+		const loginTimes: {[k: string]: number} = {};
+		for (const {friend: friendID, last_login, allowing_login: hideLogin} of [...friends].sort()) {
+			const friend = Users.get(friendID);
```
AnnikaCodes, line 59: Won't this show users' alts? I usually use alts to get away from people.

server/chat-plugins/friends.ts
```diff
@@ -0,0 +1,539 @@
...
+			buf += ` (${Chat.toDurationString(Date.now() - login, {precision: 1})} ago)`;
+		} else if (typeof login === 'string') {
+			buf += `${login}`;
+		}
+		buf = `<div class="infobox">${buf}</div>`;
+		return toLink(buf);
+	}
+	checkCanUse(context: Chat.CommandContext | Chat.PageContext) {
+		const user = context.user;
+		if (user.locked || user.namelocked || user.semilocked || user.permalocked) {
```
AnnikaCodes, line 142: Why check `.permalocked`? permalocked users will always be locked anyway.

mia-pi-git: Being extra careful, mostly.

AnnikaCodes: If permalocked users don't have `.locked` truthy, there is a much larger issue than them being able to use the friends list.

server/chat.ts
```diff
@@ -1099,9 +1100,16 @@ export class CommandContext extends MessageContext {
 					const groupName = Config.groups[Config.pmmodchat] && Config.groups[Config.pmmodchat].name || Config.pmmodchat;
 					throw new Chat.ErrorMessage(this.tr`On this server, you must be of rank ${groupName} or higher to PM users.`);
 				}
-				if (targetUser.settings.blockPMs &&
-					(targetUser.settings.blockPMs === true || !Users.globalAuth.atLeast(user, targetUser.settings.blockPMs)) &&
-					!user.can('lock') && targetUser.id !== user.id) {
+				const targetFriends = Chat.Friends.cache.get(targetUser.id);
```
AnnikaCodes, line 1103: So when the cache expires your friends will be randomly unable to PM you and then once the DB query finishes they can PM you? That seems confusing.

lib/cache.ts
mia-pi-git: I explained in dms, but I'll post for posterity. This is explicitly made so that no async/await is used, so we don't hold up on PMs.

server/chat.ts
mia-pi-git: When it expires, we don't remove the data - we merely make a new DB request to make it current.

server/chat-plugins/friends.ts
mia-pi-git: No, it won't. In displayFriend it checks getExact, so if they're on an alt, it says they're on an alt (but not the name of said alt). This is the same info you can get with /user.

server/chat.ts
AnnikaCodes: Ah. Wouldn't this still prevent the sending of PMs until that first request goes through?

mia-pi-git: It's set up so that when you accept a request, or otherwise add a friend, the cache is updated automatically. If that hasn't resolved by the time you type out and send a PM, there's likely another problem.

AnnikaCodes: OK

mia-pi-git: I just edited it to await the update of the cache for each user. Should be better.

server/friends.ts
```diff
@@ -0,0 +1,398 @@
...
+		case 'transaction':
+			result.result = transactions[statement]([data]);
+			break;
+		case 'all':
+			result.result = statements[statement].all(data);
+			break;
+		}
+	} catch (e) {
+		if (!e.name.endsWith('FailureMessage')) {
+			result.error = 'Sorry! The database process crashed. We\'ve been notified and will fix this.';
```
AnnikaCodes, line 357:
```suggestion
			result.error = "Sorry! The database process crashed. We've been notified and will fix this.";
```

server/room-battle.ts
```diff
@@ -1357,7 +1357,6 @@ export const PM = new ProcessManager.StreamProcessManager(module, () => new Room
 if (!PM.isParentProcess) {
 	// This is a child process!
 	global.Config = require('./config-loader').Config;
-	global.Chat = require('./chat').Chat;
```
AnnikaCodes, line 1360: is Chat required here?

mia-pi-git: It is not.

AnnikaCodes:
This seems fine at first look.

monsanto:
This has seen a ton of review and according to Mia has been extensively tested & hotpatchable. I'm sure there are problems but they can be dealt with post-merge. Go ahead and merge this when you feel confident

AnnikaCodes:
OK let's see how this goes.

server/chat-commands/core.ts
```diff
@@ -716,7 +716,10 @@ export const commands: Chat.ChatCommands = {
 		} else if (target === 'autoconfirmed' || target === 'trusted' || target === 'unlocked') {
 			user.settings.blockPMs = target;
 			target = this.tr(target);
-			this.sendReply(this.tr`You are now blocking private messages, except from staff and ${target} users.`);
+			this.sendReply(this.tr `You are now blocking private messages, except from staff and ${target} users.`);
```
Zarel, line 719: Surprised eslint didn't catch this.

PR #7334, opened 2020-09-08
Fix visual bug with /hiderank in PMs

AnnikaCodes:
image

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -747,7 +747,7 @@ export const commands: ChatCommands = {
 		}
 		user.tempGroup = group;
 		user.updateIdentity();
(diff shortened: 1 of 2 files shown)


server/chat-commands/core.ts
```diff
@@ -747,7 +747,7 @@ export const commands: ChatCommands = {
 		}
 		user.tempGroup = group;
 		user.updateIdentity();
-		this.sendReply(`|c|~|${this.tr`Your temporary group symbol is now`} \`\`${user.tempGroup}\`\`.`);
+		this.sendReply(`|c|~|/text ${this.tr`Your temporary group symbol is now`} \`\`${user.tempGroup}\`\`.`);
```
Zarel, line 750: Should this be fixed inside `sendReply`? I assume so

AnnikaCodes: That works, yeah

Zarel:
tyty

PR #7338, opened 2020-09-08
Add a plugin to repeat phrases in chatrooms

AnnikaCodes:
This is a very common functionality used by room staff to disseminate information, and I think that given the recent issues caused by losing Kid A, making this into a chat plugin makes sense.

### DIFF
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -34,9 +34,9 @@ export const commands: ChatCommands = {
 	repeat(target, room, user) {
 		room = this.requireRoom();
 		this.checkCan('mute', null, room);
-		const [intervalText, message] = this.splitOne(target);
-		const interval = parseInt(intervalText);
-		if (isNaN(interval) || !/[0-9]{1,}/.test(intervalText) || interval < 1) {
+		const [intervalString, message] = this.splitOne(target);
+		const interval = parseInt(intervalString);
+		if (isNaN(interval) || !/[0-9]{1,}/.test(intervalString) || interval < 1) {
 			return this.errorReply(this.tr`You must specify a numerical interval of at least 1 minute.`);
 		}
 
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -75,7 +75,11 @@ export const pages: PageTable = {
 			html += `<tr><td>${repeat.phrase}</td><td>${this.tr`every ${minutes} minute(s)`}</td>`;
 			html += `<td><button class="button" name="send" value="/removerepeat ${repeat.phrase}">${this.tr`Remove`}</button></td>`;
 		}
-		html += `</table></div>`;
+		html += `</table>`;
+		if (user.can("editroom", null, room)) {
+			html += `<br /><button class="button" name="send" value="/removeallrepeats">${this.tr`Remove all repeats`}</button>`;
+		}
+		html += `</div>`;
 		return html;
 	},
 };
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -37,7 +37,7 @@ export function removeRepeat(room: BasicRoom, phrase: string) {
 
 function runRepeat(room: BasicRoom, repeat: RepeatedPhrase) {
 	if (!repeats.has(room.roomid)) {
-		repeats.set(room.roomid, new Map<string, NodeJS.Timeout | null>());
+		repeats.set(room.roomid, new Map());
 	}
 
 	const displayRepeat = (targetRoom: BasicRoom, phrase: string) => {
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -15,7 +15,7 @@ export const pages: PageTable = {
 		const roomid = args[0];
 		this.title = `Repeated phrases in ${roomid}`;
 		const room = Rooms.get(roomid);
-		if (!room) throw new Chat.ErrorMessage(`<h2>${this.tr`No such room: "${roomid}".`}</h2>`);
+		const room = this.requireRoom();
 		this.checkCan("mute", null, room);
 		if (!room.settings.repeats) return `<h1>${this.tr`There are no repeated phrases in ${roomid}.`}</h1>`;
 		let html = `<div class="ladder pad"><h1>${this.tr`Repeated phrases in ${roomid}`}</h1>`;
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -1,0 +1,108 @@
+/**
+ * Chat plugin for repeating messages in chat
+ * Based on bot functionality from Kid A and Expecto Botronum
+ * @author Annika
+ */
+
+export interface RepeatedPhrase {
+	phrase: string;
+	/** interval in milliseconds */
+	interval: number;
+}
+
+export const pages: PageTable = {
+	repeats(args, user) {
+		const roomid = args[0];
+		this.title = `Repeated phrases in ${roomid}`;
+		const room = Rooms.get(roomid);
+		if (!room) throw new Chat.ErrorMessage(`<h2>${this.tr`No such room: "${roomid}".`}</h2>`);
+		this.checkCan("mute", null, room);
+		if (!room.settings.repeats) return `<h1>${this.tr`There are no repeated phrases in ${roomid}.`}</h1>`;
+		let html = `<div class="ladder pad"><h1>${this.tr`Repeated phrases in ${roomid}`}</h1>`;
+		html += `<table><tr><th>${this.tr`Phrase`}</th><th>${this.tr`Interval`}</th><th>${this.tr`Action`}</th>`;
+		html += `<button class="button" name="send" value="/join view-repeats-${roomid}" style="float: right">`;
+		html += `<i class="fa fa-refresh"></i> ${this.tr`Refresh`}</button>`;
+		for (const repeat of room.settings.repeats) {
+			const minutes = repeat.interval / (60 * 1000);
+			html += `<tr><td>${repeat.phrase}</td><td>${this.tr`every ${minutes} minute(s)`}</td>`;
+			html += `<td><button class="button" name="send" value="/removerepeat ${repeat.phrase}">${this.tr`Remove`}</button></td>`;
+		}
+		html += `</table></div>`;
+		return html;
+	},
+};
+
+export const commands: ChatCommands = {
+	repeat(target, room, user) {
+		room = this.requireRoom();
+		this.checkCan('mute', null, room);
+		const [intervalText, message] = this.splitOne(target);
+		const interval = parseInt(intervalText);
+		if (isNaN(interval) || !/[0-9]{1,}/.test(intervalText) || interval < 1) {
+			return this.errorReply(this.tr`You must specify a numerical interval of at least 1 minute.`);
+		}
+
+		if (room.hasRepeat(message)) {
+			return this.errorReply(this.tr`The phrase "${message}" is already being repeated in this room.`);
+		}
+
+		room.addRepeat({
+			phrase: message,
+			interval: interval * 60 * 1000, // convert to milliseconds
+		});
+
+		this.modlog('REPEATPHRASE', null, `every ${interval} minute${Chat.plural(interval)}: "${message}"`);
+		this.privateModAction(
+			room.tr`${user.name} set the phrase "${message}" to be repeated every ${interval} minute(s).`
+		);
+	},
+	repeathelp() {
+		this.runBroadcast();
+		this.sendReplyBox(
+			`<code>/repeat [minutes], [phrase]</code>: repeats a given phrase every [minutes] minutes.<br />` +
+			`<code>/removerepeat [phrase]</code>: removes a repeated phrase.<br />` +
+			`/viewrepeats [optional room] - Displays all repeated phrases in a room.<br />` +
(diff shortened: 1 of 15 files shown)


KrisXV:
A ton of rooms would benefit from an optimized rfaq functionality that automatically posts boxes for sanitized PS markdown (similar to what was in my code) instead of just adding a line from the server to call rfaq commands etc.

AnnikaCodes:
> A ton of rooms would benefit from an optimized rfaq functionality that automatically posts boxes for sanitized PS markdown (similar to what was in my code) instead of just adding a line from the server to call rfaq commands etc.

Yeah, I believe @mia-pi-git is refactoring room FAQs, so I don't want to add that now, but I think that a way to repeat rfaqs would be a desirable end goal.

KrisXV:
I don't think this is something that has to be blocked by a rfaq refactor. There are things constantly used in RoA and NatDex that would prefer to not exist at all as rfaqs and were only rfaqs previously due to bot limitations.

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,108 @@
...
+	/** interval in milliseconds */
+	interval: number;
+}
+
+export const pages: PageTable = {
+	repeats(args, user) {
+		const roomid = args[0];
+		this.title = `Repeated phrases in ${roomid}`;
+		const room = Rooms.get(roomid);
+		if (!room) throw new Chat.ErrorMessage(`<h2>${this.tr`No such room: "${roomid}".`}</h2>`);
```
mia-pi-git, line 18: These two lines can just use this.requireRoom().

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,108 @@
...
+		html += `</table></div>`;
+		return html;
+	},
+};
+
+export const commands: ChatCommands = {
+	repeat(target, room, user) {
+		room = this.requireRoom();
+		this.checkCan('mute', null, room);
+		const [intervalText, message] = this.splitOne(target);
```
mia-pi-git, line 39: Why not call this intervalNum? intervalText is misleading for something that's going to be a number.

Zarel: I mean I understand. It's the text version to be parsed into a thing later.

AnnikaCodes: I wanted to indicate that it was a string (as opposed to `interval` which is a number). Maybe `intervalString`?

AnnikaCodes:
> I don't think this is something that has to be blocked by a rfaq refactor. There are things constantly used in RoA and NatDex that would prefer to not exist at all as rfaqs and were only rfaqs previously due to bot limitations.

Oh, I think I misunderstood you. Are you just suggesting having repeats be in text boxes rather than chat lines? That could be a good idea, although using the FAQ system would make it easy to broadcast them when a user asked, etc, rather than only having them be accessible through the repeat.

server/chat-plugins/repeats.ts
mia-pi-git: to me, at least, it reads like the text you want to set in the interval. But, that's just me.
intervalString sounds better, I agree.

KrisXV:
> > I don't think this is something that has to be blocked by a rfaq refactor. There are things constantly used in RoA and NatDex that would prefer to not exist at all as rfaqs and were only rfaqs previously due to bot limitations.
> 
> Oh, I think I misunderstood you. Are you just suggesting having repeats be in text boxes rather than chat lines? That could be a good idea, although using the FAQ system would make it easy to broadcast them when a user asked, etc, rather than only having them be accessible through the repeat.

Yeah i was suggesting something like that. maybe you could add the ability to add it via another command (like `addmulti`) or something to indicate a text box/multiline pattern?

Zarel:
Hm. I'm not a big fan of all the plugin state being shoved into `rooms.ts`, but I do see the value of being able to access it from `/eval` and similar.

Maybe we should allow plugins to export `state`, which is globally accessible somewhere like `Chat.pluginState[pluginName]`?

mia-pi-git:
> Hm. I'm not a big fan of all the plugin state being shoved into `rooms.ts`, but I do see the value of being able to access it from `/eval` and similar.
> 
> Maybe we should allow plugins to export `state`, which is globally accessible somewhere like `Chat.pluginState[pluginName]`?

To me, at least, it makes sense that room-bound repeats are in the room object. Seems uglier to manage them inside the plugin.

AnnikaCodes:
> Maybe we should allow plugins to export state, which is globally accessible somewhere like Chat.pluginState[pluginName]?

This seems like a good idea, I might implement it soon.

> To me, at least, it makes sense that room-bound repeats are in the room object. Seems uglier to manage them inside the plugin, when managing them in the room object makes more sense and allows for easier saving.

Yeah, I thought it was cleaner than having a `Map<roomid, Map<phrase, timeout>>` and passing `room` into all the helper functions (so that I can access room.settings and room.saveSettings).

AnnikaCodes:
> > > I don't think this is something that has to be blocked by a rfaq refactor. There are things constantly used in RoA and NatDex that would prefer to not exist at all as rfaqs and were only rfaqs previously due to bot limitations.
> > 
> > 
> > Oh, I think I misunderstood you. Are you just suggesting having repeats be in text boxes rather than chat lines? That could be a good idea, although using the FAQ system would make it easy to broadcast them when a user asked, etc, rather than only having them be accessible through the repeat.
> 
> Yeah i was suggesting something like that. maybe you could add the ability to add it via another command (like `addmulti`) or something to indicate a text box/multiline pattern?

tbh I'm fine with making them all infoboxes - the only use for direct adding is ``/announce`` which, well, I think a box is pretty disruptive to chat as well.

Zarel:
> > To me, at least, it makes sense that room-bound repeats are in the room object. Seems uglier to manage them inside the plugin, when managing them in the room object makes more sense and allows for easier saving.
> 
> Yeah, I thought it was cleaner than having a `Map<roomid, Map<phrase, timeout>>` and passing `room` into all the helper functions (so that I can access room.settings and room.saveSettings). That being said, hotpatchability might be a worthwhile tradeoff here.

I think persistence belongs to `room.settings`, but the Timer object itself belongs in the plugin. That was how I always envisioned the plugin system to go.

I think if room-bound repeats belong in the room object, it should be a core library in like `server/room-repeats.ts` with the commands in `server/chat-commands/` somewhere. But it doesn't seem like core functionality to me.

AnnikaCodes:
Another benefit of doing it this way is that it's hotpatchable!

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,163 @@
...
+	room.saveSettings();
+	runRepeat(room, repeat);
+}
+
+export function removeRepeat(room: BasicRoom, phrase: string) {
+	if (!room.settings.repeats) return false;
+	room.settings.repeats = room.settings.repeats.filter(repeat => repeat.phrase !== phrase);
+	room.saveSettings();
+
+	if (!repeats.has(room.roomid)) return;
```
KrisXV, line 32: I would probably make the return statements consistent.

KrisXV (commented):
I would add a button at the bottom of the repeats page that parses the removeallrepeats command since we have buttons for individual repeats

Zarel:
Should this wait for #7351?

server/chat-plugins/repeats.ts
AnnikaCodes: Good catch

AnnikaCodes:
> Should this wait for #7351?

I don't think it's any harder to debug than daily spotlights, but I'd also be fine with waiting.

AnnikaCodes:
@Zarel what's the status on this?

sparkychildcharlie:
What happens when you delete a room with a repeat running? Would the setInterval crash since youre running a function of a pointer to ``undefined``?

AnnikaCodes:
I have tested deleting rooms that are using a repeat and it doesn't cause crashes.

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,166 @@
...
+	const oldInterval = repeats.get(room.roomid)?.get(phrase);
+	if (oldInterval) clearInterval(oldInterval);
+	repeats.get(room.roomid)?.set(phrase, null);
+}
+
+function runRepeat(room: BasicRoom, repeat: RepeatedPhrase) {
+	if (!repeats.has(room.roomid)) {
+		repeats.set(room.roomid, new Map<string, NodeJS.Timeout | null>());
```
Zarel, line 40:
```suggestion
		repeats.set(room.roomid, new Map());
```

In theory the type can be inferred, right?

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	const displayRepeat = (targetRoom: BasicRoom, phrase: string) => {
+		targetRoom.add(`|html|<div class="infobox">${phrase}</div>`);
+		targetRoom.update();
+	};
+
+	repeats.get(room.roomid)?.set(
```
Zarel, line 48:
```suggestion
	repeats.get(room.roomid)!.set(
```

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,166 @@
...
+
+function runRepeat(room: BasicRoom, repeat: RepeatedPhrase) {
+	if (!repeats.has(room.roomid)) {
+		repeats.set(room.roomid, new Map<string, NodeJS.Timeout | null>());
+	}
+
+	const displayRepeat = (targetRoom: BasicRoom, phrase: string) => {
+		targetRoom.add(`|html|<div class="infobox">${phrase}</div>`);
+		targetRoom.update();
+	};
```
Zarel, line 46: I don't like binding parameters because it does the same thing as having them in closures, but worse.

I would actually make `Repeats` into an entire class, if I were you. Do you mind if I refactor this?

Zarel: But if I were to refactor only this function, it would look like:

```ts
function runRepeat(room: BasicRoom, repeat: RepeatedPhrase) {
    const roomid = room.roomid;
    let roomRepeats = repeats.get(roomid);
    if (!roomRepeats) {
        roomRepeats = new Map();
        roomRepeats.set(roomid, roomRepeats);
    }

    if (roomRepeats.has(repeat.phrase)) {
        handle this case somehow
    }

	roomRepeats.set(
		repeat.phrase,
		setInterval(() => {
            if (room !== Rooms.get(room.roomid)) {
                handle room destroyed
                return;
            }
            room.add(`|html|<div class="infobox">${repeat.phrase}</div>`);
            room.update();
        }, repeat.interval)
	);
}
```

Zarel:
> I have tested deleting rooms that are using a repeat and it doesn't cause crashes.

Yeah, it doesn't crash, but it does look like a memory leak to me. I think it should be cleaned up.

server/chat-plugins/repeats.ts
AnnikaCodes: If you want to refactor this you can; if not I can remove the bound parameters thing and fix the memory leak.

AnnikaCodes: @Zarel bump?

Zarel:
@AnnikaCodes sorry I was going to refactor this but it fell off my radar.

Zarel:
@AnnikaCodes how's this?

server/rooms.ts
```diff
@@ -113,7 +113,6 @@ export interface RoomSettings {
 	showEnabled?: GroupSymbol | true;
 	permissions?: {[k: string]: GroupSymbol};
 	repeats?: RepeatedPhrase[];
-	quotes?: {userid: string, quote: string, date: number}[];
```
AnnikaCodes: This looks like an artifact from an unrelated PR?

AnnikaCodes:
I don't see any issues with this.

server/chat-plugins/repeats.ts
```diff
@@ -0,0 +1,198 @@
...
+	Repeats.destroy();
+}
+
+export const pages: PageTable = {
+	repeats(args, user) {
+		const room = this.requireRoom();
+		this.title = `Repeated phrases in ${room.roomid}`;
+		this.checkCan("mute", null, room);
+		if (!room.settings.repeats) return `<h1>${this.tr`There are no repeated phrases in ${room.roomid}.`}</h1>`;
+		let html = `<div class="ladder pad"><h1>${this.tr`Repeated phrases in ${room.roomid}`}</h1>`;
```
KrisXV, line 103:
```suggestion
		if (!room.settings.repeats) throw new Chat.ErrorMessage(this.tr`There are no repeated phrases in ${room.roomid}.`);
		let html = `<div class="ladder pad"><h2>${this.tr`Repeated phrases in ${room.roomid}`}</h2>`;
```

server/rooms.ts
Zarel: You're commenting on the wrong diff (it even says "Outdated"). That's in fact an artifact from an unrelated PR brought in by merging master, but it was merged correctly.

server/chat-plugins/repeats.ts
Zarel: idk if "no repeated phrases" needs to be an error message.

What do h1 vs h2 look like? Is h2 necessarily nicer?

server/rooms.ts
AnnikaCodes: Ah, my bad.

server/chat-plugins/repeats.ts
Zarel, line 103:
```suggestion
		let html = `<div class="ladder pad"><h2>${this.tr`Repeated phrases in ${room.roomid}`}</h2>`;
```

I'll take the h2, though

AnnikaCodes:
I rebased this; @Zarel how does it look?

Zarel:
Ah, I think I forgot about this. Thanks!

PR #7341, opened 2020-09-09
Don't ban autoclose tags without spaces from html

TheMezStrikes:
Not the most elegant solution, but if it works it works.

Fixes tags like `<br/>` being banned because the tag name end gets detected via spaces, leaving tagName to be `br/`, which isn't a valid tag name

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1243,8 +1243,8 @@ export class CommandContext extends MessageContext {
 			const stack = [];
 			for (const tag of tags) {
 				const isClosingTag = tag.charAt(1) === '/';
-				const isAutocloseTag = tag.charAt(tag.length - 1) === '/';
-				const tagContent = tag.slice(isClosingTag ? 2 : 1, isAutocloseTag ? tag.length - 1 : undefined).replace(/\s+/, ' ').trim();
+				const isAutoclose = tag.charAt(tag.length - 1) === '/';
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -1243,7 +1243,8 @@ export class CommandContext extends MessageContext {
 			const stack = [];
 			for (const tag of tags) {
 				const isClosingTag = tag.charAt(1) === '/';
-				const tagContent = tag.slice(isClosingTag ? 2 : 1).replace(/\s+/, ' ').trim();
+				const isAutoclose = tag.charAt(tag.length - 1) === '/';
+				const tagContent = tag.slice(isClosingTag ? 2 : 1, isAutoclose ? tag.length - 1 : undefined).replace(/\s+/, ' ').trim();
```
Zarel, line 1247:
```suggestion
				const contentEndLoc = tag.charAt(tag.length - 1) === '/' ? -1 : undefined;
				const tagContent = tag.slice(isClosingTag ? 2 : 1, contentEndLoc).replace(/\s+/, ' ').trim();
```

I think this is slightly better for readability.

Zarel:
General reminder that the PS codebase itself prefers `<br />` over any of the alternatives.

Zarel:
Looks good, thanks!

PR #7342, opened 2020-09-09
Let admins addhtml by default

TheMezStrikes:
I figured this is probably supposed to exist, feel free to reject PR if intentional


Zarel:
It's intentionally console-only for now.

Zarel:
Wait, I'm thinking of `/addline` which is a separate permission.

I think we did eventually decide to allow this? Let me check.

Zarel:
Looks like we never added it. Yeah, that's an oversight, it should have been given to admins at the same time as ROs.

PR #7344, opened 2020-09-09
Support promoting multiple users at once in rooms

AnnikaCodes:
Suggested [here](https://www.smogon.com/forums/threads/allow-multiple-arguments-for-promotions-and-demotions.3670212/#post-8592616).

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -124,16 +124,16 @@ export const commands: ChatCommands = {
 			}
 			if (!user.can('makeroom')) {
 				if (currentGroup.id && !user.can(`room${currentGroup.id || 'voice'}` as 'roomvoice', null, room)) {
-					this.errorReply(`/${cmd} - Access denied for promoting/demoting from ${currentGroupName}.`);
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting ${name} from ${currentGroupName}.`);
 					continue;
 				}
 				if (nextSymbol !== ' ' && !user.can(`room${nextGroup.id || 'voice'}` as 'roomvoice', null, room)) {
-					this.errorReply(`/${cmd} - Access denied for promoting/demoting to ${nextGroupName}.`);
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting ${name} to ${nextGroupName}.`);
 					continue;
 				}
 			}
 			if (targetUser?.locked && room.persist && room.settings.isPrivate !== true && nextGroup.rank > 2) {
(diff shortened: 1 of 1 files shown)


Zarel:
This function has definitely gotten pretty messy. I wonder if there's a way to refactor.

server/chat-commands/moderation.ts
```diff
@@ -110,76 +92,103 @@ export const commands: ChatCommands = {
...
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting from ${currentGroupName}.`);
+					continue;
+				}
+				if (nextSymbol !== ' ' && !user.can(`room${nextGroup.id || 'voice'}` as 'roomvoice', null, room)) {
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting to ${nextGroupName}.`);
+					continue;
+				}
+			}
+			if (targetUser?.locked && room.persist && room.settings.isPrivate !== true && nextGroup.rank > 2) {
+				this.errorReply("Locked users can't be promoted.");
```
Zarel, line 136: These three error messages (access denied promoting to, access denied promoting from, and locked users) happen every loop and so should mention the username.

server/chat-commands/moderation.ts
```diff
@@ -110,76 +92,103 @@ export const commands: ChatCommands = {
...
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting ${name} from ${currentGroupName}.`);
+					continue;
+				}
+				if (nextSymbol !== ' ' && !user.can(`room${nextGroup.id || 'voice'}` as 'roomvoice', null, room)) {
+					this.errorReply(`/${cmd} - Access denied for promoting/demoting ${name} to ${nextGroupName}.`);
+					continue;
+				}
+			}
+			if (targetUser?.locked && room.persist && room.settings.isPrivate !== true && nextGroup.rank > 2) {
+				this.errorReply(`${name} is locked and therefore can't be promoted.`);
```
Zarel, line 136:
```suggestion
				this.errorReply(`${name} is locked and can't be promoted.`);
```

Zarel:
It looks like most of this command is I/O which isn't easy to refactor. `Chat.ErrorMessage` does make it possible to at least refactor permission checking somewhere nicer, though.

AnnikaCodes:
> It looks like most of this command is I/O which isn't easy to refactor. `Chat.ErrorMessage` does make it possible to at least refactor permission checking somewhere nicer, though.

yeah I'll do that when I'm not in class

Zarel:
Looks good, thanks!

PR #7348, opened 2020-09-10
Fix gen 8 fossil descriptions

TheMezStrikes:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-17#post-8525605
I don't own the games but if every source I could find says they're like this then they're probably like this


Zarel:
That's valid; these are probably typoes from earlier.

PR #7350, opened 2020-09-10
Add a library for running SQLite databases in threads

mia-pi-git:
Per @monsanto in https://github.com/smogon/pokemon-showdown/pull/7277#issuecomment-689238778.

> I'll pass on writing the code, instead I'll give some notes on what I imagine will be the interesting parts:
> 
> * SQL stuff in a separate file, should not have dependencies other than `better-sqlite3` (and maybe `sql-template-strings`)
> * Opening a database takes a path to a sqlite database and a path to a extension file.
> * The extension file exports an object with name -> function. For modlogs, this would include `regex`
> * A [new Worker](https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options) is created with the SQL stuff filename, passing the sqlite db path & extension file as `options.workerData`
> * SQL worker opens database, `require`s extension file, adds all of the functions. Could just go ahead and `PRAGMA foreign_keys=on` here as I believe every database will want this on. (Can skip `PRAGMA journal_mode=WAL`, this is a persistent database-level pragma and should be part of the schema)
> * Preparing a statement [sends the string to the worker](https://nodejs.org/api/worker_threads.html#worker_threads_worker_postmessage_value_transferlist), the worker increments an integer, maps it to the prepared statement, and messages the integer to `parentPort`. That integer is given a type like `Statement<Params, RowType>` (more expressive than what better-sqlite3 does, I think)
> * Querying follows a similar pattern, we send the statement integer, the worker maps the integer to the prepared statement, and messages back the rows.
> 
> I imagine the rest will be straightforward but just ping me if not.
> 
> It is probably easier if this is added as part of this PR (which I intend to be the start of SQLite usage in PS) as the design may need to evolve in response to feedback on the modlog. If mia is writing it maybe push access could be given to this branch. Alternatively a separate PR could be created and this PR could every now and then merge it in.


Zarel:
My concerns are:

- pre has complained to me in the past that `Worker` support in V8 is really experimental and has a lot of weird difficult-to-debug bugs
- `better-sqlite3` specifically uses blocking instead of async methods because they claim it has much better performance, and I see zero reason to disbelieve that. Using a `Worker` nullifies those performance gains – why not simply use an async sqlite library if you disagree with their performance claims, rather than adding a further layer of abstraction simply to regain the asyncness?

monsanto:
@Zarel how long ago? Do you have any specifics? 

Will respond to the second point later

monsanto:
- The primary reason to use better-sqlite is user-defined functions in JS. We can generally work around their absence with changes to the schema or C extensions. The latter requires that we somehow build them in `npm install`. An example desirable extension is [sqlite3-pcre](https://github.com/ralight/sqlite3-pcre).
- If there are concerns over heisenbugs then perhaps the smart choice for now is to specify a minimal set of functionality that can be achieved via either package, and go with node-sqlite for now.
- In node-sqlite it is [not safe to share `Database` objects in the presence of transactions](https://github.com/mapbox/node-sqlite3/issues/304). This library would need to handle pooling of these objects, perhaps using [generic-pool](https://www.npmjs.com/package/generic-pool).
- There is some recent talk about unifying node-sqlite and better-sqlite in [this thread](https://github.com/mapbox/node-sqlite3/issues/1292#issuecomment-692280258). Worth noting that the author of better-sqlite doesn't think there's much point to async now that there is workers. Not sure if I 100% agree with that given PS' unique architectural constraints, but it probably merits benchmarking.

Zarel:
Hm. I think that's fair.

Why threads and not `ProcessManager`, though? `ProcessManager` was designed for this...

monsanto:
In the microbenchmark I wrote it is more than twice as fast to send/receive messages to a thread vs a process. This is unsurprising because communication with another process additionally requires socket syscalls.

Zarel:
Annoying.

Maybe ProcessManager should have "spawn as thread" and "spawn as process" as possible options?

lib/sql.ts
```diff
@@ -0,0 +1,147 @@
...
+			const {data} = message;
+			database.exec(data);
+		}
+			break;
+		}
+		parentPort!.postMessage(results);
+	});
+}
+
+export class SQLDatabase extends Worker {
```
Zarel, line 83: We've had a lot of trouble extending builtins. I think wrapping the builtin makes much more sense.

Zarel:
Node.js's documentation describes `worker_threads` as "Stability: 1 - Experimental" which worries me in conjunction with pre's reports of getting lots of bugs he never experienced with child processes.

I think refactoring ProcessManager to work with it might be a significantly better approach, especially because ProcessManager would make it easy to switch between "subprocess", "thread", and "sync in-thread" if we _did_ encounter bugs we couldn't work around.

monsanto:
You're probably reading outdated documentation, worker threads have been stable since 12.x. I don't know enough about ProcessManager to comment on the suggested approach, should be fine as long as this implementation detail doesn't leak into the library interface.

Zarel:
You're right. I'm surprised devdocs is suddenly out of date. :/

lib/sql.ts
```diff
@@ -0,0 +1,147 @@
...
+		const result = await resultPromise;
+		return result;
+	}
+}
+/**
+ * @param options Either an object of filename, extension, or just the string filename
+ */
+export function SQL(options: SQLOptions | string) {
+	if (typeof options === 'string') options = {file: options};
+	return new SQLDatabase(options);
```
Kaiepi, line 146: A new worker for each database...? Wouldn't something like a thread pool be more optimal here?

lib/sql.ts
```diff
@@ -0,0 +1,147 @@
...
+		this.postMessage(args);
+		const resultPromise = new Promise<any>((resolve, reject) => {
+			this.once('message', result => {
+				resolve(result);
+			});
+			this.once('error', err => {
+				reject(err);
+			});
+		});
+		const result = await resultPromise;
```
Kaiepi, line 137: I don't see the point of awaiting this when the `makeRequest` is already async.

Zarel:
I've opened #7398.

@mia-pi-git do you think you can transition over to using ProcessManager for this?

mia-pi-git:
> I've opened #7398.
> 
> @mia-pi-git do you think you can transition over to using ProcessManager for this?

Will do.

server/rooms.ts
```diff
@@ -743,6 +738,14 @@ export abstract class BasicRoom {
 				}
 			}
 			this.settings.aliases = undefined;
+
+			if (this.battle) {
```
monsanto, line 742: This changeset has a bunch of unrelated changes in it, I can't review as-is

mia-pi-git: The important ones should be just lib/sql.ts and test/lib/sql.js. I'll see if i can properly rebase this, though there will be some changes i can't edit from Zarel's processwrapper branch.

mia-pi-git: Okay, removed the irrelevant ones not from Zarel's branch.

lib/sql.ts
```diff
@@ -0,0 +1,164 @@
...
+	const database = new Sqlite(file!);
+	if (extension) {
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
+		const functions: {[k: string]: (...params: any) => any} = require(`../${extension}`);
+		for (const key in functions) {
+			database.function(key, functions[key]);
+		}
+	}
+	database.pragma(`foreign_keys=on`);
+	parentPort!.on('message', (message: string) => {
```
Zarel, line 47: This does not appear to be actually using `ProcessManager`. Like I said, the value of using PM is that you can easily switch between subthreads, subprocesses, and synchronous execution. Hardcoding thread code defeats that goal.

You don't actually need to base on top of my PM code at all, you can just implement this as a PM subprocess, and then my PR will add thread support.

mia-pi-git: Apologies! Misunderstood. Is this more what you meant, now?

monsanto:
I need to brainstorm a bit and do some benchmarks before I can give further review on this PR. I probably cannot get to this tomorrow (Tuesday) but I'll try to find some time on Wed.

monsanto:
If we stay with better-sqlite3:

- As-is ProcessManager blocks some of the optimizations I want to play with, such as schema-based (de-)serialization, which is a natural fit for SQL queries. [slow-json-stringify](https://github.com/lucagez/slow-json-stringify/blob/master/benchmark.md) (name for fun) and [this benchmark](https://github.com/Adelost/javascript-serialization-benchmark) are on my mind. The culprit is unconditionally `JSON.stringify`ing the input, which requires a pass over the entire string to escape it.

- The fastest way of executing a transaction is to have the transaction code be `require`d on the worker side, and work synchronously, similar to how UDFs work. This would require having a synchronous interface as well.

mia-pi-git:
> * As-is ProcessManager blocks some of the optimizations I want to play with, such as schema-based (de-)serialization, which is a natural fit for SQL queries. [fast-json-stringify](https://github.com/fastify/fast-json-stringify), [slow-json-stringify](https://github.com/lucagez/slow-json-stringify/blob/master/benchmark.md) (name for fun) and [this benchmark](https://github.com/Adelost/javascript-serialization-benchmark) are on my mind. The culprit is unconditionally `JSON.stringify`ing the input, which requires a pass over the entire string to escape it.
> * The fastest way of executing a transaction is to have the transaction code be `require`d on the worker side, and work synchronously, similar to how UDFs work. This would require having a synchronous interface as well.

If JSON stringifying is an issue, i can always convert it from using `taskid` and `json` to just putting the resolver into an array rather than a map, so the request can be sent directly. Which, though i went with this approach for extra safety, _should_ also be fine?

lib/sql.ts
```diff
@@ -0,0 +1,208 @@
...
+
+interface SQLOptions {
+	file: string;
+	/** file to import database functions from - this should be relative to this filename. */
+	extension?: string;
+}
+
+type DataType = unknown[] | Record<string, unknown>;
+
+export type DatabaseQuery = {
```
monsanto, line 18: This seems like an implementation detail to me, do we really need to export it?

mia-pi-git: Probably not, but I can't see it hurting.

lib/sql.ts
```diff
@@ -0,0 +1,208 @@
...
+	async prepare(statement: string) {
+		const cachedStatement = this.statements.get(statement);
+		if (cachedStatement) {
+			return cachedStatement;
+		}
+		const int = await this.query({type: 'prepare', data: statement});
+		if (typeof int === 'number') this.statements.set(statement, int);
+		return int;
+	}
+	all(statement: string | number, data: DataType = {}) {
```
monsanto, line 93: Can you make some sort of `Statement` type instead of having `string | number` everywhere?

lib/sql.ts
```diff
@@ -0,0 +1,208 @@
...
+		case 'exec': {
+			const {data} = query;
+			database.exec(data);
+		}
+			break;
+		case 'transaction': {
+			const {num, data} = query;
+			const transaction = transactions.get(num);
+			if (!transaction) break;
+			results = transaction(data);
```
monsanto, line 221: I don't understand how this works -- presumably the transaction function needs to be passed some sort of connection object it can use? Are you envisioning the transaction function using the methods on `database`, and if so, how would you determine which transaction these methods are associated with?

Needs tests too

mia-pi-git: I overlooked that. I'll probably pass it the connection parameter, yes.

mia-pi-git: Done.

mia-pi-git:
@monsanto bump

lib/sql.ts
monsanto: I don't like it when implementation details are exported; if there's anything that would change if you switched from `better-sqlite3` to `sqlite`, then don't let any consumers use it. It's a code smell sort of thing, your API is messed up if anyone needs to know what this type is

mia-pi-git: Alright.

monsanto:
I only took the most cursory of looks this time around. Please see my comment about exporting implementation details, but otherwise this is fine to merge when you're ready

mia-pi-git:
(checked with chaos, was told it was fine to merge)

PR #7351, opened 2020-09-10
Chat plugins: Support exporting plugin state

AnnikaCodes:
This will make it easier to debug chat plugins. I also added state to some of the plugins that I felt could use it.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -319,7 +319,7 @@ export const commands: ChatCommands = {
 				if (requiresForce(patch)) return this.errorReply(requiresForceMessage);
 
 				global.Tournaments = require('../tournaments').Tournaments;
-				Chat.loadPluginData(Tournaments);
+				Chat.loadPluginData(Tournaments, 'tournaments');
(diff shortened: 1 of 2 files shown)


Zarel:
Looks great, thanks!

Zarel:
I wonder if all the state was simply at the top level?

So instead of `export const state = {foo, bar};` you would `export foo` and `export bar`.

And then `Chat.plugins[pluginName]` would be everything the plugin exported except `commands` and `pages`.

AnnikaCodes:
> I wonder if all the state was simply at the top level?
> 
> So instead of `export const state = {foo, bar};` you would `export foo` and `export bar`.
> 
> And then `Chat.plugins[pluginName]` would be everything the plugin exported except `commands` and `pages`. Maybe even including `commands` and `pages`.

Yeah, that seems like a better idea. Is there a reason to exclude `commands` and `pages`?

Zarel:
> Is there a reason to exclude `commands` and `pages`?

You might reclaim some RAM, I guess? Probably worth keeping them.

Zarel:
Looks good, thanks!

PR #7352, opened 2020-09-11
Support adding quotes in rooms

mia-pi-git:
This is a thing a lot of people liked with Kid A, and since it's gone i figured it would be nice to migrate and not too much work.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2536,11 +2536,13 @@ export const commands: ChatCommands = {
 		this.runBroadcast();
 		const {quote, date, userid} = room.settings.quotes[Math.floor(Math.random() * room.settings.quotes.length)];
 		const formatted = quote.split('\n').map(item => Chat.formatText(item)).join('<br />');
+		const attribute = toID(target) === 'showauthor';
 		return this.sendReplyBox(
 			`${formatted}<br />` +
-			`<hr /><small>Added by ${userid} on ${Chat.toTimestamp(new Date(date), {human: true})}</small>`
+			attribute ? `<hr /><small>Added by ${userid} on ${Chat.toTimestamp(new Date(date), {human: true})}</small>` : ''
 		);
 	},
+	randquotehelp: [`/randquote [showauthor] - Show a random quote from the room. Add 'showauthor' to see who added it and when.`],
 
 	addquote: 'quote',
 	quote(target, room, user) {
@@ -2551,7 +2553,7 @@ export const commands: ChatCommands = {
 		if (this.filter(target) !== target) {
 			return this.errorReply(`Invalid quote.`);
 		}
-		if (Object.values(room.settings.quotes).filter(item => item.quote === target).length) {
+		if (room.settings.quotes.filter(item => item.quote === target).length) {
 			return this.errorReply(`"${target}" is already quoted in this room.`);
 		}
 		if (target.length > 300) {
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2530,6 +2530,63 @@ export const commands: ChatCommands = {
 	},
 	denyshowhelp: [`/denyshow [user] - Denies the media display request of [user]. Requires: % @ # &`],
 
+	randquote(target, room, user) {
+		room = this.requireRoom();
+		if (!room.settings.quotes) return this.errorReply(`This room has no quotes.`);
+		this.runBroadcast();
+		const [quote] = Utils.shuffle(room.settings.quotes);
+		const formatted = quote.split('\n').map(item => Chat.formatText(item)).join('<br />');
+		return this.sendReplyBox(formatted);
+	},
+
+	addquote: 'quote',
+	quote(target, room, user) {
+		room = this.requireRoom();
+		target = target.trim();
+		this.checkCan('mute', null, room);
+		if (!room.settings.quotes) room.settings.quotes = [];
+		if (this.filter(target) !== target) {
+			return this.errorReply(`Invalid quote.`);
+		}
+		if (room.settings.quotes.includes(target)) {
+			return this.errorReply(`You already have a quote in this room that matches "${target}".`);
+		}
+		if (room.settings.quotes.length >= 50) {
+			return this.errorReply(`You already have 50 quotes, which is the maximum.`);
+		}
+		room.settings.quotes.push(target);
+		room.saveSettings();
+		this.privateModAction(`${user.name} added the quote "${target}"`);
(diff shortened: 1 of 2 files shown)


server/chat-commands/info.ts
```diff
@@ -2703,11 +2760,34 @@ export const pages: PageTable = {
...
+		const room = this.requireRoom();
+		// allow it for users if they can access the room
+		if (!user.inRooms.has(room.roomid) && room.settings.isPrivate && !user.isStaff) {
+			return this.errorReply(`Access denied.`);
+		}
+		if (!room.settings.quotes) {
+			return `<div class="pad"><h2>This room has no quotes.</h2></div>`;
+		}
+
+		let buffer = `<div class="pad"><h2>Quotes on ${room.title}: (${room.settings.quotes.length})</h2>`;
```
KrisXV, line 2773:
```suggestion
		let buffer = `<div class="pad">`;
		if (!room.settings.quotes?.length) {
			return `${buffer}<h2>This room has no quotes.</h2></div>`;
		}

		let buffer = `<h2>Quotes for ${room.title}: (${room.settings.quotes.length})</h2>`;
```

AnnikaCodes (commented):
I sent you a review on Discord as requested.

Zarel:
Hm, I'm starting to worry if we're stuffing too much stuff in `room.settings`. I suppose at least we might want to consider having one file per room rather than one `chatrooms.json` for all the rooms.

That'll be dealt with later, though; I believe this is ready to merge. @AnnikaCodes should I be concerned about your private review?

Zarel (approved):

server/chat-commands/info.ts
```diff
@@ -2530,6 +2530,63 @@ export const commands: ChatCommands = {
 	},
 	denyshowhelp: [`/denyshow [user] - Denies the media display request of [user]. Requires: % @ # &`],
 
+	randquote(target, room, user) {
+		room = this.requireRoom();
+		if (!room.settings.quotes?.length) return this.errorReply(`This room has no quotes.`);
+		this.runBroadcast();
+		const [quote] = Utils.shuffle(room.settings.quotes);
```
Zarel, line 2537: Wait, this is wrong. `Utils.shuffle` is a mutating shuffle which should not be applied to the quotes list. If there's no `Utils.random` you should be doing:

```suggestion
		const quote = room.settings.quotes[Math.floor(Math.random() * room.settings.quotes.length)];
```

server/chat-commands/info.ts
```diff
@@ -2530,6 +2530,63 @@ export const commands: ChatCommands = {
...
+		if (this.filter(target) !== target) {
+			return this.errorReply(`Invalid quote.`);
+		}
+		if (room.settings.quotes.includes(target)) {
+			return this.errorReply(`"${target}" is already quoted in this room.`);
+		}
+		if (room.settings.quotes.length >= 50) {
+			return this.errorReply(`This room already has 50 quotes, which is the maximum.`);
+		}
+		room.settings.quotes.push(target);
```
Zarel, line 2557: This should probably at least track the user who added it, and the date it was added.

AnnikaCodes:
> Hm, I'm starting to worry if we're stuffing too much stuff in `room.settings`. I suppose at least we might want to consider having one file per room rather than one `chatrooms.json` for all the rooms.
> 
> That'll be dealt with later, though; I believe this is ready to merge. @AnnikaCodes should I be concerned about your private review?

can you wait to merge this until I approve (or I can merge it then if you're ok with that)? I think it's probably good to go but I want to make sure that nothing from the private review slipped by and I need to go do stuff IRL for a bit.

server/chat-commands/info.ts
```diff
@@ -2530,6 +2530,63 @@ export const commands: ChatCommands = {
...
+		if (!room.settings.quotes) room.settings.quotes = [];
+		if (this.filter(target) !== target) {
+			return this.errorReply(`Invalid quote.`);
+		}
+		if (room.settings.quotes.includes(target)) {
+			return this.errorReply(`"${target}" is already quoted in this room.`);
+		}
+		if (room.settings.quotes.length >= 50) {
+			return this.errorReply(`This room already has 50 quotes, which is the maximum.`);
+		}
```
Zarel, line 2556: Pretty low maximum... maybe this should be stored in SQLite?

Zarel:
Nah, don't merge without me. I'll wait until I get an approval review from you before merging, though.

KrisXV:
> Hm, I'm starting to worry if we're stuffing too much stuff in `room.settings`. I suppose at least we might want to consider having one file per room rather than one `chatrooms.json` for all the rooms.
> 
> That'll be dealt with later, though; I believe this is ready to merge. @AnnikaCodes should I be concerned about your private review?

Though not entirely related to the PR at hand, I'm unsure why we are using `room.settings` to store things that are not settings. I think we should either find a new place to store quotes/repeats/etc. or rename `settings` to something that makes more sense.

AnnikaCodes:
Maybe have `room.settings` continue storing settings, and add a `room.data` attribute that stores data (with its own saving method, which could save to a separate file for each chatroom if desired). Probably we would also only support data for non-personal rooms.

AnnikaCodes (approved):

server/chat-commands/info.ts
```diff
@@ -2530,6 +2530,72 @@ export const commands: ChatCommands = {
...
 
+	randquote(target, room, user) {
+		room = this.requireRoom();
+		if (!room.settings.quotes?.length) return this.errorReply(`This room has no quotes.`);
+		this.runBroadcast();
+		const {quote, date, userid} = room.settings.quotes[Math.floor(Math.random() * room.settings.quotes.length)];
+		const formatted = quote.split('\n').map(item => Chat.formatText(item)).join('<br />');
+		return this.sendReplyBox(
+			`${formatted}<br />` +
+			`<hr /><small>Added by ${userid} on ${Chat.toTimestamp(new Date(date), {human: true})}</small>`
```
Zarel, line 2541: Do most quote databases show this data?

mia-pi-git: No, but i though that's what you meant by attribution? 
Since if you're not showing it here, there's little point, considering modlog exists.

Zarel: I don't think it needs to show author and date by default.

mia-pi-git: Okay.

mia-pi-git: Updated. It will now only show who it's by / when if you do !randquote showauthor.

Zarel: Hm... You could make an argument for it either way, I guess. :s

PR #7353, opened 2020-09-11
Thing of the Day: Fix bug in removing users

mia-pi-git:
`user.previousIDs` did not contain the user's _current_ ID, so removing nominations from them wouldn't work, as it checks the property that `user.previousIDs.slice()` was assigned to. This `concat`s the current `id` to the `userid` array so it's properly tracked.


Zarel:
Thanks!

PR #7354, opened 2020-09-11
Update attributions in various locations

mia-pi-git:


server/chat-plugins/help.ts
```diff
@@ -3,7 +3,8 @@
  * Uses Regex to match room frequently asked question (RFAQ) entries,
  * and replies if a match is found.
  * Supports configuration.
- * Written by mia-pi.
+ * Written by Mia.
+ * @author mia-pi-git
```
Zarel, line 7: There's no need to use your Git username for `@author`, I think you can just use whatever you want.

Zarel: https://jsdoc.app/tags-author.html

mia-pi-git: Yes, i'm just picky about how I have it.

Zarel:
Looks like we forgot about this; thanks for the bump.

PR #7355, opened 2020-09-11
Youtube: Support auto-recaching data

mia-pi-git:
I'm splitting this off from #7006, since this can happen before SQLite - that cannot.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,3 +9,4 @@ server/chat-plugins/othermetas.ts @KrisXV @TheImmortal
 server/chat-plugins/random-battles.ts @KrisXV @TheImmortal
 server/chat-plugins/scavenger*.ts @xfix @sparkychildcharlie
 server/chat-plugins/trivia.ts @AnnikaCodes
+server/chat-plugins/youtube.ts @mia-pi-git
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -12,7 +12,17 @@ import {Utils} from '../../lib/utils';
 const ROOT = 'https://www.googleapis.com/youtube/v3/';
 const STORAGE_PATH = 'config/chat-plugins/youtube.json';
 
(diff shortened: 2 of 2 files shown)


Zarel:
Why would you want it to auto-recache? Like I've said before, I think it makes more sense to just store a last update time, and ignore the cache if it's too old.

Zarel:
Otherwise it seems like you're making extra HTTP requests from Main for no reason?

sparkychildcharlie:
I have no experience with working with youtube APIs, but would auto updating all the data cause youtube to block PS's requests to it to prevent "DDoS"?  Once the list of channels expands into the hundreds, Youtube might not like being sent 100+ requests in under 100ms.  Last time I wrote a script to grab data from multiple pages at a time from serebii, I got blocked until I started pulling the data (some 200+ pages) one at a time at an interval of 1 second.

mia-pi-git:
> Why would you want it to auto-recache? Like I've said before, I think it makes more sense to just store a last update time, and ignore the cache if it's too old.

Yes, but historical data :( 
I suppose I could track a update time for each chunk of data in the array and have it make a new request if that time is too old, and attach the data to the front of the array. From there, it'd probably be better to see a time for each data point in a trend.

mia-pi-git:
On reflection, though, it would be an issue to do that cleanly in SQLite. I'm not really sure what the best solution for this is, so I guess I'll test a few different options and push the most efficient  one.

Zarel:
Depending on how often this is used, I don't think it's worth spamming YouTube with requests from Main. I think if you want to do this, it should go in a bot.

mia-pi-git:
> Depending on how often this is used, I don't think it's worth spamming YouTube with requests from Main. I think if you want to do this, it should go in a bot.

Alright, I can rework historical data to accommodate just using a last cached time.

mia-pi-git:
Nuking this as well - even with the increased queries this would suck a lot.

PR #7356, opened 2020-09-11
Fix Auth#isStaff

iscke:


AnnikaCodes (approved):

server/user-groups.ts
```diff
@@ -56,7 +56,9 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 		return super.get(user) || Auth.defaultSymbol();
 	}
 	isStaff(userid: ID) {
-		return this.has(userid) && this.get(userid) !== '+';
+		if (!this.has(userid)) return false;
+		const auth = this.get(userid);
+		return auth !== '+' && auth !== Auth.defaultSymbol();
```
Zarel, line 61: Oh, we need to hardcode Prize Winner now.

```suggestion
		return auth !== '+' && auth !== '^' && auth !== Auth.defaultSymbol();
```

Maybe the individual config entries should mark `isStaff`?

whalemer:
how's this?

Zarel:
I wonder if it should be an actual permission. We also use `lock` for a lot of permissions that aren't actually related to locks...

Zarel:
Alternatively, just do the `^` hardcode, for now?

whalemer:
having it as a proper permission feels like a misname. what does it give you that you can't already do through Auth#isStaff and setting it manually in config?

(i'm pretty sure this needs a restart to take effect, so i'm not too worried about pushing it through particularly quickly)

server/user-groups.ts
```diff
@@ -89,9 +90,9 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 			name: symbol,
 		});
 	}
-	getEffectiveSymbol(user: User): EffectiveGroupSymbol {
+	getEffectiveSymbol(user: ID | User): EffectiveGroupSymbol {
 		const group = this.get(user);
-		if (this.has(user.id) && group === Auth.defaultSymbol()) {
+		if (this.has(typeof user !== 'string' ? (user as User).id : user) && group === Auth.defaultSymbol()) {
```
AnnikaCodes, line 95:
```suggestion
		if (this.has(toID(user)) && group === Auth.defaultSymbol()) {
```

AnnikaCodes:
I think having it in the same place as `roomonly` makes sense - it's metadata about the group itself, rather than a permission in and of itself.

Zarel:
I think the first question is "what is it used for"? Like `lock`, I think there's an argument that a lot of those things should be individual permissions.

Zarel:
If we have time, let's create a list. My recent refactor migrated IP-viewing from `ban` to `ip`, which was a good start.

Things left:

- Viewing private mod actions
- Autojoining Staff?

Can one of you make a list?

Zarel:
- Viewing private mod actions
- Autojoining Staff
- Automatically grant "trusted"
- use STAFF_THROTTLE rather than regular throttle
- A quotes permission hardcode which seems unnecessary?
- Allow certain monitor bypasses
- Allow use of the helpticket system

That seems to be it for `user.isStaff`.

Zarel:
`auth.isStaff` is used for:

- `STAFF_THROTTLE` again
- trusted user checking
- chatfilter bypassing
- a Mafia chatfilter bypass

Zarel:
`'lock'` is used for:

- `/lock`, `/unlock` (obviously)
- people locked users can PM (and a lot of similar exceptions, basically only users who can lock are allowed to interact with locked users)
- permission to use scripted buttons
- permission to use `` tags
- not allowed to block PMs
- bypassing `hideBattlesFromTrainerCard`
- seeing help for private commands
- `/sharedbattles`
- `/showglobalpunishments`
- `/globalpunishments`
- seeing hidden roomauth in `/userauth`
- punishment notifications (to prevent spam, if someone in a private room roombans/warns someone outside of the private room, it will only notify if the action is done by a global)
- `/unbattleban`
- `/renamegroupchat`
- bypassing staff impersonation filter for status message
- `<<view-filters>>`
- `/chatlog`
- various helpticket permissions
- viewing IPs in modlog

whalemer:
can('mute') is used for:
- bypassing banwords
- seeing staffintros
- seeing punishments in `/whois`
- /a lot/ of generic staff-only commands (approvals, show, subroomgroupchats, showing secret subrooms, chatlogs, a lot of room activity commands)
- /a lot/ of generic moderation commands (hidetext, showblacklist, banwords list)
- seeing whitelisted users in auth
- helpfilter
- getting helpticket notifications (looks like a mistake, honestly)

AnnikaCodes:
I'm not sure that this is necessary anymore - our current method of determining of `isStaff` hasn't been subject to any complaints lately.

Zarel:
Our permissions system is still very inconsistent.

mia-pi-git:
This bug was fixed some time ago and as such, this PR can be closed.

monsanto:
Mia says its closed, so I'm closing.

PR #7357, opened 2020-09-12
Add hidestats arg to showteam and improve output

DaWoblefet:
I'm not sure if it'll ever get used as such, but !showteam is very very close to how VGC Player's Cup works - each side has the other's complete team information, except for stats. This does the following:

- Adds a "hidestats" arg which prevents EVs, IVs, and Nature from printing
- Adds the Gmax forme information if applicable
- Doesn't display the @ if the Pokemon is not holding an item

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -954,7 +954,7 @@ export const commands: ChatCommands = {
 				}
 			}
 		}
-		let resultString = Dex.stringifyTeam(teamStrings, false, hideStats);
+		let resultString = Dex.stringifyTeam(teamStrings, undefined, hideStats);
 		if (showAll) {
 			resultString = `<details><summary>${this.tr`View team`}</summary>${resultString}</details>`;
 		}
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -935,7 +935,7 @@ export const commands: ChatCommands = {
 		if (!battle) return this.errorReply(this.tr("This command can only be used in a battle."));
 		let teamStrings = await battle.getTeam(user);
 		if (!teamStrings) return this.errorReply(this.tr("Only players can extract their team."));
-		if (target && !hideStats) {
+		if (!showAll) {
 			const parsed = parseInt(target);
 			if (parsed > 6) return this.errorReply(this.tr`Use a number between 1-6 to view a specific set.`);
 			if (isNaN(parsed)) {
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -928,7 +928,7 @@ export const commands: ChatCommands = {
 	async showset(target, room, user, connection, cmd) {
 		this.checkChat();
 		const showAll = cmd === 'showteam';
-		const hideStats = target === 'hidestats';
+		const hideStats = toID(target) === 'hidestats';
 		room = this.requireRoom();
 		const battle = room.battle;
 		if (!showAll && !target) return this.parse(`/help showset`);
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -935,23 +935,21 @@ export const commands: ChatCommands = {
 		if (!battle) return this.errorReply(this.tr("This command can only be used in a battle."));
 		let teamStrings = await battle.getTeam(user);
 		if (!teamStrings) return this.errorReply(this.tr("Only players can extract their team."));
-		if (target) {
+		if (target && !hideStats) {
 			const parsed = parseInt(target);
 			if (parsed > 6) return this.errorReply(this.tr`Use a number between 1-6 to view a specific set.`);
-			if (!hideStats) {
-				if (isNaN(parsed)) {
-					const matchedSet = teamStrings.filter(set => {
-						const id = toID(target);
-						return toID(set.name) === id || toID(set.species) === id;
-					})[0];
-					if (!matchedSet) return this.errorReply(this.tr`The Pokemon "${target}" is not in your team.`);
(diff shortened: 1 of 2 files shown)


mia-pi-git (commented):
Not on a computer, so I'll do a more thorough review when I can, but I'll leave a few things I see.

server/chat-commands/core.ts
```diff
@@ -928,6 +928,7 @@ export const commands: ChatCommands = {
 	async showset(target, room, user, connection, cmd) {
 		this.checkChat();
 		const showAll = cmd === 'showteam';
+		const hideStats = target === 'hidestats';
```
mia-pi-git, line 931: This should probably check the toID(target) rather than just the target.

mia-pi-git: Also, what if you want to show just one set with stats hidden?

server/chat-commands/core.ts
```diff
@@ -937,21 +938,23 @@ export const commands: ChatCommands = {
 		if (target) {
 			const parsed = parseInt(target);
 			if (parsed > 6) return this.errorReply(this.tr`Use a number between 1-6 to view a specific set.`);
-			if (isNaN(parsed)) {
```
mia-pi-git, line 940: I can't review on the if (target) line, but you should put the !hideStats check there instead, I believe. 
if (target && !hideStats) ...

mia-pi-git: Would simplify some code.

server/chat-commands/core.ts
DaWoblefet: I could make the arguments comma-delimited, I suppose and split on that. I don't think it'd be a common use case though; I'm picturing the use case for /showteam hidestats to be purely team-only for VGC purposes.

mia-pi-git: Fair. I think it'd be nice to have, personally, but I'm not saying it has to be done.

server/chat-commands/core.ts
DaWoblefet: It could run the else statement then though, which I don't think we want, since it would try to display the info an arrayIndex of whatever "hideStats" - 1 is which it probably won't like.

DaWoblefet: Oh never mind I'm dumb, was looking at the wrong line lololol

sim/dex.ts
```diff
@@ -1445,27 +1445,30 @@ export class ModdedDex {
...
-				if (mon.ivs[stat as StatName] !== 31) ivs.push(`${mon.ivs[stat as StatName]} ${stat}`);
-			}
-			if (ivs.length) output += `IVs: ${ivs.join(' / ')}<br/>`;
-			output += mon.moves.map(move => `- ${Dex.getMove(move).name}<br/>`).join('');
-			output += '<br/>';
+			const name = mon.gigantamax ? `${species.name}-Gmax` : species.name;
+			output += nicknames ? `${nicknames?.[i]} (${name})` : name;
+			output += mon.item ? ` @ ${Dex.getItem(mon.item).name}<br />` : '<br />';
+			output += `Ability: ${Dex.getAbility(mon.ability).name}<br />`;
+			if (typeof mon.happiness === 'number' && mon.happiness !== 255) output += `Happiness: ${mon.happiness}<br />`;
```
KrisXV, line 1456:
```suggestion
			output += nicknames ? `${nicknames?.[i]} (${species.name})` : species.name;
			output += mon.item ? ` @ ${Dex.getItem(mon.item).name}<br />` : '<br />';
			output += `Ability: ${Dex.getAbility(mon.ability).name}<br />`;
			if (typeof mon.happiness === 'number' && mon.happiness !== 255) output += `Happiness: ${mon.happiness}<br />`;
			if (mon.gigantamax) output += ``Gigantamax: Yes<br />;
```

mia-pi-git (approved):
Looks good to me.

server/chat-commands/core.ts
```diff
@@ -928,13 +928,14 @@ export const commands: ChatCommands = {
...
 		room = this.requireRoom();
 		const battle = room.battle;
 		if (!showAll && !target) return this.parse(`/help showset`);
 		if (!battle) return this.errorReply(this.tr("This command can only be used in a battle."));
 		let teamStrings = await battle.getTeam(user);
 		if (!teamStrings) return this.errorReply(this.tr("Only players can extract their team."));
-		if (target) {
+		if (target && !hideStats) {
 			const parsed = parseInt(target);
 			if (parsed > 6) return this.errorReply(this.tr`Use a number between 1-6 to view a specific set.`);
```
Zarel, line 940: Maybe mention hidestats here?

server/chat-commands/core.ts
```diff
@@ -928,13 +928,14 @@ export const commands: ChatCommands = {
...
 		const showAll = cmd === 'showteam';
+		const hideStats = toID(target) === 'hidestats';
 		room = this.requireRoom();
 		const battle = room.battle;
 		if (!showAll && !target) return this.parse(`/help showset`);
 		if (!battle) return this.errorReply(this.tr("This command can only be used in a battle."));
 		let teamStrings = await battle.getTeam(user);
 		if (!teamStrings) return this.errorReply(this.tr("Only players can extract their team."));
-		if (target) {
+		if (target && !hideStats) {
```
Zarel, line 938: Actually I think maybe this should test `showAll` rather than `target && !hideStats`.

DaWoblefet:
Did I miss anything else, or is this good to go?

Zarel:
Looks good to me!

PR #7359, opened 2020-09-13
Allow groupchat hosts to rename their groupchats

AnnikaCodes:


Zarel:
Should parent roomdrivers be able to rename a subroom groupchat?

server/chat-commands/room-settings.ts
```diff
@@ -933,7 +933,7 @@ export const commands: ChatCommands = {
 		if (room.persist && isGroupchat) return this.errorReply(`This isn't a groupchat.`);
 		if (!room.persist && !isGroupchat) return this.errorReply(`Use /renamegroupchat instead.`);
 		if (isGroupchat) {
-			if (!user.can('lock')) this.checkCan('editroom', null, room);
+			this.checkCan('mute', null, room);
```
Zarel, line 936: I suppose this _is_ simpler and usually correct, but I can't say I'm a fan of abusing unrelated permissions in the general case.

AnnikaCodes: We could use `Auth#atLeast`, I guess? I think that a `renamegroupchat` permission is a bit specific, personally.

Zarel:
I'm not going to block this on the `mute` trick, but this question _is_ relevant:

> Should parent roomdrivers be able to rename a subroom groupchat?

Because I believe this changes that from "no" to "yes", which was probably unintentional.

AnnikaCodes:
Yes, that was unintentional

Zarel:
Thanks!

PR #7361, opened 2020-09-13
Quotes: Quality-of-life changes and bug fixes

KrisXV:


server/chat-commands/info.ts
```diff
@@ -2770,23 +2775,25 @@ export const pages: PageTable = {
...
 		}
 
-		buffer = `${buffer}<h2>Quotes on ${room.title}: (${room.settings.quotes.length})</h2>`;
+		buffer += `<h2>Quotes for ${room.title} (${room.settings.quotes.length}):</h2>`;
 		for (const entry of room.settings.quotes) {
 			const index = room.settings.quotes.indexOf(entry) + 1;
 			const {quote, userid, date} = entry;
-			buffer += `<div class="infobox">${index}: ${Chat.collapseLineBreaksHTML(Chat.formatText(quote))}`;
-			buffer += `<br /> Added by ${userid} on ${Chat.toTimestamp(new Date(date), {human: true})}`;
+			buffer += `<div class="infobox">${index}: ${Chat.formatText(quote).replace(/\n/g, '<br />')}`;
```
mia-pi-git, line 2793: Why not just collapse the line breaks when the quote is added? The only places where the quote is shown, it would be collapsed anyway, so it'd save some time to collapse once rather then in randquote AND in the page, imo.

server/chat-commands/info.ts
```diff
@@ -2770,23 +2775,25 @@ export const pages: PageTable = {
 	},
 	quotes(args, user) {
 		const room = this.requireRoom();
+		this.title = `[Quotes]`;
```
mia-pi-git, line 2778: Should probably note the room in the title.

server/chat-commands/info.ts
```diff
@@ -2583,17 +2585,20 @@ export const commands: ChatCommands = {
 			return this.errorReply(`Quote not found.`);
 		}
 		const [removed] = targetRoom.settings.quotes.splice(index - 1, 1);
-		this.privateModAction(`${user.name} removed quote ${index}: "${removed.quote}" (originally added by ${removed.userid})`);
-		this.modlog(`REMOVEQUOTE`, null, removed.quote);
-		return targetRoom.saveSettings();
+		this.privateModAction(`${user.name} removed quote indexed at ${index} (originally added by ${removed.userid}).`);
+		this.modlog(`REMOVEQUOTE`, null, Chat.collapseLineBreaksHTML(removed.quote).replace(/&#10;/g, ' '));
```
mia-pi-git, line 2589: See below - escape on adding rather than everywhere else.

server/chat-commands/info.ts
```diff
@@ -2562,8 +2564,8 @@ export const commands: ChatCommands = {
 		}
 		room.settings.quotes.push({userid: user.id, quote: target, date: Date.now()});
 		room.saveSettings();
-		this.privateModAction(`${user.name} added the quote "${target}"`);
-		return this.modlog(`ADDQUOTE`, null, target);
+		this.privateModAction(`${user.name} added a new quote.`);
```
mia-pi-git, line 2567: Posting in case this gets lost in sim-dev before we resolve it; I think this should still show the added quote for ease of others viewing.

server/chat-commands/info.ts
KrisXV: This change makes it so it no longer collapses on the page because the way it currently displays multi-line quotes is very cluttered, and even if i collapsed while adding, I would end up just doing the same exact functionality but with `.replace(/&amp;#10;/g, '<br />')` instead of what's currently used. Additionally, formatting upon adding would make the modlog entry and mod action look extremely ugly (`(user added a quote: a<b>bolded text</b>)`) for example

server/chat-commands/info.ts
```diff
@@ -2535,35 +2535,41 @@ export const commands: ChatCommands = {
...
 		if (room.settings.quotes.length >= 100) {
 			return this.errorReply(`This room already has 100 quotes, which is the maximum.`);
 		}
 		room.settings.quotes.push({userid: user.id, quote: target, date: Date.now()});
 		room.saveSettings();
-		this.privateModAction(`${user.name} added the quote "${target}"`);
-		return this.modlog(`ADDQUOTE`, null, target);
+		const collapsedQuote = Chat.collapseLineBreaksHTML(target).replace(/&#10;/g, ' ');
+		this.privateModAction(`${user.name} added a new quote: "${collapsedQuote}".`);
+		return this.modlog(`ADDQUOTE`, null, collapsedQuote);
```
Zarel, line 2572: Why is HTML being stored in modlog?

server/chat-commands/info.ts
```diff
@@ -2535,35 +2535,41 @@ export const commands: ChatCommands = {
...
+			return this.errorReply(`Your quote cannot exceed 8192 characters.`);
 		}
 		if (room.settings.quotes.length >= 100) {
 			return this.errorReply(`This room already has 100 quotes, which is the maximum.`);
 		}
 		room.settings.quotes.push({userid: user.id, quote: target, date: Date.now()});
 		room.saveSettings();
-		this.privateModAction(`${user.name} added the quote "${target}"`);
-		return this.modlog(`ADDQUOTE`, null, target);
+		const collapsedQuote = Chat.collapseLineBreaksHTML(target).replace(/&#10;/g, ' ');
```
Zarel, line 2570: Oh, wait, I think I understand now. You're just abusing `collapseLineBreaksHTML`. But why? Wouldn't `target.replace(/\n/g, ' ')` be shorter and also less buggy?

KrisXV: yes, idk why i didnt think i could do that

KrisXV:
Alright, done @zarel

Zarel:
Thanks!

PR #7363, opened 2020-09-13
Improve spread damage return value check

urkerab:
Simply replacing these instances with `damage[i]` doesn't work, since `damage[i]` can legally be `undefined` at ths point, typically for status moves, so I just copied the check used for the return value of `getSpreadDamage`.


Zarel:
Looks good, thanks!

PR #7364, opened 2020-09-13
Fix crash in /invite

AnnikaCodes:
Apparently sending an invite to `~` crashes PS. I don't know if this is the ideal way to fix this or if it's better to move the "this message can't be sent" thing, but this should work.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -544,11 +544,12 @@ export const commands: ChatCommands = {
 			return this.parse(`/pm ${targetUsername}, /invite ${targetRoom.roomid}`);
 		}
 
-		const targetUser = this.pmTarget!; // not room means it's a PM
+		const targetUser = this.pmTarget; // not room means it's a PM
 
 		if (!targetRoom) {
 			return this.errorReply(this.tr`The room "${target}" was not found.`);
 		}
+		if (!targetUser) return this.parse(`/help invite`);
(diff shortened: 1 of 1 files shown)


server/chat-commands/core.ts
```diff
@@ -544,11 +544,12 @@ export const commands: ChatCommands = {
 			return this.parse(`/pm ${targetUsername}, /invite ${targetRoom.roomid}`);
 		}
 
-		const targetUser = this.pmTarget!; // not room means it's a PM
+		const targetUser = this.pmTarget; // not room means it's a PM
 
 		if (!targetRoom) {
 			return this.errorReply(this.tr`The room "${target}" was not found.`);
 		}
+		if (!targetUser) return this.errorReply(this.tr`/invite - must be sent to a user or room`);
```
Zarel, line 552: This error seems unnecessarily vague to me? I would ``return this.parse(`/help invite`);``

AnnikaCodes: Okay.

Zarel:
Thanks!

Zarel:
I think, as a general rule of thumb, if a user uses a command that requires arguments without any arguments, we should be showing that command's help. That's what I've always done and I think it's clearer than any error message.

I mean, showing the error message followed by the command help is clearest, but I think a lot of the time the user is intentionally looking for the help, so we should just give it to them.

PR #7366, opened 2020-09-14
Add a command to view the registration time of a user

mia-pi-git:
This is a useful bot command that I thought was worth migrating.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2595,7 +2595,9 @@ export const commands: ChatCommands = {
 	regtime: 'registertime',
 	async registertime(target, room, user, connection) {
 		this.runBroadcast();
-		Monitor.countCommands(connection.ip);
+		if (Monitor.countNetRequests(connection.ip)) {
+			return this.errorReply(`You are using this command to quickly. Wait a bit and try again.`);
+		}
 		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
 		target = toID(target);
 		if (!target) target = user.id;
@@ -2611,7 +2613,7 @@ export const commands: ChatCommands = {
 		const date = new Date(result.registertime * 1000);
 		const regDate = Chat.toTimestamp(date, {human: true});
 		const regTimeAgo = Chat.toDurationString(Date.now() - date.getTime(), {precision: 1});
-		this.sendReplyBox(Chat.html`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`);
+		this.sendReplyBox(Utils.html`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`);
 	},
 	registertimehelp: [`/registertime OR /regtime [user] - Find out when [user] registered.`],
 
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2601,12 +2601,11 @@ export const commands: ChatCommands = {
 		if (!raw) return this.errorReply(`An error occurred when retrieving user data.`);
 		// not in a try-catch block because if this doesn't work, this is a problem that should be known
 		const result = JSON.parse(raw);
-		const time = result.registertime * 1000;
-		const [day, minutes] = Chat.toTimestamp(new Date(time), {human: true}).split(' ');
-		return this.sendReplyBox(
-			`The user '${target}' registered ` +
-			`${Chat.toDurationString(Date.now() - time, {precision: 1})} ago, ` +
-			`on the date ${day}, at ${minutes}.`
+		const time = new Date(result.registertime * 1000);
+		const regTime = Chat.toTimestamp(new Date(time), {human: true});
+		const regTimeAgo = Chat.toDurationString(Date.now() - time, {precision: 1});
+		this.sendReplyBox(
+			`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`
 		);
 	},
 	registertimehelp: [`/registertime OR /regtime [user] - Find out when [user] registered.`],
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2014,6 +2014,7 @@ export const commands: ChatCommands = {
 		}
 		if (showAll || target === 'autoconfirmed' || target === 'ac') {
 			buffer.push(this.tr`A user is autoconfirmed when they have won at least one rated battle and have been registered for one week or longer. In order to prevent spamming and trolling, most chatrooms only allow autoconfirmed users to chat. If you are not autoconfirmed, you can politely PM a staff member (staff have %, @, or # in front of their username) in the room you would like to chat and ask them to disable modchat. However, staff are not obligated to disable modchat.`);
+			if (!this.broadcasting) this.parse(`/regtime`);
 		}
 		if (showAll || target === 'ladder' || target === 'ladderhelp' || target === 'decay') {
 			buffer.push(`<a href="https://${Config.routes.root}/${this.tr`pages/ladderhelp`}">${this.tr`How the ladder works`}</a>`);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2599,10 +2599,13 @@ export const commands: ChatCommands = {
 		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
 		target = toID(target);
 		if (!target) target = user.id;
-		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(e => {
+		let rawResult;
+		try {
+			rawResult = await Net(`https://${Config.routes.root}/users/${target}.json`).get();
+		} catch (e) {
 			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
 			else throw new Chat.ErrorMessage(e.message);
-		});
+		}
 		// not in a try-catch block because if this doesn't work, this is a problem that should be known
 		const result = JSON.parse(raw);
 		const date = new Date(result.registertime * 1000);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2607,7 +2607,7 @@ export const commands: ChatCommands = {
 			else throw new Chat.ErrorMessage(e.message);
 		}
 		// not in a try-catch block because if this doesn't work, this is a problem that should be known
-		const result = JSON.parse(raw);
+		const result = JSON.parse(rawResult);
 		const date = new Date(result.registertime * 1000);
 		const regDate = Chat.toTimestamp(date, {human: true});
 		const regTimeAgo = Chat.toDurationString(Date.now() - date.getTime(), {precision: 1});
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2590,6 +2590,7 @@ export const commands: ChatCommands = {
 	},
 	showhelp: [`/show [url] - shows an image or video url in chat. Requires: whitelist % @ # &`],
 
+	regdate: 'registertime',
 	regtime: 'registertime',
 	async registertime(target, room, user) {
 		this.checkChat();
@@ -2598,12 +2599,12 @@ export const commands: ChatCommands = {
 		target = toID(target);
 		if (!target) target = user.id;
 		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(() => {});
-		if (!raw) return this.errorReply(`An error occurred when retrieving user data.`);
+		if (!raw) return this.sendReplyBox(`The user '${target}' is unregistered.`);
 		// not in a try-catch block because if this doesn't work, this is a problem that should be known
 		const result = JSON.parse(raw);
-		const time = new Date(result.registertime * 1000);
-		const regTime = Chat.toTimestamp(new Date(time), {human: true});
-		const regTimeAgo = Chat.toDurationString(Date.now() - time, {precision: 1});
+		const date = new Date(result.registertime * 1000);
+		const regDate = Chat.toTimestamp(date, {human: true});
+		const regTimeAgo = Chat.toDurationString(Date.now() - date.getTime(), {precision: 1});
 		this.sendReplyBox(
 			`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`
 		);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -11,6 +11,7 @@
 import * as net from 'net';
 import {YoutubeInterface} from '../chat-plugins/youtube';
 import {Utils} from '../../lib/utils';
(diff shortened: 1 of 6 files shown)


sparkychildcharlie (commented):
You should also make it identify accounts that arent registered (registertime = 0).  !regdate might be a good alias to include as it is the command that side servers have been using for the past many many years

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,27 @@ export const commands: ChatCommands = {
...
 
+	regtime: 'registertime',
+	async registertime(target, room, user) {
+		this.checkChat();
+		this.runBroadcast();
+		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
+		target = toID(target);
+		if (!target) target = user.id;
+		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(() => {});
+		if (!raw) return this.errorReply(`An error occurred when retrieving user data.`);
```
Zarel, line 2656: I think it's pretty bad form to refuse to say what the error was. This is one of the design principles that's in CONTRIBUTING now.

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,27 @@ export const commands: ChatCommands = {
...
+		// not in a try-catch block because if this doesn't work, this is a problem that should be known
+		const result = JSON.parse(raw);
+		const time = result.registertime * 1000;
+		const [day, minutes] = Chat.toTimestamp(new Date(time), {human: true}).split(' ');
+		return this.sendReplyBox(
+			`The user '${target}' registered ` +
+			`${Chat.toDurationString(Date.now() - time, {precision: 1})} ago, ` +
+			`on the date ${day}, at ${minutes}.`
+		);
```
Zarel, line 2665:
```suggestion
		const time = new Date(result.registertime * 1000);
		const regTime = Chat.toTimestamp(new Date(time), {human: true});
		const regTimeAgo = Chat.toDurationString(Date.now() - time, {precision: 1});
		this.sendReplyBox(
			`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`
		);
```

Please don't use `return` on something just because it's the last function.

As for your `on, at` stuff, see design standard 1: less text is better.

Zarel:
I wonder if this should just be folded into `/whois`, instead of needing an entire separate command?

Zarel:
I wonder if the ladder and everything else should also go in `/whois`... maybe it would get too messy at that point.

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,27 @@ export const commands: ChatCommands = {
...
+	regdate: 'registertime',
+	regtime: 'registertime',
+	async registertime(target, room, user) {
+		this.checkChat();
+		this.runBroadcast();
+		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
+		target = toID(target);
+		if (!target) target = user.id;
+		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(() => {});
+		if (!raw) return this.sendReplyBox(`The user '${target}' is unregistered.`);
```
Zarel, line 2657: Is that really the only possible error here? I think we should handle network failures.

mia-pi-git: It's the most common failure, and people would panic at network failures, so imo i don't think we need to scare people?

sparkychildcharlie: For the record, not getting a response from the server means that the login server is probably down (How were you using the command in the first place).

For example the unregistered account ``Zarel is cute`` would give 
```{"username":"zareliscute","userid":"zareliscute","registertime":0,"group":0,"ratings":{}}```

It doesn't return any error.  It returns actual JSON.  I did say that if ``userdata.registertime === 0`` then it's unregistered before....

Also regarding the JSON.parse(), I'm not sure how the user json api for the login server works, but if the server really was down, it's possible you might get some html code (for error 500) instead of JSON.  That would cause an actual crash because JSON.parse won't work.

((Did you actually test it and use the command on an unregistered account?))

**Comment by sparkychildcharlie** [comment]
Also to add to Zarel's comment about including this in /whois, why not put it in the /ac command.  It would be very helpful to let users know if they're missing ranked games, or if they're just not 7 days old yet.

File: server/chat-commands/info.ts (hunk 3 above)
Comment (mia-pi-git) [reply server/chat-commands/info.ts:2657]: Yes, i did test it on a non-regged ac- that's where the error came from. 
I do not get `{"username":"name","userid":"name","registertime":0,"group":0,"ratings":{}}` as a response.

Comment (mia-pi-git) [reply server/chat-commands/info.ts:2657]: The error is `not found`.

Comment (Zarel) [reply server/chat-commands/info.ts:2657]: I don't think the correct solution to "people would panic at network failures" is "we should lie to them".

...to be clear I'm saying you should detect the 404 condition and say "User is unregistered", differently from other errors which should say "Something went wrong with the request" (and probably just forward the original error message), maybe with "try again later" in there.

Comment (mia-pi-git) [reply server/chat-commands/info.ts:2657]: Ok.

**Comment by mia-pi-git** [comment]
> Also to add to Zarel's comment about including this in /whois, why not put it in the /ac command. It would be very helpful to let users know if they're missing ranked games, or if they're just not 7 days old yet.

I presume you mean /faq ac?

**Comment by sparkychildcharlie** [comment]
> > Also to add to Zarel's comment about including this in /whois, why not put it in the /ac command. It would be very helpful to let users know if they're missing ranked games, or if they're just not 7 days old yet.
> 
> I presume you mean /faq ac?

yeah that thing, it's been a while since i was involved with help room

File: server/chat-commands/info.ts [hunk 4]
```diff
@@ -2644,6 +2645,29 @@ export const commands: ChatCommands = {
...
+	async registertime(target, room, user) {
+		this.checkChat();
+		this.runBroadcast();
+		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
+		target = toID(target);
+		if (!target) target = user.id;
+		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(e => {
+			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
+			else throw new Chat.ErrorMessage(e.message);
+		});
```
Zarel, line 2659: I know you wouldn't be able to use `const` anymore, but I really think you should be using try-catch instead of this approach.

Zarel: Mostly, I don't like mixing approaches like this. I think if you're going to use async/await, you should use all the async/await patterns, like try/catch.

Zarel:
```suggestion
		const raw = await Net(`https://${Config.routes.root}/users/${target}.json`).get().catch(e => {
			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
			else throw new Chat.ErrorMessage(e.message);
		});
```
```suggestion
		let rawResult;
		try {
			rawResult = await Net(`https://${Config.routes.root}/users/${target}.json`).get();
		} catch (e) {
			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
			throw new Chat.ErrorMessage(e.message);
		}
```

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,32 @@ export const commands: ChatCommands = {
 	},
 	showhelp: [`/show [url] - shows an image or video url in chat. Requires: whitelist % @ # &`],
 
+	regdate: 'registertime',
+	regtime: 'registertime',
+	async registertime(target, room, user) {
+		this.checkChat();
+		this.runBroadcast();
```
Zarel, line 2652: Should this be restricted to only the user themself? It seems kinda creepy to broadcast the regdate of someone else...

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,32 @@ export const commands: ChatCommands = {
...
+			else throw new Chat.ErrorMessage(e.message);
+		}
+		// not in a try-catch block because if this doesn't work, this is a problem that should be known
+		const result = JSON.parse(rawResult);
+		const date = new Date(result.registertime * 1000);
+		const regDate = Chat.toTimestamp(date, {human: true});
+		const regTimeAgo = Chat.toDurationString(Date.now() - date.getTime(), {precision: 1});
+		this.sendReplyBox(
+			`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`
+		);
```
Zarel, line 2670:
```suggestion
		this.sendReplyBox(`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`);
```

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,32 @@ export const commands: ChatCommands = {
...
+		this.runBroadcast();
+		if (!user.autoconfirmed) return this.errorReply(`Only autoconfirmed users can use this command.`);
+		target = toID(target);
+		if (!target) target = user.id;
+		let rawResult;
+		try {
+			rawResult = await Net(`https://${Config.routes.root}/users/${target}.json`).get();
+		} catch (e) {
+			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
+			else throw new Chat.ErrorMessage(e.message);
```
Zarel, line 2661:
```suggestion
			throw new Chat.ErrorMessage(e.message);
```

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,32 @@ export const commands: ChatCommands = {
 	},
 	showhelp: [`/show [url] - shows an image or video url in chat. Requires: whitelist % @ # &`],
 
+	regdate: 'registertime',
+	regtime: 'registertime',
+	async registertime(target, room, user) {
+		this.checkChat();
```
Zarel, line 2651: Is this necessary? It probably shouldn't be for `/` use, and for `!` use it should already be a part of `runBroadcast`, right?

server/chat-commands/info.ts
mia-pi-git: I mean, not really? It's already public - it would also be v useful for users who ask why they aren't autoconfirmed yet.

server/chat-commands/info.ts
mia-pi-git: I think / use would be practical; but that's just me.

server/chat-commands/info.ts
Zarel: ...okay, I guess it's worth leaving it like this for now.

server/chat-commands/info.ts
Zarel: I think you're misinterpreting me. I'm saying the user shouldn't need to be able to talk to use `/regdate`.

server/chat-commands/info.ts
```diff
@@ -2644,6 +2645,30 @@ export const commands: ChatCommands = {
...
+		} catch (e) {
+			if (e.message.includes('Not found')) throw new Chat.ErrorMessage(`User '${target}' is unregistered.`);
+			throw new Chat.ErrorMessage(e.message);
+		}
+		// not in a try-catch block because if this doesn't work, this is a problem that should be known
+		const result = JSON.parse(rawResult);
+		const date = new Date(result.registertime * 1000);
+		const regDate = Chat.toTimestamp(date, {human: true});
+		const regTimeAgo = Chat.toDurationString(Date.now() - date.getTime(), {precision: 1});
+		this.sendReplyBox(Chat.html`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`);
```
Zarel, line 2668:
```suggestion
		this.sendReplyBox(Utils.html`The user '${target}' registered ${regTimeAgo} ago, at ${regDate}.`);
```

server/chat-commands/info.ts
mia-pi-git: Ah. I think it probably would be prudent, so someone (cough simplechimp cough) can't abuse it to send too many requests in a short period. But i'll defer to you.

sparkychildcharlie:
I noticed that it requires autoconfirmed to use the command - and some newer users wouldn't be able to check themselves.  I do understand why you would do it to avoid abuse, but it might sorta defeat the purpose.  It might be interesting to add some sort of special cooldown per-ip for this and any other slow commands on the server for non autoconfirmed/trusted users.  (Asking @Zarel for feedback owo)

(I have that on sorting/http fetching commands on my own server)

server/chat-commands/info.ts
Zarel: Does `checkChat` put it on a chat throttle, though? We might want to manually put it on the chat throttle?

Zarel: What's your goal in this? Because I don't believe `checkChat` will prevent simplechimp from abusing it.

Zarel:
The least we should do is probably cache its output to prevent this sort of thing. But we might also want a separate throttle so you can't just check a whole bunch of random usernames in a loop.

Zarel:
I think it makes sense to add a Net throttle specifically – flag all Net commands and throttle them separately from other commands.

We can actually just reuse the existing team validator throttle.

Zarel:
(We talked about this on Discord and agreed to use a throttle based on the teamvalidator throttle, but not actually share the throttle.)

mia-pi-git:
@Zarel bump?

server/monitor.ts
```diff
@@ -211,6 +212,16 @@ export const Monitor = new class {
 		return count > 4;
 	}
 
+	/**
+	 * Counts commands that use HTTPs requests. Returns true if too many.
```
Zarel, line 216: This documentation isn't accurate. I'm not sure we want to refactor all our monitors to `Chat.ErrorMessage`, though – I think I'd rather this just return true like the others.

mia-pi-git: Returning true it is.

mia-pi-git: Done.

server/monitor.ts
```diff
@@ -53,6 +53,7 @@ if (('Config' in global) &&
 
 export const Monitor = new class {
 	connections = new TimedCounter();
+	commands = new TimedCounter();
```
Zarel, line 56: I think `netRequests` and `countNetRequest` makes more sense? We're not counting _all_ commands, after all.

server/users.ts
```diff
@@ -361,6 +361,7 @@ export class User extends Chat.MessageContext {
 	chatQueueTimeout: NodeJS.Timeout | null;
 	lastChatMessage: number;
 	lastCommand: string;
+	lastSlowCommand: number | null;
```
Zarel, line 364:
```suggestion
```

server/users.ts
```diff
@@ -447,6 +448,7 @@ export class User extends Chat.MessageContext {
 		this.chatQueueTimeout = null;
 		this.lastChatMessage = 0;
 		this.lastCommand = '';
+		this.lastSlowCommand = null;
```
Zarel, line 451:
```suggestion
```

Zarel:
Thanks!

PR #7367, opened 2020-09-14
Searchlogs: Don't require a room

mia-pi-git:
Fixes an issue where the more buttons would throw the requireRoom error, but this doesn't actually need a room, necessarily. Also updates the commands the buttons send.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -238,8 +238,8 @@ export const LogViewer = new class {
 			buf += `<br><strong>Max results reached, capped at ${limit}</strong>`;
 			buf += `<br><div style="text-align:center">`;
 			if (total < MAX_RESULTS) {
-				buf += `<button class="button" name="send" value="/sl ${search}|${roomid}|${month}|${limit + 100}">View 100 more<br />&#x25bc;</button>`;
-				buf += `<button class="button" name="send" value="/sl ${search}|${roomid}|${month}|all">View all<br />&#x25bc;</button></div>`;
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -731,14 +731,13 @@ export const commands: ChatCommands = {
...
 	searchlogs(target, room) {
-		room = this.requireRoom();
 		target = target.trim();
 		const args = target.split(',').map(item => item.trim());
 		if (!target) return this.parse('/help searchlogs');
 		let date = 'all';
 		const searches: string[] = [];
 		let limit = '500';
-		let tarRoom = room.roomid;
+		let tarRoom = room?.roomid;
```
Zarel, line 740: This seems redundant with

```
		const curRoom = tarRoom ? Rooms.search(tarRoom) : room;
```

It feels like the entire thing should be refactored. Let me think.

Zarel:
```ts
		for (const arg of args) {
			if (arg.startsWith('room:')) {
				const id = arg.slice(5);
				room = Rooms.search(id as RoomID);
				if (!room) {
					return this.errorReply(`Room "${id}" not found.`);
				}
			} else if (arg.startsWith('limit:')) {
				limit = arg.slice(6);
			} else if (arg.startsWith('date:')) {
				date = arg.slice(5);
			} else {
				searches.push(arg);
			}
		}
		if (!room) {
			return this.parse(`/help searchlogs`);
		}
```

No need for `curRoom` or `tarRoom` which don't really make much sense to me anyway.

Zarel:
Looks good now, thanks!

PR #7368, opened 2020-09-14
Trivia: Add a new game called Mastermind

AnnikaCodes:
Requested by Aegii and the rest of the Trivia auth team.


Zarel:
Looks good.

PR #7369, opened 2020-09-15
Improve the /faq command

mia-pi-git:
QOL stuff.


server/chat-commands/info.ts
```diff
@@ -1996,6 +1996,7 @@ export const commands: ChatCommands = {
 	faq(target, room, user) {
 		if (!this.runBroadcast()) return;
 		target = target.toLowerCase().trim();
+		target = toID(target);
```
KrisXV, line 1999:
```suggestion
		target = toID(target);
```

Zarel:
Thanks!

PR #7370, opened 2020-09-15
Tournaments: Check permissions properly

KrisXV:
The conversion to the conventional commands formatting removed all permissions checks, so now regular users can modify tournaments.

This also introduced a bug where the args order for /tour settype was switched around.

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -1832,7 +1832,6 @@ export const commands: ChatCommands = {
 		forcetimer(target, room, user, connection, cmd) {
 			room = this.requireRoom();
 			checkPermissions(this, room, user);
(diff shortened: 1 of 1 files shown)


whalemer:
doesn't this stop you from changing subcommand permissions, since you need the literal `checkCan` call in the function source?

do we even need the special `room.toursEnabled` property now that `/permissions set` exists?

KrisXV:
I know this somewhat defeats the point, but I'd rather have the commands working properly at all before making them meet modern coding standards.

Zarel:
On one hand, fair, I'll merge this. On the other hand, "properly" is a strong term if you're not _actually_ going to do them properly. :p

Zarel:
The room settings should probably be transitioned to `/permissions`, I think.

PR #7371, opened 2020-09-15
Trivia: Use Room#subGame for Mastermind

AnnikaCodes:
@Zarel and @sparkychildcharlie, I think this is what you were proposing? It makes `Room#mastermindGame` track the Mastermind game, and uses `Room#game` for the rounds and finals. I also made `RoomGamePlayer`s independent of the game since all the state we really need is the room ID.

### DIFF
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -163,7 +163,7 @@ function getMastermindGame(room: Room | null) {
 	if (game.gameid !== 'mastermind') {
 		throw new Chat.ErrorMessage(room.tr`The currently running game is not Mastermind, it's ${game.title}.`);
 	}
-	return game;
+	return game as Mastermind;
 }
 
 function writeTriviaData() {
@@ -379,7 +379,7 @@ export class Trivia extends Rooms.RoomGame {
 		length: string, questions: TriviaQuestion[], creator: string,
 		isRandomMode = false, isSubGame = false
 	) {
-		super(room);
+		super(room, isSubGame);
 		this.playerTable = {};
 		this.gameid = 'trivia' as ID;
 		this.title = 'Trivia';
@@ -1323,7 +1323,7 @@ export class Mastermind extends Rooms.RoomGame {
 
 export class MastermindRound extends FirstModeTrivia {
 	constructor(room: Room, category: string, questions: TriviaQuestion[], playerID?: ID) {
-		super(room, 'first', category, 'infinite', questions, 'Automatically Created');
+		super(room, 'first', category, 'infinite', questions, 'Autosmatically Created', false, true);
 
 		this.playerCap = 1;
 		this.minPlayers = 0;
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -179,6 +180,7 @@ export abstract class BasicRoom {
 	scavLeaderboard: AnyObject;
 	giveaway?: QuestionGiveaway | LotteryGiveaway | null;
 	gtsga?: GTSGiveaway | null;
+	mastermindGame?: Mastermind | null;
```
Zarel, line 183: Is there any reason why this needs to be directly in room, instead of just `room.game.mastermindGame` or something?

server/room-game.ts
```diff
@@ -18,7 +18,7 @@
 // globally Rooms.RoomGamePlayer
 export class RoomGamePlayer {
 	readonly num: number;
-	readonly game: RoomGame;
+	readonly roomid: RoomID;
```
Zarel, line 21: Why was this changed?

server/rooms.ts
AnnikaCodes: Well, `room.game` represents a round of Mastermind here, so `mastermindGame` needs to exist while `room.game`  is null.

server/room-game.ts
AnnikaCodes: I changed this so that I could use `RoomGamePlayer`s to track Mastermind players without technically having a `RoomGame` in progress.

server/rooms.ts
Zarel: Would it be very hard to do it the other way around? Where `room.game` is the entire Mastermind game, and `room.game.round` is a round of Mastermind?

Zarel: That would also fix your `RoomGamePlayer` issue.

AnnikaCodes: That would be _possible_ but probably overcomplicated, since Mastermind rounds are just slightly modified Trivia games, and we still want Trivia to extend RoomGame.

Zarel: I... would still massively prefer that to the weirdness going on in this PR which the RoomGame system was never designed to support.

If this is a frequent problem, I think the correct way to do this would be to spec out a `.subGame` property on roomgames which can take `RoomGame` instances.

Zarel: Either way, a simple solution is just not using RoomGame the way it was designed to be used and would need to be carefully audited and end up overcomplicated no matter what.

AnnikaCodes:
I'm gonna draft this until I make `RoomGame`s work somewhere other than `room.game`.

server/rooms.ts
AnnikaCodes: This uses `Room#subGame` now.

server/room-game.ts
AnnikaCodes: This is no longer necessary.

Zarel:
Thanks! Hopefully this is all it needed.

PR #7372, opened 2020-09-15
Random Battles Fixes

ACakeWearingAHat:
fix water stab enforcement in doubles (now forces water stab OR hyper voice, no matter the format. Only affects doubles primarina.)
remove focus sash cinccino (since it's already item-hardcoded with tail slap in another place)
clean up unused hyper voice rejection
consolidate power construct with other rejected abilities
Prevent Triage from occurring on Synthesisless Comfey

### DIFF
--- a/data/mods/gen5/random-teams.ts
+++ b/data/mods/gen5/random-teams.ts
@@ -456,7 +456,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 				} else if (ability === 'Sturdy') {
 					rejectAbility = (!!counter['recoil'] && !counter['recovery']);
 				} else if (ability === 'Swarm') {
-					rejectAbility = !counter['Bug'];
+					rejectAbility = (!counter['Bug'] || abilities.includes('Hustle'));
 				} else if (ability === 'Swift Swim') {
 					rejectAbility = (!hasMove['raindance'] && !teamDetails['rain']);
 				} else if (ability === 'Technician') {
--- a/data/mods/gen7/random-teams.ts
+++ b/data/mods/gen7/random-teams.ts
@@ -116,6 +116,9 @@ export class RandomGen7Teams extends RandomTeams {
 				case 'icebeam':
 					if (hasAbility['Tinted Lens'] && !!counter['Status']) rejected = true;
 					break;
+				case 'morningsun':
+					if (counter.damagingMoves.length < 1) rejected = true;
+					break;
 				case 'perishsong':
 					if (!hasMove['protect']) rejected = true;
 					break;
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -1172,7 +1172,7 @@ export class RandomTeams {
 					rejectAbility = (hasType['Steel'] && !hasMove['fakeout']);
 				} else if (ability === 'Triage') {
 					rejectAbility = (!counter['recovery']);
- 				} else if (ability === 'Unaware') {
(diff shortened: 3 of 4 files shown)


data/random-teams.ts
```diff
@@ -1129,8 +1126,6 @@ export class RandomTeams {
 					rejectAbility = !counter['Grass'];
 				} else if (ability === 'Own Tempo') {
 					rejectAbility = isDoubles;
-				} else if (ability === 'Power Construct') {
-					rejectAbility = true;
```
TheImmortal, line 1133: I mean, this was a placeholder for when it is released with DLC2 and it'd be rejected based on certain conditions (for example: 10% would have an Aura Break set).

data/random-teams.ts
```diff
@@ -1348,7 +1343,8 @@ export class RandomTeams {
 			item = 'Leftovers';
 
 		// Better than Leftovers
-		} else if (isLead && !['Disguise', 'Sturdy'].includes(ability) && !hasMove['substitute'] && !counter['recoil'] && !counter['recovery'] && species.baseStats.hp + species.baseStats.def + species.baseStats.spd < 255 && !isDoubles) {
+		} else if (isLead && !['Disguise', 'Sturdy'].includes(ability) && !hasMove['substitute'] && !hasMove['tailslap'] && !counter['recoil'] && !counter['recovery'] &&
+			species.baseStats.hp + species.baseStats.def + species.baseStats.spd < 255 && !isDoubles) {
```
TheImmortal, line 1347: I don't think Sash Cinccino should be removed.

data/mods/gen7/random-teams.ts
```diff
@@ -116,6 +116,9 @@ export class RandomGen7Teams extends RandomTeams {
 				case 'icebeam':
 					if (hasAbility['Tinted Lens'] && !!counter['Status']) rejected = true;
 					break;
+				case 'morningsun':
+					if (counter.damagingMoves.length < 1) rejected = true;
+					break;
```
TheImmortal, line 121: What is this for?

data/random-teams.ts
TheImmortal: At least not with a really specific hardcode like this.

ACakeWearingAHat: i mean, you already hardcoded it to give it Life Orb

data/mods/gen7/random-teams.ts
ACakeWearingAHat: stealth rock morning sun dual screens solrock exists in gen 7

data/random-teams.ts
ACakeWearingAHat: and it's reported at least once a week. But you can remove it if you want.

data/mods/gen5/random-teams.ts
```diff
@@ -456,7 +456,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 				} else if (ability === 'Sturdy') {
 					rejectAbility = (!!counter['recoil'] && !counter['recovery']);
 				} else if (ability === 'Swarm') {
-					rejectAbility = !counter['Bug'];
+					rejectAbility = (!counter['Bug'] || abilities.includes('Hustle'));
```
TheImmortal, line 459: This is the wrong fix because Hustle has a much higher rating. I found the problem.

TheImmortal:
I'm going to close this because almost all of it is included in my next update.

PR #7373, opened 2020-09-15
Helptickets: Display user language and remove prefix

AnnikaCodes:
Requested by @DaWoblefet. This will need a private code change too which I can make if this is merged.


Zarel:
Looks good, thanks.

PR #7376, opened 2020-09-16
Rooms: Remove instance of for-each

mia-pi-git:
Minor thing, as per contributing.md

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1235,11 +1235,11 @@ export class GlobalRoomState {
 	}
 
 	onCreateBattleRoom(players: User[], room: GameRoom, options: AnyObject) {
-		players.forEach(player => {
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks, although I'm not sure it's worth cluttering PRs for a single style issue like this. While this is here, though, maybe you could look for other style issues? I'd feel better about this if it was combined with other fixes.

AnnikaCodes:
There are a couple of code issues that [LGTM](https://lgtm.com/projects/g/smogon/pokemon-showdown/alerts?mode=list) caught; most of them are non-issues like the `.replace('*', '')` in IPTools or various complaints about cryptography when there is no cryptography, etc, but there are a few that are legitimate code style issues and would be good to fix.

PR #7379, opened 2020-09-17
Add more Spanish translations

AnnikaCodes:
Courtesy of Naziel and Iyarito.


Zarel:
Thanks Naziel and Iyarito!

PR #7380, opened 2020-09-17
RoomGame: Support a second game in room.subGame

AnnikaCodes:
I feel like there's something more to do here but I can't think of what it would be.

### DIFF
--- a/server/room-game.ts
+++ b/server/room-game.ts
@@ -41,7 +41,7 @@ export class RoomGamePlayer {
 		this.name = (typeof user === 'string' ? user : user.name);
 		if (typeof user === 'string') user = null;
 		this.id = user ? user.id : '';
-		if (user) {
+		if (user && !this.game.isSubGame) {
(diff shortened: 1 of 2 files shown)


Zarel:
I wonder if players should be synced?

AnnikaCodes:
I mean, the whole point of Mastermind is that each round has only a subset of the main game's players. I think that maintaining one player list for both the main and sub games would be too restrictive and make this less useful.

Zarel:
Oh, I see. Interesting.

Zarel:
Hm, after a brief review, this is probably enough.

If you want to do this on hard mode, you could add assertions (i.e. by throwing for unexpected state). But considering we already get plenty of unexpected state, that might not be advisable in practice.

AnnikaCodes:
If we get a lot of unexpected state I don't think that throwing would be a good idea, at least not in production

Zarel:
Fair. I do want to investigate, but there's no need to block this on that.

PR #7381, opened 2020-09-18
Chat: Move forcerename tracking to Monitor and Punishments

mia-pi-git:
This is so they don't get cleared on hotpatch. 
Monkeypatch (NECESSARY, OR BAD THINGS CRASH) 
```ts
>> Punishments.namefilterwhitelist = Chat.namefilterwhitelist;
Punishments.saveNameWhitelist();
Monitor.forceRenames = Chat.forceRenames;
delete Chat.forceRenames;
delete Chat.namefilterwhitelist;
```


Zarel:
Is the namefilter whitelist not permanent? It should be permanent, shouldn't it?

mia-pi-git:
> Is the namefilter whitelist not permanent? It should be permanent, shouldn't it?

Looking around, it seems like it should be. I'll add it to Punishments instead.

Zarel:
I'm going to assume a restart is sufficient.

PR #7382, opened 2020-09-18
Battlesearch: Support searching for multiple users and searching months

mia-pi-git:
Various improvements requested by Dawoblefet. 
It now searches by month, not day, has a back button, supports searching for 2 users, and the formats list is reformatted.
Screenshots: multisearch (https://prnt.sc/uj8ih4), singlesearch (https://prnt.sc/uj8ioq)

### DIFF
--- a/server/chat-plugins/modlog-viewer.ts
+++ b/server/chat-plugins/modlog-viewer.ts
@@ -20,12 +20,20 @@ import {Dex} from '../../sim/dex';
 import {Config} from '../config-loader';
 import {checkRipgrepAvailability, ModlogID} from '../modlog';
 
-interface Results {
-	[k: string]: number;
+interface BattleOutcome {
+	lost: string;
+	won: string;
+	turns: string;
+}
+
+interface BattleSearchResults {
 	totalBattles: number;
-	totalWins: number;
-	totalLosses: number;
+	/** Total battle outcomes. Null when only searching for one userid.*/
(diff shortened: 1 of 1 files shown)


Zarel:
It doesn't seem to have any links to battle results. Is that just something we've never supported?

server/chat-plugins/modlog-viewer.ts
```diff
@@ -20,12 +20,20 @@ import {Dex} from '../../sim/dex';
 import {Config} from '../config-loader';
 import {checkRipgrepAvailability, ModlogID} from '../modlog';
 
-interface Results {
-	[k: string]: number;
+interface BattleOutcome {
+	lost: string;
+	 won: string,
```
Zarel, line 25: Do you still not have `npm test` working?

mia-pi-git:
> It doesn't seem to have any links to battle results. Is that just something we've never supported?

No, it just displays battle results as numbers in a total.

DaWoblefet:
Typically matches aren't saved as replays for what we're concerned about anyway. So far as I know, battlesearch has only been used in cases of ladder boosting - we never use it to say, verify if there was harassment in a game that wasn't saved as a replay. That's only available to those with console access.

Zarel:
Seems like it's unlikely to be a problem.

Zarel:
Thanks!

PR #7383, opened 2020-09-18
Remove Idle status on login

AnnikaCodes:
Currently, users are getting their status set to Idle when they log back in after a time of being logged out. According to the comment in `users.ts`, we consider logging in enough activity to remove Idle status, but this was not borne out in the code.


Zarel:
Yeah, I remember noticing that. Good fix!

PR #7387, opened 2020-09-18
Unify /guess and /choose

Zarel:
A lot of roomgames that should be using `/choose` or `/guess` but aren't. I'm going to unify them.

Would appreciate input from whoever's maintaining Trivia right now. @sparkychildcharlie?

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1058,12 +1058,17 @@ export const commands: ChatCommands = {
 		room.game.forfeit(user);
 	},
 
+	guess: 'choose',
 	choose(target, room, user) {
 		room = this.requireRoom();
 		if (!room.game) return this.errorReply(this.tr("This room doesn't have an active game."));
 		if (!room.game.choose) return this.errorReply(this.tr("This game doesn't support /choose"));
+		if (room.game.checkChat) this.checkChat();
 		room.game.choose(user, target);
 	},
+	choosehelp: [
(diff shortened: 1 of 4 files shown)


Zarel:
`/trivia answer` looks a bit too complex to factor in, but it really should...

Zarel:
We'll wait for Mastermind/subgame to get refactored first.

sparkychildcharlie:
Annika does Trivia and I do Scavs - but unifying them like this seems nice.  The commands for the two rooms are kinda tradition by now so changing them might throw people off when they try to race to sumbitntheir answers. /ta and /scavenge - long and clunky as they are - both have history and tons of reflex habits attached to both.  I can do something to make it use Chat.errorMessage throughout scavengers since im sure there are many instances if were talking about unifying stuff

AnnikaCodes:
I maintain Trivia. I definitely want `/ta` to remain as an alias since that's what users are used to typing, but I think that aliasing it to `/guess` would be fine. I agree with the overall idea of having one command for all roomgames.

Zarel:
We can keep all the old aliases; I just want `/choose` or `/guess` (maybe `/c` as a shortcut) to also work for all roomgames going forward.

AnnikaCodes:
Sounds good to me!

Zarel:
Probably also `/answer` too

Zarel:
(Technically this will only crash hotpatch if there's a currently-running game of one of these, but I'm in no hurry so this can wait until next restart.)

PR #7389, opened 2020-09-18
Punishmentfilter: Don't activate for permalocks

AnnikaCodes:
This is currently causing Staff room to be spammed.


Zarel:
This... will need a restart to be fixed, won't it?

mia-pi-git:
> This... will need a restart to be fixed, won't it?

Restart or monkeypatch & /hotpatch punishments, yes.

PR #7390, opened 2020-09-18
Replace @ts-ignore with type assertions or other less unsafe constructs

urkerab:
... or just remove it if it's become unnecessary, of course.

I was also considering changing second parameter to `calcRecoilDamage` to be the recoil rather than the move. Currently most of the implementations assert that the recoil exists, since while the callers all know that the recoil is a tuple, that information is lost to the callee, and I have copied that style to the last place that was using a `@ts-ignore`. If instead the parameter was just the recoil tuple itself, no assertion would be necessary at all.

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -364,10 +364,9 @@ export const Conditions: {[k: string]: ConditionData} = {
 			if (data.source.hasAbility('adaptability') && this.gen >= 6) {
 				data.moveData.stab = 2;
 			}
-			// @ts-ignore
-			const hitMove: ActiveMove = new this.dex.Data.Move(data.moveData);
+			const hitMove: Move = new this.dex.Data.Move(data.moveData);
 
-			this.trySpreadMoveHit([target], data.source, hitMove);
+			this.trySpreadMoveHit([target], data.source, hitMove as ActiveMove);
 		},
 	},
 	healreplacement: {
@@ -404,8 +403,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 			return success;
 		},
 		onRestart() {
-			// @ts-ignore
-			if (this.effectData.counter < this.effect.counterMax) {
+			if (this.effectData.counter < (this.effect as Condition).counterMax!) {
 				this.effectData.counter *= 3;
 			}
 			this.effectData.duration = 2;
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -364,9 +364,9 @@ export const Conditions: {[k: string]: ConditionData} = {
 			if (data.source.hasAbility('adaptability') && this.gen >= 6) {
 				data.moveData.stab = 2;
 			}
-			const hitMove: Move = new this.dex.Data.Move(data.moveData);
+			const hitMove = new this.dex.Data.Move(data.moveData) as Move as ActiveMove;
 
-			this.trySpreadMoveHit([target], data.source, hitMove as ActiveMove);
+			this.trySpreadMoveHit([target], data.source, hitMove);
 		},
 	},
 	healreplacement: {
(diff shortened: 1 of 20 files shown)


data/conditions.ts
```diff
@@ -364,10 +364,9 @@ export const Conditions: {[k: string]: ConditionData} = {
 			if (data.source.hasAbility('adaptability') && this.gen >= 6) {
 				data.moveData.stab = 2;
 			}
-			// @ts-ignore
-			const hitMove: ActiveMove = new this.dex.Data.Move(data.moveData);
+			const hitMove: Move = new this.dex.Data.Move(data.moveData);
```
Zarel, line 367:
```suggestion
			const hitMove = new this.dex.Data.Move(data.moveData) as ActiveMove;
```

Do you actually need a signature at all if you're going to do this? Also I think I prefer the cast here rather than there.

data/mods/gen6/random-teams.ts
```diff
@@ -552,9 +549,9 @@ export class RandomGen6Teams extends RandomGen7Teams {
 				// Handle Hidden Power IVs
 				if (moveid === 'hiddenpower') {
 					const HPivs = this.dex.getType(move.type).HPivs;
-					for (const iv in HPivs) {
-						// @ts-ignore
-						ivs[iv] = HPivs[iv];
+					let iv: StatName;
+					for (iv in HPivs) {
```
Zarel, line 553: The main problem with this workaround is that it doesn't work in TypeScript 4.0 anymore. What's our plan there?

data/scripts.ts
```diff
@@ -1085,8 +1079,7 @@ export const Scripts: BattleScriptsData = {
 	},
 
 	calcRecoilDamage(damageDealt, move) {
-		// @ts-ignore
-		return this.clampIntRange(Math.round(damageDealt * move.recoil[0] / move.recoil[1]), 1);
+		return this.clampIntRange(Math.round(damageDealt * move.recoil[0]! / move.recoil[1]!), 1);
```
Zarel, line 1082:
```suggestion
		return this.clampIntRange(Math.round(damageDealt * move.recoil![0] / move.recoil![1]), 1);
```

I _think_ this is what you want?

Zarel:
Looks good otherwise, though! Thanks!

data/conditions.ts
urkerab: Sorry, I already tried this, and it just results in TS2352 Conversion ... may be a mistake.

data/mods/gen6/random-teams.ts
urkerab: Sorry, I was under the impression that this was our current workaround, so I just copied it here.

data/scripts.ts
urkerab: Yes, sorry, I mixed it up when I tried to undo my trial `calRecoilDamage` change as mentioned in the original description.

data/mods/gen6/random-teams.ts
Zarel: Oh, it's fine. You don't need to fix this now. I was just wondering if you had any ideas for what we should do after TS 4.0.

data/scripts.ts
Zarel: You don't need to apologize so much. Especially since this is a really useful change overall, and all these things are easy mistakes or nitpicks.

urkerab: I feel a strange need to apologise for apologising too much...

data/conditions.ts
Zarel:
```
			const hitMove = new this.dex.Data.Move(data.moveData) as Move as ActiveMove;
```

Looks like this works. I'll push this unless you mind.

urkerab: It's not my favourite but it's better than @ts-ignore.

urkerab: ... there's actually a GitHub user with that name, just to remind you not to use it!

Zarel: I relatively strongly prefer it. I think it's much clearer that it needs two casts, while your version with the two casts spread out make people wonder why it couldn't just be one cast.

PR #7392, opened 2020-09-19
Chat filters: Support custom redtext messages

AnnikaCodes:
Requested by @DaWoblefet.

I also refactored the `FilterWord` type because I think that an Object is better than an Array here.


Zarel:
Looks good, thanks!

PR #7393, opened 2020-09-19
Chatlog: Use ProcessManager for searches

mia-pi-git:
This PR (as the title states) now runs chatlog searches in a subprocess, so it doesn't, say, take down all of PS as easily.
It also makes a bunch of minor improvements to the code in the name of readability.


mia-pi-git:
@Zarel bump?

Zarel:
Looks good, thanks!

PR #7398, opened 2020-09-19
Add thread support to ProcessManager

Zarel:


Zarel:
I've transitioned verifier over to use worker threads and they appear to work normally. Our other processes use them for crash isolation as much as performance and should probably stay as processes.

Zarel:
Test failure seems to be a bug in Node 12 threads? We'll definitely need to bump minimum Node version to Node 14.

@monsanto I suspect this is the sort of weird thread problem I was talking about.

lib/process-manager.ts
```diff
@@ -57,40 +60,88 @@ class RawSubprocessStream extends Streams.ObjectReadWriteStream<string> {
...
+		case 'cluster':
+			this.clusterWorker = cluster.fork(env);
+			this.process = this.clusterWorker.process;
+			this.clusterWorker.on('disconnect', () => {
+				this.isConnected = false;
+			});
+			this.pid = this.process.pid;
+			break;
+		case 'worker':
+			this.worker = new Worker(file, {env});
```
mia-pi-git, line 97:
```suggestion
			this.worker = new Worker(file, {workerData: env});
```
While i was transitioning #7350 to ProcessWrapper, i noticed in testing that this crashes unless changed to this.

Zarel: I think this is just JavaScript being extra dumb.

I'd do this, probably:

```suggestion
			const options = {};
			if (env) options.env = env;
			this.worker = new Worker(file, options);
```

mia-pi-git: This also crashes. It explicitly needs env to be `{workerData: env}` as i suggested, or else workerData will be undefined in the worker.

monsanto:
If keeping Node 12 support is desired you could create processes when workers are requested on <14.

lib/process-manager.ts
```diff
@@ -57,40 +60,88 @@ class RawSubprocessStream extends Streams.ObjectReadWriteStream<string> {
...
 
-interface ProcessWrapper {
-	load: number;
-	process: ChildProcess | Worker;
-	release: () => Promise<void>;
-	getProcess: () => ChildProcess;
+abstract class ProcessWrapper {
+	process: ChildProcess | null = null;
+	clusterWorker: cluster.Worker | null = null;
+	worker: worker_threads.Worker | null = null;
```
monsanto, line 76: These types are imprecise (all could be `null`) forcing you to bypass the typechecker later with the `!` operator. Can avoid this by using a union of the form `{type: 'process', data: ChildProcess} | {type: 'worker', data: worker_threads.Worker} | ...`

Zarel: I did consider that but in general I don't like having worse performance to satisfy the typechecker.

Zarel:
lol, it turns out I was wrong; it was just a test I missed because it wasn't covered by `npm test`.

lib/process-manager.ts
Zarel: Wait, that makes no sense. The documentation doesn't say anything about requiring `workerData`...

Zarel: Exactly how and why does it crash? The current code works fine for me in `verifier`. Is it Windows that's crashing, or is it the way you're using it?

Zarel: Are you using it by passing non-JSON to env? Is that why it's crashing?

mia-pi-git: I don't use Windows. What i'm doing is passing the data i would set as options.workerData property directly - the error i get is that workerData is undefined in the worker.
Code: 
```ts
	constructor(options: SQLOptions) {
		super(__filename, 'worker', options);
```

lib/process-manager.ts
Zarel: You could define `ProcessWrapper` externally to have that type, but most of the assertions are inside `ProcessWrapper` so it wouldn't help much anyway.

lib/process-manager.ts
Zarel: ...that's not what it's for.

Zarel: Of course if you put workerData in the env field, it won't appear as workerData. Did you realize that's what you were doing?

mia-pi-git: I see. My bad, then.

AnnikaCodes:
Node 12 and 13 are no longer supported by PS as of #8369.

PR #7399, opened 2020-09-19
Fix renaming battle rooms

AnnikaCodes:
Third time's the charm? The issue was that `RoomGame#roomid` wasn't being updated. I fixed that, and it seems to work fine on localhost (I believe that the replay server already handles replay uploads from battles with passwords, so I didn't put in the extra work to test this on my replay server, but I can do that if we're unsure about the replay server functionality).


Zarel (approved):
I continue to not see obvious problems here. I guess we'll have to find out when we merge.

PR #7400, opened 2020-09-20
Fix missing variable renames after PR #6670

urkerab:
The upshot is that if you use, say, the `/om cup` command, this finds `gen8350cup` as a single near match (because it thinks that this is an om called "cup" in generation 8350...), but fails to display all of the details correctly.

I tried to clarify by inlining the `format` variable but of course that results in a line of code longer than 120 characters...

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -1839,19 +1839,19 @@ export const commands: ChatCommands = {
 				if (rules.length > 0) {
 					rulesetHtml = `<details><summary>Banlist/Ruleset</summary>${rules.join("<br />")}</details>`;
 				} else {
-					rulesetHtml = `No ruleset found for ${format.name}`;
+					rulesetHtml = `No ruleset found for ${subformat.name}`;
 				}
 			}
-			let formatType: string = (format.gameType || "singles");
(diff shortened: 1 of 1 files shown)


Zarel:
I don't understand what you mean by "inlining the `format` variable". If you could post the code you wanted to do, maybe I could offer suggestions?

urkerab:
Basically I tried to replace this:
```javascript
		let formatList: string[] = [];
		const format = Dex.getFormat(targetId);
		if (['Format', 'ValidatorRule', 'Rule'].includes(format.effectType)) formatList = [targetId];
		if (!formatList.length) {
			formatList = Object.keys(Dex.formats);
		}
```
with this:
```javascript
		const formatList = ['Format', 'ValidatorRule', 'Rule'].includes(Dex.getFormat(targetId).effectType) ? [targetId] : Object.keys(Dex.formats);
```

Zarel:
Should be relatively simple with an intermediate variable.

```ts
const targetIsFormat = ['Format', 'ValidatorRule', 'Rule'].includes(Dex.getFormat(targetId).effectType);
const formatList = targetIsFormat ? [targetId] : Object.keys(Dex.formats);
```

PR #7401, opened 2020-09-20
Make modnotes global in rooms with shared modlogs

AnnikaCodes:
This mainly applies to modnotes in help tickets and battles, as well as global staff modnoting in groupchats (in case of misbehavior in a groupchat).


Zarel:
Looks good, thanks!

PR #7404, opened 2020-09-22
Add more Portuguese translations

AnnikaCodes:
Courtesy of Mitsuki and woolf.


Zarel:
tyty Mitsuki and woolf!

PR #7406, opened 2020-09-22
 Add more French and Italian translations

AnnikaCodes:
Courtesy of Alpha and Milak for Italian, and Teclis and Lionyx for French. (I wish I could set translators as authors but I don't think most of them have GitHub accounts...)


Zarel:
Thanks, Alpha, Milak, Teclis, Lionyx!

PR #7409, opened 2020-09-22
Helpfilter: Fix crash in suggestions

mia-pi-git:


server/chat-plugins/help.ts
```diff
@@ -142,7 +142,7 @@ export class HelpResponder {
 		return null;
 	}
 	getFaqID(faq: string) {
```
Zarel, line 144:
```suggestion
	getFaqID(faq: string | undefined) {
```

It seems like this signature is wrong? And/or some function that calls it is wrong?

mia-pi-git: It's not _supposed_ to be undefined, but apparently it could be, and this is a cleaner fix, i felt.

Zarel:
There's still a type signature further up the call stack which is wrong and needs to be corrected, unfortunately.

mia-pi-git:
> There's still a type signature further up the call stack which is wrong and needs to be corrected, unfortunately.

Actually, based on the crash, there isn't. `getFaqID` is directly called from the offending command, where I mistakenly forgot to assert FAQ was defined. This corrects that in a cleaner way.

Zarel:
This means that the offending command is wrong...

mia-pi-git:
Replying here too since it went unnoticed on Discord: it's not wrong, I prefer the validation is done in the shared functions since that way it covers more use-cases. It corrects itself here so that the command doesn't need more code, code that would also need to be elsewhere.

Zarel:
I'm going to merge because this is still a net improvement, but what you're saying doesn't change that the callers are wrong.

Zarel:
Actually I have time to fix this myself.

Zarel:
Thanks!

PR #7410, opened 2020-09-23
Fix typo in Dubious Disc

ivanlonel:
https://bulbapedia.bulbagarden.net/wiki/Dubious_Disc
https://www.serebii.net/itemdex/dubiousdisc.shtml

It's already spelled correctly in data/items.ts and data/text/items.ts.


Zarel:
Thanks!

PR #7411, opened 2020-09-23
Helptickets: Close a user's ticket on global ban

AnnikaCodes:


server/chat-plugins/helptickets.ts
```diff
@@ -1365,3 +1370,13 @@ export const commands: ChatCommands = {
 		`/helpticket delete [user] - Deletes a user's ticket. Requires: &`,
 	],
 };
+
+export const punishmentfilter: PunishmentFilter = (user, punishment) => {
```
Zarel, line 1374:
```suggestion
export const punishmentfilter: Chat.PunishmentFilter = (user, punishment) => {
```

Can we do this instead of exporting a new global type?

AnnikaCodes: Sure; although I think we should do that for loginfilters and statusfilters too.

Zarel:
Looks good, thanks!

PR #7412, opened 2020-09-23
Fix typo in thunderpunch

ivanlonel:
This was the only "thudnerpunch" occurrence in the repo.


Zarel:
Huh do we not have this in the Dex validator either?

PR #7413, opened 2020-09-23
Rooms: Check globalAuth for autojoining staff rooms

mia-pi-git:
According to staff, when you're hideranking, autojoining staff room won't work, so this fixes that.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1313,11 +1313,12 @@ export class GlobalRoomState {
 				i--;
 				continue;
 			}
+			const globalGroup = Users.globalAuth.get(user.id);
 			if (room.settings.staffAutojoin === true && user.isStaff ||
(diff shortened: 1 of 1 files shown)


Zarel:
`staffAutojoin` is super old.

Here's what I think it should be: If you have permission to join the room, you should join.

Zarel:
I think there's actually no need for autojoin and staffAutojoin to work differently anymore. They should both be the same setting in the future.

mia-pi-git:
> I think there's actually no need for autojoin and staffAutojoin to work differently anymore. They should both be the same setting in the future.

Alright, I'll just refactor it in this PR, then?

Zarel:
Yes, that'd be great.

Zarel:
Actually, I ended up implementing this.

PR #7414, opened 2020-09-24
IPTools: Check proxy IPs before other host types

AnnikaCodes:


Zarel:
I don't understand what this is for, but it seems harmless enough. This doesn't interfere with the Hurricane Electric hardcodes, right?

Zarel:
Oh, of course it doesn't.

Zarel:
Yeah, this is definitely slightly better. Was there any specific reason you chose this approach, though?

AnnikaCodes:
According to asgdf, some people run Tor nodes out of IPs that would otherwise be marked as residential based on the host.

AnnikaCodes:
Plus, I'd generally trust a designation on a specific IP address over a broader range designation.

Zarel:
```js
		if (this.residentialHosts.has(host) || host.endsWith('/res')) {
			return 'res';
		}
```

This part makes me worry, though. I'm surprised it's still around after your refactors. Is this function supposed to be called before or after hosts are tagged with `/res` or `/proxy` or whatever?

If before, the check is pointless; if after, the other checks are pointless and it should be replaced with `return host.split('/')[1];`

The only use I can imagine is if it's used to replace the current host tag? Or to add a host tag only to hosts that don't already have one?

AnnikaCodes:
This is called on the result from `IPTools.getHost()`, which handles reverse DNS lookups (which can tag a host with `/res`, but doesn't check PS' manually added IP and host designations). That's why we should acknowledge the rDNS result but also check it against what PS knows about that IP and host.

PR #7417, opened 2020-09-24
Helptickets: Make the HelpTicket game class hotpatchable

mia-pi-git:
This will, as the title states, make helpticket games hotpatchable by /hotpatch chat, rather than having the risk of old instances causing crashes, as we've seen before.

### DIFF
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -9,22 +9,22 @@ const TICKET_BAN_DURATION = 48 * 60 * 60 * 1000; // 48 hours
 Punishments.roomPunishmentTypes.set(`TICKETBAN`, 'banned from creating help tickets');
 
 interface TicketState {
-	creator: string;
-	userid: ID;
-	open: boolean;
-	active: boolean;
-	type: string;
-	created: number;
-	claimed: string | null;
-	ip: string;
-	resolution?: 'unknown' | 'dead' | 'unresolved' | 'resolved';
-	result: TicketResult | null;
-	claimQueue?: string[];
-	involvedStaff?: ID[];
-	createTime?: number;
-	emptyRoom?: boolean;
-	firstClaimTime?: number;
-	unclaimedTime?: number;
+	readonly creator: string;
+	readonly userid: ID;
+	readonly open: boolean;
+	readonly active: boolean;
+	readonly type: string;
+	readonly created: number;
+	readonly claimed: string | null;
+	readonly ip: string;
+	readonly resolution?: 'unknown' | 'dead' | 'unresolved' | 'resolved';
+	readonly result: TicketResult | null;
+	readonly claimQueue?: string[];
+	readonly involvedStaff?: ID[];
+	readonly createTime?: number;
+	readonly emptyRoom?: boolean;
+	readonly firstClaimTime?: number;
+	readonly unclaimedTime?: number;
 }
 type TicketResult = 'approved' | 'valid' | 'assisted' | 'denied' | 'invalid' | 'unassisted' | 'ticketban' | 'deleted';
 
@@ -414,7 +414,7 @@ try {
 				continue;
 			}
 			// Close open tickets after a restart
-			if (ticket.open && !Chat.oldPlugins.helptickets) ticket.open = false;
+			if (!Chat.oldPlugins.helptickets) continue;
 
 			// recreate tickets
 			for (const room of Rooms.rooms.values()) {
@@ -628,7 +628,6 @@ export const pages: PageTable = {
 				const helpRoom = Rooms.get(`help-${ticket.userid}`);
 				if (!helpRoom) {
 					// Should never happen
-					tickets[ticket.userid].open = false;
 					writeTickets();
 				} else {
 					if (!helpRoom.auth.has(user.id)) helpRoom.auth.set(user.id, '+');
@@ -1264,25 +1263,18 @@ export const commands: ChatCommands = {
 		close(target, room, user) {
 			if (!target) return this.parse(`/help helpticket close`);
 			let result = !(this.splitTarget(target) === 'false');
-			const tickets = getTickets();
-			const ticket = tickets[toID(this.inputUsername)];
-			if (!ticket || !ticket.open || (ticket.userid !== user.id && !user.can('lock'))) {
-				return this.errorReply(this.tr`${this.inputUsername} does not have an open ticket.`);
-			}
-			const helpRoom = Rooms.get(`help-${ticket.userid}`) as ChatRoom | null;
+
+			const helpRoom = Rooms.get(`help-${toID(target)}`) as ChatRoom | null;
 			if (helpRoom) {
 				const ticketGame = helpRoom.getGame(HelpTicket)!;
-				if (ticket.userid === user.id && !user.isStaff) {
+				if (ticketGame.userid === user.id && !user.isStaff) {
 					result = !!(ticketGame.firstClaimTime);
 				}
 				ticketGame.close(result, user);
 			} else {
-				ticket.open = false;
-				notifyStaff();
-				writeTickets();
+				return this.errorReply(this.tr`${this.inputUsername} does not have an open ticket.`);
 			}
-			ticket.claimed = user.name;
-			this.sendReply(`You closed ${ticket.creator}'s ticket.`);
+			this.sendReply(`You closed ${this.inputUsername}'s ticket.`);
 		},
 		closehelp: [`/helpticket close [user] - Closes an open ticket. Requires: % @ &`],
 
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -18,7 +18,7 @@ interface TicketState {
 	claimed: string | null;
 	ip: string;
 	resolution?: 'unknown' | 'dead' | 'unresolved' | 'resolved';
-	result?: TicketResult;
+	result: TicketResult | null;
 	claimQueue?: string[];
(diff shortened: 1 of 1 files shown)


server/chat-plugins/helptickets.ts
```diff
@@ -508,7 +516,9 @@ function checkIp(ip: string) {
 for (const room of Rooms.rooms.values()) {
 	if (!room.settings.isHelp || !room.game) continue;
 	const game = room.getGame(HelpTicket)!;
```
KrisXV, line 518:
```suggestion
	const game = room.getGame(HelpTicket);
```

mia-pi-git: Unfortunately, this has to be asserted. The above if is an or, not an and. TS thinks it can be null still.

KrisXV: gross

server/chat-plugins/helptickets.ts
```diff
@@ -1380,3 +1390,24 @@ export const punishmentfilter: Chat.PunishmentFilter = (user, punishment) => {
...
+			involvedStaff: [...game.involvedStaff],
+			createTime: game.createTime,
+			emptyRoom: game.emptyRoom,
+			firstClaimTime: game.firstClaimTime,
+			unclaimedTime: game.unclaimedTime,
+		};
+		tickets[game.ticket.userid] = entry;
+	}
+	writeTickets();
+};
```
Zarel, line 1413: I think it's much better to track these sorts of things in `game.ticket` only, instead of having to write a `destroy` handler.

Zarel: your code also desyncs `tickets` and `game.ticket` which causes a bug on restarts where it'll have old data if you hotpatch chat.

Zarel: We talked about this in Discord: the main problem is that some properties are inside `game.ticket` and some are directly in `game`. They should either all be moved to `game.ticket` or all moved to `game` (and `.ticket` would be autogenerated from `game` when `writeTickets` is called). Either approach would remove the need for this `destroy` function.

server/chat-plugins/helptickets.ts
```diff
@@ -17,10 +17,18 @@ interface TicketState {
...
+	involvedStaff?: ID[];
+	createTime?: number;
+	emptyRoom?: boolean;
+	firstClaimTime?: number;
+	unclaimedTime?: number;
 }
 type TicketResult = 'approved' | 'valid' | 'assisted' | 'denied' | 'invalid' | 'unassisted' | 'ticketban' | 'deleted';
 
-const tickets: {[k: string]: TicketState} = {};
+export const tickets: {[k: string]: TicketState} = {};
```
Zarel, line 31: Do you need this variable at all? It seems unused at this point.

Zarel: Like, clearly it's used all over the place, but all those uses are bugs now, aren't they?

mia-pi-git: Are they bugs? It seems a valid way to track state beyond just the roomgames.

Zarel: You're not actually tracking state in this, though.

Zarel: Like you should just be getting the data from the ticket directly.

server/chat-plugins/helptickets.ts
```diff
@@ -17,40 +17,30 @@ interface TicketState {
...
 	ip: string;
+	resolution?: 'unknown' | 'dead' | 'unresolved' | 'resolved';
+	result: TicketResult | null;
+	claimQueue?: string[];
+	involvedStaff?: ID[];
+	createTime?: number;
+	emptyRoom?: boolean;
+	firstClaimTime?: number;
+	unclaimedTime?: number;
 }
```
Zarel, line 28: `TicketState` should all be `readonly`. There's a few obvious bugs due to writing to this.

server/chat-plugins/helptickets.ts
```diff
@@ -1210,6 +1264,7 @@ export const commands: ChatCommands = {
 		close(target, room, user) {
 			if (!target) return this.parse(`/help helpticket close`);
 			let result = !(this.splitTarget(target) === 'false');
+			const tickets = getTickets();
 			const ticket = tickets[toID(this.inputUsername)];
```
Zarel, line 1268: For instance, in `ticket close`, too. You set `ticket.open = false` here, which does nothing.

Zarel: This code also contains the assumption that the corresponding Help room is ``Rooms.get(`help-${ticket.userid}`)`` which is also wrong.

Actually, there are a lot of false assumptions here which this refactor is revealing. Should tickets be retained beyond the Help room being deleted? I might need to investigate personally.

mia-pi-git: I don't think it should be saved beyond the ticket deletion. 
Why would the help room not be help-ticket.userid, though?

mia-pi-git: I pushed the readonly changes and tried to change the assumptions i noticed, is that better-ish?

Zarel: `/ht list` contains a lot of old tickets where the room no longer exists. Would those just all be gone in your refactor?

mia-pi-git: They would, yes. Is that not desirable? I thought that's what you were going for - I can remove it if it's not.

Zarel: When I suggested that ticket state could be autogenerated from existing tickets, I didn't realize we kept old ticket state past when we kept old tickets. I wouldn't have suggested that if I knew.

I'm not _opposed_ to changing this, but it definitely wasn't intentional on my part.

Zarel: I asked Staff, they agree it's no big deal to get rid of expired tickets.

server/chat-plugins/helptickets.ts
```diff
@@ -1057,14 +1109,14 @@ export const commands: ChatCommands = {
...
 			let ticket = tickets[user.id];
 			const ipTicket = checkIp(user.latestIp);
 			if (ticket?.open || ipTicket) {
 				if (!ticket && ipTicket) ticket = ipTicket;
 				const helpRoom = Rooms.get(`help-${ticket.userid}`);
 				if (!helpRoom) {
 					// Should never happen
-					tickets[ticket.userid].open = false;
 					writeTickets();
 				} else {
```
Zarel, line 1121: This code doesn't do anything, and is also no longer possible under the current arrangement.

server/chat-plugins/helptickets.ts
```diff
@@ -1057,14 +1109,14 @@ export const commands: ChatCommands = {
 			if (!user.named) return this.popupReply(this.tr`You need to choose a username before doing this.`);
 			const banMsg = HelpTicket.checkBanned(user);
 			if (banMsg) return this.popupReply(banMsg);
+			const tickets = getTickets();
 			let ticket = tickets[user.id];
```
Zarel, line 1113:
```suggestion
			let helpRoom = Rooms.get(`help-${user.id}`);
			let ticket = helpRoom?.ticket;
```

server/chat-plugins/helptickets.ts
```diff
@@ -569,14 +620,14 @@ export const pages: PageTable = {
 
 			const banMsg = HelpTicket.checkBanned(user);
 			if (banMsg) return connection.popup(banMsg);
+			const tickets = getTickets();
 			let ticket = tickets[user.id];
```
Zarel, line 624:
```suggestion
			let helpRoom = Rooms.get(`help-${user.id}`);
			let ticket = helpRoom?.ticket;
```

server/chat-plugins/helptickets.ts
Zarel: (There's more code needed to make this work, but you should be able to take it from here.)

server/chat-plugins/helptickets.ts
```diff
@@ -1343,16 +1380,14 @@ export const commands: ChatCommands = {
 			// This is a utility only to be used if something goes wrong
 			this.checkCan('makeroom');
 			if (!target) return this.parse(`/help helpticket delete`);
-			const ticket = tickets[toID(target)];
-			if (!ticket) return this.errorReply(this.tr`${target} does not have a ticket.`);
-			const targetRoom = Rooms.get(`help-${ticket.userid}`);
+			const targetRoom = Rooms.get(`help-${toID(target)}`);
 			if (targetRoom) {
 				targetRoom.getGame(HelpTicket)!.deleteTicket(user);
```
Zarel, line 1385: Is it possible for the room not to have a ticket?

server/chat-plugins/helptickets.ts
```diff
@@ -1210,24 +1255,18 @@ export const commands: ChatCommands = {
 		close(target, room, user) {
 			if (!target) return this.parse(`/help helpticket close`);
 			let result = !(this.splitTarget(target) === 'false');
```
Zarel, line 1257: Why isn't CI complaining about an unused variable? I don't see this used anywhere.

server/chat-plugins/helptickets.ts
mia-pi-git: I don't believe so?

Zarel: I guess we'll find out.

mia-pi-git:
Don't know if this is really relevant anymore. Mostly we just use text tickets now anyway. Closing this for now, since I see little point in it.

PR #7418, opened 2020-09-25
Auth: Make whitelist/prizewinner not give trusted status

mia-pi-git:
This seemed a better solution than hardcoding `!['^', ' '].includes(this.get(userid))`. Bug happens because they have a rank in the auth table (which it checks) other than `+`, `' '`, and we didn't account for that implementing whitelist. 
This needs a pretty simple monkeypatch in order to take affect (no, it won't crash things, but giving trusted wrongly is a bit of an issue): 
```ts
>> Users.Auth.prototype.isStaff = function(userid) {
return this.has(userid) && Users.Auth.atLeast(this.get(userid), '%');
}
```

### DIFF
--- a/server/user-groups.ts
+++ b/server/user-groups.ts
@@ -56,7 +56,7 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 		return super.get(user) || Auth.defaultSymbol();
 	}
 	isStaff(userid: ID) {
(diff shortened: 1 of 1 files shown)


Zarel:
I think there was an older PR we never resolved fixing this, but it got bogged down in questions of how exactly `isStaff` would work. This is a good incremental change, though, so I'm accepting it here.

PR #7420, opened 2020-09-25
Fix Flooring in /statcalc

PartMan7:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8607763

I've used ~~ instead of Math.floor for performance reasons, but it shouldn't be much of a difference either way.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -1411,7 +1411,7 @@ export const commands: ChatCommands = {
 		if (realSet) {
 			if (!baseSet) {
 				if (calcHP) {
-					baseStat = Math.ceil((100 * realStat - 10 - level * (~~(ev / 4) + iv + 100)) / (2 * level));
+					baseStat = Math.ceil((100 * realStat - 10 - level * (Math.floor(ev / 4) + iv + 100)) / (2 * level));
 				} else {
 					if (!positiveMod) {
 						realStat *= (2 + modifier) / 2;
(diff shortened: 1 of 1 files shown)


urkerab:
If you were worried about performance then wouldn't you use `(ev >> 2)` in the first place?

PartMan7:
I'll do that, then. Didn't know about it. 'o.o

monsanto:
I think optimizations which make the code harder to read should be accompanied by a benchmark proving the gain is worth it.

server/chat-commands/info.ts
```diff
@@ -1411,15 +1411,15 @@ export const commands: ChatCommands = {
...
+					baseStat = Math.ceil((100 * realStat - 10 - level * (Math.floor(ev / 4) + iv + 100)) / (2 * level));
 				} else {
 					if (!positiveMod) {
 						realStat *= (2 + modifier) / 2;
 					} else {
 						realStat *= 2 / (2 + modifier);
 					}
 
-					baseStat = Math.ceil((100 * Math.ceil(realStat) - nature * (level * (ev / 4 + iv) + 500)) / (2 * level * nature));
+					baseStat = Math.ceil((100 * Math.ceil(realStat) - nature * (level * (Math.floor(ev / 4) + iv) + 500)) / (2 * level * nature));
```
Zarel, line 1422:
```suggestion
					baseStat = Math.ceil(
						(100 * Math.ceil(realStat) - nature * (level * (Math.floor(ev / 4) + iv) + 500)) /
						(2 * level * nature)
					);
```

This is probably the most readable way to come in under the line limit.

Zarel:
Thanks!

PR #7422, opened 2020-09-26
Add the alias /profile to /whois

mia-pi-git:
https://www.smogon.com/forums/threads/profile-command.3666431/ as per this.


Zarel:
ty

PR #7424, opened 2020-09-26
makegroupchat - prevent HTML injection

sparkychildcharlie:
Was messing around with HTML injection for quotes in room titles and noticed this one


Zarel:
Thanks!

PR #7425, opened 2020-09-26
Quotes: escape some HTML

sparkychildcharlie:
While no one on main is dumb enough to put HTML in a room title, we can't guarantee the same for other servers.

### DIFF
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -139,7 +139,7 @@ export const pages: PageTable = {
 			return `${buffer}<h2>This room has no quotes.</h2></div>`;
 		}
 
-		buffer += `<h2>Quotes for ${Utils.escapeHTML(room.title)} (${roomQuotes.length}):</h2>`;
+		buffer += Utils.html`<h2>Quotes for ${room.title} (${roomQuotes.length}):</h2>`;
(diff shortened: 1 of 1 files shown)


server/chat-plugins/quotes.ts
```diff
@@ -139,7 +139,7 @@ export const pages: PageTable = {
 			return `${buffer}<h2>This room has no quotes.</h2></div>`;
 		}
 
-		buffer += `<h2>Quotes for ${room.title} (${roomQuotes.length}):</h2>`;
+		buffer += `<h2>Quotes for ${Utils.escapeHTML(room.title)} (${roomQuotes.length}):</h2>`;
```
AnnikaCodes, line 142:
```suggestion
		buffer += Utils.html`<h2>Quotes for ${room.title} (${roomQuotes.length}):</h2>`;
```

sparkychildcharlie: dont see any difference but whatever

sparkychildcharlie: performance wise mine might be a little tiny bit faster since the roomQuotes.length part can never be anything but ``/[0-9]+/``

mia-pi-git (approved):
Good catch.

sparkychildcharlie:
shit misclick

Zarel:
Thanks!

PR #7426, opened 2020-09-27
Moderation: Refactor room promotions to a function

AnnikaCodes:
As per https://github.com/smogon/pokemon-showdown/pull/7344#issuecomment-690748979.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -19,6 +19,85 @@ const HOURMUTE_LENGTH = 60 * 60 * 1000;
 /** Require reasons for punishment commands */
 const REQUIRE_REASONS = true;
 
+/**
+ * Promotes a user within a room. Returns a User object if a popup should be shown to the user,
+ * and null otherwise. Throws a Chat.ErrorMesage on an error.
+ *
+ * @param promoter the User object of the user who is promoting
+ * @param room the Room in which the promotion is happening
+ * @param userid the ID of the user to promote
+ * @param symbol the GroupSymbol to promote to
+ * @param username the username of the user to promote
+ * @param force whether or not to forcibly promote
+ */
+export function runPromote(
+	promoter: User,
+	room: Room,
+	userid: ID,
+	symbol: GroupSymbol,
+	username?: string,
+	force?: boolean
+) {
+	const targetUser = Users.getExact(userid);
+	username = username || userid;
+	if (!username) return;
+
+	if (userid.length > 18) {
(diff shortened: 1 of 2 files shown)


server/chat-commands/moderation.ts
```diff
@@ -95,70 +174,27 @@ export const commands: ChatCommands = {
...
-				}
-			}
-			if (targetUser?.locked && room.persist && room.settings.isPrivate !== true && nextGroup.rank > 2) {
-				this.errorReply(`${name} is locked and can't be promoted.`);
+			const oldSymbol = room.auth.getDirect(userid);
+			let shouldPopup;
+			try {
+				shouldPopup = runPromote(user, room, userid, nextSymbol, toPromote, force);
+			} catch (err) {
+				if (err.name?.endsWith('ErrorMessage')) this.errorReply(err.message);
```
KrisXV, line 185: shouldnt you still `throw new Chat.ErrorMessage(err.message);`?

Zarel: No, if you're promoting many users, you don't want one failure to stop the other promotions.

test/server/chat-commands/moderation.js
```diff
@@ -0,0 +1,68 @@
...
+		this.user = makeUser('Annika', '127.0.0.1');
+		this.user.setGroup('&');
+
+		this.targetUser = makeUser('Heart of Etheria', '127.0.0.1');
+	});
+
+	it('should not promote users with usernames >18 characters long', () => {
+		assert.throws(() => moderation.runPromote(this.user, this.room, 'a'.repeat(19), '+', undefined, true));
+		assert.doesNotThrow(() => moderation.runPromote(this.user, this.room, 'a'.repeat(18), '+', undefined, true));
+	});
```
Zarel, line 36: Err wait why?

Zarel: It seems like long usernames should either be allowed or banned, "allowed but can't be promoted" seems like an unnecessary condition to me.

AnnikaCodes: People can use `/forceroompromote` to promote usernames that are longer than 18 characters, and thus not allowed, but since we are able to forcibly promote users who are offline and unregistered, this check makes sense so you can't promote some username who cannot possibly log in.

Zarel: Oh, I see.

Zarel:
Looks reasonable. I'm not sure if I like this approach, but I don't dislike it, either.

PR #7427, opened 2020-09-27
Add suspect tests plugin

TheImmortal:
A chat-plugin to allow suspect test notices to be added and removed via a command rather than modifying the formats file.

Date isn't necessary per say but I included it for the purpose of being able to see suspect deadlines in `/suspects`.

Shoutout Annika for solving some of my TS issues.

### DIFF
--- a/server/chat-plugins/suspect-tests.ts
+++ b/server/chat-plugins/suspect-tests.ts
@@ -9,7 +9,7 @@ interface SuspectTest {
 	url: string;
 }
 
-export const suspectTests: {[format: string]: SuspectTest} = JSON.parse(FS(SUSPECTS_FILE).readIfExistsSync() || "{}");
+const suspectTests: {[format: string]: SuspectTest} = JSON.parse(FS(SUSPECTS_FILE).readIfExistsSync() || "{}");
 
 function saveSuspectTests() {
 	FS(SUSPECTS_FILE).writeUpdate(() => JSON.stringify(suspectTests));
@@ -56,7 +56,7 @@ export const commands: ChatCommands = {
 				!['https://www.smogon.com/forums/threads/', 'https://www.smogon.com/forums/posts/']
 					.some(prefix => urlActual.startsWith(prefix))
 			) {
-				return this.errorReply("Suspect test URLs must be Smogon threads or posts.");
+				throw new Chat.ErrorMessage("Suspect test URLs must be Smogon threads or posts.");
 			}
 
 			this.privateGlobalModAction(`${user.name} ${suspectTests[format.id] ? "edited the" : "added a"} ${format.name} suspect test.`);
--- a/server/chat-plugins/suspect-tests.ts
+++ b/server/chat-plugins/suspect-tests.ts
@@ -1,6 +1,6 @@
 import {FS} from '../../lib/fs';
 
-const SUSPECTS_FILE = 'config/chat-plugins/suspects.json';
+const SUSPECTS_FILE = 'config/suspects.json';
 
 interface SuspectTest {
 	tier: string;
--- a/server/chat-plugins/suspect-tests.ts
+++ b/server/chat-plugins/suspect-tests.ts
@@ -52,10 +52,7 @@ export const commands: ChatCommands = {
 			const dateActual = `${month}/${day}`;
 
 			const urlActual = url.trim();
-			if (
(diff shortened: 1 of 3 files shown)


DaWoblefet:
Bless you and your entire family

server/chat-plugins/suspect-tests.ts
```diff
@@ -0,0 +1,101 @@
...
+			if (!isValidDate) return this.errorReply("Dates must be in the format MM/DD.");
+			const dateActual = `${month}/${day}`;
+
+			const urlActual = url.trim();
+			if (
+				!['https://www.smogon.com/forums/threads/', 'https://www.smogon.com/forums/posts/']
+					.some(prefix => urlActual.startsWith(prefix))
+			) {
+				return this.errorReply("Suspect test URLs must be Smogon threads or posts.");
+			}
```
KrisXV, line 60:
```suggestion
			if (/^https:\/\/www\.smogon\.com\/forums\/(posts|threads)\//.test(prefix)) {
				throw new Chat.ErrorMessage("Suspect test URLs must be Smogon threads or posts.");
			}
```

server/chat-plugins/suspect-tests.ts
```diff
@@ -0,0 +1,101 @@
...
+
+			this.privateGlobalModAction(`${user.name} ${suspectTests[format.id] ? "edited the" : "added a"} ${format.name} suspect test.`);
+			this.globalModlog('SUSPECTTEST', null, `by ${user.id}: ${suspectTests[format.id] ? "edited" : "added"} ${format.name}`);
```
KrisXV, line 63:
```suggestion
			this.globalModlog('SUSPECTTEST ${suspectTests[format.id] ? "EDIT" : "ADD"}', null, `by ${user.id}: ${format.name}`);
```

or ADDSUSPECTTEST, something of the sort; doing `SUSPECTTEST [action]` lets doing `/ml SUSPECTTEST` show all modlog entries related to it, so its more of a QoL matter ig

server/chat-plugins/suspect-tests.ts
```diff
@@ -0,0 +1,101 @@
...
+
+		delete: 'remove',
+		remove(target, room, user) {
+			this.checkCan('gdeclare');
+
+			const format = toID(target);
+			if (!suspectTests[format]) return this.errorReply(`There is no suspect test for '${target}'. Check spelling?`);
+
+			this.privateGlobalModAction(`${user.name} removed the ${suspectTests[format].tier} suspect test.`);
+			this.globalModlog('SUSPECTTEST', null, `by ${user.id}: removed ${suspectTests[format].tier}`);
```
KrisXV, line 82: ditto (and remove `removed` from the ML entry)

sim/battle.ts
```diff
@@ -13,6 +13,7 @@ import {Side} from './side';
 import {State} from './state';
 import {BattleQueue, Action} from './battle-queue';
 import {Utils} from '../lib/utils';
+import {suspectTests} from '../server/chat-plugins/suspect-tests';
```
mia-pi-git, line 16: Isn't sim/ supposed to be separate from the server? 
Couldn't you just require it or fs / JSON.parse it?

KrisXV: I don't see a reason to duplicate the work when it's easier to just export it

Zarel: Oh yes sim should be separate from server.

server/chat-plugins/suspect-tests.ts
TheImmortal: This is what it shows currently.

image

Seems fine? I think it's more readable than your suggestion.

Zarel:
Looks good. Did you want to handle the new suspect test system here? Maybe also add functionality for rules changes (like bans/unbans) in suspect tests?

TheImmortal:
> Looks good. Did you want to handle the new suspect test system here? Maybe also add functionality for rules changes (like bans/unbans) in suspect tests?

I did this for the notices mostly because [look at the amount of notices in our commit history](https://github.com/smogon/pokemon-showdown/commits/master/config). Expanding it beyond this can be done later on, if someone wants to. But I don't think there is an immediate need because majority of suspects don't modify the banlist.

Zarel:
Cool, I think this is pretty low-risk currently so I'm going to merge now. Feel free to push improvements directly.

PR #7429, opened 2020-09-27
Refactor staffAutojoin

Zarel:
`room.settings.autojoin` and `room.settings.staffAutojoin` are now merged.

Staff autojoin rooms are now simply regular autojoin rooms with a modjoin setting. You simply autojoin all autojoin modjoin rooms that you're allowed to join.

Fixes #7413

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -78,7 +78,6 @@ export interface RoomSettings {
 	auth: {[userid: string]: GroupSymbol};
 	creationTime: number;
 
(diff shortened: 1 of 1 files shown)


urkerab:
(This also fixes auth display in staff rooms in side servers, both for new side servers and for existing side servers which haven't already updated their staff rooms from secret to hidden rooms.)

Zarel:
Looks hotpatch-safe to me, so let's go!

PR #7430, opened 2020-09-27
Update German translation for /groups

xMarcell:
I added the german translation for the newly added Prize Winner rank.


Zarel:
tyty

PR #7432, opened 2020-09-28
Update Portuguese translation for /groups

xMarcell:
I added the portuguese translation for the newly added Prize Winner Rank; courtesy of Mitsuki :D


Zarel:
ty

PR #7433, opened 2020-09-28
ScavGames - add October twist

sparkychildcharlie:
New Twist: Bonus Round
(This also fixes a tiny oversight in a past twist as well.)

It would be purrrrfect if we can get this merged and patched in by the end of tomorrow :3


Zarel:
Looks good, thanks!

PR #7434, opened 2020-09-28
Introduce Chat.oldPlugins

Zarel:
This provides a new way to persist chat plugin data across hotpatches: `Chat.oldPlugins`.

In a plugin, you can now do:

```
export const cache = Chat.oldPlugins.pluginname?.cache || {};
```

to create a cache that will persist across hotpatches.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -330,6 +330,7 @@ export const commands: ChatCommands = {
 				}
 				if (requiresForce(patch)) return this.errorReply(requiresForceMessage);
 
+				const oldPlugins = Chat.plugins;
 				Chat.destroy();
 
 				const processManagers = ProcessManager.processManagers;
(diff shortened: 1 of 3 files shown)


monsanto:
Why not make the API something like `Chat.getPluginCache : (pluginName: string) => Record<string, any>`

mia-pi-git:
Seconded, `oldPlugins` is a bit hard to tell what it does. If not `getPluginCache`, then maybe something like `cachedPlugin` or `lastPluginCache`?

Zarel:
> Why not make the API something like `Chat.getPluginCache : (pluginName: string) => Record<string, any>`

This API is more general and supports schema migrations and getting whatever other information you need from older plugins.

For instance, this PR adds a change to `helptickets` that wouldn't be possible with your system.

Zarel:
In case this was unclear: This lets you get all data from the previous plugin, not just caches. It's just a way to access the `Chat.plugins` from before the hotpatch.

monsanto:
> This API is more general and supports schema migrations and getting whatever other information you need from the plugin before the hotpatch.

My objection is primarily to exposing the implementation detail that `oldPlugins` is an object, and that `oldPlugins.<pluginName>` can be missing. There's no reason chat plugins should break if you, say, wanted to use a `Map` as the backing storage for `oldPlugins` (which is probably what I would have done)

Zarel:
(We talked this out over Discord.)

Zarel:
While @monsanto doesn't fully agree with me, neither do his suggested alternatives actually satisfy the use-case this API was designed for, so there's not much I can do about that.

PR #7435, opened 2020-09-28
Modlog: Parse entries in the child process

AnnikaCodes:
This moves `parseModlog` into the child process so that it doesn't lag the main process. Also I fixed a crash I saw in Dev room.

### DIFF
--- a/server/modlog.ts
+++ b/server/modlog.ts
@@ -331,8 +331,7 @@ export class Modlog {
 			maxLines: maxLines,
 			onlyPunishments: onlyPunishments,
 		};
-		const rawResponse = await PM.query(query);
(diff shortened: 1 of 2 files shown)


server/modlog.ts
```diff
@@ -352,14 +351,22 @@ export class Modlog {
 	}
 }
 
-export const PM = new QueryProcessManager<ModlogQuery, string[] | undefined>(module, async data => {
+// if I don't do this TypeScript thinks that (ModlogResult | undefined)[] is a function
+// and complains about an "nexpected newline between function name and paren"
+// even though it's a type not a function...
```
Zarel, line 356: I've had this problem too! I'm in the process of upgrading eslint which should hopefully fix this.

Zarel:
tyty

PR #7436, opened 2020-09-28
Modlog viewer: Improve UI

mia-pi-git:
Some requests from gstaff: 
- Param detection assumes that the first argument is a roomid since that's how it used to work
- Moves the  `<- Current server time` indicator back up to the top

### DIFF
--- a/server/chat-plugins/modlog-viewer.ts
+++ b/server/chat-plugins/modlog-viewer.ts
@@ -539,7 +539,7 @@ export const commands: ChatCommands = {
 		let lines;
 		const search: ModlogSearch = {};
 		const targets = target.split(',');
-		for (const [i, option] of targets) {
+		for (const [i, option] of targets.entries()) {
 			let [param, value] = option.split('=').map(part => part.trim());
 			if (!value) {
 				// We should guess what parameter they meant
--- a/server/chat-plugins/modlog-viewer.ts
+++ b/server/chat-plugins/modlog-viewer.ts
@@ -110,7 +110,7 @@ function prettifyResults(
 	const title = `[${roomid}] ${searchCmd}`;
 	const lines = resultArray.length;
 	let curDate = '';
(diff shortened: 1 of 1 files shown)


server/chat-plugins/modlog-viewer.ts
```diff
@@ -538,13 +538,17 @@ export const commands: ChatCommands = {
...
-		for (const option of target.split(',')) {
+		const targets = target.split(',');
+		for (const option of targets) {
 			let [param, value] = option.split('=').map(part => part.trim());
 			if (!value) {
 				// We should guess what parameter they meant
 				value = param.trim();
 				if (/^[0-9]{1,3}\.[0-9]{1,3}/.test(value)) {
 					param = 'ip';
+				} else if (targets.indexOf(option) === 0) {
```
Zarel, line 549:
```suggestion
		for (const [i, option] of targets) {
			let [param, value] = option.split('=').map(part => part.trim());
			if (!value) {
				// We should guess what parameter they meant
				value = param.trim();
				if (/^[0-9]{1,3}\.[0-9]{1,3}/.test(value)) {
					param = 'ip';
				} else if (i === 0) {
```

Zarel: Please do not ever search an array you're iterating through; there's nearly always a better way to do it.

server/chat-plugins/modlog-viewer.ts
```diff
@@ -539,14 +539,14 @@ export const commands: ChatCommands = {
 		let lines;
 		const search: ModlogSearch = {};
 		const targets = target.split(',');
-		for (const option of targets) {
+		for (const [i, option] of targets) {
```
Zarel, line 542:
```suggestion
		for (const [i, option] of targets.entries()) {
```

WAIT SHIT

Zarel:
Thanks!

PR #7437, opened 2020-09-29
UNO: Fix timestamp and switch to Chat.ErrorMessage

KrisXV:


Zarel:
Looks good; thanks!

You can still define a function `timestamp()` if that would look better to you.

PR #7438, opened 2020-09-29
Chatlog: Handle ErrorMessage properly

mia-pi-git:
Doesn't, rn.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -657,12 +657,12 @@ export const PM = new QueryProcessManager<AnyObject, string | undefined>(module,
 			return LogViewer.error(`Config.chatlogreader is not configured.`);
 		}
 	} catch (e) {
-		Monitor.crashlog(e, 'A chatlog search query', data);
-		return (
-			`<div class="pad"><p class="message-error">` +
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -657,12 +657,12 @@ export const PM = new QueryProcessManager<AnyObject, string | undefined>(module,
...
-			`<div class="pad"><p class="message-error">` +
-			`Sorry! Your chatlog search crashed. We've been notified and will fix this.` +
-			`</p></div>`
-		);
+		if (e.name?.endsWith('ErrorMessage')) {
+			return LogViewer.error(e.message);
+		} else {
+			Monitor.crashlog(e, 'A chatlog search query', data);
+			return LogViewer.error(`Sorry! Your chatlog search crashed. We've been notified and will fix this.`);
+		}
```
Zarel, line 665:
```suggestion
		}
		Monitor.crashlog(e, 'A chatlog search query', data);
		return LogViewer.error(`Sorry! Your chatlog search crashed. We've been notified and will fix this.`);
```

No need to `else` after a `return`.

Zarel:
tyty

PR #7439, opened 2020-09-29
Add run on repl.it badge to README

tacogods:
This pull request adds a  badge to the . This will allow users to easily run this repository in their browser, without having to set up an environment. You can learn more about Repl.it [here](https://repl.it).

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "npm start"
--- a/README.md
+++ b/README.md
@@ -213,3 +213,4 @@ Contributors
 Special thanks
 
 - See http://pokemonshowdown.com/credits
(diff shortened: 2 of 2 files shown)


Zarel:
Are you associated with Repl.it? It seems like a weird choice to create a GitHub account just for this PR, instead of using an existing account for everything.

Anyway, your `npm start` thing doesn't seem to work. I tried starting on port 8000 (which didn't seem to make it accessible) and port 80 (which gave a permission error). We use `cloud-env` which Repl.it doesn't seem to use. Is there a specific thing we should be doing?

Zarel:
Since the run configuration doesn't work and the submitter doesn't seem to be responding, I'm going to close this.

Thanks for the interest, and repl.it does look cool! This isn't a rejection; feel free to reopen this at any time!

PR #7441, opened 2020-09-29
Update Portuguese translation wording

1Mitsuki:
After a brief discussion with woolf, we agreed our initial word choice for harassment was leaning too much towards sexual harassment, so I've changed it to something more broad that better encompasses the multiple instances the ticket is for.


Zarel:
tyty

PR #7442, opened 2020-09-30
Apply chatfilters to !code

AnnikaCodes:


server/chat-commands/info.ts
```diff
@@ -2610,13 +2610,15 @@ export const commands: ChatCommands = {
...
-
-		const code = Chat.getReadmoreCodeBlock(target);
 		this.runBroadcast(true);
+
 		if (this.broadcasting) {
-			return `/raw <div class="infobox">${code}</div>`;
+			const filteredCode = Chat.filter(this, target, user, room, connection, this.pmTarget);
+			if (!filteredCode) return;
+			target = filteredCode;
+			return `/raw <div class="infobox">${Chat.getReadmoreCodeBlock(target)}</div>`;
```
mia-pi-git, line 2619:
```suggestion
			target = this.filter(target, this.pmTarget);
			if (!target) return this.errorReply(`Invalid code.`);
			return `/raw <div class="infobox">${Chat.getReadmoreCodeBlock(target)}</div>`;
```

mia-pi-git: Less sure on the error, but let's not silently fail.

AnnikaCodes:
```ts
filter(message: string, targetUser: User | null = null) {
	if (!this.room) return null;
	return Chat.filter(this, message, this.user, this.room, this.connection, targetUser);
}
```
This will always return null in PMs.

Zarel:
I'm afraid chatfilters might be overzealous. Is this a problem in practice?

mia-pi-git:
> I'm afraid chatfilters might be overzealous. Is this a problem in practice?

I mean currently you can say anything, and get away with it with little to no moderation - that seems like a problem?

Zarel:
It requires roomvoice, though, and can be hidetexted.

Zarel:
We might want to filter them in certain cases like idk battles and groupchats? But I don't want filters causing problems in actual code or something.

Zarel:
Especially in staff rooms where you might want to discuss filter evasion, or filter code.

mia-pi-git:
> It requires roomvoice, though, and can be hidetexted.

The main concern i can think of is PMs, where anyone can use it. 

> Especially in staff rooms where you might want to discuss filter evasion, or filter code.

Don't the relevant monitors already not apply in staff rooms?

AnnikaCodes:
Hmmm. They do apply in staff rooms, but won't punish the user or block messages. We definitely want to filter them in PMs, for reasons I'll explain elsewhere. Battles and groupchats might make sense, too.

server/chat-commands/info.ts
```diff
@@ -2610,13 +2610,16 @@ export const commands: ChatCommands = {
...
-		const code = Chat.getReadmoreCodeBlock(target);
 		this.runBroadcast(true);
-		if (this.broadcasting) {
-			return `/raw <div class="infobox">${code}</div>`;
+
+		const isPMOrPersonalRoom = this.room?.settings.isPersonal !== false;
+
+		if (this.broadcasting && isPMOrPersonalRoom) {
+			target = Chat.filter(this, target, user, room, connection, this.pmTarget)!;
+			if (!target) return this.errorReply(`Invalid code.`);
```
Zarel, line 2619: Am I misunderstanding this code? I think you meant:

```suggestion
		if (this.broadcasting) {
			if (isPMOrPersonalRoom) {
				target = Chat.filter(this, target, user, room, connection, this.pmTarget)!;
				if (!target) return this.errorReply(`Invalid code.`);
			}
```

AnnikaCodes: yea i wasnt thinking straight

server/chat-commands/info.ts
```diff
@@ -2610,13 +2610,18 @@ export const commands: ChatCommands = {
...
-
-		const code = Chat.getReadmoreCodeBlock(target);
 		this.runBroadcast(true);
+
+		const isPMOrPersonalRoom = this.room?.settings.isPersonal !== false;
+
 		if (this.broadcasting) {
-			return `/raw <div class="infobox">${code}</div>`;
+			if (isPMOrPersonalRoom) {
+				target = Chat.filter(this, target, user, room, connection, this.pmTarget)!;
```
Zarel, line 2619: tbh do we still need `connection` if we refactor `Chat.filter` to use `Chat.ErrorMessage`?

Zarel: I think if we move this `if`-block directly under `isPMOrPersonalRoom`, no other changes are necessary.

Zarel: On second thought, I think this is better.

Zarel:
I think `Chat.filter` should probably only take `this` and `target`. The other params seem kinda redundant because they can just be grabbed from the context, right?

server/chat-commands/info.ts
AnnikaCodes: > tbh do we still need connection if we refactor Chat.filter to use Chat.ErrorMessage?

Unless we're doing something weird in `config.js`, I don't think we even need `Chat.ErrorMessage`; no chatfilters in public or private code use `connection`.

Zarel: Yeah, that's why I changed tune to just saying `this` (`CommandContext`) is enough; you can pull anything you want out of the context if you actually need it.

PR #7443, opened 2020-09-30
Modlog: Support exact user searches

AnnikaCodes:
Requires hotpatching modlog and chat

### DIFF
--- a/server/modlog.ts
+++ b/server/modlog.ts
@@ -57,7 +57,7 @@ interface ModlogQuery {
 
 export interface ModlogSearch {
 	note?: {searches: string[], isExact?: boolean};
-	user?: string;
+	user?: {search: string, isExact?: boolean};
 	ip?: string;
 	action?: string;
 	actionTaker?: string;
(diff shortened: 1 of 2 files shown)


server/modlog.ts
```diff
@@ -57,7 +57,7 @@ interface ModlogQuery {
 
 export interface ModlogSearch {
```
Zarel, line 58: LINQ when ;_;

Zarel: Honestly, though, having a data structure representing a query might be nice, instead of needing to implement query parsing repeatedly.

Zarel:
tyty

server/modlog.ts
AnnikaCodes:
```ts

interface ModlogQuery<T> {
	statement: Database.Statement<T>;
	args: T[];
	returnsResults?: boolean;
}
```
I'm doing this in the SQLite PR.

Zarel: Unfortunately not a very generalizable data structure :(

It would be nice to have something generic enough that all the searchers including dexsearch could share code.

AnnikaCodes: ~~just move the Pokémon info to a database~~

It seems like that would be a bit much generalization? I suppose something like key:value pairs where each value is an array of possible search terms one of which for each key must be matched could work; but I'm not sure how helpful that'd be.

Zarel: Moving it to a database would be even worse! The only way to search a database is by making SQL queries, which are annoying to make safely in basically every programming language. You can't just expose the query directly because that leads to SQL injection.

So no matter what, you need to translate user input into a data structure representing a query, and then translate that data structure into the actual SQL code. Either way, you need a generalized data structure for that. Here's how I'd do it:

A search is an array of filters. A filter is one of:

- key = value
- key contains value
- key is one of a list of values
- key > value
- key < value
- key >= value
- key <= value
- the inverse of a filter above

This seems to apply to most search types. It's not nearly as flexible as SQL (the big omission is OR searches), but it should cover most real-world use-cases.

AnnikaCodes: Moving it to a database wasn't a serious suggestion.

A filter array could work though.

PR #7444, opened 2020-09-30
Modlog viewer: Update regex for new modlog syntax

AnnikaCodes:
It looks like #7436 didn't do this which is causing problems when trying to view older results.


Zarel:
tyty

PR #7403, opened 2020-09-21
Modlog: Refactor to use ModlogEntry objects

AnnikaCodes:
This new format will let us migrate to SQLite more easily.

PR #7446, opened 2020-09-30
Modlog: Fix entries with an IP address and no user

AnnikaCodes:
I swear I handled this in #7403, but I guess not. Anyway, IP addresses without a username are a thing, and I also fixed some formatting issues with locks.


Zarel:
Looks good, thanks!

PR #7449, opened 2020-10-01
Modlog: Default to searching all fields 

AnnikaCodes:
As per [this suggestion](https://www.smogon.com/forums/threads/regarding-the-recent-modlog-update.3671232/). Requires hotpatching both modlog and chat.


Zarel:
tyty

PR #7452, opened 2020-10-01
Add properties to see which Pokemon fainted last/this turn

KrisXV:


sim/battle.ts
```diff
@@ -1509,7 +1509,9 @@ export class Battle {
 			trappedBySide.push(sideTrapped);
 			stalenessBySide.push(sideStaleness);
 			side.faintedLastTurn = side.faintedThisTurn;
+			side.pokemonFaintedLastTurn = side.pokemonFaintedThisTurn;
 			side.faintedThisTurn = false;
+			side.pokemonFaintedThisTurn = null;
```
Zarel, line 1514: Do we need all of these? Can `faintedThisTurn` and `pokemonFaintedThisTurn` be merged into one property?

Zarel:
Looks good, thanks!

PR #7454, opened 2020-10-02
PU: update banlist

xInstruct:
https://www.smogon.com/forums/threads/sword-and-shield-tiering-announcements.3657492/page-4#post-8614913


Zarel:
@KrisXV?

PR #7456, opened 2020-10-02
Scavengers: Improve /edithunt

aQrator:
Two minor changes for the Scavengers plugin:

- In /viewhunt, change the ``/`` for alts to a ``;`` to allow easier copying.
- For /edithunt, allow commas in answers.

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -1592,6 +1592,7 @@ const ScavengerCommands: ChatCommands = {
 		) {
 			return this.errorReply("You cannot edit the hints and answers if you are not the host.");
 		}
+
 		const [question, type, ...value] = target.split(',');
 		if (!game.onEditQuestion(parseInt(question), toID(type), value.join(',').trim())) {
 			return this.sendReply("/scavengers edithunt [question number], [hint | answer], [value] - edits the current scavenger hunt.");
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -551,7 +551,6 @@ export class ScavengerHunt extends Rooms.RoomGame {
 
 		let answer: string[] = [];
 		if (question_answer === 'answer') {
-			if (value.includes(',')) return false;
 			answer = value.split(';').map(p => p.trim());
 		}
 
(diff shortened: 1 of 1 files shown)


server/chat-plugins/scavengers.ts
```diff
@@ -1593,9 +1593,11 @@ const ScavengerCommands: ChatCommands = {
...
 			return this.errorReply("You cannot edit the hints and answers if you are not the host.");
 		}
-
-		const [question, type, ...value] = target.split(',');
-		if (!game.onEditQuestion(parseInt(question), toID(type), value.join(',').trim())) {
+		const parts = target.split(',');
+		const question = parseInt(parts[0]);
+		const type = parts[1];
+		const value = parts.slice(2).join(',').trim();
+		if (!game.onEditQuestion(parseInt(question), toID(type), value)) {
```
mia-pi-git, line 1600:
```suggestion
		const [question, type, value] = target.split(',');
		if (!game.onEditQuestion(question, toID(type), value.trim())) {
```

aQrator:
I will redo this better. My apologies

mia-pi-git:
You don't need to close it, it's just that it already should do what you're trying to do - that's what the `...value` is for.

aQrator:
You're right in that it does what it should do there. I need to make the edit elsewhere.

server/chat-plugins/scavengers.ts
aQrator: Yeah, I'll change this back

server/chat-plugins/scavengers.ts
```diff
@@ -1593,7 +1592,6 @@ const ScavengerCommands: ChatCommands = {
 		) {
 			return this.errorReply("You cannot edit the hints and answers if you are not the host.");
 		}
-
```
AnnikaCodes, line 1596: Why are you removing a line of whitespace?

AnnikaCodes:
Have the Scavengers Room Owners approved this? It's definitely something that room staff need to be aware of.

server/chat-plugins/scavengers.ts
aQrator: There was initially a change there that I have now removed. The whitespace was a residual effect of that change. If preferred, it can be put back

aQrator:
ROs have been informed. Staff is aware that I was working on this, and will be informed if the change goes through. (I am a moderator in scavengers)

sparkychildcharlie:
+1

server/chat-plugins/scavengers.ts
```diff
@@ -1593,7 +1592,6 @@ const ScavengerCommands: ChatCommands = {
...
 		const [question, type, ...value] = target.split(',');
```
Zarel, line 1595:
```suggestion

				const [question, type, ...value] = target.split(',');
```

I think I'd prefer it back. I personally think whitespace after a `return` is more readable.

Zarel:
Looks like a pretty simple change. If everyone who cares agrees, we can merge this.

aQrator:
Here's the approval by RO pants. If anything else is required, do let me know
afbeelding

Zarel:
Thanks!

PR #7458, opened 2020-10-02
Update German translation for /roomhelp

xMarcell:
I updated the german translation for the /roomhelp command, by adding the /show explanation, while also fixing a problem with a line not showing as translated.


Zarel:
tyty

PR #7459, opened 2020-10-02
Chatlog: Update permission checks

mia-pi-git:
Requested by Aeonic.


AeonicX7:
I did indeed request this - /roomlog already lets gstaff see private room logs so this makes sense to do.

Zarel:
Yeah, it does look like this better matches `/roomlog`. Thanks!

PR #7460, opened 2020-10-02
Helptickets: Display shared battles in staffintro

AnnikaCodes:
[Requested by global staff](https://www.smogon.com/forums/threads/improving-help-tickets.3642984/page-6#post-8607392).

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -12,6 +12,20 @@ import * as net from 'net';
 import {YoutubeInterface} from '../chat-plugins/youtube';
 import {Utils} from '../../lib/utils';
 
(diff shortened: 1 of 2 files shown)


Zarel:
Looks good!

Zarel:
Importing from chat plugins may cause problems further down the line, but we can deal with them then, I think.

AnnikaCodes:
Yeah, I don't usually like to import from *plugins* (my philosophy is that you should ideally be able to remove anything in the `chat-plugins/` directory and have PS work fine), but `info.ts` is in `chat-commands/` so I'm less wary of that. It'd be easy enough to move to `Rooms` but that would lose hotpatchability :x.

Zarel:
I think it's easily monkeypatched, so if that's your only concern it's not a big deal.

PR #7461, opened 2020-10-02
Use attributes of CommandContext in Chat.filter

AnnikaCodes:
This does not change the parameters passed to individual filters, in the name of backwards compatibility with old filters.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2616,7 +2616,7 @@ export const commands: ChatCommands = {
 
 		if (this.broadcasting) {
 			if (isPMOrPersonalRoom) {
-				target = Chat.filter(this, target, user, room, connection, this.pmTarget)!;
+				target = Chat.filter(this, target)!;
 				if (!target) return this.errorReply(`Invalid code.`);
 			}
 			return `/raw <div class="infobox">${Chat.getReadmoreCodeBlock(target)}</div>`;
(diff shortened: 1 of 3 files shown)


server/chat.ts
```diff
@@ -1376,21 +1376,22 @@ export const Chat = new class {
...
 	readonly filters: ChatFilter[] = [];
-	filter(
-		context: CommandContext,
-		message: string,
-		user: User,
-		room: Room | null,
-		connection: Connection,
-		targetUser: User | null = null
-	) {
+	filter(context: CommandContext, message: string) {
```
Zarel, line 1379: I think I'd prefer this the other way around: `message, context`.

server/chat-commands/info.ts
```diff
@@ -2616,7 +2616,7 @@ export const commands: ChatCommands = {
 
 		if (this.broadcasting) {
 			if (isPMOrPersonalRoom) {
-				target = Chat.filter(this, target, user, room, connection, this.pmTarget)!;
+				target = Chat.filter(this, target)!;
```
Zarel, line 2619: I saw a `context.filter` function defined. Can it be used in these places? Just `target = this.filter(target)!;`?

server/chat.ts
AnnikaCodes: Okay.

server/chat-commands/info.ts
AnnikaCodes: Yeah, it can.

Zarel:
Thanks! This looks good!

PR #7464, opened 2020-10-03
/showteam: Support nicknames

mia-pi-git:


Zarel:
It looks like there was already a nicknames parameter? Was it just not working in the past?

mia-pi-git:
> It looks like there was already a nicknames parameter? Was it just not working in the past?

Replied on discord, but i figured it'd be smart to reply here too - it just wasn't passed in the past, see the `undefined` param.

PR #7468, opened 2020-10-04
Battlesearch: Make turn limit optional

mia-pi-git:
Per #sim-dev.

### DIFF
--- a/server/chat-plugins/modlog-viewer.ts
+++ b/server/chat-plugins/modlog-viewer.ts
@@ -441,7 +441,7 @@ async function getBattleSearch(
 	const user = connection.user;
 	if (!user.can('forcewin')) return connection.popup(`/battlesearch - Access Denied`);
 
-	const response = await PM.query({userids, turnLimit: turnLimit, month, tierid});
+	const response = await PM.query({userids, turnLimit, month, tierid});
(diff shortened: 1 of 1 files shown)


server/chat-plugins/modlog-viewer.ts
```diff
@@ -434,29 +434,34 @@ function buildResults(
...
+	connection: Connection, userids: string[], month: string,
+	tierid: ID, turnLimit?: number
 ) {
 	userids = userids.map(toID);
 	const user = connection.user;
 	if (!user.can('forcewin')) return connection.popup(`/battlesearch - Access Denied`);
 
-	const response = await PM.query({userids, turnLimit, month, tierid});
-	connection.send(buildResults(response, userids as ID[], turnLimit, month, tierid));
+	const response = await PM.query({userids, turnLimit: turnLimit, month, tierid});
```
Zarel, line 444: Was this a typo?

```suggestion
	const response = await PM.query({userids, turnLimit, month, tierid});
```

Is something wrong with our CI? I swear this is something eslint checks...

Zarel:
Looks good after that formatting thing. Thanks!

server/chat-plugins/modlog-viewer.ts
mia-pi-git: Was a typo, yeah.

PR #7469, opened 2020-10-04
Fix tour autostart with less than 2 players

DaWoblefet:
This was working and broke sometime in the past month or so? Anyway, this is fixes it so tournaments automatically forceend if there's not enough users when the autostart timer reaches 0.


Zarel:
It feels like this should be done a different way.

Zarel:
Thanks for the contribution! Yours does also fix it, I just fixed it a different way that should make it harder for future updates to accidentally break.

DaWoblefet:
That works well for me! Thanks for improving it.

PR #7470, opened 2020-10-04
Add more letter substitutions to filter

asgdf:
I'm not sure if I should be adding upper/lowercase versions separately.


AnnikaCodes:
PS' word filters are case insensitive regular expressions (`/i` flag), so you should only need to add the lowercase variants. However, I'm not sure if all special characters are properly handled by Node's regex functionality; it's possible that some slip through (we could test this if there's some that you're curious about).

PR #7472, opened 2020-10-04
Add skipped failing event validator tests

DaWoblefet:
Adds 3 failing validator tests related to events:
1. Evolutions of Shiny-locked event Pokemon should not be able to be Shiny (e.g. Urshifu, Cosmoem)
2. Unreleased G-Max Pokemon should not be allowed (Melmetal)
3. Events should not have access to past-generation moves (e.g. Shiny Zeraora is Gen 8-only, so it can't have Knock Off since it only gets that in Gen 7)

### DIFF
--- a/test/sim/team-validator.js
+++ b/test/sim/team-validator.js
@@ -528,6 +528,32 @@ describe('Team Validator', function () {
 		assert(illegal);
 	});
 
+	it.skip('should not allow evolutions of Shiny-locked events to be Shiny', function () {
+		const team = [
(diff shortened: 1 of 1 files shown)


TheImmortal:
Tagging @KrisXV to fix 2 because #7001 broke it.

Zarel:
Should I leave this open until 2 is fixed?

DaWoblefet:
Nah, it's skipped tests, they can be merged with no consequence. Kris or whomever can address them whenever you feel like it.

Zarel:
Sure.

PR #7474, opened 2020-10-05
Chatlog: Notify the user that a search is running

mia-pi-git:
Minor thing, but since battlesearch has this behavior and chatlog searches are slow, this is useful to have.
Also adds some aliases to /searchlogs.


Zarel:
Thanks!

PR #7475, opened 2020-10-05
Make !groups case-insensitive

PartMan7:


server/chat-commands/info.ts
```diff
@@ -1507,6 +1507,7 @@ export const commands: ChatCommands = {
 
 	groups(target, room, user) {
 		if (!this.runBroadcast()) return;
+		target = target.toLowerCase();
```
AnnikaCodes, line 1510:
```suggestion
		target = toID(target);
```

AnnikaCodes: Everything we compare `target` to is effectively an `ID`.

AnnikaCodes (approved):
Good catch!

PR #7476, opened 2020-10-06
Chat#checkHTML: Ban <iframe> tags

DaWoblefet:
iframes don't work in /htmlbox etc. right now anyway and I don't think there's a good reason to allow it.

Screenshot of current behavior:
image


Zarel:
Yeah, that sounds like a useful change, thanks!

Zarel:
To be clear, this list was never supposed to be a security thing (Caja already filters out all of these); it was always a sort of "this tag doesn't do anything, so whatever you're trying to do, it won't work". Feel free to add any other tags along those lines!

PR #7477, opened 2020-10-06
Support persisting polls and announcements

mia-pi-git:
Not sure if queue should be persisted? Went with yes to be safe.


server/chat-plugins/announcements.ts
```diff
@@ -42,7 +53,47 @@ export class Announcement {
...
+		clearTimeout(this.timeout);
+		this.timeoutMins = 0;
+		delete this.timerEnd;
+		return this;
+	}
+}
+
+for (const room of Rooms.rooms.values()) { // hotpatching!
+	if (room.settings.minorActivity?.activityId === 'announcement') {
+		room.minorActivity = new Announcement(room, room.settings.minorActivity);
```
Zarel, line 96: The saving seems entirely unnecessary? Why not just

```suggestion
	if (room.minorActivity?.activityId === 'announcement') {
		room.minorActivity = new Announcement(room, room.minorActivity.toJSON());
```

Zarel: (The same approach applies to hotpatching helptickets.)

Zarel: Oh, you're also trying to persist through restarts, aren't you? In which case this should stay as it is.

server/chat-plugins/announcements.ts
```diff
@@ -4,20 +4,31 @@
  */
 import {Utils} from '../../lib/utils';
 
+const MINUTE = 60000;
+
+export interface AnnouncementData {
+	activityId?: 'announcement';
```
Zarel, line 10:
```suggestion
	activityId: 'announcement';
```

Shouldn't this be required?

server/chat-plugins/announcements.ts
Zarel: Do you clear the old timer on hotpatch? It doesn't seem like you do.

server/chat-plugins/announcements.ts
mia-pi-git: I can make it required, though it means adding it when /poll create is used. But i suppose that's not a big deal.

server/chat-plugins/announcements.ts
mia-pi-git: It does now.

server/chat-plugins/announcements.ts
Zarel: Oh, you deal with that by having them be different types, not by adding it when `/poll create` is used.

Zarel: `AnnouncementData extends AnnouncementOptions`

server/chat-plugins/announcements.ts
```diff
@@ -42,7 +53,48 @@ export class Announcement {
...
+		clearTimeout(this.timeout);
+		this.timeoutMins = 0;
+		delete this.timerEnd;
+		return this;
+	}
+}
+
+for (const room of Rooms.rooms.values()) { // hotpatching!
+	if (room.settings.minorActivity?.activityId === 'announcement') {
+		if (room.minorActivity?.timeout) clearTimeout(room.minorActivity.timeout);
```
Zarel, line 96:
```suggestion
		room.minorActivity?.destroy();
```

Should probably be an interface looking more like this tbh.

server/chat-plugins/announcements.ts
```diff
@@ -42,7 +53,48 @@ export class Announcement {
...
+	endTimer() {
+		if (!this.timeout) return;
+		clearTimeout(this.timeout);
+		this.timeoutMins = 0;
+		delete this.timerEnd;
+		return this;
+	}
+}
+
+for (const room of Rooms.rooms.values()) { // hotpatching!
```
Zarel, line 94:
```suggestion
// should handle restarts and also hotpatches
for (const room of Rooms.rooms.values()) {
```

server/chat-plugins/announcements.ts
```diff
@@ -70,8 +122,9 @@ export const commands: ChatCommands = {
 
 			const source = supportHTML ? this.checkHTML(target) : Utils.escapeHTML(target);
 
-			room.minorActivity = new Announcement(room, source);
+			room.minorActivity = new Announcement(room, {source, activityId: 'announcement'});
```
Zarel, line 125: `activityId` shouldn't be specified here; like I said, they should be two different types that extend from each other.

server/chat-plugins/announcements.ts
mia-pi-git: Oh, okay.

server/chat-plugins/announcements.ts
mia-pi-git: I don't think this needs to be a `destroy` function when you can just clear the timer, since that's the main issue here. 
`end` handles the other stuff that destroy might.

mia-pi-git: Using `endTimer` here would be better, though.

Zarel: A `destroy` function is better futureproofing. You don't know exactly what a `minorActivity` needs to do to clean up, you're just generically telling it to clean up so it can be replaced.

server/chat-plugins/poll.ts
```diff
@@ -259,9 +286,98 @@ export class Poll {
...
+	if (room.settings.minorActivity?.activityId === 'poll') {
+		room.minorActivity?.endTimer();
+		room.minorActivity = new Poll(room, room.settings.minorActivity);
+	}
+	if (room.settings.minorActivityQueue) { // rebuild queue
+		room.minorActivityQueue = [];
+		for (const queued of room.settings.minorActivityQueue) {
+			if (queued.activityId !== 'poll') continue;
+			room.minorActivityQueue.push(new Poll(room, queued));
+		}
```
Zarel, line 380: Wait you can't put a `new Poll` into `room.settings`. `room.settings` needs to be JSON.

I don't see any benefit to this, anyway. You can and should just put a `AnnouncementData | PollData` into the `minorActivityQueue`; it'd be better for everyone, right?

mia-pi-git: This isn't putting the Poll into room.settings, this is putting the JSON into the settings.

mia-pi-git: This takes the JSON from the queue and rebuilds every poll in the queue.

mia-pi-git: But, sure, i can put the JSON into the queue. That does make sense.

mia-pi-git: Done.

Zarel: Oh, you're right, I misread. But yes, if you put JSON in the queue instead, you don't need to rebuild the queue every hotpatch/restart.

server/chat-plugins/poll.ts
```diff
@@ -572,12 +662,13 @@ export const pages: PageTable = {
 				`(${this.tr`delete`})</button>`
 			);
 			buf += `<hr />`;
-			buf += `${button}<br />${poll.generateResults()}`;
+			buf += `${button}<br />${new Poll(room, poll).generateResults()}`;
```
Zarel, line 665: Errrr... this seems like it would cause bugs.

You should maybe do a `static generateResults` that can generate them from a `PollData`?

Zarel: For convenience, maybe also let `Poll implements PollData`?

Zarel: This is getting kind of involved, which I'm sorry about. I unfortunately don't think this sort of thing is mergeable, but if you're getting overwhelmed, just say so and I can work on it myself.

mia-pi-git: No worries.

mia-pi-git:
@Zarel bump?

mia-pi-git:
@Zarel bump

server/chat-plugins/announcements.ts
```diff
@@ -40,7 +55,52 @@ export class Announcement {
...
+			timerEnd: this.timerEnd,
+			activityId: 'announcement',
+		};
+		this.room.settings.minorActivity = entry;
+		this.room.saveSettings();
+		return entry;
+	}
+	runTimeout(timeout: number) {
+		this.timeoutMins = timeout;
+		this.timerEnd = Date.now() + timeout * 60000;
```
DaWoblefet, line 78: Probably should be using the MINUTE instead of 60000, applies elsewhere too.

server/chat-commands/admin.ts
```diff
@@ -17,6 +17,24 @@ import {Utils} from '../../lib/utils';
 
 import * as ProcessManager from '../../lib/process-manager';
 
+<<<<<<< HEAD
+=======
```
KrisXV, line 21: .

Zarel (approved):
It's mergeable in this state.

Zarel:
Thanks!

PR #7478, opened 2020-10-06
Fix /help for room promotion commands

AnnikaCodes:
I'm not sure if a regex is the right way to do this but it seems more compact than listing every possible variation, and I figure that we care more about readability than performance here.


Zarel:
Looks good, thanks!

Zarel:
One day we should probably use the same methods of getting commands so this stuff isn't duplicated, though.

Zarel:
By which I mean, we should probably use the same code for parsing `/roomdriver`, `/help roomdriver`, and `/permissions /roomdriver`.

It's my understanding that currently there's not as much code shared as there could be.

PR #7481, opened 2020-10-06
Battlesearch: Move to its own file

mia-pi-git:


server/chat-plugins/battlesearch.ts
```diff
@@ -0,0 +1,412 @@
...
+ */
+import * as child_process from 'child_process';
+import * as util from 'util';
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {QueryProcessManager} from '../../lib/process-manager';
+import {Repl} from '../../lib/repl';
+import {Dex} from '../../sim/dex';
+import {Config} from '../config-loader';
```
Zarel, line 12: Yeesh, I didn't know this was done this way. I wasn't aware Config was designed to be imported this way...

Zarel: I think we should import the old way for `Dex` and `Config`. That's more consistent to the rest of the codebase and allows them to be hotpatched with their corresponding command.

server/chat-plugins/battlesearch.ts
```diff
@@ -0,0 +1,412 @@
...
+import {checkRipgrepAvailability} from '../modlog';
```
Zarel, line 13: I guess this is why they were the same file in the past.

I wonder if there's some global we can stick this in. Like `Chat` or something.

mia-pi-git: tbh we could probably just use the existing `Config` setting that it generates instead of the function, no?

Zarel: Oh yeah, `config-loader` already exists; we can probably stick it there?

server/chat-plugins/battlesearch.ts
mia-pi-git: What do you mean, "the old way"?

Zarel: Like we do in `room-battle` etc:

https://github.com/smogon/pokemon-showdown/blob/52fb644d9eb57acbeaaaa2b60cf670cec21c6e5f/server/room-battle.ts#L1267-L1271

server/chat-plugins/battlesearch.ts
mia-pi-git: Done.

Zarel:
(Test failures – just look at how I fixed them in #7537)

mia-pi-git:
> (Test failures – just look at how I fixed them in #7537)

Resolved.

mia-pi-git:
@Zarel bump?

server/chat-plugins/battlesearch.ts
```diff
@@ -0,0 +1,410 @@
...
+ * Battle search - handles searching battle logs.
+ */
+import * as child_process from 'child_process';
+import * as util from 'util';
+
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {QueryProcessManager} from '../../lib/process-manager';
+import {Repl} from '../../lib/repl';
+import {checkRipgrepAvailability} from '../config-loader';
```
Zarel, line 11: This is not what I meant by "put it in config loader" but I guess it's fine for now; no need to stall this PR any longer.

Zarel:
Thanks!

server/config-loader.ts
```diff
@@ -7,6 +7,10 @@
 
 import * as defaults from '../config/config-example';
 import type {GroupInfo, EffectiveGroupSymbol} from './user-groups';
+import * as child_process from 'child_process';
+import {normalize as normalizePath} from 'path';
```
Zarel, line 11: This is also not an import PS should actually do.

Our standard is:

```js
import * as path from 'path';

path.normalize()
```

PR #7483, opened 2020-10-06
Users: Set AC status after checking punishments

AnnikaCodes:
This seems like something that should be implemented based on odd behavior with namelocked users being able to get autoconfirmed status. If not then feel free to close it.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -745,14 +745,7 @@ export class User extends Chat.MessageContext {
 		if (userType !== '1') {
 			registered = true;
 
-			if (userType === '3') {
-				this.isSysop = true;
-				this.isStaff = true;
-				this.trusted = userid;
-				this.autoconfirmed = userid;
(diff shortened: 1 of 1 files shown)


Zarel:
I'm confused. `loginfilter` expects `user.autoconfirmed` to already be set so it can be used for filtering.

What's the point of this?

Zarel:
Does this change namelocked users getting ac? We could hardcode a change for that, but is that a problem in practice?

AnnikaCodes:
Oh, I forgot that loginfilter needs it. I suppose it's not that big of a deal in practice. It came up in staff room and it seemed like unwanted behavior to me, but if you're namelocked it doesn't really matter if you're autoconfirmed.

Zarel:
Hm. I'm closing, but feel free to reopen if you still want to make a new change that accounts for loginfilter.

PR #7484, opened 2020-10-07
Add Neutralizing Gas switchout message

DaWoblefet:
Client support was already there. It feels a bit weird to do -fieldend, but trying to do a normal -end doesn't work and makes the client spit out mean red errors :(. There's no other ability implemented like this right now (Desolate Land and co. should be but they aren't). I also cleaned up a couple comments (entrance abilities are working correctly, another typo).

_Technically_ speaking the Pokemon should switch out, then the "effects of ngas left" message should play, then all the entrance abilities should resolve, then the switch completes, but I don't intend to actually fix the mechanic right now, just taking the low-hanging fruit of adding the message.

image


data/abilities.ts
```diff
@@ -2192,16 +2192,17 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
 			pokemon.abilityData.ending = false;
 		},
 		onEnd(source) {
+			this.add('-fieldend', 'ability: Neutralizing Gas');
+
 			// FIXME this happens before the pokemon switches out, should be the opposite order.
 			// Not an easy fix since we cant use a supported event. Would need some kind of special event that
 			// gathers events to run after the switch and then runs them when the ability is no longer accessible.
-			// (If your tackling this, do note extreme weathers have the same issue)
+			// (If you're tackling this, do note extreme weathers have the same issue)
```
Zarel, line 2205: hi HoeenHero ;)

Zarel:
Hm, after taking a look, I believe `-end` is the correct message.

What error do you get? I should be able to fix it.

DaWoblefet:
TypeError: Cannot read property 'pokemon' of undefined
at Battle.getPokemon (http://play.pokemonshowdown.com/js/battle.js?3556fc3f:3037:6)
at Battle.runMinor (http://play.pokemonshowdown.com/js/battle.js?3556fc3f:2466:18)

image

Zarel:
Do you have a link to that replay?

DaWoblefet:
No, I was testing on localhost

I assume it's because the Weezing left the field. Of note, the Ability shouldn't actually announce anyway - it just plays the message on cart.

DaWoblefet:
oh nvm I'm dumb, just take the inputlog:

>version 3736aaf709a50ae6a2cd8dcf6acdc5fdefcd3a5d
>version-origin 3e0a47c4b33769d1f5ab3bc35895044e00936288
>start {"formatid":"gen8customgame","seed":[59826,53354,60770,20973]}
>player p1 {"name":"jfksldjlskaj","avatar":"102","team":"Lunala||choicescarf|shadowshield|moongeistbeam,teleport|Timid|4,,,252,,252||,0,,,,||50|]Marshadow||choiceband|technician|spectralthief,shadowsneak|Jolly|4,252,,,,252||||50|]Dragapult||choicespecs|infiltrator|shadowball,dracometeor|Timid|4,,,252,,252||,0,,,,||50|]Necrozma|necrozmadawnwings|kasibberry|prismarmor|moongeistbeam,calmmind,heatwave,protect|Modest|244,,,252,,12||,0,,,,||50|]Mimikyu||lifeorb|disguise|shadowclaw,shadowsneak,swordsdance,protect|Adamant|4,252,,,,252||||50|]Drifblim||focussash|unburden|shadowball,tailwind,destinybond,protect|Modest|4,,,252,,252||,0,,,,||50|","rating":0}
>player p2 {"name":"DaWoblefet","avatar":"pokefan-gen4","team":"Dubwool|||fluffy|bodypress,cottonguard,rest,sleeptalk||252,252,252,252,252,252||,0,,,,|||]Weezing|weezinggalar||neutralizinggas|aromatherapy||56,,,,,||,0,,,,|||","rating":0}
>p1 team 6, 2, 3, 4, 5, 1
>p2 team 2, 1
>p1 move protect
>p2 switch 2

Zarel:
I would have also taken a replay file (you can still download replays from localhost), but the inputlog works too.

DaWoblefet:
bump @Zarel to look into the crash on client if we want to do this PR the correct way

DaWoblefet:
OK so the reason I had this wrong 3 months ago was because I didn't add the actual Pokemon in this.add(). Now that this is fixed, should be an easy merge for whoever wants to do so.

PR #7486, opened 2020-10-07
Remove the broadcast cooldown for !showteam

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/allow-showteam-to-be-broadcasted-with-no-cooldown.3671496/


Zarel:
Thanks! This is a good next step.

But I wonder if we should find some way to prevent the same user from spamming it.

Distrib-ps:
I'll try to watch this. I don't have a lot of knowledge yet, I'm just starting out but I'm looking into the subject.

Zarel:
Yeah it's something I can do myself; you don't need to worry about it.

PR #7487, opened 2020-10-07
Translate Repeat phrases and Weekban to German

xMarcell:
I translated and added the weekban line for /punishments and repeat phrases.


Zarel:
Thanks!

PR #7488, opened 2020-10-08
Repeats: Refactor storing and deleting repeats, let people repeat room FAQs

KrisXV:


Zarel:
Looks good to me, but I'll give Annika a dayish to give her own opinion.

Zarel (approved):

server/chat-plugins/repeats.ts
```diff
@@ -75,7 +78,9 @@ export const Repeats = new class {
 					this.clearRepeats(room);
 					return;
 				}
-				room.add(`|html|<div class="infobox">${repeat.phrase}</div>`);
+				const phrase = !repeat.faq ? repeat.phrase :
+					Chat.formatText(roomFaqs[room.roomid][repeat.faq], true).replace(/\n/g, '<br />');
```
AnnikaCodes, line 82:
```suggestion
				const phrase = repeat.faq ? Chat.formatText(roomFaqs[room.roomid][repeat.faq], true).replace(/\n/g, '<br />') :
					repeat.phrase;
```
Personally, I think it would be more readable to make `Chat.formatText(..., true).replace(/\n/g, '<br />');` a function, since it's kind of unclear that it's transforming the rfaq text into something that we can repeat, but that's up to you.

server/chat-plugins/repeats.ts
```diff
@@ -146,28 +159,76 @@ export const commands: ChatCommands = {
 		this.runBroadcast();
 		this.sendReplyBox(
 			`<code>/repeat [minutes], [phrase]</code>: repeats a given phrase every [minutes] minutes.<br />` +
+			`<code>/repeatfaq [minutes], [FAQ name/alias]</code>: repeats a given Room FAQ every [minutes] minutes.<br />` +
 			`<code>/removerepeat [phrase]</code>: removes a repeated phrase.<br />` +
```
AnnikaCodes, line 163: This should probably mention that you can specify a room in PMs.

server/chat-plugins/repeats.ts
```diff
@@ -146,28 +159,76 @@ export const commands: ChatCommands = {
...
-			`/viewrepeats [optional room] - Displays all repeated phrases in a room.<br />` +
+			`<code>/removerepeatfaq [faq]</code>: removes a repeated FAQ.<br />` +
```
AnnikaCodes, line 164:
```suggestion
			`<code>/removerepeatfaq [FAQ name/alias]</code>: removes a repeated FAQ.<br />` +
```
I like consistency

server/chat-plugins/repeats.ts
```diff
@@ -4,10 +4,13 @@
  * @author Annika, Zarel
  */
 
+import {roomFaqs, getAlias} from './room-faqs';
+
 export interface RepeatedPhrase {
 	phrase: string;
 	/** interval in milliseconds */
 	interval: number;
+	faq?: string;
```
AnnikaCodes, line 13: I wonder if it would make more sense to give each repeat an ID of some kind that can be used to remove it. That would solve the issue of not being able to remove repeats with commas in them, and then I think you could just set `phrase` to be the FAQ text.

AnnikaCodes: Something like ``/repeat [minutes], [name], [phrase]``.

AnnikaCodes (changes requested):
This seems like it'll be very useful; thanks for working on it!

server/chat-plugins/repeats.ts
AnnikaCodes: (Either of us could also do this in a separate PR; I'm fine with this implementation for now.)

server/chat-plugins/repeats.ts
KrisXV: i think that should be handled in another PR since it's used within numerous files so it's better as a utils/chat function

server/chat-plugins/repeats.ts
KrisXV: I intentionally made the split message very odd so that people wouldnt do it for other rooms in PMs/rooms

server/chat-plugins/repeats.ts
KrisXV: Good idea, but I'll leave that to another PR

server/chat-plugins/repeats.ts
AnnikaCodes: Ah, OK, that's fair.

AnnikaCodes (approved):
Looks good to me!

server/chat-plugins/repeats.ts
KrisXV: Actually, I'll do the name thing now

KrisXV:
@zarel, annika and I went over the new code in PMs on PS and she said the approval still stands, so I believe this is ready.

Zarel:
Cool, thanks!

PR #7489, opened 2020-10-08
/showteam: Properly handle nick/form names

mia-pi-git:


server/chat-commands/core.ts
```diff
@@ -952,7 +952,10 @@ export const commands: ChatCommands = {
 				teamStrings = [indexedSet];
 			}
 		}
-		const nicknames = teamStrings.map(item => item.species !== item.name ? item.name : item.species);
+		const nicknames = teamStrings.map(item => {
+			const species = Dex.getSpecies(item.species).baseSpecies;
+			return species !== item.name ? item.name : species;
+		});
```
Zarel, line 958: I get the impression `item` isn't actually an `Item` instance. Can you rename this variable for clarity?

Zarel:
Thanks!

PR #7493, opened 2020-10-08
Translations: Convert JSON to TypeScript

AnnikaCodes:
I also made `language` attributes IDs since that's already how they were being used.


translations/traditionalchinese/minor-activities.ts
```diff
@@ -0,0 +1,62 @@
+import {Translations} from "../../server/chat";
```
Zarel, line 1:
```suggestion
import type {Translations} from "../../server/chat";
```

(and so on for all the others)

server/chat.ts
```diff
@@ -97,6 +97,15 @@ export type PunishmentFilter = (user: User | ID, punishment: Punishment) => void
...
 export type HostFilter = (host: string, user: User, connection: Connection, hostType: string) => void;
 
+export interface TRStrings {
+	[k: string]: string;
+}
+
+export interface Translations {
+	name?: string;
+	strings: TRStrings;
+}
```
Zarel, line 107:
```suggestion
export interface Translations {
	name?: string;
	strings: {[english: string]: string};
}
```

It doesn't look like you actually use the `TRStrings` type anywhere.

AnnikaCodes: Hm, I think that was just a leftover from before.

.eslintrc-no-types.json
```diff
@@ -192,7 +192,7 @@
     {
       "files": [
         "./config/*.ts", "./data/**/*.ts", "./lib/*.ts", "./server/**/*.ts", "./sim/**/*.ts",
-        "./tools/set-import/*.ts",  "./tools/modlog/*.ts"
+         "./tools/set-import/*.ts",  "./tools/modlog/*.ts", "./translations/**/*.ts"
```
Zarel, line 195:
```suggestion
        "./tools/set-import/*.ts",  "./tools/modlog/*.ts", "./translations/**/*.ts"
```

Zarel:
Thanks!

PR #7496, opened 2020-10-09
Chat: Propagate promises for loading translations

AnnikaCodes:
Although I was unable to replicate the crash about translating to a nonexistent language locally, this should fix it (and is a good change regardless).


Zarel:
Thanks!

PR #7497, opened 2020-10-09
RPS: Improve matchup handling

mia-pi-git:


server/chat-plugins/rock-paper-scissors.ts
```diff
@@ -373,6 +360,11 @@ export const commands: ChatCommands = {
 			game.unpause(user);
 		},
 
+		start() {
+			const game = getGame(this);
+			game.start();
+		},
```
Zarel, line 366: Why do you need a start command?

mia-pi-git: Actually, shouldn't be necessary. Will remove.

Zarel:
Thanks!

PR #7498, opened 2020-10-09
Modlog viewer: Improve IP handling

AnnikaCodes:
- prevents non-global roomstaff from seeing IPs in modnotes or legacy modlogs
- makes modnoted IP addresses link to whatismyipaddress.com


Zarel:
Thanks!

PR #7499, opened 2020-10-09
Moderation: Notify Staff for RO promos

mia-pi-git:
[17:17:15] &The Immortal: now that i think about it
[17:17:31] &The Immortal: we should probably have ro promote notices here


Zarel:
Thanks!

PR #7501, opened 2020-10-09
Fix exploit in dexsearch

Volco:
a length of 300 just allows someone to `[do a thing]` over and over until it crashes a server from lag. This isn't as big of an issue on main as it is on side server due to main likely having a very high-end host in comparison to the average side server but it's still something I think should be patched since just to prevent any chance of it being a future issue else where.

If you want me to change the allowed argument limit  or handle it differently lmk, this was just the simplest and quickest fix I thought of

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -78,6 +78,10 @@ export const commands: ChatCommands = {
 		this.checkBroadcast();
 		if (!target) return this.parse('/help dexsearch');
 		target = target.slice(0, 300);
+		// make sure there isn't more than 10 arguments to prevent exploitability
+		if (target.split(",").length > 10) {
+			target = target.split(",").splice(0, 10).join(",");
+		}
 		const targetGen = parseInt(cmd[cmd.length - 1]);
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
I'm working on a more comprehensive fix which is simply to terminate queries to the ProcessManager that run for too long. This fixes the exploit in a more general case than just in `/dexsearch`.

mia-pi-git:
This also would prevent legitimate queries that just happen to be long.

Volco:
I'll just leave this open for whatever Zarel decides to do with it regardless

KrisXV:
I edited the exploit out of your message @Volco because we don't want to tell people how to perform a DoS attack.

Volco:
that's fair

Zarel:
I talked to one of your server's sysadmins and the length of 300 can't actually crash your server from lag, so I'm rejecting this. I don't believe an argument limit of 10 is necessary.

Zarel:
image

Volco:
so i looked into it after apparently it was fixed before the PR as i didn't realize dawn's dexsearch outdated since i never check that file since i have almost no reason to

PR #7502, opened 2020-10-10
ProcessManager: Support timing out queries

AnnikaCodes:


Zarel:
Hm. This starts one timer per query, which might be a bit much? I wonder if there's a way that involves sweeping rather than starting that many timers.

AnnikaCodes:
I could add a `lastResponse` parameter to `processWrapper` and then we could add checks to kill processes that have pending queries and haven't returned a response recently?

Zarel:
Hm, I think it would make sense as `currentRequestStartTime` which would be `0` once the response was received.

xfix:
> Hm. This starts one timer per query, which might be a bit much? I wonder if there's a way that involves sweeping rather than starting that many timers.

If your concern is performance, JavaScript timers are pretty much free, I would assume most JavaScript implementations to use some sort of priority queue for those (it would make sense to do so). Additionally, `query` is used for expensive queries only.

I personally prefer the original version, seems much less complicated.

Zarel:
I trust xfix. Sorry to make you go through this, but could you go back to the previous implementation?

AnnikaCodes:
Sure!

lib/process-manager.ts
```diff
@@ -460,17 +460,32 @@ export abstract class ProcessManager {
...
+		this.timeout = timeout;
 
 		processManagers.push(this);
 	}
-	query(input: T) {
+	async query(input: T) {
 		const process = this.acquire() as QueryProcessWrapper;
-		if (!process) return Promise.resolve(this._query(input));
-		return process.query(input);
+		const timeout = setTimeout(() => {
```
xfix, line 477: I would avoid starting a timeout when `process` doesn't exist. Yes, it will be immediately cleared because there is no `await` for `this._query`, but that doesn't mean we should unnecessarily add onto a queue just to remove it immediately after.

```javascript
		if (!process) return this._query(input);
```

xfix (changes requested):

lib/process-manager.ts
AnnikaCodes: Yes, that would be a good plan.

lib/process-manager.ts
```diff
@@ -460,17 +460,35 @@ export abstract class ProcessManager {
...
-		return process.query(input);
+
+		if (!process) return this._query(input);
+
+		const timeout = setTimeout(() => {
+			const debugInfo = process.debug || "No debug information found.";
+			process.destroy();
+			throw new Error(
+				`A query originating in ${this.basename} took too long to complete; the process has been killed.\n${debugInfo}`
+			);
```
Zarel, line 485:
```suggestion
			Monitor.crashlog(new Error(
				`A query originating in ${this.basename} took too long to complete; the process has been killed.\n${debugInfo}`
			));
```

(Forcing `uncaughtException` to handle a crash tends to cause subtle bugs, and should be a last resort.)

Zarel: WAIT NEVERMIND THIS IS PROCESS-MANAGER

lib/process-manager.ts
```diff
@@ -460,17 +460,35 @@ export abstract class ProcessManager {
 
 export class QueryProcessManager<T = string, U = string> extends ProcessManager {
 	_query: (input: T) => U | Promise<U>;
+	timeout: number;
 
-	constructor(module: NodeJS.Module, query: (input: T) => U | Promise<U>) {
+	/**
+	 * @param timeout The number of milliseconds to wait before terminating a query. Defaults to 900000 ms (15 minutes).
+	 */
+	constructor(module: NodeJS.Module, query: (input: T) => U | Promise<U>, timeout: number = 15 * 60 * 1000) {
```
Zarel, line 468:
```suggestion
	constructor(module: NodeJS.Module, query: (input: T) => U | Promise<U>, timeout = 15 * 60 * 1000) {
```

Zarel (approved):
Good other than the unnecessary type in the constructor.

Zarel:
Thanks!

PR #7504, opened 2020-10-10
Update to TypeScript 4.0

Zarel:
Making a PR so everyone knows that Jeopardy has a type issue that TS 4.0 caught.


Zarel:
Specifically, that Jeopardy has a type issue that TS 4.0 caught and that I've only suppressed. Jeopardy is currently unmaintained and this error doesn't seem to be causing problems so it can be left alone, but in the future, please do not lie about whether or not a type can be null except synchronously while the constructor is running.

Also, a general note brought up by the upgrade: You should actually not ever use `delete` except in objects that you are explicitly treating as hashtables. In objects that are actually objects, you always want to do `obj.prop = null` rather than `delete obj.prop`, for performance reasons.

PR #7506, opened 2020-10-11
Add various mechanics tests

DaWoblefet:
Tons of failing tests from the bug reports thread. I think I added 60 tests in total, 29 of them skipped failing mechanics tests. Everything up to post 789 is handled now.


test/sim/moves/encore.js
```diff
@@ -30,11 +30,11 @@ describe('Encore', function () {
 		// If a user's previous move was Focus Punch and it is Encored into Focus Punch while attempting to
 		// execute the move, the regular "you can't be hit" effect for Focus Punch will be enforced.
 		battle.makeChoices('move focuspunch 1, move teleport', 'move encore 1, move extremespeed 1');
-		assert.equal(battle.p2.active[0].hp, hp);
+		assert.strictEqual(battle.p2.active[0].hp, hp);
```
Zarel, line 33: nooo never use `strictEqual` don't fall for it....

DaWoblefet: ope, I used that in all of these tests because VS Code was showing assert.equal() was deprecated and said to use strictEqual instead
image


should I keep using assert.equal(), or would you rather me do like assert(value=== anotherValue)? I don't really like the latter since it doesn't show the two comparisons

Zarel: Yeah, that's a VS Code bug that was fixed 4 days ago and a new version with the fix hasn't been published yet. You'll just have to deal with it being crossed out until the fix is published. :(

I've published #7515, which should make it clearer what to do here.

DaWoblefet: Yes, thank you!

DaWoblefet:
@Zarel once you merge https://github.com/smogon/pokemon-showdown/pull/7515 it will cause merge conflicts for one of us. Would you like to merge your PR first and then have me rebase, or would you like to do it the other way around?

Zarel:
It doesn't seem to have caused a merge conflict. Note that Git is pretty smart: you need to edit the same line, not just the same file, to cause a merge conflict.

Zarel:
Unless there's something else you're afraid of. Should I wait for you to merge?

DaWoblefet:
Nope, go ahead.

Zarel:
Thanks!

PR #7507, opened 2020-10-11
TextFormatter: Support replacing \n without trusted

KrisXV:


test/server/chat.js
```diff
@@ -88,5 +88,17 @@ describe('Chat', function () {
 			Chat.formatText(`:)`, true),
 			`:)`
 		);
+		assert.equal(
+			Chat.formatText(`a\nb\nc`),
+			`a\nb\nc`
+		);
+		assert.equal(
```
AnnikaCodes, line 95: These should all be `strictEqual` but that is for another PR

AnnikaCodes:
The changes to my plugin are fine. Are there places where we want to format text with the trusted flag without replacing `\n`? I wouldn't expect there to be.

test/server/chat.js
KrisXV: I'm aware

KrisXV:
> The changes to my plugin are fine. Are there places where we want to format text with the trusted flag without replacing `\n`? I wouldn't expect there to be.

It auto replaces `\n` if `isTrusted === true`

test/server/chat.js
Zarel: NO THEY SHOULDN'T

Notice the top of this file:

```js
```

TypeScript marking these as wrong is a TypeScript bug.

Zarel: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/assert.d.ts

lol, this was fixed 3 days ago.

KrisXV: It's not typescript marking these wrong, `strict` still works but is is deprecated

Zarel:
I think the confusion is because the PR is called "Allow" rather than "Support", which suggests there are permission changes rather than just API changes.

Zarel:
Anyway thanks, and please do NOT replace `equal` with `strictEqual` anywhere in the codebase.

test/server/chat.js
Zarel: What do you mean?

Zarel: `require('assert').strict` is not only not deprecated, `require('assert')` _is_ deprecated...

AnnikaCodes: Oh, sorry, I didn't know that.

PR #7508, opened 2020-10-11
Remove background from /scav queue

PartMan7:
https://cdn.discordapp.com/attachments/467865863240810516/761475882904584212/Screenshot_2020-10-02_Showdown_.png

Gotten RO confirmation from Pants.
https://cdn.discordapp.com/attachments/762324232948023316/764693429711929344/Screenshot_2020-10-11_Showdown_.png


Zarel:
General reminder that you fix this sort of thing by adding a foreground color whenever you add a background color.

PartMan7:
I'd suggested that too; pants had set up a poll for the preferred option and the option with it being completely removed won out with 83%.

Zarel:
From Discord:

> yeah
> but I just mean in general for future reference
> if you ever add a background to any element anywhere ever, always also add a text color

PR #7513, opened 2020-10-12
Modlog: Support logging to a SQLite database

AnnikaCodes:
This PR takes the next step in migrating modlogs to SQLite: starting dual logging to a SQLite database as well as text files.

I am currently implemented note ID search simply by streaming to regex in JavaScript; this seemed like the least bad option based on the discussions I've had.


Zarel:
> I am currently implemented note ID search simply by streaming to regex in JavaScript; this seemed like the least bad option based on the discussions I've had.

I would normally separately store a note ID, and then do a `LIKE` search in SQL. Was this considered, and if so, why was it rejected?

AnnikaCodes:
I rejected the idea of a `note_ids` table because I am not aware of a satisfactory way to extract the IDs from old notes into a table, although we did discuss that as a possible solution for future modnotes.

Or did you mean having another column in the table for the `toID`d note, and using a `LIKE` query on that? That's definitely feasible and would eliminate the need to define our own `regex` function in SQLite. I'd want to see what @monsanto thinks about that solution, though.

Zarel:
> Or did you mean having another column in the table for the `toID`d note, and using a `LIKE` query on that? That's definitely feasible and would eliminate the need to define our own `regex` function in SQLite.

Yes, that's what I meant.

monsanto:
This was suggested for killing the dependency on better-sqlite3. The reason not to do it would be the extra storage space required, but I don't think it really matters either way.

monsanto:
I think I can do fast note ID search with a custom FTS5 tokenizer. This would require that I write a native sqlite extension in C or Rust. I would prefer doing the latter if there are appropriate bindings available. It is unclear how well this would work, but if it does work, then we will get the best of all possible worlds.

Since this would complicate install for users we could have a configuration option for enabling it, like we currently do for ripgrep. If its off we could just disable searching notes entirely, or fallback to slow `%LIKE%`.

monsanto:
Give this a shot. Make sure you have `libsqlite3-dev` installed or whatever the equivalent is on your platform. From `/native` run `./build.sh` and `sqlite3 ':memory:' < test.sql`

If this works then 
 
- Load `fts_id_tokenizer.o`
- Modify the schema to do `CREATE VIRTUAL TABLE modlog_fts USING fts5(note, content=modlog, content_rowid=modlog_id, tokenize='id_tokenizer')`
- Whenever you insert, also insert into `modlog_fts`
- Match on the note with `... FROM modlog_fts WHERE note MATCH '<query>'`

See https://sqlite.org/fts5.html for more information

AnnikaCodes:
```
annika@annika-mac ~/p/native (modlog-sql-new) [1]> /usr/local/opt/sqlite/bin/sqlite3 < test.sql 
** ids available: **
hello this is a test
with 1 numbers 2
Heizölrückstoßabdämpfung

** search for thisis: **
hello this is a test

** search for 1numbers2: **
with 1 numbers 2

** search for lrck: **
Heizölrückstoßabdämpfung
annika@annika-mac ~/p/native (modlog-sql-new)>
```

Looks good to me; I'll update the modlog logic to use it later today.

server/modlog.ts
```diff
@@ -153,21 +155,32 @@ export class Modlog {
...
 
 
-		this.modlogInsertionQuery = this.database.prepare(
-			`INSERT INTO modlog (timestamp, roomid, visual_roomid, action, userid, autoconfirmed_userid, ip, action_taker_userid, note)` +
-			` VALUES ($time, $roomID, $visualRoomID, $action, $userid, $autoconfirmedID, $ip, $loggedBy, $note)`
-		);
+		let insertionQuerySource = `INSERT INTO modlog (timestamp, roomid, visual_roomid, action, userid, autoconfirmed_userid, ip, action_taker_userid, note)`;
+		insertionQuerySource += ` VALUES ($time, $roomID, $visualRoomID, $action, $userid, $autoconfirmedID, $ip, $loggedBy, $note)`;
+		if (this.useFTS) {
+			insertionQuerySource += `INSERT INTO modlog_fts (timestamp, roomid, visual_roomid, action, userid, autoconfirmed_userid, ip, action_taker_userid, note)`;
```
monsanto, line 179: I haven't tried this yet but this doesn't look right to me. `modlog_fts` should have columns `note` and (implicitly) `rowid`. When you insert a modlog entry, get its `modlog_id` (from SQL, `last_insert_rowid()`) and `INSERT INTO modlog_fts (rowid, note) VALUES (<last row id>, <same note as you inserted into the modlog table>)`. In `modlog_fts`, `note` is the tokenized version whereas in `modlog` it is the literal note string you specified

databases/schemas/modlog-fts.sql
```diff
@@ -0,0 +1,33 @@
+SELECT load_extension("native/fts_id_tokenizer.o");
+
+CREATE TABLE IF NOT EXISTS modlog (
```
monsanto, line 3: I see no reason to duplicate this definition, you can just have the `load_extension` + `create virtual table` in a separate file that is run after `modlog.sql`

server/modlog.ts
```diff
@@ -498,8 +511,13 @@ export class Modlog {
 
 			args.push(...Array(6).fill(search.anyField));
 
-			query += ` OR regex(?, note)`;
-			args.push(this.generateIDRegex(toID(search.anyField)));
+			if (this.useFTS) {
+				query += ` OR note MATCH ?`;
```
monsanto, line 515: No way that this works, you need to `JOIN modlog_fts ON modlog_fts.rowid = modlog_id` and match on `modlog_fts.note`

.eslintrc-no-types.json
```diff
@@ -192,7 +192,7 @@
     {
       "files": [
         "./config/*.ts", "./data/**/*.ts", "./lib/*.ts", "./server/**/*.ts", "./sim/**/*.ts",
-        "./tools/set-import/*.ts",  "./tools/modlog/*.ts", "./translations/**/*.ts"
```
monsanto, line 195: Please just commit style nits like this directly

server/modlog.ts
```diff
@@ -324,24 +466,103 @@ export class Modlog {
...
+
+		if (search.anyField) {
+			query += ` AND action LIKE '%' || ? || '%'`;
+
+			query += ` OR userid LIKE '%' || ? || '%'`;
+			query += ` OR autoconfirmed_userid LIKE '%' || ? || '%'`;
+			query += ` OR EXISTS(SELECT * FROM alts WHERE alts.modlog_id = modlog.modlog_id AND userid LIKE '%' || ? || '%')`;
+
+			query += ` OR ip LIKE '%' || ? || '%'`;
+			query += ` OR action_taker_userid LIKE '%' || ? || '%'`;
```
monsanto, line 510: Substring is always slow, if you want to do anything beyond a prefix match then you need to add an FTS column or manually tokenize.

server/modlog.ts
```diff
@@ -324,24 +466,103 @@ export class Modlog {
...
+
+		if (search.action) {
+			query += ` AND action LIKE '%' || ? || '%'`;
+			args.push(search.action);
+		} else if (onlyPunishments) {
+			query += ` AND action IN (${this.formatArray(PUNISHMENTS, args)})`;
+		}
+
+		if (userid) {
+			query += ` AND (userid LIKE '%' || ? || '%' OR autoconfirmed_userid LIKE '%' || ? || '%' OR EXISTS(SELECT * FROM alts WHERE alts.modlog_id = modlog.modlog_id AND userid LIKE '%' || ? || '%'))`;
```
monsanto, line 531: I don't think SQLite is smart enough to do `OR` efficiently, the optimal query plan does parallel index scan + ordered merge (see the PM log where I manually applied this optimization). This is also the strategy you would use to search all fields at the same time, so maybe it is worth waiting on a refactor of this function

AnnikaCodes: How would you advise I refactor it? (All your other suggestions should be done.)

server/modlog.ts
```diff
@@ -497,59 +500,60 @@ export class Modlog {
 		if (rooms.includes('global')) roomChecker = `(roomid LIKE 'global-%' OR ${roomChecker})`;
 
 		let query = `SELECT *, (SELECT group_concat(userid, ',') FROM alts WHERE alts.modlog_id = modlog.modlog_id) as alts FROM modlog`;
+		if (this.useFTS) query += ` LEFT JOIN modlog_fts ON modlog_fts.rowid = modlog_id`;
```
monsanto, line 503: Any reason for using a left join here over a regular join?

AnnikaCodes: because modlog entries with no note won't have a row in the FTS table

server/modlog.ts
```diff
@@ -497,59 +500,60 @@ export class Modlog {
...
 		query += ` WHERE ${roomChecker}`;
 
 		if (search.anyField) {
-			query += ` AND action LIKE '%' || ? || '%'`;
+			query += ` AND action LIKE ? || '%'`;
```
monsanto, line 507: Is this satisfactory? We have a lot of actions that are a combination of two words, like `GTS GIVEAWAY`. Would they expect a search on "giveaway" to return a GTS GIVEAWAY row?

server/modlog.ts
```diff
@@ -497,59 +500,60 @@ export class Modlog {
...
 
 			args.push(...Array(6).fill(search.anyField));
 
 			if (this.useFTS) {
-				query += ` OR note MATCH ?`;
+				query += ` OR modlog_fts.note MATCH ?`;
 				args.push(toID(search.anyField));
 			} else {
-				query += ` OR note LIKE '%' || ? || '%'`;
+				query += ` OR note LIKE ? || '%'`;
```
monsanto, line 522: We shouldn't have an index on note anyway, so this will always be `O(n)` for non-FTS case.

Would it be simpler if we just always required FTS, but used the default tokenization (word boundaries) for when you don't want to build the native extension?

AnnikaCodes: Yeah, that would be simpler.

server/modlog.ts
AnnikaCodes: I'll post in WAIL.

monsanto:
Unfortunately still doesn't work

```
monsanto@monsanto:~/downloads/pslogs/main$ ~/smogon/pokemon-showdown/tools/modlog/convert --db ./modlog.db --logdir ./modlog --from txt --to sqlite
Converting the text modlog files in ./modlog from txt to sqlite...
(node:9362) UnhandledPromiseRejectionWarning: RangeError: Too many parameter values were provided
    at ModlogConverterTxt.<anonymous> (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/converter.js:546:40)
    at ModlogConverterTxt.sqliteTransaction [as insertionTransaction] (/mnt/ssd1/smogon/pokemon-showdown/node_modules/better-sqlite3/lib/transaction.js:58:24)
    at insertEntries (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/converter.js:578:10)
    at ModlogConverterTxt.toSQLite (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/converter.js:597:47)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9362) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:9362) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
monsanto@monsanto:~/downloads/pslogs/main$ ~/smogon/pokemon-showdown/tools/modlog/convert --db ./modlog.db --logdir ./modlog --from txt --to sqlite
Converting the text modlog files in ./modlog from txt to sqlite...
(node:9410) UnhandledPromiseRejectionWarning: SqliteError: table modlog_fts already exists
    at new ModlogConverterTxt (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/converter.js:533:18)
    at Function.convert (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/converter.js:684:22)
    at Object.<anonymous> (/mnt/ssd1/smogon/pokemon-showdown/tools/modlog/convert:45:17)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
(Use `node --trace-warnings ...` to show where the warning was created)
(node:9410) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:9410) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

AnnikaCodes:
Hm, I don't know if that is FTS related. Fixed it, though!

monsanto:
Did you forget to push your fix?

AnnikaCodes:
yeah lmao

tools/modlog/converter.ts
```diff
@@ -530,18 +537,35 @@ export class ModlogConverterTxt {
...
 		);
-		this.FTSInsertionQuery = this.database.prepare(`INSERT INTO modlog_fts (rowid, note) VALUES (?, ?)`);
+		this.noteInsertionQuery = this.database.prepare(`INSERT INTO modlog_fts (rowid, note) VALUES (?, ?)`);
+		this.useridInsertionQuery = this.database.prepare(`INSERT INTO modlog_fts (rowid, userid) VALUES (?, ?)`);
+		this.autoconfirmedInsertionQuery = this.database.prepare(
+			`INSERT INTO modlog_fts (rowid, autoconfirmed_userid) VALUES (?, ?)`
+		);
+		this.actionTakerInsertionQuery = this.database.prepare(
+			`INSERT INTO modlog_fts (rowid, action_taker_userid) VALUES (?, ?)`
+		);
```
monsanto, line 559: Surely if you ran these in succession, you'd get a constraint error? Is there something I'm missing here? You've tested that this works, right?

AnnikaCodes: No, the converter works fine without constraint errors on my end.

monsanto: I see... this seems to be undocumented behavior with contentless FTS tables
```
sqlite> create virtual table foo using fts5(bar, baz);
sqlite> insert into foo (rowid, bar) values (1, 'test');
sqlite> insert into foo (rowid, baz) values (1, 'test');
Error: constraint failed
sqlite> create table foo2 (bar, baz);
sqlite> create virtual table foo_fts using fts5(bar, baz, content=foo2);
sqlite> insert into foo2 values ('test', 'test');
sqlite> insert into foo_fts (rowid, bar) values (1, 'test');
sqlite> insert into foo_fts (rowid, baz) values (1, 'test');
sqlite> select * from foo_fts('test');
test|test
```
I'd be careful about relying on this sort of thing, the FTS module seems to accept whatever you throw at it, even if it causes problems down the line:
```
sqlite> insert into foo_fts(rowid, bar) values (100, 'crash please');
sqlite> select * from foo_fts;
test|test
sqlite> select * from foo_fts('crash');
Error: database disk image is malformed
```
Is there an issue with inserting multiple columns in the same statement? Wouldn't it simplify the code a bit too?

AnnikaCodes: sure, that's fine. That's a bit worrying regarding illegal insertions being allowed through FTS though.

server/modlog.ts
```diff
@@ -137,8 +147,118 @@ export class Modlog {
...
+
+		// Set up tables, etc
+		this.database.exec(FS(MODLOG_SCHEMA_PATH).readIfExistsSync());
+		if (Config.modlogftsextension) {
+			this.database.exec(`SELECT load_extension('native/fts_id_tokenizer.o')`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id, tokenize='id_tokenizer')`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid, tokenize='id_tokenizer')`);
+		} else {
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id)`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid)`);
```
monsanto, line 176: When you include the converter, each FTS table definitions have been duplicated 4 times, which is a pretty big red flag.

You can at least reduce the duplication here removing the branch, having 'tokenize' be either `'id_tokenizer'` or `'unicode61'` depending on whether the extension exists.

The JavaScript branch can be removed completely by having a `variables`/`settings` table which stores key/value pairs. Then `Config.modlogftsextension` could be moved from JS to the DB. We'll need something like this for storing the schema version anyway for migrations.

tools/modlog/converter.ts
```diff
@@ -419,12 +421,220 @@ export function rawifyLog(log: ModlogEntry) {
...
+		if (useFTSExtension || Config.modlogftsextension) {
+			this.database.exec(`SELECT load_extension('native/fts_id_tokenizer.o')`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id, tokenize='id_tokenizer')`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid, tokenize='id_tokenizer')`);
+		} else {
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id)`);
+			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid)`);
+		}
+
+		this.insertionQuery = this.database.prepare(
```
monsanto, line 545: The fact that all of the queries are duplicated in the converter suggests that the modlog API is insufficient. The converter is a temporary thing, so I guess this isn't urgent, but it's a red flag

server/modlog.ts
monsanto: Actually, it seems like `tokenize` can't take an SQL expression, which would require duplicating the table definitions if we moved it all to SQL. I'm not sure what the right thing to do here is. One idea would be to have something like `%TOKENIZER%` in the SQL file and `.replace` it when we load the script.

AnnikaCodes: `.replace()` seems reasonable

tools/modlog/converter.ts
AnnikaCodes: oh, I just need to grab the transactions from `Modlog`. That'll be DRYer

server/modlog.ts
```diff
@@ -166,16 +166,14 @@ export class Modlog {
...
-		this.database.exec(FS(MODLOG_SCHEMA_PATH).readIfExistsSync());
+		let tokenizer = 'unicode61';
 		if (Config.modlogftsextension) {
 			this.database.exec(`SELECT load_extension('native/fts_id_tokenizer.o')`);
-			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id, tokenize='id_tokenizer')`);
-			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid, tokenize='id_tokenizer')`);
-		} else {
-			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS modlog_fts USING fts5(note, userid, autoconfirmed_userid, action_taker_userid, content=modlog, content_rowid=modlog_id)`);
-			this.database.exec(`CREATE VIRTUAL TABLE IF NOT EXISTS alts_fts USING fts5(userid, content=alts, content_rowid=rowid)`);
+			tokenizer = 'id_tokenizer';
```
monsanto, line 172: This is backwards

AnnikaCodes: What do you mean?

AnnikaCodes: I intended the `Config.modlogftsextension` option to be set when we were using the built extension. 
```ts

let tokenizer = 'unicode61';
if (Config.modlogftsextension) {
	this.database.exec(`SELECT load_extension('native/fts_id_tokenizer.o')`);
	tokenizer = 'id_tokenizer'; 
 }
```
This is the code I currently have; I don't think it will use `id_tokenizer` unless it's built?

monsanto: Sorry, you are correct, I misread the unified diff

server/modlog.ts
```diff
@@ -281,40 +281,55 @@ export class Modlog {
...
+		this.writeSQL(entries);
+		this.writeText(entries);
 	}
 
-	writeText(roomid: string, entry: ModlogEntry, overrideID?: string) {
-		const stream = this.streams.get(roomid as ModlogID);
-		if (!stream) throw new Error(`Attempted to write to an uninitialized modlog stream for the room '${roomid}'`);
+	writeSQL(entries: Iterable<ModlogEntry>) {
+		for (const entry of entries) {
+			this.insertionTransaction({
```
monsanto, line 297: This creates N transactions instead of inserting N inside 1 transaction

monsanto:
- I'm relatively confident in this plan of action and intend to merge this soon.

- Let's just remove the indexes for now, the base schema is sound, and since there won't be any querying for now we can just hold off. Having the dependency in will let me merge mia's work under an experimental flag.

- Like I mentioned on mia's PR, it isn't a good idea to have `IF NOT EXISTS` on the SQL. The initial schema code needs to run exactly once, and then after that migrations need to run exactly once.

- I would make the Config `usesqlitemodlog` instead of `nosqlitemodlog` and default to false. We don't want developers not looped into this feature & other servers to be aware of our work until it is ready for the prime time.

- The converter (and maybe other things at some point?) need to be able to import the `Modlog` class without it trying to open the modlog at `databases/modlog.db`.

If you handle the last two, I can handle the other two, and we can merge.

AnnikaCodes:
I did the last two.

monsanto:
Fine to merge, need to install extra dependencies, modify config, and hotpatch modlog

AnnikaCodes:
One sec. I don't think hotpatching modlog directly will work now that we construct Modlog in `rooms.ts`.

AnnikaCodes:
Okay whoever deploys needs to hotpatch chat first (to load the new hotpatching code) and then hotpatch modlog

monsanto:
That is a nasty hack, it would be nice if a follow-up commit better isolated the modlog from /hotpatch

AnnikaCodes:
Yes, once we restart and can depend on `Rooms` itll be simpler. I'd personally be inclined to put the `MODLOG_PATH` and stuff in `Config` instead but our precedent with these sorts of things is to use a `const` at the top of a file so I dont know if there's a reason we do that instead. Maybe to avoid cluttering `config.js`?

monsanto:
I don't think you need to export these at all, they are implementation details that /hotpatch should not be aware of. The process manager lines above also fall under that umbrella. A better implementation strategy would be for modules that support hotpatch to come with methods to reload themselves.

AnnikaCodes:
Sure, that sounds good. Id rather do that a separate commit though.

PR #7514, opened 2020-10-13
Chatlog: Support viewing log stats for a room

mia-pi-git:
Useful feature from several bots that PS can do better.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -485,6 +485,10 @@ export const LogSearcher = new class {
 
 		return `^` + searches.map(term => `(?=.*${term})`).join('');
 	}
+	constructUserRegex(user: string) {
+		const id = toID(user);
+		return `.${[...id].join('[^a-zA-Z0-9]*')}[^a-zA-Z0-9]*`;
+	}
 
 	fsSearch(roomid: RoomID, search: string, date: string, limit: number | null) {
 		const isAll = (date === 'all');
@@ -579,6 +583,7 @@ export const LogSearcher = new class {
 		if (!raw) {
 			search = this.constructSearchRegex(search);
 		}
+		const resultSep = args?.includes('-m') ? '--' : '\n';
 		try {
 			const options = [
 				'-e', search,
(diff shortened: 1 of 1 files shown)


Zarel:
I think peach said something about `rg -C` or `rg -c` or something? Should I wait for that change?

mia-pi-git:
> I think peach said something about `rg -C` or `rg -c` or something? Should I wait for that change?

I pushed that already. See line 701.

server/chat-plugins/chatlog.ts
```diff
@@ -666,17 +687,135 @@ export const LogSearcher = new class {
...
+			buf += `<small><a roomid="view-roomstats-${roomid}--${prevMonth}${user ? `--${user}` : ''}">Previous month</a></small>`;
+		}
+		if (FS(`logs/chat/${roomid}/${nextMonth}`).existsSync()) {
+			buf += ` <small><a roomid="view-roomstats-${roomid}--${nextMonth}${user ? `--${user}` : ''}">Next month</a></small>`;
+		}
+		buf += `<hr /><ol>`;
+		if (user) {
+			const sortedDays = Object.keys(results).sort((a, b) => {
+				return new Date(b).getTime() - new Date(a).getTime();
+			});
```
Zarel, line 771:
```suggestion
			const sortedDays = Object.keys(results).sort((a, b) => (
				new Date(b).getTime() - new Date(a).getTime()
			));
```

I feel like I've suggested this a lot lately.

server/chat-plugins/chatlog.ts
```diff
@@ -666,17 +687,135 @@ export const LogSearcher = new class {
...
+			for (const date in results) {
+				for (const userid in results[date]) {
+					if (!totalResults[userid]) totalResults[userid] = 0;
+					totalResults[userid] += results[date][userid];
+				}
+			}
+			const resultKeys = Object.keys(totalResults);
+			const sortedResults = resultKeys.sort((a, b) => {
+				return totalResults[b] - totalResults[a];
+			}).slice(0, MAX_TOPUSERS);
```
Zarel, line 788:
```suggestion
			const sortedResults = resultKeys.sort((a, b) => (
				totalResults[b] - totalResults[a]
			)).slice(0, MAX_TOPUSERS);
```

Zarel:
https://eslint.org/docs/rules/arrow-body-style#require-braces-in-arrow-function-body-arrow-body-style

We should be turning this on, apparently.

Zarel:
Thanks. The `resultSep` stuff is kinda confusing but it seems to be working around a ripgrep quirk?

mia-pi-git:
> Thanks. The `resultSep` stuff is kinda confusing but it seems to be working around a ripgrep quirk?

Yeah, it's not my favorite but ripgrep uses different separators for searches with -m as opposed to ones without.

mia-pi-git:
It uses `---` to deliminate context blocks.

PR #7515, opened 2020-10-13
Stop using assert.strict.strictEqual

Zarel:
It turns out that when I switched us from `assert` to `assert.strict`, I didn't actually update any existing tests or tell anyone:

https://github.com/smogon/pokemon-showdown/commit/0df0d234f2ae65f69e17a15c938a5c7c9a9352ac

So apparently everyone else just kept on using `strictEqual`.

This will be a PR and also throw an error if people continue trying to use it, which should make it much clearer what PS policy is on this.

A lot of the problem may be that TypeScript marks assert.strict.equal as deprecated when it's not. This was fixed 4 days ago:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/48452

But this probably hasn't made it to a thing yet. Until then, you'll have to deal with TS marking your tests as deprecated, but it shouldn't be too long.

### DIFF
--- a/test/assert.js
+++ b/test/assert.js
@@ -10,7 +10,8 @@
 
 'use strict';
 
-const assert = require('assert').strict;
+const legacyAssert = require('assert');
+const assert = legacyAssert.strict;
 const AssertionError = assert.AssertionError;
 
 assert.bounded = function (value, range, message) {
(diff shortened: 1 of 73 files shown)


DaWoblefet:
While you're here, you may as well deprecate assert.ok() - if I recall correctly we don't like that one either.

Zarel:
Good catch; thanks!

PR #7516, opened 2020-10-13
Ban Japanese Gen 1 Events in International Formats

PlagueVonKarma:
Not sure if this is how to go about it, but since it's impossible for international Game Boy games to trade with Japanese ones, I don't see why these are actually legal on PS. You straight-up can't get these on any international game, which I believe is the base for simulation? I believe you could make some kind of Event Clause or learnset thing for these, but this seems to be the easier way to go about it? 

Feel free to throw this out and ignore if it's the incorrect way to go about it. They're all super niche, it's more of an exhaustive thing. Fly is already banned by Standard due to the semi-invulnerability glitch, but in case it's patched by the council or semi-invulnerability clause is ever instated it's probably best to have this anyway. 

Overall, this bans;
- Flying Pikachu, and by extension Raichu (September 1997 Corocoro Event)
- Dragon Rage Magikarp (Shogakukan "Tamamushi University Hyper Test" Event)
- Pay Day Fearow (Shogakukan Pokemon Stamp Event)
- Pay Day Rapidash (Shogakukan Pokemon Stamp Event)

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -2860,7 +2860,7 @@ export const Formats: FormatList = [
 		mod: 'gen1',
 		searchShow: false,
 		ruleset: ['Standard'],
-		banlist: ['Magikarp + Dragon Rage', 'Rapidash + Pay Day', 'Fearow + Pay Day', 'Pikachu + Fly', 'Raichu + Fly']
+		banlist: ['Magikarp + Dragon Rage', 'Rapidash + Pay Day', 'Fearow + Pay Day', 'Pikachu + Fly', 'Raichu + Fly'],
 	},
 	{
 		name: "[Gen 1] UU",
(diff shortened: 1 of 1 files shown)


DisasterAreaPS:
right idea but i think the wrong way to go about it? probably should go in learnsets somewhere??

PlagueVonKarma:
> right idea but i think the wrong way to go about it? probably should go in learnsets somewhere??

Yeah, I've not really done this before but I didn't really know where else to go with it.

_Of course I missed the commas again..._

xfix:
Maybe we just shouldn't list those events in `eventData` data. There is no support in Pokemon Showdown for Japanese first generation Pokemon games - in particular we don't support Blizzard with 30% chance to freeze.

PlagueVonKarma:
> Maybe we just shouldn't list those events in `eventData`. There is no support otherwise in Pokemon Showdown for Japanese first generation Pokemon games - in particular we don't support Blizzard with 30% chance to freeze.

I personally agree. The JP games have no support whatsoever; Swift accuracy glitch isn't there, etc. The only time I've seen it supported is on an RBY side server.

Zarel:
This should probably be implemented with an event flag and a validator update. Are the individual `eventData` entries identifiable as Japanese?

KrisXV:
No, but in general I would like to widely improve the support for old gens for Japanese mechanics and also Stadium 2, etc.

PlagueVonKarma:
> No, but in general I would like to widely improve the support for old gens for Japanese mechanics and also Stadium 2, etc.
I would happily assist with research for this if necessary, and there's a lot of the work done on RBY 2k20's PS Server. Stad2 has never had extensive research. 

On a side note, Gen 1 events don't seem to be listed on the learnset files at all. Or at least, not in any eventData area.

KrisXV:
> > No, but in general I would like to widely improve the support for old gens for Japanese mechanics and also Stadium 2, etc.
> 
> I would happily assist with research for this if necessary, and there's a lot of the work done on RBY 2k20's PS Server. Stad2 has never had extensive research.
> 
> On a side note, Gen 1 events don't seem to be listed on the learnset files at all. Or at least, not in any eventData area.

They're in the `mods/gen2/learnsets.ts` file

Zarel:
Can one of you submit a PR marking the Japanese events in `mods/gen2/learnsets.ts`?

Zarel:
(I can take it from there. It doesn't need to pass unit tests if you can't figure out how to make your markings pass TypeScript.)

PR #7518, opened 2020-10-13
Repeats: Support non-time-based repeats

AnnikaCodes:
This PR adds support for repeating a phrase every n messages, as well as a more generic interface for running a function every n messages in a room.

I'm not sure if this will cause performance issues because it loops over `nthMessageHandlers` on every call to `add`; is there a better way to do that?


Zarel:
I would probably not want to do it every `n` lines with chat. There are certain messages that shouldn't be separated, like the clear thingy.

One thing you can do is to do it with `Chat.parse`, check the room.log length before and after, and if it's changed, count that as a message.

AnnikaCodes:
(This is done.)

AnnikaCodes:
@Zarel bump

AnnikaCodes:
@Zarel bump

Zarel:
I'm sorry I've taken so long to review this. I just got a lot of regrets from accepting repeats at all, and I really don't want "more tools to spam the user". I feel the same way about this as about https://www.smogon.com/forums/threads/pinned-messages-at-the-top-of-chatroom.3673227/

Roomintros and polls already seem like too much to me, tbh.

AnnikaCodes:
Is this just not going to be added as a feature? @TheMezStrikes, do you want to elaborate on why you suggested that I implement this?

KrisXV:
It prevents repeat spam of less-active rooms during dead hours for one.

Zarel (approved):

Zarel:
I approve this for the next restart. I'm not happy with this, but I suppose it's not worse than supporting repeats at all.

PR #7519, opened 2020-10-14
Chatlog: Use new username html tag

mia-pi-git:


Zarel:
This is good but needs to wait one week or a server restart.

mia-pi-git:
> This is good but needs to wait one week or a server restart.

It should be fine by now, I think (Unless i'm missing something?). I asked Marty to redeploy the client before I made this PR.

Zarel:
I also redeployed the client. But you need to wait for a server restart or a week, for everyone to refresh the client after the redeploy.

Zarel (approved):

server/chat-plugins/chatlog.ts
```diff
@@ -325,8 +325,11 @@ export const LogViewer = new class {
 				if (opts !== 'all') return `<div class="notice">[uhtml box hidden]</div>`;
 				return `<div class="notice">${message.slice(message.indexOf(',') + 1)}</div>`;
 			}
-			const group = !name.startsWith(' ') ? `<small>${name.charAt(0)}</small>` : ``;
-			return `<div class="chat"><small>[${timestamp}] </small><strong>${group}${Utils.escapeHTML(name.slice(1))}:</strong> <q>${Chat.formatText(message)}</q></div>`;
+			const group = !name.startsWith(' ') ? name.charAt(0) : ``;
+			return (
+				`<div class="chat"><small>[${timestamp}] ${group}</small><username>${Utils.escapeHTML(name.slice(1))}:` +
+				`</username> <q>${Chat.formatText(message)}</q></div>`
+			);
```
Zarel, line 332:
```suggestion
			return `<div class="chat">` +
				Utils.html`<small>[${timestamp}] ${group}</small><username>${name.slice(1)}:</username> ` +
				`<q>${Chat.formatText(message)}</q>` +
				`</div>`;
```

When possible, we prefer for opening and closing tags to be in the same line.

Zarel:
ty

PR #7520, opened 2020-10-14
Helptickets: Don't show battles for self-reports

AnnikaCodes:
Requested by DaWoblefet


Zarel:
@DaWoblefet can you explain why not? it still seems like it could be useful sometimes, and I can't imagine it _harms_ anything?

DaWoblefet:
@Zarel this is for if you click Report on yourself. It tries to display battles that you had against _yourself_. Which is impossible. Even if it were possible, the probability that you experienced harassment against yourself while playing against yourself is very low, I'd say :P

Zarel:
Oh, I see. Thanks!

PR #7521, opened 2020-10-14
Update punishment expiration messaging to give a more accurate time

Bardiches:
Currently, bans/locks/etc. only tell you that they will expire "in a few days". This updates it such that it tells you how many days remain and how many hours if there's less than a day left.

This also refactors the punishment API a bit. Now all of the APIs that punish someone return not only the list of affected users, but also the punishment.

First time contributing, let me know what you think!

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -693,7 +693,7 @@ export const commands: ChatCommands = {
 		this.addModAction(`${name} was banned ${week ? ' for a week' : ''} from ${room.title} by ${user.name}.${reason}`);
 
 		const time = week ? Date.now() + 7 * 24 * 60 * 60 * 1000 : null;
-		const affected = Punishments.roomBan(room, targetUser, time, null, target);
+		const {affected} = Punishments.roomBan(room, targetUser, time, null, target);
 
 		if (!room.settings.isPrivate && room.persist) {
 			const acAccount = (targetUser.autoconfirmed !== userid && targetUser.autoconfirmed);
@@ -806,13 +806,14 @@ export const commands: ChatCommands = {
 
 		// Use default time for locks.
 		const duration = week ? Date.now() + 7 * 24 * 60 * 60 * 1000 : (month ? Date.now() + 30 * 24 * 60 * 60 * 1000 : null);
+		let punishment;
 		let affected = [];
 
 		if (targetUser) {
 			const ignoreAlts = Punishments.sharedIps.has(targetUser.latestIp);
-			affected = await Punishments.lock(targetUser, duration, null, ignoreAlts, userReason);
+			({punishment, affected} = await Punishments.lock(targetUser, duration, null, ignoreAlts, userReason));
 		} else {
-			affected = await Punishments.lock(userid, duration, null, false, userReason);
+			({punishment, affected} = await Punishments.lock(userid, duration, null, false, userReason));
 		}
 
 		const globalReason = (target ? `${userReason} ${proof}` : '');
@@ -846,7 +847,7 @@ export const commands: ChatCommands = {
 			}
 
 			if (appeal) message += `\n\nIf you feel that your lock was unjustified, you can ${appeal}.`;
-			message += `\n\nYour lock will expire in a few days.`;
+			message += `\n\nYour lock will expire ${Punishments.getPunishmentExpirationDescription(punishment)}.`;
 			targetUser.send(message);
 
 			const roomauth = Rooms.global.destroyPersonalRooms(userid);
@@ -1000,11 +1001,17 @@ export const commands: ChatCommands = {
 			userReason = target.substr(0, proofIndex).trim();
 		}
 
-		targetUser.popup(`|modal|${user.name} has globally banned you.${(userReason ? `\n\nReason: ${userReason}` : ``)} ${(Config.appealurl ? `\n\nIf you feel that your ban was unjustified, you can appeal:\n${Config.appealurl}` : ``)}\n\nYour ban will expire in a few days.`);
+		const {punishment, affected} = await Punishments.ban(targetUser, null, null, false, userReason);
+
+		targetUser.popup(
+			`|modal|${user.name} has globally banned you.` +
+			(userReason ? `\n\nReason: ${userReason}` : '') +
+			(Config.appealurl ? `\n\nIf you feel that your ban was unjustified, you can appeal:\n${Config.appealurl}` : ``) +
+			`\n\nYour ban will expire ${Punishments.getPunishmentExpirationDescription(punishment)}.`
+		);
 
 		this.addGlobalModAction(`${name} was globally banned by ${user.name}.${(userReason ? ` (${userReason})` : ``)}`);
 
-		const affected = await Punishments.ban(targetUser, null, null, false, userReason);
 		const acAccount = (targetUser.autoconfirmed !== userid && targetUser.autoconfirmed);
 		let displayMessage = '';
 		if (affected.length > 1) {
@@ -1745,7 +1752,7 @@ export const commands: ChatCommands = {
 
 		this.privateModAction(`${name} was blacklisted from ${room.title} by ${user.name}. ${target ? ` (${target})` : ''}`);
 
-		const affected = Punishments.roomBlacklist(room, targetUser, null, null, target);
+		const {affected} = Punishments.roomBlacklist(room, targetUser, null, null, target);
 
 		if (!room.settings.isPrivate && room.persist) {
 			const acAccount = (targetUser.autoconfirmed !== userid && targetUser.autoconfirmed);
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -469,7 +469,7 @@ export const Punishments = new class {
 		}
 
 		if (!bypassPunishmentfilter) Chat.punishmentfilter(user, punishment);
-		return affected;
+		return {punishment, affected};
 	}
 
 	async punishInner(user: User, punishment: Punishment, userids: Set<ID>, ips: Set<string>, mobileIps: Set<string>) {
@@ -546,7 +546,7 @@ export const Punishments = new class {
 		}, id, PUNISHMENT_FILE);
 
 		Chat.punishmentfilter(userid, punishment);
-		return affected;
+		return {punishment, affected};
 	}
 
 	unpunish(id: string, punishType: string) {
@@ -610,7 +610,7 @@ export const Punishments = new class {
 			}
 		}
 
-		return affected;
+		return {punishment, affected};
 	}
 
 	roomPunishInner(roomid: RoomID, user: User, punishment: Punishment, userids: Set<string>, ips: Set<string>) {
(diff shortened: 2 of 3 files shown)


server/punishments.ts
```diff
@@ -1358,24 +1374,31 @@ export const Punishments = new class {
...
+		const expiresIn = new Date(punishment[2]).getTime() - Date.now();
+		const expiresDays = Math.round(expiresIn / 1000 / 60 / 60 / 24);
+		const expiresHours = Math.round(expiresIn / 1000 / 60 / 60);
+		if (expiresDays >= 1) {
+			return `in around ${Chat.count(expiresDays, "days")}`;
+		} else if (expiresHours >= 1) {
+			return `in around ${Chat.count(expiresHours, "hours")}`;
+		} else {
+			return 'within the next hour';
+		}
```
mia-pi-git, line 1401: Just use ``Chat.toDurationString`` for this.

server/punishments.ts
```diff
@@ -742,14 +742,16 @@ export const Punishments = new class {
 		if (!expireTime) expireTime = Date.now() + LOCK_DURATION;
 		const punishment = ['LOCK', id, expireTime, reason] as Punishment;
 
-		const affected = await Punishments.punish(user, punishment, ignoreAlts, bypassPunishmentfilter);
+		const {
+			punishment: outputPunishment, affected,
+		} = await Punishments.punish(user, punishment, ignoreAlts, bypassPunishmentfilter);
```
AnnikaCodes, line 747:
```suggestion
		const {
			punishment: outputPunishment,
			affected,
		} = await Punishments.punish(user, punishment, ignoreAlts, bypassPunishmentfilter);
```

server/punishments.ts
```diff
@@ -772,11 +774,17 @@ export const Punishments = new class {
...
 		if (Users.get(user)?.locked) return false;
 		const name = typeof user === 'string' ? user : (user as User).name;
+		let outputPunishment;
+		let affected;
 		if (namelock) {
 			punishment = `NAME${punishment}`;
-			await Punishments.namelock(user, expires, toID(namelock), false, `Autonamelock: ${name}: ${reason}`);
+			({
+				punishment: outputPunishment, affected,
+			} = await Punishments.namelock(user, expires, toID(namelock), false, `Autonamelock: ${name}: ${reason}`));
```
AnnikaCodes, line 783:
```suggestion
			{
				punishment: outputPunishment,
				affected,
			} = await Punishments.namelock(user, expires, toID(namelock), false, `Autonamelock: ${name}: ${reason}`);
```
What are the parentheses for?

server/punishments.ts
```diff
@@ -1358,24 +1374,31 @@ export const Punishments = new class {
 	 * @return Descriptive text for the remaining time until the punishment expires, if any.
 	 */
 	checkLockExpiration(userid: string | null) {
-		if (!userid) return ``;
+		if (!userid) return '';
+
+		const user = Users.get(userid);
+		if (user?.permalocked) return '(expires never)';
```
AnnikaCodes, line 1380:
```suggestion
		if (user?.permalocked) return '(never expires; you are permalocked)';
```

server/punishments.ts
```diff
@@ -1575,12 +1598,14 @@ export const Punishments = new class {
 				const message = `${(user as User).name || userid} was locked for having punishments in ${punishments.length} rooms: ${punishmentText}`;
 				const isWeek = await Rooms.Modlog.getGlobalPunishments(userid, AUTOWEEKLOCK_DAYS_TO_SEARCH) >= AUTOWEEKLOCK_THRESHOLD;
 
-				void Punishments.autolock(user, 'staff', 'PunishmentMonitor', reason, message, isWeek);
+				const autoLockResult = await Punishments.autolock(user, 'staff', 'PunishmentMonitor', reason, message, isWeek);
```
AnnikaCodes, line 1601:
```suggestion
				const autolockResult = await Punishments.autolock(user, 'staff', 'PunishmentMonitor', reason, message, isWeek);
```

AnnikaCodes (changes requested):
More generally, I don't think that the punishing methods themselves need to return the punishment (although I'm not opposed to it); the duration is determined by the caller of the `Punishments` method, so returning a `Punishment` doesn't seem terribly useful.

AnnikaCodes:
I like the overall idea of this change, though!

Bardiches:
> More generally, I don't think that the punishing methods themselves need to return the punishment (although I'm not opposed to it); the duration is determined by the caller of the `Punishments` method, so returning a `Punishment` doesn't seem terribly useful (are there other usecases for this that you envision?).

@AnnikaCodes In the case where the expiration time is not given, it's useful to have access to the expiration time without having to inspect the default values. E.g. You call `Punishment.battleban` without an expire time, you'll end up with `Date.now() + BATTLEBAN_DURATION` and it may be useful to the caller (for generating the pop-up) to know this.

AnnikaCodes:
Hm, that makes sense. It might be clearer to remove all of the default times from `punishments.ts` but I'm fine with this.

server/punishments.ts
Bardiches: They're necessary for destructuring assignment like this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

server/punishments.ts
Bardiches: Sure, makes sense. Updated in the most recent version.

server/punishments.ts
AnnikaCodes: Oh, I didn't realize it was different if you had already declared the variable.

Bardiches: Yeah, surprised me too, never used this pattern before

DaWoblefet:
Generally speaking I don't mind this idea personally, but I think it was left vague as a matter of policy, not necessarily because we never added the functionality. It might be bad for locked users to have a timestamp of exactly when they can start harassing another person again, for example. I'll reach out to policy folks and get some opinions.

Bardiches:
> Generally speaking I don't mind this idea personally, but I think it was left vague as a matter of policy, not necessarily because we never added the functionality. It might be bad for locked users to have a timestamp of exactly when they can start harassing another person again, for example. I'll reach out to policy folks and get some opinions.

@DaWoblefet I could reduce the precision if that's helpful. It just seems a bit vague/unhelpful to have it say "a few days" when it could technically be much, much longer.

DaWoblefet:
@Bardiches I spoke with some of our policy staff and most people seem to think it's ok. There were a few with reservations but I think worst case scenario will be getPunishmentExpirationDescription() would just change the precision call in toDurationString(). Pretty sure you can proceed.

AnnikaCodes (approved):
This seems fine to me; thanks!

Zarel:
We originally left punishment time vague for a lot of reasons, but one of them is that we have a policy of not expiring global punishments while the user is online. This means that users will remain locked/banned even after the punishment expires, until they've logged out and been offline for an hour.

I don't know if this is still the case (although I don't remember any policy discussions changing that), but if it is, that will need to be fixed. I'm open to that change, but that _would_ need to be run past policy staff as well.

server/chat-commands/moderation.ts
```diff
@@ -806,13 +806,14 @@ export const commands: ChatCommands = {
 
 		// Use default time for locks.
 		const duration = week ? Date.now() + 7 * 24 * 60 * 60 * 1000 : (month ? Date.now() + 30 * 24 * 60 * 60 * 1000 : null);
+		let punishment;
 		let affected = [];
 
 		if (targetUser) {
 			const ignoreAlts = Punishments.sharedIps.has(targetUser.latestIp);
-			affected = await Punishments.lock(targetUser, duration, null, ignoreAlts, userReason);
+			({punishment, affected} = await Punishments.lock(targetUser, duration, null, ignoreAlts, userReason));
```
Zarel, line 814:
```suggestion
			({punishment, affected}) = await Punishments.lock(targetUser, duration, null, ignoreAlts, userReason);
```

I would prefer this style for destructuring.

server/chat-commands/moderation.ts
```diff
@@ -1000,11 +1001,12 @@ export const commands: ChatCommands = {
 			userReason = target.substr(0, proofIndex).trim();
 		}
 
-		targetUser.popup(`|modal|${user.name} has globally banned you.${(userReason ? `\n\nReason: ${userReason}` : ``)} ${(Config.appealurl ? `\n\nIf you feel that your ban was unjustified, you can appeal:\n${Config.appealurl}` : ``)}\n\nYour ban will expire in a few days.`);
+		const {punishment, affected} = await Punishments.ban(targetUser, null, null, false, userReason);
+
+		targetUser.popup(`|modal|${user.name} has globally banned you.${(userReason ? `\n\nReason: ${userReason}` : ``)} ${(Config.appealurl ? `\n\nIf you feel that your ban was unjustified, you can appeal:\n${Config.appealurl}` : ``)}\n\nYour ban will expire ${Punishments.getPunishmentExpirationDescription(punishment)}.`);
```
Zarel, line 1006:
```suggestion
		targetUser.popup(
			`|modal|${user.name} has globally banned you.` +
			(userReason ? `\n\nReason: ${userReason}` : ``) +
			(Config.appealurl ? `\n\nIf you feel that your ban was unjustified, you can appeal:\n${Config.appealurl}` : ``) +
			`\n\nYour ban will expire ${Punishments.getPunishmentExpirationDescription(punishment)}.`
		);
```

Just bikeshedding, but we might as well make this more readable.

Zarel:
Having now read the code: this code doesn't seem to change the lack of automatic expiration.

That's a pretty big change which should probably go in a new PR, if you wanted to do that. Until then, I would not be more specific than "under one day" for global punishments, given this. The rest of the changes here are probably fine.

server/chat-commands/moderation.ts
Bardiches: Does that syntax work? Just tried it in the Node REPL and it didn't seem to work.

Bardiches:
> Having now read the code: this code doesn't seem to change the lack of automatic expiration.
> 
> That's a pretty big change which should probably go in a new PR, if you wanted to do that. Until then, I would not be more specific than "under one day" for global punishments, given this. The rest of the changes here are probably fine.

@Zarel Not sure where this change would be exactly and not a huge fan of creating separate logic for global bans. Wouldn't they ideally just press "reconnect" which would have the server check again?

server/chat-commands/moderation.ts
Zarel: Oh, sorry, my mistake.

Zarel:
> @Zarel Not sure where this change would be exactly and not a huge fan of creating separate logic for global bans. Wouldn't they ideally just press "reconnect" which would have the server check again?

No, the user needs to be offline for at least an hour for a lock/ban to expire.

AnnikaCodes:
@Bardiches I'm sorry about the lack of communication.

PS punishments now expire after their duration is up, regardless of whether the user is online or not.

AnnikaCodes:
@Bardiches, there's been a lot of changes to the punishment code but if you resolve the merge conflicts we'd be happy to accept this PR.

mia-pi-git:

PR #7522, opened 2020-10-15
Connections: Track current chat page

mia-pi-git:
Needed for #7333.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -422,7 +422,10 @@ export const commands: ChatCommands = {
 	part(target, room, user, connection) {
 		const targetRoom = target ? Rooms.search(target) : room;
 		if (!targetRoom) {
-			if (target.startsWith('view-')) return;
+			if (target.startsWith('view-')) {
+				connection.currentPages?.delete(target.slice(5));
+				return;
+			}
 			return this.errorReply(`The room '${target}' does not exist.`);
 		}
 		user.leaveRoom(targetRoom, connection);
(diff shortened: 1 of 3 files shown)


Zarel:
Didn't we agree to have a `Set` of open pages? Whatever happened to that?

mia-pi-git:
> Didn't we agree to have a `Set` of open pages? Whatever happened to that?

I don't remember that, but i can certainly do it.

mia-pi-git:
Done, @Zarel.

server/chat-commands/moderation.ts
```diff
@@ -422,7 +422,10 @@ export const commands: ChatCommands = {
 	part(target, room, user, connection) {
 		const targetRoom = target ? Rooms.search(target) : room;
 		if (!targetRoom) {
-			if (target.startsWith('view-')) return;
+			if (target.startsWith('view-')) {
+				connection.currentPages?.delete(target.slice(5));
```
Zarel, line 426:
```suggestion
				connection.currentPages?.delete(target.slice(5));
				if (!connection.currentPages?.size()) connection.currentPages = null;
```

server/chat-commands/moderation.ts
```diff
@@ -422,7 +422,11 @@ export const commands: ChatCommands = {
...
+				if (!connection.currentPages?.size()) connection.currentPages = null;
```
Zarel, line 427:
```suggestion
				if (!connection.currentPages?.size) connection.currentPages = null;
```

Ugh, I never remember if `set.size` is a function.

Zarel:
Can you rename from `currentPages` to `openPages`?

Zarel:
Looks good.

PR #7523, opened 2020-10-15
Fix spread move mechanics in ADV Doubles

KrisXV:
In ADV, spread moves that hit exactly two opponents have a spread modifier of 0.5x instead of 0.75x. Moves that have more than two targets (Explosion, Earthquake, etc.) don't get affected by spread modification at all.

### DIFF
--- a/data/mods/gen3/scripts.ts
+++ b/data/mods/gen3/scripts.ts
@@ -207,7 +207,9 @@ export const Scripts: ModdedBattleScriptsData = {
 				}
 				if (damage === this.NOT_FAIL) pokemon.moveThisTurnResult = null;
 			}
-			if (move.spreadHit) this.attrLastMove('[spread] ' + hitSlots.join(','));
+			if (move.spreadHit || move.target === "allAdjacent") {
+					this.attrLastMove('[spread] ' + hitSlots.join(','));
+				}
 		} else {
 			target = targets[0];
 			let lacksTarget = !target || target.fainted;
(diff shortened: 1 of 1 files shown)


urkerab:
`spreadHit` is also used to tell the client how to animate the move correctly. Is there another way of distinguishing between these cases, perhaps by changing it to false | number? (Also, what happens if you use Earthquake when you have no ally remaining?)

KrisXV:
> `spreadHit` is also used to tell the client how to animate the move correctly. Is there another way of distinguishing between these cases, perhaps by changing it to false | number? (Also, what happens if you use Earthquake when you have no ally remaining?)

you can just hardcode `if (move.spreadHit || [moveid array].includes(moveid))`

data/mods/gen3/scripts.ts
```diff
@@ -109,7 +190,9 @@ export const Scripts: ModdedBattleScriptsData = {
 				this.add('-notarget', pokemon);
 				return false;
 			}
-			if (targets.length > 1) move.spreadHit = true;
+			// In Generation 3, moves that hit both foes and the user's ally,
+			// like Earthquake and Explosion, don't get affected by spread modifiers
+			if (targets.length === 2) move.spreadHit = true;
```
Marty-D, line 195: Doesn't this consider foe slots that are currently empty? I'd specifically check for unfainted slots and `move.target === 'allAdjacentFoes'` to be safe, then replace the animation check below just with `targets.length > 1` for good measure.

PR #7524, opened 2020-10-15
Add alias for Shell Side Arm

Nolnis:
Adds Side Shell Arm as an alias for Shell Side Arm as per https://www.smogon.com/forums/threads/make-searching-side-shell-arm-bring-back-shell-side-arm.3671775/


DaWoblefet:
While you're at it, you should totally add ssa as an alias too.

Zarel:
Thanks!

PR #7525, opened 2020-10-15
Room events: Add a command to view categories

AnnikaCodes:
Suggested [here](https://www.smogon.com/forums/threads/add-the-ability-to-view-categories-to-roomevents.3671887/#post-8626771).

### DIFF
--- a/server/chat-plugins/room-events.ts
+++ b/server/chat-plugins/room-events.ts
@@ -476,6 +476,19 @@ export const commands: ChatCommands = {
 			room.saveSettings();
 		},
 
+		viewcategories: 'categories',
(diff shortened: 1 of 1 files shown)


Zarel:
Isn't the problem that the regular list view doesn't show categories? Shouldn't the regular list view show categories?

Zarel:
Or am I misinterpreting the request?

AnnikaCodes:
The issue is that people want a list of categories (including those with no events in them), rather than a list of events that displays the categories each event is in (and can be filtered by category).

Zarel:
Oh okay.

Zarel:
Thanks!

PR #7526, opened 2020-10-15
Support replay URLs in /join

AnnikaCodes:


server/chat-commands/moderation.ts
```diff
@@ -412,6 +412,7 @@ export const commands: ChatCommands = {
 		if (target.startsWith('http://')) target = target.slice(7);
 		if (target.startsWith('https://')) target = target.slice(8);
 		if (target.startsWith(`${Config.routes.client}/`)) target = target.slice(Config.routes.client.length + 1);
+		if (target.startsWith(`${Config.routes.replays}/`)) target = `battle-${target.slice(Config.routes.replays.length + 1)}`;
```
mia-pi-git, line 415: Config.routes doesn't start with https://, so you may want to check for that as well.

AnnikaCodes:
```ts
		if (target.startsWith('http://')) target = target.slice(7);
		if (target.startsWith('https://')) target = target.slice(8);
```
It's checked for just above.

Zarel:
Looks good, thanks!

AnnikaCodes:
@Zarel bump

Zarel:
Huh idk how I forgot about this.

PR #7529, opened 2020-10-16
Youtube: Support searching for videos

mia-pi-git:
Requested by @KrisXV for a plugin they're doing.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -212,6 +212,16 @@ export class YoutubeInterface {
 		}
 		return room;
 	}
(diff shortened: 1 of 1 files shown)


server/chat-plugins/youtube.ts
```diff
@@ -212,6 +212,16 @@ export class YoutubeInterface {
 		}
 		return room;
 	}
+	async searchVideo(name: string, limit?: number): Promise<string[]> {
```
Zarel, line 215:
```suggestion
	async searchVideo(name: string, limit?: number): Promise<string[] | undefined> {
```

Zarel:
Looks good, thanks!

PR #7530, opened 2020-10-17
Chatfilters: Fix wordfilter storage

AnnikaCodes:
This fixes the bug where public reasons are converted to replacements.


Zarel:
Thanks!

PR #7445, opened 2020-09-30
Support playing Rock Paper Scissors 

mia-pi-git:
Per https://www.smogon.com/forums/threads/rock-paper-scissors-lizard-spock-in-pms.3670873/

PR #7531, opened 2020-10-17
CommandContext: Add a method for getting roomgames

mia-pi-git:
As discussed in #7445.

### DIFF
--- a/server/chat-plugins/blackjack.ts
+++ b/server/chat-plugins/blackjack.ts
@@ -20,8 +20,6 @@ type SymbolName = 'Hearts' | 'Diamonds' | 'Clubs' | 'Spades';
 export class Blackjack extends Rooms.RoomGame {
 	room: Room;
 
-	blackjack: boolean;
-
 	roomID: RoomID;
 
 	autostart: NodeJS.Timer | null;
@@ -88,7 +86,6 @@ export class Blackjack extends Rooms.RoomGame {
 
 		this.roomID = this.room.roomid;
 		this.title = `Blackjack (${room.title})`;
-		this.blackjack = true;
 		this.state = 'signups';
 
 		this.lastMessage = '';
@@ -678,10 +675,7 @@ export const commands: ChatCommands = {
 		start(target, room, user) {
 			room = this.requireRoom();
 			this.checkCan('minigame', null, room);
-			const game = room.getGame(Blackjack);
-			if (!game || !game.blackjack) {
-				return this.errorReply("There is no game of blackjack currently ongoing in this room.");
-			}
+			const game = this.getGame(Blackjack);
 			if (game.state !== 'signups') return this.errorReply("This game of blackjack has already started.");
 
 			this.privateModAction(`The game of blackjack was started by ${user.name}.`);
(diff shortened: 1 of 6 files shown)


server/chat.ts
```diff
@@ -1344,6 +1344,15 @@ export class CommandContext extends MessageContext {
 		}
 		return this.room;
 	}
+	// eslint-disable-next-line @typescript-eslint/type-annotation-spacing
+	getGame<T extends RoomGame>(constructor: new (...args: any[]) => T) {
+		const room = this.requireRoom();
+		if (!room.game) throw new Chat.ErrorMessage(`There is no game going on in this room.`);
+		const game = room.getGame(constructor);
+		// must be a different game
+		if (!game) throw new Chat.ErrorMessage(`There is already a game of ${room.game.title} going on.`);
```
Zarel, line 1353: These error messages should probably mention what game _should_ be going on. `constructor.name` would do this, but we could add a static property like `gameName` if we wanted something nicer.

server/chat.ts
```diff
@@ -1344,6 +1344,15 @@ export class CommandContext extends MessageContext {
 		}
 		return this.room;
 	}
+	// eslint-disable-next-line @typescript-eslint/type-annotation-spacing
+	getGame<T extends RoomGame>(constructor: new (...args: any[]) => T) {
```
Zarel, line 1348: I would call this `requireGame` instead of `getGame`.

server/chat-plugins/blackjack.ts
```diff
@@ -20,8 +20,6 @@ type SymbolName = 'Hearts' | 'Diamonds' | 'Clubs' | 'Spades';
 export class Blackjack extends Rooms.RoomGame {
 	room: Room;
 
-	blackjack: boolean;
-
 	roomID: RoomID;
```
Zarel, line 23: Are these properties necessary? They should already be defined in `RoomGame`, right?

mia-pi-git: In the name of checking later: my response that i sent on discord (roughly)

Possibly? I don't maintain blackjack, so i didn't want to delete anything that wasn't directly relevant to my change.

server/chat.ts
Zarel:
```suggestion
		if (!room.game) throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this room has no game).`);
		const game = room.getGame(constructor);
		// must be a different game
		if (!game) throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this game is ${room.game.title}).`);
```

server/chat.ts
```diff
@@ -1347,10 +1347,10 @@ export class CommandContext extends MessageContext {
 	// eslint-disable-next-line @typescript-eslint/type-annotation-spacing
 	requireGame<T extends RoomGame>(constructor: new (...args: any[]) => T) {
 		const room = this.requireRoom();
-		if (!room.game) throw new Chat.ErrorMessage(`There is no game going on in this room.`);
+		if (!room.game) throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this room has no game).`);
```
Zarel, line 1350:
```suggestion
		if (!room.game) {
			throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this room has no game).`);
		}
```

server/chat.ts
```diff
@@ -1347,10 +1347,10 @@ export class CommandContext extends MessageContext {
...
 		const game = room.getGame(constructor);
 		// must be a different game
-		if (!game) throw new Chat.ErrorMessage(`There is already a game of ${room.game.title} going on.`);
+		if (!game) throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this game is ${room.game.title}).`);
```
Zarel, line 1353:
```suggestion
		if (!game) {
			throw new Chat.ErrorMessage(`This command requires a game of ${constructor.name} (this game is ${room.game.title}).`);
		}
```

Zarel:
Thanks!

PR #7534, opened 2020-10-17
Moderation: Make weekbans override 2-day roombans

AnnikaCodes:
For consistency with `/weeklock`.


Zarel:
Thanks!

Zarel:
Maybe this should be a generic "all commands should override themselves if there's more than one day of expiration difference".

PR #7535, opened 2020-10-18
COTW: Support saving channel data of winning channels

mia-pi-git:
I'm not a _huge_ fan of the approach of making the handlers held in a top-level const, but it simplifies some code, and makes it the least messy.

### DIFF
--- a/server/chat-plugins/thing-of-the-day.ts
+++ b/server/chat-plugins/thing-of-the-day.ts
@@ -1,5 +1,6 @@
 import {FS, FSPath} from '../../lib/fs';
 import {Utils} from '../../lib/utils';
+import {YouTube} from './youtube';
 
 const MINUTE = 60 * 1000;
 const PRENOM_BUMP_TIME = 2 * 60 * MINUTE;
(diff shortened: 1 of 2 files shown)


server/chat-plugins/thing-of-the-day.ts
```diff
@@ -50,7 +71,7 @@ class OtdHandler {
 	lastPrenom: number;
 	winners: AnyObject[];
 	constructor(
-		id: string, name: string, room: ChatRoom, filename: string, keys: string[], keyLabels: string[], week = false
+		id: string, name: string, room: ChatRoom, filename: string,	keys: string[], keyLabels: string[], week = false,
```
Zarel, line 74: image

This part is kinda ominous...

Zarel: Oh, it looks like you changed a space into a tab. I'm surprised that passes CI.

```suggestion
		id: string, name: string, room: ChatRoom, filename: string, keys: string[], keyLabels: string[], week = false,
```

Zarel: The trailing comma is also wrong, though; why isn't the validator picking up on that?

Zarel:
Thanks.

PR #7537, opened 2020-10-18
Add more style linting rules

Zarel:
This is a PR so everyone knows what's up, and in case people want to object to one of these.

`arrow-body-style` means

```ts
// this is wrong:
[1, 2, 3].map(a => { return a + 1; });

// this is right:
[1, 2, 3].map(a => a + 1);
```
 
`prefer-object-spread` means:

```ts
// this is wrong:
const fooCopy = Object.assign({}, foo);

// this is right:
const fooCopy = {...foo};
```

In addition, a lot of you seem to be using `{...foo}` in places you could just use `foo`. Remember, you don't need to clone an object unless you want to modify it for some reason. `JSON.stringify` etc do not modify objects.

`no-parameter-properties` means:

```ts
// this is wrong:
class Foo {
  constructor(public bar: string) {}
}

// this is right:
class Foo {
  bar: string;
  constructor(bar: string) {
    this.bar = bar;
  }
}
```

This is something we already do, but I'm turning it on just in case we have a contributor used to a different convention.

### DIFF
--- a/data/mods/ssb/abilities.ts
+++ b/data/mods/ssb/abilities.ts
@@ -638,9 +638,7 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 					Castform: ['Rain Dance', 'Sunny Day', 'Hail', 'Weather Forecast'],
 				};
 				// Store percentage of PP left for each moveSlot
-				const carryOver = pokemon.moveSlots.slice().map(m => {
-					return m.pp / m.maxpp;
-				});
+				const carryOver = pokemon.moveSlots.slice().map(m => m.pp / m.maxpp);
 				// Incase theres ever less than 4 moves
 				while (carryOver.length < 4) {
 					carryOver.push(1);
--- a/data/mods/ssb/abilities.ts
+++ b/data/mods/ssb/abilities.ts
@@ -638,7 +638,7 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 					Castform: ['Rain Dance', 'Sunny Day', 'Hail', 'Weather Forecast'],
 				};
 				// Store percentage of PP left for each moveSlot
-				const carryOver = pokemon.moveSlots.slice().map(m => m.pp / m.maxpp);
+				const carryOver = pokemon.moveSlots.map(m => m.pp / m.maxpp);
 				// Incase theres ever less than 4 moves
 				while (carryOver.length < 4) {
 					carryOver.push(1);
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -384,7 +384,7 @@ export const commands: ChatCommands = {
 		connection.popup(buffer.join("\n\n"));
 	},
 
-	autojoin(target, room, user, connection) {
+	async autojoin(target, room, user, connection) {
 		const targets = target.split(',');
 		if (targets.length > 16 || connection.inRooms.size > 1) {
 			return connection.popup("To prevent DoS attacks, you can only use /autojoin for 16 or fewer rooms, when you haven't joined any rooms yet. Please use /join for each room separately.");
@@ -400,27 +400,25 @@ export const commands: ChatCommands = {
 			})
 		);
 
-		return Promise.all(promises).then(() => {
-			connection.autojoins = autojoins.join(',');
-		});
+		await Promise.all(promises);
+		connection.autojoins = autojoins.join(',');
 	},
 
 	joim: 'join',
 	j: 'join',
-	join(target, room, user, connection) {
+	async join(target, room, user, connection) {
 		if (!target) return this.parse('/help join');
 		if (target.startsWith('http://')) target = target.slice(7);
 		if (target.startsWith('https://')) target = target.slice(8);
 		if (target.startsWith(`${Config.routes.client}/`)) target = target.slice(Config.routes.client.length + 1);
 		if (target.startsWith('psim.us/')) target = target.slice(8);
-		return user.tryJoinRoom(target as RoomID, connection).then(ret => {
-			if (ret === Rooms.RETRY_AFTER_LOGIN) {
-				connection.sendTo(
-					target as RoomID,
-					`|noinit|namerequired|The room '${target}' does not exist or requires a login to join.`
-				);
-			}
-		});
+		const ret = user.tryJoinRoom(target as RoomID, connection);
+		if (ret === Rooms.RETRY_AFTER_LOGIN) {
+			connection.sendTo(
+				target as RoomID,
+				`|noinit|namerequired|The room '${target}' does not exist or requires a login to join.`
+			);
+		}
 	},
 	joinhelp: [`/join [roomname] - Attempt to join the room [roomname].`],
 
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -872,7 +872,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(`User ${targetUser.name} is namelocked, not locked. Use /unnamelock to unnamelock them.`);
 		}
 		let reason = '';
-		if (targetUser?.locked && targetUser.locked.charAt(0) === '#') {
+		if (targetUser?.locked && targetUser.locked.startsWith('#')) {
 			reason = ` (${targetUser.locked})`;
 		}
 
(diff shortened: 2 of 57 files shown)


test/lib/dashycode.js
```diff
@@ -56,15 +56,15 @@ describe('Dashycode', function () {
...
+		}
 	});
 
 	it('should decode all codepoints accurately', function () {
-		return [...encoded.keys()].reduce((p, dashycode) => (
-			p.then(v => decode(dashycode))
-		), Promise.resolve());
+		for (const dashycode of encoded.keys()) {
+			decode(dashycode);
+		}
```
Zarel, line 67: @Kaiepi I don't like `reduce` and this is exactly why. This is 800x more readable as a for-loop.

server/roomlogs.ts
```diff
@@ -231,16 +231,13 @@ export class Roomlog {
...
-			return Promise.all([
-				roomlogExists && !newRoomlogExists ?
-					FS(roomlogPath + `/${this.roomid}`).rename(roomlogPath + `/${newID}`) :
-					undefined,
-			]);
-		});
+		]);
+		if (roomlogExists && !newRoomlogExists) {
+			await FS(roomlogPath + `/${this.roomid}`).rename(roomlogPath + `/${newID}`);
+		}
```
Zarel, line 240: @AnnikaCodes I think most of these "you don't need to use `promise.then`" situations are you. This isn't a big deal and it's fine if you forget (unfortunately there's no way to make the linter detect these), but I hope this example helps illustrate how you would do this with `async`/`await`.

server/chat-plugins/scavengers.ts
```diff
@@ -155,6 +155,7 @@ class PlayerLadder extends Ladder {
 	}
 
 	// add the different keys to the history - async for larger leaderboards
+	// FIXME: this is not what "async" means
 	softReset() {
```
Zarel, line 159: @sparkychildcharlie Making something async does not prevent it from causing freezes, or do anything at all, by itself.

If you split it into multiple Promises, though, it could help. But since this is one big Promise, it's actually no difference. Promises only change the order things happen in, they don't add multitasking by themselves.

The part that's confusing is that there are a lot of built-in functions, like everything in `fs`, which do add multitasking to Promises. But those have multi-tasking implemented inside them, it's not _just_ the Promise that makes it multitask. In PS, we have a library (process-manager) that will add multitasking to a Promise. But just creating a Promise is not enough.

I don't know if I'm explaining this well. This YouTube video might help: https://www.youtube.com/watch?v=8aGhZQkoFbQ

server/chat-commands/moderation.ts
```diff
@@ -400,27 +400,25 @@ export const commands: ChatCommands = {
...
 		if (target.startsWith('psim.us/')) target = target.slice(8);
-		return user.tryJoinRoom(target as RoomID, connection).then(ret => {
-			if (ret === Rooms.RETRY_AFTER_LOGIN) {
-				connection.sendTo(
-					target as RoomID,
-					`|noinit|namerequired|The room '${target}' does not exist or requires a login to join.`
-				);
-			}
-		});
+		const ret = user.tryJoinRoom(target as RoomID, connection);
```
xfix, line 415:
```suggestion
		const ret = await user.tryJoinRoom(target as RoomID, connection);
```

server/chat-plugins/youtube.ts
```diff
@@ -429,9 +429,8 @@ export const pages: PageTable = {
 				return this.errorReply(`There are currently no categories in the Youtube channel database.`);
 			}
 			const sorted: {[k: string]: string[]} = {};
-			const channels = Object.assign({}, YouTube.data.channels);
-			for (const id of Object.keys(channels)) {
-				const channel = channels[id];
+			const channels = YouTube.data.channels;
+			for (const [id, channel] of Object.entries(channels)) {
```
Zarel, line 433: @mia-pi-git Do you remember why you used `Object.assign` here? `channels` doesn't seem to be modified ever.

xfix (changes requested):

server/chat.ts
```diff
@@ -2016,7 +2014,7 @@ export const Chat = new class {
 		return `<div class="message"><ul class="utilichart">${buf}<li style="clear:both"></li></ul></div>`;
 	}
 	getDataMoveHTML(move: Move) {
-		if (typeof move === 'string') move = Object.assign({}, Dex.getMove(move));
+		if (typeof move === 'string') move = Dex.getMove(move);
```
Zarel, line 2017: @Spandamn Do you remember why you cloned these? It seems unnecessary.

data/mods/ssb/abilities.ts
```diff
@@ -638,9 +638,7 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 					Castform: ['Rain Dance', 'Sunny Day', 'Hail', 'Weather Forecast'],
 				};
 				// Store percentage of PP left for each moveSlot
-				const carryOver = pokemon.moveSlots.slice().map(m => {
-					return m.pp / m.maxpp;
-				});
+				const carryOver = pokemon.moveSlots.slice().map(m => m.pp / m.maxpp);
```
Zarel, line 641:
```suggestion
				const carryOver = pokemon.moveSlots.map(m => m.pp / m.maxpp);
```

`slice` is redundant with `map`

xfix (approved):

server/chat.ts
urkerab: Since `move` can't even be a string these days, the whole line seems unnecessary.

test/lib/dashycode.js
urkerab: Would use of a Promise would mean that the assertion would have run after the test had finished, so a test failure wouldn't even show up? Also, can you write something like:
```suggestion
		return [...encoded.keys()].reduce((p, dashycode) => (
			p.then(v => decode(dashycode))
		), Promise.resolve());
		for (const [dashycode] of encoded) {
			decode(dashycode);
		}
```

server/roomlogs.ts
AnnikaCodes: Yeah, I'll try to convert to `await` logic in the future.

test/lib/dashycode.js
xfix: @urkerab Promise here is misleading, `decode`/`encode` isn't even an async function.

urkerab: I didn't mean to suggest it was; I was just concerned as to whether the Promise was making the code execute at the wrong time, thus making the test useless.

Kaiepi: It's been a long time since I wrote this... basically:

> Promise here is misleading.

The point of the promises here is unrelated to `encode`/`decode`. These tests are exhaustive. IIRC, these were prone to blocking other tests when I first wrote it. With a rough benchmark (25 runs of `npx mocha -g '(en|de|trans)cod|att'` in a row), if the promises make any difference, they're within the margin of error, so I'm fine with them going.

> I was just concerned as to whether the Promise was making the code execute at the wrong time, thus making the test useless.

Mocha supports [promise-based](https://mochajs.org/#working-with-promises) asynchronous tests.

> I don't like reduce and this is exactly why. This is 800x more readable as a for-loop.

This for loop isn't equivalent to the reduce... as a for loop, that would look something like:

```javascript
let p = Promise.resolve();
for (const dashycode of encoded.keys()) {
	p = p.then(v => decode(dashycode));
}
return p;
```

To me, this doesn't read better than it did before. JavaScript isn't something I work with very often nowadays though, so I'll take your word for it.

server/chat-plugins/youtube.ts
mia-pi-git: I... don't remember doing this at all. Should be fine to remove.

test/lib/dashycode.js
xfix: @Kaiepi That's not how JavaScript promises work, making a promise doesn't spawn a new thread, this isn't Raku :).

Zarel: @Kaiepi What xfix said. Your implementation does not actually yield. It's exactly the same as my code, only slower.

If you really wanted to yield, you could do it like this:

```ts
const yield = () => new Promise<void>(resolve => {
	setTimeout(() => resolve(), 0);
});

for (const dashycode of encoded.keys()) {
	await yield();
	decode(dashycode);
}
```

Yielding is also pointless in tests; there's nothing to yield to. It's not like we're receiving messages from users or anything.

PR #7540, opened 2020-10-18
Calculate natures with 16-bit truncation

urkerab:
Latest research shows that in Pure Hackmons, Eternatus-Eternamax can have enough EVs such that its stat can overflow when calculating its nature. If with a neutral nature it would have 596 in a stat, the nature adjustment calculates 596 * 110 / 100, however this is a 16-bit calculation and the 596 * 110 overflows. (Its stat can actually go slightly higer than 596 with predictable results.)

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2011,15 +2011,18 @@ export class Battle {
 	}
 
 	natureModify(stats: StatsTable, set: PokemonSet): StatsTable {
+		// Natures are calculated with 16-bit truncation.
+		// This only affects Eternatus-Eternmax in Pure Hackmons.
+		const tr = this.trunc;
 		const nature = this.dex.getNature(set.nature);
 		let stat: keyof StatsTable;
 		if (nature.plus) {
 			stat = nature.plus;
-			stats[stat] = Math.floor(stats[stat] * 1.1);
+			stats[stat] = tr(tr(stats[stat] * 110, 16) / 100);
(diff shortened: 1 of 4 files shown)


DaWoblefet:
I only took a quick glance, but make sure this is ignored in Custom Game - we make sure to ignore overflow there.

urkerab:
Custom Game replaces `trunc` with `Math.floor`; this was implemented as part of the damage calculation.

sim/battle.ts
```diff
@@ -1994,6 +1994,39 @@ export class Battle {
...
+		if ('hp' in baseStats) {
+			const stat = baseStats['hp'];
+			modStats['hp'] = tr(tr(2 * stat + set.ivs['hp'] + tr(set.evs['hp'] / 4) + 100) * set.level / 100 + 10);
+		}
+		return this.natureModify(modStats as StatsTable, set);
+	}
+
+	natureModify(stats: StatsTable, set: PokemonSet): StatsTable {
+		// Natures are calculated with 16-bit truncation.
+		// This only affects Eternatus-Eternmax in Pure Hackmons.
```
TheImmortal, line 2015: Well it would affect other stuff too, like Shuckle in Tier Shift, so this comment isn’t completely true. I would just say this affects Pokémon in OMs only, if you really want to put a comment.

Zarel:
Thanks!

PR #7541, opened 2020-10-19
Fix Allowing Hidden Power with an undefined hpType

KrisXV:


sim/team-validator.ts
```diff
@@ -749,8 +749,8 @@ export class TeamValidator {
 		for (const moveName of set.moves) {
 			const move = dex.getMove(moveName);
 			if (move.id === 'hiddenpower' && move.type !== 'Normal') {
-				if (!set.hpType) {
-					set.hpType = move.type;
+				if (!set.hpType || this.toID(set.hpType) === 'undefined') {
```
Zarel, line 752: I believe this is entirely unnecessary. The `sim/dex` changes are all that are needed; this just slows down validation.

KrisXV: i blame sleep deprivation

PR #7536, opened 2020-10-18
Youtube: Move video data fetching to a different function

mia-pi-git:
Requested by Kris for their plugin for the Studio. Partial credit to them for re-writing some of the code.

PR #7542, opened 2020-10-19
Add The Studio's chat plugin

KrisXV:
I have a key generated locally, this just won't be ready until #7536 is merged

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2562,8 +2562,8 @@ export const commands: ChatCommands = {
 		const [link, comment] = Utils.splitFirst(target, ',');
 
 		let buf;
-		if (YoutubeInterface.linkRegex.test(link)) {
-			const YouTube = new YoutubeInterface();
+		const YouTube = new YoutubeInterface();
+		if (YouTube.linkRegex.test(link)) {
 			buf = await YouTube.generateVideoDisplay(link);
 		} else {
 			try {
--- a/server/chat-plugins/the-studio.ts
+++ b/server/chat-plugins/the-studio.ts
@@ -412,8 +412,7 @@ export const commands: ChatCommands = {
 
 	addrec: 'addrecommendation',
 	addrecommendation(target, room, user) {
-		room = this.requireRoom();
-		Recs.checkRoom(room);
+		room = this.requireRoom('thestudio' as RoomID);
 		this.checkCan('show', null, room);
 		const [artist, title, url, description, ...tags] = target.split('|').map(x => x.trim());
 		if (!(artist && title && url && description && tags?.length)) {
@@ -455,8 +454,7 @@ export const commands: ChatCommands = {
 
 	suggestrec: 'suggestrecommendation',
 	async suggestrecommendation(target, room, user) {
-		room = this.requireRoom();
-		Recs.checkRoom(room);
+		room = this.requireRoom('thestudio' as RoomID);
 		if (!target) {
 			return this.parse('/help suggestrecommendation');
 		}
--- a/server/chat-plugins/the-studio.ts
+++ b/server/chat-plugins/the-studio.ts
@@ -13,22 +13,23 @@ const DEFAULT_IMAGES = [
 	'https://lastfm.freetls.fastly.net/i/u/300x300/2a96cbd8b46e442fc41c2b86b821562f.png',
 ];
 
-interface SavedRecommendation {
+interface Recommendation {
 	artist: string;
 	title: string;
 	url: string;
 	description: string;
 	tags: string[];
+	userData: {
+		name: string;
+		avatar?: string;
+	};
+	likes: number;
 }
 
-interface SuggestedRecommendation extends SavedRecommendation {
-	submittee: string;
-}
-
-type Recommendation = {suggested: SuggestedRecommendation[], saved: SavedRecommendation[]};
+type Recommendations = {suggested: Recommendation[], saved: Recommendation[]};
 
 const lastfm: {[userid: string]: string} = JSON.parse(FS(LASTFM_DB).readIfExistsSync() || "{}");
-const recommendations: Recommendation = JSON.parse(FS(RECOMMENDATIONS).readIfExistsSync() || "{}");
+const recommendations: Recommendations = JSON.parse(FS(RECOMMENDATIONS).readIfExistsSync() || "{}");
 
 if (!recommendations.saved) recommendations.saved = [];
 if (!recommendations.suggested) recommendations.suggested = [];
@@ -72,7 +73,7 @@ export class LastFMInterface {
 			buf += `<br />`;
 			const trackName = `${track.artist?.['#text'] ? `${track.artist['#text']} - ` : ''}${track.name}`;
 			const videoIDs = await YouTube.searchVideo(trackName, 1);
-			if (!videoIDs.length) {
+			if (!videoIDs?.length) {
 				throw new Chat.ErrorMessage(`Something went wrong with the YouTube API.`);
 			}
 			buf += `<a href="https://youtu.be/${videoIDs[0]}">${Utils.escapeHTML(trackName)}</a>`;
@@ -158,17 +159,14 @@ export class LastFMInterface {
 	}
 }
 
-class Recommendations {
-	room: Room;
-	pending: AnyObject[];
-	timer: NodeJS.Timer | null;
-	lastRecommendation: AnyObject | null;
+class RecommendationsInterface {
+	lastRecommendation?: Recommendation;
+	intervalTime: number | null;
+	interval: NodeJS.Timer | null;
 
-	constructor(room: Room) {
-		this.room = room;
-		this.pending = [];
-		this.timer = null;
-		this.lastRecommendation = null;
+	constructor() {
+		this.intervalTime = null;
+		this.interval = null;
 	}
 
 	getRoom(context: CommandContext) {
@@ -179,15 +177,25 @@ class Recommendations {
 		return room;
 	}
 
-	add(artist: string, title: string, url: string, description: string, ...tags: string[]) {
+	async add(room: Room, artist: string, title: string, url: string, description: string, username: string, tags: string[],  avatar?: string) {
 		artist = artist.trim();
 		title = title.trim();
 		if (recommendations.saved.find(x => toID(x.title) === toID(title) && toID(x.artist) === toID(artist))) {
 			throw new Chat.ErrorMessage(`The song titled '${title}' by ${artist} is already recommended.`);
 		}
+		if (!/^https?:\/\//.test(url)) url = `https://${url}`;
+		if (!YouTube.validURL(url)) {
+			throw new Chat.ErrorMessage(`Please provide a valid YouTube link.`);
+		}
+		// JUST in case
 		if (!recommendations.saved) recommendations.saved = [];
-		recommendations.saved.push({artist, title, url, description, tags});
+		const rec: Recommendation = {artist, title, url, description, tags, userData: {name: username}, likes: 0};
+		if (avatar) rec.userData.avatar = avatar;
+		recommendations.saved.push(rec);
 		saveRecommendations();
+		await this.render(room, rec);
+		this.collapse(room, this.lastRecommendation);
+		this.lastRecommendation = rec;
 	}
 
 	delete(artist: string, title: string) {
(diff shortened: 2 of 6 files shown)


KrisXV:
alright, this is ready for review

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+	},
+	lastfmhelp: [
+		`/lastfm [username] - Displays the last scrobbled song for the person using the command or for [username] if provided.`,
+		`To link up your last.fm account, check out "/help registerlastfm".`,
+	],
+
+	async track(target, room, user) {
+		if (!target) return this.parse('/help track');
+		this.checkChat(target);
+		const [track, artist] = this.splitOne(target);
```
mia-pi-git, line 382:
```suggestion
		const [track, artist] = Utils.splitFirst(target, ',');
```
I'm deprecating ``this.splitOne`` soon.

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+			return this.parse(`/help suggestrecommendation`);
+		}
+
+		const cleansedTags = tags.map(x => x.trim());
+		Recs.suggest(artist, title, url, description, user.name, cleansedTags, String(user.avatar));
+		this.sendReply(`Your suggestion for '${title}' by ${artist} has been submitted.`);
+		room.sendRankedUsers(
+			`|html|${await Recs.render({artist, title, url, description, userData: {name: user.name, avatar: String(user.avatar)}, tags: cleansedTags, likes: 0}, true)}`,
+			'%'
+		);
```
mia-pi-git, line 453:
```suggestion
		const html = await Recs.render({artist, title, url, description, userData: {name: user.name, avatar: String(user.avatar)}, tags: cleansedTags, likes: 0}, true)
		room.sendRankedUsers(`|html|${html}`, '%');
```
Slightly cleaner.

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+		this.sendReply(`You liked '${title}' by ${artist}.`);
+	},
+	likerecommendationhelp: [
+		`/likerecommendation artist | title - Upvotes a recommendation for the provided artist and title.`,
+	],
+
+	viewrecs: 'viewrecommendations',
+	viewrecommendations(target, room, user) {
+		const targetRoom = Rooms.search('thestudio') || room;
+		Recs.checkRoom(targetRoom);
```
mia-pi-git, line 548: Why not just do this in one function, passing commandcontext, like youtube does?

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+	],
+};
+
+export const pages: PageTable = {
+	async recommendations(query, user, connection) {
+		const room = this.requireRoom();
+		if (!room.checkModjoin(user)) {
+			throw new Chat.ErrorMessage(`Access denied.`);
+		}
+		if (!user.inRooms.has(room.roomid)) throw new Chat.ErrorMessage(`You must be in ${room.title} to view this page.`);
```
mia-pi-git, line 575: Why block them from viewing it if they aren't in the room? Perhaps minor, but no other plugin does this, and people may want to view it without being in it, and `checkModjoin` should handle the rest of it.

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
+import {FS} from '../../lib/fs';
+import {Net} from '../../lib/net';
+import {Utils} from '../../lib/utils';
+import {YoutubeInterface} from './youtube';
+
+const YouTube = new YoutubeInterface();
```
mia-pi-git, line 6: Is there a reason to construct a new one rather than just using the exported instance?

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+		this.sendReply(LastFM.addAccountName(user.id, target.trim()));
+	},
+	registerlastfmhelp: [
+		`/registerlastfm [username] - Adds the provided [username] to the last.fm database for scrobbling.`,
+		`Usernames can only be 2-15 characters long, must start with a letter, and can only contain letters, numbers, hyphens, and underscores.`,
+	],
+
+	async lastfm(target, room, user) {
+		this.runBroadcast(true);
+		this.splitTarget(target, true);
```
mia-pi-git, line 368: Also gonna deprecate this very soon. Dunno if you wanna handle this here.

server/chat-plugins/the-studio.ts
KrisXV: bc it doesnt work as cleanly for setting `this.room` for parsing commands from a page

server/chat-plugins/the-studio.ts
KrisXV: because the commands still trigger even if the user isnt in the room (similarly for other plugins like quotes)

server/chat-plugins/the-studio.ts
KrisXV: Might as well just import the interface so i can use the static function, and theres no harm in not using the exported class from the file since i dont need the channel interfaces

server/chat-plugins/the-studio.ts
KrisXV: i dont think this should be deprecated, but if you do it, i'd rather have you rewrite it upon deprecating/refactoring since i need it to work for `this.targetUsername`

mia-pi-git: I'm deprecating targetUsername as well.

KrisXV: well I need that

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+}
+
+class RecommendationsInterface {
+	getRandomRecommendation() {
+		const recs = recommendations.saved;
+		return recs[Math.floor(Math.random() * recs.length)];
+	}
+
+	checkRoom(room: Room | null) {
+		if (!room || room.roomid !== 'thestudio') {
```
AnnikaCodes, line 181:
```suggestion
		if (room?.roomid !== 'thestudio') {
```
Also I don't necessarily think this should be in a class; it seems more like a helper for the chat commands that could go at the top level rather than something for recommendations specifically.

KrisXV: Why shouldn't it be in the Recommendation class? Recommendations are The Studio-specific, so it makes perfect sense in my opinion to put the function to check if the room that the class is used for IN the class.

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+		if (fromSuggestions) recs = recommendations.suggested;
+		return recs.find(x => toID(x.artist) === toID(artist) &&
+			toID(x.title) === toID(title) &&
+			(submitter ? toID(x.userData.name) === toID(submitter) : true));
+	}
+
+	getIndex(artist: string, title: string, submitter?: string, fromSuggestions = false) {
+		let recs = recommendations.saved;
+		if (fromSuggestions) recs = recommendations.suggested;
+		const index = recs.findIndex(x => toID(x.artist) === toID(artist) &&
```
AnnikaCodes, line 340: This is not easy to read. I'd suggest refactoring it into a full arrow function (with curly braces and a `return`).

server/chat-plugins/the-studio.ts
KrisXV: It's also used in other chat plugins similarly, so those would need changed too.

server/chat-plugins/the-studio.ts
KrisXV: the new eslint rules forbade that

KrisXV: it was originally like that, there's even a commit by me to change it to this because of it

KrisXV: https://github.com/smogon/pokemon-showdown/pull/7542/commits/fd8777b096a2790495b3d5249a176e9cd966398d

server/chat.ts
```diff
@@ -515,6 +515,7 @@ export class CommandContext extends MessageContext {
 			}).catch(err => {
 				if (err.name?.endsWith('ErrorMessage')) {
 					this.errorReply(err.message);
+					this.update();
```
AnnikaCodes, line 518: I feel like I would expect `CommandContext#errorReply` to update. I suppose it's usually used at the end of a command so it doesn't matter but it seems odd. Is updating the room too much a performance issue?

server/chat-plugins/the-studio.ts
AnnikaCodes: Hm, okay. I thought it would be permissible if there were multiple return statements (e.g. `if (!submitter) return true;`).

server/chat.ts
KrisXV: when i tested locally, broadcasting `!lastfm`, `!track`, etc. would not display the error message until I updated the room. I really didn't bother to look into it because I mostly did it to stop complaints of errors 'lagging', but I'd rather look into it more in-depth in another commit.

server/chat-plugins/the-studio.ts
AnnikaCodes: I'm fine with it either way; there are chat plugins that put these things at the top level too. that was more of a personal nitpick, i should've made that more clear, mb

server/chat-plugins/the-studio.ts
Zarel: Your suggestion breaks line length rules.

```suggestion
		const html = await Recs.render({
			artist, title, url, description,
			userData: {name: user.name, avatar: String(user.avatar)},
			tags: cleansedTags, likes: 0,
		}, true)
		room.sendRankedUsers(`|html|${html}`, '%');
```

server/chat-plugins/the-studio.ts
Zarel: To be clear, I was the one who wanted to deprecate `splitTarget` and `targetUsername`.

There's no need for you to worry about that, though. The migration path will come when it comes, and will not lose features.

server/chat-plugins/the-studio.ts
Zarel: This should probably just be an argument for `requireRoom` to require a specific roomid.

Actually doesn't `requireRoom` already have that feature? Why does it need to be duplicated?

server/chat-plugins/the-studio.ts
Zarel: You can still make it more readable.

server/chat-plugins/the-studio.ts
```diff
@@ -0,0 +1,613 @@
...
+				toID(x.title) === toID(title) &&
+				(submitter ? toID(x.userData.name) === toID(submitter) : true));
```
Zarel, line 342:
```suggestion
		const index = recs.findIndex(rec => (
			toID(rec.artist) === toID(artist) &&
			toID(rec.title) === toID(title) &&
			(submitter ? toID(rec.userData.name) === toID(submitter) : true)
		));
```

Zarel: (And similar for `get`)

server/chat-plugins/the-studio.ts
Zarel: Actually maybe don't. I forgot that `splitOne` has other features like defaulting to `,` and trimming.

Zarel: @mia-pi-git In general, there's no need to comment on PRs to remove things _before_ they're deprecated. The deprecation PR itself explains exactly what approach you want to change to and why; without it, these sorts of change suggestions are unnecessary distractions.

server/chat.ts
```diff
@@ -515,6 +515,7 @@ export class CommandContext extends MessageContext {
...
 					return false;
 				}
 				if (err.name.endsWith('Interruption')) {
```
Zarel, line 521:
```suggestion
				if (err.name.endsWith('Interruption')) {
					this.update();
```

server/chat.ts
Zarel: This is the correct fix.

@AnnikaCodes The individual `reply`/`add`/`send` commands don't update, because it's more efficient (and saves bandwidth) to send one update with everything at once, than to send things one at a time.

Calling `update` here, at the end of the chat command handler, is the correct place to do it.

AnnikaCodes: Alright, that makes sense.

KrisXV:
i did everything btw

server/chat-commands/info.ts
```diff
@@ -2562,8 +2562,8 @@ export const commands: ChatCommands = {
 		const [link, comment] = Utils.splitFirst(target, ',');
 
 		let buf;
-		if (YoutubeInterface.linkRegex.test(link)) {
-			const YouTube = new YoutubeInterface();
+		const YouTube = new YoutubeInterface();
+		if (YouTube.linkRegex.test(link)) {
```
Zarel, line 2566: this seems bad, I thought we fixed this?

Zarel: @mia-pi-git?

PR #7543, opened 2020-10-19
Reject HP undefined

xInstruct:


sim/team-validator.ts
```diff
@@ -456,7 +456,9 @@ export class TeamValidator {
 		if (set.happiness !== undefined && isNaN(set.happiness)) {
 			problems.push(`${name} has an invalid happiness value.`);
 		}
-		if (set.hpType && (!dex.getType(set.hpType).exists || ['normal', 'fairy'].includes(toID(set.hpType)))) {
+		if (!dex.getType(set.hpType).exists) {
+			set.hpType = dex.getHiddenPower(set.ivs).type || '';
```
HoeenCoder, line 460:
```suggestion
		if (!dex.getType(set.hpType || '').exists) {
			set.hpType = dex.getHiddenPower(set.ivs).type;
```

Zarel:
Shouldn't this be auto-calculated later in the validation process? What specific bug is this intended to fix?

DaWoblefet:
Presumably the one that's been crashing for 2 months where if you bring a G-Max Pokemon with "Hidden Power" (not Hidden Power Ice, etc.) it crashes upon clicking Hidden Power because the typing is undefined. You can also currently do like Hidden Power Ice!?!?!?!?!?!?! in the teambuilder and crash PS that way too.

xInstruct:
Capture
this one

Zarel:
It looks like that crash is already fixed by #7541. This PR doesn't fix it, but instead just breaks the validation further.

HoeenCoder:
#7541 was patched in last night and were still having issues though

PR #7544, opened 2020-10-19
Validator: Fix crash in stabmons move legality

mia-pi-git:
This seems like the right way to fix this? For some reason, `species.types` is undefined.

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -955,7 +955,7 @@ export const Formats: {[k: string]: FormatData} = {
 				if (species.forme || species.otherFormes) {
 					const baseSpecies = dex.getSpecies(species.baseSpecies);
 					const originalForme = dex.getSpecies(species.changesFrom || species.name);
-					types = originalForme.types;
+					types = originalForme.types || [];
 					if (baseSpecies.otherFormes) {
 						for (const formeName of baseSpecies.otherFormes) {
 							if (baseSpecies.prevo) {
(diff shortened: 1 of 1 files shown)


data/rulesets.ts
```diff
@@ -955,7 +955,7 @@ export const Formats: {[k: string]: FormatData} = {
 				if (species.forme || species.otherFormes) {
 					const baseSpecies = dex.getSpecies(species.baseSpecies);
 					const originalForme = dex.getSpecies(species.changesFrom || species.name);
-					types = originalForme.types;
+					types = originalForme.types || [];
```
Zarel, line 958: This should never happen...

Zarel:
Which crash is this?

Zarel:
Do you have a repro?

mia-pi-git:
> Do you have a repro?

PM'd on Discord.

data/rulesets.ts
```diff
@@ -964,7 +964,7 @@ export const Formats: {[k: string]: FormatData} = {
 							}
 							const forme = dex.getSpecies(formeName);
 							if (forme.changesFrom === originalForme.name && !forme.battleOnly) {
-								types = types.concat(forme.types);
+								types.push(...forme.types);
```
Zarel, line 967: This is a bug!!!!

Zarel: (`species.types` is not mutable)

Zarel: This would add these types after every validation.

So after three validations Meloetta would have types `Normal/Psychic/Normal/Fighting/Normal/Fighting/Normal/Fighting`.

PR #7546, opened 2020-10-20
Add alias for National Dex

Distrib-ps:
https://www.smogon.com/forums/threads/make-the-tier-commands-for-national-dex-formats-more-consistant.3672067/


Zarel:
Looks good; thanks!

PR #7549, opened 2020-10-20
CommandContext: Make requireRoom support specifying a room

mia-pi-git:
This was discussed in #7542, i believe, but since it doesn't exist i thought it would be useful to support this.

### DIFF
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -144,10 +144,6 @@ if (triviaData.questions.some(q => !('type' in q))) {
 /** from:to Map */
 export const pendingAltMerges = new Map<ID, ID>();
 
-function isTriviaRoom(room: Room) {
-	return room.roomid === 'trivia';
-}
-
 function getTriviaGame(room: Room | null) {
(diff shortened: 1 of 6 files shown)


server/chat-plugins/trivia.ts
```diff
@@ -1490,8 +1486,7 @@ const triviaCommands: ChatCommands = {
 	new(target, room, user, connection, cmd) {
 		const randomizeQuestionOrder = !cmd.includes('sorted');
 
-		room = this.requireRoom();
-		if (!isTriviaRoom(room)) return this.errorReply(this.tr`This command can only be used in the Trivia room.`);
+		room = this.requireRoom(`trivia` as RoomID);
```
AnnikaCodes, line 1489:
```suggestion
		room = this.requireRoom('trivia' as RoomID);
```

Zarel:
Maybe don't require the cast to `RoomID`. It's nearly always a string literal passed directly, anyway.

Zarel:
Hm. I could see arguments for both approaches. I think this is actually more consistent. Thanks!

PR #7550, opened 2020-10-21
Youtube: Persist interval time

mia-pi-git:
Requested by Youtube ROs.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -41,6 +41,7 @@ interface VideoData {
 interface ChannelData {
 	channels: {[k: string]: ChannelEntry};
 	categories: string[];
+	intervalTime?: number;
 }
 
 function loadData() {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/youtube.ts
```diff
@@ -252,6 +255,23 @@ export class YoutubeInterface {
...
+				if (!room) return; // do nothing if the room doesn't exist anymore
+				const res = await YouTube.randChannel();
+				room.add(`|html|<div class="infobox">${res}</div>`).update();
+			})();
+		}, interval);
+		return this.interval;
+	}
 }
 
 export const YouTube = new YoutubeInterface(channelData);
```
Zarel, line 277:
```suggestion
export const YouTube = new YoutubeInterface(channelData);

export function destroy() {
	if (YouTube.interval) clearInterval(YouTube.interval);
}
```

I don't see a cleanup for the interval. Do you need something like this?

Zarel:
Thanks.

PR #7551, opened 2020-10-21
Smogon master

hsahovic:


hsahovic:
This PR was opened by mistake - I wanted to target `hsahovic:master` instead of `smogon:master`, but GitHub defaults got me.

Zarel:
Interesting fork, though.

I don't think it needs to be a fork. `Config.fakeladder` should be updated to disable all the rate limits you care about, and we can add a new `Config.fakelogin` to disable the login handshake as well, if you'd like. I think that'd probably be easier than having you maintain such a simple fork.

If you were to pull request those changes, I'd love to accept them into master.

hsahovic:
Thanks! That sounds like a good idea. I'll look into it sometime this week :) Are there other things you think could improve the overall performance (in terms of games per second)?

Zarel:
You may wish to use the sim stream API directly:

https://github.com/smogon/pokemon-showdown/tree/master/sim

Not having the server overhead may help.

PR #7554, opened 2020-10-21
Moderation: visualize mute times with less precision

mia-pi-git:
This part of this change seems to annoy everyone, and it's not useful for the main chat.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -586,7 +586,10 @@ export const commands: ChatCommands = {
 		if (targetUser.id in room.users) {
 			targetUser.popup(`|modal|${user.name} has muted you in ${room.roomid} for ${Chat.toDurationString(muteDuration)}. ${target}`);
 		}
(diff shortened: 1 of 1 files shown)


Zarel:
This was a bug which will be fixed soon and fixed everywhere, not just in mute messages.

Zarel:
Did... did someone hotpatch my change? I was hoping I could fix it before it got hotpatched.

mia-pi-git:
It was hotpatched, yes.

Zarel:
Can someone just fix it properly? :[

Zarel:
Ugh I'll fix it in the web editor I guess.

mia-pi-git:
> Can someone just fix it properly? :[

Is this not the proper way..? The problem is only here.

Zarel:
Note to self, fixing it on my iPad is easier than fixing it in the web editor.

Zarel:
Also note to self, fixing it in a monkeypatch is easier than fixing it in the web editor.

PR #7555, opened 2020-10-21
Fix G-Max moves not showing up in /ms Gmax

PartMan7:
Happened because G-Max moves had move.isNonstandard as 'Gigantamax'.


PartMan7:
Low priority bump.

Zarel:
This looks probably correct to me. Thanks!

PR #7557, opened 2020-10-21
Make /clearstatus log to Staff room

AnnikaCodes:
`/forcerename` and other global punishments do this.


AnnikaCodes:
@Zarel bump

Zarel:
Thanks!

PR #7558, opened 2020-10-23
Support banning users from using groupchats

AnnikaCodes:


AnnikaCodes:
@Zarel bump

Zarel:
Is this a global punishment? I think the main thing I'd be worried about is that PS's punishment system doesn't support having multiple concurrent global punishments. So we wouldn't be able to handle having simultaneously a long groupchat ban and also a short lock.

server/punishments.ts
```diff
@@ -923,6 +943,101 @@ export const Punishments = new class {
...
+					Punishments.bannedGroupchatParticipants[targetRoom.roomid] = new Set(
+						// Room#users is a UserTable where the keys are IDs,
+						// but typed as strings so that they can be used as object keys.
+						Object.keys(targetRoom.users) as ID[]
+					);
+				}
+			}
+		}
+
+		Punishments.roomPunish("groupchat", user, punishment);
```
AnnikaCodes, line 978: Yes, I already scoped it to a fake room.

Zarel: Oh wow, nice.

Zarel: I should probably have finished reading before posting preliminary thoughts.

Zarel:
I wonder if this should just be a generally "untrustworthy" tag?

We already have "probation" which kind of like this? Although not exactly. But we might want to _revise_ probation to be this?

On the other hand, different users arguably can be trusted with different things, and maybe there isn't all that much overlap? It might be worth (at least for now) leaving them separate and seeing how much overlap there ends up being.

If you want to keep it separate, though, now that I've read through the code (sorry!), I think the code for it is good.

Do you think it would be better separated or as a combined status/punishment?

AnnikaCodes:
By "separate" do you mean separate from probation?

My understanding of probation is that it completely prevents a user from being autoconfirmed, which would block them from using groupchats, but also prevent them from talking in most chatrooms due to modchat. This seems rather heavy handed; like you said, different users can be trusted with different things, and I think that removing a user's autoconfirmed status for a week or a month is overkill for someone who doesn't moderate their groupchats well.

Also, if we were to have a general untrustworthy flag, I'd prefer a punishment-like system; currently, probation can only be added/removed by Upper Staff, since it's a standing just like permalock, but I think that the whole point of this command was to give policy drivers and mods, who seem to do most of the day-to-day moderation, more tools to deal with groupchats.

I do like the concept of an untrustworthy flag, but I'm not sure what its real-world application would be.

Zarel:
Yeah, I suppose we should punt on that.

PR #7567, opened 2020-10-24
Info: Clean up YouTube import

mia-pi-git:


server/chat-commands/info.ts
```diff
@@ -2511,7 +2511,6 @@ export const commands: ChatCommands = {
 			buf = Utils.html`<img src="${request.link}" width="${width}" height="${height}" />`;
 			if (resized) buf += Utils.html`<br /><a href="${request.link}" target="_blank">full-size image</a>`;
 		} else {
-			const YouTube = new YoutubeInterface();
```
Zarel, line 2514: Why not:

```ts
const YouTube = Chat.plugins.youtube.YouTube;
```

?

mia-pi-git:
Irrelevant since I pushed this a while ago.

PR #7570, opened 2020-10-24
Fix IPTools' handling of IP addresses

Kaiepi:
The IPv4 address implementation `IPTools` has is broken. Presentation-formatted IP addresses produced by `IPTools` are stored persistently in `config/hosts.csv` and `config/proxies.csv`, which get involved when resolving hosts pertaining to IP addresses. This means there is no guarantee that anyone's host on PS is accurate as things stand now.

This makes the first steps to fixing `IPTools`' IP address handling, but is incomplete in its current state, does nothing to mitigate the damage that's been done, and will break things further if merged now. How best to complete this merits some discussion beforehand.

### DIFF
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -109,29 +109,35 @@ export const IPTools = new class {
 
 	ipToNumber(ip: string) {
 		ip = ip.trim();
-		if (ip.includes(':') && !ip.includes('.')) {
-			// IPv6
-			return -1;
-		}
 		if (ip.startsWith('::ffff:')) ip = ip.slice(7);
 		else if (ip.startsWith('::')) ip = ip.slice(2);
-		let num = 0;
-		const parts = ip.split('.');
-		for (const part of parts) {
-			num *= 256;
-			num += parseInt(part);
+
+		const octets = ip.split('.').map((o) => parseInt(o)).filter((o) => !isNaN(o) && o >= 0 && o <= 0xFF);
+		if (octets.length !== 4) return -1;
+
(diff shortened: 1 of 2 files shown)


Kaiepi:
In order for this to be complete enough to merge, `IPTools#numberToIP` needs to validate IP addresses passed, and there needs to be a way to notify server administrators when `config/hosts.csv` and `config/proxies.csv` contain invalid IP presentations. How would the latter best be handled?

AnnikaCodes:
I'm a bit confused as to what this fixes.

Your comments about how "there is no guarantee that anyone's host on PS is accurate as things stand now" make it sound like you're concerned about the CSV files containing poor mappings of IP addresses to hosts, but the actual code here seems to address a concern about poor validation, which is certainly good to address, but wouldn't actually affect people's hosts — no one can log in from an invalid IP address, so mappings of invalid IP ranges wouldn't affect people's hosts, although they are still malformed data (and could lead server administrators to believe that they've marked IPs as proxy/mobile/residential when they actually made a typo).

`config/hosts.csv` and `config/proxies.csv` aren't supposed to be manually edited; the Hosts chatplugin provides tools to modify the data stored in those files. It's possible that the validation there is _incomplete_, but it does _exist_.
(It's also worth noting that we had several invalid IP ranges stored before the Hosts chatplugin existed, and until they were removed they didn't seem to cause any issues.)

Kaiepi:
I have zero excuse to be spewing complete fucking 𝑮𝑨𝑹𝑩𝑨𝑮𝑬 like this after all this time. Utterly fucking shameful in every way possible.

PR #7572, opened 2020-10-25
Build: Delete built files with no .ts counterpart

mia-pi-git:
This is to ensure that files that don't exist in the TS directory don't exist in the built directory, to avoid crashes that pull data from old files.
Not sure if this is the best solution, but it seemed the best way to clear out deleted files as quickly and simply as possible.

### DIFF
--- a/build
+++ b/build
@@ -92,12 +92,13 @@ function needsSucrase(source, dest, path = '') {
 			const files = fs.readdirSync(source + path);
 			for (const oldFile of oldFiles) {
 				if (!oldFile.endsWith('.js') || files.includes(oldFile)) continue;
-				if (!files.includes(oldFile.slice(0, -2) + 'ts')) {
+				if (!files.includes(oldFile.slice(0, -2) + 'ts') || force) {
 					fs.unlink(dest + path + '/' + oldFile, function (err) {
 						if (err) throw err;
 					});
 				}
 			}
+			if (force) return true;
 			for (const file of files) {
 				if (needsSucrase(source, dest, path + '/' + file)) {
 					return true;
@@ -113,7 +114,7 @@ function needsSucrase(source, dest, path = '') {
 
 function sucrase(src, out, opts) {
 	try {
-		if (!force && src !== './config' && !needsSucrase(src, out)) {
+		if (src !== './config' && !needsSucrase(src, out)) {
 			return false;
 		}
 	} catch (e) {}
--- a/build
+++ b/build
@@ -86,7 +86,15 @@ function needsSucrase(source, dest, path = '') {
 	if (!path.includes('.')) {
 		// probably dir
 		try {
+			const oldFiles = fs.readdirSync(dest + path);
 			const files = fs.readdirSync(source + path);
+			for (const oldFile of oldFiles) {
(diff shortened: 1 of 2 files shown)


mia-pi-git:
....rewriting it in another way it is, then.

build
```diff
@@ -100,12 +100,30 @@ function needsSucrase(source, dest, path = '') {
...
+	} catch (e) {}
+}
+
 function sucrase(src, out, opts) {
 	try {
 		if (!force && src !== './config' && !needsSucrase(src, out)) {
 			return false;
 		}
 	} catch (e) {}
+	clearPath(out, src); // empty the dir to get rid of built js files with no ts couterpart
```
urkerab, line 126: As of Node.js 12.10.0, fs.rmdirSync supports a `recursive` option, so you can use:

```suggestion
	fs.rmdirSync(out, { recursive: true });
```

mia-pi-git: I thought about this, but it presents problems with things like tools/, where built stuff is in the same dir as non-built stuff. Additionally, not sure if windows supports this (we had trouble before) so I thought it might be wiser to be a bit more fine-tuned with it.

AnnikaCodes: yes, `FS#rmdir` only really works on UNIX-like systems. https://github.com/smogon/pokemon-showdown/pull/7187#issuecomment-671115093

urkerab: Ah, fair enough.

Zarel:
This should probably go inside `needsSucrase`, instead of iterating every file _twice_...

build
```diff
@@ -52,9 +52,7 @@ function copyOverDataJSON(file) {
 	var dest = './.data-dist/' + file;
 	fs.readFile(source, function (err, text) {
 		if (err) throw err;
-		fs.writeFile(dest, text, function (err) {
-			if (err) throw err;
-		});
+		fs.writeFileSync(dest, text);
```
Zarel, line 55: <_<

build
```diff
@@ -86,7 +84,14 @@ function needsSucrase(source, dest, path = '') {
 	if (!path.includes('.')) {
 		// probably dir
 		try {
+			const oldFiles = fs.readdirSync(dest + path);
 			const files = fs.readdirSync(source + path);
+			for (const oldFile of oldFiles) {
+				if (!oldFile.endsWith('.js') || files.includes(oldFile)) continue;
+				if (!files.includes(oldFile.slice(0, -2) + 'ts')) {
+					fs.unlinkSync(dest + path + '/' + oldFile);
```
Zarel, line 92: Does not need to be sync.

build
```diff
@@ -53,7 +53,7 @@ function copyOverDataJSON(file) {
 	fs.readFile(source, function (err, text) {
 		if (err) throw err;
 		fs.writeFile(dest, text, function (err) {
-			if (err) throw err;
+			throw err;
```
Zarel, line 56:
```suggestion
			if (err) throw err;
```

Zarel:
There is unfortunately not much I can do about this if we can't figure out the build error. Drafting for now.

mia-pi-git:
Closing since this was basically done.

PR #7573, opened 2020-10-25
Fix Crown Tundra Pokedex validation

LegoFigure11:
This prevents the ruleTable from allowing Pokemon from outside of the Crown Tundra dex, which fixes the bug of every single BSS-legal mon being allowed in Tundra Tourney (See [the official site](https://battle.pokemon-home.com/regulation/000000106/el) for more)


data/rulesets.ts
```diff
@@ -310,7 +310,7 @@ export const Formats: {[k: string]: FormatData} = {
 			];
 			const species = this.dex.getSpecies(set.species || set.name);
 			if (!tundraDex.includes(species.baseSpecies) && !tundraDex.includes(species.name) &&
-				this.ruleTable.isBannedSpecies(species)) {
+				!this.ruleTable.has('+' + species.id)) {
 				return [`${species.baseSpecies} is not in the Crown Tundra Pokédex.`];
 			}
```
KrisXV, line 315: this isnt actually how the ruletable works, i would just remove the third check entirely

LegoFigure11: Ah, I just copied from all the other pokedex rulesets. Should I remove it from those too?

KrisXV: The other rulesets aren't relevant to to this pull request, I would just open a separate PR to properly check validation after this one

LegoFigure11: Cheers. I'm not sure I'm familiar enough with the ruletable to "properly check validation" in all the other rulesets, but I removed the third check from this for now!

PR #7575, opened 2020-10-26
Make Eerie Spell's secondary effect secondary

urkerab:
As researched by Zaggyo, Eerie Spell's PP deduction is a secondary effect for the purposes of Sheer Force and Shield Dust.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -4006,17 +4006,19 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 5,
 		priority: 0,
 		flags: {protect: 1, mirror: 1, sound: 1, authentic: 1},
-		onHit(target) {
-			if (!target.hp) return;
-			const move = target.lastMove;
-			if (!move || move.isZ || move.isMax) return;
+		secondary: {
+			chance: 100,
+			onHit(target) {
+				if (!target.hp) return;
+				const move = target.lastMove;
+				if (!move || move.isZ || move.isMax) return;
 
-			const ppDeducted = target.deductPP(move.id, 3);
-			if (!ppDeducted) return;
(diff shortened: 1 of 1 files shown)


Marty-D:
Shouldn't this be using `onAfterHit` so a Rowap KO won't let it activate? Do secondaries support that?

urkerab:
Unfortunately I don't think they do; secondaries currently run before damaging hit events. The best idea I have is to have the secondary add a volatile which runs later?

Zarel:
Thanks! I assume this is better than not doing it.

Zarel:
Re: Rowap order, is the problem that the games run some secondaries, then some primaries, then other secondaries?

Marty-D:
I was going to suggest the Clangorous Soulblaze strategy of using onAfterHit normally, having an empty `secondary` and then deleting the onAfterHit during Sheer Force and Shield Dust.

The problem is they randomly decided to have moves with secondaries that aren't percentage based be affected by Sheer Force and Shield Dust starting last gen (https://github.com/smogon/pokemon-showdown/issues/3339). Most secondaries happen before the target faints, some happen after that, and some happen after all hits of the move (which is your classic "Sheer Force blocks these effects if the move was boosted" section). Eerie Spell, apparently, is in a custom table saying "this is affected by Sheer Force" like those other moves, but also its secondary normally happens after the target faints while Rowap happens before.

Zarel:
That's a pretty good idea, actually. Although I'd suggest having the `onAfterHit` handler itself just check if the move still has secondaries, rather than hardcoding the deletion into Sheer Force and Shield Dust.

Marty-D:
It's non-trivial to do that because of King's Rock and friends, and even if you filter out volatileStatus flinches then it stays truthy anyway.

PR #7576, opened 2020-10-26
Update Lunar Dance to Gen 8 mechanics

urkerab:
As reported [here](https://www.smogon.com/forums/posts/8637536).

### DIFF
--- a/data/mods/gen7/moves.ts
+++ b/data/mods/gen7/moves.ts
@@ -545,6 +545,24 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		isNonstandard: null,
 	},
+	lunardance: {
+		inherit: true,
+		condition: {
+			duration: 2,
+			onSwitchInPriority: 1,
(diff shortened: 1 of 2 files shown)


data/mods/gen7/moves.ts
```diff
@@ -545,6 +545,24 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+			duration: 2,
+			onSwitchInPriority: 1,
+			onSwitchIn(target) {
+				if (!target.fainted) {
+					target.heal(target.maxhp);
+					target.setStatus('');
+					for (const moveSlot of target.moveSlots) {
+						moveSlot.pp = moveSlot.maxpp;
+					}
+					this.add('-heal', target, target.getHealth, '[from] move: Healing Wish');
```
DaWoblefet, line 560: Probably copied too much from the Healing Wish implementation ;)

urkerab: Indeed, and on the next line too. Whoops!

PR #7577, opened 2020-10-26
sendReplyBox: Fix interaction with hidelines

sparkychildcharlie:
This makes non-broadcasted reply infoboxes come from user ``~`` (This is questionable - shouldn't it be & by now since the beautiful admin sign ~ is replaced with the super ugly &) instead of the user themselves.  This makes it so that un-broadcasted information searches won't be hidden by any /hidetext, hidelines command.  This is rather important in game rooms that rely on /ds for searches.  If someone accidentally leaks something that might be an answer and the room's staff has to quickly suppress their text, losing progress of all their previous searches would be detrimental to their progress in the current game.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -770,7 +770,7 @@ export class CommandContext extends MessageContext {
 		this.add(`|html|<div class="infobox">${htmlContent}</div>`);
 	}
 	sendReplyBox(htmlContent: string) {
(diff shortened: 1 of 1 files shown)


sparkychildcharlie:
just a pet peeve when a lot of my /ds searches got wiped on my own side server when I suppressed my own messages.... and I had to start all over again.

mia-pi-git:
This is intentional, though. I find it more useful to be able to hidetext big boxes, personally. That's part of the reason i made this interaction.

Zarel:
Thanks!

PR #7580, opened 2020-10-26
Fix Slow Start's interaction with Neutralizing Gas

KrisXV:
I don't know if this is the best approach to this but it's currently the only way I've tried that works.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2268,6 +2268,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		// TODO Will abilities that already started start again? (Intimidate seems like a good test case)
 		onPreStart(pokemon) {
 			this.add('-ability', pokemon, 'Neutralizing Gas');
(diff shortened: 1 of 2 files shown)


urkerab:
This bug seems to affects other Abilities too (I tested Flash Fire but there are probably others such as Unburden). Maybe you need to look for Abilities affected by NG and send them the End event?

DaWoblefet:
Yes, there are currently several skipped tests in the repo for those sorts of persistent effects like Unburden, Slow Start, Illusion, and Gluttony that don't play well with NGas right now.

Zarel:
The only other thing I can think of is to make NGas's volatile do nothing if the ability isn't active.

That might actually be simpler, actually.

Zarel:
But I believe all abilities should be implemented that way: their volatile does nothing, but the ability itself checks if the volatile is there and uses that to determine how it works.

PR #7581, opened 2020-10-26
Ban Japanese Gen 1 Events in Int Formats: Part 2

PlagueVonKarma:
In reference to this, which I assume can be closed now: https://github.com/smogon/pokemon-showdown/pull/7516

Zarel asked for a PR marking the JP Events so he can take it from there? I've marked it with comments as so: "//JP Event". A simple Ctrl+F should have them pop up.

### DIFF
--- a/data/mods/gen2/learnsets.ts
+++ b/data/mods/gen2/learnsets.ts
@@ -998,7 +998,7 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 			whirlwind: ["1M"],
 		},
 		eventData: [
-			{generation: 1, level: 20, moves: ["growl", "leer", "furyattack", "payday"]},
+			{generation: 1, level: 20, moves: ["growl", "leer", "furyattack", "payday"]}, //JP Event
 		],
 	},
 	ekans: {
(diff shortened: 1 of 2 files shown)


Zarel:
Thanks! Yeah this is why I said the format isn't important; it's not too hard for me to convert them to the correct format.

PlagueVonKarma:
> Thanks! Yeah this is why I said the format isn't important; it's not too hard for me to convert them to the correct format.

I forgot to note that Gen 2 can't have these specific events either, as GSC has the same "can't trade internationally without save game corruption" issue as RBY. The moves were never re-released in Gen 2 either, not even Flying Pikachu.

Zarel:
It's fine, the code I committed covers both. Feel free to tell me if you catch any bugs with it, though.

PlagueVonKarma:
> It's fine, the code I committed covers both. Feel free to tell me if you catch any bugs with it, though.

I think the message could be a bit misleading with Gen 2, as it implies that you're playing Yellow. I gave it a look and it was a bit out there. You could say "X can't use moves from Japanese Gen 1 games. This format simulates the international Crystal Version, which can't trade with Japanese games.".

On a side note, the message itself could possibly receive a clarity enhancement for Gen 1, as it doesn't exclude Blue or JP Yellow. Something like "X can't use moves from Japanese Gen 1 games. This format simulates the international Yellow Version, which can't trade with Japanese games.". 

Not sure if it's necessary, though.

PR #7586, opened 2020-10-27
The Studio: Handle Net errors better

mia-pi-git:


server/chat-plugins/the-studio.ts
```diff
@@ -303,7 +313,10 @@ class RecommendationsInterface {
 	}
 
 	async render(rec: Recommendation, suggested = false) {
-		const videoInfo = await YouTube.getVideoData(rec.url);
+		let videoInfo = null;
+		try {
+			videoInfo = await YouTube.getVideoData(rec.url);
+		} catch (e) {}
```
KrisXV, line 319: I would prefer this gives an actual error reply

mia-pi-git: Ok - you previously didn't do that so i assumed it wasn't necessary. One sec.

Zarel:
Why do `YouTube.searchVideo` and `YouTube.getVideoData` crash at all? It seems like these should probably be better handled internally?

PR #7592, opened 2020-10-27
Indicate pending promises in /eval

xfix:
After this change running a command like this:

```
/eval new Promise(resolve => { setTimeout(() => resolve(42), 1000) })
```

will print

```
<< Promise pending
```

before saying

```
<< Promise -> 42
```

This will make it clearer that the promise is resolving.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -924,6 +924,7 @@ export const commands: ChatCommands = {
 			if (result?.then) {
 				uhtmlId = `eval-${room.nextGameNumber()}`;
 				this.sendReply(`|uhtml|${uhtmlId}|${generateHTML('<', 'Promise pending')}`);
+				this.update();
 				result = `Promise -> ${Utils.visualize(await result)}`;
 				this.sendReply(`|uhtmlchange|${uhtmlId}|${generateHTML('<', result)}`);
 			} else {
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -906,8 +906,14 @@ export const commands: ChatCommands = {
 			this.broadcasting = true;
 			this.broadcastToRoom = true;
 		}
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -906,8 +906,14 @@ export const commands: ChatCommands = {
...
 		}
-		this.sendReply(`|html|<table border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">&gt;&gt;&nbsp;</td><td>${Chat.getReadmoreCodeBlock(target)}</td></tr><table>`);
+		const generateHTML = (direction: string, contents: string) => (
+			`<table border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">` +
+				Utils.escapeHTML(direction).repeat(2) +
+				`&nbsp;</td><td>${Chat.getReadmoreCodeBlock(contents)}</td></tr><table>`
+		);
+		this.sendReply(`|html|${generateHTML('>', target)}`);
 		logRoom?.roomlog(`>> ${target}`);
+		let uhtmlId: string | null = null;
```
mia-pi-git, line 916: Is there any reason you use `null` and `=== null` here when you could just use `let uthmlId: string;` and `!uhtmlId ? ... : ...`?

xfix: Not really, fixing.

mia-pi-git (approved):
Looking good.

server/chat-commands/admin.ts
```diff
@@ -916,15 +922,19 @@ export const commands: ChatCommands = {
 			/* eslint-enable no-eval, @typescript-eslint/no-unused-vars */
 
 			if (result?.then) {
+				uhtmlId = `eval-${room.nextGameNumber()}`;
+				this.sendReply(`|uhtml|${uhtmlId}|${generateHTML('<', 'Promise pending')}`);
 				result = `Promise -> ${Utils.visualize(await result)}`;
```
Zarel, line 927: This would break for `!eval` and `>>`.

```suggestion
				this.sendReply(`|uhtml|${uhtmlId}|${generateHTML('<', 'Promise pending')}`);
				this.update();
				result = `Promise -> ${Utils.visualize(await result)}`;
```

Zarel:
Thanks!

PR #7594, opened 2020-10-28
/whois: Only show the IP of the current connection

AnnikaCodes:
This protects users who get their unregistered accounts logged into by malicious users.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -170,8 +170,11 @@ export const commands: ChatCommands = {
 				buf += `<br />Semilocked: ${targetUser.semilocked}`;
 			}
 		}
-		if (user === targetUser ? user.can('ipself') : user.can('ip', targetUser)) {
-			const ips = targetUser.ips.map(ip => {
+		const canIPCheck = user.can('ip', targetUser);
+		if (canIPCheck || (user === targetUser && user.can('ipself'))) {
+			// The 'ipself' permission only allows users to see the IP address of their current connection.
+			// This maintains anonymity in case of account sharing.
+			const ips = (canIPCheck ? targetUser.ips : [connection.ip]).map(ip => {
 				const status = [];
 				const punishment = Punishments.ips.get(ip);
 				if (user.can('alts') && punishment) {


Zarel:
Unregistered accounts should never get merged, right? Is there a bug here?

AnnikaCodes:
I don't believe that's currently the case. Am I reading `users.ts` wrong?

Zarel:
Hm, this is probably bugged:

https://github.com/smogon/pokemon-showdown/blob/8d3e06854fe00b5d5d55c585bc109e88d9a4bb42/server/users.ts#L728-L737

Line 732 should probably always apply.

Zarel:
I think the only fix needed is to remove `&& conflictUser.connected`?

```ts
if (conflictUser && !conflictUser.registered) {
	if (newlyRegistered && userType !== '1') {
		if (conflictUser !== this) conflictUser.resetName();
	} else {
		this.send(`|nametaken|${name}|Someone is already using the name "${conflictUser.name}".`);
		return false;
	}
}
```

Zarel:
Or maybe we could loosen it to `&& conflictUser.latestIp !== this.latestIp`

AnnikaCodes:
yes, I think the latter is correct, since people may lurk as a Guest before picking a name

AnnikaCodes:
Should I just close this pr and push that then?

Zarel:
My main worry is two different users accidentally using the same unregistered username on the same IP (e.g. school computer lab).

Ideally we use cookies for this.

AnnikaCodes:
Couldn't we just do ``&& (conflictUser.latestIp !== this.latestIp || conflictUser.connected)``? That way, two different people can't share an unregistered name, and malicious users on another IP can't steal others' IP info.

I suppose the advantage to my original `/whois` change is that it also protects people who intentionally share registered accounts, but may not trust the person they're sharing with to know their approximate location.

Zarel:
I think I'd prefer `&& (conflictUser.latestIp !== this.latestIp || conflictUser.connected)`.

I think anyone who intentionally shares a registered account is someone I'm not going to explicitly bother supporting (they can do it if they want, but I'm not going to protect them from each other).

AnnikaCodes:
That's fair.

PR #7596, opened 2020-10-28
Ticketbans: Use the user object

mia-pi-git:


server/chat-plugins/helptickets.ts
```diff
@@ -1289,7 +1291,7 @@ export const commands: ChatCommands = {
 				targetUser.popup(`|modal|${user.name} has banned you from creating help tickets.${(target ? `\n\nReason: ${target}` : ``)}\n\nYour ban will expire in a few days.`);
 			}
 
-			const affected = HelpTicket.ban(userid, target);
+			const affected = HelpTicket.ban(this.targetUser || userid, target);
```
Zarel, line 1294:
```suggestion
			const affected = HelpTicket.ban(targetUser || userid, target);
```

?

Zarel:
Looks good to me. Thanks!

PR #7598, opened 2020-10-28
Punishments: Make locks expire on their own

AnnikaCodes:
Credit for the idea goes to HoeenHero.


AnnikaCodes:
@Zarel bump

Zarel:
_Probably_ good? I didn't see anything on a cursory review but I want to give it a deeper review before next restart.

PR #7599, opened 2020-10-29
calculate rank on server

dvdreedy:


server/ladders-remote.ts
```diff
@@ -149,4 +150,34 @@ export class LadderStore {
 	static async visualizeAll(username: string) {
 		return [`<tr><td><strong>Please use the official client at play.pokemonshowdown.com</strong></td></tr>`];
 	}
+
+	public calculateElo (p1: User | null, p2: User | null, score: number, formatId: string ) {
```
mia-pi-git, line 154:
```suggestion
	calculateElo(p1: User | null, p2: User | null, score: number, formatId: string) {
```
PS doesn't use `public`.

server/ladders-remote.ts
```diff
@@ -81,11 +81,12 @@ export class LadderStore {
...
-				room.add(`||Ladder updated, but score could not be retrieved.`);
-			} else {
-				room.add(`||Ladder (probably) updated, but score could not be retrieved (${error.message}).`);
-			}
+			const newRank = this.calculateElo(p1,p2,p1score,formatid);
+            if (error.message === 'stream interrupt') {
+                room.add(`||Ladder updated, your new rank is (${newRank}).`);
+            } else {
+                room.add(`||Ladder (probably) updated, your new rank is (${newRank}).`);
+            }
```
mia-pi-git, line 89:
```suggestion
			const newRank = this.calculateElo(p1,p2,p1score,formatid);
         if (error.message === 'stream interrupt') {
        		room.add(`||Ladder updated, your new rank is (${newRank}).`);
        	} else {
         	room.add(`||Ladder (probably) updated, your new rank is (${newRank}).`);
         }
```

server/ladders-remote.ts
```diff
@@ -149,4 +150,34 @@ export class LadderStore {
...
 		return [`<tr><td><strong>Please use the official client at play.pokemonshowdown.com</strong></td></tr>`];
 	}
+
+	public calculateElo (p1: User | null, p2: User | null, score: number, formatId: string ) {
+
+        let K = 50;
+        let elo = p1?.mmrCache[formatId];
+        const foeElo = p2?.mmrCache[formatId];
+        if(!elo || !foeElo)
+        return "unknown, please try again later";
```
AnnikaCodes, line 160:
```suggestion
        if (!elo || !foeElo) return "unknown, please try again later";
```

AnnikaCodes: Also this should probably `throw`.

dvdreedy:
need to actually ttest this

server/ladders-remote.ts
```diff
@@ -149,4 +150,31 @@ export class LadderStore {
...
 	}
+
+	calculateElo(p1: User | null, p2: User | null, score: number, formatId: string) {
+		let K = 50;
+		const elo = p1?.mmrCache[formatId];
+		const foeElo = p2?.mmrCache[formatId];
+		if (!elo || !foeElo) return "unknown, please try again later";
+
+		// dynamic K-scaling (optional)
+		if (elo < 1200) {
```
AnnikaCodes, line 161: This formula doesn't seem accurate to the [server-side one](https://github.com/smogon/pokemon-showdown-client/blob/master/lib/ntbb-ladder.lib.php#L395).

AnnikaCodes (changes requested):

server/ladders-remote.ts
```diff
@@ -81,10 +81,11 @@ export class LadderStore {
 		let problem = false;
 
 		if (error) {
+			const newRank = this.calculateElo(p1, p2, p1score, formatid);
 			if (error.message === 'stream interrupt') {
-				room.add(`||Ladder updated, but score could not be retrieved.`);
+				room.add(`||Ladder updated, your new rank is (${newRank}).`);
```
mia-pi-git, line 86: Bit of a necropost but it feels like, since this does the same thing both times, it could be just one `room.add` call.
(use the first one, the `(${newRank})` makes the second one awkward.)

mia-pi-git:
Closing this since we did it in another commit and this just got left open.

PR #7601, opened 2020-10-29
Move natures into their own file

KrisXV:


sim/dex.ts
```diff
@@ -813,19 +781,29 @@ export class ModdedDex {
...
+			nature = new Data.Nature(natureData);
+			if (nature.gen > this.gen) nature.isNonstandard = 'Future';
+		} else {
+			nature = new Data.Nature({id, name, exists: false});
+		}
 
+		if (nature.exists) {
+			if (!!nature.plus !== !!nature.minus) {
+				throw new Error(`Nature with a '${nature.plus ? 'plus' : 'minus'}' property but no '${nature.plus ? 'minus' : 'plus'}' property: ${nature.name}`);
+			}
```
Zarel, line 804: This belongs in `test/data` I think. The engine itself doesn't require both, does it?

Zarel: Either way, I'd rather test this once for free than every time someone uses `Dex.getNature`.

Zarel:
I believe everything here is good. I'll handle the tests in a follow-up commit.

sim/dex.ts
KrisXV: oh mb

PR #7604, opened 2020-10-29
Fix Overflow Stat Clause and turn it into a mod

KrisXV:


DaWoblefet:
This occurs in at least Gen 7+8, probably farther back but we can't check without ROM hacking.

KrisXV:
O

PR #7609, opened 2020-10-30
Don't allow forcerenamed usernames to be reused

AnnikaCodes:
https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/page-17#post-8642512


Zarel:
Thanks!

PR #7611, opened 2020-10-31
Allow bots to include "click here" in HTML messages

xfix:
This allows GitHub bot to report commits which happen to include this phrase such as this one.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1208,7 +1208,8 @@ export class CommandContext extends MessageContext {
 	checkHTML(htmlContent: string | null) {
 		htmlContent = ('' + (htmlContent || '')).trim();
 		if (!htmlContent) return '';
-		if (/>here.?</i.test(htmlContent) || /click here/i.test(htmlContent)) {
+		const auth = this.room ? this.room.auth : Users.globalAuth;
+		if (auth.get(this.user.id) !== '*' && (/>here.?</i.test(htmlContent) || /click here/i.test(htmlContent))) {
 			throw new Chat.ErrorMessage('Do not use "click here"');
 		}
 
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -1208,7 +1208,8 @@ export class CommandContext extends MessageContext {
 	checkHTML(htmlContent: string | null) {
 		htmlContent = ('' + (htmlContent || '')).trim();
 		if (!htmlContent) return '';
-		if (/>here.?</i.test(htmlContent) || /click here/i.test(htmlContent)) {
+		const auth = this.room ? this.room.auth : Users.globalAuth;
+		if (auth.get(this.user.id) !== '*' && (/>here.?</i.test(htmlContent) || /click here/i.test(htmlContent))) {
```
mia-pi-git, line 1212: This would allow whitelisted users (which is just voice without the symbol) to use click here. Maybe just auth.get(...) === '*'?

Zarel: Are you misreading this?

Zarel:
I would rather discourage this anti-pattern in bots still. I realize this causes problems in user-generated content but I guess bot authors could find/replace `here` with `her&#101;`.

The question is, is this too dumb of a solution? How do I balance that with knowing bot authors will start using "click here" without guide rails?

server/chat.ts
mia-pi-git: Yes, yes I am. I could have sworn it was a + earlier....

xfix:
Hm, yeah, fair point, without a hint to not use "click here"  people will use "click here". In fact there are examples of this on Showdown, for example, here is Portuguese room with unrecognized "click here" as it's written in another language.

image

Or from Eventos, a subroom of Spanish room.

image

For bots, replacing `here` with `her&#101;` seems to be fine for user provided content, I don't think there is a HTML tag or attribute including `here` in its name so doing this as a final transformation after generating HTML output should work fine. HTML does have an entity called `&therefore;`, but the issue with that one could be easily avoided by using `&#8756;` or `∴` instead.

xfix:
I think I will be closing it and applying changes in my own bot instead.

Zarel:
On the other hand, this sounds like a good reason to add `aquí` to our banlist.

PR #7613, opened 2020-10-31
Modlog: Avoid optional properties in ModlogEntry 

AnnikaCodes:


server/modlog.ts
```diff
@@ -74,19 +74,21 @@ export interface ModlogSearch {
...
-	ip?: string;
-	isGlobal?: boolean;
-	loggedBy?: ID;
-	note?: string;
+	roomID: string;
+	visualRoomID: string;
+	userid: ID | null;
+	autoconfirmedID: ID | null;
+	alts: ID[];
+	ip: string;
```
monsanto, line 82: Is `ip` always present? I see you `|| ''` elsewhere. If not, I would choose a strategy consistent with `userid`--empty string isn't a valid IP address.

monsanto (changes requested):
Need corresponding `NOT NULL` entries in schema, otherwise looks good

server/modlog.ts
AnnikaCodes: no, it's not, i'll make it null

monsanto (approved):

monsanto:
Until we have some migration support the procedure to hotpatch things with schema changes is set `Config.usesqlitemodlog = false`, delete `databases/modlog.db*`, hotpatch, `Config.usesqlitemodlog = true`

PR #7614, opened 2020-11-01
Abilities: add isPermanent flag

DaWoblefet:
Based on [new USUM research](https://pastebin.com/9fvC7PzX), when combined with research on Gen 8 abilities, this adds a neverChange Ability flag to the following:
As One (both variants), Battle Bond, Comatose, Disguise, Gulp Missile, Ice Face, Multitype, Power Construct, RKS System, Schooling, Shields Down, Stance Change, Zen Mode

Here's cart footage demonstrating this research in SwSh: [[1](https://www.youtube.com/watch?v=b0gdBvKxvEI)] [[2](https://www.youtube.com/watch?v=UlNyofLOd1s)] and here's some more [research documentation](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8430587)

Generally speaking, this indicates they can't be replaced, copied, or negated in any way. I also updated the lists for every ability effect to get things in sync. There were quite a number of desyncs due to time, such as:

- Gulp Missile and Ice Face weren't on most Ability blocking lists
- Zen Mode wasn't on some ability blocking lists
- Wandering Spirit and Mummy didn't have proper lists at all
- Various other minor one-offs

I still have quite a bit to do, including:

- Oldgens (e.g. Zen Mode)
- Making sure descriptions are right
- Figuring out how to have Entrainment / Simple Beam / Worry Seed into Truant and Worry Seed into Insomnia return something that indicates to use "It doesn't affect [Pokemon]..." rather than "But it failed!" without causing the move to always use that failure message
- A couple more mechanics tests on cart (Receiver vs what we call 'noability', Ability index 0, and a couple other confirmations)


data/abilities.ts
```diff
@@ -2151,14 +2158,18 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
+		neverChanges: true,
 		name: "Multitype",
 		rating: 4,
 		num: 121,
 	},
 	mummy: {
 		name: "Mummy",
 		onDamagingHit(damage, target, source, move) {
-			if (move.flags['contact'] && source.ability !== 'mummy') {
+			if (source.getAbility().neverChanges && source.ability !== 'mummy') {
```
MacChaeger, line 2169:
```suggestion
			if (source.getAbility().neverChanges || source.ability === 'mummy') {
```
The old code filtered out Pokémon that had the ability Mummy. What you have now would filter out every Pokémon that _doesn’t_ have Mummy, I think.

urkerab:
I had a go at this as PR #6263; I don't know whether you can mark this PR as superseding that one, or at least steal some code from it.

data/abilities.ts
urkerab: He's doing an early return for when the attacker's Ability can't be changed into Mummy.

data/moves.ts
```diff
@@ -15264,13 +15256,13 @@ export const Moves: {[moveid: string]: MoveData} = {
...
-			const bannedAbilities = [
-				'asoneglastrier', 'asonespectrier', 'battlebond', 'comatose', 'disguise', 'multitype', 'powerconstruct', 'rkssystem', 'schooling', 'shieldsdown', 'simple', 'stancechange', 'truant', 'zenmode',
-			];
-			if (bannedAbilities.includes(pokemon.ability)) {
+		onTryHit(target) {
+			if (target.getAbility().neverChanges || target.ability === 'simple') {
 				return false;
 			}
+			if (target.ability === 'truant') {
+				return false; // TODO how to return differently?
```
urkerab, line 15264: - If the immunity should be checked before accuracy (if relevant), use `onTryImmunity` instead of `onTryHit`.
- If the immunity should be checked here, then return `null`, but also add code to show the target was immune.

data/moves.ts
```diff
@@ -15383,11 +15375,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {protect: 1, mirror: 1, authentic: 1, mystery: 1},
 		onTryHit(target, source) {
-			const bannedAbilities = [
-				'asoneglastrier', 'asonespectrier', 'battlebond', 'comatose', 'disguise', 'gulpmissile', 'hungerswitch', 'iceface', 'illusion', 'multitype', 'neutralizinggas', 'powerconstruct', 'rkssystem', 'schooling', 'shieldsdown', 'stancechange', 'wonderguard', 'zenmode',
-			];
+			const additionalBannedAbilities = ['hungerswitch', 'illusion', 'neutralizinggas', 'wonderguard'];
```
urkerab, line 15378: Huh, so Flower Gift, Forecast, Imposter, Power of Alchemy, Receiver and Trace can be Skill Swapped?

urkerab (commented):
It seems to be that there are two special categories of Abilities:
- Those that can't be suppressed e.g. by Gastro Acid or Neutralizing Gas, or replaced by Entrainment etc.
- Those that can be suppressed and replaced but can't be copied (although Wonder Guard is a bit of an exception there, as I believe it can be Traced)

I don't know whether it's worth adding a flag for the latter.

data/abilities.ts
Zarel: No, I think MacChaeger is right.

I could cache the `getAbility` though.

```suggestion
			const sourceAbility = source.getAbility();
			if (sourceAbility.neverChanges || sourceAbility.id === 'mummy') {
```

Zarel: Actually this would depend on: "If the attacker's Ability is Gastro Acid Mummy, will the defender's Mummy activate?" If yes, my code is correct.

Zarel:
Thanks! When we designed the systems, we noticed that each move/ability/item effect had different lists of compatible abilities, so while we always wanted to share lists, we never knew how. Now that we know the games just use multiple lists, we can use the same shared list they do.

data/abilities.ts
urkerab: @MacChaeger's code is correct, but his comment was confusing.

data/moves.ts
DaWoblefet: Yep, that's right. I thought Imposter in particular was weird that it could be Skill Swapped, but it makes sense (say you send out Imposter into a Substitute or something). Skill Swap uses the same list for both the source and the target, which like I have here is just the general bans and Illusion + Wonder Guard (though behavior in Gen 8 indicates that NGas and Hunger Switch also fall in this group). Ain't it bizarre? Skill Swap is kind of out there on its own.

DaWoblefet:
> I had a go at this as PR #6263; I don't know whether you can mark this PR as superseding that one, or at least steal some code from it.

I did this because of our new knowledge of how USUM works, which I linked at the start of this PR (here it is again: https://pastebin.com/9fvC7PzX). I'm not a huge fan of 2 completely separate lists because of how Game Freak internally does its checks. It does a lot of inheritance with exceptions. For example, Entrainment inherits from Trace, and Trace inherits from Role Play except Wonder Guard is allowed and ability index 0 is banned (but Entrainment doesn't ban ability index 0), then Role Play inherits from the general bans and has its own list, some of which overlaps with the general bans and some doesn't.

It wouldn't be unreasonable to have a second flag;  the neverChanges one is much more straightforward.

Zarel:
If you have time, it looks like we settled on `isPermanent` for the flag name.

data/abilities.ts
DaWoblefet: Zarel's correct, so I've gone and done it his way.

data/moves.ts
DaWoblefet: Yep, onTryImmunity works perfectly for what I want, thanks!

DaWoblefet:
> I still have quite a bit to do, including:
> 
> * Oldgens (e.g. Zen Mode)
> * Making sure descriptions are right
> * Figuring out how to have Entrainment / Simple Beam / Worry Seed into Truant and Worry Seed into Insomnia return something that indicates to use "It doesn't affect [Pokemon]..." rather than "But it failed!" without causing the move to always use that failure message
> * A couple more mechanics tests on cart (Receiver vs what we call 'noability', Ability index 0, and a couple other confirmations)

- I'm not actually 100% certain on the Zen Mode mechanics from 5-6. If descriptions are right, then it looks like Zen Mode was only in the "Role Play table" in those gens; that is, Role Play (target), Trace, and Entrainment (user) will fail, but Skill Swap, Role Play (user), Entrainment (target), Gastro Acid, Simple Beam, Worry Seed, and Mummy will succeed in messing with it (@Marty-D ?)
- Descriptions should be correct; there's new stuff for Gen 8 and I added Gen 7 descriptions where appropriate
- urkerab helped there, onTryImmunity was what I needed
- Receiver / PoA just can't take "No Ability", but if another Pokemon faints afterwards with an ability it can take, it'll do so

data/text/abilities.ts
```diff
@@ -1376,8 +1378,9 @@ export const AbilitiesText = {
...
+		desc: "On switch-in, or when this Pokemon acquires this ability, this Pokemon copies a random adjacent opposing Pokemon's Ability. However, if one or more adjacent Pokemon has the Ability \"No Ability\", Trace won't copy anything even if there is another valid Ability it could normally copy. Otherwise, if there is no Ability that can be copied at that time, this Ability will activate as soon as an Ability can be copied. Abilities that cannot be copied are the previously mentioned \"No Ability\", as well as As One, Battle Bond, Comatose, Disguise, Flower Gift, Forecast, Gulp Missile, Hunger Switch, Ice Face, Illusion, Imposter, Multitype, Neutralizing Gas, Power Construct, Power of Alchemy, Receiver, RKS System, Schooling, Shields Down, Stance Change, Trace, Wonder Guard, and Zen Mode.",
```
Marty-D, line 1381: Wonder Guard shouldn't be mentioned here, ditto for the line below.

Marty-D:
Descriptions for these effects were accurate for Gen 5 and 6 at minimum, yeah.

Zarel (commented):
Looks good to me

Zarel:
(Someone tell me when it's safe to merge, or either of you could merge it yourself tbh.)

DaWoblefet:
I need to still wrap up oldgen Zen Mode and then I'm done with everything I can think of. @urkerab was there anything in your original PR that I've missed?

urkerab:
Have the Gen 6 overrides for Simple Beam, Skill Swap and Worry Seed become redundant?

> was there anything in your original PR that I've missed?

In case you're interested, I had added code to make Gastro Acid wear off when Baton Passed via something with an ability that wouldn't be affected by it, and also I had added code to make `/dt` show whether an ability was affected by Gastro Acid and/or Mold Breaker.

data/text/abilities.ts
DaWoblefet: Fixed

DaWoblefet:
> Have the Gen 6 overrides for Simple Beam, Skill Swap and Worry Seed become redundant?
> 
> > was there anything in your original PR that I've missed?
> 
> In case you're interested, I had added code to make Gastro Acid wear off when Baton Passed via something with an ability that wouldn't be affected by it, and also I had added code to make `/dt` show whether an ability was affected by Gastro Acid and/or Mold Breaker.

Yep, same with Gastro Acid, so I removed them. I also added both of the changes you did (shamelessly copy-pasted). Thanks!

This should be ready to merge, I'll let Marty press the button just to make sure he's seen everything he wants.

data/moves.ts
```diff
@@ -5901,6 +5901,9 @@ export const Moves: {[moveid: string]: MoveData} = {
 				this.add('-endability', pokemon);
 				this.singleEvent('End', pokemon.getAbility(), pokemon.abilityData, pokemon, pokemon, 'gastroacid');
 			},
+			onCopy(pokemon) {
+				if (this.dex.getAbility(pokemon.ability).isPermanent) pokemon.removeVolatile('gastroacid');
```
urkerab, line 5905: Actually that code is out of date and should probably say
```suggestion
				if (pokemon.getAbility().isPermanent) pokemon.removeVolatile('gastroacid');
```

data/mods/gen6/moves.ts
```diff
@@ -71,6 +71,20 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 		},
 	},
+	entrainment: {
+		inherit: true,
+		onTryHit(target, source) {
+			const bannedSourceAbilities = [
+				'flowergift', 'forecast', 'illusion', 'imposter', 'multitype', 'stancechange', 'trace', 'zenmode',
```
urkerab, line 78: So the difference here is that Zen Mode still can't be copied, but because it's not permanent, the Gen 8 code doesn't apply? I guess it would be a bit of a hack to add Zen Mode to the Gen 8 exception lists for Entrainment and Role Play so that it works automagically in Gen 6...

DaWoblefet: hmm, I think that makes a good deal more sense to me. In exchange for minor duplication in modern gens, you don't have to inherit at all for Trace / Entrainment / Role Play, which I think is a net gain trade-off; I imagine if some new functionality were discovered with Entrainment, someone would be more likely to forget to update the mod than we lose by adding one more item in a list that could potentially be duplicated

Marty-D:
Nicely done, thanks! 👍

PR #7617, opened 2020-11-01
Config: Move Bot rank below Driver 

AnnikaCodes:
This allows users to be banned from modjoined autojoined rooms (which is useful for keeping global bots out of Staff room when some of their owners aren't global staff).

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -628,9 +628,11 @@ exports.grouplist = [
 		modchat: true,
 		hiderank: true,
 	},
-	// Bots are ranked below Driver so that they can be kept out of modjoin % rooms.
-	// This was motivated by rhe desire to promote Global Bots whose owners do not hav Staff room access.
 	{
+		// Bots are ranked below Driver/Mod so that Global Bots can be kept out
+		// of modjoin % rooms (namely, Staff).
+		// (They were previously above Driver/Mod so they can have game management
+		// permissions drivers don't, but these permissions can be manually given.)
 		symbol: '*',
 		id: "bot",
 		name: "Bot",
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -586,23 +586,6 @@ exports.grouplist = [
 		joinbattle: true,
 	},
 	{
-		symbol: '*',
-		id: "bot",
-		name: "Bot",
-		inherit: '%',
-		jurisdiction: 'u',
-
-		addhtml: true,
-		tournaments: true,
-		declare: true,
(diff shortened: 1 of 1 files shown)


Zarel:
I would rather do this the other way around: Global bots aren't allowed in Staff by default, and then roombotting them allows them in Staff.

PS simply isn't designed for long-term bans (blacklists are limited to one year) and bans really should not be used as an access control system.

Zarel:
Investigation has turned up that we've never had bots below driver/mod.

After a brief Staff discussion, we seem to be reasonably unanimous: we should move bots below driver/mod (and document why they're there in a comment).

AnnikaCodes:
Alright, I'll update the PR to make that change in config-example

Zarel:
Further discussion in Staff says that game management should be given to bots by default, since that was apparently the reason it was elevated above driver/mod in the first placxe.

Zarel:
I think (not sure) that tours management is a separate permission. Can you investigate that?

config/config-example.js
```diff
@@ -645,6 +628,27 @@ exports.grouplist = [
...
+	// This was motivated by rhe desire to promote Global Bots whose owners do not hav Staff room access.
+	{
+		symbol: '*',
+		id: "bot",
+		name: "Bot",
+		inherit: '%',
+		jurisdiction: 'u',
+
+		addhtml: true,
+		tournaments: true,
```
AnnikaCodes, line 641: Yeah, pretty sure it's the `tournaments` permission

config/config-example.js
```diff
@@ -645,6 +628,27 @@ exports.grouplist = [
 		modchat: true,
 		hiderank: true,
 	},
+	// Bots are ranked below Driver so that they can be kept out of modjoin % rooms.
+	// This was motivated by rhe desire to promote Global Bots whose owners do not hav Staff room access.
+	{
```
Zarel, line 633: _Multiple_ spelling errors here. :p

```suggestion
	{
		// Bots are ranked below Driver/Mod so that Global Bots can be kept out
		// of modjoin % rooms (namely, Staff).
		// (They were previously above Driver/Mod so they can have game management
		// permissions drivers don't, but these permissions can be manually given.)
```

config/config-example.js
Zarel: Thanks!

Zarel:
Looks good! Can I get you to apply this to Main?

AnnikaCodes:
did

PR #7619, opened 2020-11-02
Support disabling modlogs

AnnikaCodes:
This allows PS to run without depending on SQLite, which I understood to be something Zarel wanted. I'm open to suggestions for a better variable name than `Config.usesqlitemodlog`.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -417,8 +417,8 @@ export const commands: ChatCommands = {
 				void IPTools.loadHostsAndRanges();
 				this.sendReply("IPTools has been hot-patched.");
 			} else if (target === 'modlog') {
-				if (!Config.usesqlitemodlog) {
-					return this.errorReply("Modlogs are not enabled on this server; please set up `Config.usesqlitemodlog`.");
+				if (!Config.usesqlite) {
+					return this.errorReply("Modlogs are not enabled on this server; please set up `Config.usesqlite`.");
 				}
 				patch = 'modlog';
 				if (lock['modlog']) {
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -429,8 +429,8 @@ export const commands: ChatCommands = {
 				}
 				if (requiresForce(patch)) return this.errorReply(requiresForceMessage);
 
-				const streams = Rooms.Modlog?.streams;
-				const sharedStreams = Rooms.Modlog?.sharedStreams;
+				const streams = Rooms.Modlog.streams;
+				const sharedStreams = Rooms.Modlog.sharedStreams;
 
 				const processManagers = ProcessManager.processManagers;
 				for (const manager of processManagers.slice()) {
@@ -444,8 +444,8 @@ export const commands: ChatCommands = {
 				);
 				this.sendReply("Modlog has been hot-patched.");
 
-				if (streams) Rooms.Modlog.streams = streams;
-				if (sharedStreams) Rooms.Modlog.sharedStreams = sharedStreams;
+				Rooms.Modlog.streams = streams;
+				Rooms.Modlog.sharedStreams = sharedStreams;
 				this.sendReply("Modlog streams have been re-initialized.");
 			} else if (target.startsWith('disable')) {
 				this.sendReply("Disabling hot-patch has been moved to its own command:");
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -441,8 +441,8 @@ export const commands: ChatCommands = {
 				);
 				this.sendReply("Modlog has been hot-patched.");
 
-				if (streams) Rooms.Modlog!.streams = streams;
-				if (sharedStreams) Rooms.Modlog!.sharedStreams = sharedStreams;
+				if (streams) Rooms.Modlog.streams = streams;
+				if (sharedStreams) Rooms.Modlog.sharedStreams = sharedStreams;
 				this.sendReply("Modlog streams have been re-initialized.");
 			} else if (target.startsWith('disable')) {
 				this.sendReply("Disabling hot-patch has been moved to its own command:");
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -440,7 +440,6 @@ export const commands: ChatCommands = {
 					Rooms.MODLOG_DB_PATH || `${__dirname}/../../databases/modlog.db`
 				);
 				this.sendReply("Modlog has been hot-patched.");
-
 				Rooms.Modlog.streams = streams;
 				Rooms.Modlog.sharedStreams = sharedStreams;
 				this.sendReply("Modlog streams have been re-initialized.");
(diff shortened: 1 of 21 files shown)


server/chat-commands/admin.ts
```diff
@@ -417,14 +417,17 @@ export const commands: ChatCommands = {
 				void IPTools.loadHostsAndRanges();
 				this.sendReply("IPTools has been hot-patched.");
 			} else if (target === 'modlog') {
+				if (!Config.usesqlitemodlog) {
```
monsanto, line 420: I would call this something like `usesqlite` because 1) other things will be using SQLite and 2) using the same variable as our temporary one breaks the temporarily useful ability to turn off SQLite logging before doing a migration

server/config-loader.ts
```diff
@@ -24,6 +24,15 @@ export function load(invalidate = false) {
 	const config = ({...defaults, ...require('../config/config')}) as ConfigType;
 	// config.routes is nested - we need to ensure values are set for its keys as well.
 	config.routes = {...defaults.routes, ...config.routes};
+
+	// Automatically disable SQLite modlogs if better-sqlite3 isn't installed
+	try {
+		require('better-sqlite3');
+	} catch (e) {
+		reportError(`Modlogs have been disabled because better-sqlite3 is not installed or could not be loaded.`);
+		config.usesqlitemodlog = false;
```
monsanto, line 33: If `usesqlite` is set to `true` in config this should abort startup, only if unset continue operating in degraded mode

server/modlog.ts
```diff
@@ -476,8 +476,8 @@ type ModlogResult = ModlogEntry | undefined;
 export const PM = new QueryProcessManager<ModlogTextQuery, ModlogResult[]>(module, async data => {
 	const {rooms, regexString, maxLines, onlyPunishments} = data;
 	try {
-		const results = await Rooms.Modlog.runTextSearch(rooms, regexString, maxLines, onlyPunishments);
-		return results.map((line: string, index: number) => parseModlog(line, results[index + 1]));
+		const results = await Rooms.Modlog?.runTextSearch(rooms, regexString, maxLines, onlyPunishments);
```
monsanto, line 479: I don't like that this change infects all callers, it isn't a nice interface to use. I think `Rooms.Modlog` should always be set and these calls should just be no-ops if you don't have SQLite enabled

monsanto (changes requested):

AnnikaCodes:
I'm pretty sure I implemented those changes; tests are failing because of the team validator (???)

Zarel:
I assume that's because of d83118162b0d3e3f0. Rebasing/merging master should fix the test.

server/chat-commands/admin.ts
```diff
@@ -420,14 +420,17 @@ export const commands: ChatCommands = {
 				void IPTools.loadHostsAndRanges();
 				this.sendReply("IPTools has been hot-patched.");
 			} else if (target === 'modlog') {
+				if (!Config.usesqlite) {
```
monsanto, line 423: I'd just kill this block of code for now. Until we actually move everything over to SQLite the existing text modlogs should continue to work without interruption.

server/chat-plugins/modlog-viewer.ts
```diff
@@ -205,6 +209,10 @@ export const commands: ChatCommands = {
 	pl: 'modlog',
 	timedmodlog: 'modlog',
 	modlog(target, room, user, connection, cmd) {
+		if (!Config.usesqlite) {
```
monsanto, line 212: Ditto

server/chat-commands/admin.ts
```diff
@@ -420,14 +420,17 @@ export const commands: ChatCommands = {
...
+				}
 				patch = 'modlog';
 				if (lock['modlog']) {
 					return this.errorReply(`Hot-patching modlogs has been disabled by ${lock['modlog'].by} (${lock['modlog'].reason})`);
 				}
 				if (requiresForce(patch)) return this.errorReply(requiresForceMessage);
 
-				const streams = Rooms.Modlog.streams;
-				const sharedStreams = Rooms.Modlog.sharedStreams;
+				const streams = Rooms.Modlog?.streams;
```
monsanto, line 432: `Rooms.Modlog` should not be undefined, please see all other uses too

monsanto (changes requested):

AnnikaCodes:
Alright

server/modlog.ts
```diff
@@ -477,7 +479,7 @@ export const PM = new QueryProcessManager<ModlogTextQuery, ModlogResult[]>(modul
 	const {rooms, regexString, maxLines, onlyPunishments} = data;
 	try {
 		const results = await Rooms.Modlog.runTextSearch(rooms, regexString, maxLines, onlyPunishments);
-		return results.map((line: string, index: number) => parseModlog(line, results[index + 1]));
+		return results?.map((line: string, index: number) => parseModlog(line, results[index + 1])) || [];
```
monsanto, line 482: Is this still necessary?

server/chat-commands/admin.ts
```diff
@@ -440,6 +440,7 @@ export const commands: ChatCommands = {
 					Rooms.MODLOG_DB_PATH || `${__dirname}/../../databases/modlog.db`
 				);
 				this.sendReply("Modlog has been hot-patched.");
+
```
monsanto, line 443:
```suggestion
```

test/server/modlog.js
```diff
@@ -101,24 +101,26 @@ describe('Modlog', () => {
 		});
 	});
 
-	describe("Modlog#rename", async () => {
-		const entry = {note: 'This is in a modlog that will be renamed!', action: 'UNITTEST'};
+	(Config.usesqlite ? describe : describe.skip)("Modlog#rename", () => {
+		it('should rename modlogs', async () => {
```
monsanto, line 105: As far as I can tell this (the `it` change) has equivalent semantics, so if you prefer this way then: sure. This is the kind of change you should just make without requesting review from me, though

test/tools/modlog/converter.js
```diff
@@ -32,19 +30,24 @@ const garfieldCopypasta = [
 	` In some sense, even his cat Garfield is an object to him, a thing. The first ideology that comes to mind that comes to mind when thinking of objects in the tangible world... is pragmatism. slur => slur)`,
 ].join('');
 
-describe('Modlog conversion script', () => {
+(Config.usesqlite ? describe : describe.skip)('Modlog conversion script', () => {
+	before(() => {
+		this.converter = require('../../../tools/modlog/converter');
+		this.ml = require('../../../.server-dist/modlog');
+	});
```
monsanto, line 37: Don't see the point in this change, should be safe to import those files even if SQLite unavailable

server/modlog.ts
```diff
@@ -477,7 +479,7 @@ export const PM = new QueryProcessManager<ModlogTextQuery, ModlogResult[]>(modul
 	const {rooms, regexString, maxLines, onlyPunishments} = data;
 	try {
 		const results = await Rooms.Modlog.runTextSearch(rooms, regexString, maxLines, onlyPunishments);
-		return results.map((line: string, index: number) => parseModlog(line, results[index + 1]));
+		return results.map((line: string, index: number) => parseModlog(line, results[index + 1])) || [];
```
monsanto, line 482:
```suggestion
		return results.map((line: string, index: number) => parseModlog(line, results[index + 1]));
```

PR #7620, opened 2020-11-02
Modlog: Encapsulate hotpatching logic

AnnikaCodes:


AnnikaCodes:
@monsanto bump

Zarel:
I don't know if I like this. The chat command is nice because it has access to both the new and the old stuff.

server/modlog.ts
```diff
@@ -450,6 +450,26 @@ export class Modlog {
...
+			throw new Error(`Attempt to call Modlog#hotpatch on a modlog instance other than Rooms.Modlog`);
+		}
+
+		const streams = Rooms.Modlog.streams;
+		const sharedStreams = Rooms.Modlog.sharedStreams;
+
+		const processManagers = ProcessManager.processManagers;
+		for (const manager of processManagers.slice()) {
+			if (manager.filename.startsWith(__filename)) void manager.destroy();
+		}
```
Zarel, line 464: Why the manual search for PMs, anyway? I believe our usual solution to this problem is to export `Rooms.Modlog.PM`.

AnnikaCodes: Chat plugins search manually; I can certainly just put the PM in `Rooms` if thats better.

monsanto:
This isn't totally what I had in mind--Zarel is correct that access to both the old and new is necessary. 

@Zarel the issue is that internal details of hotpatchable modules leak into `/hotpatch`. This means 1) we have to export more than necessary & 2) when the details change, we have to `/hotpatch` both chat & the desired module. This was a problem for us when deploying the SQLite modlog.

I think a better solution is for hotpatchable modules to export `setup` and `teardown` methods. On startup, the `setup` method is called on each module. On `/hotpatch`, `teardown` is called on the old module, the require cache is cleared, and `setup` is called on the new module. This should be general enough that there shouldn't be a need to make module-specific edits to `/hotpatch` again.

Thoughts?

server/modlog.ts
Zarel: I might not be understanding correctly, so correct me if I'm wrong: You're restarting chat-plugin processes but only for chat plugins that use the Modlog library?

That seems really fragile/buggy. I would recommend just always hotpatching chat when hotpatching modlog (like how hotpatching formats also hotpatches battles).

AnnikaCodes: No, `/hotpatch chat` manually searches.

Zarel: Oh, it turns out I just merged hotpatching formats/battles a while ago. But the point is, I think you should just hotpatch chat; partially hotpatching chat sounds unnecessarily dangerous.

AnnikaCodes: If you want to hotpatch chat every time why not just make `Modlog` part of `Chat`

Zarel:
> @Zarel the issue is that internal details of hotpatchable modules leak into `/hotpatch`. This means 1) we have to export more than necessary & 2) when the details change, we have to `/hotpatch` both chat & the desired module. This was a problem for us when deploying the SQLite modlog.

I don't think hotpatching chat as well as the desired module is a big deal, but I do agree that better encapsulation could make the code more readable/maintainable.

> I think a better solution is for hotpatchable modules to export `setup` and `teardown` methods. On startup, the `setup` method is called on each module. On `/hotpatch`, `teardown` is called on the old module, the require cache is cleared, and `setup` is called on the new module. This should be general enough that there shouldn't be a need to make module-specific edits to `/hotpatch` again.

We already support `teardown` in chat plugins, although we call it `destroy`. We relatively consistently use `destroy` as a name for teardown/deallocation stuff and I'd prefer to keep using it here. Setup code can just be put in the top level; there's not too much need for an explicit method for it. Chat plugins support `if (Chat.plugins.pluginname)` if you want to detect that you're in a hotpatch and not a complete server restart.

I think I agree that Modlog (and any other module with complex hotpatch needs) could benefit from having some of its teardown moved to a `destroy` method. I'm not sure I see the need for anything else.

Zarel:
To be clear, when I say "I'm not sure I see the need for anything else", that doesn't mean "no". Annika, as long as you're maintaining Modlog, you don't need my approval to structure it however is most convenient for you (as long as you follow our code style guidelines).

AnnikaCodes:
I think that Modlog should have a teardown method, for sure. Perhaps a compromise for the setup logic could be to move modlog into Chat (which would eliminate the need for a separate modlog patch; moderation commands are already in `Chat` and I think it's equally logical as `Rooms`), and add a cache for the streams similar to what we do with `Chat#oldPlugins`.

Thoughts?

Zarel:
I think that would be an improvement.

monsanto:
Sure.

The friends list has a similar structure to the modlog (server/friends.ts, and a chat-plugin) and would make sense to hotpatch in chat as well. I don't know precisely what you have in mind for re-organization but it would be nice if /hotpatch chat Just Works(tm) for that & similarly structured things too.

AnnikaCodes:
@Zarel bump

AnnikaCodes:
@Zarel or @monsanto bump

server/modlog.ts
```diff
@@ -450,6 +458,13 @@ export class Modlog {
 		};
 	}
 
+	async destroy() {
+		if (this === Chat.modlog) {
```
Zarel, line 462: What's this guarding against?

AnnikaCodes: A modlog instance that's not the main one for the server (for example, one created by the converter) shouldn't set `Chat.modlogStreamPersistence` or destroy the ProcessManager since those are specific to `Chat.modlog`.

Zarel: Makes sense.

Zarel: Actually, is `Chat.modlogStreamPersistence` needed at all? It seems like reading `Chat.modlog.streams` and `Chat.modlog.sharedStreams` is enough, right?

AnnikaCodes: That would work; `Chat.modlogStreamPersistence` is just another level of abstraction.

AnnikaCodes:
bump

AnnikaCodes:
@Zarel @monsanto bump

AnnikaCodes:
Uhhhh... there are multi battle changes in here for some reason? Anyway, do we still want this?

AnnikaCodes:
This is less important without the flat-file support, but still something I should probably fix up sooner or later.

PR #7621, opened 2020-11-02
Make /addline broadcastable and rename to /pline

mia-pi-git:


Zarel:
I thought we renamed `runBroadcast` to `broadcastable`?

Zarel:
btw can you remove all the other `rawpacket` usages from the repo in this PR?

server/chat-commands/admin.ts
```diff
@@ -131,10 +131,11 @@ export const commands: ChatCommands = {
 		`/changerankuhtml [rank], [name], [message] - Changes the message previously shown with /addrankuhtml [rank], [name]. Requires: * # &`,
 	],
 
-	addline(target, room, user) {
+	showraw(target, room, user) {
```
Zarel, line 134: Still haven't been able to think of a good name. I don't like `showraw` because `raw` currently means "HTML" in PS's codebase.

Zarel:
I haven't heard much objection to the name `pline` so I'm merging now. If you would like to object, the term isn't set in stone.

However, if no one objects, I'll probably update the documentation to use "pline" at some point. I think having our own made-up term will be much clearer than using words like "message" and "line" and "packet" that have many meanings.

PR #7622, opened 2020-11-02
Add /vsl alias to viewspotlights

mia-pi-git:
https://www.smogon.com/forums/threads/vsl-for-viewspotlights.3672561/


Zarel:
I didn't hear objections so I guess we're taking this. Thanks!

PR #7623, opened 2020-11-02
Support basic text formatting in polls/announcements

mia-pi-git:
https://www.smogon.com/forums/threads/allow-non-html-polls-to-have-basic-formatting.3653845/


Zarel:
Thanks!

PR #7624, opened 2020-11-02
Tournaments: Start timer in all battles when forcetimer is activated

mia-pi-git:
https://www.smogon.com/forums/threads/turn-on-timer-in-all-existing-tour-battles-if-forcetimer-is-turned-on-mid-tour.3655471/


Zarel:
Thanks!

PR #7625, opened 2020-11-02
Support Config.noguestsecurity

Zarel:
This Config option, intended purely for development test servers, allows users to log in without a signed token, by using `/trn [USERNAME]`. This allows logins without a login server.

The rest of the server has been changed to make sure unregistered users never have any powers they shouldn't have.

### DIFF
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -391,6 +391,16 @@ exports.inactiveuserthreshold = 1000 * 60 * 60;
 exports.autolockdown = true;
 
 /**
+ * noguestsecurity - purely for development servers: allows logging in without
+ * a signed token: simply send `/trn [USERNAME]`. This allows using PS without
+ * a login server.
+ *
+ * Logging in this way will make you considered an unregistered user and grant
+ * no authority.
+ */
+exports.noguestsecurity = false;
+
+/**
  * Custom avatars.
  * This allows you to specify custom avatar images for users on your server.
  * Place custom avatar files under the /config/avatars/ directory.
--- a/config/config-example.js
+++ b/config/config-example.js
@@ -396,7 +396,7 @@ exports.autolockdown = true;
  * a login server.
  *
  * Logging in this way will make you considered an unregistered user and grant
- * no authority.
+ * no authority. You cannot log into a trusted (g+/r%) user account this way.
  */
 exports.noguestsecurity = false;
 
--- a/server/user-groups.ts
+++ b/server/user-groups.ts
@@ -213,11 +213,14 @@ export class RoomAuth extends Auth {
 		super();
 		this.room = room;
 	}
-	get(user: ID | User): GroupSymbol {
+	get(userOrID: ID | User): GroupSymbol {
+		const user = typeof userOrID === 'string' ? null : userOrID as User;
+		const id = user ? user.id : userOrID as ID;
+		if (user && !user.registered) return Auth.defaultSymbol();
+
 		const parentAuth: Auth | null = this.room.parent ? this.room.parent.auth :
 			this.room.settings.isPrivate !== true ? Users.globalAuth : null;
-		const parentGroup = parentAuth ? parentAuth.get(user) : Auth.defaultSymbol();
-		const id = typeof user === 'string' ? user : (user as User).id;
+		const parentGroup = parentAuth ? parentAuth.get(userOrID) : Auth.defaultSymbol();
 
 		if (this.has(id)) {
 			// authority is whichever is higher between roomauth and global auth
--- a/server/user-groups.ts
+++ b/server/user-groups.ts
@@ -214,9 +214,7 @@ export class RoomAuth extends Auth {
 		this.room = room;
 	}
 	get(userOrID: ID | User): GroupSymbol {
-		const user = typeof userOrID === 'string' ? null : userOrID as User;
-		const id = user ? user.id : userOrID as ID;
-		if (user && !user.registered) return Auth.defaultSymbol();
+		const id = typeof userOrID === 'string' ? userOrID : (userOrID as User).id;
 
 		const parentAuth: Auth | null = this.room.parent ? this.room.parent.auth :
 			this.room.settings.isPrivate !== true ? Users.globalAuth : null;
--- a/server/users.ts
+++ b/server/users.ts
@@ -752,13 +752,14 @@ export class User extends Chat.MessageContext {
 			// unregistered users can only merge in limited situations
 			let canMerge = registered && conflictUser.registered;
 			if (
-				!registered && !conflictUser.registered && conflictUser.latestIp !== this.latestIp &&
+				!registered && !conflictUser.registered && conflictUser.latestIp === this.latestIp &&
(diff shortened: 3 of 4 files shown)


Zarel:
@monsanto @xfix I'd very much like some assurance that I'm not accidentally introducing some sort of vulnerability.

I want this to be safe to enable on development servers without actually letting malicious users take over your server (DoSing is fine).

server/user-groups.ts
```diff
@@ -213,11 +213,14 @@ export class RoomAuth extends Auth {
 		super();
 		this.room = room;
 	}
-	get(user: ID | User): GroupSymbol {
+	get(userOrID: ID | User): GroupSymbol {
+		const user = typeof userOrID === 'string' ? null : userOrID as User;
+		const id = user ? user.id : userOrID as ID;
+		if (user && !user.registered) return Auth.defaultSymbol();
```
xfix, line 219: This means that unregistered users won't get player rank in battles, is this intended?

Zarel:
Refs #7618

monsanto:
We need to be extra sure to not do permission checks with usernames in the server, or at least ensure that we are guarding the check with another check that the username is registered. I know the private PM log code has username checks, if that was to ever be upstreamed. I think the friends list does too. I think the rest of the username checks are in the client/website and are gated by the login server.

Besides that & Annika's concern seems OK to me

mia-pi-git:
> I think the friends list does too.

The friends list is gated on being autoconfirmed.

server/user-groups.ts
Zarel: Ah, no, that's not intended. But it may be preferable?

Zarel:
> Testing this locally, if I logged in as Annika with authentication, then in a private browsing window did `/trn Annika` I'd have access to the same user object with console access and so forth. Also, even if I didn't already have a session, I would gain trusted access, which is not ideal.

I thought we had code to prevent unregistered users from merging? I guess you bypassed because you had the same IP? I can add a guard for that.

server/user-groups.ts
Zarel: We could hardcode an exception for temporary rooms.

Zarel:
One thing I mulled was to just not allow people to take names of trusted users. I decided against that because I thought it might interfere with poke-env's use-case. But maybe I should revisit that post.

@hsahovic would that work? Only allow empty tokens for users that don't have any authority on the server?

hsahovic:
In principle, I think so. What would count as a user with authority on the server?

Zarel:
> In principle, I think so. What would count as a user with authority on the server?

"Known" means "any roomauth at all". (Which includes if a user has battled within the last ten minutes, no one else can use that account.)

There's also the weaker "trusted" which means "global voice or higher, or roomdriver or higher in a public/hidden room".

hsahovic:
In that case, I think that this would be too restrictive - I can think of several use cases which would require authenticating the same user several times in a couple of minutes. Could such a limitation be avoided?

Zarel:
It sounds like the "trusted" line would work for you, then?

monsanto:
I'd prefer "trusted" if possible, but if that is too restrictive, even just disallowing sysops would be a nice defense-in-depth thing to do

hsahovic:
Yeah, it sounds like "trusted" would work. I'm a bit shaky on what global voice / roomdriver entails though.

Zarel:
Are you familiar with PS's rank system?

`+` Voice (general recognition with no moderation powers)
`%` Driver (junior moderator)
`@` Moderator
`&` Administrator

They can be global ranks or room ranks.

hsahovic:
Do new users have global voice?

AnnikaCodes:
They do not.

hsahovic:
In this case "Trusted" shoud be fine :)

server/users.ts
```diff
@@ -587,28 +587,18 @@ export class User extends Chat.MessageContext {
...
-					console.log(`desynced roomgame ${roomid} renaming ${this.id} -> ${userid}`);
-					continue;
-				}
-				if (room.game.allowRenames || !this.named) continue;
-				this.popup(`You can't change your name right now because you're in ${room.game.title}, which doesn't allow renaming.`);
-				return false;
+	async validateToken(token: string, name: string, userid: ID, connection: Connection) {
+		if (!token && Config.noguestsecurity) {
+			if (Users.isTrusted(userid)) {
+				this.send(`|nametaken|${name}|You need an authentication token to be a trusted user.`);
```
AnnikaCodes, line 593:
```suggestion
				this.send(`|nametaken|${name}|You need an authentication token to log in as a trusted user.`);
```

AnnikaCodes: The original error messages reads to me as "you can log in, you just won't be marked as trusted"

server/users.ts
```diff
@@ -730,27 +745,36 @@ export class User extends Chat.MessageContext {
...
+			// unregistered users can only merge in limited situations
+			let canMerge = registered && conflictUser.registered;
+			if (
+				!registered && !conflictUser.registered && conflictUser.latestIp !== this.latestIp &&
+				!conflictUser.connected
+			) {
+				canMerge = true;
+			}
+			if (!canMerge) {
+				if (newlyRegistered && registered) {
```
Zarel, line 761:
```suggestion
				if (registered && !conflictUser.registered) {
```

server/users.ts
```diff
@@ -730,27 +745,36 @@ export class User extends Chat.MessageContext {
...
-			if (newlyRegistered && userType !== '1') {
-				if (conflictUser !== this) conflictUser.resetName();
-			} else {
-				this.send(`|nametaken|${name}|Someone is already using the name "${conflictUser.name}".`);
-				return false;
+		if (conflictUser) {
+			// unregistered users can only merge in limited situations
+			let canMerge = registered && conflictUser.registered;
+			if (
+				!registered && !conflictUser.registered && conflictUser.latestIp !== this.latestIp &&
```
Zarel, line 755:
```suggestion
				!registered && !conflictUser.registered && conflictUser.latestIp === this.latestIp &&
```

config/config-example.js
```diff
@@ -390,6 +390,16 @@ exports.inactiveuserthreshold = 1000 * 60 * 60;
  */
 exports.autolockdown = true;
 
+/**
+ * noguestsecurity - purely for development servers: allows logging in without
+ * a signed token: simply send `/trn [USERNAME]`. This allows using PS without
+ * a login server.
+ *
+ * Logging in this way will make you considered an unregistered user and grant
+ * no authority.
```
Zarel, line 399:
```suggestion
 * Logging in this way will make you considered an unregistered user and grant
 * no authority. You cannot log into a trusted (g+/r%) user account this way.
```

Zarel:
Safe to hotpatch! Won't take effect upon a restart, though.

server/user-groups.ts
Zarel: (I ended up replacing this with a trusted user check, which no longer has this problem.)

Zarel:
3/3!

PR #7626, opened 2020-11-02
Introduce Config.noipchecks

mia-pi-git:
See https://github.com/smogon/pokemon-showdown/pull/7618#issuecomment-720556024.

### DIFF
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -354,7 +354,6 @@ export class HelpTicket extends Rooms.RoomGame {
 		return Punishments.roomUnpunish('staff', user, 'TICKETBAN');
 	}
 	static checkBanned(user: User) {
-		if (Config.noipchecks) return false;
 		const staffRoom = Rooms.get('staff');
 		if (!staffRoom) return;
 		const punishment = Punishments.getRoomPunishType(staffRoom, user.id);
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -363,16 +363,14 @@ export class HelpTicket extends Rooms.RoomGame {
 		// skip if the user is autoconfirmed and on a shared ip
 		if (Punishments.sharedIps.has(user.latestIp) && user.autoconfirmed) return false;
 
-		if (!Config.noipchecks) {
-			for (const ip of user.ips) {
-				const curPunishment = Punishments.roomIps.get('staff')?.get(ip);
-				if (curPunishment && curPunishment[0] === 'TICKETBAN') {
-					const [, userid,, reason] = curPunishment;
-					return (
-						`You are banned from creating help tickets` +
-						`${userid !== user.id ? `, because you have the same IP as ${userid}` : ''}. ${reason ? `Reason: ${reason}` : ''}`
-					);
-				}
+		for (const ip of user.ips) {
+			const curPunishment = Punishments.roomIps.get('staff')?.get(ip);
+			if (curPunishment && curPunishment[0] === 'TICKETBAN') {
+				const [, userid,, reason] = curPunishment;
+				return (
+					`You are banned from creating help tickets` +
+					`${userid !== user.id ? `, because you have the same IP as ${userid}` : ''}. ${reason ? `Reason: ${reason}` : ''}`
+				);
 			}
 		}
 		return false;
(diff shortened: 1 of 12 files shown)


server/chat-plugins/helptickets.ts
```diff
@@ -363,14 +363,16 @@ export class HelpTicket extends Rooms.RoomGame {
...
+		if (!Config.noipchecks) {
+			for (const ip of user.ips) {
+				const curPunishment = Punishments.roomIps.get('staff')?.get(ip);
+				if (curPunishment && curPunishment[0] === 'TICKETBAN') {
+					const [, userid,, reason] = curPunishment;
+					return (
+						`You are banned from creating help tickets` +
+						`${userid !== user.id ? `, because you have the same IP as ${userid}` : ''}. ${reason ? `Reason: ${reason}` : ''}`
+					);
+				}
```
Zarel, line 375: I thought we agreed to still apply ip checks to punishments?

mia-pi-git: Oh, forgot to remove this one. One sec.

server/chat-plugins/lottery.ts
```diff
@@ -46,12 +46,23 @@ function endLottery(roomid: RoomID, winners: string[]) {
...
+}
+
 function addUserToLottery(roomid: RoomID, user: User) {
 	const lottery = lotteries[roomid];
 	if (!lottery) return;
 	const participants = lottery.participants;
-	const userSignedup = participants[user.latestIp] || Object.values(participants).map(toID).includes(user.id);
-	if (!userSignedup) {
+	const signedUp = isSignedUp(roomid, user);
+	if (!signedUp) {
```
Zarel, line 65:
```suggestion
	if (!isSignedUp(roomid, user)) {
```

Presumably you don't need the const anymore.

server/chat-plugins/the-studio.ts
```diff
@@ -374,7 +375,7 @@ class RecommendationsInterface {
 		if (!rec.liked) {
 			rec.liked = {ips: [], userids: []};
 		}
-		if (rec.liked.ips.includes(liker.latestIp) || rec.liked.userids.includes(liker.id)) {
+		if ((!ALLOW_ALTS && rec.liked.ips.includes(liker.latestIp)) || rec.liked.userids.includes(liker.id)) {
```
Zarel, line 378: Why the const? It looks like it's only used in one place?

Zarel: (The const makes it so Config updates don't apply instantly.)

server/monitor.ts
```diff
@@ -213,6 +217,7 @@ export const Monitor = new class {
 	 * Counts group chat creation. Returns true if too much.
 	 */
 	countGroupChat(ip: string) {
+		if (Config.noipchecks) return;
```
Zarel, line 220: Most of these should be `return false` rather than `return`. Look at the function signatures.

server/tournaments/index.ts
```diff
@@ -12,7 +12,7 @@ const AUTO_START_MINIMUM_TIMEOUT = 30 * 1000;
 const MAX_REASON_LENGTH = 300;
 const MAX_CUSTOM_NAME_LENGTH = 100;
 const TOURBAN_DURATION = 14 * 24 * 60 * 60 * 1000;
-const ALLOW_ALTS = false;
+const ALLOW_ALTS = !!Config.noipchecks;
```
Zarel, line 15: tbh if you could convert all existing `ALLOW_ALTS` to `Config.noipchecks` that'd be great.

Zarel: (Because of the hotpatch thingy)

mia-pi-git: Done.

Zarel:
1/3!

server/ladders.ts
```diff
@@ -465,7 +465,7 @@ class Ladder extends LadderStore {
 		// users must be different
 		if (user1 === user2) return false;
 
-		if (Config.fakeladder) {
+		if (Config.noipchecks) {
```
Zarel, line 468: Hm I wonder if `noipchecks` should necessary disable the rest of matchmaking? The way this is implemented, users can get matched with each other twice in a row, as well as get matched with people with wildly different Elo ratings. poke-env is probably happy with this, but it should probably be documented?

PR #7631, opened 2020-11-03
NatDex Mono:  Fix an oversight

DarkPhoenix911:
Ban Zekrom.  Ban it!  Send it to the shadow realm!!

PR #7632, opened 2020-11-03
Fix an oversight in NatDex Mono

DarkPhoenix911:
Send Lunala to the shadow realm too!


mia-pi-git:
Why not just do this in #7631? Doesn't need to be 2 PRs.

Zarel:
For these sorts of things, could you please link to the thread in which the ban was announced?

DarkPhoenix911:
Yeah you're right Mia.  Was on mobile and this way was just easier.  Fixed it though.

PR #7634, opened 2020-11-04
Split up dex-data into more files

Zarel:
This commit introduces:
- `dex-abilities.ts`
- `dex-conditions.ts`
- `dex-formats.ts`
- `dex-items.ts`
- `dex-moves.ts`
- `dex-species.ts`

These files centralize definitions from `dex-data` and `global-types`.


Zarel:
Next steps:

- Move definitions from `Move` to `DataMove` so they can be merged into a single `Move` type/class (and similarly for the others)
- Start work on new Dex API (`Dex.moves.get()`, `Dex.moves.all()`)

sim/dex.ts
```diff
@@ -1596,3 +1533,13 @@ dexes['base'] = new ModdedDex(undefined, true);
...
+export namespace Dex {
+	export type Species = import('./dex-species').Species;
+	export type Item = import('./dex-items').Item;
+	export type Move = import('./dex-moves').Move;
+	export type Ability = import('./dex-abilities').Ability;
+
+	export type HitEffect = import('./dex-moves').HitEffect;
+	export type SecondaryEffect = import('./dex-moves').SecondaryEffect;
+	export type RuleTable = import('./dex-formats').RuleTable;
+}
```
Zarel, line 1545: Am very annoyed I have to do this but I don't see any other way. TypeScript being TypeScript... :/

data/scripts.ts
```diff
@@ -1048,7 +1050,7 @@ export const Scripts: BattleScriptsData = {
 		if (!moveData.secondaries) return;
 		for (const target of targets) {
 			if (target === false) continue;
-			const secondaries: SecondaryEffect[] =
+			const secondaries: Dex.SecondaryEffect[] =
```
KrisXV, line 1053: ? Shouldn’t this be `this.dex`?

Zarel: No, it's a type.

Zarel: It does not even seem possible to define a type in `this.dex` :(

KrisXV: oh im dumb

also, rip

PR #7637, opened 2020-11-04
Support Config.nothrottle

mia-pi-git:
As per #7618. Think i got everything related to this.


config/config-example.js
```diff
@@ -257,6 +257,12 @@ exports.reportbattlejoins = true;
  */
 exports.monitorminpunishments = 3;
 
+/**
+ * Turns off all time-based throttles - rename, challenges, laddering, etc.
+ */
+
```
Zarel, line 263:
```suggestion
```

Zarel:
2/3!

PR #7638, opened 2020-11-04
Properly update Diamond Storm boosts for Gen 7

urkerab:
As researched by @DaWoblefet, Diamond Storm boosts only apply once per move, like Clangorous Soulblaze. To get this to work correctly:

- Moved boost from `secondary` to `selfBoost` except in Gen 6
- Added support for `chance` on `selfBoost`, including Serene Grace and Fire+Water Pledge combo rainbow
- Sheer Force deletes the `selfBoost` if the move also has secondaries (dummy secondary added to Diamond Storm to enable this)

### DIFF
--- a/sim/dex-moves.ts
+++ b/sim/dex-moves.ts
@@ -186,7 +186,7 @@ export interface MoveData extends EffectData, MoveEventMethods, HitEffect {
 	struggleRecoil?: boolean;
 	secondary?: SecondaryEffect | null;
 	secondaries?: SecondaryEffect[] | null;
-	self?: HitEffect | null;
+	self?: SecondaryEffect | null;
 
 	// Hit effect modifiers
 	// --------------------
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3180,6 +3180,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 					if (secondary.chance) secondary.chance *= 2;
 				}
 			}
+			if (move.self?.chance) move.self.chance *= 2;
 		},
 		name: "Serene Grace",
 		rating: 3.5,
@@ -3233,6 +3234,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			if (move.secondaries) {
 				delete move.secondaries;
 				// Technically not a secondary effect, but it is negated
+				delete move.self;
 				if (move.id === 'clangoroussoulblaze') delete move.selfBoost;
 				// Actual negation of `AfterMoveSecondary` effects implemented in scripts.js
 				move.hasSheerForce = true;
--- a/data/mods/gen6/moves.ts
+++ b/data/mods/gen6/moves.ts
@@ -16,6 +16,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	},
 	diamondstorm: {
 		inherit: true,
+		self: null,
 		secondary: {
 			chance: 50,
 			self: {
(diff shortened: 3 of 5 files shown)


Zarel:
Does this apply to all secondaries? Should we instead manually detect `secondary.self` and roll it during the self phase rather than the secondary phase?

urkerab:
Secondaries normally run once per hit, rather than once per move.

urkerab:
That being said, there's an edge case scenario - Diamond Storm with Parental Bond skill swapped on to the user, with one active foe, while the rainbow Pledge combo is active. Hopefully it only boosts the user once. If not, then we need a new approach to secondaries...

DaWoblefet:
Bad news - Diamond Storm will raise Defense twice during Parental Bond in Pledge rainbow (you gain +2 with the first hit, then +2 with the second hit).

sim/dex-moves.ts
```diff
@@ -170,7 +170,7 @@ export interface MoveData extends EffectData, MoveEventMethods, HitEffect {
 	heal?: number[] | null;
 	forceSwitch?: boolean;
 	selfSwitch?: string | boolean;
-	selfBoost?: {boosts?: SparseBoostsTable};
+	selfBoost?: SecondaryEffect | null;
```
Zarel, line 173: It seems like it would make more sense to call this `selfSecondary`?

Marty-D:
We could get around the Parental Bond thing by abusing the fact that spread moves and multi-hit moves are mutually exclusive. Just keep the original implementation as-is and add the `selfBoost` on top of it, running `secondaries` only if the Diamond Storm is `multihit`, and `selfBoost` only if it isn't `multihit`.

urkerab:
Actually I was confused because I thought that we processed spread moves by applying hits to each target in turn, but in fact we apply hits to each target in turn. This means that the processing of the `self` property actually happens at the right time to make Diamond Storm boosts happen once per hit of Parental Bond. Of course, I still need to remove it when Sheer Force is active (another special case unfortunately) and also make the chance apply of course.

DaWoblefet:
Doubly confirming that Sheer Force negates the Defense boost potential and powers up Diamond Storm btw.

DaWoblefet:
Also, if possible, could we merge this even if the edge cases are buggy? As long as Diamond Storm will function properly on Diancie, that will be good enough to get us started. I mention this because Doubles OU is debating a Diancie ban and this mechanic impacts that deliberation significantly, as you might guess.

Zarel:
The main thing I want to understand before I merge is why this is called `selfBoost` rather than `selfSecondary`. It seems to be implemented as similar to a secondary effect in every way except it only happens once when using a spread move.

Zarel:
Especially since you don't put boosts directly inside it, but rather in `selfBoost.boosts`. Calling it `selfSecondary.boosts` would make it line up better with `secondary.boosts`, which would also be easier to understand.

urkerab:
@DaWoblefet Sorry for the delay, but I've now updated the code:

- Stat boost happens after each hit of Parental Bond
- Stat boost chance applies
- Serene Grace or Pledge Rainbow doubles stat boost chances
- Sheer Force removes stat boost

Marty-D (approved):
I'm also confused, since when does `self` run after all hits of a spread move and not between hits? The only reason I implemented `selfBoost` originally is because `self` didn't work for Clanging Scales and later much more importantly Clangorous Soulblaze. Now only Scale Shot requires it? Time to figure out how to refactor it out of existence.

This looks good then, thanks urkerab!

urkerab:
> I'm also confused, since when does `self` run after all hits of a spread move and not between hits?

Maybe this got "fixed" by the spread move event order refactor?

urkerab:
In fact, `selfBoost` probably doesn't work for Clanging Scales + Parental Bond.

PR #7639, opened 2020-11-04
Don't reveal Illusion in Camomons

urkerab:
Also removed `onAfterMega` as it's not relevant in Gen 8.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -922,10 +922,10 @@ export const Formats: FormatList = [
 			return {...species, types: types};
 		},
 		onSwitchIn(pokemon) {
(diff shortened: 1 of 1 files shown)


TheImmortal:
I think it's relevant for Camomons / Mix and Mega mashup?

KrisXV:
I would prefer keeping onAfterMega so this can be added to old gens metas as a rule and still work, so you might want to consider moving the Camomons stuff into its own Rule in `rulesets.ts`.

Also, I spoke with TI about this a little bit ago, and iirc we ultimately decided on not doing this?

urkerab:
I've restored `onAfterMega`. I'm not sure what else you might have been looking for though.

PR #7640, opened 2020-11-05
Support `npm run fix`

Zarel:


Zarel:
We decided on `fix` rather than `lint-fix` because if we ever had other fixers, we'd probably want to integrate them into this command anyway.

This should be more convenient than trying to remember or figure out how to get eslint to fix things.

PR #7644, opened 2020-11-05
Require a digit to follow port number in links

xfix:
This fixes a bug which caused lines like `rooms.js: does room stuff` to be incorrectly linked.


Zarel:
Thanks!

PR #7645, opened 2020-11-05
Allow the Help auto-response filter to be used in all rooms

mia-pi-git:
This was suggested when the Help filter was first created, and I see no reason other rooms can't benefit from it.


server/chat-plugins/responder.ts
```diff
@@ -0,0 +1,517 @@
...
+	answererData = JSON.parse(FS(PATH).readSync());
+} catch (e) {}
+
+function loadOldData() {
+	const oldDataPath = 'config/chat-plugins/help.json';
+	const oldPath = FS(oldDataPath);
+	if (oldPath.existsSync() && Config.helpFilterRoom) {
+		const oldData = JSON.parse(oldPath.readSync());
+		answererData[toID(Config.helpFilterRoom)] = oldData;
+		oldPath.unlinkIfExistsSync();
```
Zarel, line 31: Maybe wait to make sure it works correctly on the live server?

mia-pi-git: I thoroughly tested it locally. I can remove the unlink, though.

mia-pi-git: Removed the unlink. Now it'll check it, but only set the current data to the old data if it hasn't been changed since.

PR #7646, opened 2020-11-05
Remove hotpatch notifications from Upper Staff

xfix:
Hotpatch notifications in Upper Staff aren't very useful considering Staff already gets those.


Zarel:
They're easier to see in US though :(

PR #7648, opened 2020-11-05
Add a --no-security flag

mia-pi-git:
Not sure if i like how the start script handles this.

### DIFF
--- a/pokemon-showdown
+++ b/pokemon-showdown
@@ -86,8 +86,8 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 	case 'start':
 		{
 			process.argv.splice(2, 1);
-			if (process.argv[2] === '--skip-build') {
-				process.argv.splice(2, 1);
+			if (process.argv.includes('--skip-build')) {
+				process.argv.splice(process.argv.indexOf('--skip-build'), 1);
 				built = true;
 			}
 			if (!built) build();
--- a/pokemon-showdown
+++ b/pokemon-showdown
@@ -87,7 +87,6 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 		{
 			process.argv.splice(2, 1);
 			if (process.argv.includes('--skip-build')) {
-				process.argv.splice(process.argv.indexOf('--skip-build'), 1);
 				built = true;
 			}
 			if (!built) build();
--- a/server/config-loader.ts
+++ b/server/config-loader.ts
@@ -15,6 +15,10 @@ export type ConfigType = typeof defaults & {
 	greatergroupscache: {[combo: string]: GroupSymbol},
 	[k: string]: any,
 };
+/** Map<process flag, config settings for it to turn on> */
+const FLAG_PRESETS = new Map([
+	['--no-security', ['nothrottle', 'noguestsecurity', 'noipchecks']],
+]);
 
 const CONFIG_PATH = require.resolve('../config/config');
 
@@ -30,6 +34,12 @@ export function load(invalidate = false) {
 			require('better-sqlite3');
 		} catch (e) {
 			throw new Error(`better-sqlite3 is not installed or could not be loaded, but Config.usesqlite is enabled.`);
+		}
+	}
+
+	for (const [preset, values] of FLAG_PRESETS) {
+		if (process.argv.includes(preset)) {
+			for (const value of values) config[value] = true;
 		}
 	}
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -163,7 +163,11 @@ if (require.main === module) {
 	// in the case of app.js being imported as a module (e.g. unit tests),
 	// postpone launching until app.listen() is called.
 	let port;
-	if (process.argv[2]) port = parseInt(process.argv[2]);
+	for (const arg of process.argv) {
+		const parsed = parseInt(arg);
+		if (isNaN(parsed)) continue;
+		port = parsed;
+	}
 	Sockets.listen(port);
 }
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -164,9 +164,10 @@ if (require.main === module) {
 	// postpone launching until app.listen() is called.
 	let port;
 	for (const arg of process.argv) {
-		const parsed = parseInt(arg);
-		if (isNaN(parsed)) continue;
-		port = parsed;
+		if (/^[0-9]+$/.test(arg)) {
+			port = parseInt(arg);
+			break;
+		}
 	}
 	Sockets.listen(port);
 }


Slayer95:
Every decision that decreases security should be a conscious choice and absolutely cannot be implemented through generic catch-all flags. If someone needs to regularly dismiss several security features they should write their own bash script to exec PS with their required args.

pokemon-showdown
```diff
@@ -86,8 +86,8 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 	case 'start':
 		{
 			process.argv.splice(2, 1);
-			if (process.argv[2] === '--skip-build') {
-				process.argv.splice(2, 1);
+			if (process.argv.includes('--skip-build')) {
+				process.argv.splice(process.argv.indexOf('--skip-build'), 1);
```
Zarel, line 90: You don't have the splice for `--no-security`. I suspect it might not be necessary.

Zarel: Probably should have all the argv's in one place?

Zarel:
> Every decision that decreases security should be a conscious choice and absolutely cannot be implemented through generic catch-all flags. If someone needs to regularly dismiss several security features they should write their own bash script to exec PS with their required args.

To be clear, none of the options disabled here are particularly dangerous. They're all ones you want disabled while doing local development testing. The name is intentionally a bit scary so people don't try turning it on their prod servers, but they're things I think a wide variety of people would want disabled for test use-cases.

server/config-loader.ts
```diff
@@ -33,6 +37,12 @@ export function load(invalidate = false) {
 		}
 	}
 
+	for (const [preset, values] of FLAG_PRESETS) {
+		if (process.argv.includes(preset)) {
+			for (const value of values) config[value] = true;
+		}
+	}
```
Zarel, line 44: How is the port detected? That's probably relevant if you're not going to remove the settings from `process.argv`.

mia-pi-git: I'd rather just remove the flag from process.argv and let port be handled normally, if I'm understanding you correctly?

mia-pi-git: a9b4303 should do it.

Zarel: I don't like how you do the removal and detection in separate loops. It's easy enough just to skip everything starting with `-` when detecting the port.

mia-pi-git: There, does that work better?

server/index.ts
```diff
@@ -163,7 +163,11 @@ if (require.main === module) {
 	// in the case of app.js being imported as a module (e.g. unit tests),
 	// postpone launching until app.listen() is called.
 	let port;
-	if (process.argv[2]) port = parseInt(process.argv[2]);
+	for (const arg of process.argv) {
+		const parsed = parseInt(arg);
+		if (isNaN(parsed)) continue;
+		port = parsed;
```
Zarel, line 169:
```suggestion
		if (/^[0-9]+$/.test(arg)) {
			port = parseInt(arg);
			break;
		}
```

Remember, `parseInt('98.3 oifajew')` is `98`.

Zarel:
@Slayer95 I _am_ open to a different name, though. Something like "dev-mode".

@xfix @monsanto do either of you think this is a bad idea for security reasons?

Slayer95:
Of course the name "no-security" doesn't make any sense. "Dev-Mode", on the other hand, doesn't convey anything about what it actually does, and it makes it seem as if this flag had no risks other than maybe stability, or performance degradation due to profiling (which the developer might expect if they turn on a "dev mode").

I shall also take the chance to point out that the setting ``noguestsecurity`` can easily become an attack vector against bots that may connect to such a server and accept console commands from impersonators. And yes, I know for certain that there are plenty of bots out there which would be vulnerable.

Slayer95:
Now, if we go back to the drawing board, and consider the use case:

> they're all ones you want disabled while doing local development testing

Then shouldn't we look for a setting which deals with that more literally such as some "trust-loopback", which disables checks for connections from ``127.0.0.1/8`` or ``::1/128`` ?

Zarel:
> Then shouldn't we look for a setting which deals with that more literally such as some "trust-loopback", which disables checks for connections from `127.0.0.1/8` or `::1/128` and also sets the bind address?

My concern is that you might very well want to test e.g. a bot hosted on a different server. I don't see a need for finer-grained permissions when they're only used in development test servers and none of them are particularly dangerous in that context; I think it's more worthwhile to have one setting that immediately flips all this stuff for anything that doesn't test security.

xfix:
I don't think we should trust loopback without requiring a server owner to explicilty pick such an option, you may as well host a Showdown server with, say, nginx proxy pass (I know I did that back when I hosted my server).

I like the name `--no-security` myself, makes it rather clear you shouldn't use it in production.

Slayer95:
> you may as well host a Showdown server with, say, nginx proxy pass

PS supports ``X-Forwarded-For`` through ``Config.noproxy`` though, and sets the connection ``ip`` accordingly.

As for the rest of both your comments, guess we can only disagree. ``</thread>``

xfix:
> PS supports `X-Forwarded-For` through `Config.noproxy` though, and sets the connection `ip` accordingly.

That's assuming you know about it, I know I personally ran a server for a month before realizing this option exists as misconfiguring it isn't immediately obvious, and even if you do notice everybody having `::1` as an IP address doesn't mean it's clear what needs to be changed. Also note that `config-example.js` discourages changing the value of this option.

```javascript
/**
 * proxyip - proxy IPs with trusted X-Forwarded-For headers
 *   This can be either false (meaning not to trust any proxies) or an array
 *   of strings. Each string should be either an IP address or a subnet given
 *   in CIDR notation. You should usually leave this as `false` unless you
 *   know what you are doing
 * @type {false | string[]}.
 */
exports.proxyip = false;
```

That being said, personally I'm not a fan of `noguestsecurity`. If you want to test stuff with unregistered users, you may as well type `/nick ` and smash random letters on keyboard (this is very likely to be an unregistered account), and if you need to test something where permissions matter, well, this won't help with that.

The other options are fine with me, and definitely something Showdown should support.

Zarel:
> That being said, personally I'm not a fan of `noguestsecurity`. If you want to test stuff with unregistered users, you may as well type `/nick ` and smash random letters on keyboard (this is very likely to be an unregistered account), and if you need to test something where permissions matter, well, this won't help with that.

This is mostly for `poke-env`, which wants to write AI battle bots without going through the hassle of a login handshake.

We could maybe require the username prefix "Guest", for slightly more security?

xfix:
Fair then.

Zarel:
I imagine `noguestsecurity` would also be useful to others who want to write unit tests for their bots or clients, but don't want the tests to require network use (it's also possible to generate your own login keypair, but that's undocumented and a lot more effort).

Zarel:
So we decided against #7690. I do like the untrusted symbol. Maybe a permanent `!` trumping all other symbols, in front of untrusted users?

Untrusted users being marked however, maybe `user.trusted === false`?

hsahovic:
What is the status on this PR?
I tested the config values that were introduced beforehand (`nothrottle`, `noguestsecurity` and `noipchecks`) and they are performing as well as expected :)

Zarel:
I have some nitpicks but it's worth merging as-is and nitpicks can be fixed later.

Zarel:
Thanks, Mia!

PR #7652, opened 2020-11-06
Chatlog: Support room aliases in /linecount

AnnikaCodes:


server/chat-plugins/chatlog.ts
```diff
@@ -1047,10 +1047,10 @@ export const commands: ChatCommands = {
 	roomstats: 'linecount',
 	linecount(target, room, user) {
 		let [roomid, month, userid] = target.split(',').map(item => item.trim());
-		const tarRoom = roomid ? toID(roomid) : room?.roomid;
-		if (!tarRoom) return this.errorReply(`You must specify a room.`);
+		const tarRoom = (roomid ? Rooms.get(toID(roomid)) : room);
```
mia-pi-git, line 1050: if you want aliases, you want to use `Rooms.search` here, no?

mia-pi-git: Since otherwise `Rooms.get` on its own doesn't handle aliases.

AnnikaCodes: Oh, right. I forgot Rooms#get doesn't account for aliases.

Zarel: Easy mistake to make, because `Users.get` and `Dex.getMove` both do account for aliases. Maybe Rooms should be fixed. On the other hand, I think the defaults for a lot of these should be "no aliases" – only user-facing functions need to grab aliases.

PR #6931, opened 2020-06-29
Adjust fractional priority for very latest research

urkerab:
PR #6928:

> I also assumed that Custap Berry, should it ever be released, will activate in preference to Quick Draw.

Me and my big mouth... Anubis later proved me wrong again. This backs out the changes to Quick Draw and applies them to Custap Berry instead.

PR #7653, opened 2020-11-07
Fix custap berry with gluttony

urkerab:
Stupid error in PR #6931 where I inserted the `priority <= 0 && ` without bracketing the existing `||` conditions first (confusing someone into helpfully bracketing them the wrong way). Since the only other ways to get a positive fractional priority is either with Quick Claw (and therefore not Custap Berry) or Quick Draw (and therefore not Gluttony), this doesn't affect normal play, but some OMs might misbehave, if anyone wanted to combine these for some reason.


Zarel:
Thanks!

PR #7654, opened 2020-11-07
Fix two-turn submoves + Ally Switch in doubles

MacChaeger:
Previously, two-turn moves called by another move such as Metronome would target the user’s ally if Ally Switch was used to make the user and its ally trade places on the second turn. This change makes the move correctly choose a random target on the first turn instead of the second and makes the battle engine respect that choice instead of defaulting to what the user’s previous move targeted (when the two-turn move was called by a self-targeting move, this would make the two-turn move target the user which would conveniently be interpreted as a random opponent except when the user and ally traded places).

### DIFF
--- a/sim/side.ts
+++ b/sim/side.ts
@@ -445,12 +445,16 @@ export class Side {
 
 		const lockedMove = pokemon.getLockedMove();
 		if (lockedMove) {
-			const lockedMoveTarget = pokemon.lastMoveTargetLoc || 0;
+			let lockedMoveTargetLoc = pokemon.lastMoveTargetLoc || 0;
+			const lockedMoveID = toID(lockedMove);
+			if (pokemon.volatiles[lockedMoveID] && pokemon.volatiles[lockedMoveID].targetLoc) {
+				lockedMoveTargetLoc = pokemon.volatiles[lockedMoveID].targetLoc;
+			}
 			this.choice.actions.push({
 				choice: 'move',
 				pokemon,
-				targetLoc: lockedMoveTarget,
-				moveid: toID(lockedMove),
+				targetLoc: lockedMoveTargetLoc,
+				moveid: lockedMoveID,
 			});
 			return true;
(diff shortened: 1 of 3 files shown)


MacChaeger:
*I say “correctly”, but @DaWoblefet is looking into what should happen when one or more of the potential slots that could be randomly targeted is unoccupied at the time when it would be chosen because the Pokémon there were KO’d.

sim/side.ts
```diff
@@ -445,12 +445,16 @@ export class Side {
 
 		const lockedMove = pokemon.getLockedMove();
 		if (lockedMove) {
-			const lockedMoveTarget = pokemon.lastMoveTargetLoc || 0;
```
Zarel, line 448: I remember you were asking me how this worked, and I couldn't figure it out, either. I'm glad you figured it out!

(I think this should probably be documented in the `lockedmove` and `twoturnmove` conditions in `conditions.ts`.

Zarel:
Looks great to me! Tell me when you get the research stuff figured out.

DaWoblefet:
I believe this covers the use cases asked about:

- A regular Dig / Future Sight will respect the target that the player selected. That is, using Dig / Future Sight into the opponent's left slot will still attempt to target the opponent's left slot, even if the opponent's left slot is fainted or both opponents are fainted. Same for the right slot.
- A called Future Sight (I did all my testing via Sleep Talk), when one of the opponent's slots is fainted will always target the remaining unfainted slot. It is not random.
- A called Future Sight when both of the opponent's slots are fainted will target either opponent at random.

I cannot easily test called Dig since it's uncallable by Assist and Sleep Talk. However, I would expect identical behavior to Future Sight.

MacChaeger:
What I already have should work for that regarding two turn moves, but future sight is another matter and I think outside the scope of this PR?

Marty-D (approved):
👍

PR #7655, opened 2020-11-07
Add a command to disable commands

mia-pi-git:
Requested by.. many people now, it feels like.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -565,13 +565,16 @@ export const commands: ChatCommands = {
 		}
 		const parsed = this.parseCommand(`/${target}`);
 		if (!parsed) {
-			return this.errorReply(`/${target} - Command not found.`);
+			return this.errorReply(`Command "/${target}" is in an invalid format.`);
 		}
 		const {handler, cmd} = parsed;
-		if (handler?.disabled) {
-			return this.errorReply(`That command is already disabled`);
+		if (!handler) {
+			return this.errorReply(`Command "/${target}" not found.`);
 		}
-		handler!.disabled = true;
+		if (handler.disabled) {
+			return this.errorReply(`Command "/${target}" is already disabled`);
+		}
+		handler.disabled = true;
 		this.addGlobalModAction(`${user.name} disabled the command /${cmd}.`);
 		this.globalModlog(`DISABLECOMMAND`, null, target);
 	},
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -557,6 +557,26 @@ export const commands: ChatCommands = {
 		this.sendReply("learnsets.js saved.");
 	},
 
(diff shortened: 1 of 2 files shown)


Zarel:
So this persists past hotpatches but not restarts. Is that a good line to draw? If we don't need it to persist past hotpatches, it can literally just be:

```ts
Chat.commands[command] = function () {
  this.errorReply("Sorry, this command isn't working due to technical difficulties, try again later.");
};
```

I think if it _does_ persist past hotpatches, hotpatches should announce "x command is disabled; can it be re-enabled?"

But I definitely don't want to _normalize_ taking a long time to fix disabled commands? I think if it's not fixed by one hotpatch, something is seriously wrong and we should probably push the command disabling into the repo (so side servers don't break). So I think the "not persist past hotpatches" solution might actually be the best solution.

Zarel:
(pre wanted the command disabling to persist past hotpatches and restarts, but that's because for a while we had crashes stay unfixed for weeks at a time. This is not actually normal and we should be fixing crashes within a dayish. In the future, if crashes stay unfixed for over a day, please bring them to my attention and I'll take care of them.)

server/chat.ts
```diff
@@ -472,6 +473,11 @@ export class CommandContext extends MessageContext {
 
 		try {
 			if (this.handler) {
+				if (this.handler.disabled) {
+					throw new Chat.ErrorMessage(
+						`The command /${this.cmd} is disabled. Please try again later.`
```
Zarel, line 478:
```suggestion
						`The command /${this.cmd} is temporarily unavailable due to technical difficulties. Please try again in a few hours.`
```

Zarel:
Oh, it would be nice to warn about re-enablings on hotpatch.

server/chat-commands/admin.ts
```diff
@@ -557,6 +557,26 @@ export const commands: ChatCommands = {
...
+		}
+		const parsed = this.parseCommand(`/${target}`);
+		if (!parsed) {
+			return this.errorReply(`/${target} - Command not found.`);
+		}
+		const {handler, cmd} = parsed;
+		if (handler?.disabled) {
+			return this.errorReply(`That command is already disabled`);
+		}
+		handler!.disabled = true;
```
Zarel, line 574: Why is `handler` optional? That `?.` is probably wrong.

Zarel, line 574:
```suggestion
		const parsed = this.parseCommand(`/${target}`);
		if (!parsed) {
			return this.errorReply(`Command "/${target}" is in an invalid format.`);
		}
		const {handler, cmd} = parsed;
		if (!handler) {
			return this.errorReply(`Command "/${target}" not found.`);
		}
		if (handler.disabled) {
			return this.errorReply(`Command "/${target}" is already disabled`);
		}
		handler.disabled = true;
```

Zarel:
Thanks! But the warn thingy!

PR #7657, opened 2020-11-07
Chatlog: Automatically determine regex engine

xfix:
This makes ripgrep use PCRE2 for searching logs only when necessary.


Zarel:
Thanks!

PR #7658, opened 2020-11-07
Chatlog: Handle next result buttons properly

mia-pi-git:
At the moment, this uses the full regex, which is messy and doesn't work right.


Zarel:
Thanks!

PR #7659, opened 2020-11-08
Moderation: Properly log /hidetext's message

mia-pi-git:


Zarel:
```
// if the user hiding their own text, it would clear the "cleared" message,
// so we can't attribute it in that case
```

How did you deal with this issue?

mia-pi-git:
> ```
> // if the user hiding their own text, it would clear the "cleared" message,
> // so we can't attribute it in that case
> ```
> 
> How did you deal with this issue?

It adds the mod action after room.hideText is run.

Zarel:
> It adds the mod action after `room.hideText` is run.

Wouldn't that put the "show hidden messages" button in the wrong place?

mia-pi-git:
> > It adds the mod action after `room.hideText` is run.
> 
> Wouldn't that put the "show hidden messages" button in the wrong place?

Yes. Resolved by just using `this.roomlog`. Edited the comment as requested.

server/chat-commands/moderation.ts
```diff
@@ -1660,13 +1660,13 @@ export const commands: ChatCommands = {
 
 		// if the user hiding their own text, it would clear the "cleared" message,
 		// so we can't attribute it in that case
+		// and sending the message after the hideText call puts the "show lines" button in the wrong place
```
Zarel, line 1663:
```suggestion
		// and sending the message after `|unlink|` puts the "show lines" button in the wrong place
```

Zarel:
ty!

PR #7660, opened 2020-11-09
Admin: Warn the user of disabled commands on hotpatch/lockdown

mia-pi-git:
Not sure if this should make /hotpatch require force. Personally, i think it should, but i'll leave it up for debate.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -17,6 +17,21 @@ import {Utils} from '../../lib/utils';
 
 import * as ProcessManager from '../../lib/process-manager';
 
+function getDisabledCommands(table: ChatCommands = Chat.commands): string[] {
+	const disabled = [];
+	for (const k in table) {
+		const handler = table[k];
+		if (Array.isArray(handler) || typeof handler === 'string') continue;
+		if (typeof handler === 'object') {
+			disabled.push(...getDisabledCommands(handler));
+		}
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -559,10 +581,10 @@ export const commands: ChatCommands = {
 
 	disablecommand(target, room, user) {
 		this.checkCan('makeroom');
-		target = toID(target);
 		if (!toID(target)) {
 			return this.parse(`/help disablecommand`);
 		}
+		if (['!', '/'].some(c => target.startsWith('c'))) target = target.slice(1);
```
iscke, line 587: did you mean for the `'c'`?

mia-pi-git: lol no i'm dumb

Zarel:
Thanks!

Everyone else should feel free to comment on "should this require forcehotpatch". I don't have a strong opinion here.

AnnikaCodes:
Tbh I think that instead of `/forcehotpatch` it should require you to type the `/hotpatch` command again, just like `/unbanall`. If I update private code and hotpatch it I might overwrite the disabled commands without noticing.

Zarel:
Honestly forcehotpatch has always been dumb, to me. I only added it because pre wanted it so much. Does anyone else even benefit from it? I personally would rather it be gone.

Zarel:
Making `/hotpatch` always do a rebuild before hotpatching might be a better approach tbh.

PR #7661, opened 2020-11-09
CommandContext: Move parseCommand into Chat

mia-pi-git:
This is useful logic that would be nice to have elsewhere (not needing a command context).


Zarel:
Looks good!

PR #7664, opened 2020-11-09
Helptickets: Use globalModAction for /ht ban

mia-pi-git:
Nitpick. addModAction means basically nothing since the room is deleted immediately.


Zarel:
tbh is that a good reason to switch to `globalModAction`? Do people in the Staff room actually care?

PR #7667, opened 2020-11-09
Info: Make /rules account for other possible use-cases

mia-pi-git:
Apparently a lot of people make the mistake of `/rules [room]` giving the rules for the given room, and of `/rules format` giving rules for that format. So, couldn't hurt to make that work.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -1971,7 +1971,8 @@ export const commands: ChatCommands = {
 	},
 
 	rule: 'rules',
-	rules(target, room, user) {
+	roomrules: "rules",
+	rules(target, room, user, connection, cmd) {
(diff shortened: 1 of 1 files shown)


Zarel:
A lot of rooms are also formats. How do you propose we deal with this?

mia-pi-git:
> A lot of rooms are also formats. How do you propose we deal with this?

Maybe it'd be best to prompt the user asking which they want, and then provide a `/roomrules` for room rules, and use /tier if they want a format?

Zarel:
Hm, I think that's probably the best solution.

mia-pi-git:
> Hm, I think that's probably the best solution.

Done.

Zarel:
Thanks!

PR #7669, opened 2020-11-09
Punishments: Don't warn / mark staff users in banned groupchats

mia-pi-git:
Found out this was a thing when i groupchatbanned someone in the target GC.


Zarel:
The change to `getSubRooms` seems to make this not hotpatchable?

mia-pi-git:
> The change to `getSubRooms` seems to make this not hotpatchable?

That was a botched merge, I believe. One sec.

mia-pi-git:
> > The change to `getSubRooms` seems to make this not hotpatchable?
> 
> That was a botched merge, I believe. One sec.

Yeah, now we can /hotpatch punishments and /hotpatch roomsp.

mia-pi-git:
But, this is probably pointless now. GCBans are deprecated.

Zarel:
Unfortunate. :(

PR #7672, opened 2020-11-10
National Dex Monotype: Unban Aegislash

DarkPhoenix911:
Freeing the once-banned Sword and Shield Mon once more!  
https://www.smogon.com/forums/threads/monotype-om-mega-thread.3660464/post-8652440


Zarel:
Thanks!

PR #7675, opened 2020-11-10
Rooms: Properly rename private battle rooms

mia-pi-git:
This is causing the /roomban crash.
Relevant monkeypatch: 
```ts
>> Rooms.GameRoom.prototype.makePublic = function() {
		this.settings.isPrivate = false;
		if (!this.roomid.endsWith('pw')) return true;
		const {id} = this.getReplayData();
		this.rename(this.title, `battle-${id}`);
	}


Zarel:
tyty!

PR #7677, opened 2020-11-10
The Cafe: Stop holding a reference to the room long-term

mia-pi-git:
This causes crashes if the room doesn't exist.


Zarel:
ty!

PR #7679, opened 2020-11-10
 Exclude room-specific commands for other rooms in /permissions

mia-pi-git:
Does what it says on the tin. This is much easier now with `this.requireRoom(roomid)`.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -43,7 +43,7 @@ export type ChatHandler = (
 	message: string
 ) => void;
 export type AnnotatedChatHandler = ChatHandler & {
-	requiresRoom: boolean,
+	requiresRoom: boolean | RoomID,
 	hasRoomPermissions: boolean,
 	broadcastable: boolean,
 	cmd: string,
@@ -1720,7 +1720,7 @@ export const Chat = new class {
 			if (typeof entry !== 'function') continue;
 
 			const handlerCode = entry.toString();
-			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`).\)/.exec(handlerCode)?.[1] || /this\.requireRoom\(/.test(handlerCode);
+			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`) /.exec(handlerCode)?.[1] as RoomID || /this\.requireRoom\(/.test(handlerCode);
 			entry.hasRoomPermissions = /\bthis\.(checkCan|can)\([^,)\n]*, [^,)\n]*,/.test(handlerCode);
 			entry.broadcastable = cmd.endsWith('help') || /\bthis\.(?:(check|can|run)Broadcast)\(/.test(handlerCode);
 			entry.isPrivate = /\bthis\.(?:privately(Check)?Can|commandDoesNotExist)\(/.test(handlerCode);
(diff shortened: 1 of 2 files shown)


server/chat.ts
```diff
@@ -1720,7 +1721,8 @@ export const Chat = new class {
 			if (typeof entry !== 'function') continue;
 
 			const handlerCode = entry.toString();
-			entry.requiresRoom = /\bthis\.requires?Room\(/.test(handlerCode);
+			entry.requiresRoom = /\bthis\.require?Room\(/.test(handlerCode);
+			entry.roomSpecific = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`).\)/.exec(handlerCode)?.[1];
```
Zarel, line 1725: Why not just turn `requiresRoom` into `RoomID | null`? I don't see any backwards compatibility issues.

mia-pi-git: Done.

server/chat.ts
```diff
@@ -1720,7 +1720,7 @@ export const Chat = new class {
 			if (typeof entry !== 'function') continue;
 
 			const handlerCode = entry.toString();
-			entry.requiresRoom = /\bthis\.requires?Room\(/.test(handlerCode);
+			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`).\)/.exec(handlerCode)?.[1] || false;
```
Zarel, line 1723: Will this be `true` for `requireRoom()` with no arguments?

Zarel: Looks like no. That'll need to be fixed.

mia-pi-git: Fixed.

server/chat.ts
```diff
@@ -1720,7 +1720,7 @@ export const Chat = new class {
 			if (typeof entry !== 'function') continue;
 
 			const handlerCode = entry.toString();
-			entry.requiresRoom = /\bthis\.requires?Room\(/.test(handlerCode);
+			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`).\)/.exec(handlerCode)?.[1] || /this\.requireRoom\(/.test(handlerCode);
```
Zarel, line 1723: `entry.requiresRoom` seems to still be a `boolean`? That should be updated to `boolean | RoomID`, right?

Zarel, line 1723: What's with the `.\)`? There should be a cast to `RoomID` after the string, right? There's no reason to search for the closing paren anyway, right?

server/chat.ts
```diff
@@ -1720,7 +1720,7 @@ export const Chat = new class {
 			if (typeof entry !== 'function') continue;
 
 			const handlerCode = entry.toString();
-			entry.requiresRoom = /\bthis\.requires?Room\(/.test(handlerCode);
+			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`) /.exec(handlerCode)?.[1] as RoomID || /this\.requireRoom\(/.test(handlerCode);
```
Zarel, line 1723: Why space?
```suggestion
			entry.requiresRoom = /requireRoom\((?:'|"|`)(.*?)(?:'|"|`)/.exec(handlerCode)?.[1] as RoomID || /this\.requireRoom\(/.test(handlerCode);
```

Zarel:
Build failure seems unrelated to this PR; merging so I don't forget.

PR #7680, opened 2020-11-11
Refactor text files

KrisXV:


Marty-D (approved):

Zarel:
I like this a lot; I'm just waiting for smogon/pokemon-showdown-client#1664

Zarel:
On second thought, there's no need to wait. Thanks!

PR #7683, opened 2020-11-11
Remove ^ from staff list in /scav top

PartMan7:
Requested by Litt

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -1927,7 +1927,7 @@ const ScavengerCommands: ChatCommands = {
 		const ladder = await Leaderboard.visualize('points') as AnyObject[];
 		this.sendReply(
 			`|uhtml${isChange ? 'change' : ''}|scavladder|<div class="ladder" style="overflow-y: scroll; max-height: 300px;"><table style="width: 100%"><tr><th>Rank</th><th>Name</th><th>Points</th></tr>${ladder.map(entry => {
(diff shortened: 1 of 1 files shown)


PartMan7:
W-what was that?

sparkychildcharlie:
Room.auth.get returns the higher rank between global and roomauth.  Use room.auth.getDirect instead to get only the room rank

PartMan7:
Ah, I see.

Any idea why it's crashing, though?

KrisXV:
Just resync your branch, `this.parse` is being silly

server/chat-plugins/scavengers.ts
```diff
@@ -1927,7 +1927,7 @@ const ScavengerCommands: ChatCommands = {
 		const ladder = await Leaderboard.visualize('points') as AnyObject[];
 		this.sendReply(
 			`|uhtml${isChange ? 'change' : ''}|scavladder|<div class="ladder" style="overflow-y: scroll; max-height: 300px;"><table style="width: 100%"><tr><th>Rank</th><th>Name</th><th>Points</th></tr>${ladder.map(entry => {
-				const isStaff = room!.auth.has(toID(entry.name));
+				const isStaff = ['+', '%', '@', '*', '#'].includes(room!.auth.getDirect(toID(entry.name)));
```
mia-pi-git, line 1930:
```suggestion
				const roomRank = room!.auth.getDirect(toID(entry.name));
				const isStaff = Users.Auth.atLeast(roomRank, '+');
```
Bit cleaner and accounts for other ranks that may be added.

Zarel:
Looks good!

PR #7684, opened 2020-11-11
Handle unpropagated `this.parse` calls

KrisXV:
(also fixes the build)

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -414,11 +414,11 @@ export const commands: ChatCommands = {
 			return this.sendReply(this.tr("What?! How are you not more excited to battle?! Try /battle! to show me you're ready."));
 		}
 		if (!target) target = "randombattle";
-		return void this.parse(`/search ${target}`);
+		return this.parse(`/search ${target}`);
 	},
 
 	avatar(target, room, user) {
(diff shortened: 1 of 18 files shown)


server/chat-commands/core.ts
```diff
@@ -414,11 +414,11 @@ export const commands: ChatCommands = {
 			return this.sendReply(this.tr("What?! How are you not more excited to battle?! Try /battle! to show me you're ready."));
 		}
 		if (!target) target = "randombattle";
-		return this.parse(`/search ${target}`);
+		return void this.parse(`/search ${target}`);
```
Zarel, line 417: Err, `return` already correctly propagates `this.parse`; the `void` is unnecessary.

KrisXV: idk how i messed that up so bad, oops

Zarel:
For some reason the build errors didn't show up for me when I did `npm test` or `npm run lint`. I guess a bug in eslintcache? I dealt with similar problems with other `no-floating-promises` errors randomly appearing and disappearing. I guess I'll remove the return value for now.

server/chat-commands/core.ts
Zarel: The `void` is actually bad because it stops the promise from being propagated like it should be.

PR #7685, opened 2020-11-11
Fix Gen 1 Psywave mechanics

PlagueVonKarma:
This implements the Psywave Desync Patch from the tiering decision found [here](https://www.smogon.com/forums/threads/rby-psywave-imperfect-dvs-and-thawing-desync-glitches.3668900/)
For details on how the desync works, go [here](https://www.youtube.com/watch?v=5KmTCdnWzVI)

[Here is a replay from the RBY 2k20 PS Server showing the implementation in action.](https://replay.pokemonshowdown.com/rby-gen1customgame-897)

I'm by no means a _good_ programmer, nor do I consider myself one, so I think my code may be a bit suboptimal.

### DIFF
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -606,9 +606,8 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			if (psywaveDamage <= 0) {
 				this.hint("Desync Clause activated.");
 				return false;
-			} else {
-				return psywaveDamage;
 			}
+			return psywaveDamage;
 		},
 	},
 	rage: {
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -602,12 +602,12 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		basePower: 1,
 		damageCallback(pokemon) {
-			let psywaveDamage = (this.random(0, (1.5 * pokemon.level)))
+			const psywaveDamage = (this.random(0, (1.5 * pokemon.level)));
 			if (psywaveDamage <= 0) {
 				this.hint("Desync Clause activated.");
 				return false;
-			} else { 
-				return psywaveDamage
+			} else {
+				return psywaveDamage;
 			}
 		},
 	},
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -602,7 +602,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		basePower: 1,
 		damageCallback(pokemon) {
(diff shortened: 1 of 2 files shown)


PlagueVonKarma:
I realized that this would have introduced a Stadium issue, so I added a commit that prevents Desync Clause from applying there. Besides, it has a minimum of 1 damage rather than 0 there, so this should tidy it up nicely.

data/mods/gen1/moves.ts
```diff
@@ -601,6 +601,15 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	psywave: {
 		inherit: true,
 		basePower: 1,
+		damageCallback(pokemon) {
+			const psywaveDamage = (this.random(0, (1.5 * pokemon.level)));
```
Zarel, line 605:
```suggestion
			const psywaveDamage = (this.random(0, this.trunc(1.5 * pokemon.level)));
```

Zarel: (`1.5 * pokemon.level` is rounded down before rolling the dice, right?)

data/mods/gen1/moves.ts
```diff
@@ -601,6 +601,15 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+			if (psywaveDamage <= 0) {
+				this.hint("Desync Clause activated.");
+				return false;
+			} else {
+				return psywaveDamage;
+			}
```
Zarel, line 611:
```suggestion
			if (psywaveDamage <= 0) {
				this.hint("Desync Clause activated.");
				return false;
			}
			return psywaveDamage;
```

data/mods/stadium/moves.ts
```diff
@@ -55,6 +55,13 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 		},
 	},
+	psywave: {
+		inherit: true,
+		basePower: 1,
+		damageCallback(pokemon) {
+			return (this.random(1, (1.5 * pokemon.level)));
```
Zarel, line 62:
```suggestion
			return this.random(1, this.trunc(1.5 * pokemon.level));
```

Zarel:
I'll leave this open for @DaWoblefet or @Marty-D to comment, but I personally think this is good.

Zarel (approved):

DaWoblefet:
No need for my approval with super oldgen mechanics, I'm no expert with those (like anything before Gen 5). You could write a test if you weren't sure, but if it looks good to you two that's fine with me.

Zarel:
Tests are annoying with RNG mechanics, and these usually don't need tests anyway (the code describes what it's supposed to do well enough).

Zarel:
We'll accept this as-is. Thanks!

PR #7686, opened 2020-11-11
Datasearch: Disallow 'all' search broadcasting in battles

mia-pi-git:
Per @asgdf.


Zarel:
We do support "read more" now. Maybe we should just use that?

PR #7687, opened 2020-11-12
Settings: Properly make battle rooms public

mia-pi-git:


Zarel:
It's still weird to only call this in battles.

This should maybe just all be merged into a single function `setPrivate` in all rooms, that sets `this.settings.isPrivate` and also updates the roomid if necessary.

PR #7690, opened 2020-11-12
Users: Only allow guest usernames when Config.nosecurity is in use

mia-pi-git:


server/users.ts
```diff
@@ -596,6 +596,10 @@ export class User extends Chat.MessageContext {
 				this.send(`|nametaken|${name}|You need an authentication token to log in as a trusted user.`);
 				return null;
 			}
```
Zarel, line 598: This part can be deleted because trusted users can't start with `guest`.

Zarel:
See discussion in #7648 starting at https://github.com/smogon/pokemon-showdown/pull/7648#issuecomment-723416352

@Slayer95, @xfix, do you think this would be an significant improvement?

@hsahovic this shouldn't be a problem for you, right?

hsahovic:
All of this looks good :) To be clear, bypassing token validation will only be possible with usernames prefixed with `guest`?

xfix:
This is fine with me as long `--no-security` won't turn on `noguestsecurity`. For development purposes disabling rate limits and IP checks could be useful, but when testing staff commands `noguestsecurity` is just annoying.

For example, for testing Scavengers module having IP checks disabled is useful because only one user on a given IP address can participate in a hunt, however you don't want to disable staff accounts, as only staff can start hunts.

Slayer95:
Yeah this concept solves my concerns about ``Config.noguestsecurity`` (though my -1 on the preset 3-in-1 flag stands FWIW).

However, this namespace is a bit subtle. ``/^guest.*$/`` is forbidden by the login server, while ``/^guest\d+$/`` is controlled by the sim server. That leaves ``/^guest[a-zA-Z]+$`` as fair play ground for ``Config.noguestsecurity``.

Zarel:
> All of this looks good :) To be clear, bypassing token validation will only be possible with usernames prefixed with `guest`?

Yes.

Zarel:
> However, this namespace is a bit subtle. `/^guest.*$/` is forbidden by the login server, while `/^guest\d+$/` is controlled by the sim server. That leaves `/^guest[a-z]+$` as fair play ground for `Config.noguestsecurity`.
> 
> ... Unless PS goes out of its way generating `guestNum`s randomly to avoid collisions?

Oh, shit, yeah, we'd have to only allow `/^guest[a-z].+$`. @mia-pi-git?

server/users.ts
```diff
@@ -592,8 +592,8 @@ export class User extends Chat.MessageContext {
 	}
 	async validateToken(token: string, name: string, userid: ID, connection: Connection) {
 		if (!token && Config.noguestsecurity) {
-			if (Users.isTrusted(userid)) {
-				this.send(`|nametaken|${name}|You need an authentication token to log in as a trusted user.`);
+			if (!/^guest[a-z]/.test(userid)) {
+				this.send(`|nametaken|${name}|You must use a username starting with 'Guest' if you're not using a token.`);
```
Zarel, line 596:
```suggestion
				this.send(`|nametaken|${name}|You must use a username starting with 'Guest' followed by a letter if you're not using a token.`);
```

hsahovic:
I can work with the restriction on the usernames compatible with `--no-security`, but it will come at a price in terms of QOL when working with `poke-env`; right now, usernames are derived from the agent classes used by the developers (these are names like `RandomPlayer 1`, `MaxDamagePlayer 4`, etc). This will need to be updated towards names that may be more confusing for users of the package.

I am not sure I understand why such a restriction is necessary for a flag that's explicitly incompatible with production, but as you probably gathered I'm far from perfectly understanding ps' organization and structure.

Zarel:
Yeah, I was concerned about that, too, which is why I didn't implement it this way at first. I think the problem is that I'm trying to reconcile your use case with other development use cases where the `guest` prefix adds useful security.

I think I'm not going to require the prefix. I don't think there's serious risk from it, and your QoL is important to me as well.

I think maybe a better idea is a rank symbol for "distrusted" to be used here.

PR #7691, opened 2020-11-13
/whois: Only show semilock reasons to global staff

mia-pi-git:


Zarel:
Why?

Zarel:
If you're roomstaff and a user can't talk in your room, you should know that.

mia-pi-git:
I changed it to hiding the reasons - is that better?

Zarel:
Sure, but did someone request this? What sorts of reasons did we want to hide from romostaff?

mia-pi-git:
> Sure, but did someone request this? What sorts of reasons did we want to hide from roomstaff?

See #private-dev.

server/chat-commands/info.ts
```diff
@@ -211,7 +211,7 @@ export const commands: ChatCommands = {
 			}
 
 			if (targetUser.semilocked) {
-				buf += `<br />Semilocked: ${targetUser.semilocked}`;
+				buf += `<br />Semilocked: ${user.can('lock') ? targetUser.semilocked : "(reason hidden)"}`;
```
Zarel, line 214:
```suggestion
				buf += `<br />Semilocked: ${user.trusted ? targetUser.semilocked : "(reason hidden)"}`;
```

I think I prefer a lower threshold like `trusted`. Is that fine with you?

Zarel:
After further staff discussion, it seems most think the `trusted` threshold is too low, so it looks like we're going with this.

Thanks!

PR #7692, opened 2020-11-14
Tests: add a function to save a replay of a battle

MacChaeger:
The intended use for this is to add a `common.saveReplay(battle);` line to tests you want to watch back in the client so you can see why a given assertion might be failing. I don't know enough about the testing suite to figure out how to make it automatically clear out the `replays` folder when `npm test` gets run, but I don't think that's entirely necessary anyway since one might want to compare two replays from different versions of the code.

Now that I think about it, it might be worth automatically storing replays for every applicable test and replacing them all when `npm test` is run to potentially detect regressions by checking for battle logs that have changed even though their assertions all pass. I don't know how to do that either, though.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -108,9 +108,10 @@ class TestTools {
 
 	/**
 	 * Saves the log of the given battle as a bare-bones replay file in the `test\replays` directory
-	 * To view the replay, simply drag and drop the file into a PS! client window
+	 * You can view the replay by opening the file in any browser or by dragging and dropping the
+	 * file into a PS! client window.
 	 *
-	 * @param {Sim.Battle} [battle]
+	 * @param {Sim.Battle} battle
 	 * @param {string} [fileName]
 	 */
 	saveReplay(battle, fileName) {
@@ -119,7 +120,17 @@ class TestTools {
 		const filePath = path.resolve(__dirname, `./replays/${fileName}-${Date.now()}.html`);
 		const out = fs.createWriteStream(filePath, {flags: 'a'});
 		out.on('open', () => {
-			out.write(`<script type="text/plain" class="log">${battleLog}</script>`);
+			out.write(`<!DOCTYPE html>
+			<meta charset="utf-8" />
+			<!-- version 1 -->
+			
+			<div class="wrapper replay-wrapper" style="max-width:1180px;margin:0 auto">
+			<div class="battle"></div><div class="battle-log"></div><div class="replay-controls"></div><div class="replay-controls-2"></div>
+			<script type="text/plain" class="battle-log-data">${battleLog}</script>
+			</div>
+			<script>
+			let daily = Math.floor(Date.now()/1000/60/60/24);document.write('<script src="https://play.pokemonshowdown.com/js/replay-embed.js?version'+daily+'"></'+'script>');
+			</script>`);
 			out.end();
 		});
 	}
--- a/test/common.js
+++ b/test/common.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const path = require('path');
+const fs = require('fs');
 const assert = require('./assert');
 const Sim = require('./../.sim-dist');
 const Dex = Sim.Dex;
@@ -102,6 +104,24 @@ class TestTools {
 		}
 
 		return new Sim.Battle(battleOptions);
+	}
+
+	/**
+	 * Saves the log of the given battle as a bare-bones replay file in the `test\replays` directory
+	 * To view the replay, simply drag and drop the file into a PS! client window
+	 *
+	 * @param {Sim.Battle} [battle]
+	 * @param {string} [fileName]
+	 */
+	saveReplay(battle, fileName) {
+		const battleLog = battle.getDebugLog();
+		if (!fileName) fileName = 'test-replay';
+		const filePath = path.resolve(__dirname, `./replays/${fileName}-${Date.now()}.html`);
+		const out = fs.createWriteStream(filePath, {flags: 'a'});
+		out.on('open', () => {
+			out.write(`<script type="text/plain" class="log">${battleLog}</script>`);
+			out.end();
+		});
 	}
 }
 
--- a/test/common.js
+++ b/test/common.js
@@ -120,17 +120,11 @@ class TestTools {
 		const filePath = path.resolve(__dirname, `./replays/${fileName}-${Date.now()}.html`);
 		const out = fs.createWriteStream(filePath, {flags: 'a'});
 		out.on('open', () => {
-			out.write(`<!DOCTYPE html>
-			<meta charset="utf-8" />
-			<!-- version 1 -->
-			
-			<div class="wrapper replay-wrapper" style="max-width:1180px;margin:0 auto">
-			<div class="battle"></div><div class="battle-log"></div><div class="replay-controls"></div><div class="replay-controls-2"></div>
-			<script type="text/plain" class="battle-log-data">${battleLog}</script>
-			</div>
-			<script>
-			let daily = Math.floor(Date.now()/1000/60/60/24);document.write('<script src="https://play.pokemonshowdown.com/js/replay-embed.js?version'+daily+'"></'+'script>');
-			</script>`);
+			out.write(
+				`<!DOCTYPE html>\n` +
+				`<script type="text/plain" class="battle-log-data">${battleLog}</script>\n` +
+				`<script src="https://play.pokemonshowdown.com/js/replay-embed.js"></script>\n`
+			);
 			out.end();
 		});
 	}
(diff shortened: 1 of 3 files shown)


Slayer95:
> Now that I think about it, it might be worth automatically storing replays for every applicable test and replacing them all when npm test is run to potentially detect regressions by checking for battle logs that have changed even though their assertions all pass. I don't know how to do that either, though.

That's a snapshot test, and it's still [controversial](https://blog.usejournal.com/whats-wrong-with-snapshot-tests-37fbe20dfe8e) whether they are all that good. The butterfly effect of a commit that changes the order of RNG rolls would be huge for all the snapshots, for instance.

Here's a how-to compatible with Mocha:
https://medium.com/blogfoster-engineering/how-to-use-the-power-of-jests-snapshot-testing-without-using-jest-eff3239154e5

MacChaeger:
I wouldn’t actually expect something that changes RNG _order_ to be that significant if it doesn’t involve damage rolls directly, as unless it actually changes the number of RNG calls it shouldn’t actually have a butterfly effect. Even if it does, we should be keeping an eye out for big changes to test outcomes because of that. After all, it’s not like we’ve never had a regression that went unnoticed because a prior RNG change secretly broke a test. The idea would be not to rely on the testing infrastructure itself to actually compare changes to battle logs, but rather to just store them. We would instead rely on git itself to track those files and detect changes to them, which would essentially eliminate most of the problems raised in that article.

They wouldn’t be hidden. We would actually be able to see what was changed exactly, line by line, and we we wouldn’t be committing to having bugged snapshots cause failing tests because changed snapshots wouldn’t actually make tests necessarily fail. With Git tracking them, even if someone did commit a change that broke a test without realizing it, we could look back through our version history to see when that test’s log changed, and how. This would only be used for the battle engine, so dependencies changing small aesthetic things would hopefully never be a problem. Our battle tests are reasonably well documented in terms of expected behavior, so it should hopefully be clear to anyone watching a test’s replay whether or not it’s happening as intended. We don’t need to worry about this adding a false sense of security as long as people write tests to go with their bug fixes when they make them, and frankly our tests are so extensive yet specific that they already give a false sense of security, IMO.

That’s not to say there wouldn’t be any problems at all. Opening and writing to and closing so many files would significantly slow down the full test, and having the remote repo modify its own files as part of an automatic testing procedure seems bad, especially when those files are theoretically never getting changed by said procedure. I imagine there’s some way to get the remote tests not to overwrite the replays, but I doubt I could figure that out on my own.

MacChaeger:
We’ll certainly find out just how false their sense of security is when server-side multi battle support gets added and breaks a bunch of things despite passing all the tests. 😂

Zarel:
I'm opposed to snapshot testing for the reasons described in that article.

Anything that changes mechanics, like fixing certain bugs, and implementing multi-battles, are _likely_ to change most snapshots. Are you really going to go through 150 snapshots to individually make sure every single change isn't a bug?

test/common.js
```diff
@@ -103,6 +105,24 @@ class TestTools {
 
 		return new Sim.Battle(battleOptions);
 	}
+
+	/**
+	 * Saves the log of the given battle as a bare-bones replay file in the `test\replays` directory
```
Zarel, line 110: if you made it slightly less barebones, you could just double-click it.

Zarel: Why not just make it work like real replay files?

test/common.js
```diff
@@ -103,6 +105,24 @@ class TestTools {
...
+	 * @param {Sim.Battle} [battle]
+	 * @param {string} [fileName]
+	 */
+	saveReplay(battle, fileName) {
+		const battleLog = battle.getDebugLog();
+		if (!fileName) fileName = 'test-replay';
+		const filePath = path.resolve(__dirname, `./replays/${fileName}-${Date.now()}.html`);
+		const out = fs.createWriteStream(filePath, {flags: 'a'});
+		out.on('open', () => {
+			out.write(`<script type="text/plain" class="log">${battleLog}</script>`);
```
Zarel, line 122:
```html
<!DOCTYPE html>
<meta charset="utf-8" />
<!-- version 1 -->

<div class="wrapper replay-wrapper" style="max-width:1180px;margin:0 auto">
<div class="battle"></div><div class="battle-log"></div><div class="replay-controls"></div><div class="replay-controls-2"></div>
<script type="text/plain" class="battle-log-data">
${battleLog}
</script>
</div>
<script>
let daily = Math.floor(Date.now()/1000/60/60/24);document.write('<script src="https://play.pokemonshowdown.com/js/replay-embed.js?version'+daily+'"></'+'script>');
</script>
```

is all you need.

Zarel: I should probably should make it not require the divs.

MacChaeger:
> 
> 
> I'm opposed to snapshot testing for the reasons described in that article.
> 
> Anything that changes mechanics, like fixing certain bugs, and implementing multi-battles, are _likely_ to change most snapshots. Are you really going to go through 150 snapshots to individually make sure every single change isn't a bug?

My second paragraph explains that none of the reasons from that article would apply to what I'm suggesting, which isn't really snapshot testing, just snapshot recording. I also really don't expect most code changes to actually change more than a few clearly relevant battle logs. If we implement multi battles correctly, then none of the existing tests should change even a little (except tests where Unnerve is involved, since the multi battle PR includes removing the extraneous side data from the activation message). That's the general idea: any change made to mechanics shouldn't change the behavior of any tests that don't directly involve those mechanics, and even refactors that touch basically everything shouldn't change existing behaviors. If they do, it's important that we understand not only that they do, but _how_ they do. Even if we don't comb through every one of 150 (which is probably an unusual number; if it's higher than like 50, it's probably going to be closer to 500) changed replays (not that I think that would happen without a huge change that would affect practically every battle on the server), if we find a regression later, having these logs will make it easier to track down when an why it happened.

Slayer95:
The implementation of a one-off method you may easily call to store a replay would be convenient, because it's a one-off matter. However, if you intend to scale them up and use replays / snapshots as part of the test infrastructure, then the choice between automated and manual just puts open-source contributors between the sword and the wall, facing the nitty-gritty details of PS protocol.

> potentially detect regressions by checking for battle logs that have changed even though their assertions all pass

In fact, the true and proven way to figure out regressions is to write a new test showcasing the issue and checking it with ``git bisect`` to find the first BAD commit. IMHO your proposal to save replays to be analysed by a human in an hypothetical future is redundant with what ``git`` already offers.

MacChaeger:
> 
> 
> The implementation of a one-off method you may easily call to store a replay would be convenient, because it's a one-off matter. However, if you intend to scale them up and use replays / snapshots as part of the test infrastructure, then the choice between automated and manual just puts open-source contributors between the sword and the wall, facing the nitty-gritty details of PS protocol.
> In fact, the true and proven way to figure out regressions is to write a new test showcasing the issue and checking it with `git bisect` to find the first BAD commit. IMHO your proposal to save replays to be analysed by a human in an hypothetical future is redundant with what `git` already offers.

That's a thing? Never mind, then, I guess.

Slayer95:
Bumping to add that this function would help test implementors ensure that their suite _really_ does what they have in mind, and it's not testing some other random stuff resulting in false positives/negatives.

test/common.js
Zarel: (bump)

Zarel: I guess I can refactor `replay-embed` first before asking for this.

test/common.js
```diff
@@ -103,6 +105,35 @@ class TestTools {
...
+			<meta charset="utf-8" />
+			<!-- version 1 -->
+			
+			<div class="wrapper replay-wrapper" style="max-width:1180px;margin:0 auto">
+			<div class="battle"></div><div class="battle-log"></div><div class="replay-controls"></div><div class="replay-controls-2"></div>
+			<script type="text/plain" class="battle-log-data">${battleLog}</script>
+			</div>
+			<script>
+			let daily = Math.floor(Date.now()/1000/60/60/24);document.write('<script src="https://play.pokemonshowdown.com/js/replay-embed.js?version'+daily+'"></'+'script>');
+			</script>`);
```
Zarel, line 133:
```suggestion
			out.write(
				`<!DOCTYPE html>\n` +
				`<script type="text/plain" class="battle-log-data">${battleLog}</script>\n` +
				`<script src="https://play.pokemonshowdown.com/js/replay-embed.js"></script>`
			);
```

I pushed some code to simplify this: https://github.com/smogon/pokemon-showdown-client/commit/d44b94e8414d02141ad072b465fe7dcce3254c5d

Zarel:
I won’t personally use this, but if you’d like this for convenience, I don’t have a problem with it.

PR #7694, opened 2020-11-15
Rename "Desync Clause" to "Desync Clause Mod"

PlagueVonKarma:
This was a minor mistake when I was writing this thing up. Since Desync Clause isn't something you replicate in RBY, it is more accurate to describe it as a Mod, in line with the ADV Switch Priority Clause Mod.

### DIFF
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -604,7 +604,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		damageCallback(pokemon) {
 			const psywaveDamage = (this.random(0, this.trunc(1.5 * pokemon.level)));
 			if (psywaveDamage <= 0) {
-				this.hint("Desync Clause Mod activated.");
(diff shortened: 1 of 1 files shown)


data/mods/gen1/moves.ts
```diff
@@ -604,7 +604,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		damageCallback(pokemon) {
 			const psywaveDamage = (this.random(0, this.trunc(1.5 * pokemon.level)));
 			if (psywaveDamage <= 0) {
-				this.hint("Desync Clause activated.");
+				this.hint("Desync Clause Mod activated.");
```
Zarel, line 607:
```suggestion
				this.hint("Desync Clause Mod activated!");
```

If you want to be consistent with my commit (and with Pokémon's general rule of putting exclamation marks on most everything), this is an exclamation mark. ;)

Zarel:
tyty!

PR #7695, opened 2020-11-15
Rooms: Merge privacy setting code into setPrivate

mia-pi-git:
As per https://github.com/smogon/pokemon-showdown/pull/7687#issuecomment-725795911.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -668,8 +668,8 @@ export const commands: ChatCommands = {
 		const targetRoom = Rooms.search(target);
 		if (!targetRoom) throw new Error(`Error in room creation.`);
 		if (cmd === 'makeprivatechatroom') {
-			targetRoom.setPrivate(true);
 			if (!targetRoom.persist) throw new Error(`Private chat room created without settings.`);
+			targetRoom.setPrivate(true);
 			const upperStaffRoom = Rooms.get('upperstaff');
 			if (upperStaffRoom) {
 				upperStaffRoom.add(`|raw|<div class="broadcast-green">Private chat room created: <b>${Utils.escapeHTML(target)}</b></div>`).update();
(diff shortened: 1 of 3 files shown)


server/rooms.ts
```diff
@@ -703,6 +703,27 @@ export abstract class BasicRoom {
...
+			for (const user of Object.values(this.users)) {
+				if (!user.named) {
+					user.leaveRoom(this.roomid);
+					user.popup(`The battle <<${this.roomid}>> has been made private; you must log in to watch private battles.`);
+				}
+			}
+			if (this.roomid.endsWith('pw')) return true;
+			this.rename(this.title, `${this.roomid}-${password}pw` as RoomID, true);
+		} else if (!privacy && 'makePublic' in this) {
+			(this as GameRoom).makePublic();
```
Zarel, line 724: Why didn't you merge `makePublic` into this?

server/chat-commands/room-settings.ts
```diff
@@ -668,10 +668,8 @@ export const commands: ChatCommands = {
 		const targetRoom = Rooms.search(target);
 		if (!targetRoom) throw new Error(`Error in room creation.`);
 		if (cmd === 'makeprivatechatroom') {
-			targetRoom.settings.isPrivate = true;
+			targetRoom.setPrivate(true);
 			if (!targetRoom.persist) throw new Error(`Private chat room created without settings.`);
```
Zarel, line 672:
```suggestion
			if (!targetRoom.persist) throw new Error(`Private chat room created without settings.`);
			targetRoom.setPrivate(true);
```

server/chat-commands/room-settings.ts
```diff
@@ -668,10 +668,8 @@ export const commands: ChatCommands = {
 		const targetRoom = Rooms.search(target);
 		if (!targetRoom) throw new Error(`Error in room creation.`);
 		if (cmd === 'makeprivatechatroom') {
-			targetRoom.settings.isPrivate = true;
 			if (!targetRoom.persist) throw new Error(`Private chat room created without settings.`);
```
Zarel, line 671: Does this even need to be thrown? This just using `/makeprivatechatroom` on a groupchat/battle, right?

mia-pi-git: Looks like it. This shouldn't be necessary from what it seems, i agree.

Zarel: After some testing, this doesn't look like a problem, but it's definitely pretty confusing code.

PR #7697, opened 2020-11-15
Throw unrelated errors caught in runPromote

HoeenCoder:
We shouldn't be suppressing errors caught here that are really errors. Caused an obscure bug this morning that I had to hunt down.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -189,7 +189,9 @@ export const commands: ChatCommands = {
 			try {
 				shouldPopup = runPromote(user, room, userid, nextSymbol, toPromote, force);
 			} catch (err) {
-				if (err.name?.endsWith('ErrorMessage')) this.errorReply(err.message);
+				// Throw actual crashes, but don't throw chat ErrorMessages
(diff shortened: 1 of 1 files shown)


server/chat-commands/moderation.ts
```diff
@@ -189,7 +189,9 @@ export const commands: ChatCommands = {
 			try {
 				shouldPopup = runPromote(user, room, userid, nextSymbol, toPromote, force);
 			} catch (err) {
-				if (err.name?.endsWith('ErrorMessage')) this.errorReply(err.message);
+				// Throw actual crashes, but don't throw chat ErrorMessages
+				if (!err.name?.endsWith('ErrorMessage')) throw err;
+				this.errorReply(err.message);
 				continue;
```
Zarel, line 195:
```suggestion
				if (err.name?.endsWith('ErrorMessage')) {
					this.errorReply(err.message);
					continue;
				}
				throw err;
```

This is how I'd implement this.

Zarel:
Looks like everything else that handles `Chat.ErrorMessage` correctly uses the crashlogger. Thanks!

PR #7698, opened 2020-11-15
Update BST after changing stats to allow Scalemons Mod mashup

urkerab:
Reported [here](https://www.smogon.com/forums/posts/8646471).

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1056,10 +1056,12 @@ export const Formats: FormatList = [
 			mixedSpecies.evos = [];
 			mixedSpecies.eggGroups = crossSpecies.eggGroups;
 			mixedSpecies.abilities = crossSpecies.abilities;
+			mixedSpecies.bst = 0;
 			let i: StatName;
 			for (i in species.baseStats) {
 				const statChange = crossSpecies.baseStats[i] - crossPrevoSpecies.baseStats[i];
 				mixedSpecies.baseStats[i] = this.clampIntRange(species.baseStats[i] + statChange, 1, 255);
+				mixedSpecies.bst += mixedSpecies.baseStats[i];
 			}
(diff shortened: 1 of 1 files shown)


urkerab:
(Ugh, `species.bst` is readonly but `species.baseStats.X` isn't. Fun...)

MacChaeger:
Somewhat related: having random-teams run the ModifySpecies event as it builds teams so scalemons and other similar mods that modify base stats can be taken into account for things like level in hackmons cup

config/formats.ts
```diff
@@ -1370,12 +1374,14 @@ export const Formats: FormatList = [
 			};
 			const tier = this.toID(species.tier) || 'ou';
 			if (!(tier in boosts)) return;
-			const pokemon: Species = this.dex.deepClone(species);
+			const pokemon = this.dex.deepClone(species);
+			pokemon.bst = 0;
 			const boost = boosts[tier];
 			let statName: StatName;
-			for (statName in pokemon.baseStats) {
+			for (statName in pokemon.baseStats as StatsTable) {
```
KrisXV, line 1381: why is this necessary?

urkerab: Because `pokemon` is no longer a `Species` (so that I can change its `bst`), `pokemon.baseStats` is no longer a `StatsTable`, so `StatName` is no longer the key of it, resulting in the error "The left-hand side of a 'for...in' statement must be of type 'string' or 'any'".

Zarel:
How important is caching the BST? Should we just have a `Dex.getBST(species)` function?

KrisXV:
> How important is caching the BST? Should we just have a `Dex.getBST(species)` function?

it seems to be done pretty often throughout the PR, so this might be helpful, though I'm unsure how it'd be approached

PR #7700, opened 2020-11-15
Runners: support 4P (and max games for exhaustive)

MacChaeger:
Ran 10000 games of [Gen 8] Multi Hackmons Cup at 6 Pokemon per player (24 pokemon per battle) on the latest version of the server-side multi battle branch, and got 0 crashes.
`[gen8multihackmonscup] P:0 (11/761) I:203 (4/357) A:285 (3/267) M:106 (1/664) = 10000`
Then again, apparently it wasn't actually an exhaustive test because 11 species of Pokemon never got used at all, even though it should have used somewhere between 120,000 and 240,000 individual Pokemon. I can guess those 11 aren't terribly important, though, if they can't appear even in Hackmons Cup.

### DIFF
--- a/sim/tools/exhaustive-runner.ts
+++ b/sim/tools/exhaustive-runner.ts
@@ -81,12 +81,15 @@ export class ExhaustiveRunner {
 		do {
 			this.games++;
 			try {
+				const is4P = dex.getFormat(this.format).gameType === 'multi';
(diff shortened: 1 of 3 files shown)


MacChaeger:
Bump

Zarel:
Thanks!

PR #7701, opened 2020-11-16
SS Ubers: Ban Dynamax

PlagueVonKarma:
In accordance with this tiering decision: https://www.smogon.com/forums/threads/ss-ubers-stage-4-dynamax-voting.3673382/post-8658893

Not sure if I'm allowed to do this, string me up if it's not


Zarel:
Err... did you mess this up or something?

DaWoblefet:
Nah, Marty just beat Plague to it

PR #7702, opened 2020-11-16
Gen 1: Announce Desync Clause Mod, fix Stadium Sleep Clause

PlagueVonKarma:
This is to be added to the "Standard" ruleset, so that "Desync Clause Mod" will show up when you start an RBY game on PS. I'll add this to the gen1/rulesets.ts "Standard" and any Gen 1 formats that don't use it, which should cover everything. 

This isn't meant to actually _do_ anything on its own, only note that the "mod" exists and will trigger in certain scenarios. 

I've written the rule description in a way that it will need updating whenever another desync is patched. There's still quite a few of these, though, and I believe the RBY Council still needs to go over a bunch. This includes Bide and Mirror Move + Partial Trapping. There's also some game crashes involving specific defensive stats but those are different. 

While doing this, I noticed that Stadium OU was removing Smogon Sleep Clause and replacing it with Stadium Sleep Clause, which seemed pretty inefficient, so I thought it would be better to have this in the "Standard" rules instead.

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -830,7 +830,7 @@ export const Formats: {[k: string]: FormatData} = {
 		effectType: 'Rule',
 		name: 'Desync Clause Mod',
 		desc: 'In the event a move would cause a desync, the move will fail instead. This rule covers Psywave and Counter.',
-		// The effect itself is implemented in gen1/moves.ts. 
+		// The effect itself is implemented in gen1/moves.ts.
 		onBegin() {
 			this.add('rule', 'In the event a move would cause a desync, the move will fail instead.');
 		},
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -829,7 +829,7 @@ export const Formats: {[k: string]: FormatData} = {
 	desyncclausemod: {
 		effectType: 'Rule',
 		name: 'Desync Clause Mod',
-		desc: 'In the event a move would cause a desync, the move will fail instead. This rule covers Psywave and Counter.',
+		desc: 'If a desync would happen, the move fails instead. This rule currently covers Psywave and Counter.',
 		onBegin() {
 			this.add('rule', 'Desync Clause Mod: Desyncs changed to move failure.');
 		},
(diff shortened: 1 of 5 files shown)


data/rulesets.ts
```diff
@@ -826,6 +826,15 @@ export const Formats: {[k: string]: FormatData} = {
 			this.add('rule', 'Switch Priority Clause Mod: Faster Pokémon switch first');
 		},
 	},
+	desyncclausemod: {
+		effectType: 'Rule',
+		name: 'Desync Clause Mod',
+		desc: 'In the event a move would cause a desync, the move will fail instead. This rule covers Psywave and Counter.',
+		// The effect itself is implemented in gen1/moves.ts. 
+		onBegin() {
+			this.add('rule', 'In the event a move would cause a desync, the move will fail instead.');
```
Zarel, line 835: Design principle 1 is in full effect, especially in rule descriptions where the length limit is _very_ low.

```suggestion
			this.add('rule', 'Desync Clause Mod: Desyncs become move failure.');
```

data/rulesets.ts
```diff
@@ -826,6 +826,15 @@ export const Formats: {[k: string]: FormatData} = {
...
 		},
 	},
+	desyncclausemod: {
+		effectType: 'Rule',
+		name: 'Desync Clause Mod',
+		desc: 'In the event a move would cause a desync, the move will fail instead. This rule covers Psywave and Counter.',
+		// The effect itself is implemented in gen1/moves.ts.
+		onBegin() {
+			this.add('rule', 'Desync Clause Mod: Desyncs changed to move failure.');
+		},
```
Zarel, line 836:
```suggestion
		onBegin() {
			this.add('rule', 'Desync Clause Mod: Desyncs changed to move failure.');
		},
		// Hardcoded in gen1/moves.ts
		// Can't be disabled (no precedent for how else to handle desyncs)
```
(Design principle 1 again; also good to match how other rules are commented.)

data/rulesets.ts
```diff
@@ -826,6 +826,16 @@ export const Formats: {[k: string]: FormatData} = {
 			this.add('rule', 'Switch Priority Clause Mod: Faster Pokémon switch first');
 		},
 	},
+	desyncclausemod: {
+		effectType: 'Rule',
+		name: 'Desync Clause Mod',
+		desc: 'In the event a move would cause a desync, the move will fail instead. This rule covers Psywave and Counter.',
```
Zarel, line 832:
```suggestion
		desc: 'If a desync would happen, the move fails instead. This rule currently covers Psywave and Counter.',
```

principle 1

Zarel:
This looks good to me now.

Zarel:
Looks good now; thanks!

PR #7704, opened 2020-11-17
/regtime: Fix crash

mia-pi-git:


Zarel:
In what situations does this crash?

PR #7705, opened 2020-11-17
Fix sub player and IDEA mechanics (Mafia)

zgeorgecsat:
- Forceadd allows players who have subbed out or stopped cohosting to re-enter the game. 
- Plurality shifts to the subbed in player if it was on the subbed out player, and the mafia page is updated after the subbed out player is removed from the player list.
- /mafia dt gi/ci/etc gives the rolelists for each.
- Role lists have spaces between each role. 
- The clear button during ideapick phase is fixed.

### DIFF
--- a/server/chat-plugins/mafia.ts
+++ b/server/chat-plugins/mafia.ts
@@ -1178,6 +1178,8 @@ class MafiaTracker extends Rooms.RoomGame {
 				if (this.dead[p].restless && this.dead[p].lynching === oldPlayer.id) this.dead[p].lynching = newPlayer.id;
 			}
 		}
+		if(this.hasPlurality === oldPlayer.id) this.hasPlurality = newPlayer.id;
+		
 		if (newUser?.connected) {
 			for (const conn of newUser.connections) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/mafia.ts
```diff
@@ -1514,7 +1516,11 @@ class MafiaTracker extends Rooms.RoomGame {
 		const targetString = self ? `You are` : `${user.id} is`;
 		if (!this.room.users[user.id]) return `${targetString} not in the room.`;
 		for (const id of [user.id, ...user.previousIDs]) {
-			if (this.playerTable[id] || this.played.includes(id)) return `${targetString} already in the game.`;
+			if (this.playerTable[id] || this.dead[id]) {
+				return `${targetString} already in the game.`;
```
AnnikaCodes, line 1520: This function should probably be refactored to use `Chat.ErrorMessage` rather than returning a string on failure and seemingly false on success (?)

iscke: (there's a few places in which mafia doesn't use new features like that, i/someone'll deal with that later)

iscke:
looks great, thanks!

PR #7706, opened 2020-11-18
Correct multi-hit distribution and confusion odds

DaWoblefet:
Based on new research, multi-hit distribution is now known to be 35-35-15-15 rather than 1/3-1/3-1/6-1/6, and confusion is now known to be 33/100 rather than 1/3.

See videos:
https://www.youtube.com/watch?v=Il1sMnFUh9s
https://www.youtube.com/watch?v=TMtpLva4_qM

I think the way I corrected multi-hit move distribution is accurate, but it looks kind of dumb; not sure if there's a preferable way to do it.

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -168,7 +168,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 				return;
 			}
 			this.add('-activate', pokemon, 'confusion');
-			if (!this.randomChance(1, 3)) {
+			if (!this.randomChance(33, 100)) {
 				return;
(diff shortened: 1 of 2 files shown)


Zarel:
This is definitely the _easiest_ way to change it. I don't think it's worth redesigning how the roll works for readability; the comment's enough.

Thanks!

MacChaeger:
So... what about previous generations and move descriptions?

DaWoblefet:
dang it, forgot about those, I'll pick those up and push directly

PR #7708, opened 2020-11-18
Helptickets: Mark tickets with offline users separately

mia-pi-git:
As per https://www.smogon.com/forums/threads/3642984/page-6#post-8648367


Zarel:
Should be good, thanks!

PR #7709, opened 2020-11-19
Filters: Properly stop forcerenamed names from being used

mia-pi-git:
https://www.smogon.com/forums/threads/staff-suggestions-bugs.3514540/page-17#post-8642512
(before this change, it was just stopping users that had been forcerenamed from using pokemon nicknames)


AnnikaCodes (approved):

Zarel:
ty

PR #7710, opened 2020-11-19
Chat monitor: Support unit tests

AnnikaCodes:

PR #7711, opened 2020-11-20
Propagate promises for Punishments#autolock

AnnikaCodes:
This allows to test chat monitors and other things involving autolocks without needing hacky workarounds to wait until a the autolock code is executed.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -644,7 +644,7 @@ export const commands: ChatCommands = {
 		return user.update();
 	},
 
-	status(target, room, user, connection, cmd) {
+	async status(target, room, user, connection, cmd) {
 		if (user.locked || user.semilocked) {
 			return this.errorReply(this.tr("Your status cannot be updated while you are locked or semilocked."));
 		}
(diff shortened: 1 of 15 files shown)


server/chat-commands/core.ts
```diff
@@ -654,7 +654,7 @@ export const commands: ChatCommands = {
 		if (target.length > maxLength) {
 			return this.errorReply(this.tr `Your status is too long; it must be under ${maxLength} characters.`);
 		}
-		target = this.statusfilter(target);
+		target = await this.statusfilter(target);
```
Zarel, line 657: I really don't think we should be propagating this.

We're adding a bunch of lag to all our commands for... what, exactly?

Zarel: Improved unit test capabilities? I'm not sure this is worth it. Can we frontload enough of autolock such that the important stuff just happens synchronously?

AnnikaCodes: We can't exactly frontload all the important stuff, but we can add an extra line to set `user.locked` before the full punishment codes runs, which i did in #7710 since you seem to prefer that approach. And yes, less fragile unit tests (not having to delay assertions by 25ms).

PR #7713, opened 2020-11-20
Dexsearch: support resists move and weak move

MacChaeger:
I borrowed some code from the `/effectiveness` command, so this one should now support searches such as `/ds resists freeze dry`, `/ds resists flying press`, and `/ds !weak thousand arrows`.


Zarel:
Hm, this seems unlikely to break anything. Thanks!

PR #7715, opened 2020-11-21
Support automatic expiry for monthlocks

AnnikaCodes:
As discussed in PMs with Zarel, this pokes every punishment timer daily, allowing for punishments longer than Node's maximum timeout length to automatically expire.

### DIFF
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -39,6 +39,9 @@ const AUTOLOCK_POINT_THRESHOLD = 8;
 const AUTOWEEKLOCK_THRESHOLD = 5; // number of global punishments to upgrade autolocks to weeklocks
 const AUTOWEEKLOCK_DAYS_TO_SEARCH = 60;
 
+/** The longest amount of time any individual timeout will be set for. */
+const MAX_PUNISHMENT_TIMER_LENGTH = 24 * 60 * 60 * 1000; // 24 hours
+
 /**
  * The number of users from a groupchat whose creator was banned from using groupchats
  * who may join a new groupchat before the GroupchatMonitor activates.
@@ -765,7 +768,7 @@ export const Punishments = new class {
 		const affected = await Punishments.punish(user, punishment, ignoreAlts, bypassPunishmentfilter);
 
 		for (const curUser of affected) {
-			Punishments.setPunishmentTimer(curUser, expireTime, punishment[1]);
+			Punishments.pokePunishmentTimer(curUser, punishment);
 			curUser.locked = punishment[1];
 			curUser.updateIdentity();
 		}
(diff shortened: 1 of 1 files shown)


Zarel:
Hm, I imagined it as a global punishment timer that ran every 24 hours, and then only individual users whose punishments were within 24 hours would get their individual punishment timer set.

I suppose your implementation is easier and maybe more readable, and probably has similar performance.

Any hotpatching concerns?

AnnikaCodes:
It hotpatches fine locally without any crashes (I think that setTimeout holds onto a reference to its callback)

server/punishments.ts
```diff
@@ -861,17 +864,24 @@ export const Punishments = new class {
...
-	 * Validates the length and sets the punishment timer for a user.
+	 * Sets the punishment timer for a user,
+	 * to either MAX_PUNISHMENT_TIMER_LENGTH or the amount of time left on the punishment.
 	 */
-	setPunishmentTimer(user: User, expireTime: number, punishedID: ID | PunishType) {
-		if (expireTime === 0 || expireTime === Infinity || expireTime > Chat.MAX_TIMEOUT_DURATION) return;
-		user.punishmentTimer = setTimeout(Punishments.automaticUnlock, expireTime - Date.now(), user, punishedID);
+	pokePunishmentTimer(user: User, punishment: Punishment) {
+		const length = Math.min(punishment[2] - Date.now(), MAX_PUNISHMENT_TIMER_LENGTH);
+		user.punishmentTimer = setTimeout(Punishments.runExpiry, length, user, punishment);
```
Zarel, line 884: I never like `setTimeout` with arguments, I think it has worse readability.

```suggestion
		user.punishmentTimer = setTimeout(() => {
			Punishments.runExpiry(user, punishment);
		}, length);
```

On the other hand, your approach _is_ what let the previous one hotpatch correctly. Still, though. Probably better just to `clearTimeout`.

Also, why isn't `pokePunishmentTimer` and `runExpiry` the same function? I'd call the merged function `checkPunishmentTime`.

server/punishments.ts
```diff
@@ -861,17 +864,22 @@ export const Punishments = new class {
...
-	 * Validates the length and sets the punishment timer for a user.
+	 * Sets the punishment timer for a user,
+	 * to either MAX_PUNISHMENT_TIMER_LENGTH or the amount of time left on the punishment.
+	 * It also expires a punishment if the time is up.
 	 */
-	setPunishmentTimer(user: User, expireTime: number, punishedID: ID | PunishType) {
-		if (expireTime === 0 || expireTime === Infinity || expireTime > Chat.MAX_TIMEOUT_DURATION) return;
-		user.punishmentTimer = setTimeout(Punishments.automaticUnlock, expireTime - Date.now(), user, punishedID);
+	checkPunishmentTime(user: User, punishment: Punishment) {
+		const [, id, expireTime] = punishment;
```
Zarel, line 873:
```suggestion
		if (user.punishmentTimer) {
			clearTimeout(user.punishmentTimer);
			user.punishmentTimer = null;
		}

		const [, id, expireTime] = punishment;
```

server/punishments.ts
```diff
@@ -861,17 +864,22 @@ export const Punishments = new class {
...
+
+		if (expireTime < Date.now()) {
```
Zarel, line 875:
```suggestion
		if (expireTime - 1 <= Date.now()) {
```

A bit more leeway so you don't set timers for 1ms if there's some minor timer jitter.

server/punishments.ts
```diff
@@ -861,17 +864,22 @@ export const Punishments = new class {
...
+			if (user.locked === id) Punishments.unlock(user.id);
+		} else {
+			const length = Math.min(expireTime - Date.now(), MAX_PUNISHMENT_TIMER_LENGTH);
+			user.punishmentTimer = setTimeout(() => {
+				Punishments.checkPunishmentTime(user, punishment);
+			}, length);
+		}
```
Zarel, line 882:
```suggestion
		const timeLeft = expireTime - Date.now();
		if (timeLeft <= 1) {
			if (user.locked === id) Punishments.unlock(user.id);
			return;
		}
		const waitTime = Math.min(timeLeft, MAX_PUNISHMENT_TIMER_LENGTH);
		user.punishmentTimer = setTimeout(() => {
			Punishments.checkPunishmentTime(user, punishment);
		}, waitTime);
```

Zarel: Mostly, this is about not calling `Date.now()` multiple times. In this case it's not a big deal, but it's just a good habit to get into. Naming variables `timeLeft` and `waitTime` I think improves readability by a lot.

Zarel:
Looks good!

PR #7717, opened 2020-11-21
Fix type immunity check in /ds weak and resists and disallow status moves

MacChaeger:
It broke something! Oops!


MacChaeger:
It broke a bit more than one something, hang on while I add those fixes.

MacChaeger:
Yeah I have no idea what those failing tests are for.

MacChaeger:
Wait, they were only failing locally? Whuh?

Zarel:
Thanks!

PR #7719, opened 2020-11-23
Fix Emergency Exit/Wimp Out false positives

MacChaeger:


DaWoblefet:
The new tests all simulate correct behavior as far as I can tell. For Assurance, I might recommend using Life Orb instead of Wild Charge recoil, but it works fine.

test/sim/moves/assurance.js
```diff
@@ -0,0 +1,24 @@
...
+
+	it('should double its base power if the target already took damage this turn', function () {
+		battle = common.createBattle([
+			[{species: 'Morpeko', ability: 'hungerswitch', moves: ['assurance']}],
+			[{species: 'Regieleki', ability: 'transistor', moves: ['wildcharge']}],
+		]);
+		battle.makeChoices();
+		const regi = battle.p2.active[0];
+		const recoilRange = [113, 133].map(d => Math.floor(d / 4));
+		const assuRange = [214, 253];
```
MacChaeger, line 21: I had to really resist calling this `assRange` instead.

sim/battle.ts
```diff
@@ -2533,6 +2533,7 @@ export class Battle {
 
 	runAction(action: Action) {
 		const pokemonOriginalHP = action.pokemon?.hp;
+		let residualPokemon: (readonly [Pokemon, number])[] | null = null;
```
urkerab, line 2536: Could default this to `[]` perhaps?

sim/pokemon.ts
```diff
@@ -193,8 +193,13 @@ export class Pokemon {
...
 	moveThisTurnResult: boolean | null | undefined;
-	/** used for Assurance check */
-	hurtThisTurn: boolean;
+	/**
+	 * The undynamaxed HP value this Pokemon was reduced to by damage this turn,
+	 * or false if it hasn't taken damage yet this turn
+	 *
+	 * Used for Assurance, Emergency Exit, and Wimp Out
+	 */
+	hurtThisTurn: number | false;
```
urkerab, line 202: Is there any mileage in having this be zero instead of false?

MacChaeger: Functionally it would be pretty much the same either way, but `false` is easier to understand as “was not hurt this turn” than `0` is, in my opinion. This is a sort of hybrid property used for two admittedly somewhat unrelated checks, so I think it’s fitting for its type to be hybrid as well.

sim/battle.ts
MacChaeger: I don’t see the advantage of that. You’d just be checking `residualPokemon.length` instead of `residualPokemon`.

sim/pokemon.ts
urkerab: I just thought the extra cast looked odd that's all (I don't know if it's a common thing to do in PS! code).

sim/battle.ts
urkerab: Except that the `for` loop does that _for_ you anyway...

MacChaeger: Oh, right, I always forget for ofs work like that. That seems neater, then.

sim/pokemon.ts
Zarel: CONTRIBUTING specifies that this should be `number | null`, not `number | false`. Can you make that change?

data/scripts.ts
```diff
@@ -749,7 +749,7 @@ export const Scripts: BattleScriptsData = {
 				// The previous check was for `move.multihit`, but that fails for Dragon Darts
 				const curDamage = targets.length === 1 ? move.totalDamage : d;
 				if (typeof curDamage === 'number' && targets[i].hp) {
-					if (targets[i].hp <= targets[i].maxhp / 2 && targets[i].hp + curDamage > targets[i].maxhp / 2) {
+					if (targets[i].hp <= targets[i].maxhp / 2 && +targets[i].hurtThisTurn + curDamage > targets[i].maxhp / 2) {
```
Zarel, line 752:
```suggestion
					if (targets[i].hp <= targets[i].maxhp / 2 && (targets[i].hurtThisTurn || 0) + curDamage > targets[i].maxhp / 2) {
```

Zarel: I prefer `|| 0` to the cast.

Zarel:
Great work! Other than our nitpicks, this looks like a really good fix!

Zarel:
Looks great! Thanks!

PR #7720, opened 2020-11-23
Fix disqualifyUser bug in tournaments

Volco:
turns out since "inProgressMatch" only sets on 1 of the players the sub room deletions on line 733 only happen if the user it's set to is disqualified. Not sure if the inProgressMatch setting that way intentional or not so i just figured I'd pr this fix and if it's preferred for me to make it so inProgressMatch is set on both players instead of just one i can do that instead.

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -749,7 +749,13 @@ export class Tournament extends Rooms.RoomGame {
 		if (matchTo) {
 			matchTo.isBusy = false;
 			const matchRoom = matchTo.inProgressMatch!.room;
-			matchRoom.parent = null;
+			if (matchRoom.parent) {
+				matchRoom.parent.subRooms?.delete(matchRoom.roomid);
+				if (!matchRoom.parent.subRooms?.size) {
+					matchRoom.parent.subRooms = null;
+				}
+				matchRoom.parent = null;
+			}
 			this.completedMatches.add(matchRoom.roomid);
 			if (matchRoom.battle) matchRoom.battle.forfeit(player.id);
 			matchTo.inProgressMatch = null;


server/tournaments/index.ts
```diff
@@ -749,7 +749,13 @@ export class Tournament extends Rooms.RoomGame {
 		if (matchTo) {
 			matchTo.isBusy = false;
 			const matchRoom = matchTo.inProgressMatch!.room;
-			matchRoom.parent = null;
+			if (matchRoom.parent) {
```
MacChaeger, line 752: Won't this still crash with `property 'parent' does not exist on undefined` or something like that?

Volco: i figured it'll be fine since the current implementation is setting ``matchRoom.parent = null`` without checking if ``matchRoom`` even existed i could add the ``?`` to ``matchRoom`` though to be safe though

Volco: let me add a ``?`` to be safe

Volco: actually should be fine considering how much matchRoom is called without making sure it exists but i can add ``?`` if someone wants me to

Zarel:
It's plausible that we should have a `setParent` function which makes sure all the state stays consistent, like how `setPrivate` works.

Volco:
so how would you prefer this to be handled then?

Zarel:
I'm currently working on this myself.

Volco:
ah okay.

Zarel:
Thanks, by the way! After further investigation, I've confirmed this is the correct fix.

`setParent` is still important to prevent further bugs, but this is a good fix to have for now.

Volco:
No problem.

PR #7721, opened 2020-11-23
Refactor subroom code

Zarel:
This introduces a new function, `setParent`, to handle the details of setting up subrooms.

`roomid`, `parent`, and `subRooms` are now read-only, so they can't be accidentally be set directly rather than through their setters (`rename`, `setParent`, and `clearSubRooms`).

I don't think setters should be used for this, because I think it's important to be clear that `rename` and `setParent` will change a lot of other state and induce network activity.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -880,16 +880,9 @@ export const commands: ChatCommands = {
 			}
 		}
 
-		if (room.subRooms) {
-			for (const subRoom of room.subRooms.values()) subRoom.parent = null;
-		}
-
 		room.add(`|raw|<div class="broadcast-red"><b>This room has been deleted.</b></div>`);
-		room.update(); // |expire| needs to be its own message
-		room.add(`|expire|This room has been deleted.`);
-		this.sendReply(`The room "${title}" was deleted.`);
 		room.update();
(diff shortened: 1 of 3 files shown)


Zarel:
(This is the thing I was talking about in #7720)

Zarel:
I'm not a huge fan of the `as any` casts so that `.parent` is mutable only from inside `setParent`, but I don't think there are better approaches. I'm open to suggestions, though.

MacChaeger:
What about making `subRooms` a getter that refers to a `protected` property of some other name?

Zarel:
> What about making `subRooms` a getter that refers to a `protected` property of some other name?

`protected` isn't exactly what we want. We don't want other class methods to modify `subRooms`, either.

MacChaeger:
> > What about making `subRooms` a getter that refers to a `protected` property of some other name?
> 
> `protected` isn't exactly what we want. We don't want other class methods to modify `subRooms`, either.

I dunno. It seems like using that many casts in that many places would lead people to think that that’s the intended way of modifying it, and the comments you have now don’t say _not_ to do that.

Zarel:
I'd agree that more comments could be useful here.

AnnikaCodes:
@Zarel status on this?

Zarel:
@AnnikaCodes ready for merge any restart.

PR #7723, opened 2020-11-24
/usersearch: Sort users by offline/online status

mia-pi-git:
Requested by @DaWoblefet

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -492,7 +492,7 @@ export const commands: ChatCommands = {
 			);
 		}
 		Utils.sortBy(results, name => name.startsWith('  \u25C9'));
-		if (!results.length) return this.errorReply(`No users found.`);
+		if (!results.length) results = [`No users found.`];
 		return this.sendReplyBox(
 			`Users with a name matching '${target}':<br />${results.join('; ')}`
 		);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -492,7 +492,7 @@ export const commands: ChatCommands = {
 			);
 		}
 		Utils.sortBy(results, name => name.startsWith('  \u25C9'));
-		if (!results.length) results = [`No users found.`];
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Dammit. Accidentally forcepushed before a commit.

server/chat-commands/info.ts
```diff
@@ -491,6 +491,11 @@ export const commands: ChatCommands = {
 				Utils.html`${curUser.connected ? ` \u25C9 ` : ` \u25CC `} ${curUser.name}`
 			);
 		}
+		results.sort((aName, bName) => {
+			if (aName.includes(`\u25CC`) && bName.includes(`\u25C9`)) return 1;
+			return -1;
+		});
```
Zarel, line 497: This is not an ordering and can't be used in a sort. :/

It always returns `-1` if both users are online (or both users are offline), which is incorrectly ordered (they can't both be after the other). You should return `0` if both are online or both are offline:

```ts
		results.sort((aName, bName) => (
			(aName.includes(`\u25CC` ? 1 : 0) - (aName.includes(`\u25CC` ? 1 : 0)
		);
```

All of this is, of course, entirely unnecessary because I wrote a sort helper in Utils so you'd never have to deal with that again.

```suggestion
		Utils.sortBy(results, name => name.startsWith('  \u25C9'));
```

server/chat-commands/info.ts
```diff
@@ -491,6 +491,8 @@ export const commands: ChatCommands = {
 				Utils.html`${curUser.connected ? ` \u25C9 ` : ` \u25CC `} ${curUser.name}`
 			);
 		}
+		Utils.sortBy(results, name => name.startsWith('  \u25C9'));
+		if (!results.length) return this.errorReply(`No users found.`);
```
Zarel, line 495:
```suggestion
		if (!results.length) results = [`No users found.`];
```

Gives a nicer error message which reminds you what you searched for.

Zarel:
Thanks!

PR #7724, opened 2020-11-24
/randombattles: Check Gmax forme for moves

MacChaeger:
There are several Pokemon species for which only the Gmax forme has a random battles/random doubles move pool, and the /randbats and /randdubs commands would report that these Pokemon had no moves unless users specified said Gmax forme.

### DIFF
--- a/server/chat-plugins/random-battles.ts
+++ b/server/chat-plugins/random-battles.ts
@@ -280,16 +280,18 @@ export const commands: ChatCommands = {
 			}
 			return this.sendReplyBox(`<span style="color:#999999;">Moves for ${species.name} in ${formatName}:</span><br />${lgpeMoves}`);
 		}
-		if (!species.randomBattleMoves) {
+		let randomMoves = species.randomBattleMoves;
+		if (!randomMoves) {
 			const gmaxSpecies = dex.getSpecies(`${args[0]}gmax`);
 			if (!gmaxSpecies.exists || !gmaxSpecies.randomBattleMoves) {
 				return this.errorReply(`Error: No moves data found for ${species.name}${`gen${dex.gen}` in GEN_NAMES ? ` in ${GEN_NAMES[`gen${dex.gen}`]}` : ``}.`);
 			}
 			species = gmaxSpecies;
+			randomMoves = gmaxSpecies.randomBattleMoves;
 		}
 		const moves: string[] = [];
 		// Done because species.randomBattleMoves is readonly
-		for (const move of species.randomBattleMoves!) { // <- TypeScript bug: species.randomBattleMoves can't be undefined
+		for (const move of randomMoves) {
 			moves.push(move);
 		}
 		const m = moves.sort().map(formatMove);
--- a/server/chat-plugins/random-battles.ts
+++ b/server/chat-plugins/random-battles.ts
@@ -260,7 +260,7 @@ export const commands: ChatCommands = {
 			dex = Dex.mod(format.mod);
 			if (format.mod === 'letsgo') isLetsGo = true;
 		}
-		const species = dex.getSpecies(args[0]);
+		let species = dex.getSpecies(args[0]);
 		if (!species.exists) {
 			return this.errorReply(`Error: Pok\u00e9mon '${args[0].trim()}' does not exist.`);
 		}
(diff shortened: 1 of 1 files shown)


server/chat-plugins/random-battles.ts
```diff
@@ -281,11 +281,15 @@ export const commands: ChatCommands = {
...
+			const gmaxSpecies = dex.getSpecies(`${args[0]}gmax`);
+			if (!gmaxSpecies.exists || !gmaxSpecies.randomBattleMoves) {
+				return this.errorReply(`Error: No moves data found for ${species.name}${`gen${dex.gen}` in GEN_NAMES ? ` in ${GEN_NAMES[`gen${dex.gen}`]}` : ``}.`);
+			}
+			species = gmaxSpecies;
 		}
 		const moves: string[] = [];
 		// Done because species.randomBattleMoves is readonly
-		for (const move of species.randomBattleMoves) {
+		for (const move of species.randomBattleMoves!) { // <- TypeScript bug: species.randomBattleMoves can't be undefined
```
Zarel, line 292:
```suggestion
		for (const move of species.randomBattleMoves!) {
```

The `!` already means "this can be defined, but the mechanism by which it can be defined is too hard for TypeScript to check". I'm fine with comments that explain _why_ it's too hard for TS to check or something, but your comment as it stands doesn't seem to add anything.

Zarel: If you'd prefer not to have the assertion, you can also have a local `let randomBattleMoves = species.randomBattleMoves` variable.

MacChaeger: I added the comment because it seems like TypeScript should be able to tell that it’s defined, given the preceding checks, but OK.

server/chat-plugins/random-battles.ts
```diff
@@ -281,11 +281,15 @@ export const commands: ChatCommands = {
 			return this.sendReplyBox(`<span style="color:#999999;">Moves for ${species.name} in ${formatName}:</span><br />${lgpeMoves}`);
 		}
 		if (!species.randomBattleMoves) {
```
KrisXV, line 283:
```suggestion
		let randomMoves = species.randomBattleMoves;
		if (!randomMoves) {
```

server/chat-plugins/random-battles.ts
```diff
@@ -281,11 +281,15 @@ export const commands: ChatCommands = {
...
-			return this.errorReply(`Error: No moves data found for ${species.name}${`gen${dex.gen}` in GEN_NAMES ? ` in ${GEN_NAMES[`gen${dex.gen}`]}` : ``}.`);
+			const gmaxSpecies = dex.getSpecies(`${args[0]}gmax`);
+			if (!gmaxSpecies.exists || !gmaxSpecies.randomBattleMoves) {
+				return this.errorReply(`Error: No moves data found for ${species.name}${`gen${dex.gen}` in GEN_NAMES ? ` in ${GEN_NAMES[`gen${dex.gen}`]}` : ``}.`);
+			}
+			species = gmaxSpecies;
```
KrisXV, line 288:
```suggestion
			species = gmaxSpecies;
			randomMoves = gmaxSpecies.randomBattleMoves;
```

server/chat-plugins/random-battles.ts
KrisXV, line 292:
```suggestion
		for (const move of randomMoves) {
```

KrisXV (commented):
repeat this for /randdubs too

server/chat-plugins/random-battles.ts
KrisXV: This also fixes the bug i believe

Zarel: TypeScript is _sometimes_ nice enough to check properties, but it should never be relied upon.

Zarel: (Kris's fix is the one I suggested.)

Zarel:
Looks good!

Zarel:
Which Pokémon have Gmax moves but no regular moves, though?

MacChaeger:
A whole bunch. I'd ask TI about it.

PR #7726, opened 2020-11-24
Moderation: Properly disallow forcerenamed offline names

mia-pi-git:
Not really sure why this was done in the first place? Seems to be weird though.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1522,16 +1522,17 @@ export const commands: ChatCommands = {
 		this.checkCan('forcerename', targetUser);
 
 		let forceRenameMessage;
+		const lastRename = (Monitor.forceRenames.get(targetUser.id) || 0);
 		if (targetUser.connected) {
 			forceRenameMessage = `was forced to choose a new name by ${user.name}${(reason ? `: ${reason}` : ``)}`;
 			this.globalModlog('FORCERENAME', targetUser, reason);
-			Monitor.forceRenames.set(targetUser.id, (Monitor.forceRenames.get(targetUser.id) || 0) + 1);
+			Monitor.forceRenames.set(targetUser.id, lastRename + 1);
 			Ladders.cancelSearches(targetUser);
 			targetUser.send(`|nametaken||${user.name} considers your name inappropriate${(reason ? `: ${reason}` : ".")}`);
 		} else {
 			forceRenameMessage = `would be forced to choose a new name by ${user.name} but is offline${(reason ? `: ${reason}` : ``)}`;
 			this.globalModlog('FORCERENAME OFFLINE', targetUser, reason);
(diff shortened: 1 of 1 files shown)


server/chat-commands/moderation.ts
```diff
@@ -1522,16 +1522,17 @@ export const commands: ChatCommands = {
...
 			this.globalModlog('FORCERENAME', targetUser, reason);
-			Monitor.forceRenames.set(targetUser.id, (Monitor.forceRenames.get(targetUser.id) || 0) + 1);
+			Monitor.forceRenames.set(targetUser.id, lastRename + 1);
 			Ladders.cancelSearches(targetUser);
 			targetUser.send(`|nametaken||${user.name} considers your name inappropriate${(reason ? `: ${reason}` : ".")}`);
 		} else {
 			forceRenameMessage = `would be forced to choose a new name by ${user.name} but is offline${(reason ? `: ${reason}` : ``)}`;
 			this.globalModlog('FORCERENAME OFFLINE', targetUser, reason);
-			if (!Monitor.forceRenames.has(targetUser.id)) Monitor.forceRenames.set(targetUser.id, 0);
+			Monitor.forceRenames.set(targetUser.id, lastRename + 1);
```
Zarel, line 1535: It seems like this could just be moved down two lines, instead of needing to be replaced twice? or even moved up above `if (targetUser.connected)`.

Zarel:
```suggestion
```

server/chat-commands/moderation.ts
```diff
@@ -1522,16 +1522,17 @@ export const commands: ChatCommands = {
 		this.checkCan('forcerename', targetUser);
 
 		let forceRenameMessage;
+		const lastRename = (Monitor.forceRenames.get(targetUser.id) || 0);
 		if (targetUser.connected) {
 			forceRenameMessage = `was forced to choose a new name by ${user.name}${(reason ? `: ${reason}` : ``)}`;
 			this.globalModlog('FORCERENAME', targetUser, reason);
-			Monitor.forceRenames.set(targetUser.id, (Monitor.forceRenames.get(targetUser.id) || 0) + 1);
+			Monitor.forceRenames.set(targetUser.id, lastRename + 1);
```
Zarel, line 1529:
```suggestion
```

server/chat-commands/moderation.ts
```diff
@@ -1522,16 +1522,17 @@ export const commands: ChatCommands = {
 		this.checkCan('forcerename', targetUser);
 
 		let forceRenameMessage;
+		const lastRename = (Monitor.forceRenames.get(targetUser.id) || 0);
```
Zarel, line 1525:
```suggestion
		Monitor.forceRenames.set(targetUser.id, (Monitor.forceRenames.get(targetUser.id) || 0) + 1);
```

Zarel:
Thanks!

PR #7727, opened 2020-11-24
/dexsearch: Fix immunity check for Thousand Arrows

MacChaeger:
Unlike /effectiveness and /coverage, /dexsearch doesn't need to support arbitrary mods, so Thousand Arrows can be hard-coded as the only move to ignore any immunities. Simplifying the check like this was pretty much necessary for me to actually be able to understand how the logic should be structured to accommodate the opposing exceptions for Thousand Arrows and Sheer Cold.


Zarel:
Looks good, thanks!

PR #7729, opened 2020-11-25
Correct Protean interactions & Gravity-blocked status Z-moves

DaWoblefet:
Originally, I was just going to add a couple additional Protean tests on The Immortal's recommendation. Thanks to additional [USUM research](https://pastebin.com/bDtvxdT2), we now have a more complete list of what I call in the checks done before move success "[move failure checks part 1](https://docs.google.com/spreadsheets/d/1nn0b7Na9UEoE_NChLLlknV3ERacNeenV4TiMieTpS8U/edit#gid=0)". I just decided to implement it. The only new checks we learned about were:

- Aurora Veil without hail
- Splash / Magnet Rise / Telekinesis used in Gravity - this is why Z-Splash in Gravity boosts to +3 attack but still fails
- Teleport (pre-Gen 8)
- Magnet Rise when the user has Smack Down or Ingrain status
- Sky Drop if the thing it had picked up already fainted

However, Showdown was very inconsistent on how many of its effects were in onTry and onPrepareHit, which are what you need to be pre-Protean. I swapped most of these over and all the tests are passing, but this will definitely be a Marty thing to review.

The only ones I don't know how to do for sure are Counter, Mirror Coat, Metal Burst, and Bide. Their onTryHits don't seem to as easily be moved to onTry unless I'm missing something.

There's also a weird thing with the Protect checks where they do !!this.queue.willAct() and I think this is the "moved last in the turn" check but I'm not 100% sure.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -7304,6 +7304,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 					return false;
 				}
 			},
+
 			onResidualOrder: 22,
 			onEnd() {
 				this.add('-fieldend', 'move: Gravity');
@@ -8590,7 +8591,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 				return null;
 			}
 			this.attrLastMove('[still]');
-			this.add('-fail', pokemon, 'move: Hyperspace Fury');
+			this.add('-fail', source, 'move: Hyperspace Fury');
 			return null;
 		},
 		self: {
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -773,7 +773,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {snatch: 1},
 		sideCondition: 'auroraveil',
-		onTryHitSide() {
+		onTry() {
 			if (!this.field.isWeather('hail')) return false;
 		},
 		condition: {
@@ -913,8 +913,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {},
 		stallingMove: true,
 		volatileStatus: 'banefulbunker',
-		onTryHit(target, source, move) {
-			return !!this.queue.willAct() && this.runEvent('StallMove', target);
+		onPrepareHit(pokemon) {
+			return !!this.queue.willAct() && this.runEvent('StallMove', pokemon);
 		},
 		onHit(pokemon) {
 			pokemon.addVolatile('stall');
@@ -2182,10 +2182,12 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 5,
 		priority: 0,
 		flags: {snatch: 1, sound: 1, dance: 1},
-		onTryHit(pokemon, target, move) {
+		onTry(pokemon, target, move) {
 			if (pokemon.hp <= (pokemon.maxhp * 33 / 100) || pokemon.maxhp === 1) {
 				return false;
 			}
+		},
+		onTryHit(pokemon, target, move) {
 			if (!this.boost(move.boosts as SparseBoostsTable)) return null;
 			delete move.boosts;
 		},
@@ -2761,7 +2763,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 3,
 		flags: {},
 		sideCondition: 'craftyshield',
-		onTryHitSide(side, source) {
+		onTry(side, source) {
 			return !!this.queue.willAct();
 		},
 		condition: {
@@ -2973,7 +2975,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {protect: 1, reflectable: 1, mirror: 1},
 		status: 'slp',
-		onTryMove(pokemon, target, move) {
+		onTry(pokemon, target, move) {
 			if (pokemon.species.name === 'Darkrai' || move.hasBounced) {
 				return;
 			}
@@ -4306,8 +4308,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {},
 		stallingMove: true,
 		volatileStatus: 'endure',
-		onTryHit(pokemon) {
-			return this.queue.willAct() && this.runEvent('StallMove', pokemon);
+		onPrepareHit(pokemon) {
+			return !!this.queue.willAct() && this.runEvent('StallMove', pokemon);
 		},
 		onHit(pokemon) {
 			pokemon.addVolatile('stall');
@@ -5497,7 +5499,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 2,
 		flags: {},
 		volatileStatus: 'followme',
-		onTryHit(target) {
+		onTry(target) {
 			if (target.side.active.length < 2) return false;
 		},
 		condition: {
@@ -9226,7 +9228,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {},
 		stallingMove: true,
 		volatileStatus: 'kingsshield',
-		onTryHit(pokemon) {
+		onPrepareHit(pokemon) {
 			return !!this.queue.willAct() && this.runEvent('StallMove', pokemon);
 		},
 		onHit(pokemon) {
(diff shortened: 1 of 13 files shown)


test/sim/abilities/protean.js
```diff
@@ -28,18 +32,90 @@ describe('Protean', function () {
...
+		// More examples: https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8548957
 	});
 
-	it('should not change the user\'s type when using Fling with no item', function () {
-		battle = common.createBattle();
-		battle.setPlayer('p1', {team: [{species: "Kecleon", ability: 'protean', moves: ['fling']}]});
-		battle.setPlayer('p2', {team: [{species: "Wobbuffet", moves: ['counter']}]});
-		battle.makeChoices('move fling', 'move counter');
-		assert(battle.p1.active[0].hasType('Normal'));
+	it(`should not change the user's type when abilities that fail earlier than Protean will activate`, function () {
```
MacChaeger, line 87: This is worded strangely. I would change it to something like:
```suggestion
	it(`should not change the user's type when abilities that activate earlier than Protean cause user's move to fail`, function () {
```

test/sim/abilities/protean.js
```diff
@@ -28,18 +32,90 @@ describe('Protean', function () {
...
+			{species: "Groudon", ability: 'desolateland', moves: ['powder']},
+			{species: "Tsareena", ability: 'dazzling', moves: ['sleeptalk']},
+			{species: "Golduck", ability: 'damp', moves: ['sleeptalk']},
+		]]);
+
+		const kecleon = battle.p1.active[0];
+
+		// Primordial Sea
+		battle.makeChoices('move mindblown', 'move sleeptalk');
+		assert(kecleon.hasType('Normal'), "Primordial Sea changed the user's type when it should not have.");
```
MacChaeger, line 101: I also think these should probably say something more along the lines of "<ability> allowed the user's type to change" since Protean would be the one actually changing the user's type.

test/sim/abilities/protean.js
DaWoblefet: Good idea!

test/sim/abilities/protean.js
DaWoblefet: I want to make it clear which effect is causing the move to fail. Otherwise, if it's 1 failing and they all have the same error come back, it'd be annoying to track down which of the 5 was wrong. I did a compromise of "Protean + thing".

data/moves.ts
```diff
@@ -10108,10 +10110,18 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {snatch: 1, gravity: 1},
 		volatileStatus: 'magnetrise',
+		onTry(source, target, move) {
+			if (target.volatiles['smackdown'] || target.volatiles['ingrain']) return false;
+
+			// Additional Gravity check for Z-move variant
```
urkerab, line 10116: What I suspect you need is an additional event handler (which event depends on when the Z-move fails relative to other effects such as Protean) for Gravity to use to block these moves. It does depend somewhat on whether Gravity should be blocking called airborne moves. (For instance, you could hack a Pokémon with Protean, Sleep Talk and Telekinesis while its ally sends it to sleep and sets Gravity. Similarly, should Throat Chop block called sound moves or Heal Block block called healing moves, and if so, at what point do they fail, relative to other effects such as Protean?)

DaWoblefet: Like I linked in my pastebin above, the game literally does these Gravity checks in the same event handler as the others effects I touched, so I think it's more accurate to include them here directly.

Gravity, Throat Chop, Heal Block, and whether a move is blocked in a Sky Battle are all ran on submoves. They fail in the pre-move success checks The move proper would deduct PP, but the user won't become Choice locked or switch from Stance Change or set a target. Some tests from my Google Doc (this is Gen 8)

first action is Sleep Talk while asleep -> calls Snore -> blocked by Throat Chop -> Pokémon not locked into Sleep Talk (PP deducted)
Sleep Talk while asleep -> calls Snore -> blocked by Throat Chop -> (1 PP deducted despite both opponents having Pressure)
Sleep Talk while asleep -> calls Snore into Pressure -> 2 PP deducted

SadisticMystic's SM notes say Choice locking occurred prior to submove calls in 7.

DaWoblefet:
Thank you for cleaning up my ``onTry``s @TheImmortal!

data/moves.ts
urkerab: It's the triplication of code that I was most concerned about, but the centralisation would also make it easy to adapt for called airborne moves under Gravity.

As for ensuring that the Gravity event handler runs at the appropriate time, based on my reading of your comments and links, this actually needs to be `onModifyMove` anyway so that it happens before the move is announced and targetted. (Certainly for Telekinesis the Smack Down and Ingrain check can't happen until after the move is targetted!)

(Note that currently on the sim Stance Change never triggers for Sleep Talk calling an attack, so if that's wrong it probably also needs to switch to using the `onModifyMove` event.)

Zarel:
All these changes look good to me. Is this ready to merge?

Marty-D (changes requested):
You'll notice Sucker Punch sends `-fail` manually because returning `false` from `onTry` doesn't send any message.

data/moves.ts
DaWoblefet: > It's the triplication of code that I was most concerned about, but the centralisation would also make it easy to adapt for called airborne moves under Gravity.
That's a fair point, and one you could have poked Game Freak about. :P

> As for ensuring that the Gravity event handler runs at the appropriate time, based on my reading of your comments and links, this actually needs to be `onModifyMove` anyway so that it happens before the move is announced and targetted. (Certainly for Telekinesis the Smack Down and Ingrain check can't happen until after the move is targetted!)
Strange as it might be, the Gravity check really is supposed to be here: it really does announce the move, set the target, and deduct PP. Here's a video of behavior with me targeting a Pressure Pokemon and losing 2 PP while Z-Telekinesis gets stopped by Gravity: https://www.youtube.com/watch?v=Lppqg2ifVR8 The Gravity checks really are just are in this location.

> (Note that currently on the sim Stance Change never triggers for Sleep Talk calling an attack, so if that's wrong it probably also needs to switch to using the `onModifyMove` event.)
TI fixed this in an independent commit!

DaWoblefet:
@Marty-D cleaned those up as well as a few missing [still] messages to hide the target.

I still don't know what to do about Bide, Counter, Mirror Coat, and Metal Burst.

data/moves.ts
urkerab: > That's a fair point, and one you could have poked Game Freak about. :P

So they only special-cased those three moves? What about Supersonic Skystrike? Or does some other check defeat it? What if you try to (say) Z-Assist a Supersonic Skystrike?

> https://www.youtube.com/watch?v=Lppqg2ifVR8

Sorry, I misunderstood that the two documents needed to be read in combination.

urkerab: Even after rewriting several times that still came out wrong. I meant I misunderstood how the documents needed to be combined.

DaWoblefet: All good, it's very bizarre (and newly discovered!) behavior.

> So they only special-cased those three moves? What about Supersonic Skystrike? Or does some other check defeat it? What if you try to (say) Z-Assist a Supersonic Skystrike?

That's right; Z-Fly, Z-Bounce etc. wouldn't be impacted by this. It's only an explicit Gravity check for those 3 moves.

urkerab: (What was also confusing me is that Supersonic Skystrike bypasses Gravity, but it does that anyway because it's not marked as airborne!)

data/moves.ts
```diff
@@ -16040,8 +16001,6 @@ export const Moves: {[moveid: string]: MoveData} = {
 		sleepUsable: true,
 		onTry(source) {
 			if (source.status !== 'slp' && !source.hasAbility('comatose')) {
-				this.attrLastMove('[still]');
-				this.add('-fail', source);
 				return false;
 			}
 		},
```
urkerab: (Could this not be changed to something like `return source.status === 'slp' || source.hasAbility('comatose');`?)

DaWoblefet: Yes, with Marty's new refactor of onTry() to have false automatically send 

``this.attrLastMove('[still]');
this.add('-fail', source); ``

on returning false, a few more of these can be further cleaned up. I'll see what remains.

data/scripts.ts
```diff
@@ -572,11 +572,16 @@ export const Scripts: BattleScriptsData = {
 	tryMoveHit(target, pokemon, move) {
 		this.setActiveMove(move, pokemon, target);
 
-		if (!this.singleEvent('Try', move, null, pokemon, target, move)) {
+		let hitResult = this.singleEvent('Try', move, null, pokemon, target, move);
+		if (!hitResult) {
+			if (hitResult === false) {
+				this.add('-fail', pokemon);
+				this.attrLastMove('[still]');
+			}
```
urkerab, line 580: Ah right, this was added to the other call site to fix future moves (well it makes more sense than individually annotating each move). But at this point, is there any difference between the `Try` and `PrepareHit` singleEvents? (Aside from the parameter order of course.)

DaWoblefet: There is some stuff after ``Try`` but before Protean, notably future attacks. Might not hurt to repurpose it, but that's beyond the scope of what I want to do in this PR.

Marty-D:
I still want to double check past gens in a bit to make sure any existing overrides are handled correctly, but other than that I don't think Bide/Counter have to be dealt with in this PR. No need to get this caught in development hell.

Marty-D (approved):

PR #7731, opened 2020-11-25
Residual damage Wimp Out timing is correct

MacChaeger:
Apparently my previous Emergency Exit change also fixed what would have been the last skipped Emergency Exit test. However, this also adds a new skipped test for a minor visual bug I don't really know how to fix, which is that when a Pokemon with Wimp Out or Emergency Exit forces another Pokemon out with Dragon Tail or Circle Throw and takes recoil damage that drops it down to less than 50% HP, the ability visibly activates before the target is switched out, though the player controlling the attacker in that case is correctly asked to switch their Pokemon out after the target's replacement is dragged in. The only error is in the message timing.

### DIFF
--- a/test/sim/abilities/emergencyexit.js
+++ b/test/sim/abilities/emergencyexit.js
@@ -83,15 +83,17 @@ describe(`Emergency Exit`, function () {
 		assert(!battle.p2.activeRequest.forceSwitch);
 	});
 
-	it.skip(`should request switch-out before end-of-turn fainted Pokemon`, function () {
-		battle = common.createBattle([
-			[{species: "Golisopod", ability: 'emergencyexit', item: 'toxicorb', moves: ['substitute', 'sleeptalk', 'liquidation']}, {species: "Magikarp", moves: ['splash']}],
-			[{species: "Charizard", item: 'rockyhelmet', moves: ['bellydrum', 'sleeptalk', 'dragonclaw']}, {species: "Mew", moves: ['sleeptalk']}],
-		]);
-		battle.makeChoices('move substitute', 'move bellydrum');
-		battle.makeChoices('move sleeptalk', 'move sleeptalk');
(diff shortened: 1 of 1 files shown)


test/sim/abilities/emergencyexit.js
```diff
@@ -83,15 +83,17 @@ describe(`Emergency Exit`, function () {
...
-		battle = common.createBattle([
-			[{species: "Golisopod", ability: 'emergencyexit', item: 'toxicorb', moves: ['substitute', 'sleeptalk', 'liquidation']}, {species: "Magikarp", moves: ['splash']}],
-			[{species: "Charizard", item: 'rockyhelmet', moves: ['bellydrum', 'sleeptalk', 'dragonclaw']}, {species: "Mew", moves: ['sleeptalk']}],
-		]);
-		battle.makeChoices('move substitute', 'move bellydrum');
-		battle.makeChoices('move sleeptalk', 'move sleeptalk');
-		battle.makeChoices('move liquidation', 'move dragonclaw');
+	it(`should request switch-out before end-of-turn fainted Pokemon`, function () {
+		battle = common.createBattle([[
+			{species: "Golisopod", item: 'blacksludge', ability: 'emergencyexit', moves: ['payback']},
```
DaWoblefet, line 88: Can you use an OHKO move instead of Payback? It'd make it more obvious the Swoobat is intended to die.

DaWoblefet (commented):
Everything else looks great!

test/sim/abilities/emergencyexit.js
Zarel: tbh I'm not sure how you can make it any more obvious than `assert.fainted`.

DaWoblefet: either way is fine, that's a very minor nitpick

PR #7732, opened 2020-11-25
Youtube: Add support for watching videos as a group

mia-pi-git:
Requested by YT staff.


Zarel:
...how are you syncing them? And shouldn't you make the video bigger or something?

mia-pi-git:
> ...how are you syncing them? And shouldn't you make the video bigger or something?

I'm not syncing them since that's a massive technical headache and people want to be able to pause the video on their own, etc. as for size, we (me + yt staff) worked out the design and we landed on this.

PR #7733, opened 2020-11-26
Don't activate Flower Gift/Ice Face when KOed

MacChaeger:


TheImmortal:
The correct fix is probably the Update event not running on fainted Pokemon.

@Marty-D ?

Marty-D:
This happens all over the place already, maybe could use a refactor one day but I think that's out of scope for this change.

PR #7739, opened 2020-11-28
Fix Max Guard Bypass with Multiple Targets

aviettran:
Still learning the codebase. Feedback encouraged.


DaWoblefet:
The tests indicate correct behavior for sure, but the implementation does look a bit weird.

aviettran:
I am open to suggestions. My main motivation here is to learn the codebase.

Marty-D:
I believe doing it this way will activate immunity Abilities which also use `TryHit` (such as Motor Drive or Volt Absorb), but this is actually fine for Teatime since it's supposed to behave like Perish Song in this way. The effects that check Plus / Minus would only matter for Other Metas that allow multiple Abilities, and Flower Shield only with convoluted Electrify shenanigans and type or Ability swapping, but I imagine someday someone would report a bug.

So I'd recommend checking for the Max Guard volatile before running `TryHit` for all these effects except Teatime. @DaWoblefet do these get loudly blocked or silently blocked?

urkerab:
(That's a bit of a shame for Gear up and Magnetic Flux because changing their target type from "allySide" to "allies" would have automatically run the `TryHit` events for regular spread moves. I don't think we have a more suitable target type for Teatime though.)

DaWoblefet:
> I believe doing it this way will activate immunity Abilities which also use `TryHit` (such as Motor Drive or Volt Absorb), but this is actually fine for Teatime since it's supposed to behave like Perish Song in this way. The effects that check Plus / Minus would only matter for Other Metas that allow multiple Abilities, and Flower Shield only with convoluted Electrify shenanigans and type or Ability swapping, but I imagine someday someone would report a bug.
> 
> So I'd recommend checking for the Max Guard volatile before running `TryHit` for all these effects except Teatime. @DaWoblefet do these get loudly blocked or silently blocked?

All of the Max Guard-blocked effects are properly announced with a "[Pokemon] protected itself!". Here's Block, Flower Shield, Gear Up, Magnetic Flux, Psych Up, Teatime, and Transform: https://youtu.be/5C6JfhhXn2w?t=1383

Max Guard does happen in the checks done before move success prior to something like Lightning Rod or Magic Bounce, so the aforementioned Electrified Flower Shield would not trigger that Max Guarded Pokemon's Lightning Rod.

aviettran:
Based on this discussion, I have:

1. Added the maxguard volatile check before the TryHit event
2. Added an assertion to check the battle log: "[Pokemon] protected itself!"

test/sim/moves/maxguard.js
```diff
@@ -34,7 +34,7 @@ describe('Max Guard', function () {
 		assert.false.fainted(minun);
 	});
 
-	it.skip('should block certain moves that bypass Protect', function () {
+	it('should block certain moves that bypass Protect', function () {
```
MacChaeger, line 37: I think it would make more sense to merge more of the tests into this one. It seems a bit silly to have a test just for Gear Up when this test already has a practically identical Magnetic Flux test. Surely we could just give the Pokémon in this test a few more moves to use? It’s not like they have to be legal, either. We can easily have a Kartana with Plus and like 5 moves.

Marty-D (approved):
Looks good to me, well done and thanks for the changes! 👍

DaWoblefet:
Seconding, thanks!

urkerab:
(More precisely, Gear Up and Magnetic Flux behave as if their list of targets was "all allies with an active ability of Plus or Minus", which means that you don't have to worry about the target having an immunity ability at that point, since we know what their ability is. And Partners in Crime can just deal with it.)

PR #7740, opened 2020-11-28
Scavengers: Expand the functionality of /scav timer

aQrator:
Adjust the _/scav timer_ function to allow for turning off the timer, and to enable fractional times once again.

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -577,10 +577,10 @@ export class ScavengerHunt extends Rooms.RoomGame {
 			this.timer = null;
 			this.timerEnd = null;
 		}
-		
-		if (minutes == 0) { 
+
+		if (minutes === 0) {
 			return 'off';
-		}		
+		}
 		if (minutes > 24 * 60) { // 24 hours
 			throw new Chat.ErrorMessage(`Time limit must be under 24 hours (you asked for ${Chat.toDurationString(minutes * 60000)}).`);
 		}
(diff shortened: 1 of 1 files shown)


aQrator:
Proof of the request by the RO LittEleven
afbeelding

server/chat-plugins/scavengers.ts
```diff
@@ -577,7 +577,10 @@ export class ScavengerHunt extends Rooms.RoomGame {
 			this.timer = null;
 			this.timerEnd = null;
 		}
-
+		
+		if (minutes == 0) { 
+			return 'off';
+		}
```
AnnikaCodes, line 583:
```suggestion

		if (minutes === 0) { 
			return 'off';
		}		
```

server/chat-plugins/scavengers.ts
```diff
@@ -1483,9 +1486,16 @@ const ScavengerCommands: ChatCommands = {
...
 		const game = room.getGame(ScavengerHunt);
 		if (!game) return this.errorReply(`There is no scavenger hunt currently running.`);
+		
+		let minutes;
+		if (target == "off") {
+			minutes = 0;
+		}
+		else {
+			minutes = parseFloat(target);
+		}
```
AnnikaCodes, line 1496:
```suggestion

		const minutes = (toID(target) === 'off' ? 0 : parseFloat(target));
```

AnnikaCodes: (To be clear, `game.setTimer(0)` should clear the timeout, not set it to 0 milliseconds, right?)

AnnikaCodes (changes requested):

server/chat-plugins/scavengers.ts
aQrator: That is correct

AnnikaCodes: Also, if you were to do this with an if statement, I'd do it like this:
```ts
let minutes = 0;
if (target !== 'off) minutes = parseFloat(target);
```
or
```ts
let minutes;
if (target === 'off') {
     minutes = 0;
} else {
     minutes = parseFloat(target);
}
```

server/chat-plugins/scavengers.ts
aQrator: Adjusted alongside other errors

server/chat-plugins/scavengers.ts
aQrator: Thanks. I will remember this in the future. For now, I've just used the 1-line const declaration

AnnikaCodes (approved):
Looks fine to me

aQrator:
Thank you. When can we expect this update to go live?

AnnikaCodes:
> Thank you. When can we expect this update to go live?

idt I'm supposed to merge things, so give it a few days so that Zarel (or xfix) can review, and merge and someone can hotpatch the change in

sparkychildcharlie:
looks good

xfix (approved):

PR #7741, opened 2020-11-29
Punishments: Show users a warn message on login for offline warns

mia-pi-git:
https://www.smogon.com/forums/threads/3673524/#post-8662678


Zarel:
Thanks! Should we care about persistence past restarts?

PR #7742, opened 2020-11-29
Support using more moderation commands offline

mia-pi-git:
https://www.smogon.com/forums/threads/3673524/


Zarel:
Seems good. What's the build error?

mia-pi-git:
@Zarel, a test is failing but i have no idea why (it works fine locally?)

Zarel:
It means you broke `user.can`.

`user1.can('mute', user2)` should return `true` if `user1` is a driver and `user2` is a voice.

Are you sure it works locally? I can definitely repro this:

```
zarel@Adenosine ~/W/psserver (offline-punishments)>
mocha -g "should allow 'u' permissions"


  !

  0 passing (206ms)
  1 failing

  1) Users features
       Users
         User
           #can
             should allow 'u' permissions on lower ranked users:

      AssertionError [ERR_ASSERTION]: targeting self
      + expected - actual

      -true
      +false

      at Context.<anonymous> (test/server/users.js:163:13)
      at processImmediate (node:internal/timers:462:21)
```

mia-pi-git:
> It means you broke `user.can`.
I figured as much :p

> `user1.can('globalban', user1)` should return `false` if `user1` is a global mod.
> 
> Are you sure it works locally? I can definitely repro this:

https://prnt.sc/vufsgy

mia-pi-git:
I have now un-broken user.can.

server/chat-commands/moderation.ts
```diff
@@ -785,23 +812,24 @@ export const commands: ChatCommands = {
...
 		} else {
 			name = this.targetUsername;
 			userid = toID(this.targetUsername);
 		}
 
+
+		if (Users.isTrusted(userid)) {
+			if (force) {
+				const from = runCrisisDemote(userid);
+				Monitor.log(`[CrisisMonitor] ${name} was locked by ${user.name} and demoted from ${from!.join(", ")}.`);
```
Zarel, line 824: Why `from!`?

test/server/users.js
```diff
@@ -159,15 +159,16 @@ describe('Users features', function () {
 				});
 				it(`should allow 'u' permissions on lower ranked users`, function () {
 					const user = new User();
-					user.tempGroup = '@';
+					user.forceRename("Moderator");
+					user.setGroup('@');
```
Zarel, line 163: Why `setGroup`? Did you change how `tempGroup` works?

mia-pi-git: No, I just changed that in my attempts to figure out what was wrong and then never changed it back, since it sets tempGroup and i made User.can work with tempGroup.

Zarel: Can you change it back now, then?

mia-pi-git: Alright.

server/user-groups.ts
```diff
@@ -117,7 +117,16 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 		const auth: Auth = room ? room.auth : Users.globalAuth;
 
 		const symbol = auth.getEffectiveSymbol(user);
-		let targetSymbol = (typeof target === 'string' || !target) ? target : auth.get(target);
+
+		let targetSymbol = target as GroupSymbol;
```
Zarel, line 121: `target` is actually not necessarily a `GroupSymbol` here. I think the previous cast was in the correct location.

Zarel:
Okay, we'll assume this is ready. Seems like it'll require a restart? I plan to restart soon, anyway.

mia-pi-git:
> Okay, we'll assume this is ready. Seems like it'll require a restart? I plan to restart soon, anyway.

Well, it doesn't crash hotpatch iirc. We _would_ need to hotpatch chat, hotpatch punishments, and monkeypatch Users.Auth, though.

PR #7745, opened 2020-11-29
Users: Reapply hostfilter locks after user merge

mia-pi-git:


asgdf:
It should probably only check latestIp, otherwise we would go back to people having to log out and refresh after getting off proxies

Zarel:
@mia-pi-git?

mia-pi-git:
Oh, we talked about this and then i got wrapped up in other things. I'll commit that now, it's pretty easy.

server/users.ts
```diff
@@ -997,6 +997,9 @@ export class User extends Chat.MessageContext {
 		this.userMessage = oldUser.userMessage || this.userMessage || '';
 
 		oldUser.markDisconnected();
+
+		const latestConn = this.connections.find(c => c.ip === this.latestIp);
+		void Punishments.checkIp(this, latestConn!);
```
Zarel, line 1002:
```suggestion
		const latestConnection = this.connections[this.connections.length - 1];
		if (latestConnection) void Punishments.checkIp(this, latestConnection);
```

Zarel:
Wait, why is this necessary at all? Don't the locks already apply to the individual connections?

Zarel:
After some discussion in Staff, I think this is the wrong direction:

[14:51] #Zarel: like, you should only be hostfilter locked if all your connections are hostfilter locked
[14:51] #Zarel: we can mark the connections with hostfilter lock status if that would make it easier

asgdf:
I'd just like to point out that I consider this a very high-priority fix, on account of the entire hostfilter system not working at all at the moment

Zarel:
> I'd just like to point out that I consider this a very high-priority fix, on account of the entire hostfilter system not working at all at the moment

That's awkward because this requires a restart to apply which is definitely not happening today...

Zarel:
Actually, it might not.......

mia-pi-git:
> > I'd just like to point out that I consider this a very high-priority fix, on account of the entire hostfilter system not working at all at the moment
> 
> That's awkward because this requires a restart to apply which is definitely not happening today...

Well, this can be monkeypatched / hotpatched (ish?) - just would be difficult.

asgdf:
To be clear we'll survive some more days if need be, I just wanted to caution against the regression bugfix turning into a refactor PR

asgdf:
It seems Annika has pushed a fix for now (eceb8549ddfc26ddaf1a9c6810455357296d6626), so this is no longer as urgent - refactor away!

Zarel:
OK. I think this should be closed since it is bad code. The current code is pretty fucked up but this just piles more hacks upon hacks instead of being a proper refactor.

Zarel:
In particular, I'm unhappy with this `if` statement:

```ts
// If either user is unlocked and neither is locked by name, remove the lock.
// Otherwise, keep any locks that were by name.
if (
	(!oldUser.locked || !this.locked) &&
	oldUser.locked !== oldUser.id &&
	this.locked !== this.id &&
	// Only unlock if no previous names are locked
	!oldUser.previousIDs.some(id => !!Punishments.search(id)
		.filter(punishment => punishment[2][0] === 'LOCK' && punishment[2][1] === id)
		.length)
) {
```

Zarel:
We should not need to do five Punishment searches to merge two users. :[

Zarel:
I think the solution is for the user to track if it's been "strongly locked" or "weakly locked".

PR #7748, opened 2020-11-29
Correct eslint.options.configFile for vscode

aviettran:
Fixes the error in vscode for the ESLint Plugin:

Error: ENOENT: no such file or directory, open '******/pokemon-showdown/.eslintrc-types.json'


Zarel:
Thanks!

PR #7751, opened 2020-11-30
Nicknamefilter: Support hiding nicks from filters

mia-pi-git:


Zarel:
Looks like this requires a restart?

Zarel:
Hm, having read the whole thing, I think it's safe to merge now for a restart whenever; it wouldn't break hotpatching. Would you agree?

Zarel (approved):

mia-pi-git:
Yeah, i said this on discord but if someone else wants to do it, i'll leave it here. This doesn't need a restart but does need a monkeypatch (https://gist.github.com/mia-pi-git/1768d1bc0b55694a1b158502b5b47290)

Zarel:
ty!

PR #7755, opened 2020-12-01
Add missing </details>

PartMan7:
Doesn't really matter since it gets automatically added, but doesn't feel right without it. 'o.o


Zarel:
ty!

PR #7756, opened 2020-12-01
Make `/weak type1/type2` work

PartMan7:
Seems like this should exist, especially because the help for the command shows: `/weakness [type 1]/[type 2] - Provides a type or type combination's resistances, weaknesses, and immunities, ignoring abilities.`


Zarel:
Looks good, thanks!

PR #7757, opened 2020-12-01
Users: Properly assign isStaff to staff roomauth

mia-pi-git:
(We already monkeypatched this on the server and it works, just PR'ing over pushing to be safe)


Zarel:
Thanks!

PR #7758, opened 2020-12-02
Bugfix: Disallow Snatch from calling sound moves while Throat Chopped; healing while Heal Blocked

aviettran:


data/moves.ts
```diff
@@ -15961,6 +15961,18 @@ export const Moves: {[moveid: string]: MoveData} = {
 				}
 				snatchUser.removeVolatile('snatch');
 				this.add('-activate', snatchUser, 'move: Snatch', '[of] ' + source);
+				// Call select onBeforeMove handlers
+				const handlersToRun = {"throatchop": 1, "healblock": 1};
+				const onBeforeMoveHandlers = this.findPokemonEventHandlers(snatchUser, "onBeforeMove")
+					.filter(handler => handlersToRun.hasOwnProperty(handler.effect.id));
+				const handlerResult = onBeforeMoveHandlers.reduce((result, handler) => {
+					if (handler.callback !== undefined) {
+						return handler.callback.bind(this)(snatchUser, null, move) && result;
```
AnnikaCodes, line 15970: I don't know what the purpose of `findPokemonEventHandlers` is, so I don't know how best to refactor this to use arrow functions or avoid the use of `this`, but our [contributing guidelines](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#ES5-and-ES6) advise against using `Function#bind` for performance reasons.

Slayer95:
So it's only Snatch that is affected by Throat Chop / Heal Block?
i.e. Are sound moves called by e.g. muted Metronome users supposed to work?

data/moves.ts
Slayer95: Probably the best way to implement this would be through a new flavor of ``Battle#runEvent()`` which passes a list of filtered ``effect.id``s to ``findPokemonEventHandlers``.

Slayer95:
Right. Snatch sets ``move.sourceEffect === 'snatch'``, right? In that case it should be possible to implement ``onTryMove`` handlers for ``Heal Block`` and ``Throat Chop`` that make moves fail.

data/moves.ts
MacChaeger: Usually in cases like these we would just make a new event, right?

Slayer95: Also

Slayer95: Also, or simply hardcode the restrictions.

urkerab:
> So it's only Snatch that is affected by Throat Chop / Heal Block?
i.e. Are sound moves called by e.g. muted Metronome users supposed to work? In that case it should be possible to implement `onTryMove` handlers for Heal Block and Throat Chop that make moves fail.

I'd hope not, in which case adding handlers for Heal Block and Throat Chop (and probably also Gravity) would be the way to go. (Bulbapedia's description of Heal Block says, "Restricted moves can still be called by moves that call other moves (such as Metronome), but they are prevented from being used.")

The choice of handler depends on when exactly the failure point for called moves is; I'm not sure I've entirely understood @DaWoblefet's documents on the subject, but my best bet is that it belongs in `onModifyMove` handlers.

aviettran:
Updated this PR per @Slayer95's recommendations to use onTryMove.

DaWoblefet:
Snatch is pretty weird and you all are asking the right questions!

- All submoves (Sleep Talk, Metronome) run Heal Block -> Throat Chop -> Gravity -> Sky Battle checks for their moves. So a Metronome Recover is blocked by Heal Block, and a Sleep Talk Roar is blocked by Throat Chop. To clarify, the called moves themselves are what's stopped.
- For Snatch to execute, it first checks if the move is Snatchable, then if the attack is a Z-move (it'll pass over Z-moves). It makes sure the target isn't in Sky Drop after that. Next, Snatch checks for Heal Block, then Throat Chop, but not Gravity (relevant only for the move Magnet Rise, but impossible to come up anyway bc the Magnet Rise would fail from Gravity before it reaches Snatch in the first place). Snatch then completes, and it begins to execute the called move in the targeting step I think (minimally, it doesn't re-run Attract, and you can fail things like +3 Stockpile or full HP Rest).
-- The tl;dr of that is Heal Block and Throat Chop's checks are hardcoded into Snatch. They don't use the traditional way of doing it.

@SadisticMystic should probably fact check me on Snatch.

urkerab:
> Snatch is pretty weird and you all are asking the right questions!

The impression is that the code wasn't so much written as congealed. (This is typical behaviour of commercial code; on the other hand, Zarel likes to refactor, which tends to keep the code fresher.)

> - All submoves (Sleep Talk, Metronome) run Heal Block -> Throat Chop -> Gravity -> Sky Battle checks for their moves. So a Metronome Recover is blocked by Heal Block, and a Sleep Talk Roar is blocked by Throat Chop. To clarify, the called moves themselves are what's stopped.

Right, so ideally this new code shouldn't be special-casing Snatch?

> - For Snatch to execute, it first checks if the move is Snatchable, then if the attack is a Z-move (it'll pass over Z-moves). It makes sure the target isn't in Sky Drop after that. Next, Snatch checks for Heal Block, then Throat Chop, but not Gravity (relevant only for the move Magnet Rise, but impossible to come up anyway bc the Magnet Rise would fail from Gravity before it reaches Snatch in the first place). Snatch then completes, and it begins to execute the called move in the targeting step I think (minimally, it doesn't re-run Attract, and you can fail things like +3 Stockpile or full HP Rest). However, Protean doesn't run again, so something like Snatch -> take Rest leaves you Dark-type after all's said and done, not Psychic-type.

Is that Protean interaction specific to Snatch? As for the sim's coding, Attract already runs in the BeforeMove event group, which includes things like being frozen, so it won't affect submoves.

>   -- The tl;dr of that is Heal Block and Throat Chop's checks are hardcoded into Snatch. They don't use the traditional way of doing it. However, I can't think of any reason why it wouldn't be ok to use the traditional system for PS's sake.

I guess for Snatch it's impossible to distinguish whether the move fails before or after the targeting step, as only self-targeting moves can be snatched. How about other submoves (Sleep Talk, Metronome etc.)?

DaWoblefet:
> Right, so ideally this new code shouldn't be special-casing Snatch?

If you mean "should Snatch call Heal Block -> Throat Chop as well" then yeah. The missing Gravity and Sky Battle checks are irrelevant.

> Is that Protean interaction specific to Snatch? As for the sim's coding, Attract already runs in the BeforeMove event group, which includes things like being frozen, so it won't affect submoves.

Yes, the Protean interaction is specific to Snatch, but I'm not sure why. I think it's because Snatch doesn't announce "[Pokemon] used [move]!" like Sleep Talk does. It just says ["[user] snatched target]'s move!"](https://youtu.be/ekVbPPaIYZA?t=85) after the soon-to-be-snatched Pokemon uses something like Tailwind. To be fair, though, it could also be because submoves are run in a different area than Snatch. Submove checks like Sleep Talk / Metronome are run way earlier than Protean, and Snatch is run after the user would change typing from Protean, so maybe that has something to do with it. To clarify further, something like Nasty Plot getting stolen by Snatch would still cause the original Nasty Plot user to change typing from Protean.

> I guess for Snatch it's impossible to distinguish whether the move fails before or after the targeting step, as only self-targeting moves can be snatched. How about other submoves (Sleep Talk, Metronome etc.)?

Oh yeah, that's easy - Sleep Talk Thunderbolt redirects into Follow Me or LightningRod, so it's known by the targeting step. Snatch has to know how to target _somehow_ but like you say, it's hard to figure out empirically.

Also I must say @aviettran, I appreciate you getting into the weeds with these niche submove interactions :D

MacChaeger:
Can Curse be snatched by a ghost type if a non-ghost type tries to use it?

urkerab:
>> Right, so ideally this new code shouldn't be special-casing Snatch?
>
> If you mean "should Snatch call Heal Block -> Throat Chop as well" then yeah. The missing Gravity and Sky Battle checks are irrelevant.

No, I meant should Heal Block and Throat Chop (and Gravity) checks apply to other submoves.

>> Is that Protean interaction specific to Snatch?
>
> Yes, the Protean interaction is specific to Snatch, but I'm not sure why. Submove checks like Sleep Talk / Metronome are run way earlier than Protean, and Snatch is run after the user would change typing from Protean, so maybe that has something to do with it. To clarify further, something like Nasty Plot getting stolen by Snatch would still cause the original Nasty Plot user to change typing from Protean.

Interesting... if I can code that up I'll write a PR for it. (Currently the sim has the Snatch user getting the Protean type change on the snatched move.)

> Can Curse be snatched by a ghost type if a non-ghost type tries to use it?

I don't think Curse has the `snatch` flag, does it?

Slayer95:
According to my reading of the discussion, the actionable steps would be:
- Expand the overly-specific checks in the ``Heal Block`` and ``Throat Chop`` volatiles to cover every caller move.
- Ensure that Gravity blocks Bounce, Fly, Flying Press, High Jump Kick, Jump Kick, Magnet Rise, Sky Drop, Splash, and Telekinesis when those moves are called by Sleep Talk, etc.
- Follow-up with improvements to Protean+Snatch.

DaWoblefet:
> Can Curse be snatched by a ghost type if a non-ghost type tries to use it?

They probably thought of this; Curse can't be Snatched at all.

> No, I meant should Heal Block and Throat Chop (and Gravity) checks apply to other submoves.

Right, they should. If anything, Snatch is the weird one :P

> if you are Tricked a choice item in between using Snatch and snatching a move in your moveset, you're not actually choiced yet, are you?

Right, that doesn't Choice lock you. 

@Slayer95 that looks to be a good summary, yeah.

aviettran:
@Slayer95 can you comment on the newly drafted approach using a new volatile? If it looks okay, I will add an onTryMove to Gravity.

@DaWoblefet is there an exhaustive list of "caller moves" that are applicable? I can blindly add the volatile to each, but clearly things like Nature Power are not applicable (but could be in the future?).

urkerab:
I think he meant for you to remove the check for Snatch from your new events for the existing volatiles for Heal Block and Throat Chop, as these checks should apply to all caller moves, not just Snatch.

DaWoblefet:
> @DaWoblefet is there an exhaustive list of "caller moves" that are applicable? I can blindly add the volatile to each, but clearly things like Nature Power are not applicable (but could be in the future?).

Assist, Copycat, Me First, Metronome, Mirror Move, Nature Power, and Sleep Talk are all moves that call other moves, what I like to call "submoves". They all do the Heal Block -> Throat Chop -> Gravity -> Sky Battle checks discussed. In fact, submoves are the only way to bring up the Sky Battle failure message as far as I know! Not like we support Sky Battles but ya know :P

Slayer95:
> I think he meant for you to remove the check for Snatch from your new events for the existing volatiles for Heal Block and Throat Chop, as these checks should apply to all caller moves, not just Snatch.

Yeah, and by extension there is no need to create a new volatile for all those moves as far as I can tell.

MacChaeger:
> Assist, Copycat, Me First, Metronome, Mirror Move, Nature Power, and Sleep Talk are all moves that call other moves, what I like to call "submoves". They all do the Heal Block -> Throat Chop -> Gravity -> Sky Battle checks discussed. In fact, submoves are the only way to bring up the Sky Battle failure message as far as I know! Not like we support Sky Battles but ya know :P

I might be misunderstanding you, but [the "official" convention is actually to call the moves called by those moves "submoves".](https://dex.pokemonshowdown.com/articles/submoves)

aviettran:
1. Removed the volatile
2. Added onTryMove to Gravity
3. Added Sleep Talk + Gravity Test

Slayer95:
>but my best bet is that it belongs in onModifyMove handlers.

It's not clear to me whether @DaWoblefet implied a positive or negative answer to this. Uhm. @Marty-D ?

> Conflicting files
test/sim/moves/snatch.js

You should rebase the PR on ``master``, since 6daa05781b already got merged.

Marty-D:
> 
> 
> > but my best bet is that it belongs in onModifyMove handlers.
> 
> It's not clear to me whether @DaWoblefet implied a positive or negative answer to this. Uhm. @Marty-D ?

Based on https://www.smogon.com/forums/posts/8096570/ I would agree with urkerab, since calling moves redo the earlier steps prior to Stance Change activating, and it's currently using `ModifyMove` (in Gen 7+).

aviettran:
@Marty-D I am a bit lost. Should this change be utilizing onModifyMove instead of onTryMove?

Marty-D:
I think so, to be accurate with the order of other effects that are currently implemented. I don't know of any existing usage of ModifyMove that ends processing of a move though; these might be the first cases.

aviettran:
@Marty-D complete. I also confirmed that the move exits if the ModifyMove fails.

Marty-D:
Perfect, thanks for dealing with all these revisions! 👍

PR #7761, opened 2020-12-02
Disable <youtube> tag for groupchats.

sirKiraUzumaki:


Zarel:
This seems like a policy decision, to me. Was this requested by staff? Can you describe the reasoning behind it?

mia-pi-git:
This wasn't requested. As i said in Staff room, there's no real problem with allowing this in groupchats - only a global bot could use it in groupchats anyway, and we trust them to not abuse it. There's also the fact that this is easily bypassed by just using !show.
So i don't really see the point here.

AnnikaCodes:
Seconding Zarel and Mia. In general, you should get approval from someone for these types of policy-related changes.

Zarel:
I've asked Staff, and consensus is that it seems kind of pointless considering `/show` exists. I'm going to close this because it's probably not going to be approved, but I'm still interested in hearing why you want this. Has this been abused in a groupchat you've been in?

Also, don't let the rejection get you down. I still appreciate the effort! In the future, I recommend discussing policy changes like this in the Dev chatroom or Discord so you don't spend too much effort on something we won't accept.

sirKiraUzumaki:
> This wasn't requested. As i said in Staff room, there's no real problem with allowing this in groupchats - only a global bot could use it in groupchats anyway, and we trust them to not abuse it. There's also the fact that this is easily bypassed by just using !show.
> So i don't really see the point here.

Groupchat's host can also embed the video easily by using `!htmlbox <youtube src="VideoLink"></youtube>` but I think that most of the users are not aware of this youtube tag.

I was not aware of the piracy-related change, so pardon me for that. But it is up to you guys to decide to allow it or not.

PR #7767, opened 2020-12-04
Youtube: Track current video time in groupwatch

mia-pi-git:


Zarel:
You can track a start time but please don't add another timer.

mia-pi-git:
> You can track a start time but please don't add another timer.

Updated.

server/chat-plugins/youtube.ts
```diff
@@ -308,7 +308,7 @@ export class YoutubeInterface {
 export class GroupWatch extends Rooms.RoomGame {
 	url: string;
 	info: VideoData;
-	started = false;
+	started: false | number = false;
```
Zarel, line 311:
```suggestion
	started: number | null = null;
```

PR #7768, opened 2020-12-06
Protean should affect the victim of Snatch rather than the user

urkerab:
According to the move checks in one of the research threads (and also repeated by @DaWoblefet in PR https://github.com/smogon/pokemon-showdown/pull/7758#issuecomment-738547967), Snatch actually happens just after Protean, so I moved its event to onAnyPrepareHit at lower priority, with fixups so that the `PrepareHit` event can actually cause the move to stop executing. I then made Protean/Libero not run on the Snatched move. After checking with researchers I also ensured that choice locking doesn't happen for Snatched moves (you could be Tricked a choice item in between using Snatch and snatching a move). (I don't know whether it would be worth refactoring `hasBounced` into new property which covers both bounced and snatched moves.)

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -1780,7 +1780,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	},
 	libero: {
 		onPrepareHit(source, target, move) {
-			if (move.hasBounced) return;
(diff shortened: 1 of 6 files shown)


TheImmortal:
Looks fine but @DaWoblefet want to add the relevant tests?

DaWoblefet:
Should be good to go.

PR #7777, opened 2020-12-08
Prevent /tour forcepublic spamming rooms

KrisXV:
There's currently no check for if a tournament is already being forced public, which I noticed when testing the feature on my bot.

> [06:01:47] (Tournament public battles were turned ON by Unown)
Tournament battles forced public: ON
[06:01:47] (Tournament public battles were turned ON by Unown)
Tournament battles forced public: ON
[06:01:47] (The tournament auto disqualify timer was set to 2 by Unown)
The tournament's automatic disqualify timer has been set to 2 minutes.
[06:01:48] (The tournament auto start timer was set to 3 by Unown)
The tournament will automatically start in 3 minutes.
[06:01:48] (Tournament public battles were turned ON by Unown)
Tournament battles forced public: ON

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -1782,13 +1782,19 @@ const commands: ChatCommands = {
 			target = target.trim();
 			const option = target || 'on';
 			if (this.meansYes(option)) {
+				if (tournament.forcePublic) {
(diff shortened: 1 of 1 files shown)


Zarel:
I preferred the previous success messages, tbh. Much clearer.

PR #7782, opened 2020-12-09
Ease restrictions in Monotype Random Battle

MacChaeger:


TheImmortal:
Why?

PartMan7:
https://cdn.discordapp.com/attachments/762324232948023316/786289094518112256/G5MR.png

TheImmortal:
Unnecessary changes for a straightforward fix.

MacChaeger:
So you’re just gonna ignore that you’re rejecting Arceus 16 times out of 17 in gen 5 mono rands for no reason? Or that the crash reports just report that the team didn’t have enough Pokémon and don’t actually report what Pokémon it did have, which most of the time is the only thing you would need the seed for anyway?

TheImmortal:
> So you’re just gonna ignore that you’re rejecting Arceus 16 times out of 17 in gen 5 mono rands for no reason?

No that’s a good catch. I’ll do it in another commit.

data/mods/gen3/random-teams.ts
```diff
@@ -572,7 +572,7 @@ export class RandomGen3Teams extends RandomGen4Teams {
 			// To prevent this, we prevent more than one Wobbuffet in a single battle.
 			if (set.ability === 'Shadow Tag') this.hasWobbuffet = true;
 		}
-		if (pokemon.length < 6 && !isMonotype) throw new Error(`Could not build a random team for ${this.format} (seed=${seed})`);
+		if (pokemon.length < 6 && !isMonotype) throw new Error(`Could not build a random team for ${this.format} (seed=${seed}) (team:${pokemon.map((mon) => mon.name)})`);
```
Zarel, line 575: The reason it doesn't show the team is because you can just repro with `./pokemon-showdown generate-team`. I don't think this adds anything.

MacChaeger: If it tells you the team right away you don’t even need to reproduce it, and that should save time.

PR #7785, opened 2020-12-10
Bug fix/heal block independent fail

aviettran:


aviettran:
Haven't tested this in client, since Heal Block isn't valid in any of the formats. What's a good way to test?

DaWoblefet:
Hop on localhost and battle in something like Gen 7 Custom Double Battle.

aviettran:
Looks like Gen 7 Doubles OU works, thanks. Not sure quite sure what the proper method of displaying custom text for a move failure is. If someone can give me a quick primer it'd save me some time.

DaWoblefet:
I would also like to know this fwiw.

aviettran:
I figured it out. An educational experience. The client has the sim as a dependency, so I pulled the client repo and did a full build with the changes from this PR. I was then able to do a full test. Moving this PR out of draft.

image

Marty-D (approved):
Nicely done once again, thanks! If there's a card associated with this I'll go move it over now.

PR #7786, opened 2020-12-10
Chat monitor: Improve the handling of usernames

AnnikaCodes:
This PR ignores word boundaries in evasion regexes when filtering usernames, preventing the use of usernames with evasions of slurs in them.

I also added a unit test for this behavior, to avoid regressions.


AnnikaCodes:
@Zarel  bump

Zarel (approved):
Looks good! You can merge this without further review once you handle the merge conflict.

PR #7787, opened 2020-12-10
Use BigInt for PRNG seed

xfix:
This makes `nextFrame` method so much simpler.

### DIFF
--- a/sim/prng.ts
+++ b/sim/prng.ts
@@ -130,76 +130,24 @@ export class PRNG {
 	 * c = 0x00269EC3
 	 * m = 2^64
 	 * ````
-	 *
-	 * Javascript doesnt handle such large numbers properly, so this function does it in 16-bit parts.
-	 * ````
-	 * x_{n + 1} = (x_n * a) + c
-	 * ````
-	 *
-	 * Let any 64 bit number:
-	 * ````
-	 * n = (n[0] << 48) + (n[1] << 32) + (n[2] << 16) + n[3]
-	 * ````
-	 *
-	 * Then:
-	 * ````
-	 * x_{n + 1} =
-	 * 	((a[3] x_n[0] + a[2] x_n[1] + a[1] x_n[2] + a[0] x_n[3] + c[0]) << 48) +
-	 * 	((a[3] x_n[1] + a[2] x_n[2] + a[1] x_n[3] + c[1]) << 32) +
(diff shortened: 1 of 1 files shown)


Zarel:
I would prefer not using this specific modern feature because it makes interop more complicated.

Zarel:
I... I see benefits to this. I do agree it makes the code simpler. I'm not a fan of how you seem to have dropped support for the old array format.

I think this is _marginally_ useful, but I don't think it's anywhere near useful enough to be worth dropping backwards compatibility for.

Zarel:
Heck, I'm not even sure it's worth complicating `state.ts`.

xfix:
Yeah, fair, reduced the scope of changes to `nextFrame`, hopefully that's acceptable.

xfix:
Or I guess I can use `reverse()`.

Zarel:
I... I guess I'm still not sure if I think this is worth dropping support for older JavaScript engines. Like, I'm aware that we already dropped support for Node versions not supporting BigInt, but it might be nice to keep the old one around.

The conversion between `PRNGSeed` and bigint is also kind of subtle, I don't know if it's worth requiring people to know `BigInt64Array` and the subtleties of big/little-endianness.

If we were concerned about readability, I would refactor it this way:

```ts
/** calculates `a * b + c` */
multiplyAdd(a: PRNGSeed, b: PRNGSeed, c: PRNGSeed) {
	const out: PRNGSeed = [0, 0, 0, 0];
	let carry = 0;

	for (let outIndex = 3; outIndex >= 0; outIndex--) {
		for (let bIndex = outIndex; bIndex < 4; bIndex++) {
			const aIndex = 3 - (bIndex - outIndex);

			carry += a[aIndex] * b[bIndex];
		}
		carry += c[outIndex];

		out[outIndex] = carry & 0xFFFF;
		carry >>>= 16;
	}

	return out;
}

nextFrame(seed: PRNGSeed, framesToAdvance = 1): PRNGSeed {
	const a: PRNGSeed = [0x5D58, 0x8B65, 0x6C07, 0x8965];
	const c: PRNGSeed = [0, 0, 0x26, 0x9EC3];

	for (int i = 0; i < framesToAdvance; i++) {
		seed = this.multiplyAdd(seed, a, c);
	}

	return seed;
}
```

PR #7789, opened 2020-12-11
Fix Autoresponder Refresh button bug

tmagicturtle:
Currently if you go into `/autoresponder view` and choose to view autoresponder Stats, it opens a menu where no date is specified and you can choose a date from there. If, however, you press the Refresh button (where no date is specified), it opens a new page called "view-autoresponder-room-stats-" (note the extra hyphen at the end). This is because the "refresh" function is called (on line 458) with [date], which would equal [""]. This is a truthy value, unlike [] and unlike "". This patch changes the conditional statement to look at the first item in the array's truthfulness - it is either an empty string (falsey) or a non-empty string (truthy) - instead of always being truthy.

### DIFF
--- a/server/chat-plugins/responder.ts
+++ b/server/chat-plugins/responder.ts
@@ -440,16 +440,13 @@ export const pages: PageTable = {
 		const canChange = user.can('ban', null, room);
 		let buf = '';
 		const refresh = (type: string, extra?: string[]) => {
+			if (extra) extra = extra.filter(Boolean);
 			let button = `<button class="button" name="send" value="/join view-autoresponder-${room.roomid}-${type}`;
-			if (extra) {
-				button += `${extra.length ? `-${extra.join('-')}` : ''}" style="float: right">`;
-			} else {
-				button += `" style="float: right">`;
-			}
+			button += `${extra?.length ? `-${extra.join('-')}` : ''}" style="float: right">`;
 			button += `<i class="fa fa-refresh"></i> Refresh</button><br />`;
 			return button;
 		};
-		const back = `<br /><a roomid="view-autoresponder-${room.roomid}-">Back to all</a>`;
+		const back = `<br /><a roomid="view-autoresponder-${room.roomid}">Back to all</a>`;
 		switch (args[0]) {
 		case 'stats':
 			args.shift();
--- a/server/chat-plugins/responder.ts
+++ b/server/chat-plugins/responder.ts
@@ -456,7 +456,7 @@ export const pages: PageTable = {
 				return `<h2>Invalid date.</h2>`;
 			}
 			buf = `<div class="pad"><strong>Stats for the ${room.title} auto-response filter${date ? ` on ${date}` : ''}.</strong>`;
-			buf += `${back}${refresh('stats', (date ? [date] : []))}<hr />`;
+			buf += `${back}${refresh('stats', [date])}<hr />`;
 			const stats = roomData.stats;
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
This change seems fine, although I'm not familiar with the autoresponder plugin.
That being said, `[]` is not a truthy value :)

server/chat-plugins/responder.ts
```diff
@@ -441,7 +441,11 @@ export const pages: PageTable = {
 		let buf = '';
 		const refresh = (type: string, extra?: string[]) => {
 			let button = `<button class="button" name="send" value="/join view-autoresponder-${room.roomid}-${type}`;
-			button += `${extra ? `-${extra.join('-')}` : ''}" style="float: right">`;
+			if (extra) {
```
mia-pi-git, line 444: This might be better suited by running a `.filter(Boolean)` on the `extra` arg, if it exists, and then just checking .length?

server/chat-plugins/responder.ts
```diff
@@ -441,7 +441,11 @@ export const pages: PageTable = {
...
+				button += `${extra.length ? `-${extra.join('-')}` : ''}" style="float: right">`;
+			} else {
+				button += `" style="float: right">`;
+			}
```
AnnikaCodes, line 448:
```suggestion
			if (extra?.length) button += `-${extra.join('-')}`;
			button += `" style="float: right">`;
```
I agree with Mia's suggestion on the need for `.filter(Boolean)`; an empty array or `[""]` would still cause the extra hyphen to be appended.

server/chat-plugins/responder.ts
mia-pi-git:
```ts
const refresh = (type: string, extra?: string[]) => {
			if (extra) extra = extra.filter(Boolean);
			let button = `<button class="button" name="send" value="/join view-autoresponder-${room.roomid}-${type}`;
			button += `${extra?.length ? `-${extra.join('-')}` : ''}" style="float: right">`;
			button += `<i class="fa fa-refresh"></i> Refresh</button><br />`;
			return button;
		};
		const back = `<br /><a roomid="view-autoresponder-${room.roomid}">Back to all</a>`;```

File: server/chat-plugins/responder.ts (hunk 2 above)
Comment (mia-pi-git) [reply server/chat-plugins/responder.ts:448]: (See my most recent code block comment, should cover this)

Comment (AnnikaCodes) [reply server/chat-plugins/responder.ts:448]: Although it looks like your code on line 462 prevents those arrays from being passed in, unless `refresh()` is called elsewhere.

File: server/chat-plugins/responder.ts [hunk 3]
```diff
@@ -455,7 +456,7 @@ export const pages: PageTable = {
 				return `<h2>Invalid date.</h2>`;
 			}
 			buf = `<div class="pad"><strong>Stats for the ${room.title} auto-response filter${date ? ` on ${date}` : ''}.</strong>`;
-			buf += `${back}${refresh('stats', [date])}<hr />`;
+			buf += `${back}${refresh('stats', (date ? [date] : []))}<hr />`;
```
AnnikaCodes, line 459: is this necessary, since you are already filtering the array later

tmagicturtle: Undone. Thank you!

server/chat-plugins/responder.ts
tmagicturtle: Done. Thank you!

mia-pi-git (approved):
Looking good.

PR #7791, opened 2020-12-11
Metronome Battle: Ban Steel Memory

MacChaeger:
Steel types are all supposed to be banned from Metronome Battle, but a Silvally other than Silvally-Steel could hold a Steel Memory with RKS System to be Steel type while fooling the validator into thinking it's not.


MacChaeger:
Not sure who runs Metronome, but they should review this.

VineST:
This has been a known issue in Metronome; we never fixed it because Silvally-Steel without the item slot is nothing to write home about. Of course, plenty of worse steels get banned by the blanket ban, like all NFEs, so for consistency's sake, this is a worthwhile patch.

xfix:
This issue is harmless I feel like, but if we ban trash Steel-type mons we may as well get this merged. Leaving the decision to merge to @TheImmortal.

xfix (approved):

PR #7792, opened 2020-12-11
Chat: Support specifying priority for chatfilters

AnnikaCodes:
As per [this discussion](https://discord.com/channels/534298043176976395/581910956167266324/7841468372539146750).

### DIFF
--- a/server/chat-plugins/chat-monitor.ts
+++ b/server/chat-plugins/chat-monitor.ts
@@ -320,58 +320,55 @@ void FS(MONITOR_FILE).readIfExists().then(data => {
 
 /* The sucrase transformation of optional chaining is too expensive to be used in a hot function like this. */
 /* eslint-disable @typescript-eslint/prefer-optional-chain */
-export const chatfilter: ChatFilter = {
-	priority: 0,
-	handler(message, user, room) {
-		let lcMessage = message
-			.replace(/\u039d/g, 'N').toLowerCase()
-			// eslint-disable-next-line no-misleading-character-class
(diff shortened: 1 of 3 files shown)


AnnikaCodes:
bump

AnnikaCodes:
@Zarel bump (the test failures are unrelated)

AnnikaCodes:
bump

server/chat-plugins/chat-monitor.ts
```diff
@@ -320,55 +320,58 @@ void FS(MONITOR_FILE).readIfExists().then(data => {
 
 /* The sucrase transformation of optional chaining is too expensive to be used in a hot function like this. */
 /* eslint-disable @typescript-eslint/prefer-optional-chain */
-export const chatfilter: ChatFilter = function (message, user, room) {
```
Zarel, line 323: One thing that would be interesting here (and also not change indentation level) would be:

```ts
export const chatfilter: ChatFilter = function (message, user, room) {
  ...
};
chatfilter.priority = 0;
```

Which can be left off for most chatfilters at default priority.

Zarel: I would leave it up to you, but I would definitely strongly prefer that approach.

AnnikaCodes: That's definitely simpler.

AnnikaCodes: I'll change to that

Zarel:
Thanks!

PR #7795, opened 2020-12-12
Bug fix: Future Move Failure Message

aviettran:
Regarding the card:

> Moves that call other moves + Future Sight / Doom Desire cause the target to be itself? Also it seems to be deterministic and always targets opponent's left slot? See replay - DW

I also tested the determinism with a few seeds and it is indeed random.


DaWoblefet:
Hahaha, I can't believe the error was just in the string and not the mechanic, that's hilarious

PR #7797, opened 2020-12-13
Support generating sourceMaps for source debugging

aviettran:
**Justification**

Sucrase, by default, does not generate source maps. In order for IDEs (in my case, vscode) to allow for breakpoints on source files (\*.ts), [source maps must exist](https://code.visualstudio.com/docs/nodejs/nodejs-debugging). The only alternative, which is quite tedious, is to find the built Javascript (\*-dist/\*\*/\*) corresponding to the source file (\*.ts) and to create the corresponding breakpoint there.

**Proposed Solution**

Sucrase does not allow its cli tool to generate source maps (https://github.com/alangpierce/sucrase/issues/223), nor does it seem to be a priority. It must be done programmatically. The proposed solution is to convert the cli-based transcoding process into a programmatic one ~~which, optionally,~~ generates source maps.

~~**Proposed Configuration**
Use the node module dotenv to allow the developer to create a .env file in order to set the environment variable "NODE_ENV" to "development". This will enable the generation of sourceMaps. This was just a standard and quick way for me to get this PR up and running. I am very open to alternative suggestions.~~

**Proposed Build-Artifacts**

For each output JavaScript file (\*.js), generate a corresponding \*.js.map file and place it in a new directory called "sourceMaps".

**Example:**

**Source:**
data/moves.ts

**Output:**
.data-dist/moves.js
.data-dist/sourceMaps/moves.js.map

In addition to these new \*.map files, each output JavaScript file will have a newly appended comment at the end of the file:
 //# sourceMappingURL=sourceMaps/moves.js.map

~~CI pipelines and production builds _should not_ have the environment "NODE_ENV" set to "development" and will, therefor, not have any change in behavior as a result of this PR.~~

### DIFF
--- a/.eslintrc-no-types.json
+++ b/.eslintrc-no-types.json
@@ -307,6 +307,13 @@
         "mocha": true
       },
       "files": ["test/**/*.js"]
+    },
+    {
+      "files": ["build"],
+      "rules": {
+        "strict": "off",
+        "no-var": "off"
+      }
     }
   ]
 }
--- a/.eslintrc-no-types.json
+++ b/.eslintrc-no-types.json
@@ -311,7 +311,6 @@
     {
       "files": ["build"],
       "rules": {
-        "strict": "off",
         "no-var": "off"
       }
     }
--- a/.gitignore
+++ b/.gitignore
@@ -16,10 +16,12 @@ package-lock.json
 databases/*.db*
 *.o
 *.o.dSYM
+.env
 # Typescript build artifacts
 .*-dist/
 tools/set-import/importer.js
 tools/modlog/converter.js
+**/sourceMaps
 
 # visual studio live share
 .vs
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ package-lock.json
 databases/*.db*
 *.o
 *.o.dSYM
-.env
 # Typescript build artifacts
 .*-dist/
 tools/set-import/importer.js
--- a/build
+++ b/build
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+"use strict";
 
 try {
 	// technically this was introduced in Node 11, but we'll ask for the most recent LTS with it to be safe
@@ -8,10 +9,10 @@ try {
 	process.exit(1);
 }
 
-var child_process = require('child_process');
-var fs = require('fs');
-var path = require('path');
-var force = process.argv[2] === '--force';
+const child_process = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const force = process.argv[2] === '--force';
 
 process.chdir(__dirname);
 
@@ -26,21 +27,21 @@ function replace(file, replacements) {
 		if (stats.isFile()) {
 			if (!file.endsWith('.js')) return;
 			fs.readFile(file, "utf-8", function (err, text) {
+				if (err) throw err;
+				let anyMatch = false;
+				for (let i = 0; i < replacements.length; i++) {
+					anyMatch = anyMatch || text.match(replacements[i].regex);
+					if (anyMatch) text = text.replace(replacements[i].regex, replacements[i].replace);
+				}
+				if (!anyMatch) return;
+				fs.writeFile(file, text, function (err) {
 					if (err) throw err;
-					var anyMatch = false;
-					for (var i = 0; i < replacements.length; i++) {
-						anyMatch = anyMatch || text.match(replacements[i].regex);
-						if (anyMatch) text = text.replace(replacements[i].regex, replacements[i].replace);
-					}
-					if (!anyMatch) return;
-					fs.writeFile(file, text, function (err) {
-						if (err) throw err;
-					});
 				});
+			});
 		} else if (stats.isDirectory()) {
 			fs.readdir(file, function (err, files) {
 				if (err) throw err;
-				for (var i = 0; i < files.length; i++) {
+				for (let i = 0; i < files.length; i++) {
 					replace(path.join(file, files[i]), replacements);
 				}
 			});
@@ -48,8 +49,8 @@ function replace(file, replacements) {
 	});
 }
 function copyOverDataJSON(file) {
-	var source = './data/' + file;
-	var dest = './.data-dist/' + file;
+	const source = './data/' + file;
+	const dest = './.data-dist/' + file;
 	fs.readFile(source, function (err, text) {
 		if (err) throw err;
 		fs.writeFile(dest, text, function (err) {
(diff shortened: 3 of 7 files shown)


build
```diff
@@ -8,10 +9,13 @@ try {
 	process.exit(1);
 }
 
-var child_process = require('child_process');
-var fs = require('fs');
-var path = require('path');
-var force = process.argv[2] === '--force';
```
AnnikaCodes, line 14: IIRC these are intended to be `var`, not `const`, so that really old node versions will reach a point where a more descriptive error can be thrown.

AnnikaCodes: I don't really think that decision still makes sense several years later, but that's a separate change.

Zarel: Yeah, we should probably put the explanation in comments at the top of the file, so this doesn't keep happening.

Zarel:
How do source maps impact performance? If it's light enough, we might prefer to always enable source maps. In general, we prefer not to have differences between development and production builds (so that debugging issues in prod isn't harder than it needs to be).

Zarel:
(That would also sidestep the `dotenv` dependency which I'm not a big fan of adding.)

build
aviettran: Noted.

aviettran:
Build times are not significantly impacted:

aviettran:~/pokemon-showdown$ time ./build --force

real    0m6.779s
user    0m6.031s
sys     0m1.375s
aviettran:~/pokemon-showdown$ export NODE_ENV=development
aviettran:~/pokemon-showdown$ time ./build --force

real    0m7.071s
user    0m5.922s
sys     0m1.500s

Zarel:
Indeed. Would you be willing to remove all the env stuff, then? Just build source maps all the time?

.gitignore
```diff
@@ -16,10 +16,12 @@ package-lock.json
 databases/*.db*
 *.o
 *.o.dSYM
+.env
 # Typescript build artifacts
 .*-dist/
 tools/set-import/importer.js
 tools/modlog/converter.js
+**/sourceMaps
```
urkerab, line 24: Isn't this already covered by the `.*-dist` exclusion?

aviettran: `tools/set-import/importer.js` and `converter.js` are listed explicitly, so rather than repeat them, I just added a rule for **/sourceMaps.

build
aviettran: I reverted any eslint auto-fixed changes and added exceptions to `.eslintrc-no-types.json`.

aviettran:
@Zarel I removed anything conditional from the `sourceMap` generation logic and removed `dotenv`. I also removed the `sourceRoot` property from `*.map`, since it can hypothetically get deployed somewhere other than where it was built. The relative directory structure will have to stay intact for maps to work, though.

.eslintrc-no-types.json
```diff
@@ -307,6 +307,13 @@
         "mocha": true
       },
       "files": ["test/**/*.js"]
+    },
+    {
+      "files": ["build"],
+      "rules": {
+        "strict": "off",
```
Zarel, line 314: `"use strict"` is fine, though, right? No compatibility issues there.

build
```diff
@@ -8,6 +8,7 @@ try {
 	process.exit(1);
 }
 
+var {transform} = require('sucrase');
```
Zarel, line 11:
```suggestion
var transform = require('sucrase').transform;
```

(Destructuring is also not supported by early Node versions)

aviettran: Ok. What is the minimum version we are supporting? Let me know so I can test.

build
```diff
@@ -101,13 +102,74 @@ function needsSucrase(source, dest, path = '') {
...
+			['node_modules', '.git'].includes(child) ||
+			options.excludeDirs.includes(child)
+		) {
+			continue;
+		}
+		const srcChildPath = path.join(srcDirPath, child);
+		const outChildPath = path.join(outDirPath, child);
+		const pathToSourceChild = path.join(pathToSource, '..');
+		if (fs.statSync(srcChildPath).isDirectory()) {
+			const innerOptions = {...options};
```
Zarel, line 131:
```suggestion
			const innerOptions = Object.assign({}, options);
```

`Object.assign` also postdates early Node, but it's not syntactical so we'd hit the error before hitting this.

Honestly, if we want to write modern code here, we might want to put it in a different file wrapping sucrase, rather than in this file with a lot of restrictions on what syntax is allowed.

build
```diff
@@ -101,13 +102,74 @@ function needsSucrase(source, dest, path = '') {
...
+			innerOptions.srcDirPath = srcChildPath;
+			innerOptions.outDirPath = outChildPath;
+			innerOptions.pathToSource = pathToSourceChild;
+			const innerFiles = findFiles(innerOptions);
+			outArr.push(...innerFiles);
```
Zarel, line 136:
```suggestion
			Array.prototype.push.apply(outArr, innerFiles);
```

build
```diff
@@ -101,13 +102,74 @@ function needsSucrase(source, dest, path = '') {
...
+		} else if (extensions.some((ext) => srcChildPath.endsWith(ext))) {
+			const outPath = outChildPath.replace(
+				/\.\w+$/,
+				`.${options.outExtension}`
```
Zarel, line 140:
```suggestion
				"." + options.outExtension
```

build
```diff
@@ -101,13 +102,74 @@ function needsSucrase(source, dest, path = '') {
...
+		const outChildPath = path.join(outDirPath, child);
+		const pathToSourceChild = path.join(pathToSource, '..');
+		if (fs.statSync(srcChildPath).isDirectory()) {
+			const innerOptions = {...options};
+			innerOptions.srcDirPath = srcChildPath;
+			innerOptions.outDirPath = outChildPath;
+			innerOptions.pathToSource = pathToSourceChild;
+			const innerFiles = findFiles(innerOptions);
+			outArr.push(...innerFiles);
+		} else if (extensions.some((ext) => srcChildPath.endsWith(ext))) {
```
Zarel, line 137:
```suggestion
		} else if (extensions.some(function (ext) { return srcChildPath.endsWith(ext); })) {
```

build
```diff
@@ -101,13 +102,74 @@ function needsSucrase(source, dest, path = '') {
...
 	try {
 		if (!force && src !== './config' && !needsSucrase(src, out)) {
 			return false;
 		}
 	} catch (e) {}
-	shell(`npx sucrase ${opts || ''} -q ${src} -d ${out} --transforms typescript,imports --enable-legacy-typescript-module-interop`);
+	const sucraseOptions = {transforms: ['typescript', 'imports'], enableLegacyTypeScriptModuleInterop: true, ...opts};
+	const files = findFiles({outDirPath: out, srcDirPath: src, sucraseOptions: sucraseOptions, excludeDirs: ['sourceMaps', ...excludeDirs], outExtension: 'js'});
+	for (const file of files) {
+		const sucraseOptionsFile = {...sucraseOptions, sourceMapOptions: {compiledFilename: file.outPath}, filePath: path.join(file.pathToSource, file.srcPath)};
```
Zarel, line 166:
```suggestion
		const sucraseOptionsFile = Object.assign(
			{},
			sucraseOptions,
			{sourceMapOptions: {compiledFilename: file.outPath}, filePath: path.join(file.pathToSource, file.srcPath)}
		);
```

build
aviettran: I took your suggestion and moved a couple functions out. Let me know if it works, or if you want me to put the lib somewhere else.

Zarel:
Looks great! Thanks!

PR #7798, opened 2020-12-13
Fix help for /rps command.

sirKiraUzumaki:
Also rps game is bugged.

### DIFF
--- a/server/chat-plugins/rock-paper-scissors.ts
+++ b/server/chat-plugins/rock-paper-scissors.ts
@@ -347,18 +347,17 @@ export const commands: ChatCommands = {
 			const game = this.requireGame(RPSGame);
 			game.unpause(user);
 		},
-
-		help() {
-			this.runBroadcast();
-			const strings = [
-				`/rockpaperscissors OR /rps<br />`,
-				`/rps create OR /rps challenge [user], [rounds] - Challenges a user to a game of Rock Paper Scissors with [rounds] (or 3) rounds`,				`/rps leave - Leave the game.`,
-				`/rps start - Start the Rock Paper Scissors game.`,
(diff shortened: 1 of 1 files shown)


mia-pi-git:
I'm aware it has issues - I've been rewriting it.
However - what does this PR do, exactly?
You don't seem to make any edits to the help text?

sirKiraUzumaki:
> I'm aware it has issues - I've been rewriting it.
> However - what does this PR do, exactly?
> You don't seem to make any edits to the help text?
Oh great that you are rewriting it.
Currently /h rps didn't work because the help code was in the wrong position.
But now you are rewriting I don't think so this pr is necessary, you can add this later in your pr.

server/chat-plugins/rock-paper-scissors.ts
```diff
@@ -347,18 +347,17 @@ export const commands: ChatCommands = {
...
-				`/rps create OR /rps challenge [user], [rounds] - Challenges a user to a game of Rock Paper Scissors with [rounds] (or 3) rounds`,				`/rps leave - Leave the game.`,
-				`/rps start - Start the Rock Paper Scissors game.`,
-				`/rps end - End the Rock Paper Scissors game`,
-				`/rps pause - Pauses the game, if it's in progress.`,
-				`/rps resume - Resumes the game, if it's paused.`,
-			];
-			return this.sendReplyBox(strings.join('<br />'));
-		},
+	},
+	help() {
```
mia-pi-git, line 351: .....where you're moving this to means you make this the new /help command, rather than the one in `chat-commands/core.ts`. It is in the right position currently.

mia-pi-git:
Closing this since in fact it works as it should atm (this just breaks normal /help).

PR #7799, opened 2020-12-13
Modlog: Remove Rooms global in PM

mia-pi-git:


server/modlog.ts
```diff
@@ -467,6 +469,7 @@ export class Modlog {
 // even though it's a type not a function...
 type ModlogResult = ModlogEntry | undefined;
 
+export const mainModlog = new Modlog(MODLOG_PATH, MODLOG_DB_PATH);
```
Zarel, line 472: Does this API need to be changed?

Zarel: Oh, I see, it's used by Modlog.

Zarel: I would call it `modlog` lowercase `m` unless there's a reason otherwise?

Zarel: I guess it conflicts with other variables called `modlog` in this file? This is fine.

Zarel:
@AnnikaCodes `Rooms` is not designed to be imported outside of the main process, and importing it in Modlog subprocesses causes them to spawn unused `room-battle` sub-subprocesses. That's what this PR fixes. Can you review to make sure it works as you intended?

AnnikaCodes (approved):
These changes look fine. #7620 moves modlog entirely to `Chat`, which I think is a much better solution than `mainModlog` trickery, but if this PR will end up being faster to merge then I think it's a good interim solution. Thanks!

PR #7800, opened 2020-12-13
Trivia: Use an Object for questions instead of a sorted array

AnnikaCodes:
This will make future features (like custom categories) much simpler, and eliminate odd behaviors, like the `ALL_CATEGORIES` object requiring alphabetically-sorted keys.

### DIFF
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -1776,7 +1776,7 @@ const triviaCommands: ChatCommands = {
 		for (const category in submissions) {
 			for (const [i, entry] of submissions[category].entries()) {
 				total++;
-				innerBuffer += `<tr><td><strong>${submissions[category].indexOf(entry)}</strong></td><td>${entry.category}</td><td>${entry.question}</td><td>${entry.answers.join(", ")}</td><td>${entry.user}</td></tr>`;
+				innerBuffer += `<tr><td><strong>${i}</strong></td><td>${entry.category}</td><td>${entry.question}</td><td>${entry.answers.join(", ")}</td><td>${entry.user}</td></tr>`;
 			}
 		}
 		if (!innerBuffer) return this.sendReply(this.tr`No questions await review.`);
--- a/server/chat-plugins/trivia.ts
+++ b/server/chat-plugins/trivia.ts
@@ -1772,17 +1772,17 @@ const triviaCommands: ChatCommands = {
 		if (!submissions) return this.sendReply(this.tr`No questions await review.`);
 
 		let innerBuffer = '';
-		let i = 0;
+		let total = 0;
 		for (const category in submissions) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/trivia.ts
```diff
@@ -1793,19 +1770,22 @@ const triviaCommands: ChatCommands = {
...
-			`<tr><td colspan="6"><strong>${Chat.count(submissionsLen, "</strong> questions")} awaiting review:</td></tr>` +
-			`<tr><th>#</th><th>${this.tr`Category`}</th><th>${this.tr`Question`}</th><th>${this.tr`Answer(s)`}</th><th>${this.tr`Submitted By`}</th></tr>`;
 
+		let innerBuffer = '';
 		let i = 0;
-		while (i < submissionsLen) {
-			const entry = submissions[i];
-			buffer += `<tr><td><strong>${(++i)}</strong></td><td>${entry.category}</td><td>${entry.question}</td><td>${entry.answers.join(", ")}</td><td>${entry.user}</td></tr>`;
+		for (const category in submissions) {
+			for (const entry of submissions[category]) {
```
mia-pi-git, line 1777:
```suggestion
			for (const [i, entry] of submissions[category].entries()) {
```

server/chat-plugins/trivia.ts
```diff
@@ -1793,19 +1770,22 @@ const triviaCommands: ChatCommands = {
...
+				i++;
+				innerBuffer += `<tr><td><strong>${submissions[category].indexOf(entry)}</strong></td><td>${entry.category}</td><td>${entry.question}</td><td>${entry.answers.join(", ")}</td><td>${entry.user}</td></tr>`;
```
mia-pi-git, line 1779: Use the 'i' from above suggestion here instead of indexOf.

server/chat-plugins/trivia.ts
AnnikaCodes: We need `i` outside of the loop, on line 1785. (It's the total number of questions in all categories that await review.)

server/chat-plugins/trivia.ts
AnnikaCodes: Oh, I see what you mean, separating the index from the total.

server/chat-plugins/trivia.ts
mia-pi-git: Call it index or something, then?

mia-pi-git: Or tbh maybe just rename the other variable, since "i" seems unclear for that.

AnnikaCodes: Yes, I named it `total` instead.

Zarel:
I will trust that this is a good change, and also back up Trivia data just in case.

PR #7801, opened 2020-12-14
Fix userid search regexes for battlesearch/chatlog

mia-pi-git:


Zarel:
Not sure what this does but I'll trust that it's an improvement. Thanks!

PR #7802, opened 2020-12-14
Enable incremental compilation in `tsc`

AnnikaCodes:
This makes `npm test` faster.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ databases/*.db*
 .*-dist/
 tools/set-import/importer.js
 tools/modlog/converter.js
+*.tsbuildinfo
 
 # visual studio live share
 .vs
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,8 @@
         "moduleResolution": "node",
         "strict": true,
         "allowJs": true,
-        "checkJs": true
+        "checkJs": true,
+        "incremental": true
     },
     "types": ["node"],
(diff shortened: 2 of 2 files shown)


Zarel:
Neat, thanks!

urkerab:
How do you forcibly rebuild? I had an error in `data/abilities.ts` caused by a typo in `sim/dex-conditions.ts` and the only way I could proceed was to undo this change locally.

urkerab:
... although I made the same mistake again and this time `tsc` noticed my change, so whatever...

mia-pi-git:
> How do you forcibly rebuild? I had an error in `data/abilities.ts` caused by a typo in `sim/dex-conditions.ts` and the only way I could proceed was to undo this change locally.

``node build --force``.

urkerab:
... it was the `tsc` that was failing, does `node build` even affect that?

mia-pi-git:
Oh, i was only reading the "forcibly rebuild" part, sorry.

AnnikaCodes:
You can probably just remove `*.tsbuildinfo`.

Zarel:
We should probably make it so `build --force` deletes `*.tsbuildinfo`?

AnnikaCodes:
That doesn't make much sense to me, since `node build` doesn't use `tsc`. Maybe `npm run clean` or something?

Zarel:
I think that makes sense to me.

PR #7806, opened 2020-12-15
Update accuracy calculation for latest Gen 7 research

urkerab:
According to @DaWoblefet's latest research:
- Accuracy modifiers are chained, Gravity first, then Abilities, then Items
- Within each category modifiers execute in usual speed order
- Then, relative boosts are applied, so that 0/-6, +3/-3 and +6/0 have the same effect
- Finally, Micle Berry is applied, if appropriate.

The old code was moved to Gen 4 since the commit that added it says that this is the gen that it was based on.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -4080,10 +4080,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 260,
 	},
 	victorystar: {
-		onFoeModifyAccuracyPriority: -1,
-		onFoeModifyAccuracy(accuracy) {
-			if (typeof accuracy !== 'number') return;
-			return this.chainModify([0x119A, 0x1000]);
+		onAnyModifyAccuracyPriority: -1,
+		onAnyModifyAccuracy(accuracy, target, source) {
+			if (source.side === this.effectData.target.side && typeof accuracy === 'number') {
+				return this.chainModify([0x119A, 0x1000]);
+			}
 		},
 		name: "Victory Star",
 		rating: 2,
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -487,11 +487,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 172,
 	},
 	compoundeyes: {
-		onSourceModifyAccuracyPriority: 9,
+		onSourceModifyAccuracyPriority: -1,
 		onSourceModifyAccuracy(accuracy) {
 			if (typeof accuracy !== 'number') return;
 			this.debug('compoundeyes - enhancing accuracy');
-			return accuracy * 1.3;
+			return this.chainModify([0x14CD, 0x1000]);
 		},
 		name: "Compound Eyes",
 		rating: 3,
@@ -1430,10 +1430,10 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		onModifyAtk(atk) {
 			return this.modify(atk, 1.5);
 		},
-		onSourceModifyAccuracyPriority: 7,
+		onSourceModifyAccuracyPriority: -1,
 		onSourceModifyAccuracy(accuracy, target, source, move) {
 			if (move.category === 'Physical' && typeof accuracy === 'number') {
-				return accuracy * 0.8;
+				return this.chainModify([0x0CCD, 0x1000]);
 			}
 		},
 		name: "Hustle",
@@ -3076,12 +3076,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		onImmunity(type, pokemon) {
 			if (type === 'sandstorm') return false;
 		},
-		onModifyAccuracyPriority: 8,
+		onModifyAccuracyPriority: -1,
 		onModifyAccuracy(accuracy) {
 			if (typeof accuracy !== 'number') return;
 			if (this.field.isWeather('sandstorm')) {
 				this.debug('Sand Veil - decreasing accuracy');
-				return accuracy * 0.8;
+				return this.chainModify([0x0CCD, 0x1000]);
 			}
 		},
 		name: "Sand Veil",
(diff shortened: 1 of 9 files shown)


data/abilities.ts
```diff
@@ -3799,12 +3799,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
 	},
 	tangledfeet: {
-		onModifyAccuracyPriority: 6,
+		onModifyAccuracyPriority: -1,
 		onModifyAccuracy(accuracy, target) {
 			if (typeof accuracy !== 'number') return;
 			if (target?.volatiles['confusion']) {
 				this.debug('Tangled Feet - decreasing accuracy');
-				return accuracy * 0.5;
+				return this.chainModify(0.5);
```
DaWoblefet, line 3807: I think chainModify automatically turns this into a 4096, but it may be better for consistency to use 0x800, 0x1000.

DaWoblefet:
Do we have a good way to write tests for expected values of functions like this or BP? I think tests would be good but you wouldn't want to do it in the traditional way with a battle, since that would be dependent on the RNG seed.

MacChaeger:
There's a `battle.onEvent` function you can use to put assertions into event handlers. The [Stomping Tantrum tests ](https://github.com/smogon/pokemon-showdown/blob/master/test/sim/moves/stompingtantrum.js)have examples to refer to.

data/abilities.ts
urkerab: I checked other places e.g. Defeatist and they all seem to use the readable value if it has an exact binary representation.

DaWoblefet: As long as it actually uses 2048 when chain modifying, it's ok.

DaWoblefet:
I'll handle writing some tests then.

DaWoblefet:
Two tests to be fixed @urkerab: 
1. Victory Star isn't applying for the user (1st test)
2. I think applying stat boosts / drops is dividing by 0 somewhere. I think that's what the Dancer test failure and the 2nd test failure are.

data/scripts.ts
```diff
@@ -452,31 +452,23 @@ export const Scripts: BattleScriptsData = {
...
-						} else if (boost < 0) {
-							accuracy *= boostTable[-boost];
-						}
+						const boosts = this.runEvent('ModifyBoost', target, null, null, {...target.boosts});
+						boost = this.clampIntRange(boost - boosts['evasion'], -6, 6);
+					}
+					if (boost > 0) {
+						accuracy = this.trunc(accuracy * (boost + 3) / 3);
+					} else if (boost < 0) {
+						accuracy = this.trunc(accuracy * 3 / (boost + 3));
```
MacChaeger, line 469: At -3 accuracy, isn’t this a division by 0?

MacChaeger:
```suggestion
						accuracy = this.trunc(accuracy * 3 / (boost * -1 + 3));
```

urkerab: Yeah, this should be `3 - boost`.

urkerab: That seems to fix the Dancer test failure, at least.

test/sim/misc/accuracy.js
```diff
@@ -0,0 +1,67 @@
...
+			{species: 'Bonsly', ability: 'tangledfeet', moves: ['doubleteam', 'sleeptalk']},
+			{species: 'Pyukumuku', ability: 'noguard', moves: ['confuseray', 'sandattack', 'sleeptalk']},
+		]]);
+
+		battle.makeChoices('move sandattack -2, move sleeptalk', 'move doubleteam, move sandattack 2');
+		battle.makeChoices('auto', 'move sleeptalk, move confuseray -1');
+
+		battle.onEvent('Accuracy', battle.format, function (accuracy, target, source, move) {
+			if (move.id !== 'fireblast') return;
+			assert.equal(accuracy, 52);
```
urkerab, line 62: My calculations:
Base accuracy: 85%
Chained modifiers: Gravity, Victory Star, Compound Eyes, Tangled Feet: 4892 / 4096
Modified accuracy: 101.52 rounded to 102%
Boosts halve the accuracy to 51%

Of the possible permutations of modifiers, the following chained modifiers are possible:
- 4893 from Tangled Feet/Compound Eyes, Victory Star, Gravity
- 4892 from Tangled Feet/Compound Eyes, Gravity, Victory Star
- 4892 from Gravity/Victory Star, Tangled Feet/Compound Eyes
- All other orderings produce 4891 which results in an overall accuracy of 50%.

DaWoblefet: Oops that's right, I got so caught up on the infinity I forgot to use the right number! I fixed.

data/abilities.ts
```diff
@@ -4080,10 +4080,10 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 260,
 	},
 	victorystar: {
-		onAllyModifyMove(move) {
-			if (typeof move.accuracy === 'number') {
-				move.accuracy *= 1.1;
-			}
+		onFoeModifyAccuracyPriority: -1,
+		onFoeModifyAccuracy(accuracy) {
```
MacChaeger, line 4084: Wait, really? _This_ is why Victory Star doesn’t work? So it actually has not only been not working,  it’s actually been _lowering_ the teams evasiveness instead? Geez. I think you should probably fix the `ModifyAccuracy` event so that `onAllyModifyAccuracy` works here; it’s not called `ModifyEvasion` after all.

MacChaeger: And maybe just for fun let’s add a test to make sure allied Victory Star never boosts the opponent’s accuracy.

urkerab: It used to use the `ModifyMove` event, which targets the user of the move, which is why it affects an ally at that point. I'll need to double-check what happens if you target your own ally, so that Victory Star still boosts accuracy in this case; I might need to switch to using `onAnyModifyAccuracy` and check that the attacker and ability holder are on the same side.

urkerab: Yeah, I had to switch to explicitly checking that the attacker is the ability holder or an ally. Should be fixed now.

test/sim/misc/accuracy.js
```diff
@@ -0,0 +1,86 @@
...
+
+		battle.makeChoices();
+
+		battle = common.createBattle([
+			[{species: 'Butterfree', ability: 'victorystar', moves: ['fireblast']}],
+			[{species: 'Regirock', moves: ['sleeptalk']}],
+		]);
+
+		battle.onEvent('Accuracy', battle.format, function (accuracy) {
+			assert.equal(accuracy, 94, 'Victory Star Fire Blast should be 93% accurate');
```
Marty-D, line 31: I'm guessing the assertion is correct but the description isn't? Ditto the Wide Lens one below.

Marty-D (approved):
Looks good to me besides those test assertion nitpicks. After that's fixed I think this is ready to go, nice work and thanks for moving the appropriate stuff to Gen 4! 👍

test/sim/misc/accuracy.js
DaWoblefet: Yes, I'm bad

PR #7807, opened 2020-12-15
Turn off Format On Save in VS Code

aviettran:
This setting is off by default in vscode, but is commonly enabled at the User-level (it's under "Commonly Used"). Since this repo does not include a standardized formatter, it should be explicitly disabled.


AnnikaCodes:
bump

Zarel:
Thanks!

PR #7809, opened 2020-12-15
Display called moves before becoming Z-Move

bdepst:
According to the list of mechanics bugs, there was an issue where ["Z-moves of moves that call other moves should display the called move prior to becoming a Z-move" ](https://github.com/smogon/pokemon-showdown/projects/3#card-23557676)

In data/scripts.ts I've added checks in useMoveInner(), adding the appropriate messages for moves that are to become Z-Moves and which have been called by move-calling moves. I'm not sure if this is the preferred place in the code to handle displaying that text, but I couldn't see any easier way to display it with the correct values (perhaps caused by my unfamiliarity with the codebase). Hopefully my contribution is okay, as it passes all the unit tests, but I'm not quite sure how to check if these messages are being displayed properly when I'm running the simulator locally (much more familiar with playing in the client). Fortunately, it's mostly a visual change as opposed to something that could break the whole battle system, but please let me know if there's anything I should change. Thanks!

(Very new to this, but wanted to contribute to a project that I've spent so many hours enjoying over the years <3)

### DIFF
--- a/data/scripts.ts
+++ b/data/scripts.ts
@@ -157,6 +157,24 @@ export const Scripts: BattleScriptsData = {
 			this.singleEvent('ModifyType', move, null, pokemon, target, move, move);
 			if (move.type !== 'Normal') sourceEffect = move;
 		}
+		if (zMove && sourceEffect) {
+			if (['metronome', 'assist', 'sleeptalk'].includes(sourceEffect.id)) {
+				this.add('message', pokemon.name + ' will change ' + move.name + ' chosen by ' + sourceEffect.name + ' to its Z-Move!');
+			}
+			if (['copycat'].includes(sourceEffect.id)) {
+				this.add('message', pokemon.name + ' will change ' + move.name + ' copied by Copycat to its Z-Move!');
+			}
+			if (['mefirst'].includes(sourceEffect.id)) {
+				this.add('message', pokemon.name + ' will change ' + move.name + ' it stole using Me First to its Z-Move!');
+			}
+			if (['mirrormove'].includes(sourceEffect.id)) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Posting the strings here from Gen 7 text dump:
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] chosen by Metronome to its Z-Move!
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] chosen by Assist to its Z-Move!
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] chosen by Sleep Talk to its Z-Move!
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] copied by Copycat to its Z-Move!
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] it stole using Me First to its Z-Move!
[VAR PKNICK(0000)] will change [VAR MOVE(0001)] to its Z-Move using Mirror Move!

Nature Power doesn't have its own special hardcoded string technically speaking, but this implementation is fine. See youtube vid as reference: https://youtu.be/1mdiuFr96KY?t=145

This isn't typically how we handle custom text though. Usually that belongs in data/text/moves.ts. For example, this is the Metronome string:

move: "Waggling a finger let it use [MOVE]!"

So ideally I'd prefer to see it added there, and then in the this.add() we add the protocol message instead. @Marty-D do you have an opinion on how this would best be done?

urkerab:
Actually I think the protocol already gives you enough information, so all you need is the new strings, and then on the client, check whether the move is a Z-effect, and if so look for a Z-move message for the calling move rather then the regular move message. (But check that Z-Weather Ball still works though.)

aposp (approved):
LGTM

AnnikaCodes:
The PR and author have both been inactive for a year; closing this. Feel free to reopen if you have changes.

PR #7810, opened 2020-12-15
Support automodchat when all roomstaff are offline

mia-pi-git:
https://www.smogon.com/forums/threads/674410/page-2#post-8684167


server/rooms.ts
```diff
@@ -934,9 +938,44 @@ export abstract class BasicRoom {
...
+	runAutoModchat() {
+		if (!this.settings.autoModchat || this.settings.autoModchat.active) return;
+		// they are staff and online
+		const staff = Object.values(this.users).filter(u => this.auth.isStaff(u.id));
+		if (!staff.length && this.settings.autoModchat) {
+			const {rank, time} = this.settings.autoModchat;
+			this.modchatTimer = setTimeout(() => {
+				this.settings.modchat = rank;
+				this.add(
+					`|raw|<div class="broadcast-blue"><b>This room has had no active staff for ${Chat.toDurationString(time)},` +
```
DaWoblefet, line 955: Use  < strong > instead of < b >. Also I'd drop the exclamation mark, since I don't think automodchat is _that_ exciting :P

Zarel:
I'm nearly petty enough to automatically make the room a hidden room if it's public...

Zarel:
Hm. I don't think this should be a setting; it should probably just be instantaneous.

Zarel:
And required for all public rooms.

mia-pi-git:
> And required for all public rooms.

Not everyone wants to use this. Additionally, people might go offline accidentally?

Zarel:
Not a fan. I'll allow it for now.

Zarel:
To be clear: Not a fan of the policy. Your code is great! :)

PR #7811, opened 2020-12-15
Allow users to pm themselves while /blockpms is active

sirKiraUzumaki:
Suggestion: https://www.smogon.com/forums/threads/allow-users-to-chat-themselves-with-blockpms-active.3672227/

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -984,7 +984,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !Users.globalAuth.atLeast(user, targetUser.settings.blockPMs)) &&
-					!user.can('lock')) {
+					!user.can('lock') && toID(targetUser) != toID(user)) {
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -984,7 +984,7 @@ export class CommandContext extends MessageContext {
 				}
 				if (targetUser.settings.blockPMs &&
 					(targetUser.settings.blockPMs === true || !Users.globalAuth.atLeast(user, targetUser.settings.blockPMs)) &&
-					!user.can('lock')) {
+					!user.can('lock') && toID(targetUser) != toID(user)) {
```
mia-pi-git, line 987:
```suggestion
					!user.can('lock') && targetUser.id != user.id) {
```

mia-pi-git: (if you use toID on a user object, it returns user.id, so it's just cleaner this way)

AnnikaCodes: This should also be `!==` not `!=`.

Zarel:
Thanks!

PR #7813, opened 2020-12-16
Bugfix: Conversion 2 Submove Interaction

aviettran:
> Conversion 2 vs. moves that call other moves should use the called move, not the original move. e.g. Sleep Talk -> Thunderbolt should roll Ground, Electric, Grass, or Dragon, NOT Steel, Rock, or Ghost. See [replay ](https://replay.pokemonshowdown.com/gen7customgame-969010846)- DW

Are there any other moves in which the differentiation between `runMove` and `useMove` is significant?

Moves with references to `lastMove`: copycat, disable, eeriespell, encore, gmaxdepletion, grudge, iceball, instruct, mimic, mirrormove, rollout, skech, spite, torment, uproar

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -2440,11 +2440,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {authentic: 1},
 		onHit(target, source) {
-			if (!target.lastMove) {
+			if (!target.lastMoveUsed) {
 				return false;
 			}
 			const possibleTypes = [];
-			const attackType = target.lastMove.type;
+			const attackType = target.lastMoveUsed.type;
 			for (const type in this.dex.data.TypeChart) {
 				if (source.hasType(type)) continue;
 				const typeCheck = this.dex.data.TypeChart[type].damageTaken[attackType];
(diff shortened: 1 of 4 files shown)


urkerab:
That begs the question, at what point does the called move count as the last move used? For example, what happens if the called move gets blocked by Throat Chop?

aviettran:
@DaWoblefet can you weigh in on whether this change needs additional research, or if it's good to go?

DaWoblefet:
From the effects you've mentioned here, here's their behavior in Gen 7 as best I can tell:

Copycat - move must have at least passed the Powder check (before Micle / our onTry fail event with things like Fake Out failure)
Disable / Encore / Instruct / Mimic / Sketch / Spite / Torment - move must have at least consumed PP

Mirror Move does things a little differently than the Encore-style checks but I think it's also before consuming PP. It wouldn't hurt for me to directly test this.

Eerie Spell and G-Max Depletion are likely same as Spite. Grudge necessarily has to work because if you are KOed for Grudge to trigger, a move exists that was successful. idk what Rollout / Ice Ball are doing with it; Defense Curl just has to be used at some point prior to the move's execution. Same for Uproar.

aviettran:
So if I'm interpreting your post correctly:

- Copycat should copy the calling move, e.g., Sleep Talk, not the sub-move
- Disable, Encore, etc apply to the calling move
- You'll check Mirror Move, but it sounds like it will use the calling move, not the submove

I also checked Rollout and Iceball again and they are doing struggle checks, which aren't applicable for submoves.

So no further changes needed?

Marty-D (approved):
👍

PR #7816, opened 2020-12-16
Add two simple aliases for /suggestions

TheMezStrikes:
because they're intuitive and I keep thinking they already exist


Zarel:
Works for me, thanks!

PR #7817, opened 2020-12-16
Throttle how many rooms a user can join at once

mia-pi-git:
Discussed in Staff room.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -419,7 +419,8 @@ export const commands: ChatCommands = {
 		if (target.startsWith(`${Config.routes.replays}/`)) target = `battle-${target.slice(Config.routes.replays.length + 1)}`;
 		if (target.startsWith('psim.us/')) target = target.slice(8);
 		// isn't in tryJoinRoom so you can still join your own battles / gameRooms etc
-		if (!user.can('altsself') && !target.startsWith('view-') && curRooms.length >= 50) {
+		const numRooms = [...Rooms.rooms.values()].filter(r => user.id in r.users).length;
+		if (!user.can('altsself') && !target.startsWith('view-') && numRooms >= 50) {
 			return connection.sendTo(target as RoomID, `|noinit||You can only join 50 rooms at a time.`);
 		}
 		const ret = await user.tryJoinRoom(target as RoomID, connection);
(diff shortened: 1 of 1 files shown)


Slayer95:
This is not how you fail joining. You need to send ``|noinit|``

server/chat-commands/moderation.ts
```diff
@@ -412,11 +412,17 @@ export const commands: ChatCommands = {
 	j: 'join',
 	async join(target, room, user, connection) {
 		if (!target) return this.parse('/help join');
+		const curRooms = [...Rooms.rooms.values()].filter(r => user.id in r.users);
+		target = target.trim();
```
Zarel, line 416: I'd put this above the check for empty target.

Zarel, line 416:
```suggestion
		target = target.trim();
 		if (!target) return this.parse('/help join');
```

server/chat-commands/moderation.ts
```diff
@@ -412,11 +412,17 @@ export const commands: ChatCommands = {
...
 		if (target.startsWith('http://')) target = target.slice(7);
 		if (target.startsWith('https://')) target = target.slice(8);
 		if (target.startsWith(`${Config.routes.client}/`)) target = target.slice(Config.routes.client.length + 1);
 		if (target.startsWith(`${Config.routes.replays}/`)) target = `battle-${target.slice(Config.routes.replays.length + 1)}`;
 		if (target.startsWith('psim.us/')) target = target.slice(8);
+		// isn't in tryJoinRoom so you can still join your own battles / gameRooms etc
+		if (!user.can('altsself') && !target.startsWith('view-') && curRooms.length >= 50) {
+			return connection.sendTo(target as RoomID, `|noinit||You can only join 50 rooms at a time.`);
+		}
```
Zarel, line 425:
```suggestion
		// isn't in tryJoinRoom so you can still join your own battles / gameRooms etc
		const numRooms = [...Rooms.rooms.values()].filter(r => user.id in r.users).length;
		if (!user.can('altsself') && !target.startsWith('view-') && numRooms >= 50) {
			return connection.sendTo(target as RoomID, `|noinit||You can only join 50 rooms at a time.`);
		}
```

server/chat-commands/moderation.ts
```diff
@@ -411,12 +411,18 @@ export const commands: ChatCommands = {
 	joim: 'join',
 	j: 'join',
 	async join(target, room, user, connection) {
-		if (!target) return this.parse('/help join');
+		target = target.trim();
+ 		if (!target) return this.parse('/help join');
```
Zarel, line 415:
```suggestion
		if (!target) return this.parse('/help join');
```

Zarel:
Looks good! Thanks!

PR #7819, opened 2020-12-17
Bug fix: pursuit switch move interaction

aviettran:
> If Pursuit is used on the same turn a Pokemon uses Volt Switch, U-turn, or Parting Shot (not Baton Pass), the user of the switching move should take damage, then the player should have the option to choose a new Pokemon. Currently, the player is able to select a new Pokemon prior to taking Pursuit damage. See replay. DaWoblefet tested in Gens 5, 6, and 7, and Marty said it's always been this way

### DIFF
--- a/test/sim/moves/pursuit.js
+++ b/test/sim/moves/pursuit.js
@@ -73,7 +73,7 @@ describe(`Pursuit`, function () {
 
 	it(`should deal damage prior to attacker selecting a switch in after u-turn etc`, function () {
 		battle = common.createBattle([[
-			{species: 'tyranitar', moves: ['pursuit']},
+			{species: 'parasect', moves: ['pursuit']},
 		], [
 			{species: 'emolga', moves: ['voltswitch']},
 			{species: 'zapdos', moves: ['batonpass']},
--- a/test/sim/moves/pursuit.js
+++ b/test/sim/moves/pursuit.js
@@ -70,4 +70,21 @@ describe(`Pursuit`, function () {
 		battle.makeChoices('move Pursuit mega -2, switch 3', 'auto');
 		assert.equal(furret.maxhp - furret.hp, 66);
 	});
+
+	it(`should deal damage prior to attacker selecting a switch in after u-turn etc`, function () {
+		battle = common.createBattle([[
+			{species: 'tyranitar', moves: ['pursuit']},
+		], [
+			{species: 'emolga', moves: ['voltswitch']},
+			{species: 'zapdos', moves: ['batonpass']},
+		]]);
+		battle.makeChoices('move Pursuit', 'move voltswitch');
+		assert.false.fullHP(battle.p2.pokemon[0]);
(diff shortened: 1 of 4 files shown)


DaWoblefet:
One of the things Marty and I wanted done with this is for the Pokemon to visually disappear from the field as well after Volt Switch etc., waiting for the new Pokemon to be sent in. Does this also accomplish that?

aviettran:
They do not. It also looks like there are issues with the client. It waits until the end of the turn, even though the request happens earlier. I'll move this to draft while I investigate.

aviettran:
I am moving this out of draft. I ran some tests in client and fixed a few things. I am not confident I've tested every edge case, so I'd appreciate it if someone could review. If someone could also, at a high level, explain what would be involved in having the switching Pokemon disappear, I can also look into that. I assume that would go in the client repo.

DaWoblefet:
Here's an example on cart: https://youtu.be/EcudwHuQJQ4?t=6993

Notice how Rillaboom uses U-turn, leaves the field by returning to its Poke Ball, and then I select another mon to switch in. From the opponent's perspective, my Pokemon has effectively disappeared.

The reason it's useful to display the Pokemon actually returning is because of situations where the Pokemon _doesn't_ return to the player. For example, if you use Volt Switch into Eject Button, the Eject Button mon switches out, but Volt Switch doesn't. So having the visual cue is useful to explain more obviously what happened to the other player.

aviettran:
I'm thinking we should merge this change if there are no outstanding issues. The client work will have to be a separate PR anyway, and I'm not quite up to speed with that repo at this time.

DaWoblefet:
@Marty-D, looks good to merge for me, how about you?

test/sim/moves/pursuit.js
```diff
@@ -70,4 +70,22 @@ describe(`Pursuit`, function () {
...
+			{species: 'emolga', moves: ['voltswitch']},
+			{species: 'zapdos', moves: ['batonpass']},
+		]]);
+		battle.makeChoices('move Pursuit', 'move voltswitch');
+		assert.false.fullHP(battle.p2.pokemon[0]);
+		battle.choose('p2', 'switch 2');
+		assert.equal(battle.p2.pokemon[0].name, "Zapdos");
+		battle.makeChoices('move Pursuit', 'move batonpass');
+		battle.choose('p2', 'switch 2');
+		assert.fullHP(battle.p2.pokemon[1], 'should not hit Pokemon that has used Baton Pass');
```
Marty-D, line 87: I imagine it's possible for Emolga to faint to the first Pursuit should it happen to be a critical hit, which will make this check fail. I'd suggest using a weaker attacker and/or a less frail defender.

aviettran: Swapped ttar out for parasect.

Marty-D (approved):
Great, thanks!

PR #7821, opened 2020-12-18
Fix spinlock in /nomarkshared view

mia-pi-git:


AnnikaCodes (changes requested):

server/ip-tools.ts
```diff
@@ -177,7 +177,10 @@ export const IPTools = new class {
 		const aParts = a.split('.');
 		const bParts = b.split('.');
 		while (diff === 0) {
-			diff = (parseInt(aParts[i]) || 0) - (parseInt(bParts[i]) || 0);
+			const aPart = parseInt(aParts[i]);
+			const bPart = parseInt(bParts[i]);
+			if (isNaN(aPart) || isNaN(bPart)) throw new Error("Invalid IP passed to IPTools.ipSort.");
+			diff = (aPart || 0) - (bPart || 0);
```
AnnikaCodes, line 183:
```suggestion
			diff = aPart - bPart;
```
parseInt returns an integer or NaN, nothing falsy except 0

AnnikaCodes (changes requested):

AnnikaCodes (approved):
Thanks!

server/chat-plugins/hosts.ts
```diff
@@ -135,7 +135,7 @@ export const pages: PageTable = {
 		} else {
 			buf += `<div class="ladder"><table><tr><th>IP</th><th>Reason</th></tr>`;
 			const sortedSharedIPBlacklist = [...Punishments.sharedIpBlacklist];
-			sortedSharedIPBlacklist.sort((a, b) => IPTools.ipSort(a[1], b[1]));
+			sortedSharedIPBlacklist.sort((a, b) => IPTools.ipSort(a[0], b[0]));
```
Zarel, line 138: Should probably be `ipCompare`? Also I wonder if `sortBy` would be easier?

Zarel:
```ts
Utils.sortBy(sortedSharedIPBlacklist, ip => ip.split('.').map(octet => parseInt(octet)));
```

PR #7822, opened 2020-12-20
Feature: Make Pivot moves a /ms category

aviettran:
https://www.smogon.com/forums/threads/make-pivot-moves-a-ms-category.3672772/

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -45,6 +45,7 @@ interface MoveOrGroup {
 	targets: {[k: string]: boolean};
 	recoil: boolean;
 	skip: boolean;
+	pivot: boolean;
 }
 
 type Direction = 'less' | 'greater' | 'equal';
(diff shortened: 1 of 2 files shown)


server/chat-plugins/datasearch.ts
```diff
@@ -1841,6 +1883,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 	} else {
 		resultsStr += "No moves found.";
 	}
+	if (isTest) return {results, reply: resultsStr};
```
Zarel, line 1886: It seems to me like this should rather be refactored into two functions, one that returns `results`, and one that transforms `results` into `resultsStr`.

(Not for this PR, but probably a good follow-up PR.)

Zarel (approved):
I don't see anything obviously wrong with this. I'll leave this open for a dayish in case someone else wants to review, because I'm not very familiar with this code.

Zarel:
Thanks!

PR #7823, opened 2020-12-20
Refactor Room#add and Room#addRaw to automatically update

AnnikaCodes:
This simplifies the common use case of wanting to add something to a room and send that to the users in the room immediately.

Thoughts?

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2580,7 +2580,7 @@ export const commands: ChatCommands = {
 		} else {
 			buf += `</small></div>`;
 		}
-		room.add(`|c| ${request.name}|/raw ${buf}`);
+		room.addQueue(`|c| ${request.name}|/raw ${buf}`);
 		this.privateModAction(`${user.name} approved showing media from ${request.name}.`);
 	},
 	approveshowhelp: [`/approveshow [user] - Approves the media display request of [user]. Requires: % @ # &`],
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -325,7 +325,7 @@ export const commands: ChatCommands = {
 		await rebuild(this);
 
 		const lock = Monitor.hotpatchLock;
-		const hotpatches = ['chat', 'formats', 'loginserver', 'punishments', 'dnsbl', 'modlog'];
+		const hotpatches = ['chat', 'formats', 'loginserver', 'punishments', 'iptools', 'modlog'];
 		const version = await Monitor.version();
 
 		let patch = target;
@@ -439,8 +439,8 @@ export const commands: ChatCommands = {
 				global.Punishments = require('../punishments').Punishments;
 				this.sendReply("DONE");
 			} else if (target === 'dnsbl' || target === 'datacenters' || target === 'iptools') {
-				patch = 'dnsbl';
-				this.sendReply("Hotpatching ip-tools...");
+				patch = 'iptools';
+				this.sendReply("Hotpatching IP tools...");
 
 				global.IPTools = require('../ip-tools').IPTools;
 				void IPTools.loadHostsAndRanges();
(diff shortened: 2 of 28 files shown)


mia-pi-git:
Why, though? We can just use room.add(...).update(), no?
I think this is actually longer by a bit.

KrisXV:
> Why, though? We can just use room.add(...).update(), no?
> I think this is actually longer by a bit, so tbh I fail to see why this is super needed.

It's the exact same amount of characters.

mia-pi-git:
> > Why, though? We can just use room.add(...).update(), no?
> > I think this is actually longer by a bit, so tbh I fail to see why this is super needed.
> 
> It's the exact same amount of characters.

Okay, well, my point still stands that this seems like more complication than is needed, given how room.add works.

AnnikaCodes:
I personally think it's cleaner so I figured I'd make a pr to see what other developers thought. If nobody else wants it that's perfectly fine, it's a minor thing.

KrisXV:
I personally prefer this because I'd rather not have to call .update() every time I want to use room.add(...).

Slayer95:
Might as well repurpose ``.add()`` to also update the room, and make a new API for adding text without sending it to the users (maybe ``addQueue()``).

Slayer95:
Oh. I almost forgot. ``addRaw().update()`` is a [sister API](https://github.com/smogon/pokemon-showdown/commit/a2d2a523019c298a66e88e8bef2be70a352144ae). Presumably they should be updated together.

Zarel:
There's a lot of uses of `add` in this PR which should be `addQueue` if you really want to use that approach.

Zarel:
(Over 50% of all current uses of `add` should be `addQueue`.)

Zarel:
That's the reason we `add` defaulted not to automatically update: So that if you called `add` multiple times in a row, the messages would be combined into a single packet.

One other option that would keep the API simple would be just to add the room to e.g. `Rooms.global.pendingUpdate: Set<Room>`, and then flush that queue at the end of every command.

Slayer95:
> One other option that would keep the API simple would be just to add the room to e.g. Rooms.global.pendingUpdate: Set<Room>, and then flush that queue at the end of every command.

That wouldn't be much different from the status quo, since every single use would need to be non-trivially audited against message ordering bugs -formerly common in those commands that use the modlog-, etc.

For instance: 831dec5b6 would cause a ``core.ts`` message not be sent until someone joins/leaves the room or sends a different message, forcing an update. Similarly, changing ``add()`` to ``addQueue()`` in plugin methods would make timeouts/intervals harder. It's not likely to be the right abstraction level to handle chat log queues.

server/chat-commands/info.ts
```diff
@@ -2580,7 +2580,7 @@ export const commands: ChatCommands = {
 		} else {
 			buf += `</small></div>`;
 		}
-		room.add(`|c| ${request.name}|/raw ${buf}`);
+		room.addQueue(`|c| ${request.name}|/raw ${buf}`);
```
mia-pi-git, line 2583: Should be an `add`, not `addQueue`.

Slayer95:
> One other option that would keep the API simple would be just to add the room to e.g. Rooms.global.pendingUpdate: Set<Room>, and then flush that queue at the end of every command.

My preferred approach in this line, which would handle the issue with plugins and timers, but not liberate us from an audit effort, would be:

```js
scheduleUpdate() {
  this.needsUpdate = true;
  setImmediate(() => {
    if (!this.needsUpdate) return;
    this.update();
    this.needsUpdate = false;
  });
}

add(message) {
  this.log.add(message);
  this.scheduleUpdate();
  return this;
}
```

Zarel:
> My preferred approach in this line

I like your approach a lot, @Slayer95! I think it solves all our problems.

I would add a micro-optimization where we set `room.needsUpdate = true` at the beginning of a chat command, to save a `setImmediate` when adding text to the current room.

AnnikaCodes:
I don't think this is a big enough improvement from the status quo for me to want to work on, and I don't like having PRs I don't want to work on languishing.

PR #7824, opened 2020-12-21
Alias /srgc to /subroomgroupchat

sirKiraUzumaki:
Adding a simple alias "srgc" for commands /subroomgroupchat.

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -693,6 +693,7 @@ export const commands: ChatCommands = {
 	],
 
 	subroomgroupchat: 'makegroupchat',
+	srgc: 'makegroupchat',
 	mgc: 'makegroupchat',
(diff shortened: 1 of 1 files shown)


Zarel:
I'm OK with this if there's enough interest. Was there a Suggestions thread somewhere?

sirKiraUzumaki:
> I'm OK with this if there's enough interest. Was there a Suggestions thread somewhere?

No there wasn't a suggestion thread.

sirKiraUzumaki:
I have also dropped a message in discord but lemme also clear this here. As /subroomgroupchat command to lengthy to typed. Sometimes typos can be faced and then you need to type in again (especially this sucks when you are typing on mobile).

Zarel:
Hm. It matches other groupchat commands and at four characters long it's unlikely to conflict, so I think if you find it useful, that's good enough for me. Thanks!

PR #7826, opened 2020-12-21
Translations: Properly ignore non-translation directories

mia-pi-git:
At the moment, sourceMaps is getting loaded as a language. This is obviously not right, so i replaced the current directory check (which it passes because sourceMaps is a directory) with a simple caps check, since as the comment says none of the languages should have caps in the dir title.


Zarel:
I don't think I like this? A lot of OSes have a tendency to put stray files and directories in, like Windows's `Thumbs.db` or macOS's `.DS_Store`. While this does get them, I don't think we should _rely_ on them being capitalized this way.

I would probably specifically look for a directory name which is an ID.

PR #7827, opened 2020-12-21
Document restriction rules

urkerab:


Zarel:
I was hoping to have examples of restrictions here. Like, I definitely was very confused about what a restriction was until I saw an example. I think it would help a lot here.

urkerab:
As in a list of OMs and how they use restrictions?

Zarel:
Not the entire list, but just a few examples so people can better understand what restrictions are for.

urkerab:
OK, I picked 4 OMs and used them as examples.

Zarel:
This is great! Much clearer than before! Thanks!

PR #7829, opened 2020-12-22
Improve /help dc

Distrib-ps:


server/chat-plugins/hosts.ts
```diff
@@ -282,18 +282,17 @@ export const commands: ChatCommands = {
...
+			`<code>/ipranges view [type]</code>: view the list of a particular type of IP range (<code>residential</code>, <code>mobile</code>, or <code>proxy</code>). Requires: hosts manager @ & <br />` +
+			`<code>/ipranges add [type], [low IP]-[high IP], [host]</code>: add IP ranges (can be multiline). Requires: hosts manager &</br></summary>` +
+			`<code>/ipranges widen [type], [low IP]-[high IP], [host]</code>: add IP ranges, allowing a new range to completely cover an old range. Requires: hosts manager & <br />` +
+			`For example: <code>/ipranges add proxy, 5.152.192.0-5.152.223.255, redstation.com</code>. <br />` +
+			`Get datacenter info from <code>/whois</code>; <code>[low IP]</code>, <code>[high IP]</code> are the range in the last inetnum. <br />` +
+			`<code>/ipranges remove [low IP]-[high IP]</code>: remove IP range(s). Can be multiline. Requires: hosts manager & <br />` +
+			`For example: <code>/ipranges remove 5.152.192.0, 5.152.223.255</code>. <br />` +
+			`<code>/ipranges rename [low IP]-[high IP], [host]</code>: changes the host an IP range resolves to. Requires: hosts manager &`
+		);
```
AnnikaCodes, line 295: I think that `/ipranges view [type]` and `/ipranges add` are the only two that need to be above the readmore (`/ipranges view` with no type is impractical on Main since there are so many IP ranges).

AnnikaCodes: Also, you should use `<br />` consistently (personally, I'd keep using `Array#join` for the rest of the help).

Distrib-ps: Something like that ?
http://prntscr.com/w81ots , http://prntscr.com/w81p9b

AnnikaCodes: Yes, although you don't want to duplicate the help.

AnnikaCodes:
Looks good, thanks!

PR #7830, opened 2020-12-24
Utils: Visualize Maps/Sets in a cleaner fashion

mia-pi-git:
This is nice so that we can see the output of Maps / Sets without having to use Array.from or [...thing].

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -97,7 +97,6 @@ export const Utils = new class {
 		if (!baseName) baseName = constructor;
 
 		switch (baseName) {
-		// other constructor cases can be handled here!
 		case 'Map':
 			if (depth > 2) return `Map`;
 			const mapped = [...value.entries()].map(
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -90,6 +90,22 @@ export const Utils = new class {
 		} else {
 			constructor = 'null';
 		}
+
+		// if it has a toString, check that to see if it tells us if there's a base class, else look at the constructor name
+		let baseName = value?.toString ? /\[object (.*)\]/.exec(value.toString())?.[1] : null;
+		if (!baseName) baseName = constructor;
+
+		switch (baseName) {
+		// other constructor cases can be handled here!
+		case 'Map':
+			const mapped = [...value.entries()].map(
+				val => `${this.visualize(val[0], depth + 1)} => ${this.visualize(val[1], depth + 1)}`
+			);
+			return `${constructor} (${value.size}) { ${mapped.join(', ')} }`;
+		case 'Set':
+			return `${constructor} (${value.size}) { ${[...value].map(v => this.visualize(v), depth + 1).join(', ')} }`;
+		}
+
 		if (value.toString) {
 			try {
 				const stringValue = value.toString();
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -92,17 +92,20 @@ export const Utils = new class {
 		}
 
 		// if it has a toString, check that to see if it tells us if there's a base class, else look at the constructor name
+		// (this is to pull the base class so it works for subclasses)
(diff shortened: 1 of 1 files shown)


lib/utils.ts
```diff
@@ -90,6 +90,22 @@ export const Utils = new class {
 		} else {
 			constructor = 'null';
 		}
+
+		// if it has a toString, check that to see if it tells us if there's a base class, else look at the constructor name
+		let baseName = value?.toString ? /\[object (.*)\]/.exec(value.toString())?.[1] : null;
+		if (!baseName) baseName = constructor;
```
Zarel, line 96: I still don't understand what this is for.

lib/utils.ts
```diff
@@ -90,6 +90,22 @@ export const Utils = new class {
...
+
+		switch (baseName) {
+		// other constructor cases can be handled here!
```
Zarel, line 99: This seems kind of superfluous as far as comments go?

```suggestion
```

You could add this comment to any `switch` statement. It's just how `switch`es work. I don't think old language features like `switch` need to be documented this way.

Zarel: Remind me to add a detailed explanation on what comments are for, in CONTRIBUTING, at some point.

lib/utils.ts
```diff
@@ -90,6 +90,22 @@ export const Utils = new class {
...
+		case 'Map':
+			const mapped = [...value.entries()].map(
+				val => `${this.visualize(val[0], depth + 1)} => ${this.visualize(val[1], depth + 1)}`
+			);
+			return `${constructor} (${value.size}) { ${mapped.join(', ')} }`;
+		case 'Set':
+			return `${constructor} (${value.size}) { ${[...value].map(v => this.visualize(v), depth + 1).join(', ')} }`;
```
Zarel, line 106:
```suggestion
		case 'Map':
			const mapped = [...value].map(
				([k, v]) => `${this.visualize(k, depth + 1)}: ${this.visualize(v, depth + 1)}`
			);
			return `${constructor}<${value.size}>(${mapped.join(', ')})`;
		case 'Set':
			return `${constructor}<${value.size}>(${[...value].map(v => this.visualize(v), depth + 1).join(', ')})`;
```

At some point we should revisit how lists are displayed, but this should be better than before.

lib/utils.ts
Zarel: Did!

https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#dont-teach-javascript

lib/utils.ts
```diff
@@ -90,6 +90,25 @@ export const Utils = new class {
 		} else {
 			constructor = 'null';
 		}
+
+		// if it has a toString, check that to see if it tells us if there's a base class, else look at the constructor name
+		// (this is to pull the base class so it works for subclasses)
```
Zarel, line 95:
```suggestion
		// If it has a toString, try to grab the base class from there
		// (This is for Map/Set subclasses like user.auth)
```

lib/utils.ts
```diff
@@ -90,6 +90,25 @@ export const Utils = new class {
...
+		let baseName = value?.toString ? /\[object (.*)\]/.exec(value.toString())?.[1] : null;
+		if (!baseName) baseName = constructor;
+
+		switch (baseName) {
```
Zarel, line 99:
```suggestion
		const baseClass = (value?.toString && /\[object (.*)\]/.exec(value.toString())?.[1]) || constructor;

		switch (baseClass) {
```

lib/utils.ts
```diff
@@ -90,6 +90,25 @@ export const Utils = new class {
...
+		// other constructor cases can be handled here!
```
Zarel, line 100:
```suggestion
```
Did you forget about this? This is now rule C1 in CONTRIBUTING:

https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#c1-dont-teach-javascript

Zarel:
Thanks!

PR #7832, opened 2020-12-24
Poll: Fix deletequeue

Distrib-ps:
Pr for fix /deletequeue
recording

### DIFF
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -350,23 +350,23 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
-			target = target.trim();
-			const [slotString, roomid] = target.split(',');
-			const Room = roomid ? Rooms.search(roomid) : room;
-			if (!Room) return this.errorReply(`Invalid room.`);
-			if (!Room.persist) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			const targetRoom = roomid ? Rooms.search(roomid) : room;
+			if (!targetRoom) return this.errorReply(`Invalid room.`);
+			if (!targetRoom.persist) {
 				return this.errorReply("This command is unavailable in temporary rooms.");
 			}
-			this.room = Room;
-			this.checkCan('mute', null, Room);
-			if (!Room.minorActivityQueue) {
+			this.room = targetRoom;
+			room = this.requireRoom();
+			this.checkCan('mute', null, room);
+			if (!targetRoom.minorActivityQueue) {
 				return this.errorReply(this.tr`The queue is already empty.`);
 			}
-			if (cmd === 'deletequeue' && Room.minorActivityQueue.length !== 1 && !target) {
+			if (cmd === 'deletequeue' && targetRoom.minorActivityQueue.length !== 1 && !target) {
 				return this.parse('/help deletequeue');
 			}
 			if (!target) {
-				Room.minorActivityQueue = null;
+				targetRoom.minorActivityQueue = null;
 				this.modlog('CLEARQUEUE');
 				this.sendReply(this.tr`Cleared poll queue.`);
 			} else {
@@ -376,7 +376,7 @@ export const commands: ChatCommands = {
 				if (isNaN(slot)) {
 					return this.errorReply(this.tr`Can't delete poll at slot ${slotString} - "${slotString}" is not a number.`);
 				}
-				if (!Room.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
+				if (!targetRoom.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
 
 				curRoom.minorActivityQueue!.splice(slot - 1, 1);
 				if (!curRoom.minorActivityQueue?.length) curRoom.minorActivityQueue = null;
@@ -388,7 +388,8 @@ export const commands: ChatCommands = {
 				});
 				curRoom.sendMods(this.tr`(${user.name} deleted the queued poll in slot ${slot}.)`);
 				curRoom.update();
-				if (roomid) this.parse(`/j view-pollqueue-${Room.roomid}`);
+				if (roomid) this.parse(`/j view-pollqueue-${curRoom}`);
+				if (update) this.parse(`/j view-pollqueue-${curRoom}`);
 			}
 		},
 		deletequeuehelp: [
@@ -567,6 +568,7 @@ export const commands: ChatCommands = {
 export const pages: PageTable = {
 	pollqueue(args, user) {
 		const room = this.requireRoom();
+
 		let buf = `<div class="pad"><strong>${this.tr`Queued polls:`}</strong>`;
 		buf += `<button class="button" name="send" value="/join view-pollqueue-${room.roomid}" style="float: right">`;
 		buf += `<i class="fa fa-refresh"></i> ${this.tr`Refresh`}</button><br />`;
@@ -578,7 +580,7 @@ export const pages: PageTable = {
 			const number = i + 1; // for translation convienence
 			const button = (
 				`<strong>${this.tr`#${number} in queue`} </strong>` +
-				`<button class="button" name="send" value="/poll deletequeue ${i + 1},${room.roomid}">` +
+				`<button class="button" name="send" value="/poll deletequeue ${i + 1},${room.roomid},updatelist">` +
 				`(${this.tr`delete`})</button>`
 			);
 			buf += `<hr />`;
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -357,16 +357,17 @@ export const commands: ChatCommands = {
 				return this.errorReply("This command is unavailable in temporary rooms.");
 			}
 			this.room = targetRoom;
+			room = targetRoom;
 			room = this.requireRoom();
 			this.checkCan('mute', null, room);
-			if (!targetRoom.minorActivityQueue) {
+			if (!room.minorActivityQueue) {
 				return this.errorReply(this.tr`The queue is already empty.`);
 			}
-			if (cmd === 'deletequeue' && targetRoom.minorActivityQueue.length !== 1 && !target) {
+			if (cmd === 'deletequeue' && room.minorActivityQueue.length !== 1 && !target) {
 				return this.parse('/help deletequeue');
 			}
 			if (!target) {
-				targetRoom.minorActivityQueue = null;
+				room.minorActivityQueue = null;
 				this.modlog('CLEARQUEUE');
 				this.sendReply(this.tr`Cleared poll queue.`);
 			} else {
@@ -376,7 +377,7 @@ export const commands: ChatCommands = {
 				if (isNaN(slot)) {
 					return this.errorReply(this.tr`Can't delete poll at slot ${slotString} - "${slotString}" is not a number.`);
 				}
-				if (!targetRoom.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
+				if (!room.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
 
 				curRoom.minorActivityQueue!.splice(slot - 1, 1);
 				if (!curRoom.minorActivityQueue?.length) curRoom.minorActivityQueue = null;
--- a/server/chat-plugins/poll.ts
+++ b/server/chat-plugins/poll.ts
@@ -350,27 +350,33 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
-			room = this.requireRoom();
-			this.checkCan('mute', null, room);
-			if (!room.minorActivityQueue) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/poll.ts
```diff
@@ -573,7 +578,7 @@ export const pages: PageTable = {
 			const number = i + 1; // for translation convienence
 			const button = (
 				`<strong>${this.tr`#${number} in queue`} </strong>` +
-				`<button class="button" name="send" value="/poll deletequeue ${i + 1},${room.roomid},updatelist">` +
```
mia-pi-git, line 576: Why? This is important, don't remove this.

server/chat-plugins/poll.ts
```diff
@@ -350,27 +350,33 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
-			room = this.requireRoom();
-			this.checkCan('mute', null, room);
-			if (!room.minorActivityQueue) {
+			target = target.trim();
+			const [slotString, roomid] = target.split(',');
+			const Room = roomid ? Rooms.search(roomid) : room;
```
mia-pi-git, line 355: Should be `room`, not Room.

mia-pi-git, line 355: This is additionally not the fix you want. This does not work.
This is what you want instead.
```ts
const [slotString, roomid, update] = target.split(',');
room = roomid ? Rooms.search(roomid) : room;
room = this.requireRoom();

```

None of these other changes are necessary.

mia-pi-git (changes requested):
This does not work as is, and in fact breaks more functionality.

server/chat-plugins/poll.ts
Distrib-ps: If I add only this code, there is an error: http://prntscr.com/w99uye

mia-pi-git: I mean that's pretty simple to fix.
Just make it into:
room = roomid ? Rooms.search(roomid) : room;
this.room = room;
room = this.requireRoom();

Distrib-ps: Sorry I'm trying to learn at the same time

Distrib-ps:
```suggestion
					clearqueue: 'deletequeue',
		deletequeue(target, room, user, connection, cmd) {
			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
			const targetRoom = roomid ? Rooms.search(roomid) : room;
			if (!targetRoom) return this.errorReply(`Invalid room.`);
			if (!targetRoom.persist) {
				return this.errorReply("This command is unavailable in temporary rooms.");
			}
			this.room = targetRoom;
			room = this.requireRoom();
			this.checkCan('mute', null, room);
			if (!targetRoom.minorActivityQueue) {
				return this.errorReply(this.tr`The queue is already empty.`);
			}
			if (cmd === 'deletequeue' && targetRoom.minorActivityQueue.length !== 1 && !target) {
				return this.parse('/help deletequeue');
			}
			if (!target) {
				targetRoom.minorActivityQueue = null;
				this.modlog('CLEARQUEUE');
				this.sendReply(this.tr`Cleared poll queue.`);
			} else {
				const [slotString, roomid, update] = target.split(',');
				const slot = parseInt(slotString);
				const curRoom = roomid ? (Rooms.search(roomid) as ChatRoom | GameRoom) : room;
				if (!curRoom) return this.errorReply(this.tr`Room "${roomid}" not found.`);
				if (isNaN(slot)) {
					return this.errorReply(this.tr`Can't delete poll at slot ${slotString} - "${slotString}" is not a number.`);
				}
				if (!targetRoom.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);

				curRoom.minorActivityQueue!.splice(slot - 1, 1);
				if (!curRoom.minorActivityQueue?.length) curRoom.minorActivityQueue = null;

				curRoom.modlog({
					action: 'DELETEQUEUE',
					loggedBy: user.id,
					note: slot.toString(),
				});
				curRoom.sendMods(this.tr`(${user.name} deleted the queued poll in slot ${slot}.)`);
				curRoom.update();
				if (update) this.parse(`/j view-pollqueue-${curRoom}`);
			}
		},
```
It's good like that ?

mia-pi-git: Remove line 373. Duplicated code.

mia-pi-git: You also need the "update" variable to update the page if it's there.

mia-pi-git: At line 391 you check for if (roomid) (parse join), but that should actually be if (update), as well.

server/chat-plugins/poll.ts
```diff
@@ -382,6 +388,7 @@ export const commands: ChatCommands = {
 				});
 				curRoom.sendMods(this.tr`(${user.name} deleted the queued poll in slot ${slot}.)`);
 				curRoom.update();
+				if (roomid) this.parse(`/j view-pollqueue-${curRoom}`);
```
AnnikaCodes, line 391: What is this line for?

Distrib-ps: Oops it was for testing.
You know how to fix the 120 character limit problem ?

server/chat-plugins/poll.ts
```diff
@@ -350,27 +350,33 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			const targetRoom = roomid ? Rooms.search(roomid) : room;
+			if (!targetRoom) return this.errorReply(`Invalid room.`);
+			if (!targetRoom.persist) {
+				return this.errorReply("This command is unavailable in temporary rooms.");
+			}
+			this.room = targetRoom;
```
mia-pi-git, line 359: Additionally, set `room = targetRoom`, then change all the uses of `targetRoom` back to `room`. Cleaner that way.

server/chat-plugins/poll.ts
```diff
@@ -350,27 +350,33 @@ export const commands: ChatCommands = {
...
 			} else {
-				const [slotString, roomid, update] = target.split(',');
 				const slot = parseInt(slotString);
 				const curRoom = roomid ? (Rooms.search(roomid) as ChatRoom | GameRoom) : room;
 				if (!curRoom) return this.errorReply(this.tr`Room "${roomid}" not found.`);
 				if (isNaN(slot)) {
 					return this.errorReply(this.tr`Can't delete poll at slot ${slotString} - "${slotString}" is not a number.`);
 				}
-				if (!room.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
+				if (!targetRoom.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
```
mia-pi-git, line 379:
```suggestion
				if (!room.minorActivityQueue[slot - 1]) return this.errorReply(this.tr`There is no poll in queue at slot ${slot}.`);
```

Changing this back to `room` like I said above will fix the max-len error,

Distrib-ps:
I put a message here.
I don't know if she was forgotten. Thx

server/chat-plugins/poll.ts
```diff
@@ -350,6 +350,14 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			const targetRoom = roomid ? Rooms.search(roomid) : room;
+			if (!targetRoom) return this.errorReply(`Invalid room.`);
+			if (!targetRoom.persist) {
+				return this.errorReply("This command is unavailable in temporary rooms.");
+			}
+			this.room = targetRoom;
```
mia-pi-git, line 359:
```suggestion
			this.room = room = targetRoom;
```
Probably cleaner.

server/chat-plugins/poll.ts
```diff
@@ -350,6 +350,14 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			const targetRoom = roomid ? Rooms.search(roomid) : room;
+			if (!targetRoom) return this.errorReply(`Invalid room.`);
```
mia-pi-git, line 355: Not necessary since requireRoom below does that.

server/chat-plugins/poll.ts
```diff
@@ -350,6 +350,14 @@ export const commands: ChatCommands = {
...
+			if (!targetRoom.persist) {
```
mia-pi-git, line 356: Do this below the requireRoom call.

mia-pi-git (changes requested):

Distrib-ps:
Like this?

server/chat-plugins/poll.ts
```diff
@@ -350,6 +350,11 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			const targetRoom = roomid ? Rooms.search(roomid) : room;
+			if (!targetRoom) return this.errorReply(`Invalid room.`);
+			this.room = room = targetRoom;
+			room = targetRoom;
```
mia-pi-git, line 357: You can remove this line.

server/chat-plugins/poll.ts
```diff
@@ -363,7 +367,6 @@ export const commands: ChatCommands = {
 				this.modlog('CLEARQUEUE');
 				this.sendReply(this.tr`Cleared poll queue.`);
 			} else {
-				const [slotString, roomid, update] = target.split(',');
 				const slot = parseInt(slotString);
 				const curRoom = roomid ? (Rooms.search(roomid) as ChatRoom | GameRoom) : room;
 				if (!curRoom) return this.errorReply(this.tr`Room "${roomid}" not found.`);
```
mia-pi-git, line 372: This can be removed. Same opts are done above.

server/chat-plugins/poll.ts
mia-pi-git: This check if targetRoom is falsy is still unnecessary - requireRoom does the same thing below.

Distrib-ps: Yep, Done !

Distrib-ps: Removing the lines, the deleteqeue does not work. @mia-pi-git

mia-pi-git: How so?

mia-pi-git: Oh. you removed too much of the code.
Restore the targetRoom stuff, you should only have removed the `if (!targetRoom)` line.

mia-pi-git:
```ts
if (roomid) {
const targetRoom = Rooms.search(target);
if (targetRoom) {
this.room = targetRoom;
}
}
room = this.requireRoom(); // etc
```

Distrib-ps: Done, thx !

mia-pi-git (approved):
Looks like that's everything from me.

server/chat-plugins/poll.ts
```diff
@@ -350,6 +350,13 @@ export const commands: ChatCommands = {
 
 		clearqueue: 'deletequeue',
 		deletequeue(target, room, user, connection, cmd) {
+			const [slotString, roomid, update] = target.split(',').map(i => i.trim());
+			if (roomid) {
+				const targetRoom = roomid ? Rooms.search(roomid) : room;
+				if (targetRoom) {
+					this.room = targetRoom;
```
AnnikaCodes, line 357: Wouldn't setting `this.room` make it difficult to tell if the command succeeded or failed, since its output would be in a different room than it was used in? Or is this command only used from the poll management page? (I dislike using pages personally, and prefer commands. I don't know if others share this preference.)

mia-pi-git: This is meant for the page, yes.

mia-pi-git: `this.room` has to be set for it to work.

server/chat-plugins/poll.ts
```diff
@@ -363,10 +370,8 @@ export const commands: ChatCommands = {
 				this.modlog('CLEARQUEUE');
 				this.sendReply(this.tr`Cleared poll queue.`);
 			} else {
-				const [slotString, roomid, update] = target.split(',');
 				const slot = parseInt(slotString);
 				const curRoom = roomid ? (Rooms.search(roomid) as ChatRoom | GameRoom) : room;
-				if (!curRoom) return this.errorReply(this.tr`Room "${roomid}" not found.`);
```
Zarel, line 369: Why was this deleted? This will cause a crash if it doesn't exist, won't it?

mia-pi-git: we really shouldn't be using `curRoom` at all, `room` is updated to the new room above.

server/chat-plugins/poll.ts
Zarel: I'm pretty sure `this.room` does not actually need ot be set for it to work.

server/chat-plugins/poll.ts
Zarel: You approved this... :/

PR #7835, opened 2020-12-24
Bug fix: Metal Burst Submove Interaction

aviettran:
> If a move that calls another move calls Metal Burst it should behave exactly like a normal Metal Burst ie. deal 1.5x damage back to the Pokemon that most recently targeted it. See replay
> 
> Added by scheibo

Question: Smogon states that this move can deal 1 damage if the user of the move did not take damage from an attacker. Does this include all attacks, including failures or non-damaging moves?

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10885,7 +10885,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		basePower: 0,
 		damageCallback(pokemon) {
 			const lastDamagedBy = pokemon.getLastDamagedBy();
-			if (lastDamagedBy) {
+			if (lastDamagedBy !== undefined) {
 				return (lastDamagedBy.damage * 1.5) || 1;
 			}
 			return 0;
@@ -10897,7 +10897,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {protect: 1, mirror: 1},
 		onTryHit(target, source, move) {
 			const lastDamagedBy = source.getLastDamagedBy();
-			if (!lastDamagedBy) return false;
+			if (lastDamagedBy === undefined) return false;
 		},
 		onModifyTarget(targetRelayVar, source, target, move) {
 			const lastDamagedBy = source.getLastDamagedBy();
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10886,7 +10886,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		damageCallback(pokemon) {
 			const lastDamagedBy = pokemon.getLastDamagedBy();
 			if (lastDamagedBy !== undefined) {
-				return (lastDamagedBy.damage * 1.5) || 1;
+				return (lastDamagedBy.damage * 1.5);
 			}
 			return 0;
 		},
@@ -10897,7 +10897,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {protect: 1, mirror: 1},
 		onTryHit(target, source, move) {
 			const lastDamagedBy = source.getLastDamagedBy();
-			if (lastDamagedBy === undefined) return false;
+			if (lastDamagedBy === undefined || !lastDamagedBy.thisTurn) return false;
 		},
 		onModifyTarget(targetRelayVar, source, target, move) {
 			const lastDamagedBy = source.getLastDamagedBy();
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10886,7 +10886,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		damageCallback(pokemon) {
 			const lastDamagedBy = pokemon.getLastDamagedBy();
 			if (lastDamagedBy !== undefined) {
-				return (lastDamagedBy.damage * 1.5);
+				return (lastDamagedBy.damage * 1.5) || 1;
 			}
 			return 0;
 		},
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10902,7 +10902,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		onModifyTarget(targetRelayVar, source, target, move) {
 			const lastDamagedBy = source.getLastDamagedBy();
 			if (lastDamagedBy !== undefined) {
-				targetRelayVar.target = lastDamagedBy.source;
+				targetRelayVar.target = source.side.foe.active[lastDamagedBy.position];
 			}
 		},
 		secondary: null,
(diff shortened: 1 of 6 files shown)


DaWoblefet:
> Question: Smogon states that this move can deal 1 damage if the user of the move did not take damage from an attacker. Does this include all attacks, including failures or non-damaging moves?

I think this is mistaken. Do you have the quote? I thought it might be for situations where the user took 0 damage (e.g. False Swipe at 1 HP, Sturdy Shedinja), but that just causes the move to "But it failed!" as well. Generally speaking, if nothing hit the Metal Burst user, the move should fail.

aviettran:
> Deals damage to the last opposing Pokemon to hit the user with an attack this turn equal to 1.5 times the HP lost by the user from that attack, rounded down. If the user did not lose HP from the attack, this move deals 1 HP of damage instead. If that opposing Pokemon's position is no longer in use and there is another opposing Pokemon on the field, the damage is done to it instead. Only the last hit of a multi-hit attack is counted. Fails if the user was not hit by an opposing Pokemon's attack this turn.

DaWoblefet:
> I think this is mistaken. Do you have the quote? I thought it might be for situations where the user took 0 damage (e.g. False Swipe at 1 HP, Sturdy Shedinja), but that just causes the move to "But it failed!" as well. Generally speaking, if nothing hit the Metal Burst user, the move should fail.

Oops, looks like I messed up and was originally correct - it only deals 1 damage in scenarios like Sturdy Shedinja or False Swipe at 1 HP.  Turns out Metal Burst / Counter / Mirror Coat pass over ally damage. So you can't Metal Burst a hit from your ally. Testing with the opponent hitting Sturdy Shedinja, though, Metal Burst does deal 1 damage in that situation. So the description is simply missing that it doesn't work on allies.

Marty-D:
It says pretty clearly "opposing Pokemon" on the first and last lines.

Marty-D:
You'll also notice Gen 4 is different from Gen 5 and 6, which is different from Gen 7 and 8
https://www.smogon.com/dex/dp/moves/metal-burst/

test/sim/moves/metalburst.js
```diff
@@ -0,0 +1,40 @@
...
+		const breloomHpTurn1 = battle.p2.active[0].hp;
+		assert.equal(breloomHpTurn1, battle.p2.active[0].maxhp - battle.dex.getMove('sonicboom').damage * 1.5);
+		battle.makeChoices('move sleeptalk', 'move spore');
+		battle.makeChoices('move sleeptalk', 'move sonicboom');
+		assert.equal(battle.p2.active[0].hp, breloomHpTurn1 - battle.dex.getMove('sonicboom').damage * 1.5);
+	});
+
+	it('should target the opposing Pokemon that hit the user with an attack most recently that turn', function () {
+		// The seed should select venusaur if the test would otherwise fail
+		battle = common.createBattle({gameType: 'doubles', seed: [3, 4, 5, 6]});
```
DaWoblefet, line 27: This shouldn't need an RNG seed. Metal Burst doesn't randomly target.

test/sim/moves/metalburst.js
```diff
@@ -0,0 +1,40 @@
...
+
+let battle;
+
+describe('Metal Burst', function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it('should run conditions for submove', function () {
+		battle = common.createBattle({seed: [1, 2, 3, 4]}); // Snorlax needs to sleep for more than 1 turn
```
DaWoblefet, line 14: Swap this to doubles so it doesn't need an RNG seed.

test/sim/moves/metalburst.js
aviettran: You're right, it shouldn't. This seed is so that if the code is _incorrect_, the test will actually fail instead of passing as a false negative.

aviettran:
This PR is lacking the 1 damage logic from the conversation above. Does anyone know a good way to determine whether a move should have dealt damage? I'm not sure if the `onDamagingHit` event used for Counter/Mirror Move accounts for this. I haven't written a test against it but it looks like there are several `damage !== 0` checks in `scripts.ts`.

test/sim/moves/metalburst.js
aviettran: Complete.

test/sim/moves/metalburst.js
DaWoblefet: Gotcha; if it _were_ random, it _would_ target Venusaur. Looks fine then.

Marty-D:
I believe all those `damage !== 0` checks are just to differentiate it from other falsy values like `false` and `null`. 0 damage has to go through every step correctly because it does actually count as a hit for the purposes of secondary effects of the attack, or effects like Rough Skin.

aviettran:
Updates:

- Fixed target by location (not by Pokemon) in case of a switch
- Fixed minimum 1 damage when taking a damaging hit

DaWoblefet:
This looks good to me.

data/moves.ts
```diff
@@ -10884,39 +10884,26 @@ export const Moves: {[moveid: string]: MoveData} = {
...
-				this.effectData.position = null;
-				this.effectData.damage = 0;
-			},
-			onRedirectTargetPriority: -1,
-			onRedirectTarget(target, source, source2) {
-				if (source !== this.effectData.target) return;
-				return source.side.foe.active[this.effectData.position];
-			},
-			onDamagingHit(damage, target, source, effect) {
-				if (source.side !== target.side) {
```
Marty-D, line 10915: This is an important check that's gone missing; damage taken from your ally shouldn't be considered at all for counter moves not named Bide.

aviettran: Added. Please review.

Marty-D:
I was going to suggest exactly what you did in your readability change, excellent! 😄

MacChaeger:
I wish I had asked this sooner but will the `onRedirectTarget` -> `onModifyTarget` affect how metal burst interacts with Follow Me, Lightning Rod, Stalwart, etc.?

Marty-D:
Theoretically no, since it's placed before getMoveTargets.

MacChaeger:
Metal Burst originally had a negative `onRedirectTargetPriority` so if the ‘ModifyTarget` event is before the `RedirectTarget` event then this would change the order they happen in.

Marty-D:
Seems fine https://github.com/smogon/pokemon-showdown/commit/1b046208ba5346ca74f9eeb7c940bdef770e8a56

PR #7836, opened 2020-12-24
Refactor /updateserver

AnnikaCodes:
Now, `/updateserver` updates both public and private code, and `/updateserver private` only updates private code.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -902,7 +902,9 @@ export const commands: ChatCommands = {
 			success = await updateserver(this, Config.privatecodepath);
 			this.addGlobalModAction(`${user.name} used /updateserver private`);
 		} else {
-			if (validPrivateCodePath) success = await updateserver(this, Config.privatecodepath);
+			if (target !== 'public' && validPrivateCodePath) {
+				success = await updateserver(this, Config.privatecodepath);
+			}
 			success = success && await updateserver(this, path.resolve(`${__dirname}/../..`));
 			this.addGlobalModAction(`${user.name} used /updateserver`);
 		}
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -31,6 +31,64 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
 	});
 }
 
+async function updateserver(context: CommandContext, codePath: string) {
+	const exec = (command: string) => bash(command, context, codePath);
+
+	context.sendReply(`Fetching newest version of code in the repository ${codePath}...`);
+
+	let [code, stdout, stderr] = await exec(`git fetch`);
+	if (code) throw new Error(`updateserver: Crash while fetching - make sure this is a Git repository`);
+	if (!stdout && !stderr) {
+		context.sendReply(`There were no updates.`);
+		Monitor.updateServerLock = false;
+		return;
+	}
+
+	[code, stdout, stderr] = await exec(`git rev-parse HEAD`);
+	if (code || stderr) throw new Error(`updateserver: Crash while grabbing hash`);
+	const oldHash = String(stdout).trim();
+
+	[code, stdout, stderr] = await exec(`git stash save "PS /updateserver autostash"`);
+	let stashedChanges = true;
+	if (code) throw new Error(`updateserver: Crash while stashing`);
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -824,78 +882,36 @@ export const commands: ChatCommands = {
...
-				if (code) {
-					// conflict while popping stash
-					await exec(`git reset HEAD .`);
-					await exec(`git checkout .`);
-					throw new Error(`restore`);
-				}
-			}
+		this.sendReply(`Rebuilding...`);
+		await rebuild(this);
+		this.sendReply(`DONE: code is up to date and rebuilt!`);
```
Zarel, line 907:
```suggestion
		this.sendReply(`DONE`);
```

I know the old message wasn't as terse, but I think I do prefer it this way; I don't think it says anything important that wasn't already said (especially since it doesn't even mention if it updated public/private/both).

AnnikaCodes: I think anyone who has the permissions to use this command on Main can just read the code anyway, so I don't really care how terse it is.

server/chat-commands/admin.ts
```diff
@@ -31,6 +31,64 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
...
+			[code] = await exec(`git stash pop`);
+			if (code) {
+				// conflict while popping stash
+				await exec(`git reset HEAD .`);
+				await exec(`git checkout .`);
+				throw new Error(`restore`);
+			}
+		}
+
+		context.sendReply(`SUCCESSFUL, server updated.`);
```
Zarel, line 83: This seems redundant with the `DONE` at the end of the command.

Zarel:
```suggestion
```

server/chat-commands/admin.ts
```diff
@@ -824,78 +882,36 @@ export const commands: ChatCommands = {
...
-				if (code) {
-					// conflict while popping stash
-					await exec(`git reset HEAD .`);
-					await exec(`git checkout .`);
-					throw new Error(`restore`);
-				}
-			}
+		this.sendReply(`Rebuilding...`);
+		await rebuild(this);
+		this.sendReply(`DONE`);
```
Zarel, line 907: It seems to output `DONE` whether or not the updateserver succeeded. I think `updateserver` should probably return success state, to be output here.

server/chat-commands/admin.ts
```diff
@@ -31,6 +31,64 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
...
+				throw new Error(`restore`);
+			}
+		}
+
+		context.sendReply(`SUCCESSFUL, server updated.`);
+	} catch (e) {
+		// failed while rebasing or popping the stash
+		await exec(`git reset --hard ${oldHash}`);
+		if (stashedChanges) await exec(`git stash pop`);
+		context.sendReply(`FAILED, old changes restored.`);
```
Zarel, line 88:
```suggestion
		return true;
	} catch (e) {
		// failed while rebasing or popping the stash
		await exec(`git reset --hard ${oldHash}`);
		if (stashedChanges) await exec(`git stash pop`);
		return false;
```

This is how I would implement this: so that the final `DONE` or `FAILED, old changes restored.` is sent after the rebuild.

server/chat-commands/admin.ts
```diff
@@ -824,78 +885,38 @@ export const commands: ChatCommands = {
...
+		let success = true;
+		if (target === 'private') {
+			if (!validPrivateCodePath) {
+				throw new Chat.ErrorMessage("`Config.privatecodepath` must be set to an absolute path before using /updateserver private.");
+			}
+			success = await updateserver(this, Config.privatecodepath);
+			this.addGlobalModAction(`${user.name} used /updateserver private`);
 		} else {
-			this.sendReply(`Saving changes...`);
+			if (validPrivateCodePath) success = await updateserver(this, Config.privatecodepath);
```
Zarel, line 905:
```suggestion
			if (target !== 'public' && validPrivateCodePath) {
				success = await updateserver(this, Config.privatecodepath);
			}
```

Zarel (approved):
Other than `/updateserver public`, I think this is done. Good job!

Zarel:
Thanks!

PR #7844, opened 2020-12-28
Improve /help youtube

Distrib-ps:
The Poll and dc have been improved. I did the same for Youtube.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -573,25 +573,27 @@ export const commands: ChatCommands = {
 			game.start();
 		},
 	},
-	youtubehelp: [
-		`YouTube commands:`,
-		`/randchannel [optional category]- View data of a random channel from the YouTube database.` +
-			` If a category is given, the random channel will be in the  given category.`,
-		`/youtube addchannel [channel] - Add channel data to the YouTube database. Requires: % @ #`,
-		`/youtube removechannel [channel]- Delete channel data from the YouTube database. Requires: % @ #`,
-		`/youtube channel [channel] - View the data of a specified channel. Can be either channel ID or channel name.`,
-		`/youtube video [video] - View data of a specified video. Can be either channel ID or channel name.`,
-		`/youtube update [channel], [name] - sets a channel's PS username to [name]. Requires: % @ #`,
-		`/youtube repeat [time] - Sets an interval for [time] minutes, showing a random channel each time. Requires: # &`,
-		`/youtube addcategory [name] - Adds the [category] to the channel category list. Requires: @ # &`,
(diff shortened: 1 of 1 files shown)


server/chat-plugins/youtube.ts
```diff
@@ -573,25 +573,27 @@ export const commands: ChatCommands = {
...
-		`/youtube decategorize [channel name] - Removes the category for the [channel], if there is one. Requires: @ # &`,
-		`/youtube categores - View all channels sorted by category.`,
-		`/youtube groupwatch [link], [title] - Creates a group watch of the [url] with the given [title]. Requires % @ & #`,
-		`/youtube startwatch - Starts the group watch in the current room, if there is one. Requires % @ & #`,
-		`/youtube stopwatch - Ends the current group watch, if there is one in the current room. Requires % @ & #`,
-	],
+	youtubehelp() {
+		const help = [
+			`<code>/randchannel [optional category]</code> - View data of a random channel from the YouTube database.` +
+				` If a category is given, the random channel will be in the  given category.`,
```
AnnikaCodes, line 579:
```suggestion
			`<code>/randchannel [optional category]</code> - View data of a random channel from the YouTube database. If a category is given, the random channel will be in the  given category.`,
```

server/chat-plugins/youtube.ts
```diff
@@ -573,25 +573,27 @@ export const commands: ChatCommands = {
...
+			`<code>/youtube addchannel [channel]</code> - Add channel data to the YouTube database. Requires: % @ #`,
+			`<code>/youtube removechannel [channel]</code>- Delete channel data from the YouTube database. Requires: % @ #</summary>` +
+			`<code>/youtube channel [channel]</code> - View the data of a specified channel. Can be either channel ID or channel name.`,
```
AnnikaCodes, line 582:
```suggestion
			`<code>/youtube removechannel [channel]</code> - Delete channel data from the YouTube database. Requires: % @ #</summary>`,
			`<code>/youtube channel [channel]</code> - View the data of a specified channel. Can be either channel ID or channel name.`,
```

mia-pi-git:
Is there a gain for this?
Code formatting makes it almost harder to read, as well as the fact this adds more bandwidth and size to the help command when it doesn't need it.
I'll leave this for a bit but personally I would deny this without a better argument.

Distrib-ps:
> Is there a gain for this?
> Code / readmore formatting makes it almost harder to read, as well as the fact this adds more bandwidth and size to the help command when it doesn't need it.
> This isn't a command that gets broadcasted much or ever (only YT staff need it) so there's not much gain to the readmore other than more inconvenience.
> I'll leave this for a bit but personally I would deny this without a better argument.


I just wanted to do like the other help command.
Yes, youtube staff only use it.
So I close this pr.

PR #7846, opened 2020-12-28
Throttle Rooms.global.writeChatRoomData

mia-pi-git:
See #sim-dev.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1034,6 +1034,7 @@ export class GlobalRoomState {
 	maxUsers: number;
 	maxUsersDate: number;
 	formatList: string;
+	writingData?: boolean;
 
 	constructor() {
 		this.settingsList = [];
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -1034,6 +1034,7 @@ export class GlobalRoomState {
 	maxUsers: number;
 	maxUsersDate: number;
 	formatList: string;
+	writingData?: boolean;
```
Zarel, line 1037:
```suggestion
	pendingDataWrite: NodeJS.Timer | null = null;
```

the rest is left as an exercise for the reader

Zarel:
I think we ended up agreeing to replace this inside `FS`.

mia-pi-git:
Yeah, as i'm rewriting this throttle inside FS there's no point to this PR.

PR #7847, opened 2020-12-28
Allow alternative parameters for dexsearches

tmagicturtle:
Maybe I'm not understanding something or there's some non-obvious reason for it, but with the new pivot search, I didn't really understand why it had special coding to allow only "batonpass" (and not "baton pass" or other search options). I especially didn't understand why it didn't just use the normal "or" search (e.g. flash fire|baton pass). This got me wondering about the other searches disallowing alternatives. Surely there's a use for a search like "recovery|regenerator" and I don't really see why it specifically disallowed alternatives.

The only other thing I could not figure out was allowing alternatives for mega, fully evolved, and gmax. Surely there's a use for "mega|gmax" or "fully evolved|def>100" or similar. Again, unless there's a reason I'm not aware of, I think these are artificial limitations on the dexsearch function and should be removed.


server/chat-plugins/datasearch.ts
```diff
@@ -806,17 +800,9 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
...
-				if (parameters.length > 1) {
-					if (parameters[1].trim().toLowerCase() === 'batonpass') {
-						includeBatonPass = true;
-					} else {
-						return {error: "The parameter 'pivot' cannot have alternative parameters other than 'batonpass'."};
-					}
-				}
 				for (const move in Dex.data.Moves) {
 					const moveData = Dex.getMove(move);
-					if (moveData.selfSwitch && (includeBatonPass || moveData.id !== 'batonpass')) {
```
KrisXV, line 819: why are you removing the baton pass parameter?

tmagicturtle: That's the way the pivot parameter already works. If we're going to include baton pass that's fine, but the decision to not include it wasn't mine.

tmagicturtle: The reasoning at https://www.smogon.com/forums/threads/make-pivot-moves-a-ms-category.3672772/ is that most formats ban Baton Pass, so if BP was included, searching by pivot would have little value.

tmagicturtle:
Just checking if there's any additional thoughts on this, it's been some time.

tmagicturtle:
I am asking once again for your review and support.

AnnikaCodes (approved):
This seems fine to me.

AnnikaCodes:
You have to fix the merge conflicts, though.

server/chat-plugins/datasearch.ts
```diff
@@ -836,17 +830,9 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
...
-					if (parameters[1].trim().toLowerCase() === 'batonpass') {
-						includeBatonPass = true;
-					} else {
-						return {error: "The parameter 'pivot' cannot have alternative parameters other than 'batonpass'."};
-					}
-				}
 				for (const move in mod.data.Moves) {
-					const moveData = mod.moves.get(move);
-					if (moveData.selfSwitch && (includeBatonPass || moveData.id !== 'batonpass')) {
+					const moveData = mod.getMove(move);
```
Distrib-ps, line 834:
```suggestion
					const moveData = mod.moves.get(move);
```

AnnikaCodes:
Sorry I missed this, it looks good. Thanks!

PR #7848, opened 2020-12-29
Youtube: Modify uses of youtube tag to comply with API ToS

mia-pi-git:
This is a bandage solution. 
Ideally, we will "revert" this once we can find a suitable solution with the tag that complies with Google's API ToS.


AnnikaCodes (approved):
I'm not too familiar with the YouTube code but this seems to be for an urgent ToS compliance issue and the fix looks good to me.

PR #7853, opened 2020-12-29
Fix move-disabling mechanics for dynamaxed Pokemon

MacChaeger:
To sum up what I've seen from research, it should work like this:
- Taunt and Assault Vest disable Max Guard
- No Max Move can be used if the base move is out of PP
- No other effects can individually disable Max Moves, but if each and every one of a dynamaxed Pokemon's base moves is disabled by way of being out of PP, or by the effects of Encore, Disable, Taunt, Assault Vest, or Belch or Stuff Cheeks disabling themselves, then the Pokemon will be forced to use Struggle anyway

I removed the no-PP hint because I think this is at least the third time it's been wrong, and Metronome Battle players have been seeing that message every time they dynamax at 1 PP left to win Struggle wars, so when that doesn't work anymore you can bet there will be at least irate post in the bugs thread about it.

### DIFF
--- a/data/scripts.ts
+++ b/data/scripts.ts
@@ -72,10 +72,6 @@ export const Scripts: BattleScriptsData = {
 			if (!lockedMove) {
 				if (!pokemon.deductPP(baseMove, null, target) && (move.id !== 'struggle')) {
 					this.add('cant', pokemon, 'nopp', move);
-					const gameConsole = [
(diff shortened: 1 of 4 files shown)


DaWoblefet:
Tests all look good, research summary is accurate too.

DaWoblefet (approved):

Marty-D (approved):
Oh yeah, I reviewed this earlier and forgot to merge when I got back to my desktop. 👍

PR #7854, opened 2020-12-30
Add horizontal scrollbar for !code

sirKiraUzumaki:
P5f29EtVnJho
When you put more than the limited characters in !code it overflows out the box and then you just can't read that.
By adding infobox-limited it will add a horizontal scrolling bar.

And sorry for those creepy lines.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2713,7 +2713,7 @@ export const commands: ChatCommands = {
 				target = this.filter(target)!;
 				if (!target) return this.errorReply(`Invalid code.`);
 			}
-			return `/raw <div class="infobox">${Chat.getReadmoreCodeBlock(target)}</div>`;
+			return `/raw <div class="infobox"><div class="infobox-limited">${Chat.getReadmoreCodeBlock(target)}</div></div>`;
 		} else {
 			this.sendReplyBox(Chat.getReadmoreCodeBlock(target));
 		}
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.ts
```diff
@@ -2713,7 +2713,7 @@ export const commands: ChatCommands = {
 				target = this.filter(target)!;
 				if (!target) return this.errorReply(`Invalid code.`);
 			}
-			return `/raw <div class="infobox">${Chat.getReadmoreCodeBlock(target)}</div>`;
+			return `/raw <div class="infobox"><div class="infobox-limited">${Chat.getReadmoreCodeBlock(target)}</div></div>`;
```
mia-pi-git, line 2716:
```suggestion
			return `/raw <div class="infobox-limited">${Chat.getReadmoreCodeBlock(target)}</div>`;
```

doing this just creates two boxes.

mia-pi-git: Also, how do you plan to address it when it's not broadcasting (using sendReplyBox)?

sirKiraUzumaki: It won't create two boxes, I have checked it.

Zarel:
This should auto-wrap like all other chat text. The correct fix is probably in client CSS somewhere.

Zarel:
OH IT'S BECAUSE OF THE DISPLAY:TABLE HACK

Zarel:
So I ended up doing a really dumb hack in Scavengers:

https://github.com/smogon/pokemon-showdown/blob/674eb1449f38d6852faf8687eec119122566834e/server/chat-plugins/scavengers.ts#L660-L673

It's probably better to just move this to `Chat` than to show a horizontal scrollbar.

Zarel:
(I'll do this)

PR #7855, opened 2020-12-30
Add multi-hit as a parameter to /ms

PartMan7:


PartMan7:
Oh, I messed that up, one sec...

PartMan7:
*bump*

Zarel:
Movesearch is still a huge mess which we should really fix at some point... @_@

Zarel:
Thanks!

PR #7860, opened 2020-12-31
Bug Fix: Neutralizing Gas Ending Issues

aviettran:
> Neutralizing Gas wearing off should cause entrance Abilities to reactivate in order of Speed

> Neutralizing Gas should not wear off twice after suppression

neutralizinggas test from #7857

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2312,8 +2312,14 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			// (If your tackling this, do note extreme weathers have the same issue)
 
 			// Mark this pokemon's ability as ending so Pokemon#ignoringAbility skips it
-			source.abilityData.ending = true;
-			for (const pokemon of this.getAllActive()) {
+			if (source.abilityData.ending) {
+				return;
+			} else {
+				source.abilityData.ending = true;
+			}
+			const sortedActive = this.getAllActive();
+			this.speedSort(sortedActive);
+			for (const pokemon of sortedActive) {
 				if (pokemon !== source) {
 					// Will be suppressed by Pokemon#ignoringAbility if needed
 					this.singleEvent('Start', pokemon.getAbility(), pokemon.abilityData, pokemon);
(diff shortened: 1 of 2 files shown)


DaWoblefet:
I'm going to test the following just to make sure:
- Check if it uses raw / summary screen Speed or Speed with modifiers (I expect Speed with modifiers)
- Check if non-entrance Abilities also activate in order of Speed. E.g. would the Speed of two Defiant & Intimidate Pokemon determine if Defiant gets an Attack boost? (I expect no)
- Does Unnerve still have priority? (I expect yes)

These would be because I expect whatever event we run for Abilities coming in naturally (e.g. the start of the game, or two switchins after a double KO) is what Neutralizing Gas actually does. I'll double check and get back to you.

DaWoblefet:
Good thing I tested! My assumption was wrong about Unnerve - it doesn't have priority on reactivation, unlike regular switchins. Neutralizing Gas does use Speed sorting with modifiers (e.g. Choice Scarf is included). It also doesn't turn on the abilities one at a time. It's more like just entrance abilities are turned on one at a time. For example, if I have fast Intimidate vs. slow Defiant, Intimidate will still trigger Defiant and give a +1 boost.

So I think the direct call to Speed sort like you've done is actually preferable here. I will add a test for that sort of Defiant / Intimidate behavior to ensure nothing's wrong there.

DaWoblefet:
Ugh, I messed up a push. Sorry about that xfix, you can ignore.

DaWoblefet:
On topic, I added Defiant and Unnerve tests. Of note is that I had to skip the Unnerve test because apparently PS has another bug with Unnerve and Neutralizing Gas. On cart, if Unnerve was active, then negated by Neutralizing Gas, then Neutralizing Gas leaves, Unnerve doesn't visibly reactivate - it does negate Berries and all, it just doesn't display the message. But if Unnerve never activated initially and Neutralizing Gas lets it start, then it does display a message. I have two tests for that behavior.

DaWoblefet (approved):
I'm ok leaving the Unnerve issue unsolved for now. This is a pretty relevant fix for Doubles play so I'm very happy with it.

Marty-D (approved):
👍

PR #7861, opened 2020-12-31
Remove 1 extra box from YT Channel Display Generator

sirKiraUzumaki:
Idk if 2 boxes were added intentionally or it was a mistake.

### DIFF
--- a/server/chat-plugins/youtube.ts
+++ b/server/chat-plugins/youtube.ts
@@ -106,7 +106,7 @@ export class YoutubeInterface {
 		const id = this.getId(link);
 		const {name, description, icon, videos, subs, views, username} = await this.get(id);
 		// credits bumbadadabum for most of the html
-		let buf = `<div class="infobox"><table style="margin:0px;"><tr>`;
+		let buf = `<table style="margin:0px;"><tr>`;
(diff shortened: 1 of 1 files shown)


mia-pi-git (dismissed):
Box is necessary for /yt channels all. 
The proper solution would be to not use sendReplyBox in most of the commands, but instead `this.sendReply('|html|....)`.

server/chat-plugins/youtube.ts
```diff
@@ -282,7 +282,7 @@ export class YoutubeInterface {
 				const room = Rooms.get('youtube');
 				if (!room) return; // do nothing if the room doesn't exist anymore
 				const res = await YouTube.randChannel();
-				room.add(`|html|<div class="infobox">${res}</div>`).update();
+				room.add(`|html|${res}</div>`).update();
```
Zarel, line 285: That `</div>` seems extraneous?

Zarel:
```suggestion
				room.add(`|html|${res}`).update();
```

Zarel (approved):
Looks good other than the extraneous `</div>`

server/chat-plugins/youtube.ts
sirKiraUzumaki: Oh yea, thanks!

Zarel:
Thanks!

PR #7862, opened 2020-12-31
Mafia: Use errorReply and requireGame more

zgeorgecsat:


iscke:
looks good, thanks!

while you're here, can you refactor `MafiaTracker#canJoin` to use `Chat.ErrorMessage`?

server/chat-plugins/mafia.ts
```diff
@@ -2252,8 +2242,8 @@ export const commands: ChatCommands = {
 			} else {
 				args.shift();
 			}
-			const game = targetRoom.getGame(MafiaTracker);
-			if (!game) return user.sendTo(targetRoom, `|error|There is no game of mafia running in this room.`);
+			this.room = targetRoom;
```
Zarel, line 2245: This is a bit awkward. It would mess up the response text functions like `errorReply`.

Zarel: Maybe drop support from using the function in other rooms? Or support passing a room to `requireGame`?

zgeorgecsat: If I'm understanding correctly, you're concerned that errorReply would send to targetRoom instead of the original room because `this.room = targetRoom` set it that way. I believe that's the goal, since targetRoom is the room where the mafia game is

Zarel: Hm, okay. Is this for a clickable button somewhere? That makes sense, then.

Zarel:
Thanks!

PR #7864, opened 2021-01-01
Update translations

palt0:
Apologies for the generic title. Opening 2 separate PRs seemed spammy, but I divided the changes in 2 commits just in case you prefer it that way.
1. Add Italian translations for non-time-based repeats and a phrase we skipped.
2. Add missing phrases to the reference English translations. (On a side note, the Prize Winner phrase is the only `/groups` string that ends with a dot.)


Zarel:
@AnnikaCodes? I think you would have more of an idea of what to do with the missing phrases than I would.

AnnikaCodes:
They should be added to all of the actively translated `main.ts` files. I don't think it's necessary to copy files into the other language directories if they're completely untranslated, but if there's an effort to translate the particular file then the phrases should be consistent across all languages (to make it easier for translators to see what they need to do).

palt0:
Sure.

AnnikaCodes:
Thanks!

PR #7865, opened 2021-01-01
Repeats: Fix private mod actions

palt0:
`/repeatfaq` works as expected but the private mod actions that it sends depending on whether a repeat is set by minutes or by messages should be swapped.


AnnikaCodes (approved):
Good catch!

PR #7866, opened 2021-01-01
Make Unnerve activate after Skill Swap and Neutralising Gas wearing off

urkerab:
As recently mentioned by @DaWoblefet, if Unnerve has not yet activated, then it should activate when Neutralising Gas wears off.

It should also activate when Skill Swapped.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -4068,6 +4068,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	unnerve: {
 		onPreStart(pokemon) {
 			this.add('-ability', pokemon, 'Unnerve', pokemon.side.foe);
+			this.effectData.unnerved = true;
+		},
+		onStart(pokemon) {
+			if (this.effectData.unnerved) return;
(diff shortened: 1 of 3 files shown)


DaWoblefet:
This got discussed in https://github.com/smogon/pokemon-showdown/pull/7860, where I added a couple tests describing the behavior. Unnerve only announces after Neutralizing Gas ends _if_ it had never gone off in the first place.

On cart, if Unnerve was active, then negated by Neutralizing Gas, then Neutralizing Gas leaves, Unnerve doesn't visibly reactivate - it does negate Berries and all, it just doesn't display the message. But if Unnerve never activated initially and Neutralizing Gas lets it start, then it does display a message. I have two tests for that behavior.

urkerab:
If I paste the last test from that PR and change the `if.skip` to `if.only` then the test passes. I've also checked and the first test from that PR still passes after this change.

DaWoblefet:
Viett's PR got merged, so if you could rebase and unskip that test in this PR, we should be good!

urkerab:
Test unskipped.

DaWoblefet (approved):

PR #7867, opened 2021-01-02
Mafia: Minor script fixes

zgeorgecsat:
- If there was a plurality shift upon someone getting hammered, the lynch count shown now reflects that plurality shift.
- The night responses dropdown was removed from the mafia page as it was clunky and unnecessary.
- If a user tries to use an argument with `/mafia action`, they will get an error reply.
- Cohosts are now listed under the host on the mafia page.
- `/mafia players` now gives an alphabetized player list.
- An announcement in room is made when Closed Setup is turned on.
- Multiple people can be killed at the same time with `/mafia kill`.


server/chat-plugins/mafia.ts
```diff
@@ -3510,52 +3534,83 @@ export const commands: ChatCommands = {
...
-
-			const [targetUser, durationString] = this.splitOne(target);
-			const targetUserID = toID(targetUser);
-			const duration = parseInt(durationString);
+			this.checkCan('warn', null, room);
+
+			target = this.splitTarget(target, false);
+			const [string1, string2] = this.splitOne(target);
+			let duration, reason;
+			if (parseInt(string1)) {
```
iscke, line 3546: the flexible arg order here seems unnecessary

server/chat-plugins/mafia.ts
```diff
@@ -3510,52 +3534,83 @@ export const commands: ChatCommands = {
...
+				}
 			}
 
-			if (isUnban) {
-				delete hostBans[targetUserID];
-				this.modlog(`MAFIAUNHOSTBAN`, null, `${targetUserID}`);
-			} else {
-				if (isNaN(duration) || duration < 1) return this.parse('/help mafia hostban');
-				if (duration > 7) return this.errorReply(`Bans cannot be longer than 7 days.`);
+			if (cmd === 'hostban') MafiaTracker.hostBan(room, targetUser, reason, duration);
```
iscke, line 3577: nit, this should probably have brackets

server/chat-plugins/mafia.ts
```diff
@@ -2679,11 +2695,16 @@ export const commands: ChatCommands = {
...
 			this.room = targetRoom;
 			const game = this.requireGame(MafiaTracker);
-			if (game.hostid !== user.id && !game.cohosts.includes(user.id)) this.checkCan('mute', null, targetRoom);
+			if (game.hostid !== user.id && !game.cohostids.includes(user.id)) this.checkCan('mute', null, targetRoom);
 			if (!toID(args.join(''))) return this.parse('/help mafia revive');
 			let didSomething = false;
 			for (const targetUsername of args) {
-				if (game.revive(user, toID(targetUsername), cmd === 'forceadd')) didSomething = true;
+				const targetUser = Users.get(toID(targetUsername));
+				if (targetUser && MafiaTracker.isGameBanned(targetUser)) {
```
iscke, line 2703: this check (and the one in join) should probably be moved into `canJoin`

server/chat-plugins/mafia.ts
```diff
@@ -2566,14 +2572,20 @@ export const commands: ChatCommands = {
...
-			game.eliminate(player, cmd);
-			game.logAction(user, `killed ${player.name}`);
+			if (!args[0]) return this.parse('/help mafia kill');
+			for (const targetUsername of args) {
+				const player = game.playerTable[toID(targetUsername)];
+				if (player) {
+					game.eliminate(player, cmd);
+					game.logAction(user, `killed ${player.name}`);
+				} else {
+					user.sendTo(this.room, `|error|${targetUsername} is not a living player.`);
```
iscke, line 2586: nit, this can just be `this.errorReply` now (because we set `this.room` to be the proper room)

server/chat-plugins/mafia.ts
iscke: also is it intentional that gamebans are global, while hostbans only apply in the mafia room?

server/chat-plugins/mafia.ts
```diff
@@ -106,7 +101,9 @@ const VALID_IMAGES = [
 
 let MafiaData: MafiaData = Object.create(null);
 let logs: MafiaLog = {leaderboard: {}, mvps: {}, hosts: {}, plays: {}, leavers: {}};
-let hostBans: MafiaHostBans = Object.create(null);
+
+Punishments.roomPunishmentTypes.set('GAMEBAN', 'banned from playing games');
```
iscke, line 105: i feel like these should probably be namespaced (MAFIAGAMEBAN / MAFIAHOSTBAN), and the description should probably mention that they're mafia games specifically

AnnikaCodes (commented):
I'm not at all involved with Mafia development but since my review has been requested I'll say that at a glance this change looks fine; Hoeen and peach are the relevant Mafia script coders.

server/chat-plugins/mafia.ts
```diff
@@ -355,9 +334,9 @@ class MafiaTracker extends Rooms.RoomGame {
 		this.sendHTML(this.roomWindow());
 	}
 
-	join(user: User) {
+	join(room: Room, user: User) {
```
iscke, line 337: this can just be `this.room`

server/chat-plugins/mafia.ts
```diff
@@ -383,6 +362,30 @@ class MafiaTracker extends Rooms.RoomGame {
...
+	static gameBan(room: Room, user: User, reason: string, duration: number) {
+		Punishments.roomPunish(room, user, ['MAFIAGAMEBAN', toID(user), Date.now() + (duration * 24 * 60 * 60 * 1000), reason]);
+	}
+
+	static ungameBan(room: Room, user: User) {
+		Punishments.roomUnpunish(room, toID(user), 'MAFIAGAMEBAN', false);
+	}
+
+	static isHostBanned(room: Room, user: User) {
+		return Punishments.getRoomPunishType(room, toID(user)) === 'MAFIAHOSTBAN' || this.isGameBanned(room, user);
```
iscke, line 378: nit, you should probably just get the punishment once and check it against hostban/gameban in this function

server/chat-plugins/mafia.ts
```diff
@@ -1049,7 +1052,7 @@ class MafiaTracker extends Rooms.RoomGame {
 		this.updatePlayers();
 	}
 
-	revive(user: User, toRevive: string, force = false) {
+	revive(room: Room, user: User, toRevive: string, force = false) {
```
iscke, line 1055: just use `this.room`

server/chat-plugins/mafia.ts
```diff
@@ -1517,25 +1520,27 @@ class MafiaTracker extends Rooms.RoomGame {
 		return output;
 	}
 
-	canJoin(user: User, self = false, force = false) {
+	canJoin(room: Room, user: User, self = false, force = false) {
```
iscke, line 1523: ditto

iscke:
looks good, thanks!

PR #7869, opened 2021-01-03
Added emotes like how you can do in discord

harbar20:
Like I can do `:smile:` instead of copy-pasting the smile unicode thing. 

I literally copy-pasted this directly from [here](https://github.com/Jackinev/Bully-Showdown-Server) so it's 99% likely there are lots of issues. Also, in emoticons.json, there's lots of unnecessary emotes, so they can be removed at request. Also if formatting's an issue that can be fixed too. Basically anything lol, idek if it'll work as is.

Hopefully it does.


mia-pi-git:
I believe we've discussed this before. 
The consensus seems to be most people don't want this, as PS's simplicity is appreciated.
Additionally, this is a very clunky way of doing it, and provides absolutely no way for rooms to disable it. As such, it would be spammed and abused almost instantly.

AeonicX7:
> Also quick talk with Aeonic, seems to be a similar no from policy?

Yeah. Not a fan of this idea for a variety of reasons, most of which were covered pretty well in that thread.

harbar20:
Ah. The reasons in that thread make complete sense. Should I close the PR?

mia-pi-git:
Seems like that's the consensus.

AnnikaCodes:
For future reference, all new code accepted to this repository should be written in TypeScript.

harbar20:
3456ertydfghxcvb

PR #7870, opened 2021-01-03
Bug Fix: Protean/Libero on first of two turn moves

aviettran:
> Protean / Libero should activate on both turns of a charge move like Fly or Phantom Force


Marty-D (approved):
Thanks!

PR #7874, opened 2021-01-03
Support logging battle logs to a Postgres database

mia-pi-git:
This should help speed up battlesearch, battlelog, and /sharedloggedbattles.
The library (postgres.ts) is there for ease of connection / wrapping / keeping things functioning without having the module installed.
The schema includes all items that logBattle logs so that hopefully (?) we can use this for usage stats as well, but less sure on that.


DaWoblefet:
Would this currently be doing dual logging? I will preempt Marty to ask about how this will affect the usage stats script.

mia-pi-git:
> Would this currently be doing dual logging? I will preempt Marty to ask about how this will affect the usage stats script.

Yes, dual logging. 
I also already talked to Marty about this. The response was that pulling from postgres for the scripts won't be feasible until pre finishes rewriting them.

databases/schemas/battle-logs.sql
```diff
@@ -0,0 +1,23 @@
...
+	p1id TEXT NOT NULL,
+	p2id TEXT NOT NULL,
+	p1 TEXT NOT NULL,
+	p2 TEXT NOT NULL,
+	p1team JSONB[] NOT NULL,
+	p2team JSONB[] NOT NULL,
+   log TEXT[] NOT NULL,
+   inputLog TEXT[] NOT NULL,
+   turns SMALLINT NOT NULL,
+   endType TEXT NOT NULL,
```
monsanto, line 14: 3rd try at getting you to change this to an enum :P

lib/postgres.ts
```diff
@@ -0,0 +1,28 @@
...
+import type {Pool as PostgresPool} from 'pg';
+import type {SQLStatement} from 'sql-template-strings';
+
+const DEFAULT_CONFIG = 'Config' in global ? Config.usepostgres : null;
+
+export class PostgresDatabase {
+	private pool: PostgresPool;
+	connected = false;
+	constructor(config = DEFAULT_CONFIG) {
+		this.pool = config ? new (require('pg').Pool)(config) : null!;
```
monsanto, line 16: Weren't we going to try using Odyssey to pool outside PS?

mia-pi-git: I think so? but i set it as a pool for now since i was confused whether you were sure on that or not.

lib/postgres.ts
```diff
@@ -0,0 +1,28 @@
...
+	}
+	async query(statement: string | SQLStatement, values?: any[]) {
+		if (!this.pool) {
+			return null;
```
monsanto, line 20: Why?

mia-pi-git: So that it's easier to tell / error properly when someone isn't using postgres (plus, i think `null` is PS's preferred return failure type?)

lib/postgres.ts
```diff
@@ -0,0 +1,28 @@
...
+		}
+		if (!this.connected) {
+			await this.pool.connect();
+			this.connected = true;
+		}
```
monsanto, line 25: This doesn't make any sense, this.pool.connect will return a new connection, which you discard. this.pool.query by itself does what you want--make a connection, run a query, and then disconnect. The connected flag is superfluous too

package.json
```diff
@@ -42,7 +44,7 @@
       "pre-push": "npm run lint"
     }
   },
-  "bin": "./pokemon-showdown",
+  "bin": "pokemon-showdown",
```
monsanto, line 47: ?

mia-pi-git: Accidental. Not sure why this keeps happening.

lib/postgres.ts
mia-pi-git: Ah, i see.

server/chat-plugins/chatlog.ts
```diff
@@ -1023,8 +1025,49 @@ export class RipgrepLogSearcher extends Searcher {
...
+	database: PostgresDatabase;
+	constructor() {
+		super();
+		this.usePM = false;
+		this.database = new PostgresDatabase();
+	}
+	async findBattleLog(tier: ID, number: number) {
+		const SQL = require('sql-template-strings');
+		const results = await this.database.query(
+			SQL`SELECT log FROM battle_logs WHERE roomid = ${number} AND format = ${tier}`
```
monsanto, line 1038: Another discord comment that was missed -- if `roomid` is a primary key, then `AND format =` is redundant. You need to decide whether to remove that, or to add `format` to a primary composite key

lib/postgres.ts
monsanto: If postgres isn't enabled it shouldn't even be possible to call `query`, right? If this is the case you should fail loudly by throwing an error

server/chat-plugins/chatlog.ts
```diff
@@ -1019,8 +1021,48 @@ export class RipgrepLogSearcher extends Searcher {
...
+		);
+		if (!results || !results.length) return null;
+		return results[0].log;
+	}
+	async getSharedBattles(userids: string[]) {
+		const SQL = require('sql-template-strings');
+		const query = SQL`SELECT roomid, format FROM battle_logs WHERE `;
+		query.append(SQL`(p1id = ${userids[0]} AND p2id = ${userids[1]})`);
+		query.append(` OR `);
+		query.append(SQL`(p1id = ${userids[1]} AND p2id = ${userids[0]})`);
```
monsanto, line 1044: Missed discord comment: you have to have an index on these or it will be too slow

server/chat-plugins/battlesearch.ts
```diff
@@ -24,58 +23,108 @@ interface BattleSearchResults {
...
+export class PostgresBattleSearcher extends BattleSearchHandler {
+	database = new PostgresDatabase();
+	async runSearch(userids: ID[], month: string, tierid: ID, turnLimit?: number) {
+		const SQL = require('sql-template-strings');
+		const query = SQL`SELECT * FROM battle_logs `;
+		if (userids.length === 1) {
+			query.append(SQL`WHERE (p1id = ${userids[0]} OR p2id = ${userids[0]})`);
+		} else {
+			const [id1, id2] = userids;
+			query.append(SQL`WHERE (p1id = ${id1} OR p2id = ${id1}) AND (p1id = ${id2} OR p2id = ${id2})`);
```
monsanto, line 114:
```suggestion
			query.append(SQL`WHERE (p1id = ${id1} AND p2id = ${id2}) OR (p1id = ${id2} AND p2id = ${id1})`);
```

server/chat-plugins/battlesearch.ts
```diff
@@ -106,140 +155,211 @@ export async function runBattleSearch(userids: ID[], month: string, tierid: ID,
...
+		for (const {date} of rows) {
+			const [day] = toDate(date).split(' ');
+			const month = day.slice(0, -3);
+			if (!months.includes(month)) months.push(month);
+		}
+		return months;
+	}
+	async listTiers(month: string) {
+		const results: string[] = [];
+		const rows = await this.database.query(`SELECT format, date FROM battle_logs`);
```
monsanto, line 183: You are selecting every single row from `battle_logs`

server/chat-plugins/battlesearch.ts
```diff
@@ -106,140 +155,211 @@ export async function runBattleSearch(userids: ID[], month: string, tierid: ID,
...
-			} else {
-				results[day].totalTies++;
-			}
-			// explicitly state 0 of stats if none
-			for (const id of userids) {
-				if (!results[day].totalLosses[id]) results[day].totalLosses[id] = 0;
-				if (!results[day].totalWins[id]) results[day].totalWins[id] = 0;
+	async listMonths() {
+		const rows = await this.database.query(
+			`SELECT date FROM battle_logs`
```
monsanto, line 171: You are selecting every row from `battle_logs`

mia-pi-git: is there a way to only select one of each differing row?

monsanto: `select distinct`, but you need to run the query analyzer on it to see if it will be efficient enough. It may work efficiently with an index on the appropriate columns. If not, you need to make a different table with this stuff & reference it from `battle_logs`

lib/postgres.ts
```diff
@@ -0,0 +1,47 @@
...
+	}
+	async query(statement: string | SQLStatement, values?: any[]) {
+		if (!this.pool) {
+			throw new Error(`Attempting to use postgres without 'pg' installed`);
+		}
+		let result;
+		try {
+			result = await this.pool.query(statement, values);
+		} catch (e) {
+			throw new Error(e.message);
```
monsanto, line 24: Is this because of the wrong stack thing you mentioned on discord? if so needs a comment

server/room-battle.ts
```diff
@@ -892,6 +893,42 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+		const winner = this.endType === 'tie' ? '' : toID(logData.winner);
+		const {p1, p2} = logData;
+		const p1id = toID(p1);
+		const p2id = toID(p2);
+		const loser = winner ? (p1id === winner ? p2id : p1id) : "";
+		const format = Dex.getFormat(this.format).id;
+		const SQL = require('sql-template-strings');
+
+		let rowid;
+		await Rooms.RoomBattle.logDatabase.transaction(async worker => {
```
monsanto, line 911: i think it would make more sense for `transaction` to return the result of its callback, and then do `rowid = await Rooms.RoomBattle.logDatabase.transaction(....)`

lib/postgres.ts
```diff
@@ -0,0 +1,47 @@
...
+			config = require('../config/config').usepostgres;
+			if (!config) throw new Error('');
+		} catch (e) {}
+		return config;
+	}
+	async transaction(callback: (conn: PG.PoolClient) => any) {
+		const conn = await this.pool.connect();
+		await conn.query(`BEGIN;`);
+		try {
+			await callback(conn);
```
monsanto, line 42: Doesn't handle case where you need to retry transaction

lib/postgres.ts
```diff
@@ -0,0 +1,63 @@
...
+	async transaction(callback: (conn: PG.PoolClient) => any, depth = 0): Promise<any> {
+		const conn = await this.pool.connect();
+		await conn.query(`BEGIN;`);
+		let result;
+		try {
+			// eslint-disable-next-line callback-return
+			result = await callback(conn);
+		} catch (e) {
+			await conn.query(`ROLLBACK;`);
+			const code = parseInt(e?.code);
```
monsanto, line 46: These aren't integers, there's a reason I did string comparisons in the sample code I gave you: https://www.postgresql.org/docs/13/errcodes-appendix.html

The two codes you're checking for happen to be integers, but I see no reason to do this parse step first

lib/postgres.ts
```diff
@@ -0,0 +1,63 @@
...
+			// two concurrent transactions conflicted, try again
+			if (code === 40001 && depth <= 10) {
+				result = await this.transaction(callback, depth++);
```
monsanto, line 49: Don't you mean `depth + 1`?

lib/postgres.ts
```diff
@@ -0,0 +1,63 @@
...
+			// serialization failures to be reported as failed
+			// unique constraint checks. Only retrying once since
+			// it could be our fault (thanks chaos for this info / the first half of this comment)
+			} else if (code === 23505 && !depth) {
+				result = await this.transaction(callback, depth++);
+			} else {
+				throw e;
+			}
+		}
+		await conn.query(`COMMIT;`);
```
monsanto, line 60: If you catch, then you will rollback, so if you reach this line of code you won't be in a transaction

AnnikaCodes:
This code will be handling a large amount of data; battle logs currently take up the majority of the disk space on `sim3`. Before deploying this, we'd need to make sure the Postgres server was able to handle the data, and have a plan about what to do (the current selective compression policy for the oldest logs wouldn't directly apply to Postgres -- would the old logs be deleted? exported and compressed? would we set up a new database on occasion? would we just keep adding storage to the db server ad infinitum?). These are decisions that need to be made anyway, but we should think about them before deploying.

lib/postgres.ts
```diff
@@ -0,0 +1,62 @@
...
+import type {SQLStatement} from 'sql-template-strings';
+
+export class PostgresDatabase {
+	private pool: PG.Pool;
+	constructor(config = PostgresDatabase.getConfig()) {
+		this.pool = config ? new (require('pg').Pool)(config) : null!;
+	}
+	async query(statement: string | SQLStatement, values?: any[]) {
+		if (!this.pool) {
+			throw new Error(`Attempting to use postgres without 'pg' installed`);
```
AnnikaCodes, line 18: chaos mentioned this somewhere else but it's worth repeating: should pg be optional if it's not hard-to-install like sqlite?

server/chat-plugins/battlesearch.ts
```diff
@@ -87,159 +157,230 @@ export async function runBattleSearch(userids: ID[], month: string, tierid: ID,
...
+	const b = bKey.split('-').map(n => parseInt(n));
+	if (a[0] !== b[0]) return b[0] - a[0];
+	return b[1] - a[1];
+}
 
-	const response = await PM.query({userids, turnLimit, month, tierid});
-	connection.send(buildResults(response, userids as ID[], month, tierid, turnLimit));
+function toDate(str: string) {
+	return Chat.toTimestamp(new Date(Number(str)));
 }
```
AnnikaCodes, line 385: Have you tested these queries on a large database of battles? Are they performant? It would be nice to get battlesearch right this time; right now, staff rely on an external tool to manually check battle logs by format and date.

server/chat-plugins/chatlog.ts
```diff
@@ -532,6 +421,119 @@ export abstract class Searcher {
 		}
 		return `^` + searches.filter(Boolean).map(term => `(?=.*${term})`).join('');
 	}
+	async findBattleLog(tier: ID, number: number): Promise<string[] | null> {
```
AnnikaCodes, line 424: Did you change this function at all, or just move it?

AnnikaCodes:
(Also, please fix the merge conflicts :)

PR #7877, opened 2021-01-04
Chat: add support for private pages

mia-pi-git:
Similar to command support - should be useful rather than having to do user.can + this.errorReply('page doesn't exist') etc.


Zarel:
Looks good, thanks!

PR #7878, opened 2021-01-04
FS: Fix hotpatching and add more throttling

mia-pi-git:
- `writeUpdate` state is now stored in a global variable, so hotpatching doesn't crash it
- throttling now writes on the tail (so two throttled `writeUpdate` calls will write one update, not two)
- room settings, punishments, and helptickets are now throttled

-----

Discussed with @Zarel, decided this would be the best way to stop the writeUpdate crashes.
(for example, `A main process Promise crashed: Error: ENOENT: no such file or directory, rename '/home/ps/main/config/tickets.json.NEW' -> '/home/ps/main/config/tickets.json'`, etc)

### DIFF
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -28,7 +28,7 @@ import {ReadStream, WriteStream} from './streams';
 const ROOT_PATH = pathModule.resolve(__dirname, '..');
 const DEFAULT_THROTTLE = 5 * 1000; // 5 seconds
 
-export interface PendingUpdate {
+interface PendingUpdate {
 	isWriting: boolean; // true: waiting on a call to FS.write, false: waiting on a throttle
 	pendingDataFetcher: (() => string | Buffer) | null;
 	pendingOptions: AnyObject | null;
@@ -152,7 +152,7 @@ export class FSPath {
 	 * called, if `writeUpdate` is called many times in a short period.
 	 *
 	 * `options.throttle`, if it exists, will make sure updates are not
-	 * written more than once every `options.throttle` milliseconds, else the default is 5 seconds.
+	 * written more than once every `options.throttle` milliseconds.
 	 *
 	 * No synchronous version because there's no risk of race conditions
 	 * with synchronous code; just use `safeWriteSync`.
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -161,7 +161,7 @@ export class FSPath {
 		if (Config.nofswriting) return;
 		const pendingUpdate: PendingUpdate | undefined = __fsState.pendingUpdates.get(this.path);
 
-		const throttleTime = Date.now() + (options.throttle || 0);
+		const throttleTime = options.throttle ? Date.now() + options.throttle : 0;
 
 		if (pendingUpdate) {
 			pendingUpdate.pendingDataFetcher = dataFetcher;
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -189,8 +189,7 @@ export class FSPath {
 	}
 
 	writeUpdateNow(dataFetcher: () => string | Buffer, options: AnyObject) {
-		// @ts-ignore
-		const throttleTime = Date.now() + (options.throttle ? options.throttle : DEFAULT_THROTTLE);
+		const throttleTime = options.throttle ? Date.now() + options.throttle : 0;
 		const update = {
 			isWriting: true,
 			pendingDataFetcher: null,
--- a/lib/fs.ts
+++ b/lib/fs.ts
@@ -26,7 +26,6 @@ import * as pathModule from 'path';
 import {ReadStream, WriteStream} from './streams';
 
 const ROOT_PATH = pathModule.resolve(__dirname, '..');
-const DEFAULT_THROTTLE = 5 * 1000; // 5 seconds
 
 interface PendingUpdate {
 	isWriting: boolean; // true: waiting on a call to FS.write, false: waiting on a throttle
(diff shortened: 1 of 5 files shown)


server/global-variables.d.ts
```diff
@@ -53,4 +56,5 @@ declare global {
 	const Users: typeof UsersType.Users;
 	const Verifier: typeof VerifierType;
 	const toID: typeof DexType.toID;
+	const __fsState: {pendingUpdates: Map<string, PendingUpdate>};
```
Zarel, line 59: I would actually explicitly not declare this here; since the entire point of not using `FS` as a global is so that it's not used outside the FS package.

lib/fs.ts
```diff
@@ -26,16 +26,21 @@ import * as pathModule from 'path';
...
 	pendingOptions: AnyObject | null;
 	throttleTime: number; // throttling until time (0 for no throttle)
 	throttleTimer: NodeJS.Timer | null;
 }
 
-const pendingUpdates = new Map<string, PendingUpdate>();
+if (!global.__fsState) {
+	global.__fsState = {
+		pendingUpdates: new Map<string, PendingUpdate>(),
+	};
```
Zarel, line 42:
```suggestion
declare const global: {__fsState: {pendingUpdates: Map<string, PendingUpdate>}};

if (!global.__fsState) {
	global.__fsState = {
		pendingUpdates: new Map(),
	};
```

This _should_ work to locally override `global`.

Zarel:
Drafting until `options.throttle` is merged with `options.writeNow`.

lib/fs.ts
```diff
@@ -145,17 +152,16 @@ export class FSPath {
 	 * called, if `writeUpdate` is called many times in a short period.
 	 *
 	 * `options.throttle`, if it exists, will make sure updates are not
-	 * written more than once every `options.throttle` milliseconds.
+	 * written more than once every `options.throttle` milliseconds, else the default is 5 seconds.
```
Zarel, line 155: The default's now no throttle, right?

lib/fs.ts
```diff
@@ -26,16 +26,23 @@ import * as pathModule from 'path';
 import {ReadStream, WriteStream} from './streams';
 
 const ROOT_PATH = pathModule.resolve(__dirname, '..');
+const DEFAULT_THROTTLE = 5 * 1000; // 5 seconds
 
-interface PendingUpdate {
+export interface PendingUpdate {
```
Zarel, line 31:
```suggestion
interface PendingUpdate {
```
We ended up not exporting this, right?

lib/fs.ts
```diff
@@ -145,17 +152,16 @@ export class FSPath {
...
 	 * with synchronous code; just use `safeWriteSync`.
 	 */
 	writeUpdate(dataFetcher: () => string | Buffer, options: AnyObject = {}) {
 		if (Config.nofswriting) return;
-		const pendingUpdate: PendingUpdate | undefined = pendingUpdates.get(this.path);
+		const pendingUpdate: PendingUpdate | undefined = __fsState.pendingUpdates.get(this.path);
 
-		// @ts-ignore
-		const throttleTime = options.throttle ? Date.now() + options.throttle : 0;
+		const throttleTime = Date.now() + (options.throttle || 0);
```
Zarel, line 164: I don't think this needed to be changed.

lib/fs.ts
```diff
@@ -168,38 +174,50 @@ export class FSPath {
 			return;
 		}
 
-		this.writeUpdateNow(dataFetcher, options);
+		if (!options.throttle) {
```
Zarel, line 177:
```suggestion
		if (!throttleTime) {
```

lib/fs.ts
Zarel:
```suggestion
	 * written more than once every `options.throttle` milliseconds.
```

lib/fs.ts
```diff
@@ -26,16 +26,23 @@ import * as pathModule from 'path';
 import {ReadStream, WriteStream} from './streams';
 
 const ROOT_PATH = pathModule.resolve(__dirname, '..');
+const DEFAULT_THROTTLE = 5 * 1000; // 5 seconds
```
Zarel, line 29:
```suggestion
```

lib/fs.ts
```diff
@@ -168,38 +174,50 @@ export class FSPath {
...
+			throttleTime,
+			throttleTimer: setTimeout(() => this.checkNextUpdate(), throttleTime - Date.now()),
+		};
+		__fsState.pendingUpdates.set(this.path, update);
 	}
 
 	writeUpdateNow(dataFetcher: () => string | Buffer, options: AnyObject) {
 		// @ts-ignore
-		const throttleTime = options.throttle ? Date.now() + options.throttle : 0;
+		const throttleTime = Date.now() + (options.throttle ? options.throttle : DEFAULT_THROTTLE);
```
Zarel, line 194:
```suggestion
		const throttleTime = options.throttle ? Date.now() + options.throttle : 0;
```

lib/fs.ts
```diff
@@ -168,38 +174,50 @@ export class FSPath {
...
+			pendingDataFetcher: dataFetcher,
+			pendingOptions: options,
+			throttleTime,
+			throttleTimer: setTimeout(() => this.checkNextUpdate(), throttleTime - Date.now()),
+		};
+		__fsState.pendingUpdates.set(this.path, update);
 	}
 
 	writeUpdateNow(dataFetcher: () => string | Buffer, options: AnyObject) {
 		// @ts-ignore
```
Zarel, line 193:
```suggestion
```

server/global-variables.d.ts
```diff
@@ -16,6 +16,8 @@ import {Tournaments as TournamentsType} from './tournaments';
 
 import {Dex as DexType} from '../sim/dex';
 
+import type {PendingUpdate} from '../lib/fs';
+
```
Zarel, line 20:
```suggestion
```

Zarel:
Thanks!

PR #7879, opened 2021-01-05
Various mechanics fixes

urkerab:
- Cursed Body should be able to disable Z-moves
- Pressure should not deduct additional PP from Sticky Web
- Pressure should deduct additional PP from Z-moves and Max moves
- Sturdy should trigger before Focus Sash but after False Swipe
- Revelation Dance should use Arceus's real type
- Shell Bell should heal from all damage

You might want to review each commit separately.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3679,7 +3679,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 				return null;
 			}
 		},
-		onDamagePriority: -100,
+		onDamagePriority: -30,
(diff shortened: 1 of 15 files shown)


Marty-D (approved):
Neat, thanks! I'll merge them separately too.

MacChaeger:
Is Cursed Body supposed to disable Max Moves? This change seems to have made it so https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8709336

PR #7880, opened 2021-01-06
Fix Intimidate failure text

DaWoblefet:
Inner Focus, Oblivious, Scrappy, and Own Tempo were not displaying "[Pokemon]'s Attack was not lowered!", instead a generic immunity message. Additionally, I've removed the string for Own Tempo in ``block``; I tried it out on testclient without it and it still displayed "[Pokemon] cannot be confused!" on Confuse Ray so that should be ok, and it fixes the bug with Own Tempo blocking Intimidate using its confusion blocking message.

### DIFF
--- a/data/text/abilities.ts
+++ b/data/text/abilities.ts
@@ -923,8 +923,6 @@ export const AbilitiesText: {[k: string]: AbilityText} = {
 			desc: "This Pokemon cannot be confused. Gaining this Ability while confused cures it.",
 			shortDesc: "This Pokemon cannot be confused.",
 		},
-
-		block: "  [POKEMON] cannot be confused!",
 	},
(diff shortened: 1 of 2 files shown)


data/text/abilities.ts
```diff
@@ -923,8 +923,6 @@ export const AbilitiesText: {[k: string]: AbilityText} = {
 			desc: "This Pokemon cannot be confused. Gaining this Ability while confused cures it.",
 			shortDesc: "This Pokemon cannot be confused.",
 		},
-
-		block: "  [POKEMON] cannot be confused!",
```
urkerab, line 927: But the text does appear in the text dump... is it a left-over from an old gen?

PR #7882, opened 2021-01-06
Grassy Terrain should heal before Leftovers

urkerab:
The test was a little fragile until I worked out that it was looking for the wrong amount of healing, so Alakazam was still healing first, even though the test thought it wasn't.

### DIFF
--- a/data/mods/gen7/moves.ts
+++ b/data/mods/gen7/moves.ts
@@ -304,6 +304,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			onResidual() {
 				this.eachEvent('Terrain');
 			},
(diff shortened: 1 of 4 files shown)


DaWoblefet:
Do conditions of moves not partially inherit? It seems weird that we would have to update this in two places. The only thing that should be relevant to not inherit would be the BP modifier from 6144 -> 5325.

urkerab:
Unfortunately you have to copy the entire condition if you want to tweak it slightly. (I don't know how easy it would be to make conditions inherit.)

DaWoblefet (approved):

Marty-D (approved):
👍

PR #7883, opened 2021-01-07
Helptickets: Fix bugs with unbanning

mia-pi-git:
Should fix the bug mentioned in #private-dev.

### DIFF
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -397,7 +397,7 @@ export class HelpTicket extends Rooms.RoomGame {
 				`${punished !== userid ? `, because you have the same IP as ${userid}` : ''}. ${reason ? `Reason: ${reason}` : ''}`
 			);
 		}
-		return `You are banned from creating tickets.`;
+		return `You are banned from creating help tickets.`;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


server/chat-plugins/helptickets.ts
```diff
@@ -1365,17 +1382,18 @@ export const commands: ChatCommands = {
 			if (!target) return this.parse('/help helpticket unban');
 
 			this.checkCan('lock');
-			const targetUser = Users.get(target, true);
-			if (!targetUser) return this.errorReply(`User not found.`);
+			let targetUser: User | ID | null = Users.get(target);
+			if (!targetUser) targetUser = toID(target);
```
Zarel, line 1386: :/ can we not? It feels like this should be two different variables...

server/chat-plugins/helptickets.ts
```diff
@@ -364,28 +364,41 @@ export class HelpTicket extends Rooms.RoomGame {
...
 	}
+	static getBanMessage(userid: ID, punishment: Punishment) {
+		if (userid !== punishment[0]) {
+			const [, punished,, reason] = punishment;
+			return (
+				`You are banned from creating help tickets` +
+				`${punished !== userid ? `, because you have the same IP as ${userid}` : ''}. ${reason ? `Reason: ${reason}` : ''}`
+			);
+		}
+		return `You are banned from creating tickets.`;
```
Zarel, line 400:
```suggestion
		return `You are banned from creating help tickets.`;
```

Zarel:
Thanks!

PR #7884, opened 2021-01-07
Bugfix: Various Future Sight / Doom Desire Fixes

aviettran:
Four separate issues. The commits are pretty tidy so you can view each individually for details.

### DIFF
--- a/data/scripts.ts
+++ b/data/scripts.ts
@@ -345,6 +345,7 @@ export const Scripts: BattleScriptsData = {
 				this.add('-fail', pokemon);
 				this.attrLastMove('[still]');
 			}
+			if (move.isFutureMove) return true;
 			return false;
 		}
 
--- a/data/scripts.ts
+++ b/data/scripts.ts
@@ -297,7 +297,7 @@ export const Scripts: BattleScriptsData = {
 		return true;
 	},
 	/** NOTE: includes single-target moves */
-	trySpreadMoveHit(targets, pokemon, move) {
+	trySpreadMoveHit(targets, pokemon, move, notActive) {
 		if (targets.length > 1 && !move.smartTarget) move.spreadHit = true;
 
 		const moveSteps: ((targets: Pokemon[], pokemon: Pokemon, move: ActiveMove) =>
@@ -335,7 +335,7 @@ export const Scripts: BattleScriptsData = {
 			[moveSteps[2], moveSteps[4]] = [moveSteps[4], moveSteps[2]];
 		}
 
-		this.setActiveMove(move, pokemon, targets[0]);
+		if (!notActive) this.setActiveMove(move, pokemon, targets[0]);
 
 		const hitResult = this.singleEvent('Try', move, null, pokemon, targets[0], move) &&
 			this.singleEvent('PrepareHit', move, {}, targets[0], pokemon, move) &&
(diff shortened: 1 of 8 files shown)


data/scripts.ts
```diff
@@ -335,7 +335,7 @@ export const Scripts: BattleScriptsData = {
 			[moveSteps[2], moveSteps[4]] = [moveSteps[4], moveSteps[2]];
 		}
 
-		this.setActiveMove(move, pokemon, targets[0]);
+		if (!notActive) this.setActiveMove(move, pokemon, targets[0]);
```
urkerab, line 338: Maybe this really belongs in `useMoveInner` somewhere? (That way you wouldn't need it in `tryMoveHit` either.)

data/scripts.ts
```diff
@@ -345,6 +345,7 @@ export const Scripts: BattleScriptsData = {
 				this.add('-fail', pokemon);
 				this.attrLastMove('[still]');
 			}
+			if (move.isFutureMove) return true;
 			return false;
```
urkerab, line 349: We already have `this.NOT_FAIL` for other moves that don't double the power of Stomping Tantrum. Maybe this should say
```suggestion
			return hitResult === this.NOT_FAIL;
```
(with other related changes as necessary)

test/sim/moves/futuresight.js
```diff
@@ -85,11 +85,12 @@ describe('Future Sight', function () {
 		battle.makeChoices();
 		battle.makeChoices();
 		battle.makeChoices();
+		battle.makeChoices();
```
urkerab, line 88: I thought Future Sight hit at the end of the third turn, so you shouldn't need a fourth turn here. Are you sure you're really fixing the bug, and not just giving Wynaut a second chance to set Future Sight?

aviettran: This is 3 turns. The third choice seems to be switching to Roggenrola after Shedinja faints.

data/scripts.ts
aviettran: The futuremove condition calls trySpreadMoveHit directly. Is it advisable to change this?

data/scripts.ts
aviettran: trySpreadMoveHit returns a boolean. We _could_ refactor it, but I think it's appropriate that it return true, i.e., futuresight _does_ succeed when it is used. Thoughts?

test/sim/moves/futuresight.js
urkerab: Ah, in that case I think a comment on the switch choice would make this clearer.

data/scripts.ts
urkerab: I'm not changing the return value of `trySpreadMoveHit` though am I?

data/scripts.ts
urkerab: (I'm insufficiently familiar with what effects can modify future moves to be able to contribute meaningfully to that question.)

test/sim/moves/futuresight.js
aviettran: resolved

data/scripts.ts
aviettran: Gotcha. Resolved.

test/sim/moves/futuresight.js
urkerab: Thanks!

data/scripts.ts
urkerab: Great, although personally I really don't like the `if (...) return true; return false;` especially when the condition actually has boolean type.

aviettran: updated

Marty-D (approved):
👍

PR #7887, opened 2021-01-07
Fix capitalization in Charizard-Mega-Y

PartMan7:
D:


Zarel:
Thanks! We might want to add this to our data validator if we care about stuff like this.

PR #7891, opened 2021-01-08
Cursed Body ignores Max Moves

urkerab:
When enabling Cursed Body against Z-moves in PR #7879 I accidentally allowed it to process against Max moves as well.


Zarel:
Thanks!

PR #7892, opened 2021-01-08
Max Guard should block Shadow Force

PartMan7:
Serebii says it should be blocked, Bulbapedia has no mention of it.


DaWoblefet:
Here's a proper source confirming your change is accurate: https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8647503

PartMan7:
Oh, TI committed after the PR. Closing this.

PR #7893, opened 2021-01-08
Add Megas For All and Double Trouble

HematiteH2O:
Uh, hi!
I'm the coder for the current PMOTS, Megas for All, and I was just hoping to help get smooth the transition process to get it hosted on Showdown, preferably before we resume making regular updates to it?

I've already tested this patch locally, and I can confirm that it's fully compatible and works perfectly on main Showdown as I have it.
I offered this patch to the more recognized contributor KrisXV a few days ago, and he said he would rather copy it over himself so that he could submit it alongside the Leader's Choice mod. That said, he was planning on copy/pasting directly from my patch anyway, so I really hope it's all the same if I just do this myself at this point?
It's been three days since then and more than a week now since the PMOTS was announced, and M4A is about to start a new slate.
I get that it's not a priority for people outside of pet mods, but it's kind of frustrating at this point.

I'm rrreasonably sure that I would kind of need to be allowed to make pull requests for this anyway in the future?
It would be really, really silly if I were maintaining the format myself on DH like I always do and someone I don't even know had to spend the next three months mirroring the changes on Showdown manually after the fact, haha - I feel like it would be a lot easier on everyone if I could just handle this myself.

If you need me to do or to clarify anything about this, I'm Hematite on Smogon and Hematite#4609 on Discord!
I just wanted to give the context for this pull request because I'm a new face to this repository and would not be recognized.


HematiteH2O:
_There_ we go. Sorry, I've never done this before because the check on DH doesn't actually work. Kind of felt like playing Whack-a-Mole for a bit there. OTL
But yeah, I think this is all clear now, except the one about line 2 in scripts.ts, which... I don't actually know what that error means.
Hopefully whoever reviews this can handle the one line? It still compiles fine and works in battle, so I don't believe it's a problem, exactly, but passing the last of the checks can only be a good thing, I'm sure...
Anyway, uh, I think that's about all I can do here (unless you have any concerns, in which case please do get in touch)!

data/mods/m4av7/scripts.ts
```diff
@@ -0,0 +1,267 @@
+export const Scripts: {[k: string]: ModdedBattleScriptsData} = {
+	init: function () {
```
mia-pi-git, line 2:
```suggestion
	init() {
```

KrisXV:
I had to copy and paste your code and then fix all of the build errors/ further optimizing the code. That is why it has been taking me so long.

Zarel:
There's a lot more than one error. It looks like you added a custom event, which will be pretty difficult to integrate. I don't think we have a good way to add those in a way that makes TypeScript happy.

data/mods/m4av7/abilities.ts
```diff
@@ -0,0 +1,1493 @@
...
+			}
+		},
+		name: "Cold Sweat",
+		rating: 4,
+		num: -1006,
+	},
+	trashcompactor: {
+		desc: "This Pokémon is immune to all entry hazards. If it lands on any type of entry hazard, it clears the hazard and Stockpiles 1.",
+		shortDesc: "Hazard immunity. Clears hazards, Stockpiles 1 if switched in on them.",
+		onAfterMega(pokemon) {
```
Zarel, line 342: Actually, I can't tell if `AfterMega` is a custom event or just not in the `AbilityData` definition. Feel free to add it (I guess to `EventMethods`?

Zarel: Oh yeah, it's a real event. Just add it to `EventMethods` in `dex-conditions` @KrisXV

Zarel:
Hello, @HematiteH2O! Welcome to contributing to Pokémon Showdown! Thanks for your work here!

You are certainly welcome to send your code directly to us, as you did here (it would certainly make it faster to contribute), but we do have a hard requirement that you need to pass our unit tests and follow our contribution rules.

Fortunately for you, since you'd be contributing to the official repository, if you ever have trouble passing our unit tests, the rest of us would be happy to help with that process! Our Discord (which I see you've found) is the best place for that.

Zarel:
Thanks, everyone!

PR #7898, opened 2021-01-09
Megas for All: Fix Delibird's ability

HematiteH2O:
Okay, this worked when I tried it locally and it also appeared to pass npm test on my end!
This is necessary for Mega Delibird's Ability like I noted on the Discord. C:


Zarel:
Looks good!

PR #7899, opened 2021-01-10
Ripen should double Jaboca and Rowap damage

urkerab:
I can't use the `Damage` event because the damage has already been rounded by that point, so instead I've set a flag in the item data for maximum flexibility (because who knows what SSB or some other OM or Pet Mod is going to want to do in the future).

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -118,7 +118,7 @@ export const Items: {[itemid: string]: ItemData} = {
 				pokemon.eatItem();
 			}
 		},
-		onTryEatItem(item, pokemon) {
+		onTryEatItem(itemData, pokemon) {
 			if (!this.runEvent('TryHeal', pokemon)) return false;
 		},
 		onEat(pokemon) {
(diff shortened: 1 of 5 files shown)


TheImmortal:
imo it should be ``ripen`` not ``ripe``

data/items.ts
```diff
@@ -2587,7 +2587,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		onDamagingHit(damage, target, source, move) {
 			if (move.category === 'Physical') {
 				if (target.eatItem()) {
-					this.damage(source.baseMaxhp / 8, source, target);
+					this.damage(source.baseMaxhp / (this.effectData.ripe ? 4 : 8), source, target);
```
Marty-D, line 2590:
```suggestion
					this.damage(source.baseMaxhp / (target.hasAbility('ripen') ? 4 : 8), source, target);
```
Doesn't this accomplish the same thing? I know you mention flexibility but I'm not sure theoretical future uses are worth adding EffectState all over the place in this case.

DaWoblefet: Bump @urkerab, I am also generally a fan of just hardcoding the Ripen Ability for Jaboca and Rowap.

urkerab:
I'll let someone else do this then.

PR #7900, opened 2021-01-10
Add a command for searching groupchats

mia-pi-git:
See https://www.smogon.com/forums/threads/3514540/page-18#post-8709906.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -1336,8 +1336,7 @@ export const commands: ChatCommands = {
 	gcsearch: 'groupchatsearch',
 	async groupchatsearch(target, room, user) {
 		this.checkCan('lock');
-		// eslint-disable-next-line no-useless-escape
-		target = target.toLowerCase().replace(/[^a-z0-9\-]+/g, '');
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -1333,4 +1333,29 @@ export const commands: ChatCommands = {
 		const [type, userid] = target.split(',').map(toID);
 		return this.parse(`/j view-logsaccess-${type || 'all'}${userid ? `-${userid}` : ''}`);
 	},
+	gcsearch: 'groupchatsearch',
+	async groupchatsearch(target, room, user) {
+		this.checkCan('lock');
+		// eslint-disable-next-line no-useless-escape
```
Zarel, line 1339: ?

mia-pi-git: For some reason, ESLint sees the \- as a useless escape, and so fails checks on it.

KrisXV: You don't need to escape `-` in character classes unless it's between, a and z, A and Z, or 0 and 9.

Zarel: That's because it _is_ a useless escape.

Zarel:
Thanks!

PR #7903, opened 2021-01-11
Permissions: Fix bug in adding namespaces

mia-pi-git:
https://www.smogon.com/forums/threads/add-roomevents-as-a-group-permission-in-permissions.3676240/#post-8714596.


Zarel:
Thanks!

PR #7904, opened 2021-01-11
Info: Fix /show and /requestshow interaction error

mia-pi-git:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-44#post-8709275

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2515,10 +2515,7 @@ export const commands: ChatCommands = {
 		if (!room.settings.requestShowEnabled) {
 			return this.errorReply(`Media approvals are disabled in this room.`);
 		}
-		const showPerm = room.settings.permissions?.['/show'];
-		if (user.can('showmedia', null, room) && showPerm && room.auth.atLeast(user, showPerm)) {
-			return this.errorReply(`Use !show instead.`);
-		}
+		if (user.can('showmedia', null, room, '/show')) return this.errorReply(`Use !show instead.`);
 		if (room.pendingApprovals?.has(user.id)) return this.errorReply('You have a request pending already.');
 		if (!toID(target)) return this.parse(`/help requestshow`);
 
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.ts
```diff
@@ -2515,7 +2515,10 @@ export const commands: ChatCommands = {
 		if (!room.settings.requestShowEnabled) {
 			return this.errorReply(`Media approvals are disabled in this room.`);
 		}
-		if (user.can('showmedia', null, room)) return this.errorReply(`Use !show instead.`);
+		const showPerm = room.settings.permissions?.['/show'];
+		if (user.can('showmedia', null, room) && showPerm && room.auth.atLeast(user, showPerm)) {
+			return this.errorReply(`Use !show instead.`);
+		}
```
Zarel, line 2521:
```suggestion
		if (user.can('showmedia', null, room, '/show')) return this.errorReply(`Use !show instead.`);
```

This is easier, right?

Zarel:
I asked you to test the simplified version and you said it wasn't working correctly? Can you clarify?

mia-pi-git:
> I asked you to test the simplified version and you said it wasn't working correctly? Can you clarify?

As in, `user.can('showmedia', null, room, '/show')` with the correct settings to replicate the rest of this bug, was returning false. I'll look more into it later today.

AnnikaCodes:
What's the status of this PR?

mia-pi-git:
> What's the status of this PR? The bug report link is no longer working, so I can't even see what the undesirable interaction was :/.

Can't repro the issue anymore. Probably fine to merge?

AnnikaCodes:
If you can't reproduce the issue, why is the PR fixing the bug necessary?

mia-pi-git:
> If you can't reproduce the issue, why is the PR fixing the bug necessary?

I meant that I cannot reproduce the issue that stalled the PR.

AnnikaCodes:
Oh, okay. This is fine then, thanks!

PR #7906, opened 2021-01-11
Polls: Properly calculate total votes in multi polls

mia-pi-git:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-40#post-8685683


Zarel:
I don't understand how this fixes it.

Zarel:
Specifically: The problem isn't that the vote totals _added up_ to more than 100%. That's expected and useful for a multipoll. The problem is that a single answer was more than 100%.

PR #7907, opened 2021-01-11
thing-of-the-day: Fix minor bug

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8710248

### DIFF
--- a/server/chat-plugins/thing-of-the-day.ts
+++ b/server/chat-plugins/thing-of-the-day.ts
@@ -187,7 +187,7 @@ class OtdHandler {
 
 		const nomObj = {
 			nomination: nomination,
-			name: user.name,
+			name: toID(user),
 			userids: user.previousIDs.concat(user.id),
 			ips: user.ips.slice(),
 		};
--- a/server/chat-plugins/thing-of-the-day.ts
+++ b/server/chat-plugins/thing-of-the-day.ts
@@ -230,7 +230,7 @@ class OtdHandler {
 		const entries = [];
 
 		for (const value of this.nominations.values()) {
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Why toID this? Why not just escape the HTML?

server/chat-plugins/thing-of-the-day.ts
```diff
@@ -187,7 +187,7 @@ class OtdHandler {
 
 		const nomObj = {
 			nomination: nomination,
-			name: user.name,
+			name: toID(user),
```
AnnikaCodes, line 190: This shouldn't be a user ID - you should just escape it with `Utils.escapeHTML` (either here or when it's displayed).

AnnikaCodes (changes requested):

server/chat-plugins/thing-of-the-day.ts
```diff
@@ -230,7 +230,7 @@ class OtdHandler {
 		const entries = [];
 
 		for (const value of this.nominations.values()) {
-			entries.push(`<li><b>${value.nomination}</b> <i>(Submitted by ${value.name})</i></li>`);
+			entries.push(`<li><b>${value.nomination}</b> <i>(Submitted by ${Utils.escapeHTML(value.name)})</i></li>`);
```
AnnikaCodes, line 233:
```suggestion
			entries.push(Utils.html`<li><b>${value.nomination}</b> <i>(Submitted by ${value.name})</i></li>`);
```
IIRC we prefer this style.

Zarel:
Thanks!

PR #7908, opened 2021-01-11
Groupchats: Add a button to staffintro to delete the groupchat

mia-pi-git:
Requested / convenient.


Zarel:
I'm not sure "convenient" is the word I would use for being able to accidentally delete a groupchat with a single misclick...

PartMan7:
What about making the value ``/htmlbox Are you sure you want to delete the groupchat? <button ...close>Hai</button>``

I'm aware that the usual PS protocol is to avoid this, but there's no way to undo deleting a groupchat.

Zarel:
Very cool and creative use of `/htmlbox`.

Sadly you can't send uhtml to yourself, so that would still add an unclosable Delete button.

I think a better workaround would be to supporting requiring a confirmation client-side. Maybe with `<button data-requireconfirm="Delete">`.

mia-pi-git:
Closing because this is barely relevant anymore, with groupchat deprecation.

PR #7911, opened 2021-01-12
Unnerve should activate before hazards after phazing

urkerab:
This makes the `PreStart` event happen before the `Start` event, which also incidentally prevents the related double Unnerve activation message.


Zarel:
Thanks!

PR #7913, opened 2021-01-13
Punishments: Support blacklisting ranges from being marked as shared

mia-pi-git:
approved by Aeo - useful feature for ranges with bad users.


AnnikaCodes (approved):
This change looks good to me after reviewing in PMs; I'll wait to merge this for a bit in case others have thoughts.

Zarel:
Looks good; thanks!

PR #7914, opened 2021-01-14
Add missing move to Dracovish BSS set

jeremydorne:


LegoFigure11:
This set intentionally has only three moves.

LegoFigure11:
See https://www.smogon.com/forums/threads/bss-factory.3675374/#post-8701013

PR #7915, opened 2021-01-14
Help Tickets: Improve roomintro CSS

DaWoblefet:
Goes from
image

to 

image


Zarel:
It feels like this is the wrong solution. Something else is bugged if not deleting margins results in a horizontal scrollbar.

Zarel:
It looks like the problem is that `<h2>` is the turn separator, and it wasn't designed to go inside infoboxes.

It's probably easiest to just switch to `<h3>` or something?

PR #7916, opened 2021-01-15
Brick Break should destroy screens if type immunity is bypassed

urkerab:
Currently for Gen 5 or later the sim hard-codes the type immunity when checking whether Brick Break should destroy screens, but it doesn't for Gen 4. This suggests that the order of events was different back in the day. Sadly I can't figure out what the correct order of events is for Gen 4, so I'll just have to limit myself for fixing Gen 5 or later, by removing the explicit type immunity check and relying on type immunity having been checked before this event runs. This fixes using Brick Break against Ghost types in a number of scenarios, such as if the move is Electrified, or type immunity is bypassed through such means as Scrappy or the Ring Target.

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -165,13 +165,6 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		recoil: [1, 3],
 	},
-	brickbreak: {
-		inherit: true,
-		onTryHit(pokemon) {
-			pokemon.side.removeSideCondition('reflect');
-			pokemon.side.removeSideCondition('lightscreen');
-		},
-	},
 	bulletseed: {
 		inherit: true,
 		basePower: 10,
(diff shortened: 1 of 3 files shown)


Marty-D:
Oh, this has probably been broken since the spread moves refactor then. I was confused about your messages earlier implying this doesn't change Gen 4's implementation but now I see it just doesn't work anymore. Thanks!

urkerab:
I don't know if or when Gen 4 ever worked, but I can't work out how it's supposed to work, otherwise I would have taken a crack at fixing it while I was here. In the mean time, the Gen 4 code became redundant, but I didn't want to confuse you into thinking that's because the Gen 4 mechanics were correct.

Marty-D:
The singleEvent and runEvent TryHits used to be in the same place, which is what let the old implementation work. It may not have worked in combination with Foresight and such but it definitely worked against your average Ghost type.

PR #7917, opened 2021-01-15
Fix Soundproof blocking self sound moves in Gen 8

urkerab:
Note that the tests are subtly different:
- In the first test, nobody would lose if both Soundproof work, and if you take both Soundproof away, then player 2 wins (as per the previous existing test), so this shows that only Player 2's Soundproof didn't take effect
- In the second test, Player 2 would win if Player 1's Soundproof didn't work, so this shows that Player 1's Soundproof took effect in this gen where it didn't for Player 2 in the previous test

### DIFF
--- a/data/mods/gen7/abilities.ts
+++ b/data/mods/gen7/abilities.ts
@@ -66,6 +66,15 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 		inherit: true,
 		onBoost() {},
 	},
+	soundproof: {
+		inherit: true,
+		onTryHit(target, source, move) {
+			if (move.flags['sound']) {
+				this.add('-immune', target, '[from] ability: Soundproof');
+				return null;
+			}
+		},
+	},
 	technician: {
 		inherit: true,
 		onBasePowerPriority: 19,
--- a/test/sim/moves/perishsong.js
+++ b/test/sim/moves/perishsong.js
@@ -38,4 +38,28 @@ describe('Perish Song', function () {
 		for (let i = 0; i < 4; i++) { battle.makeChoices(); }
 		assert.equal(battle.winner, 'Player 2');
 	});
+
+	it(`should not affect other Pokemon with the ability Soundproof`, function () {
+		battle = common.createBattle([[
+			{species: 'Weavile', ability: 'soundproof', moves: ['perishsong']},
+		], [
+			{species: 'Slowpoke', ability: 'soundproof', moves: ['sleeptalk']},
+		]]);
+
+		battle.makeChoices();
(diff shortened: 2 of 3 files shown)


DaWoblefet:
999999% sure this is because of Clangorous Soul lol

Confirming that similar behavior happens with something like Howl. Soundproof will block an ally's Howl, but not your own.

test/sim/moves/perishsong.js
```diff
@@ -38,4 +38,26 @@ describe('Perish Song', function () {
...
+
+	it(`should not affect other Pokemon with the ability Soundproof`, function () {
+		battle = common.createBattle([[
+			{species: 'Weavile', ability: 'soundproof', moves: ['sleeptalk']},
+		], [
+			{species: 'Slowpoke', ability: 'soundproof', moves: ['perishsong']},
+		]]);
+
+		for (let i = 0; i < 4; i++) { battle.makeChoices(); }
+		assert.equal(battle.winner, 'Player 1');
```
DaWoblefet, line 50: I think it would be much more straightforward to just check the if the Pokemon has the Perish Song volatile. That way, the test is independent of Perish Song's fainting behavior. You can do ``.volatiles['perishsong']`` to check if it currently has Perish Song (see Protective Pads test vs. Perish Body as an example).

data/mods/gen7/abilities.ts
```diff
@@ -66,6 +66,15 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 		inherit: true,
 		onBoost() {},
 	},
+	soundproof: {
+		inherit: true,
+		onTryHit(target, source, move) {
+			if (move.target !== 'self' && move.flags['sound']) {
```
Marty-D, line 72:
```suggestion
			if (move.flags['sound']) {
```
This was just a naive way to support Clangorous Soul before I made the connection with Howl and Perish Song. https://github.com/smogon/pokemon-showdown/commit/1082258f84e3bf26390279e0168670fc4b705314

urkerab: Fair enough... I've rebased the backout of that commit onto my branch to make it clearer.

test/sim/moves/perishsong.js
urkerab: I was just copying the other test because I suck at writing tests... I'll see what I can do.

urkerab: Well, I did it, but I guess I should have copied and pastd  `battle.p1.active[0].volatiles['perishsong]'` to save me all of that typing...

Marty-D (approved):
👍

DaWoblefet (approved):

PR #7922, opened 2021-01-17
Megas for All: slate 26 + various balance changes

HematiteH2O:
Sweet, I think this passes the checks now! (or at least it does on my end!)
To summarize the changes:

* we banned Spectrier
* we removed four moves from Starmie and added new moves to Leavanny, Klinklang, Exploud, Toxtricity and Noivern
* we changed the stats of Bibarel, Leavanny and Bisharp
* because of slate 26, there are now four new Mega Evolutions (Exploud, Noivern and both Toxtricity) and three Mega Stones accordingly, and two new Abilities for Exploud and Noivern
* then one old Ability (Diamond Dust for Aurorus) was revised as a new weather condition, and I had to alter moves and Abilities that are affected by hail for compatibility with it

That should be all for us to be live again through slate 26! Let me know if anything seems amiss and I will do whatever I can to correct it! C:


Zarel (approved):
Nothing looks obviously wrong to me. I'll merge if Kris is too busy to review over the next dayish.

PR #7930, opened 2021-01-20
Fix Kerfluffle, Jumbao, Astrolotl evo methods

jcheng412:
On the Showdown dex, Kerfluffle's page says its evolution method is "level-up" when it should be "level-up with high Friendship during the day", and Jumbao's page says its evolution method is "level-up" when it should be "level-up with Play Rough". Also, Astrolotl's page says its evolution method is "level-up at level 36" when it should be "level-up at level 35".
Source on Kerfluffle evo method: https://www.smogon.com/forums/threads/cap-22-pre-evo-final-product.3594807/
Source on Jumbao evo method: https://www.smogon.com/forums/threads/cap-24-prevo-final-product.3641453/
Source on Astrolotl evo method (look in Level-up tab): https://www.smogon.com/forums/threads/cap-27-prevo-part-7-movepool-submissions.3667129/#post-8533961


Marty-D (approved):
Good catches, thanks!

PR #7931, opened 2021-01-20
Change help for makegroupchat

sirKiraUzumaki:


server/chat-commands/room-settings.ts
```diff
@@ -784,6 +784,7 @@ export const commands: ChatCommands = {
 	makegroupchathelp: [
 		`/makegroupchat [roomname] - Creates an invite-only group chat named [roomname].`,
 		`/subroomgroupchat [roomname] - Creates a subroom groupchat of the current room. Can only be used in a public room you have staff in.`,
+		`Requires: % @ # &`,
```
xInstruct, line 787: I mean, global voices can make groupchats too.

Might want to modify it to include that I am but not sure how to do it succinctly.

sirKiraUzumaki: maybe it can be like this: Requires: Trusted Users

sirKiraUzumaki: maybe I should add it in line 785

AnnikaCodes, line 787:
```suggestion
		`Only users who are staff in a public room or global auth can make groupchats.`,
```
This is how I would word it.

AnnikaCodes:
Also, as per [our commit message guidelines](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#cm3-grammar), your PR name shouldn't end in a period.

sirKiraUzumaki:
> Also, as per [our commit message guidelines](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#cm3-grammar), your PR name shouldn't end in a period.

Opps sorry, I will make sure not to do this again.

AnnikaCodes (approved):
Looks fine to me. Thanks!

PR #7935, opened 2021-01-21
Fix excluded params in /ms

Kalest3:
The pivot parameters apparently, when excluded with !, were returning all movements, including pivot moves. Same with recoil moves. This PR fix it.


DaWoblefet:
Ummm this should have been fixed already in https://github.com/smogon/pokemon-showdown/pull/7859

The behavior seems accurate on PS too:
image

Is there something else this PR does?

Kalest3:
After aviettran PR, with /nms !pivot, you can see all the movements in the Pokémon. However, it does not exclude pivot moves (you can see that U-turn and Flip Turn are still there). These moves are excluded with my PR.

server/chat-plugins/datasearch.ts
```diff
@@ -1725,20 +1727,22 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
...
-					hasRecovery = !!move.drain || !!move.flags.heal;
-				} else if (recoveryType === "zrecovery") {
-					hasRecovery = (move.zMove?.effect === 'heal');
-				}
-				if (hasRecovery === alts.other[recoveryType]) {
-					matched = true;
-					break;
-				}
-			}
+			if (!zrecoveryUndefined || recoveryUndefined === false) {
```
AnnikaCodes, line 1730:
```suggestion
			if (!zrecoveryUndefined) {
```
If `something` is equal to `false` then `!something` will always be true.

Kalest3: !recovery doesn't work with this change

image

AnnikaCodes: Oh, I misread `recoveryUndefined` as `zrecoveryUndefined`.

Kalest3: well, i tried !recoveryUndefined, and !pivot and !recoil still with errors. I tried recoveryUndefined (without !), and zrecovery doesn't works

AnnikaCodes (commented):
I'm not familiar with the details of movesearch.

server/chat-plugins/datasearch.ts
Zarel: Is there a reason one of them is `=== false` and the other one is `!`? And why they're comparing with undefined instead of the usual existence check?

Kalest3: Indeed. `=== false` is useless as long as both variables are `!`
However, i didn't get positive results on my local server without `undefined`

server/chat-plugins/datasearch.ts
```diff
@@ -1725,20 +1727,22 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
...
+					 hasRecovery = !!move.drain || !!move.flags.heal;
+				 } else if (recoveryType === "zrecovery") {
+					 hasRecovery = (move.zMove?.effect === 'heal');
+				 }
+				 if (hasRecovery === alts.other[recoveryType]) {
+					 matched = true;
+					 break;
+				 }
+			    }
+		    }
```
AnnikaCodes, line 1743:
```suggestion
			if (!zrecoveryUndefined || !recoveryUndefined) {
				for (const recoveryType in alts.other) {
					let hasRecovery = false;
					if (recoveryType === "recovery") {
						hasRecovery = !!move.drain || !!move.flags.heal;
					} else if (recoveryType === "zrecovery") {
						hasRecovery = (move.zMove?.effect === 'heal');
					}
					if (hasRecovery === alts.other[recoveryType]) {
						matched = true;
						break;
					}
				}
			}
```
This isn't indented properly.

AnnikaCodes:
Er, @Zarel @KrisXV, do either of you have opinions on this? I haven't worked on movesearch in a while, and so I'm a little apprehensive about merging this fix.

AnnikaCodes:
bump?

PR #7937, opened 2021-01-22
Add machine-learning based monitors

mia-pi-git:
Aeonic said it was fine to go ahead and PR this now. 
Yes, the static train method is a bit hacky, but training is so heavy on the processes (and it blocks the process entirely since it's sync) that it's the best solution that doesn't involve also holding up the messages.

### DIFF
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -131,6 +131,7 @@ export const chatfilter: ChatFilter = function (message, user, room) {
 			}
 		}
 	})();
+	return undefined;
 };
 
 export const PM = new QueryProcessManager<NetQuery, any>(module, query => {
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -6,7 +6,7 @@
  * @author mia-pi-git
  */
 
-import {QueryProcessManager, QueryProcessWrapper} from '../../lib/process-manager';
+import {QueryProcessManager} from '../../lib/process-manager';
 import {FS} from '../../lib/fs';
 import {Utils} from '../../lib/utils';
 import {Config} from '../config-loader';
@@ -83,7 +83,9 @@ export class NeuralNetChecker {
 		// do the training in its own process
 		const result = await PM.queryTemporaryProcess({type: 'train', data});
 		// load it into the main process that we're querying
-		await PM.query({type: 'load', data: PATH});
+		for (const process of PM.processes) {
+			process.query({type: 'load', data: PATH});
+		}
 		return result;
 	}
 }
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -84,7 +84,7 @@ export class NeuralNetChecker {
 		const result = await PM.queryTemporaryProcess({type: 'train', data});
 		// load it into the main process that we're querying
 		for (const process of PM.processes) {
-			process.query({type: 'load', data: PATH});
+			await process.query({type: 'load', data: PATH});
 		}
 		return result;
 	}
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -79,15 +79,9 @@ export class NeuralNetChecker {
 		// acceptable to drop since training is very slow
 		return result;
 	}
-	static trainingProcesses: QueryProcessWrapper[] = [];
 	static async train(data: TrainingLine[]) {
-		const process = PM.createProcess();
-		this.trainingProcesses.push(process);
 		// do the training in a separate process (which we never add to PM.processes so it doesn't get pinged for other requests)
-		const result = await process.query({type: 'train', data});
-		// kill process
-		this.trainingProcesses.splice(this.trainingProcesses.indexOf(process), 1);
-		process.destroy();
+		const result = await PM.queryTemporaryProcess({type: 'train', data});
 		// load it into the main process that we're querying
 		await PM.query({type: 'load', data: PATH});
 		return result;
@@ -165,6 +159,7 @@ if (!PM.isParentProcess) {
 	// we only want to spawn one network, when it's the subprocess
 	// otherwise, we use the PM for interfacing with the network
 	net = new NeuralNetChecker(PATH);
+	// eslint-disable-next-line no-eval
 	Repl.start('net-filters', cmd => eval(cmd));
 } else {
 	PM.spawn(NUM_PROCESSES);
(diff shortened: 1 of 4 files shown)


server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+			this.model = null;
+		}
+		if (path) this.load(path);
+	}
+	async train(data: TrainingLine[], iterations?: number) {
+		// 100 has good perf but is still effective
+		if (!iterations) iterations = 100;
+		const now = Date.now();
+		await FS(PATH).copyFile(PATH + '.backup');
+		this.model?.train(data, {iterations});
```
Zarel, line 46: something something design principle silent failure

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+		const process = PM.createProcess();
+		this.trainingProcesses.push(process);
+		// do the training in a separate process (which we never add to PM.processes so it doesn't get pinged for other requests)
+		const result = await process.query({type: 'train', data});
+		// kill process
+		this.trainingProcesses.splice(this.trainingProcesses.indexOf(process), 1);
+		process.destroy();
+		// load it into the main process that we're querying
+		await PM.query({type: 'load', data: PATH});
+		return result;
```
Zarel, line 82: I explained in DMs, but there's no reason for you to use a process manager for this. This is exactly what `child_process` is for.

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,217 @@
...
+			const minCount = Config.netfilterlimit || 20;
+			if (hits[room.roomid][user.id] >= minCount) {
+				Rooms.get('upperstaff')?.add(
+					`|c|&|/log [ERPMonitor] Suspicious messages detected in <<${room.roomid}>>`
+				).update();
+				hits[room.roomid][user.id] = 0; // so they can't spam messages
+			}
+		}
+	})();
+	return undefined;
```
Zarel, line 122:
```suggestion
```

server/chat-plugins/net-filters.ts
Zarel: Also: this is already a subprocess! PM was never written for sub-subprocesses.

mia-pi-git: No? This is run from the command in the main process.

Zarel: Oh, I see. Still, this is what `child_process` is for.

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,227 @@
...
+ * @author mia-pi-git
+ */
+
+import {QueryProcessManager, QueryProcessWrapper} from '../../lib/process-manager';
+import {FS} from '../../lib/fs';
+import {Utils} from '../../lib/utils';
+import {Config} from '../config-loader';
+
+const PATH = "config/chat-plugins/net.json";
+const MAX_PROCESSES = Config.netfilterprocesses || 1;
```
Zarel, line 15: Shouldn't this be `NUM_PROCESSES`? In what sense is this a max?

server/chat-plugins/net-filters.ts
mia-pi-git: Modified this in the checkAllowed() function so that it should always loudly fail in the command if it exists.

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,228 @@
...
+			this.model = null;
+		}
+		if (path) this.load(path);
+	}
+	async train(data: TrainingLine[], iterations?: number) {
+		// 100 has good perf but is still effective
+		if (!iterations) iterations = 100;
+		const now = Date.now();
+		await FS(PATH).copyFile(PATH + '.backup');
+		if (!this.model) return;
```
Zarel, line 55: This is still a silent failure...

server/chat.ts
```diff
@@ -92,7 +92,7 @@ export type ChatFilter = ((
 	connection: Connection,
 	targetUser: User | null,
 	originalMessage: string
-) => string | false | null | undefined) & {priority?: number};
+) => string | false | null | undefined | void) & {priority?: number};
```
Zarel, line 95: Oh! That's why you had `return undefined`. I think the `return undefined` opt-in might be better, then.

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,249 @@
...
+		return 'success';
+	}
+}, PM_TIMEOUT);
+
+if (!PM.isParentProcess) {
+	global.Config = Config;
+	// we only want to spawn one network, when it's the subprocess
+	// otherwise, we use the PM for interfacing with the network
+	net = new NeuralNetChecker(PATH);
+	Repl.start('net-filters', cmd => eval(cmd));
```
Zarel, line 168:
```suggestion
	// eslint-disable-next-line no-eval
	Repl.start('net-filters', cmd => eval(cmd));
```

server/chat-plugins/net-filters.ts
mia-pi-git: Updated to handle new refactor!

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,244 @@
...
+		if (!this.model) return result;
+		try {
+			result = this.model.run(data);
+		} catch (e) {}
+		// usually means someone didn't train it, carry on
+		// acceptable to drop since training is very slow
+		return result;
+	}
+	static async train(data: TrainingLine[]) {
+		// do the training in a separate process (which we never add to PM.processes so it doesn't get pinged for other requests)
```
Zarel, line 83:
```suggestion
		// do the training in its own process
```

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,244 @@
...
+		const result = await PM.queryTemporaryProcess({type: 'train', data});
+		// load it into the main process that we're querying
+		await PM.query({type: 'load', data: PATH});
```
Zarel, line 86: Wait, this only works if you have exactly one process... PM isn't designed to be used like that. :s

Zarel: I guess it's fine but `Config.netfilterprocesses` can't exist if you do this.

server/chat-plugins/net-filters.ts
```diff
@@ -0,0 +1,244 @@
+/**
+ * Neural net chatfilters.
+ * These are in a separate file so that they don't crash the other filters.
+ * (issues with globals, etc)
+ * by Mia.
+ * @author mia-pi-git
+ */
+
+import {QueryProcessManager, QueryProcessWrapper} from '../../lib/process-manager';
```
Zarel, line 9:
```suggestion
import {QueryProcessManager} from '../../lib/process-manager';
```

server/chat-plugins/net-filters.ts
mia-pi-git: Alright. We probably can do with that for now? Or I could just iterate each process and query each of them to load them?

Zarel:
I don't see anything obviously wrong, here. Let's see how this goes.

PR #7938, opened 2021-01-22
Fix bug with Hidden Power desc for past gens

Spandamn:


TheImmortal:
I think @Zarel would say use `move.realMove` instead of hardcoding

Spandamn:
it wasnt much of a difference really

Spandamn:
well merge it i guess

DaWoblefet:
@Spandamn I just tried your fix locally and it doesn't work:

image

Spandamn:
Hmmm, looks like a client-server bug

PR #7939, opened 2021-01-22
Default sendhtmlpage to all connections

TheMezStrikes:
figured this was fine given Zarel's comment on https://www.smogon.com/forums/posts/8687858


Zarel:
Thanks!

PR #7944, opened 2021-01-23
UNO: Fix turn order after a disqualification

Distrib-ps:
Fix this bug : https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8727895


AnnikaCodes:
Have you tested this locally?

Distrib-ps:
Yes

PR #7948, opened 2021-01-25
Chat: Implement Discord spoilers

monsanto:
Mostly so A/M can say "spoilers work like Discord", instead of teaching them a new syntax.

Added a new span type `'|'` instead of reusing `'spoiler'` so as to avoid unwanted interactions in `popAllSpans`.


mia-pi-git:
Should be noted you'll need to change the client rendering as well (this is only for like, serverside stuff, not chat messages :p)

monsanto:
Also, the rest of PS staff may want to discuss this before it is merged, so we can let it sit for a bit. Just felt like getting the work out of the way

Zarel (approved):
I agreed with this when chaos suggested it, and I'm happy with the implementation. I'll wait in case the staff has problems with it, but as far as I'm aware, this is fine in its current state.

AnnikaCodes:
Has this been mentioned in the Staff room? As far as I know, only programmers and the Upper Staff discord are aware of it. I support it fwiw.

AnnikaCodes:
I asked Staff room and they're fine with this as long as we help rooms who use regex to ban spoiler tags.

Zarel:
Yeah, there's no reason to stall on this. The only hard part will be recompiling for client.

PR #7949, opened 2021-01-25
Calculator: Support h, o and b suffixes for base 16, 8 and 2 numbers

DayanYT:
Make the calculator support hexadecimal numbers like these: 1h; 16h; Fh etc.
Tested this locally, it works.

### DIFF
--- a/server/chat-plugins/calculator.ts
+++ b/server/chat-plugins/calculator.ts
@@ -97,7 +97,12 @@ function solveRPN(rpn: string[]) {
 			if (!resultStack.length) throw new SyntaxError(`Unknown syntax error`);
 			resultStack.push(-resultStack.pop()!);
 		} else if (!"^%*/+-".includes(token)) {
-			let num = Number(token);
+			let num;
+			if (token.endsWith('h')) {
+			num = Number("0x" + token.slice(0, -1)); //replacing h with 0x
+			} else { 
+			num = Number(token); 
+			}
 			if (isNaN(num) && token.toUpperCase() in Math) {
 				// @ts-ignore
 				num = Math[token.toUpperCase()];
@@ -141,7 +146,7 @@ export const commands: ChatCommands = {
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
 		let base = 10;
-		if (target.includes('0x')) {
+		if (target.includes('0x') || target.includes('h')) {
 			base = 16;
 		} else if (target.includes('0o')) {
 			base = 8;
@@ -151,7 +156,7 @@ export const commands: ChatCommands = {
 
 		const baseMatchResult = (/\b(?:in|to)\s+([a-zA-Z]+)\b/).exec(target);
 		if (baseMatchResult) {
-			switch (toID(baseMatchResult[1])) {
+			switch (toID(baseMatchResult[1])) {				
 			case 'decimal': case 'dec': base = 10; break;
 			case 'hexadecimal': case 'hex': base = 16; break;
 			case 'octal': case 'oct': base = 8; break;
--- a/server/chat-plugins/calculator.ts
+++ b/server/chat-plugins/calculator.ts
@@ -92,17 +92,22 @@ function parseMathematicalExpression(infix: string) {
 
 function solveRPN(rpn: string[]) {
 	const resultStack: number[] = [];
-	for (const token of rpn) {
+	for (let token of rpn) {
 		if (token === 'negative') {
 			if (!resultStack.length) throw new SyntaxError(`Unknown syntax error`);
 			resultStack.push(-resultStack.pop()!);
 		} else if (!"^%*/+-".includes(token)) {
-			let num;
 			if (token.endsWith('h')) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/calculator.ts
```diff
@@ -97,7 +97,12 @@ function solveRPN(rpn: string[]) {
 			if (!resultStack.length) throw new SyntaxError(`Unknown syntax error`);
 			resultStack.push(-resultStack.pop()!);
 		} else if (!"^%*/+-".includes(token)) {
-			let num = Number(token);
+			let num;
+			if (token.endsWith('h')) {
+			num = Number("0x" + token.slice(0, -1)); //replacing h with 0x
+			} else { 
+			num = Number(token); 
+			}
```
AnnikaCodes, line 105:
```suggestion
			if (token.endsWith('h')) token = `0x${token.slice(0, -1)}`;
			let num = Number(token);
```
This is how I would do it (converting `const token` to `let token`).

AnnikaCodes: I think this is more readable than `let num;` but if you'd rather do that, please at least indent the inner part of your `if` statements.

server/chat-plugins/calculator.ts
```diff
@@ -151,7 +156,7 @@ export const commands: ChatCommands = {
 
 		const baseMatchResult = (/\b(?:in|to)\s+([a-zA-Z]+)\b/).exec(target);
 		if (baseMatchResult) {
-			switch (toID(baseMatchResult[1])) {
+			switch (toID(baseMatchResult[1])) {
```
AnnikaCodes, line 159:
```suggestion
			switch (toID(baseMatchResult[1])) {
```

AnnikaCodes:
I wonder if `o` should also be supported as a suffix for base 8 numbers?

Zarel:
Is the `h` suffix common for hex numbers?

> Some authors prefer a text subscript, such as 159<sub>decimal</sub> and 159<sub>hex</sub>, or 159<sub>d</sub> and 159<sub>h</sub>.

(Wikipedia)

Interesting.

> In programming languages, octal literals are typically identified with a variety of prefixes, including the digit 0, the letters o or q, the digit–letter combination 0o, or the symbol & or $. In Motorola convention, octal numbers are prefixed with @, whereas a small (or capital) letter o or q is added as a postfix following the Intel convention.

(Wikipedia)

Looks like the `h`, `o`, and `b` suffixes are Intel convention, and it's probably fine to use them here, too.

server/chat-plugins/calculator.ts
```diff
@@ -141,11 +151,11 @@ export const commands: ChatCommands = {
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
 		let base = 10;
-		if (target.includes('0x')) {
+		if (target.includes('0x') || target.includes('h')) {
 			base = 16;
-		} else if (target.includes('0o')) {
+		} else if (target.includes('o')) {
```
AnnikaCodes, line 156: This would be a problem if the calculator were ever expanded to support more functions (`cos` comes to mind; this would make all expressions involving cosines use base 8). I don't think we're planning to do that, but it's worth noting.

AnnikaCodes (approved):
Aside from my minor nitpick, I think this looks good.

server/chat-plugins/calculator.ts
```diff
@@ -141,11 +151,11 @@ export const commands: ChatCommands = {
 	calculate(target, room, user) {
 		if (!target) return this.parse('/help calculate');
 		let base = 10;
-		if (target.includes('0x')) {
+		if (target.includes('0x') || target.includes('h')) {
```
Zarel, line 154:
```suggestion
		if (target.includes('0x') || /\bh\b/.test(target)) {
```

server/chat-plugins/calculator.ts
Zarel:
```suggestion
		} else if (/\bo\b/.test(target)) {
```

server/chat-plugins/calculator.ts
```diff
@@ -141,11 +151,11 @@ export const commands: ChatCommands = {
...
 			base = 16;
-		} else if (target.includes('0o')) {
+		} else if (target.includes('o')) {
 			base = 8;
-		} else if (target.includes('0b')) {
+		} else if (target.includes('b')) {
```
Zarel, line 158:
```suggestion
		} else if (/\bb\b/.test(target)) {
```

server/chat-plugins/calculator.ts
Zarel: Probably the easiest way to prevent that false positive.

Zarel: Actually, in hindsight, that's not a good approach, either.

Zarel:
Thanks! I pushed a better base detection system here: https://github.com/smogon/pokemon-showdown/commit/93e921360eb85d8a20d3bc87ab52fffa31684ac8

PR #7889, opened 2021-01-07
SSB4: Bugfixes and Balance Patches

xInstruct:
Turns out some of the "viable" moves from Data Corruption weren't viable at all, and were removed and added some other viable moves that were missed.

Also cut down on archas spam

PR #7954, opened 2021-01-27
SSB4: Balance changes and Bugfixes, pt 2

xInstruct:
I was originally going to do it in #7889 but I didn't want the already written code there to rot.


xInstruct:
I'd prefer to have this reviewed and possibly implemented into PS before Feb, pls.

Thanks in advance

xInstruct:
Alright this is all set

PR #7957, opened 2021-01-28
Users: Automatically expire inactive connections

mia-pi-git:
This should hopefully help alleviate sockets load + clear up the bug with users staying online despite disconnecting.


server/users.ts
```diff
@@ -50,6 +50,7 @@ import {Auth, GlobalAuth, PLAYER_SYMBOL, HOST_SYMBOL, RoomPermission, GlobalPerm
 const MINUTES = 60 * 1000;
 const IDLE_TIMER = 60 * MINUTES;
 const STAFF_IDLE_TIMER = 30 * MINUTES;
+const CONNECTION_EXPIRY_TIME = 3 * 60 * MINUTES;
```
Zarel, line 53:
```suggestion
const CONNECTION_EXPIRY_TIME = 24 * 60 * MINUTES;
```

This should be much less likely to accidentally hit something it shouldn't.

Zarel:
Fine by me.

PR #7959, opened 2021-01-29
Gen 1: Move Graveler to NFE

PlagueVonKarma:
I originally brought this up with FOMG, and since then, clarification has been given. The RBY Council supports moving Graveler to NFE since Golem has dropped to UU (and will drop to NU should it be made official). There's a screenshot **[here](https://i.imgur.com/dhEt44O.png)**, I guess. 

FOMG originally told Kris but nothing seems to have happened for the past 2 and a half days, so I decided to make the PR myself rather than saddle him with more work. 

Should be mentioned that NU Live Tours in RoA were using Graveler all this time kekw


KrisXV:
that was my bad i actually just blanked on it, and the build error is irrelevant to this so merging

PR #7960, opened 2021-01-29
Calculate and display Elo change on server

DylanSp:
`ladders-local.ts` already had the calculations for Elo change; I factored that out into a separate file that's called from both `ladders-local` and `ladders-remote`. `ladders-remote` now calculates Elo change immediately and displays it while waiting for the login server to return from updating the ladder.

Fixes relevant bullet point on #2444.


server/elo.ts
```diff
@@ -0,0 +1,27 @@
+export function calculateElo(previousUserElo: number, score: number, foeElo: number): number {
+	// The K factor determines how much your Elo changes when you win or
+	// lose games. Larger K means more change.
+	// In the "original" Elo, K is constant, but it's common for K to
+	// get smaller as your rating goes up
+	let K = 50;
+
+	// dynamic K-scaling (optional)
+	if (previousUserElo < 1200) {
```
DylanSp, line 9: Just noticed that these numbers differ from the loginserver (see https://github.com/smogon/pokemon-showdown-client/blob/master/lib/ntbb-ladder.lib.php#L395). Do `ladders-local` and `ladders-remote` need different dynamic K-scaling?

DylanSp: Per Zarel in Discord, they're intentionally different due to the high activity of OU/randbats on the main server. I'll adjust this PR.

DylanSp: Separated out, main server calc now matches the loginserver numbers with 2c8d3c5.

server/ladders-remote.ts
```diff
@@ -70,25 +70,46 @@ export class LadderStore {
...
 		});
+
+		// calculate new Elo scores and display to room while loginserver updates the ladder
+		const [p1OldElo, p2OldElo] = (await Promise.all([this.getRating(p1!.id), this.getRating(p2!.id)])).map(Math.round);
+		const p1NewElo = Math.round(this.calculateElo(p1OldElo, p1score, p2OldElo));
+		const p2NewElo = Math.round(this.calculateElo(p2OldElo, 1 - p1score, p1OldElo));
+
+		room.update();
+		room.send(`||Ladder updating...`);
+
```
Zarel, line 88:
```suggestion
```

These are no longer necessary now that the rest is synchronous.

server/ladders-remote.ts
```diff
@@ -70,25 +70,46 @@ export class LadderStore {
...
-		room.update();
-		room.send(`||Ladder updating...`);
-		const [data, error] = await LoginServer.request('ladderupdate', {
+
+		const ladderUpdatePromise = LoginServer.request('ladderupdate', {
 			p1: p1name,
 			p2: p2name,
 			score: p1score,
 			format: formatid,
 		});
```
Zarel, line 79: Could this just be moved down to where `await ladderUpdatePromise` is?

server/ladders-remote.ts
```diff
@@ -149,4 +140,37 @@ export class LadderStore {
 	static async visualizeAll(username: string) {
 		return [`<tr><td><strong>Please use the official client at play.pokemonshowdown.com</strong></td></tr>`];
 	}
+
+	/**
+	 * Calculates Elo for quick display, matching the formula on loginserver
+	 */
+	// see lib/ntbb-ladder.lib.php in the pokemon-showdown-client repo for the login server implementation
+	// *intentionally* different from calculation in ladders-local, due to the high activity on the main server
```
Zarel, line 148: Wait, in hindsight I think the only difference is because of ladder decay, which isn't implemented here.

Is the only difference here the `1300` vs the `1350`? Maybe we *should* merge them.

DylanSp: That's not the only difference; the numbers for scaling when elo < 1200 are also different.
ladders-local:
```
if (score < 0.5) {
  K = 10 + (previousUserElo - 1000) * 40 / 200;
} else if (score > 0.5) {
  K = 90 - (previousUserElo - 1000) * 40 / 200;
}
```

ladders-remote:
```
if (score < 0.5) {
  K = 20 + (previousUserElo - 1000) * 30 / 100;
} else if (score > 0.5) {
  K = 80 - (previousUserElo - 1000) * 30 / 100;
}
```

ladders-remote uses 20/80 instead of 10/90, and multiplies by 30 instead of 40.

server/ladders-remote.ts
DylanSp: This allows the loginserver's processing to happen concurrently with the server running and displaying the Elo changes. It's admittedly a small optimization, given that the Elo calc doesn't take much time; up to you if you want to change it.

server/ladders-remote.ts
DylanSp: Updated.

server/ladders-remote.ts
Zarel: Interesting! We probably want to investigate the formulas more, later. But that's probably a job for a different PR.

server/ladders-remote.ts
Zarel: Oh, I see! I think this is fine.

Zarel:
This is great! Thanks!

PR #7961, opened 2021-01-29
Chat: Track slow async commands

mia-pi-git:
The room was Zarel's suggestion - more accessible.
As a side note, I set errors to be logged since I figure knowing something lasted >3000ms before crashing is still useful knowledge.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -127,6 +127,7 @@ import {formatText, linkRegex, stripFormatting} from './chat-formatter';
 
 // @ts-ignore no typedef available
 import ProbeModule = require('probe-image-size');
+import { Monitor } from './monitor';
 const probe: (url: string) => Promise<{width: number, height: number}> = ProbeModule;
 
 const EMOJI_REGEX = /[\p{Emoji_Modifier_Base}\p{Emoji_Presentation}\uFE0F]/u;
@@ -1629,8 +1630,21 @@ export const Chat = new class {
 
 		const initialRoomlogLength = room?.log.getLineCount();
 		const context = new CommandContext({message, room, user, connection});
+		const start = Date.now();
 		const result = context.parse();
-
+		if (typeof result?.then === 'function') {
+			void result.then(() => {
+				const delta = Date.now() - start;
+				if (delta > 3000) {
+					this.logSlowMessage(delta, context);
+				}
+			});
+		} else {
+			const delta = Date.now() - start;
+			if (delta > 1000) {
+				this.logSlowMessage(delta, context);
+			}
+		}
 		if (room && room.log.getLineCount() !== initialRoomlogLength) {
 			room.messagesSent++;
 			for (const [handler, numMessages] of room.nthMessageHandlers) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Created the room on Showdown.

server/users.ts
```diff
@@ -1676,7 +1676,8 @@ function socketReceive(worker: StreamWorker, workerid: number, socketid: string,
 	}
 	const deltaTime = Date.now() - startTime;
 	if (deltaTime > 1000) {
-		Monitor.warn(`[slow] ${deltaTime}ms - ${user.name} <${connection.ip}>: ${roomId}|${message}`);
+		const logRoom = Rooms.get('slowlog');
+		(logRoom ? logRoom.add : Monitor.warn)(`[slow] ${deltaTime}ms - ${user.name} <${connection.ip}>: ${roomId}|${message}`);
```
Zarel, line 1680: Why do you have one log in `chat` and one log in `users`? They should both be in one place, shouldn't they?

mia-pi-git: lol i just said that in DMs 
ill move it to chat as well

Zarel: Putting them both in `users` is generally better, because they're both in `users`, but we might need to be better at propagating the promises.

server/chat.ts
```diff
@@ -1629,8 +1630,21 @@ export const Chat = new class {
...
+				if (delta > 3000) {
+					this.logSlowMessage(delta, context);
+				}
+			});
+		} else {
+			const delta = Date.now() - start;
+			if (delta > 1000) {
+				this.logSlowMessage(delta, context);
+			}
+		}
```
Zarel, line 1647:
```suggestion
		if (typeof result?.then === 'function') {
			void result.then(() => {
				const timeUsed = Date.now() - start;
				if (timeUsed > 3000) {
					this.logSlowMessage(delta, context);
				}
			});
		} else {
			const timeUsed = Date.now() - start;
			if (timeUsed > 1000) {
				this.logSlowMessage(delta, context);
			}
		}
```

server/chat.ts
```diff
@@ -1640,6 +1654,19 @@ export const Chat = new class {
...
+			`[slow] ${deltaTime}ms - ${context.user.name} (${context.connection.ip}): ` +
+			`<${context.room ? context.room.roomid : context.pmTarget ? `PM:${context.pmTarget?.name}` : 'CMD'}> ` +
+			`${context.message.replace(/\n/ig, ' ')}`
+		);
+		if (logRoom) {
+			logRoom.add(`|c|&|/log ` + logMessage).update();
+		} else {
+			Monitor.warn(logMessage);
+		}
+	}
```
Zarel, line 1669:
```suggestion
	logSlowMessage(timeUsed: number, context: CommandContext) {
		const logRoom = Rooms.get('slowlog');
		const logMessage = (
			`[slow] ${timeUsed}ms - ${context.user.name} (${context.connection.ip}): ` +
			`<${context.room ? context.room.roomid : context.pmTarget ? `PM:${context.pmTarget?.name}` : 'CMD'}> ` +
			`${context.message.replace(/\n/ig, ' ')}`
		);
		if (logRoom) {
			logRoom.add(`|c|&|/log ` + logMessage).update();
		} else {
			Monitor.warn(logMessage);
		}
	}
```

server/chat.ts
```diff
@@ -127,6 +127,7 @@ import {formatText, linkRegex, stripFormatting} from './chat-formatter';
 
 // @ts-ignore no typedef available
 import ProbeModule = require('probe-image-size');
+import { Monitor } from './monitor';
```
Zarel, line 130:
```suggestion
```

PR #7963, opened 2021-01-29
Add a command to change battle seeds

mia-pi-git:
Discussed in #sim-dev.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -968,9 +968,7 @@ export const commands: ChatCommands = {
 	async reseedbattle(target, room, user) {
 		room = this.requireRoom();
 		if (!room.battle) return this.errorReply(`This command can only be used in battle.`);
-		if (room.auth.get(user.id) !== Users.HOST_SYMBOL) {
-			this.canUseConsole();
-		}
+		this.checkCan('editbattle');
 		let seed;
 		if (target) {
 			const parts = target.split(',').map(parseInt);
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -965,6 +965,26 @@ export const commands: ChatCommands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ &`],
 
+	async reseedbattle(target, room, user) {
+		room = this.requireRoom();
+		if (!room.battle) return this.errorReply(`This command can only be used in battle.`);
+		if (room.auth.get(user.id) !== Users.HOST_SYMBOL) {
+			this.canUseConsole();
+		}
+		let seed;
+		if (target) {
+			const parts = target.split(',').map(parseInt);
+			if (target.length !== 4) return this.errorReply(`Invalid seed length - must be 4.`);
+			if (parts.some(i => isNaN(i))) {
+				return this.errorReply(`Invalid seed - must be all numbers`);
+			}
+			seed = parts as [number, number, number, number];
+		}
+		const newSeed = await room.battle.reseed(seed);
+		room.addByUser(user, `${user.name} reseeded this battle`).update();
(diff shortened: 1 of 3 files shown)


DaWoblefet:
What is stopping from either player arbitrarily rerolling the seed during a tournament match?

mia-pi-git:
> What is stopping from either player arbitrarily rerolling the seed during a tournament match? Or a ladder match, etc.?

This checks host symbol, which is different than the player symbol, and only exists on a player if an admin forcepromotes them or if it's an imported replay.

Zarel:
I mentioned before this is should be a part of `editbattle`.

server/chat-commands/core.ts
```diff
@@ -965,6 +965,26 @@ export const commands: ChatCommands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ &`],
 
+	async reseedbattle(target, room, user) {
+		room = this.requireRoom();
+		if (!room.battle) return this.errorReply(`This command can only be used in battle.`);
+		if (room.auth.get(user.id) !== Users.HOST_SYMBOL) {
```
Zarel, line 971: We have an entire permission system for this...

server/room-battle.ts
```diff
@@ -1203,6 +1217,12 @@ export class RoomBattleStream extends BattleStream {
 			this.battle.inputLog.push(`>chat ${message}`);
 			this.battle.add('chat', `${message}`);
 			break;
+		case 'reseed':
```
Zarel, line 1220: Putting this in `RoomBattleStream` instead of `BattleStream` makes it incompatible with `./pokemon-showdown simulate-battle`, which is bad.

sim/battle.ts
```diff
@@ -286,6 +286,12 @@ export class Battle {
 		this.prng = new PRNG(this.prng.startingSeed);
 	}
 
+	reseed(newSeed?: [number, number, number, number]) {
```
Zarel, line 289: We already have a `resetRNG` function which could have been used for this. We also have a `PRNGSeed` type so you don't need to write out the entire tuple.

sim/battle.ts
```diff
@@ -286,6 +286,12 @@ export class Battle {
...
+		this.prng = new PRNG(newSeed);
+		(this.prngSeed as any) = this.prng;
```
Zarel, line 291: This does absolutely nothing... `this.prng` is a pointer.

server/room-battle.ts
```diff
@@ -730,12 +730,18 @@ export class RoomBattle extends RoomGames.RoomGame {
 		for (const player of this.players) player.wantsTie = false;
 
 		switch (lines[0]) {
-		case 'requesteddata':
+		case 'reseeded': {
```
Zarel, line 733: I guess this is only necessary to privately tell the host what the seed was. I think that's basically unnecessary anyway (the host can grab it from the inputlog after the battle is over if they really care), so my implementation doesn't use this.

server/chat-commands/core.ts
```diff
@@ -965,6 +965,24 @@ export const commands: ChatCommands = {
 	},
 	importinputloghelp: [`/importinputlog [inputlog] - Starts a battle with a given inputlog. Requires: + % @ &`],
 
+	async reseedbattle(target, room, user) {
```
Zarel, line 968: I said before, but this should be part of the `/editbattle` command, not its own command.

server/chat-commands/core.ts
```diff
@@ -965,6 +965,24 @@ export const commands: ChatCommands = {
...
+		room = this.requireRoom();
+		if (!room.battle) return this.errorReply(`This command can only be used in battle.`);
+		this.checkCan('editbattle');
```
Zarel, line 971: It turns out editbattle uses the `forcewin` permission; it doesn't have its own permission.

PR #7965, opened 2021-01-29
Roomsettings: Support a per-room default mod for /dt

mia-pi-git:
Useful utility.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -118,6 +118,7 @@ export interface RoomSettings {
 	repeats?: RepeatedPhrase[];
 	autoModchat?: {rank: GroupSymbol, time: number, active: boolean};
 	tournaments?: TournamentRoomSettings;
+	detailsMod?: string;
 
 	scavSettings?: AnyObject;
 	scavQueue?: QueuedHunt[];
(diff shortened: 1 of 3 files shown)


Zarel:
Don't rooms already have a default format for other commands? Did we never implement that?

Zarel:
I think a default format makes more sense than a default mod.

server/rooms.ts
```diff
@@ -118,6 +118,7 @@ export interface RoomSettings {
 	repeats?: RepeatedPhrase[];
 	autoModchat?: {rank: GroupSymbol, time: number, active: boolean};
 	tournaments?: TournamentRoomSettings;
+	detailsMod?: string;
```
Zarel, line 121: I would call this `defaultFormat` and plan for it to be eventually used for all relevant commands, such as `/learn`.

mia-pi-git:
> Don't rooms already have a default format for other commands? Did we never implement that?

No, we did not. You may be thinking of data command tier display? But that's just singles / doubles.

Zarel:
Like I said in DMs, this should be `defaultFormat`.

xInstruct:
> Don't rooms already have a default format for other commands? Did we never implement that?

afaik the only thing that rooms can edit /dt is the smogon tier (e.g. LC room has it to use doubles tiers as opposed to standard singles)

Zarel:
(I'm working on this now)

DaWoblefet:
This was superseded by https://github.com/smogon/pokemon-showdown/pull/8237

PR #7968, opened 2021-01-30
PM: Support querying in a new process

Zarel:
Better subprocess query system for #7937.

I mentioned in #7937 that it should be a regular `child_process`, but Mia's not _wrong_ that PM has some useful features that make it more convenient than `child_process`. This change allows a `QueryProcessManager` to easily run a query in a new child process.

### DIFF
--- a/lib/process-manager.ts
+++ b/lib/process-manager.ts
@@ -445,29 +445,39 @@ export abstract class ProcessManager<T extends ProcessWrapper = ProcessWrapper> 
 		}
 	}
 	unspawn() {
-		const released = [];
-		const processes = this.processes;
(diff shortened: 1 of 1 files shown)


lib/process-manager.ts
```diff
@@ -517,6 +525,12 @@ export class QueryProcessManager<T = string, U = string> extends ProcessManager
 		clearTimeout(timeout);
 		return result;
 	}
+	queryTemporaryProcess(input: T, force?: boolean) {
```
Zarel, line 528: I don't know how to feel about these function names, mostly.

`queryNewProcess`?

PR #7972, opened 2021-01-31
Lib: Add an index file for all exports

mia-pi-git:
Discussed with @Zarel. Ideally, this can be a standalone module next.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -15,179 +15,180 @@
  *   that if it's English-specific, it should be left out of here.
  */
 
-type Comparable = number | string | boolean | Comparable[] | {reverse: Comparable};
+export type Comparable = number | string | boolean | Comparable[] | {reverse: Comparable};
 
-export const Utils = new class {
-	/**
-	 * Safely converts the passed variable into a string. Unlike '' + str,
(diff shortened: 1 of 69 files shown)


lib/utils.ts
```diff
@@ -199,155 +200,157 @@ export const Utils = new class {
...
+}
+
+export function waitUntil(time: number): Promise<void> {
+	return new Promise(resolve => {
+		setTimeout(() => resolve(), time - Date.now());
+	});
+}
+
+// backwards compatibility
+export const Utils = module.exports;
```
Zarel, line 356: I'd prefer not to mix CommonJS and ES modules. It's probably better to just name every function again here.

mia-pi-git: Alright.

KrisXV (approved):

Zarel:
tbh I didn't expect you to still import everything individually. It doesn't seem much different from before. :s

Zarel:
Writing down what we discussed in DMs:

I do still want to, at some point, release a generalized library called `libio`, which exports a variable named `IO`, with most things in `lib` under it.

But I think this is pretty fine for now. I think we should think harder of how we'd want to package `libio` if we wanted to go much further along that route.

PR #7973, opened 2021-01-31
Accuracy on Bright Powder+Lax Incense

Flethan:
Changed the accuracy modifier on Bright Powder and Lax Incense from 0x0F1C (eff. 94.4%) to 0x0E66 (eff. 89.99%).


urkerab:
I typed `3868` into my hex converter instead of `3686`. I must be dyslexic...

Marty-D:
Good catch, thanks!

PR #7974, opened 2021-01-31
NN filters: Fix error handling and make improvements

mia-pi-git:
Few bug fixes:
- checks if a save exists when copying to a backup
- saves replays of the battle if it alerts US to suspicious behavior (if it's a battle)
- /netfilter train now ignores blank lines
- child process has proper error handling.


Zarel:
Whenever you say "Net" I think of the `lib/net` library. Let's switch to saying "NN"? It's a common abbreviation for Neural Network.

mia-pi-git:
NN it is, then.

PR #7978, opened 2021-02-01
Spotlights: Store image data

mia-pi-git:
Should improve the performance when rendering (not having to request for each spotlight is less laggy!)


Zarel:
ty

PR #7980, opened 2021-02-01
Youtube: Notify room of starting groupwatches

mia-pi-git:
Requested by YT staff.


Zarel:
I can imagine users hating this, tbh.

Zarel:
If this makes a bunch of users disable notifications I'm going to be very mad. :/

PR #7981, opened 2021-02-01
NN Filters: Add a command to train on logs from a room

mia-pi-git:
Easier to mass train this way.

### DIFF
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -22,7 +22,7 @@ interface NetQuery {
 	type: "run" | "train" | "save" | "load" | 'trainfrom';
 	options?: AnyObject;
 }
-// @ts-ignore
+// @ts-ignore in case the optional dependency is not installed
 type LSTM = import('brain.js').recurrent.LSTM;
 
 interface TrainingLine {
--- a/server/chat-plugins/net-filters.ts
+++ b/server/chat-plugins/net-filters.ts
@@ -7,10 +7,10 @@
  */
 
 import {QueryProcessManager} from '../../lib/process-manager';
-import {FS} from '../../lib/fs';
-import {Utils} from '../../lib/utils';
+import {FS, Utils} from '../../lib/';
 import {Config} from '../config-loader';
 import {Repl} from '../../lib/repl';
+import {linkRegex} from '../chat-formatter';
 
 const PATH = "config/chat-plugins/net.json";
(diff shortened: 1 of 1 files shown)


server/chat-plugins/net-filters.ts
```diff
@@ -38,8 +39,12 @@ function modelExists() {
 	return true;
 }
 
+function toRoomID(room: RoomID | Room) {
+	return ("" + room).toLowerCase().replace(/[^a-z0-9\-]+/g, '');
```
Zarel, line 43:
```suggestion
	return ("" + room).toLowerCase().replace(/[^a-z0-9-]+/g, '');
```

server/chat-plugins/net-filters.ts
```diff
@@ -7,21 +7,22 @@
...
 const WHITELIST = ["mia"];
 
 interface NetQuery {
-	data: string | TrainingLine[];
-	type: "run" | "train" | "save" | "load";
+	data: string | TrainingLine[] | AnyObject;
+	type: "run" | "train" | "save" | "load" | 'trainfrom';
 	options?: AnyObject;
 }
+type Model = import('brain.js').recurrent.LSTM;
```
Zarel, line 25: You can `ts-ignore` this if you want it to be `any` if `brain.js` isn't installed, I think?

I might also suggest calling it `LSTM` to match what brain.js calls it.

mia-pi-git: Alright.

server/chat-plugins/net-filters.ts
```diff
@@ -50,14 +55,40 @@ export class NeuralNetChecker {
 	}
 	async train(data: TrainingLine[], iterations?: number) {
 		// 100 has good perf but is still effective
```
Zarel, line 57: something something people hate comments because they're frequently out of date

server/chat-plugins/net-filters.ts
```diff
@@ -7,21 +7,23 @@
...
 const WHITELIST = ["mia"];
 
 interface NetQuery {
-	data: string | TrainingLine[];
-	type: "run" | "train" | "save" | "load";
+	data: string | TrainingLine[] | AnyObject;
+	type: "run" | "train" | "save" | "load" | 'trainfrom';
 	options?: AnyObject;
 }
+// @ts-ignore
```
Zarel, line 25:
```suggestion
// @ts-ignore in case the optional dependency is not installed
```

PR #7983, opened 2021-02-02
Fix crash in calculator.ts

Volco:


Zarel:
Oh, heh, I guess it makes sense that TypeScript supports it but sucrase doesn't.

Zarel:
Thanks!

Volco:
No problem

PR #7989, opened 2021-02-03
/battlefactory: Fix searching for nonexistent tiers

xInstruct:
Searching for a Monotype or LC set in Gen 6 battle factory causes PS to puke so here's a fix.

### DIFF
--- a/server/chat-plugins/random-battles.ts
+++ b/server/chat-plugins/random-battles.ts
@@ -102,7 +102,7 @@ function battleFactorySets(species: string | Species, tier: string | null, gen =
 	if (!isBSS) {
 		if (!tier) return {e: `Please provide a valid tier.`};
 		if (!(toID(tier) in TIERS)) return {e: `That tier isn't supported.`};
-		if (['Mono', 'LC'].includes(tier) && genNum < 7) {
+		if (['Mono', 'LC'].includes(TIERS[toID(tier)]) && genNum < 7) {
 			return {e: `That tier is not included for that generation of Battle Factory.`};
 		}
 		const t = statsFile[TIERS[toID(tier)]];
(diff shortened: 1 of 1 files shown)


server/chat-plugins/random-battles.ts
```diff
@@ -102,6 +102,7 @@ function battleFactorySets(species: string | Species, tier: string | null, gen =
 	if (!isBSS) {
 		if (!tier) return {e: `Please provide a valid tier.`};
 		if (!(toID(tier) in TIERS)) return {e: `That tier isn't supported.`};
+		if (['Mono', 'LC'].includes(tier) && genNum < 7) return {e: `That tier is not included for that generation of Battle Factory.`};
```
KrisXV, line 105:
```suggestion
		if (['Mono', 'LC'].includes(tier) && genNum < 7) {
			return {e: `That tier is not included for that generation of Battle Factory.`};
		}
```

server/chat-plugins/random-battles.ts
```diff
@@ -102,6 +102,9 @@ function battleFactorySets(species: string | Species, tier: string | null, gen =
 	if (!isBSS) {
 		if (!tier) return {e: `Please provide a valid tier.`};
 		if (!(toID(tier) in TIERS)) return {e: `That tier isn't supported.`};
+		if (['Mono', 'LC'].includes(tier) && genNum < 7) {
```
KrisXV, line 105:
```suggestion
		if (['Mono', 'LC'].includes(TIERS[toID(tier)]) && genNum < 7) {
```

server/chat-plugins/random-battles.ts
```diff
@@ -102,6 +102,9 @@ function battleFactorySets(species: string | Species, tier: string | null, gen =
...
+			return {e: `That tier is not included for that generation of Battle Factory.`};
```
KrisXV, line 106:
```suggestion
			return {e: `${TIERS[toID(tier)]} is not included in [Gen ${genNum}] Battle Factory.`};
```

KrisXV (approved):

KrisXV:
ill come up with a less hardcode-y change later but this works for now

PR #7990, opened 2021-02-03
ProcessManager: Handle Monitor#slow in subprocesses

mia-pi-git:
Fixes ``A simulator process crashed: TypeError: Monitor.slow is not a function``

### DIFF
--- a/lib/process-manager.ts
+++ b/lib/process-manager.ts
@@ -106,16 +106,18 @@ export class QueryProcessWrapper<T, U> implements ProcessWrapper {
 	process: ChildProcess;
 	taskId: number;
 	pendingTasks: Map<number, (resp: U) => void>;
+	messageCallback: ((message: string) => any) | null;
 	pendingRelease: Promise<void> | null;
 	resolveRelease: (() => void) | null;
 	debug?: string;
 
-	constructor(file: string) {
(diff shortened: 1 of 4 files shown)


lib/process-manager.ts
```diff
@@ -124,6 +124,11 @@ export class QueryProcessWrapper<T, U> implements ProcessWrapper {
 				throw error;
 			}
 
+			if (message.startsWith(`SLOW\n`)) {
+				Monitor.slow(message.slice(5));
```
Zarel, line 128: `Monitor` is not a Node.js global and can't be called from inside PM.

What we might want is to set up a listener inside PM for custom debug messages, which we could customize from the server.

DaWoblefet:
@Zarel apparently side servers are reporting lots of crashes that this PR would fix; it hasn't been reported publicly but apparently lots of people are encountering it.

mia-pi-git:
> @Zarel apparently side servers are reporting lots of crashes that this PR would fix; it hasn't been reported publicly but apparently lots of side server owners are encountering it.

(this also _has_ been an issue on main, search `TypeError: Monitor.slow is not a function` in crashlog emails)

Zarel:
Thanks!

PR #7992, opened 2021-02-04
Fix crash in hangman

Volco:
If someone adds random words but then they're all deleted afterwards causes a crash whenever ``/hangman random`` is used in that room.


Volco:
Since someone did this

PR #7993, opened 2021-02-04
NN filters: Support canceling queries

mia-pi-git:
Useful thing i found was necessary.


server/chat-plugins/net-filters.ts
```diff
@@ -309,14 +320,35 @@ export const commands: ChatCommands = {
...
+		cancel: 'stop',
+		stop(target, room, user) {
+			checkAllowed(this);
+			let count = 0;
+			if (!PM.releasingProcesses.length) {
+				return this.errorReply(`No train tasks are pending`);
+			}
+			for (const subProcess of PM.releasingProcesses) {
+				for (const [task, resolve] of subProcess.pendingTasks) {
+					resolve({error: `The pending train query was cancelled.`});
```
Zarel, line 340: This is all the releasing processes, not just pending train queries, though.

server/chat-plugins/net-filters.ts
```diff
@@ -309,14 +320,35 @@ export const commands: ChatCommands = {
...
+					subProcess.pendingTasks.delete(task);
+					count++;
+				}
+				// kill is done here to ensure the process is actually killed right now
+				// fsr the disconnect() done in destroy doesn't stop the running task
```
Zarel, line 345: That's intentional; we probably want a separate kill command.

server/chat-plugins/net-filters.ts
```diff
@@ -309,14 +320,35 @@ export const commands: ChatCommands = {
...
+				subProcess.process.kill();
+				subProcess.destroy();
```
Zarel, line 347: Does this actually clean the process up from `releasingProcesses` and `processes`?

server/chat-plugins/net-filters.ts
mia-pi-git: Yeah, that's intentional - i know to watch for the processes to be careful, and i know most releasing processes are the train ones usually. No other process should be releasing except during hotpatch, and those would release much quicker.

server/chat-plugins/net-filters.ts
mia-pi-git: I believe so? I recall seeing it locally

mia-pi-git: https://prnt.sc/yc7jpw

server/chat-plugins/net-filters.ts
Zarel: It seems like this is bad just as a correctness thing. It seems like training should be a different PM if you want to do stuff like this.

server/chat-plugins/net-filters.ts
Zarel: Hm, this is probably a bit unintuitive, that `.destroy` will clean itself up from `.releasingProcesses` but not from `.processes`.

Zarel: I'm not sure if there's anything we can do about that, though.

Zarel: Actually, it's probable that `.process.disconnect` would handle removing it from `.processes`. So maybe this is fine overall! I should better comment `destroy` though.

server/chat-plugins/net-filters.ts
```diff
@@ -192,35 +221,46 @@ export const PM = new QueryProcessManager<NetQuery, any>(module, async query =>
...
+		});
+		// we only want to spawn one network, when it's the subprocess
+		// otherwise, we use the PM for interfacing with the network
+		net = new NeuralNetChecker(PATH);
+		// eslint-disable-next-line no-eval
+		Repl.start(`netfilters-${k}-${process.pid}`, cmd => eval(cmd));
+	} else {
+		PM.spawn(NUM_PROCESSES[k as keyof typeof processes]);
+	}
 }
```
Zarel, line 264:
```suggestion
const PM = processes.main;
const PMTraining = process.training;
if (!PM.isParentProcess) {
	global.Config = Config;

	global.Monitor = {
		crashlog(error: Error, source = `A netfilter ${k} process`, details: AnyObject | null = null) {
			const repr = JSON.stringify([error.name, error.message, source, details]);
			process.send!(`THROW\n@!!@${repr}\n${error.stack}`);
		},
	};
	process.on('uncaughtException', err => {
		if (Config.crashguard) {
			Monitor.crashlog(err, 'A net filter child process');
		}
	});
	// we only want to spawn one network, when it's the subprocess
	// otherwise, we use the PM for interfacing with the network
	net = new NeuralNetChecker(PATH);
	// eslint-disable-next-line no-eval
	Repl.start(`netfilters-${k}-${process.pid}`, cmd => eval(cmd));
} else {
	PM.spawn(NUM_PROCESSES.main);
	PMTraining.spawn(NUM_PROCESSES.training);
}
```

This fixes your double-response problem.

Zarel: In hindsight, this doesn't fully fix it, but it's a start.

server/chat-plugins/net-filters.ts
```diff
@@ -192,19 +221,29 @@ export const PM = new QueryProcessManager<NetQuery, any>(module, async query =>
...
+
+export const processes = {
+	// this one runs longer because training is SLOW
+	training: new ProcessManager.QueryProcessManager(module, handleQuery, PM_TIMEOUT),
+	// this one is fine
+	main: new ProcessManager.QueryProcessManager(module, handleQuery),
+}
 
+const PM = processes.main;
+const PMTraining = process.training;
```
Zarel, line 241:
```suggestion
const PM = new ProcessManager.QueryProcessManager(module, handleQuery),
// this one runs longer because training is SLOW
const PMTraining = new ProcessManager.QueryProcessManager(module, handleQuery, PM_TIMEOUT),
```

server/chat-plugins/net-filters.ts
```diff
@@ -309,14 +346,40 @@ export const commands: ChatCommands = {
...
+					count++;
+				}
+				// kill is done here to ensure the process is actually killed right now
+				// fsr the disconnect() done in destroy doesn't stop the running task
+				subProcess.destroy();
+				subProcess.process.removeAllListeners('disconnect'); // if we don't do this it thinks it's a crash
+				subProcess.process.kill();
+				PM.processes.splice(PM.processes.indexOf(subProcess), 1);
+			}
+			PM.spawn(NUM_PROCESSES['training']);
```
Zarel, line 373: These should just use `PMTraining` directly. Also why are you manually destroying the processes?

Zarel: It feels like a lot of this code goes inside `lib/process-manager.ts`

server/chat-plugins/net-filters.ts
```diff
@@ -6,14 +6,15 @@
...
 import {Config} from '../config-loader';
-import {Repl} from '../../lib/repl';
 import {linkRegex} from '../chat-formatter';
 
 const PATH = "config/chat-plugins/net.json";
-const NUM_PROCESSES = Config.netfilterprocesses || 1;
+const NUM_PROCESSES = {
+	training: 1,
+	main: 1,
+}
```
Zarel, line 17:
```suggestion
};
```

lib/process-manager.ts
```diff
@@ -557,6 +557,12 @@ export class QueryProcessManager<T = string, U = string> extends ProcessManager<
 			process.exit();
 		});
 	}
+	destroyProcess(process: QueryProcessWrapper<T, U>) {
+		process.destroy();
+		process.process.removeAllListeners('disconnect');
+		process.process.kill();
+		this.processes.splice(this.processes.indexOf(process), 1);
```
Zarel, line 564: Should probably make sure that `this.processes.indexOf(process)` doesn't return `-1`.

lib/process-manager.ts
```diff
@@ -553,9 +553,16 @@ export class QueryProcessManager<T = string, U = string> extends ProcessManager<
...
-			process.exit();
-		});
+		process.on('disconnect', () => process.exit());
+	}
+	destroyProcess(process: QueryProcessWrapper<T, U>) {
+		process.destroy();
+		process.process.removeAllListeners('disconnect');
+		process.process.kill();
+		const idx = this.processes.indexOf(process);
+		if (idx > 0) {
```
Zarel, line 563:
```suggestion
		if (idx >= 0) {
```

😬

lib/process-manager.ts
```diff
@@ -553,9 +553,16 @@ export class QueryProcessManager<T = string, U = string> extends ProcessManager<
 				response => process.send!(`${taskId}\n${JSON.stringify(response)}`)
 			);
 		});
-		process.on('disconnect', () => {
-			process.exit();
-		});
+		process.on('disconnect', () => process.exit());
```
Zarel, line 556: ಠ_ಠ

mia-pi-git: b-but it was already there ;_;

Zarel: No, I mean, you changed it for no reason.

The old version doesn't depend on the return value of `process.exit()`, the new version returns it.

mia-pi-git: Ah.

mia-pi-git:
lol. not needed anymore

PR #7994, opened 2021-02-04
SSB4: Bugfixes

xInstruct:
aaaaaaaaaaa

Update description to match this, and fixes a different thing


Volco:
The fact that my mon has a bug like this is funny as fuck

PR #8000, opened 2021-02-06
Fix helptickets game.ticket setting to undefined

Volco:


HoeenCoder:
Could `game.ticket` in theory be undefined after a hotpatch but before this change is applied? Since were having issues with `game.ticket` being undefined.

Volco:
I don't think that's possible due to the only spot ``game.ticket`` is set is line 550 unless the ``game.ticket`` is getting changed via eval which would be a different issue

Volco:
what do you mean, like an existing ``game.ticket`` already being undefined then this change being added?

HoeenCoder:
I mean `game.ticket` being undefined through some other measure (the hotpatch itself?) and then this statement running (resulting in no change since `game.ticket` is already undefined).

Volco:
seems unlikely given the only place ``game.ticket`` is ever changed aside from when it's first applied is line 550

Volco:
Bump

Also @HoeenCoder when you're free can you look at this?
https://github.com/smogon/pokemon-showdown/pull/7992

PR #8002, opened 2021-02-06
Fix minor activity timer not queuing polls

Volco:
the ``ts-ignores`` are because when i added ``class: Poll`` it says some properties are missing and the other one is "type MinorActivity has no construct signatures". If you recommend handling the errors some way else lmk.

personally I wish it was possible to import poll.ts to room-minor-activity without it crashing on boot but if it was then the ``throw new Error`` line would have never been necessary

### DIFF
--- a/server/room-minor-activity.ts
+++ b/server/room-minor-activity.ts
@@ -32,6 +32,7 @@ export interface MinorActivityData {
 	totalVotes?: number;
 	isQuiz?: boolean;
 	answers: string[] | {name: string, votes: number, correct?: boolean}[];
+	class?: MinorActivity;
 }
 
 // globally Rooms.MinorActivity
@@ -95,7 +96,10 @@ export abstract class MinorActivity {
 			});
 
 			if (!MinorActivityClass) {
-				if (pollData.activityid === 'poll') throw new Error('No minorActivity class provided');
+				// hacky fix to make poll queues work with timers
+				const Poll = pollData.class;
+				// @ts-ignore
+				if (pollData.activityid === 'poll') room.setMinorActivity(new Poll(room, pollData));
 			} else {
 				room.setMinorActivity(new MinorActivityClass(room, pollData));
 			}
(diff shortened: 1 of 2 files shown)


server/room-minor-activity.ts
```diff
@@ -95,7 +96,10 @@ export abstract class MinorActivity {
 			});
 
 			if (!MinorActivityClass) {
-				if (pollData.activityid === 'poll') throw new Error('No minorActivity class provided');
+				// hacky fix to make poll queues work with timers
+				const Poll = pollData.class;
+				// @ts-ignore
+				if (pollData.activityid === 'poll') room.setMinorActivity(new Poll(room, pollData));
```
AnnikaCodes, line 102: The reason you can't import `Poll` here is that `poll.ts` expects the `Rooms` global to be set up. A less hacky way to solve this is to utilize `require`.
```suggestion
				if (pollData.activityid === 'poll') {
					const {Poll} = require('./chat-plugins/poll');
					room.setMinorActivity(new Poll(room, pollData));
				}
```

AnnikaCodes: As far as I know, `Rooms` will always exist by the time we're ending a MinorActivity, so this seems safe to me.

Volco: let me test it real fast just to make sure. for some reason i though require weren't allowed in TS files. probably my memory just being bad per usual

AnnikaCodes (approved):
Looks good to me.

server/room-minor-activity.ts
```diff
@@ -95,7 +95,10 @@ export abstract class MinorActivity {
 			});
 
 			if (!MinorActivityClass) {
-				if (pollData.activityid === 'poll') throw new Error('No minorActivity class provided');
```
Zarel, line 98: I think I want to know more about why `MinorActivityClass` doesn't exist here. It definitely looks like we need some sort of global registry of minor activity classes (probably also games) to make this easier to use.

Zarel:
Not a fix for the core problem, but it sure fixes the symptoms, so thanks!

PR #8003, opened 2021-02-07
Convert modifiers from hex to decimal

DaWoblefet:
Working with hex in modifiers is cumbersome. It's pretty outdated, and I assume only used for consistency with the [old BW damage formula article](https://www.smogon.com/bw/articles/bw_complete_damage_formula). However, requiring the extra step of converting to hex leads to issues where simple mistakes can be made - for example, the Aura multiplier was wrong (listed as 0x1547, which is 5447 in decimal, but it should be 5448). Using decimal allows for better maintainability.

This PR doesn't guarantee all the modifiers are accurate, but the conversions from the old hex values to the decimal values should be accurate.

### DIFF
--- a/data/mods/ssb/conditions.ts
+++ b/data/mods/ssb/conditions.ts
@@ -2143,7 +2143,7 @@ export const Conditions: {[k: string]: ModdedConditionData & {innateName?: strin
 		name: "Big Storm Coming Mod",
 		duration: 1,
 		onBasePower() {
-			return this.chainModify([1228, 4096]);
+			return this.chainModify([1229, 4096]);
 		},
 	},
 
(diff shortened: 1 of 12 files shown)


data/mods/ssb/conditions.ts
```diff
@@ -2143,7 +2143,7 @@ export const Conditions: {[k: string]: ModdedConditionData & {innateName?: strin
 		name: "Big Storm Coming Mod",
 		duration: 1,
 		onBasePower() {
-			return this.chainModify([0x4CC, 0x1000]);
+			return this.chainModify([1228, 4096]);
```
xInstruct, line 2146: You PMed me this earlier, and yeah I think it should be 1229 over 1228. `0.300...` is better than `0.299...`
```suggestion
			return this.chainModify([1229, 4096]);
```
for your convenience

Marty-D (approved):
The Dark/Fairy Aura modifier being wrong is because in 2013 or whenever, it was actually thought to be 5447/4096. If everyone is reporting things in terms of 4096 these days then sure, makes sense. 👍

PR #8004, opened 2021-02-07
Modlog: Use SQLite for searches

AnnikaCodes:
This is very much a work in progress - the indexes probably need some fine-tuning (and we may need to use the `UNION` of `SELECT`s optimization rather than relying on SQLite to properly index `OR`s). Also, this runs SQL `SELECT` queries in the main process, which is unwise; I believe Mia is working on a library to run SQL in a separate process (or thread?). If not, then I'll need to make this compatible with a QueryProcessManager.

I'm making a PR because I had the motivation to work on this over the weekend and this way it'll be out in public; this is still pretty tentative. I will try to work on this when I can but I'm not in a great place mentally and school-wise IRL so no promises.

I elected to leave the text-searching functionality in temporarily; I do not think that PS should maintain flat-file modlogs in the long term, but if there are any bugs in the initial deployment, being able to turn off SQL modlog searching simply by setting `Config.usesqlitemodlog` to `false` could be very helpful. It can and should be removed in a later commit.

If it moves forward and gets to a state where it's ready to be deployed:
 - The indexes will need to be manually added to the database
 - The modlog conversion script will need to be run on the preexisting text modlogs prior to the start of dual logging (November 1, 2020). This will necessitate some way of stopping new modlogs from being written as they would be missed; for transparency and acountability, I'd recommend doing this during downtime.


AnnikaCodes:
@monsanto @Zarel bump? This isn't urgent at all (modlog works fine right now), but I'd like to make sure you're aware of this PR.

AnnikaCodes:
I just tried to work on this and the conversion script isn't picking up on all the rooms...
```
sqlite> SELECT DISTINCT roomid FROM modlog WHERE roomid NOT LIKE '%-%';
1v1
1v1oldgens
1v1typechallenge
2am
2v2
42
absoluteabsolposters
academics
adminlog
aetheroom
agpl
agsnakedraft
agsnakedraftmock
ama
amber
amberleague
amphyrp
animalcrossing
animeandmanga
animeandmangalive
animemanga
animetheory
anythinggoes
arabic
arcade
arcticlugias
aroomoficeandfire
art
artscrafts
artsculture
ataraxia
aura
b101staff
basshouse
battle
lobby
```

Not sure what's wrong here but it's a bug I need to fix.

AnnikaCodes:
I fixed the above issue; the only issue with the converter now is that there are some discrepancies (for example, Lobby having more modlog entries in SQLite than lines in the `modlog_lobby.txt` file); I don't know if these are bugs in my code or glitches with the old system of modlog. Still, outside perspectives are appreciated.

I still think SQL modlog searches can be made faster (it's pretty slow rn), but I'm out of time today for programming. Next time I work on this, I will look into improving indices and/or queries.

AnnikaCodes:
Alright, I think I figured out a way to deploy this without overall downtime. 
- set `Config.modlogsqlitetimeout` to 10000
- hotpatch the modlog, which will migrate the database (adding the new indices and the `is_global`) and activate SQL searching
  - modlogs from before November 2020 will be temporarily inaccessible
- run the conversion script with `--newest-allowed-timestamp 1604268462363`
  - this will convert the old modlogs; SQLite writes are fast enough (and rare enough; main's modlog averages less than four writes per minute) that with a slightly increased timeout on waiting for the writer lock, the conversion can happen without downtime.
  - there is no concern about newly written modlogs being lost because they are already dual logging on SQLite
- if anything goes wrong I can revert this commit and since text logging was never removed, modlogs will still work fine without using SQLite at all

mia-pi-git (commented):
Seems mostly fine. My only comment would be that splitting modlog into a new directory seems unnecessary complex - why not just export the transactions from server/modlog.ts?

AnnikaCodes:
> Seems mostly fine. My only comment would be that splitting modlog into a new directory seems unnecessary complex - why not just export the transactions from server/modlog.ts?

(We discussed this on Discord.)

AnnikaCodes:
OK, let's do this.

PR #8007, opened 2021-02-08
Gen 1: Stadium Haze should cure user's status

PlagueVonKarma:
So, to be clear: this doesn't fix the whole host of bugs this move has. There's a lot of them.

This fix intends to give Haze its "refresh" effect that it has exclusively in Stadium. Not only does it cure the opponent's status, but it also cures the user's as well. This is an important change for Vaporeon in particular, as it makes a decent status absorber. There _may_ be a better way to implement this, I'm not sure, but I'm guessing this'll receive edits during review. My code seems a touch inefficient?

Here is my implementation working on the RBY Server and my subsequent freakout: 
https://replay.pokemonshowdown.com/rby-gen1stadiumou-1735

This addresses part of my vitriolic criticisms of the Haze implementation here:
https://www.smogon.com/forums/threads/stadium-format-is-now-available-on-ps.3526616/post-8503308

For a video of what the implementation should look like, go here: 
https://www.youtube.com/watch?v=LoTM-tsop68
This will also verify what you see on my replay and such. 

Credit to Beelzemon 2003 for assistance with the initial testing of Stadium Haze.

### DIFF
--- a/data/mods/stadium/moves.ts
+++ b/data/mods/stadium/moves.ts
@@ -47,7 +47,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			for (const pokemon of this.getAllActive()) {
 				pokemon.clearBoosts();
 				if (pokemon.status !== 'tox') {
-					//This should cure the status of both Pokemon, and subsequently recalculate stats to remove the Paralysis/Burn Speed Drop.
+					// This should cure the status of both Pokemon, and subsequently recalculate stats to remove the Paralysis/Burn Speed Drop.
 					pokemon.cureStatus();
 					pokemon.recalculateStats();
 				}
--- a/data/mods/stadium/moves.ts
+++ b/data/mods/stadium/moves.ts
@@ -41,7 +41,6 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		// FIXME: onBeforeMove() {},
 	},
 	haze: {
-		haze: {
 		inherit: true,
 		onHit(target, source) {
 			this.add('-clearallboost');
--- a/data/mods/stadium/moves.ts
+++ b/data/mods/stadium/moves.ts
@@ -48,6 +48,8 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 				pokemon.clearBoosts();
 				// This should cure the status of both Pokemon, and subsequently recalculate stats to remove the Paralysis/Burn Speed Drop.
 				pokemon.cureStatus();
+				pokemon.removeVolatile(id);
+				this.add('-end', pokemon, id);
 				target.recalculateStats!();
 			}
 		},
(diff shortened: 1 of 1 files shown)


PlagueVonKarma:
Finally!! Sorry for the dumb mistakes, it worked fine on my server.

data/mods/stadium/moves.ts
```diff
@@ -40,6 +40,32 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+					// This should cure the status of both Pokemon, and subsequently recalculate stats to remove the Paralysis/Burn Speed Drop.
+					pokemon.cureStatus();
+					target.recalculateStats!();
+				}
+				if (pokemon.status === 'tox') {
+					pokemon.setStatus('psn');
+				}
+				for (const id of Object.keys(pokemon.volatiles)) {
+					if (id === 'residualdmg') {
+						pokemon.volatiles[id].counter = 0;
```
MacChaeger, line 59: The video you linked seems to show the Tox status being fully cured by Haze, not turned into a weaker poison. Am I missing something?

PlagueVonKarma: Ahhhh shit, you're right. I never noticed that when testing, I always assumed it was the same as RBY. That can be cut then.

MacChaeger: It’s weird how it seems to not be cured immediately, though.

PlagueVonKarma: Yeah, I don't get that myself. Magnemite doesn't take the residual damage though, so the status is definitely removed in its entirety. Maybe the game can only update one of the graphics at a time or something? I dunno, I'm kinda spitballing here.

Marty-D (approved):
👍

data/mods/stadium/moves.ts
```diff
@@ -40,6 +40,22 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+			this.add('-clearallboost');
+			for (const pokemon of this.getAllActive()) {
+				pokemon.clearBoosts();
+				// This should cure the status of both Pokemon, and subsequently recalculate stats to remove the Paralysis/Burn Speed Drop.
+				pokemon.cureStatus();
+				for (const id of Object.keys(pokemon.volatiles)) {
+					pokemon.removeVolatile(id);
+					this.add('-end', pokemon, id);
+				}
+				target.recalculateStats!();
```
urkerab, line 55: Should this be `pokemon` rather than `target`?

KrisXV: you have a knack for waiting until its just too late to comment on things like this

PlagueVonKarma: It should be both.

KrisXV: yeah its in the for loop so it's `pokemon` is what urkerab was saying; I'll push a fix

urkerab: Sorry, I would have commented the first time I looked at this PR but I didn't see it then...

KrisXV: > Sorry, I would have commented the first time I looked at this PR but I didn't see it then...

@urkerab I should clarify I was just joking, those kinds of catches are obviously always appreciated; my joke just didn't land like I had anticipated it would. Sorry!

PR #8008, opened 2021-02-08
Suspect Test and Update

DarkPhoenix911:
Adding Nat Dex Mono Suspect Test (Dracovish) and updating thread links.


KrisXV:
This is not how we add suspect tests nowadays, and the discussion threads were already up to date

PR #8016, opened 2021-02-09
Uno: Fix crash

Distrib-ps:


server/chat-plugins/uno.ts
```diff
@@ -246,8 +246,11 @@ export class UNO extends Rooms.RoomGame {
 				this.sendToRoom(`|raw|${Utils.escapeHTML(name)} has not picked a color, the color will stay as <span style="color: ${textColors[this.topCard.changedColor]}">${this.topCard.changedColor}</span>.`);
 			}
 		}
-		if (this.awaitUno === userid) this.awaitUno = null;
-
+		if (this.awaitUno === userid) this.awaitUno = null; {
```
mia-pi-git, line 249: uh, what? Does this work locally?

Distrib-ps: For me yes:
Now: http://prntscr.com/yvvqsd
Before: http://prntscr.com/yvvu2v

mia-pi-git: Okay then.
```suggestion
		if (this.awaitUno === userid) this.awaitUno = null;
```

server/chat-plugins/uno.ts
```diff
@@ -246,8 +246,11 @@ export class UNO extends Rooms.RoomGame {
...
+			if (!this.topCard) {
+				throw new Chat.ErrorMessage(`Unable to disqualify ${name}.`);
+			}
+		}
```
mia-pi-git, line 253:
```suggestion
		if (!this.topCard) {
			throw new Chat.ErrorMessage(`Unable to disqualify ${name}.`);
		}
```

Distrib-ps:
It's good like that? @mia-pi-git

Zarel:
Who even worked on Uno recently? This seems like a pretty esoteric error message, too – why are we sometimes unable to disqualify players?

Distrib-ps:
I’ve been working on it recently (including a pending pr https://github.com/smogon/pokemon-showdown/pull/8001).
And yes, maybe an explanation would be better. Maybe "The uno hasn't started yet." ?

PR #8020, opened 2021-02-12
Suspect Tests: Add a whitelist

KrisXV:
Annika said this was cool but that I should make a pull request in case Zarel or someone has a different opinion

### DIFF
--- a/server/chat-plugins/suspect-tests.ts
+++ b/server/chat-plugins/suspect-tests.ts
@@ -1,6 +1,7 @@
 import {FS} from '../../lib/fs';
 
 const SUSPECTS_FILE = 'config/suspects.json';
+const WHITELIST = ["kris"];
 
 interface SuspectTest {
 	tier: string;
@@ -13,6 +14,12 @@ export const suspectTests: {[format: string]: SuspectTest} = JSON.parse(FS(SUSPE
 
 function saveSuspectTests() {
 	FS(SUSPECTS_FILE).writeUpdate(() => JSON.stringify(suspectTests));
+}
+
+function checkPermissions(context: CommandContext) {
+	const user = context.user;
+	if (WHITELIST.includes(user.id)) return true;
+	context.checkCan('gdeclare');
 }
 
 export const commands: ChatCommands = {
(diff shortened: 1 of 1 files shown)


KrisXV:
um oops

server/chat-plugins/suspect-tests.ts
```diff
@@ -1,6 +1,7 @@
 import {FS} from '../../lib/fs';
 
 const SUSPECTS_FILE = 'config/suspects.json';
+const WHITELIST = ["kris"];
```
HoeenCoder, line 4: Do we want this to be a configuration setting rather than a hardcoded entry?

HoeenCoder (commented):
Looks good to me overall.

server/chat-plugins/suspect-tests.ts
AnnikaCodes: I would prefer that, personally; that way, it's specific to each server.

KrisXV: It would be nice but Zarel prefers this method

KrisXV:
bump @Zarel

server/chat-plugins/suspect-tests.ts
Zarel: It sucks for other servers but makes it easier for people to PR changes to the list. I'm neutral, as long as I'm not the one maintaining the configuration setting – it all depends on how frequently people make changes to the list, and how other servers do things like this.

In the past, we've also based it on "roomauth in Dev". Would that work here? Whitelist all dev roommods?

AnnikaCodes: Dev roommod is a pretty large group (annika, austin, bmelts, chaos, dawoblefet, hoeenhero, instruct, jetou, joim, kaphotics, kris, marty, mia, panpawn, peach, pyuk, sirdonovan, slayer95, theimmortal, urkerab, v4, viett, xfix). Given that we originally only permitted upperstaff to add suspects, this seems like a big jump. I think Dev roommods are generally either inactive or trustworthy, but I question why we would expose suspect management to such a large group?

In my mind, Dev roommod has been used in conjunction with other things (such as global staff or room staff), not as a primary permission check, but just to verify that the user is technically competent and won't break things. (This was proposed for Anubis to manage hosts and Mia/Kris to add regexes to the Help autoresponder.)

I can maintain the config setting if we decide to go that route.

Zarel:
I don't mind the whitelist. I just want to be clear that I have no problem with a Config setting if I'm not maintaining it.

Thanks!

PR #8022, opened 2021-02-12
Give usersearch a page

KrisXV:
approved by @HoeenCoder 

This was originally requested by OM as a way to easily click and forcerename problematic names, and I moved it into its own chat plugin in the process

### DIFF
--- a/server/chat-plugins/usersearch.ts
+++ b/server/chat-plugins/usersearch.ts
@@ -1,0 +1,96 @@
+import {Utils} from '../../lib';
+
+const ONLINE_SYMBOL = ` \u25C9 `;
+const OFFLINE_SYMBOL = ` \u25CC `;
+
+function searchUsernames(target: string, page = false) {
+	const results: {offline: string[], online: string[]} = {
+		offline: [],
+		online: [],
+	};
+	for (const curUser of Users.users.values()) {
+		if (!curUser.id.includes(target) || curUser.id.startsWith('guest')) continue;
+		const escapedName = Utils.escapeHTML(curUser.name);
+		const buttonHTML = `<button class="button" name="send" value="/forcerename ${escapedName}&#10;/j view-usersearch-${target}">Forcerename</button>`;
+		if (curUser.connected) {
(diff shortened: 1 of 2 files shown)


AnnikaCodes (approved):
This looks fine to me overall.

server/chat-plugins/usersearch.ts
```diff
@@ -0,0 +1,96 @@
+import {Utils} from '../../lib';
+
+const ONLINE_SYMBOL = ` \u25C9 `;
+const OFFLINE_SYMBOL = ` \u25CC `;
```
AnnikaCodes, line 4: These are used in several places; I wonder if they should be attached to the `Users` or `Chat` globals (maybe in a separate PR).

server/chat-plugins/usersearch.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function searchUsernames(target: string, page = false) {
+	const results: {offline: string[], online: string[]} = {
+		offline: [],
+		online: [],
+	};
+	for (const curUser of Users.users.values()) {
+		if (!curUser.id.includes(target) || curUser.id.startsWith('guest')) continue;
+		const escapedName = Utils.escapeHTML(curUser.name);
+		const buttonHTML = `<button class="button" name="send" value="/forcerename ${escapedName}&#10;/j view-usersearch-${target}">Forcerename</button>`;
```
mia-pi-git, line 14: Please don't add forcerename buttons to the command, that would be _so_ easy to click on accidentally.

KrisXV: That would defeat the point of this

AnnikaCodes: Maybe add them to the page, but not the command?

This seems like a good thing to discuss in <<staff>>.

KrisXV: They are only on the page?

KrisXV: spoke w/ mia, she just misread the code

AnnikaCodes:
Thanks!

PR #8023, opened 2021-02-12
Chatlog: Improve logsearch speed by streaming results

mia-pi-git:
As it is, we just hold a giant string in a buffer, so breaking that stream up and transforming it bit-by-bit should be quicker.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -5,11 +5,14 @@
  * @license MIT
  */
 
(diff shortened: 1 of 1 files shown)


Zarel:
I would be very surprised if this makes a performance difference. Have you tested this?

Zarel:
Transforming it and holding it in a buffer should be relatively instantaneous; the limiting factor is searching which is done by ripgrep and this doesn't improve on.

PR #8027, opened 2021-02-13
Hangman: Improve hangman terms

Distrib-ps:


server/chat-plugins/hangman.ts
```diff
@@ -428,24 +429,34 @@ export const commands: ChatCommands = {
...
 			buf += `<div class="infobox">`;
-			buf += `<strong>${t}</strong>:<br />`;
-			buf += roomTerms[t].map(
-				hint => `<button class="button" name="send" value="/hangman rr ${t},${hint},room:${room.roomid}">${hint}</button>`
-			);
+			buf += `<h3>${t}</h3><hr />`;
+			if (user.can('mute', null, room, 'hangman addrandom')) {
+				buf += roomTerms[t].map(
+					hint => `<button class="button" name="send" value="/hangman rr ${t},${hint},room:${room.roomid}">${hint}</button>`
+				);
```
Zarel, line 453: It seems kind of bad for clicking on the hint to delete it. Especially if there's another delete button that deletes all the hints.

Zarel: I guess this PR doesn't make that problem _worse_, though.

Distrib-ps: > It seems kind of bad for clicking on the hint to delete it. Especially if there's another delete button that deletes all the hints.

Should this option be removed?

Zarel: Plausibly. I'll look into it.

PR #3280, opened 2017-02-23
Update Inheritance Hack

Spandamn:
This is a worse hack, but it will prevent locked users from cheating by getting their nicknames suppressed.
I also wanna know: What would be the problems that this could possibly face? I've tested this and it works fine (But I dont yet know about any problems)

PR #8030, opened 2021-02-14
Change protocol for passing donor details in Inheritance

Spandamn:
The ability hack that i came up with years ago is no longer required, since Gen 8does not have Hidden Powers.
@KrisXV

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -688,9 +688,9 @@ export const Formats: FormatList = [
 				return [`${name} has an illegal set with an ability from ${this.dex.getSpecies(pokemonWithAbility[0]).name}.`];
 			}
 
-			// Protocol: Include the data of the donor species in the `ability` data slot.
-			// Afterwards, we are going to reset the name to what the user intended.
-			set.ability = `${set.ability}0${canonicalSource}`;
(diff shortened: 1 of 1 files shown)


Spandamn:
> Well, not impossible, but it would break things if someone tried to use Hidden Power.

If someone tries to use Hidden Power, wont it just say wrong move at the validator

Slayer95:
> ability hack that i came up with years ago

Fairly sure that was my own idea. But ok?

Spandamn:
> > ability hack that i came up with years ago
> 
> Fairly sure that was my own idea. But ok?

mightve been, its been a while since gen 7

Spandamn:
#3280 this was the one and holy shit its been 4 years

Spandamn:
Tried it out, validator says Invalid set of you select Hidden Power as a move

KrisXV:
I don't want to merge this; it breaks compatibility with past-gens inheritance and national dex inheritance

Zarel:
I think Kris makes a good point. You probably want to continue using this hack for now.

Zarel:
Closing because there's been no further discussion since Kris mentioned the NatDex/past-gens issue.

PR #8031, opened 2021-02-14
Focus Punch can't be Flinched

urkerab:
As researched by @DaWoblefet.

Although you lose focus, you are never flinched, and can still move normally by way of Dancer, but on the other hand you lose out on the Steadfast boost.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -5484,6 +5484,9 @@ export const Moves: {[moveid: string]: MoveData} = {
 					pokemon.volatiles['focuspunch'].lostFocus = true;
 				}
 			},
+			onTryAddVolatile(status, pokemon) {
+				if (status.id === 'flinch') return null;
+			},
 		},
 		secondary: null,
 		target: "normal",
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Regarding the Steadfast boost: you also don't get one if anything in the [checks done before move success](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8548957) happens prior to the flinch check. In Gen 8, this includes remaining asleep, remaining frozen, or having no PP remaining for the move, as well as Focus Punch losing focus.

Just so it's clear for anyone else reading, the recent discovery with Focus Punch is that the "lost focus" and "will flinch" flags are dichotomous. Normally, the Focus Punch failure just happens to be before the flinch failure. But when you tack on using another move during the turn, such as by Dancer, the Focus Punch "lost focus" flag is set and it never gets around to setting the flinch flag. The lost focus message will only play if the Pokemon is actually attempting to execute Focus Punch, so what ends up happening is the Dancer attack goes on through just fine even though in the ordinary case, the Pokemon would flinch from Dancer. See these short videos by way of comparison:
https://www.youtube.com/watch?v=BJ3eQLSVrA8
https://www.youtube.com/watch?v=JkNM1SbZKL4

Zarel:
Crazy!

PR #8032, opened 2021-02-14
SSB: Change set generation to rely on weaknesses against types

xInstruct:
Approved by myself.

Although this will deviate the ratio of deviation between the most generated set vs the least generated set, i think this will be a welcome change that helps fix how ssb teams fair against each other.

### DIFF
--- a/data/mods/ssb/scripts.ts
+++ b/data/mods/ssb/scripts.ts
@@ -403,7 +403,7 @@ export const Scripts: ModdedBattleScriptsData = {
 		}
 
 		if (move.selfSwitch && this.getAllActive().some(x => x.hasAbility('skilldrain'))) {
-			this.hint(`Self-switching don't trigger when a Pokemon with Skill Drain is active.`);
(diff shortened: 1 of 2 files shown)


xInstruct:
tested, works as intended

data/mods/ssb/scripts.ts
```diff
@@ -402,8 +402,8 @@ export const Scripts: ModdedBattleScriptsData = {
 			}
 		}
 
-		if ((move.forceSwitch || move.selfSwitch) && this.getAllActive().some(x => x.hasAbility('skilldrain'))) {
-			this.hint(`Self-switching and force switch moves don't trigger when a Pokemon with Skill Drain is active.`);
+		if (move.selfSwitch && this.getAllActive().some(x => x.hasAbility('skilldrain'))) {
+			this.hint(`Self-switching don't trigger when a Pokemon with Skill Drain is active.`);
```
AnnikaCodes, line 406:
```suggestion
			this.hint(`Self-switching doesn't trigger when a Pokemon with Skill Drain is active.`);
```

PR #8033, opened 2021-02-14
Moderation: Add  `spoilers` support to more commands

mia-pi-git:
Requested feature, plus this makes it easier to do the `spoilers` thing on other commands.


Zarel:
(build error)

mia-pi-git:
> (build error)

(fixed)

Zarel:
It's really too bad that we're using `spoiler:` for this, instead of something that makes more sense like "private:" or "staff:" or something.

PR #8035, opened 2021-02-15
Usersearch: Make page forcerenames log to staff

KrisXV:
(I also fixed the title @mia-pi-git yw)

approved/requested by aeonic

While this isn't the most optimal solution, it's probably the best at the moment because it a) solves the issue of all /forcerenames for users in Lobby being parsed from the usersearch pages log to Lobby (this isn't good because it clutters the Lobby's modlog and causes their roomstaff to see things they shouldn't have to) and all /forcerenames from staff not in Lobby being PMed to ~, which is annoying for different reasons such as unnecessarily opening a new pm, and b) allows people to run `/frinstaff` to tackle obscene names outside of Usersearch pages before roomstaff have to see those names while shaving off a few additional seconds that it would normally take to copy the name and switch to the Staff room.

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1537,7 +1537,7 @@ export const commands: ChatCommands = {
 	],
 
 	fr: 'forcerename',
-	forcerename(target, room, user, connection, cmd) {
+	forcerename(target, room, user) {
 		if (!target) return this.parse('/help forcerename');
 
 		const reason = this.splitTarget(target, true);
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1537,9 +1537,11 @@ export const commands: ChatCommands = {
 	],
 
 	fr: 'forcerename',
-	forcerename(target, room, user) {
+	forcerenameinstaff: 'forcerename',
+	forcerename(target, room, user, connection, cmd) {
 		if (!target) return this.parse('/help forcerename');
 
+		if (cmd.endsWith('staff') && Rooms.search('staff')) room = this.room = Rooms.search('staff')!;
 		const reason = this.splitTarget(target, true);
 		const targetUser = this.targetUser;
 		const targetID = toID(this.targetUsername);
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1537,6 +1537,7 @@ export const commands: ChatCommands = {
 	],
 
 	fr: 'forcerename',
+	frinstaff: 'forcerename',
 	forcerenameinstaff: 'forcerename',
 	forcerename(target, room, user, connection, cmd) {
 		if (!target) return this.parse('/help forcerename');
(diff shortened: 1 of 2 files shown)


server/chat-commands/moderation.ts
```diff
@@ -1537,9 +1537,12 @@ export const commands: ChatCommands = {
 	],
 
 	fr: 'forcerename',
-	forcerename(target, room, user) {
+	frinstaff: 'forcerename',
+	forcerenameinstaff: 'forcerename',
+	forcerename(target, room, user, connection, cmd) {
 		if (!target) return this.parse('/help forcerename');
 
+		if (cmd.endsWith('staff') && Rooms.search('staff')) room = this.room = Rooms.search('staff')!;
```
Zarel, line 1545: I wonder if we should have a `/msgroom staff, /fr username` so we don't have to rewrite this sort of thing in every room button.

server/chat-commands/moderation.ts
```diff
@@ -1537,9 +1537,12 @@ export const commands: ChatCommands = {
 	],
 
 	fr: 'forcerename',
-	forcerename(target, room, user) {
+	frinstaff: 'forcerename',
```
Zarel, line 1540: seems kinda long for a shortcut. `/fris`? `/frs`?

Zarel:
> all /forcerenames for users in Lobby being parsed from the usersearch pages log to Lobby (this isn't good because it clutters the Lobby's modlog and causes their roomstaff to see things they shouldn't have to)

I'm confused by this. Either lobby staff want to see forcerenames or they don't. If they don't, we could just log all forcerenames to staff, and stop logging them in individual rooms entirely...

KrisXV:
I believe you're overgeneralizing what I'm saying. "Why would a lobby staff want to see a forcerename of a user who's either not in lobby at all or currently not active in lobby?" was the point I was making.

Zarel:
Again, I'm mostly talking about the `/frinstaff` alias. Why do staff need to decide where they want to log something every time they fr someone?

server/chat-plugins/usersearch.ts
```diff
@@ -11,7 +11,7 @@ function searchUsernames(target: string, page = false) {
 	for (const curUser of Users.users.values()) {
 		if (!curUser.id.includes(target) || curUser.id.startsWith('guest')) continue;
 		const escapedName = Utils.escapeHTML(curUser.name);
-		const buttonHTML = `<button class="button" name="send" value="/forcerename ${escapedName}&#10;/j view-usersearch-${target}">Forcerename</button>`;
+		const buttonHTML = `<button class="button" name="send" value="/forcerenameinstaff ${escapedName}&#10;/j view-usersearch-${target}">Forcerename</button>`;
```
Zarel, line 14:
```suggestion
		const buttonHTML = `<button class="button" name="send" value="/msgroom staff, /forcerename ${escapedName}&#10;/j view-usersearch-${target}">Forcerename</button>`;
```

server/chat-commands/moderation.ts
Zarel, line 1545:
```suggestion
	forcerename(target, room, user, connection, cmd) {
		if (!target) return this.parse('/help forcerename');

```

Consensus seems to be no need for `/frinstaff`.

KrisXV:
done @Zarel

PR #8042, opened 2021-02-16
Use /msgroom in more places

KrisXV:


server/chat-commands/core.ts
```diff
@@ -485,13 +485,15 @@ export const commands: ChatCommands = {
...
-	async msgroom(target, room, user, connection) {
+	async msgroom(target, room, user, connection, cmd) {
 		const [targetId, message] = Utils.splitFirst(target, ',').map(i => i.trim());
 		if (!targetId || !message) {
 			return this.parse(`/help msgroom`);
 		}
 		const targetRoom = Rooms.search(toID(targetId));
-		if (!targetRoom) return this.errorReply(`Room not found`);
+		if (!targetRoom) return this.errorReply(`Room not found.`);
+		if (!targetRoom.users[user.id]) return this.errorReply(`You must be in the room "${targetRoom.title}" to message it.`);
```
Zarel, line 495: I don't think this is necessary; I think this just makes it harder on Pages. Presumably moderation commands should still work when you're not in a room.

KrisXV: fair

KrisXV: Main thing i don't like about this is you can currently use it to just send casual messages like 'hi' to rooms you're not in

server/chat-commands/core.ts
```diff
@@ -485,13 +485,15 @@ export const commands: ChatCommands = {
...
+		if (message.startsWith(`/${cmd}`)) return this.errorReply(`You may not use /msgroom to parse itself in other rooms.`);
```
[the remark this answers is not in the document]
KrisXV: you're right; i tried fiddling with `this.handler.name` and `this.handler.aliases` but there's little need for this to get an alias, so i just hardcoded it

server/chat-commands/core.ts
```diff
@@ -491,7 +491,10 @@ export const commands: ChatCommands = {
 			return this.parse(`/help msgroom`);
 		}
 		const targetRoom = Rooms.search(toID(targetId));
-		if (!targetRoom) return this.errorReply(`Room not found`);
+		if (!targetRoom) return this.errorReply(`Room not found.`);
+		if (message.trim().startsWith('/msgroom')) {
```
Zarel, line 495:
```suggestion
		if (message.trim().startsWith('/msgroom ')) {
```

server/chat-commands/core.ts
```diff
@@ -491,7 +491,10 @@ export const commands: ChatCommands = {
...
+			return this.errorReply(`You may not use /msgroom to parse itself in other rooms.`);
```
Zarel, line 496:
```suggestion
			return this.errorReply(`Please do not nest /msgroom inside itself.`);
```

Zarel (approved):
Looks good other than my two remaining suggestions!

KrisXV:
@Zarel what other suggestions? I already addressed them all.

Zarel:
> @Zarel what other suggestions? I already addressed them all.

I'm sorry, I can't tell if you're joking or not. If you're not joking, I was talking about the two you just applied.

PR #8045, opened 2021-02-16
Lib: Add a static file server

mia-pi-git:
edit: oops 
anyway yeah this is basically just node-static without the vuln.

### DIFF
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -107,7 +107,7 @@ export class NetServer {
 	}
 	checkPath(pathname: string) {
 		const path = pathModule.basename(pathname);
-		if (path.startsWith('.') || path.includes('../')) return;
+		if (path.startsWith('.') || pathname.includes('..')) return;
 		return path;
 	}
 }
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -107,7 +107,7 @@ export class NetServer {
 	}
 	checkPath(pathname: string) {
 		const path = pathModule.basename(pathname);
-		if (path.startsWith('.')) return;
+		if (path.startsWith('.') || path.includes('../')) return;
 		return path;
 	}
 }
--- a/lib/server.ts
+++ b/lib/server.ts
@@ -1,0 +1,99 @@
+import * as http from 'http';
+import * as https from 'https';
+import {FS, FSPath} from './fs';
+import * as Utils from './utils';
+
+type FileListener = (this: NetServer, req: http.IncomingMessage, response: http.ServerResponse, body: AnyObject) => any;
+
+interface NetServerOptions {
+	https?: https.ServerOptions;
+	notFound?: string;
+	server?: http.Server;
+	default?: FileListener;
+	port?: number;
+}
+
+export class ServerError extends Error {
+	status: number;
+	constructor(message: string, status = 404) {
+		super(message);
+		this.status = 404;
+		Error.captureStackTrace(this, Error);
+	}
+}
+
+export class NetServer {
+	opts: NetServerOptions;
+	dir: string;
+	listeners: {[path: string]: {callback: FileListener, isDir?: boolean}};
+	notFound: FSPath;
+	constructor(dir: string, opts: NetServerOptions = {}) {
+		this.dir = dir;
+		this.opts = opts;
+		this.listeners = {};
+		this.notFound = opts.notFound ? FS(opts.notFound) : FS(`server/static/404.html`);
+	}
+	get(path: string, callback: FileListener, isDir = false) {
+		this.listeners[path] = {callback, isDir};
+	}
+	getListener(path: string) {
+		if (this.listeners[path]) return this.listeners[path];
+		for (const k in this.listeners) {
+			const listener = this.listeners[k];
+			if (listener.isDir && k.startsWith(path)) {
+				return listener;
+			}
+		}
+	}
+	serve(
+		req: http.IncomingMessage,
+		response: http.ServerResponse,
+		errorCallback?: (this: NetServer, error: Error) => any,
+	) {
+		if (req.method !== 'GET') {
+			response.writeHead(405, 'Invalid request type');
+			return response.end();
+		}
+		let pathname = req.url || `/404.html`;
+		if (!pathname.startsWith('/')) pathname = `/${pathname}`;
+		const listener = this.getListener(pathname);
+		const body: AnyObject = {};
+		if (listener) {
+			const [, args] = Utils.splitFirst(pathname, '?');
+			if (args) {
+				const parts = args.split('&');
+				for (const part of parts) {
+					const [name, data] = part.split('=');
+					body[name] = data;
+				}
+			}
+			void listener.callback.call(this, req, response, body);
+		}
+
+		try {
+			const path = FS(this.dir +  pathname);
+			if (!path.existsSync() || path.isDirectorySync()) {
+				return response.end(this.notFound.readIfExistsSync());
+			}
+			response.writeHead(200);
+			response.end(path.readSync());
+		} catch (e) {
+			response.writeHead(404);
+			response.end();
+			if (errorCallback) {
+				return errorCallback.call(this, e);
+			}
+			throw new ServerError(`Error serving request: ${req.url} (${this.dir}) - ${e.message}`);
+		}
+
+	}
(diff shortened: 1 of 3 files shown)


server/sockets.ts
```diff
@@ -216,17 +216,15 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
...
 			const roomidRegex = /^\/(?:[A-Za-z0-9][A-Za-z0-9-]*)\/?$/;
-			const cssServer = new StaticServer('./config');
-			const avatarServer = new StaticServer('./config/avatars');
-			const staticServer = new StaticServer('./server/static');
+			const cssServer = new NetServer('./config');
+			const avatarServer = new NetServer('./config/avatars');
+			const staticServer = new NetServer('./server/static');
 			const staticRequestHandler = (req: http.IncomingMessage, res: http.ServerResponse) => {
-				// console.log(`static rq: ${req.socket.remoteAddress}:${req.socket.remotePort} -> ${req.socket.localAddress}:${req.socket.localPort} - ${req.method} ${req.url} ${req.httpVersion} - ${req.rawHeaders.join('|')}`);
+				Monitor.debug(`static rq: ${req.socket.remoteAddress}:${req.socket.remotePort} -> ${req.socket.localAddress}:${req.socket.localPort} - ${req.method} ${req.url} ${req.httpVersion} - ${req.rawHeaders.join('|')}`);
```
Zarel, line 224:
```suggestion
				// console.log(`static rq: ${req.socket.remoteAddress}:${req.socket.remotePort} -> ${req.socket.localAddress}:${req.socket.localPort} - ${req.method} ${req.url} ${req.httpVersion} - ${req.rawHeaders.join('|')}`);
```

Too spammy for prod.

server/sockets.ts
```diff
@@ -244,7 +242,7 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 
 					server.serve(req, res, e => {
 						if (e && (e as any).status === 404) {
-							staticServer.serveFile('404.html', 404, {}, req, res);
+							staticServer.serveFile('/404.html', res, {code: 404, thisDir: true});
```
Zarel, line 245: :|

You told me `404.html` was the default, but it's clearly explicitly specified here!

lib/server.ts
```diff
@@ -0,0 +1,98 @@
...
+				for (const part of parts) {
+					const [name, data] = part.split('=');
+					body[name] = data;
+				}
+			}
+			void listener.callback.call(this, req, response, body);
+		}
+
+		try {
+			const path = FS(this.dir + pathname);
```
Zarel, line 74: This is a vuln! Imagine if someone went to `/../../../etc/passwd`. They could see your passwords!

Zarel (changes requested):
major vuln

server/sockets.ts
mia-pi-git: (for the record, this only logs if config.loglevel is set to something it isn't on prod)

server/sockets.ts
mia-pi-git: There's code in the module to default to it - i just left this in as a fallback.

lib/server.ts
```diff
@@ -0,0 +1,113 @@
...
+		if (!FS(file).existsSync()) {
+			return response.end(this.notFound.readIfExistsSync());
+		}
+		const filename = opts.thisDir ? this.dir + file : file;
+		const content = FS(filename).readIfExistsSync();
+		response.end(content);
+	}
+	checkPath(pathname: string) {
+		const path = pathModule.basename(pathname);
+		if (path.startsWith('.')) return;
```
Zarel, line 110: I don't believe this fixes the vuln I was talking about. This will only ban files starting with `.`, not directories.

mia-pi-git: That work a bit better?

lib/server.ts
```diff
@@ -0,0 +1,113 @@
...
+		if (!FS(file).existsSync()) {
+			return response.end(this.notFound.readIfExistsSync());
+		}
+		const filename = opts.thisDir ? this.dir + file : file;
+		const content = FS(filename).readIfExistsSync();
+		response.end(content);
+	}
+	checkPath(pathname: string) {
+		const path = pathModule.basename(pathname);
+		if (path.startsWith('.') || path.includes('../')) return;
```
Zarel, line 110: This doesn't change anything. A `basename` will never contain `/`. This will also fail on Windows, which doesn't use `/` as a path separator.

Zarel: Do you know what `basename` does?

server/sockets.ts
Zarel: I mean, I don't think there's code in `node-static` to default to it, which was your original reasoning for adding code to your own static file server to default to it.

lib/server.ts
mia-pi-git: yes, this was a mistake, i meant to make the includes('../') check `pathname` but i messed up.

Zarel (approved):

lib/server.ts
Zarel: At least part of the problem was that you called the basename `path`. Calling it `filename` might have prevented this bug. This is why it's important to give variables good names.

Zarel (changes requested):
I'm unapproving this because I see several major bugs, such as trying to write responses after `response.end()`

Zarel:
This also uses Sync methods which it really shouldn't, and the API isn't Promise when it should be.

Zarel:
The redundant `serve` and `serveFile` methods are also pretty suspicious.

AnnikaCodes:
@mia-pi-git, are you planning to do more work on this PR?

mia-pi-git:
Don't think we still care about this.

Zarel:
Reopening while we try to decide what to do between this and #11294

lib/server.ts
```diff
@@ -0,0 +1,113 @@
...
+		const body: AnyObject = {};
+		if (listener) {
+			const [, args] = Utils.splitFirst(pathname, '?');
+			if (args) {
+				const parts = args.split('&');
+				for (const part of parts) {
+					const [name, data] = part.split('=');
+					body[name] = data;
+				}
+			}
```
Slayer95, line 70: ``querystring`` module deals with this.

Slayer95:
Well, this API is nice, but security handling is way too sloppy. Sync fs access is a big issue, but also easily fixable. Anyway, the least we need to ensure is that there is no arbitrary directory traversal (the node-static vulnerability, which for the npm module it isn't actually as "arbitrary" as the vuln name sounds.)

Merits of this PR over #11249:
- Identical API to node-static
- Centralized (albeit not correct) path checking.
- Rejects non-read-only HTTP methods (but it's overly zealous, and it also rejects HEAD and OPTIONS, which can result in excessive traffic.).

Anyway, I don't know why I am examining all these custom static servers. Again, why don't we use your fork, @Zarel ? ``~.~``

Zarel:
My fork has some unnecessary subdependencies, and no TypeScript support, and a lot of features we don't actually need. I figure a very basic static server is probably easier.

The main alternative would be using node-static's master branch which has had the fix for a few years now.

Slayer95:
> The main alternative would be using node-static's master branch which has had the fix for a few years now.

Sure, who's against this?

Zarel:
I'm not _against_ it per se but I figure it'd be a nice opportunity to have a lightweight TypeScripted alternative.

mia-pi-git:
> I'm not _against_ it per se but I figure it'd be a nice opportunity to have a lightweight TypeScripted alternative.

Gonna toss my support in for just installing the master branch via NPM. Easier and fewer things for us to maintain.

Zarel:
Maintaining a simple server that has the features we want and doesn't have the features we don't want is much easier than maintaining a depedency on an unstable thing that'll randomly break or have bugs that are difficult to debug because it's not our code, not TypeScript, many more lines, etc.

Like I do think dependencies are worth it for complicated stuff. But Dieter's PR is like under 200 lines.

Slayer95:
> a lot of features we don't actually need

Please clarify. At least caching avatars is a feature that can't just be discarded as "unneeded". Compression is yea not a big deal for PS purposes. But what other bloat is there in node-static?

Zarel:
> But what other bloat is there in node-static?

The `minimatch` dep, mostly. The callback juggling is also outdated by modern standards.

Zarel:
Superseded by #11295

Feel free to suggest changes there.

PR #8047, opened 2021-02-16
Add 'ps' output to /processes

mia-pi-git:
Zarel mentioned that we should do this a while ago (iirc? thought it was brought up), and it's a useful enough tool for debugging it couldn't hurt to have it here.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -610,7 +610,7 @@ export const commands: ChatCommands = {
 
 		const processes = new Map<string, ProcessData>();
 
-		const psOutput = child_process.execSync('ps -o pid,%mem,time,command', {cwd: `${__dirname}/../..`}).toString();
+		const psOutput = child_process.execSync('ps -o pid,%cpu,time,command', {cwd: `${__dirname}/../..`}).toString();
 		const rows = psOutput.split('\n').slice(1); // first line is the title
 		for (const row of rows) {
 			if (!row.trim()) continue;
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -649,7 +649,7 @@ export const commands: ChatCommands = {
 			}
 		}
 		buf += `<br />`;
-		buf += `<strong>Other processes:</strong><br />`;
+		buf += `<details class="readmore"><summary><strong>Other processes:</strong></summary>`;
 
 		for (const [pid, process] of processes) {
 			buf += `<strong>${pid}</strong> - <code>${process.cmd}</code>`;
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -602,16 +608,57 @@ export const commands: ChatCommands = {
 			this.checkCan('lockdown');
 		}
 
-		let buf = `<strong>${process.pid}</strong> - Main<br />`;
+		const processes = new Map<string, ProcessData>();
+
+		const psOutput = child_process.execSync('ps -o pid,%mem,time,command', {cwd: `${__dirname}/../..`}).toString();
```
Zarel, line 613: Is `cwd` necessary? `ps` isn't touching the filesystem, is it?

mia-pi-git: I'm not quite sure? I figured better safe than sorry, especially since `bash` does it above as well.

Zarel: Also, why not `%cpu`? That's the whole reason we wanted it in he first place, right?

mia-pi-git: lol yeah i switched them - fixed.

server/chat-commands/admin.ts
```diff
@@ -602,16 +608,57 @@ export const commands: ChatCommands = {
 			this.checkCan('lockdown');
 		}
 
-		let buf = `<strong>${process.pid}</strong> - Main<br />`;
+		const processes = new Map<string, ProcessData>();
+
+		const psOutput = child_process.execSync('ps -o pid,%cpu,time,command', {cwd: `${__dirname}/../..`}).toString();
+		const rows = psOutput.split('\n').slice(1); // first line is the title
```
Zarel, line 614:
```suggestion
		const rows = psOutput.split('\n').slice(1); // first line is the table header
```

server/chat-commands/admin.ts
```diff
@@ -602,16 +608,57 @@ export const commands: ChatCommands = {
...
+		buf += `<strong>Other processes:</strong><br />`;
+
+		for (const [pid, process] of processes) {
+			buf += `<strong>${pid}</strong> - <code>${process.cmd}</code>`;
+			if (process.cpu) buf += ` (CPU: ${process.cpu}`;
+			if (process.time) {
+				buf += `${process.cpu ? `, ` : ' ('}time: ${process.time})`;
+			}
+			buf += `<br />`;
+		}
```
Zarel, line 661: This should be in a readmore; it's too spammy to show by default.

Zarel: It plausibly shouldn't be shown at all, but I guess we can try it out for now and see how it goes.

Zarel:
This is exactly what I wanted! Thanks a ton!

PR #8050, opened 2021-02-17
Add Clanger as an Alias for Kommo-o

ACakeWearingAHat:
This has been used by both the UU community and the sim in general as the colloquial term for kommo-o for years at this point. Ask Estarossa, Lilburr, and other UU community members for confirmation if needed.

### DIFF
--- a/data/aliases.ts
+++ b/data/aliases.ts
@@ -670,6 +670,7 @@ export const Aliases: {[alias: string]: string} = {
 	cathy: "Trevenant",
 	chandy: "Chandelure",
 	chomp: "Garchomp",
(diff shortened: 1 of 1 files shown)


Zarel:
There were a few others in the suggestions thread. Do you want to add them all at once?

ACakeWearingAHat:
Honestly I did this independently from the suggestions thread but yeah, that sounds good

ACakeWearingAHat:
and by that i mean feel free to alias whatever has enough approval; i am not in enough communities to comment on most of the ones in the thread

Zarel:
Hm, let's just accept this separately, then.

PR #8051, opened 2021-02-18
Make Scalemons apply before Tier Shift when they are mashed up

urkerab:
Without a priority, the two event handlers run in random order, which is frustrating the OM mashup community, as Tier Shift Scalemons gives completely different results to Scalemons Tier Shift.

I also added priorities to Flipped and 350 Cup for consistency, although they're not mashed up enough for anyone to care that strongly about it.


KrisXV:
i really dont know how i missed this pr, sorry

PR #8052, opened 2021-02-18
Add aliases for Tauros, Mewtwo, and Tentacruel

PlagueVonKarma:
I think I did this right...nag me if I screwed up. 

The RBY community has always called Tauros "The Bull" - or more informally, "bull" - as informal nomenclature. It dates back to Azure Heights, literally over a _decade_ old. M2 is typically an abbreviation for Mewtwo I see in all kinds of communities, and cruel is sometimes used for Tentacruel. If you need confirmation on any of this, poke the RBY Community with a stick and you'll see it.

I suppose something more tangible would be good, though. "Bull" is just used on the RBY dex's set for Tauros. As for "M2", there's a famous "BooM2" set that you can find on places like Pokemon Perfect, or you can poke a regular Uber conversation in any generation. I don't have anything in mind for Tentacruel but PS chats would probably turn it up. 

Had this stuff noted here, all names I've tested have been crossed out;
https://www.smogon.com/forums/threads/alias-megathread.3676993/post-8730138

If I think of any others I'll either add them here or make a separate PR if this is merged.


AnnikaCodes (approved):
I'm fine with this, thanks!

PR #8056, opened 2021-02-19
Hangman: Correctly check guesses with non-alphabetic answer

cheesetheorange:
Currently word guesses on hangmans with non-alphabetic characters in the answer check the length of the sanitized guess against the unsanitized answer, leading to correct answers being rejected as invalid.

### DIFF
--- a/server/chat-plugins/hangman.ts
+++ b/server/chat-plugins/hangman.ts
@@ -113,7 +113,7 @@ export class Hangman extends Rooms.RoomGame {
 	}
 
 	guessWord(word: string, guesser: string) {
-		const ourWord = toID(this.word);
+		const ourWord = toID(this.word.replace(/[^A-Za-z ]/g, ''));
 		const guessedWord = toID(word);
 		if (ourWord === guessedWord) {
 			for (const [i, letter] of this.wordSoFar.entries()) {


Zarel:
Err, `toID` already does most of this. Is this about numbers, or is there a different bug?

Can you give an example of a question/answer pair that fails?

cheesetheorange:
ah yeah i missed testing with non-alphanumeric, but this addresses numbers (any question with a number in it can't be solved with guessword as of now)

server/chat-plugins/hangman.ts
```diff
@@ -113,7 +113,7 @@ export class Hangman extends Rooms.RoomGame {
 	}
 
 	guessWord(word: string, guesser: string) {
-		const ourWord = toID(this.word);
+		const ourWord = toID(this.word.replace(/[^A-Za-z ]/g, ''));
```
Zarel, line 116:
```suggestion
		const ourWord = toID(this.word).replace(/[0-9]/g, '');
```

server/chat-plugins/hangman.ts
```diff
@@ -113,7 +113,7 @@ export class Hangman extends Rooms.RoomGame {
...
 		const guessedWord = toID(word);
```
Zarel, line 117:
```suggestion
		const ourWord = toID(this.word.replace(/[0-9]+/g, ''));
		const guessedWord = toID(word.replace(/[0-9]+/g, ''));
```

I understand what your code is doing now, but I think this is more readable (and allows passing in unsanitized guesses for future-proofing).

AnnikaCodes:
@Zarel?

AnnikaCodes:
This seems fine to me (although my naive impulse would be to include numbers in the sanitized word) but I don't work on Hangman very much.

Zarel:
Thanks!

Hangman traditionally has letters only; including numbers would make it confusing, I think.

PR #8059, opened 2021-02-21
Implement RandomTeams#multipleSamplesNoReplace

AnnikaCodes:


data/random-teams.ts
```diff
@@ -225,6 +225,23 @@ export class RandomTeams {
 		return this.fastPop(list, index);
 	}
 
+	/**
+	 * Removes n random elements from an unsorted array and returns them.
+	 * If n is less than or equal to the array's length, returns the array unchanged.
```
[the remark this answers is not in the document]
AnnikaCodes: Hmmm, running `sampleNoReplace` that many times could lead to an array filling up with `undefined` values; it might be better to throw an error if we try to sample more times than the array has elements, leaving the caller to decide how they want to handle a short array?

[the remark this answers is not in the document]
Zarel: `throw`ing might be a good idea in general, but I think it would make the benefit over just calling `sampleNoReplace` really marginal.

"Returns the entire array unchanged" isn't true, though; presumably the array is still sorted randomly.

data/random-teams.ts
```diff
@@ -225,6 +225,20 @@ export class RandomTeams {
...
+	 */
+	multipleSamplesNoReplace<T>(list: T[], n: number): T[] {
+		const samples = [];
+		while (samples.length < n && list.length) {
+			const index = this.random(list.length);
+			samples.push(this.fastPop(list, index));
```
Zarel, line 236:
```suggestion
			samples.push(this.sampleNoReplace(list));
```

data/random-teams.ts
```diff
@@ -225,6 +225,19 @@ export class RandomTeams {
 		return this.fastPop(list, index);
 	}
 
+	/**
+	 * Removes n random elements from an unsorted array and returns them.
+	 * If n is less than or equal to the array's length, returns the array unchanged.
```
Zarel, line 230:
```suggestion
	 * Removes n random elements from an unsorted array and returns them.
	 * If n is less than or equal to the array's length, removes and returns all the elements
	 * in the array (so the returned array could have length < n).
```

Zarel: Good point, edited.

Zarel:
Neat, this looks good to me.

PR #8063, opened 2021-02-21
Use optional chaining in more places

AnnikaCodes:


Zarel:
Have you benchmarked? This is unlikely to have an impact, but some before/after numbers for full-test would make me feel better.

AnnikaCodes:
I ran full-test three times before and after (to account for differing load on my computer because I was going in and out of Zoom meetings, etc), removing the -dist folders each time. The average for before was 3:40.6 and after was 3:47.3. That's not too much of a difference given the variation between times, but I can certainly look for optional chains that are causing performance hits and remove them.
```
npm run full-test  213.28s user 4.82s system 108% cpu 3:20.33 total (before) 
npm run full-test  275.04s user 8.26s system 114% cpu 4:06.77 total (before)
npm run full-test  224.02s user 5.64s system 106% cpu 3:36.16 total (before)

npm run full-test  265.99s user 7.10s system 117% cpu 3:52.42 total (after)
npm run full-test  272.42s user 7.73s system 116% cpu 4:01.22 total (after)
npm run full-test  220.34s user 5.59s system 108% cpu 3:28.91 total (after)
```

Zarel:
Nah, if it's not visible on there, it's not worth worrying about. Thanks for checking for me!

PR #8066, opened 2021-02-22
Datasearch: Improve dexsearchhelp & movesearchhelp

Distrib-ps:


AnnikaCodes:
@Zarel do you have opinions on this?

Distrib-ps:
Bump

AnnikaCodes (approved):
I'm not too opinionated about the design of help messages, but this looks OK to me. It might be worth moving "The order of the parameters does not matter." to the summary; that seems at least as relevant to the average user than a list of supported tiers?

Zarel:
Sorry, I think I missed this.

As far as I can tell, all this does is hide a bunch of the help behind a readmore. Is there a reason for this?

Zarel:
I'm not against this but I'd prefer to understand more about why people want this.

Distrib-ps:
Requested on discord :

Zarel:
@AnnikaCodes? I'll let you decide I guess.

AnnikaCodes:
I wanted a readmore because it makes broadcasting the help for these commands much less spammy. I think this would be a good usecase for your `readmore-broadcast` idea.

Zarel:
There's no need to block this PR on that. It's been open a long time, let's merge and implement `readmore-broadcast` some other time.

Zarel:
Thanks, @Distrib-ps!

PR #8067, opened 2021-02-23
Add and improve various move failure messages

DaWoblefet:
This PR does a number of small things:

- Convert Damp, Queenly Majesty, and Dazzling protocol to properly use ``-block`` protocol
- Add a failure message to Corrosive Gas
- Add the always-plays and failure message of Teatime
- Add the proper message for a Pokemon attempting to use an attack mid-Disable
- Add the proper fail message to Aura Wheel

While I'm at it, I'd like to properly add the failure messages for Rest vs. Insomnia / Vital Spirit and Own Tempo vs. confusion, but I  those require a client-side update too, so I'm drafting this for now until I figure out how to properly test my server-side changes there.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3694,7 +3694,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	suctioncups: {
 		onDragOutPriority: 1,
 		onDragOut(pokemon) {
-			this.add('-activate', pokemon, 'ability: Suction Cups');
+			this.add('-immune', pokemon, 'ability: Suction Cups');
 			return null;
 		},
 		name: "Suction Cups",
@@ -3843,7 +3843,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	telepathy: {
 		onTryHit(target, source, move) {
 			if (target !== source && target.side === source.side && move.category !== 'Status') {
-				this.add('-activate', target, 'ability: Telepathy');
+				this.add('-immune', target, 'ability: Telepathy');
 				return null;
 			}
 		},
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -573,7 +573,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		onAnyTryMove(target, source, effect) {
 			if (['explosion', 'mindblown', 'mistyexplosion', 'selfdestruct'].includes(effect.id)) {
 				this.attrLastMove('[still]');
-				this.add('cant', this.effectData.target, 'ability: Damp', effect, '[of] ' + target);
+				this.add('-block', this.effectData.target, 'ability: Damp', effect, target);
 				return false;
 			}
 		},
@@ -626,7 +626,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			const dazzlingHolder = this.effectData.target;
 			if ((source.side === dazzlingHolder.side || move.target === 'all') && move.priority > 0.1) {
 				this.attrLastMove('[still]');
-				this.add('cant', dazzlingHolder, 'ability: Dazzling', move, '[of] ' + target);
+				this.add('-block', dazzlingHolder, 'ability: Queenly Majesty', move, target);
 				return false;
 			}
 		},
@@ -2830,7 +2830,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			const dazzlingHolder = this.effectData.target;
 			if ((source.side === dazzlingHolder.side || move.target === 'all') && move.priority > 0.1) {
 				this.attrLastMove('[still]');
-				this.add('cant', dazzlingHolder, 'ability: Queenly Majesty', move, '[of] ' + target);
+				this.add('-block', dazzlingHolder, 'ability: Queenly Majesty', move, target);
 				return false;
 			}
 		},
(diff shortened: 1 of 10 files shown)


Zarel:
Should we take this opportunity to make `-block` etc consistent?

If I remember, the end goal was to make it work like this:

- `-fail` - some move or other effect fails due to its own mechanics (Counter failing because the opponent hasn't moved)
- `-block` - some condition causes some move or other effect to fail (Mirror Coat failing because the opponent has Wonder Guard)
- `-immune` - only used for type immunities
- `-activate` - should not be used in situations where `-fail`/`-block` apply - used only when a condition does something that isn't already described by a different protocol message (Spite reducing PP)
- `-ability`, `-item` - the equivalent of `-start` but for abilities/items; should not be used where `-activate` applies (should only be used if the ability/item was changed, or revealed without any other effect; if it's revealed while having an effect, `-activate` applies) (Mold Break or Air Balloon announcing itself on switch-in)

DaWoblefet:
I'm down to include that as part of the process!

Zarel:
There's also an interesting argument that we don't need `-block` because it means the same thing as `-immune` so we should just always just use `-immune`. I don't know if that would be more or less confusing, though.

DaWoblefet:
Based on my thoughts in Discord, these are my thoughts on current protocol.

Failure   cause | Current protocol | Proposed protocol
-- | -- | --
Choice   lock / Gorilla Tactics active and trying to use a move not locked into | fail | cant
Inner   Focus, Own Tempo, Oblivious, and Scrappy blocking Intimidate | fail | immune
Primordial   Sea / Desolate Land blocking Fire / Water moves | fail | block
Aura   Wheel, Dark Void, and Hyperspace Fury as wrong Pokemon | fail | fail
Bide   with no energy stored | fail | fail
Burn Up   when not Fire-type | fail | fail
Corrosive   Gas when target lacks item | N/A | fail
Weight-based   moves into Dynamax | fail | block
Heal   Block against target that already is affected | fail | fail
Rest at   full HP / user has Insomnia or Vital Spirit | fail | fail
Sky Drop   and target above 200kg | fail | fail
Repeated   Soak | fail | fail (but should be corrected in modern gens)
Substitute | fail | fail
Move   failure on Sub (except stat failures) | fail | fail
Stat   failure on Sub | ? | immune
Teatime   with no Berries eaten | N/A | fail
Uproar   blocking Rest / sleep moves | fail | immune
Generic   'Try' or 'PrepareHit' failures | fail | fail
Generic   'TryHit' failures | fail | fail
Generic   'TryHitField' or 'TryHitSide' failures | fail | fail
If   damage returns false?!?!? Idk | fail | fail
If user   HP is 100% and healing move | fail | fail
If some   kind of healing fails, idk | fail | fail
Some   ultra generic failure if a Pokemon didn't do anything | fail | fail
Forced   switch failure | fail | fail
Truant   loaf | cant | cant
Full   para | cant | cant
Fast   asleep | cant | cant
Frozen | cant | cant
Flinched | cant | cant
Must   recharge | cant | cant
Attract   full infatuation | cant | cant
Disable   move is disabled | N/A | cant
Focus   Punch lost focus | cant | cant
Gravity   blocks move | cant | cant
Heal   Block blocks move | cant | cant
Imprison   blocks move | cant | cant
Magnet   Rise / Splash / Telekinesis Z-move failure in Gravity | cant | fail
Shell   Trap and user wasn't hit | cant | fail
Taunt   blocks move | cant | cant
Throat   Chop move failure | cant | cant
Move has   no PP | cant | cant
Aroma   Veil / Flower Veil / Pastel Veil / Sweet Veil | block | immune
Damp /   Queenly / Dazzling | cant | block
Dynamax   phazing moves | block | block
All   Abilities with -immune | immune | immune
Pollen   Puff into full HP | immune | fail
Rototiller   into ungrounded target | immune | immune
Sky Drop   and target has Flying-type | immune | immune
Telekinesis   into Diglett etc | immune | immune
Powder | immune | immune
TryImmunity | immune | immune
Prankster | immune | immune
OHKO   moves vs Dynamax or higher  level | immune | immune
Suction   Cups | activate | immune
Telepathy | activate | immune
Safety   Goggles | activate | immune
Electric   Terrain | activate | immune
Ingrain | activate | immune
Mist | activate | immune
Misty   Terrain | activate | immune
Psychic   Terrain | activate | immune
Safeguard | activate | immune

Zarel:
Hm, so, as mentioned earlier, I think it's impossible to pick a sensical difference between `-immune` and `-block`. I can't figure out your rule here.

As for `cant` vs `-fail`, the difference has always been:

- Replaces "[POKEMON] used [MOVE]" → `cant`
- Otherwise → `-fail`

You seem to be correctly reflecting this, though. Your other changes appear to be the correct fixes, but again, I can't figure out your line between `-immune` and `-block`, and I suspect it won't be the same as others' lines between `-immune` and `-block`, and there's no real reason for them _not_ to be the same message.

DaWoblefet:
I think merging them is good as well. I don't have a torch to carry one way or another, and simplifying protocol is good! I'll start tackling that now as well as the rest of the list, and note changes in another comment.

DaWoblefet:
The last three commits are untested - I want to go through and verify stuff should work like I expect it to.

Zarel:
If we've decided to make them all `-immune`, we might want to consider also renaming the entries in `data/text/`.

AnnikaCodes:
@DaWoblefet are you planning to do more work on this PR?

mia-pi-git:
https://discord.com/channels/630837856075513856/630845310033330206/1165829327456325662

PR #8072, opened 2021-02-24
Moderation: Don't hidetext for locks in tickets

mia-pi-git:
https://www.smogon.com/forums/threads/3514540/page-19#post-8763488


Zarel:
Thanks!

PR #8073, opened 2021-02-24
CC1v1: Reban Dynamax

xInstruct:
Just assuming this was an accident.


KrisXV:
It wasn't

AnnikaCodes:
ah, ok.

I didn't expect the rule inheritance to work like that; that's on me

xInstruct:
so uh

PR #8077, opened 2021-02-27
Add Gen 1 Hackmons Cup

KrisXV:


data/mods/gen1/random-teams.ts
```diff
@@ -342,6 +348,112 @@ export class RandomGen1Teams extends RandomGen2Teams {
...
+				}
+			} while (moves.length < 4);
+
+			// Random EVs
+			const evs = {hp: 0, atk: 0, def: 0, spa: 0, spd: 0, spe: 0};
+			const s: StatName[] = ['hp', 'atk', 'def', 'spa', 'spd', 'spe'];
+			for (const x of s) {
+				if (x === 'spd') continue;
+				evs[x] = this.random(256);
+			}
```
AnnikaCodes, line 398:
```suggestion
			const evs = {
				hp: this.random(256),
				atk: this.random(256),
				def: this.random(256),
				spa: this.random(256),
				spd: 0,
				spe: this.random(256)
			};
```
Is there a reason why this wouldn't work? It's a lot cleaner.

config/formats.ts
```diff
@@ -1629,6 +1629,31 @@ export const Formats: FormatList = [
...
+		name: "[Gen 1] Hackmons Cup",
+		desc: `Randomized teams of level-balanced Pok&eacute;mon with absolutely any moves, types, and stats.`,
+
+		mod: 'gen1',
+		team: 'randomHC',
+		searchShow: false,
+		ruleset: ['HP Percentage Mod', 'Cancel Mod', 'Desync Clause Mod', 'Sleep Clause Mod', 'Freeze Clause Mod', 'Team Preview'],
+		onBegin() {
+			for (const pokemon of this.getAllPokemon()) {
+				this.add(`c|\u2606${pokemon.side.name}|${pokemon.name}'s types: ${(pokemon.set as any).hc.types.join('/')}`);
```
AnnikaCodes, line 1642: I think it's a bit weird to make the messages come from the player - during testing you managed to get it to just show up as the small text in battles like the rules. If you're going to use `c|` I'd rather it come from `&Staff` or `&Battle` rather than the plater.

data/mods/gen1/random-teams.ts
KrisXV: that's how current gen hackmons cup does it

AnnikaCodes: It looks like it does that because of some Gen 6 shenanigans (apparently Gen 6 Hackmons Cup has a 510-EV cap, but other gens remove that). Let's go with my suggestion, then; it does the same thing while being more readable and better performing.

KrisXV:
did @AnnikaCodes

AnnikaCodes (approved):
Ok, this looks good. Thanks.

PR #8079, opened 2021-02-27
Fix crash in /randset

mia-pi-git:
Was caused by the mon not having a learnset for this gen.

### DIFF
--- a/server/chat-plugins/random-battles.ts
+++ b/server/chat-plugins/random-battles.ts
@@ -826,7 +826,7 @@ export const commands: ChatCommands = {
 		}
 		const dex = Dex.forFormat(format);
 		if (!dex.getLearnsetData(species.id).learnset) {
-			return this.errorReply(`Learnsets for ${species.name} in ${format.name} not found.`);
+			throw new Chat.ErrorMessage(`Learnsets for ${species.name} in ${format.name} not found.`);
 		}
 
 		const set = Dex.getTeamGenerator(format.id).randomSet(species.id);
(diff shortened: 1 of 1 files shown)


server/chat-plugins/random-battles.ts
```diff
@@ -824,6 +824,10 @@ export const commands: ChatCommands = {
 		if (!species.exists) {
 			throw new Chat.ErrorMessage(`${pokemonString} is not a valid Pokémon for that format.`);
 		}
+		const dex = Dex.forFormat(format);
+		if (!dex.getLearnsetData(species.id).learnset) {
+			return this.errorReply(`Learnsets for ${species.name} in ${format.name} not found.`);
```
AnnikaCodes, line 829:
```suggestion
			throw new Chat.ErrorMessage(`Learnsets for ${species.name} in ${format.name} not found.`);
```

AnnikaCodes (dismissed):
Looks good, thanks!

mia-pi-git:
> D130CAB1-E34D-4239-9C1E-32042304C9D9
> It’s not a matter of which gen the Pokémon is in. Seems like someone tried to use `/randset` on a Pokestar Pokémon, which have no learnset in any gen.

Yes, i'm aware. That's what i'm trying to fix, by adding a more general solution to check if they have a learnset in the target format at all.

AnnikaCodes:
Both checks should use the modded dex.

AnnikaCodes (approved):
Looks good, thanks!

PR #8080, opened 2021-02-27
Help tickets: Refresh ticket list in 'more tickets' button on click

mia-pi-git:
Minor qol.


Zarel:
I think we should probably have some way for making links do this.

Zarel:
(Links are more accessible in other ways.)

PR #8082, opened 2021-02-28
NU: Ban Mienshao

Distrib-ps:
https://www.smogon.com/forums/threads/mienshao-is-now-banned-from-ss-nu.3679135/


HoeenCoder:
Thanks!

PR #8086, opened 2021-03-01
Let "Allow Tradeback" operate for any generation

KrisXV:
A ton of rooms would benefit from this; for example, Ruins of Alph is trying to do Gen 6 OU tours + [insert Gen 7 mon], and this would be much easier than adding `!Obtainable Moves` to the tour rules

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1558,7 +1558,7 @@ export class TeamValidator {
 		const dex = this.dex;
 		let name = set.species;
 		const species = dex.getSpecies(set.species);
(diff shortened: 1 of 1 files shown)


KrisXV:
i know this is only 2 days old but bumping, it would help out a lot with a ton of various "What if" tours in the RoA room (what if tours are just tournaments in one gen with a future gen mon added; we have an event next week) and would be helpful for OM mashups as well

Zarel:
I don't think the rest of the code is written with this in mind so it might not function correctly, but sure, let's give this a try.

PR #8092, opened 2021-03-02
Restore /potd

mia-pi-git:
It's in `admin.ts` because i was unsure of where else to put it, and admin already has the ProcessManager import. (plus it's console only so...)

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -106,7 +106,6 @@ async function rebuild(context: CommandContext) {
 export const commands: ChatCommands = {
 	potd(target, room, user) {
 		this.canUseConsole();
-		target = toID(target);
 		const species = Dex.getSpecies(target);
 		if (species.id === Config.potd) {
 			return this.errorReply(`The PotD is already set to ${species.name}`);
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -104,6 +104,25 @@ async function rebuild(context: CommandContext) {
 
 
 export const commands: ChatCommands = {
+	potd(target, room, user) {
+		this.canUseConsole();
+		target = toID(target);
+		const species = Dex.getSpecies(target);
+		if (!species.exists) return this.errorReply(`Pokemon ${target} not found.`);
+		if (!Dex.getLearnsetData(species.id).learnset) {
+			return this.errorReply(`That Pokemon has no learnset and cannot be used as the PotD.`);
+		}
+		// this _could_ be an instanceof check but worst case we send it to another PM
+		// that's in that file and it does nothing
+		const simProcesses = ProcessManager.processManagers.filter(p => p.basename.includes('room-battle'));
+		for (const sim of simProcesses) {
+			for (const sub of sim.processes) {
+				sub.getProcess().send(`EVAL\n\nConfig.potd = '${species.id}'`);
+			}
+		}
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -104,6 +104,25 @@ async function rebuild(context: CommandContext) {
...
+			return this.errorReply(`That Pokemon has no learnset and cannot be used as the PotD.`);
+		}
+		// this _could_ be an instanceof check but worst case we send it to another PM
+		// that's in that file and it does nothing
+		const simProcesses = ProcessManager.processManagers.filter(p => p.basename.includes('room-battle'));
+		for (const sim of simProcesses) {
+			for (const sub of sim.processes) {
+				sub.getProcess().send(`EVAL\n\nConfig.potd = '${species.id}'`);
+			}
+		}
```
Zarel, line 122:
```suggestion
		}
		Config.potd = species.id;
```

Zarel, line 122: Why are you manually searching through PMs? Isn't this just `Rooms.PM`?

server/chat-commands/admin.ts
```diff
@@ -104,6 +104,25 @@ async function rebuild(context: CommandContext) {
 
 
 export const commands: ChatCommands = {
+	potd(target, room, user) {
+		this.canUseConsole();
+		target = toID(target);
```
Zarel, line 109: What's this for?

server/chat-commands/admin.ts
```diff
@@ -104,6 +104,24 @@ async function rebuild(context: CommandContext) {
...
+		if (species.id === Config.potd) {
+			return this.errorReply(`The PotD is already set to ${species.name}`);
+		}
+		if (!species.exists) return this.errorReply(`Pokemon ${target} not found.`);
+		if (!Dex.getLearnsetData(species.id).learnset) {
+			return this.errorReply(`That Pokemon has no learnset and cannot be used as the PotD.`);
+		}
+		Config.potd = species.id;
+		for (const sim of Rooms.PM.processes) {
+			sim.getProcess().send(`EVAL\n\nConfig.potd = '${species.id}'`);
```
Zarel, line 120:
```suggestion
		for (const process of Rooms.PM.processes) {
			process.getProcess().send(`EVAL\n\nConfig.potd = '${species.id}'`);
```

server/chat-commands/admin.ts
Zarel: `getSpecies` already has an ID conversion, and the only other place `target` is used is an error message that doesn't need to be in ID form.

```suggestion
```

server/chat-commands/admin.ts
mia-pi-git: honestly tho why is this `Rooms.PM` - why not `Rooms.RoomBattle.PM`? It's confusing.

server/chat-commands/admin.ts
```diff
@@ -104,6 +104,23 @@ async function rebuild(context: CommandContext) {
 
 
 export const commands: ChatCommands = {
+	potd(target, room, user) {
+		this.canUseConsole();
+		const species = Dex.getSpecies(target);
+		if (species.id === Config.potd) {
+			return this.errorReply(`The PotD is already set to ${species.name}`);
+		}
+		if (!species.exists) return this.errorReply(`Pokemon ${target} not found.`);
```
Zarel, line 113:
```suggestion
		if (!species.exists) return this.errorReply(`Pokemon "${target}" not found.`);
```

PR #8094, opened 2021-03-04
Add Spikemuth Cup Online Competition

LegoFigure11:
https://www.smogon.com/forums/threads/spikemuth-cup.3679445/
Apparently they ballsed up the implementation and it's not actually what  it's meant to be, so this implements the currently playable downloadable  ruleset and NOT the one described in the HOME app or the thread OP (see post 2 & 3)

If there's a better way of doing this, feel free to close this and PR it yourself - I won't have any time over the next couple of days to look after this PR.


LegoFigure11:
Closing as gf have updated the rule set on their end, will reopen once I’ve updated this to match.

cantsay:
TPCi fixed the errors and the ruleset is now clear. Can this be updated to match and go live ASAP?

galar dex + IoA dex + CT dex including all formes. NO dynamax

LegoFigure11:
Sorry that’s on me - I’ll do this in ~6 hours if no one gets to it first.

PR #8095, opened 2021-03-04
Admin: Support hotpatching User / Room / ProcessManager prototypes

mia-pi-git:
Should make some stuff a lot easier. 
I removed the duplicated `if (lock[patch])` stuff since it's rather annoyingly duplicated with no real point.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -40,6 +40,20 @@ export function escapeRegex(str: string) {
 	return str.replace(/[\\^$.*+?()[\]{}|]/g, '\\$&');
 }
 
+export function listMethods(obj: any, checker?: (prop: string, val: any) => boolean) {
+	const methods = new Set<string>();
+	let current = obj;
+	do {
+		const curProps = Object.getOwnPropertyNames(current);
+		for (const prop of curProps) {
+			if (typeof obj[prop] !== 'function') continue;
+			if (checker && !checker(prop, obj[prop])) continue;
+			methods.add(prop);
+		}
+	} while ((current = Object.getPrototypeOf(current)));
+	return [...methods.keys()];
+}
+
 /**
  * Escapes HTML in a string.
 */
@@ -368,5 +382,5 @@ export const Utils = {
 	shuffle, deepClone, clearRequireCache,
 	randomElement, forceWrap, splitFirst,
 	stripHTML, visualize, getString,
-	escapeRegex,
+	escapeRegex, listMethods,
 };
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -40,20 +40,6 @@ export function escapeRegex(str: string) {
 	return str.replace(/[\\^$.*+?()[\]{}|]/g, '\\$&');
 }
 
-export function listMethods(obj: any, checker?: (prop: string, val: any) => boolean) {
-	const methods = new Set<string>();
-	let current = obj;
-	do {
-		const curProps = Object.getOwnPropertyNames(current);
-		for (const prop of curProps) {
-			if (typeof obj[prop] !== 'function') continue;
-			if (checker && !checker(prop, obj[prop])) continue;
-			methods.add(prop);
-		}
-	} while ((current = Object.getPrototypeOf(current)));
-	return [...methods.keys()];
-}
-
 /**
  * Escapes HTML in a string.
 */
@@ -382,5 +368,5 @@ export const Utils = {
 	shuffle, deepClone, clearRequireCache,
 	randomElement, forceWrap, splitFirst,
 	stripHTML, visualize, getString,
-	escapeRegex, listMethods,
+	escapeRegex,
 };
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -34,7 +34,7 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
 	});
 }
 
-function listMethods(obj: any) {
+function listNonEnumerableKeys(obj: any) {
 	const methods = new Set<string>();
 	let current = obj;
 	do {
@@ -46,7 +46,7 @@ function listMethods(obj: any) {
 			methods.add(prop);
 		}
 	} while ((current = Object.getPrototypeOf(current)));
-	return [...methods.keys()];
+	return [...methods];
 }
 
 /**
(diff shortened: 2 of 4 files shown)


server/chat-commands/admin.ts
```diff
@@ -389,14 +395,17 @@ export const commands: ChatCommands = {
...
 
 		const lock = Monitor.hotpatchLock;
-		const hotpatches = ['chat', 'formats', 'loginserver', 'punishments', 'dnsbl', 'modlog'];
+		const hotpatches = [
+			'chat', 'formats', 'loginserver', 'punishments', 'dnsbl', 'modlog',
+			'processmanager', 'roomsp', 'usersp',
+		];
+		if (lock[target]) {
+			return this.errorReply(`That hotpatch has been disabled by ${lock[target].by} (${lock[target].reason})`);
+		}
```
Zarel, line 404: Certain hotpatch commands have aliases (like `/hotpatch commands` → `chat`) that you haven't accounted for, and even the ones you did account for (like `battles`), the approach seems very bug-prone. And while we've merged hotpatching formats/battles, I don't think it's unthinkable that we will have hotpatch commands that combine other hotpatch commands together, which this approach doesn't support well.

server/chat-commands/admin.ts
```diff
@@ -34,6 +34,12 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
 	});
 }
 
+const BLACKLISTED_PROPS = ['valueOf', 'constructor'];
+
+function isWhitelistedProp(prop: string) {
+	return !(prop.includes('__') || prop.toLowerCase().includes('prop') || BLACKLISTED_PROPS.includes(prop));
+}
```
Zarel, line 41: You should be using `hasOwnProperty`, not a literal blacklist.

server/chat-commands/admin.ts
mia-pi-git: Ok, I can restore that.

server/chat-commands/admin.ts
mia-pi-git: That doesn't seem to solve what i want, here?

server/chat-commands/admin.ts

lib/utils.ts
```diff
@@ -40,6 +40,20 @@ export function escapeRegex(str: string) {
...
+		const curProps = Object.getOwnPropertyNames(current);
+		for (const prop of curProps) {
+			if (typeof obj[prop] !== 'function') continue;
+			if (checker && !checker(prop, obj[prop])) continue;
+			methods.add(prop);
+		}
+	} while ((current = Object.getPrototypeOf(current)));
+	return [...methods.keys()];
+}
+
```
Zarel, line 56: I don't think this belongs in Utils. It's very specific to PS's hotpatching system, `chat-commands/admin` definitely strikes me as the correct place. Passing in an external checker also seems to be unnecessary abstraction (lower readability for no gain).

server/chat-commands/admin.ts
Zarel: Hm, I think I see. The whitelist should be documented. I'll suggest a change once you move `listMethods` over.

Zarel: I also don't see a need for you to check for functions specifically. Do you know why you did that?

mia-pi-git: I don't know if prototypes include non-functions? Nor do i know how it interacts, so I went for safety.

Zarel: Prototypes can include anything. In practice they shouldn't include non-functions (and the `class` syntax disallows non-functions), but they're supported by JS and should be copied if they do exist.

mia-pi-git: Ah, alright.

server/chat-commands/admin.ts
```diff
@@ -34,6 +34,21 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
...
+	let current = obj;
+	do {
+		const curProps = Object.getOwnPropertyNames(current);
+		for (const prop of curProps) {
+			if (prop.includes('__') || prop.toLowerCase().includes('prop') || ['valueOf', 'constructor'].includes(prop)) {
+				continue;
+			}
+			methods.add(prop);
+		}
+	} while ((current = Object.getPrototypeOf(current)));
```
Zarel, line 48: Wait, if you want all properties, just use `keys` instead of `getOwnPropertyNames`...

Zarel:
```ts
do {
	const curProps = Object.getOwnPropertyNames(current);
	for (const prop of curProps) {
		foo(prop);
	}
} while ((current = Object.getPrototypeOf(current)));
```

is a really long-winded way to do the exact same thing as

```ts
for (const prop in current) {
	foo(prop);
}
```

Zarel: Oh, I see, this is to include non-enumerable keys, because methods are non-enumerable by default now.

server/chat-commands/admin.ts
```diff
@@ -439,11 +463,100 @@ export const commands: ChatCommands = {
...
+				}
+				this.sendReply('DONE');
+			} else if (target === 'usersp' || target === 'roomsp') {
+				if (lock[target]) {
+					return this.errorReply(`Hot-patching ${target} has been disabled by ${lock[target].by} (${lock[target].reason})`);
+				}
+				let newProto: any, oldProto: any, message: string;
+				switch (target) {
+				case 'usersp':
+					newProto = require('../users').User.prototype;
```
Zarel, line 511: You don't clear the require cache. Does this even do anything?

server/chat-commands/admin.ts
```diff
@@ -34,6 +34,26 @@ function bash(command: string, context: CommandContext, cwd?: string): Promise<[
...
+			methods.add(prop);
+		}
+	} while ((current = Object.getPrototypeOf(current)));
+	return [...methods];
+}
+
+function keysToCopy(obj: object) {
+	return keysIncludingNonEnumerable(obj).filter(
+		// `__` matches sucrase init methods
+		// FIXME: document what 'prop' is for
```
Zarel, line 52: Please push an update documenting why these are excluded.

server/chat-commands/admin.ts
mia-pi-git: (the require cache is cleared above by Utils.clearRequireCache)

Zarel: Oh! You're doing a weird thing for lib but the normal clearing works here; I see.

PR #8097, opened 2021-03-05
Update pokedex.ts

Lucasrang:


Distrib-ps:
Why add this pokemon? He does not exist. There are only on megas for all (https://github.com/smogon/pokemon-showdown/blob/fcdfbca875b9db6990d1854ba21cc56fb8ee0c2e/data/mods/megasforall/pokedex.ts#L628)

PR #8104, opened 2021-03-07
Magician should not steal Weakness Policy on super-effective hits

Cureja:
https://github.com/smogon/pokemon-showdown/projects/3#card-52059185
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-42#post-8696614

Added a specific condition to check for Magician stealing Weakness Policy. Previously was stealing item without checking if the item got triggered.

I don't think there is an easier away around this issue other than a hard coded if statement because it has to do with trigger order of the item function (onDamagingHit, item.ts line 6665) and Magician ability function (onSourceHit, abilities.ts line 1665). I'm not sure if there needs to be change done for other items or item stealing abilities, but the bug only describes Magician and Weakness Policy.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -1918,6 +1918,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			if (!move || !target) return;
 			if (target !== source && move.category !== 'Status') {
 				if (source.item || source.volatiles['gem'] || move.id === 'fling') return;
+				if (target.hasItem('weaknesspolicy') && source.getMoveHitData(move).typeMod > 0 && target.hp > 0) return;
 				const yourItem = target.takeItem(source);
 				if (!yourItem) return;
 				if (!source.setItem(yourItem)) {


DaWoblefet:
This has a skipped test in magician.js; try unskipping it just to make sure it passes the test (change it.skip to just it)

Generally speaking though, I think this is a bit too much hardcoding. The real fix would be to make the event that runs Weakness Policy happen earlier than Magician. Generally speaking, Weakness Policy should run at the same timing as effects like Rocky Helmet dealing damage or Justified raising attack. Magician, on the other hand, runs much later, in the area of Moxie boosting Attack, or Eject Button switching a Pokemon out. 

I think the problem is that Magician is currently using a unique event handler (onSourceHit) when it should be using a more general one. Offhand, onAfterMoveSecondary would be a more appropriate place for Magician; @Marty-D do you have thoughts?

DaWoblefet:
This is of course generally problematic, as PS does not yet support post-move success checks very well. One of my latest research projects is doing a proper analysis of this order based on Gen 7 & 8 data.

Just to toss it here preemptively, these are the effects that occur at the same timing as Magician:
```
Maranga Berry
Red Card
Kee Berry
Eject Button
Color Change
Moxie
Beast Boost
Magician
Berserk
Scald autothaw on target
Sparkling Aria healing burn
Secret Power
Wake-Up Slap
Smelling Salts
Knock Off
Thief/Covet
Pluck/Bug Bite
Circle Throw/Dragon Tail
Smack Down
Fell Stinger
Thousand Arrows
Thousand Waves
```

Marty-D:
Yes, the solution is to rearrange the events so that everything triggers in the correct order. Some related discussion here: https://github.com/smogon/pokemon-showdown/commit/1f6e1e8d457f5cc37612be78b711e22470fea45b

Also note that Weakness Policy was swapped from `Hit` to `DamagingHit` recently to fix its interaction with Burning Jealousy, which may have broken this interaction. https://github.com/smogon/pokemon-showdown/commit/2b0c77939fee287d824115755c57c7f563b6631e

Cureja:
I will test the previous version to see if how it affects the timing interaction with Weakness Policy and some of the other items on the list and also try the aforementioned timing changes you both mentioned. Thanks for the comprehensive comments on the pull request and sorry that my previous fix was hard-coded as I didn't have enough knowledge of the ordering events.

AnnikaCodes:
@Cureja how is this going?

PR #8107, opened 2021-03-08
Update comment to reflect the build script

AnnikaCodes:
As far as I know, our documentation recommends using `node pokemon-showdown` to start PS, not `node build && node .server-dist/index.js`.
Thus, I don't think the latter is considered starting PS normally.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -2,8 +2,8 @@
  * Main file
  * Pokemon Showdown - http://pokemonshowdown.com/
  *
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

Zarel:
(In the past, we did support `node server` as the way to skip the build. Nowadays the launcher does enough other things that it's `./pokemon-showdown --no-build` I think.)

PR #8112, opened 2021-03-11
Chat monitor: Refactor to handle private filters better

mia-pi-git:
This does two things. One, it abstracts filter loading into a function (which is run a tick after the module is done), and makes it so missing filter words don't throw, as to avoid leaking private into. Loading is done a tick after the module to allow for other files to be loaded (the crashes were because they weren't being loaded before the filter loading was done, afaict.)

### DIFF
--- a/server/chat-plugins/chat-monitor.ts
+++ b/server/chat-plugins/chat-monitor.ts
@@ -288,42 +288,46 @@ Chat.registerMonitor('shorteners', {
  */
 
 export function loadFilters() {
+	let data;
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Bump, please, this is a relevant issue.

AnnikaCodes (approved):
This looks good to me; I'll leave it up for a day or two more in case anyone wants to comment.

Zarel:
Looks good to me.

PR #8118, opened 2021-03-14
Add alias for /poll queuehtml

tmagicturtle:
Some staff have expressed confusion on the syntax of queueing HTML polls, since it is inconsistent with /poll htmlcreate, where html- comes before create (compared to queuehtml, where -html comes after). Instead of changing it entirely and causing additional confusion for people who got used to one command, this allows either command to be used. Further, the command for queuing HTML polls was not listed in the help for the command.


Zarel:
ty!

PR #8119, opened 2021-03-15
Thing of the Day: Refactor storage / remove hardcodes

mia-pi-git:
I don't like the LEGACY_DATA hardcode, but I can't really think of a better way to migrate it. We should be able to remove that eventually.
This should make it so other rooms can use OTDs without having to make a code change, and also merges it so all OTDs are stored in a singular file (with settings), which _should_ make it easier to migrate in the future.


Zarel (approved):
Looks good, but we should probably backup config before hotpatching this, in case something breaks.

I'll do that when I have time, but anyone else should feel free to do that (backup config, hotpatch, make sure everything still works) for me.

PR #8121, opened 2021-03-16
SSB: Remove Easter Egg

xInstruct:
Requested

If you are wondering why this isn't in my current batch PR, I am sick as shit and the less headaches the better


Zarel:
Fair enough. You could also just ask me to merge the batch, for future reference.

PR #8122, opened 2021-03-17
Enable tsc warnings about unreachable code

AnnikaCodes:
To me, this seems like a useful thing to check in linting, with no disadvantage that I can see (I found no measurable performance difference in running `tsc` after removing the `.tsbuildinfo` file).

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,7 +8,8 @@
         "strict": true,
         "allowJs": true,
         "checkJs": true,
-        "incremental": true
+        "incremental": true,
+        "allowUnreachableCode": false
     },
     "types": ["node"],
     "exclude": [


Zarel:
Too bad it can't be actually done in the linter. I'd like for this to be skipped when testing, like we currently do for `const` etc.

AnnikaCodes:
> I'd like for this to be skipped when testing, like we currently do for `const` etc.

What do you mean? I assume you're talking about `prefer-const`, but that is checked in `npm test`.

Zarel:
> > I'd like for this to be skipped when testing, like we currently do for `const` etc.
> 
> What do you mean? I assume you're talking about `prefer-const`, but that is checked in `npm test`.

It's a warning in `npm test` (does not prevent unit tests from running) and an error in `npm run full-test`.

AnnikaCodes:
tsc errors don't stop unit tests from running in our current configuration

PR #8127, opened 2021-03-20
Gen 5 Random Battle updates

ACakeWearingAHat:
This patch updates Gen 5 Random Battle with many of the less controversial improvements proposed

- Superpower gone from the elemental monkeys bc setup + superp is not a fun time
- Swampert is now its usual supporty self
- fix the ability thing that apparently exists in every gen that makes equally rated abilities have unequal appearance rates
- focus blast nidos are a crime against nature
- things like wish chansey and flamethrower blissey just fully do not work in rands and it took way too long for anyone to realize this
- weak armor mandibuzz is bad
- intimidate staraptor is bad
- that one sceptile that should be unburden but isnt is gone
- quick feet is like, actually a thing again on mightyena sometimes
other misc move changes

### DIFF
--- a/data/mods/gen5/random-teams.ts
+++ b/data/mods/gen5/random-teams.ts
@@ -147,7 +147,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 				hasMove['lightscreen'] || hasMove['reflect']
 			)};
 		case 'uturn':
-			// Infernape doesn't want Expert Belt
+			// Infernape doesn't want mixed sets with U-turn
 			const infernapeCase = species.id === 'infernape' && counter.Special;
 			return {cull: counter.setupType || !!counter.speedsetup || hasMove['batonpass'] || infernapeCase};
 		case 'voltswitch':
--- a/data/mods/gen5/random-teams.ts
+++ b/data/mods/gen5/random-teams.ts
@@ -198,7 +198,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 		case 'glare': case 'headbutt':
 			return {cull: hasMove['bodyslam']};
 		case 'healbell':
-			return {cull: hasMove['magiccoat']};
+			return {cull: counter.speedsetup || hasMove['magiccoat']};
 		case 'moonlight': case 'painsplit': case 'recover': case 'roost': case 'softboiled': case 'synthesis':
 			return {cull: ['leechseed', 'rest', 'wish'].some(m => hasMove[m])};
 		case 'substitute':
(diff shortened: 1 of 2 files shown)


ACakeWearingAHat:
apologies i could've sworn i had renamed the commit ahead of time, as usual

data/mods/gen5/random-teams.ts
```diff
@@ -147,7 +147,9 @@ export class RandomGen5Teams extends RandomGen6Teams {
 				hasMove['lightscreen'] || hasMove['reflect']
 			)};
 		case 'uturn':
-			return {cull: counter.setupType || !!counter.speedsetup || hasMove['batonpass']};
+			// Infernape doesn't want Expert Belt
```
AnnikaCodes, line 150: If this hardcode is about Expert Belt, not about U-turn's usefulness on Infernape, I'd rather hardcode it in item generation.

data/mods/gen5/random-teams.ts
```diff
@@ -390,7 +392,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 			return 'Lustrous Orb';
 		}
 		if (counter.damagingMoves.length >= 4 && ability !== 'Sturdy') {
-			return (species.name === 'Deoxys-Attack' || counter.Normal || !!counter.priority) ? 'Life Orb' : 'Expert Belt';
+			return (hasMove['uturn']) ? 'Expert Belt' : 'Life Orb';
```
AnnikaCodes, line 395:
```suggestion
			return hasMove['uturn'] ? 'Expert Belt' : 'Life Orb';
```

data/mods/gen5/random-teams.ts
```diff
@@ -250,7 +252,7 @@ export class RandomGen5Teams extends RandomGen6Teams {
 		case 'Immunity':
 			return hasAbility['Toxic Boost'];
 		case 'Intimidate':
-			return hasMove['rest'];
+			return (hasMove['rest'] || species.id === 'staraptor');
```
AnnikaCodes, line 255:
```suggestion
			return hasMove['rest'] || species.id === 'staraptor';
```
(you don't need to wrap the entire `return`ed value in parentheses)

data/mods/gen5/random-teams.ts
AnnikaCodes: Is this just for the Infernape thing, or do U-turners generally want to avoid LO (because of hazard damage ig)?

data/mods/gen5/random-teams.ts
ACakeWearingAHat: infernape also doesn't want Life Orb U-turn, so the best way to prevent expert belt would be to prevent u-turn from generating in places that would otherwise give it Life Orb

AnnikaCodes: Why not add `species.id !== 'infernape'` to the `if (counter.damagingMoves.length >= 4 && ability !== 'Sturdy'))` check for ExpertBelt/LifeOrb? Or make the item that Infernape _does_ want higher priority?

ACakeWearingAHat: because then it'd get mixed u-turn, which it doesn't want

ACakeWearingAHat: Updated description to be more accurate to move issues

AnnikaCodes: Okay, that makes a lot more sense!

AnnikaCodes:
The unit tests are Angry(TM) about something in formats.ts; can you run `npx tsc` locally?

AnnikaCodes (approved):
I'm fine with this.

PR #8130, opened 2021-03-22
Don't allow setting modchat above one's rank

AnnikaCodes:
Discussed in the Upper Staff discord.

I've hardcoded this within the `/modchat` command, similar to the hardcode added in 92bc1bdb2d46ce26f84b993121a4f2385890c35e to prevent users above the current modchat rank from turning it off. As far as I know, modchat is the only permission that should work like this. If, in the future, we desire more commands that apply only to ranks at and below the user's rank, we could make this a formal permission, but that seems overkill for now.


Zarel:
Works for me, thanks!

PR #8131, opened 2021-03-22
Scavengers: Make !scav queue broadcastable

PartMan7:
Currently, the code is meant to call ``!scav viewqueue`` when no parameters are passed, but the runBroadcast line was missing in that bit. Tested to confirm that it broadcasts !scav queue, and that it doesn't broadcast while queuing a hunt with !scav queue [format].

RO confirmation from @aQrator

### DIFF
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -1691,7 +1691,7 @@ const ScavengerCommands: ChatCommands = {
 		}
 		if (!target && this.cmd !== 'queuerecycled') {
 			if (this.cmd === 'queue') {
-				if (!this.runBroadcast()) return false;
+				// Necessary for broadcasting: this.runBroadcast();
 				const commandHandler = ScavengerCommands.viewqueue as ChatHandler;
 				commandHandler.call(this, target, room, user, this.connection, this.cmd, this.message);
 				return;
--- a/server/chat-plugins/scavengers.ts
+++ b/server/chat-plugins/scavengers.ts
@@ -1691,6 +1691,7 @@ const ScavengerCommands: ChatCommands = {
 		}
 		if (!target && this.cmd !== 'queuerecycled') {
 			if (this.cmd === 'queue') {
+				if (!this.runBroadcast()) return false;
 				const commandHandler = ScavengerCommands.viewqueue as ChatHandler;
 				commandHandler.call(this, target, room, user, this.connection, this.cmd, this.message);
 				return;


aQrator:
RO confirmation from me

xfix:
I'm somewhat suspicious about necessity of doing so considering `viewqueue` command itself is broadcastable, and `/scav queue` simply calls the entire function.

xfix:
Oh, I see what's going on, the broadcast check is very hacky.

```javascript
entry.broadcastable = cmd.endsWith('help') || /\bthis\.(?:(check|can|run)Broadcast)\(/.test(handlerCode);
```

@Zarel, do you have any particular opinion about this? `/scav queue` calls `viewqueue` command internally, and is not recognized as broadcastable because `queue`'s code doesn't match the regex.

server/chat-plugins/scavengers.ts
```diff
@@ -1691,6 +1691,7 @@ const ScavengerCommands: ChatCommands = {
 		}
 		if (!target && this.cmd !== 'queuerecycled') {
 			if (this.cmd === 'queue') {
+				if (!this.runBroadcast()) return false;
 				const commandHandler = ScavengerCommands.viewqueue as ChatHandler;
```
mia-pi-git, line 1695: Is there a reason this doesn't use `this.parse`?
It feels like a better solution would be for it to use that, and to make the regex follow parse calls.
Covers a more general use-case.

AnnikaCodes:
In the past I've just added a comment `// Necessary for broadcasting: this.runBroadcast(`; personally, I'm in favor of adding an option to have annotated commands as an alternative to running regexes over the source code to determine command metadata. 

Although, for this particular usecase, Mia's suggestion might be more in line with PS's current design.

mia-pi-git:
> personally, I'm in favor of adding an option to have annotated commands as an alternative to running regexes over the source code to determine command metadata.

I do wonder, what's the benefit of this when you can just add ``runBroadcast()``? Both accomplish the same thing, and would be adding one line of code vs one line of code.

server/chat-plugins/scavengers.ts
```diff
@@ -1691,6 +1691,7 @@ const ScavengerCommands: ChatCommands = {
 		}
 		if (!target && this.cmd !== 'queuerecycled') {
 			if (this.cmd === 'queue') {
+				if (!this.runBroadcast()) return false;
```
AnnikaCodes, line 1694: Also, `runBroadcast()` throws if there's an problem, instead of returning, so you don't need to `return false;`.

AnnikaCodes:
> > personally, I'm in favor of adding an option to have annotated commands as an alternative to running regexes over the source code to determine command metadata.
> 
> I do wonder, what's the benefit of this when you can just add `runBroadcast()`? Both accomplish the same thing, and would be adding one line of code vs one line of code.

Because the code doesn't need to execute `runBroadcast()`, and doing so could be negative (especially if you need to broadcast without going through checkBroadcast, or if you check broadcasting yourself). But more generally, we shouldn't be running code that we don't need to run.

Zarel:
The comment is a good workaround for now. I don't think a better workaround is urgent. I'll think about it, though.

Zarel:
Thanks!

PR #8133, opened 2021-03-22
Hangman: Support tagging premade phrases

mia-pi-git:
https://www.smogon.com/forums/threads/hangman-groups-tags.3680445/#post-8793303

### DIFF
--- a/server/chat-plugins/hangman.ts
+++ b/server/chat-plugins/hangman.ts
@@ -442,7 +442,7 @@ export const commands: ChatCommands = {
 			this.checkCan('mute', null, room);
 			let [term, ...tags] = target.split(',');
 			term = toID(term);
-			tags = tags.map(i => i.trim()).filter(Boolean);
+			tags = tags.map(i => toID(i)).filter(Boolean);
 			if (!term || !tags || !tags.length) {
 				return this.parse('/help hangman');
 			}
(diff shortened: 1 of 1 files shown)


server/chat-plugins/hangman.ts
```diff
@@ -8,10 +8,30 @@ const HANGMAN_FILE = 'config/chat-plugins/hangman.json';
...
+	tags?: string[];
+}
+
+export let hangmanData: {[roomid: string]: {[phrase: string]: HangmanEntry}} = {};
 
 try {
 	hangmanData = JSON.parse(FS(HANGMAN_FILE).readSync());
+	const roomids = Object.keys(hangmanData);
+	let save = false;
+	for (const roomid of roomids) {
```
Zarel, line 22:
```suggestion
	let save = false;
	for (const roomid in hangmanData) {
```

Zarel:
Should tags be IDs? That might protect against certain mistakes.

Zarel:
You're certainly using `toID` to _compare_ them. It seems to me they should be stored as IDs so people know what's going on.

mia-pi-git:
> You're certainly using `toID` to _compare_ them. It seems to me they should be stored as IDs so people know what's going on.

I was thinking it might be useful to have them not be IDs, for visual purposes, but this is also a good point. I'll convert it in a moment.

mia-pi-git:
Done.

Zarel:
> I was thinking it might be useful to have them not be IDs, for visual purposes, but this is also a good point. I'll convert it in a moment.

Yeah, I guessed it might be something like that. But for instance Twitter hashtags are ID-like. I think it's probably good for tags to be IDs so people have a visual representation of how they work.

server/chat-plugins/hangman.ts
Zarel: .

PR #8134, opened 2021-03-23
Gen 3/4: Remove HAs in `getSpecies` instead of scripts files

KrisXV:


Zarel:
Why?

I think the current approach has some nicenesses, such as that iterating `Dex.mod('gen4').data.Pokedex` is more accurate.

KrisXV:
You’re mutating immutable data as it currently stands l; if you want niceness in iterating values of Dex.mod('gen4').data.Pokedex then hardcode the changes in mods/gen4/pokedex.ts imo

Zarel:
Honestly that's fair, it should be safe here but we probably can't rely on that.

Zarel:
By which I mean, it _is_ a bit easy to accidentally do something unsafe.

Rezzo64:
This commit is currently causing an error on building the client: 

```
Building `data/teambuilder-tables.js`... /home/sean/Documents/Projects/pokemon-showdown-client/caches/pokemon-showdown/dist/data/mods/gen4/scripts.js:29
      delete this.modData("Pokedex", i).abilities["H"];
                                       ^

TypeError: Cannot delete property 'H' of #<Object>
```

PR #8138, opened 2021-03-25
Refactor scripts to battle-actions

Zarel:
This introduces a new class, BattleActions, available as `battle.actions`, for most of the functions that were previously in `data/scripts.ts`.

This makes it so that "go to definition" will now work correctly for functions previously in scripts; we no longer need UnimplementedError, and there's now a clean conceptual separation between `battle` and `battle-actions` (whereas the previous distinction between `battle` and `scripts` was basically nonexistent).

Migration Guide
----------------

- Most of the contents of `data/scripts.ts` have been moved to `sim/battle-actions.ts`

- If you have a mod with a `scripts` file, you will need to move most of its contents into an `actions` sub-object. For instance, where you previously had:
  ```ts
  getDamage() {},
  modifyDamage() {},
  ```
  this should now be:
  ```ts
  actions: {
    getDamage() {},
    modifyDamage() {},
  },
  ```

- Because the list is so huge, I don't want to individually list every function. Instead, just check the individual files: Any function defined in `battle-actions.ts` should be in the `actions` sub-object, while any function still in `battle.ts` should stay where it currently is. TypeScript will also guide you on this, if you're using it.

- The functions in the `actions` sub-object now have `BattleActions` as `this`, rather than `Battle`. Mostly, this means that what was previously `this` is now `this.battle`, except that `this.battle.dex` can be shortened to `this.dex`, and `this.battle.actions` can be shortened to `this`.

- Using those functions will also require `.actions`. So, for instance, `battle.runMove` is now `battle.actions.runMove`.

- `battle.isAdjacent(pokemon1, pokemon2)` has been renamed `pokemon1.isAdjacent(pokemon2)`.

- `pokemon.nearbyAllies()` has been renamed `pokemon.adjacentAllies()`.

- `pokemon.nearbyFoes()` has been renamed `pokemon.adjacentFoes()`.

- `battle.getAtLoc(pokemon, targetLoc)` has been renamed `pokemon.getAtLoc(targetLoc)`


Zarel:
Merge conflicts are, of course, a huge problem here. :( I'll do one merge after each review, probably.

sim/battle-actions.ts
```diff
@@ -0,0 +1,1620 @@
...
+};
+
+export class BattleActions {
+	battle: Battle;
+	dex: ModdedDex;
+	constructor(battle: Battle) {
+		this.battle = battle;
+		this.dex = battle.dex;
+		if (this.dex.data.Scripts.actions) Object.assign(this, this.dex.data.Scripts.actions);
+	}
```
urkerab, line 55: (For OMs)
```suggestion
	constructor(battle: Battle) {
		this.battle = battle;
		this.dex = battle.dex;
		if (this.dex.data.Scripts.actions) Object.assign(this, this.dex.data.Scripts.actions);
		if (battle.format.actions) Object.assign(this, battle.format.actions);
	}
```
(Unfortunately you can't use the pattern that other classes use because all gens have battle actions.)

data/mods/gen1/scripts.ts
```diff
@@ -68,531 +58,742 @@ export const Scripts: ModdedBattleScriptsData = {
...
+				attack = attacker.getStat(atkType, true);
+			}
+			if (move.ignoreDefensive) {
+				this.battle.debug('Negating (sp)def boost/penalty.');
+				// No screens
+				defense = target.getStat(defType, true);
+			}
+
+			// When either attack or defense are higher than 256, they are both divided by 4 and moded by 256.
+			// This is what cuases the roll over bugs.
```
AnnikaCodes, line 730:
```suggestion
			// This is what causes the roll over bugs.
```
Might as well fix this typo while you're changing the file anyway.

Zarel: I mean, "moded" is like "plused" – it's bad grammar but it might be shorter/easier for people to understand than "and then take the remainder modulo 256".

data/mods/gen1/scripts.ts
```diff
@@ -68,531 +58,742 @@ export const Scripts: ModdedBattleScriptsData = {
...
+				attack = attacker.getStat(atkType, true);
+			}
+			if (move.ignoreDefensive) {
+				this.battle.debug('Negating (sp)def boost/penalty.');
+				// No screens
+				defense = target.getStat(defType, true);
+			}
+
+			// When either attack or defense are higher than 256, they are both divided by 4 and moded by 256.
+			// This is what causes the roll over bugs.
```
Zarel, line 730:
```suggestion
			// When either attack or defense are higher than 256, both are divided by 4.
			// If that's still over 256, it rolls over (%256), which is what causes rollover bugs.
```

data/mods/gen1/scripts.ts
urkerab: At least in some dictionaries, the verb "to plus" exists, so "plused" isn't bad grammar. However the existing verb "to mod" used as an abbreviation of either "moderate" or "modify" has the simple past tense "modded" rather than "moded" (which would have been the simple past tense of "to mode"). While you could have used "reduced mod 256", I like the new comment anyway.

PR #8139, opened 2021-03-25
Stadium: Apply BRN Atk drop once instead of 5x

Zarel:
@PlagueVonKarma confirmed that burns are supposed to drop Atk to 50% and not 3%, but just in case I'm missing something obvious, we'll have a pull request about this.

### DIFF
--- a/data/mods/gen1stadium/scripts.ts
+++ b/data/mods/gen1stadium/scripts.ts
@@ -26,8 +26,8 @@ export const Scripts: ModdedBattleScriptsData = {
 				this.baseStoredStats[statName] = this.storedStats[statName] = Math.floor(stat);
 				this.modifiedStats![statName] = Math.floor(stat);
 				// Re-apply drops, if necessary.
-				if (this.status === 'par') this.modifyStat!('spe', 0.25);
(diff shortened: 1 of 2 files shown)


data/mods/gen1stadium/scripts.ts
```diff
@@ -26,8 +26,8 @@ export const Scripts: ModdedBattleScriptsData = {
 				this.baseStoredStats[statName] = this.storedStats[statName] = Math.floor(stat);
 				this.modifiedStats![statName] = Math.floor(stat);
 				// Re-apply drops, if necessary.
-				if (this.status === 'par') this.modifyStat!('spe', 0.25);
-				if (this.status === 'brn') this.modifyStat!('atk', 0.5);
+				if (this.status === 'par' && statName === 'spe') this.modifyStat!('spe', 0.25);
+				if (this.status === 'brn' && statName === 'atk') this.modifyStat!('atk', 0.5);
```
Zarel, line 30: Both of these are unintentional, but Paralysis was accidentally working correctly, thanks to a quirk where Speed is the last stat, and `modifiedStats` is reset right before the second-to-last quartering. Still, this code is much more readably correct.

PR #8140, opened 2021-03-25
Add "m4a" as an alias for "Megas for all"

DayanYT:
Tested locally, works


DayanYT:
Wait, why did it try to patch the calculator again

PR #8142, opened 2021-03-28
Tag Mythical and Legendary Pokemon

xInstruct:
This is sorta just a big WIP atm as I try and feel stuff out, so mind the clutter.

Also has an example validator rule as a Proof-of-concept.

~~Am I in over my head? Probably.~~


DaWoblefet:
I think we decided on "Restricted" or "Restricted Legendary" > purely "Legendary". While legendary may more accurately represent the block of Pokemon internally, for our purposes it would provide additional clarity for what the tag is planned to be used for.

xInstruct:
Yeah true, I should probably tag alt formes of restricted legendary pokémon too, probably.

xInstruct:
also hi @Zarel

sim/dex-species.ts
```diff
@@ -152,6 +152,10 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 	readonly heightm: number;
 	/** Color. */
 	readonly color: string;
+	/** Tags, miscellaneous data.
+	* This just checks if a Pokemon has some sort of Legendary or Mythical status right now though.
+	*/
+	readonly tags?: string[];
```
Zarel, line 158: I would fill this out (`this.tags = data.tags || []`) so you don't need the `?.`

Zarel: I would also make it `SpeciesTag[]`

And then have `type SpeciesTag = "Mythical" | "Restricted Legendary";`

xInstruct:
Hey would it be possible to also tag Sub-Legendaries as well?

People who wanna do absolutely no legends at all could use this as a simple way to ban them.

DaWoblefet:
> Hey would it be possible to also tag Sub-Legendaries as well?
> 
> People who wanna do absolutely no legends at all could use this as a simple way to ban them.

Good idea!

sim/dex-species.ts
xInstruct: I've done this. But also added a Sub-Legendary alternative in keeping w/ data

(reason stated below)

xInstruct:
I'm sorry, I lost any and all motivation to complete this. Someone else can finish what I started.

xInstruct:
Ugh, I'm really sorry. The context behind this was mostly just an exorbitant amount of depression. I'm going back to work on this more in say a week or so after some much needed vacationing.

Zarel:
Fair enough. I'm glad you're feeling better!

xInstruct:
All tags have their respective validation coded in, among other things. The proof of concept alt gbu does work too!!!

darrenswhite:
@xInstruct would it be worth tagging all forms instead of just the base forms?

xInstruct:
> @xInstruct would it be worth tagging all forms instead of just the base forms?

Good question, my answer is that I just didn't want to in the begging since that'd an exponential amount of work especially for mons like arceus and silvally

DaWoblefet:
It should be unnecessary as well pretty sure.

xInstruct:
> It should be unnecessary as well pretty sure.

ye because im grabbing the basespecies which will save a lot of work in the long run

sim/team-validator.ts
```diff
@@ -1286,6 +1286,10 @@ export class TeamValidator {
 			}
 		}
 
+		if (dex.species.get(species.baseSpecies).tags?.length > 0) {
```
KrisXV, line 1289:
```suggestion
		if (dex.species.get(species.baseSpecies).tags?.length) {
```
0 is falsy so if the length === 0 the if statement wont go through

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
 			if (banReason === '') return null;
 		}
 
+		if (dex.species.get(species.baseSpecies).tags?.length > 0) {
```
KrisXV, line 1381: ditto

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
...
+			if (dex.species.get(species.baseSpecies).tags[0] === "Mythical") {
```
KrisXV, line 1382:
```suggestion
			if (dex.species.get(species.baseSpecies).tags.includes("Mythical")) {
```

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
...
+				banReason = ruleTable.check('pokemontag:mythical');
+				if (banReason) {
+					return `${species.name} is a Mythical Pokemon, which is ${banReason}.`;
+				}
+				if (banReason === '') return null;
+			} else if (dex.species.get(species.baseSpecies).tags[0] === "Restricted Legendary") {
```
KrisXV, line 1388: ditto

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
...
+				banReason = ruleTable.check('pokemontag:restrictedlegendary');
+				if (banReason) {
+					return `${species.name} is a Restricted Legendary Pokemon, which is ${banReason}.`;
+				}
+				if (banReason === '') return null;
+			} else if (dex.species.get(species.baseSpecies).tags[0] === "Sub-Legendary") {
```
KrisXV, line 1394: ditto

darrenswhite:
> > It should be unnecessary as well pretty sure.
> 
> ye because im grabbing the basespecies which will save a lot of work in the long run

Hmm makes sense to reduce redundant data.

What happens if the alt form has different tags? They will become out of sync so you would either need to check both sets of tags or include the base species tags in the alt form tags.

darrenswhite:
> > What happens if the alt form has different tags? They will become out of sync so you would either need to check both sets of tags or include the base species tags in the alt form tags.
> 
> I hope game freak doesnt pull that but the baseform is used everywhere with legendary validation so the alt form tag currently gets ignored

With the legendary validation yes, but this could be applied to all species (in the future) as the `SpeciesTag` type could be expanded to include new tags.

Plus it would make it consistent with the other fields. For example, both `Giratina` forms have the same `types` and are set in both forms.

xInstruct:
> With the legendary validation yes, but this could be applied to all species (in the future) as the `SpeciesTag` type could be expanded to include new tags.
> 
> Plus it would make it consistent with the other fields. For example, both `Giratina` forms have the same `types` and are set in both forms.

If that ever happens it'll be better to refactor to all forms, but for now this'll work for what we're trying to do rn

xInstruct:
tentatively putting it out of draft

sim/team-validator.ts
```diff
@@ -1286,6 +1286,10 @@ export class TeamValidator {
 			}
 		}
 
+		if (dex.species.get(species.baseSpecies).tags?.length) {
```
darrenswhite, line 1289: No need for `?` here as tags is always defined

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
 			if (banReason === '') return null;
 		}
 
+		if (dex.species.get(species.baseSpecies).tags?.length) {
```
darrenswhite, line 1381: Same as above

sim/team-validator.ts
```diff
@@ -1374,6 +1378,28 @@ export class TeamValidator {
...
+			if (dex.species.get(species.baseSpecies).tags.includes("Mythical")) {
```
darrenswhite, line 1382: nit: you could extract tags into a const to save getting it multiple times.

darrenswhite (changes requested):

sim/dex-species.ts
```diff
@@ -162,6 +164,10 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 	readonly heightm: number;
 	/** Color. */
 	readonly color: string;
+	/** Tags, miscellaneous data.
+	* This just checks if a Pokemon has some sort of Legendary or Mythical status right now though.
+	*/
```
Zarel, line 169:
```suggestion
	 * This just checks if a Pokemon has some sort of Legendary or Mythical status right now though.
	 */
```

Zarel: on second thought nvm we need more changes than this anyway.

Zarel, line 169:
```suggestion
	/**
	 * Tags, boolean data. Currently just legendary/mythical status.
	 */
```

sim/team-validator.ts
xInstruct: Sorry I have a habit of including redundant failsafes

sim/team-validator.ts
xInstruct: we could add it in line 1290

sim/team-validator.ts
Zarel: (You forgot to fix this...)

```suggestion
		if (dex.species.get(species.baseSpecies).tags.length) {
```

Zarel: Also you should really search the `tierSpecies` tags list.

xInstruct:
Sorry, I completely forgot about this PR. I will work on it ASAP.

PR #8143, opened 2021-03-28
SSB: Small tweaks

xInstruct:
Boat's move should bypass mbounce. Perish gets meme text.

### DIFF
--- a/data/mods/ssb/moves.ts
+++ b/data/mods/ssb/moves.ts
@@ -3424,6 +3424,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			this.add('-anim', source, "Trick", target);
 		},
 		onHit(target, source, effect) {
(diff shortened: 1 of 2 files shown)


data/mods/ssb/moves.ts
```diff
@@ -3424,6 +3424,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			this.add('-anim', source, "Trick", target);
 		},
 		onHit(target, source, effect) {
+			this.add(`c|${getName('Perish Song')}|/html <img src="https://i.imgflip.com/3rt1d8.png" />`);
```
AnnikaCodes, line 3427: This should probably include width and height attributes?

xInstruct: it doesnt particularly need it

xInstruct: yeah this is fine as is tbh

PR #8145, opened 2021-03-29
Allow berries to be eaten when a Pokémon with Unnerve switches out to another Pokémon with Unnerve

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-18659708

The fix was also added to both "as one" abilities.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -187,7 +187,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.effectData.unnerved = false;
 		},
 		onFoeTryEatItem() {
-			return !this.effectData.unnerved
+			return !this.effectData.unnerved;
 		},
 		onSourceAfterFaint(length, target, source, effect) {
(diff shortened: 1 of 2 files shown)


data/abilities.ts
```diff
@@ -182,7 +182,15 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.add('-ability', pokemon, 'As One');
 			this.add('-ability', pokemon, 'Unnerve', pokemon.side.foe);
 		},
-		onFoeTryEatItem: false,
+		onEnd() {
+			this.effectData.unnerved = false;
+		},
+		onFoeTryEatItem() {
+			if (!this.effectData.unnerved) {
```
DaWoblefet, line 189: These can be simplified to something like ``return !this.effectData.unnerved``, right?

AnnikaCodes: yes.

Marty-D (approved):
Looks good to me, nicely done! 👍

PR #8150, opened 2021-03-31
Doubles: Fix Aromatherapy/Sap Sipper interaction 

pacmanboss256:
aromatherapy now boosts allies with sap sipper in double battles

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -504,8 +504,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 			this.add('-activate', source, 'move: Aromatherapy');
 			let success = false;
 			for (const ally of pokemon.side.pokemon) {
-				if (ally !== source && ((ally.hasAbility('sapsipper')) ||
-						(ally.volatiles['substitute'] && !move.infiltrates))) {
+				if (ally !== source && (ally.hasAbility('sapsipper'))) {
+					ally.boostBy({atk: 1});
+					continue;
+				}
+				if (ally !== source && ((ally.volatiles['substitute'] && !move.infiltrates))) {
 					continue;
 				}
 				if (ally.cureStatus()) success = true;
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -504,7 +504,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 			this.add('-activate', source, 'move: Aromatherapy');
 			let success = false;
 			for (const ally of pokemon.side.pokemon) {
-				if (ally !== source && (ally.hasAbility('sapsipper'))) {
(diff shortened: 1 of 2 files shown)


data/moves.ts
```diff
@@ -504,8 +504,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 			this.add('-activate', source, 'move: Aromatherapy');
 			let success = false;
 			for (const ally of pokemon.side.pokemon) {
-				if (ally !== source && ((ally.hasAbility('sapsipper')) ||
-						(ally.volatiles['substitute'] && !move.infiltrates))) {
+				if (ally !== source && (ally.hasAbility('sapsipper'))) {
```
AnnikaCodes, line 507:
```suggestion
				if (ally !== source && ally.hasAbility('sapsipper')) {
```

data/moves.ts
```diff
@@ -504,8 +504,11 @@ export const Moves: {[moveid: string]: MoveData} = {
...
+					ally.boostBy({atk: 1});
+					continue;
+				}
+				if (ally !== source && ((ally.volatiles['substitute'] && !move.infiltrates))) {
```
AnnikaCodes, line 511:
```suggestion
				if (ally !== source && ally.volatiles['substitute'] && !move.infiltrates) {
```
`&&` is associative

urkerab:
> aromatherapy now boosts allies with sap sipper in double battles

When you say "now", in which game did the behaviour change?

pacmanboss256:
should be gen 8

Marty-D:
Wasn't the bug reported that this interaction only works sometimes? This change seems like it'll make it work all the time and sometimes twice.

pacmanboss256:
The bug report just says it doesn't work, nothing about partially working

DaWoblefet:
There is a skipped test for this behavior in tests/sim/abilities/sapsipper.js. Make sure to unskip to check if it passes that.

I am pretty sure the bug is slot-dependent.

pacmanboss256:
i ran the test and it passed

Marty-D:
So, yes
```
Turn 1

Goodra used Sleep Talk!
But it failed!

The opposing Goodra used Sleep Talk!
But it failed!

Blissey used Aromatherapy!
[Goodra's Sap Sipper]
Goodra's Attack rose!
A soothing aroma wafted through the area!
But it failed!
[DEBUG] move failed because it did nothing.

The opposing Blissey used Aromatherapy!
A soothing aroma wafted through the area!
But it failed!
[DEBUG] move failed because it did nothing.
```
The first Blissey that moved was to the right of its ally Goodra, which is why it worked.

PR #8154, opened 2021-04-01
Fix Adrenaline Orb's interaction with Contrary and maxed boosts, and fix Adrenaline Orb -5 test

pacmanboss256:
Adrenaline orb will no longer activate on intimidate if the holder is at -6 attack, +6 speed, or has the ability contrary while being at +6 attack


data/items.ts
```diff
@@ -67,6 +67,9 @@ export const Items: {[itemid: string]: ItemData} = {
 			basePower: 30,
 		},
 		onAfterBoost(boost, target, source, effect) {
+			if ((target.boosts['atk'] === -6) || (target.boosts['spe'] === 6)) return;
+			// check contrary users
+			if (target.hasAbility('Contrary') && (target.boosts['atk'] === 6)) return;
```
urkerab, line 72: I don't think this is right because `target` has already been bosted at this point so this fails when the target was boosted to 6. Try checking `boost` to see whether there was actually a stat change inflicted.

pacmanboss256: It passes all 7 tests

pacmanboss256:
bump

data/items.ts
Marty-D: I'm pretty sure urkerab meant check what `boost` was in this function as part of these checks, not during the boosts.

Marty-D (changes requested):
The checks in onBoost seem to be unnecessary if you just fix your initial idea.

pacmanboss256 (commented):
added the fix

test/sim/items/adrenalineorb.js
```diff
@@ -60,7 +60,23 @@ describe('Adrenaline Orb', function () {
...
+			{species: "Incineroar", ability: 'intimidate', moves: ['sleeptalk']},
+		]]);
+
+		battle.makeChoices();
+		battle.makeChoices('move curse', 'move topsyturvy');
+		assert.statStage(battle.p1.active[0], 'spe', 1);
+		battle.makeChoices('move splash', 'switch 2');
+		assert.statStage(battle.p1.active[0], 'spe', 2);
+		assert.false.holdsItem(battle.p1.active[0]);
+	});
```
Zarel, line 77: I believe your code performs incorrectly in this case. @Marty-D, can you confirm?

pacmanboss256:
bump

pacmanboss256 (commented):
Marty requested changes but those aren't relevant now

Marty-D:
Your implementation still assumes some stuff:
- that Simple doesn't exist
- that the holder was holding the item and had `ModifyBoost` run on it at some point before the Intimidate happens; your use of `target.itemData.oldatk` is always going to be undefined on the first pass.

What you want is to check what the boost was right before it happens, store only that, and forget about checking that it wasn't +5/-5. Check that it _was_ +6/-6.

pacmanboss256:
what function do you recommend i replace onModifyBoost with?

Marty-D:
Probably onBoost?

data/items.ts
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import {Pokemon} from "../sim";
```
Marty-D, line 2: What is this for?

data/items.ts
```diff
@@ -66,7 +69,16 @@ export const Items: {[itemid: string]: ItemData} = {
...
 			basePower: 30,
 		},
+		onBoostPriority: 1,
+		onBoost(boost, target) {
+			target.itemState.lastAtk = target.boosts['atk'];
+		},
 		onAfterBoost(boost, target, source, effect) {
+			const noAtkChange = boost.atk && boost.atk < 0 && target.boosts['atk'] === -6 && target.itemState.lastAtk === -6;
+			const noContraryAtkChange = boost.atk && boost.atk > 0 && target.boosts['atk'] === 6 && target.itemState.lastAtk === 6;
+			if (target.boosts['spe'] === 6 || noAtkChange || (target.hasAbility('contrary') && noContraryAtkChange)) {
```
Marty-D, line 79: No need to check for Contrary here, the only way Intimidate could have raised your Attack is if you had Contrary.

data/items.ts
pacmanboss256: es lint errors

data/items.ts
pacmanboss256: i just wanted to make sure defiant didnt mess anything up here

Marty-D: If you have Defiant then you couldn't have been at +6 before and after when the boost was positive.

data/items.ts
Marty-D: I mean, you aren't using it anywhere and had to disable eslint _because_ you aren't using it. Just remove it entirely?

Marty-D (approved):
Thanks!

PR #8159, opened 2021-04-02
Reremove accidentally readded removed line

ACakeWearingAHat:
(Reversion of https://github.com/smogon/pokemon-showdown/commit/23cc9519bb2bec73b02af7d5d7bec6f6404cba4d)

Regular Venusaur should not have a doubles set, as it has no corresponding level.


AnnikaCodes:
Oh, it looks like I accidentally added it while trying to add Doubles sets for CAPs. That's my bad

PR #8161, opened 2021-04-02
Tournaments: Allow regular users to check tour rules

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/allow-voiceds-and-regulars-to-check-tour-rules.3681028/

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -1556,6 +1556,9 @@ const commands: ChatCommands = {
 		customrules(target, room, user, connection, cmd) {
 			room = this.requireRoom();
 			const tournament = this.requireGame(Tournament);
+			if (cmd === 'banlist') {
+				return this.errorReply('The new syntax is: /tour rules -bannedthing, +un[banned|restricted]thing, *restrictedthing, !removedrule, addedrule');
+			}
 			if (!target) {
 				this.sendReply("Usage: /tour rules <list of rules>");
 				this.sendReply("Rules can be: -bannedthing, +un[banned|restricted]thing, *restrictedthing, !removedrule, addedrule");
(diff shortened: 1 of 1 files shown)


server/tournaments/index.ts
```diff
@@ -1555,17 +1555,17 @@ const commands: ChatCommands = {
...
 			if (!target) {
 				this.sendReply("Usage: /tour rules <list of rules>");
 				this.sendReply("Rules can be: -bannedthing, +un[banned|restricted]thing, *restrictedthing, !removedrule, addedrule");
 				this.parse('/tour viewrules');
 				return this.sendReplyBox(`<details><summary>Source</summary><code style="white-space: pre-wrap; display: table; tab-size: 3">/tour rules ${tournament.customRules}</code></details>`);
 			}
+			this.checkCan('tournaments', null, room);
+			if (cmd === 'banlist') {
+				return this.errorReply('The new syntax is: /tour rules -bannedthing, +un[banned|restricted]thing, *restrictedthing, !removedrule, addedrule');
+			}
```
AnnikaCodes, line 1568: Why is this moved? Do we want `/tour banlist` with no further arguments to show the rules? That seems fine to me, but it wasn't mentioned in the suggestion and I figured I'd check if that was intentional.

Distrib-ps: In the suggestion we ask that we can have the code.  So I put part of the code above the "checkCan" so that a user could use the command. A normal user cannot use the /tour banlist.

AnnikaCodes: The suggestion is to let regular users use `/tour rules` to see the rules. You make that possible, and you also change `/tour banlist`'s behavior.

Currently, it's deprecated, and just results in an errorReply. However, if you use `/tour banlist` with no arguments after your changes, the `if (!target)` block will run (since it's before the `checkCan` and the `if (cmd === 'banlist)`), and it will show the tournament's rules and return, without an error message describing the new syntax. I'm not opposed to that behavior, but I want to make sure that is intentional.

Distrib-ps: In this context, yes

Zarel:
Thanks!

PR #8164, opened 2021-04-03
Quotes: Support accessing a quote by index

HuntressRubyRose:
added the ability to display a quote by index

### DIFF
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -57,7 +57,7 @@ export const commands: ChatCommands = {
 		const {quote} = roomQuotes[index - 1];
 		return this.sendReplyBox(Chat.getReadmoreBlock(quote));
 	},
-	getquotehelp: [`/nquote [index] - Show a specifix quote from the room.`],
+	getquotehelp: [`/getquote [index] - Show a specifix quote from the room.`],
 
 	addquote: 'quote',
 	quote(target, room, user) {
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -49,15 +49,12 @@ export const commands: ChatCommands = {
 		const roomQuotes = quotes[room.roomid];
 		if (!roomQuotes) return this.errorReply(`No quotes found for ${room.roomid}.`);
 		const index = parseInt(target.trim());
-		if (isNaN(index)) {
+		if (isNaN(index) || !roomQuotes[index - 1]) {
 			return this.errorReply(`Invalid index.`);
 		}
-		if (!roomQuotes[index - 1]) {
-			return this.errorReply(`Out of bounds.`);
-		}
-		const {quote, date, userid} = roomQuotes[index - 1];
+		const {quote, date} = roomQuotes[index - 1];
 		const time = Chat.toTimestamp(new Date(date), {human: true});
-		return this.sendReplyBox(`${Chat.getReadmoreBlock(quote)}`);
+		return this.sendReplyBox(Chat.getReadmoreBlock(quote));
 	},
 	nquotehelp: [`/nquote [index] [showauthor] - Show a specifix quote from the room. Add 'showauthor' to see who added it and when.`],
 
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -48,6 +48,7 @@ export const commands: ChatCommands = {
 	nquote(target, room, user) {
 		room = this.requireRoom();
 		const roomQuotes = quotes[room.roomid];
+		if (!roomQuotes) return this.errorReply(`No quotes found for ${room.roomid}.`);
 		const index = parseInt(target.trim());
 		if (isNaN(index)) {
 			return this.errorReply(`Invalid index.`);
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -52,7 +52,7 @@ export const commands: ChatCommands = {
 		if (isNaN(index)) {
 			return this.errorReply(`Invalid index.`);
 		}
-		if(!roomQuotes[index - 1]) {
+		if (!roomQuotes[index - 1]) {
 			return this.errorReply(`Out of bounds.`);
 		}
 		this.runBroadcast(true);
--- a/server/chat-plugins/quotes.ts
+++ b/server/chat-plugins/quotes.ts
@@ -44,6 +44,23 @@ export const commands: ChatCommands = {
 		return this.sendReplyBox(`${Chat.getReadmoreBlock(quote)}${attribution}`);
 	},
 	randquotehelp: [`/randquote [showauthor] - Show a random quote from the room. Add 'showauthor' to see who added it and when.`],
+	
+	nquote(target, room, user) {
+		room = this.requireRoom();
+		const roomQuotes = quotes[room.roomid];
+		const index = parseInt(target.trim());
+		if (isNaN(index)) {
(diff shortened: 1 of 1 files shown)


mia-pi-git:
I believe I asked in Dev, but is there much of a practical use for this? They're all available on the quotes page.
Maybe post a suggestion and see if there's support?

server/chat-plugins/quotes.ts
```diff
@@ -44,6 +44,23 @@ export const commands: ChatCommands = {
 		return this.sendReplyBox(`${Chat.getReadmoreBlock(quote)}${attribution}`);
 	},
 	randquotehelp: [`/randquote [showauthor] - Show a random quote from the room. Add 'showauthor' to see who added it and when.`],
+	
+	nquote(target, room, user) {
+		room = this.requireRoom();
+		const roomQuotes = quotes[room.roomid];
```
mia-pi-git, line 50:
```suggestion
		const roomQuotes = quotes[room.roomid];
		if (!roomQuotes) return this.errorReply(`No quotes found for ${room.roomid}.`);
```

server/chat-plugins/quotes.ts
```diff
@@ -44,6 +44,23 @@ export const commands: ChatCommands = {
...
+		const index = parseInt(target.trim());
+		if (isNaN(index)) {
+			return this.errorReply(`Invalid index.`);
+		}
+		if(!roomQuotes[index - 1]) {
+			return this.errorReply(`Out of bounds.`);
+		}
+		this.runBroadcast(true);
```
mia-pi-git, line 58: Why bypass the broadcast timer? No need to show this more than once, no?

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
 		room = this.requireRoom();
 		const roomQuotes = quotes[room.roomid];
 		if (!roomQuotes?.length) return this.errorReply(`This room has no quotes.`);
-		this.runBroadcast(true);
```
HoeenCoder, line 40: Why make this command not broadcastable?

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
...
+		room = this.requireRoom();
+		const roomQuotes = quotes[room.roomid];
+		if (!roomQuotes) return this.errorReply(`No quotes found for ${room.roomid}.`);
+		const index = parseInt(target.trim());
+		if (isNaN(index)) {
+			return this.errorReply(`Invalid index.`);
+		}
+		if (!roomQuotes[index - 1]) {
+			return this.errorReply(`Out of bounds.`);
+		}
```
HoeenCoder, line 57:
```suggestion
		if (isNaN(index) || !roomQuotes[index - 1]) {
			return this.errorReply(`Invalid index.`);
		}
```
Logical Operators like `||` will short circuit if needed. This means that if the whole condition is 100% going to be true or false, the rest of the condition will not be checked. In this case if `index` is NaN, it won't attempt to check if there is a quote at `index - 1`. Only if index is not NaN will that be checked.

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
...
+		const {quote, date, userid} = roomQuotes[index - 1];
```
HoeenCoder, line 58:
```suggestion
		const {quote, date} = roomQuotes[index - 1];
```

If you don't want to use a value, you can just exclude it.

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
...
+		const time = Chat.toTimestamp(new Date(date), {human: true});
```
HoeenCoder, line 59: Remove this variable because its unused.

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
...
+		return this.sendReplyBox(`${Chat.getReadmoreBlock(quote)}`);
```
HoeenCoder, line 60:
```suggestion
		return this.sendReplyBox(Chat.getReadmoreBlock(quote));
```
Just cleaning this up a bit, no need to place this in a string if the result is the entire string.

server/chat-plugins/quotes.ts
```diff
@@ -37,13 +37,29 @@ export const commands: ChatCommands = {
...
+	},
+	nquotehelp: [`/nquote [index] [showauthor] - Show a specifix quote from the room. Add 'showauthor' to see who added it and when.`],
```
HoeenCoder, line 62: It doesn't appear you are supporting the `[showauthor]` parameter, this should either be updated or support added.

HoeenCoder (commented):
Hello! Thank you for submitting a pull request for this. We really appreciate that your willing to help us out. I have suggested some changes below, can you take a look when you have some time?

server/chat-plugins/quotes.ts
```diff
@@ -44,6 +44,19 @@ export const commands: ChatCommands = {
...
+		if (!roomQuotes) return this.errorReply(`No quotes found for ${room.roomid}.`);
+		const index = parseInt(target.trim());
+		if (isNaN(index) || !roomQuotes[index - 1]) {
+			return this.errorReply(`Invalid index.`);
+		}
+		this.runBroadcast(true);
+		const {quote} = roomQuotes[index - 1];
+		return this.sendReplyBox(Chat.getReadmoreBlock(quote));
+	},
+	getquotehelp: [`/getquote [index] - Show a specifix quote from the room.`],
```
AnnikaCodes, line 59:
```suggestion
	getquotehelp: [`/getquote [index] - Show a specific quote from the room.`],
```

HoeenCoder (approved):
Approving and not merging due to the question of if this should be implemented, I'm fine with it myself for broadcasting specific quotes.

server/chat-plugins/quotes.ts
AnnikaCodes:
```suggestion
		this.runBroadcast();
```
(This is how you avoid bypassing the broadcast timer.)

Zarel:
I think a big question is whether we preserve index numbers. For instance, if you delete a quote, do all the other quotes get renumbered? If so, we should probably not merge this.

mia-pi-git:
Relevant:
[13:06:00] @Mia: preserving old indeces would create probably bad behavior
[13:06:02] @Mia: yeah, same, PartMan
[13:06:09] @Mia: indices*
[13:06:13] %PartMan: D:
[13:06:16] @Mia: but i don't see the harm in them being reordered
[13:06:24] HuntressRubyRose: I don't either
[13:06:24] %PartMan: I've been running PartBot with dynamic indices for nearly two years now, haven't ever had any complaints.
[13:06:26] @Mia: but, like
[13:06:37] @Mia: at that point, we also lose any meaning for said command
[13:06:43] @Mia: imo
[13:06:46] @Mia: because you'd have to go into the page to find it
[13:06:49] @Mia: and at that point why not just c/p

Zarel:
So... I think it sounds like we're not going to do this. Unfortunate!

@HuntressRubyRose Don't feel discouraged! This isn't personal, and I think your code is great and the initiative you took to submit this is admirable! We just don't see a need for this feature.

If you would like to contribute more, you might want to talk about your ideas with us in the development Discord server: https://psim.us/devdiscord – we can help guide you about what sorts of contributions we'd probably accept or not.

(By the way, I'm also a fan of RWBY.)

PR #8171, opened 2021-04-03
Fix hazards not applying if the target side is fainted

HoeenCoder:
The 4 player battle changes broke hazards in general, they wont set if the target side has no un-fainted pokemon. This is because no target is provided, unlike in the past. The issue appears to be a combination of changes in two commits https://github.com/smogon/pokemon-showdown/commit/956daf89b6eb73f16846690d032fb7f77052046d which made `foeSide` moves start using an array of targets, and this from the 4-player patch which is used to get that array of targets but ignores fainted pokemon https://github.com/smogon/pokemon-showdown/commit/3b5e8cbfc2546afb19a0abce9f14c48c4e63e315#diff-6c26dba8e16b8854c67b0fc13f79e9811908c5f36d673084e350c8b89252ac81R230 .

This is my proposed patch, there may be a better way though (eg: new methods getAllFoes getAllAllies or w/e). PRing because I'm looking for thoughts.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -619,8 +619,8 @@ export class Pokemon {
 		return this.side.allies().filter(ally => this.isAdjacent(ally));
 	}
 
-	foes(): Pokemon[] {
-		return this.side.foes();
+	foes(all?: boolean): Pokemon[] {
+		return this.side.foes(all);
 	}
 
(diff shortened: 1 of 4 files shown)


DaWoblefet:
This implementation works in singles, but not in FFAs themselves (the battle appears to crash?).

Zarel:
Thanks!

PR #8172, opened 2021-04-03
Support searching multi formats

mia-pi-git:
Teammate ready isn't stored in the primary battleReady since i figured it'd be easier to access externally, as opposed to sorting. It also means we can set it _before_ we queue, as opposed to having to draw it apart and do some mess with validating in sync. (plus, if the search fails we can cache it while they wait, etc)


server/users.ts
```diff
@@ -45,6 +45,7 @@ const DEFAULT_TRAINER_SPRITES = [1, 2, 101, 102, 169, 170, 265, 266];
 
 import {FS, Utils, ProcessManager} from '../lib';
 import {Auth, GlobalAuth, PLAYER_SYMBOL, HOST_SYMBOL, RoomPermission, GlobalPermission} from './user-groups';
+import {BattleReady} from './ladders';
```
Zarel, line 48:
```suggestion
import type {BattleReady} from './ladders';
```

server/chat-commands/core.ts
```diff
@@ -1526,6 +1526,64 @@ export const commands: ChatCommands = {
...
+		let requestMap = Ladders.requests.get(targetUser.id);
+		if (!requestMap) {
+			requestMap = new Map();
+			Ladders.requests.set(targetUser.id, requestMap);
+		}
+		requestMap.set(user.id, format.id);
+		targetUser.send(`|requestmulti|${user.name}|${format.id}`);
+		this.connection.send(
+			`|pm|${user.getIdentity()}|${targetUser.getIdentity()}|/text` +
+			`You invited ${targetUser.name} to join you in ${format.name}.`
```
KrisXV, line 1554: Shouldn't there be a space either after `text` or before `You`?

KrisXV (commented):
That's all I see. @Zarel can you please take a look?

server/chat-commands/core.ts
mia-pi-git: yes. done.

KrisXV:
@mia-pi-git fix merge conflict imo 

bump @Zarel

mia-pi-git:
> fix merge conflict imo

It is done.

server/chat-commands/core.ts
```diff
@@ -1645,6 +1645,64 @@ export const commands: Chat.ChatCommands = {
 		}
 	},
 
+	requestpartner(target, room, user) {
+		target = this.splitTarget(target);
```
KrisXV, line 1649: id make this and any other commands that might use it use the new `splitUser` stuff

mia-pi-git: I'll do that when it's merged.

KrisXV: Why wait until after, especially if this might get merged first? The function already exists in the main code, so there's no reason to not do it here

KrisXV: Yeah, all my function does is rewrite the uses of the deprecated version

mia-pi-git: Ah. I thought your PR added it as well.

server/ladders.ts
```diff
@@ -79,7 +91,13 @@ class Ladder extends LadderStore {
 		super(formatid);
 	}
 
-	async prepBattle(connection: Connection, challengeType: ChallengeType, team: string | null = null, isRated = false) {
+	async prepBattle(
+		connection: Connection,
+		challengeType: ChallengeType,
+		team: string | null = null,
+		isRated = false,
+		noPartner = false
```
KrisXV, line 99:
```suggestion
		partner = false
```

server/ladders.ts
```diff
@@ -179,6 +197,13 @@ class Ladder extends LadderStore {
 			return null;
 		}
 
+		if (Dex.formats.get(this.formatid).gameType === 'multi' && !noPartner) {
```
KrisXV, line 200:
```suggestion
		if (Dex.formats.get(this.formatid).gameType === 'multi' && !partner) {
```
`!noPartner` would sound like it'd equate to `true` as in you do have a partner.

mia-pi-git:
This now uses the new challenge protocol.

mia-pi-git:
@Zarel? Please?

PR #8176, opened 2021-04-04
LC: Ban Woobat

Distrib-ps:
PR for https://www.smogon.com/forums/threads/ss-lc-suspect-process-woobat-voting.3681457/post-8808892


KrisXV:
You need to actually ban Woobat from the LC format as well

Zarel:
Shouldn't its tier be LC Uber and not NFE?

KrisXV:
no, LC Uber doesnt exist anymore

Zarel:
I don't like using the same tier name for middle evolutions and first-stages banned from LC.

Issue #8178, opened 2021-04-05
Towards a better Dex API


Zarel:
Our current Dex API revolves around:

```
Dex.getMove(moveName)
Dex.getAbility(abilityName)
Dex.getItem(itemName)
Dex.getSpecies(speciesName)
```

Everything else is sort of just haphazardly squeezed in there.

The most noticeably missing feature is the ability to iterate through a list of all moves, or all abilities. We currently achieve this with `for (const moveid in Dex.data.Moves)`, but this is not really intended to be an officially supported API, so much as a workaround. It also doesn't work in the client, which has a much older API `for const moveid in BattleMovedex)`.

I've been meaning to standardize iteration APIs. Some time ago, I polled developers on the two options:

1. `Dex.getMove(moveName)` and `Dex.allMoves()`
2. `Dex.moves.get(moveName)` and `Dex.moves.all()`

Most people preferred the latter, and it certainly makes the code easier, but I figured I'd post one more issue declaring my intent, and looking for any final objections, before actually moving forward with it.

There's still one fear, which is that it won't be _very_ consistent. We'll still have plenty of APIs which use `getWhatever()` functions. But that's probably pretty minor in the grand scheme of things.

Anyway. The plan is to rename `Dex.getMove` to `Dex.moves.get` (and so on for the others). Questions, comments?

Zarel:
A related question is that we probably want four functions: one for a loose match, one for an exact ID match (for performance), one for iterating legal entries, and one for iterating all entries.

The one I'd default to:

`Dex.moves.get(name)`, `Dex.moves.getByID(id)`, `Dex.moves.all()`, `Dex.moves.all(true)`

Other options include:

`Dex.moves.search(name)`, `Dex.moves.get(id)`, `Dex.moves.allLegal()`, `Dex.moves.all()`

`get(id)` is a weird one: we use plenty of Maps where `get` needs to be an ID, but in most of the rest of our code, `get` usually takes a name. But our API should probably be more Map-like.

Zarel:
You might ask: Can we implement it so `for (const move of Dex.moves)` works?

The answer is: yes, but for performance reasons, it wouldn't be usable on client. I would rather an API that works the same way on client and server than one that's slightly nicer but only works on the server.

[...]

PR #8181, opened 2021-04-06
Update Dex API

Zarel:
This is the change that renames:

- `Dex.getMove` -> `Dex.moves.get`
- `Dex.getAbility` -> `Dex.abilities.get`
- `Dex.getItem` -> `Dex.items.get`
- `Dex.getSpecies` -> `Dex.species.get`
- `Dex.getEffect` -> `Dex.conditions.get`
- `Dex.getNature` -> `Dex.natures.get`
- `Dex.getType` -> `Dex.types.get`
- `Dex.getFormat` -> `Dex.formats.get`

In addition, some other APIs have been updated:

- `getByID` methods have also been added to every other table.
- `Dex.moves.all()` now gets an array of all moves
  - Plus equivalent methods for `abilities`, `items`, `species`, `formats`, `natures`, `types`
  - Note: there's no `Dex.conditions.all()`
- new API: `Dex.stats` for naming/iterating stats
- `Dex.getEffectByID` -> `Dex.conditions.getByID`
- `Dex.getType` -> `Dex.types.get`
- `Dex.data.Formats` -> `Dex.data.Rulesets`
- `Dex.formats` -> now an array `Dex.formats.all()`
- `Dex.getRuleTable` -> `Dex.formats.getRuleTable`
- `Dex.validateFormat` -> `Dex.formats.validate`

Team functions have been split off into a new `sim/teams` package:

- `Dex.packTeam` -> `Teams.pack`
- `Dex.fastUnpackTeam` -> `Teams.unpack`
- `Dex.generateTeam` -> `Teams.generate`
- `Dex.stringifyTeam` -> `Teams.export`

`Teams.export` has also been rewritten to better match how it works in client.

This implements #8178


Zarel:
Planned:

- `Dex.moves.all()` and so on, for the tables.

- `Dex.formats` is currently an Object table, which will need to be converted if we want it to work like the other tables.

- Add a `Dex.stats` table.

I probably want to implement as many of these as possible ASAP, so they can all be done in one server restart.

data/FORMES.md
```diff
@@ -192,6 +192,6 @@ Dynamax and Gigantamax are not considered formes by the games themselves. PS imp
 `pokedex.js`
```
AnnikaCodes, line 192: We should probably change this to `.ts`?

AnnikaCodes: There are other references to `.js` files that have since been TypeScripted; should I just make a separate commit to fix it?

sim/dex-abilities.ts
```diff
@@ -50,3 +49,56 @@ export class Ability extends BasicEffect implements Readonly<BasicEffect> {
...
+				(ability as any).isNonstandard = null;
+			}
+		} else {
+			ability = new Ability({
+				id, name: id, exists: false,
+			});
+		}
+
+		if (ability.exists) this.abilityCache.set(id, ability);
+		return ability;
```
AnnikaCodes, line 102:
```suggestion
		if (this.dex.data.Aliases.hasOwnProperty(id)) {
			ability = this.get(this.dex.data.Aliases[id]);
		} else if (id && this.dex.data.Abilities.hasOwnProperty(id)) {
			const abilityData = this.dex.data.Abilities[id] as any;
			const abilityTextData = this.dex.getDescs('Abilities', id, abilityData);
			ability = new Ability({
				name: id,
				...abilityData,
				...abilityTextData,
			});
			if (ability.gen > this.dex.gen) {
				(ability as any).isNonstandard = 'Future';
			}
			if (this.dex.currentMod === 'letsgo' && ability.id !== 'noability') {
				(ability as any).isNonstandard = 'Past';
			}
			if ((this.dex.currentMod === 'letsgo' || this.dex.gen <= 2) && ability.id === 'noability') {
				(ability as any).isNonstandard = null;
			}
		} else {
			ability = new Ability({
				id, name: id, exists: false,
			});
		}

		if (ability.exists) this.abilityCache.set(id, ability);
		return ability;
```
Why mix else-if and returning from an `if` statement? This seems cleaner.

data/FORMES.md
Zarel: Probably, yeah.

sim/dex-abilities.ts
Zarel: Actually are the two alias checks intended to do different things?

AnnikaCodes: I only see the one alias check? There's the alias branch, the extant-ability branch, and the nonexistent-ability branch.

Zarel: Oh yeah I see that now.

PR #8182, opened 2021-04-06
Negate Unburden boost when Neutralizing Gas is active.

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-52059168

Simply adds checking ignoringAbility in the unburden condition code. Ran both Neutralizing Gas and Unburden tests and they all passed. This probably requires a change in the client tooltips code so unburden's speed boost is not shown when it is suppressed.


Marty-D (approved):
Great idea, thanks!

PR #8183, opened 2021-04-06
Fix autoconfirmed translations

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8811931


AnnikaCodes (approved):
Good catch!

PR #8185, opened 2021-04-07
Gen 1: Rename [Gen 1] OU (Tradeback) to [Gen 1] Tradebacks OU

PlagueVonKarma:
This is to coincide with my suggestion [here](https://www.smogon.com/forums/threads/rename-gen-1-ou-tradeback-to-gen-1-tradebacks-ou.3681565/).

I believe this name change allows for players to use team folders with Tradebacks, which reduces a lot of complication with RoA Room Tours and any potential spotlight ladders in the future. I _think_ this lets the Strategy Dex feed sample sets via import/export as well? I'm a boomer who doesn't know how stuff works...

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -3267,7 +3267,7 @@ export const Formats: FormatList = [
 		banlist: ['UU', 'NUBL'],
 	},
 	{
-		name: "[Gen 1] OU (Tradeback)",
+		name: "[Gen 1] Tradebacks OU",
 		desc: `RBY OU with movepool additions from the Time Capsule.`,
 		threads: [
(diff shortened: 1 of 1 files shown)


Zarel (approved):

Zarel:
I guess I'll give Annika and Marty and everyone else a dayish to decide if they dislike this. Maybe the Suggestions thread will have some hot takes or something by then.

AnnikaCodes:
I don't really care what the format is called; if people want Tradebacks OU, that's fine with me.

Marty-D (approved):

Zarel:
I think that's enough approvals. If anyone objects we can consider reverting it, but at this point I can't imagine it being an issue.

PR #8188, opened 2021-04-08
IPTools: Automatically add Tor IPs to the proxy list

mia-pi-git:
Discussed in PMs with Zarel.

### DIFF
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -222,6 +222,7 @@ export const IPTools = new class {
 	 */
 	ranges: (AddressRange & {host: string})[] = [];
 	singleIPOpenProxies: Set<string> = new Set();
+	torProxyIps = new Set<string>();
 	proxyHosts: Set<string> = new Set();
 	residentialHosts: Set<string> = new Set();
 	mobileHosts: Set<string> = new Set();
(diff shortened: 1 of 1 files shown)


Zarel:
They should probably be specially marked as Tor, so they can be removed if they no longer exist?

server/ip-tools.ts
```diff
@@ -222,6 +222,7 @@ export const IPTools = new class {
 	 */
 	ranges: (AddressRange & {host: string})[] = [];
 	singleIPOpenProxies: Set<string> = new Set();
```
Zarel, line 224:
```suggestion
	singleIPOpenProxies = new Set<string>();
```

server/ip-tools.ts
```diff
@@ -222,6 +222,7 @@ export const IPTools = new class {
...
+	torProxyIps = new Set<string>();
 	proxyHosts: Set<string> = new Set();
 	residentialHosts: Set<string> = new Set();
 	mobileHosts: Set<string> = new Set();
```
Zarel, line 228:
```suggestion
	proxyHosts = new Set<string>();
	residentialHosts = new Set<string>();
	mobileHosts = new Set<string>();
```

Zarel:
Thanks!

PR #8189, opened 2021-04-08
Admin: Add ram info to /processes

mia-pi-git:


server/chat-commands/admin.ts
```diff
@@ -18,6 +18,14 @@ interface ProcessData {
...
 	time?: string;
+	ram?: string;
+}
+
+function checkCanUse(context: CommandContext) {
+	const devRoom = Rooms.get('development');
+	if (!(devRoom && Users.Auth.atLeast(devRoom.auth.getDirect(context.user.id), '%'))) {
+		context.checkCan('lockdown');
+	}
 }
```
Zarel, line 29: This is a really awkward way to do it, moves the original `checkCan` outside of where we can see it, requires passing around contexts which we really shouldn't be doing, and is also named badly in a way that doesn't mention development at all.

Zarel:
```suggestion
function hasDevAuth(user: User) {
	const devRoom = Rooms.get('development');
	return devRoom && Users.Auth.atLeast(devRoom.auth.getDirect(user.id), '%');
}
```

server/chat-commands/admin.ts
```diff
@@ -742,24 +750,27 @@ export const commands: ChatCommands = {
 	],
 
 	async processes(target, room, user) {
-		const devRoom = Rooms.get('development');
-		if (!(devRoom && Users.Auth.atLeast(devRoom.auth.getDirect(user.id), '%'))) {
-			this.checkCan('lockdown');
-		}
+		checkCanUse(this);
```
Zarel, line 753:
```suggestion
		if (!hasDevAuth(user)) this.checkCan('lockdown');
```

server/chat-commands/admin.ts
```diff
@@ -408,7 +416,7 @@ export const commands: ChatCommands = {
 
 	memusage: 'memoryusage',
 	memoryusage(target) {
-		this.checkCan('lockdown');
+		checkCanUse(this);
```
Zarel, line 419:
```suggestion
		if (!hasDevAuth(user)) this.checkCan('lockdown');
```

server/chat-commands/admin.ts
```diff
@@ -408,7 +416,7 @@ export const commands: ChatCommands = {
 
 	memusage: 'memoryusage',
 	memoryusage(target) {
```
Zarel, line 418:
```suggestion
	memoryusage(target, room, user) {
```

Zarel:
Thanks!

PR #8191, opened 2021-04-10
Validator: Disallow duplicate moves

DaWoblefet:
Currently, when a user creates a Pokemon with multiple duplicate moves, the validator doesn't say anything is wrong and just removes the duplicated move from the user's moveset automatically. I've definitely made the mistake in the past where I quickly make a Pokemon's set, don't realize I really only have 3 moves, and don't figure it out until the battle. This also fixes an issue with automatic move changes (i.e. Iron Head -> Behemoth moves) where a user accidentally adds both Iron Head and the Behemoth move without realizing Iron Head becomes Behemoth Blade / Bash. This way, the mistake is exposed at the teambuilder level, so they don't think get confused about why a move is missing.

Sample:
image

Fixes https://github.com/smogon/pokemon-showdown/projects/3#card-51711810

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -265,7 +265,6 @@ export const Rulesets: {[k: string]: FormatData} = {
 					moves.push(moveId);
 				}
 			}
-			set.moves = moves;
 		},
 	},
 	hoennpokedex: {
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -254,7 +254,6 @@ export const Rulesets: {[k: string]: FormatData} = {
 			}
 
 			// limit one of each move
-			const moves = [];
 			if (set.moves) {
 				const hasMove: {[k: string]: true} = {};
 				for (const moveId of set.moves) {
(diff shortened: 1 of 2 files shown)


data/rulesets.ts
```diff
@@ -260,12 +260,11 @@ export const Rulesets: {[k: string]: FormatData} = {
 				for (const moveId of set.moves) {
 					const move = this.dex.moves.get(moveId);
 					const moveid = move.id;
-					if (hasMove[moveid]) continue;
+					if (hasMove[moveid]) return [species.baseSpecies + " has multiple copies of " + move.name + "."];
```
AnnikaCodes, line 263:
```suggestion
					if (hasMove[moveid]) return [`${species.baseSpecies} has multiple copies of ${move.name}.`];
```
Is there a reason you didn't use template strings here?

DaWoblefet: I forgot lol

data/rulesets.ts
```diff
@@ -260,12 +260,11 @@ export const Rulesets: {[k: string]: FormatData} = {
...
 					hasMove[moveid] = true;
 					moves.push(moveId);
```
AnnikaCodes, line 265: What's the point of having a `moves` variable? You're already iterating over `set.moves`, and you're no longer reassigning `set.moves`.

DaWoblefet: Another good catch!

Zarel:
Thanks!

PR #8192, opened 2021-04-10
Random Battles Improvements

ACakeWearingAHat:
A series of improvements for legibility, function, and code simplicity.

-Added comments to random-teams to indicate special cases and explain organization

-Organized many things

-Removed some unnecessary hardcodes

-Neutralizing Gas should really not be at the same rating as Moody, Shadow Tag, and Wonder Guard, especially when it necessitates a dumb hardcode for Weezings in rands. 4 is a much better fit by all logic.

-Separated the doubles Butterfree sets to actually accommodate for Quiver Dance being forced

-Nasty Plot PZ is free

-Speedsetup moves are now further clarified as a group, only referring to moves that *solely* boost Speed. This causes Extreme Speed + Shift Gear genesect to exist again and makes Thunder Wave + Shift Gear Klinklang happen in doubles, but those are acceptable enough to the point where I'd rather not make special cases for them.

-Dual Wingbeat is no longer forced on Dragon/Flying types. Outrage is, instead. Further clarification and simplification of recent update.

-Performed some minor level updates in No Dmax to fix some levels that were supposed to be initially adjusted at the format's release (Zacian and Dugtrio)

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -41,9 +41,9 @@ const SpecialSetup = [
 const MixedSetup = [
 	'clangoroussoul', 'growth', 'happyhour', 'holdhands', 'noretreat', 'shellsmash', 'workup',
 ];
-// Moves which boost Speed:
+// Moves which only boost Speed:
 const SpeedSetup = [
-	'agility', 'autotomize', 'flamecharge', 'rockpolish', 'shiftgear',
+	'agility', 'autotomize', 'flamecharge', 'rockpolish',
 ];
 // Moves that shouldn't be the only STAB moves:
 const NoStab = [
@@ -120,16 +120,12 @@ export class RandomTeams {
 				if (!counter.Dark) return true;
 				return hasMove['suckerpunch'] && (movePool.includes('knockoff') || movePool.includes('wickedblow'));
 			},
-			Dragon: (movePool, hasMove, hasAbility, hasType, counter, species) => {
-				if (movePool.includes('outrage') && species.baseStats.atk >= 115) return true;
-				return (
-					!(counter.setupType === 'Physical' && hasType['Flying']) &&
-					!counter.Dragon &&
-					!hasMove['dragonascent'] &&
-					!hasMove['substitute'] &&
-					!(hasMove['rest'] && hasMove['sleeptalk'])
-				);
-			},
+			Dragon: (movePool, hasMove, hasAbility, hasType, counter) => (
+				!counter.Dragon &&
+				!hasMove['dragonascent'] &&
+				!hasMove['substitute'] &&
+				!(hasMove['rest'] && hasMove['sleeptalk'])
+			),
 			Electric: (movePool, hasMove, hasAbility, hasType, counter) => !counter.Electric || movePool.includes('thunder'),
 			Fairy: (movePool, hasMove, hasAbility, hasType, counter) => (
 				!counter.Fairy &&
@@ -142,7 +138,7 @@ export class RandomTeams {
 				return !hasMove['bellydrum'] && (!counter.Fire || (!enteiException && movePool.includes('flareblitz')));
 			},
 			Flying: (movePool, hasMove, hasAbility, hasType, counter) => (
-				!counter.Flying && [
+				!counter.Flying && !hasType['Dragon'] && [
 					'airslash', 'bravebird', 'dualwingbeat', 'oblivionwing',
 				].some(moveid => movePool.includes(moveid))
 			),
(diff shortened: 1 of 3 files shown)


data/random-teams.ts
```diff
@@ -41,9 +41,9 @@ const SpecialSetup = [
 const MixedSetup = [
 	'clangoroussoul', 'growth', 'happyhour', 'holdhands', 'noretreat', 'shellsmash', 'workup',
 ];
-// Moves which boost Speed:
+// Moves which only boost Speed:
 const SpeedSetup = [
-	'agility', 'autotomize', 'flamecharge', 'rockpolish', 'shiftgear',
+	'agility', 'autotomize', 'flamecharge', 'rockpolish',
 ];
```
AnnikaCodes, line 47: If this actually means something now, and isn't just a hardcode, Aura Wheel and Rapid Spin should be included.

data/random-teams.ts
```diff
@@ -922,7 +915,14 @@ export class RandomTeams {
...
 			return {cull: counter.speedsetup || (counter.setupType && !bugSwordsDanceCase) || (isDoubles && hasMove['leechlife'])};
 
-		// Ineffective to have two particular moves together
+		/**
+		 * Ineffective to have both moves together
+		 *
+		 * These are sorted in order of:
+		 * Normal>Fire>Water>Electric>Grass>Ice>Fighting>Poison>Ground>Flying>Psychic>Bug>Rock>Ghost>Dragon>Dark>Fairy
+		 * and then subsorted alphabetically.
+		 * This type order is arbitrary and referenced from some sort of webpage. I do not know which webpage.
```
AnnikaCodes, line 924: This probably comes from [PokemonDB's typechart](https://pokemondb.net/type).

data/random-teams.ts
```diff
@@ -1308,19 +1307,25 @@ export class RandomTeams {
...
+				// For Doubles Alolan Persian
 				movePool.includes('snarl')
 			);
 		case 'Tinted Lens':
 			return (
+				// For Sigilyph
 				hasMove['defog'] ||
-				(species.id === 'butterfree') ||
+				// For Butterfree
+				(hasMove['hurricane'] && hasAbility['Compound Eyes']) ||
```
AnnikaCodes, line 1318: If this is "for Butterfree," why not hardcode it? Would you complain if Vivillon became legal and it got Compound Eyes with Hurricane? I personally think this is a good general cull, and would be fine with Compound Eyes Vivillon.

data/random-teams.ts
```diff
@@ -1308,19 +1307,25 @@ export class RandomTeams {
...
 				(counter.Status > 2 && !counter.setupType)
 			);
 		case 'Torrent':
+			// For Inteleon-Gmax and Primarina
 			return (hasMove['focusenergy'] || hasMove['hypervoice']);
```
AnnikaCodes, line 1323: I'd rather cull based on the abilities (Sniper and Liquid Voice) that are preferred to Torrent.
```suggestion
			return hasAbility['Sniper'] || hasAbility['Liquid Voice'];
```

data/random-teams.ts
```diff
@@ -1905,10 +1910,11 @@ export class RandomTeams {
 			};
 			const customScale: {[k: string]: number} = {
 				// These Pokemon are too strong and need a lower level
-				zaciancrowned: 66, calyrexshadow: 68, xerneas: 70, necrozmaduskmane: 72, glalie: 78, octillery: 84, wobbuffet: 86,
-				jolteon: 84, swoobat: 84, haxorus: 80, kyogre: 73, inteleon: 80, slurpuff: 84, polteageist: 84,
+				zaciancrowned: 66, calyrexshadow: 68, xerneas: 70, necrozmaduskmane: 72, zacian: 72, kyogre: 73, glalie: 78,
+				haxorus: 80, inteleon: 80, octillery: 84, jolteon: 84, swoobat: 84, dugtrio: 84, slurpuff: 84, polteageist: 84,
+				wobbuffet: 86,
```
AnnikaCodes, line 1915: Are you changing levels here, or just sorting them?

data/random-teams.ts
```diff
@@ -120,16 +120,12 @@ export class RandomTeams {
...
-				return (
-					!(counter.setupType === 'Physical' && hasType['Flying']) &&
-					!counter.Dragon &&
-					!hasMove['dragonascent'] &&
-					!hasMove['substitute'] &&
-					!(hasMove['rest'] && hasMove['sleeptalk'])
-				);
-			},
+			Dragon: (movePool, hasMove, hasAbility, hasType, counter) => (
+				!counter.Dragon &&
```
AnnikaCodes, line 124: I don't fully understand the change in logic here, but the unit test for forcing Outrage still passes, so I guess it works. It looks like you're not forcing Flying STAB on Flying/Dragon types?

data/random-teams.ts
ACakeWearingAHat: that'd be it!

data/random-teams.ts
ACakeWearingAHat: vivillon doesn't have tinted lens. this was changed in order to reflect that in doubles and only in doubles, both butterfree and gmax butterfree want compound eyes. it still fulfills the purpose it had with a species call in singles.

data/random-teams.ts
ACakeWearingAHat: that wouldn't work. Primarina only wants Liquid Voice in doubles, and even then only when it has Hyper Voice. Hyper Voice is the proper cull here.

Same goes for Sniper; Regular inteleons without Focus Energy don't want to have Torrent rejected. Only the ones *with* focus energy want sniper. The rest want Torrent.

It's a simple matter of how hasAbility only checks for *available* abililties and not anything else

data/random-teams.ts
ACakeWearingAHat: the changed levels are in the description. otherwise, sorted in ascending order.

data/random-teams.ts
ACakeWearingAHat: yes, check the description.

your fix last update just added a hardcode to negate another hardcode for the same thing. i just removed both of them and made it simpler.

data/random-teams.ts
ACakeWearingAHat: this would be fine as a general cull, though. i just noted what it was specifically for at this point in time

data/random-teams.ts
ACakeWearingAHat: this goes across all gens. rapid spin is excluded for that reason, since it'd mess with oldgens otherwise. plus, some mons want sr+spin even in gen 8. Rapid Spin doesnt "only boost Speed", it also removes hazards.

i'll include aura wheel though, shouldn't affect anything too much? i think

ACakeWearingAHat: never mind, adding aura wheel messes with morpeko's moveset a lot actually. parting shot is rejected with speedsetup

data/random-teams.ts
AnnikaCodes: ok

data/random-teams.ts
AnnikaCodes: Oh, do we prefer Torrent + Hydro Pump in singles? this is fine, then, I suppose.

data/random-teams.ts
AnnikaCodes: OK.

data/random-teams.ts
AnnikaCodes, line 924:
```suggestion
		 * This type order is arbitrary and referenced from https://pokemondb.net/type.
```

data/random-teams.ts
```diff
@@ -41,9 +41,9 @@ const SpecialSetup = [
 const MixedSetup = [
 	'clangoroussoul', 'growth', 'happyhour', 'holdhands', 'noretreat', 'shellsmash', 'workup',
 ];
-// Moves which boost Speed:
+// Moves which only boost Speed:
```
AnnikaCodes, line 44:
```suggestion
// Some moves that boost Speed:
```

AnnikaCodes:
Thanks!

PR #8196, opened 2021-04-12
Random Battles: Fix Toxtricity

ACakeWearingAHat:
happens when we remove stuff from speedsetup. as of last update shift gear's gotta be manually called whenever we want to do anything with it since it's not categorized as setupType or speedsetup now.

idea: define speedsetup as "all speed-boosting moves that are not otherwise categorized as setupType and would need to be called in the generator at any point specifically due to their speed boosting properties"

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -905,7 +905,7 @@ export class RandomTeams {
 		case 'taunt':
 			return {cull: hasMove['encore'] || hasMove['nastyplot'] || hasMove['swordsdance']};
 		case 'thunderwave': case 'voltswitch':
-			const cullInDoubles = hasMove['electroweb'] || hasMove['nuzzle'];
+			const cullInDoubles = isDoubles && (hasMove['electroweb'] || hasMove['nuzzle']);
 			return {cull: counter.setupType || counter.speedsetup || hasMove['shiftgear'] || hasMove['raindance'] || cullInDoubles};
 		case 'toxic':
 			return {cull: counter.setupType || ['sludgewave', 'thunderwave', 'willowisp'].some(m => hasMove[m])};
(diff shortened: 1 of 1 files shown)


data/random-teams.ts
```diff
@@ -906,7 +906,13 @@ export class RandomTeams {
...
 		case 'thunderwave': case 'voltswitch':
 			const cullInDoubles = hasMove['electroweb'] || hasMove['nuzzle'];
-			return {cull: counter.setupType || counter.speedsetup || hasMove['raindance'] || (isDoubles && cullInDoubles)};
+			return {cull:
+				counter.setupType ||
+				counter.speedsetup ||
+				hasMove['shiftgear'] ||
+				hasMove['raindance'] ||
+				(isDoubles && cullInDoubles),
+			};
```
Zarel, line 915:
```suggestion
			return {
				cull: counter.setupType ||
					counter.speedsetup ||
					hasMove['shiftgear'] ||
					hasMove['raindance'] ||
					(isDoubles && cullInDoubles),
			};
```

Very minor, but we prefer this style.

ACakeWearingAHat: oh! i was just matching some of the other parts of the file, but sure

data/random-teams.ts
```diff
@@ -906,7 +906,10 @@ export class RandomTeams {
 			return {cull: hasMove['encore'] || hasMove['nastyplot'] || hasMove['swordsdance']};
 		case 'thunderwave': case 'voltswitch':
 			const cullInDoubles = hasMove['electroweb'] || hasMove['nuzzle'];
```
AnnikaCodes, line 908:
```suggestion
			const cullInDoubles = isDoubles && (hasMove['electroweb'] || hasMove['nuzzle']);
```

data/random-teams.ts
```diff
@@ -906,7 +906,10 @@ export class RandomTeams {
...
-			return {cull: counter.setupType || counter.speedsetup || hasMove['raindance'] || (isDoubles && cullInDoubles)};
+			return {
+				cull: counter.setupType || counter.speedsetup || hasMove['shiftgear'] || hasMove['raindance'] ||
+					(isDoubles && cullInDoubles),
+			};
```
AnnikaCodes, line 912:
```suggestion
			return {cull: counter.setupType || counter.speedsetup || hasMove['shiftgear'] || hasMove['raindance'] || cullInDoubles};
```

AnnikaCodes (approved):
Aside from my style nitpick this is fine. Thanks!

PR #8198, opened 2021-04-13
Use a dispatch system for sockets

Zarel:
Instead of using a mega-switch, instead use a dispatch system for the sockets protocol.

A big advantage is that we don't need to be weird about variable declarations using this method. We can use `const`s!

### DIFF
--- a/server/sockets.ts
+++ b/server/sockets.ts
@@ -145,6 +145,124 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 
 	isTrustedProxyIp: (ip: string) => boolean;
 
+	receivers: {[k: string]: (this: ServerStream, data: string) => void} = {
+		'$'(data) {
+			// $code
+			// eslint-disable-next-line no-eval
+			eval(data.substr(1));
+		},
+		'!'(data) {
+			// !socketid
+			// destroy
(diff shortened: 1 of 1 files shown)


Zarel:
What do you all think about migrating the rest of our mega-switches to use this method?

mia-pi-git:
Personally, I'd favor that as well - cleaner, and easier to modify.

server/sockets.ts
```diff
@@ -408,137 +526,9 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
...
 	_write(data: string) {
 		// console.log('worker received: ' + data);
-		let socket: import('sockjs').Connection | undefined = undefined;
-		let socketid = '';
-		let room: Map<string, import('sockjs').Connection> | undefined = undefined;
-		let roomid = '' as RoomID;
-		let roomChannel: Map<string, ChannelID> | undefined = undefined;
-		let channelid: ChannelID = 0;
-		let nlLoc = -1;
-		let message = '';
```
Zarel, line 418: These are gone forever! Always the most awkward parts of mega-switches. I've experimented with several approaches in the past and they've all sucked.

Zarel:
This won't take effect until next restart so I'm not in a huge rush to merge.

PR #8201, opened 2021-04-15
Scavengers: Escape usernames in /scav status

PartMan7:
In particular, just on the huntmaker's name; the rest works already.


PartMan7:
Oh, whoops

PR #8202, opened 2021-04-15
Use onAfterMoveSecondary for Magician effect

Karthik99999:
This was done to fix the timing of item activation for Magician, which currently steals these items before they can activate. The bug report for this Magician interaction stated this happening to Weakness Policy, but Magician should interact in the same way with the other items I have changed. Should I add additional tests for the other items like Snowball, Cell Battery etc.?

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -1939,7 +1939,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 98,
 	},
 	magician: {
-		onAfterMoveSecondary(target, source, move) {
+		onAfterMoveSecondarySelf(source, target, move) {
 			if (!move || !target) return;
 			if (target !== source && move.category !== 'Status') {
 				if (source.item || source.volatiles['gem'] || move.id === 'fling') return;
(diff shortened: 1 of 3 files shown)


Karthik99999:
After taking a look at #8104, I tried changing magician to use onAfterMoveSecondary, and it worked while the items I changed were reverted back to using onDamagingHit.

DaWoblefet:
This looks correct to me! PS's onAfterMoveSecondary is in the Sheer Force-negated block which is where Magician is supposed to live. I'll let Marty sign off.

DaWoblefet (approved):

Marty-D:
There don't seem to be any existing tests for Magician actually working correctly, so it isn't very surprising that this skipped test passes since it's expecting Magician not to work. `AfterMoveSecondary` is for effects where the move's target is the Pokemon with the Ability. I think `AfterMoveSecondarySelf` with the source and target parameters swapped will do what you want.

Marty-D (approved):
That was quick; looks good, thanks! 👍

PR #8203, opened 2021-04-16
Implement Series 9 VGC 2021

DaWoblefet:
Although Series 8 officially doesn't switch over until the end of the month, all official Series 8 tournament are already completed. Reaching out to various community leaders, it seems there's only one semi-impactful grassroots tournament left (but it's on cartridge, not Showdown). As a result, many casual and competitive players are champing at the bit waiting to formally practice Series 9, as several major tournaments are upcoming even this month. I ran a poll in the VGC room for a few days to see if general opinion matched what the vast majority of competitors are telling me, and a supermajority were in favor of switching to Series 9 early:
image

Previously I was told in the past we couldn't have two active VGC ladders at the same time without doing some tomfoolery to get around it (last time, we named one of the series "Battle Stadium Doubles" which ultimately caused a lot of confusion). Assuming that is still the case, I've swapped the format name to be Gen 8 VGC 2021 Series 8 as challenge-only, and added a new ladder for Series 9. In the best case, we have two active ladders simultaneously.

EDIT: Kris said it was ok to run both ladders simultaneously (Series 8 through the end of the month). Whoo-hoo!


data/aliases.ts
```diff
@@ -5,7 +5,7 @@ export const Aliases: {[alias: string]: string} = {
 	mono: "[Gen 8] Monotype",
 	ag: "[Gen 8] Anything Goes",
 	bss: "[Gen 8] Battle Stadium Singles",
-	vgc: "[Gen 8] VGC 2020",
+	vgc: "[Gen 8] VGC 2021",
```
KrisXV, line 8: this wont do anything since theres no format named vgc 2021, you can alias this to series 9 if youd like

Zarel:
Yeah, we've had this problem before (some people want to practice for the next series before the previous series has fully completed). In the future, let's just always run two simultaneous VGC ladders in this situation.

PR #8204, opened 2021-04-17
Fix Transform source Ability leak

DaWoblefet:
Currently, when Transform sets the user's ability to that of the target's, it sends an ``-endability`` protocol message which leaks the Transform user's Ability. The way I approached this was by turning off that -endability message for all form changes. I'm _pretty_ sure this is fine, because it effectively was doing this anyway with requiring the source to be a move (I think we still need that check so Ability stuff doesn't start leaking all over the place, but I'm not super sure).

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1699,7 +1699,7 @@ export class Pokemon {
 		}
 		if (!this.battle.runEvent('SetAbility', this, source, this.battle.effect, ability)) return false;
 		this.battle.singleEvent('End', this.battle.dex.abilities.get(oldAbility), this.abilityData, this, source);
-		if (this.battle.effect && this.battle.effect.effectType === 'Move') {
+		if (this.battle.effect && this.battle.effect.effectType === 'Move' && !isFromFormeChange) {
 			this.battle.add('-endability', this, this.battle.dex.abilities.get(oldAbility), '[from] move: ' +
 				this.battle.dex.moves.get(this.battle.effect.id));
 		}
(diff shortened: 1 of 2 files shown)


Zarel:
Didn't we deprecate `|-endability|`? It got reconned to `|-volatile|Gastro Acid`, didn't it?

urkerab:
Apparently at some point it used to show that the user's old ability was being replaced by Entrainment, Roly Play, Simple Beam or Worry Seed. (The new ability is announced by the caller, which looks as if it would be incorrect for abilities that announce themselves, but I don't claim to know the correct sequence of activations.)

DaWoblefet:
> Didn't we deprecate `|-endability|`? It got reconned to `|-volatile|Gastro Acid`, didn't it?

It's documented in SIM-PROTOCOL as still in use, but only for Gastro Acid. I think that means we can remove it here in ``setAbility()`` entirely.

DaWoblefet:
Oh wait never mind, that doesn't work. Something like Role Play or Entrainment needs to show the old Ability being removed.

Zarel:
I remember in Gen 5, ability changes showed the old ability turning into the new ability.

Is Transform explicitly an exception to that in all gens?

DaWoblefet:
Yeah, Transform _never_ reveals the user's base ability. In Gen 7+, via the info screens, you learn what the _target's_ Ability was (i.e. what you Transformed into), but that's completely independent from this.

Zarel:
I assume the animation for old ability turning into new ability is split across multiple places. We might want them all in one place so this is clearer.

PR #8205, opened 2021-04-17
Add room sections

KrisXV:
This involves several protocol changes:

- `|queryresponse|userdetails|` has a new `customgroup?: string` field, which will be `"Section Leader"` for section leaders. Clients should display this in the same format as room groups and global groups.
- `|queryresponse|rooms|` no longer has `pspl` or `official` fields. Instead, the `chat` field's list of rooms gains two new fields for each room:
  - `section?: string`, the section name (`Official` for official rooms)
  - `spotlight?: string`, a spotlight name (to replace the old PSPL room system, but allowing you to set the label however you want)
- `|queryresponse|rooms|` has a new `sectionTitles: string[]` field, a list of sections in the order clients should display them in


server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,94 @@
+import {Utils} from '../../lib';
+
+export type RoomSection = 'none' | 'nonpublic' | 'officialrooms' | 'officialtiers' | 'communityprojects' |
```
mia-pi-git, line 3: Is there a reason this is a new file? It feels better suited to move the types / sections into rooms.ts, and the commands into room-settings (seeing as this _is_ a setting)

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,94 @@
...
+			if (staff && !staff.auth.has(userid)) this.parse(`/msgroom staff,/roomvoice ${userid}`);
+			if (targetUser) {
+				targetUser.popup(`You were appointed Section Leader by ${user.name}.`);
+				if (!targetUser.inRooms.has('staff')) this.parse(`/msgroom staff,/invite ${name}`);
+				if (staff && !targetUser.inRooms.has(staff.roomid)) {
+					this.parse(`/msgroom staff,/invite ${name}`);
+				}
+			}
+		} else {
+			Users.sectionLeaders.delete(userid);
```
mia-pi-git, line 73: Why isn't this in `Users.globalAuth`?

mia-pi-git: Or, I suppose since this would override the two keys, why not attach it to the room object directly, in room.auth (or the like)?

KrisXV: because you can't have `Users.globalAuth.get('kris')` and then have it return `Section Leader` and `+`

server/user-groups.ts
```diff
@@ -344,3 +344,55 @@ export class GlobalAuth extends Auth {
 		return true;
 	}
 }
+
+export class SectionLeaders extends Set<ID> {
```
mia-pi-git, line 348: See dms. imo, this is better suited `Map<userid, section the user leads>` that's attached to Users.globalAuth. (cleaner, and keeps everything in the same place)

mia-pi-git: Sharing from discord DMs:
> mia — Today at 12:46 AM
> ehh
> i feel like, since section leader is partially global
> it's better suited to being near the rest of the auth tables
> imo
> but that's just me
> plus, in that case like
> we wouldn't need a user.sectionLeader
> we could just like
```ts
Users.globalAuth.isSectionLeader = function(name: string) {
return this.sectionLeaders.has(toID(name)):
}```

(or, as was mentioned, just attach the SectionLeaders class to global auth)

File: server/chat-commands/room-sections.ts [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+export const sections: RoomSection[] = [
+	'none', 'nonpublic', 'officialrooms', 'officialtiers', 'communityprojects', 'gaming', 'languages', 'entertainment', 'lifehobbies', 'onsitegames',
+];
+
+export const sectionNames: {[k in RoomSection]: string} = {
+	none: 'none',
+	nonpublic: 'Non-public',
+	officialrooms: 'Official rooms',
+	officialtiers: 'Official tiers',
+	communityprojects: 'Community projects',
```
mia-pi-git, line 15: Would probably be better to caps these. 
```suggestion
	communityprojects: 'Community Projects',
```

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,94 @@
...
+	gaming: 'Gaming',
+	languages: 'Languages',
+	entertainment: 'Entertainment',
+	lifehobbies: 'Life & hobbies',
```
mia-pi-git, line 19:
```suggestion
	lifehobbies: 'Life & Hobbies',
```

server/chat-commands/room-sections.ts
KrisXV: i purposely didnt do it to match how ps currently does things, + i think all caps is ugly, especially next to hyphens

server/chat-commands/room-sections.ts
KrisXV: (to clarify, almost everything in this file is exported and interacts directly with the client, so i secluded it for cleanliness)

server/user-groups.ts
```diff
@@ -344,3 +346,59 @@ export class GlobalAuth extends Auth {
 		return true;
 	}
 }
+
+/**
+ * userid:section
+ */
+class SectionLeaders extends Map<ID, RoomSection> {
```
AnnikaCodes, line 353: This seems to share a lot of code with globalAuth, especially regarding save/load. I'm fine with it but it might be something to look into refactoring in the future (having a `UserMap<T>` class or something).

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+		const userid = toID(this.targetUsername);
+		const name = targetUser ? targetUser.name : this.targetUsername;
+		const demoting = cmd === 'desectionleader';
+		const leadsSection = Users.globalAuth.sectionLeaders.getSection(targetUser || userid);
+		if (Users.globalAuth.sectionLeaders.has(targetUser || userid) && !demoting && section === leadsSection) {
+			throw new Chat.ErrorMessage(`${name} is already a Section Leader of ${sectionNames[section]}.`);
+		} else if (!Users.globalAuth.sectionLeaders.has(targetUser || userid) && demoting) {
+			throw new Chat.ErrorMessage(`${name} is not a Section Leader.`);
+		}
+		const staff = Rooms.get('staff');
```
AnnikaCodes, line 62:
```suggestion
		const staffRoom = Rooms.get('staff');
```

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+			if (targetUser) {
+				targetUser.popup(`You were appointed Section Leader by ${user.name}.`);
+				if (!targetUser.inRooms.has('staff')) this.parse(`/msgroom staff,/invite ${name}`);
+				if (staff && !targetUser.inRooms.has(staff.roomid)) {
+					this.parse(`/msgroom staff,/invite ${name}`);
+				}
+			}
+		} else {
+			Users.globalAuth.sectionLeaders.delete(userid);
+			this.parse();
```
AnnikaCodes, line 77: What does this do?

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+			this.privateGlobalModAction(`${name} was demoted from Section Leader by ${user.name}.`);
+			this.globalModlog(`DESECTION LEADER`, userid);
+			if (staff && staff.auth.getDirect(userid) === '+') this.parse(`/msgroom staff,/roomdeauth ${userid}`);
```
AnnikaCodes, line 80:
```suggestion
			if (staff?.auth.getDirect(userid) === '+') this.parse(`/msgroom staff,/roomdeauth ${userid}`);
```
`undefined` !== `'+'`

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+				targetUser.popup(`You were appointed Section Leader by ${user.name}.`);
+				if (!targetUser.inRooms.has('staff')) this.parse(`/msgroom staff,/invite ${name}`);
+				if (staff && !targetUser.inRooms.has(staff.roomid)) {
+					this.parse(`/msgroom staff,/invite ${name}`);
+				}
+			}
+		} else {
+			Users.globalAuth.sectionLeaders.delete(userid);
+			this.parse();
+			this.privateGlobalModAction(`${name} was demoted from Section Leader by ${user.name}.`);
```
AnnikaCodes, line 78: This should probably include their former section?

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		const [targetStr, sectionid] = this.splitOne(target);
+		this.splitTarget(targetStr);
+		const section = room.sanitizeSection(sectionid);
+		const targetUser = this.targetUser;
+		const userid = toID(this.targetUsername);
+		const name = targetUser ? targetUser.name : this.targetUsername;
+		const demoting = cmd === 'desectionleader';
+		const leadsSection = Users.globalAuth.sectionLeaders.getSection(targetUser || userid);
+		if (Users.globalAuth.sectionLeaders.has(targetUser || userid) && !demoting && section === leadsSection) {
```
AnnikaCodes, line 57:
```suggestion
		if (Users.globalAuth.sectionLeaders.has(targetUser || userid) && !demoting) {
```
Your code assumes people will only lead one section at a time, and your help message says that they should be demoted from Section Leader first. I spoke to US and we're not planning to have people lead two sections at once. (I'm fine with having `/sectionleader` change the section they lead, but then the help should be changed.)

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
...
+			if (targetUser.trusted && !Users.isTrusted(targetUser.id)) {
+				targetUser.trusted = '';
+			}
+		}
+	},
+	sectionleaderhelp: [
+		`/sectionleader [target user], [sectionid] - Appoints [target user] Section Leader.`,
+		`/desectionleader [target user] - Demotes [target user] from Section Leader.`,
+		`Valid sections: ${sections.join(', ')}`,
+		`If you wish to change someone's section to another one, just /desectionleader then /sectionleader.`,
```
AnnikaCodes, line 96:
```suggestion
		`If you want to change which section someone leads, demote them and then re-promote them in the desired section.`,
```

server/chat-commands/room-sections.ts
```diff
@@ -0,0 +1,99 @@
+import {Utils} from '../../lib';
```
AnnikaCodes, line 1: Does this really need its own chat plugin? Sim authority/recognition generally goes in in `moderation.ts`.

server/chat-commands/room-sections.ts
mia-pi-git: Unresolving this since it appears I am not the only one with this question:

server/chat-commands/room-sections.ts
KrisXV: remnant of code i forgot to remove while debugging

server/chat-commands/moderation.ts
```diff
@@ -1363,6 +1389,59 @@ export const commands: ChatCommands = {
...
+		this.checkCan('gdeclare');
+		room = this.requireRoom();
+		const demoting = cmd === 'desectionleader';
+		if (!target || (target.split(',').length < 2 && !demoting)) return this.parse(`/help sectionleader`);
+
+		const [targetStr, sectionid] = this.splitOne(target);
+		this.splitTarget(targetStr);
+		const targetUser = this.targetUser;
+		const userid = toID(this.targetUsername);
+		const section = !demoting ? room.sanitizeSection(sectionid) : Users.globalAuth.sectionLeaders.getSection(userid)!;
```
AnnikaCodes, line 1403:
```suggestion
		const section = demoting ? Users.globalAuth.sectionLeaders.getSection(userid)! : room.sanitizeSection(sectionid);
```

server/chat-commands/room-settings.ts
```diff
@@ -1114,47 +1143,26 @@ export const commands: ChatCommands = {
...
 		if (!room.persist) {
-			return this.errorReply(`/psplwinnerroom - This room can't be marked as a PSPL Winner room`);
+			return this.errorReply(`/psplwinnerroom - This room can't be marked as a PSPL Winner room.`);
 		}
 		if (this.meansNo(target)) {
 			if (!room.settings.pspl) return this.errorReply(`This chat room is already not a PSPL Winner room.`);
-			delete room.settings.pspl;
 			this.addModAction(`${user.name} made this chat room no longer a PSPL Winner room.`);
-			this.modlog('PSPLROOM');
+			this.modlog('UNPSPLROOM');
```
AnnikaCodes, line 1159:
```suggestion
			this.globalModlog('UNPSPLROOM');
```

server/chat-commands/room-settings.ts
```diff
@@ -1114,47 +1143,26 @@ export const commands: ChatCommands = {
...
 			delete room.settings.pspl;
 			room.saveSettings();
 		} else {
 			if (room.settings.pspl) return this.errorReply("This chat room is already a PSPL Winner room.");
-			room.settings.pspl = true;
 			this.addModAction(`${user.name} made this chat room a PSPL Winner room.`);
-			this.modlog('UNPSPLROOM');
+			this.modlog('PSPLROOM');
```
AnnikaCodes, line 1165:
```suggestion
			this.globalModlog('PSPLROOM');
```

server/chat-commands/room-settings.ts
```diff
@@ -1114,47 +1143,26 @@ export const commands: ChatCommands = {
...
 		if (!room.persist) {
-			return this.errorReply(`/psplwinnerroom - This room can't be marked as a PSPL Winner room`);
+			return this.errorReply(`/psplwinnerroom - This room can't be marked as a PSPL Winner room.`);
 		}
 		if (this.meansNo(target)) {
 			if (!room.settings.pspl) return this.errorReply(`This chat room is already not a PSPL Winner room.`);
-			delete room.settings.pspl;
 			this.addModAction(`${user.name} made this chat room no longer a PSPL Winner room.`);
-			this.modlog('PSPLROOM');
+			this.globalModlog('UNPSPLROOM');
```
Distrib-ps, line 1159:
```suggestion
			this.globalModlog('UNPSPLROOM', null);
```

server/chat-commands/room-settings.ts
```diff
@@ -1114,47 +1143,26 @@ export const commands: ChatCommands = {
...
 			delete room.settings.pspl;
 			room.saveSettings();
 		} else {
 			if (room.settings.pspl) return this.errorReply("This chat room is already a PSPL Winner room.");
-			room.settings.pspl = true;
 			this.addModAction(`${user.name} made this chat room a PSPL Winner room.`);
-			this.modlog('UNPSPLROOM');
+			this.globalModlog('PSPLROOM');
```
Distrib-ps, line 1165:
```suggestion
			this.globalModlog('PSPLROOM', null);
```

server/chat-commands/room-settings.ts
mia-pi-git: null isn't required here - this has no effect.

Distrib-ps: > null isn't required here - this has no effect.
null is for : Expected 2-4 arguments, but got 1

mia-pi-git: Huh. That certainly shouldn't be required. It seems like the optimal solution is to mark the user parameter as optional, instead (as it used to be?)

AnnikaCodes: Yes. See d5a4c7d9e2a14ad83503eb5c455f9063fc8e60a8.

server/chat-commands/moderation.ts
```diff
@@ -414,6 +418,28 @@ export const commands: ChatCommands = {
...
+		const sections: {[k in RoomSection]: Set<string>} = Object.create(null);
+		for (const section of RoomSections.sections) {
+			if (['none', 'nonpublic'].includes(section)) continue;
+			sections[section] = new Set<string>();
+		}
+		for (const [id, username] of usernames) {
+			const sectionid = Users.globalAuth.sectionLeaders.getSection(id)!;
+			if (!sections[sectionid]) continue;
+			sections[sectionid].add(username);
+		}
```
AnnikaCodes, line 433:
```suggestion
		for (const [id, username] of usernames) {
			const sectionid = Users.globalAuth.sectionLeaders.getSection(id);
			if (!sectionid) continue;
			if (!sections[sectionid]) sections[sectionid] = new Set();
			sections[sectionid].add(username);
		}
```

server/chat-commands/moderation.ts
```diff
@@ -1363,6 +1389,59 @@ export const commands: ChatCommands = {
...
+		const staffRoom = Rooms.get('staff');
+		if (!demoting) {
+			Users.globalAuth.sectionLeaders.add(userid, section, name);
+			this.addGlobalModAction(`${name} was appointed Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+			this.globalModlog(`SECTION LEADER`, userid, section);
+			if (!staffRoom?.auth.has(userid)) this.parse(`/msgroom staff,/forceroompromote ${userid},+`);
+			if (targetUser) {
+				targetUser.popup(`You were appointed Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+			}
```
AnnikaCodes, line 1418:
```suggestion
			targetUser?.popup(`You were appointed Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
```

server/chat-commands/moderation.ts
```diff
@@ -1363,6 +1389,59 @@ export const commands: ChatCommands = {
...
+		} else {
+			Users.globalAuth.sectionLeaders.delete(userid);
+			this.privateGlobalModAction(`${name} was demoted from Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+			this.globalModlog(`DESECTION LEADER`, userid, section);
+			if (staffRoom?.auth.getDirect(userid) === '+') this.parse(`/msgroom staff,/roomdeauth ${userid}`);
+			if (targetUser) {
+				targetUser.popup(`You were demoted from Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+			}
```
AnnikaCodes, line 1426:
```suggestion
			targetUser?.popup(`You were demoted from Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
```

server/chat-commands/room-settings.ts
```diff
@@ -708,9 +730,14 @@ export const commands: ChatCommands = {
...
 		room = this.requireRoom();
 		this.checkCan('makeroom');
-		const id = toID(target);
+		let [id, sectionTarget] = this.splitOne(target).map(toID);
+		if (!sectionTarget) sectionTarget = (cmd.includes('public') ? 'none' : 'nonpublic') as ID;
 		if (!id || this.cmd === 'makechatroom') return this.parse('/help makechatroom');
-		if (!Rooms.global.addChatRoom(target)) {
+		const section = room.sanitizeSection(sectionTarget);
+		if (!cmd.includes('public') && section !== 'nonpublic') {
+			throw new Chat.ErrorMessage(`Non-public rooms cannot be placed into room sections.`);
```
AnnikaCodes, line 738:
```suggestion
			throw new Chat.ErrorMessage(`Only public rooms can be placed into room sections.`);
```

server/rooms.ts
```diff
@@ -824,6 +829,26 @@ export abstract class BasicRoom {
...
+	adjustSection(newSection: string) {
+		if (!this.persist) {
+			throw new Chat.ErrorMessage(`You cannot change the section of temporary rooms.`);
+		}
+		const section = this.sanitizeSection(newSection);
+		const oldSection = this.settings.section;
+		if (oldSection === section) {
+			throw new Chat.ErrorMessage(`${this.title}'s room section is already set to "${RoomSections.sectionNames[oldSection]}".`);
+		}
+		this.settings.section = this.section = section;
```
AnnikaCodes, line 848: Why does the section have to be included both in the settings and the `Room` object? One place should suffice.

AnnikaCodes:
> GitHub Actions has encountered an internal error when running your job.

Well that's odd.

KrisXV:
> > GitHub Actions has encountered an internal error when running your job.
> 
> Well that's odd.

I re ran the tasks and now it’s good

server/rooms.ts
```diff
@@ -217,7 +219,7 @@ export abstract class BasicRoom {
 	 */
 	nthMessageHandlers: Map<MessageHandler, number>;
 
-	constructor(roomid: RoomID, title?: string, options: Partial<RoomSettings> = {}) {
+	constructor(roomid: RoomID, title?: string, section: RoomSection = 'none', options: Partial<RoomSettings> = {}) {
```
Zarel, line 222: Why is `section` its own param instead of just being a roomsetting (and defaulting to 'none')?

KrisXV: It's so admins can specify sections upon room creation if they want

server/rooms.ts
```diff
@@ -1368,14 +1399,12 @@ export class GlobalRoomState {
 				title: room.title,
 				desc: room.settings.desc || '',
 				userCount: room.userCount,
+				section: room.settings.section || 'none',
```
Zarel, line 1402: `|| undefined` would save bandwidth and shouldn't be difficult for the client to parse.

KrisXV: the 'none' is for rooms that havent been placed into a section yet, though thinking back this should probably go into client code when building the room list for side servers on old code, so fair

server/chat-commands/room-settings.ts
```diff
@@ -708,37 +730,44 @@ export const commands: ChatCommands = {
 	makechatroom(target, room, user, connection, cmd) {
 		room = this.requireRoom();
 		this.checkCan('makeroom');
-		const id = toID(target);
+		let [title, sectionTarget] = this.splitOne(target).map(toID);
```
Zarel, line 733:
```suggestion
		let [title, sectionTarget] = this.splitOne(target);
```

server/rooms.ts
```diff
@@ -824,6 +829,29 @@ export abstract class BasicRoom {
...
 		}
 	}
+	sanitizeSection(section: string) {
+		const target = toID(section);
+		if (!RoomSections.sections.includes(target as any)) {
+			throw new Chat.ErrorMessage(`"${target}" is not a valid room section. Valid categories include: ${RoomSections.sections.join(', ')}`);
+		}
+		return target as RoomSection;
+	}
+	adjustSection(newSection: string) {
```
Zarel, line 839: Is there a reason this isn't `setSection`?

KrisXV: not really

server/rooms.ts
```diff
@@ -824,6 +829,29 @@ export abstract class BasicRoom {
 			}
 		}
 	}
+	sanitizeSection(section: string) {
```
Zarel, line 832: Is there a reason this isn't `validateSection`?

Zarel: ("sanitize" means something different from what you're doing here)

KrisXV: again not really

server/rooms.ts
```diff
@@ -248,6 +251,7 @@ export abstract class BasicRoom {
 			title: this.title,
 			auth: Object.create(null),
 			creationTime: Date.now(),
+			section: 'none',
```
Zarel, line 254: Why not just default to not having a section entry?

```suggestion
```

server/chat-commands/room-settings.ts
```diff
@@ -813,6 +842,7 @@ export const commands: ChatCommands = {
 			creationTime: Date.now(),
 			modjoin: parent ? null : '+',
 			parentid: parent,
+			section: 'nonpublic',
```
Zarel, line 845: Do we need a section named `'nonpublic'`?

server/rooms.ts
KrisXV: did

server/chat-commands/room-settings.ts
KrisXV: no

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
 const SLOWCHAT_MAXIMUM = 60;
 const SLOWCHAT_USER_REQUIREMENT = 10;
 
+export const sections = [
+	'officialrooms', 'officialtiers', 'communityprojects', 'languages', 'gaming', 'leisure', 'lifehobbies', 'onsitegames',
```
Zarel, line 19: Should this be hardcoded?

Zarel: I suppose it's fine for now.

server/rooms.ts
```diff
@@ -56,7 +57,9 @@ interface ChatRoomTable {
 	title: string;
 	desc: string;
 	userCount: number;
+	section?: RoomSection;
 	subRooms?: string[];
+	prizewinner?: true;
```
Zarel, line 62: After sleeping on it, I think `spotlight` is a better term than `prizewinner`.

Specifically, `spotlight?: string`, so it can be `spotlight: "PSPL Winner",` but would make it easier for us to make spotlights with different names in the future. Perhaps also rename the commands to be similarly generic, but that part's less important than the protocol.

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
...
+] as const;
+
+export type RoomSection = typeof sections[number];
+
+export const RoomSections: {sectionNames: {[k in RoomSection]: string}, sections: readonly RoomSection[]} = {
+	sectionNames: {
+		officialrooms: 'Official rooms',
+		officialtiers: 'Official tiers',
+		communityprojects: 'Community projects',
```
Zarel, line 28: I think these names make Official Rooms and Official Tiers sound more related than they actually are, and Upper Staff agreed.

New names:

```suggestion
		official: 'Official',
		smogontiers: 'Smogon tiers',
		communityprojects: 'Community projects',
```

Zarel: I also wanted "Smogon formats" and "Community projects and formats" but I couldn't build consensus for it.

server/rooms.ts
```diff
@@ -1352,11 +1383,13 @@ export class GlobalRoomState {
...
-			official: ChatRoomTable[], pspl: ChatRoomTable[], chat: ChatRoomTable[], userCount: number, battleCount: number,
+			chat: ChatRoomTable[],
+			sectionTitles: {[k: string]: string},
+			userCount: number,
+			battleCount: number,
 		} = {
-			official: [],
-			pspl: [],
 			chat: [],
+			sectionTitles: RoomSections.sectionNames,
```
Zarel, line 1392: Might it be better to just use the section title directly in the chat list, instead of needing to bundle an id:title map?

Zarel: I did the math and: yes, that would be better. Since it also makes it easier to parse, I strongly recommend going with that approach instead.

server/rooms.ts
```diff
@@ -1368,18 +1401,13 @@ export class GlobalRoomState {
 				title: room.title,
 				desc: room.settings.desc || '',
 				userCount: room.userCount,
+				section: room.settings.section || undefined,
```
Zarel, line 1404:
```suggestion
				section: room.settings.section ? RoomSections.sectionNames[room.settings.section] : undefined,
```

server/rooms.ts
```diff
@@ -1394,7 +1422,7 @@ export class GlobalRoomState {
 		if (Rooms.rooms.has(id)) return false;
 
 		const settings = {
-			title,
+			title: title,
```
Zarel, line 1425: wut

KrisXV: oh it was from a scrapped change, oops

server/rooms.ts
KrisXV: How would i build the dropdown menu on the client then?

server/rooms.ts
KrisXV: This still has a chance to return undefined if somehow a room's section gets mutated; wouldnt it make more sense to pass along `RoomSections.sectionName[room.settings.section] || room.settings.section` (room.settings.section is typed `RoomSection | undefined`)?

server/rooms.ts
KrisXV: I'll change to spotlight, but the generalizing `/psplwinnerroom` etc should probably be in a different PR

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
...
+		languages: 'Languages',
+		gaming: 'Gaming',
+		leisure: 'Leisure',
```
Zarel, line 31:
```suggestion
		entertainment: 'Entertainment',
```

server/rooms.ts
Zarel: You could send the title rather than the ID. Shouldn't be very different.

Zarel: I can also do it myself.

server/rooms.ts
Zarel: Yeah, that's fine.

server/rooms.ts
Zarel: Sure.

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
...
+export const sections = [
+	'official', 'smogontiers', 'communityprojectsformats', 'languages', 'entertainment', 'gaming', 'lifehobbies', 'onsitegames',
+] as const;
+
+export type RoomSection = typeof sections[number];
+
+export const RoomSections: {sectionNames: {[k in RoomSection]: string}, sections: readonly RoomSection[]} = {
+	sectionNames: {
+		official: 'Official',
+		smogontiers: 'Smogon tiers',
```
Zarel:
```suggestion
		smogonformats: 'Smogon formats',
```

To match "Community formats", although apparently we're considering merging them in the UI.

server/user-groups.ts
Zarel: After some discussion, I think I strongly prefer this to just be a room rank in Staff rather than an entirely separate class with its own persistence mechanism.

KrisXV: Any particular symbol?

server/user-groups.ts
```diff
@@ -355,4 +358,31 @@ export class GlobalAuth extends Auth {
...
+			return this.delete(id);
+		}
+		const user = Users.get(id);
+		if (user) {
+			user.updateIdentity();
+			Rooms.global.checkAutojoin(user);
+		}
+		this.save();
+		return true;
+	}
```
Zarel, line 387: So, this is in some sense difficult to avoid, but if you set someone to trusted, then set and delete their section, they won't be trusted anymore.

I don't think it's worth trying to remember the trusted state from before the promotion somewhere. But I think the best solution here is probably to add a message about that when demoting them.

KrisXV: Would you mind doing this? I won’t be home for a bit.

server/chat-commands/moderation.ts
```diff
@@ -1363,6 +1386,55 @@ export const commands: ChatCommands = {
...
+		const staffRoom = Rooms.get('staff');
+		if (!demoting) {
+			Users.globalAuth.setSection(userid, section);
+			this.addGlobalModAction(`${name} was appointed Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+			this.globalModlog(`SECTION LEADER`, userid, section);
+			if (!staffRoom?.auth.has(userid)) this.parse(`/msgroom staff,/forceroompromote ${userid},▸`);
+			targetUser?.popup(`You were appointed Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
+		} else {
+			Users.globalAuth.deleteSection(userid);
+			this.privateGlobalModAction(`${name} was demoted from Section Leader of ${RoomSections.sectionNames[section]} by ${user.name}.`);
```
Zarel, line 1416:
```suggestion
			if (Users.globalAuth.get(userid) === ' ') {
				this.sendReply(`They are also no longer manually trusted. If they should be, use /trustuser`);
			}
```

server/user-groups.ts
Zarel: Restart isn't for another half-day. I added the suggestion but you're in no hurry to commit it. I think the rest of this PR is done now.

server/chat-commands/core.ts
```diff
@@ -1704,12 +1704,14 @@ export const commands: ChatCommands = {
...
 			if (targetUser.locked) group = Config.punishgroups?.locked?.symbol ?? '\u203d';
 			if (targetUser.namelocked) group = Config.punishgroups?.namelocked?.symbol ?? '✖';
+			const sectionleader = Users.globalAuth.sectionLeaders.has(targetUser.id);
 			const userdetails: AnyObject = {
 				id: target,
 				userid: targetUser.id,
 				name: targetUser.name,
 				avatar: targetUser.avatar,
 				group: group,
+				sectionleader,
```
Zarel, line 1714: Hm, I still don't know how to feel about this.

Zarel, line 1714:
```suggestion
			const userdetails: AnyObject = {
				id: target,
				userid: targetUser.id,
				name: targetUser.name,
				avatar: targetUser.avatar,
				group: group,
				sectionleader: Users.globalAuth.sectionLeaders.has(targetUser.id),
```

Zarel: I wonder if this should be something like:

```ts
customgroup: Users.globalAuth.sectionLeaders.has(targetUser.id) ? "Section Leader" : undefined,
```

Zarel: (Similar to spotlight, it would let side servers set their own messages in the group section.)

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
...
+export const RoomSections: {sectionNames: {[k in RoomSection]: string}, sections: readonly RoomSection[]} = {
+	sectionNames: {
+		official: 'Official',
+		smogonformats: 'Smogon formats',
+		communityprojectsformats: 'Community projects & formats',
+		languages: 'Languages',
+		entertainment: 'Entertainment',
+		gaming: 'Gaming',
+		lifehobbies: 'Life & hobbies',
+		onsitegames: 'On-site games',
```
Zarel, line 33:
```suggestion
		official: 'Official',
		battleformats: 'Battle formats',
		languages: 'Languages',
		entertainment: 'Entertainment',
		gaming: 'Gaming',
		lifehobbies: 'Life & hobbies',
		onsitegames: 'On-site games',
```

server/chat-commands/room-settings.ts
```diff
@@ -15,6 +15,26 @@ const SLOWCHAT_MINIMUM = 2;
 const SLOWCHAT_MAXIMUM = 60;
 const SLOWCHAT_USER_REQUIREMENT = 10;
 
+export const sections = [
+	'official', 'smogonformats', 'communityprojectsformats', 'languages', 'entertainment', 'gaming', 'lifehobbies', 'onsitegames',
```
Zarel, line 19:
```suggestion
	'official', 'battleformats', 'languages', 'entertainment', 'gaming', 'lifehobbies', 'onsitegames',
```

server/chat-commands/room-settings.ts
Zarel: Upper Staff decided that putting these all in one section made the most sense.

Zarel:
Is there a need for section IDs to exist at all? I think it might be easier to use only section names.

Zarel:
I guess for commands so people don't have to match capitalization. Yeah, that's probably fine.

server/user-groups.ts
Zarel: (We decided to go with `▸`, to represent a collapsed section.)

PR #8206, opened 2021-04-17
Add Plasma Fists hint

DaWoblefet:
Approved suggestion by Aeonic: https://www.smogon.com/forums/threads/have-a-message-clarifying-plasma-fists-interactions.3680958/

Plasma Fists turning Normal-type moves into Electric-type moves is a somewhat common mistaken bug report.


Zarel:
Thanks!

PR #8210, opened 2021-04-18
SSB4: Support Same Type Clause

xInstruct:
Approved by ~~corrupt project head~~ myself

### DIFF
--- a/data/mods/ssb/random-teams.ts
+++ b/data/mods/ssb/random-teams.ts
@@ -871,7 +871,12 @@ export class RandomStaffBrosTeams extends RandomTeams {
 	randomStaffBrosTeam(options: {inBattle?: boolean} = {}) {
 		const team: PokemonSet[] = [];
 		const debug: string[] = []; // Set this to a list of SSB sets to override the normal pool for debugging.
-		const pool = debug.length ? debug : Object.keys(ssbSets);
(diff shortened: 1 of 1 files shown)


data/mods/ssb/random-teams.ts
```diff
@@ -871,44 +871,53 @@ export class RandomStaffBrosTeams extends RandomTeams {
...
-				for (const type of weaknesses) {
-					if (typePool[type] === undefined) typePool[type] = 0;
-					if (typePool[type] >= 3) {
-						// Reject
-						rejected = true;
-						break;
+				if (ruleTable.has('sametypeclause')) {
+					if (!isFiltered) {
+						const monotype = this.sample(types);
+						pool = pool.filter(x => this.dex.species.get(ssbSets[x].species).types.includes(monotype));
```
AnnikaCodes, line 890: Shouldn't you select randomly from all types, rather than picking a type of the first Pokémon selected? That's how regular Monotype Random Battle does it, and it would simplify the code (and make each type equally likely to be chosen).

xInstruct: good point

xInstruct: taken care of, and made the code a lot more concise because of it

AnnikaCodes (approved):
Looks good, thanks!

PR #8211, opened 2021-04-18
Gen 1: Don't allow Pokemon to recharge if frozen

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8768936

This implements the above mechanic, where a pokemon that is frozen during its recharge turn can't do anything while it is still frozen, including switching out or actually recharging. I reimplemented the freeze move disabling in the `mustrecharge` condition because of the priority for onBeforeMove in `mustrecharge`.

### DIFF
--- a/data/mods/gen1/conditions.ts
+++ b/data/mods/gen1/conditions.ts
@@ -87,8 +87,9 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		effectType: 'Status',
 		onStart(target) {
 			this.add('-status', target, 'frz');
+			delete target.volatiles['mustrecharge']?.duration;
 		},
-		onBeforeMovePriority: 10,
+		onBeforeMovePriority: 12,
 		onBeforeMove(pokemon, target, move) {
 			this.add('cant', pokemon, 'frz');
 			pokemon.lastMove = null;
@@ -220,17 +221,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 	},
 	mustrecharge: {
 		inherit: true,
-		onBeforeMove(pokemon) {
-			if (pokemon.status === 'frz') {
-				this.effectData.duration += 1;
-				this.add('cant', pokemon, 'frz');
-				pokemon.lastMove = null;
-				return false;
-			}
-			this.add('cant', pokemon, 'recharge');
-			pokemon.removeVolatile('mustrecharge');
-			return null;
-		},
 		onStart() {},
 	},
 	lockedmove: {
--- a/data/mods/gen1/conditions.ts
+++ b/data/mods/gen1/conditions.ts
@@ -87,7 +87,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		effectType: 'Status',
 		onStart(target) {
 			this.add('-status', target, 'frz');
-			delete target.volatiles['mustrecharge']?.duration;
 		},
 		onBeforeMovePriority: 12,
 		onBeforeMove(pokemon, target, move) {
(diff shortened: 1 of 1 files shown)


urkerab:
> I reimplemented the freeze move disabling in the `mustrecharge` condition because of the priority for `onBeforeMove` in `mustrecharge`.

Wouldn't it be easier to increase the priority for `frz` to 12?

Zarel:
What @urkerab said.

Karthik99999:
> Wouldn't it be easier to increase the priority for frz to 12?

I tried this, but that skipped over `mustrecharge`'s `onBeforeMove` check, which meant the duration didn't get the +1 it should get to prevent it from ending. That might mean there's a better way to implement the `mustrecharge` duration from never ending during freeze, but I don't know of a different way.

urkerab:
> That might mean there's a better way to implement the `mustrecharge` duration from never ending during freeze

Fortunately the `mustrecharge` code itself already removes the volatile, so you probably don't need a duration. (Don't take my word for that though.)

Zarel:
`mustrecharge` should probably have a duration of 2 so it automatically goes away without needing the recharge, in that case.

Zarel:
Oh, other way around, you want to _prevent_ it from ending. Just remove the duration completely, if so.

Zarel:
Make it so the `onBeforeMove` handler removes the duration instead.

Karthik99999:
I was able to delete the `mustrecharge` duration during the `onStart` function for freeze and get the same effect. There is a visual bug that deletes the "Must Recharge" tag from the mon even if the duration was deleted, but that's a client side fix.

urkerab:
> `mustrecharge` should probably have a duration of 2 so it automatically goes away without needing the recharge, in that case.

That behaviour is only needed by Sky Drop, so for Gen 1 you could might be able to hard code zero duration and rely on the volatile removing itself?

data/mods/gen1/conditions.ts
```diff
@@ -87,8 +87,9 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		effectType: 'Status',
 		onStart(target) {
 			this.add('-status', target, 'frz');
+			delete target.volatiles['mustrecharge']?.duration;
```
Zarel, line 90: You should remove `duration` from `mustrecharge` directly, unless there exists another effect that suppresses "Must Recharge" but still makes it end early?

data/mods/gen1/conditions.ts
```diff
@@ -220,17 +221,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 	},
 	mustrecharge: {
 		inherit: true,
```
Zarel:
```suggestion
		inherit: true,
		duration: 0,
```

(Like this.)

Karthik99999:
Removed the recharge duration, was just hesitant about removing it since I don't know gen 1 mechanics that well.

Zarel:
> Removed the recharge duration, was just hesitant about removing it since I don't know gen 1 mechanics that well.

Completely fair! @PlagueVonKarma if you could verify this for us, that'd help a lot. But I definitely would assume "You always need to recharge once" is more likely to be true than "You only need to recharge after turn 2 if you're frozen".

PlagueVonKarma:
> > Removed the recharge duration, was just hesitant about removing it since I don't know gen 1 mechanics that well.
> 
> Completely fair! @PlagueVonKarma if you could verify this for us, that'd help a lot. But I definitely would assume "You always need to recharge once" is more likely to be true than "You only need to recharge after turn 2 if you're frozen".

I believe you're right.

Marty-D (approved):
I, too, am hesitant to touch Gen 1 since it's so brittle, but this seems correct to me. Thanks!

PR #8213, opened 2021-04-19
PU: Ban Clawitzer, Kingler and Omastar

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/post-8826858


AnnikaCodes:
Looks good, thanks!

PR #8214, opened 2021-04-20
Support forcing modchat on ladder tour battles

mia-pi-git:
Requested here https://www.smogon.com/forums/threads/allow-for-auto-modchatting-on-a-prefix.3682737/#post-8829088.
Monkeypatches are here https://gist.github.com/mia-pi-git/61fc91671107113a252d9d39791bad87.
I was told it would be optimal to have this live on the server by the 25th, as there's a tour that'd benefit from this.

### DIFF
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -504,7 +504,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 	ended: boolean;
 	active: boolean;
 	replaySaved: boolean;
-	forcePublic: string | null = null;
+	forcedSettings: {modchat?: string | null, privacy?: string | null} = {};
 	playerTable: {[userid: string]: RoomBattlePlayer};
 	players: RoomBattlePlayer[];
 	p1: RoomBattlePlayer;
@@ -1070,12 +1070,19 @@ export class RoomBattle extends RoomGames.RoomGame {
 
 		if (user) {
 			this.room.auth.set(player.id, Users.PLAYER_SYMBOL);
-			if (this.rated && !this.forcePublic) {
-				this.forcePublic = user.battlesForcedPublic();
+			if (this.rated) {
+				this.checkForcedSettings(user);
 			}
 		}
 		if (user?.inRooms.has(this.roomid)) this.onConnect(user);
 		return player;
+	}
+
+	checkForcedSettings(user: User) {
+		this.forcedSettings = {
+			modchat: this.forcedSettings.modchat || user.forcedPublic('modchat'),
+			privacy: this.forcedSettings.privacy || user.forcedPublic('privacy'),
+		};
 	}
 
 	makePlayer(user: User) {
--- a/server/room-battle.ts
+++ b/server/room-battle.ts
@@ -1087,7 +1087,7 @@ export class RoomBattle extends RoomGames.RoomGame {
 			} else if (playerOptions && playerOptions.hidden) {
 				privacySetter.add(playerOptions.user.id);
 			}
-			if (playerOptions) this.checkForcedUserSettings(playerOptions.user);
+			if (playerOptions?.user) this.checkForcedUserSettings(playerOptions.user);
 		}
 
 		if (privacySetter.size) {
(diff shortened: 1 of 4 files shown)


server/users.ts
```diff
@@ -1480,9 +1480,13 @@ export class User extends Chat.MessageContext {
 			this.registered ? `[registered]` :
 			``;
 	}
-	battlesForcedPublic() {
-		if (!Config.forcedpublicprefixes) return null;
-		for (const prefix of Config.forcedpublicprefixes) {
+	forcedPublic(key: 'modchat' | 'privacy') {
```
Zarel, line 1483: This should probably be `battleForcedSetting`?

Zarel: Also, putting it in `room-battle` might make maintenance easier than having the functionality spread out over lots of different files.

server/rooms.ts
```diff
@@ -1935,13 +1933,19 @@ export const Rooms = {
 			} else if (playerOptions && playerOptions.hidden) {
 				privacySetter.add(playerOptions.user.id);
 			}
+			if (playerOptions) battle.checkForcedSettings(playerOptions.user);
+		}
+
+		// Special battles have modchat set to Player from the beginning
+		if (p1Special || battle.forcedSettings.modchat) {
+			room.settings.modchat = '\u2606';
```
Zarel, line 1941: This can also be moved inside `checkForcedSettings`.

server/rooms.ts
```diff
@@ -1935,13 +1933,19 @@ export const Rooms = {
...
 		}
 
 		if (privacySetter.size) {
-			if (battle.forcePublic) {
+			if (battle.forcedSettings.privacy) {
```
Zarel, line 1945: Honestly, this all can.

server/users.ts
mia-pi-git: I'll make it `static` - That functionality _is_ needed in other places.

server/rooms.ts
mia-pi-git: Done.

server/rooms.ts
mia-pi-git: Should this be? It seems fine where it is (also, passing the options down seems messy and not worth it?).

mia-pi-git: Never mind, I just went and did it.

server/rooms.ts
```diff
@@ -1922,39 +1922,11 @@ export const Rooms = {
...
-					room.settings.modjoin = '%';
-					room.add(`|raw|<div class="broadcast-red"><strong>This battle is invite-only!</strong><br />Users must be invited with <code>/invite</code> (or be staff) to join</div>`);
-				}
-			}
-		}
+		battle.checkPrivacySettings(options);
 
 		for (const p of players) {
 			if (p) {
+				battle.checkForcedUserSettings(p);
```
Zarel, line 1929: This can also be moved inside `checkPrivacySettings`, right?

server/room-battle.ts
```diff
@@ -16,6 +16,7 @@ import {execSync} from "child_process";
 import {BattleStream} from "../sim/battle-stream";
 import * as RoomGames from "./room-game";
 import type {Tournament} from './tournaments/index';
+import {RoomSettings} from './rooms';
```
Zarel, line 19:
```suggestion
import type {RoomSettings} from './rooms';
```

server/room-battle.ts
```diff
@@ -1070,14 +1071,68 @@ export class RoomBattle extends RoomGames.RoomGame {
...
+		const privacySetter = new Set<ID>([]);
+		for (const p of ['p1', 'p2', 'p3', 'p4'] as const) {
+			const playerOptions = options[p];
+			if (playerOptions && playerOptions.inviteOnly) {
+				inviteOnly = true;
+				privacySetter.add(playerOptions.user.id);
+			} else if (playerOptions && playerOptions.hidden) {
+				privacySetter.add(playerOptions.user.id);
+			}
+			if (playerOptions?.user) this.checkForcedUserSettings(playerOptions.user);
```
Zarel, line 1090:
```suggestion
			if (playerOptions) {
				if (playerOptions.inviteOnly) {
					inviteOnly = true;
					privacySetter.add(playerOptions.user.id);
				} else if (playerOptions.hidden) {
					privacySetter.add(playerOptions.user.id);
				}
				if (playerOptions.user) this.checkForcedUserSettings(playerOptions.user);
			}
```

Zarel (approved):

Zarel:
ty!

This will be merged for the restart tonight.

PR #8215, opened 2021-04-21
Add more support for custom punishments

mia-pi-git:
This adds a few useful things.
One, punishmentTypes now support callbacks (that are called when a user has the punishment they're set for and has just logged in.) This should make it easier to _do_ things with custom punishments without extra code for a loginfilter that also checks if they have the type.
It also adds support for adding more properties to punishment types.
Lastly, it ensures that when a user has a custom global punishment, it doesn't randomly lock them (as that is frequently undesired behavor) and instead just calls the callback.

### DIFF
--- a/server/punishments.ts
+++ b/server/punishments.ts
@@ -58,13 +58,12 @@ export type Punishment = [string, ID | PunishType, number, string];
 
 /**
  * Info on punishment types. Can be used for either room punishment types or global punishments
- * extended by plugins. It can either be just a string (corresponding to the display in /alts),
- * or a [string, callback] - where the string is the /alts display, and the callback is what to do if the user
+ * extended by plugins. The `desc` is the /alts display, and the `callback` is what to do if the user
  * _does_ have the punishment (can be used to add extra effects in lieu of something like a loginfilter.)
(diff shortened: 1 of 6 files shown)


Zarel (approved):
Looks good! I like this.

server/punishments.ts
```diff
@@ -55,6 +55,19 @@ const GROUPCHAT_MONITOR_INTERVAL = 10 * 60 * 1000; // 10 minutes
...
+
+/**
+ * Info on punishment types. Can be used for either room punishment types or global punishments
+ * extended by plugins. It can either be just a string (corresponding to the display in /alts),
+ * or a [string, callback] - where the string is the /alts display, and the callback is what to do if the user
+ * _does_ have the punishment (can be used to add extra effects in lieu of something like a loginfilter.)
+ * Rooms will be specified for room punishments, otherwise it's null for global punishments
+ */
+export interface PunishInfo {
+	display: string;
```
Zarel, line 67: The main thing I don't like is `display`. I think `desc` would be clearer here.

mia-pi-git: Desc it is. I renamed it.

PR #8216, opened 2021-04-21
Make gmax chip damage occur in speed order

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-51997606

Would've liked to fix the other gmax chip damage related bugs, but this should do for now.


Marty-D:
Although this was my idea I don't particularly love it since it only takes care of the one issue and makes the others harder to fix. I'd like to know if @pyuk-bot's idea panned out or whether she thinks it's still feasible before committing to this solution for now.

Zarel:
One thing we could do is separate `onResidual` from `onSideResidual` and `onFieldResidual`.

Zarel:
This would require residual events to be even weirder than they already are, though.

DaWoblefet:
This was superseded by https://github.com/smogon/pokemon-showdown/pull/8222.

PR #8218, opened 2021-04-22
Use data/tags for validator

Zarel:
Previously, the validator had its own hardcoded tag system. This removes the hardcoding and makes it so any pokemon tag in `tags.ts` can be banned or unbanned.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -521,7 +521,7 @@ export const Formats: FormatList = [
 			'Garchomp', 'Gliscor', 'Greninja', 'Gyarados-Mega', 'Heatran', 'Kartana', 'Kommo-o', 'Kyurem', 'Landorus-Therian', 'Lopunny-Mega',
 			'Magnezone', 'Mawile-Mega', 'Medicham-Mega', 'Pelipper', 'Rillaboom', 'Scizor-Mega', 'Serperior', 'Slowbro-Base', 'Swampert-Mega',
 			'Tangrowth', 'Tapu Fini', 'Tapu Koko', 'Tapu Lele', 'Toxapex', 'Tyranitar', 'Victini', 'Volcarona', 'Zapdos-Base',
-			'NDUUBL', // National Dex UUBL
+			'ND UUBL', // National Dex UUBL
 			'Drizzle', 'Drought',
 			// Slowbronite is banned so it doesn't validate on Galarian Slowbro
 			'Slowbronite',
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -521,10 +521,8 @@ export const Formats: FormatList = [
 			'Garchomp', 'Gliscor', 'Greninja', 'Gyarados-Mega', 'Heatran', 'Kartana', 'Kommo-o', 'Kyurem', 'Landorus-Therian', 'Lopunny-Mega',
 			'Magnezone', 'Mawile-Mega', 'Medicham-Mega', 'Pelipper', 'Rillaboom', 'Scizor-Mega', 'Serperior', 'Slowbro-Base', 'Swampert-Mega',
 			'Tangrowth', 'Tapu Fini', 'Tapu Koko', 'Tapu Lele', 'Toxapex', 'Tyranitar', 'Victini', 'Volcarona', 'Zapdos-Base',
-			'nduubl', // National Dex UUBL
-			'Aerodactyl-Mega', 'Azumarill', 'Blacephalon', 'Diancie-Mega', 'Gallade-Mega', 'Gardevoir-Mega', 'Gengar', 'Gyarados', 'Hawlucha',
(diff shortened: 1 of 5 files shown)


config/formats.ts
```diff
@@ -521,10 +521,8 @@ export const Formats: FormatList = [
 			'Garchomp', 'Gliscor', 'Greninja', 'Gyarados-Mega', 'Heatran', 'Kartana', 'Kommo-o', 'Kyurem', 'Landorus-Therian', 'Lopunny-Mega',
 			'Magnezone', 'Mawile-Mega', 'Medicham-Mega', 'Pelipper', 'Rillaboom', 'Scizor-Mega', 'Serperior', 'Slowbro-Base', 'Swampert-Mega',
 			'Tangrowth', 'Tapu Fini', 'Tapu Koko', 'Tapu Lele', 'Toxapex', 'Tyranitar', 'Victini', 'Volcarona', 'Zapdos-Base',
-			'nduubl', // National Dex UUBL
-			'Aerodactyl-Mega', 'Azumarill', 'Blacephalon', 'Diancie-Mega', 'Gallade-Mega', 'Gardevoir-Mega', 'Gengar', 'Gyarados', 'Hawlucha',
-			'Heracross-Mega', 'Hydreigon', 'Latias', 'Latias-Mega', 'Latios', 'Manaphy', 'Pinsir-Mega', 'Slowbro-Mega', 'Thundurus',
-			'Venusaur-Mega', 'Xurkitree', 'Zapdos-Galar', 'Drizzle', 'Drought',
+			'NDUUBL', // National Dex UUBL
```
KrisXV, line 524:
```suggestion
			'ND UUBL', // National Dex UUBL
```
I don't really care which spacing is used, but this is just to keep it consistent with the `name` property in `tags.ts`

sim/dex-formats.ts
```diff
@@ -596,11 +611,23 @@ export class DexFormats {
...
+		this.getTagRules(ruleTable);
 
 		if (!repeals) format.ruleTable = ruleTable;
 		return ruleTable;
 	}
 
+	getTagRules(ruleTable: RuleTable) {
+		const tagRules = [];
+		for (const ruleid of ruleTable.keys()) {
+			if (ruleid.startsWith('+pokemontag:') || ruleid.startsWith('-pokemontag:') || ruleid.startsWith('*pokemontag:')) {
```
KrisXV, line 623:
```suggestion
			if (/^[+-*]pokemontag:/.test(ruleid)) {
```
Is there a reason to not use regex here? I don't know if so.

sim/dex-formats.ts
```diff
@@ -596,11 +611,23 @@ export class DexFormats {
...
+				if (ruleid.slice(12) === 'allpokemon') continue;
+				tagRules.push(ruleid);
+			}
+		}
+		ruleTable.tagRules = tagRules;
```
Zarel, line 628: Note to self: This should be sorted so obtainability rules are always highest priority.

`-Unobtainable, +Legendary` should not let you use Arceus in Gen 1, for instance – `-Unobtainable` must trump every rule that isn't an obtainability rule like `+Past`.

sim/dex-formats.ts
```diff
@@ -641,6 +668,12 @@ export class DexFormats {
 		}
 	}
 
+	validPokemonTag(tagid: ID) {
+		const tag = Tags.hasOwnProperty(tagid) && Tags[tagid];
+		if (!tag) return false;
+		return !!(tag.speciesFilter || tag.genericFilter);
+	}
```
KrisXV, line 675: Should there be a test for this? What's the point of having a property in `tags.ts` if it doesn't have one of these.

sim/dex-formats.ts
Zarel: It has to be `[+*-]` - `-` means something different when it's in the middle of a range. But otherwise yes, you're right, regex is better.

KrisXV: Oh fair, I just kinda typed swiftly

sim/dex-formats.ts
Zarel: `moveFilter` also exists. I also have `speciesNumCol` and `moveNumCol` iirc, which are planned to be used for dexsearch.

sim/dex-formats.ts
Zarel:
```suggestion
			if (/^[+*-]pokemontag:/.test(ruleid)) {
```

xInstruct:
I feel like we can do #8412 in a similar vein in this style, than just assigning an array that holds one value

xInstruct:
I feel like we can do #8142 in a similar vein in this style, than just assigning an array that holds one value

Zarel:
> I feel like we can do #8142 in a similar vein in this style, than just assigning an array that holds one value

Yeah I also wanted to have a new `tierTags` system. Kris wanted it too but gave up on it at some point. :/

Zarel:
I'd like to thank the academy, also my unit tests.

PR #8219, opened 2021-04-22
Fix interaction of Destiny Bond and Red Card

dhw0:
Fixed by checking the faint queue between the end of a multi-hit move and recoil damage, causing the effects of Destiny Bond to be applied before the effects of Red Card are applied. I'm not sure if evaluating the faint queue first will cause incorrect behavior (like ordering shenanigans with something like Soul Heart in doubles), but it seems to pass all the tests.

bug report: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-23#post-8585749
mechanics bugs card: https://github.com/smogon/pokemon-showdown/projects/3#card-52058755

### DIFF
--- a/test/sim/items/redcard.js
+++ b/test/sim/items/redcard.js
@@ -10,7 +10,7 @@ describe('Red Card', function () {
 		battle.destroy();
 	});
 
-	it.skip(`should not trigger if the target should be KOed from Destiny Bond and also not crash the client`, function () {
+	it(`should not trigger if the target should be KOed from Destiny Bond and also not crash the client`, function () {
 		battle = common.createBattle({gameType: 'doubles'}, [[
 			{species: "Aggron", item: 'redcard', moves: ['rockslide']},
 			{species: "Wynaut", ability: 'prankster', level: 1, moves: ['destinybond']},
--- a/test/sim/items/redcard.js
+++ b/test/sim/items/redcard.js
@@ -23,4 +23,18 @@ describe('Red Card', function () {
 		battle.makeChoices();
 		assert.holdsItem(battle.p1.active[0]);
 	});
+
+	it(`should trigger if the target is still in battle`, function () {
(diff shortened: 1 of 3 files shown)


test/sim/items/redcard.js
```diff
@@ -23,4 +23,18 @@ describe('Red Card', function () {
...
+			{species: "Aggron", item: 'redcard', moves: ['rockslide']},
+			{species: "Wynaut", ability: 'prankster', level: 1, moves: ['sleeptalk']},
+		], [
+			{species: "Conkeldurr", moves: ['sleeptalk']},
+			{species: "Gardevoir", moves: ['strugglebug']},
+			{species: "Corsola", moves: ['sleeptalk']},
+		]]);
+
+		battle.makeChoices();
+		assert.false(battle.p1.active[0].item, "red card should be used");
```
DaWoblefet, line 38:
```suggestion
		assert.false.holdsItem(battle.p1.active[0], "Red Card should be consumed.");
```

test/sim/items/redcard.js
```diff
@@ -23,4 +23,18 @@ describe('Red Card', function () {
...
+	});
```
DaWoblefet, line 39: I would add something to ensure that battle.p2.active[1] is Corsola; something like assert.species() should work

Marty-D:
If we're going by the UPC timing, shouldn't the faint queue be checked somewhere between `DamagingHit` and recoil damage? Doesn't make much sense to check it after those, and only when Sheer Force isn't involved.

DaWoblefet:
I'll bet moving this to the proper place actually ends up fixing a number of other post-faint bugs, haha

Marty-D (approved):
Looks good, thanks! 👍 

I'll merge when I have some free time to babysit the server for a while in case unexpected things happen and I have to revert. I, too, find all the tests passing to be a false sense of security, especially for "simple" one-liners like this.

PR #8220, opened 2021-04-22
Refactor everything to use Utils.sortBy

Zarel:
A few uses of `array.sort()` have been left alone:

- `array.sort((a, b) => b - a)[0]` to get the max of an array (this can be refactored to `Math.max(...array)` but would crash on large arrays - thanks, JavaScript)

- sorting in `data/` because they aren't supposed to import anything

- `set-importer` because I still have no clue what that's for and what dependencies it is/isn't allowed to have

- `sort()` with no arguments used as a lexical sort (at which point `sortBy` offers no benefits)

All other cases have been replaced with `Utils.sortBy`, which should be a massive increase in readability.

Sort orders should be much more readable now, without needing to puzzle through sign issues. The order is always low-to-high, A-to-Z, true-to-false.

### DIFF
--- a/server/chat-plugins/helptickets.ts
+++ b/server/chat-plugins/helptickets.ts
@@ -471,33 +471,18 @@ export function notifyStaff() {
 	const room = Rooms.get('staff');
 	if (!room) return;
 	let buf = ``;
-	const keys = Object.keys(tickets).sort((aKey, bKey) => {
-		const a = tickets[aKey];
-		const b = tickets[bKey];
-		if (a.offline) {
-			return (b.offline ? 1 : -1);
-		}
-		if (a.open !== b.open) {
-			return (a.open ? -1 : 1);
-		} else if (a.open && b.open) {
-			if (a.active !== b.active) {
-				return (a.active ? -1 : 1);
-			}
-			if (!!a.claimed !== !!b.claimed) {
-				return (a.claimed ? 1 : -1);
-			}
-			return a.created - b.created;
-		}
-		return 0;
-	});
+	const sortedTickets = Utils.sortBy(Object.values(tickets), ticket => [
+		!ticket.offline,
+		ticket.open,
+		ticket.open ? [ticket.active, !ticket.claimed, ticket.created] : 0,
+	]);
 	let count = 0;
 	let hiddenTicketUnclaimedCount = 0;
 	let hiddenTicketCount = 0;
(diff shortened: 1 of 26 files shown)


server/chat-plugins/helptickets.ts
```diff
@@ -868,27 +853,16 @@ export const pages: PageTable = {
...
-						return (a.active ? -1 : 1);
-					}
-					return a.created - b.created;
-				}
-				return b.created - a.created;
-			});
+			const sortedTickets = Utils.sortBy(Object.values(tickets), ticket => [
+				ticket.open,
+				ticket.open ? [ticket.active, ticket.created] : -ticket.created,
+			]);
```
Zarel, line 859: This is a good representative sample of the kind of code that `sortBy` massively simplifies and makes more readable.

I hope it's as easy for everyone else as it is for me to read this as "open active tickets by earliest first, then open inactive tickets by earliest first, then closed tickets by earliest last".

Zarel:
I think a big readability gain here is that you have to do a lot of thinking to tell the difference between `arr.sort((a, b) => a - b)` and `arr.sort((a, b) => b - a)` (which one is low-to-high? which one is high-to-low?)

But it's much easier to grasp that `Utils.sortBy(arr)` is low-to-high so `Utils.sortBy(arr, n => -n)` is high-to-low.

server/chat-plugins/mafia.ts
```diff
@@ -948,30 +939,24 @@ class Mafia extends Rooms.RoomGame {
...
-				if (l1.dir === 'up') return (l1.lastLynch < l2.lastLynch ? -1 : 1);
-				return (l1.lastLynch > l2.lastLynch ? -1 : 1);
-			}
-		});
-		this.hasPlurality = topLynches[0];
+		topLynches = Utils.sortBy(topLynches, ([key, lynch]) => [
+			lynch.dir === 'down',
+			lynch.dir === 'up' ? lynch.lastLynch : -lynch.lastLynch,
+		]);
+		[this.hasPlurality] = topLynches[0];
```
Zarel, line 959: Another complex sort order simplified by `sortBy`. "Down in order of high-to-low lastLynch, then up in order of low-to-high lastLynch".

Zarel:
(It's mostly the removal of `ipSort` that I think crashes hotpatch.)

PR #8224, opened 2021-04-24
Grassy terrain only weakens moves against grounded targets

urkerab:
It also doesn't weaken moves against semi-invulnerable targets, which means that users of Dig take regular doubled damage from Earthquake.


DaWoblefet:
Don't forget to update in Gen 7.

DaWoblefet:
Looks like the test itself is wrong, lol.

DaWoblefet:
Alright @urkerab I pushed updates to Grassy Terrain tests to master; you'll want to unskip the halving BP test in grassyterrain.js.

DaWoblefet:
Thanks urkerab!

PR #8227, opened 2021-04-25
Commands: Add /botmsg and inheritance to /noreply

sirDonovan:
This would be nice to have for command-heavy bot pages.


Zarel:
I'm not sure I like hiding what's going on there...

server/chat-commands/core.ts
```diff
@@ -539,7 +539,7 @@ export const commands: ChatCommands = {
 			return this.errorReply(this.tr`User ${targetUsername} is offline.`);
 		}
 
-		return this.parse(target);
+		return this.parse(target, this.isQuiet);
```
Zarel, line 542: Instead of passing `this.isQuiet` as an argument, should `this.parse` automatically inherit the `isQuiet` setting of the parent context?

sirDonovan: A potential issue with that would be hiding the help text if a command were used incorrectly.

Zarel: Fair enough. I think it should at least be an option argument rather than a positional argument.

sirDonovan: Updated.

Zarel:
Although perhaps the correct answer is for `/help` and error messages to bypass `/noreply`?

Zarel:
I'm also somewhat afraid this would get abused to trick people into PMing other users bad things.

Some sort of prefix like `[scripted bot message]` or some check for a bot rank might be necessary here.

Zarel:
Perhaps a different command entirely, like `/botcmd`

server/chat.ts
```diff
@@ -444,11 +444,11 @@ export class CommandContext extends MessageContext {
 	}
 
 	// TODO: return should be void | boolean | Promise<void | boolean>
-	parse(msg?: string, quiet?: boolean): any {
+	parse(msg?: string, options?: {isQuiet?: boolean}): any {
 		if (typeof msg === 'string') {
 			// spawn subcontext
 			const subcontext = new CommandContext(this);
-			if (quiet) subcontext.isQuiet = true;
+			if (options && options.isQuiet) subcontext.isQuiet = true;
```
mia-pi-git, line 451:
```suggestion
			if (options?.isQuiet) subcontext.isQuiet = true;
```

sirDonovan:
`CommandContext` now inherits `isQuiet` but bypasses it for `/help` and error replies. I also added a check for bot rank when a button contains `/noreply /msg`.

Zarel:
I think I'd prefer `/botmsg` to `/noreply msg`. I don't think of showing the sent PM as a "reply" exactly.

sirDonovan:
Updated to `/botmsg` and also removed the `/noreply` bypass for errors as that's intended for things like https://github.com/smogon/pokemon-showdown-client/pull/1600

Zarel:
This part doesn't seem to have been implemented, and it also seems like the most important part:

> I'm also somewhat afraid this would get abused to trick people into PMing other users bad things.
> 
> Some sort of prefix like `[scripted bot message]` or some check for a bot rank might be necessary here.

The check for a bot rank is probably best.

Zarel:
Oh, I have a pretty simple fix for this: Require `/botmsg` to be used in the room the bot is auth of.

PR #8229, opened 2021-04-25
[Gen 8] Pokebilities: Block unsafe pseudo-abilities when run under a different mod

WeWuzNidokangz:
Prior to this change, I was able to replicate the expected problem where Trace activated multiple times (not 100% but often) with these sets:-

Side A:

Porygon2 @ Choice Band  
Ability: Analytic  
EVs: 252 HP / 4 SpA / 252 SpD  
Calm Nature  
IVs: 0 Atk  
- Recover  
- Discharge  
- Foul Play  
- Rest

Side B:

Clefable @ Leftovers  
Ability: Magic Guard  
EVs: 252 HP / 152 Def / 104 SpD  
Bold Nature  
IVs: 0 Atk  
- Wish  
- Moonblast  
- Thunder Wave  
- Protect

Challenge code:

/challenge [Gen 8] OU @@@[Gen 8] Pokebilities

The change prevents the problematic abilities from being added only when the battle's mod is not 'pokebilities'.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -646,6 +646,8 @@ export const Formats: FormatList = [
 			'Diglett-Base', 'Dugtrio-Base', 'Gothita', 'Gothitelle', 'Gothorita', 'Trapinch', 'Wobbuffet', 'Wynaut',
 		],
 		onBegin() {
+			const isNativeMod = this.dex.currentMod === 'pokebilities';
+			const nonNativeUnsafeAbilities = ['trace', 'mirrorarmor'];
 			for (const pokemon of this.getAllPokemon()) {
 				if (pokemon.ability === this.toID(pokemon.species.abilities['S'])) {
 					continue;
@@ -653,7 +655,8 @@ export const Formats: FormatList = [
 				pokemon.m.innates = Object.keys(pokemon.species.abilities)
 					.filter(key => key !== 'S' && (key !== 'H' || !pokemon.species.unreleasedHidden))
 					.map(key => this.toID(pokemon.species.abilities[key as "0" | "1" | "H" | "S"]))
-					.filter(ability => ability !== pokemon.ability);
+					.filter(ability => (ability !== pokemon.ability) &&
+						(isNativeMod || !nonNativeUnsafeAbilities.includes(ability)));
 			}
 		},
 		onSwitchInPriority: 2,


Zarel:
Is this worth it?

Is the problem that this causes a crash or something?

Zarel:
Is it desirable to run pokebilities under a different mod? Perhaps the validator should just error out with "You must use this mod with this format".

KrisXV:
@zarel Similar to Shared Power, this causes a Stack Overflow if its added onto randbats or something. I'd rather just add a format property or something that prevents certain formats/rulesets from being added onto certain `customRules` things in challenges/tours.

Zarel:
We could just hardcode not allowing formats with different mods to be inherited from at all.

KrisXV:
> We could just hardcode not allowing formats with different mods to be inherited from at all.

I disagree with this because people like to put camomons onto different old gen formats for example

Zarel:
> I disagree with this because people like to put camomons onto different old gen formats for example

Then make it a rule! Is random crashes really the solution here?

WeWuzNidokangz:
In the case of Camomons, I think making the battle effects into a rule would be a great idea, but it seems as if it might be considerably more complex to try to do that with Pokebilities. I don't know if Showdown supports allowing rules to specify things like ModdedAbilityData in the way that the mods of these formats do. Even if that could be supported, it seems like it would raise a similar situation at a different junction because it is possible that a battle could have multiple active rules trying to define potentially conflicting e.g. ModdedAbilityData at the same time, and it seems like one winner would have to picked somehow.

Regarding use cases, I'm sorry that the challenge code I gave above isn't a good example, as it was just meant as an easy way to test this situation. A more realistic use case that users might try would be adding the Pokebilities battle effects onto something like Random Battle to play a randoms game where each Pokemon has multiple abilities. Many users greatly appreciate Showdown providing this kind of functionality, e.g. for this forum tour that used Shared Power in a very similar way:-

https://www.smogon.com/forums/threads/random-battles-mayhem-gauntlet-won-by-longhiep341-see-post-228.3679686/

I thought Kris's solution for Shared Power (https://github.com/smogon/pokemon-showdown/commit/ebf13740f4e45f3d4e9eee3a6004fd3304579c9e) was a very well-considered compromise in maintaining this kind of functionality at the same time as adding a mod to fine-tune some abilities in the base OM, so I hoped that something similar could be considered for Pokebilities.

Zarel:
After some testing, it seems like `effect.name === "Intimidate"` is going to be better than `effect.id === 'intimidate' || effect.id === 'ability:intimidate'`.

I'd accept that sort of change into `data/`, so you don't need to maintain mods for changes like that.

WeWuzNidokangz:
Oh...you mean e.g. changing data/abilities.ts in ways like this?

```
// Don't bounce self stat changes, or boosts that have already bounced
			if (target === source || !boost || effect.id === 'mirrorarmor') return;
```
=>
```
// Don't bounce self stat changes, or boosts that have already bounced
			if (target === source || !boost || effect.name === "mirrorarmor") return;
```

That seems like it could prevent Mirror Armor from being able to overflow (as well as likely being a fix for a lot of smaller issues).

However, in the case of Trace, it seems like a larger change might be needed to achieve similar functionality without relying on a mod. In particular it seems like there would need to be some way of removing the "ability:trace" volatile instead of replacing the Pokemon's real ability if the volatile remaining is the cause of the overflow...

Zarel:
> That seems like it could prevent Mirror Armor from being able to overflow (as well as likely being a fix for a lot of smaller issues).

Yeah, that was my idea.

> However, in the case of Trace, it seems like a larger change might be needed to achieve similar functionality without relying on a mod. In particular it seems like there would need to be some way of removing the "ability:trace" volatile instead of replacing the Pokemon's real ability if the volatile remaining is the cause of the overflow...

I would probably accept this into the main simulator. If the engine supports `addVolatile("ability:trace")` then the engine should support that not causing crashes.

WeWuzNidokangz:
Thank-you for being so accommodating. I would like to try to submit a change like that if there is a good chance of it being accepted because it would likely solve multiple other issues along with this, as well as pre-emptively protect against similar problems occurring in the likely case that Partners In Crime (another OM in which Pokemon can have multiple abilities simulataneously) becomes OMotM/LCotM in the near future.

If am able to make a working change where the default data supports volatile pseudo-abilities and the battle effects of Pokebilities and Shared Power become rules, should I try to submit it as a different PR, and only fall back to this one if the larger change can't be accepted?

Zarel:
I think a new PR would be appropriate.

AnnikaCodes:
I'm closing this since #8254 is currently underway to rulify Pokébilities.

PR #8233, opened 2021-04-26
Fix Shell Side Arm information leak

DaWoblefet:
Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8833264


Zarel:
ty!

PR #8234, opened 2021-04-26
Rooms : Fix read-more in crash report

Distrib-ps:
Before:

After:


Zarel:
Unfortunately I have a better fix in the server. It's currently uncommitted because I was holding off on testing it until next restart.

Distrib-ps:
> Unfortunately I have a better fix in the server. It's currently uncommitted because I was holding off on testing it until next restart.

Ok I didn't know.  I close this pr.

PR #8235, opened 2021-04-26
Give the negative operator a precedence

urkerab:
Reported [here](https://www.smogon.com/forums/posts/8835883). I also gave `(` a precedence to simplify the loop.


Zarel:
ty!

PR #8237, opened 2021-04-27
Support setting a default format/mod for commands in rooms

Zarel:
Affected commands: `/data`, `/coverage`, `/effectiveness`, `/weakness`, `/learn`

Supersedes #7965

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -543,10 +543,10 @@ export const commands: Chat.ChatCommands = {
 		const gen = parseInt(cmd.substr(-1));
 		if (gen) target += `, gen${gen}`;
 
+		const {dex, format, targets} = this.splitFormat(target, true);
+
 		let buffer = '';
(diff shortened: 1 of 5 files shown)


Zarel:
@mia-pi-git you didn't understand why I wanted to support setting a default format as well as a default mod.

The reason is because a lot of commands work differently in different formats. For instance:

- Setting the default format to VGC would make `/learn` exclude past-gen learn methods
- Setting the default format to Tier Shift would make `/data` show Tier Shift stats

Zarel:
@KrisXV do you think `dataCommandTierDisplay` should be folded into this?

DaWoblefet:
Being able to use /learn etc. for VGC purposes at all would be awesome!

KrisXV:
> @KrisXV do you think `dataCommandTierDisplay` should be folded into this?

probably, but this wouldnt be accurate for national dex anymore; we prefer the numbers

PR #8238, opened 2021-04-27
Handle extraneous parentheses again

urkerab:
The calculator used to ignore extra `)`s but PR #8235 broke that. (Sorry about that.)


Zarel:
Can you explain what you mean? Probably ignoring extra `)` is a good thing since that sounds like a typo, right?

Zarel:
(I'm merging in case I'm misinterpreting and this is an important regression.)

urkerab:
With the old calculator, you could type something like `1+1)*2` and still get 4, and I broke that.

Zarel:
Oh, neat.

PR #8239, opened 2021-04-27
Sockets: Add a configurable socket validator

mia-pi-git:


server/sockets.ts
```diff
@@ -54,6 +54,16 @@ export const Sockets = new class {
 				break;
 			}
 
+			case '&': {
+				// roomid\nmessage - message to a roomid from the child process
```
Zarel, line 58: How does this differ from `<`?

Zarel:
```suggestion
				// &roomid, pline
				// forcibly add a pline directly to a room ID
```

server/sockets.ts
```diff
@@ -54,6 +54,14 @@ export const Sockets = new class {
...
+				const [roomid, message] = Utils.splitFirst(message, '\n');
+				const room = Rooms.get(roomid);
+				if (room) room.add(message).update();
```
Zarel, line 62:
```suggestion
				Rooms.get(roomid)?.add(message).update();
```

server/sockets.ts
```diff
@@ -54,6 +54,14 @@ export const Sockets = new class {
 				break;
 			}
 
+			case '&': {
+				// &roomid, pline
+				// forcibly add a pline directly to a room ID
```
Zarel, line 59:
```suggestion
				// forcibly add a pline directly to a room ID
				// (intended for logging)
```

server/sockets.ts
```diff
@@ -495,6 +504,11 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 				}
 			}
 		}
+		if (Config.validateSocket && !Config.validateSocket.call(this, socket, socketip)) {
```
Zarel, line 507:
```suggestion
		if (Config.validateSocket && Config.validateSocket.call(this, socket, socketip) === false) {
```

Probably `undefined` should not force a socket to be deleted.

server/sockets.ts
```diff
@@ -495,6 +504,11 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
...
+			socket.destroy();
+			this.sockets.delete(socketid);
+			return;
+		}
```
Zarel, line 511: I'm... still not sure this hook is necessary. I _guess_ it can't hurt.

server/sockets.ts
```diff
@@ -54,6 +54,15 @@ export const Sockets = new class {
 				break;
 			}
 
+			case '&': {
+				// &roomid, pline
+				// forcibly add a pline directly to a room ID
+				// (intended for logging)
+				const [roomid, message] = Utils.splitFirst(message, '\n');
```
Zarel, line 61:
```suggestion
				const [roomid, message] = Utils.splitFirst(data, '\n');
```

server/sockets.ts
```diff
@@ -495,6 +504,11 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 				}
 			}
 		}
+		if (Config.validateSocket && Config.validateSocket.call(this, socket, socketip) === false) {
+			socket.destroy();
+			this.sockets.delete(socketid);
+			return;
+		}
```
Zarel, line 511: On further consideration, I still want this removed. This needs to be implemented in the main process, not here.

```suggestion
```

Zarel: (The logger can remain; it has plenty of other uses for debugging.)

PR #8240, opened 2021-04-28
Implement /botmsg

Zarel:
Supersedes #8227

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -433,7 +433,7 @@ export const commands: Chat.ChatCommands = {
 		target = this.checkChat(target);
 		Chat.sendPM(`/botmsg ${target}`, user, targetUser, targetUser);
 	},
-	msghelp: [`/botmsg [username], [message] - Send a private message to a bot without feedback. For room bots, must use in the room the bot is auth in.`],
+	botmsghelp: [`/botmsg [username], [message] - Send a private message to a bot without feedback. For room bots, must use in the room the bot is auth in.`],
 
 	nick() {
 		this.sendReply(`||New to the Pokémon Showdown protocol? Your client needs to get a signed assertion from the login server and send /trn`);
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -415,6 +415,26 @@ export const commands: Chat.ChatCommands = {
 		`If a [highlight] is specified, only highlights them if they have that term on their highlight list.`,
 	],
 
+	botmsg(target, room, user, connection) {
+		if (!target || !target.includes(',')) return this.parse('/help botmsg');
+		target = this.splitTarget(target);
+		const targetUser = this.targetUser;
+		const targetUsername = this.targetUsername;
+
+		if (!targetUser || !targetUser.connected) {
+			return this.popupReply(`The bot "${targetUsername}" is offline.`);
+		}
+
+		const auth = this.room ? this.room.auth : Users.globalAuth;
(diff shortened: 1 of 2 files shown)


Zarel:
@sirDonovan does this work for your use-case? The `/botmsg` at the beginning of the message sent to the bot is one way for bots to tell that it's not a privileged command, and also in case it wants to know that it came from a `/botmsg` command, but if it's inconvenient I can remove it.

server/chat-commands/admin.ts
```diff
@@ -415,6 +415,26 @@ export const commands: Chat.ChatCommands = {
 		`If a [highlight] is specified, only highlights them if they have that term on their highlight list.`,
 	],
 
+	botmsg(target, room, user, connection) {
+		if (!target || !target.includes(',')) return this.parse('/help botmsg');
+		target = this.splitTarget(target);
+		const targetUser = this.targetUser;
+		const targetUsername = this.targetUsername;
```
KrisXV, line 422: Would it make more sense to make `splitTarget` return these two variables (or something like `{newTarget, targetUser, targetUsername}`), similar to how datsearch's `getMod` function returns the mod and adjusted target, or similar to how `extractFormat()` is used? Would probably be for another PR, but it's used this way in a lot of other commands.

server/chat.ts
```diff
@@ -1292,19 +1292,22 @@ export class CommandContext extends MessageContext {
 					if ((!this.room || this.room.settings.isPersonal || this.room.settings.isPrivate === true) && !this.user.can('lock')) {
 						const buttonName = / name ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
 						const buttonValue = / value ?= ?"([^"]*)"/i.exec(tagContent)?.[1];
-						const msgCommandRegex = /^\/(?:msg|pm|w|whisper) /i;
+						const msgCommandRegex = /^\/(?:msg|pm|w|whisper|botmsg) /;
+						const botmsgCommandRegex = /^\/msgroom (?:[a-z0-9-]+), \/botmsg /;
```
KrisXV, line 1296:
```suggestion
						const botmsgCommandRegex = /^\/msgroom (?:[a-z0-9-]+), ?\/botmsg /;
```

server/chat-commands/admin.ts
Zarel: You mean `splitFormat`, and yes, I think `splitTarget` sucks and should be refactored.

Zarel: We can call it `splitUser`.

sirDonovan:
@Zarel this works for me, thanks!

server/chat-commands/admin.ts
```diff
@@ -415,6 +415,26 @@ export const commands: Chat.ChatCommands = {
...
+
+		const auth = this.room ? this.room.auth : Users.globalAuth;
+		if (auth.get(targetUser) !== '*') {
+			return this.popupReply(`The user "${targetUsername}" is not a bot in this room.`);
+		}
+
+		target = this.checkChat(target);
+		Chat.sendPM(`/botmsg ${target}`, user, targetUser, targetUser);
+	},
+	msghelp: [`/botmsg [username], [message] - Send a private message to a bot without feedback. For room bots, must use in the room the bot is auth in.`],
```
sirDonovan, line 436:
```suggestion
	botmsghelp: [`/botmsg [username], [message] - Send a private message to a bot without feedback. For room bots, must use in the room the bot is auth in.`],
```

Zarel:
Nice!

PR #8241, opened 2021-04-28
Rule: Add No Switch

Distrib-ps:
Pr for https://www.smogon.com/forums/threads/no-switch-rule.3653057/

For this rule, pokemon cannot switch normally.
But he can switch with an attack (teleport, baton pass)...

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1361,9 +1361,11 @@ export const Rulesets: {[k: string]: FormatData} = {
 		effectType: 'Rule',
 		name: 'No Switch',
 		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
+		onBegin() {
+			this.add('rule', 'No Switch : Players cannot switch their Pok&eacute;mon');
+		},
 		onTrapPokemon(pokemon) {
 			pokemon.trapped = true;
-			this.add('rule', '-activate', pokemon, 'trapped');
 		},
 	},
 };
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1357,7 +1357,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Bans move combinations on Pok\u00e9mon that weren't legal in Japanese versions of Gen 1.",
 		// Implemented in mods/gen1jpn/rulesets.ts
 	},
-	noswitch: {
+	noswitching: {
 		effectType: 'Rule',
 		name: 'No Switching',
 		desc: 'All Pok\u00e9mon are trapped (cannot switch naturally, but can as the effect of an item, move, or Ability).',
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1360,7 +1360,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 	noswitch: {
 		effectType: 'Rule',
 		name: 'No Switch',
-		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
+		desc: 'All Pok\u00e9mon are trapped (cannot switch naturally, but can as the effect of an item, move, or Ability).',
 		onBegin() {
 			this.add('rule', 'No Switch: All Pok\u00e9mon are trapped');
 		},
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1360,7 +1360,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 	noswitch: {
 		effectType: 'Rule',
 		name: 'No Switch',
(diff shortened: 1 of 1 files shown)


data/rulesets.ts
```diff
@@ -1357,4 +1357,13 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Bans move combinations on Pok\u00e9mon that weren't legal in Japanese versions of Gen 1.",
 		// Implemented in mods/gen1jpn/rulesets.ts
 	},
+	noswitch: {
+		effectType: 'Rule',
+		name: 'No Switch',
+		desc: 'It is not possible to switch Pok&eacute;mon. Except with an item, moove or Ability provided for this purpose.',
```
DaWoblefet, line 1363:
```suggestion
		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
```

DaWoblefet:
How does this interact with Ghost-types and Shed Shell ignoring trapping effects? Should they?

Distrib-ps:
> How does this interact with Ghost-types and Shed Shell ignoring trapping effects? Should they?
Ghost-types not ignoring effects. But Shed Shell ignoring trapping effects. I don’t know if we should ban  the switch Ability/moves/items of switch.

xInstruct:
> How does this interact with Ghost-types and Shed Shell ignoring trapping effects? Should they?

I 100% think this should be Gen 4 Ingrain trapping. because when I think of a No Switch Mod, I think the button to switch should be hidden entirely.

data/rulesets.ts
```diff
@@ -1357,4 +1357,13 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Bans move combinations on Pok\u00e9mon that weren't legal in Japanese versions of Gen 1.",
 		// Implemented in mods/gen1jpn/rulesets.ts
 	},
+	noswitch: {
+		effectType: 'Rule',
+		name: 'No Switch',
+		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
+		onTrapPokemon(pokemon) {
+			pokemon.trapped = true;
+			this.add('rule', '-activate', pokemon, 'trapped');
```
Zarel, line 1366: Uh this is going to look super ugly and also be very spammy (appearing for every pokemon every turn)

Distrib-ps: Done, thx.

data/rulesets.ts
```diff
@@ -1357,4 +1357,15 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Bans move combinations on Pok\u00e9mon that weren't legal in Japanese versions of Gen 1.",
 		// Implemented in mods/gen1jpn/rulesets.ts
 	},
+	noswitch: {
+		effectType: 'Rule',
+		name: 'No Switch',
```
Zarel, line 1362: Perhaps "No Switching" would be a better name?

Zarel: I'll ask.

data/rulesets.ts
```diff
@@ -1357,4 +1357,15 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
+		onBegin() {
+			this.add('rule', 'No Switch : Players cannot switch their Pok&eacute;mon');
```
Zarel, line 1365:
```suggestion
			this.add('rule', 'No Switch: All Pok\u00e9mon are trapped');
```

Zarel: Let me guess, you're French? Is that the language that has the space before `:`? Anyway, in English, we don't.

Distrib-ps: > Let me guess, you're French? Is that the language that has the space before `:`? Anyway, in English, we don't.

Yes! Absolutely. I didn’t know.

data/rulesets.ts
```diff
@@ -1357,4 +1357,15 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Bans move combinations on Pok\u00e9mon that weren't legal in Japanese versions of Gen 1.",
 		// Implemented in mods/gen1jpn/rulesets.ts
 	},
+	noswitch: {
+		effectType: 'Rule',
+		name: 'No Switch',
+		desc: 'Players cannot switch their Pok&eacute;mon, except with an item, move, or Ability.',
```
Zarel, line 1363:
```suggestion
		desc: 'All Pok\u00e9mon are trapped (cannot switch naturally, but can as the effect of an item, move, or Ability).',
```

Zarel:
image

Distrib-ps:
Done @Zarel

Zarel:
Thanks!

PR #8244, opened 2021-04-28
Punishments: Cflood ban IPs with too many concurrent connections 

mia-pi-git:


HoeenCoder:
While not a bad idea, testing indicates this isn't currently the issue.

server/monitor.ts
```diff
@@ -154,7 +154,26 @@ export const Monitor = new class {
...
 		IPTools.dnsblCache.clear();
 	}
+	/**
+	 * Check to see how many connections share this IP.
+	 * Returns true if it should be terminated for having too many concurrent connections.
+	 */
+	countConcurrentConnections(ip: string) {
+		const cfloodMax = Punishments.sharedIps.has(ip) ? 50 : (Config.maxconns || 30);
+		let count = 0;
+		for (const conn of Users.connections.values()) {
```
AnnikaCodes, line 164: Iterating over 10,000+ connections every time a new user connects doesn't seem like great performance, especially if this isn't the issue. I don't mind trying this (if it proves to be a performance issue we could use a Map of IPs:number to connections or something), but it doesn't seem like it would be too useful (currently, no IP addresses have more than eight connections).

mia-pi-git: OK.

PR #8245, opened 2021-04-29
SSB: Fix fart's signature move from refactor

xInstruct:


data/mods/ssb/moves.ts
```diff
@@ -3415,7 +3415,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			this.add('-anim', source, "Trick", target);
 		},
 		onHit(target, source, effect) {
-			this.add(`c|${getName('Perish Song')}|/html <img src="https://i.imgflip.com/3rt1d8.png" />`);
+			this.add(`c|${getName('Perish Song')}|/html <img src="https://i.imgflip.com/3rt1d8.png" height=50% width=50%/>`);
```
Zarel, line 3418: um isn't this 50% of the screen, not 50% of the original image size?

Zarel: Probably just specify exact numbers

xInstruct: idk, it looks fine on my screen when I tested it. But if you want i can still do it.

Zarel: Yes, please. It's important for standards compliance and screen deformation, and has bugs the first time it's downloaded (which you might not notice because you've already downloaded it before).

```suggestion
			this.add(`c|${getName('Perish Song')}|/html <img src="https://i.imgflip.com/3rt1d8.png" width="262" height="150" />`);
```

Zarel: (This is the same reason why we require width/height in images in roomintros and `!showimage` before its refactor.)

Zarel:
ty!

PR #8246, opened 2021-04-29
Refactor splitTarget to splitUser

KrisXV:


translations/german/helptickets.ts
```diff
@@ -139,7 +139,7 @@ export const translations: Translations = {
 		"Help Ticket": "Help-Ticket",
 		"Issue": "Problem",
 		"A Global Staff member will be with you shortly.": "Ein globales Staffmitglied wird gleich da sein.",
-		"${this.inputUsername} does not have an open ticket.": "${this.inputUsername} besitzt kein aktives Ticket.",
+		"${inputUsername} does not have an open ticket.": "${inputUsername} besitzt kein aktives Ticket.",
```
AnnikaCodes, line 142: You don't have to change the stuff in the `${}`; template strings just work positionally AFAICT. (I tested this in <<croutons>> if you want to see what I mean.)

KrisXV: O

Zarel: They look better without the `this.` anyway.

Zarel: They look better without the `this.` anyway.

server/chat-commands/admin.ts
```diff
@@ -275,19 +275,18 @@ export const commands: Chat.ChatCommands = {
...
-		const targetUser = this.targetUser!;
-		this.checkPMHTML(targetUser);
+		const {targetUser, targetUsername, rest} = this.splitUser(target);
+		this.checkHTML(rest);
+		this.checkPMHTML(targetUser, targetUsername);
 
 		// Apply the infobox to the message
 		target = `/raw <div class="infobox">${target}</div>`;
-		const message = `|pm|${user.getIdentity()}|${targetUser.getIdentity()}|${target}`;
+		const message = `|pm|${user.getIdentity()}|${targetUser!.getIdentity()}|${target}`;
```
Zarel, line 284: Instead of needing `targetUser!`, we could have a `requireUser` which automatically throws `Chat.ErrorMessage` if the user doesn't exist.

I bet it would save a lot of code.

server/chat-plugins/wifi.ts
```diff
@@ -844,32 +844,30 @@ const cmds: Chat.ChatCommands = {
...
 		}
 
-		Giveaway.ban(room, targetUser, target);
+		Giveaway.ban(room, targetUser, rest);
 		if (room.giveaway) room.giveaway.kickUser(targetUser);
-		this.modlog('GIVEAWAYBAN', targetUser, target);
-		if (target) target = ` (${target})`;
+		this.modlog('GIVEAWAYBAN', targetUser, rest);
+		if (rest) target = ` (${rest})`;
 		this.privateModAction(`${targetUser.name} was banned from entering giveaways by ${user.name}.${target}`);
```
Zarel, line 860: This is likely a bug. I'm guessing you used `target` instead of `rest` because `rest` is a const. But that introduces a bug when `rest` is empty.

(The correct fix would be to use `let` instead of `const`. It's fine if you don't change the other values!)

KrisXV: idk what i did here, i definitely didnt intend to approach it like that

server/chat-commands/admin.ts
```diff
@@ -277,7 +277,7 @@ export const commands: Chat.ChatCommands = {
 
 		const {targetUser, rest: html} = this.requireUser(target);
 		this.checkHTML(html);
-		this.checkPMHTML(targetUser);
+		this.checkPMHTML(targetUser, targetUser.name);
```
Zarel, line 280: `targetUser.name` seems unnecessary here.

server/chat-commands/core.ts
```diff
@@ -633,9 +633,7 @@ export const commands: Chat.ChatCommands = {
 			this.errorReply(this.tr`You forgot the comma.`);
 			return this.parse('/help msg');
 		}
-		target = this.splitTarget(target);
-		const targetUser = this.targetUser;
-		const targetUsername = this.targetUsername;
+		const {targetUser, targetUsername, rest} = this.splitUser(target);
```
Zarel, line 636: `message`

Zarel:
Neat, thanks!

PR #8248, opened 2021-04-29
Mirror Coat stalwart doubles target fix

pacmanboss256:
I don't know what to say here....

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2770,9 +2770,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		onModifyMovePriority: 1,
 		onModifyMove(move, pokemon, target) {
 			// most of the implementation is in Battle#getTarget
-			if (!(move.target === 'scripted')) {
-				move.tracksTarget = true;
-			}
+			move.tracksTarget = move.target !== 'scripted';
 		},
 		name: "Propeller Tail",
 		rating: 0,
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2769,12 +2769,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	propellertail: {
 		onModifyMovePriority: 1,
 		onModifyMove(move, pokemon, target) {
+			// most of the implementation is in Battle#getTarget
 			if (!(move.target === 'scripted')) {
 				move.tracksTarget = true;
 			}
 		},
-		// the actual implementation is in Battle#getTarget
-		// move.tracksTarget is not actually useless
 		name: "Propeller Tail",
 		rating: 0,
 		num: 239,
@@ -3519,12 +3518,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	stalwart: {
 		onModifyMovePriority: 1,
 		onModifyMove(move, pokemon, target) {
+			// most of the implementation is in Battle#getTarget
 			if (!(move.target === 'scripted')) {
 				move.tracksTarget = true;
 			}
 		},
-		// the actual implementation is in Battle#getTarget
-		// move.tracksTarget is not actually useless
 		name: "Stalwart",
 		rating: 0,
 		num: 242,
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3514,11 +3514,6 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 100,
 	},
 	stalwart: {
-		onModifyMove(move) {
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
-		},
 		name: "Stalwart",
 		rating: 0,
 		num: 242,
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2769,7 +2769,6 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	propellertail: {
 		onModifyMovePriority: 1,
 		onModifyMove(move, pokemon, target) {
-			// this works with every test....
 			if (!(move.target === 'scripted')) {
 				move.tracksTarget = true;
 			}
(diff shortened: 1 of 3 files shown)


pacmanboss256:
Apparently the `onModifyMove` function in the stalwart and propeller tail ability code did something that messed up the target. deleting the functions there fixes it.

DaWoblefet:
lol nice

DaWoblefet:
Alright @pacmanboss256, it looks like removing ``move.tracksTarget = true;`` causes Stalwart to break with Follow Me and Storm Drain.

pacmanboss256:
that line said it was useless and commented out, so i wonder if i can find a fix for that

DaWoblefet:
Given that it's passing with Ally Switch and the switch-out switch-back-in test, my guess is that the redirection check for Storm Drain, Follow Me, etc. is overriding something earlier than Stalwart can jump in. When you look it over, remember that Dragon Darts is "smartTarget" and Stalwart etc. is "tracksTarget".

DaWoblefet:
I _really_ don't like hardcoding Counter and Mirror Coat here. You're going to encounter the same issue with Metal Burst, probably Magic Coat (idk what that even is supposed to do), and just hardcoding in certain moves doesn't seem like the solution. I would guess that modifying something in ``getTarget()`` in battle.ts, ``useMove()`` in script.ts (particular around the time the ``ModifyMove`` event is ran) or potentially both.

If we did want to use the hardcode solution, at least make sure you hardcode it in both places.

pacmanboss256:
i think in this case something with stalwart using the move modified the target AFTER counter/mirrorcoat modified the target to the correct one, which set target to undefined (?) and ended up random, since the target redirect is before the move is used and modified. 
metalburst, counter and mirror coat already track the target to be the last one that dealt damage, so the target isn't selected in the first place.

DaWoblefet:
Hmm, I can get behind this, since it's no longer hardcoding the moves themselves.

data/abilities.ts
```diff
@@ -2767,11 +2767,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 232,
 	},
 	propellertail: {
-		onModifyMove(move) {
```
DaWoblefet, line 2770: You need to make sure the new Stalwart code is in Propeller Tail too.

pacmanboss256: oh right durrr
my bad

data/abilities.ts
```diff
@@ -3514,11 +3510,15 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 100,
 	},
 	stalwart: {
-		onModifyMove(move) {
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
+		onModifyMovePriority: 1,
+		onModifyMove(move, pokemon, target) {
+			// this works with every test....
```
DaWoblefet, line 3515: I'd remove this comment, it doesn't add anything

data/abilities.ts
```diff
@@ -3514,11 +3510,15 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
+			if (!(move.target === 'scripted')) {
+				move.tracksTarget = true;
+			}
 		},
+		// the actual implementation is in Battle#getTarget
+		// move.tracksTarget is not actually useless
```
DaWoblefet, line 3521: I'd remove this comment, it doesn't add anything

data/abilities.ts
```diff
@@ -3514,11 +3517,14 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
+		onModifyMovePriority: 1,
+		onModifyMove(move, pokemon, target) {
+			if (!(move.target === 'scripted')) {
+				move.tracksTarget = true;
+			}
 		},
+		// the actual implementation is in Battle#getTarget
```
DaWoblefet, line 3526: I'd change this to something like "most of the implementation is done in Battle#getTarget". Move the comment inside of onModifyMove (above the if statement).

DaWoblefet:
Make sure the comments in both Stalwart and Propeller Tail are consistent.

DaWoblefet (approved):

data/abilities.ts
```diff
@@ -2767,10 +2767,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
-		onModifyMove(move) {
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
+		onModifyMovePriority: 1,
+		onModifyMove(move, pokemon, target) {
+			// most of the implementation is in Battle#getTarget
+			if (!(move.target === 'scripted')) {
+				move.tracksTarget = true;
+			}
```
Marty-D, line 2775:
```suggestion
			move.tracksTarget = move.target !== 'scripted';
```

data/abilities.ts
```diff
@@ -3514,10 +3516,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
...
```
Marty-D, line 3524:
```suggestion
			move.tracksTarget = move.target !== 'scripted';
```

data/abilities.ts
```diff
@@ -2767,10 +2767,10 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 232,
 	},
 	propellertail: {
-		onModifyMove(move) {
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
+		onModifyMovePriority: 1,
+		onModifyMove(move, pokemon, target) {
```
Marty-D, line 2771:
```suggestion
		onModifyMove(move) {
```

data/abilities.ts
```diff
@@ -3514,10 +3514,10 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 100,
 	},
 	stalwart: {
-		onModifyMove(move) {
-			// this doesn't actually do anything because ModifyMove happens after the tracksTarget check
-			// the actual implementation is in Battle#getTarget
-			move.tracksTarget = true;
+		onModifyMovePriority: 1,
+		onModifyMove(move, pokemon, target) {
```
Marty-D, line 3518:
```suggestion
		onModifyMove(move) {
```

Marty-D (approved):
👍

PR #8250, opened 2021-04-29
Wait for next time damage is taken to activate Gluttony after Neutralizing Gas leaves the field.

Karthik99999:


Karthik99999:
should I add a test for gluttony working properly?

DaWoblefet:
> should I add a test for gluttony working under normal circumstances?

Yeah, that'd be good.

Marty-D (approved):
👍

PR #8251, opened 2021-04-29
Chat: Make all plugins share a processmanager

mia-pi-git:


AnnikaCodes:
Is this still relevant? I haven't heard about chat plugins causing lag in a while.

mia-pi-git:
> Is this still relevant? I haven't heard about chat plugins causing lag in a while.

Loading them causes lag, but that's a separate issue. I think it'd be useful to keep this code on hand, but it's not very necessary at present.

PR #8253, opened 2021-04-30
Implement /closehtmlpage

sirDonovan:
This command restores the ability for bots to close their HTML pages. Prior to 0e0389bdc89e3e5b7d5ad9ef237fc9ee30a07477, this was achieved by sending `/sendhtmlpage user, pageid, |deinit|`.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -313,10 +313,13 @@ export const commands: Chat.ChatCommands = {
 		`/pmuhtmlchange [user], [name], [html] - Changes html that was previously PMed to [user] to [html]. Requires * # &`,
 	],
 
+	closehtmlpage: 'sendhtmlpage',
 	changehtmlpageselector: 'sendhtmlpage',
 	sendhtmlpage(target, room, user, connection, cmd) {
 		room = this.requireRoom();
 		this.checkCan('addhtml', null, room);
+
+		const closeHtmlPage = cmd === 'closehtmlpage';
 
 		const {targetUser, rest} = this.requireUser(target);
 		let [pageid, content] = this.splitOne(rest);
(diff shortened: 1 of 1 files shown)


Zarel:
I think closing a page without user interaction is somewhat user-hostile. I think I would prefer to simply update with a message like "Someone challenged you and then cancelled their challenge" or something to that effect.

Or is this for close buttons? In theory you can already script those.

sirDonovan:
This is mainly for closing a room tournament or game page once the activity ends. These already have close buttons for mid-activity use but auto-closing was requested soon after these types of pages were implemented.

I understand your concern, but I feel it's important for bots to have the option for cases where it's the desired behavior.

Zarel:
I can understand someone wanting to auto-close, but I think it'd be better to have a big "close" button in that case. That'd better match how actual battles work in PS too.

Zarel:
Considering that I've never gotten a request to make battles auto-close when you're done playing, I think you may be overestimating how much users actually want that to happen.

sirDonovan:
We ran this poll for 24 hours:

poll

The pages are more comparable to battle controls than battle rooms as a whole. The visuals and logs are sent to the chat rooms.

Zarel:
Fair enough. I'll accept this.

server/chat-commands/admin.ts
```diff
@@ -313,19 +313,24 @@ export const commands: Chat.ChatCommands = {
...
+
 		const {targetUser, rest} = this.requireUser(target);
 		let [pageid, content] = this.splitOne(rest);
 		let selector: string | undefined;
 		if (cmd === 'changehtmlpageselector') {
 			[selector, content] = this.splitOne(content);
 			if (!selector) return this.parse(`/help ${cmd}`);
 		}
-		if (!pageid || !content) return this.parse(`/help ${cmd}`);
+		if (!pageid || (closeHtmlPage ? content : !content)) {
```
Zarel, line 331: This can actually be further simplified to `closeHtmlPage === !!content`, but I think the `?:` method is most readable.

Zarel: This is the one that lets me read "if it's `/closehtmlpage`, having content is an error, otherwise, not having content is the error".

sirDonovan:
Thanks!

PR #8255, opened 2021-04-30
Rename VGC 2021 ladder and upgrade Battle Stadium Singles to Series 9

DaWoblefet:
Approved by chemcoop. This should start May 1 (merging with the other tier changes is fine).

Also removes "VGC 2021" (Series 8), as it is also no longer needed for anything.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -270,7 +270,7 @@ export const Formats: FormatList = [
 			validate: [3, 6],
 			battle: 3,
 		},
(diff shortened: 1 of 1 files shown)


KrisXV:
Can you rename the Series 9 ladder?

KrisXV:
thx

DaWoblefet:
@KrisXV we explicitly wanted the name left as Series 9 so ladder ranks carried over

PR #8258, opened 2021-05-01
Fix Me First's interaction with recharge moves

pacmanboss256:
for some reason a test in room-battle.js times out and I have zero clue why

It fails even with the one line i added commented out


pacmanboss256:
edit: i have zero clue what happened here but it seems to pass everything now

DaWoblefet:
I think this is fine.

Zarel:
Thanks!

Zarel:
@DaWoblefet Perhaps `willMove` should simply not return recharge turns? Would that break any other interaction?

PR #8261, opened 2021-05-02
Implement Zacian/Zamazenta/Xerneas forme leak

Karthik99999:
Corresponding client side changes here: https://github.com/smogon/pokemon-showdown-client/pull/1772

More info on the mechanic is in this twitter thread: https://twitter.com/DaWoblefet/status/1353213137785380864

This does a couple things:

- Adds a new `onStart` event for formats, which is triggered right after `|start` is sent to the client. The descriptions of `onBegin` and `onStart` were changed to clarify their use
- disables `onStart` in Team Preview for gen 7 and below, as this mechanic only exists in gen 8.
- Sends a new protocol `updatepokemon` to change the icons

Let me know if the code can be cleaned up anywhere.


DaWoblefet:
Karthik and I talked a bit about this, and this implementation is really really good imo for standard Gen 8 play. It's somewhat problematic with hackmons, because PS converts Zacian etc. to be its "correct" forme at the Team Preview level. Typically, a Pokemon like Urshifu-Rapid-Strike is Urshifu-Rapid-Strike from the player's perspective at the Team Preview, but from the opponent's perspective, you just see a generic sprite. That's the not the way Zacian works; Zacian is Zacian-H up until the battle starts. But hackmons permits hacking in a Zacian-C at the Team Preview level, and it is _revealed as Zacian-C at Team Preview_. So, properly speaking, non-hacked Zacian will always be Zacian-H at Team Preview, but a hacked Zacian-C will be Zacian-C.

Ultimately though, it's impossible to actually implement the hackmons behavior without taking away teambuilder convenience. Technically in hackmons, you could hack in a Zacian-C with Intrepid Sword and give it Rusted Sword and it will display as Zacian-C in Team Preview. But that means players in regular formats can't create their Zacian in the teambuilder like this:

```
Zacian-Crowned @ Rusted Sword  
Ability: Intrepid Sword  
Level: 50  
EVs: 92 HP / 212 Atk / 4 Def / 4 SpD / 196 Spe  
Jolly Nature  
- Behemoth Blade  
- Sacred Sword  
- Substitute  
- Protect
```

To implement hackmons behavior correctly, the teambuilder would have to reject this as invalid and only allow Zacian-H with Iron Head, which would be contrary to how we handle most other form change conveniences (we'd probably break most AG teams if we did).

Subjectively, I think this minor regression in hackmons behavior is offset by the significant gain in cartridge accuracy for formats like VGC that only allow you to bring part of your team. Karthik's implementation is pretty good!

Marty-D (approved):

sim/battle.ts
```diff
@@ -2306,6 +2306,14 @@ export class Battle {
 			}
 
 			this.add('start');
+			if (format.onStart) format.onStart.call(this);
+			for (const rule of this.ruleTable.keys()) {
+				if (rule.startsWith('+') || rule.startsWith('-') || rule.startsWith('!')) continue;
```
KrisXV, line 2311:
```suggestion
				if (/^[+*-]/.test(rule)) continue;
```

Zarel:
#8267 will fix the whole `onBegin`/`onStart` issue, so let's table this for a day while that gets resolved.

Karthik99999:
sorry about the requested reviews! maybe i shouldn't have just pulled from upstream...

Karthik99999:
first time fixing merge conflicts for a pr, everything should be back to normal now (I hope)

Zarel:
> To implement hackmons behavior correctly, the teambuilder would have to reject this as invalid and only allow Zacian-H with Iron Head in non-hackmons play, which would be contrary to how we handle most other form change conveniences (we'd probably break most AG teams if we changed this).

I don't think this is a problem in practice. We already do this with megas, right? `Blaziken-Mega` means a hacked Mega Blaziken in Hackmons formats, and a regular Blaziken in regular formats.

Zarel:
ty!

PR #8262, opened 2021-05-03
Backport Psychic Terrain hint for priority on non-grounded targets

Karthik99999:
Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8842643


DaWoblefet:
Thanks!

PR #8264, opened 2021-05-03
Improve self-hit confusion damage

DaWoblefet:
See [battle mechanics research post](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8843892) for more details; this behavior happens for sure in Gen 7/8 and almost certainly happens in Gen 5/6.

I'm not sure what I should do for Gens 3 and 4 (Gen 2 overrides the condition already). According to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Confusion_(status_condition)), if I had to guess it looks like confusion damage is using a regular damage calculation, but there's a ton of caveats for each gen. We definitely don't support all those caveats currently and I am not interested in properly researching it for Gens 3 and 4 to sort it out. For now I've just blindly copy-pasted Gen 6's old confusion condition into Gen 4.

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -172,7 +172,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 				return;
 			}
 			this.activeTarget = pokemon;
-			const damage = this.actions.getDamage(pokemon, pokemon, 40);
+			const damage = this.actions.getConfusionDamage(pokemon, 40);
 			if (typeof damage !== 'number') throw new Error("Confusion damage not dealt");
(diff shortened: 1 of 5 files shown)


Zarel (approved):
I like this implementation! I'll leave this open for a day in case Marty has an opinion.

Marty-D:
I've determined that Gen 4 uses the regular damage formula to calculate confusion, so it's correct to ignore it for this implementation. There's one more thing I wanted to check on tomorrow and then I'll merge. 👍

Marty-D (approved):
Looks good, thanks!

PR #7929, opened 2021-01-20
 Server Side for Multi Battles (New)

Spandamn:
Old:   #7607 
Client: smogon/pokemon-showdown-client#1649
Last SS:
image

PR #8266, opened 2021-05-04
Rename and fix maxTeamSize

Zarel:
`maxTeamSize` is a bad variable name (not that `teamLength.battle` is any better, but that'll get fixed in a future refactor).

- Rename `maxTeamSize` to `chosenTeamSize`, to better indicate that this is the size after Team Preview, and that it is also the minimum size after Team Preview.

- Don't limit team sizes to 6 if `teamLength.battle` isn't specified. This removes an unnecessary `teamLength.battle` requirement in all Custom Game formats.

- Stop requiring `maxTeamSize` as a parameter for `battle.getRequests`. It's not even used except as a hint to the Preact client, and was never state in the first place.

- Stop supporting partial `side.chooseTeam`. This is an unused feature and removing it massively simplifies the code and fixes a bug in `cupLevelLimit` which definitely was not written with the understanding that `chooseTeam` could be partial.

- Fix [a bug in #7929](https://github.com/smogon/pokemon-showdown/pull/7929/files#r625493011) which seemed to misunderstand what `teamsize` was for.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -284,7 +284,6 @@ export const Formats: FormatList = [
 		defaultLevel: 100,
 		teamLength: {
 			validate: [1, 24],
-			battle: 24,
 		},
 		// no restrictions, for serious (other than team preview)
 		ruleset: ['Team Preview', 'Cancel Mod'],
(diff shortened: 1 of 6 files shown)


Zarel:
This code is unfortunately so old that it's difficult to tell who originally wrote most of it. I know the unnecessary `getRequests` parameter was probably me, and a few of the others were likely also me.

PR #8268, opened 2021-05-04
EV limit rule, WIP

pacmanboss256:


Zarel:
Might be a bit early for this >_>

Zarel:
Anyway, you don't want to attach these values to the `ruleTable`. The values I'm attaching are used directly by the team validator, but an EV limit validator should be entirely self-contained, like how `Force Monotype` works.

PR #8272, opened 2021-05-05
Fix typos/grammar in CONTRIBUTING.MD

Timbuktu22:


Zarel:
Two of these changes are real mistakes; the others are just a matter of stylistic choice.

CONTRIBUTING.md
```diff
@@ -44,7 +44,7 @@ For simplicity (mostly to make relicensing easier), client code should be also b
 Design standards
 ------------------------------------------------------------------------
 
-We strive to be maximally intuitive and accessible. "That's what they all say", but the currently-popular flat design trend straight-up sacrifices usability for aesthetics, and we try to take the other side of that trade-off.
+We strive to be maximally intuitive and accessible. "That's what they all say," but the currently-popular flat design trend straight-up sacrifices usability for aesthetics, and we try to take the other side of that trade-off.
```
Zarel, line 47: Quote formatting is quite the controversial thing in English orthography.

I'm aware the Chicago Manual of Style wants the comma on the inside here, but it's also common style to put the comma wherever it makes most sense. Linking the phrase to the rest of the sentence isn't _part_ of the phrase, so leaving it outside is valid and I prefer it that way.

Zarel:
```suggestion
We strive to be maximally intuitive and accessible. Sure, "that's what they all say", but the currently-popular flat design trend straight-up sacrifices usability for aesthetics, and we try to take the other side of that trade-off.
```

It might be clearer like this.

CONTRIBUTING.md
```diff
@@ -60,7 +60,7 @@ In addition, blind users in particular navigate by link text, so a blind user wi
 
 ### D3. Remove unnecessary clicks
 
-Whenever you give a user a button to click, always think "in what situations would a user want to click this? in what situations would a user not want to click this?" Dialogs like "Are you sure?" can often be replaced with just doing the thing with an "Undo" button. Buttons to show more details can often be replaced with simply showing more details by default.
+Whenever you give a user a button to click, always think "In what situations would a user want to click this? In what situations would a user not want to click this?" Dialogs like "Are you sure?" can often be replaced with just doing the thing with an "Undo" button. Buttons to show more details can often be replaced with simply more details by default.
```
Zarel, line 63: So there are two ways to handle quotes, depending on whether or not they're complete sentences:

> Nails? Think "hammer".

> Nails? Think, "A hammer would be useful here."

I'm using the first way here, which is complicated because we _are_ using multiple sentences. But if you're going to switch to complete sentence style, you should fully switch to complete sentence style.

As for "showing", it's definitely clearer with it than without it. I don't see any grammatical error with that one.

```suggestion
Whenever you give a user a button to click, always think, "In what situations would a user want to click this? In what situations would a user not want to click this?" Dialogs like "Are you sure?" can often be replaced with just doing the thing with an "Undo" button. Buttons to show more details can often be replaced with simply showing more details by default.
```

CONTRIBUTING.md
```diff
@@ -136,7 +136,7 @@ Sometimes, you have information about a variable/function (such as how and when
 let numConnections: number | null = null;
 ```
 
-Doc comments start with `/**` and end with `*/`. In VS Code, doc comments will show up when you hover your mouse over the variable/function name, anywhere it's used. If your information would be useful there, please put it in a doc comment.
+Doc comments start with `/**` and end with `*/`. In VS Code, doc comments will show up when you hover your mouse over the variable/function name anywhere it's used. If your information would be useful there, please put it in a doc comment.
```
Zarel, line 139: I don't know why you removed the comma here. There are some cases where you need a comma or can't have one, but in many cases, commas are optional. This is one of those cases, and the comma adds clarity which is why I had it in.

CONTRIBUTING.md
```diff
@@ -214,7 +214,7 @@ If you want to have more than one commit in Git master's history after merge (i.
 
 Here is a guide for squashing, if you need help with that: https://redew.github.io/rebaseguide/
 
-If while rebasing, you somehow unintentionally break your pull request, do not close it and make a new one to replace it. Instead, you can ask in the Development chatroom for help on trying to fix it; it can almost always be fixed.
+If while rebasing you somehow unintentionally break your pull request, do not close it and make a new one to replace it. Instead, you can ask in the Development chatroom for help on trying to fix it; it can almost always be fixed.
```
Zarel, line 217:
```suggestion
If, while rebasing, you somehow unintentionally break your pull request, do not close it and make a new one to replace it. Instead, you can ask in the Development chatroom for help on trying to fix it; it can almost always be fixed.
```

The one comma might be weird, but this is another one of those situations where commas are optional and it adds more clarity with them in than out.

CONTRIBUTING.md
Zarel:
```suggestion
Doc comments start with `/**` and end with `*/`. In VS Code, doc comments will show up when you hover your mouse over the variable/function name, anywhere it's used. If your information would be useful there, please put it in a doc comment.
```

CONTRIBUTING.md
Timbuktu22: I interpreted a disagreement between gerunds, but I did not interpret "Buttons to show more details" as a gerund phrase rather than just "Buttons" itself. Thank you for pointing out the clarity issue, though, as that is most important.

Timbuktu22:
Thought I'd make some suggestions on top of the actual errors, but I see your point, so thank you for clarifying. Should be good to go with your changes too.

Zarel:
Fair enough.

PR #8275, opened 2021-05-07
Doubles two turn move now targets a slot turn 1

pacmanboss256:


pacmanboss256:
as referenced here [https://github.com/smogon/pokemon-showdown/projects/3#card-51993101](url)

data/conditions.ts
```diff
@@ -269,15 +269,17 @@ export const Conditions: {[k: string]: ConditionData} = {
...
 		onStart(target, source, effect) {
 			this.effectState.move = effect.id;
+			let slot: PokemonSlot | null = target.getSlot();
 			target.addVolatile(effect.id);
 			let moveTarget: Pokemon | null = source;
 			if (effect.sourceEffect && this.dex.moves.get(effect.id).target === 'normal') {
 				// this move was called by another move such as metronome and needs a random target to be determined now
 				// won't randomly choose an empty slot if there's at least one valid target
 				moveTarget = this.getRandomTarget(target, effect.id);
+				if (moveTarget) slot = moveTarget?.getSlot();
```
AnnikaCodes, line 279:
```suggestion
				if (moveTarget) slot = moveTarget.getSlot();
```
(You don't need to optional chain, since this code is only executed if `moveTarget` is truthy.)

data/conditions.ts
```diff
@@ -269,15 +269,17 @@ export const Conditions: {[k: string]: ConditionData} = {
 		duration: 2,
 		onStart(target, source, effect) {
 			this.effectState.move = effect.id;
+			let slot: PokemonSlot | null = target.getSlot();
```
AnnikaCodes, line 272: Why is this `| null`? `getSlot` always returns a `PokemonSlot`.

pacmanboss256: just in case the move is called by a sub-move

AnnikaCodes: Is the type signature for `getSlot()` not correct?
```ts
getSlot(): PokemonSlot {
	const positionOffset = Math.floor(this.side.n / 2) * this.side.active.length;
	const positionLetter = 'abcdef'.charAt(this.position + positionOffset);
	return (this.side.id + positionLetter) as PokemonSlot;
}
```

pacmanboss256:
I was told to target a slot instead of a pokemon for the first turn

pacmanboss256:
so what is going wrong here

pacmanboss256:
ok i need to mess with rng

PR #8276, opened 2021-05-07
Tests: Use assert.equal and assert.notEqual more

AnnikaCodes:


test/sim/moves/curse.js
```diff
@@ -132,7 +132,10 @@ describe('XY/ORAS Curse targetting when becoming Ghost the same turn', function
 		} else {
 			// Cartridge glitch
 			assert.notEqual(cursePartner.hp, cursePartner.maxhp); // Partner afflicted by Curse
-			assert(foeHP[0] === p2active[0].maxhp && foeHP[1] === p2active[1].maxhp); // Foes unaffected by Curse
+
+			// Foes unaffected by Curse
+			assert.equal(foeHP[0], p2active[0].maxhp);
```
Slayer95, line 137:
```js
assert.fullHP
```

test/sim/abilities/flashfire.js
```diff
@@ -75,6 +75,6 @@ describe('Flash Fire [Gen 3-4]', function () {
 		const flashFireMon = battle.p1.active[0];
 		flashFireMon.setStatus('frz');
 		battle.makeChoices();
-		assert(flashFireMon.hp !== flashFireMon.maxhp);
+		assert.notEqual(flashFireMon.hp, flashFireMon.maxhp);
```
DaWoblefet, line 78:
```suggestion
		assert.false.fullHP(flashFireMon.hp, flashFireMon.maxhp);
```

DaWoblefet (approved):

test/sim/moves/curse.js
```diff
@@ -132,7 +132,10 @@ describe('XY/ORAS Curse targetting when becoming Ghost the same turn', function
 		} else {
 			// Cartridge glitch
 			assert.notEqual(cursePartner.hp, cursePartner.maxhp); // Partner afflicted by Curse
```
Zarel, line 134:
```suggestion
			assert.false.fullHP(cursePartner); // Partner afflicted by Curse
```

Zarel:
Personally I find the `===` syntax more readable (it's also terser) but I don't particularly care enough to argue about it.

One thing we could do is add a compile step to our tests and then compile the `===` syntax to the `.equals` syntax... but I don't want to slow down the test process. Honestly, if we ever did start building unit tests, though, I would want to do that.

AnnikaCodes:
I like `assert.equals` because it automatically shows the expected and received values when a test fails, not for readability reasons.

Zarel:
I know, that's why compiling `===` to the `.equals` syntax is the best of both worlds.

PR #8278, opened 2021-05-07
Fix "All Pokemon" ban not banning all pokemon in natdex.

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8847996

Moves checking for `pokemontag:allpokemon` right before the nonexistent checks are done. Worked well in testing, though the testing wasn't that rigorous. Just experimented a little with natdex tours.

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1285,7 +1285,7 @@ export class TeamValidator {
 		if (!tierSpecies.canGigantamax && set.gigantamax) {
 			return `${tierSpecies.name} cannot Gigantamax but is flagged as being able to.`;
 		}
-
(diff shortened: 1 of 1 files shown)


Karthik99999:
now that I think about it, doing this probably goes against the whole "more specific beats less specific". correct way to fix this is probably implementing natdex in a different way

Zarel:
This would definitely break e.g. `- all pokemon, + CAP` to have a CAP-only metagame, for instance.

Zarel:
Correct fix would be to stuff all bans/unbans in the `tagRules` array.

PR #8279, opened 2021-05-07
Random Battles: Use more Dex functions

KrisXV:


data/mods/gen6/random-teams.ts
```diff
@@ -785,7 +785,7 @@ export class RandomGen6Teams extends RandomGen7Teams {
 
 			// Choose next 4 moves from learnset/viable moves and add them to moves list:
 			while (moves.length < 4 && movePool.length) {
-				const moveid = this.sampleNoReplace(movePool);
+				const moveid = this.sampleNoReplace(movePool) as ID;
```
AnnikaCodes, line 788: Why all the casts?

AnnikaCodes: `movePool` should be an array of `ID`s already?

AnnikaCodes: @KrisXV?

AnnikaCodes: Oh, I see. `movePool` can't be an array of IDs because it would require casting on every `.includes()` check. `rejectedPool` should still be declared as an `ID[]`, though; that shouldn't cause any typing problems.

AnnikaCodes:
@KrisXV?

Zarel:
:( I really would have liked this to get merged...

KrisXV:
If you check the dev discord you’d see annika is doing it….

PR #8282, opened 2021-05-08
Life orb recoil now happens when future moves hit rather than on usage

pacmanboss256:
keeping this as a draft for now so i can fix it because its probably wrong


DaWoblefet:
I'm not a fan of this solution. If Life Orb recoil is running _at all_ on first turn of its usage, it means that future attacks are running the rest of checks done before and after move success when they should not. See [this thread](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8548957) to get a rough idea. So I would wager PS is also doing things like allowing Future Sight to change type with Protean / Libero. It likely wouldn't hurt to do extra research though, since we've got all that new research from USUM to explore.

PR #8285, opened 2021-05-08
Implement /sendprivatehtmlbox

sirDonovan:
Similar use as `/sendhtmlpage` for bots but for cases where the HTML is small or should be easier to reach from the chat room.

send-private-html

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -420,9 +420,9 @@ export const commands: Chat.ChatCommands = {
 		`If a [highlight] is specified, only highlights them if they have that term on their highlight list.`,
 	],
 
-	changeprivateuhtml: 'sendprivatehtml',
-	sendprivateuhtml: 'sendprivatehtml',
-	sendprivatehtml(target, room, user, connection, cmd) {
+	changeprivateuhtml: 'sendprivatehtmlbox',
+	sendprivateuhtml: 'sendprivatehtmlbox',
+	sendprivatehtmlbox(target, room, user, connection, cmd) {
 		room = this.requireRoom();
 		this.checkCan('addhtml', null, room);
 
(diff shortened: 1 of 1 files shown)


server/chat-commands/admin.ts
```diff
@@ -420,6 +420,57 @@ export const commands: Chat.ChatCommands = {
...
+	},
+	sendprivatehtmlhelp: [
+		`/sendprivatehtml [userid], [html] - Sends [userid] the private [html]. Requires: * # &`,
+	],
+	sendprivateuhtmlhelp: [
+		`/sendprivateuhtml [userid], [name], [html] - Sends [userid] the private [html] that can change. Requires: * # &`,
+	],
+	changeprivateuhtmlhelp: [
+		`/changeprivateuhtml [userid], [name], [html] - Changes the message previously sent with /sendprivateuhtml [userid], [name], [html]. Requires: * # &`,
+	],
```
mia-pi-git, line 472:
```suggestion
	sendprivatehtmlhelp: [
		`/sendprivatehtml [userid], [html] - Sends [userid] the private [html]. Requires: * # &`,
	],
	sendprivateuhtmlhelp: [
		`/sendprivateuhtml [userid], [name], [html] - Sends [userid] the private [html] that can change. Requires: * # &`,
	],
	changeprivateuhtmlhelp: [
		`/changeprivateuhtml [userid], [name], [html] - Changes the message previously sent with /sendprivateuhtml [userid], [name], [html]. Requires: * # &`,
	],
```

It might be cleaner to compress these into one help under `sendprivatehtml`? I think that's generally what we do elsewhere.

server/chat-commands/admin.ts
```diff
@@ -420,6 +420,57 @@ export const commands: Chat.ChatCommands = {
...
+	sendprivateuhtml: 'sendprivatehtml',
+	sendprivatehtml(target, room, user, connection, cmd) {
+		room = this.requireRoom();
+		this.checkCan('addhtml', null, room);
+
+		const {targetUser, rest} = this.requireUser(target);
+
+		if (targetUser.locked && !this.user.can('lock')) {
+			this.errorReply("This user is currently locked, so you cannot send them private HTML.");
+			return false;
```
AnnikaCodes, line 433:
```suggestion
			throw new Chat.ErrorMessage("This user is currently locked, so you cannot send them private HTML.");
```
This seems more consistent

Zarel:
Why even have the box around regular HTML if you can just bypass it with UHTML? -_-

sirDonovan:
That seems to be the standard looking at `/addhtmlbox` & `/adduhtml`,  `/addrankhtmlbox` & `/addrankuhtml`, or `/pminfobox` & `/pmuhtml`.

Zarel:
I'm sure they have lots of historical basis but at some point we probably want to actually fix this.

server/chat-commands/admin.ts
```diff
@@ -420,6 +420,57 @@ export const commands: Chat.ChatCommands = {
 		`If a [highlight] is specified, only highlights them if they have that term on their highlight list.`,
 	],
 
+	changeprivateuhtml: 'sendprivatehtml',
+	sendprivateuhtml: 'sendprivatehtml',
+	sendprivatehtml(target, room, user, connection, cmd) {
```
Zarel, line 425: Anyway this should be `sendprivatehtmlbox` if it's supposed to have the box.

sirDonovan: Good point

Zarel:
Thanks! We'll have to deal with the box issue later. :s

PR #8286, opened 2021-05-09
Prevent bringing less Pokemon in Doubles and Triples battles

xInstruct:
Caused a crash earlier, but this should prevent it

### DIFF
--- a/sim/dex-formats.ts
+++ b/sim/dex-formats.ts
@@ -245,6 +245,9 @@ export class RuleTable extends Map<string, string> {
 		if (this.pickedTeamSize && this.pickedTeamSize < gameTypeMinTeamSize) {
 			throw new Error(`Chosen team size ${this.pickedTeamSize}${this.blame('pickedteamsize')} must be at least ${gameTypeMinTeamSize} for a ${format.gameType} game.`);
 		}
+		if (this.maxTeamSize && this.maxTeamSize < gameTypeMinTeamSize) {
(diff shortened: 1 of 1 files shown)


xInstruct:
actually, theres a better way to do this

sim/dex-formats.ts
```diff
@@ -686,6 +686,14 @@ export class DexFormats {
 				if (subformat.hasValue === 'positive-integer') {
 					if (parseInt(value) <= 0) throw new Error(`In rule "${ruleSpec}", "${value}" must be positive.`);
 				}
+				if (subformat.id === 'pickedteamsize' || subformat.id === 'maxteamsize') {
+					if (format.gameType === 'doubles' && parseInt(value) < 2) {
+						throw new Error(`In a doubles battle, at least 2 Pokemon is needed.`);
+					}
+					if (format.gameType === 'triples' && parseInt(value) < 3) {
+						throw new Error(`In a triples battle, at least 3 Pokemon is needed.`);
```
AnnikaCodes, line 694:
```suggestion
						throw new Error(`In a doubles battle, at least 2 Pokemon are needed.`);
					}
					if (format.gameType === 'triples' && parseInt(value) < 3) {
						throw new Error(`In a Triples battle, at least 3 Pokemon are needed.`);
```

AnnikaCodes: This should probably be in the `else if (subformat.hasValue === 'integer' || subformat.hasValue === 'positive-integer')` block? That's where the rest of the number validation is

xInstruct: Did it a much better way, this is now irrelevant

Zarel:
You're missing a check for `maxTeamSize >= minTeamSize` and `maxTeamSize >= pickedTeamSize`.

PR #8287, opened 2021-05-09
Use |-swapsideconditions| for Court Change

Karthik99999:
Corresponding client change: smogon/pokemon-showdown-client#1783

Uses a new pline `-sideswitch`, which swaps side conditions between sides. Right now it only uses it for non ffa battles, mostly cause I don't really understand the ffa code rn, and was hoping I could get some help on that one.

https://user-images.githubusercontent.com/32044378/117557824-1cf54480-b045-11eb-9be0-e11c2e20ea69.mp4

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -2727,7 +2727,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 					}
 					success = true;
 				}
-				this.add('-sideswitch', '[silent]');
+				this.add('-sideswitch');
 			}
 			if (!success) return false;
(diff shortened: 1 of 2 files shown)


Karthik99999:
maybe this can be a good time to remove the randomness of court change in FFAs like @Zarel suggested here https://github.com/smogon/pokemon-showdown/pull/8259#issuecomment-831868687?

Zarel (approved):
Looks good, but we might want to unify FFA?

Karthik99999:
bumping. I think it would be better to just merge this now and deal with ffa later

PR #8288, opened 2021-05-09
Properly implement Gen 1 Wrap

InAShellnut:
So as covered here, https://www.smogon.com/forums/threads/rby-fight-button-simulation-improvements-fix-partial-trapping-improve-counter.3673280/, Wrap is currently not properly implemented since there is not a "Fight" button, now while yes Pokemon Showdown does implement it as if the Pokemon being wrapped opens up the FIGHT menu first, then attacks. The issue is that when a wrapped player tries to open the "Fight" Menu, the menu doesn't even come up in the first place, which tells the player being wrapped if they can attack or not, which is very important information when it comes to mind games and predicting switches. As an example, if you know the opponent's wrap has ended, you can predict they will switch to something defensive and can capitalize on that with your own switch. On cartridge, the player being wrapped would have this information because they could check if they can move or not by attempting to open the fight menu and seeing if it opens or not, on Pokemon Showdown they cannot. This is the issue with Showdown's current implementation that Plague Von Karma tried to explain, it is hiding information that should be available to the player being wrapped.

How this code fixes Gen 1 Wrap is by locking the user's moves to just a "Fight" option when they are wrapped while also giving them the freedom to switch. Since a player would always cancel the "Fight" option when being wrapped since it gives them information of if they can attack or not, this implementation should be correct from an information perspective. If I need to clarify anything, please ask, thanks.

### DIFF
--- a/data/mods/gen1/conditions.ts
+++ b/data/mods/gen1/conditions.ts
@@ -192,6 +192,7 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 			this.add('cant', pokemon, 'partiallytrapped');
 			return false;
 		},
+		onLockMove: 'fight',
 	},
 	partialtrappinglock: {
 		name: 'partialtrappinglock',
--- a/data/mods/gen1/conditions.ts
+++ b/data/mods/gen1/conditions.ts
@@ -190,9 +190,9 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		onBeforeMovePriority: 4,
 		onBeforeMove(pokemon) {
 			this.add('cant', pokemon, 'partiallytrapped');
+			pokemon.maybePartialTrappingLock = true;
 			return false;
 		},
-		onLockMove: 'fight',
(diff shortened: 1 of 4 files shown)


PlagueVonKarma:
Just going to say I went over this with Shell prior to him sparking this project off. He's absolutely right; I also don't like the info leak that's currently on PS and this would fix the problem. I believe this fix would also partially fix the problems with Stadium's Wrap? 

The trapper's Wrap works in the same way on the cartridge, but the "Wrap" button already serves this purpose in my opinion. You could technically extend this to the player but it may get a bit messy.

InAShellnut:
Edit: nvm, i need to fix its implementation with cancel mod.

InAShellnut:
I forgot that cancel mod wouldn't let you cancel a switch if your not arena trapped as otherwise info would leak, thus locking fight would by these rules prevent you from cancelling and switching, I'm stupid....

mia-pi-git:

PR #8291, opened 2021-05-09
mold breaker refactor

pacmanboss256:
Mold breaker now ignores specific abilities rather than by using event checks
No longer specifies abilities that were not supposed to be ignored (isUnbreakable)


pacmanboss256:
rough idea of making mold breaker only ignore certain abilities rather than not ignore certain abilities and ignore everything else

pacmanboss256:
will fix sunsteel strike once this is merged

sim/battle.ts
```diff
@@ -710,37 +710,10 @@ export class Battle {
 				// it's changed; call it off
 				continue;
 			}
-			if (effect.effectType === 'Ability' && !effect.isUnbreakable &&
+			if (effect.effectType === 'Ability' && effect.ignorable &&
```
urkerab, line 713: Rather than ignoring `isUnbreakable` you should just remove it completely.

server/chat-commands/info.ts
```diff
@@ -803,7 +803,7 @@ export const commands: Chat.ChatCommands = {
 						Gen: String(ability.gen) || 'CAP',
 					};
 					if (ability.isPermanent) details["&#10003; Not affected by Gastro Acid"] = "";
-					if (ability.isUnbreakable) details["&#10003; Not affected by Mold Breaker"] = "";
+					if (ability.ignorable) details["&#10003; Ignored by Mold Breaker"] = "";
```
Zarel, line 806: What's with the flip? Are there more abilities affected by Mold Breaker or not affected?

Zarel: (Out of the abilities that could potentially be affected by Mold Breaker.)

pacmanboss256: oh its flipped because isUnbreakable will be deleted

data/mods/gen4/abilities.ts
```diff
@@ -376,7 +377,7 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 	stickyhold: {
 		inherit: true,
 		onTakeItem(item, pokemon, source) {
-			if (this.suppressingAttackEvents(pokemon)) return;
+			if (this.suppressingAbility(pokemon)) return;
```
Marty-D, line 380: Here, the whole line is actually useless.

Marty-D (changes requested):
- Rebound needs to be considered
- `isBreakable: false` is unnecessary on real Abilities
- Please check past gens for cases like Gen 4 Simple which override the whole effect

pacmanboss256 (commented):
all set

Marty-D:
> Please check past gens for cases like Gen 4 Simple which override the whole effect

Gen 4 Thick Fat
Gen 3 Lightning Rod (realistically this one doesn't matter, but no reason to take functionality away from Custom Games)

Marty-D:
While reviewing I wondered if Aura Break, Dark Aura, and Fairy Aura would actually have their `Start` events suppressed in-game, so I checked and they do! But PS runs `Start` separately in `runSwitch`, so you'll need to add `suppressingAbility` to those three Abilities (kind of like how Sticky Hold did it) for completeness.

Marty-D (approved):
Looks good, thanks! 👍

server/chat-commands/info.ts
Zarel: I'm saying why should it be deleted?

Zarel: I think it'd be better if it was flipped back.

PR #8293, opened 2021-05-10
Sockets: Monitor and reconnect slow sockets

mia-pi-git:


mia-pi-git:
This got fixed in a different way? I think? Either way, probably unnecessary.

PR #8294, opened 2021-05-11
Validator: Improve Gen 1 level validation

AnnikaCodes:
cc @PlagueVonKarma

### DIFF
--- a/data/mods/gen2/learnsets.ts
+++ b/data/mods/gen2/learnsets.ts
@@ -83,6 +83,9 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 		eventData: [
 			{generation: 2, level: 5, shiny: 1, moves: ["tackle", "growl", "ancientpower"]},
 		],
+		encounters: [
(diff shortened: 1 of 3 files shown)


PlagueVonKarma:
Few notes for why this is a thing, in case people aren't sure;
- Currently, you can use L1 Pokemon in RBY, an impossible level that causes crashes on the cart.
- You cannot breed in RBY w/out Tradebacks, so every Pokemon has a minimum level to adhere to. This is important for a couple of side formats like Pika Cup, which is currently borderline unplayable due to the restrictions.
- These legality issues affect newgens to some degree; for example, Aerodactyl can only have Rage at L5+ even factoring in Tradeback learnsets.
- The validator currently unfairly bans Pokemon like Dragonair from level-limited formats because it thinks the Pokemon are evolve-only. Ergo, an L15 Dragonair, which you can fish up in Fuschia City Safari Zone, can't be used without this fix.

Zarel:
Thanks!

PR #8295, opened 2021-05-11
Scavengers: Ladder isn't monthly

PartMan7:
Changed all occurrences of monthly to a standardized `current`


AnnikaCodes:
Thanks!

PR #8296, opened 2021-05-12
Update to Sw/Sh 1.3.2 mechanics

KrisXV:
I'm waiting on a test from Anubis to see if the pre-1.3.2 functionality for Trick/Switcheroo on Kyogre/Groudon was in DLC1.

### DIFF
--- a/data/mods/gen8dlc1/rulesets.ts
+++ b/data/mods/gen8dlc1/rulesets.ts
@@ -1,18 +1,7 @@
 export const Rulesets: {[k: string]: ModdedFormatData} = {
 	teampreview: {
-		effectType: 'Rule',
-		name: 'Team Preview',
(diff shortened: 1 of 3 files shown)


DaWoblefet:
> I'm waiting on a test from Anubis to see if the pre-1.3.2 functionality for Trick/Switcheroo on Kyogre/Groudon was in DLC1.

It does not matter because Kyogre / Groudon were not in DLC1. Even if it was the behavior, it does not matter.

data/mods/gen8dlc1/rulesets.ts
```diff
@@ -0,0 +1,25 @@
+export const Rulesets: {[k: string]: ModdedFormatData} = {
+	teampreview: {
```
Karthik99999, line 2: Can't you just inherit everything and only add the `onFieldStart` function?

Marty-D (approved):
Thanks!

PR #8297, opened 2021-05-12
Fix faint check mid-attack

Marty-D:
Only bother checking for a winner if the move's user is also fainted at the time, and check for a winner after the move completes.

### DIFF
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -313,6 +313,8 @@ export class BattleActions {
 			}
 		}
 		if (noLock && pokemon.volatiles['lockedmove']) delete pokemon.volatiles['lockedmove'];
+		this.battle.faintMessages();
+		this.battle.checkWin();
 	}
 	/**
 	 * useMove is the "inside" move caller. It handles effects of the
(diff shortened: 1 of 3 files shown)


DaWoblefet:
I'd add a test for Life Orb recoil playing before faint messages are processed.

Marty-D:
There's no functional difference between recoil moves and Life Orb here. Life Orb just happens later, but still between the target's faint message and the end of runMove.

sim/battle-actions.ts
```diff
@@ -313,6 +313,8 @@ export class BattleActions {
 			}
 		}
 		if (noLock && pokemon.volatiles['lockedmove']) delete pokemon.volatiles['lockedmove'];
+		this.battle.faintMessages();
+		this.battle.checkWin();
```
Zarel, line 317: Wait, I'm confused. `faintMessages` still runs `checkWin` when invoked like this, so the second `checkWin` is useless, right?

sim/battle.ts
```diff
@@ -2219,7 +2219,7 @@ export class Battle {
 			}
 		}
 
-		if (this.checkWin(faintData)) return true;
+		if (checkWin && this.checkWin(faintData)) return true;
 
 		if (faintData && length) {
 			this.runEvent('AfterFaint', faintData.target, faintData.source, faintData.effect, length);
```
Zarel, line 2225: How _should_ `checkWin` interact with `AfterFaint`?

sim/battle-actions.ts
Marty-D: If the `faintMessages` mid-attack fainted the target, there's nothing in the faint queue when it gets here. Without running `checkWin` separately, the first test I added fails because the battle isn't actually over.

sim/battle.ts
Marty-D: If the battle's over these effects don't happen.

PR #8300, opened 2021-05-13
Fix Counter redirecting Encored self-target moves

pacmanboss256:
Counter/mirror coat target redirection carried over because of the decreased priority, this fix makes it so the target is not redirected if the move being used is not counter or mirror coat.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -4275,6 +4275,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 					// it failed
 					return false;
 				}
+				this.effectState.slot = target.getSlot();
+				this.effectState.user = target;
 				this.effectState.move = move.id;
 				this.add('-start', target, 'Encore');
 				if (!this.queue.willMove(target)) {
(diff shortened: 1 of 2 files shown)


pacmanboss256:
referenced here https://github.com/smogon/pokemon-showdown/projects/3#card-52058554

Zarel:
This seems like the wrong place to fix it? `OverrideAction` should probably be the place that re-rolls the target if incompatible?

data/moves.ts
```diff
@@ -11267,6 +11270,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 			},
 			onRedirectTargetPriority: -1,
 			onRedirectTarget(target, source, source2) {
+				if (source.moveThisTurn !== 'mirrorcoat') return;
```
Zarel, line 11273:
```suggestion
			onRedirectTarget(target, source, source2, move) {
				if (move.id !== 'mirrorcoat') return;
```

data/moves.ts
```diff
@@ -4275,6 +4276,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 					// it failed
 					return false;
 				}
+				this.effectState.slot = target.getSlot();
+				this.effectState.user = target;
```
Zarel, line 4280:
```suggestion
```

data/moves.ts
```diff
@@ -2645,6 +2645,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 			},
 			onRedirectTargetPriority: -1,
 			onRedirectTarget(target, source, source2) {
+				if (source.moveThisTurn !== 'counter') return;
```
Zarel, line 2648:
```suggestion
			onRedirectTarget(target, source, source2, move) {
				if (move.id !== 'counter') return;
```

pacmanboss256:
bump

Zarel:
Thanks!

PR #8301, opened 2021-05-14
Use different animation depending on Shell Side Arm category.

Karthik99999:
Uses Sludge Bomb for special, Poison Jab for physical. I left the hints in for accessibility reasons, but I can remove it if you guys want me to.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -15121,15 +15121,12 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
-		onTryMove() {
-			this.attrLastMove('[still]');
-		},
 		onPrepareHit(target, source, move) {
 			if (!source.isAlly(target)) {
 				if (move.category === 'Physical') {
-					this.add('-anim', source, 'Poison Jab', target);
+					this.attrLastMove('[anim] Poison Jab');
 				} else {
-					this.add('-anim', source, 'Sludge Bomb', target);
+					this.attrLastMove('[anim] Sludge Bomb');
 				}
 			}
 		},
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I like this!

xInstruct:
This is an issue in Staff Bros, so this is 100% gonna come up. There's gonna be false positives where the animation would hit but would actually miss.

Karthik99999:
> This is an issue in Staff Bros, so this is 100% gonna come up. There's gonna be false positives where the animation would hit but the move itself would actually miss.

Well this is a problem with showdown in general, cause move animations just move a little bit to the side if it misses instead of just not playing. Though I guess this is the one circumstance where this showdown behavior matters.

data/moves.ts
```diff
@@ -15121,6 +15121,18 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
+		onTryMove() {
+			this.attrLastMove('[still]');
```
Zarel, line 15125: I'm a bit uncomfortable with the animation removal and the animation adding to be on two separate events. Why aren't these both in `onPrepareHit`?

Also `attrLastMove('[anim] Poison Jab')` works, right?

Zarel: Also, may want to `attrLastMove('[anim] Shell Side Arm Physical')` and then alias it to Poison Jab in `move-animations`.

xInstruct: > attrLastMove('[anim] Shell Side Arm Physical')

wait thats a thing????

Karthik99999: I just copied the staff bros behavior tbh. And yes `attrLastMove('[anim] Poison Jab')` does work.

Zarel: @xInstruct lol yes Staff Bros should probably be using it, too.

Zarel:
ty!

PR #8305, opened 2021-05-15
Fix Neutralizing Gas ending if another is active

DaWoblefet:
Previously, if multiple Neutralizing Gas were active, switching one out would cause the Neutralizing Gas "The effects of the neutralizing gas wore off!" to play when it shouldn't. Ability events were being re-run too unnecessarily (they wouldn't do anything since it was still being negated). Note that it is unlike Primal weather, where Neutralizing Gas _does_ announce itself on switchin, even if the effect is already active.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2360,6 +2360,14 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			}
 		},
 		onEnd(source) {
+			let nGasActiveElsewhere = false;
+			for (const pokemon of this.getAllActive()) {
+				if (pokemon !== source && pokemon.hasAbility('Neutralizing Gas')) {
+					nGasActiveElsewhere = true;
+					break;
+				}
+			}
+			if (nGasActiveElsewhere) return;
 			this.add('-end', source, 'ability: Neutralizing Gas');
 
 			// FIXME this happens before the pokemon switches out, should be the opposite order.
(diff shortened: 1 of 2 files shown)


data/abilities.ts
```diff
@@ -2360,6 +2360,14 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			}
 		},
 		onEnd(source) {
+			let nGasActiveElsewhere = false;
+			for (const pokemon of this.getAllActive()) {
+				if (pokemon !== source && pokemon.hasAbility('Neutralizing Gas')) {
```
urkerab, line 2365: Not sure that `hasAbility` is a good idea here; compare `ignoringAbility`.

It might be worth sharing the code for neutralising gas being active between here and there, as long as you set this ability state to be ending first.

In fact, I think there might be a spurious `-end` bug with an NGas user switching out while under the effect of Gastro Acid, which setting the ability state to be ending first would fix.

Marty-D: hasAbility() seems okay to use in this case since it doesn't ignore itself and covers Gastro Acid. The only ways for `End` to run are losing the Ability somehow, switching out, or becoming affected by Gastro Acid. If you check that Pokemon other than this one actually have a functional Neutralizing Gas then it should be fine?

What seems unnecessary here is using `nGasActiveElsewhere`; you can just `return` right away if this line is true.

Zarel:
Thanks!

Zarel:
Marty is right, though; I probably merged too fast, but that can be a follow-up commit.

DaWoblefet:
Improved the logic in https://github.com/smogon/pokemon-showdown/commit/ddd5d6ac15490804f241f0f86a49ae11325224aa

PR #8306, opened 2021-05-15
Gen VII: Ignore user's Ability with Mold Breaker moves

DaWoblefet:
Not sure if this was the correct place or not, but it does seem to work!

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -308,7 +308,7 @@ export class Battle {
 	}
 
 	suppressingAbility(target?: Pokemon) {
-		return this.activePokemon && this.activePokemon.isActive && (this.activePokemon !== target || this.gen === 7) &&
+		return this.activePokemon && this.activePokemon.isActive && (this.activePokemon !== target || this.gen < 8) &&
 			this.activeMove && this.activeMove.ignoreAbility;
 	}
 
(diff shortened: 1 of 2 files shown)


sim/battle.ts
```diff
@@ -308,7 +308,7 @@ export class Battle {
 	}
 
 	suppressingAbility(target?: Pokemon) {
-		return this.activePokemon && this.activePokemon.isActive && this.activePokemon !== target &&
+		return this.activePokemon && this.activePokemon.isActive && (this.activePokemon !== target || this.gen === 7) &&
```
Marty-D, line 311:
```suggestion
		return this.activePokemon && this.activePokemon.isActive && (this.activePokemon !== target || this.gen < 8) &&
```

sim/battle.ts
```diff
@@ -726,7 +726,7 @@ export class Battle {
 				// it's changed; call it off
 				continue;
 			}
-			if (effect.effectType === 'Ability' && effect.isBreakable !== false &&
+			if (effect.effectType === 'Ability' && effect.isBreakable &&
```
Marty-D, line 729:
```suggestion
			if (effect.effectType === 'Ability' && effect.isBreakable !== false &&
```
This can't be removed until mods have their Abilities tagged properly.

Marty-D (approved):
Nice, thanks! 👍

PR #8307, opened 2021-05-15
Fix the order of many end-turn resolution effects

DaWoblefet:
I noticed Grassy Terrain was happening before Volcalith because of it missing an ``onResidualSubOrder``, decided to audit the rest of the end-turn resolution order numbers and got to fix quite a few. List is partially based on my [post from October 2019](https://www.smogon.com/forums/threads/ultra-sun-ultra-moon-battle-mechanics-research-read-post-2.3620030/post-8264465) as well as notes from [new USUM research](https://github.com/smogon/pokemon-showdown/pull/8222#issuecomment-825941312). As far as new effects go:
- Octolock is sandwiched between binding moves and Taunt according to Zaggyo
- G-Max chip _technically_ doesn't have an ending message on cart, but there's no way it's anything but the same block as Pledge moves / screens etc. I just gave it values higher than Aurora Veil.

I did not touch mods (like SSB) or anything before Gen 5, but these orders should be backported into Gens 5-7.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -237,7 +237,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 188,
 	},
 	baddreams: {
-		onResidualOrder: 27,
+		onResidualOrder: 28,
 		onResidualSubOrder: 2,
 		onResidual(pokemon) {
 			if (!pokemon.hp) return;
@@ -1387,7 +1387,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	},
 	harvest: {
 		name: "Harvest",
-		onResidualOrder: 27,
+		onResidualOrder: 28,
 		onResidualSubOrder: 2,
 		onResidual(pokemon) {
 			if (this.field.isWeather(['sunnyday', 'desolateland']) || this.randomChance(1, 2)) {
@@ -1458,7 +1458,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 37,
 	},
 	hungerswitch: {
-		onResidualOrder: 28,
+		onResidualOrder: 29,
 		onResidual(pokemon) {
 			if (pokemon.species.baseSpecies !== 'Morpeko' || pokemon.transformed) return;
 			const targetForme = pokemon.species.name === 'Morpeko' ? 'Morpeko-Hangry' : 'Morpeko';
@@ -2161,7 +2161,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 104,
 	},
 	moody: {
-		onResidualOrder: 27,
+		onResidualOrder: 28,
 		onResidualSubOrder: 2,
 		onResidual(pokemon) {
 			let stats: BoostID[] = [];
@@ -2619,7 +2619,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		num: 124,
 	},
 	pickup: {
-		onResidualOrder: 27,
+		onResidualOrder: 28,
 		onResidualSubOrder: 2,
 		onResidual(pokemon) {
 			if (pokemon.item) return;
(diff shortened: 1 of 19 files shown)


DaWoblefet:
This is the PS order now (no parentheses = onResidualOrder):
1. (onFieldResidualOrder) Weather effects 
2. Nothing (most lists separated weather ending and things like Sand damage / Solar Power as step 2)
3. Future moves / Wish
4. Nothing (most lists separated Wish and future attacks, but they should be the same step)
5. Some more stuff

(onResidualSubOrder)
1. Side conditions - G-Max Chip / Sea of Fire
2. Field conditions - Grassy Terrain
3. Abilities - Shed Skin / Healer / Hydration
4. Items - Leftovers / Black Sludge

6. Aqua Ring
7. Ingrain
8. Leech Seed
9. Poison / Toxic poison
10. Burn
11. Nightmare
12. Curse
13. Binding moves
14. Octolock
15. Taunt
16. Encore
17. Disable
18. Magnet Rise
19. Telekinesis
20. Heal Block
21. Embargo
22. Throat Chop
23. Yawn
24. Perish Song
25. Roost
26. (onSideResidualOrder) A bunch of side field conditions ending

(onSideResidualSubOrder)

1. Reflect
2. Light Screen
3. Safeguard
4. Mist
5. Tailwind
6. Lucky Chant
7. Pledge rainbow
8. Pledge Sea of Fire
9. Pledge swamp
10. Aurora Veil
11. G-Max Chip

27. (onFieldResidualOrder) A bunch of other field effects ending

(onFieldResidualSubOrder)
1. Trick Room
2. Gravity
3. Water Sport
4. Mud Sport
5. Wonder Room
6. Magic Room
7. All of the Terrains

28. Some more stuff

(onResidualSubOrder)
1. Uproar
2. Abilities - Bad Dreams, Harvest, Moody, Pickup, Slow Start, Speed Boost
3. Items - Flame Orb / Toxic Orb / Sticky Barb

29. Hunger Switch / Power Construct / Shields Down / Schooling / Zen Mode

Zarel:
pyuk changed terrains to work on residual but left weathers alone. Should weathers also be moved to the new system?

DaWoblefet:
> pyuk changed terrains to work on residual but left weathers alone. Should weathers also be moved to the new system?

What's the difference between weather and Terrains in this context? It's both onFieldResidualOrder.

Zarel:
Weather uses `onWeather`, while terrain uses `onResidual`. Weather should probably be changed to also use `onResidual`.

pacmanboss256:
solar power uses onWeather as well, not sure when the HP loss takes place but that might be important

DaWoblefet:
Whatever is going on with onWeather, all the behavior is correct:

- Weather wears off before the final turn of the effect
- Weather occurs before Future attacks
- Weather residual effects are run in order of Speed

So I don't think there's a problem here unless I'm missing something else.

Marty-D:
This is as good a time as any to actually do Gen 3 and 4 properly, which are very similar to each other but very different from later gens. There are a few things I want to check and then I'll sort them out in this PR. Looks good so far!

Marty-D (approved):
Well that was a marathon. I'll look at Gen 2 tomorrow but it should only be a small handful of things.

Zarel:
@DaWoblefet weather might not be broken, but moving it from `onWeather` to `onResidual` should make it easier to maintain. You'd have it all in one Order/SubOrder list, instead of having a separate list for only weather residuals.

DaWoblefet:
> @DaWoblefet weather might not be broken, but moving it from `onWeather` to `onResidual` should make it easier to maintain. You'd have it all in one Order/SubOrder list, instead of having a separate list for only weather residuals.

I'll leave that task for a separate PR.

Marty-D (approved):
Woo! Good to go, thanks for trudging through 5-8! 👍

PR #8308, opened 2021-05-15
Implement Dynamax levels

Karthik99999:
Client Side: smogon/pokemon-showdown-client#1805

It works pretty much as expected. I don't think I missed anything but if I did let me know. Also unsure if I put the dynamax level in the "correct" spot in the packed team format (kinda just shoved it in at the end with all the other misc data).


sim/pokemon.ts
```diff
@@ -956,6 +958,7 @@ export class Pokemon {
 			}
 		}
 		if (!atLeastOne) return;
+		if (this.dynamaxLevel) result.dynamaxLevel = this.dynamaxLevel;
```
KrisXV, line 961: What does this do here?

sim/teams.ts
```diff
@@ -98,6 +98,12 @@ export interface PokemonSet {
 	 * because `ivs` contain post-Battle-Cap values.
 	 */
 	hpType?: string;
+	/**
+	 * Dynamax Level. Affects the amount of HP gained when
+	 * dynamaxed.
```
KrisXV, line 103:
```suggestion
	 * Dynamax Level. Affects the amount of HP gained when Dynamaxed.
```

KrisXV (approved):

AnnikaCodes:
Looks good, thanks, and sorry for the delay!

PR #8310, opened 2021-05-16
Smogon master thing

PlagueVonKarma:


PlagueVonKarma:
sry i clicked something stupid by mistake

PR #8313, opened 2021-05-16
Add various mechanics tests

DaWoblefet:
This adds 10 new skipped tests, fixes some of the Sky Drop glitch tests that were failing because of bad syntax and not because the condition wouldn't have worked, adds passing unit tests for things missing it that probably should have it (e.g. Gulp Missile), and cleaned up many tests to use our standard style of checking things (obviously not comprehensive, I just tackled a file if I happened to be working in it). [Updated as of here with bug reports](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8860308).

As of this PR, I think we have 39 skipped mechanics tests and 3 major sets of tests not implemented (the three are glitches: Symbiosis Eject Button, Sky Drop Gravity, and Rollout Storage). We're generally speaking doing pretty good about keeping these in check! Client definitely has more repeated mechanics bugs overall throughout the bug reports thread.

### DIFF
--- a/test/sim/misc/target-resolution.js
+++ b/test/sim/misc/target-resolution.js
@@ -323,4 +323,23 @@ describe('Target Resolution', function () {
 		battle.makeChoices('auto', 'move phantomforce 1, move sheercold 1');
 		assert.false.fullHP(battle.p1.active[1], 'Altaria should not be at full HP, because Phantom Force was redirected and targeted it.');
 	});
+
+	it.skip(`should cause Rollout to target the same slot after being called as a submove`, function () {
+		// hardcoded RNG seed to show the erroneous targeting behavior
+		battle = common.createBattle({gameType: 'doubles', seed: [1, 2, 3, 4]}, [[
+			{species: 'purrloin', ability: 'compoundeyes', moves: ['rollout', 'sleeptalk']},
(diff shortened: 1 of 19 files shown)


DaWoblefet:
Ready for review!

Marty-D (approved):
Looks good, thanks! 👍

test/sim/misc/target-resolution.js
```diff
@@ -323,4 +323,23 @@ describe('Target Resolution', function () {
...
+		// hardcoded RNG seed to show the erroneous targeting behavior
+		battle = common.createBattle({gameType: 'doubles', seed: [1, 2, 3, 4]}, [[
+			{species: 'purrloin', ability: 'compoundeyes', moves: ['rollout', 'sleeptalk']},
+			{species: 'regieleki', moves: ['healbell', 'spore']},
+		], [
+			{species: 'aggron', moves: ['sleeptalk']},
+			{species: 'slowbro', moves: ['sleeptalk']},
+		]]);
+
+		battle.makeChoices('move sleeptalk, move spore -1', 'auto');
```
MathyFurret, line 337: This test is wrong, as sleep prevents Rollout from locking. I am working on this bug so I will write a new test using Copycat instead.

PR #8314, opened 2021-05-16
Fix Gen 3 Weather Ball and Gen 4 Trick

pacmanboss256:
Gen 3 weather ball no longer triggers mirror coat

Gen 4 trick fails when the target has multitype

Also added a test for sleep talk PP deduction in gen 4

### DIFF
--- a/data/mods/gen3/moves.ts
+++ b/data/mods/gen3/moves.ts
@@ -137,10 +137,9 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 			onDamagePriority: -101,
 			onDamage(damage, target, source, effect) {
-				const weatherball = (effect.id === 'weatherball' && this.field.effectiveWeather() !== '');
 				if (
 					effect.effectType === 'Move' && !source.isAlly(target) &&
-					this.getCategory(effect.id) === 'Physical' && !weatherball
+					(effect.category === 'Physical' || effect.id === 'hiddenpower')
 				) {
 					this.effectState.slot = source.getSlot();
 					this.effectState.damage = 2 * damage;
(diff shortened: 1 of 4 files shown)


data/mods/gen3/moves.ts
```diff
@@ -413,7 +413,8 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 			onDamagePriority: -101,
 			onDamage(damage, target, source, effect) {
-				if (effect.effectType === 'Move' && !source.isAlly(target) && this.getCategory(effect.id) === 'Special') {
+				if (effect.effectType === 'Move' && !source.isAlly(target) &&
+				this.getCategory(effect.id) === 'Special' && effect.id !== 'weatherball') {
```
Zarel, line 417:
```suggestion
				if (
					effect.effectType === 'Move' && !source.isAlly(target) &&
					this.getCategory(effect.id) === 'Special' && effect.id !== 'weatherball'
				) {
```
Really wish eslint could enforce this :s

Marty-D:
None of this is correct.
- Mirror Coat is supposed to work on Weather Ball if the attack was a Special type, it's Counter which should not
- Trick also fails if the user has Multitype, and Switcheroo exists in Gen 4 too
- The Sleep Talk PP thing is regarding Choice items and Encore, not failing in general

pacmanboss256:
bump

data/mods/gen3/moves.ts
```diff
@@ -414,7 +418,11 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 			onDamagePriority: -101,
 			onDamage(damage, target, source, effect) {
-				if (effect.effectType === 'Move' && !source.isAlly(target) && this.getCategory(effect.id) === 'Special') {
+				const weatherball = (effect.id === 'weatherball' && this.field.effectiveWeather() !== '');
+				if (
+					effect.effectType === 'Move' && !source.isAlly(target) &&
+					(this.getCategory(effect.id) === 'Special' || weatherball)
+				) {
```
Marty-D, line 425: Weather Ball can be Rock type. Re-implementing all this Weather Ball logic in Counter and Mirror Coat doesn't seem like a great idea though.

pacmanboss256: what do you suggest i do

Karthik99999: Pretty sure you can just check for `effect.category` and see if it's special. I'm pretty sure the only reason `effect.category` is not used in the first place is because of hidden power always being considered physical for counter/mirror coat, but I could be wrong on that.

pacmanboss256: yep that seems to work, however i will need to hardcode hidden power i think

Marty-D (approved):
👍

PR #8315, opened 2021-05-16
Fix Struggle to be unaffected by Disable

DaWoblefet:
I bet you could think of some epic endless battle strats with this if we left it alone though!

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -583,7 +583,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	cursedbody: {
 		onDamagingHit(damage, target, source, move) {
 			if (source.volatiles['disable']) return;
-			if (!move.isMax && !move.isFutureMove) {
+			if (!move.isMax && !move.isFutureMove && move.id !== 'struggle') {
 				if (this.randomChance(3, 10)) {
 					source.addVolatile('disable', this.effectState.target);
 				}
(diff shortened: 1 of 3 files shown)


Zarel:
Thanks!

urkerab:
> I bet you could think of some epic endless battle strats with this if we left it alone though!

https://www.youtube.com/watch?v=uGwLzQuNTOo

Zarel:
Actually, is this the correct fix? This shouldn't be a hardcode, right? Because Disable should just fail if it tries to disable a move that you don't have in your move list?

urkerab:
Cursed Body can disable a move that you don't have in your move list.

DaWoblefet:
> Actually, is this the correct fix? This shouldn't be a hardcode, right? Because Disable should just fail if it tries to disable a move that you don't have in your move list?

Fwiw that's how Struggle and Disable interact on the actual games (being explicitly restricted). There is an associated check with Disable for not being able to Disable moves that aren't in your movelist, but as urkerab says, Cursed Body doesn't have a check like that, which is actually super funny thinking about it. But for Struggle's case, it means that even if you hack Struggle onto a Pokemon's moveslot, both Cursed Body and Disable can't actually Disable it.

Zarel:
Good to know, thanks! But it does sound like we don't have that associated check for Disable, so should add that check, too?

DaWoblefet:
> Good to know, thanks! But it does sound like we don't have that associated check for Disable, so should add that check, too?

I tested it out and the behavior already seems to be in place. For example, using Pound -> Dancer -> call Swords Dance and then attempt to Disable will correctly disable Pound. Minimally, including ``target.moves.includes(target.lastMove.id)`` as one of the conditions for Disable failing seems to break correct behavior, so I'll leave it alone.

PR #8319, opened 2021-05-19
Chat: Support exporting custom handlers from plugins

mia-pi-git:
More convenient this way.

### DIFF
--- a/server/chat-plugins/daily-spotlight.ts
+++ b/server/chat-plugins/daily-spotlight.ts
@@ -256,7 +256,7 @@ export const commands: Chat.ChatCommands = {
 	},
 };
 
-export const hooks: Chat.Handlers = {
+export const handlers: Chat.Handlers = {
 	onRenameRoom(oldID, newID) {
 		if (spotlights[oldID]) {
 			if (!spotlights[newID]) spotlights[newID] = {};
--- a/server/chat-plugins/daily-spotlight.ts
+++ b/server/chat-plugins/daily-spotlight.ts
@@ -256,7 +256,7 @@ export const commands: Chat.ChatCommands = {
 	},
 };
 
-export const onRenameRoom: Rooms.RenameHandler = (oldID, newID) => {
+export const onRenameRoom: Chat.Handlers.RenameRoom = (oldID, newID) => {
 	if (spotlights[oldID]) {
 		if (!spotlights[newID]) spotlights[newID] = {};
 		Object.assign(spotlights[newID], spotlights[oldID]);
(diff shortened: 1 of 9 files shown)


server/chat-plugins/daily-spotlight.ts
```diff
@@ -256,7 +256,7 @@ export const commands: Chat.ChatCommands = {
 	},
 };
 
-export const onRenameRoom: Rooms.RenameHandler = (oldID, newID) => {
+export const onRenameRoom: Chat.Handlers.RenameRoom = (oldID, newID) => {
```
Zarel, line 259: What if:

```ts
export const handlers: Chat.Handlers = {
  onRenameRoom(oldID, newID) {
  }
};
```

Would provide some more safety, too!

Zarel: Although also consider "listeners" or "hooks"

mia-pi-git: Done. Renamed to `hooks` as well.

Zarel: I should probably not have suggested "hooks" since it's my least favorite of the three. I like "listeners" or "handlers" better.

Zarel: Or even "events".

mia-pi-git: Alright. Personally, I prefer hooks since it's shortest, but handlers will do?

mia-pi-git: Done.

Zarel: Yeah, handlers is probably best.

server/chat.ts
```diff
@@ -1831,8 +1839,11 @@ export const Chat = new class {
 		if (plugin.punishmentfilter) Chat.punishmentfilters.push(plugin.punishmentfilter);
 		if (plugin.nicknamefilter) Chat.nicknamefilters.push(plugin.nicknamefilter);
 		if (plugin.statusfilter) Chat.statusfilters.push(plugin.statusfilter);
-		if (plugin.onRenameRoom) Chat.renameHandlers.push(plugin.onRenameRoom);
-		if (plugin.onCloseRoom) Chat.closeRoomHandlers.push(plugin.onCloseRoom);
```
Zarel, line 1835: No backwards compatibility?

mia-pi-git: Done.

Zarel:
Thanks!

PR #8322, opened 2021-05-21
Tournaments: Add feedback to /tour runautodq

sirDonovan:


AnnikaCodes:
This is probably a good idea, thanks!

PR #8324, opened 2021-05-22
Stomping tantrum smack down fix

pacmanboss256:


data/moves.ts
```diff
@@ -16238,7 +16239,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		secondary: {
 			dustproof: true,
 			chance: 100,
-			onHit(target) {
+			onHit(target, source, move) {
```
Karthik99999, line 16242:
```suggestion
			onHit(target) {
```
no need to add `source` and `move` when you're not using them (and when they weren't there in the first place)

Zarel (approved):
Looks safe enough. I like it. @Marty-D?

Marty-D:
Shouldn't this just be part of `cancelMove()`? I can't imagine subsequent Smack Downs on already smacked down Fly/Bounce users don't behave like this.

pacmanboss256:
Would this affect other moves that use `cancelMove()` or do they all not trigger stomping tantrum

Zarel:
Uses of `cancelMove` that may be relevant:

- Gravity canceling the second turn of Bounce, Fly, or Sky Drop
- Smack Down canceling the second turn of Bounce or Fly
- Pursuit moving itself up in priority if someone tries to switch
- Quash moving a move down in priority in Gen 7 and earlier

Quash and Pursuit can be refactored if they don't match. So the only question left is Gravity.

Zarel:
I would be very surprised if Gravity worked differently from Smack Down, though. So sure, let's put it inside `cancelMove` and change Pursuit/Quash to manually rearrange queue order.

DaWoblefet:
Yeah, to be explicit, Fly being interrupted by Gravity doesn't double Stomping Tantrum.

pacmanboss256:
not sure how to change pursuit, and quash doesnt use `cancelMove()`

pacmanboss256:
ignore the commit message branch names got messed up in vscode

PR #8325, opened 2021-05-22
Gen 3: Always break the opponents screens with Brick Break

Karthik99999:
Wasn't mentioned as a past gen bug, but saw it while reading up on the other old gen Brick Break bug on Bulbapedia and noticed this behavior wasn't implemented.

### DIFF
--- a/test/sim/moves/brickbreak.js
+++ b/test/sim/moves/brickbreak.js
@@ -94,13 +94,13 @@ describe('Brick Break', function () {
 		assert.false(battle.p2.sideConditions['reflect']);
 	});
 
-	it('should break the foe\'s Reflect when used against an ally in Gen 3', function () {
+	it(`should break the foe's Reflect when used against an ally in Gen 3`, function () {
 		battle = common.gen(3).createBattle({gameType: 'doubles'}, [[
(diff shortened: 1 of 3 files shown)


pacmanboss256:
i was working on this and it didnt seem to work like that

test/sim/moves/brickbreak.js
```diff
@@ -93,4 +93,20 @@ describe('Brick Break', function () {
 		battle.makeChoices('move brickbreak', 'move electrify');
 		assert.false(battle.p2.sideConditions['reflect']);
 	});
+
+	it('should break the foe\'s Reflect when used against an ally in Gen 3', function () {
```
DaWoblefet, line 97:
```suggestion
	it(`should break the foe's Reflect when used against an ally in Gen 3`, function () {
```

pacmanboss256:
make sure to update gen 4 as well

Karthik99999:
> make sure to update gen 4 as well

This behavior I'm trying to implement (where the opponent's screens are always broken regardless of target) is only in gen 3.

Zarel (approved):
Looks like a good solution to me. @Marty-D?

Marty-D (approved):
Yep, this is great, thanks!

PR #8326, opened 2021-05-22
Gen 3-4: Always break opponent screens with brick break

pacmanboss256:


data/mods/gen3/moves.ts
```diff
@@ -98,6 +98,9 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		accuracy: 100,
 		ignoreAccuracy: true,
 	},
+	brickbreak: {
+		inherit: true,
```
Karthik99999, line 102: you don't need to define inherit here, gen 3 already inherits from gen 4

test/sim/moves/brickbreak.js
```diff
@@ -49,6 +49,7 @@ describe('Brick Break', function () {
 		assert(battle.p2.sideConditions['reflect']);
 
 		battle.makeChoices('move brickbreak', 'move splash');
+		common.saveReplay(battle, 'brickbreak.js');
```
Karthik99999, line 52: not needed

pacmanboss256: whoops that was for me to test

pacmanboss256:
this worked on my end so idk whats up, but i have to redo this to make sure the move hits

pacmanboss256:
ignore the hardcoding

PR #8327, opened 2021-05-22
Allow Weezing-Galar to have a VC transfer move without its Hidden Ability

Karthik99999:
hardcoded but I don't think there's another solution (aside from remove Galarian Weezing's HA in the gen 7 mod but that seems dumb)

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1013,7 +1013,7 @@ export class TeamValidator {
 				generation: 2,
 				level: isMew ? 5 : isCelebi ? 30 : 3, // Level 1/2 Pokémon can't be obtained by transfer from RBY/GSC
 				perfectIVs: isMew || isCelebi ? 5 : 3,
-				isHidden: !!(!isWeezingGalar && this.dex.mod('gen7').species.get(species.id).abilities['H']),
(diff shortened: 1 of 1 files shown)


Zarel:
How do you teach it this move in-game?

Karthik99999:
> How do you teach it this move in-game?

Kanto Koffing, which you can obtain from the virtual console with its gen 1-2 specific moves (such as Zap Cannon), can evolve into Galarian Weezing after being transfered to SwSh. Since this Koffing doesn't have its hidden ability due to not having one at the time of the VC transfer release, the resulting Galarian Weezing is allowed to not have its hidden ability as well.

Zarel:
Can you give an example of a VC transfer move?

KrisXV:
> Can you give an example of a VC transfer move?

Zap Cannon, Bide, Rage

Zarel:
I suppose Rage works.

KrisXV:
Zap Cannon is the only move legal in Sw/Sh though

Zarel:
I was asking to write the unit test, but it turns out there's already a Zap Cannon Koffing unit test.

urkerab:
I was thinking that the 7V source would need to include the species being transferred, so that the validator would ensure that if we were in fact transferring a 7V Koffing that there weren't any moves specific to 7V Weezing moves in the moveset, but no such move exists, so that's not a problem.

PR #8328, opened 2021-05-23
Chat-monitor: Centralize functions and migrate to JSON

mia-pi-git:
I centralized the code to make it easier to use, but the main purpose of this PR is to move to safer loading code (see comment at 130) and to a safer loading format (JSON) to prevent current issues. More details on the issue in #private-dev.

### DIFF
--- a/server/chat-plugins/chat-monitor.ts
+++ b/server/chat-plugins/chat-monitor.ts
@@ -1,7 +1,8 @@
 import {FS, Utils} from '../../lib';
 import type {FilterWord} from '../chat';
 
-const MONITOR_FILE = 'config/chat-plugins/chat-monitor.tsv';
+const LEGACY_MONITOR_FILE = 'config/chat-plugins/chat-monitor.tsv';
+const MONITOR_FILE = 'config/chat-plugins/chat-filter.json';
 const WRITE_THROTTLE_TIME = 5 * 60 * 1000;
 
 // Substitution dictionary adapted from https://github.com/ThreeLetters/NoSwearingPlease/blob/master/index.js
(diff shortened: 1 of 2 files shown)


server/chat-plugins/chat-monitor.ts
```diff
@@ -1,7 +1,8 @@
 import {FS, Utils} from '../../lib';
 import type {FilterWord} from '../chat';
 
-const MONITOR_FILE = 'config/chat-plugins/chat-monitor.tsv';
+const LEGACY_MONITOR_FILE = 'config/chat-plugins/chat-monitor.tsv';
+const MONITOR_FILE = 'config/chat-plugins/chat-filter.tsv';
```
KrisXV, line 5:
```suggestion
const MONITOR_FILE = 'config/chat-plugins/chat-filter.json';
```
?

mia-pi-git: whoops, typo

server/chat-plugins/chat-monitor.ts
```diff
@@ -45,74 +46,155 @@ const EVASION_DETECTION_SUBSTITUTIONS: {[k: string]: string[]} = {
...
-export function generateRegex(word: string, isEvasion = false, isShortener = false, isReplacement = false) {
-	try {
-		if (isEvasion) {
-			return constructEvasionRegex(word);
-		} else {
-			return new RegExp((isShortener ? `\\b${word}` : word), (isReplacement ? 'igu' : 'iu'));
+		const data = JSON.parse(FS(MONITOR_FILE).readIfExistsSync() || "{}");
+		for (const k in data) {
+			filterWords[k] = [];
+			// previously, this checked to be sure the monitor existed in Chat.monitors and that there was
```
DaWoblefet, line 130: /* */ yn

Zarel: I think a row of `//`s is fine; I don't think we have a use case for `/*` other than doc comments where they're `/**`.

Zarel:
Sure, let's try this out.

PR #8330, opened 2021-05-24
/ipsearch: Support low-high syntax for ranges

mia-pi-git:
Requested functionality, along with cleaning of the code.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -426,7 +426,7 @@ export const commands: Chat.ChatCommands = {
 		this.checkCan('rangeban');
 
 		let [ip, roomid] = this.splitOne(target);
-		const targetRoom = roomid && !IPTools.ipRegex.test(roomid) ? Rooms.get(roomid) : null;
+		const targetRoom = roomid ? Rooms.get(roomid) : null;
 		if (typeof targetRoom === 'undefined') {
 			return this.errorReply(`The room "${roomid}" does not exist.`);
 		}
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -426,48 +426,51 @@ export const commands: Chat.ChatCommands = {
 		this.checkCan('rangeban');
 
 		let [ip, roomid] = this.splitOne(target);
-		const targetRoom = roomid ? Rooms.get(roomid) : null;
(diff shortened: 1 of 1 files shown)


Zarel:
What is a "specific range"? You don't seem to have changed the help at all.

server/chat-commands/info.ts
```diff
@@ -426,48 +426,51 @@ export const commands: Chat.ChatCommands = {
 		this.checkCan('rangeban');
 
 		let [ip, roomid] = this.splitOne(target);
-		const targetRoom = roomid ? Rooms.get(roomid) : null;
+		const targetRoom = roomid && !IPTools.ipRegex.test(roomid) ? Rooms.get(roomid) : null;
```
Zarel, line 429:
```suggestion
		const targetRoom = roomid ? Rooms.get(roomid) : null;
```

server/chat-commands/info.ts
```diff
@@ -426,48 +426,51 @@ export const commands: Chat.ChatCommands = {
 		this.checkCan('rangeban');
 
 		let [ip, roomid] = this.splitOne(target);
```
Zarel:
```suggestion
		let [ipOrHost, roomid] = this.splitOne(target);
```

Zarel: nvm, too much to suggest, I'll just push the rename

server/chat-commands/info.ts
```diff
@@ -430,44 +430,47 @@ export const commands: Chat.ChatCommands = {
...
-				if (results.length > 100 && !isAll) continue;
-				if (!curUser.latestIp.startsWith(ip)) continue;
-				if (targetRoom && !curUser.inRooms.has(targetRoom.roomid)) continue;
-				results.push(`${curUser.connected ? ONLINE_SYMBOL : OFFLINE_SYMBOL} ${curUser.name}`);
-			}
-			if (results.length > 100 && !isAll) {
-				return this.sendReply(`More than 100 users match the specified IP range. Use /ipsearchall to retrieve the full list.`);
-			}
+		} else if (target.split('-').every(i => IPTools.ipRegex.test(i))) {
+			const [low, high] = target.split(',').map(i => i.trim());
```
Zarel, line 451:
```suggestion
			const [low, high] = target.split('-').map(i => i.trim());
```

ಠ_ಠ

PR #8331, opened 2021-05-24
IPTools: Add sanity checks

AnnikaCodes:
This is probably a good idea.

### DIFF
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -176,7 +176,9 @@ export const IPTools = new class {
 	stringToRange(range: string): AddressRange | null {
 		if (!IPTools.isValidRange(range)) return null;
 		if (range.endsWith('*')) {
-			const [a, b, c] = range.replace('*', '').split('.');
+			const parts = range.replace('*', '').split('.');
+			if (parts.length > 3) return null;
+			const [a, b, c] = parts;
 			return {
 				minIP: IPTools.ipToNumber(`${a || '0'}.${b || '0'}.${c || '0'}.0`),
 				maxIP: IPTools.ipToNumber(`${a || '255'}.${b || '255'}.${c || '255'}.255`),
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -175,7 +175,6 @@ export const IPTools = new class {
 		return this.ipRegex.test(ip.trim()) && /^[1-3]?[0-9]$/.test(bits.trim());
 	}
 	stringToRange(range: string): AddressRange | null {
-		if (!IPTools.isValidRange(range)) return null;
 		if (range.endsWith('*')) {
 			const parts = range.replace('.*', '').split('.');
 			if (parts.length > 3) return null;
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -164,9 +164,7 @@ export const IPTools = new class {
 
 			const minIPNumber = IPTools.ipToNumber(minIP);
 			const maxIPNumber = IPTools.ipToNumber(maxIP);
-			if (minIPNumber < 0 || minIPNumber > maxIPNumber) return false;
-
-			return this.ipRegex.test(minIP.trim()) && this.ipRegex.test(maxIP.trim());
+			return minIPNumber >= 0 && minIPNumber < maxIPNumber;
 		}
 
 		// "127.0.0.0/24" format
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -119,7 +119,7 @@ export const IPTools = new class {
 		for (const part of parts) {
 			num *= 256;
 
-			const partAsInt = parseInt(part);
+			const partAsInt = Utils.parseExactInt(part);
 			if (isNaN(partAsInt) || partAsInt < 0 || partAsInt > 255) return -1;
 			num += partAsInt;
 		}
@@ -146,7 +146,7 @@ export const IPTools = new class {
 			return {minIP: ip, maxIP: ip};
 		}
 		const low = IPTools.ipToNumber(cidr.slice(0, index));
-		const bits = parseInt(cidr.slice(index + 1));
+		const bits = Utils.parseExactInt(cidr.slice(index + 1));
 		// fun fact: IPTools fails if bits <= 1 because JavaScript
 		// does << with signed int32s.
 		const high = low + (1 << (32 - bits)) - 1;
@@ -173,9 +173,8 @@ export const IPTools = new class {
 		const [ip, bits] = cidrParts;
 		return this.ipRegex.test(ip.trim()) && /^[1-3]?[0-9]$/.test(bits.trim());
 	}
-	/** does not check for validity; use `validRange` for that; supports both range formats */
 	stringToRange(range: string): AddressRange | null {
-		if (!range) return null;
+		if (!IPTools.isValidRange(range)) return null;
 		if (range.endsWith('*')) {
 			const [a, b, c] = range.replace('*', '').split('.');
 			return {
--- a/server/ip-tools.ts
+++ b/server/ip-tools.ts
@@ -109,18 +109,18 @@ export const IPTools = new class {
 		ip = ip.trim();
 		if (ip.includes(':') && !ip.includes('.')) {
 			// IPv6, which PS does not support
-			return -1;
+			return null;
 		}
 		if (ip.startsWith('::ffff:')) ip = ip.slice(7);
 		else if (ip.startsWith('::')) ip = ip.slice(2);
 		let num = 0;
 		const parts = ip.split('.');
-		if (parts.length !== 4) return -1;
+		if (parts.length !== 4) return null;
 		for (const part of parts) {
 			num *= 256;
 
 			const partAsInt = Utils.parseExactInt(part);
-			if (isNaN(partAsInt) || partAsInt < 0 || partAsInt > 255) return -1;
+			if (isNaN(partAsInt) || partAsInt < 0 || partAsInt > 255) return null;
 			num += partAsInt;
 		}
 		return num;
@@ -142,14 +142,14 @@ export const IPTools = new class {
 		const index = cidr.indexOf('/');
 		if (index <= 0) {
 			const ip = IPTools.ipToNumber(cidr);
-			if (ip < 0) return null;
+			if (ip === null) return null;
 			return {minIP: ip, maxIP: ip};
 		}
 		const low = IPTools.ipToNumber(cidr.slice(0, index));
 		const bits = Utils.parseExactInt(cidr.slice(index + 1));
 		// fun fact: IPTools fails if bits <= 1 because JavaScript
 		// does << with signed int32s.
-		if (!bits || bits < 2 || bits > 32) return null;
+		if (low === null || !bits || bits < 2 || bits > 32) return null;
 		const high = low + (1 << (32 - bits)) - 1;
 		return {minIP: low, maxIP: high};
 	}
@@ -165,7 +165,7 @@ export const IPTools = new class {
 
 			const minIPNumber = IPTools.ipToNumber(minIP);
 			const maxIPNumber = IPTools.ipToNumber(maxIP);
-			return minIPNumber >= 0 && minIPNumber < maxIPNumber;
+			return minIPNumber !== null && maxIPNumber !== null && minIPNumber < maxIPNumber;
 		}
 
 		// "127.0.0.0/24" format
@@ -177,23 +177,26 @@ export const IPTools = new class {
 	stringToRange(range: string): AddressRange | null {
 		if (!IPTools.isValidRange(range)) return null;
 		if (range.endsWith('*')) {
-			const parts = range.replace('*', '').split('.');
+			const parts = range.replace('.*', '').split('.');
 			if (parts.length > 3) return null;
 			const [a, b, c] = parts;
-			return {
-				minIP: IPTools.ipToNumber(`${a || '0'}.${b || '0'}.${c || '0'}.0`),
-				maxIP: IPTools.ipToNumber(`${a || '255'}.${b || '255'}.${c || '255'}.255`),
-			};
+			const minIP = IPTools.ipToNumber(`${a || '0'}.${b || '0'}.${c || '0'}.0`);
+			const maxIP = IPTools.ipToNumber(`${a || '255'}.${b || '255'}.${c || '255'}.255`);
+			if (minIP === null || maxIP === null) return null;
+			return {minIP, maxIP};
 		}
 		const index = range.indexOf('-');
 		if (index <= 0) {
-			return range.includes('/') ? IPTools.getCidrRange(range) : {
-				minIP: IPTools.ipToNumber(range),
-				maxIP: IPTools.ipToNumber(range),
-			};
+			if (range.includes('/')) return IPTools.getCidrRange(range);
+			const ip = IPTools.ipToNumber(range);
+			if (ip === null) return null;
+
+			return {maxIP: ip, minIP: ip};
 		}
 		const minIP = IPTools.ipToNumber(range.slice(0, index));
 		const maxIP = IPTools.ipToNumber(range.slice(index + 1));
+
+		if (minIP === null || maxIP === null) return null;
 		return {minIP, maxIP};
 	}
 
@@ -224,7 +227,11 @@ export const IPTools = new class {
 		} else {
 			ranges = rangeString.map(IPTools.stringToRange).filter(x => x) as AddressRange[];
 		}
-		return (ip: string) => IPTools.checkPattern(ranges, IPTools.ipToNumber(ip));
+		return (ip: string) => {
+			const ipNumber = IPTools.ipToNumber(ip);
+			if (ipNumber === null) return false; // default to assuming that an invalid IP is not in the range
+			return IPTools.checkPattern(ranges, ipNumber);
+		};
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


Zarel:
Interesting that d8f492 didn't conflict with this at all.

AnnikaCodes:
yeah, do you want me to move the `stringToRange` validation into `isValidRange`?

Zarel:
It depends on how much we care about perf. The `stringToRange` validation has a few regex executions. Unlikely to be a huge deal, but we should probably validate them _somehow_.

test/server/ip-tools.js
```diff
@@ -33,6 +33,20 @@ describe("IP tools", () => {
 		assert.deepEqual(cidrRange, stringRange);
 	});
 
+	it('should not parse invalid ranges', () => {
+		assert.equal(IPTools.stringToRange('42.42.10.0 - 42.42.5.0'), null);
+		assert.equal(IPTools.stringToRange('250.0.0.0 - 260.0.0.0'), null);
```
Zarel, line 38:
```suggestion
		assert.equal(IPTools.stringToRange('250.0.0.0 - 260.0.0.0'), null);
		assert.equal(IPTools.stringToRange('250.0.0.0/43'), null);
		assert.equal(IPTools.stringToRange('250.0.0.0/16x'), null);
		assert.equal(IPTools.stringToRange('250.0.0.0.1/24'), null);
		assert.equal(IPTools.stringToRange('250.0.0.0.1 - 250.0.0.2'), null);
```

This PR might not be able to validate all of these, but it probably should.

Zarel:
Oh, wait. Yes, you should do that, but I think currently your validation is less thorough than `isValidRange`.

test/server/ip-tools.js
AnnikaCodes: Sure.

test/server/ip-tools.js
```diff
@@ -33,6 +33,27 @@ describe("IP tools", () => {
...
+		assert.equal(IPTools.stringToRange('250.0.0.0/43'), null);
+		assert.equal(IPTools.stringToRange('250.0.0.0/16x'), null);
+		assert.equal(IPTools.stringToRange('250.0.0.0.1/24'), null);
+		assert.equal(IPTools.stringToRange('250.0/24'), null);
+		assert.equal(IPTools.stringToRange('250.0.0.0.1 - 250.0.0.2'), null);
+		assert.equal(IPTools.stringToRange('250.0.0.0.*'), null);
+		assert.equal(IPTools.stringToRange('250.0.0'), null);
```
Zarel, line 45:
```suggestion
		assert.equal(IPTools.stringToRange('250.0.0.0.1 - 250.0.0.2'), null);
		assert.equal(IPTools.stringToRange('250.0.0.0.*'), null);
```

(Actually, leaving off octets is valid; `127.1` is shorthand for `127.0.0.1`)

test/server/ip-tools.js
```diff
@@ -33,6 +33,27 @@ describe("IP tools", () => {
...
+	});
+
```
Zarel, line 47:
```suggestion

	it('should correctly parse shorthand IPs', () => {
		assert.equal(IPTools.ipToNumber('127.0.0.1'), IPTools.ipToNumber('127.1'));
	});

```

Zarel: It's somewhat questionable what we should do here. Rejecting them for not being valid IPs is also plausibly valid.

test/server/ip-tools.js
```diff
@@ -33,6 +33,30 @@ describe("IP tools", () => {
...
+	it('should correctly parse shorthand IPs', () => {
+		assert.equal(IPTools.ipToNumber('127.1'), IPTools.ipToNumber('127.0.0.1'));
+		assert.equal(IPTools.ipToNumber('192.168.257'), IPTools.ipToNumber('192.168.1.1'));
+	});
+
+	it('should reject invalid IPs', () => {
+		assert.equal(IPTools.ipToNumber('256.0.0.0'), -1);
+		assert.equal(IPTools.ipToNumber('42.0.0.1111'), -1);
+		assert.equal(IPTools.ipToNumber('42.0.hi.0'), -1);
```
Zarel, line 54:
```suggestion
		assert.equal(IPTools.ipToNumber('42.0.hi.0'), -1);
		assert.equal(IPTools.ipToNumber('256.0.0.0.1'), -1);
```

test/server/ip-tools.js
AnnikaCodes: I'm not really familiar with IP shorthand, but I don't think it would be used internally by PS. The question is whether it's worth supporting for user-facing commands (if so, the `ipRegex` should probably be updated to support it).

Zarel: It's probably not worth supporting. I've never seen anyone use it in my life; I just know of its existence because it used to be a fun way to get around internet filters - if an IP is banned in one format, just try another format!

AnnikaCodes: lmao, sounds like fun, but not something we need to support

server/ip-tools.ts
```diff
@@ -117,7 +117,10 @@ export const IPTools = new class {
 		const parts = ip.split('.');
```
Zarel, line 117:
```suggestion
		const parts = ip.split('.');
		if (parts.length !== 4) return -1;
```

Zarel:
This file should probably define a `parseExactInt` and use it in place of `parseInt`.

`parseInt` doesn't do validation:

```
> parseInt('12jaiow')
12
```

Or maybe it should be in Utils?

```ts
/** Like parseInt, but returns NaN if the int isn't already in normalized form */
function parseExactInt(str: string): number {
  if (!/^-?[1-9][0-9]*$/.test(str)) return NaN;
  return parseInt(str);
}
```

AnnikaCodes:
That definitely feels like a `Utils` function; it would be useful outside of IPTools.

test/server/ip-tools.js
```diff
@@ -33,6 +33,26 @@ describe("IP tools", () => {
...
+		assert.equal(IPTools.isValidRange('250.0.0.0.1/24'), false);
+		assert.equal(IPTools.isValidRange('250.0.0.0.1 - 250.0.0.2'), false);
+		assert.equal(IPTools.isValidRange('250.0.0.0.*'), false);
+	});
+
+	it('should reject invalid IPs', () => {
+		assert.equal(IPTools.ipToNumber('256.0.0.0'), -1);
+		assert.equal(IPTools.ipToNumber('42.0.0.1111'), -1);
+		assert.equal(IPTools.ipToNumber('42.0.hi.0'), -1);
+		assert.equal(IPTools.ipToNumber('256.0.0.0.1'), -1);
```
Zarel, line 50:
```suggestion
		assert.equal(IPTools.ipToNumber('256.0.0.0.1'), -1);
		assert.equal(IPTools.ipToNumber('256.0.0hi.1'), -1);
```

server/ip-tools.ts
```diff
@@ -157,6 +161,11 @@ export const IPTools = new class {
 		const ips = range.split('-');
 		if (ips.length === 2) {
 			const [minIP, maxIP] = ips;
+
+			const minIPNumber = IPTools.ipToNumber(minIP);
+			const maxIPNumber = IPTools.ipToNumber(maxIP);
+			if (minIPNumber < 0 || minIPNumber > maxIPNumber) return false;
+
 			return this.ipRegex.test(minIP.trim()) && this.ipRegex.test(maxIP.trim());
```
Zarel, line 169: If `ipToNumber` does validation as it's intended to, the regex should be unnecessary.

```suggestion
			return minIPNumber >= 0 && minIPNumber < maxIPNumber;
```

Zarel:
I do, in fact, think `stringToRange` should do validation, though. It should be easier than having `isValidRange` do validation, if all the parsing functions do in fact return `null` for errors.

server/ip-tools.ts
```diff
@@ -129,20 +133,20 @@ export const IPTools = new class {
 			num = (num - part) / 256;
 			ipParts.unshift(part.toString());
 		}
+		if (ipParts.length !== 4) return null;
```
Zarel, line 136: image

```suggestion
		while (ipParts.length < 4) ipParts.unshift(0);
		if (ipParts.length !== 4) return null;
```

server/ip-tools.ts
```diff
@@ -108,16 +108,20 @@ export const IPTools = new class {
 	ipToNumber(ip: string) {
```
Zarel, line 108: Should this return `number | null`? `null` is a better error state (easier for TypeScript to catch if you forgot to handle it).

server/ip-tools.ts
```diff
@@ -129,20 +133,20 @@ export const IPTools = new class {
...
-			};
+			const ip = IPTools.ipToNumber(cidr);
+			if (ip < 0) return null;
+			return {minIP: ip, maxIP: ip};
 		}
 		const low = IPTools.ipToNumber(cidr.slice(0, index));
-		const bits = parseInt(cidr.slice(index + 1));
+		const bits = Utils.parseExactInt(cidr.slice(index + 1));
 		// fun fact: IPTools fails if bits <= 1 because JavaScript
 		// does << with signed int32s.
```
Zarel, line 151:
```suggestion
		// fun fact: IPTools fails if bits <= 1 because JavaScript
		// does << with signed int32s.
		if (!bits || bits < 2 || bits > 32) return null;
```

server/ip-tools.ts
```diff
@@ -166,9 +173,8 @@ export const IPTools = new class {
 		const [ip, bits] = cidrParts;
 		return this.ipRegex.test(ip.trim()) && /^[1-3]?[0-9]$/.test(bits.trim());
 	}
-	/** does not check for validity; use `validRange` for that; supports both range formats */
 	stringToRange(range: string): AddressRange | null {
-		if (!range) return null;
+		if (!IPTools.isValidRange(range)) return null;
 		if (range.endsWith('*')) {
 			const [a, b, c] = range.replace('*', '').split('.');
```
Zarel, line 179:
```suggestion
			const parts = range.replace('*', '').split('.');
			if (parts.length > 3) return null;
			const [a, b, c] = parts;
```

server/ip-tools.ts
AnnikaCodes: Yeah, probably

server/ip-tools.ts
```diff
@@ -166,25 +174,29 @@ export const IPTools = new class {
 		const [ip, bits] = cidrParts;
 		return this.ipRegex.test(ip.trim()) && /^[1-3]?[0-9]$/.test(bits.trim());
 	}
-	/** does not check for validity; use `validRange` for that; supports both range formats */
 	stringToRange(range: string): AddressRange | null {
-		if (!range) return null;
+		if (!IPTools.isValidRange(range)) return null;
```
Zarel, line 178:
```suggestion
```

I'm pretty sure this is no longer necessary, and also `isValidRange` might be replaceable with `stringToRange(str) !== null`. If you have free time, you might want to benchmark that to see if it's faster.

test/server/ip-tools.js
```diff
@@ -33,6 +33,27 @@ describe("IP tools", () => {
...
+		assert.equal(IPTools.isValidRange('250.0.0.0.1 - 250.0.0.2'), false);
+		assert.equal(IPTools.isValidRange('250.0.0.0.*'), false);
+	});
+
+	it('should reject invalid IPs', () => {
+		assert.equal(IPTools.ipToNumber('256.0.0.0'), null);
+		assert.equal(IPTools.ipToNumber('42.0.0.1111'), null);
+		assert.equal(IPTools.ipToNumber('42.0.hi.0'), null);
+		assert.equal(IPTools.ipToNumber('256.0.0.0.1'), null);
+		assert.equal(IPTools.ipToNumber('256.0.0hi.1'), null);
```
Zarel, line 51:
```suggestion
		assert.equal(IPTools.ipToNumber('256.0.0hi.1'), null);
		assert.equal(IPTools.ipToNumber('256.0.1'), null);
```

Since we've committed not to take shorthand IPs, let's make sure they're `null`.

AnnikaCodes: Fair enough.

test/server/ip-tools.js
```diff
@@ -33,6 +33,27 @@ describe("IP tools", () => {
...
+
+		assert.equal(IPTools.numberToIP(4294967296), null);
+		assert.equal(IPTools.numberToIP(-1), null);
```
Zarel, line 55:
```suggestion
		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(1)), 1);
		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(56468451)), 56468451);
```

test/server/ip-tools.js
```diff
@@ -33,6 +33,29 @@ describe("IP tools", () => {
...
+		assert.equal(IPTools.ipToNumber('256.0.0.0'), null);
+		assert.equal(IPTools.ipToNumber('42.0.0.1111'), null);
+		assert.equal(IPTools.ipToNumber('42.0.hi.0'), null);
+		assert.equal(IPTools.ipToNumber('256.0.0.0.1'), null);
+		assert.equal(IPTools.ipToNumber('256.0.0hi.1'), null);
+		assert.equal(IPTools.ipToNumber('256.0.1'), null);
+
+		assert.equal(IPTools.numberToIP(4294967296), null);
+		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(1)), 1);
+		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(56468451)), 56468451);
```
Zarel, line 56:
```suggestion
		assert.equal(IPTools.numberToIP(-1), null);
	});

	it('should be able to convert IPs in both directions', () => {
		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(0)), 0);
		assert.equal(IPTools.numberToIP(IPTools.ipToNumber('0.0.0.1'))), '0.0.0.1');
		assert.equal(IPTools.ipToNumber(IPTools.numberToIP(56468451)), 56468451);
```

Zarel: (On second thought, this belongs in its own block of tests.)

Zarel:
Drafting until you finish the `ipToNumber` `null` conversion.

AnnikaCodes:
> I'm pretty sure this is no longer necessary, and also isValidRange might be replaceable with stringToRange(str) !== null. If you have free time, you might want to benchmark that to see if it's faster.

Yes, it is faster.

AnnikaCodes:
The conversion is done.

server/ip-tools.ts
```diff
@@ -108,83 +108,79 @@ export const IPTools = new class {
...
+			if (isNaN(partAsInt) || partAsInt < 0 || partAsInt > 255) return null;
+			num += partAsInt;
 		}
 		return num;
 	}
 
 	numberToIP(num: number) {
 		const ipParts: string[] = [];
 		while (num) {
+			if (num < 0) return null;
```
Zarel, line 132:
```suggestion
		if (num < 0 || num >= 256 ** 4 || num !== Math.trunc(num)) return null;
		while (num) {
```

server/ip-tools.ts
```diff
@@ -108,83 +108,79 @@ export const IPTools = new class {
...
 			const part = num % 256;
 			num = (num - part) / 256;
 			ipParts.unshift(part.toString());
 		}
+		while (ipParts.length < 4) ipParts.unshift('0');
+		if (ipParts.length !== 4) return null;
```
Zarel, line 138:
```suggestion
```

(obsoleted by earlier suggestion)

server/users.ts
```diff
@@ -159,7 +159,12 @@ function findUsers(userids: ID[], ips: string[], options: {forPunishment?: boole
 		for (const myIp of ips) {
 			if (user.ips.includes(myIp) || (
 				(myIp.includes('*') || myIp.includes('-')) &&
-				user.ips.map(IPTools.ipToNumber).some(ip => {
+				user.ips.map(ip => {
+					const ipNumber = IPTools.ipToNumber(ip);
+					// this should never happen, but if a user object has an invalid IP address, that's pretty concerning
+					if (ipNumber === null) throw new Error(`Invalid IP address on ${user.id}: '${ip}'`);
+					return ipNumber;
+				}).some(ip => {
```
Zarel, line 167: I believe this is known to happen sometimes due to a SockJS bug, and I really don't want to take down the entire server in that case.

mia-pi-git: We already handle that case in

Zarel: I think a lot of these checks can be replaced with having functions like `checkPattern` or `stringToRange` accept `null`.

server/users.ts
```diff
@@ -159,7 +159,12 @@ function findUsers(userids: ID[], ips: string[], options: {forPunishment?: boole
...
 					const range = IPTools.stringToRange(myIp);
 					return range && IPTools.checkPattern([range], ip);
```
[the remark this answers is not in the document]
AnnikaCodes: I don't think we need to cache the ranges in the user object, but we can definitely generate the range in the `for` loop instead of in the `.some`.

server/users.ts
```diff
@@ -157,12 +157,11 @@ function findUsers(userids: ID[], ips: string[], options: {forPunishment?: boole
 			continue;
```
Zarel, line 157: So above `for (const user`, we should add:

```ts
const patterns = ips.map(ip => {
  const range = IPTools.stringToRange(ip);
  if (!range) throw new Chat.ErrorMessage(`Invalid IP or range ${ip}`);
  return range;
});
```

server/users.ts
```diff
@@ -157,12 +157,11 @@ function findUsers(userids: ID[], ips: string[], options: {forPunishment?: boole
...
 		}
 		for (const myIp of ips) {
```
Zarel, line 159: And then this entire for-loop can be

```ts
if (user.ips.some(ip => ipMatcher(ip))) {
  matches.push(user);
}
```

server/users.ts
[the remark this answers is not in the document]
Zarel: The range is specific to the call, by "cache the range" I just mean "put it in a variable at the top level of the function" rather than re-generating it O(n*k) times

server/users.ts
AnnikaCodes: OK.

Zarel (approved):
Looks good! You have plenty of time before the next restart to make more functions accept `number | null` or `string | null` if you wanted to do that.

PR #8333, opened 2021-05-24
Repl: Make it easier to heapdump processes

mia-pi-git:


mia-pi-git:
Oops.

lib/repl.ts
```diff
@@ -42,6 +42,17 @@ export const Repl = new class {
 		if (!process.listeners('SIGINT').length) {
 			process.once('SIGINT', () => process.exit(128 + 2));
 		}
+		(global as any).heapdump = (targetPath?: string) => {
+			if (!targetPath) targetPath = `${filename}-${Date.now()}`;
```
Zarel, line 46:
```suggestion
			if (!targetPath) targetPath = `${filename}-${new Date().toISOString()}.heapsnapshot`;
```

mia-pi-git: (heapsnapshot is automatically appended)

Zarel: (edited)

Zarel:
Looks good, thanks!

PR #8334, opened 2021-05-25
Sky Drop: Fix message when target is fainted

Karthik99999:
Checks to see if the current target is the one originally picked up by sky drop. I could also check for `this.effectState.target.fainted`, but I thought this would encompass more stuff.


DaWoblefet (approved):

Zarel:
tyty!

PR #8336, opened 2021-05-26
Gen 1: Move Victreebel and Lapras to OU by technicality

PlagueVonKarma:
[So the new RBY OU VR is out](https://www.smogon.com/forums/threads/rby-ou-viability-ranking-mark-ii.3572352/page-16#post-8867170), and it's in a state where Lapras and Victreebel would drop.

However, a new policy is being instated by the RBY Council [where drops are delayed by 2 VR Revisions to increase tier stability](https://www.smogon.com/forums/threads/tiering-in-rby.3682580/post-8867192). I've been chatting with Hipmonlee about it, and to display the "held back" mons we want them put in OU by technicality for the duration. 

image
Here he is agreeing to the OU by technicality proposal and me remembering I can do this 😂


AnnikaCodes:
I'm not familiar with Gen 1 tiering policies, but I trust you to get it right.

PR #8343, opened 2021-05-30
Support HTML in roomfaqs

mia-pi-git:
https://www.smogon.com/forums/threads/show-full-size-images-html-support-within-rfaq.3683406/

### DIFF
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -94,8 +94,7 @@ export const Repeats = new class {
 				return;
 			}
 			const repeatedPhrase = repeat.faq ?
-				visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) :
-				Chat.formatText(phrase, true);
+				visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) : Chat.formatText(phrase, true);
 			const formattedText = repeat.isHTML ? phrase : repeatedPhrase;
 			targetRoom.add(`|html|<div class="infobox">${formattedText}</div>`);
 			targetRoom.update();
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -94,8 +94,8 @@ export const Repeats = new class {
 				return;
 			}
 			const repeatedPhrase = repeat.faq ?
-					visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) :
-					Chat.formatText(phrase, true);
+				visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) :
+				Chat.formatText(phrase, true);
 			const formattedText = repeat.isHTML ? phrase : repeatedPhrase;
 			targetRoom.add(`|html|<div class="infobox">${formattedText}</div>`);
 			targetRoom.update();
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -93,7 +93,9 @@ export const Repeats = new class {
 				this.clearRepeats(targetRoom);
 				return;
 			}
-			const repeatedPhrase = repeat.faq ? visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) : Chat.formatText(phrase, true);
+			const repeatedPhrase = repeat.faq ?
+					visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) :
+					Chat.formatText(phrase, true);
 			const formattedText = repeat.isHTML ? phrase : repeatedPhrase;
 			targetRoom.add(`|html|<div class="infobox">${formattedText}</div>`);
 			targetRoom.update();
--- a/server/chat-plugins/repeats.ts
+++ b/server/chat-plugins/repeats.ts
@@ -4,7 +4,7 @@
  * @author Annika, Zarel
  */
 
-import {roomFaqs, getAlias} from './room-faqs';
+import {roomFaqs, getAlias, visualizeFaq} from './room-faqs';
 import type {MessageHandler} from '../rooms';
 
 export interface RepeatedPhrase {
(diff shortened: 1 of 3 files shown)


server/chat-plugins/room-faqs.ts
```diff
@@ -3,12 +3,50 @@ import {FS, Utils} from '../../lib';
 export const ROOMFAQ_FILE = 'config/chat-plugins/faqs.json';
 const MAX_ROOMFAQ_LENGTH = 8192;
 
-export const roomFaqs: {[k: string]: {[k: string]: string}} = JSON.parse(FS(ROOMFAQ_FILE).readIfExistsSync() || "{}");
+export const roomFaqs: {[k: string]: {[k: string]: RoomFaq}} = (() => {
```
AnnikaCodes, line 6: Why does this use IIFE?

server/chat-plugins/room-faqs.ts
```diff
@@ -3,12 +3,50 @@ import {FS, Utils} from '../../lib';
...
+export function visualizeFaq(faq: RoomFaq) {
+	if (faq.html) {
+		return faq.source;
+	}
+	return Chat.formatText(faq.source, true);
+}
+
 /**
  * Aliases are implemented as a "regular" FAQ entry starting with a >. EX: {a: "text", b: ">a"}
  * This is done to allow easy checking whether a key is associated with
```
AnnikaCodes, line 52: This comment is no longer accurate.

server/chat-plugins/room-faqs.ts
```diff
@@ -17,14 +55,19 @@ function saveRoomFaqs() {
...
 	return false;
 }
 
 export const commands: Chat.ChatCommands = {
+	addhtmlfaq: 'addfaq',
 	addfaq(target, room, user, connection) {
 		room = this.requireRoom();
+		const useHTML = this.cmd.includes('html');
 		this.checkCan('ban', null, room);
+		if (useHTML && !user.can('declare', null, room, this.fullCmd)) {
```
AnnikaCodes, line 68:
```suggestion
		if (useHTML && !user.can('addhtml', null, room, this.fullCmd)) {
```

server/chat-plugins/repeats.ts
```diff
@@ -93,8 +93,10 @@ export const Repeats = new class {
 				this.clearRepeats(targetRoom);
 				return;
 			}
-			const repeatedPhrase = repeat.faq ? roomFaqs[targetRoom.roomid][repeat.id] : phrase;
-			const formattedText = repeat.isHTML ? phrase : Chat.formatText(repeatedPhrase, true);
+			const repeatedPhrase = repeat.faq ?
+				visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) :
+				Chat.formatText(phrase, true);
```
AnnikaCodes, line 98:
```suggestion
				visualizeFaq(roomFaqs[targetRoom.roomid][repeat.id]) : Chat.formatText(phrase, true);
```

server/chat-plugins/repeats.ts
```diff
@@ -229,7 +231,7 @@ export const commands: Chat.ChatCommands = {
 
 		Repeats.addRepeat(room, {
 			id: topic as ID,
-			phrase: roomFaqs[room.roomid][topic],
+			phrase: roomFaqs[room.roomid][topic]?.source,
```
AnnikaCodes, line 234: We should validate that the rfaq exists and display an error message to the user instead of optional chaining here.

server/chat-plugins/repeats.ts
```diff
@@ -136,7 +138,7 @@ export const pages: Chat.PageTable = {
 		html += `<table><tr><th>${this.tr`Identifier`}</th><th>${this.tr`Phrase`}</th><th>${this.tr`Raw text`}</th><th>${this.tr`Interval`}</th><th>${this.tr`Action`}</th>`;
 		for (const repeat of room.settings.repeats) {
 			const minutes = repeat.interval / (repeat.isByMessages ? 1 : 60 * 1000);
-			const repeatText = repeat.faq ? roomFaqs[room.roomid][repeat.id] : repeat.phrase;
+			const repeatText = repeat.faq ? roomFaqs[room.roomid][repeat.id]?.source : repeat.phrase;
```
AnnikaCodes, line 141: Is the rfaq not guaranteed to exist?

server/chat-plugins/room-faqs.ts
Zarel: In theory I fine with this. Although if you're going to define a function for `convertFaq` you might just also define a function for `loadFaqs`

server/chat-plugins/repeats.ts
AnnikaCodes: Then you don't need to optional chain.

mia-pi-git: I noticed (i removed the chain)

server/chat-plugins/room-faqs.ts
```diff
@@ -3,28 +3,66 @@ import {FS, Utils} from '../../lib';
...
+				data[k][name] = convertFaq(data[k][name]);
+				save = true;
+			}
+		}
+	}
+	if (save) saveRoomFaqs();
+	return data;
+})();
+
+interface RoomFaq {
```
KrisXV, line 21:
```suggestion
interface RoomFAQ {
```

server/chat-plugins/room-faqs.ts
```diff
@@ -3,28 +3,66 @@ import {FS, Utils} from '../../lib';
...
+				data[k][name] = convertFaq(data[k][name]);
+				save = true;
+			}
+		}
+	}
+	if (save) saveRoomFaqs();
+	return data;
+})();
+
+interface RoomFAQ {
```
AnnikaCodes, line 21: I'd prefer something like
```ts
type RoomFAQ = {alias: string} | {source: string, isHTML?: boolean};
``` 
but it's not a big deal to have the alias in the `source` I guess.

AnnikaCodes (approved):
Looks good to me!

PR #8344, opened 2021-05-30
Make sure berserk uses damage taken for dragon darts 

pacmanboss256:


DaWoblefet:
Doesn't this solution make Dragon Darts not trigger Berserk if the 2nd hit brings it under 50%?

pacmanboss256:
hmmmmm maybe yeah

data/abilities.ts
```diff
@@ -341,7 +341,8 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			if (!source || source === target || !target.hp || !move.totalDamage) return;
 			const lastAttackedBy = target.getLastAttackedBy();
 			if (!lastAttackedBy) return;
-			const damage = move.multihit ? move.totalDamage : lastAttackedBy.damage;
+			const darts = move.id === 'dragondarts' && source.getSmartTargets(target, move).length > 1;
```
Marty-D, line 344: Using `getSmartTargets` here doesn't make sense; it's not necessarily going to give you the same result as when it was used before damage happened.

pacmanboss256:
bump

Marty-D:
Shouldn't Berserk activate in general when hit by one dart from above 50%?

pacmanboss256:
yes, i just realized that a few minutes ago. Is there a way to check how many times a move hit a specific pokemon?

Marty-D:
Dragon Darts is just fundamentally broken, then. In the test what you're getting out of `getLastAttackedBy().damage` is 11, which is the damage actually dealt to Togedemaru, not Drampa. Meanwhile, the damage dealt to Togedemaru is stored as 0.

PR #8345, opened 2021-05-30
Support viewing stats on the activity of rooms

mia-pi-git:
Requested by Teclis.

### DIFF
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -702,7 +702,7 @@ export abstract class Searcher {
 	visualizeStats(stats: RoomStats) {
 		const titles: {[k: string]: string} = {
 			deadTime: 'Average time between lines',
-			deadPercent: 'Average % of the day spent more than 5m inactive',
+			deadPercent: 'Average % of the day spent more than 5 minutes inactive',
 			linesPerUser: 'Average lines per user',
 			averagePresent: 'Average users present',
 			totalLines: 'Average lines per day',
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -822,7 +822,7 @@ export abstract class Searcher {
 			}
 			}
 		}
-		results.deadTime = waitIncrements.length ? waitIncrements.reduce((c, p) => c + p) / waitIncrements.length : 0;
+		results.deadTime = waitIncrements.length ? this.calculateDead(waitIncrements) : 0;
 		results.deadPercent = !results.totalLines ? 100 : (waitIncrements.length / results.totalLines) * 100;
 		results.linesPerUser = (results.totalLines / Object.keys(results.users).length) || 0;
 		results.averagePresent = results.averagePresent / userstatCount;
@@ -832,6 +832,13 @@ export abstract class Searcher {
 			this.roomstatsCache.set(day, results);
 		}
 		return results;
+	}
+	private calculateDead(waitIncrements: number[]) {
+		let num = 0;
+		for (const k of waitIncrements) {
+			num += k;
+		}
+		return num / waitIncrements.length;
 	}
 }
 
--- a/server/chat-plugins/chatlog.ts
+++ b/server/chat-plugins/chatlog.ts
@@ -29,6 +29,33 @@ interface ChatlogSearch {
 	args?: string[];
 }
 
+interface RoomStats {
+	/**
+	 * Lines per user.
+	 */
+	lines: {[k: string]: number};
+	// guessed from |J| (number of joins)
+	users: {[k: string]: number};
+	days: number;
+	/**
+	 * Average wait time between each line ("dead")
+	 */
+	deadTime: number;
+	/**
+	 * Average percent of the day that it's inactive
+	 */
+	deadPercent: number;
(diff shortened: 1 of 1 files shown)


server/chat-plugins/chatlog.ts
```diff
@@ -626,6 +663,176 @@ export abstract class Searcher {
...
+			case 'userstats': {
+				const [rawTotal] = rest;
+				const total = parseInt(rawTotal.split(':')[1]);
+				results.averagePresent += total;
+				userstatCount++;
+				break;
+			}
+			}
+		}
+		results.deadTime = waitIncrements.length ? waitIncrements.reduce((c, p) => c + p) / waitIncrements.length : 0;
```
Zarel, line 825: Can this just be a for-loop? I'm not a fan of `reduce` for things like this.

mia-pi-git: Done.

server/chat-plugins/chatlog.ts
```diff
@@ -626,6 +663,183 @@ export abstract class Searcher {
...
+			buf += this.visualizeStats(day);
+			buf += `</div>`;
+		}
+		buf += '</details>';
+		return LogViewer.linkify(buf);
+	}
+	visualizeStats(stats: RoomStats) {
+		const titles: {[k: string]: string} = {
+			deadTime: 'Average time between lines',
+			deadPercent: 'Average % of the day spent more than 5m inactive',
```
AnnikaCodes, line 705:
```suggestion
			deadPercent: 'Average % of the day spent more than 5 minutes inactive',
```

server/chat-plugins/chatlog.ts
```diff
@@ -626,6 +663,183 @@ export abstract class Searcher {
...
+			linesPerUser: 'Average lines per user',
+			averagePresent: 'Average users present',
+			totalLines: 'Average lines per day',
+		};
+		let buf = `<div class="ladder pad"><table><tr><th>`;
+		buf += Object.keys(titles).map(k => titles[k]).join('</th><th>');
```
AnnikaCodes, line 711:
```suggestion
		buf += Object.values(titles).join('</th><th>');
```

server/chat-plugins/chatlog.ts
```diff
@@ -626,6 +663,183 @@ export abstract class Searcher {
...
+	}
+	async activityStats(room: RoomID, month: string) {
+		const days = (await FS(`logs/chat/${room}/${month}`).readdir()).map(f => f.slice(0, -4));
+		const stats: RoomStats[] = [];
+		for (const day of days) {
+			const curStats = await this.dayStats(room, day);
+			if (!curStats) continue;
+			stats.push(curStats);
+		}
+		// now, having collected the stats for each day, we need to merge them together
```
AnnikaCodes, line 738: We should probably add the stats to the total as we iterate over the days. This means we only have one loop.

AnnikaCodes (commented):
This seems fine at a glance; how well does it perform? I'm worried that this will become battlesearch 2.0.

mia-pi-git:
> This seems fine at a glance; how well does it perform? I'm worried that this will become battlesearch 2.0.

I ran it on my collection of Help room logs from 2014-2020, it performed relatively fine (few seconds lag). Teclis has been advised he should absolutely not run it on big rooms.

AnnikaCodes:
I'm skeptical of adding more laggy commands that we have to tell US to only use in certain circumstances. A few seconds lag is bad enough, and your laptop has an SSD, which is almost certainly faster than the spinning disks we use to store Main's logs. How do you feel about limiting it to running on only 6 or 12 months at a time? I assume there's a need for Policy admins specifically to use this command; otherwise, I could just do what I did for battlesearch and write my own (more efficient) standalone tool to run on sim3.

mia-pi-git:
> I'm skeptical of adding more laggy commands that we have to tell US to only use in certain circumstances. A few seconds lag is bad enough, and your laptop has an SSD, which is almost certainly faster than the spinning disks we use to store Main's logs. How do you feel about limiting it to running on only 6 or 12 months at a time? I assume there's a need for Policy admins specifically to use this command; otherwise, I could just do what I did for battlesearch and write my own (more efficient) standalone tool to run on sim3.

Big rooms translates to things like Lobby, OU, tours, and honestly those are the big ones. All else should be fine. Additionally, we only run this on one month at a time. Yes, policy admins specifically need this - that's why I made a command rather than a script.

mia-pi-git:
In case it's not clear: this runs as a file read stream in a child process. We take one string, parse it, then throw it away. By lag, I mean "time it takes to run." This will not affect the main process whatsoever.

AnnikaCodes:
I'm told this is in a child process and doesn't affect an end-user's experience. I'm fine with it, then, but if Marty reports that it interferes with usage stats (as battlesearch does), we may have to disable it (or restrict it to console admins with a warning to keep an eye on load).

mia-pi-git:
> interferes with usage stats (as battlesearch does)

It also shouldn't do this. We don't run ripgrep (instead, we use a file stream) so it shouldn't affect processing, afaik.

AnnikaCodes:
I think this is probably fine.

PR #8346, opened 2021-05-31
/eval: Clients opt into viewing its output

Slayer95:
Leverage ``/showdebug`` for that purpose.

Status quo: ≪development≫ chat is disrupted by evals and may even become unreadable.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -1249,7 +1249,7 @@ export const commands: Chat.ChatCommands = {
 			this.broadcastToRoom = true;
 		}
 		const generateHTML = (direction: string, contents: string) => (
-			`<table class="debug" border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">` +
+			`<table class="debug" style="overflow:auto; max-height:20em;" border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">` +
 				Utils.escapeHTML(direction).repeat(2) +
 				`&nbsp;</td><td>${Chat.getReadmoreCodeBlock(contents)}</td></tr><table>`
 		);
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -1249,7 +1249,7 @@ export const commands: Chat.ChatCommands = {
 			this.broadcastToRoom = true;
 		}
 		const generateHTML = (direction: string, contents: string) => (
-			`<table border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">` +
+			`<table class="debug" border="0" cellspacing="0" cellpadding="0"><tr><td valign="top">` +
 				Utils.escapeHTML(direction).repeat(2) +
 				`&nbsp;</td><td>${Chat.getReadmoreCodeBlock(contents)}</td></tr><table>`
 		);


mia-pi-git:
This doesn't seem super necessary, really. I think the solution is just improving handling for eval to ensure that big evals are hidden behind a readmore (read: number of characters to use a readmore rather than >3 lines).

AnnikaCodes:
I don't have any strong opinions on this except to second pyuk's point about making `/eval` enable showdebug if that's the solution used. Personally, I don't really mind long eval output in the Development room; it makes it easier to stay on the same page when debugging a problem (when compared to using `/eval` without broadcasting or broadcasting it in <<upperstaff>>). I can understand why people would dislike it, though.

Slayer95:
I support all of the above.
- Adding ``overflow:auto; max-height: 20em;`` would make the output small and scrollable.
- ``/eval`` turning on ``/showdebug`` does make sense.

Zarel:
What we should be doing is refactoring `visualize` so its horizontal size doesn't get too long. A lot of the problem most recently was with someone looking at a really large map/set whose elements were all on one line (so readmore couldn't hide it).

In the meantime, putting the readmore before everything else could also work. I'll do that instead.

In general I don't think this solves the problem. Anyone involved in development is going to want to see eval output; hiding it complicates the code and benefits practically no one.

Slayer95:
I'm confused as to why this *"Adding overflow:auto; max-height: 20em; would make the output small and scrollable"* doesn't work for you.

> Anyone involved in development is going to want to see eval output; and benefits practically no one.
I also don't think that's necessarily true. Only when evals apply hotfixes are they relevant to most people in the dev room; in other cases they are just noise.

> hiding it complicates the code
Is this a joke? The diff is an HTML class.

mia-pi-git:
> Now, as for why I present this approach:
> I'd argue most people -everybody but a few devs- aren't or shouldn't be concerned about what console users are evaling. So it's users who are already flabbergasted by reading evaled code (like "omg there are hackers in my game" Alt+F4). And yeah, I understand that devs do want to be kept on the same page regarding hotfixes, and that's why I intend to make them opt-in.

Honestly; do we really need to worry about regular users here? We only run evals in Dev and Upperstaff, so it's not like it's willy-nilly. People are only in Dev because they want to be informed anyway.

Zarel:
> I also don't think that's necessarily true. Only when evals apply hotfixes are they relevant to most people in the dev room; in other cases they are just noise.

Even if we assume that's true (which would be surprising to me), `/showdebug` has a lot of other problems, too; the big one being that it's not discoverable (debug messages are entirely invisible with no "show" button, making you very confused about what other people are talking about).

It's hard to turn on, nearly impossible to realize when you want to turn it on, and hard to turn off. This breaks [principle D6 (Feedback)](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#d6-feedback-is-important).

The real problem here is that long lines are long. And that's solvable through a number of different ways.

> I'm confused as to why this _"Adding overflow:auto; max-height: 20em; would make the output small and scrollable"_ doesn't work for you.

Compared to a readmore, it makes it harder to read for people who want to read it, and harder to skip for people who don't want to read it.

I don't understand what you have against just moving the readmore button upwards:

image

Slayer95:
> People are only in Dev because they want to be informed anyway.

> Also visit us in the Dev chatroom:
https://psim.us/dev
Source: PS client code.

I don't think a wall of live-evaled code is the best impression to give newcomers.

> /showdebug has a lot of other problems, too; the big one being that it's not discoverable

This is true.

> This breaks principle D6 (Feedback).

Client-side PR fixes seamlessly fixes that though.

> I don't understand what you have against just moving the readmore button upwards:

Nothing at all. I wanted a more serious comparison than "complicates the code".

Zarel:
> I'm confused as to why this _"Adding overflow:auto; max-height: 20em; would make the output small and scrollable"_ doesn't work for you.

Oh, by the way, this comes up so often that it's already in CONTRIBUTING, it's [design principle D4](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#d4-remove-unnecessary-scrolling-and-mouse-movement).

> Client-side PR fixes seamlessly fixes that though.

Not fully; I newly edited into CONTRIBUTING a more detailed explanation of D6:

: Options can exist, but options that hide the fact that you turned them on should be avoided.

Zarel:
I would go as far as to say that, as Dev RO, I think eval output is on-topic for the room. If you want to talk about development in an environment without eval output, please make your own room.

PR #8347, opened 2021-05-31
Update CUSTOM-RULES.md

DoctorTeuz:
After the second DLC "Galar Pokedex" and -Nonexistent are no more the same thing. As tested Shodown added the difference

### DIFF
--- a/config/CUSTOM-RULES.md
+++ b/config/CUSTOM-RULES.md
@@ -170,7 +170,7 @@ Custom rules can have more complicated behavior. They can also include other rul
 
 `Alola Pokedex` - allow only Pokémon in the Alola Pokédex (US/UM)
 
-(There is no `Galar Pokedex` rule, `-Nonexistent` covers it in Gen 8.)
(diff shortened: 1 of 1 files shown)


config/CUSTOM-RULES.md
```diff
@@ -170,7 +170,7 @@ Custom rules can have more complicated behavior. They can also include other rul
 
 `Alola Pokedex` - allow only Pokémon in the Alola Pokédex (US/UM)
 
-(There is no `Galar Pokedex` rule, `-Nonexistent` covers it in Gen 8.)
+`Galar Pokedex` - allow only Pokémon in the Galar Pokédex (SW/SH) [Ban pokemon you can catch in the adventures but are not listed in the pokedex like Ultracreatures, Landorus, etc...]
```
KrisXV, line 173:
```suggestion
`Galar Pokedex` - allow only Pokémon in the Galar Pokédex (SW/SH) [Ban Pokémon you can catch in the adventures but are not listed in the Pokédex like Ultra Beasts and Landorus]
```

PR #8348, opened 2021-05-31
Stomping Tantrum and CancelMove fix

pacmanboss256:
fixed PR#8324

### DIFF
--- a/sim/battle-queue.ts
+++ b/sim/battle-queue.ts
@@ -317,6 +317,7 @@ export class BattleQueue {
 				return true;
 			}
 		}
+		pokemon.moveThisTurnResult = null;
 		return false;
 	}
 
--- a/test/sim/moves/stompingtantrum.js
+++ b/test/sim/moves/stompingtantrum.js
@@ -90,7 +90,7 @@ describe('Stomping Tantrum', function () {
 		battle.makeChoices('move stompingtantrum', 'auto');
 	});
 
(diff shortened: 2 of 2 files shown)


Zarel:
You'd still need to refactor Pursuit and Quash not to use `cancelMove`.

Zarel:
(Or wait for me to, I suppose, which I'll get to eventually.)

Zarel:
(A simple temporary measure would be to just add a `forRelocation` argument to `cancelMove` which Pursuit and Quash would set to `true`.)

pacmanboss256:
it turns out that quash doesnt use cancelMove

Marty-D:
> Quash moving a move down in priority in Gen 7 and earlier

Zarel mentioned Gen 7 in the other PR because that's where Quash is still using `cancelMove`.

Anyway, do they really need refactoring? What are we trying to defend against by doing so? Stuff like Pursuit accidentally triggering double damage Stomping Tantrum the turn after it hit a Pokemon switching out? I'm not convinced that happens, but maybe there should be tests written to that effect.

pacmanboss256:
i dont think it matters actually because switching out with pursuit means you need to bring a new pokemon in, and then you would spend the turn that would double stomping tantrum swapping back into the original pokemon.

Marty-D:
I thought it would be clear I'm talking about the Pursuit user also using Stomping Tantrum, since that is the Pokemon that had its move canceled.

pacmanboss256:
ah

PR #8349, opened 2021-05-31
Gen 5: Refactor to use unreleasedHidden

xInstruct:
> [01:44] #Zarel: it should definitely set unreleasedHidden to true
[01:44] #Zarel: rather than literally nuking the HA

Welp, here I go on doing this. Will be a bit busy in a bit so just want this off the ground.

### DIFF
--- a/data/mods/gen5/pokedex.ts
+++ b/data/mods/gen5/pokedex.ts
@@ -518,10 +518,10 @@ export const Pokedex: {[k: string]: ModdedSpeciesData} = {
 	},
 	larvesta: {
 		inherit: true,
-		abilities: {0: "Flame Body"},
+		unreleasedHidden: true,
 	},
 	volcarona: {
 		inherit: true,
-		abilities: {0: "Flame Body"},
+		unreleasedHidden: true,
 	},
 };
--- a/data/mods/gen5/pokedex.ts
+++ b/data/mods/gen5/pokedex.ts
@@ -113,7 +113,7 @@ export const Pokedex: {[k: string]: ModdedSpeciesData} = {
 	},
 	zapdos: {
 		inherit: true,
-		abilities: {0: "Pressure",  H: "Lightning Rod"},
+		abilities: {0: "Pressure", H: "Lightning Rod"},
 		unreleasedHidden: true,
 	},
 	moltres: {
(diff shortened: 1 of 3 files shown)


KrisXV:
Theres a shitton more pokemon this applies to btw

also:

[01:14:21] #Zarel: uhhhhhhh
[01:15:00] #Zarel: !learn gen5, chandelure, ha, flamethrower
In Gen 5, HA Chandelure can't learn Flamethrower because:
Chandelure's move Flamethrower can only be learned in gens without Hidden Abilities.
[01:15:06] #Zarel: incorrect error message but it works
[01:15:22] #Zarel: confused why it doesn't just do unreleasedHidden though
[01:16:32] #Zarel: ok it looks like I just never pulled this hotpatch
[01:20:26] #Zarel: Bingus, I'm not entirely sure if the system needs to be changed

xInstruct:
yea..

Zarel:
This doesn't make much sense unless you want to also override with the never-released hidden abilities coded into Gen 5. Which we might want to do (for the Gen 5 DW format!) but I wouldn't do this without also doing that.

xInstruct:
> This doesn't make much sense unless you want to also override with the never-released hidden abilities coded into Gen 5. Which we might want to do (for the Gen 5 DW format!) but I wouldn't do this without also doing that.

That is the plan! Hope anyone reading this enjoys STAG chandelure

xInstruct:
Alright, DW abilities should be implemented properly, alongside a few mons who have unreleased abilities outside of gen 5. 

Anubis did PM me about incorrect shiny validation with some hidden abilities, but that will be another fish to fry later.

heatran sucks btw

Zarel:
Can you clarify the incorrect shiny validation issue?

xInstruct:
> Can you clarify the incorrect shiny validation issue? Is it a regression?


02:30:07] Centiskorch: btw did you also see the list of male/genderless-only gen 5 HAs
[02:30:24] Instruct: Yea
[02:30:46] Centiskorch: I said this to Marty in private but PS doesn't enforce the shiny lock on those mons either.
[02:33:29] Instruct: Alright, ill see if i can apply that as well
[02:33:31] Instruct: thanks
[02:34:03] Centiskorch: You can probably talk to him about it. He said something about an easier way would be to find what TI did before.

Zarel:
Sounds like it's not a regression. OK, thanks!

PR #8354, opened 2021-06-04
Confusion self-hits will not trigger berries

pacmanboss256:


DaWoblefet:
Perhaps it's just the variable name, but this implementation doesn't make it so that hitting yourself in confusion _never_ triggers Sitrus, etc. for the rest of the turn, right? It only shouldn't be able to run the Sitrus, etc. event on that confusion hit.

While you're at it, this same implementation may be relevant for something like Emergency Exit too - I'm not sure what PS's behavior is, but it shouldn't run on confusion hits either. 

The "correct" way of doing this (running very few battle events during the self-hit confusion routine) would probably involve a pretty substantial refactor, and this is a fix that I've been wanting corrected on Showdown for ages so as long as it does what it's supposed to, I'm game for this implementation.

Marty-D:
Isn't this just delaying the activation until the test passes? Berries will activate again as soon as the next `Update` event runs, which is the actual problem here.

pacmanboss256:
It originally would prevent the berry from triggering at any time on the current turn; should i move the self hit flag to the next damage taken?

Marty-D:
This old PR is much closer to the right approach: https://github.com/smogon/pokemon-showdown/pull/4045

I don't remember specifically what was wrong with it that it had to be reverted, something to do with Knock Off.

pacmanboss256:
It now switches self hit flag off when it takes any sort of damage

Marty-D:
A good place to start would be reading all the comments on the PR I linked, particularly the ones that describe in-game tests (which wouldn't hurt to implement either).

pacmanboss256:
not sure when the next possible time for a berry to activate after confusion is, but I assume it is immediately after any damage is taken

Marty-D (changes requested):
I'm not sure why you insist on implementing this as a hack on top of a hack when I pointed you in the direction of QuiteQuiet's PR from 4 years ago, which is the correct way of approaching the problem. Berries activate when they are obtained or otherwise start newly existing on a Pokemon, or in reaction to a damage event that isn't confusion. Using the Update event for Berries is completely wrong, so implementing anti-confusion measures in addition to that is just making it worse.

pacmanboss256:
ill fix this tomorrow

data/abilities.ts
```diff
@@ -184,8 +184,12 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.add('-ability', pokemon, 'Unnerve');
 			this.effectState.unnerved = true;
 		},
-		onEnd() {
+		onEnd(pokemon) {
 			this.effectState.unnerved = false;
+			if (this.effectState.unnerved) return;
```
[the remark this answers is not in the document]
pacmanboss256: the unnerve changes were from another pr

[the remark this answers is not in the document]
pacmanboss256: edited the comment, also this pr refactors healing berries on top of confusion interaction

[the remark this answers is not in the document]
pacmanboss256: if two unnerve pokemon are on the field, don't eat the berry because when a pokemon hits itself in confusion it eats the berry either when it takes damage or when all opposing unnerve pokemon leave the field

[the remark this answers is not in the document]
pacmanboss256: oh right this broke when I changed it from onUpdate. I can do that. Is each instance of unnerve on the field going to change the same effect state or is there one for each pokemon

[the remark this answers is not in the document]
pacmanboss256: the confusion self hit flag is set to false whenever unnerve leaves the field, I did not change any existing unnerve implementations

pacmanboss256:
For some reason the berries are triggered on residual and not on foe switch out, and I am not sure why

pacmanboss256:
The scope of this PR is expanding into a full berry refactor most likely

Marty-D:
Is there a reason you're still using the `hitSelf` workaround from your first version of this? What's wrong with checking `effect.id === 'confused'` or something? The residual stuff shouldn't need to exist either, just have Berries activate in reaction to damage.

pacmanboss256:
The residual stuff was for poison damage or other damage that the OnDamagingHit event doesn't catch. I am currently implementing a new event that runs whenever any damage is taken

pacmanboss256:
I'm keeping the flag so I don't have to mess with the effect id as an input when i want to run the event because calling it as a parameter is super messy

pacmanboss256:
Bumping this because I would like to have someone check for possible edge cases here about the new event.

pacmanboss256:
Bump

pacmanboss256 (commented):
outdated request, should have removed this

data/abilities.ts
```diff
@@ -184,8 +184,21 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.add('-ability', pokemon, 'Unnerve');
 			this.effectState.unnerved = true;
 		},
-		onEnd() {
+		onEnd(pokemon) {
 			this.effectState.unnerved = false;
+			const berries = ['aguavberry', 'figyberry', 'iapapaberry', 'magoberry', 'wikiberry',
+				'sitrusberry', 'oranberry', 'liechiberry', 'ganlonberry', 'salacberry', 'petayaberry', 'apicotberry', 'lansatberry', 'starfberry'];
+			let allyUnnerve = false;
+			for (const ally of pokemon.allies()) {
```
DaWoblefet, line 192: I think this would cause problems with FFAs; another opponent could have Unnerve and prevent eating Berries.

server/punishments.ts
```diff
@@ -1353,6 +1353,7 @@ export const Punishments = new class {
 
 		for (const curUser of affected) {
 			// ensure there aren't roombans so nothing gets mixed up
+			// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
DaWoblefet, line 1356: Not sure why this is here (maybe need to rebase)

sim/battle-actions.ts
```diff
@@ -1635,6 +1636,8 @@ export class BattleActions {
 		// int(int(int(2 * L / 5 + 2) * A * P / D) / 50);
 		const baseDamage = tr(tr(tr(tr(2 * level / 5 + 2) * basePower * attack) / defense) / 50);
 
+		if (baseDamage > 0) target.hitSelf = false;
```
DaWoblefet, line 1639: This isn't an indicator of the Pokemon took damage right? Base damage isn't the same as the actual damage roll. Moreover, if an attack didn't use base damage (say Seismic Toss) wouldn't this check get bypassed?

sim/pokemon.ts
```diff
@@ -1573,11 +1574,15 @@ export class Pokemon {
 
 	eatItem(force?: boolean, source?: Pokemon, sourceEffect?: Effect) {
 		if (!this.item) return false;
+
```
DaWoblefet, line 1577: Probably don't need all the extra newlines.

tools/simulate/index.js
```diff
@@ -37,12 +37,12 @@ if (process.argv[2]) {
 }
 
 require('ts-node').register({project: './tsconfig.json', files: true, transpileOnly: true, transpiler: 'ts-node/transpilers/swc-experimental'});
-const Dex = require('../../sim/dex').Dex;
+const Dex = require('../../.sim-dist/dex').Dex;
```
DaWoblefet, line 40: Same as punishments file change, probably accidentally got mixed in.

test/sim/items/sitrusberry.js
```diff
@@ -41,15 +41,23 @@ describe('Sitrus Berry', function () {
 		assert.equal(battle.p1.active[0].hp, 1);
 	});
 
-	it.skip(`should not heal 25% HP if a confusion self-hit would bring the user into Berry trigger range`, function () {
-		battle = common.createBattle([[
+	it(`should not heal 25% HP if a confusion self-hit would bring the user into Berry trigger range`, function () {
```
DaWoblefet, line 44: Can you add comments throughout this test when the Deoxys is hitting itself in confusion? It's not clear now.

test/sim/items/sitrusberry.js
```diff
@@ -41,15 +41,23 @@ describe('Sitrus Berry', function () {
...
-		assert.holdsItem(holder);
-		assert.false.fullHP(holder);
+		battle.makeChoices('move sleeptalk', 'move sleeptalk');
+		const hasItemA = Boolean(holder.item);
+		const hpA = holder.hp;
+		battle.makeChoices('move sleeptalk', 'move falseswipe');
+		const hpB = holder.hp;
+		const hasItemB = Boolean(holder.item);
+		assert.notEqual(hasItemA, hasItemB);
+		assert.notEqual(hpA, hpB);
```
DaWoblefet, line 61: Isn't this trivially true after you've attacked and used False Swipe?

test/sim/items/sitrusberry.js
```diff
@@ -41,15 +41,23 @@ describe('Sitrus Berry', function () {
...
 		], [
-			{species: 'Sableye', ability: 'prankster', moves: ['confuseray']},
+			{species: 'Sableye', ability: 'prankster', moves: ['confuseray', 'sleeptalk', 'falseswipe']},
 		]]);
 		const holder = battle.p1.active[0];
 		battle.makeChoices('move sleeptalk', 'move confuseray');
-		assert.holdsItem(holder);
-		assert.false.fullHP(holder);
+		battle.makeChoices('move sleeptalk', 'move sleeptalk');
+		const hasItemA = Boolean(holder.item);
```
DaWoblefet, line 55: You can just do ``assert.holdsItem(holder)`` and ``assert.false.holdsItem(holder)`` rather than this Boolean business.

DaWoblefet (changes requested):

Marty-D (changes requested):
I'm not going to review this PR with the botched merge, and will just close it in two weeks if it's still not in a reviewable state.

pacmanboss256:
Hey so I botched the merge and then I had finals and now I am on vacation, was going to fix when I get back

pacmanboss256:
bump

pacmanboss256:
bump

data/items.ts
```diff
@@ -385,6 +397,7 @@ export const Items: {[itemid: string]: ItemData} = {
 			basePower: 30,
 		},
 		onUpdate(pokemon) {
```
MathyFurret, line 399: You forgot to change Berry Juice to use TakeDamage instead

sim/battle.ts
```diff
@@ -1859,17 +1859,21 @@ export class Battle {
...
 				break;
 			case 'powder':
+				if (target.hitSelf) target.hitSelf = false;
 				this.add('-damage', target, target.getHealth, '[silent]');
 				break;
 			case 'confused':
 				this.add('-damage', target, target.getHealth, '[from] confusion');
 				break;
 			default:
+				if (target.hitSelf) target.hitSelf = false;
```
MathyFurret, line 1876: I'm still unclear on why you can't just pass `effect` to the `runEvent` you have below, then check `effect.id` in each handler as Marty said.

Even if you want to go this route, it'd be a lot DRYer to:
- Declare a variable and initialize it to `false`
- Set it to `true` only in the `confused` case
- Set `target.hitSelf` to this value at the end

Then you don't even need to change `getConfusionDamage`.

sim/battle.ts
```diff
@@ -1925,6 +1930,9 @@ export class Battle {
 			if (!source) source = this.event.source;
 			if (!effect) effect = this.effect;
 		}
+		// check if damage is caused by a confusion self hit
+		if (target && this.effect.id !== 'confused') target.hitSelf = false;
```
MathyFurret, line 1934: you're already setting hitSelf in spreadDamage.

MathyFurret (changes requested):

sim/battle.ts
pacmanboss256: some berries don't use onTakeDamage, such as enigma and custap. Does `this.effect.id` work the same as `effect.id` as the parameter or is that different

pacmanboss256:
bump

data/abilities.ts
```diff
@@ -184,7 +184,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.add('-ability', pokemon, 'Unnerve');
 			this.effectState.unnerved = true;
 		},
-		onEnd() {
+		onEnd(pokemon) {
```
AnnikaCodes, line 187: Why add a variable you don't use?

pacmanboss256: I have a few things in this pr that i still have to clean up, been super busy with exams/work/college stuff so i put it off. hoping to finish it this weekend

AnnikaCodes:
@Marty-D do you have more review here?

sim/battle-actions.ts
```diff
@@ -1054,6 +1054,7 @@ export class BattleActions {
 		const pokemonOriginalHP = pokemon.hp;
 		if (damagedDamage.length && !isSecondary && !isSelf) {
 			this.battle.runEvent('DamagingHit', damagedTargets, pokemon, move, damagedDamage);
+			this.battle.runEvent('TakeDamage', damagedTargets, pokemon, null, damagedDamage);
```
Marty-D, line 1057: Might as well put this line below the AfterHit part so you can remove the hardcode in Knock Off.

data/items.ts
```diff
@@ -230,7 +235,13 @@ export const Items: {[itemid: string]: ItemData} = {
...
 			type: "Ground",
 		},
-		onUpdate(pokemon) {
+		onStart(pokemon) {
+			if (pokemon.hp <= pokemon.maxhp / 4 || (pokemon.hp <= pokemon.maxhp / 2 && pokemon.hasAbility('gluttony'))) {
+				pokemon.eatItem();
+			}
+		},
+		onTakeDamage(damage, pokemon, source, effect) {
+			if (effect?.id === 'confusion') return;
```
Marty-D, line 244: This check seems to be completely unnecessary everywhere you put it because of the way you set up the new event.

data/items.ts
```diff
@@ -1028,7 +1040,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		onFractionalPriorityPriority: -2,
 		onFractionalPriority(priority, pokemon) {
 			if (
-				priority <= 0 &&
+				priority <= 0 && this.effect.id !== 'confusion' &&
```
Marty-D, line 1043: It's especially unnecessary here, where damage is not the triggering factor in activation.

test/sim/items/sitrusberry.js
```diff
@@ -41,15 +41,20 @@ describe('Sitrus Berry', function () {
...
 			{species: 'Deoxys-Attack', item: 'sitrusberry', moves: ['sleeptalk']},
 		], [
-			{species: 'Sableye', ability: 'prankster', moves: ['confuseray']},
+			{species: 'Sableye', ability: 'prankster', moves: ['confuseray', 'sleeptalk', 'falseswipe']},
 		]]);
 		const holder = battle.p1.active[0];
 		battle.makeChoices('move sleeptalk', 'move confuseray');
+		battle.makeChoices('move sleeptalk', 'move sleeptalk'); //Confusion hit is here
 		assert.holdsItem(holder);
-		assert.false.fullHP(holder);
```
Marty-D, line 53: I think this check is worth keeping.

Marty-D:
This is almost in a workable state! Just address the comments, clean up the random addition/deletion of newlines, rebase, and this could be good to go soon. 👍

sim/pokemon.ts
```diff
@@ -1590,11 +1589,11 @@ export class Pokemon {
 	}
 
 	eatItem(force?: boolean, source?: Pokemon, sourceEffect?: Effect) {
-		if (!this.item || this.itemState.knockedOff) return false;
+		if (!this.item) return false;
 		if ((!this.hp && this.item !== 'jabocaberry' && this.item !== 'rowapberry') || !this.isActive) return false;
-
 		if (!sourceEffect && this.battle.effect) sourceEffect = this.battle.effect;
 		if (!source && this.battle.event && this.battle.event.target) source = this.battle.event.target;
+
```
pacmanboss256, line 1596: I kept this in for readability, can remove it

data/items.ts
```diff
@@ -1,4 +1,4 @@
-export const Items: {[itemid: string]: ItemData} = {
+export const Items: { [itemid: string]: ItemData } = {
```
Marty-D, line 1: Why is this here?

data/items.ts
```diff
@@ -1532,6 +1549,7 @@ export const Items: {[itemid: string]: ItemData} = {
 			type: "Bug",
 		},
 		onHit(target, source, move) {
+			if (this.effect.id === 'confusion') return false;
```
Marty-D, line 1552: How would hitting yourself in confusion trigger Enigma Berry?

data/items.ts
```diff
@@ -2663,6 +2696,7 @@ export const Items: {[itemid: string]: ItemData} = {
 			type: "Fairy",
 		},
 		onAfterMoveSecondary(target, source, move) {
+			if (this.effect.id === 'confusion') return false;
```
Marty-D, line 2699: Ditto Enigma Berry.

data/items.ts
```diff
@@ -3457,7 +3506,7 @@ export const Items: {[itemid: string]: ItemData} = {
 			basePower: 100,
 			type: "Rock",
 		},
-		onResidual(pokemon) {
+		onResidual(pokemon, source, effect) {
```
Marty-D, line 3509: Unused variables.

sim/pokemon.ts
Marty-D: Nothing in this file should have changed, you even have a regression in here.

pacmanboss256: i think the regression was for removing the knock off hardcode

Marty-D: No, it's from https://github.com/smogon/pokemon-showdown/commit/0e692d4a9e5e11db9d650eb7a4ea84f99f43b071 and it's important.

pacmanboss256: oh my bad.

pacmanboss256:
I think this is actually complete now, all the formatting fixes are complete.

pacmanboss256:
bump

pacmanboss256:
Works with everything in gen 9, hope  I can finally get this merged. Been super busy with college stuff and I sort of forgot about it.

pacmanboss256:
bump

pacmanboss256:
bump

pacmanboss256:
Bumping, this has been done for a while now and needs final approval

pacmanboss256:
bump

data/abilities.ts
```diff
@@ -397,13 +397,18 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			return true;
 		},
 		onAfterMoveSecondary(target, source, move) {
+			const healingItems = [
```
AnnikaCodes, line 400: It seems like this is the third use of the same `healingItems` array in this file. Maybe it should be a file-level constant?

test/sim/items/sitrusberry.js
```diff
@@ -41,19 +41,25 @@ describe('Sitrus Berry', function () {
...
+			{species: 'Sableye', ability: 'prankster', moves: ['confuseray', 'sleeptalk', 'falseswipe']},
 		]]);
 		const holder = battle.p1.active[0];
 		battle.makeChoices('move sleeptalk', 'move confuseray');
+		battle.makeChoices('move sleeptalk', 'move sleeptalk'); //Confusion hit is here
 		assert.holdsItem(holder);
 		assert.false.fullHP(holder);
+		battle.makeChoices('move sleeptalk', 'move falseswipe');
+		common.saveReplay(battle, "test");
+		//Should eat the berry after the false swipe damage
```
AnnikaCodes, line 58:
```suggestion
		// Should eat the berry after the false swipe damage
```

AnnikaCodes:
Style nits aside, I'd like to help you get this merged, but I don't generally sign off on mechanics changes. You should probably fix the merge conflicts and then maybe @DaWoblefet can take a look?

pacmanboss256:
bump

pacmanboss256:
bump

PR #8355, opened 2021-06-04
Make Bonus Type modify species after Camomons

WeWuzNidokangz:
Currently, in a battle with both Camomons and Bonus Type battle effects, Camomons will randomly activate last per-switch in, overwriting the Pokemon's typing in full and disabling its Bonus Type. This change makes Bonus Type always activate last so that it is compatible with Camomons.

Similar to https://github.com/smogon/pokemon-showdown/pull/8051


Zarel:
Looks good; thanks!

PR #8357, opened 2021-06-06
Fix Future Sight Life Orb Interaction

aviettran:


Marty-D:
Good idea! Just from a cursory glance, I think Magician and Shell Bell will need checks to make sure they don't activate in this case. Gen 4 future moves will need some kind of override to make sure Life Orb doesn't activate at all.

aviettran:
@Marty-D good ideas. @DaWoblefet if you can give me a definitive answer, I can make the changes. 

Thanks

aviettran:
Another question: what is the behavior when a life orb carrier switches in?

DaWoblefet:
Shell Bell doesn't run at all, even if the user is active throughout the whole duration of Future Sight. Same with Magician.

Life Orb will damage the Future Sight user even if the user switches out and back in. It won't ever run _another_ Pokemon's Life Orb. If the user is not currently active, Life Orb will neither damage the user in the back nor increase the damage dealt.

From USUM research, we know that the following won't ever work with future attacks:
Abilities: Analytic, Cursed Body, Magician, Parental Bond
Items: Eject Button, Metronome, Jaboca / Rowap Berry, Red Card, Shell Bell
Moves: Destiny Bond, Grudge

I'll let Marty comment if you have any questions about what to do in Gen 4.

aviettran:
@Marty-D please review

Marty-D (approved):
Neat, pretty clever way around Magician and Shell Bell. Looks good, thanks! 👍

PR #8360, opened 2021-06-08
Gen 2: Nintendo Cup 2000 should use gen1stadium2

PlagueVonKarma:
NC2000 was played using Stadium 2.

Src: http://pokemon.s20.xrea.com/2nd/rule.html#nintendo 
"対戦環境は「ポケモンスタジアム金銀」。"
This roughly translates to "the battle environment is stadium gold and silver".

There is also an official source here stating that RGBY can be used, which is only possible via Stadium 2
https://www.nintendo.co.jp/n01/n64/software/nus_p_np3j/league/


Zarel:
Wait wait shouldn't it be `gen2stadium2`???

Marty-D:
It is; I ignored the commit title and wrote my own.

PR #8365, opened 2021-06-13
Granular Room Permissions Settings

tmagicturtle:
Hi, I thought it would be useful for /permissions to be able to turn on/off command options for each rank, rather than setting a minimum. In particular this would be useful for Bots, since (by good reason) they are ranked below Drivers, but would benefit from additional permissions like auto-moderation, Room/Staff intro modification (Officer Jenny does this for tracking watchlisted users for promotions in staff intros), and other useful commands.

My one concern is in the implementation of the user-groups.ts#hasPermission function. I am not sure what the jurisdiction thing is about (a room/global issue?) so I am not sure if simply returning the value of the room permission introduces a security risk. I would like some assistance here (and obviously if this is fine I would remove the non-executed code below it). Any assistance would be appreciated.

### DIFF
--- a/server/user-groups.ts
+++ b/server/user-groups.ts
@@ -150,17 +150,20 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 				if (roomPermissions[`/${cmd}`]) {
 					// this checks sub commands and command objects, but it checks to see if a sub-command
 					// overrides (should a perm for the command object exist) first
+					return roomPermissions[`/${cmd}`][symbol];
 					if (!auth.atLeast(user, roomPermissions[`/${cmd}`])) return false;
 					jurisdiction = 'u';
 					foundSpecificPermission = true;
 				} else if (roomPermissions[`/${namespace}`]) {
 					// if it's for one command object
+					return roomPermissions[`/${namespace}`][symbol];
 					if (!auth.atLeast(user, roomPermissions[`/${namespace}`])) return false;
 					jurisdiction = 'u';
 					foundSpecificPermission = true;
 				}
 			}
 			if (!foundSpecificPermission && roomPermissions[permission]) {
+				return (roomPermissions[permission][symbol]);
 				if (!auth.atLeast(user, roomPermissions[permission])) return false;
 				jurisdiction = 'u';
 			}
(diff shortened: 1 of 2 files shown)


server/user-groups.ts
```diff
@@ -150,17 +150,20 @@ export abstract class Auth extends Map<ID, GroupSymbol | ''> {
 				if (roomPermissions[`/${cmd}`]) {
 					// this checks sub commands and command objects, but it checks to see if a sub-command
 					// overrides (should a perm for the command object exist) first
+					return roomPermissions[`/${cmd}`][symbol];
```
mia-pi-git, line 153: Does returning here work? It shouldn't. 
You may want to check 
```ts
if ((roomPermissions[`/${cmd}`][symbol] && auth.get(user) !== roomPermissions[`/${cmd}`][symbol]) || !auth.atLeast(user, roomPermissions[`/${cmd}`])) return false```

Comment (mia-pi-git) [reply server/user-groups.ts:153]: (That's long, yes, but it should work more generally.)

Comment (mia-pi-git) [reply server/user-groups.ts:153]: Presently, this looks like it breaks other custom permissions.

**Review by mia-pi-git — CHANGES_REQUESTED** [review]

**Comment by Zarel** [comment]
I would really prefer not to have them be so granular without a better reason...

**Comment by Zarel** [comment]
If literally the only use for this huge amount of code complexity is bots, perhaps we should consider either moving bots up, or hardcoding them in some way.

**Comment by mia-pi-git** [comment]
> If literally the only use for this huge amount of code complexity is bots, perhaps we should consider either moving bots up, or hardcoding them in some way.

We've moved them up before, down before, up before, and now down to here. i don't think moving is necessarily the solution.

**Comment by Zarel** [comment]
Yeah, I know; I just went and looked at the comment because this has come up before:

```
// Bots are ranked below Driver/Mod so that Global Bots can be kept out
// of modjoin % rooms (namely, Staff).
// (They were previously above Driver/Mod so they can have game management
// permissions drivers don't, but these permissions can be manually given.)
```

I think the correct action here is to hardcode giving bots all permissions with a custom permission set.

AnnikaCodes:
@tmagicturtle, are you planning to do more work on this PR?

tmagicturtle:
No, I suppose not. Sorry for leaving it open, I'll close it now.

PR #8367, opened 2021-06-17
Support default format for randbats commands

Karthik99999:
I also made `/roomdefaultformat off` actually work.

### DIFF
--- a/server/chat-plugins/random-battles.ts
+++ b/server/chat-plugins/random-battles.ts
@@ -614,21 +614,15 @@ export const commands: Chat.ChatCommands = {
 
 		const args = target.split(',');
 		if (!args[0]) return this.parse(`/help randombattles`);
-		let dex = Dex;
-		let isLetsGo = false;
-		if (args[1] && toID(args[1]) in Dex.dexes) {
-			dex = Dex.dexes[toID(args[1])];
-			if (toID(args[1]) === 'letsgo') isLetsGo = true;
-		} else if (room?.battle) {
-			const format = Dex.formats.get(room.battle.format);
-			dex = Dex.mod(format.mod);
-			if (format.mod === 'letsgo') isLetsGo = true;
-		}
+
+		const {dex, format} = this.splitFormat(target, true);
+		let isLetsGo = !!(format?.mod === 'letsgo');
+
 		const species = dex.species.get(args[0]);
 		if (!species.exists) {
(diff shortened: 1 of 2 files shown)


server/chat-plugins/random-battles.ts
```diff
@@ -614,21 +614,18 @@ export const commands: Chat.ChatCommands = {
...
-		if (args[1] && toID(args[1]) in Dex.dexes) {
-			dex = Dex.dexes[toID(args[1])];
-			if (toID(args[1]) === 'letsgo') isLetsGo = true;
-		} else if (room?.battle) {
-			const format = Dex.formats.get(room.battle.format);
-			dex = Dex.mod(format.mod);
-			if (format.mod === 'letsgo') isLetsGo = true;
+
+		let {dex} = this.splitFormat(target, true);
+		if (!FS(`${dex.dataDir}/random-teams.js`).readIfExistsSync()) {
```
Zarel, line 619: No sync filesystem access in the main process, please!

Also are you reading the entire file just to check if it exists?

Zarel: I'm actually not sure what this is for in general...

Zarel: > No sync filesystem access in the main process, please!

I think some people reading this might point out there's a decent amount of it right now, so let me clarify that it's fine for initialization because it simplifies a lot of code. But I don't want it outside of initialization.

AnnikaCodes:
I'm fine with this feature in general, but I share Zarel's confusion and concern about the `if (!FS(`${dex.dataDir}/random-teams.js`).readIfExistsSync()) {` blocks.

Karthik99999:
yeah the FS stuff is probably not a good idea. any better way I can check if the mod has a random battles module? or do i not need to do that?

WeWuzNidokangz:
Maybe checking if Dex.formats has any Format for which the mod matches and team is not undefined?

AnnikaCodes:
I believe that modded dexes still provide the `randomBattleMoves` for the Pokémon in that gen? If I'm right, the checks should be unnecessary.

Zarel:
(Drafting until the filesystem access is figured out; I'm pretty sure the check is unnecessary.)

Karthik99999:
Ok so I thought the stadium mods would have it's own tiering (and therefore a seperate formats-data without the random battle moves), but I guess they don't. so yeah the checks are not needed

Zarel:
Thanks!

PR #8369, opened 2021-06-21
Use ts-node instead of Sucrase

AnnikaCodes:


Zarel:
Did we ever figure out the question of how we publish the library on NPM? I'm fine with falling back to Babel or something in that case.

package.json
```diff
@@ -2,28 +2,29 @@
   "name": "pokemon-showdown",
   "description": "The server for the Pokémon Showdown battle simulator",
   "version": "0.11.7",
-  "main": ".sim-dist/index.js",
+  "main": "sim/index.ts",
```
Zarel, line 5: Like I don't think this actually works in an NPM package.

package.json
```diff
@@ -2,28 +2,29 @@
...
-    "better-sqlite3": "^7.1.0",
     "brain.js": "^2.0.0-beta.2",
     "cloud-env": "^0.2.3",
     "node-static": "^0.7.11",
     "nodemailer": "^6.4.6",
     "permessage-deflate": "^0.1.7",
     "sql-template-strings": "^2.2.2",
-    "sqlite": "^3.0.6"
+    "sqlite": "^3.0.6",
+    "sucrase": "^3.15.0"
```
Zarel, line 21: ?

package.json
```diff
@@ -2,28 +2,29 @@
...
   "dependencies": {
+    "@swc/core": "^1.2.61",
     "@types/better-sqlite3": "^5.4.0",
     "probe-image-size": "^5.0.0",
     "sockjs": "^0.3.21",
-    "sucrase": "^3.15.0"
+    "ts-node": "^10.0.0"
   },
   "optionalDependencies": {
-    "better-sqlite3": "^7.1.0",
```
Zarel, line 13: ?

Are we just not using this at all anymore?

pokemon-showdown
```diff
@@ -1,41 +1,25 @@
...
+	require('ts-node').register({project: './tsconfig.json', files: true, transpileOnly: true, transpiler: 'ts-node/transpilers/swc-experimental'});
+}
+
+// enable require()ing TypeScript files
+try {
+	registerTSNode();
+} catch (e) {
+	console.log(`Installing dependencies...`);
+	var decl = ['--decl', 'decl'].includes(process.argv[2]);
+	require('child_process').execSync('npm install' + (decl ? '' : ' --production'), {stdio: 'inherit'});
```
Zarel, line 20:
```suggestion
	require('child_process').execSync('npm install --production'), {stdio: 'inherit'});
```

The `decl` option is specific to `./build` when building the NPM package; it's unnecessary here. See the rest of the conversation on building the NPM package.

package.json
AnnikaCodes: Oh yeah I should probably change that back.

package.json
AnnikaCodes: Oh sorry `npm` screwed up my package.json and I didn't fix it properly.

AnnikaCodes:
I left the Sucrase build infrastructure intact; I figure it's probably simplest to continue to use that for the NPM package. If you want to switch to using Babel or `tsc` I'm okay with that as well.

package.json
AnnikaCodes: Sucrase is an optional dependency now because it's only needed to build the NPM package.

Zarel:
Good point; `tsc` would probably also work here. I agree that that should be a separate PR.

pokemon-showdown
```diff
@@ -44,8 +27,7 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 	// host on port 9000:
 	// $ ./pokemon-showdown 9000
 
-	if (!built) build();
```
Zarel, line 47: Can you benchmark the performance impact on playing battles? Like if I do `cat inputlog.txt | ./pokemon-showdown simulate-battle` a few times before and now?

AnnikaCodes: Running one inputlog through `simulate-battle` takes about 3.3 times as long with ts-node (~2.3 seconds vs ~0.7 seconds). This makes sense, though; ts-node transpiles the code each time PS starts, resulting in a high upfront cost to starting up.

Zarel: That's... worrying. Maybe we should also keep the old build process around for the command-line tools, so it's just the server that uses `ts-node`?

AnnikaCodes: Sure, I'm good with that (although I think it might be worth switching to `tsc` sometime for both that and the NPM package).

tools/set-import/index.js
```diff
@@ -52,7 +53,7 @@ Chat.plural = function (num, plural = 's', singular = '') {
 	return (num !== 1 ? plural : singular);
 };
 // Sigh. Yay globals!
-global.toID = require('../../.sim-dist/dex').Dex.getId;
+global.toID = require('.././.sim/dex').Dex.getId;
```
Zarel, line 56: Is this even necessary anymore? The Sim API no longer requires a global `toID`.

pokemon-showdown
Zarel: Agreed.

tools/set-import/index.js
AnnikaCodes: I don't know; I've never worked on `set-import`. My find and replace script messed up here, though!

AnnikaCodes:
I'm not sure why the BattleStream tests are failing. It's asserting that the room-battle `PM.processes` has length 0, when it actually has length one. Isn't this correct, though? I thought we spawned a room-battle subprocess when starting up the server.

Zarel:
Are the unit tests running on `ts-node` now?

AnnikaCodes:
Yes.

tools/set-import/index.js
Zarel: Yeah, that's the first thing I noticed, but it occurs to me that you can just remove this line entirely now.

Zarel:
> I'm not sure why the BattleStream tests are failing. It's asserting that the room-battle `PM.processes` has length 0, when it actually has length one. Isn't this correct, though? I thought we spawned a room-battle subprocess when starting up the server.

The subprocess spawning is optional, and that unit test is a test to support setting the subprocess count to 0. (Having no subprocess also makes it easier to write unit tests since you can get at the Battle object directly.)

tools/set-import/index.js
AnnikaCodes: Alright. I'll also remove it from `tools/simulate`.

AnnikaCodes:
Hmmm, unit tests disable the ProcessManager, but for some reason it's spawning processes anyway...

I'll look into it more.

AnnikaCodes:
Ah, I found the issue. Apparently setting `require('../../lib/process-manager').disabled` doesn't actually make the value accessible within `process-manager.ts` (I'm guessing this is because of `ts-node`'s modifications to `require`, but I honestly don't know; JavaScript `require` behavior is not my area of expertise). I got unit tests to pass, though!

Zarel:
How does this affect hotpatching? Much easier than before?

AnnikaCodes:
No, it's basically the same, except that we don't have to rebuild first (since the new TS files will be transpiled when they're `require()`d by `/hotpatch`).

Zarel:
(That does not appear to have build errors.)

AnnikaCodes:
It fixed them locally, but not in CI. I'm working on it.

AnnikaCodes:
OK, this works now.

Zarel (approved):
I think this is really cool and should also make hotpatches a lot faster and make us deal with build issues much less.

Applying this will of course require a restart, which you can do whenever you like, or I can do it the next time I get the opportunity.

AnnikaCodes:
If it hasn't happened by next weekend, I'll restart PS then.

PR #8370, opened 2021-06-22
/botmsg: Permit ROs

PartMan7:
Bots face some rather inconvenient issues in subroomgroupchats.

### DIFF
--- a/server/chat-commands/admin.ts
+++ b/server/chat-commands/admin.ts
@@ -471,7 +471,7 @@ export const commands: Chat.ChatCommands = {
 		let {targetUser, rest: message} = this.requireUser(target);
 
 		const auth = this.room ? this.room.auth : Users.globalAuth;
-		if (auth.get(targetUser) !== '*') {
+		if (!['*', '#'].includes(auth.get(targetUser))) {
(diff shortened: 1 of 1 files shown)


Zarel:
Sure, we can revert if this causes a problem.

Zarel:
But why would they? Couldn't they inherit Bot status from the parent room?

TheMezStrikes:
Something similar has been an issue in the past with a few commands, and precedent seems to be adding ROs to the pool of valid groups instead of reading the parent room rank.
The big issue is that in a few of the On-Site Games rooms (and also in Scrappie somewhere), bots make subroomgroupchats for games and thusly end up as Room Owner and permission errors arise.

Zarel:
Oh, I see!

It really does sound like we need a lot of "Bot" exceptions to make permissions work right.

TheMezStrikes:
The most elegant solution would probably be to read permissions in both sub- and parent room and return success if either of them contains the permission you're looking for. (That doesn't solve this specific PR as it's not a permission issue but a hardcoded rank check, but it'd solve a lot of the other issues)

PR #8372, opened 2021-06-24
Gen 1: Fix Transform and Endless Battle Clause

TheSupervillain:
https://www.smogon.com/forums/threads/endless-battle-clause-enhancements-for-rby.3673455/
Note: Gen 1 stadium's unique transform restrictions are currently not implemented, but it's currently future-proofed when it happens.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1553,12 +1553,13 @@ export class Battle {
 		}
 
 		// Gen 1 Endless Battle Clause triggers
-		if (this.gen <= 1 && this.sides.every(side => side.pokemonLeft === 1)) {
+		if (this.gen <= 1) {
 			let onlyTransform = true;
 			let onlyGhostsWithStruggle = true;
 			let onlyFrozen = true;
 			for (const side of this.sides) {
 				for (const pokemon of side.pokemon) {
+					if (pokemon.fainted) continue;
 					if (pokemon.status !== 'frz' as ID) {
 						onlyFrozen = false;
 					}
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1554,36 +1554,47 @@ export class Battle {
 
 		// Gen 1 Endless Battle Clause triggers
 		if (this.gen <= 1) {
-			let onlyTransform = true;
-			let onlyGhostsWithStruggle = true;
-			let onlyFrozen = true;
+			let sidesUnableToMove = 0;
 			for (const side of this.sides) {
+				const reasonsUnable = ['struggle', 'frz', 'transform'];
+				if (this.dex.currentMod === 'gen1stadium') {
+					reasonsUnable.splice(reasonsUnable.indexOf('struggle'), 1);
+				}
 				for (const pokemon of side.pokemon) {
 					if (pokemon.fainted) continue;
-					if (pokemon.status !== 'frz' as ID) {
-						onlyFrozen = false;
+					if (pokemon.status !== 'frz' as ID && reasonsUnable.includes('frz')) {
+						reasonsUnable.splice(reasonsUnable.indexOf('frz'), 1);
 					}
 
+					/* I want to reevaluate this block because it doesn't cover every case of transform
 					if (!(pokemon.set.moves.map(toID).includes('transform' as ID) && pokemon.set.moves.length <= 1) ||
             (this.dex.currentMod === 'gen1stadium' && pokemon.species.id === 'ditto')) {
 						// Ditto naturally cannot cause an endless battle in Stadium, according to research
 						onlyTransform = false;
 					}
-
-					if (!pokemon.getTypes().includes('Ghost') || this.dex.currentMod === 'gen1stadium') {
-						onlyGhostsWithStruggle = false;
-					} else {
+          */
+					if (reasonsUnable.includes('struggle') && pokemon.status !== 'frz' as ID) {
 						let currentPP = 0;
 						for (const move of pokemon.moveSlots) {
 							currentPP += move.pp;
 						}
 						if (currentPP >= 1) {
-							onlyGhostsWithStruggle = false;
+							reasonsUnable.splice(reasonsUnable.indexOf('struggle'), 1);
 						}
 					}
 				}
+				if (reasonsUnable.includes('struggle')) {
+					for (const pokemon of side.foe.pokemon) {
+						if (pokemon.fainted) continue;
+						if (!pokemon.getTypes().includes('Ghost')) {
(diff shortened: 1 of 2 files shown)


TheSupervillain:
Transform is definitely an odd case, because it replenishes PP theoretically infinitely. I think staleness should be applied here, but I don't want to move forward yet if there are some issues with how staleness is already used, mostly because it's heavily associated with the berries.

sim/battle.ts
```diff
@@ -1552,6 +1552,43 @@ export class Battle {
 			this.add('bigerror', `You will auto-tie if the battle doesn't end in ${turnsLeftText} (on turn 1000).`);
 		}
 
+		// Gen 1 Endless Battle Clause triggers
+		if (this.gen <= 1) {
+			let onlyTransform = true;
+			let onlyGhostsWithStruggle = true;
+			let onlyFrozen = true;
```
Zarel, line 1559: I think this should be one variable, rather than three. Having just one `nothingCanMove` variable would catch cases such as "a frozen ghost vs a pokemon with only Struggle".

sim/battle.ts
```diff
@@ -1552,6 +1552,43 @@ export class Battle {
...
+			for (const side of this.sides) {
+				for (const pokemon of side.pokemon) {
+					if (pokemon.fainted) continue;
+					if (pokemon.status !== 'frz' as ID) {
+						onlyFrozen = false;
+					}
+
+					if (!(pokemon.set.moves.map(toID).includes('transform' as ID) && pokemon.set.moves.length <= 1) ||
+            (this.dex.currentMod === 'gen1stadium' && pokemon.species.id === 'ditto')) {
+						// Ditto naturally cannot cause an endless battle in Stadium, according to research
```
Zarel, line 1569: How does Ditto work in Stadium?

sim/battle.ts
TheSupervillain: yeah sounds good.

sim/battle.ts
TheSupervillain: I DMed you this on discord. see https://www.smogon.com/forums/threads/endless-battle-clause-enhancements-for-rby.3673455/#post-8739130

sim/battle.ts
TheSupervillain: I think I did it, albeit in a different direction than you anticipated.

PlagueVonKarma:
It may be worth implementing the Stadium Transform changes for Ditto just for completeness. If they can't be implemented here, I suppose I could do it another time? 

The following cases will make Stadium Transform fail, checked in this order;
- Ditto VS a Pokemon with Transform in its moveset (simply fails)
- Any Pokemon VS Ditto ("It couldn't transform into DITTO!" comes up)

Any other cases work as normal, so opposing Transform Mews is still an infinite battle waiting to happen.

TheSupervillain:
> It may be worth implementing the Stadium Transform changes for Ditto just for completeness. If they can't be implemented here, I suppose I could do it another time?

I can implement it here now.

sim/battle.ts
```diff
@@ -1552,6 +1552,54 @@ export class Battle {
...
+						if (!pokemon.getTypes().includes('Ghost')) {
+							reasonsUnable.splice(reasonsUnable.indexOf('struggle'), 1);
+							break;
+						}
+					}
+				}
+				if (reasonsUnable.length) sidesUnableToMove++;
+			}
+			if (sidesUnableToMove > 1) {
+				this.add('-message', `This battle cannot progress, therefore it will end prematurely.`);
```
Zarel, line 1598:
```suggestion
			const noProgressPossible = this.sides.every(side => {
				const foeAllGhosts = side.foe.pokemon.every(pokemon => pokemon.types.includes('Ghost'));
				const foeAllTransform = side.foe.pokemon.every(pokemon => (
					// true if transforming into this pokemon would lead to an endless battle
					// Transform will fail (depleting PP) if used against Ditto in Stadium 1
					(this.dex.currentMod !== 'gen1stadium' || pokemon.species.id !== 'ditto') &&
					// there are some subtleties such as a Mew with only Transform and auto-fail moves,
					// but it's unlikely to come up in a real game so there's no need to handle it
					pokemon.moves.every(moveid => moveid === 'transform')
				));
				return side.pokemon.every(pokemon => (
					// frozen pokemon can't thaw in gen 1 without outside help
					pokemon.status === 'frz' ||
					// a pokemon can't lose PP if it Transforms into a pokemon with only Transform
					(pokemon.moves.every(moveid => moveid === 'transform') && foeAllTransform) ||
					// Struggle can't damage yourself if every foe is a Ghost
					pokemon.moveSlots.every(slot => slot.pp === 0) && foeAllGhosts
				));
			});
			if (noProgressPossible) {
				this.add('-message', `This battle cannot progress. Endless Battle Clause activated!`);
```

Here, this should be shorter and more readable.

TheSupervillain: THANK YOU SO MUCH

sim/pokemon.ts
```diff
@@ -1126,7 +1126,9 @@ export class Pokemon {
 		const species = pokemon.species;
 		if (pokemon.fainted || pokemon.illusion || (pokemon.volatiles['substitute'] && this.battle.gen >= 5) ||
 			(pokemon.transformed && this.battle.gen >= 2) || (this.transformed && this.battle.gen >= 5) ||
-			species.name === 'Eternatus-Eternamax') {
+			species.name === 'Eternatus-Eternamax' || (this.battle.dex.currentMod === 'gen1stadium' && species.name === 'Ditto') ||
+      (this.battle.dex.currentMod === 'gen1stadium' && this.species.name === 'Ditto' &&
+      pokemon.moves.includes('transform'))) {
```
Zarel, line 1131: This is a bit of an unnecessarily huge if statement. Let's just put this in another if statement?

TheSupervillain: Yeah it is, I don't mind splitting it off from the other stuff

sim/pokemon.ts
```diff
@@ -1130,6 +1130,12 @@ export class Pokemon {
 			return false;
 		}
 
+		if ((this.battle.dex.currentMod === 'gen1stadium' && species.name === 'Ditto') ||
+			(this.battle.dex.currentMod === 'gen1stadium' && this.species.name === 'Ditto' &&
+			pokemon.moves.includes('transform'))) {
```
Zarel, line 1135:
```suggestion
		if (this.battle.dex.currentMod === 'gen1stadium' && (
			species.name === 'Ditto' ||
			(this.species.name === 'Ditto' && pokemon.moves.includes('transform'))
		)) {
```

Zarel:
ty!

PR #8373, opened 2021-06-24
Refactor overriding stats used for damaging moves

EvGym:
reworking the code to allow more future flexibility, and to move more of the responsibility for determining the offensive and defensive stats out of scripts and onto the moves/effects that cause the changes.

Wasn't able to move all of Wonder Room's code onto the move, but the situation is vastly improved.

Moves can now have their offensive and defensive stats be the source or targets: hp, current hp, attack, defense, special attack, special defense, or speed stat.

Due to the nature of offensive boosting items caring about the category and not the offensive stat I have added support for the attack stat and defense stat to be directly boosted to both continue to support this behavior and its possible expansion in the future.


Zarel:
It looks like you added a new `ModifyOffensiveStat` event? It's unclear to me what this improves; could you describe that more?

DaWoblefet:
In fairness, it looks like a lot of work on this PR was done before additional USUM research, but there isn't really such a thing as priority in stats, unless you want to count the list here as "priority" (https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8685465). However, that seems to be picked up from the codebase itself; another thing to refactor at some point! (beyond the scope of this PR I guess)

Marty-D:
I don't see the point of this when none of the existing implementations were removed. Don't we have enough code duplication all over the place already?

Re: Wonder Room, you should not move any of its implementation around if you don't know how it works. (https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-59#post-8707046)

EvGym:
> It looks like you added a new `ModifyOffensiveStat` event? It's unclear to me what this improves; could you describe that more?

Sure! The crux of the issue is that items like choice band and choice specs modify the damage of the move regardless of the offensive stat used which is not obvious to anyone who hasn't dug into the code before, mainly dug into scripts.ts. In fact I learned that fact writing this refactor. That is why I wanted to separate it from the atk boost to make it more clear what is happening in the code.

For example, eviolite doesn't boost the power of bodypress, but if you looked at body press and what evolite did in the code you would think it would. Evolite boosts your defensive stat, bodypress uses your defense as your offensive stat. That is the point of that function. I added a similar function for defense in case Nintendo decides to do a similar effect in the future with defence.

DaWoblefet:
> Sure! The crux of the issue is that items like choice band and choice specs modify the damage of the move regardless of the offensive stat used which is not obvious to anyone who hasn't dug into the code before, mainly dug into scripts.ts. In fact I learned that fact writing this refactor. That is why I wanted to separate it from the atk boost to make it more clear what is happening in the code.

Since this is damage formula stuff, just to make sure everything's clear here, Choice items aren't damage modifiers - they're attack modifiers. You probably meant that more colloquially (Choice Band/Specs increase damage), but I want to make sure that's clearly understood for anyone reading that.

> For example, eviolite doesn't boost the power of bodypress, but if you looked at body press and what evolite did in the code you would think it would. Evolite boosts your defensive stat, bodypress uses your defense as your offensive stat. That is the point of that function. I added a similar function for defense in case Nintendo decides to do a similar effect in the future with defence.

Body Press is a special snowflake to be sure, but so is Foul play, and I'm not really sure if this resolves any ambiguity with that. You can still ask "does Foul Play need the user or the target to hold Choice Band to get 1.5x Attack?" and I don't know if these changes make it any more obvious. I guess this helps with Body Press a little but I don't know if it's that obscure of a mechanic. Can you give some more examples?

EvGym:
> I don't see the point of this when none of the existing implementations were removed. Don't we have enough code duplication all over the place already?
> 
> Re: Wonder Room, you should not move any of its implementations around if you don't know how it works. (https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-59#post-8707046)

some of the existing implementations are removed. There is no longer a defensiveCategory, useSourceDefensiveAsOffensive, or even useTargetOffensive. This will not only make the effects of the move more explicit but improve readability. Effects will be able to own their own changes instead of making scripts handle them. Wonder Room still escapes this due to not swapping combat stages and I don't see a good way to fix it due to how boosts correlate with offensive damage.

If the extra functions do bother you I am willing to remove them. I added them because I wanted to make the code more explicit, since choice band doesn't just boost the atk stat, but the offensive stat of physical moves. I figured defense might get something similar in the future and prepared for that as well.

I still find it strange that choice band and huge power boosts body press and fur coat doesn't.

I did read into how Wonder Room works. However, it appears I misunderstood it and have re-looked into it. I now understand that it just swaps the base stats but leaves combat stages unchanged. I will make sure to correct this when I get back to my computer, thank you for the help! I had left the code in getStat() due to my previous understanding but I need to re-add it to calculateStat() and remove the code I added to the move itself. I will see if I can come up with a better fix for Wonder Room in the future as it is still very messy, but that will probably be a separate PR.

EvGym:
> In fairness, it looks like a lot of work on this PR was done before additional USUM research, but there isn't really such a thing as priority in stat modifier application, unless you want to count the list here as "priority" (https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8685465). It shouldn't actually need a property assigned to it, except for maybe Hustle. However, that seems to be picked up from the sim's codebase itself; another thing to refactor at some point! (beyond the scope of this PR I guess)

I just tried to copy the other stat changes so they match their priority. Since it is based on multiplication, and I have yet to see a stat modification that uses addition, it really should matter. Better safe than sorry though. It might matter in the future and it's better to have it now than have to add it later.

EvGym:
> Since this is damage formula stuff, just to make sure everything's clear here, Choice items aren't damage modifiers - they're attack modifiers. You probably meant that more colloquially (Choice Band/Specs increase damage), but I want to make sure that's clearly understood for anyone reading that.
>

yes. They modify the attack stat used in the damage formula, not the damage itself. It is good to check this stuff.

> Body Press is a special snowflake to be sure, but so is Foul Play, and I'm not really sure if this resolves any ambiguity with that. You can still ask "does Foul Play need the user or the target to hold Choice Band to get 1.5x Attack?" and I don't know if these changes make it any more obvious. I guess this helps with Body Press a little but I don't know if it's that obscure of a mechanic. Can you give some more examples?
>

I thought it was rather obvious that it uses the target's choice band and not your own. Maybe that is me though.
I think
"userOffensiveStat: 'target:atk'" (this is the code attached to foul play to make it work)
would be rather clear. Do you have any idea on how to improve it further? There is some more examples I could pull out of ssb for this specific situation.
Psyshock looks like this:
"userrDefensiveStat: 'target:def'"
So it should more directly line up with foul play to make it extra clear it truly uses the target for the calculation.

Marty-D:
>I still find it strange that choice band and huge power boosts body press and fur coat doesn't.

Foul Play only replaces the user's Attack stat and Attack stat stage in the damage formula with the target's Attack stat and Attack stat stage. Everything else in the damage formula stays the same - the user's Ability boost, the user's item boost, whatever.
It's exactly the same for Body Press except the user's Defense stat and stat stage are inserted instead of its Attack. Things that modify Attack continue to modify Attack, it just happens to be a different number.

Choice Band only modifies the Attack stat; "doesn't just boost the atk stat, but the offensive stat of physical moves" is completely meaningless because the offensive stat of physical moves _is_ the Attack stat. The problem is you fundamentally don't understand how the damage formula interacts with effects, and these changes make it more inconvenient for people who do understand these things to contribute code.

EvGym:
> Foul Play only replaces the user's Attack stat and Attack stat stage in the damage formula with the target's Attack stat and Attack stat stage. Everything else in the damage formula stays the same - the user's Ability boost, the user's item boost, whatever.
> It's exactly the same for Body Press except the user's Defense stat and stat stage are inserted instead of its Attack. Things that modify Attack continue to modify Attack, it just happens to be a different number.
> 
> Choice Band only modifies the Attack stat; "doesn't just boost the atk stat, but the offensive stat of physical moves" is completely meaningless because the offensive stat of physical moves _is_ the Attack stat. The problem is you fundamentally don't understand how the damage formula interacts with effects, and these changes make it more inconvenient for people who do understand these things to contribute code.
>

At a second glance, you are correct, it does use the user's attack modifications. I will make sure to correct that as well. Thank you for pointing that out! I am still rather new to developing on the showdown codebase so I am always welcoming of any feedback that helps me understand it more. Let me know if you have any ideas on how to improve the code to better show this off as it is still very bad at showing that foul play uses the user's abilities and items. Does the code support comments by properties? There are many that could use them, though that may be a separate PR.

Though you are incorrect about how the choice items work. the choice band does boost the stat used to calculate damage for body press that isn't attack. This may be a disagreement of semantics though. However, we can both agree body press uses the user's def stat value and def combat stages during the calculation for damage and that choice band still boosts it, which is not clear in the code or even the items/moves themselves.

These changes are simple enough they shouldn't impede future additions to the code. It might look complex now in file changes, but that is just because the old code is there to show the changes made. They should only make it easier to work with the code and implement future curve balls that get send our way.

EvGym:
I double-checked. I think there was a typescript update that is causing checks to fail, but none of these failures involve my code. I am not going to touch them as I want to keep this pull request focused.

EvGym:
@AnnikaCodes 
My code is now complete and finished.  It has all the latest showdown changes and I am happy with it. Now, all it needs is more reviews to make sure I didn't miss something.

DaWoblefet:
I'll let Marty have the final say. I do like the idea of being able to prepare for future Sp. Def Body Press variants or something similar. I'm actually kind of surprised we don't have Wonder Room tests in the repo; I'll add some to https://github.com/smogon/pokemon-showdown/pull/8494 while I'm at it so we can make sure no behavior regresses here.

EvGym:
> I'll let Marty have the final say. I do like the idea of being able to prepare for future Sp. Def Body Press variants or something similar. I'm actually kind of surprised we don't have Wonder Room tests in the repo; I'll add some to #8494 while I'm at it so we can make sure no behavior regresses here.

That would be greatly appreciated. I tested it locally because I was paranoid, but it would be good to double-check it and have something there for the future.

data/mods/gen2/scripts.ts
```diff
@@ -560,35 +559,59 @@ export const Scripts: ModdedBattleScriptsData = {
...
+			if (move.useBaseDefensiveStatAndBoosts) {
+				defType = move.useBaseDefensiveStatAndBoosts.substr(7) as AllStatIDs;
+			}
 			let unboosted = false;
 			let noburndrop = false;
 
 			if (isCrit) {
 				if (!suppressMessages) this.battle.add('-crit', target);
-				// Stat level modifications are ignored if they are neutral to or favour the defender.
+				// Stat level modifications are ignored if they are neutral to or favour) the defender.
```
Marty-D, line 588: ?

Marty-D:
I think the changes from `pokemon` to `source` should have been done separately from this, if at all. It makes it very annoying to find what's actually changing here. I like that the huge amount of code duplication you had is gone, at least.

Not a big fan of "future-proofing" past gens, but I understand why you did it like this. What's the argument for not refactoring all of that stuff out later? OMs can maybe use it?

Tentatively approving, subject to DaWoblefet's upcoming Wonder Room test passing and whichever Smogon Tournament is in progress finishing. Would like @Zarel to give this a look over meanwhile, but I'll merge after those conditions otherwise.

data/mods/gen1stadium/scripts.ts
```diff
@@ -529,15 +528,40 @@ export const Scripts: ModdedBattleScriptsData = {
...
-			let atkType: StatIDExceptHP = (move.category === 'Physical') ? 'atk' : 'spa';
-			const defType: StatIDExceptHP = (move.defensiveCategory === 'Physical') ? 'def' : 'spd';
-			if (move.useSourceDefensiveAsOffensive) atkType = defType;
-			let attack = attacker.getStat(atkType);
-			let defense = defender.getStat(defType);
+			let level = source.level;
+			const attacker =
+				(move.useBaseOffensiveStatAndBoosts && move.useBaseOffensiveStatAndBoosts.includes('target')) ? target : source;
+			const defender =
+				(move.useBaseDefensiveStatAndBoosts && move.useBaseDefensiveStatAndBoosts.includes('source')) ? source : target;
```
KrisXV, line 535:
```suggestion
			const attacker = move.useBaseOffensiveStatAndBoosts?.includes('target') ? target : source;
			const defender = move.useBaseDefensiveStatAndBoosts?.includes('source') ? source : target;
```

data/mods/gen1/scripts.ts
```diff
@@ -692,14 +691,40 @@ export const Scripts: ModdedBattleScriptsData = {
...
-			if (move.useTargetOffensive) attacker = target;
-			const atkType: StatIDExceptHP = (move.category === 'Physical') ? 'atk' : 'spa';
-			const defType: StatIDExceptHP = (move.defensiveCategory === 'Physical') ? 'def' : 'spd';
-			let attack = attacker.getStat(move.useSourceDefensiveAsOffensive ? defType : atkType);
-			let defense = defender.getStat(defType);
+			let level = source.level;
+			const attacker =
+				(move.useBaseOffensiveStatAndBoosts && move.useBaseOffensiveStatAndBoosts.includes('target')) ? target : source;
+			const defender =
+				(move.useBaseDefensiveStatAndBoosts && move.useBaseDefensiveStatAndBoosts.includes('source')) ? source : target;
```
KrisXV, line 698:
```suggestion
			const attacker = move.useBaseOffensiveStatAndBoosts?.includes('target') ? target : source;
			const defender = move.useBaseDefensiveStatAndBoosts?.includes('source') ? source : target;
```

data/mods/gen1stadium/scripts.ts
```diff
@@ -529,15 +528,40 @@ export const Scripts: ModdedBattleScriptsData = {
...
+				attack = attacker.getStat(atkType);
+			}
+
+			if (defType === 'hp') {
+				defense = defender.maxhp;
+			} else if (defType === 'currenthp') {
+				defense = defender.hp;
+			} else {
+				defense = defender.getStat(defType);
+			}
```
KrisXV, line 563: ditto here

data/mods/gen2/scripts.ts
```diff
@@ -560,35 +559,59 @@ export const Scripts: ModdedBattleScriptsData = {
...
-			let attacker = pokemon;
-			const defender = target;
-			if (move.useTargetOffensive) attacker = target;
-			let atkType: StatIDExceptHP = (move.category === 'Physical') ? 'atk' : 'spa';
-			const defType: StatIDExceptHP = (move.defensiveCategory === 'Physical') ? 'def' : 'spd';
-			if (move.useSourceDefensiveAsOffensive) atkType = defType;
+			const attacker =
+				(move.useBaseOffensiveStatAndBoosts && move.useBaseOffensiveStatAndBoosts.includes('target')) ? target : source;
+			const defender =
+				(move.useBaseDefensiveStatAndBoosts && move.useBaseDefensiveStatAndBoosts.includes('source')) ? source : target;
```
KrisXV, line 573: again, apply optional chaining

data/mods/gen2/scripts.ts
```diff
@@ -560,35 +559,59 @@ export const Scripts: ModdedBattleScriptsData = {
...
+				attack = attacker.getStat(atkType, unboosted, noburndrop);
+			}
+
+			if (defType === 'hp') {
+				defense = defender.maxhp;
+			} else if (defType === 'currenthp') {
+				defense = defender.hp;
+			} else {
+				defense = defender.getStat(defType, unboosted);
+			}
```
KrisXV, line 614: ditto

data/mods/gen2stadium2/scripts.ts
```diff
@@ -329,27 +328,51 @@ export const Scripts: ModdedBattleScriptsData = {
...
-			let attacker = pokemon;
-			const defender = target;
-			if (move.useTargetOffensive) attacker = target;
-			let atkType: StatIDExceptHP = (move.category === 'Physical') ? 'atk' : 'spa';
-			const defType: StatIDExceptHP = (move.defensiveCategory === 'Physical') ? 'def' : 'spd';
-			if (move.useSourceDefensiveAsOffensive) atkType = defType;
+			const attacker =
+				(move.useBaseOffensiveStatAndBoosts && move.useBaseOffensiveStatAndBoosts.includes('target')) ? target : pokemon;
+			const defender =
+				(move.useBaseDefensiveStatAndBoosts && move.useBaseDefensiveStatAndBoosts.includes('source')) ? pokemon : target;
```
KrisXV, line 334: ditto

data/mods/gen2stadium2/scripts.ts
```diff
@@ -329,27 +328,51 @@ export const Scripts: ModdedBattleScriptsData = {
...
+				attack = attacker.getStat(atkType, unboosted, noburndrop);
+			}
+
+			if (defType === 'hp') {
+				defense = defender.maxhp;
+			} else if (defType === 'currenthp') {
+				defense = defender.hp;
+			} else {
+				defense = defender.getStat(defType, unboosted);
+			}
```
KrisXV, line 375: ditto

data/mods/ssb/moves.ts
```diff
@@ -853,7 +853,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		onPrepareHit(target, source) {
 			this.add('-anim', source, 'Spirit Break', target);
 		},
-		useSourceDefensiveAsOffensive: true,
+		useBaseOffensiveStatAndBoosts: 'source:spd',
```
KrisXV, line 856: this would probably be better as an array or something `['source', 'spd']` for example

data/mods/ssb/moves.ts
```diff
@@ -907,6 +907,24 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+					} else {
+						move.useBaseOffensiveStatAndBoosts = 'target:atk';
+					}
+				} else {
+					if (attacker.getStat('atk') > attacker.getStat('spa')) {
+						move.useBaseOffensiveStatAndBoosts = 'source:spa';
+					} else {
+						move.useBaseOffensiveStatAndBoosts = 'source:atk';
+					}
+				}
```
KrisXV, line 926: optional chaining, and using arrays this could be further simplified

data/moves.ts
```diff
@@ -19436,13 +19436,27 @@ export const Moves: {[moveid: string]: MoveData} = {
...
+						if (source.boosts['def'] || source.boosts['spd']) {
+							this.hint(move.name + " uses Sp. Def boosts when Wonder Room is active.");
+						}
+					} else if (move.useBaseOffensiveStatAndBoosts.includes("spd")) {
+						move.useBaseOffensiveStatAndBoosts.replace("spd", "def");
+						if (source.boosts['def'] || source.boosts['spd']) {
+							this.hint(move.name + " uses Def boosts when Wonder Room is active.");
+						}
+					}
+				}
```
KrisXV, line 19452: id probably just slap an `if (!move.useBaseOffensiveStatAndBoosts) return;` at the top for less indentation

sim/dex-moves.ts
```diff
@@ -147,6 +147,10 @@ export interface MoveData extends EffectData, MoveEventMethods, HitEffect {
 	contestType?: string;
 	noPPBoosts?: boolean;
 
+	// stats used for the move
+	useBaseOffensiveStatAndBoosts?: `${'target:' | 'source:'}${AllStatIDs}`;
+	useBaseDefensiveStatAndBoosts?: `${'target:' | 'source:'}${AllStatIDs}`;
```
KrisXV, line 152:
```suggestion
	useBaseOffensiveStatAndBoosts?: ['target' | 'source', AllStatIDs];
	useBaseDefensiveStatAndBoosts?: ['target' | 'source', AllStatIDs];
```

KrisXV (changes requested):

data/mods/gen2/scripts.ts
EvGym: No idea why that is there or how I missed that

> I think the changes from `pokemon` to `source` should have been done separately from this, if at all. It makes it very annoying to find what's actually changing here. I like that the huge amount of code duplication you had is gone, at least.
> 
> Not a big fan of "future-proofing" past gens, but I understand why you did it like this. What's the argument for not refactoring all of that stuff out later? OMs can maybe use it?
> 
> Tentatively approving, subject to DaWoblefet's upcoming Wonder Room test passing and whichever Smogon Tournament is in progress finishing. Would like @Zarel to give this a look over meanwhile, but I'll merge after those conditions otherwise. 👍

Consistency. I wanted them all to use the same code when possible. Sorry about including the pokemon -> source change, I thought it made the code more readable. I will try and segment my changes more in the future.

EvGym:
No idea why that is there or how I missed that

> I think the changes from `pokemon` to `source` should have been done separately from this, if at all. It makes it very annoying to find what's actually changing here. I like that the huge amount of code duplication you had is gone, at least.
> 
> Not a big fan of "future-proofing" past gens, but I understand why you did it like this. What's the argument for not refactoring all of that stuff out later? OMs can maybe use it?
> 
> Tentatively approving, subject to DaWoblefet's upcoming Wonder Room test passing and whichever Smogon Tournament is in progress finishing. Would like @Zarel to give this a look over meanwhile, but I'll merge after those conditions otherwise. 👍

Consistency. I wanted them all to use the same code when possible. Sorry about including the pokemon -> source change, I thought it made the code more readable. I will try and segment my changes more in the future.

data/mods/gen2/scripts.ts
EvGym: Oops. I will fix that in the next commit. Nice catch

EvGym: fixed

data/mods/gen1stadium/scripts.ts
EvGym: Done. Didn't know that existed, very cool!

data/mods/gen1stadium/scripts.ts
EvGym: I assume you are referencing optional chaining? It will not work here due to what is needed to be done for the alternative cases. If I am just missing something please let me know. I feel like this is the better and easier to read solution to the problem.

data/mods/gen2/scripts.ts
EvGym: Done

data/mods/gen2/scripts.ts
EvGym: (same comment as above)
I assume you are referencing optional chaining? It will not work here due to what is needed to be done for the alternative cases. If I am just missing something please let me know. I feel like this is the better and easier to read solution to the problem.

data/mods/gen2stadium2/scripts.ts
EvGym: Done

data/mods/gen2stadium2/scripts.ts
EvGym: (same comment as above)
I assume you are referencing optional chaining? It will not work here due to what is needed to be done for the alternative cases. If I am just missing something please let me know. I feel like this is the better and easier to read solution to the problem.

data/mods/ssb/moves.ts
EvGym: agreed. Don't know why I, and none of the people on the discord server who talked about this, never considered just making it an array. We treat it as one anyway. Good catch.

data/mods/ssb/moves.ts
EvGym: Done

data/moves.ts
EvGym: Not needed. When using ?. it returns undefined which is treated as false. This makes it so I can remove that if statement altogether.

data/mods/gen1stadium/scripts.ts
KrisXV: no, i mean the inline if statements

data/mods/gen2/scripts.ts
KrisXV: inline if

data/mods/gen2stadium2/scripts.ts
KrisXV: inline if

data/moves.ts
```diff
@@ -19436,13 +19436,25 @@ export const Moves: {[moveid: string]: MoveData} = {
...
+					move.useBaseOffensiveStatAndBoosts[1] = "spd";
+					if (source.boosts['def'] || source.boosts['spd']) {
+						this.hint(move.name + " uses Sp. Def boosts when Wonder Room is active.");
+					}
+				} else if (move.useBaseOffensiveStatAndBoosts?.includes("spd")) {
+					move.useBaseOffensiveStatAndBoosts[1] = "def";
+					if (source.boosts['def'] || source.boosts['spd']) {
+						this.hint(move.name + " uses Def boosts when Wonder Room is active.");
+					}
+				}
```
KrisXV, line 19450:
```suggestion
				if (!move.useBaseOffensiveStatAndBoosts) return;
				const statAndBoosts = move.useBaseOffensiveStatAndBoosts[1];
				if (!['def', 'spd'].includes(statAndBoosts)) return;
				move.useBaseOffensiveStatAndBoosts[1] = statAndBoosts === 'def' ? 'spd' : 'def';
				this.hint(`${move.name} uses ${statAndBoosts === 'def' ? '' : 'Sp. '}Def boosts when Wonder Room is active.`);
```

data/moves.ts
KrisXV: it makes the code much more readable, even if it's not needed, which i think is preferable

data/moves.ts
KrisXV: ^ this is what i meant

data/mods/ssb/moves.ts
KrisXV: Those slip-ups happen sometimes, and it functions the same regardless, so it probably wasn't at the top of everyone's minds when the main focus was making sure it worked.

data/moves.ts
EvGym: oh, fancy. Good idea.

data/mods/gen2stadium2/scripts.ts
EvGym: Fixed, thank you.

data/mods/gen2/scripts.ts
EvGym: Fixed, thank you.

data/mods/gen1stadium/scripts.ts
EvGym: Fixed, thank you.

data/moves.ts
EvGym: after seeing your suggested rework, I agree. I didn't get what you were suggesting at first.

Zarel (approved):
When this PR was first opened, it seemed pretty poorly written, but its current state is actually pretty nice.

While I don't feel great about it, I can't think of any obvious improvements, and I think as-is it's better than what's in `smogon:master` right now.

Zarel:
Is there any precedent for max and current HP being used as attacking stats? They seem rather nonstandard.

sim/global-types.ts
```diff
@@ -23,6 +23,7 @@ type ID = '' | string & {__isID: true};
 type PokemonSlot = '' | string & {__isSlot: true};
 interface AnyObject {[k: string]: any}
 
+type AllStatIDs = 'currenthp' | StatID;
```
Zarel, line 26: I consider `AllStatIDs` a pretty bad name for this, tbh. Probably just use `StatID | 'currenthp'`, you only really need it in like two places. Honestly probably use `StatIDExceptHP | 'maxhp' | 'currenthp'` for extra clarity.

EvGym:
I have learned a lot about coding in typescript since starting this PR and about how pokemon works. I had only really coded in c++ and, minorly, in python. (Even with python experience, mixed type arrays still feel like heresy.)

I also never realized that foul play and body press benefited from the choice band. I used to think foul play would use the opponent's choice band. Turns out I am not the only one as I was able to stump a ton of people with this trivia. Didn't even think to question it.

All of that + some good suggestions about how to restructure it and fix it led to this pull request.

sim/global-types.ts
EvGym: ah, Good point. I was using it in a lot more places before I swapped things to using the array. Good catch

EvGym: How about StatIDAndCurrentHP?

EvGym:
> Is there any precedent for max and current HP being used as attacking stats? They seem rather nonstandard.

There isn't but I added them because it was easy. I figured it couldn't hurt.

sim/global-types.ts
Zarel: In this case, since it's not generic, I would go with `DamageCalcStatID`.

But I think given the other context, I would rather remove `maxhp` and `currenthp` from the type entirely.

Zarel:
> There isn't but I added them because it was easy. I figured it couldn't hurt.

I see. I understand why you would think that, but I disagree that it couldn't hurt. Having spent a long time working on this simulator, I think current/max HP are very different numbers than other stats. The game uses a different formula to calculate max HP, it doesn't boost in stages, it can't be affected by Natures, and no calculation that involves "a stat" will use HP.

> There is plenty of times hp has affected offensive power though so I figure it is within the realm of possibility.

There _are_ moves that involve HP, such as Reversal, but the calculation never uses HP as a stat in a stat calculation.

Now, I'd agree that the harm is relatively small (the main harm is the type complexity, and a small readability/maintenance burden), so I'd usually allow features like this if they seem like they could potentially be useful to OMs (custom game formats thought up by fans).

So I think the important points here:

- HP isn't calculated the same way as other stats, and so an OM using it as if it were a stat is going to be frustrating because the damage calculation will not "make sense".
- Current HP is even weirder of a number to use. It has the same flaws as max HP, but also fluctuates way more.
- In practice, any OM that wants to incorporate HP into its damage calculation should simply have a custom BP formula like Reversal or custom damage formula like Super Fang. That will be much easier for players to understand than a custom stat calculation that uses HP as a stat.
- The added type complexity will be frustrating for future use: the fact that the type of this variable _isn't_ `StatIDExceptHP` means that it can't be used in functions that expect `StatIDExceptHP`, even though it will never be anything else.

To conclude: I appreciate the thought and effort you put into this, but I think it'd be better for our codebase if you simply used `StatIDExceptHP`.

data/mods/ssb/moves.ts
```diff
@@ -907,6 +907,24 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+					} else {
+						move.useBaseOffensiveStatAndBoosts[1] = 'atk';
+					}
+				} else {
+					if (attacker.getStat('atk') > attacker.getStat('spa')) {
+						move.useBaseOffensiveStatAndBoosts[1] = 'spa';
+					} else {
+						move.useBaseOffensiveStatAndBoosts[1] = 'atk';
+					}
+				}
```
KrisXV, line 926:
```suggestion
				const offensiveAndBoosts = move.useBaseOffensiveStatAndBoosts;
				if (!offensiveAndBoosts) return;
				if (!['atk', 'spa'].includes(offensiveAndBoosts[1])) return;
				if (offensiveAndBoosts[0] === 'target') {
					if (!defender) return;
					move.useBaseOffensiveStatAndBoosts[1] = defender.getStat('atk') > defender.getStat('spa') ? 'spa' : 'atk';
				} else {
					move.useBaseOffensiveStatAndBoosts[1] = attacker.getStat('atk') > attacker.getStat('spa') ? 'spa' : 'atk';
				}
```

EvGym: Nice catch

sim/dex-moves.ts
```diff
@@ -147,6 +147,10 @@ export interface MoveData extends EffectData, MoveEventMethods, HitEffect {
 	contestType?: string;
 	noPPBoosts?: boolean;
 
+	// stats used for the move
+	useBaseOffensiveStatAndBoosts?: ['target' | 'source', StatIDExceptHP];
+	useBaseDefensiveStatAndBoosts?: ['target' | 'source', StatIDExceptHP];
+
```
Zarel, line 153: These should be moved down to "Hit effect modifiers", where you deleted `defensiveCategory`.

Zarel: I wonder if `overrideOffensiveStat` and `overrideDefensiveStat` would be more readable.

data/mods/gen2stadium2/scripts.ts
```diff
@@ -329,25 +328,30 @@ export const Scripts: ModdedBattleScriptsData = {
...
 			let unboosted = false;
 			let noburndrop = false;
 
 			if (isCrit) {
 				if (!suppressMessages) this.battle.add('-crit', target);
 				// Stat level modifications are ignored if they are neutral to or favour the defender.
 				// Reflect and Light Screen defensive boosts are only ignored if stat level modifications were also ignored as a result of that.
-				if (attacker.boosts[atkType] <= defender.boosts[defType]) {
+				if ((attacker.boosts[atkType] <= defender.boosts[defType])) {
```
Zarel, line 349: ?

Zarel, line 349: ?

sim/dex-moves.ts
Zarel: Yeah, I looked over everything. I think this would be much more readable if these were separate properties, instead of one combined property.

Zarel: One thing I might also be open to is a group:

```ts
overrideStats: {
  offensive: StatIDExceptHP,
  offensivePokemon: 'source' | 'target',
  defensive: StatIDExceptHP,
  defensivePokemon: 'source' | 'target',
};
```

I personally think grouping is unnecessary here, but if you like grouping, I wouldn't be _too_ opposed. Splitting into two tables is also a compromise, if you really want as minimal changes as possible.

Zarel:
(For the record, I don't mind writing my suggested changes myself, but I'm holding off in case you have objections or alternative ideas.)

data/mods/gen2stadium2/scripts.ts
Zarel: Guessing this is just a typo.

```suggestion
				if (attacker.boosts[atkType] <= defender.boosts[defType]) {
```

Zarel: Guessing this is just a typo.

```suggestion
				if (attacker.boosts[atkType] <= defender.boosts[defType]) {
```

EvGym:
I appreciate it. The suggested changes don't seem to get feedback from a linter so it is probably better that way.

sim/dex-moves.ts
EvGym: I want to try and communicate in some form that it doesn't use the opponent's items to boost stats since that isn't super obvious. overrideStats will have that problem as well. I do think splitting them is a good idea though. It will bulk things up a little, but increased readability is more important. I will have to make the pokemon value default to the source of the attack for offense and the target for defense.

I can see the appeal of grouping, but unlike conditions and secondary effects they won't rely on each other so I think it is ultimately unnecessary. The only benefit is it can help make the names shorter individually.

how about this as the names:
```
OverrideOffensiveStatPokemon: 'source' | 'target',
OverrideDefensiveStatPokemon: 'source' | 'target',
OverrideOffensiveStatAndBoosts: StatIDExceptHP,
OverrideDefensiveStatAndBoost: StatIDExceptHP
```
They aren't perfect as they don't exactly communicate that foul play would benefit from a choice band used by the attacker and not the defender but at that point, the changes to the variable names would be too much for it to be worth it. At the very least it should make it a bit more clear that choice band boosts body press.

data/mods/gen2stadium2/scripts.ts
EvGym: Left overs from removing the hp stats from the override stat support. Good catch

sim/dex-moves.ts
Zarel: > I want to try and communicate in some form that it doesn't use the opponent's items to boost stats since that isn't super obvious.

I think that's too much information to fit in a variable name. I'd recommend a doc comment.

EvGym: I am going to push a change. Let me know what you think

Zarel: Looks great!

Zarel: Looks great!

data/mods/gen1/scripts.ts
```diff
@@ -693,13 +692,21 @@ export const Scripts: ModdedBattleScriptsData = {
...
+			const defender = move.overrideDefensivePokemon === 'source' ? source : target;
+
+			let atkType: StatIDExceptHP = move.category === 'Physical' ? 'atk' : 'spa';
+			if (move.overrideOffensiveStat) {
+				atkType = move.overrideOffensiveStat;
+			}
+			let defType: StatIDExceptHP = move.category === 'Physical' ? 'def' : 'spd';
+			if (move.overrideDefensiveStat) {
+				defType = move.overrideDefensiveStat;
+			}
```
Zarel, line 705:
```suggestion
			const isPhysical = move.category === 'Physical';
			const atkType = move.overrideOffensiveStat || (isPhysical ? 'atk' : 'spa');
			const defType = move.overrideDefensiveStat || (isPhysical ? 'def' : 'spd');
```

Zarel: And with all _those_ improvements, this code can get a lot simpler, too!

Remember to also apply this to every other generation!

Zarel:
Earlier, I said in my review:

> While I don't feel great about it, I can't think of any obvious improvements

I take this back. After more reviews, I _did_ think of obvious improvements, and now I _do_ feel great about this!

data/mods/gen1/scripts.ts
EvGym: Clever! I will make sure to roll these changes over to the other scripts as well. I will have time to make the commit in ~4 hours.

sim/dex-moves.ts
```diff
@@ -203,7 +203,16 @@ export interface MoveData extends EffectData, MoveEventMethods, HitEffect {
...
+	/**
+	 * controls which pokemon it draws the attack stats from. This does not use the pokemon's ability or item
+	 */
+	overrideOffensivePokemon?: 'target' | 'source';
+	overrideOffensiveStat?: StatIDExceptHP;
+	/**
+	 * controls which pokemon it draws the defense stats from. This does not use the pokemon's ability or item
+	 */
+	overrideDefensivePokemon?: 'target' | 'source';
+	overrideDefensiveStat?: StatIDExceptHP;
```
Zarel, line 215:
```suggestion
	/** Pokemon for the attack stat. Ability and Item damage modifiers still come from the real attacker. */
	overrideOffensivePokemon?: 'target' | 'source';
	overrideOffensiveStat?: StatIDExceptHP;
	/** Pokemon for the defense stat. Ability and Item damage modifiers still come from the real defender. */
	overrideDefensivePokemon?: 'target' | 'source';
	overrideDefensiveStat?: StatIDExceptHP;
```

We still need an explanation for Choice Band. Does it modify the Attack stat?

DaWoblefet:
https://github.com/smogon/pokemon-showdown/pull/8494 is in, so you can rebase to verify that Wonder Room still works correctly with this new implementation.

EvGym:
Unless I broke something and didn't realize it, it should work. It was one of the most paranoid points in my testing.

Zarel:
image

Love to see this on battle-actions. You know you're doing a good job when you simplify so much. I think this is ready for merge. @Marty-D?

Zarel:
LOL actually we need to hold on a bit, Metronome tests are failing now.

EvGym:
Let em review. That usually means I had a typo

data/mods/gen1/scripts.ts
```diff
@@ -693,13 +692,16 @@ export const Scripts: ModdedBattleScriptsData = {
...
-			if (move.useTargetOffensive) attacker = target;
-			const atkType: StatIDExceptHP = (move.category === 'Physical') ? 'atk' : 'spa';
-			const defType: StatIDExceptHP = (move.defensiveCategory === 'Physical') ? 'def' : 'spd';
-			let attack = attacker.getStat(move.useSourceDefensiveAsOffensive ? defType : atkType);
+			const attacker = move.overrideOffensivePokemon === 'target' ? target : source;
+			const defender = move.overrideDefensivePokemon === 'source' ? source : target;
+
+			const isPhysical = move.category === 'Physical';
+			const atkType: StatIDExceptHP = move.overrideOffensiveStat || (isPhysical ? 'atk' : 'spa');
+			const defType: StatIDExceptHP = move.overrideOffensiveStat || (isPhysical ? 'def' : 'spd');
```
Marty-D, line 700: Missed another one here!

Marty-D (approved):
Looks good to go after that last typo disappears. Nice work guys! 👍

data/mods/gen1/scripts.ts
Zarel:
```suggestion
			const defType: StatIDExceptHP = move.overrideDefensiveStat || (isPhysical ? 'def' : 'spd');
```

data/mods/ssb/moves.ts
```diff
@@ -907,6 +907,16 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+			onModifyMove(move, attacker, defender) {
+				if (!move.overrideOffensiveStat) return;
+				if (!['atk', 'spa'].includes(move.overrideOffensiveStat)) return;
+				if (move.overrideOffensivePokemon === 'target') {
+					if (!defender) return;
+					move.overrideOffensiveStat = defender.getStat('atk') > defender.getStat('spa') ? 'spa' : 'atk';
+				} else {
+					move.overrideOffensiveStat = attacker.getStat('atk') > attacker.getStat('spa') ? 'spa' : 'atk';
+				}
+			},
```
Zarel, line 919:
```suggestion
			onModifyMove(move, source, target) {
				if (!move.overrideOffensiveStat) return;
				if (!['atk', 'spa'].includes(move.overrideOffensiveStat)) return;
				const attacker = move.overrideOffensivePokemon === 'target' ? target : source;
				if (!attacker) return;
				move.overrideOffensiveStat = attacker.getStat('atk') > attacker.getStat('spa') ? 'spa' : 'atk';
			},
```

Is `if (!move.overrideOffensiveStat) return;` necessary? It seems wrong to me.

data/mods/gen1/scripts.ts
EvGym: gah, I thought I fixed that one.

data/mods/ssb/moves.ts
EvGym: That is a good rewrite. Makes sense.
It is necessary but it shouldn't be returning like that. That is an issue. I will fix that. It should just set it to atk or spa based on the move's catagory.

sim/dex-moves.ts
EvGym: added this

data/mods/ssb/moves.ts
```diff
@@ -907,6 +907,15 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 				}
 				return 5;
 			},
+			onModifyMove(move, source, target) {
+				if (!move.overrideOffensiveStat) {
+					move.overrideOffensiveStat = move.category === "Physical" ? 'atk' : 'spa';
+				}
+				if (!['atk', 'spa'].includes(move.overrideOffensiveStat)) return;
```
Zarel, line 914: I don't think this is necessary.

```suggestion
				if (move.overrideOffensiveStat && !['atk', 'spa'].includes(move.overrideOffensiveStat)) return;
```

This is enough, right?

data/moves.ts
```diff
@@ -19437,13 +19437,20 @@ export const Moves: {[moveid: string]: MoveData} = {
 				}
 				return 5;
 			},
+			onModifyMove(move, source, target) {
+				if (!move.overrideOffensiveStat) return;
+				const statAndBoosts = move.overrideOffensiveStat;
+				if (!['def', 'spd'].includes(statAndBoosts)) return;
+				move.overrideOffensiveStat = statAndBoosts === 'def' ? 'spd' : 'def';
+				this.hint(`${move.name} uses ${statAndBoosts === 'def' ? '' : 'Sp. '}Def boosts when Wonder Room is active.`);
```
Zarel, line 19445: This seems to have the same issue. `move.overrideOffensiveStat` is undefined for most moves, right? But Wonder Room should affect most moves, right?

data/mods/ssb/moves.ts
EvGym: no because (I am assuming you meant to put a ! before the move.overrideOffensiveStat) it would make it so bane terrain doesn't work on moves that don't have their offensive stat overwritten.

That being said I can make the second if statement an else if as it cannot be true if the first is true, but it would make the code harder to read and wouldn't be run enough for it to be an issue. I can still make that chance if you think it is needed

data/moves.ts
EvGym: It will. This particular bit of code is there only to handle cases where the offensive stat is overwritten to a defensive stat. Attack stats are already swapped by the base wonderroom effect as coded in the stat fetching functions (I would LOVE to move that code onto wonderroom but it can't think of a good way to do it), but defense doesn't have this implementation so we need to add code to handle that edge case.

Zarel: Oh! I understand now.

sim/dex-moves.ts
```diff
@@ -448,9 +468,14 @@ export class DataMove extends BasicEffect implements Readonly<BasicEffect & Move
...
-		this.useTargetOffensive = !!data.useTargetOffensive;
-		this.useSourceDefensiveAsOffensive = !!data.useSourceDefensiveAsOffensive;
+		this.overrideOffensiveStat =
+			typeof data.overrideOffensiveStat === 'string' ? (data.overrideOffensiveStat as StatIDExceptHP) : undefined;
+		this.overrideOffensivePokemon =
+			data.overrideOffensivePokemon ? (data.overrideOffensivePokemon === "target" ? "target" : "source") : undefined;
+		this.overrideDefensiveStat =
+			typeof data.overrideDefensiveStat === 'string' ? (data.overrideDefensiveStat as StatIDExceptHP) : undefined;
+		this.overrideDefensivePokemon =
+			data.overrideDefensivePokemon ? (data.overrideDefensivePokemon === "source" ? "source" : "target") : undefined;
```
Zarel, line 478:
```suggestion
		this.overrideOffensiveStat = data.overrideOffensiveStat || undefined;
		this.overrideOffensivePokemon = data.overrideOffensivePokemon || undefined;
		this.overrideDefensiveStat = data.overrideDefensiveStat || undefined;
		this.overrideDefensivePokemon = data.overrideDefensivePokemon || undefined;
```

You're not supposed to have validation here.

data/moves.ts
```diff
@@ -19437,13 +19437,20 @@ export const Moves: {[moveid: string]: MoveData} = {
 				}
 				return 5;
 			},
+			onModifyMove(move, source, target) {
+				if (!move.overrideOffensiveStat) return;
```
Zarel, line 19441:
```suggestion
				// This is just for moves that use defensive stats; see below for most of the implementation
				if (!move.overrideOffensiveStat) return;
```

sim/dex-moves.ts
EvGym: Didn't think that would work. I had used other possibly undefined strings as references when setting these up. Good to know.

data/moves.ts
EvGym: That is a bit confusing as all moves use defensive stats. I added a comment that specified that it was for when it used defensive stats as offensive stats

Zarel:
Thanks!

PR #8377, opened 2021-06-28
Gen 5: Blue Basculin should be able to have Reckless

Karthik99999:
Second bug on the post: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-59#post-8900393

While not confirmed on the post, I tested this myself by catching a Blue Basculin from the wild in Pokemon White, and it had Reckless. Don't know whether Rock Head or Reckless should be considered the "Special" ability.

Drafting for now, as I need someone to confirm if transferring the Basculin into Pokemon Bank actually changes Reckless to Rock Head.

Screenshot 2021-06-28 130932

### DIFF
--- a/data/mods/gen5/pokedex.ts
+++ b/data/mods/gen5/pokedex.ts
@@ -459,6 +459,10 @@ export const Pokedex: {[k: string]: ModdedSpeciesData} = {
 		inherit: true,
 		types: ["Grass"],
 	},
+	basculinbluestriped: {
+		inherit: true,
+		abilities: {0: "Rock Head", 1: "Adaptability", H: "Mold Breaker", S: "Reckless"},
+	},
 	krookodile: {
 		inherit: true,
 		baseStats: {hp: 95, atk: 117, def: 70, spa: 65, spd: 70, spe: 92},


Marty-D:
Yeah, this was known before B2W2 were released, and at the time it appeared that the in-game trade was the bug, but the real problem was the wild encounters. We didn't have a workable system to deal with 4 Abilities back then, so when it got corrected in B2W2 wild encounters it just got tossed in the "I guess this doesn't matter anymore" pile. But it does kind of matter since Rock Head is still locked out of egg moves. A Blue-Striped's Reckless does become Rock Head upon transfer to Gen 6.

Karthik99999:
So I guess the other thing that would need to be done here is handle Rock Head Basculin not having access to egg moves, but I'd rather let @Zarel handle validator stuff lol.

Zarel:
I can handle the validator if I have a better idea of what's going on. If I'm understanding correctly, Reckless is breedable and Rock Head isn't in Gen 5, but Rock Head is the only ability in newer gens?

Yeah, that should be doable without an explicit event.

Marty-D:
> 
> 
> I can handle the validator if I have a better idea of what's going on. If I'm understanding correctly, Reckless is breedable and Rock Head isn't in Gen 5, but Rock Head is the only ability in newer gens?
> 
> Yeah, that should be doable without an explicit event.

Yeah that's the gist of it. I guess technically that also means the minimum Rock Head level is also the minimum wild level in B2W2 (which is 5), since the in-game trade in White is always level 25.

Looks good @Karthik99999, thanks!

PR #8378, opened 2021-06-29
Sparkling Aria: Only cure status if the user is alive

TheSupervillain:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-42#post-8750235

Should be accurate to the gif in the provided post (from what I can see, anyways)


TheSupervillain:
I'll just close this for now while I tinker by myself, I just decided to do this while tired at 3am.

pacmanboss256:
I was trying this exact same thing but it doesnt seem to work because of when secondaries are triggered

PR #8381, opened 2021-07-04
Sparkling Aria shouldn't cure burn if the user faints

pacmanboss256:


Zarel:
The problem with titles like this is that "no longer" is kind of ambiguous. Was it originally correct, and then it no longer doing so is a bug? Or was it originally bugged, and no longer doing so is the fix?

Generally the ways I find unambiguous are "Fix ... to ..." or "... should ...".

The PR description should also say things like "In Sword/Shield, this happens" and "Previously, in PS, this happens, which is a bug". I constantly have to ask clarifications because things aren't phrased that way.

data/moves.ts
```diff
@@ -16242,10 +16242,16 @@ export const Moves: {[moveid: string]: MoveData} = {
 		secondary: {
 			dustproof: true,
 			chance: 100,
-			onHit(target) {
-				if (target.status === 'brn') target.cureStatus();
+			onHit(target, source, move) {
+				if (target.status === 'brn') {
+					target.addVolatile('cureburn', source, move);
```
Zarel, line 16247: Why are you using a volatile to track this? Does Sparkling Aria not cure burns if the target gets burned during the attack?

pacmanboss256: the target does not get cured if the user of the move faints on the same turn. I needed the volatile so it still is affected by sheer force but not shield dust, but the actual curing is after "onDamage" as in innards out

Zarel: Oh, it's for Sheer Force.

data/moves.ts
```diff
@@ -16242,10 +16242,16 @@ export const Moves: {[moveid: string]: MoveData} = {
...
+				}
 			},
 		},
+		onAfterMove(source, target, move) {
+			if (!source.fainted && target.getVolatile('cureburn')) target.cureStatus();
+			target.removeVolatile('cureburn');
```
urkerab, line 16253:
```suggestion
			if (target.removeVolatile('sparklingaria') && target.status === 'brn' && !source.fainted) target.cureStatus();
```
(see above)

test/sim/moves/sparklingaria.js
```diff
@@ -12,18 +12,17 @@ describe('Sparkling Aria', function () {
 
 	it(`should cure the target's burn`, function () {
 		battle = common.createBattle([[
-			{species: 'Wynaut', ability: 'compoundeyes', moves: ['will-o-wisp', 'sparklingaria']},
+			{species: 'Wynaut', ability: 'compoundeyes', moves: ['will-o-wisp', 'sparklingaria', 'sleeptalk']},
```
DaWoblefet, line 15: Don't forget to remove the extra Sleep Talk.

DaWoblefet (approved):
I think this will end up fixing itself once I finish documenting post-move success checks, but seems fine for now.

Marty-D (changes requested):
Sparkling Aria is a spread move and `AfterMove` only happens once.

pacmanboss256:
bump

pacmanboss256:
Bump

Marty-D:
No need to keep bumping this. I want to check something first and then I'll merge if it's fine.

pacmanboss256:
Gotcha.

pacmanboss256:
its been over 2 weeks, what was it you wanted to check?

data/moves.ts
```diff
@@ -16242,9 +16242,14 @@ export const Moves: {[moveid: string]: MoveData} = {
...
 			chance: 100,
-			onHit(target) {
-				if (target.status === 'brn') target.cureStatus();
-			},
+			volatileStatus: 'sparklingaria',
+		},
+		onAfterMove(source, target, move) {
+			for (const pokemon of this.getAllActive()) {
+				if (pokemon !== source && pokemon.removeVolatile('sparklingaria') && pokemon.status === 'brn' && !source.fainted) {
+					target.cureStatus();
```
Marty-D, line 16250: `target` is only one specific Pokemon in an `AfterMove` context; you want `pokemon` here too.

Marty-D (approved):
Okay this should be fine now, thanks!

PR #8382, opened 2021-07-05
Fix build

pacmanboss256:
Endless Battle Clause ends after turn 999, not on turn 1000. Updated the test to reflect that and fixed a typo

### DIFF
--- a/test/sim/misc/endlessbattleclause.js
+++ b/test/sim/misc/endlessbattleclause.js
@@ -192,7 +192,7 @@ describe('Endless Battle Clause (slow)', () => {
 	});
 });
 
-// Endless Battle Caluse doesn't take effect for 100 turns, so we artificially skip turns
+// Endless Battle Clause doesn't take effect for 100 turns, so we artificially skip turns
 // to get the turn counter to be in the range which could possibly trigger the clause
 function skipTurns(battle, turns) {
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks! At some point this should fix to actually end at turn 1000, though.

pacmanboss256:
I can do that in a second

test/sim/misc/endlessbattleclause.js
```diff
@@ -183,7 +183,7 @@ describe('Endless Battle Clause (slow)', () => {
 			{species: "Blissey", moves: ['splash']},
 			{species: "Vaporeon", moves: ['splash']},
 		]});
-		for (let i = 0; i < 999; i++) {
+		for (let i = 0; i < 998; i++) {
```
AnnikaCodes, line 186: Is the battle ending after 999 turns the desired behavior? If not, I don't think modifying tests to test incorrect behavior is a good idea (this test will in fact fail if EBC invokes after 1000 turns); if it's an issue to have the build failing because EBC is broken, I would prefer to skip the test rather than changing it to describe incorrect behavior.

pacmanboss256: I made a new PR to fix it

PR #8383, opened 2021-07-05
Fix endless battle clause

pacmanboss256:


pacmanboss256:
@AnnikaCodes this passed on my end

AnnikaCodes:
You'll need to revert the other PR's changes to the test. It passes locally because you haven't rebased against upstream yet.

Zarel:
Hm, this will say "in 2 turns, in 1 turn, in 0 turns, ended" over the course of the last four turns, which is probably undesirable.

Ideally the countdown should go something like:

- in 2 turns
- this turn
- ended

Or perhaps it _would_ be clearer if we ended at the beginning of turn 1000 rather than the end. Plausibly we should poll people for their intuitions first.

pacmanboss256:
I can draft this for now

AnnikaCodes:
@pacmanboss256 are you still working on this?

pacmanboss256:
I think we left this as a policy decision, there really anything to change besides when it activates.

AnnikaCodes:
I'll close this, then.

PR #8390, opened 2021-07-10
Datasearch: Minor refactor

PartMan7:
This fixes/changes a couple things:

a) Long dexsearch queries now reject with an error instead of blindly trimming and giving stuff like ``retur could not be found in any of the search categories``.
b) Maxgen now works properly (before this, the maxgen parameter only worked if it was either the first term or if there were no spaces after the preceding comma).
c) Mod now allows spaces around the equals sign (`mod = SSB`) instead of matching properly and failing on the split.
d) Gen0 no longer works in abilitysearch / itemsearch (not that relevant, but since current behaviour is to block Gen 1/2, seems a bit weird to allow 0).
e) Gen X and Maxgen X now work in abilitysearch / itemsearch (before this, they'd give a blank 'no results found').
f) Gen X and Maxgen X now support multi-digit generations (well in advance before Gen 10 releases, but doesn't hurt to be ready).

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -86,10 +86,10 @@ export const commands: Chat.ChatCommands = {
 	async dexsearch(target, room, user, connection, cmd, message) {
 		this.checkBroadcast();
 		if (!target) return this.parse('/help dexsearch');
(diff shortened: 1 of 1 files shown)


Zarel:
Looks good; thanks!

PR #8392, opened 2021-07-11
Ignore 10000 speed upper limit in Custom Game

TheSupervillain:
https://www.smogon.com/forums/threads/change-custom-game-speed-mechanics.3686665/

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -562,7 +562,7 @@ export class Pokemon {
 			stat = this.battle.runEvent('Modify' + statTable[statName], this, null, null, stat);
 		}
 
-		if (statName === 'spe' && stat > 10000) stat = 10000;
+		if (statName === 'spe' && stat > 10000 && !this.battle.format.id.includes('customgame')) stat = 10000;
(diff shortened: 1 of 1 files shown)


sim/pokemon.ts
```diff
@@ -562,7 +562,7 @@ export class Pokemon {
 			stat = this.battle.runEvent('Modify' + statTable[statName], this, null, null, stat);
 		}
 
-		if (statName === 'spe' && stat > 10000) stat = 10000;
+		if (statName === 'spe' && stat > 10000 && !this.battle.format.id.includes('customgame')) stat = 10000;
```
DaWoblefet, line 565: Pretty sure this should be ``!this.debugMode``.

TheSupervillain: Well, `!this.battle.debugMode`, but yeah I get your point.

Zarel:
Good work!

But `debugMode` should not affect mechanics. Can you change this to detect a `trunc` override, instead?

Zarel:
Thanks!

PR #8393, opened 2021-07-13
Fix bugs with Sucrase build process

Karthik99999:
I feel like there should be a `.tools-dist` now.

Fixing this stuff got `.server-dist/index.js` to run, but there are still some more bugs with file paths, since the places where FS is called to access a file is not replaced like how something in `require` is. It would be quite cumbersome to account for sucrase in every part of the code that uses FS, so there has to be a better solution.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -23,6 +23,8 @@ To reload chat commands:
 
 */
 
+import * as path from 'path';
+
 import type {RoomPermission, GlobalPermission} from './user-groups';
 import type {Punishment} from './punishments';
 import type {PartialModlogEntry} from './modlog';
@@ -137,7 +139,9 @@ import ProbeModule = require('probe-image-size');
 const probe: (url: string) => Promise<{width: number, height: number}> = ProbeModule;
 
 const EMOJI_REGEX = /[\p{Emoji_Modifier_Base}\p{Emoji_Presentation}\uFE0F]/u;
-const TRANSLATION_DIRECTORY = `${__dirname}/../translations`;
+// to account for Sucrase
+const TRANSLATION_PATH = __dirname.endsWith('.server-dist') ? `../.translations-dist` : `../translations`;
+const TRANSLATION_DIRECTORY = path.resolve(__dirname, TRANSLATION_PATH);
 
 class PatternTester {
 	// This class sounds like a RegExp
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -23,8 +23,6 @@ To reload chat commands:
 
 */
 
-import * as path from 'path';
-
 import type {RoomPermission, GlobalPermission} from './user-groups';
 import type {Punishment} from './punishments';
 import type {PartialModlogEntry} from './modlog';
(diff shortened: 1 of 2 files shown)


AnnikaCodes:
Does `.server-dist/index.js` need to run? It's my understanding that actually running the server should be done with `ts-node`, whereas Sucrase is used for command-line tools and the NPM package, which would just be requiring specific parts of PS.

I'm fine with `.tools-dist` personally. You could make the `FS` constructor add `-dist` to the relevant paths if you wanted, but I don't think that's terribly necessary.

Karthik99999:
tbh getting index.js to run is more of a personal thing, since ts-node uses much more memory and I think there should still be the option of using the sucrase build to run ps on a server with lower memory. but otherwise yeah index.js does not need to be run and the FS thing shouldn't matter for the current intended use of the sucrase build.

Zarel:
Why does `.tools-dist` need to exist? It should be unused when building the NPM package, and we should be using `ts-node` for the rest.

Zarel:
Oh, `ts-node` memory consumption. There really isn't a good solution for this sort of problem, I guess. :(

AnnikaCodes:
AFAIK the long-term plan is to use `tsc` to build the NPM package; we should probably use it for low-memory servers as well. It is slower, but if you care that much about fast rebuilds you probably have a large enough server that `ts-node`'s overhead is fairly negligible (case in point: PS main).

Zarel:
What's `ts-node`'s overhead? Main doesn't seem to be _very_ happy with it.

AnnikaCodes:
It uses more memory for small testing servers, presumably because it compiles things as they are `require`d. I don't think it would cause the lag since it _should_ only be using CPU when it has to transpile requires/imports.

Zarel:
High memory usage → more GC use → more lag.

If the memory usage is only on startup, it's not a problem for us, though. I hope it is?

server/chat.ts
```diff
@@ -137,7 +137,9 @@ import ProbeModule = require('probe-image-size');
 const probe: (url: string) => Promise<{width: number, height: number}> = ProbeModule;
 
 const EMOJI_REGEX = /[\p{Emoji_Modifier_Base}\p{Emoji_Presentation}\uFE0F]/u;
-const TRANSLATION_DIRECTORY = `${__dirname}/../translations`;
+// to account for Sucrase
+const TRANSLATION_PATH = __dirname.endsWith('.server-dist') ? `../.translations-dist` : `../translations`;
+const TRANSLATION_DIRECTORY = `${__dirname}/${TRANSLATION_PATH}`;
```
Zarel, line 142: I wonder if we should put `translations/` inside `server/`

Zarel: They certainly don't translate anything outside of `server/` at the moment.

Zarel: And they shouldn't; sim translations go in `data/text/`.

Zarel:
Anyway these changes won't regress everything. But it definitely raises a lot of other questions we shouldn't forget about.

PR #8394, opened 2021-07-13
Add /notifyuser

PartMan7:
See [here](https://www.smogon.com/forums/threads/only-modnote-notifyrank-if-it-highlights-everyone.3687633/).

### DIFF
--- a/server/chat-commands/moderation.ts
+++ b/server/chat-commands/moderation.ts
@@ -1655,7 +1655,7 @@ export const commands: Chat.ChatCommands = {
 			}
 			if (notification.length > 300) return this.errorReply(`Notifications should not exceed 300 characters.`);
 			const message = `|tempnotify|${id}|${title}|${notification}`;
-			room.send(message);
+			room.sendUser(targetUser, message);
 		}
 	},
 	notifyuserhelp: [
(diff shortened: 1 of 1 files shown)


server/chat-commands/moderation.ts
```diff
@@ -1636,6 +1636,33 @@ export const commands: Chat.ChatCommands = {
...
+			room.sendUser(targetUser, `|tempnotifyoff|${id}`);
+		} else {
+			let [title, notification] = this.splitOne(titleNotification);
+			if (!title) title = `${room.title} notification!`;
+			if (!user.can('addhtml')) {
+				title += ` (notification from ${user.name})`;
+			}
+			if (notification.length > 300) return this.errorReply(`Notifications should not exceed 300 characters.`);
+			const message = `|tempnotify|${id}|${title}|${notification}`;
+			room.send(message);
```
AnnikaCodes, line 1658: Shouldn't this only be sent to the user, with `Room#sendUser`?

PartMan7: Right, whoops

I'd only tested with two users in the room

AnnikaCodes:
This seems fine, thanks!

AnnikaCodes (approved):
I'll wait a few days for others to comment and then merge.

Zarel:
Looks good to me, too.

Zarel (approved):

PR #8397, opened 2021-07-14
Gen 2: Move Nintendo Cup 2000 Move Legality to gen2stadium2

PlagueVonKarma:
Requested by @KrisXV to fix [this](https://www.smogon.com/forums/threads/pok%C3%A9mon-stadium-2-simulators-bugs-glitches-and-other-mistakes.3685286/#post-8917009).


Zarel:
Why does this work? Does stadium2 not inherit from gen2? Either way, even as a stopgap I don't think the rule should be deleted from gen2.

PlagueVonKarma:
> Why does this work? Does stadium2 not inherit from gen2? Either way, even as a stopgap I don't think the rule should be deleted from gen2.

According to Kris this should work at least? Apparently it's not importing the rule/learnset stuff properly and this is the solution? Just doing what he told me to do at the end of the day.

In regards to the thing w/ removing it from gen2, the format is never played in GSC. If you want it to stay there, I guess it could?

KrisXV:
Stadium2 inherits from gen 2 but that doesn’t seem to apply to rulesets. This also doesn’t need to be in gen 2 because it’ll never be used there

Zarel:
If you want it to error out when someone tries using it in gen 2, make it do that. It should either error out, or work as expected.

Marty-D:
The problem isn't that Stadium 2 doesn't inherit properly (it does); the problem is the part that checks Crystal tutors is only happening for the Pokemon in the `illegalCombos` list.

Zarel:
I guess that means this doesn't actually fix the problem? I'll draft to make that clear.

PR #8398, opened 2021-07-15
Stadium 2 Destiny Bond, Perish Song, and Recoil Moves Bugfixes. 

InAShellnut:
The current implementation of Perish Song and Destiny Bond on the Stadium 2 simulator causes them to always fail instead of only failing if the user has 1 Pokemon left, this pull fixes that and they now both work properly. Additionally, there was an issue with recoil moves not doing Recoil Damage if they KOed any Pokemon when it should only do no recoil damage if you KO their last Pokemon.

Bugs documented here: https://www.smogon.com/forums/threads/pok%C3%A9mon-stadium-2-simulators-bugs-glitches-and-other-mistakes.3685286/


Marty-D (commented):
Thanks!

PR #8399, opened 2021-07-15
Remove LC Uber as a valid tier in /ds help text

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8917371


Karthik99999:
If the actual solution is to make LC Uber work, I would try to do that but I have no clue what is going on in the code for this.

Zarel:
LC Uber _should_ be a valid tier but currently isn't.

PR #8401, opened 2021-07-16
Gen 2 Stadium: Fix crash

TheSupervillain:
Fixes `A battle crashed: TypeError: Cannot read property 'counter' of undefined`


Marty-D:
I feel like this Haze implementation was just copied from Gen 1, where it's also questionably done. Removing _all_ volatiles seems like a very wrong approach; it should remove only whatever Haze actually affects in the games.

What happens with Toxic in Stadium 2? I see the Gen 1 part of Haze that resets the Toxic counter wasn't copied here, so presumably it doesn't affect Toxic at all?

Marty-D:
If I understand the comment directly above Haze correctly, the only things it's supposed to remove other than stat stages in Stadium 2 are the `brnattackdrop` and `parspeeddrop` volatiles. So just focusing on those instead of cherry-picking stuff _not_ to remove out of literally every volatile would be the way to go, I'd think.

PR #8402, opened 2021-07-16
Gen 3: Type boosting items should boost the attacking stat

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-60#post-8917932

https://github.com/pret/pokeemerald/blob/9358ab33a5eacf7fb7aa573a59d26ebabe889058/src/pokemon.c#L3125

### DIFF
--- a/data/mods/gen2/items.ts
+++ b/data/mods/gen2/items.ts
@@ -5,7 +5,7 @@ export const Items: {[k: string]: ModdedItemData} = {
 	},
 	blackbelt: {
 		inherit: true,
-		onBasePower() {},
+		onModifyAtk() {},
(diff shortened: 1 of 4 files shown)


DaWoblefet:
I've got a skipped test in thick-fat.js that may pass with this change now - can you check if it's affected?

Karthik99999:
> I've got a skipped test in thick-fat.js that may pass with this change now - can you check if it's affected?

It passes, but according to the damage calc the damage range should be the same regardless of whether never-melt ice boosts bp or special attack so ¯\_(ツ)_/¯

Marty-D (approved):
Looks good, thanks!

PR #8403, opened 2021-07-16
Stadium 2: Send -unboost for stat drops

Karthik99999:
Third bug in this list https://www.smogon.com/forums/threads/pok%C3%A9mon-stadium-2-simulators-bugs-glitches-and-other-mistakes.3685286/


Zarel (approved):
Looks good! Leaving this open for a bit in case Marty/Wob want to weigh in.

Marty-D (approved):
Agreed, thanks!

PR #8404, opened 2021-07-16
Friends: Move caching to User#friends

mia-pi-git:
Is hotpatchable. Mostly PRing because I was asked to.


mia-pi-git:
Would be ideal if this could be merged soon.

server/chat-plugins/friends.ts
```diff
@@ -268,8 +272,8 @@ export const commands: Chat.ChatCommands = {
 				sendPM(`/uhtmlchange sent,`, targetUser.id);
 				sendPM(`/uhtmlchange undo,`, targetUser.id);
 			}
-			await Chat.Friends.cache.update(user.id);
-			await Chat.Friends.cache.update(target);
+			await Chat.Friends.updateCache(user);
```
Zarel, line 275: `updateCache` calls should probably be inside the actual handlers (i.e. inside `Friends.approveRequest`).

server/chat-plugins/friends.ts
```diff
@@ -268,8 +272,8 @@ export const commands: Chat.ChatCommands = {
 				sendPM(`/uhtmlchange sent,`, targetUser.id);
```
Zarel, line 272: Is there a need to wait for the database before sending UI updates?

mia-pi-git: That's why I would prefer to run them at the end of commands rather than in handlers - so we can await it and ensure the cache is updated (so we can avoid issues).

server/chat-plugins/friends.ts
mia-pi-git: See below.

server/chat-plugins/friends.ts
mia-pi-git: In case i misunderstood, we wait for the database because verification that we can actually perform this action is mostly done there.

server/friends.ts
```diff
@@ -67,13 +67,16 @@ function canPM(sender: User, receiver: User | null) {
...
 	file: string;
-	cache: Cache<Set<string>>;
 	constructor(file: string = DEFAULT_FILE) {
 		this.file = file === ':memory:' ? file : path.resolve(file);
-		this.cache = new Cache<Set<string>>(async user => {
-			const data = await this.getFriends(user as ID);
-			return new Set(data.map(f => f.friend));
-		});
+	}
+	async updateCache(user: User) {
```
AnnikaCodes, line 73: I feel like `updateUser` or `updateUserCache` or `updateUserObject` might be a better name here; when I hear `updateCache` I would think of a central cache rather than the `User#friends` set.

AnnikaCodes (approved):
I'm fine with this; it's a lot better than the odd synchronous caching in the other PR. @Zarel, do you have opinions about this?

server/friends.ts
mia-pi-git: I renamed it to `updateUserCache`.

Zarel:
I have other changes I'd want to make but this doesn't block that. I think this is probably fine although you're right about the name.

PR #8406, opened 2021-07-18
Fix Trademarked crash, improve validation

WeWuzNidokangz:
Trademarked rules: https://www.smogon.com/forums/threads/trademarked.3656980/#post-8295897

Somebody reported a crash to me in a Trademarked battle where they must have used a custom challenge to unrestrict Skill Swap, causing a stack overflow when the ability activation self-propagates indefinitely.

As well as making unsafe trademarks into hardcoded restrictions to prevent that, this change also splits up the failure conditions and gives them separate errors so that they are easier for users to understand (users sometimes had to PM me before to explain to them why their teams were invalid when e.g. they used the same unrestricted move as in the ability slot and in the moveset, since it currently throws the same error as if the trademark is restricted or non-existent). Finally, sleep moves are restricted instead of being hardcoded, since they aren't unsafe and can be displayed in the /om format data this way.

I don't think switching moves can immediately crash the battle the battle but they could lead to unlimited actions that ignore the 1000 turn limit, so it seems correct to keep that hardcoded restriction.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1472,9 +1472,10 @@ export const Formats: FormatList = [
 			'Arena Trap', 'Moody', 'Power Construct', 'Shadow Tag', 'Baton Pass',
 		],
 		restricted: [
-			'Baneful Bunker', 'Block', 'Copycat', 'Corrosive Gas', 'Detect', 'Destiny Bond', 'Disable', 'Encore', 'Fairy Lock', 'Ingrain', 'Instruct',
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

WeWuzNidokangz:
I should add that in spite of what the FAQs in the Trademarked OP say, Entrainment only activates once and is safe since you can't entrain a Pokemon with the same ability. If this PR can be accepted, I'll make a post updating them on what is safe.

WeWuzNidokangz:
(Sorry to ping reviewers with an irrelevant change, please ignore.)

KrisXV:
thanks!

PR #8410, opened 2021-07-24
Allow case-insensitive type names for Force Monotype rule

Karthik99999:
https://www.smogon.com/forums/threads/dont-require-capitalization-when-using-force-monotype.3688012/


Zarel:
The problem here is mostly that we do require good capitalization when writing `formats.ts`.

If there were a way to enforce them on one but not the other, that could work. One way is to require the normalized version returned from `validateRule`.

Karthik99999:
bump?



> The problem here is mostly that we do require good capitalization when writing `formats.ts`.

I feel like this isn't something that needs to be tested in code

Karthik99999:
Closing since this was done in another commit

KrisXV:
my bad, never saw this lol

PR #8411, opened 2021-07-25
Commands: Add feedback to /notifyuser

sirDonovan:


AnnikaCodes:
I'm fine with this; the reason feedback wasn't added originally is because the command was intended for bots, but this change doesn't hamper bots.

PR #8412, opened 2021-07-27
Remove 'lynch' from mafia.ts 

ayiadvance:
Final removal of the word "lynch" from mafia.ts due to racial considerations. All uses of the word have been changed to "vote" and all future references to the action will use "vote" instead of "lynch".

### DIFF
--- a/server/chat-plugins/mafia.ts
+++ b/server/chat-plugins/mafia.ts
@@ -765,7 +765,7 @@ class Mafia extends Rooms.RoomGame {
 		}
 
 		if (!player && this.dead[userid] && this.dead[userid].restless) player = this.dead[userid];
-		if (!player?.voting) return this.sendUser(userid, `|error|You are not voting anyone.`);
+		if (!player?.voting) return this.sendUser(userid, `|error|You are not voting for anyone.`);
 		if (player.lastVote + 2000 >= Date.now() && !force) {
 			return this.sendUser(
 				userid,
@@ -797,7 +797,7 @@ class Mafia extends Rooms.RoomGame {
 	}
 
 	/**
-	 * Returns HTML code that contains information on the current vote vote.
+	 * Returns HTML code that contains information about the current vote.
 	 */
 	voteBox() {
 		if (!this.started) return `<strong>The game has not started yet.</strong>`;
--- a/server/chat-plugins/mafia.ts
+++ b/server/chat-plugins/mafia.ts
@@ -55,14 +55,14 @@ interface MafiaRole {
 	image: string;
 }
 
-interface MafiaLynch {
-	// number of people on this lynch
+interface MafiaVote {
+	// number of people on this vote
 	count: number;
 	// number of votes, accounting for doublevoter etc
 	trueCount: number;
-	lastLynch: number;
+	lastVote: number;
 	dir: 'up' | 'down';
-	lynchers: ID[];
+	voters: ID[];
 }
 
 interface MafiaIDEAData {
@@ -170,10 +170,10 @@ class MafiaPlayer extends Rooms.RoomGamePlayer {
 	game: Mafia;
 	safeName: string;
 	role: MafiaRole | null;
-	lynching: ID;
+	voting: ID;
 	/** false - can't hammer (priest), true - can only hammer (actor) */
 	hammerRestriction: null | boolean;
-	lastLynch: number;
+	lastVote: number;
 	treestump: boolean;
 	restless: boolean;
 	silenced: boolean;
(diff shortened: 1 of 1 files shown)


server/chat-plugins/mafia.ts
```diff
@@ -683,191 +683,191 @@ class Mafia extends Rooms.RoomGame {
...
-		if (player && this.forceLynch && !force) {
+		// autoselfvote blocking doesn't apply to restless spirits
+		if (player && this.forceVote && !force) {
 			return this.sendUser(userid, `|error|You can only shift your vote, not unvote.`);
 		}
 
 		if (!player && this.dead[userid] && this.dead[userid].restless) player = this.dead[userid];
-		if (!player?.lynching) return this.sendUser(userid, `|error|You are not voting anyone.`);
-		if (player.lastLynch + 2000 >= Date.now() && !force) {
+		if (!player?.voting) return this.sendUser(userid, `|error|You are not voting anyone.`);
```
AnnikaCodes, line 768:
```suggestion
		if (!player?.voting) return this.sendUser(userid, `|error|You are not voting for anyone.`);
```
I think this flows better.

server/chat-plugins/mafia.ts
```diff
@@ -683,191 +683,191 @@ class Mafia extends Rooms.RoomGame {
...
+		player.voting = '';
+		player.lastVote = Date.now();
 		this.hasPlurality = null;
-		this.updatePlayersLynches();
+		this.updatePlayersVotes();
 	}
 
 	/**
-	 * Returns HTML code that contains information on the current lynch vote.
+	 * Returns HTML code that contains information on the current vote vote.
```
AnnikaCodes, line 800:
```suggestion
	 * Returns HTML code that contains information about the current vote.
```

Zarel:
I think "vote" is a bit unclear. Maybe something like "exile" or "execute" or "kill" or something that clarifies what it's a vote for?

Zarel:
"Hang"?

ayiadvance:
> I think "vote" is a bit unclear. Maybe something like "exile" or "execute" or "kill" or something that clarifies what it's a vote for?

We already began transitioning from lynch to vote so I think making another transition would be rather awkward at this point. The reason we chose vote over another word is that vote was already in the users' vocabulary, and saying "vote out" as a substitute for "lynch" was already in common practice.

Zarel:
Oh, I like "vote out". It's just "vote" itself is grammatically awkward in a lot of places, which I think you did notice.

ayiadvance:
Err I think vote makes more sense across the board. Think the disconnect might be because, the word "lynch" had two different yet similar meanings in terms of mafia.

The first definition is the act of voting someone for a lynch. "I'm lynching Abdel" means I'm voting for Abdel to be this days elimination. This was replaced by simply vote: "I'm voting Abdel". It's a bit more awkward to be like "I'm voting out Abdel" just because a "vote out" is an action by a group of people, not a single person. The second definition, then, is the person who had the most votes the day prior. "Abdel was lynched yesterday" means that Abdel had the most votes and is no longer in the game. Here, the better replacement is obviously "voted out". 

The meaning of lynch that most of the commands refer to is the first definition, hence the shift to "vote" instead of "vote out". In places where the second definition is used, such as the message when a player has received the most amount of votes and is immediately eliminated, "voted out" is the term that is used. Using "vote" in general also is less of a mouthful and makes abbreviations easier, and makes related terms such as "unvoteable" a lot simpler. 

We also changed to include "vote" as an option for commands a few months ago, so I really don't want to force the playerbase to go through another transition. The term “vote” has also long been used by other mafia communities, so it’s a natural term to transition to.

AnnikaCodes:
"I'm voting to kill Abdel"?

ayiadvance:
In most situations, yes, though there are themes in which the vote is for other things (which is why "vote" is arguably an even better general term than "lynch" was in the first place)

Zarel:
Yeah, sure, I just meant that I don't think "vote" is a good replacement for "lynch" in both senses. I like the idea of replacing it with "vote" in one sense and "vote out" in the other.

ayiadvance:
Yeah, in most senses where "vote out" would be better, that's what's used. Most of those cases would have already been changed in #8081 in which all of the text visible to users was changed from lynch to vote  but if anything looks off lmk (I don't actually think there are many places to use "vote out" instead of "vote" apart from the hammer message)

KrisXV:
I’d like to point out that in other mafia like games such as among us, “vote” is coined to mean “vote out”, so it should make sense regardless.

ayiadvance:
> Yeah, sure, I just meant that I don't think "vote" is a good replacement for "lynch" in both senses. I like the idea of replacing it with "vote" in one sense and "vote out" in the other.

So is there anything else that needs to be changed before this goes through?

Zarel:
No, I think you probably misinterpreted me. I think the current state of the PR is good.

PR #8414, opened 2021-07-29
Add a chat plugin for reporting GitHub actions

mia-pi-git:
Useful feature, should be able to replace the bot in terms of stability (with the same functionality).
To set this up, one needs to add a webhook to post `push` and `pull_request` actions to `{our url}:{port}/github/callback`, and then set `Config.github` to the `port` to listen on and the secret set in the webhook.

### DIFF
--- a/server/chat-plugins/github.ts
+++ b/server/chat-plugins/github.ts
@@ -93,9 +93,9 @@ export const GitHub = new class {
 			const repoName = this.getRepoName(repo);
 			const id = commit.id.substring(0, 6);
 			messages.development.push(
-				Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href=\"${url}\"><span style="color:#606060">${id}</span></a> ${message} <span style="color:#909090">(${username})</span>`
+				Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href="${url}"><span style="color:#606060">${id}</span></a> ${message} <span style="color:#909090">(${username})</span>`
 			);
-			messages.staff.push(Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href=\"${url}\">${message}</a> <span style="color:#909090">(${username})</span>`);
+			messages.staff.push(Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href="${url}">${message}</a> <span style="color:#909090">(${username})</span>`);
 		}
 		for (const k in messages) {
 			this.report(k as RoomID, repo, messages[k as RoomID]);
@@ -111,7 +111,7 @@ export const GitHub = new class {
 		const userName = this.getUsername(result.sender.login);
 		const title = result.pull_request.title;
 		let buf = Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <span style="color:#909090">${userName}</span> `;
-		buf += Utils.html`${action} <a href=\"${url}\">PR#${result.number}</a>: ${title}`;
+		buf += Utils.html`${action} <a href="${url}">PR#${result.number}</a>: ${title}`;
 		this.report('development', repo, buf);
 	}
 	report(roomid: RoomID, repo: string, messages: string[] | string) {
@@ -171,7 +171,7 @@ export const commands: Chat.ChatCommands = {
 			}
 			gitData.bans[toID(username)] = reason || " "; // to ensure it's truthy
 			GitHub.save();
-			this.privateModAction(`${user.name} banned their GitHub user ${username} from having their GitHub actions reported to this room.`);
+			this.privateModAction(`${user.name} banned the GitHub user ${username} from having their GitHub actions reported to this server.`);
 			this.modlog('GITHUB BAN', username, reason);
 		},
 		unban(target, room, user) {
@@ -183,7 +183,7 @@ export const commands: Chat.ChatCommands = {
 			delete gitData.bans[target];
 			if (!Object.keys(gitData.bans).length) delete gitData.bans;
 			GitHub.save();
-			this.privateModAction(`${user.name} allowed the GitHub user ${target} to have their GitHub actions reported to this room.`);
+			this.privateModAction(`${user.name} allowed the GitHub user ${target} to have their GitHub actions reported to this server.`);
 			this.modlog('GITHUB UNBAN', target);
 		},
 		bans() {
@@ -200,7 +200,7 @@ export const commands: Chat.ChatCommands = {
 			if (!gitData.usernames) gitData.usernames = {};
 			gitData.usernames[toID(gitName)] = username;
 			GitHub.save();
-			this.privateModAction(`${user.name} set ${gitName}'s name on GitHub actions to be ${username}.`);
+			this.privateModAction(`${user.name} set ${gitName}'s name on reported GitHub actions to be ${username}.`);
 			this.modlog('GITHUB SETNAME', null, `'${gitName}' to '${username}'`);
 		},
 		clearname: 'removeusername',
--- a/server/chat-plugins/github.ts
+++ b/server/chat-plugins/github.ts
@@ -1,0 +1,266 @@
+/**
+ * Chat plugin to view GitHub events in a chatroom.
+ * By Mia, with design / html from xfix's original bot, https://github.com/xfix/GitHub-Bot-Legacy/
+ * @author mia-pi-git
+ */
+
+import {FS, Utils} from '../../lib';
+
+const STAFF_REPOS = ['pokemon-showdown', 'pokemon-showdown-client', 'Pokemon-Showdown-Dex'];
+const COOLDOWN = 10 * 60 * 1000;
+
+export const gitData: GitData = JSON.parse(FS("config/chat-plugins/github.json").readIfExistsSync() || "{}");
+
+interface GitHookHandler {
+	on(event: 'pull_request', callback: (repo: string, ref: string | undefined, result: PullRequest) => void): void;
+	on(event: 'push', callback: (repo: string, ref: string, result: Push) => void): void;
+	on(event: string, callback: (repo: string, ref: string | undefined, result: any) => void): void;
+	listen(): void;
+	server: import('http').Server;
+}
+
+interface Push {
+	commits: Commit[];
+	sender: {login: string};
+	compare: string;
+}
+
+interface PullRequest {
+	action: string;
+	number: number;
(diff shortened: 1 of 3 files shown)


server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
+/**
+ * Chat plugin to view GitHub events in a chatroom.
+ * By Mia, with design / html from xfix's original bot, https://github.com/xfix/GitHub-Bot-Legacy/
+ * @author mia-pi-git
+ */
+
+import {FS, Utils} from '../../lib';
+
+const STAFF_REPOS = ['pokemon-showdown', 'pokemon-showdown-client', 'Pokemon-Showdown-Dex'];
```
KrisXV, line 9: The Dex isn’t on the smogon org

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+			const {message, url} = commit;
+			let shortened = /.+/.exec(message)?.[0] || "";
+			if (message !== shortened) {
+				shortened += '…';
+			}
+			const username = this.getUsername(commit.author.name);
+			const repoName = this.getRepoName(repo);
+			const id = commit.id.substring(0, 6);
+			messages.development.push(
+				Utils.html`[<font color='FF00FF'>${repoName}</font>] <a href=\"${url}\"><font color='606060'>${id}</font></a> ${shortened} <font color='909090'>(${username})</font>`
```
DaWoblefet, line 100: Although PS isn't generally HTML5 anyway, we should be using <span> tags with the style tag applied rather than the deprecated <font>. Something like ``<span style="color:#hexvalue">text</span>`` should work fine.

mia-pi-git: Fixed.

server/chat-plugins/github.ts
mia-pi-git: Said this in DMs, but repeating in public: as long as Zarel sets up the hook on the Dex repo, the org doesn't really matter.

AnnikaCodes, line 9: This feels very Main-specific. Can we put it in a config file or let Upper Staff configure it with a command?

mia-pi-git: I don't think it needs to be US, but yeah a command would probably be better.

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+	handlePush(repo: string, ref: string, result: Push) {
+		const branch = /[^/]+$/.exec(ref)?.[0] || "";
+		if (branch !== 'master') return;
+		const messages = {
+			staff: [] as string[],
+			development: [] as string[],
+		};
+		for (const commit of result.commits) {
+			const {message, url} = commit;
+			let shortened = /.+/.exec(message)?.[0] || "";
```
AnnikaCodes, line 92: errrrr... what's this for? `/.+/` should just match any string?

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+			const username = this.getUsername(commit.author.name);
+			const repoName = this.getRepoName(repo);
+			const id = commit.id.substring(0, 6);
+			messages.development.push(
+				Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href=\"${url}\"><span style="color:#606060">${id}</span></a> ${shortened} <span style="color:#909090">(${username})</span>`
+			);
+			messages.staff.push(Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <a href=\"${url}\">${shortened}</a> <span style="color:#909090">(${username})</span>`);
+		}
+		for (const k in messages) {
+			this.report(k as 'staff', repo, messages[k as 'staff']);
```
AnnikaCodes, line 105: Casting to `'staff'` is a little odd; why not just hardcode the `staff` and `development` rooms, or make the `report` method accept a roomid:Array<string> object?

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+		const action = this.isValidAction(result.action);
+		if (!action) return;
+		const repoName = this.getRepoName(repo);
+		const userName = this.getUsername(result.sender.login);
+		const title = result.pull_request.title;
+		let buf = Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <span style="color:#909090">${userName}</span> `;
+		buf += Utils.html`${action} <a href=\"${url}\">PR#${result.number}</a>: ${title}`;
+		this.report('development', repo, buf);
+	}
+	report(roomid: 'staff' | "development", repo: string, messages: string[] | string) {
```
AnnikaCodes, line 121:
```suggestion
	report(roomid: 'staff' | 'development', repo: string, messages: string[] | string) {
```

server/chat-plugins/github.ts
mia-pi-git: I chose to cast to Staff since it was that or just casting to 'staff' | 'development'. I didn't really see the need to cast to both and make it a little bit more annoying to read.

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+		if (this.isRateLimited(result.number)) return;
+		if (this.isGitbanned(result)) return;
+		const url = result.pull_request.html_url;
+		const action = this.isValidAction(result.action);
+		if (!action) return;
+		const repoName = this.getRepoName(repo);
+		const userName = this.getUsername(result.sender.login);
+		const title = result.pull_request.title;
+		let buf = Utils.html`[<span style="color:#FF00FF">${repoName}</span>] <span style="color:#909090">${userName}</span> `;
+		buf += Utils.html`${action} <a href=\"${url}\">PR#${result.number}</a>: ${title}`;
```
AnnikaCodes, line 118: Why are the quotes for `href` escaped, but not the quotes for `style`?

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+			if (room.auth.has(toID(GitHub.getUsername(username)))) {
+				return this.errorReply("That user is Dev roomauth. If you need to do this, demote them and try again.");
+			}
+			if (!gitData.bans) gitData.bans = {};
+			if (gitData.bans[toID(username)]) {
+				return this.errorReply(`${username} is already gitbanned.`);
+			}
+			gitData.bans[toID(username)] = reason || " "; // to ensure it's truthy
+			GitHub.save();
+			this.privateModAction(`${user.name} banned their GitHub user ${username} from having their GitHub actions reported to this room.`);
```
AnnikaCodes, line 178:
```suggestion
			this.privateModAction(`${user.name} banned the GitHub user ${username} from having their GitHub actions reported to this server.`);
```

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+		unban(target, room, user) {
+			room = this.requireRoom('development');
+			this.checkCan('mute', null, room);
+			target = toID(target);
+			if (!target) return this.parse('/help github');
+			if (!gitData.bans?.[target]) return this.errorReply("That user is not gitbanned.");
+			delete gitData.bans[target];
+			if (!Object.keys(gitData.bans).length) delete gitData.bans;
+			GitHub.save();
+			this.privateModAction(`${user.name} allowed the GitHub user ${target} to have their GitHub actions reported to this room.`);
```
AnnikaCodes, line 190:
```suggestion
			this.privateModAction(`${user.name} allowed the GitHub user ${target} to have their GitHub actions reported to this server.`);
```

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,266 @@
...
+		setname: 'addusername',
+		addusername(target, room, user) {
+			room = this.requireRoom('development');
+			this.checkCan('mute', null, room);
+			const [gitName, username] = Utils.splitFirst(target, ',').map(u => u.trim());
+			if (!toID(gitName) || !toID(username)) return this.parse(`/help github`);
+			if (!gitData.usernames) gitData.usernames = {};
+			gitData.usernames[toID(gitName)] = username;
+			GitHub.save();
+			this.privateModAction(`${user.name} set ${gitName}'s name on GitHub actions to be ${username}.`);
```
AnnikaCodes, line 207:
```suggestion
			this.privateModAction(`${user.name} set ${gitName}'s name to be ${username}.`);
```

server/chat-plugins/github.ts
AnnikaCodes: My suggestions remove the need for casting altogether.

mia-pi-git: >  roomid:Array object?

Would still require casting, no?

> why not just hardcode the staff and development rooms

I don't understand how this is different?

AnnikaCodes: > Would still require casting, no?

It shouldn't?
```ts
report(messages: {[k: string]: string[] | string}, repo: string) {
	for (const [roomid, message] of Object.entries(messages)) {
		if (!STAFF_REPOS.includes(repo) && roomid === 'staff') return;
		if (Array.isArray(message)) message = message.join('<br />');
		Rooms.get(roomid)?.add(`|html|<div class="infobox">${message}</div>`).update();
	}
}
```

server/chat-plugins/github.ts
mia-pi-git: Set it to a Config setting.

server/chat-plugins/github.ts
mia-pi-git: Removed it, wasn't sure on second look either.

server/chat-plugins/github.ts
mia-pi-git: I just made it cast to `RoomID`.

AnnikaCodes:
This looks OK to me; @Zarel do you have opinions?

server/chat-plugins/github.ts
```diff
@@ -0,0 +1,262 @@
...
+	usernames?: {[username: string]: string};
+	bans?: {[username: string]: string};
+}
+
+export const GitHub = new class {
+	readonly hook: GitHookHandler | null = null;
+	updates: {[k: string]: number} = Object.create(null);
+	constructor() {
+		try {
+			this.hook = (require('githubhook'))(Config.github);
```
AnnikaCodes, line 61: It would probably be a good idea to document `Config.github` somewhere; at least link to [the documentation](https://www.npmjs.com/package/githubhook).

PR #8415, opened 2021-07-31
Implement new STABmons policy

Karthik99999:
https://www.smogon.com/forums/threads/stabmons.3656429/page-8#post-8931262

Kris said he was busy and asked me to do this.

I will implement the client side changes in my existing PR for STABmons here: smogon/pokemon-showdown-client#1846

I manually checked that everything was working as it should. Do I have to write some tests in the team-validator tests file?

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1074,7 +1074,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 				const speciesTypes: string[] = [];
 				const moveTypes: string[] = [];
 				for (let i = this.dex.gen; i >= species.gen && i >= move.gen; i--) {
-					const dex = Dex.forGen(i);
+					const dex = this.dex.forGen(i);
 					moveTypes.push(dex.moves.get(move.name).type);
 					
 					const pokemon = dex.species.get(species.name);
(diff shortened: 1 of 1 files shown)


data/rulesets.ts
```diff
@@ -1085,21 +1086,27 @@ export const Rulesets: {[k: string]: FormatData} = {
...
 					prevo = prevoSpecies.prevo;
 				}
-				if (types.includes(move.type)) return null;
+				// Check for type changes from past generations
+				for (let i = dex.gen - 1; i >= species.gen && i >= move.gen; i--) {
+					const genDex = dex.forGen(i);
+					speciesTypes.push(...genDex.species.get(species.name).types);
+					moveTypes.push(genDex.moves.get(move.name).type);
+				}
+				if (moveTypes.some(m => speciesTypes.includes(m))) return null;
```
Karthik99999, line 1109: Should I reimplement the forme and prevo logic here? It's not required for the current mons this change affects, but maybe for future proofing?

Karthik99999: nevermind just figured out a better way to implement the generation loop

data/rulesets.ts
```diff
@@ -1071,35 +1071,41 @@ export const Rulesets: {[k: string]: FormatData} = {
...
-								const prevo = dex.species.get(baseSpecies.prevo);
-								if (prevo.evos.includes(formeName)) continue;
-							}
-							const forme = dex.species.get(formeName);
-							if (forme.changesFrom === originalForme.name && !forme.battleOnly) {
-								types = types.concat(forme.types);
+				const speciesTypes: string[] = [];
+				const moveTypes: string[] = [];
+				for (let i = this.dex.gen; i >= species.gen && i >= move.gen; i--) {
+					const dex = Dex.forGen(i);
```
KrisXV, line 1077:
```suggestion
					const dex = this.dex.forGen(i);
```

KrisXV:
This covers the most important part of the policy change, and we went over the rest in PMs, so this is good

PR #8418, opened 2021-08-02
Mafia: allow transitions between death states

ayiadvance:
Allow transitions between death states, as well as change the idea timer to 60 seconds, allow the lynch modifier of dead players to be changed, and start with reveals off.


server/chat-plugins/mafia.ts
```diff
@@ -986,14 +984,17 @@ class Mafia extends Rooms.RoomGame {
 			player.destroy();
 			return;
 		}
-		this.dead[player.id] = player;
+		if (toEliminate in this.playerTable) this.dead[toEliminate] = this.playerTable[toEliminate];
```
AnnikaCodes, line 987: Should dead players be removed from the `playerTable`?

ayiadvance: They are, further down in the method

AnnikaCodes: Oh, okay.

AnnikaCodes (approved):
This looks fine to me.

PR #8422, opened 2021-08-04
SQL: Use a QueryProcessManager subclass

mia-pi-git:
Draft because there's a critical bug in this code that causes the subprocesses to crash when started. No idea why. No errors are logged.

### DIFF
--- a/lib/sql.ts
+++ b/lib/sql.ts
@@ -202,19 +202,19 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any> 
 			}
 		}
 	}
-	all<T = any>(statement: string | Statement, data: DataType): Promise<T[]> {
+	all<T = any>(statement: string | Statement, data: DataType = []): Promise<T[]> {
 		if (typeof statement !== 'string') statement = statement.toString();
 		return this.query({type: 'all', statement, data});
 	}
-	get<T = any>(statement: string | Statement, data: DataType): Promise<T> {
+	get<T = any>(statement: string | Statement, data: DataType = []): Promise<T> {
 		if (typeof statement !== 'string') statement = statement.toString();
 		return this.query({type: 'get', statement, data});
 	}
-	run(statement: string | Statement, data: DataType): Promise<{changes: number}> {
+	run(statement: string | Statement, data: DataType = []): Promise<{changes: number}> {
 		if (typeof statement !== 'string') statement = statement.toString();
 		return this.query({type: 'run', statement, data});
 	}
-	transaction<T = any>(name: string, data: DataType): Promise<T> {
+	transaction<T = any>(name: string, data: DataType = []): Promise<T> {
 		return this.query({type: 'transaction', name, data});
 	}
 	async prepare(statement: string): Promise<Statement | null> {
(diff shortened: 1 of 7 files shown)


mia-pi-git:
I have since determined that this _only_ happens to the modlog database. Unsure of why.

mia-pi-git:
I fixed it. The bug was that in the child process, Modlog was not being instantiated, so the receiving child process wasn't there.

lib/sql.ts
```diff
@@ -16,310 +14,252 @@ export interface SQLOptions {
...
-	}
-	all(statement: Statement, data: DataType = {}) {
-		const num = typeof statement === 'number' ? statement : this.statements.get(statement);
-		if (num === undefined || ![...this.statements.values()].includes(num)) {
-			throw new Error(`Prepare a statement before using another database function with SQLDatabase.prepare.`);
+	setupDatabase() {
+		if (this.dbReady) return;
+		this.dbReady = true;
+		const {file, extension} = this.options;
+		// console.log(`loading database ${file}${extension ? ` and extension ${extension}` : ''}`);
```
AnnikaCodes, line 172:
```suggestion
```

lib/sql.ts
```diff
@@ -16,310 +14,252 @@ export interface SQLOptions {
...
-			process.send!(`STATEMENTS\n${JSON.stringify(Object.fromEntries(statements))}\n`);
-		}
-		if (onDatabaseStart) {
-			onDatabaseStart(database);
+		if (result === SQLiteErrors.DatabaseNotFound) {
+			throw new Error("SQLite database was queried, but Config.usesqlite is off");
 		}
+		return result;
 	}
-	database?.pragma(`foreign_keys=on`);
```
AnnikaCodes, line 235: does this line go anywhere? was it actually necessary?

server/modlog/index.ts
```diff
@@ -78,52 +77,62 @@ export interface ModlogEntry {
...
+	altsInsertionQuery: SQL.Statement | null = null;
+	renameQuery: SQL.Statement | null = null;
+	globalPunishmentsSearchQuery: SQL.Statement | null = null;
 
-	constructor(databasePath: string, sqliteOptions?: Partial<SQLOptions>) {
+	constructor(databasePath: string, options: Partial<SQL.Options>) {
 		this.queuedEntries = [];
 		this.databaseReady = false;
+		const dbExists = FS(databasePath).existsSync();
+		// if (process.send) console.log('trying db?');
```
AnnikaCodes, line 102:
```suggestion
```

server/modlog/index.ts
```diff
@@ -364,7 +373,7 @@ export class Modlog {
 		query += ` ${sortAndLimit.query}`;
 		args.push(...sortAndLimit.args);
 
-		return {statement: await this.database.prepare(query), queryText: query, args};
+		return {statement: (await this.database.prepare(query))!.toString(), queryText: query, args};
```
AnnikaCodes, line 376: Uhhhhhhh, why do we prepare it and then convert it to a string?

lib/sql.ts
mia-pi-git: Don't think it was needed. Consumers can choose themselves.

AnnikaCodes: OK.

AnnikaCodes:
Alright, let's hope this doesn't break anything.

PR #8424, opened 2021-08-08
Modlog: Fix double spaces when no moderation found

DaWoblefet:
Very, very minor; fixes double spaces for a couple modlog queries when nothing was found.

Before:
image

After:
image


Zarel:
Thanks!

PR #8425, opened 2021-08-08
Chat: Disallow HTTP images

DaWoblefet:
As of [Chrome 86](https://www.chromestatus.com/feature/4926989725073408), Chrome attempts to automatically upgrade images with an http source to be https before displaying the image. If the image can't be loaded in https, it'll just fail to load. Unlike Chrome, Firefox currently just marks the whole site as insecure if it finds image mixed content, indicated by the padlock having a triangle warning sign in the address bar.

By disallowing http images, we prevent users from displaying images that may not work in every browser. Compare the Chrome behavior to the Firefox behavior, with an image that can be upgraded from http -> https and an image that cannot.

Chrome 92:
image

Firefox 90:
image

Images in question:
http://images-wixmp-ed30a86b8c4ca887773594c2.wixmp.com/intermediary/f/578a8319-92b6-4d81-9d5f-d6914e6535a0/d5o541m-54dae5d4-710c-44d4-a898-71ea71d7bd28.jpg
http://www.math.com/images/logo.gif

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1236,7 +1236,6 @@ export class CommandContext extends MessageContext {
 		} else {
 			throw new Chat.ErrorMessage("Image URLs must begin with 'https://' or 'data:'; 'http://' cannot be used.");
 		}
-		
 	}
 	/**
 	 * This is a quick and dirty first-pass "is this good HTML" check. The full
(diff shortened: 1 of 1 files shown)


Zarel:
Thanks!

Zarel:
It seems uncommon enough these days that we shouldn't need to go out of our way to preserve it.

PR #8427, opened 2021-08-09
Support banning pokemon in randbats

Karthik99999:
https://www.smogon.com/forums/threads/supporting-custom-banlists-for-random-battles.3688882/

On the thread, I did express concern about crashes being easier to be purposefully caused if this is supported. Zarel did say this:

>>KarthikVGC said:
I think the only issue is that it would make it easier to cause crashes, since randbats formats with a max team size of 12 or under will crash when it is unable to create full teams for everyone. Otherwise this is trivial to add, requiring one extra line of code.
>
>You could handle this when validating rules. It would definitely be cool to have.

I'm not too sure what that means, though.


WeWuzNidokangz:
I was thinking about whether this could be plausible recently, but I was scared about it having potential for difficult to replicate/fix crashes and maybe performance problems...

I _think_ what Zarel meant in that thread would be having something like onValidateRule run on the Random Battle formats (as far as I know onValidateRule specifically is only called on rules at the moment) and check if the custom battle/tour the user requested is actually logically coherent (i.e. the team generator can/will generate enough valid sets to match the team sizes needed under the ban constraints) and return an error instead of creating a battle if it isn't. However, actually testing this seems complex, especially if bans on abilities (and other elements of regular tour code syntax?) are supported as well as Pokemon bans, as the requesting post outlines.

I thought about trying to generate a team under the custom rule conditions at validation time to test whether it's possible to create enough Pokemon to fulfill the team sizes, but thinking about it more I'm pretty sure this would be unsafe in certain cases, like with complex bans. Like, depending on the first set selected, the generator may or may not randomly lock itself out of being able to generate enough sets to match the order, even if additional sets are strictly necessary for the battle to run properly (like a second set in a doubles format). (Maybe if team-wide complex bans and maybe some other features are disallowed and throw an error at validation time as a workaround, this approach might become safe albeit expensive?)

AnnikaCodes:
I think we're overthinking this a bit. The purpose of crashing when we can't make a team is to alert programmers (read: me) that there's a bug in the team generation code that's preventing normal team generation. Because of this, I made it not crash when it fails to generate a team that's supposed to contain more than 12 Pokémon; this is not a normal use-case for the generator (it was designed to build 6-Pokémon teams), so it's not a bug (or at least not an urgent one) if it can't do so. I don't think checking for "logical coherence" is really necessary.

Since this would require two hardcodes (one for team size and one for custom bans), I would advise you to just not crash when a team can't be properly generated if rules with user-specified components are in play (such as custom banlists or `Max Team Size`, etc). If you want the user to know that it's not a bug, you could display a message along the lines of "the Random Battle team generator couldn't generate a full team under the given ruleset". I'm willing to improve the team generator to make it work better for custom rules, but I don't want to compromise the quality of normal teams. Moreover, these improvements aren't urgent and don't merit showing crash messages.

Zarel:
A simple way to build in some additional protections is to just generate a team when validating the ruleset. If it fails the first time, you can exit early.

If it only fails sometimes, that's a different problem and you can handle it by not crashing and just displaying an error message. But I think that should be able to catch most issues.

Karthik99999:
I made it so that an error isn't thrown if there are user specific bans. Might not not the best solution but it was the easiest

Karthik99999:
i thought for sure the timeout was because of all the `getRuleTable` calls... I really don't know what else to do

AnnikaCodes:
The current CI failure isn't related to tests timing out; you declare the same variable name (`type`) twice, once as a normal `const` and once in as part of a for loop.

Karthik99999:
Well now it failed because of a timeout

WeWuzNidokangz:
@Karthik99999 
Hi, I made a PR similar to this for Hackmons Cup (https://github.com/smogon/pokemon-showdown/pull/8607). It is mostly complimentary to this I think but it adds a few lines that would interfere with this, sorry (this.enforceNoDirectCustomBanlistChanges(); at the start of each mod's randomTeam() - I will remove this if your PR is merged first). If your PR is merged first, I think it needs something similar to https://github.com/smogon/pokemon-showdown/pull/8607/files#diff-02adab0a3629b17a5f6abab6293f19194f931ac8dae178d75ecbe717ab68551c to enable any banlist changes to pass the validator.

AnnikaCodes:
Is this ready for review? Can you fix the merge conflicts, please?

Karthik99999:
Fixed the merge conflicts, still don't know about those tests that keep timing out

AnnikaCodes:
I don't know what's up with the Sucrase failure; the Gen 1 test failure is probably because team generation is less performant now.

AnnikaCodes:
This is on my to-do list to investigate, but relatively low.

mia-pi-git:
@Karthik99999, are you still interested in working on this?

PR #8429, opened 2021-08-09
Fix Skydrop + Eject Button interaction

jmclemo6:
Fixes https://github.com/smogon/pokemon-showdown/projects/3#card-61614927

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -1397,6 +1397,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		onAfterMoveSecondary(target, source, move) {
 			if (source && source !== target && target.hp && move && move.category !== 'Status' && !move.isFutureMove) {
 				if (!this.canSwitch(target.side) || target.forceSwitchFlag || target.beingCalledBack) return;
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Nicely done! Thanks :)

data/items.ts
```diff
@@ -1397,6 +1397,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		onAfterMoveSecondary(target, source, move) {
 			if (source && source !== target && target.hp && move && move.category !== 'Status' && !move.isFutureMove) {
 				if (!this.canSwitch(target.side) || target.forceSwitchFlag || target.beingCalledBack) return;
+				if (target.isSkyDropped() || source.isSkyDropped()) return;
```
urkerab, line 1400: (Surely this is overkill as all that matters is that the holder of the button is sky dropped, and the move user can't be in the middle of sky drop if it's just damaged the holder.)

DaWoblefet: Yep, good catch - simplified the logic in https://github.com/smogon/pokemon-showdown/commit/3c7fc87b6bf829f87f0bbab462e177afa5770dd6.

PR #8430, opened 2021-08-10
Fix Metronome (Item) and two-turn moves interaction

jmclemo6:
Should fix https://github.com/smogon/pokemon-showdown/projects/3#card-61614782

Also adds a test for the case of a two-turn move that bypasses its charge turn and fixes the original broken test that was here.

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -3388,8 +3388,12 @@ export const Items: {[itemid: string]: ItemData} = {
 				}
 				if (this.effectState.lastMove === move.id && pokemon.moveLastTurnResult) {
 					this.effectState.numConsecutive++;
-				} else if (pokemon.volatiles['twoturnmove'] && this.effectState.lastMove !== move.id) {
-					this.effectState.numConsecutive = 1;
+				} else if (pokemon.volatiles['twoturnmove']) {
+					if (this.effectState.lastMove !== move.id) {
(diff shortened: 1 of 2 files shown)


jmclemo6:
I think this should be good to just squash merge. If I need to rebase it to make that clean, let me know.

Zarel (approved):
Looks fine to me.

Marty-D (approved):
This is good for now; eventually we just have to pull all this consecutive move functionality out of Metronome, apply the counter to every Pokemon regardless of item, and use it for Metronome, Fury Cutter, and Destiny Bond.

Thanks! 👍

jmclemo6:
> This is good for now; eventually we just have to pull all this consecutive move functionality out of Metronome, apply a counter to every Pokemon regardless of item, and use it for Metronome, Fury Cutter, and Destiny Bond.
> 
> Thanks! 👍 

Replying to this a bit late, but I have a feeling this will partially help with another Metronome bug as well (https://github.com/smogon/pokemon-showdown/projects/3#card-61614791). I can't guarantee anything due to my current schedule, but if I'll try to work on adding the counter to Pokémon directly. If anyone else wants to pick it up, just let me know here, on PS, or via the development Discord server and I can update you on what work (if any) I've been able to do.

PR #8431, opened 2021-08-10
Gen 8 Sketch rule

EnPassant4264:
Approved at https://www.smogon.com/forums/threads/clause-for-allowing-sketching-gen-viii-moves.3688966/

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1864,6 +1864,10 @@ export class TeamValidator {
 		 * (This is everything except in Gen 1 Tradeback)
 		 */
 		const noFutureGen = !ruleTable.has('allowtradeback');
+		/**
+		 * The format allows Sketch to copy moves in Gen 8
+		 */
+		const noSketchBlock = ruleTable.has('gen8sketch');
 
 		let tradebackEligible = false;
 		while (species?.name && !alreadyChecked[species.id]) {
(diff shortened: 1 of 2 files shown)


sim/team-validator.ts
```diff
@@ -1964,7 +1968,7 @@ export class TeamValidator {
 						if (dex.gen >= 5 && learnedGen <= 4 && [
 							'cut', 'fly', 'surf', 'strength', 'rocksmash', 'waterfall', 'rockclimb',
 						].includes(moveid)) {
-							cantLearnReason = `can't be transferred from Gen 3 to 4 because it's an HM move.`;
+							cantLearnReason = `can't be transferred from Gen 4 to 5 because it's an HM move.`;
```
DaWoblefet, line 1971: This seems unrelated to the rest of the PR.

EnPassant4264: It is, but it is also something that needs to be fixed (appears to have been copy/pasted from the 3 -> 4 check above). Since I'm editing the function already, I don't see a reason to create a whole new PR for a single line.

EnPassant4264: @KrisXV awaiting your opinion on this please

EnPassant4264: bump

Zarel:
Big fan of this change! Thanks for doing this, and sorry it took so long to review this.

PR #8434, opened 2021-08-14
Update item users on some items

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8944777

- Added Marowak-Alola to Thick Club
- Added Farfetch'd-Galar to Leek
- Added all Pikachu formes to Light Ball

The bug report mentioned stuff like the base formes of things like Giratina and Zacian not having their forme specific items show up, but I didn't add them cause I feel like there was a reason this was done. If there isn't then I can do those too.

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -2825,7 +2825,7 @@ export const Items: {[itemid: string]: ItemData} = {
 				return critRatio + 2;
 			}
 		},
-		itemUser: ["Farfetch\u2019d", "Sirfetch\u2019d"],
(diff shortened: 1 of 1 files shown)


DaWoblefet:
It makes sense to me not to include those forme change items - we don't want to recommend people choosing Giratina with Griseous Orb and Pressure, for example.

DaWoblefet:
Oh actually, this is still missing Totem Alolan Marowak.

Zarel:
Hm, I definitely wanted some way to specify "all formes" but this is probably a better implementation if it's only a few pokemon.

DaWoblefet:
Thanks Karthik!

PR #8438, opened 2021-08-17
Process-Manager: Crashlog bad JSON responses

mia-pi-git:


AnnikaCodes:
I don't normally merge ProcessManager-related changes, but JSON parsing is currently causing a mysterious crash and I would like to be more able to debug it. If it starts breaking things, I will revert it.

lib/process-manager.ts
```diff
@@ -141,11 +143,28 @@ export class QueryProcessWrapper<T, U> implements ProcessWrapper {
 			const resolve = this.pendingTasks.get(taskId);
 			if (!resolve) throw new Error(`Invalid taskId ${message.slice(0, nlLoc)}`);
 			this.pendingTasks.delete(taskId);
-			resolve(JSON.parse(message.slice(nlLoc + 1)));
+			const resp = this.safeJSON(message.slice(nlLoc + 1));
+			if (resp instanceof Error) {
+				// we'd fail here anyway if it crashed, so at least this way we can log it
+				return false;
```
Zarel, line 149: This should probably `resolve` something so we don't have an infinitely waiting Promise somewhere.

mia-pi-git: It might be nice to use [resolve, reject] pairs instead, so we could reject it there.

mia-pi-git: I'll just make it resolve undefined for now, though. Whenever the JSON.parse fails, it's usually undefined anyway.

lib/process-manager.ts
```diff
@@ -141,11 +143,28 @@ export class QueryProcessWrapper<T, U> implements ProcessWrapper {
...
+	safeJSON(obj: string): any {
+		try {
+			return JSON.parse(obj);
+		} catch (e) {
+			// this is in the parent, so it should usually exist, but it's possible
+			// it's also futureproofing in case other external modules require this
+			// we also specifically do not throw here because this json might be sensitive,
+			// so we only want it to go to emails
+			global.Monitor?.crashlog?.(e, `a ${path.basename(this.file)} process`, {result: obj});
+			return new Error(e);
```
Zarel, line 165: This is a bug, the first argument passed to the `Error` constructor must be a string.

What is with all this weirdness? Just return `undefined`.

PR #8439, opened 2021-08-18
SQL: Automatically prepare statements unless told otherwise

mia-pi-git:
Mostly PRing so @AnnikaCodes can review and decide if she wants her consumers to fully use this (i left in a lot of `this.somethingQuery = await this.db.prepare(....)`.
This makes it so by default, the child process will no longer require you to call .prepare before you query the child again to use the statement. There's a new parameter, `noPrepare?`, added to all the functions (`get`, `run`, `all`) that use prepared statements. If this is passed, it will indeed require you to call `.prepare` before using it, and if it is not prepared, it will throw.
Otherwise, if this isn't passed, the child process will try to find the cached statement, and if not, prepare it and cache it.

### DIFF
--- a/lib/sql.ts
+++ b/lib/sql.ts
@@ -178,7 +178,7 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any> 
 		const statement = query.noPrepare ?
 			this.state.statements.get(query.statement) :
 			this.cacheStatement(query.statement);
-		if (!statement) throw new Error(`Missing cached statement ${query.statement} where required`);
+		if (!statement) throw new Error(`Missing cached statement "${query.statement}" where required`);
 		return statement;
 	}
 	setupDatabase() {
--- a/lib/sql.ts
+++ b/lib/sql.ts
@@ -7,7 +7,6 @@ import type * as sqlite from 'better-sqlite3';
 import {FS} from './fs';
 
 export const DB_NOT_FOUND = null;
-export const STATEMENT_NOT_FOUND = false;
 
 export interface SQLOptions {
 	file: string;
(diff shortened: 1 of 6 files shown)


AnnikaCodes:
For my code, I would like to continue to pre-prepare statements that are used in transactions and in multiple places (so that there is a single place the query text is kept to make modifications easier), but for queries that only are called in one method I'm fine with just calling Chat.database.all on a string. However, I would like to do that on my own when I've got more free time - I'm currently too busy with real-life stuff to spend time working on Trivia and Modlog. My advice is to keep it backwards-compatible for now and remind me to look at this later (you know when I'll be more available :p).

mia-pi-git:
> r my code, I would like to continue to pre-prepare statements that are used in transactions and in multiple places (so that there is a single place the query text is kept to make modifications easier), but for queries that only are called in one method I'm fine with just calling Chat.database.all on a string.

Alright, that's basically what I have here for now. It should still be backwards-compatible as well.

server/modlog/index.ts
```diff
@@ -373,18 +372,17 @@ export class Modlog {
...
 
 		return {
-			statement: await this.database.prepare(query) as SQL.Statement,
 			queryText: query,
 			args,
 		};
 	}
 
-	async prepareSQLSearch(
+	prepareSQLSearch(
```
AnnikaCodes, line 380: You can probably remove a lot of unnecessary `async`/`await` in unit tests now.

lib/sql.ts
```diff
@@ -169,6 +162,25 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any>
 		};
 		if (!this.isParentProcess) this.setupDatabase();
 	}
+	private cacheStatement(source: string) {
+		source = source.trim();
```
AnnikaCodes, line 166:
```suggestion
```
you already trim it in extractStatement()

lib/sql.ts
```diff
@@ -169,6 +162,25 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any>
...
+		return statement;
+	}
+	private extractStatement(
+		query: DatabaseQuery & {statement: string, noPrepare?: boolean}
+	) {
+		query.statement = query.statement.trim();
+		const statement = query.noPrepare ?
+			this.state.statements.get(query.statement) :
+			this.cacheStatement(query.statement);
+		if (!statement) throw new Error(`Missing cached statement ${query.statement} where required`);
```
AnnikaCodes, line 181:
```suggestion
		if (!statement) throw new Error(`Missing cached statement "${query.statement}" where required`);
```

lib/sql.ts
mia-pi-git: Yeah, but this trims here as well as a precaution in case you're doing `noPrepare`.

AnnikaCodes (commented):
Seems mostly fine to me at a glance.

lib/sql.ts
AnnikaCodes: `cacheStatement` isn't called when `noPrepare` is true, right? you just check if it's in the map directly

mia-pi-git: Yeah, but that's still a safety check.

AnnikaCodes: `cacheStatement` is only called directly after trimming the text.

mia-pi-git: it's trimmed in extract because in the this.state.statements.get call in exact, we want to ensure safety. It's not because of cacheStatement.

PR #8441, opened 2021-08-21
Fix Emergency Exit not activating after recoil damage

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8947695

Specifically, after recoil from moves like Flare Blitz and struggle. I've added tests for both types.

### DIFF
--- a/test/sim/abilities/emergencyexit.js
+++ b/test/sim/abilities/emergencyexit.js
@@ -353,4 +353,28 @@ describe(`Emergency Exit`, function () {
 
 		assert.equal(battle.requestState, 'switch');
 	});
+
+	it(`should request a switchout after taking regular recoil damage`, function () {
+		battle = common.createBattle([[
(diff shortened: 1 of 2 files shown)


test/sim/abilities/emergencyexit.js
```diff
@@ -353,4 +353,28 @@ describe(`Emergency Exit`, function () {
...
+		]])
+		const eePokemon = battle.p1.active[0];
+		battle.makeChoices();
+		assert.atMost(eePokemon.hp, eePokemon.maxhp / 2);
+		assert.equal(battle.requestState, 'switch');
+	});
+
+	it(`should request a switchout after taking struggle recoil damage`, function () {
+		battle = common.createBattle([[
+			{species: 'Golisopod', ability: 'Emergency Exit', moves: ['sketch']},
```
DaWoblefet, line 372: You could expedite this with Assault Vest + status move.

DaWoblefet:
Looks good to me! Thanks Karthik!

PR #8442, opened 2021-08-22
Support persisting battles in a Postgres database

mia-pi-git:
This should make restarts much easier to do.
With this code, an admin can just use /kill directly, rather than /lockdown (then waiting a decent length of time for battles to end.) This is still an option with the command `/kill nosave`, which will kill the server without saving battles (and requires lockdown). Otherwise, battles are saved to a postgres database, and then after the restart are recreated. When users log in, they are joined to their battles (if they exist) and they can continue.
This is all assuming postgres is enabled via `Config.usepostgres`. If not, it defaults to the normal system.
This supersedes #7005, as this should use less memory and be more efficient.


server/chat-commands/admin.ts
```diff
@@ -1122,17 +1122,42 @@ export const commands: Chat.ChatCommands = {
...
+		this.addModAction(`${user.name} used /savebattles`);
+	},
+
+	async kill(target, room, user) {
 		this.checkCan('lockdown');
+		const noSave = toID(target) === 'nosave';
 
-		if (Rooms.global.lockdown !== true) {
-			return this.errorReply("For safety reasons, /kill can only be used during lockdown.");
+		if (Rooms.global.lockdown !== true && noSave) {
```
AnnikaCodes, line 1138: Shouldn't this also check if battles _can_ be saved and if they cannot, set noSave to true? This will allow `/kill`ing without lockdown or saving if postgresql is not installed or configured.

server/room-battle.ts
```diff
@@ -675,6 +679,10 @@ export class RoomBattle extends RoomGames.RoomGame {
 		void this.stream.write(`>${player.slot} undo`);
 	}
 	joinGame(user: User, slot?: SideID, playerOpts?: {team?: string}) {
+		if (!this.options.players?.includes(user.id)) {
+			user.popup(`You cannot join this battle, as you were not originally in it.`);
```
AnnikaCodes, line 683:
```suggestion
			user.popup(`You cannot join this battle, as you were not originally playing in it.`);
```

test/lib/postgres.js
```diff
@@ -0,0 +1,47 @@
...
+	return it;
+}
+
+// only run these if you already have postgres configured
+describe.skip("Postgres features", () => {
+	it("Should be able to connect to a database", async () => {
+		this.database = new PostgresDatabase();
+	});
+	it("Should be able to insert data", async () => {
+		assert.doesNotThrow(async () => {
```
AnnikaCodes, line 20: `doesNotThrow` doesn't work with Promises IIRC.
```suggestion
		await assert.doesNotThrowAsync(async () => {
```

mia-pi-git:
> I can’t say I thoroughly understand this code, but there doesn’t seem to be any safeguard against a battle that’s already been saved being played out more before the restart and then recreated in the previous state after the restart which would most certainly lead to complaints. Is that a valid concern?

In theory, it should be fast enough we shouldn't have a problem. If we test it on main and find it _is_ a problem, I can make something against it, but even so I wouldn't find it much of an issue. If there's enough time for a game to be played significantly out, then there's a bit of a different issue.

server/chat-commands/admin.ts
mia-pi-git: Yeah. Fixed.

mia-pi-git:
> Speaking of, this system doesn’t seem to save tournaments. What happens when it recreates a battle that’s supposed to be part of a tournament that no longer exists?
Kris is working on this separately. This code needs to ship before that, though.

AnnikaCodes:
Also, some people may want to run side servers and utilize this feature with significantly more limited resources than Main.

mia-pi-git:
> Also, some people may want to run side servers and utilize this feature with significantly more limited resources than Main.

Less worried about that, but I did a. check and 410 battles on my computer took a few seconds to save. Not sure of a good solution here other than disallowing moves on the battle after it's been saved, to until the restart is done.

mia-pi-git:
> Freezing the battle is what I would suggest, although I think it would be safer to do it just before it's saved instead of after.

Yeah, I think that's what I'll do for now.

test/lib/postgres.js
```diff
@@ -0,0 +1,47 @@
...
+		assert.doesNotThrowAsync(async () => {
+			await this.database.query(`CREATE TABLE test (col TEXT, col2 TEXT)`);
+			await this.database.query(
+				`INSERT INTO test (col, col2) VALUES ($1, $2)`,
+				['foo', 'bar'],
+			);
+		});
+	});
+	testMod('sql-template-strings')('Should support sql-template-strings', async () => {
+		assert.doesNotThrowAsync(async () => {
```
AnnikaCodes, line 29: These should be awaited.

mia-pi-git: Oops.

lib/postgres.ts
```diff
@@ -0,0 +1,76 @@
...
+import type {SQLStatement} from 'sql-template-strings';
+import * as Streams from './streams';
+
+export class PostgresDatabase {
+	private pool: PG.Pool;
+	constructor(config = PostgresDatabase.getConfig()) {
+		this.pool = config ? new (require('pg').Pool)(config) : null!;
+	}
+	async query(statement: string | SQLStatement, values?: any[]) {
+		if (!this.pool) {
```
monsanto, line 18: Is there any reason to have `pg` be an optional dependency when it is a non-native module?

lib/postgres.ts
```diff
@@ -0,0 +1,76 @@
...
+		let config: AnyObject = {};
+		try {
+			config = require('../config/config').usepostgres;
+			if (!config) throw new Error('Missing config for pg database');
+		} catch (e) {}
+		return config;
+	}
+	async transaction(callback: (conn: PG.PoolClient) => any, depth = 0): Promise<any> {
+		const conn = await this.pool.connect();
+		await conn.query(`BEGIN;`);
```
monsanto, line 40: I don't think you need to end these with semicolons

lib/postgres.ts
```diff
@@ -0,0 +1,76 @@
...
+		let result;
+		try {
+			// eslint-disable-next-line callback-return
+			result = await callback(conn);
+		} catch (e) {
+			await conn.query(`ROLLBACK;`);
+			// two concurrent transactions conflicted, try again
+			if (e?.code === '40001' && depth <= 10) {
```
monsanto, line 48: Why would `e` be undefined?

databases/schemas/stored-battles.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE stored_battles (
+	roomid TEXT NOT NULL PRIMARY KEY, -- can store both the num and the formatid
+	input_log TEXT[] NOT NULL,
+	players TEXT[] NOT NULL,
+	title TEXT NOT NULL
+);
+
+CREATE INDEX by_roomid ON stored_battles (roomid);
```
monsanto, line 8: Redundant if roomid is primary key

databases/schemas/stored-battles.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE stored_battles (
+	roomid TEXT NOT NULL PRIMARY KEY, -- can store both the num and the formatid
```
monsanto, line 2: Is this preferable to having the num and formatid in separate columns?

mia-pi-git: It's more convenient usage-wise. Is it a problem storage-wise?

databases/schemas/stored-battles.sql
```diff
@@ -0,0 +1,8 @@
...
+	input_log TEXT[] NOT NULL,
```
monsanto, line 3: Is having an array here preferable to just having newlines in the text? You seem to do `join("\n")` later on?

mia-pi-git: Yeah, I was gonna ask about that. Is there more storage used if it's an array?

server/rooms.ts
```diff
@@ -1253,6 +1253,72 @@ export class GlobalRoomState {
...
+				for (const num of ['p1', 'p2', 'p3', 'p4'] as const) {
+					if (room.battle[num]?.id) {
+						players.push(room.battle[num].id);
+					}
+				}
+			}
+			if (!players.length || !log?.length) continue; // ???
+			await logDatabase.query(
+				`INSERT INTO stored_battles (roomid, input_log, players, title) VALUES ($1, $2, $3, $4)` +
+				` ON CONFLICT (roomid) DO UPDATE SET input_log = $2, players = $3, title = $4`,
```
monsanto, line 1278: use `input_log = EXCLUDED.input_log` etc instead of rebinding the values, and maybe use SQL template strings instead of providing the array afterwards?

server/rooms.ts
```diff
@@ -1253,6 +1253,72 @@ export class GlobalRoomState {
...
+			const log = await room.battle.getLog();
+			const players: ID[] = room.battle.options.players || [];
+			if (!players.length) {
+				for (const num of ['p1', 'p2', 'p3', 'p4'] as const) {
+					if (room.battle[num]?.id) {
+						players.push(room.battle[num].id);
+					}
+				}
+			}
+			if (!players.length || !log?.length) continue; // ???
```
monsanto, line 1275: lol `// ???` is even more confusing for the reviewer than it is for you... what is your confusion?

databases/schemas/stored-battles.sql
monsanto: It looks like later in the code you do `const [, formatid] = roomid.split('-')` so it might be better to just store them in the right format to begin with

server/rooms.ts
mia-pi-git: tl;dr this should never happen.

server/rooms.ts
```diff
@@ -1253,6 +1253,72 @@ export class GlobalRoomState {
...
+				players,
+				delayedStart: true,
+			});
+			if (!room) continue; // ???
+			room.battle?.start();
+			for (const [i, p] of players.entries()) {
+				room.auth.set(p, Users.PLAYER_SYMBOL);
+				const u = Users.getExact(p);
+				if (u) {
+					room.battle?.joinGame(u, `p${i + 1}` as SideID);
```
monsanto, line 1308: How about instead of doing `room.battle?` more than once, you just `continue` if `!room || !room.battle`

databases/schemas/stored-battles.sql
mia-pi-git: Yeah, but I also use the direct roomid a few times. Suppose it's not a big deal either way.

mia-pi-git: Oh, actually, direct roomid is maybe better, since that stores the "password" / existence of a password on the roomid as well.

databases/schemas/stored-battles.sql
monsanto: I'm not sure, probably not too much if so

lib/postgres.ts
```diff
@@ -0,0 +1,76 @@
...
+		} catch (e) {
+			await conn.query(`ROLLBACK`);
+			// two concurrent transactions conflicted, try again
+			if (e.code === '40001' && depth <= 10) {
+				return this.transaction(callback, depth + 1);
+				// There is a bug in Postgres that causes some
+				// serialization failures to be reported as failed
+				// unique constraint checks. Only retrying once since
+				// it could be our fault (thanks chaos for this info / the first half of this comment)
+			} else if (e?.code === '23505' && !depth) {
```
monsanto, line 54: How come this one has `e?.code` and the other doesn't?

mia-pi-git: Missed that one.

databases/schemas/stored-battles.sql
AnnikaCodes: Being able to search for battles in a particular format without having to deal with `LIKE` would make things like [`battlesearch`](https://crates.io/crates/battlesearch) simpler, and I think that `(format, gameid, password)` would provide just as much functionality as battlesearch? Ultimately it's not very important either way.

mia-pi-git: This won't work for battlesearch. We delete the saved battles as they're loaded from the DB.

lib/postgres.ts
```diff
@@ -0,0 +1,80 @@
...
+		} catch (e) {
+			// postgres won't give accurate stacks unless we do this
+			throw new Error(e.message);
+		}
+		return result?.rows || [];
+	}
+	static getConfig() {
+		let config: AnyObject = {};
+		try {
+			config = require('../config/config').usepostgres;
```
AnnikaCodes, line 37: Why can't we use the `Config` global or have configuration information passed in to the constructor? Both of those  sound significantly better than requiring a hardcoded configuration path.

databases/schemas/stored-battles.sql
AnnikaCodes: Oh, this is meant to coexist with terabytes of JSON battle logs, not supplant it. Then it doesn't matter to me at all.

lib/postgres.ts
mia-pi-git: Passing a different configuration in the constructor. is supported. This is the default configuration, for ease of use.

AnnikaCodes: I thought `lib/` was supposed to be portable outside of PS? This makes it dependent on PS' directory structure, and I don't really think ease of use is impacted by having to call `new PostgresDatabase(Config.usepostgres)` instead of `new PostgresDatabase()`.

test/lib/postgres.js
AnnikaCodes: You resolved this, but did not actually fix the bug.

AnnikaCodes:
Does this need setup on the postgres database server? I know have something for SEM logging, but I'm not sure which box it runs on or if it could/should be used for this (should we set up a postgres server on sim3 instead?). This can also be discussed in #private-dev or #admin on Discord if it's sensitive.

test/lib/postgres.js
mia-pi-git: This is done.

server/chat-commands/admin.ts
```diff
@@ -1154,17 +1154,43 @@ export const commands: Chat.ChatCommands = {
...
 		}
 
+		if (!noSave) {
+			this.sendReply('Saving battles...');
+			Rooms.global.lockdown = true; // we don't want more battles starting while we save
+			for (const u of Users.users.values()) {
+				u.send(
+					`|pm|&|${u.getIdentity()}|/raw <div class="broadcast-red"><b>The server is restarting soon.</b><br />` +
+					`While battles are being saved, no more can be started. <br />` +
+					`After the restart, you will be able to resume your battles from where you left off.`
```
AnnikaCodes, line 1186: This should probably mention something about how your battles may be frozen after saving but before the restart? Otherwise, I think the message about your battle being paused is going to confuse people.

lib/postgres.ts
```diff
@@ -0,0 +1,80 @@
...
+		return config;
+	}
+	async transaction(callback: (conn: PG.PoolClient) => any, depth = 0): Promise<any> {
+		const conn = await this.pool.connect();
+		await conn.query(`BEGIN`);
+		let result;
+		try {
+			// eslint-disable-next-line callback-return
+			result = await callback(conn);
+		} catch (e) {
```
AnnikaCodes, line 49:
```suggestion
		} catch (e: any) {
```
This goes for the other catch statements also.

lib/postgres.ts
mia-pi-git: Oh, forgot to reply to this. getConfig is the default, so yes it would be portable, since you you can pass in an object to override the default. This is really just a QoL.

lib/postgres.ts
mia-pi-git: Done.

server/chat-commands/admin.ts

AnnikaCodes (approved):
Looks good to me.

PR #8443, opened 2021-08-23
Rooms: Hide duplicated roomintros on join

mia-pi-git:
Requested, since roomintros can be spammy. This makes it so if you use !topic, people joining after have the roomintro under it  hidden to reduce clutter.


Zarel:
I don't think this is worth the complexity cost.

You also lose some features, such as the ability to see the roomintro before/after it was changed, which I assume is a lot of the use-case of `!topic` in the first place.

AnnikaCodes:
Is there a suggestion thread or other location where the popular support and reasoning for this feature is discussed?

PR #8444, opened 2021-08-23
Fix Water Shuriken interaction with Pokemon transformed into Ash-Greninja

pacmanboss256:
as referenced here https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8954294


DaWoblefet:
Implement the magical disappearing Pokemon next oo

Thanks pacmanboss!

PR #8445, opened 2021-08-25
Rename the "mystery" and "authentic" move flags

pacmanboss256:
This PR changes the 'Mystery' flag and the 'Authentic' flag so they better represent what the flag actually means

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -778,7 +778,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			if (!['mimikyu', 'mimikyutotem'].includes(target.species.id) || target.transformed) {
 				return;
 			}
-			const hitSub = target.volatiles['substitute'] && !move.flags['authentic'] && !(move.infiltrates && this.gen >= 6);
+			const hitSub = target.volatiles['substitute'] && !move.flags['bypasssub'] && !(move.infiltrates && this.gen >= 6);
 			if (hitSub) return;
 
 			if (!target.runImmunity(move.type)) return;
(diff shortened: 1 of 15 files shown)


DaWoblefet:
I think it'd be helpful to completely modify the flag name from mystery -> ``animatesAgainstAllies`` and update all references.

Minimally, we should also make it clear that we don't currently reference this flag for PS's own animations.

pacmanboss256:
I did this on my ipad so i had to do that manually. will fix later today

Zarel:
I like `infiltrates` assuming it's not very different from Infiltrator. I believe PSdex uses `bypasssub`.

KrisXV:
`infiltrates` is wrong, Infiltrator allows mons to hit around subs and screens, whereas the `authentic` flag just hits around substitute, so `bypasssub` would probably be better (also, `move.infiltrates` already exists for Infiltrator, so it'd cause some ambiguity imo).

Zarel:
(Drafting because of the build error.)

pacmanboss256:
i can make it `bypasssub`

EvGym:
I think it might be more clear to rename animonally -> allyanimonly
It is a small detail but it is easier to read for me.
Good work on fixing these flags!

KrisXV:
> I think it might be more clear to rename animonally -> allyanimonly
> It is a small detail but it is easier to read for me.
> Good work on fixing these flags!

its not ally-anim only though, so this would be wrong.

``allyanim`` might be better though

Zarel:
I'm fine with `allyanim`. Maybe `targetallyanim` would be clearer though?

KrisXV:
what does `targetallyanim` mean? I don't see how it's more clear

Zarel:
"target" clarifies that the flag controls whether or not the animation plays when an ally is targeted. There are a lot of ways animations could involve alllies (being used by an ally doesn't seem like an unusual interpretation) so I think clarification could be nice there.

Zarel:
ty!

PR #8449, opened 2021-08-30
Handle new TypeScript version

mia-pi-git:
This is a PR for the reason that the new TS changes how try-catch handles types. Previously, it would assume in the  `catch (x)` that `x` was any. Now, it assumes it's unknown. Likewise, it can only be cast to `unknown` or `any`.
I'd like to hear people's opinions on if we should:
a) Keep it as `unknown`, and handle case-by-case.
b) Cast to any, and leave all cases as it was before.
My current approach is of casting to `any`, and handling as we did prior. Please share your thoughts if you have an opinion.


server/chat-plugins/hangman.ts
```diff
@@ -37,7 +37,7 @@ try {
 	if (save) {
 		FS(HANGMAN_FILE).writeUpdate(() => JSON.stringify(hangmanData));
 	}
-} catch (e) {}
+} catch (e: any) {}
```
AnnikaCodes, line 40: Why does _this_ need a cast?

server/chat-plugins/helptickets.ts
```diff
@@ -793,7 +793,7 @@ export async function getBattleLog(battle: string): Promise<BattleInfo | null> {
 				url: `https://${Config.routes.replays}/${battle}`,
 			};
 		}
-	} catch (e) {}
+	} catch (e: any) {}
```
AnnikaCodes, line 796: Or this?

server/chat-plugins/hangman.ts
mia-pi-git: The general response to this is that I did a mass find and replace. Might be nice to keep it as a general convention?

server/chat-plugins/helptickets.ts
mia-pi-git: https://github.com/smogon/pokemon-showdown/pull/8449#discussion_r698140489

server/chat-plugins/hangman.ts
mia-pi-git: I pushed a commit to remove the types from all patterns of `catch (e) {}`.

lib/crashlogger.ts
```diff
@@ -51,7 +51,7 @@ export function crashlogger(
 		if (!transport) {
 			try {
 				require.resolve('nodemailer');
-			} catch (e) {
+			} catch (e: any) {
```
Slayer95, line 54: Might as well use [optional catch binding](https://mariusschulz.com/blog/optional-catch-binding-in-typescript).

Zarel:
I'm surprised this requires a PR. In theory there should be a config setting to make it always glocal.

Zarel:
Yeah, just disable `useUnknownInCatchVariables` in tsconfig.

Zarel:
Although I guess it's an open question if we'd prefer the setting or the explicit `any` type.

lib/crashlogger.ts
Zarel:
```suggestion
			} catch {
```

lib/crashlogger.ts
```diff
@@ -65,7 +65,7 @@ export function crashlogger(
 		} else {
 			try {
 				transport = require('nodemailer').createTransport(Config.crashguardemail.options);
-			} catch (e) {
+			} catch (e: any) {
```
Zarel, line 68:
```suggestion
			} catch {
```

lib/sql.ts
```diff
@@ -51,7 +51,7 @@ type ErrorHandler = (error: Error, data: DatabaseQuery) => void;
 function getModule() {
 	try {
 		return require('better-sqlite3') as typeof sqlite.default;
-	} catch (e) {
+	} catch (e: any) {
```
Zarel, line 54:
```suggestion
	} catch {
```

server/chat-commands/admin.ts
```diff
@@ -114,7 +114,7 @@ async function updateserver(context: Chat.CommandContext, codePath: string) {
 		}
 
 		return true;
-	} catch (e) {
+	} catch (e: any) {
```
Zarel, line 117:
```suggestion
	} catch {
```

server/chat-commands/info.ts
```diff
@@ -2475,7 +2475,7 @@ export const commands: Chat.ChatCommands = {
 		if (!/^(https?:\/\/)?(www\.)?(youtube\.com|youtu\.be)(\/|$)/i.test(link)) {
 			try {
 				dimensions = await Chat.fitImage(link);
-			} catch (e) {
+			} catch (e: any) {
```
Zarel, line 2478:
```suggestion
			} catch {
```

server/chat-commands/info.ts
```diff
@@ -2589,7 +2589,7 @@ export const commands: Chat.ChatCommands = {
 				const [width, height, resized] = await Chat.fitImage(link);
 				buf = Utils.html`<img src="${link}" width="${width}" height="${height}" />`;
 				if (resized) buf += Utils.html`<br /><a href="${link}" target="_blank">full-size image</a>`;
-			} catch (err) {
+			} catch (err: any) {
```
Zarel, line 2592:
```suggestion
			} catch {
```

server/chat-plugins/battlesearch.ts
```diff
@@ -47,7 +47,7 @@ export async function runBattleSearch(userids: ID[], month: string, tierid: ID,
 		let output;
 		try {
 			output = await ProcessManager.exec(['rg', '-i', regexString, '--no-line-number', '-P', '-tjson', ...files]);
-		} catch (error) {
+		} catch (error: any) {
```
Zarel, line 50:
```suggestion
		} catch {
```

server/chat-plugins/chatlog.ts
```diff
@@ -66,7 +66,7 @@ export class LogReaderRoom {
 		try {
 			const listing = await FS(`logs/chat/${this.roomid}`).readdir();
 			return listing.filter(file => /^[0-9][0-9][0-9][0-9]-[0-9][0-9]$/.test(file));
-		} catch (err) {
+		} catch (err: any) {
```
Zarel, line 69:
```suggestion
		} catch {
```

server/chat-plugins/chatlog.ts
```diff
@@ -75,7 +75,7 @@ export class LogReaderRoom {
 		try {
 			const listing = await FS(`logs/chat/${this.roomid}/${month}`).readdir();
 			return listing.filter(file => file.endsWith(".txt")).map(file => file.slice(0, -4));
-		} catch (err) {
+		} catch (err: any) {
```
Zarel, line 78:
```suggestion
		} catch {
```

server/chat-plugins/chatlog.ts
```diff
@@ -212,7 +212,7 @@ export const LogReader = new class {
 				const days = (await FS(`logs/${month}/${tier}/`).readdir()).filter(this.isDay).sort();
 				firstDay = days[0];
 				break;
-			} catch (err) {}
+			} catch (e) {}
```
Zarel, line 215:
```suggestion
			} catch {}
```

server/chat-plugins/chatlog.ts
```diff
@@ -225,7 +225,7 @@ export const LogReader = new class {
 				const days = (await FS(`logs/${month}/${tier}/`).readdir()).filter(this.isDay).sort();
 				lastDay = days[days.length - 1];
 				break;
-			} catch (err) {}
+			} catch (e) {}
```
Zarel, line 228:
```suggestion
			} catch {}
```

server/chat-plugins/chatlog.ts
```diff
@@ -242,7 +242,7 @@ export const LogReader = new class {
 				Utils.sortBy(battles, getBattleNum);
 
 				return [getBattleNum(battles[0]), getBattleNum(battles[battles.length - 1])];
-			} catch (err) {
+			} catch (err: any) {
```
Zarel, line 245:
```suggestion
			} catch {
```

server/chat-plugins/daily-spotlight.ts
```diff
@@ -50,7 +50,7 @@ export async function renderSpotlight(roomid: RoomID, key: string, index: number
 				const [width, height] = await Chat.fitImage(image, 150, 300);
 				imgHTML = `<td><img src="${image}" width="${width}" height="${height}" style="vertical-align:middle;"></td>`;
 				spotlights[roomid][key][index].image = [image, width, height];
-			} catch (err) {}
+			} catch (e) {}
```
Zarel, line 53:
```suggestion
			} catch {}
```

server/chat-plugins/daily-spotlight.ts
```diff
@@ -179,7 +179,7 @@ export const commands: Chat.ChatCommands = {
 			img = img.trim();
 			try {
 				[width, height] = await Chat.fitImage(img);
-			} catch (e) {
+			} catch (e: any) {
```
Zarel, line 182:
```suggestion
			} catch {
```

server/chat-plugins/helptickets.ts
```diff
@@ -766,7 +766,7 @@ export async function getOpponent(link: string, submitter: ID): Promise<string |
 			const body = await replayUrl.get();
 			const data = JSON.parse(body);
 			return data.p1id === submitter ? data.p1id : data.p2id;
-		} catch (e) {
+		} catch (e: any) {
```
Zarel, line 769:
```suggestion
		} catch {
```

mia-pi-git:
> Although I guess it's an open question if we'd prefer the setting or the explicit `any` type.

Yeah, that's why I made this a PR (plus Wob asked me to.) I didn't know about the setting, but I think it couldn't hurt to make it open to ask. (Setting seems better though.)

Zarel:
Fine by me.

I would recommend doing the optional `catch` binding (my suggestions, but I only did the first half). `crashlogger` should also be updated to take `unknown`, so we won't need it for anything that calls out to `crashlogger`.

Zarel:
Looks like eslint or typescript-eslint don't support optional catch binding. We'll probably need to upgrade those.

Zarel:
I tried to update to `@types/node@16` but apparently they do some weird `globalThis` stuff which I couldn't figure out in like an hour of googling, so back to `@types/node@14` we go.

Zarel:
Thanks!

PR #8450, opened 2021-08-30
Gen III-IV: Fix Uproar duration

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8959754

I just made the duration get randomly set in `onStart`. Is that correct or should I be using `effectState.time` like sleep and confusion do?


Marty-D:
Oh yeah, for future reference, `this.random()` isn't immediately intuitive with two arguments. It actually gives integers in [m,n), so `this.random(2, 5)` for example will pull 2 or 3 or 4. You'll want `this.random(2, 6)` for Gen 3 and `this.random(3, 7)` for Gen 4.

Marty-D (approved):
Thanks! 👍

PR #8451, opened 2021-08-31
Smogon master

DanielJSottile:


Zarel:
OK, this is the second time you've made this mistake, so I'm going to ban you from the repo in case it happens again.

This isn't a punishment. If you ever actually need to PR us, simply e-mail `staff@pokemonshowdown.com` and I'll let you back in.

PR #8453, opened 2021-09-04
Fix timing of freeze thaws

DaWoblefet:
When using a move like Scald or Sacred Fire, the check for burn potential happens before the check for potentially thawing the target; that is, it is impossible to burn a frozen target after thawing it with something like Scald. Technically it should be its own event, but I think ``onDamagingHit`` works alright for this purpose in the meantime until research is completed on post-move success order.

In addition, the autothaw from Fire-type attacks and the autothaw from Scald, Steam Eruption, and Scorching Sands is at different timings. A Sheer Force-boosted Flamethrower would thaw its target, but a Sheer Force-boosted Scald would not.

I need to verify how this works in pastgens. I would not be super surprised if you _could_ burn a frozen target in Gens prior to 5.

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -101,8 +101,13 @@ export const Conditions: {[k: string]: ConditionData} = {
 				pokemon.setStatus('');
 			}
 		},
-		onHit(target, source, move) {
-			if (move.thawsTarget || move.type === 'Fire' && move.category !== 'Status') {
+		onAfterMoveSecondary(target, source, move) {
+			if (move.thawsTarget) {
+				target.cureStatus();
+			}
+		},
+		onDamagingHit(damage, target, source, move) {
+			if (move.type === 'Fire' && move.category !== 'Status') {
 				target.cureStatus();
 			}
 		},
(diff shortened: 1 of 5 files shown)


DaWoblefet:
Doesn't seem like you can in Gen 1: https://github.com/pret/pokeyellow/blob/b4c97b7c01f564e2f2afb303645ad19aae532d60/engine/battle/effects.asm#L346. Did about 20 tries in Gen 1 with Fire Blast on a frozen target, never got a burn.

Gen 2 seems the same way. I tried about 10 Sacred Fires into a frozen target, no burns. I can't figure out the corresponding disassembly reference, though.

pacmanboss256:
https://github.com/pret/pokegold/blob/c91db56f422a6604d00a2df4e8c9f14879f70928/engine/battle/effect_commands.asm#L3946  

has some of the gen 2 mechanics

DaWoblefet:
Checked Gen 4; like Gen 3, Serene Grace Sacred Fire failed to burn a frozen target.

Gen 5 has completely identical behavior to Gen 8. I'm not going to bother testing 6, it's certainly the same.

DaWoblefet:
There probably are some overrides that need to be done here for Gen 3 and back but I will let @Marty-D take a look at those.

Marty-D (approved):
Looks good, thanks! 👍

PR #8454, opened 2021-09-05
Inheritance: Manually implement 2 Ability Clause

Karthik99999:
https://www.smogon.com/forums/threads/inheritance.3656811/post-8967769

Fixes 2 Ability Clause not working for Inheritance due to how abilities are stored prior to the battle starting. I just reimplemented the code from 2 Ability Clause into Inheritance's `onValidateTeam` function. Should I just check if the format is Inheritance in 2 Ability Clause's code and correct the bug there instead?

I also want to note that I left 2 Ability Clause in `rulesets` for the message it sends in `onBegin`.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -707,37 +707,38 @@ export const Formats: FormatList = [
 			return null;
 		},
 		onValidateTeam(team, f, teamHas) {
-			// 2 Ability Clause
-			const abilityTable = new Map<string, number>();
-			const base: {[k: string]: string} = {
-				airlock: 'cloudnine',
-				battlearmor: 'shellarmor',
-				clearbody: 'whitesmoke',
-				dazzling: 'queenlymajesty',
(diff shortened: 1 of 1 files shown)


config/formats.ts
```diff
@@ -707,6 +707,39 @@ export const Formats: FormatList = [
...
+				if (!ability) continue;
+				if (ability in base) ability = base[ability] as ID;
+				if ((abilityTable.get(ability) || 0) >= 2) {
+					return [
+						`You are limited to two of each ability by 2 Ability Clause.`,
+						`(You have more than two ${this.dex.abilities.get(ability).name} variants)`,
+					];
+				}
+				abilityTable.set(ability, (abilityTable.get(ability) || 0) + 1);
+			}
```
KrisXV, line 741: i would encompass this in an `if` statement to check that this is only applied if the ruleTable has `2abilityclause`

PR #8456, opened 2021-09-07
Gen 1: Fix minimum level encounters

PlagueVonKarma:
This was originally reported by [Beelzemon 2003](https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/post-8969689). 

This Pull Request adds various encounter and event slots in gen2/learnsets.ts that should aid in validating Pokemon for level-limited formats if used alongside #8455. There were multiple entries that did not have the minimum obtainable levels that thus unintentionally "soft-banned" Pokemon in some metagames, such as Vulpix and Wigglytuff. 

For some reason, L16 Golem was validated in international metagames, and I can't find any reason as to why. I don't think this Pull Request fixes that as it seems to be an issue beyond this, but I added the relevant L16 encounter anyway; if the problem is the lack of an "encounters" entry, then this should do the trick. I added the other Japan-only encounters of relevance with "japan: true" tacked on, which was how we handled the Gen 1-2 events. I don't know if the validator accounts for encounters and just handles events, though. 

I also added the relevant Stadium gifts under "eventData", as Amnesia Psyduck was handled this way. If this isn't how you want it, then it's probably easily moved to "encounters" anyway. We could probably be exhaustive and add Eevee + the starters but as far as I know they're literally just what you find in in-game RBY and have no effect on gameplay. 

Sorry if there's any problems!


data/mods/gen2/learnsets.ts
```diff
@@ -6019,6 +6028,9 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 			thunderpunch: ["2M", "2L26", "1L43"],
 			toxic: ["2M", "1M"],
 		},
+		eventData: [
+			{generation: 1, level: 20, moves: ["comet punch", "agility"]},
```
Zarel, line 6032:
```suggestion
			{generation: 1, level: 20, moves: ["cometpunch", "agility"]},
```

data/mods/gen2/learnsets.ts
```diff
@@ -8164,6 +8177,7 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 			withdraw: ["2L1", "2S0", "1L1"],
 		},
 		eventData: [
+			{generation: 1, level: 20, moves: ["water gun", "withdraw"]},
```
Zarel, line 8180:
```suggestion
			{generation: 1, level: 20, moves: ["watergun", "withdraw"]},
```

data/mods/gen2/learnsets.ts
```diff
@@ -5969,6 +5975,9 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 			thief: ["2M"],
 			toxic: ["2M", "1M"],
 		},
+		eventData: [
+			{generation: 1, level: 20, moves: ["double kick", "meditate"]},
```
Zarel, line 5979:
```suggestion
			{generation: 1, level: 20, moves: ["doublekick", "meditate"]},
```

Zarel:
The build errors are because these entries need to be IDs.

data/mods/gen2/learnsets.ts
```diff
@@ -8159,12 +8172,13 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
 			takedown: ["1M"],
 			thief: ["2M"],
 			toxic: ["2M", "1M"],
-			watergun: ["2L19", "1M", "1L1"],
+			watergun: ["2L19", "1M", "1L1", "1S0"],
```
Zarel, line 8175:
```suggestion
			watergun: ["2L19", "1M", "1L1", "1S1"],
```

data/mods/gen2/learnsets.ts
```diff
@@ -8159,12 +8172,13 @@ export const Learnsets: {[k: string]: ModdedLearnsetData} = {
...
 			whirlpool: ["2M"],
-			withdraw: ["2L1", "2S0", "1L1"],
+			withdraw: ["2L1", "2S0", "1L1", "1S0"],
```
Zarel, line 8177:
```suggestion
			withdraw: ["2L1", "2S0", "1L1", "1S1"],
```

data/mods/gen2/learnsets.ts
Zarel: The number after `S` is the event number.

Marty-D (approved):
Thanks!

PR #8457, opened 2021-09-07
Update Italian translations

DomeT99:
Hi, I have changed some incorrect translations **(Italian),** I am native speaker.


data/text/default.ts
```diff
@@ -105,7 +105,7 @@ export const DefaultText: {[k: string]: DefaultText} = {
 		clearBoost: "  [POKEMON]'s stat changes were removed!",
 		clearBoostFromZEffect: "  [POKEMON] returned its decreased stats to normal using its Z-Power!",
 		invertBoost: "  [POKEMON]'s stat changes were inverted!",
-		clearAllBoost: "  All stat changes were eliminated!",
+		clearAllBoost: "  All stats changes were eliminated!",
```
DaWoblefet, line 108: This won't be changed because it's text lifted directly from the game.

data/text/default.ts
```diff
@@ -146,7 +146,7 @@ export const DefaultText: {[k: string]: DefaultText} = {
 	},
 	spd: {
 		statName: "Sp. Def",
-		statShortName: "SpD",
+		statShortName: "SpDe",
```
DaWoblefet, line 149: The common abbreviation for Sp. Def has always been SpD, so that won't change either.

data/text/default.ts
DomeT99: Sorry

PR #8459, opened 2021-09-07
Add evolution condition for Lurantis

Karthik99999:
It only evolves during the day


DaWoblefet:
Lurantis is super annoying in VGC :(

Thanks Karthik!

PR #8460, opened 2021-09-07
Gen 1: Implement Stadium Tiers

PlagueVonKarma:
Talked to @KrisXV who said I just submit this tier file and he does the rest. This is based on the [2021 Stadium OU VR](https://www.smogon.com/forums/threads/stadium-ou-poster-viability-rankings-2021.3685877/), using Lutra's proposed cutoff between D4 and E rank. This aims to clean up the teambuilder a mite bit while being somewhat consistent with LGPE OU's stuff.


Marty-D (approved):
Thanks! 👍

PR #8461, opened 2021-09-08
Trim commands in /help

PartMan7:
```
[09:45:39] @PartMan: !h status
Could not find help for '/status ' - displaying help for '/status' instead
/status [note] - Sets a short note as your status, visible when users click your username.
Use /clearstatus to clear your status message.
```


AnnikaCodes (approved):
Looks good, thanks!

PR #8462, opened 2021-09-08
Refactor MaxGen

PartMan7:
a) Is no longer case-sensitive on ds/ms
b) Allows the use of spaces (ie; Max Gen 7)
c) Permits multi-digit generation numbers
d) Extends Gen syntax in is/as to Max Gen

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -97,9 +97,9 @@ export const commands: Chat.ChatCommands = {
 		const targetGen = parseInt(cmd[cmd.length - 1]);
 		if (targetGen) target += `, mod=gen${targetGen}`;
 		const split = target.split(',').map(term => term.trim());
-		const index = split.findIndex(x => x.startsWith('maxgen'));
+		const index = split.findIndex(x => /^max ?gen/.test(x.toLowerCase()));
 		if (index >= 0) {
-			const genNum = parseInt(split[index][split[index].length - 1]);
+			const genNum = parseInt(/\d*$/.exec(split[index]));
 			if (!isNaN(genNum) && !(genNum < 1 || genNum > Dex.gen)) {
 				split[index] = `mod=gen${genNum}`;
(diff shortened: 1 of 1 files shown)


PartMan7:
**Warning**: Changes untested since I can't get even a fresh clone to run on my device

PartMan7:
Uhh, is there any way to tell TS that /\d*$/ will never return null

AnnikaCodes:
You can use `!` to mark it as not null (`/\d*$/.exec(stuff)!`). However, I don't think a `RegExpExecArray` can be an argument to `parseInt`; `RegExp#exec` doesn't just return a single string for the match....

server/chat-plugins/datasearch.ts
```diff
@@ -97,9 +97,9 @@ export const commands: Chat.ChatCommands = {
 		const targetGen = parseInt(cmd[cmd.length - 1]);
 		if (targetGen) target += `, mod=gen${targetGen}`;
 		const split = target.split(',').map(term => term.trim());
-		const index = split.findIndex(x => x.startsWith('maxgen'));
+		const index = split.findIndex(x => /^max ?gen/.test(x.toLowerCase()));
```
AnnikaCodes, line 100:
```suggestion
		const index = split.findIndex(x => /^max\s?gen/i.test(x)));
```

PartMan7: Why \s? If someone inserts a hard tab in there, is it supposed to work?

AnnikaCodes: Why shouldn't it? Should it also break if there is an nbsp?

PartMan7: max&nbsp;gen #7 let's goo

(Lemme brush my teeth, will update)

PartMan7: Oh right GitHub doesn't escape those

PartMan7:
Bump

PR #8463, opened 2021-09-08
/details: Don't show G-Max moves in old gens

PartMan7:
Only YOU can stop people from seeing GMax moves in Gen 1 games!


Zarel:
I don't see how this does that? With this changes, G-Max moves would still appear, they just wouldn't be marked as G-Max in old gens, which seems worse.

Zarel:
Oh, wait, I thought this was a move entry. This is about removing the G-Max information from a pokemon entry. Nevermind, that makes sense.

PR #8464, opened 2021-09-08
Tournaments: Improve help

PartMan7:
The major commands are kept outside, the others are in a readmore block

Screenshot 2021-09-08 at 11-28-39 Showdown

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -2128,29 +2128,33 @@ const commands: Chat.ChatCommands = {
 	tournamenthelp() {
 		if (!this.runBroadcast()) return;
 		this.sendReplyBox(
-			`- create/new &lt;format>, &lt;type>, [ &lt;comma-separated arguments>]: Creates a new tournament in the current room.<br />` +
-			`- settype &lt;type> [, &lt;comma-separated arguments>]: Modifies the type of tournament after it's been created, but before it has started.<br />` +
-			`- cap/playercap &lt;cap>: Sets the player cap of the tournament before it has started.<br />` +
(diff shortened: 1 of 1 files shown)


server/tournaments/index.ts
```diff
@@ -2128,16 +2128,18 @@ const commands: Chat.ChatCommands = {
 	tournamenthelp() {
 		if (!this.runBroadcast()) return;
 		this.sendReplyBox(
+			`Tournament Commands<br/>` +
 			`- create/new &lt;format>, &lt;type>, [ &lt;comma-separated arguments>]: Creates a new tournament in the current room.<br />` +
+			`- rules/banlist &lt;comma-separated arguments>: Sets the custom rules for the tournament before it has started.<br />` +
+			`- end/stop/delete: Forcibly ends the tournament in the current room.<br />` +
+			`- begin/start: Starts the tournament in the current room.<br /><br />` +
+			`<details class="readmore"><summary>More Commands</summary>` +
```
KrisXV, line 2136: I would make this more specific, with something like 'Tournament Configuration Commands' or smth

PartMan7: Oh I didn't get a notif for this, sorry

Zarel:
"here" should not be link text... Design standard D2...

https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#d2-buttons-should-say-what-they-do

PartMan7:
Amusingly enough, I did point that out in Dev earlier, but since it was the status quo I didn't change it. Can change if you'd like.

DaWoblefet:
> Amusingly enough, I did point that out in Dev earlier, but since it was the status quo I didn't change it. Can change if you'd like.

Please do; something like ``<a href="the link">More detailed tournament help</a>`` is good.

PR #8467, opened 2021-09-11
Gen III-IV: Fix interaction between Liquid Ooze and Dream Eater

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8972096

I used `target.lastMoveUsed` to make this work, I feel like there's a better option.

### DIFF
--- a/test/sim/abilities/liquidooze.js
+++ b/test/sim/abilities/liquidooze.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const assert = require('./../../assert');
+const common = require('./../../common');
+
+let battle;
+
+describe('Liquid Ooze', function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it('should damage the target by the amount it would have healed after it uses a draining move', function () {
+		battle = common.createBattle([[
+			{species: 'tentacruel', ability: 'liquidooze', moves: ['sleeptalk']},
+		], [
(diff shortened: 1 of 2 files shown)


test/sim/abilities/liquidooze.js
```diff
@@ -0,0 +1,73 @@
...
+
+	it('should damage the target by the amount it would have healed after it uses a draining move', function () {
+		battle = common.createBattle([[
+			{species: 'tentacruel', ability: 'liquidooze', moves: ['sleeptalk']},
+		], [
+			{species: 'serperior', moves: ['gigadrain']},
+		]]);
+		battle.makeChoices();
+		const damage = battle.p2.active[0].maxhp - battle.p2.active[0].hp;
+		assert.bounded(damage, [27, 33]);
```
DaWoblefet, line 21: For each of these tests, I think it'd be better just to do ``assert.fullHP()`` or ``assert.false.fullHP()``; we don't really care about the particular damage the target is at for any of these, just that it was damaged.

DaWoblefet (approved):
Looks fine to me. I'll see what Marty thinks.

Marty-D (approved):
`this.activeMove` is usually better for things that happen during an attack segment. In this case for example, checking the last move could cause Leech Seed not to work properly.

Other than that this is nice, thanks!

PR #8468, opened 2021-09-11
Fix Powder revealing the target of a failed Fire-type move

Karthik99999:


DaWoblefet:
Thanks Karthik! Good find.

PR #8474, opened 2021-09-19
Add mods for old VGC formats

Karthik99999:
~~I've added mods for VGC 11/12, VGC 14/15 will be done in this PR as well at some point~~

Check my comment below for more info


KrisXV:
Whats the point of a vgc 11 mod when its just vgc 12 + 5 unreleased items, you could theoretically just ban those?

Karthik99999:
@KrisXV 

Some notes:
- The following years have been added as new challenge-only formats:

| Year | Mod     |
|------|---------|
| 2009 | gen4pt  |
| 2010 | gen4    |
| 2011 | gen5bw1 |
| 2012 | gen5bw1  |
| 2014 | gen6xy  |

- We named the mods `gen4pt`, `gen5bw1` and `gen6xy`, which differs from the current mod in use for vgc 2017 (`vgc17`)
- There are some event moves that weren't released in 2011, but we considered them insignificant enough to not create a mod for 2011

DaWoblefet (approved):
I wouldn't be surprised if there's a few things we missed here or there, but Karthik did a great job with this. I'm excited to have proper oldgen VGC support!

data/mods/gen6xy/pokedex.ts
```diff
@@ -0,0 +1,42 @@
...
+		unreleasedHidden: true,
+	},
+	amaura: {
+		inherit: true,
+		unreleasedHidden: true,
+	},
+	aurorus: {
+		inherit: true,
+		unreleasedHidden: true,
+	},
```
KrisXV, line 41: Hidden ability gen 5 starters weren't released until ORAS came out, so they should probably be marked as unreleased in here

data/mods/gen6xy/moves.ts
```diff
@@ -0,0 +1,18 @@
...
+		isNonstandard: "Future",
+	},
+	originpulse: {
+		inherit: true,
+		isNonstandard: "Future",
+	},
+	precipiceblades: {
+		inherit: true,
+		isNonstandard: "Future",
+	},
```
KrisXV, line 17: all 4 of the pokemon (and their items/abilities in the case of the primals/mray) affiliated with these moves should be marked as Future (or unreleased in hoopa-confined's case, unbound would be Future/Nonexistent)

KrisXV (changes requested):
I didn't look over all of the details for the `gen6xy` mod and didn't go over the `gen5bw1` mod at all, but a lot of the stuff seems to be missing. A good way to think of gen6xy and gen5bw1 in the scope of PS is "Gen 5 before BW2 existed" and "Gen 6 before ORAS existed", or even "Gen 4 before HGSS existed" in the case of `gen4pt`.

data/mods/gen6xy/moves.ts
```diff
@@ -0,0 +1,18 @@
...
+};
```
Marty-D, line 18: I believe Thousand Arrows/Waves need to be here as Unobtainable so Smeargle can't use them?

Karthik99999: Both moves are already marked as `Unobtainable` in the base gen 6 mod

Marty-D (approved):
Looks good, thanks Karthik! 👍

PR #8478, opened 2021-09-22
Fix paralysis modifier application

DaWoblefet:
Currently, paralysis is being treated as a Speed modifier, being chained together with other Speed modifiers like Tailwind or Choice Scarf. Paralysis is actually its own independent modifier, taking place outside of that traditional calculation; this can lead to some rounding errors such as outlined in this [bug report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8979399).

This solution works for modern gens fine, but it doesn't work for past gens and mods since I'm hardcoding paralysis. I also don't in fact know for sure how paralysis is applied prior to Gen 6. In Gens 7 & 8, it's applied as a * 50/100 after other modifiers, and in Gen 6 it's the same but * 25 / 100. I assume BW is the same as Gen 6. But I don't know if it's done differently prior to then.

The other approach I thought of was creating a new event, ``onApplyParalysis`` or something, to replace the ``onModifySpe`` currently used for paralysis. But it seems kinda dumb to use a whole event for just one thing. This approach more closely models how we handle burn.

Opinions are appreciated.

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -27,7 +27,14 @@ export const Conditions: {[k: string]: ConditionData} = {
 				this.add('-status', target, 'par');
 			}
 		},
-		// Speed reduction is handled directly in the sim/pokemon.ts getStat function
+		onModifySpe(spe, pokemon) {
+			// Paralysis occurs after all other Speed modifiers, so evaluate all modifiers up to this point first
+			spe = this.finalModify(spe);
+			if (!pokemon.hasAbility('quickfeet')) {
+				spe = Math.floor(spe * 50 / 100);
+			}
+			return spe;
+		},
 		onBeforeMovePriority: 1,
 		onBeforeMove(pokemon) {
 			if (this.randomChance(1, 4)) {
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -27,11 +27,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 				this.add('-status', target, 'par');
 			}
 		},
-		onModifySpe(spe, pokemon) {
-			if (!pokemon.hasAbility('quickfeet')) {
-				return this.chainModify(0.5);
-			}
-		},
+		// Speed reduction is handled directly in the sim/pokemon.ts getStat function
 		onBeforeMovePriority: 1,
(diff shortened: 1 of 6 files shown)


Slayer95:
The ``ModifySpe`` event handler may be kept so long as the event modifier is directly manipulated (as [already done in Gen 4 Beat Up](https://github.com/smogon/pokemon-showdown/blob/4f791ef85277927c4c8b0fb6b81c53f9144f1793/data/mods/gen4/moves.ts#L95-L97))

```js
// sim/battle.ts - Battle#finalModify
finalModify(relayVar) {
	relayVar = this.modify(relayVar, this.event.modifier);
	this.event.modifier = 1;
	return relayVar;
}

// data/conditions.ts
par: {
	inherit: true,
	onModifySpePriority: -101,
	onModifySpe(spe, pokemon) {
		spe = this.finalModify(spe);
		if (!pokemon.hasAbility('quickfeet')) {
			spe = this.modify(spe, 0.25);
		}
		return spe;
	},
}
```

DaWoblefet:
> The `ModifySpe` event handler may be kept so long as the event modifier is directly manipulated (as [already done in Gen 4 Beat Up](https://github.com/smogon/pokemon-showdown/blob/4f791ef85277927c4c8b0fb6b81c53f9144f1793/data/mods/gen4/moves.ts#L95-L97))

This worked great! Excellent suggestion. These lets me leave mods alone and they'll inherit as they please.

@Marty-D do you know about when paralysis Speed reduction is applied in pastgens?

DaWoblefet:
Bumping for your review in pastgens @Marty-D

DaWoblefet:
I started work on this and have since come to appreciate how challenging this would be, as I don't know for each generation where the game rounds or floors. I'll assume it's not using the Gen 5 versions, so it'd be way out of scope for what I wanted to accomplish in this PR. Sorry for the delay on this @Marty-D.

Marty-D (approved):
No worries, I forgot to mention I didn't expect you to deal with past gens here, and just an override for now would be fine. But you've come to that conclusion yourself so 👍 .
Thanks!

PR #8479, opened 2021-09-23
Fix Ice Face behavior

im-tofa:
Ice Face should not reform when neutralizing gas leaves the field. Caused by triggering onStart for Ice Face in onEnd of Neutralizing Gas.

To fix the bug for Ice Face, the code now ignores abilities that are unaffected by Neutralizing Gas in the onEnd for Neutralizing Gas (using the isPermanent flag). 

I believe this also excludes other abilities unaffected by Neutralizing Gas (such as Schooling) from triggering onStart immediately when Neutralizing Gas leaves the field. Not sure if this is correct from a mechanics perspective; if not, I can change it to only exclude Ice Face! 

Corresponding bug report: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8949321

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2381,6 +2381,8 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			this.speedSort(sortedActive);
 			for (const pokemon of sortedActive) {
 				if (pokemon !== source) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Looks good! Thank you very much 👍

PR #8480, opened 2021-09-24
Rooms: Stop requiring connections to upload replays

mia-pi-git:
Fixes the bug where users get a popup after saving a replay mid-battle saying that they are not in the room. Should generally make this easier to use.


mia-pi-git:
(this is mostly a PR in case there's a reason we upload from the connection I'm not aware of)

mia-pi-git:
OK. Can we make that a comment, then?

PR #8487, opened 2021-10-04
Learnsets: Remove Covet from Toxicroak

Karthik99999:
multiple sources have toxicroak not learning the move


Marty-D (approved):
Wow what's the deal with all these Toxicroak errors... Thanks! 👍

PR #8488, opened 2021-10-05
Improve datasearch

mia-pi-git:
This should make /ds (and related commands) both a lot faster and a lot easier to extend. It abstracts all parameters into a single format (key=searchvalue) and then runs functions to validate / search, etc. As a result, it'll be a lot easier to work on this file since one will be able to just add a function to do what they want in the search.
This, however, does take away the specialized custom handling for things like `stat < num`, etc, since that was what made /ds so difficult to work with and nuked its performance.
I'm drafting this as I only wrote the infrastructure and the design - @KrisXV is rewriting most of the actual handlers.


DaWoblefet:
I've mentioned this before in the past, but I think we'd get a _lot_ of pushback for removing the ability to use ``>`` and ``<`` operators. I don't have a solution myself that makes datasearch easier to extend while preserving these elements.

mia-pi-git:
Yeah, I can see that. The problem is it makes dexsearch extremely hostile to work on. 
...i... guess we could... special-case it...?

mia-pi-git:
I guess we could add custom parsers that just run on arguments that don't match the `key=value` format?

mia-pi-git:
> Just as a warning, I'm also working on a dexsearch rewrite...

Is there any chance it could wait until after we're done with this? This is a rather large rewrite.

mia-pi-git:
OK. If that's the case then I'll just wait for yours instead.

PR #8489, opened 2021-10-07
Export sim/tools for AI player tool

joshbwlng:
Change-type: minor
Signed-off-by: Josh Bowling <josh@balena.io>

### DIFF
--- a/sim/index.ts
+++ b/sim/index.ts
@@ -27,4 +27,5 @@ export {TeamValidator} from './team-validator';
 
 // misc libraries
 
+export * from './tools';
 export * from '../lib';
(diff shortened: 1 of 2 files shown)


Zarel:
Was this a mistake? I don't think we would have minded this...

joshbwlng:
@Zarel Apologies for the late response, but yes I believe I created this PR by accident! I was playing around with the library and think I needed to export these tools to be used by an external consumer test module I was working on. Not working on this any more, but can re-open the PR if you think its useful. :+1:

Zarel:
Nah, no worries. I would probably have done it a different way.

PR #8492, opened 2021-10-09
Support banning specific Gigantamax formes

Karthik99999:


Zarel:
(Although maybe the Gmax should just be the tier species...)

Karthik99999:
> (Although maybe the Gmax should just be the tier species...)

This was my original approach, but I didn't do that because of this being the `checkSpecies` function

```js
if (tierSpecies.isNonstandard === 'Gigantamax') {
	return `${tierSpecies.name} is a placeholder for a Gigantamax sprite, not a real Pokémon. (This message is likely to be a validator bug.)`;
}
```

PR #8493, opened 2021-10-09
Update Passimian's Ability in Battle Factory

sexy90gxebattlefactoryplayer:
Passimian received Defiant as an ability from an event late in Gen 7 (https://twitter.com/SerebiiNet/status/1111223652178112513?s=19). Defiant is objectively better than Receiver within the context of singles, so all Passimian sets should be Defiant. Apologies if there's a more proper way of doing this - I reported this error to the Randbats Auth 3 weeks ago and didn't see it change so I decided to be the change I wanted to see in the world.


AnnikaCodes:
Sure.

PR #8496, opened 2021-10-12
Remove "any" type from EffectState#duration

Karthik99999:


Karthik99999:
> I appreciate the enthusiasm, but probably run unit tests after doing changes like this. :P

~~I use GitHub actions to run my unit tests for me~~

Yeah I probably should. I'll fix those errors.

PR #8507, opened 2021-10-28
Gen IV: Fix client messages for spread move miss

Karthik99999:
https://replay.pokemonshowdown.com/gen4vgc2010-1444280291-wzq9jvouu0ez9k6ws7pk8d4i66rrsbxpw - replay showing the bug just in case it's needed (don't skip turns)

EDIT: found smogon bug report for this https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8930317


Zarel:
This should also be fixed on client, ideally.

PR #8508, opened 2021-10-31
stop allowAVs from removing items

lgpe:
Let's Go formats already disallow any non-Mega Stone items, this code has no purpose other than preventing the use of Mega Stones (not an intended part of Candy formats).


lgpe:
missed a bracket, fixed

Zarel (approved):
Looks good to me, unless I'm missing something. @DaWoblefet?

KrisXV:
I'm not DaWoblefet but this is correct, thanks @lgpe !!

PR #8511, opened 2021-11-02
Fix Torment duration from G-Max Meltdown

KrisXV:


data/moves.ts
```diff
@@ -6511,7 +6511,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		self: {
 			onHit(source) {
 				for (const pokemon of source.foes()) {
-					if (!pokemon.volatiles['dynamax']) pokemon.addVolatile('torment');
+					if (!pokemon.volatiles['dynamax']) pokemon.addVolatile('torment', source, this.dex.getActiveMove('gmaxmeltdown'));
```
urkerab, line 6514: Doesn't `addVolatile` already default to the current effect here? Even if it doesn't, it might be better to grab the move from the third argument to `onHit` instead of creating a new copy.

data/moves.ts
```diff
@@ -18312,7 +18312,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		volatileStatus: 'torment',
 		condition: {
 			noCopy: true,
-			onStart(pokemon) {
+			onStart(pokemon, source, effect) {
+				if (effect?.id === 'gmaxmeltdown') this.effectState.duration = 3;
```
urkerab, line 18316: This could go after after the dynamax check perhaps?

data/moves.ts
KrisXV: I’m just doing it like how the other gmax moves do it, if it’s better to do it the other way then fixing the others should be a priority

data/moves.ts
KrisXV: It all works the same bye I guess I can, I’ll do it when I wake up

data/moves.ts
urkerab: G-Max Finale does already do it the way I suggested but I'd say the other G-Max moves could be improved at some point. (I checked regular moves and the only other times we use `getActiveMove` are either from inside a condition or in certain specific cases where we need a fresh move.)

KrisXV: Yeah, after this is handled I'll open another PR to fix them unless someone beats me to it

DaWoblefet (approved):
Looks good to me!

KrisXV:
ill merge it then

PR #8512, opened 2021-11-02
CRQ: Don't send friend status for yourself in userdetails

singiamtel:
You can't befriend yourself, so there's some bandwidth to be saved by not sending that info back from server

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -394,7 +394,7 @@ export const crqHandlers: {[k: string]: Chat.CRQHandler} = {
 			autoconfirmed: !!targetUser.autoconfirmed,
 			status: targetUser.getStatus(),
 			rooms: roomList,
-			friended: user.friends?.has(targetUser.id),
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Looks good, thanks!

server/chat-commands/core.ts
```diff
@@ -394,7 +394,7 @@ export const crqHandlers: {[k: string]: Chat.CRQHandler} = {
 			autoconfirmed: !!targetUser.autoconfirmed,
 			status: targetUser.getStatus(),
 			rooms: roomList,
-			friended: user.friends?.has(targetUser.id),
+			friended: user.id === targetUser.id ? undefined : user.friends?.has(targetUser.id),
```
Zarel, line 397:
```ts
friended: user.friends?.has(targetUser.id) || undefined,
```

would do the same thing for all people who aren't friends, for more bandwidth savings.

Zarel: Same goes for `status` and `autoconfirmed`.

mia-pi-git: I did this.

PR #8516, opened 2021-11-03
Gen 1 Randbats: Fix Monotype

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9021489

Fixes the above bug, where Magikarp appeared on a forced mono bug team because shitmons get pushed into the rejected but not invalid pool before typing gets checked for monotype. I also cleaned up the code a bit by using stuff like `getPokemonPool` in `randomTeam` and `randomNPokemon` in `randomCCTeam`. If I shouldn't have done that I can revert those changes and just do a basic fix for the above bug.


AnnikaCodes (approved):

AnnikaCodes:
LGTM, thanks!

PR #8518, opened 2021-11-05
Chatlog: Split into search/view files and refactor searchlog

mia-pi-git:
This PR does a few things.
- Adds `ripgrep-bin` as an optional dependency (for easy installation/use of ripgrep on any platform) 
- Splits `server/chat-plugins/chatlog.ts` into two files: `chatlog/view.ts` (handles viewing logs directly, and stats) and `chatlog/search.ts` (handles /searchlog).
- Rewrites /searchlog from the ground up - I rewrote it eyeballing performance, usability, and readability. The new code is shorter and much cleaner - should be faster as well. I also added support for ignoring terms with the syntax `!term`.
- Removes FS support for searching. With the new dependency, ripgrep is much easier to install, so there's no need for this code, which is now just a maintenance burden.


Zarel:
I'd like for this to be tested in Windows before merging, to make sure calling into ripgrep works correctly.

Zarel:
I'd like for this to be tested in Windows before merging, to make sure calling into ripgrep works correctly.

lib/ripgrep.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * An easy wrapper around ripgrep. Mostly to make install easier.
+ * @author mia-pi-git
+ */
+
+import * as Streams from './streams';
+import * as child_process from 'child_process';
+import {FS} from './fs';
+
+const ROOT = FS('./').path;
```
Zarel, line 10: 😬

I don't know if this is supposed to stay stable like this...

Zarel: You only use it as a default `options.cwd`; I think it's fine just to grab it from FS every time.

lib/ripgrep.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export function hasRipgrep() {
+	try {
+		return require('ripgrep-bin') as string;
+	} catch {
+		return false;
```
Zarel, line 16:
```suggestion
		return null;
```

mia-pi-git: Done.

lib/ripgrep.ts
mia-pi-git: Alright.

Zarel:
Wait, how are folders in chat-plugins loaded? Presumably they should only load `folder/index.ts`.

With your current layout, I think it might be easier as `chat-plugins/chatlog.ts` and `chat-plugins/chatlog-search.ts`

AnnikaCodes:
They load all files in the folders; private code needs this.

mia-pi-git:
> Wait, how are folders in chat-plugins loaded? Presumably they should only load `folder/index.ts`.
> 
> With your current layout, I think it might be easier as `chat-plugins/chatlog.ts` and `chat-plugins/chatlog-search.ts`

Yeah, this loads fine. The entries in Chat.plugins are just `search` and `view`.

mia-pi-git:
> I'd like for this to be tested in Windows before merging, to make sure calling into ripgrep works correctly.

I tested with DaWoblefet (who's on Windows), and it calls fine. Installing is a bit thorny, so we may want to toss the dep and just have users install if they want it.

mia-pi-git:
New dep works on Windows.

package.json
```diff
@@ -19,7 +19,8 @@
     "permessage-deflate": "^0.1.7",
     "sql-template-strings": "^2.2.2",
     "sqlite": "^3.0.6",
-    "sucrase": "^3.15.0"
+    "sucrase": "^3.15.0",
+    "ripgrep-bin": "^13.0.0"
```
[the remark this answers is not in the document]
mia-pi-git: That's an outdated diff - this uses `ripgrep-pkg` now, which both offers a better API and installs a ripgrep binary through NPM.

server/chat-plugins/chatlog/search.ts
```diff
@@ -0,0 +1,278 @@
...
+		if (includes.length) {
+			searchString += `matching the term${includes.length > 1 ? 's' : ''} '${includes.join("', '")}'`;
+		}
+		if (excludes.length) {
+			if (includes.length) searchString += ` and `;
+			searchString += `not matching the term${excludes.length > 1 ? 's' : ''} '${excludes.join("', '")}'`;
+		}
+		searchString += `:`;
+
+		this.title = `[Log Search] [${search.room}] ${includes} | ${excludes}`;
```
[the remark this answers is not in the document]
mia-pi-git: OK.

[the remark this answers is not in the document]
mia-pi-git: I'll just put the full query in instead, so people can still see which is which.

mia-pi-git:
@Zarel bump?

AnnikaCodes:
Please resolve merge conflicts.

mia-pi-git:
I fixed the conflicts.

mia-pi-git:
RIpgrep is not sufficient or desired anymore. This will require a more in-depth overhaul.

PR #8520, opened 2021-11-06
Dex: Remove unobtainable Gen IV Abilities from Gen III

DaWoblefet:
As [reported on the bug reports thread](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9023066), any Gen III Pokemon which gained a new Ability in Gen IV was displaying as if it was available in Gen III. I figure it's more appropriate to override the data here than to filter it out client-side on the teambuilder itself.


DaWoblefet:
hmm, I'm not sure why tests are failing - the exhaustive runner seems to have generated a species with an empty string:

``{"name":"","species":"","gender":"","item":"ultraball","ability":"hypercutter","moves":["psychic","strength","whirlpool","futuresight"],"evs":{"hp":170,"atk":129,"def":231,"spa":121,"spd":56,"spe":217},"ivs":{"hp":29,"atk":2,"def":9,"spa":31,"spd":6,"spe":18},"nature":"jolly","level":74,"happiness":29,"shiny":false}]}``

I must have typoed somewhere or something.

Marty-D:
I believe there was a reason we don't do it this way, but I'll have to look through some old logs since I don't remember the specifics.

DaWoblefet:
> I believe there was a reason we don't do it this way, but I'll have to look through some old logs since I don't remember the specifics.

One of the things I was worried about was the validator. On this implementation, Magic Guard Clefable can still learn Soft-Boiled and Counter in Gen 4 OU, and Forewarn Hypno can still learn Wish in Gen 4 OU. Whether or not they _should_ be able to is another question, but I don't think it causes any regressions in that area at least.

Zarel:
I definitely preferred to do this in `Dex.species.get` rather than needing a huge override table. This is I think mostly bandwidth but partially maintainability: it lets the files themselves focus on data that there isn't a simple easily-memorizable rules for. Like, when looking at this file, it's non-obvious to me that the only changes between gen 3 and gen 4 is that gen 4 Abilities were added to a some pokemon that only had one ability. But if it were automated, you could tell that at a glance.

DaWoblefet:
Makes sense - I'll explore a route similar to whatever we do so Hidden Abilities don't show up in Gens 4 and prior.

DaWoblefet:
This seems a lot better now.

image

sim/dex-species.ts
```diff
@@ -477,6 +477,7 @@ export class DexSpecies {
 				(!['Ditto', 'Undiscovered'].includes(species.eggGroups[0]) && !species.prevo && species.name !== 'Manaphy');
 			if (this.dex.gen === 1) species.bst -= species.baseStats.spd;
 			if (this.dex.gen < 5) delete species.abilities['H'];
+			if (this.dex.abilities.get(species.abilities['1']).gen > this.dex.gen) delete species.abilities['1'];
```
Zarel, line 480:
```suggestion
			if (this.dex.gen === 3 && this.dex.abilities.get(species.abilities['1']).gen > this.dex.gen) delete species.abilities['1'];
```

Zarel:
Yeah, the narrower scope is there for the same reason to make it easier to tell what it's for at a glance.

Zarel:
Anyway thanks!

PR #8524, opened 2021-11-09
Modlog: Extend punishment types with missing entries

FeluciaPS:
I added them to PUNISHMENTS rather than GLOBAL_PUNISHMENTS because that seemed the most internally consistent. I have no way to verify I haven't missed any punishment types so I'd appreciate if someone could extend these


server/modlog/index.ts
```diff
@@ -31,7 +31,8 @@ const PUNISHMENTS = [
 	'UNRANGEBAN', 'TRUSTUSER', 'UNTRUSTUSER', 'BLACKLIST', 'BATTLEBAN', 'UNBATTLEBAN',
 	'NAMEBLACKLIST', 'KICKBATTLE', 'UNTICKETBAN', 'HIDETEXT', 'HIDEALTSTEXT', 'REDIRECT',
 	'NOTE', 'MAFIAHOSTBAN', 'MAFIAUNHOSTBAN', 'GIVEAWAYBAN', 'GIVEAWAYUNBAN',
-	'TOUR BAN', 'TOUR UNBAN', 'UNNAMELOCK',
+	'TOUR BAN', 'TOUR UNBAN', 'UNNAMELOCK', 'FORCEWEEKLOCK', 'FORCELOCK', 'FORCEMONTHLOCK',
+	'FORCERENAME OFFLINE',
```
AnnikaCodes, line 35: `GLOBAL_PUNISHMENTS` are those punishments that count towards a user's autolocks. All of these forced punishments should probably go in that category, unless there's policy reasoning to convince me otherwise.

FeluciaPS: Thanks for clarifying, I'll move some things around tonight

FeluciaPS: Does this look right?

AnnikaCodes:
Yes; should tour bans still be handled?

AnnikaCodes:
@FeluciaPS ?

AnnikaCodes:
Thanks!

PR #8525, opened 2021-11-10
Force Prankster on Liepard set with Copycat

sexy90gxebattlefactoryplayer:
On Liepard sets without Copycat, Prankster is useless and Limber is a situational improvement. However, Prankster is needed on Copycat sets to function properly (revenging sun boosted victreebel, for example)


Zarel (approved):
Looks like a good change to me.

AnnikaCodes:
LGTM, thanks!

PR #8528, opened 2021-11-13
Refactor getIdentity not to take roomids

Zarel:
This is a PR basically only because it's not hotpatchable, so I want people to know about it. You _could_ monkeypatch it, but I'm planning to restart tonight, so I'm going to skip that.

This is partly to work around unnecessary "room not found" issues from race conditions, but this should also improve performance.

### DIFF
--- a/server/users.ts
+++ b/server/users.ts
@@ -508,17 +508,13 @@ export class User extends Chat.MessageContext {
 	popup(message: string) {
 		this.send(`|popup|` + message.replace(/\n/g, '||'));
 	}
-	getIdentity(roomid: RoomID = '') {
+	getIdentity(room: BasicRoom | null = null) {
 		const punishgroups = Config.punishgroups || {locked: null, muted: null};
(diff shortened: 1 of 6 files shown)


server/users.ts
```diff
@@ -508,17 +508,13 @@ export class User extends Chat.MessageContext {
...
+	getIdentity(room: BasicRoom | null = null) {
 		const punishgroups = Config.punishgroups || {locked: null, muted: null};
 		if (this.locked || this.namelocked) {
 			const lockedSymbol = (punishgroups.locked && punishgroups.locked.symbol || '\u203d');
 			return lockedSymbol + this.name;
 		}
-		if (roomid) {
-			const room = Rooms.get(roomid);
-			if (!room) {
-				throw new Error(`Room doesn't exist: ${roomid}`);
```
Zarel, line 520: The removal of this should save us tons of unnecessary crashes...

PR #8529, opened 2021-11-15
Chat: Add an onConfigRefresh handler

mia-pi-git:
Takes hardcodes out of the main index file, plus lets other plugins hook into config. This is only a PR because it needs a restart.

### DIFF
--- a/server/chat-plugins/username-prefixes.ts
+++ b/server/chat-plugins/username-prefixes.ts
@@ -126,3 +126,11 @@ export const commands: Chat.ChatCommands = {
 		);
 	},
 };
+
+export const handlers: Chat.Handlers = {
+	onConfigRefresh() {
+		// ensure that battle prefixes configured via the chat plugin are not overwritten
+		// by battle prefixes manually specified in config.js
+		prefixManager.refreshConfig(true);
+	},
+};
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -71,6 +71,7 @@ export interface Handlers {
 	onBattleStart?: (user: User, room: GameRoom) => void;
 	onBattleLeave?: (user: User, room: GameRoom) => void;
 	onDisconnect?: (user: User) => void;
+	onConfigRefresh?: () => void;
 }
 
 export interface ChatPlugin {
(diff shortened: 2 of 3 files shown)


Zarel:
I'm going to need way more details on what this is for.

Zarel:
Oh, I think I understand.

Hm, should chat plugins be hooking into Config at all? In theory chat plugin configuration is handled inside every chat plugin. We might need to rearchitect how configuration works at all, if chat plugins have need for it.

mia-pi-git:
> Oh, I think I understand.
> 
> Hm, should chat plugins be hooking into Config at all? In theory chat plugin configuration is handled inside every chat plugin. We might need to rearchitect how configuration works at all, if chat plugins have need for it.

It's not to configure a plugin tbc; the plugin exists to help non-sim3 admins modify a relevant part of the config.

server/chat-plugins/username-prefixes.ts
```diff
@@ -126,3 +126,11 @@ export const commands: Chat.ChatCommands = {
 		);
 	},
 };
+
+export const handlers: Chat.Handlers = {
+	onConfigRefresh() {
+		// ensure that battle prefixes configured via the chat plugin are not overwritten
+		// by battle prefixes manually specified in config.js
```
Zarel, line 133: What battle prefixes are manually specified in config.js? If they're related, maybe it should be stored in the same place the chat plugin's own data is?

Zarel:
If people without server access want to modify config, that data should probably be moved out of `config.js`.

Zarel (changes requested):
I think a single-line hardcode is better than an entire system that should never be used.

AnnikaCodes:
Battle prefixes used to be stored in `config.js` - we can probably remove the functionality to load from config.js (or at least deprecate it and remove this hook), since most battle prefixes are now stored in a JSON configuration file.

Zarel:
> I agree that chat plugins shouldn't modify config.js, but that's not what's happening here.

Yeah, I was confused by that, but I was just going off Mia saying "plugin exists to help non-sim3 admins modify a relevant part of the config".

My main issue is still that I think any setting that plugins would want to listen to should be kept out of config.js.

...We should probably move custom avatars out of config.js, incidentally.

AnnikaCodes:
Is this still necessary? Battle prefixes aren't in Config anymore, and I agree with Zarel that settings that need to be changed by people without access to sim3 should be in separate files anyway.

mia-pi-git:
Yeah, it's not.

PR #8532, opened 2021-11-19
Fix BDSP interaction with STABmons Move Legality

WeWuzNidokangz:
BDSP can't import Pokemon from Home, so it shouldn't grant moves from archaic species types here, e.g. Clefable should not be able to learn Tackle via STABmons Move Legality in these formats because it could never have been a Normal type.


KrisXV:
ty!

PR #8534, opened 2021-11-20
BDSP: Make Psycho Boost Unobtainable

Karthik99999:
Deoxys is currently unobtainable, which should make psycho boost unobtainable as well.


KrisXV:
oops forgot about this, ty!

PR #8535, opened 2021-11-21
BDSP Random Battle: Fix set generation

KrisXV:
While there is a TON of room for improvement, this removes the issue of nonexistent items being assigned to sets and makes the format playable.

### DIFF
--- a/data/mods/gen8bdsp/random-teams.ts
+++ b/data/mods/gen8bdsp/random-teams.ts
@@ -8,8 +8,6 @@ export class RandomBDSPTeams extends RandomTeams {
 		super(format, prng);
 	}
 
-	
-
 	getHighPriorityItem(
 		ability: string,
 		types: Set<string>,
@@ -61,7 +59,7 @@ export class RandomBDSPTeams extends RandomTeams {
 		if (moves.has('rest') && !moves.has('sleeptalk') && ability !== 'Shed Skin') return 'Chesto Berry';
 		if (moves.has('bellydrum')) return 'Sitrus Berry';
 	}
-	
+
 	getMediumPriorityItem(
 		ability: string,
 		moves: Set<string>,
@@ -71,7 +69,6 @@ export class RandomBDSPTeams extends RandomTeams {
 		isDoubles: boolean,
 		isNoDynamax: boolean
 	): string | undefined {
-
 		// Choice items
 		if (
 			counter.get('Physical') >= 4 && ability !== 'Serene Grace' &&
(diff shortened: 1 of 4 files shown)


data/random-teams.ts
```diff
@@ -2080,6 +2080,7 @@ export class RandomTeams {
 		const pokemonPool = [];
 		for (let species of this.dex.species.all()) {
 			if (species.gen > this.gen || exclude.includes(species.id)) continue;
+			if (this.dex.currentMod === 'gen8bdsp' && species.gen > 4) continue;
```
Karthik99999, line 2083: Wouldn't it be better to do `delete species.randomBattleMoves` and `delete species.randomBattleMoves` in `dex-species.ts` for the bdsp mod instead of adding this hardcode in?

KrisXV: Maybe but I'd prefer to reduce the amount of hardcodes in the dex files, and this file already has other gen8bdsp hardcodes

data/mods/gen8bdsp/random-teams.ts
```diff
@@ -1,7 +1,162 @@
...
+			} else {
+				return (counter.get('Physical') > counter.get('Special')) ? 'Choice Band' : 'Choice Specs';
+			}
+		}
+		if (moves.has('auroraveil') || moves.has('lightscreen') && moves.has('reflect')) return 'Light Clay';
+		if (moves.has('rest') && !moves.has('sleeptalk') && ability !== 'Shed Skin') return 'Chesto Berry';
+		if (moves.has('bellydrum')) return 'Sitrus Berry';
+	}
+
+	getMediumPriorityItem(
```
AnnikaCodes, line 63: what's changed in these item generation functions from Sw/Sh? Is it just removing the items that don't exist in BDSP?

data/random-teams.ts
AnnikaCodes: Does `Dex.species.get` with a BDSP-modded Dex still pull `randomBattleMoves` from the non-BDSP formats-data? It seems like it shouldn't (and if it doesn't, this hardcode wouldn't be necessary).

If BDSP `dex` _does_ inherit SwSh `randomBattleMoves`, then yes, I'd rather have the hardcode in this file (since it's specific to random battles). That might lead to code duplication in `/randbats`, though, which is why I think getting Dex to read only the modded randbats moves data is the better solution.

data/mods/gen8bdsp/random-teams.ts
KrisXV: removing references of moves/items/abilities mostly

AnnikaCodes:
I'd still like to figure out if there's a better way to address the hardcoding, but this is fine for now.

PR #8538, opened 2021-11-21
Chat: Properly handle subfolders when assigning plugin names

mia-pi-git:
This is so subfolders with similarly named files (like `trivia/database.ts` and a `plugin/database.ts`, any `index.ts`, etc) won't get overridden in `Chat.plugins`.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1848,9 +1848,8 @@ export const Chat = new class {
 	packageData: AnyObject = {};
 
 	getPluginName(file: string) {
-		const folder = file.replace(__dirname, '').includes('chat-commands') ? 'chat-commands' : 'chat-plugins';
-		// the .slice(1, ...) removes the / from the beginning
-		let name = file.slice(file.indexOf(folder) + folder.length).slice(1, -file.lastIndexOf('.'));
+		const nameWithExt = pathModule.relative(__dirname, file).replace(/^chat-(?:commands|plugins)./, '');
+		let name = nameWithExt.slice(0, nameWithExt.lastIndexOf('.'));
 		if (name.endsWith('/index')) name = name.slice(0, -6);
 		return name;
 	}
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -29,7 +29,7 @@ import type {PartialModlogEntry} from './modlog';
 import {FriendsDatabase, PM} from './friends';
 import {SQL, Repl, FS, Utils} from '../lib';
 import {Dex} from '../sim';
-import {resolve} from 'path';
+import * as pathModule from 'path';
 import * as JSX from './chat-jsx';
 
 export type PageHandler = (this: PageContext, query: string[], user: User, connection: Connection)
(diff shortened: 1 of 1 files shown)


server/chat.ts
```diff
@@ -29,7 +29,7 @@ import type {PartialModlogEntry} from './modlog';
 import {FriendsDatabase, PM} from './friends';
 import {SQL, Repl, FS, Utils} from '../lib';
 import {Dex} from '../sim';
-import {resolve} from 'path';
+import * as pathModule from 'path';
```
Zarel, line 32: Maybe `extname` and `resolve` should be part of `FS`?

server/chat.ts
```diff
@@ -1847,15 +1847,23 @@ export const Chat = new class {
 
 	packageData: AnyObject = {};
 
+	getPluginName(file: string) {
+		const ext = pathModule.extname(file);
+		const folder = file.includes('chat-commands') ? 'chat-commands' : 'chat-plugins';
+		// the .slice(1, ...) removes the / from the beginning
+		let name = file.slice(file.indexOf(folder) + folder.length).slice(1, -ext.length);
```
Zarel, line 1854: This will fail if PS is in a directory named `chat-commands`.

server/chat.ts
```diff
@@ -1847,15 +1847,23 @@ export const Chat = new class {
 
 	packageData: AnyObject = {};
 
+	getPluginName(file: string) {
+		const ext = pathModule.extname(file);
```
Zarel, line 1851: `lastIndexOf('.')` was working fine before... did we really need to mess up our `path` import for this?

server/chat.ts
mia-pi-git: Fixed, i believe.

server/chat.ts
mia-pi-git: Reverted.

server/chat.ts
```diff
@@ -1847,15 +1847,22 @@ export const Chat = new class {
 
 	packageData: AnyObject = {};
 
+	getPluginName(file: string) {
+		const folder = file.replace(__dirname, '').includes('chat-commands') ? 'chat-commands' : 'chat-plugins';
+		// the .slice(1, ...) removes the / from the beginning
+		let name = file.slice(file.indexOf(folder) + folder.length).slice(1, -file.lastIndexOf('.'));
+		if (name.endsWith('/index')) name = name.slice(0, -6);
+		return name;
+	}
```
Zarel, line 1856: image

Mia, please, we've talked about committing code without testing it. :(

I thought to check this because I noticed you were slicing a sliced file with the unsliced `lastIndexOf`. But I've said this repeatedly: I really don't want to be the only barrier against stuff like this hitting the repo.

Zarel: Incidentally, this is a good use for `path.relative`:

```
zarel@Asparagine ~> node
Welcome to Node.js v17.0.1.
Type ".help" for more information.
> path.relative('/foobar', '/foobar/chat-commands/asdf/asdf.ts')
'chat-commands/asdf/asdf.ts'
```

Zarel:
```suggestion
	getPluginName(file: string) {
		const nameWithExt = pathModule.relative(__dirname, file).replace(/^chat-(?:commands|plugins)./, '');
		let name = nameWithExt.slice(0, nameWithExt.lastIndexOf('.'));
		if (name.endsWith('/index')) name = name.slice(0, -6);
		return name;
	}
```

Zarel:
Has this one been tested? What are the outputs?

AnnikaCodes:
Thanks!

PR #8540, opened 2021-11-23
Repeated Smack Down should still cancel Fly and Bounce each time

urkerab:
Reported [here](https://www.smogon.com/forums/posts/9038286).

Although the first Smack Down correctly cancels Fly and Bounce, subsequent uses of Smack Down also need to remove the `twoturnmove` volatile.


DaWoblefet:
Thanks urkerab!

PR #8542, opened 2021-11-25
Gen 4-6: Fix required items for Arceus formes

Karthik99999:
Makes it so that the validator doesn't say you can use z crystals to use an Arceus forme in gens 4-6.


KrisXV:
Nice catch, thanks!

PR #8552, opened 2021-12-05
Propagate command promises

mia-pi-git:
This PR removes User.chat and breaks its functionality into 3 parts.
- Chat.receive - takes a message, room, and connection, handles multiline messages, and parses everything apart to queue them up
- Chat.queue - puts messages into chat queue where applicable, parses them if not
- Chat.parse - as normal, parses commands.
Chat.parse now is `async` and awaits all commands that return promises.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -133,6 +133,9 @@ const MAX_MESSAGE_LENGTH = 300;
 
 const BROADCAST_COOLDOWN = 20 * 1000;
 const MESSAGE_COOLDOWN = 5 * 60 * 1000;
+const THROTTLE_MULTILINE_WARN = 3;
+const THROTTLE_MULTILINE_WARN_STAFF = 6;
+const THROTTLE_BUFFER_LIMIT = 6;
 
 const MAX_PARSE_RECURSION = 10;
 
@@ -1448,7 +1451,7 @@ export class CommandContext extends MessageContext {
 	}
 	refreshPage(pageid: string) {
 		if (this.connection.openPages?.has(pageid)) {
-			this.parse(`/join view-${pageid}`);
+			void this.parse(`/join view-${pageid}`);
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


server/chat.ts
```diff
@@ -1803,29 +1806,84 @@ export const Chat = new class {
...
+	async receive(message: string, room: Room | null, connection: Connection) {
+		const user = connection.user;
+		const multilineMessage = Chat.multiLinePattern.test(message);
+		if (multilineMessage) {
+			return Chat.queue(multilineMessage, room, connection);
+		}
+
+		const lines = message.split('\n');
+		if (!lines[lines.length - 1]) lines.pop();
+		// eslint-disable-next-line @typescript-eslint/prefer-optional-chain
```
AnnikaCodes, line 1871: Why don't we use optional chaining here? It should be explained in the comment.

mia-pi-git: Oh, I didn't write that part of the code - just moved it from user.chat. I assume it's because it used to be memory-heavy. CC @Zarel?

Zarel: ts-node completely fixes the perf issue with optional chaining; we can use it freely now.

server/chat.ts
```diff
@@ -1803,29 +1806,84 @@ export const Chat = new class {
...
+		Monitor.activeIp = null;
 		return result;
 	}
+
+	async queue(message: string, room: Room | null, connection: Connection) {
+		const user = connection.user;
+		const now = Date.now();
+		const noThrottle = user.hasSysopAccess() || Config.nothrottle;
+
+		if (message.startsWith('/cmd userdetails') || message.startsWith('>> ') || noThrottle) {
```
AnnikaCodes, line 1833: Should these commands also be exempt from network throttling? That's how it worked in the old system.

server/chat.ts
```diff
@@ -1803,29 +1806,84 @@ export const Chat = new class {
...
+		const user = connection.user;
+		const multilineMessage = Chat.multiLinePattern.test(message);
+		if (multilineMessage) {
+			return Chat.queue(multilineMessage, room, connection);
+		}
+
+		const lines = message.split('\n');
+		if (!lines[lines.length - 1]) lines.pop();
+		// eslint-disable-next-line @typescript-eslint/prefer-optional-chain
+		const maxLineCount = (user.isStaff || (room && room.auth.isStaff(user.id))) ?
```
AnnikaCodes, line 1872:
```suggestion
		const maxLineCount = (user.isStaff || (room?.auth.isStaff(user.id))) ?
```

AnnikaCodes:
If I'm understanding this correctly, asynchronous commands are only awaited by `Chat.parse` if the user doesn't have a chat queue. Is that correct?

mia-pi-git:
> If I'm understanding this correctly, asynchronous commands are only awaited by `Chat.parse` if the user's chat queue is empty. Is that correct?

No, they're awaited both ways.

AnnikaCodes:
> > If I'm understanding this correctly, asynchronous commands are only awaited by `Chat.parse` if the user's chat queue is empty. Is that correct?
> 
> No, they're awaited both ways.

```ts
} else if (now < user.lastChatMessage + throttleDelay) {
	user.chatQueue = [[message, room ? room.roomid : '', connection]];
	user.startChatQueue(throttleDelay - (now - user.lastChatMessage));
```
This... doesn't seem to await it - `User#startChatQueue` just sets a timeout to parse the message later? I'm confused.

PR #8557, opened 2021-12-06
Restore abuse-monitor

mia-pi-git:
Also adds support for using commands in a room without being in the room. Heavily restricted to specific punishments only for this plugin.

### DIFF
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -25,19 +25,12 @@ const ATTRIBUTES = {
 };
 
 export const cache: {
-	[roomid: string]: {[userid: string]: number} & {
+	[roomid: string]: {
+		users: Record<string, number>,
 		staffNotified?: boolean,
 		claimed?: ID,
 	},
 } = global.Chat?.oldPlugins['abuse-monitor']?.cache || {};
-
-for (const k in cache) {
-	const entry = cache[k] as any;
-	if (entry.flags) delete entry.flags;
-	if (entry.users) {
-		cache[k] = {...entry.users, staffNotified: entry.notified};
-	}
-}
 
 const defaults: FilterSettings = {
 	threshold: 4,
@@ -252,11 +245,11 @@ export const chatfilter: Chat.ChatFilter = function (message, user, room) {
 	void (async () => {
 		const {score, flags} = await PM.query({comment: message});
 		if (score) {
-			if (!cache[roomid]) cache[roomid] = {};
-			if (!cache[roomid][user.id]) cache[roomid][user.id] = 0;
-			cache[roomid][user.id] += score;
+			if (!cache[roomid]) cache[roomid] = {users: {}};
+			if (!cache[roomid].users[user.id]) cache[roomid].users[user.id] = 0;
+			cache[roomid].users[user.id] += score;
 			let hitThreshold = 0;
-			if (cache[roomid][user.id] >= settings.threshold) {
+			if (cache[roomid].users[user.id] >= settings.threshold) {
 				cache[roomid].staffNotified = true;
 				notifyStaff();
 				hitThreshold = 1;
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -128,7 +128,7 @@ export class Limiter {
 }
 
 function isCommon(message: string) {
-	message = message.replace(/\?!\., ;:/ig, '');
+	message = message.toLowerCase().replace(/\?!\., ;:/g, '');
 	return ['gg', 'wp', 'ggwp', 'gl', 'hf', 'glhf', 'hello'].includes(message);
 }
 
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -79,57 +79,57 @@ export interface PMResult {
 }
 
 function time() {
-    return Math.floor(Date.now() / 1000);
+	return Math.floor(Date.now() / 1000);
 }
 
 export class Limiter {
-    second = time();
-    countThisSecond = 0;
-    private counts = [] as number[];
-    private period: number;
-    max: number;
-    constructor(max: number, period: number) {
-        this.max = max;
-        this.period = period;
-    }
-    shouldRequest(depth = 0) {
-        const now = time();
-        if (this.second !== now) {
-            this.counts.push(this.countThisSecond);
-            this.trimCounts();
-            this.second = now;
-            this.countThisSecond = 0;
-        }
-        const count = this.countThisSecond + 1;
-        const total = this.counts
-            .concat([count])
-            .reduce((a, b) => a+b);
-        const avg = total / (this.counts.length + 1);
-        // if it'd go over, queue instead
-        if (avg > this.max) {
-            return false;
-        }
-        // if it'd succeed, now we increment the count since 
-        // we're going to allow it to happen
-        this.countThisSecond++;
-        return true;
-    }
-    /*private enqueue(depth = 0) {
-    
+	second = time();
+	countThisSecond = 0;
+	private counts = [] as number[];
+	private period: number;
+	max: number;
+	constructor(max: number, period: number) {
+		this.max = max;
+		this.period = period;
+	}
+	shouldRequest(depth = 0) {
+		const now = time();
+		if (this.second !== now) {
+			this.counts.push(this.countThisSecond);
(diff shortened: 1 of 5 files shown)


server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,663 @@
...
+ * We use Google's Perspective API to classify messages.
+ * @see https://perspectiveapi.com/
+ * by Mia.
+ * @author mia-pi-git
+ */
+
+import {FS, Utils, Net, ProcessManager, Repl} from '../../lib';
+import {Config} from '../config-loader';
+import {toID} from '../../sim/dex-data';
+import {throttling} from '../../tools/set-import/importer';
```
Zarel, line 14: ugh pls no

Zarel: Please don't change `set-import` or introduce new dependencies in crazy ways. `tools` should be self-contained.

Zarel: The throttle in there can't be audited, anyway. Easy enough to just reimplement; maybe put an implementation in `Utils`.

mia-pi-git: Okay, done.

mia-pi-git: Moved it to utils and restored set-import.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,719 @@
...
+	private counts = [] as number[];
+	private period: number;
+	max: number;
+	constructor(max: number, period: number) {
+		this.max = max;
+		this.period = period;
+	}
+	shouldRequest(depth = 0) {
+		const now = time();
+		if (this.second !== now) {
```
Zarel, line 97: The counter is wrong if multiple seconds pass between requests.

mia-pi-git: Wrong how so? should we just have an interval that clears it once a second?

mia-pi-git: Or, wait, no, I know what to do.

mia-pi-git: Fixed.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,727 @@
...
+	shouldRequest() {
+		const now = time();
+		if (this.second !== now) {
+			const diff = now - this.second;
+			if (diff <= this.period) {
+				for (let i = 0; i < (diff - 1); i++) {
+					this.counts.push(0);
+				}
+			} else {
+				this.counts = [];
```
Zarel, line 104: This is wrong, it should be `Array(this.period).fill(0)`.

mia-pi-git: Yeah, that's what i was just thinking about - see DMs.

mia-pi-git: Fixed.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,727 @@
...
+			} else {
+				this.counts = Array(this.period).fill(0);
+			}
+			this.counts.push(this.countThisSecond);
+			this.trimCounts();
+			this.second = now;
+			this.countThisSecond = 0;
+		}
+		const count = this.countThisSecond + 1;
+		const counts = this.getCounts();
```
Zarel, line 112: Keeping these separate isn't really worth the additional complexity.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,727 @@
...
+		this.trimCounts();
+		return this.counts;
+	}
+	private trimCounts() {
+		if (this.counts.length > this.period) {
+			// oldest first
+			while (this.counts.length > this.period) this.counts.shift();
+		}
+	}
+}
```
Zarel, line 140:
```suggestion
export class RollingCounter {
	counts: number[] = [0];
	readonly size: number;
	constructor(limit: number) {
		this.size = limit;
	}
	increment() {
		this.counts[this.counts.length - 1]++;
	}
	rollOver(amount: number) {
		if (amount > this.size) {
			this.counts = Array(this.size).fill(0);
			return;
		}
		for (let i = 0; i < amount; i++) {
			this.counts.push(0);
			if (this.counts.length > this.size) this.counts.shift();
		}
	}
	mean() {
		let total = 0;
		for (const elem of this.counts) total++;
		return total / this.counts.length;
	}
}

export class Limiter {
	readonly counter: RollingCounter;
	readonly max: number;
	lastCounterRoll = time();
	constructor(max: number, period: number) {
		this.max = max;
		this.counter = new RollingCounter(period);
	}
	shouldRequest() {
		const now = time();
		this.counter.rollOver(now - this.lastCounterRoll);
		this.lastCounterRoll = now;

		if (this.counter.mean() > this.max) return false;
		this.counter.increment();
		return true;
	}
}
```

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,727 @@
...
+
+function isCommon(message: string) {
+	message = message.replace(/\?!\., ;:/ig, '');
```
Zarel, line 143:
```suggestion
	message = message.toLowerCase().replace(/\?!\., ;:/g, '');
```

mia-pi-git: Don't you want `/i`? What if someone types `WP`?

server/chat-plugins/abuse-monitor.ts
mia-pi-git: Resolved by reimplementation.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,715 @@
...
+			saveSettings();
+			this.privateGlobalModAction(`${this.user.name} set the abuse monitor trigger threshold to ${num}.`);
+			this.globalModlog('ABUSEMONITOR THRESHOLD', null, `${num}`);
+			this.sendReply(
+				`|html|Remember to use <code>/am respawn</code> to deploy the settings to the child process.`
+			);
+		},
+		resolve(target) {
+			this.checkCan('lock');
+			target = target.toLowerCase().trim().replace(/ /ig, '');
```
Zarel, line 365:
```suggestion
			target = target.toLowerCase().trim().replace(/ +/g, '');
```

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,715 @@
...
+	"INSULT": {},
+	"PROFANITY": {},
+	"THREAT": {},
+};
+
+export const cache: {
+	[roomid: string]: {[userid: string]: number} & {
+		staffNotified?: boolean,
+		claimed?: ID,
+	},
```
Zarel, line 31: Why is this `&` rather than a `users` sub-object?

mia-pi-git: I thought it was more readable, personally.

mia-pi-git: Nope, that's actually necessary for TS too.

mia-pi-git: Discussed in dms, restored cache[roomid].users.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -0,0 +1,726 @@
...
+					`Invalid punishment given. ` +
+					`Must be one of ${Object.keys(NOJOIN_COMMAND_WHITELIST).join(', ')}.`
+				);
+			}
+			this.room = tarRoom;
+			this.room.reportJoin('j', user.getIdentityWithStatus(this.room), user);
+			const result = await this.parse(`${cmd} ${rest}`, {bypassRoomCheck: true});
+			if (result) { // command succeeded - send followup
+				this.addModAction(
+					'If you have questions about this action, please contact staff by typing ``/ht`` in any chat.'
```
Zarel, line 393: Probably this should be a button.

Zarel (approved):

Zarel:
Note that this adds support for using commands in rooms you're not in.

You should really not sneak in features like these. I approved it in DMs, but it should be clearly explained in the PR summary.

server/chat-plugins/abuse-monitor.ts
mia-pi-git: Done.

PR #8558, opened 2021-12-07
Add Sleep Moves Clause to Revelationmons

TomOfTomKat:
Revelationmons, the current OMotM, should have Sleep Moves Clause implement, and does not. Proof of the clause can be found here: https://www.smogon.com/forums/threads/revelationmons-omotm-for-december.3692297/


AnnikaCodes (approved):
Looks good, thanks!

PR #8559, opened 2021-12-07
Gen2 Random battles movesets

Tarrembeau:
Correcting oversights on octillery and azumarill movepools, who actually don't learn fire blast and charm, respectively.


AnnikaCodes:
Thanks!

PR #8563, opened 2021-12-10
Avatar manager

Zarel:
This is a refactor of avatar data.

- Storage is now in `config/avatars.json` rather than `config/config.js`
- There are now new commands to add/remove custom avatars from users.
- Users now have new commands to see what avatars they have access to, and changing your avatar will now also change your default.

The first commit is the avatar manager currently live on Main; the next few commits are some experimental changes to retain support for side server avatars. The main thing left to do is code to migrate `Config.avatars`.


server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+		if (!target) return this.parse(`/help defaultavatar`);
+		const [inputUsername, inputAvatar] = this.splitOne(target);
+		if (!Users.isUsername(inputUsername)) {
+			throw new Chat.ErrorMessage(`"${inputUsername}" is not a valid username.`);
+		}
+		const userid = toID(inputUsername);
+		const avatar = Avatars.convert(inputAvatar);
+
+		const allowed = customAvatars[userid]?.allowed.filter(Boolean);
+		if (!allowed) {
```
mia-pi-git, line 706:
```suggestion
		if (!allowed.length) {
```

server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+			this.globalModlog('REMOVE AVATAR', userid, avatar);
+			this.sendReplyBox(<div>
+				{Avatars.img(avatar)}<br />
+				Removed from <username class="username">{inputUsername}</username>
+			</div>);
+		} else {
+			// delete all
+			delete customAvatars[userid];
+			Avatars.save();
+			this.globalModlog('REMOVE AVATAR', userid, allowed.join(','));
```
mia-pi-git, line 722: Maybe this entry should say `remove all`, or something close, to make it clearer all are being removed?

server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+	},
+	removeavatarhelp: 'addavatarhelp',
+
+	async avatarusers(target, room, user) {
+		target = '#' + toID(target);
+		if (!Avatars.userCanUse(user, target) && !user.can('alts')) {
+			throw new Chat.ErrorMessage(`You don't have access to avatar "${target}"`);
+		}
+
+		if (!this.runBroadcast()) return;
```
mia-pi-git, line 737: Nitpick, but strictly speaking this isn't necessary since it uses ErrorMessage.
```suggestion
		this.runBroadcast();
```

server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+				throw new Chat.ErrorMessage(`Invalid username "${username}"`);
+			}
+			await Avatars.validate('#' + toID(username));
+		}
+
+		const userids = usernames.map(toID);
+		for (const userid of userids) {
+			const avatar = '#' + userid;
+			Avatars.addPersonal(userid, avatar);
+			this.globalModlog('PAVATAR', userid, avatar);
```
mia-pi-git, line 776: `PAVATAR` doesn't seem really clear as an entry name.

server/chat-commands/avatars.tsx
Zarel: `allowed.filter(Boolean)` can never be empty.

Zarel: Hm, plausibly I could update the type to reflect this...

Zarel: Nah, that makes initialization hard. But the point is that if `allowed` were ever empty, `customAvatars[userid]` would just be deleted.

server/chat-commands/avatars.tsx
Zarel: hm

server/chat-commands/avatars.tsx
Zarel: Oh huh this was a carry-over from the old command. We should probably port all the other uses forward instead of going one-by-one.

server/chat-commands/avatars.tsx
mia-pi-git: But if it can never be empty, why do you check this? 
`!allowed` will never be true since [] is truthy.

mia-pi-git: Oops. Misread.

server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+						You have a new custom avatar!
+					</p>
+					<p>
+						{entry.allowed.map(avatar => avatar && [Avatars.img(avatar), ' '])}
+					</p>
+					Use <button class="button" name="send" value="/avatars"><code>/avatars</code></button> for usage instructions.
+				</>}`
+			);
+			delete entry.notNotified;
+			saveCustomAvatars();
```
AnnikaCodes, line 212: I find it a little confusing that the avatars are saved in a function called `tryNotify`.

server/chat-commands/avatars.tsx
```diff
@@ -0,0 +1,869 @@
...
+	allowed: [AvatarID | null, ...AvatarID[]];
+	/**
+	 * undefined = personal avatar;
+	 * null = no default (shouldn't occur in practice);
+	 * (ignored during Christmas, where the any avatar is default if it ends in `xmas`.)
+	 */
+	default?: AvatarID | null;
+}
+
+const customAvatars: {[userid: string]: AvatarEntry} = Object.create(null);
```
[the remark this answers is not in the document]
mia-pi-git: It's on Users.Avatars, isn't it?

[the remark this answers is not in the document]
AnnikaCodes: Oh, yeah, it's wrapped by the Avatar class.

server/chat-commands/avatars.tsx
Zarel: `saveCustomAvatars` saves the entire avatar table `customAvatars`, which tracks whether or not the user has been notified of an avatar. Is there a rename you'd suggest?

server/chat-commands/avatars.tsx
Zarel: Suggest an alternative?

server/chat-commands/avatars.tsx
AnnikaCodes: `tryNotifyAndSave`?

server/chat-commands/avatars.tsx
Zarel: like `'* ' + allowed.join(',')`?

server/chat-commands/avatars.tsx
Zarel: I don't think `andSave` is important enough to go in the function name? Why is that important to you?

server/chat-commands/avatars.tsx
mia-pi-git: No, I mean more like 
```suggestion
			this.globalModlog('REMOVEALLAVATARS', userid, allowed.join(','));
```

server/chat-commands/avatars.tsx
mia-pi-git:
```suggestion
			this.globalModlog('PERSONALAVATAR ADD', userid, avatar);
```

mia-pi-git: Or even just `PERSONALAVATAR`.

PR #8565, opened 2021-12-12
Add an abstraction for typed database queries

mia-pi-git:
Useful for typechecking things to avoid bugs.

### DIFF
--- a/lib/sql.ts
+++ b/lib/sql.ts
@@ -5,6 +5,8 @@
 import {QueryProcessManager} from './process-manager';
 import type * as sqlite from 'better-sqlite3';
 import {FS} from './fs';
+// @ts-ignore in case not installed
+import type {SQLStatement} from 'sql-template-strings';
 
 export const DB_NOT_FOUND = null;
 
@@ -17,6 +19,8 @@ export interface SQLOptions {
 }
 
 type DataType = unknown[] | Record<string, unknown>;
+export type SQLInput = string | number | null;
+export interface ResultRow {[k: string]: SQLInput}
 
 export interface TransactionEnvironment {
 	db: sqlite.Database;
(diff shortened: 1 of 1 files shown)


lib/sql.ts
```diff
@@ -5,6 +5,8 @@
 import {QueryProcessManager} from './process-manager';
 import type * as sqlite from 'better-sqlite3';
 import {FS} from './fs';
+// @ts-ignore in case not installed
```
Zarel, line 8: In which situation would this not be installed?

mia-pi-git: If you don't want to use it? It's optional.

lib/sql.ts
```diff
@@ -259,12 +263,154 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any>
 	}
 }
 
+export const tables = new Map<string, DatabaseTable<any>>();
+
+export class DatabaseTable<T> {
```
AnnikaCodes, line 268: Shouldn't there be a way to override this - what if some rows in a table contain integers while others contain strings?

mia-pi-git: Why would that be happening?
That seems like a failure on the part of the user. 
(Also, I'd say just typecast the query results manually where necessary).

lib/sql.ts
Zarel: It's optional? It's a library... Devs should probably have everything installed.

mia-pi-git: I thought we wanted SQLite stuff to be optional?

lib/sql.ts
AnnikaCodes: Sorry, I meant columns, not row.

lib/sql.ts
```diff
@@ -259,12 +263,154 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any>
...
+		query.append(' LIMIT 1');
+		const rows = await this.selectAll(entries, query);
+		return rows?.[0] || null;
+	}
+	selectAll(
+		entries: string | string[],
+		where?: SQLStatement
+	): Promise<T[]> {
+		const query = this.SQL`SELECT `;
+		if (typeof entries === 'string') {
```
AnnikaCodes, line 305: Errrr, should `table.selectAll('my_column');` return all values of `my_column` instead of `SELECT *`?
I guess you can do `table.selectAll(['my_column'])`?

lib/sql.ts
Zarel: This API is modeled on my own library in:

https://github.com/Zarel/telepic/tree/master/server

In mine, `selectAll<T>` is generic:

https://github.com/Zarel/telepic/blob/2c55b7b09d95491e51853d1822ba990193c94c81/server/users.ts#L56

mia-pi-git: > Shouldn't there be a way to override this - what if some column in a table contain integers while others contain strings?

Oh! I understand now. T should generally be an object, and it types the full object. IE, using your example, then we'd selectAll('*'), which returns {log_line: string, timestamp: number}[]. Even if you only select `log_line`, that still returns `{log_line: string}`.

lib/sql.ts
mia-pi-git: Yes, you can.

mia-pi-git: Note that doing `selectAll('my_column')` is the same as `selectAll(['my_column'])`.

lib/sql.ts
mia-pi-git: > In mine, `selectAll<T>` is generic:
> 
> https://github.com/Zarel/telepic/blob/2c55b7b09d95491e51853d1822ba990193c94c81/server/users.ts#L56
> 
> I'd recommend that here, too.

Can we not use a regular `query` function as a catch-all for that instead? (Or, rather, since it's SQLite - `get`, `all`, `run`, and `exec`?)

AnnikaCodes: Yes, that sounds good.

mia-pi-git: I added type overrides to most functions (`all`, `selectAll`, `selectOne`, etc)

lib/sql.ts
AnnikaCodes: > Note that doing `selectAll('my_column')` is the same as `selectAll(['my_column'])`.

The code you have appears to return values for all columns (`query.append(' * ')`) if the `entries` argument is a string, whereas when it's an array it only returns values for the specified columns. Am I missing something?

lib/sql.ts
```diff
@@ -259,12 +263,154 @@ export class SQLDatabaseManager extends QueryProcessManager<DatabaseQuery, any>
...
+		query.append(this.SQL` = ${keyId}`);
+		return this.selectOne(entries, query);
+	}
+	updateAll(toParams: Partial<T>, where?: SQLStatement, limit?: number) {
+		const to = Object.entries(toParams);
+		const query = this.SQL`UPDATE `;
+		query.append(this.name + ' SET ');
+		for (let i = 0; i < to.length; i++) {
+			const [k, v] = to[i];
+			query.append(`${k} = `);
```
AnnikaCodes, line 333: This doesn't sanitize column names, which is fine, but something to be careful of (I need to remember it for future reviews which is half the reason I'm commenting it here).

AnnikaCodes (commented):
This seems pretty good at a glance, but I would want to make sure @Zarel is OK with this (and has read/understood it) before I merge.

mia-pi-git:
Bump?

lib/sql.ts
AnnikaCodes: You resolved this but didn't actually explain my question...

mia-pi-git: it was fixed

Zarel (approved):
Looks good; this is adapted from code I wrote myself so it's not like it's mysterious to me.

AnnikaCodes:
Looks good, thanks!

PR #8566, opened 2021-12-13
Ban Barraskewda from Revelationmons

TomOfTomKat:
Revelationmons, the current OMotM, has recently banned Barraskewda, and the ladder banlist needs to be updated. Proof of ban here: https://www.smogon.com/forums/threads/revelationmons-omotm-for-december.3692297/post-9052776


KrisXV:
thx

PR #8568, opened 2021-12-15
Gen 4: Fix Griseous Orb Validation

Karthik99999:
Only Giratina can hold the Griseous Orb without hacking.


DaWoblefet:
Thanks Karthik!

PR #8569, opened 2021-12-15
Gen 3-4: Fix Knock Off mechanics

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9056030

Draft for now as I still have to write tests.

Also, I was wondering if a new client message should be created for this. Modifying `-enditem` to make it keep the item for Gen 3-4 knock off would probably break some older replays, which is why I think creating a new client message that specifically handles Gen 3-4 knock off could be a good idea.

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -832,7 +832,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			const item = target.getItem();
 			if (this.singleEvent('TakeItem', item, target.itemState, target, target, move, item)) {
 				target.itemState.knockedOff = true;
-				this.add('-enditem', target, item.name, '[from] move: Knock Off', '[of] ' + source);
+				this.add('-item', target, item.name, '[from] move: Knock Off', '[silent]');
 				this.hint("In Gens 3-4, Knock Off only makes the target's item unusable; it cannot obtain a new item.", true);
 			}
 		},
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -828,10 +828,10 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	knockoff: {
 		inherit: true,
 		onAfterHit(target, source) {
-			const item = target.takeItem();
-			if (item) {
-				this.add('-enditem', target, item.name, '[from] move: Knock Off', '[of] ' + source);
-			}
+			if (!target.item) return;
+			target.itemState.knockedOff = true;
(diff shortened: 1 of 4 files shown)


data/mods/gen4/moves.ts
```diff
@@ -827,10 +827,13 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
-		onAfterHit(target, source) {
-			const item = target.takeItem();
-			if (item) {
+		onAfterHit(target, source, move) {
+			if (!target.item || target.itemState.knockedOff) return;
+			const item = target.getItem();
+			if (this.singleEvent('TakeItem', item, target.itemState, target, target, move, item)) {
+				target.itemState.knockedOff = true;
 				this.add('-enditem', target, item.name, '[from] move: Knock Off', '[of] ' + source);
+				this.hint("Prior to Gen 5, Knock Off did not actually remove the target's item. Instead, the item would be unusable for the rest of the battle, and the target would not be able to obtain a new item.", true);
```
DaWoblefet, line 836: This is a fairly long hint for our standards, especially on something that will happen after any time Knock Off itself is used. I'd rather move the hint to a situation where unusual behavior occurs.

As for a short version:
In Gens 3-4, Knock Off only makes the target's item unusable; it cannot obtain a new item.

Karthik99999: > especially on something that will happen after any time Knock Off itself is used

I made it so that the hint only shows the first time knock off is used in a battle. I guess the hint could be moved somewhere else but I don't know where that would be.

Karthik99999:
I realized there's a good amount of stuff that checks `Pokemon#item` to see if a pokemon has an item (i.e. Unburden). Do I have to add a check to `Pokemon#itemState#knockoff` everywhere that the item is checked or is there a better solution?

data/mods/gen4/moves.ts
DaWoblefet: Oops, missed the ``true`` argument haha

Marty-D:
> I realized there's a good amount of stuff that checks `Pokemon#item` to see if a pokemon has an item (i.e. Unburden). Do I have to add a check to `Pokemon#itemState#knockoff` everywhere that the item is checked or is there a better solution?

I think with your changes to `eatItem`, `useItem`, and `takeItem`, only Frisk and Unburden require an extra check? The alternative would be creating a combined check for `item` and the `knockedOff` state and using that almost everywhere anyway.

Marty-D (approved):
Very nice, thanks! 👍

PR #8570, opened 2021-12-15
Fix setting avatars on side servers

Karthik99999:
`canUse` was replacing `.` for the avatar id which all side server custom avatars have.


Zarel:
Thanks!

PR #8572, opened 2021-12-15
Handle failing move case for Protean

RichkidRich:
https://github.com/smogon/pokemon-showdown/projects/3#card-71347448

Protean used to activate and change pokemon type before the move failed. This PR makes it so that the move fails beforehand and the pokemon type does not change.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2830,6 +2830,51 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 	protean: {
 		onPrepareHit(source, target, move) {
 			if (move.hasBounced || move.sourceEffect === 'snatch') return;
+			if (move.name === 'Mirror Coat') {
+				if (!source.volatiles['mirrorcoat']) return;
+				if (source.volatiles['mirrorcoat'].slot === null) return;
+			}
+			if (move.name === 'Counter') {
+				if (!source.volatiles['counter']) return;
+				if (source.volatiles['counter'].slot === null) return;
+			}
+			if (move.name === 'Metal Burst') {
+				const lastDamagedBy = source.getLastDamagedBy(true);
+				if (lastDamagedBy === undefined || !lastDamagedBy.thisTurn) return;
+			}
+			if (move.name === 'Fling') {
+				if (source.ignoringItem()) return;
+				const item = source.getItem();
+				if (!this.singleEvent('TakeItem', item, source.itemState, source, source, move, item)) return;
+				if (!item.fling) return;
+			}
+			if (move.name === 'Sucker Punch') {
+				const action = this.queue.willMove(target);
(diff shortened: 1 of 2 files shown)


DaWoblefet:
This isn't quite the approach that you'd want to take here - there's a lot more moves than are in the test that fail before Protean, these are just a few. You'll notice that hardcoding Mirror Coat, Counter, and Metal Burst would have been sufficient to make the test pass. Even if you did hardocde, you'd have to duplicate the hardcodes in Libero which isn't great.

As I noted in the card, the correct solution is to migrate Counter, Mirror Coat, and Metal Burst to an event that occurs _before_
Protean is run, which would likely be onTry() in this case. in USUM, the failure check for Counter, Mirror Coat, and Metal Burst is shared by the following move failures. As you can see, most (if not all) of these are implemented using the onTry() event.

Follow Me
Rage Powder
Splash vs Gravity
Stockpile (more than 3 Stockpiles already)
Spit Up / Swallow (no Stockpile)
Counter / Mirror Coat / Metal Burst when undamaged
Last Resort
Snore
Sucker Punch
Natural Gift
Magic Coat
Endure / Protect / Detect / King's Shield / Spiky Shield / Baneful Bunker
Bide
Rest
Wide Guard / Quick Guard / Crafty Shield
Mat Block
Teleport
Fling
Magnet Rise
Fake Out
First Impression
Destiny Bond
Sleep Talk
Telekinesis
Sky Drop
Hyperspace Fury
Dark Void
Beak Blast
Shell Trap
Aurora Veil

urkerab:
Some of these may have already been fixed. In particular, Fling, Sucker Punch, Steel Roller, Aura Wheel, Magnet Rise and Aurora Veil use the `onTry` or `onPrepareHit` events which should already run before Protean. I don't know offhand whether the other moves should be using those events instead, rather than trying to work around them in Protean.

DaWoblefet:
> Some of these may have already been fixed. In particular, Fling, Sucker Punch, Steel Roller, Aura Wheel, Magnet Rise and Aurora Veil use the `onTry` or `onPrepareHit` events which should already run before Protean. I don't know offhand whether the other moves should be using those events instead, rather than trying to work around them in Protean.

This is correct; only Counter, Mirror Coat, and Metal Burst are currently buggy. They were just lumped in with the old test because it's similar behavior.

urkerab:
Sorry, it took me a few minutes to research my comment so I hadn't noticed that you'd already commented in the mean time.

DaWoblefet:
I'm converting this to a draft since the approach isn't quite what we're looking for yet. Please feel free to ask questions if you need help!

DaWoblefet:
Superseded by https://github.com/smogon/pokemon-showdown/pull/8723.

PR #8573, opened 2021-12-16
Ban Blacephalon from Revelationmons

TomOfTomKat:
Revelationmons has recently voted to ban Blacephalon, and the ladder needs to be updated to reflect this ban. Proof of ban can be found here: https://www.smogon.com/forums/threads/revelationmons-omotm-for-december.3692297/post-9057202


AnnikaCodes:
Looks good, thanks!

PR #8574, opened 2021-12-16
BD/SP Random Battle: Add missing pokemons

Tarrembeau:
Update donphan and rampardos movepools as well


AnnikaCodes (approved):
Looks good, thanks!

PR #8575, opened 2021-12-18
Abuse-monitor: Support tracking stats

mia-pi-git:
Useful to be able to review.

### DIFF
--- a/databases/migrations/chat-plugins/v5.sql
+++ b/databases/migrations/chat-plugins/v5.sql
@@ -5,7 +5,6 @@ CREATE TABLE perspective_stats (
 	staff TEXT NOT NULL,
 	roomid TEXT NOT NULL PRIMARY KEY,
 	result TINYINT(1) NOT NULL,
-	date TEXT NOT NULL, -- YYYY-MM-DD - for easy stat searching.
 	timestamp INTEGER NOT NULL
 );
 
--- a/databases/migrations/chat-plugins/v5.sql
+++ b/databases/migrations/chat-plugins/v5.sql
@@ -1,0 +1,20 @@
+-- Creates stats tables for the Perspective moderation tool.
+BEGIN TRANSACTION;
+
+CREATE TABLE perspective_stats (
+	staff TEXT NOT NULL,
+	roomid TEXT NOT NULL PRIMARY KEY,
+	result TINYINT(1) NOT NULL,
+	date TEXT NOT NULL, -- YYYY-MM-DD - for easy stat searching.
+	timestamp INTEGER NOT NULL
+);
+
+CREATE INDEX staff_idx ON perspective_stats(staff);
+CREATE INDEX result_idx ON perspective_stats(result);
+CREATE INDEX date_idx ON perspective_stats(date);
+
+-- update database version
+UPDATE db_info SET value = '5' WHERE key = 'version';
+COMMIT;
+
+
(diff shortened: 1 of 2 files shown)


databases/migrations/chat-plugins/v5.sql
```diff
@@ -0,0 +1,20 @@
+-- Creates stats tables for the Perspective moderation tool.
+BEGIN TRANSACTION;
+
+CREATE TABLE perspective_stats (
+	staff TEXT NOT NULL,
+	roomid TEXT NOT NULL PRIMARY KEY,
+	result TINYINT(1) NOT NULL,
+	date TEXT NOT NULL, -- YYYY-MM-DD - for easy stat searching.
+	timestamp INTEGER NOT NULL
```
AnnikaCodes, line 9: Why do we store both the date and the timestamp? 

I'm not sure about the details of SQLite's performance, but I would imagine that an integer bounds comparison is probably faster than string pattern matching.

databases/migrations/chat-plugins/v5.sql
```diff
@@ -0,0 +1,20 @@
...
+);
+
+CREATE INDEX staff_idx ON perspective_stats(staff);
+CREATE INDEX result_idx ON perspective_stats(result);
```
AnnikaCodes, line 13: What are these two indices for?
You only seem to filter by date and room ID in your code.

databases/migrations/chat-plugins/v5.sql
mia-pi-git: Yeah, I think it was because I thought it was easier to just search for a date. Probably best to make it just do a stupidly long query to compare integers.

mia-pi-git: Updated.

databases/migrations/chat-plugins/v5.sql
mia-pi-git: Futureproofing. Had plans for it - still do - just haven't put them into practice quite yet.

mia-pi-git:
Bump?

databases/migrations/chat-plugins/v5.sql
AnnikaCodes: I would prefer to not add indices until we have code using them.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -98,6 +98,12 @@ interface BattleInfo {
 	log: string[];
 }
 
+// stolen from chatlog. necessary here, but importing chatlog sucks.
```
AnnikaCodes, line 101: Don't we already have date-handling functions in Chat? Should this also go there?

databases/migrations/chat-plugins/v5.sql
mia-pi-git: Ok.

server/chat-plugins/abuse-monitor.ts
mia-pi-git: Probably, but that should be a follow-up PR since it requires some work in chatlog.

databases/migrations/chat-plugins/v5.sql
mia-pi-git: Done.

AnnikaCodes (approved):
Seems fine.

PR #8576, opened 2021-12-18
Fix interaction of Transformed Neutralizing Gas

DaWoblefet:
On cart, Transforming into Neutralizing Gas isn't enough to actually get it to work. It doesn't stay active at all.

Leaving as a PR for now because the behavior is pretty new and so I want to make sure nothing new crops up in tests that throws a wrench in this implementation.


Zarel (approved):
Looks safe enough to me. I'd be fine just merging now; it's unlikely to cause any bigger bug than the one it fixes.

DaWoblefet:
Yep, it should be good; I can't think of an edge case for this and haven't heard of one from anyone else yet either.

PR #8579, opened 2021-12-20
Helptickets: Support adding custom response buttons to text tickets

mia-pi-git:
Requested by staff.


AnnikaCodes:
This seems okay.

PR #8581, opened 2021-12-23
Wi-Fi: Convert to TSX

KrisXV:
The HTML is mostly working, but I'm having some issues, namely with using `Chat.formatText` in the VNodes because it escapes the converted chat formatting


KrisXV:
I think this is ready for review @zarel

server/chat-plugins/wifi.tsx
```diff
@@ -280,22 +293,40 @@ class Giveaway extends Rooms.RoomGame {
...
+							<strong>Extra Information</strong><br />
+							{renderString(this.extraInfo.trim().replace(/<br \/>/g, '\n'))}
+						</td>
+					</tr> : ''}
+			</table>
+			<p style={{textAlign: 'center', fontSize: '7pt', fontWeight: 'bold'}}>
+				<u>Note:</u> You must have a Switch, Pok&eacute;mon {gameName[this.game]},
+				&#32;and Nintendo Switch Online to receive the prize.
+				&#32;Do not join if you are currently unable to trade. Do not enter if you have already won this exact Pok&eacute;mon,
+				&#32;unless it is explicitly allowed.
```
Zarel, line 327: Convention is to use `{}`, not `&#32;`.

```suggestion
				<u>Note:</u> You must have a Switch, Pok&eacute;mon {gameName[this.game]}, {}
				and Nintendo Switch Online to receive the prize. {}
				Do not join if you are currently unable to trade. Do not enter if you have already won this exact Pok&eacute;mon, {}
				unless it is explicitly allowed.
```

server/chat-plugins/wifi.tsx
```diff
@@ -141,21 +145,31 @@ class Giveaway extends Rooms.RoomGame {
...
+	generateReminder(joined?: boolean): string | Chat.VNode;
+	generateReminder() {
+		return '';
+	}
 
 	getStyle() {
-		if (this.game === 'BDSP') return `style="background-color: #aa66a9;color: #fff;padding: 2px 4px;"`;
-		return `class="broadcast-blue"`;
+		if (this.game === 'BDSP') return {style: {backgroundColor: '#aa88a9', color: '#fff', padding: '2px 4px'}};
+		return {class: "broadcast-blue"};
```
Zarel, line 155: So what was wrong with the one I suggested, where you always send `broadcast-blue` but sometimes override the colors?

KrisXV: Nothing, I just like this approach more because I don't see the point in using a class then changing half of the elements the class has

Zarel: Changing half is better than changing all, and makes the code more readable, too.

Zarel: It's also more semantic. Only changing color is better than hard-changing layout.

KrisXV: I don't see how this makes the code more readable but ok

server/chat-jsx.tsx
```diff
@@ -36,3 +37,35 @@ export type VNode = preact.VNode;
 
 export const h = preact.h;
 export const Fragment = preact.Fragment;
+
+function subRender(
+	child: preact.ComponentChild, opts?: {isTrusted?: boolean, replaceLinebreaks?: boolean}
+): preact.VNode {
+	switch (typeof child) {
+	case 'string':
+		return <div dangerouslySetInnerHTML={{__html: Chat.formatText(child, opts?.isTrusted, opts?.replaceLinebreaks)}} />;
```
Zarel, line 46: A div is often the wrong choice for this...

server/chat-jsx.tsx
```diff
@@ -36,3 +37,35 @@ export type VNode = preact.VNode;
...
+	case 'object':
+		return <>{child ? child.toString() : '' + child}</>;
+	default:
+		return preact.isValidElement(child) ? child : <>{'' + child}</>;
+	}
+}
+
+export class FormatText extends preact.Component<{isTrusted?: boolean, replaceLinebreaks?: boolean}> {
+	render() {
+		if (Array.isArray(this.props.children)) {
```
Zarel, line 59: I don't see a need to handle all these special cases. Just throw if `children` isn't a string.

server/chat-jsx.tsx
Zarel: I wonder if a fragment with `dangerouslySetInnerHTML` is possible.

server/chat-plugins/wifi.tsx
```diff
@@ -141,21 +141,32 @@ class Giveaway extends Rooms.RoomGame {
...
+	generateReminder(joined?: boolean): string | Chat.VNode;
+	generateReminder() {
+		return '';
+	}
 
 	getStyle() {
-		if (this.game === 'BDSP') return `style="background-color: #aa66a9;color: #fff;padding: 2px 4px;"`;
-		return `class="broadcast-blue"`;
+		const css: {[k: string]: string | {[k: string]: string}} = {className: "broadcast-blue"};
+		if (this.game === 'BDSP') css.style = {backgroundColor: '#aa88a9'};
```
Zarel, line 151: Always set bg and fg color simultaneously.

server/chat-plugins/wifi.tsx
```diff
@@ -767,24 +816,33 @@ export class GTS extends Rooms.RoomGame {
...
+			this.changeUhtml(
+				<p style={{textAlign: 'center', fontSize: '13pt', fontWeight: 'bold'}}>The GTS giveaway has finished.</p>
+			);
 			this.room.modlog({
 				action: 'GTS FINISHED',
 				userid: this.giver.id,
 				note: `their GTS giveaway for "${this.summary}"`,
 			});
-			this.send(`<p style="text-align:center;font-size:11pt">The GTS giveaway for a "<strong>${Utils.escapeHTML(this.lookfor)}</strong>" has finished.</p>`);
+			this.send(Chat.html`${<p style={{textAlign: 'center', fontSize: '11pt'}}>
```
Zarel, line 843: `this.send` should probably take a VNode so you don't have to do the weird wrap.

server/chat-plugins/wifi.tsx
```diff
@@ -1335,34 +1401,36 @@ export const commands: Chat.ChatCommands = {
...
-				` - Start a new lottery giveaway (voices can only host their own). Requires: + % @ # &`,
-			);
-			buf.push(`<code>/giveaway changequestion/changeanswer</code> - Changes the question/answer of a question giveaway. Requires: Being giveaway host`);
-			buf.push(`<code>/giveaway viewanswer</code> - Shows the answer of a question giveaway. Requires: Being giveaway host/giver`);
-			buf.push(`<code>/giveaway ban [user], [reason]</code> - Temporarily bans [user] from entering giveaways. Requires: % @ # &`);
-			buf.push(`<code>/giveaway end</code> - Forcibly ends the current giveaway. Requires: % @ # &`);
-			buf.push(`<code>/giveaway count [pokemon]</code> - Shows how frequently a certain Pok&eacute;mon has been given away.</details>`);
+			buf.push(<details><summary>Staff commands</summary>
+				<code>/giveaway create</code> - Pulls up a page to create a giveaway. Requires: + % @ # &amp;
+				<br /><code>
```
Zarel, line 1406: We tend to end lines with `<br />` rather than starting them.

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1442,454 @@ function makePageHeader(user: User, pageid?: string) {
...
-		stored: `<i class="fa fa-paste"></i>`,
-		'stored-add': `<i class="fa fa-paste"></i>`,
-		submitted: `<i class="fa fa-inbox"></i>`,
-		'submitted-add': `<i class="fa fa-inbox"></i>`,
+	const icons: Record<keyof typeof titles, Chat.VNode> = {
+		create: <i className="fa fa-sticky-note"></i>,
+		stored: <i className="fa fa-paste"></i>,
+		'stored-add': <i className="fa fa-paste"></i>,
+		submitted: <i className="fa fa-inbox"></i>,
+		'submitted-add': <i className="fa fa-inbox"></i>,
```
Zarel, line 1450: We use `class`, not `className`.

```suggestion
		create: <i class="fa fa-sticky-note"></i>,
		stored: <i class="fa fa-paste"></i>,
		'stored-add': <i class="fa fa-paste"></i>,
		submitted: <i class="fa fa-inbox"></i>,
		'submitted-add': <i class="fa fa-inbox"></i>,
```

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1442,454 @@ function makePageHeader(user: User, pageid?: string) {
...
 	};
-	let buf = `<button class="button" style="float:right" name="send" value="/j view-giveaways${pageid?.trim() ? `-${pageid.trim()}` : ''}"><i class="fa fa-refresh"></i> Refresh</button>`;
-	buf += `<h1>Wi-Fi Giveaways</h1>`;
+	const buf = [];
+	buf.push(<button className="button" style={{float: 'right'}} name="send" value={
```
Zarel, line 1453: Can you fix all the other `className`s too? They should all be `class`.

server/chat-plugins/wifi.tsx
KrisXV: Like I’ve asked/implied before, please make some documentation for this. I asked about exactly this in the discord and you didn’t answer me, and now you’re telling me this in a tone like I’m supposed to know better.

server/chat-plugins/wifi.tsx
KrisXV: I’ll fix this, but this is done so inconsistently throughout the sim that there was no way to know this.

server/chat-plugins/wifi.tsx
KrisXV: Yeah, that was an oversight.

server/chat-plugins/wifi.tsx
KrisXV: What do you mean?

server/chat-jsx.tsx
KrisXV: I guess that works.

server/chat-plugins/wifi.tsx
KrisXV: Yes

server/chat-jsx.tsx
KrisXV: Some of the examples I found online used divs so I went with those, I’ll switch to span if thats preferred though

server/chat-plugins/wifi.tsx
Zarel: That's very unfortunate. I've always broken lines after `<br />`. Maybe I'll need to add it to the style guide.

server/chat-plugins/wifi.tsx
Zarel: Sorry, it was meant to be neutral, but I did answer when you asked on Discord:

image

We use `class` because we use Preact and Preact uses `class`.

server/chat-plugins/wifi.tsx
Zarel:
```suggestion
		if (this.game === 'BDSP') css.style = {background: '#aa88a9', color: '#fff'};
```

In general, when writing CSS, it's useful to set foreground color at the same time as background color. While in this case, it's unlikely to matter, it often works around CSS mistakes where you end up with white-on-white or black-on-black because you messed up inheritance somewhere.

KrisXV: That makes sense, though they use the same foreground hex, so does it matter here?

server/chat-plugins/wifi.tsx
KrisXV: The last line in that log was semi-implicit that you were unsure what the standard, so it confused me with what you meant.

server/chat-jsx.tsx
Zarel: I think ideally users should just directly use `dangerouslySetInnerHTML` so you don't have an extra span _or_ an extra div.

Zarel: But if you really want to use this architecture, `span` would let chat messages be inline. Which they normally are, anyway.

server/chat-plugins/wifi.tsx
Zarel: Yes, that's why I'm saying, in this case it's unlikely to be a problem. But you can imagine, for instance, server custom CSS which makes `broadcast-blue` a dark-on-light color scheme. Then this would be unreadable if you overrode `background` but left `color` unchanged.

server/chat-plugins/wifi.tsx
Zarel: Yeah, I'm unsure about how Preact handles `for`. But I'm sure Preact prefers `class` over `className`.

server/chat-plugins/wifi.tsx
Zarel: (I also prefer `background` over `backgroundColor`. It makes it clearer that you want the entire background to be that color, rather than also having a background image, and it's also shorter.)

server/chat-jsx.tsx
Zarel: This does not seem resolved.

KrisXV: I am not throwing on arrays because components can take arrays as children

server/chat-plugins/wifi.tsx
```diff
@@ -280,22 +288,40 @@ class Giveaway extends Rooms.RoomGame {
...
+					</td>
+					<td style={{textAlign: 'center', width: '45%'}}>{rightSide}</td>
+				</tr>
+				{this.extraInfo?.trim().length ?
+					<tr>
+						<td colSpan={2} style={{textAlign: 'center'}}>
+							<strong>Extra Information</strong><br />
+							<Chat.JSX.FormatText isTrusted>{this.extraInfo.trim().replace(/<br \/>/g, '\n')}</Chat.JSX.FormatText>
+						</td>
+					</tr> : ''}
```
Zarel, line 316:
```suggestion
				{!!this.extraInfo?.trim().length && <tr>
					<td colSpan={2} style={{textAlign: 'center'}}>
						<strong>Extra Information</strong><br />
						<Chat.JSX.FormatText isTrusted>{this.extraInfo.trim().replace(/<br \/>/g, '\n')}</Chat.JSX.FormatText>
					</td>
				</tr>}
```

`&&` is the idiomatic way to do a conditional in JSX.

server/chat-plugins/wifi.tsx
```diff
@@ -733,18 +765,35 @@ export class GTS extends Rooms.RoomGame {
...
-				'') +
-			`<td style="text-align:center;width:15%">${this.sprite}</td><td style="text-align:center;width:${40 - sentModifier}%">${Chat.formatText(this.summary, true)}</td>` +
-			`<td style="text-align:center;width:${35 - sentModifier}%">${rightSide}</td></tr></table>`;
+			<strong>
+				More Pok&eacute;mon have been deposited than there are prizes in this giveaway and new deposits will not be accepted. {}
+				If you have already deposited a Pok&eacute;mon, please be patient, and do not withdraw your Pok&eacute;mon.
+			</strong> : <>
+				To participate, deposit <strong>{this.deposit}</strong> into the GTS and look for <strong>
+					{Utils.escapeHTML(this.lookfor)}
+				</strong>
```
Zarel, line 774:
```suggestion
				To participate, deposit <strong>{this.deposit}</strong> into the GTS and look for <strong>{this.lookfor}</strong>
```

I'm guessing the double-escape was unintentional?

server/chat-plugins/wifi.tsx
```diff
@@ -733,18 +765,35 @@ export class GTS extends Rooms.RoomGame {
...
+			</>;
+		return <>
+			<p style={{textAlign: 'center', fontSize: '14pt', fontWeight: 'bold', marginBottom: '2px'}}>
+				There is a GTS giveaway going on!
+			</p>
+			<p style={{textAlign: 'center', fontSize: '10pt', marginTop: 0}}>
+				Hosted by: {Utils.escapeHTML(this.giver.name)} | Left: <b>{this.left}</b>
```
Zarel, line 781:
```suggestion
				Hosted by: {this.giver.name} | Left: <b>{this.left}</b>
```

server/chat-plugins/wifi.tsx
```diff
@@ -280,22 +288,39 @@ class Giveaway extends Rooms.RoomGame {
...
+					</td>
+				</tr>
+				<tr>
+					<td style={{textAlign: 'center', width: '45%'}}>
+						<psicon item={this.ball} /> {this.sprite} <psicon item={this.ball} />
+						<Chat.JSX.FormatText isTrusted>{set}</Chat.JSX.FormatText>
+					</td>
+					<td style={{textAlign: 'center', width: '45%'}}>{rightSide}</td>
+				</tr>
+				{this.extraInfo?.trim().length && <tr>
```
Zarel, line 310:
```suggestion
				{!!this.extraInfo?.trim().length && <tr>
```

`0` in JSX is `0`; `false` in JSX is blank; that's why my suggestion included a cast.

server/chat-jsx.tsx
Zarel: Yes, but it's a bug if they do. That's why this should throw. The purpose of throwing is to tell you that your code has a bug.

server/chat-plugins/wifi.tsx
KrisXV: oh, i didnt see the !! in the suggestion mb

server/chat-jsx.tsx
Zarel: I do think taking the main text in a `text` attribute rather than in `children` would be better, if this is a problem.

KrisXV: That's fair, will change. I don't think we need to use a text attribute for now, but that could change.

Endymiona:
Jcjcjvkvobobn

server/chat-jsx.tsx
```diff
@@ -36,3 +37,16 @@ export type VNode = preact.VNode;
...
+
+export class FormatText extends preact.Component<{isTrusted?: boolean, replaceLinebreaks?: boolean}> {
+	render() {
+		const child = this.props.children;
+		if (typeof child === 'string') {
+			return <span dangerouslySetInnerHTML={
+				{__html: Chat.formatText(child, this.props.isTrusted, this.props.replaceLinebreaks)}
+			} />;
+		}
+		if (preact.isValidElement(child)) return child;
```
Zarel, line 49: Why isn't this a throw? This seems like this should throw.

Zarel: Like if you use `FormatText` and you don't use it to format your text, that clearly seems like a bug.

KrisXV: That’s fair

KrisXV: did this btw @Zarel

server/chat-jsx.tsx
```diff
@@ -28,6 +28,7 @@ export interface PSElements extends preact.JSX.IntrinsicElements {
 	spotify: {src: string};
 	username: {name?: string, class?: string};
 	psicon: {pokemon: string} | {item: string} | {type: string} | {category: string};
+	center: Record<string, unknown>;
```
Zarel, line 31: ಠ_ಠ

Zarel:
```suggestion
	center: {};
```

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1439,457 @@ function makePageHeader(user: User, pageid?: string) {
...
+											<h3 style={{textAlign: 'center'}}>Lottery</h3>
+											<hr />
+											<strong>Game:</strong> {gameName[giveaway.game]}, <strong>Giver:</strong> {giveaway.targetUserID},
+											&#32;<strong>OT:</strong> {giveaway.ot}, <strong>TID:</strong> {giveaway.tid},
+											&#32;<strong># of winners:</strong> {giveaway.winners}
+											{!!giveaway.claimed && <><br /><strong>Claimed:</strong> {giveaway.claimed}</>}<br />
+											<strong>Pok&eacute; Ball:</strong> <psicon item={giveaway.ball} />
+											<details>
+												<summary><psicon pokemon={giveaway.prize.species} /> Prize</summary>
+												{Giveaway.convertIVs(giveaway.prize, giveaway.ivs).replace(/\n/g, '<br />')}
```
Zarel, line 1770: This is not going to do what you think it does.

Zarel: The replacer replaces linebreaks with the escaped text `<br />`, not an actual `br` element.

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1439,457 @@ function makePageHeader(user: User, pageid?: string) {
...
+												</details>
+											</>}
+										</>;
+									} else {
+										giveaway = giveaway as QuestionGiveawayData;
+										return <>
+											<h3 style={{textAlign: 'center'}}>Question</h3>
+											<hr />
+											<strong>Game:</strong> {gameName[giveaway.game]}, <strong>Giver:</strong> {giveaway.targetUserID},
+											&#32;<strong>OT:</strong> {giveaway.ot}, <strong>TID:</strong> {giveaway.tid}
```
Zarel, line 1786:
```suggestion
											<strong>Game:</strong> {gameName[giveaway.game]}, <strong>Giver:</strong> {giveaway.targetUserID}, {}
											<strong>OT:</strong> {giveaway.ot}, <strong>TID:</strong> {giveaway.tid}
```

Zarel: There are like five other occurrences of `&#32;` btw, which should all be fixed.

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1439,457 @@ function makePageHeader(user: User, pageid?: string) {
...
-						}
-						buf += `</div>`;
+						return <><h2>Store a Giveaway</h2>
+							{(() => {
+								if (!type || !['question', 'lottery'].includes(type)) {
+									return <center>
+										<h3>Pick a giveaway type</h3>
+										{
+											formatFakeButton(`/view-giveaways-stored-add-lottery`, <><i class="fa fa-random"></i> Lottery</>)
+										} |  {
```
Zarel, line 1671: I'm guessing this was a typo?

KrisXV: yes

server/chat-plugins/wifi.tsx
KrisXV: this was a remnant of old code that I forgot to fix, i didn't intentionally do this

server/chat-jsx.tsx
KrisXV: the face doesn't really help with explaining why it's wrong, but i am doing this

KrisXV: Also, `{}` isn't right either; I get this error:

`31:10  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".`

Zarel: Unfortunately, in this case, it is in fact a correct value. Probably changing it to `{class?: string}` will shut up the linter, though.

Zarel: The face is because `Record<string, unknown>` is nothing like the other entries, like `{src: string}`, which are lists of attributes. If you are going to do something like this (which seems intended to remove all validation), you should at least ask me if it's a good idea.

In general, I would prefer that if you are going to shut up a validator of some sort in the PS codebase, you should first ask me if there is a better alternative.

KrisXV: Is that not what the point of a pull request is?

KrisXV: Regardless, fixing

Zarel: A pull request is an opportunity for people to catch your mistakes, but you shouldn't rely on the idea that I will carefully read every single line, especially of longer pull requests. If you intentionally do something dangerous, you should comment on that part of the PR to ask for special attention (such as asking me if there are better ways).

I do that on my pull requests, you know; I'll often point out specific lines where I'm doing weird things.

server/chat-plugins/wifi.tsx
```diff
@@ -1374,317 +1439,457 @@ function makePageHeader(user: User, pageid?: string) {
...
+								if (!type || !['question', 'lottery'].includes(type)) {
+									return <center>
+										<h3>Pick a giveaway type</h3>
+										{
+											formatFakeButton(`/view-giveaways-submitted-add-lottery`, <><i class="fa fa-random"></i> Lottery</>)
+										} | {
+											formatFakeButton(`/view-giveaways-submitted-add-question`, <><i class="fa fa-question"></i> Question</>)
+										}
+									</center>;
+								} else {
```
Zarel, line 1843: Weird that eslint didn't catch this, but the `else` is unnecessary.

KrisXV: did

Zarel:
Thanks!

PR #8583, opened 2021-12-23
Convert Max Guard to a whitelist

DaWoblefet:
Thanks to new research from Anubis, we now know that Max Guard is treated more like a whitelist rather than a "allow all moves that bypass Protect except these" type of check. Because we're pulling this data from BDSP, though, our data doesn't include any of the DLC moves; however, empirical testing helps bridge that gap. This is the list from BDSP:

> 'acupressure', 'afteryou', 'allyswitch', 'aromatherapy', 'aromaticmist', 'confide', 'copycat', 'curse', 'decorate', 'feint', 'futuresight', 'healbell', 'holdhands', 'howl', 'lifedew', 'meanlook', 'perishsong', 'playnice', 'powertrick', 'roar', 'roleplay', 'tearfullook',

These are the additional moves known from empirical testing post-DLC that also bypass Max Guard:

> 'coaching', 'doomdesire', 'gmaxoneblow', 'gmaxrapidflow', 'junglehealing', 

PS's current implementation allows a number of moves that bypass Max Guard that shouldn't, including moves like Hyperspace Fury or Helping Hand, so this change also fixes up those.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10420,16 +10420,10 @@ export const Moves: {[moveid: string]: MoveData} = {
 			},
 			onTryHitPriority: 3,
 			onTryHit(target, source, move) {
-				if (['gmaxoneblow', 'gmaxrapidflow'].includes(move.id)) return;
-				/** moves blocked by Max Guard but not Protect */
-				const overrideBypassProtect = [
-					'block', 'flowershield', 'gearup', 'magneticflux', 'phantomforce', 'psychup', 'shadowforce', 'teatime', 'transform', 'whirlwind',
+				const bypassesMaxGuard = [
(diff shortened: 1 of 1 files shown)


DaWoblefet:
This implementation should work fine, but I'm not sure if we should maybe convert this to a move property (bypassesMaxGuard or something similar) rather than being a literal hardcoded list, since 27 entries is a decent amount.

AnnikaCodes:
If it's going to be a hardcoded list, it should probably be a top-level const?

KrisXV:
> If it's going to be a hardcoded list, it should probably be a top-level const?

Why would we make a top level const for something exclusive to one move?

DaWoblefet:
I don't think we'd want to do it that way; the current approach mirrors Metronome's "do not call" list.

Endymiona:
Kskskskkdjxxjd

PR #8585, opened 2021-12-24
BDSP: Make DLC abilities nonstandard

Karthik99999:
https://twitter.com/DaWoblefet/status/1471323080437993474


KrisXV:
Why not just mark them as nonstandard?

Karthik99999:
Talked to DaWoblefet and apparently the abilities should work in custom games so I'm now marking them as nonstandard

WeWuzNidokangz:
If these abilities are made nonstandard, they can no longer pass the BDSP Pure Hackmons validator, but the video in the link makes it look as if they should be able to be brought into battle in that kind of environment, even if they don't function the same as in SS. Maybe I am missing some more recent information but the original change seems more accurate...?

DaWoblefet:
> If these abilities are made nonstandard, they can no longer pass the BDSP Pure Hackmons validator, but the video in the link makes it look as if they should be able to be brought into battle in that kind of environment, even if they don't function the same as in SS. Maybe I am missing some more recent information but the original change seems more accurate...?

I think it's analogous to dexited moves; even though you can bring, say, Hyperspace Fury, it doesn't mean it's gonna work the way you think it will. Pure Hackmons isn't losing out on anything here because Run Away would be an equivalent.

WeWuzNidokangz:
It does seem like these abilities would work similarly to Run Away here, but even if there was no functional difference between them, it seems to me that if they can be hacked into local battles, they fall into the scope of what should be accessible in PH and should still be supported options for reasons of accuracy and completeness in the same way that setting Illuminate rather than Run Away is supported even though it doesn't have a different in-battle effect. (It seems possible that there might be some very marginal mechanical implications as well, e.g. if BDSP Unseen Fist is hacked onto a set, then an opponent with Run Away presumably should be able to successfully Entrain it and not have Stomping Tantrum's BP boosted on the next turn, whereas Entrainment would fail if both had Run Away.)

That said, thank-you, I didn't understand that it was possible to actually bring dexited moves into a Sw/Sh battle before... It does seem correct to say that these are mostly analogous situations, but from what I can understand after looking into that more, it seems as if making dexited moves able to pass the validator in Sw/Sh hackmons enviroments but unclickable like Gen 7 Hyperspace Fury when hacked onto a species other than Hoopa Unbound might be a more accurate way of resolving that inconsistency(?)

KrisXV:
That seems like pointless work when simply marking them as nonstandard does the trick and allows for even more compatibility. If people want to run an ability that works like Run Away in BH or PH, they will run exactly that ability. Or Illuminate. If they want to run a useless move, they'll run Splash.

PR #8587, opened 2021-12-25
Gen 1: Fix Haze

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8973699

I'm not completely sure this is correct, since gen 1 code is confusing af, but it seemed to work fine when I tested the scenario given in the bug report.


Marty-D (approved):
Can't be worse than the existing code, thanks!

PR #8591, opened 2021-12-29
Fix interaction of failed Z heal replacements

DaWoblefet:
Z-Parting Shot and Z-Memento should succeed in healing their targets even if the move fails for some reason (e.g. Clear Body, Substitute). This change directly adds the side condition, rather than replacing it as the current active move (?) which could cause it to fail early and not use the Z-Power when it should always be doing that.

### DIFF
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -1435,7 +1435,7 @@ export class BattleActions {
 				this.battle.heal(pokemon.maxhp, pokemon, pokemon, zPower);
 				break;
 			case 'healreplacement':
-				pokemon.side.addSlotCondition(pokemon, 'healreplacement', pokemon);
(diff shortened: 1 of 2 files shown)


sim/battle-actions.ts
```diff
@@ -1435,7 +1435,7 @@ export class BattleActions {
 				this.battle.heal(pokemon.maxhp, pokemon, pokemon, zPower);
 				break;
 			case 'healreplacement':
-				move.self = {slotCondition: 'healreplacement'};
+				pokemon.side.addSlotCondition(pokemon, 'healreplacement', pokemon);
```
urkerab, line 1438: I'm not sure whether this will set the source effect correctly, so you might need to append `, move` here.

DaWoblefet: Good catch!

PR #8592, opened 2021-12-30
Gen 3-4: Fix Snatch when used by multiple pokemon

Karthik99999:
According to Bulbapedia and the long description of the move here on showdown, Snatch should be able to steal moves that were called by another user of Snatch.

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -1433,6 +1433,27 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			return !pokemon.volatiles['choicelock'] && !pokemon.volatiles['encore'];
 		},
 	},
+	snatch: {
+		inherit: true,
+		condition: {
+			duration: 1,
+			onStart(pokemon) {
+				this.add('-singleturn', pokemon, 'Snatch');
+			},
+			onAnyPrepareHitPriority: -1,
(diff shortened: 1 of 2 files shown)


data/mods/gen4/moves.ts
```diff
@@ -1433,6 +1433,27 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+		inherit: true,
+		condition: {
+			duration: 1,
+			onStart(pokemon) {
+				this.add('-singleturn', pokemon, 'Snatch');
+			},
+			onAnyPrepareHitPriority: -1,
+			onAnyPrepareHit(source, target, move) {
+				const snatchUser = this.effectState.source;
+				if (snatchUser.isSkyDropped()) return;
```
DaWoblefet, line 1446: This would be unnecessary since it's pre-Sky Drop.

Karthik99999: I saw other code check future gen stuff (specifically in the gen 4 moves files, uproar checks for throat chop), which is why I left the future gen stuff in.

DaWoblefet: I suppose it's fine if someone uses Sky Drop in Gen 4 custom game or smth. It can be left in.

Marty-D (approved):
Great, thanks!

PR #8593, opened 2021-12-30
Extend /repeat multiline pattern for all commands

FeluciaPS:
I figured instead of writing down all 6 command variations I'd just use regex for this one


AnnikaCodes (approved):
Looks good, thanks!

PR #8594, opened 2021-12-30
Use uhtml for repeats to prevent flooding

FeluciaPS:
I did not test this code, but I did check to make sure I got the uhtml protocol right (which I did). The only way this could go wrong is if somehow the `repeat#id` property doesn't exist.


AnnikaCodes:
Looks good, thanks!

PR #8595, opened 2021-12-30
Ban Blacephalon from AAA

TomOfTomKat:
Blacephalon was recently banned from AAA by suspect test, and the ban needs to be applied to the ladder. Proof of ban can be found here: https://www.smogon.com/forums/threads/ss-aaa-blacephalon-voting.3695298/page-2#post-9065969


AnnikaCodes:
Looks good, thanks!

PR #8598, opened 2022-01-01
Move Ninjask to PU

Distrib-ps:
https://www.smogon.com/forums/threads/pu-general-rules-tiering-and-announcements-thread.3660650/post-9067505


KrisXV:
I did the rest of the tier shifts with this change

PR #8599, opened 2022-01-01
Pokébilities: Ban Pheromosa

VineST:


KrisXV:
thanks for catching that

VineST:
i didnt ping you in the thread so it's totally understandable; just glad i got it in before hotpatch

PR #8602, opened 2022-01-02
Fix:red card shell bell interaction bug

chouettevan:
hope it works


Marty-D:
It doesn't; hardcoding is bad for many reasons. You're leaving out cases where Red Card doesn't activate because it's being ignored, or can't force out the holder of Life Orb/Shell Bell because it's the last Pokemon, just as some examples. The correct solution is to alter the order the events happen.

PR #8603, opened 2022-01-02
BD/SP Random Battle: Update movepools

Tarrembeau:


data/mods/gen8bdsp/formats-data.ts
```diff
@@ -227,7 +227,7 @@ export const FormatsData: {[k: string]: ModdedSpeciesFormatsData} = {
 	persian: {
 		tier: "RU",
 		doublesTier: "DUU",
-		randomBattleMoves: ["fakeout", "slash", "uturn", "bite", "aerialace"],
+		randomBattleMoves: ["fakeout", "covet", "uturn", "bite", "aerialace"],
```
AnnikaCodes, line 230: Is this really better? Is Persian without an item so often that it's worth giving up the extra damage from Slash?

Karthik99999: Covet gets boosted by Technician

AnnikaCodes: Oh, right, I forgot about Technician. Sure, that's an improvement then.

AnnikaCodes (approved):
Looks good, thanks!

PR #8606, opened 2022-01-03
Abuse-monitor: Support scaling thresholds per-room

mia-pi-git:
Feature requested by Aeonic to handle longer battles (since they can hit few false positives over many turns and cause accidental flags).

### DIFF
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -699,14 +699,23 @@ export const commands: Chat.ChatCommands = {
 			if (!toID(target)) {
 				return this.parse(`/help am`);
 			}
-			const [turns, increment, min] = Utils.splitFirst(target, ',', 2).map(r => parseInt(toID(r)));
+			const [rawTurns, rawIncrement, rawMin] = Utils.splitFirst(target, ',', 2).map(toID);
+			const turns = parseInt(rawTurns);
 			if (isNaN(turns) || turns < 0) {
 				return this.errorReply(`Turns must be a number above 0.`);
 			}
+			const increment = parseInt(rawIncrement);
 			if (isNaN(increment) || increment < 0) {
 				return this.errorReply(`The increment must be a number above 0.`);
 			}
-			settings.thresholdIncrement = {amount: increment, turns, minTurns: min || 0};
+			const min = parseInt(rawMin);
+			if (rawMin && isNaN(min)) {
+				return this.errorReply(`Invalid minimum (must be a number).`);
+			}
+			settings.thresholdIncrement = {amount: increment, turns};
+			if (min) {
+				settings.thresholdIncrement.minTurns = min;
+			}
 			saveSettings();
 			this.privateGlobalModAction(
 				`${user.name} set the abuse-monitor threshold increment ${increment} every ${Chat.count(turns, 'turns')}` +
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -51,6 +51,7 @@ export const cache: {
 
 const defaults: FilterSettings = {
 	threshold: 4,
+	thresholdIncrement: {turns: 5, amount: 1},
 	minScore: 0.65,
 	specials: {
 		THREAT: {0.96: 'MAXIMUM'},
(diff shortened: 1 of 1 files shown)


server/chat-plugins/abuse-monitor.ts
```diff
@@ -674,8 +687,35 @@ export const commands: Chat.ChatCommands = {
...
+			if (incr.minTurns) buf += ` after turn ${incr.minTurns}`;
 			this.sendReplyBox(buf);
 		},
+		ti: 'thresholdincrement',
+		thresholdincrement(target, room, user) {
+			checkAccess(this);
+			if (!toID(target)) {
+				return this.parse(`/help am`);
+			}
+			const [turns, increment, min] = Utils.splitFirst(target, ',', 2).map(r => parseInt(toID(r)));
```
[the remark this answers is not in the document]
mia-pi-git: Idiot-proofing.

server/chat-plugins/abuse-monitor.ts
```diff
@@ -674,8 +687,35 @@ export const commands: Chat.ChatCommands = {
...
+			if (isNaN(turns) || turns < 0) {
+				return this.errorReply(`Turns must be a number above 0.`);
+			}
+			if (isNaN(increment) || increment < 0) {
+				return this.errorReply(`The increment must be a number above 0.`);
+			}
+			settings.thresholdIncrement = {amount: increment, turns, minTurns: min || 0};
```
AnnikaCodes, line 709: Why set `minTurns` to 0 if there's no value specified? Shouldn't it just not be included (a la `if (min) settings.thresholdIncrement.min = min`)?
Also, what if someone specifies a non-numeric value for `min` in the command?

server/chat-plugins/abuse-monitor.ts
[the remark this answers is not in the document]
AnnikaCodes: It's already "idiot-proofed" with the `if (isNaN(...))` checks, is it not?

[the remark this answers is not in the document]
mia-pi-git: Yes, but this adds a bit more convenience by cutting off extra stuff beforehand.

server/chat-plugins/abuse-monitor.ts
AnnikaCodes: Then why type it as an optional property at all?

mia-pi-git: > Also, what if someone specifies a non-numeric value for min in the command?

```ts
> NaN || 0
0
```

mia-pi-git: > Then why type it as an optional property at all?

Doesn't really matter either way, idt. I had originally planned to make it not defined, but then realized this made the code a lot simpler.

AnnikaCodes: If someone makes a typo (say "3p" instead of "30"), it should not silently default to 0 — it should display an error message so that they realize their mistake and can retype it as "30" (or whatever they meant).

mia-pi-git: OK.

AnnikaCodes: > > Then why type it as an optional property at all?
> 
> Doesn't really matter either way, idt. I had originally planned to make it not defined, but then realized this made the code a lot simpler.

I don't like the idea of having two ways to say "no minimum turn value" - both `undefined` and `0` work. IMO the former has clearer typing while the latter has simpler (thus more readable) code. Either is fine, but you should pick one and only use it.

server/chat-plugins/abuse-monitor.ts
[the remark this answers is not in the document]
Zarel: Idiot-proofing should be about giving better error messages, not error suppression. If someone types in "12+2", you should be giving an error "no you can't do that", not to interpret it as "122".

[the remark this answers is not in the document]
Zarel: This was the lesson of JavaScript and the turn against Postel's Law: the idea that we learned, that we should, on bad input, give an error message, not try to guess what the user meant and get it wrong.

PR #8607, opened 2022-01-03
Hackmons Cup: Support custom team bans/unbans

WeWuzNidokangz:
Related to https://www.smogon.com/forums/threads/supporting-custom-banlists-for-random-battles.3688882/
and https://github.com/smogon/pokemon-showdown/pull/8427

As well as doing what it says in the title, this PR also makes some changes to custom challenge and tour validation to allow bans to be used in custom battles with generated teams in general. (Perhaps this should be 2 PRs, and if so I could try to split these things afterwards if advised, but I don't know how to handle dependencies between PRs.)

For the validation changes, since the general block on that in dex-formats.ts is lifted here, a similar error message (enforceNoDirectCustomBanlistChanges) is added to team generator methods that don't support custom bans yet so that users understand in which formats they are effective. This works similarly to the previous general method, i.e. it doesn't count bans that are nested in other formats or rules such as Camomons, so that it's still possible to add Camomons battle effects to generated team formats that don't support custom bans yet. (If those formats/rules are added to Hackmons Cup AND there are direct user-specified bans as well, the bans inside them will be included in the rule table and should affect team generation, which is a bit unintuitive and not ideal but seems to me to be the least bad way of handling this until all OM behavior that doesn't require a mod or custom validator can be encapsulated in rules independent of balance bans.)

During generation, early checks are made that should reliably prevent battles starting if custom bans mean that set pools sizes may be insufficient to build the ordered team/move size (this is important because if this cannot be guaranteed, other rules or conditions that require specific minimum team sizes, e.g. Chimera, may be affecting the battle and might crash it afterwards even if the team generator wouldn't crash by itself if it couldn't meet these constraints; this would be a particular problem in tours, where features like this are often used, as a random crash in one battle could break the entire tour).

I looked into supporting complex bans but it would probably end up needing to make the generator almost completely separate from the regular one as well as being slow, difficult to test and having limited utility so there is currently no support for that here. An error is shown if a user attempts to use complex bans.

One other thing I thought about was trying to merge the Gen 1 HC generator into the main one so that it could benefit from these changes as well (just branching on things like DVs, etc). I might try to do that in a later PR if it is likely to be accepted(?)


data/random-teams.ts
```diff
@@ -277,6 +280,53 @@ export class RandomTeams {
...
+		}
+		return false;
+	}
+
+	/**
+	 * Inform user when custom bans are unsupported in a team generator.
+	 */
+	protected enforceNoDirectCustomBanlistChanges() {
+		if (!this.hasDirectCustomBanlistChanges()) return;
+		throw new Error(`Custom bans are not currently supported in ${this.format.name}.`);
```
AnnikaCodes, line 302:
```suggestion
		if (this.hasDirectCustomBanlistChanges()) {
			throw new Error(`Custom bans are not currently supported in ${this.format.name}.`);
		}
```
Typically PS code formats these kinds checks like so; you don't need to rewrite the code that's already here (unless you want to), but next time you contribute this would be good to remember.

data/random-teams.ts
```diff
@@ -497,48 +549,267 @@ export class RandomTeams {
 		return nPokemon;
 	}
 
+	customRandomNPokemon(ruleTable: RuleTable, n: number, requiredType?: string) {
```
AnnikaCodes, line 552: There's a lot of repeated code here - why make a separate function rather than just adding an optional `ruleTable` parameter to `randomNPokemon` and thinning the pool based on that if it's specified?

data/random-teams.ts
```diff
@@ -497,48 +549,267 @@ export class RandomTeams {
...
+								if (hasAllNaturesBan) continue;
+								if (nature.isNonstandard) {
+									banReason = ruleTable.check('pokemontag:' + toID(nature.isNonstandard));
+									if (banReason) continue;
+									if (banReason !== '') {
+										if (nature.isNonstandard !== 'Unobtainable') {
+											if (hasNonexistentBan) continue;
+											if (!hasNonexistentWhitelist) continue;
+										}
+									}
```
AnnikaCodes, line 765:
```suggestion
									if (banReason !== '' && nature.isNonstandard !== 'Unobtainable') {
										if (hasNonexistentBan) continue;
										if (!hasNonexistentWhitelist) continue;
									}
```
In general, we prefer loops with `&&` conditions (split over multiple lines if needed) to nested loops, especially in cases like this where putting both conditions on the same line doesn't impact readability. (This applies to your other nested loops too.)

server/tournaments/index.ts
```diff
@@ -214,13 +215,23 @@ export class Tournament extends Rooms.RoomGame {
 	}
 
 	setCustomRules(rules: string) {
+		let format;
 		try {
 			this.fullFormat = Dex.formats.validate(`${this.baseFormat}@@@${rules}`);
+			format = Dex.formats.get(this.fullFormat, true);
+
+			// In tours of formats with generated teams, custom rule errors should be checked for here,
+			// since users can't edit their teams to avoid them at matching time
```
AnnikaCodes, line 224: Good catch!

AnnikaCodes:
I have a few style and code organization nitpicks, but your overall idea seems pretty solid. I appreciate this; custom rules for randomized formats are pretty cool and have been on my radar at a low level for a while. This could probably also be expanded to Challenge Cup in the future if there's demand.

data/random-teams.ts
WeWuzNidokangz: Thanks, I'll try to keep that in mind.

data/random-teams.ts
WeWuzNidokangz: I wasn't sure how much that was going to diverge, but I think you're right, it's unnecessary to make it a separate method now, so I merged them together. I also wasn't sure if the minSourceGen parameter was still necessary: I have kept it in the merged randomNPokemon but it doesn't seem to be referenced anywhere.

AnnikaCodes: I think I might've used in an implementation of a custom rule which became infeasible for other reasons? It's not used in private code, though; if it's not referenced anywhere it should be safe to remove.

AnnikaCodes (approved):
Alright, this seems fine to me; thanks!

PR #8612, opened 2022-01-05
Chat: Properly typecheck Chat.runHandlers arguments

mia-pi-git:
This means we shouldn't have more unpleasant issues with changing the types of handlers & missing old uses/runHandler calls.

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -77,7 +77,6 @@ interface Handlers {
 	onRoomDestroy: (roomid: RoomID) => void;
 	onBattleEnd: (battle: RoomBattle, winner: ID, players: ID[]) => void;
 	onLadderSearch: (user: User, connection: Connection, format: ID) => void;
-	// Chat.runHandlers('BattleRanked', this, winnerid, [p1rating, p2rating], [p1id, p2id]);
 	onBattleRanked: (
 		battle: Rooms.RoomBattle, winner: ID, ratings: (AnyObject | null | undefined)[], players: ID[]
 	) => void;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -65,15 +65,22 @@ export interface AnnotatedChatCommands {
 	[k: string]: AnnotatedChatHandler | string | string[] | AnnotatedChatCommands;
 }
 
-export interface Handlers {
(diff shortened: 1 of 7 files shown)


server/chat.ts
```diff
@@ -65,15 +65,22 @@ export interface AnnotatedChatCommands {
...
+	onRoomClose: (id: string, user: User, connection: Connection, page: boolean) => any;
+	onRenameRoom: (oldId: RoomID, newID: RoomID, room: BasicRoom) => void;
+	onBattleStart: (user: User, room: GameRoom) => void;
+	onBattleLeave: (user: User, room: GameRoom) => void;
+	onRoomJoin: (room: BasicRoom, user: User, connection: Connection) => void;
+	onDisconnect: (user: User) => void;
+	onRoomDestroy: (roomid: RoomID) => void;
+	onBattleEnd: (battle: RoomBattle, winner: ID, players: ID[]) => void;
+	onLadderSearch: (user: User, connection: Connection, format: ID) => void;
+	// Chat.runHandlers('BattleRanked', this, winnerid, [p1rating, p2rating], [p1id, p2id]);
```
AnnikaCodes, line 80: ?

mia-pi-git: Forgot to remove the comment when i was looking at it to write the typedef.

AnnikaCodes:
You'll need to update the `Disconnect` in users.ts (meaning this will also break on hotpatch).

AnnikaCodes:
But in general I like this change!

server/chat.ts
mia-pi-git: Removed with a commit to fix onDisconnect.

mia-pi-git:
> You'll need to update the `Disconnect` in users.ts (meaning this will also break on hotpatch).

To be clear, it won't crash hotpatch, or crash in general. It just won't work.

AnnikaCodes:
Yes, but in my book, breaking handlers is detrimental enough to merit waiting for a restart (especially since this change is mostly useful when refactoring handlers, which can be put off for a few weeks).

mia-pi-git:
Fair. Just wanted to clarify for anyone else reading.

AnnikaCodes (approved):
This looks good! If you've tested it locally it should be good to merge.

PR #8613, opened 2022-01-05
SQL: Propagate errors to the parent process

mia-pi-git:
This propagates errors to the parent process so promises properly reject. 
In doing so, it also fixes an issue with modlog creating tables several times accidentally in tests (which failed them).


AnnikaCodes:
Seems fine.

PR #8614, opened 2022-01-05
UNO: Fix the turn sequence when a user is disqualified

Kalest3:
The bug: The turn is always passed to the first in the player list when a user is disqualified from UNO


AnnikaCodes:
This doesn't seem obviously wrong, but I'm not very familiar with UNO code to review it more closely for bugs. @KrisXV or @mia-pi-git, are either of you familiar with the UNO plugin?

AnnikaCodes:
Hm, this seems fine. I'm not sure if your code that removes a +4 if the player it's used against is disqualified should be there or not, but I don't want to start debate on UNO rules, so I'm just going to merge this. Thanks!

PR #8615, opened 2022-01-07
Spotlights: Don't give an image parameter in source if no image was used

HisuianZoroark:
Prevents an undefined variable from being showcased in source if there is no image being used for the daily.

Before:
before

After:
after


mia-pi-git (approved):
Looks fine to me.

AnnikaCodes (approved):
LGTM; I'll merge when I have time to deploy changes.

PR #8616, opened 2022-01-07
1v1: Ban Power Construct

FeluciaPS:
https://www.smogon.com/forums/threads/ss-1v1-power-construct-voting.3695860/post-9075953


KrisXV:
ty

PR #8617, opened 2022-01-08
Fix Alternatium and Random Dex bugs

scoopapa:
Fixing some issues with PMOTS:
Vivillon not working due to client (renamed them to the real formes)
Spikes not working for reasons (reimplemented Power Core to bypass)
Wrong list of pokemon for Random Dex (RIP)
Past Moves being banned after adding 'Past' to banlist as part of banning all pokemon (fixed in Scripts.ts)


WeWuzNidokangz:
With the new Power Core implementation, its user can be poisoned by tspikes and will take damage from the poison.

You could override hazard conditions in a similar way to SSB to get what seems to have been the intended behavior from the previous moves.ts:-

https://github.com/smogon/pokemon-showdown/blob/7097a9ac753cbb295b484b97e2657898fd5a0932/data/mods/ssb/conditions.ts#L2408

scoopapa:
Thanks for catching that. Was able to fix it with an onSetStatus callback.

KrisXV:
ty!

PR #8619, opened 2022-01-08
Correct Max Move Base Power display

ACakeWearingAHat:
In Hackmons Cup, Max Moves have a functional Base Power of 1 but still display their old inaccurate Base Powers that they used to use. This is especially notable for Max Flare, which is NOT in fact 100 BP anymore. This provides misinformation that can lead to misplays on the part of players of Hackmons Cup.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -10300,7 +10300,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 	maxairstream: {
 		num: 766,
 		accuracy: true,
-		basePower: 10,
+		basePower: 1,
 		category: "Physical",
 		name: "Max Airstream",
 		pp: 10,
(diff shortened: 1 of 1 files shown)


Marty-D:
This is just as much misinformation since they're really 0 power. The data should not be changed; it's an implementation detail that they don't behave as their move data would otherwise indicate. If you want the display changed it should be done client-side.

Zarel:
Should we change them all to 0, then?

KrisXV:
That's not accurate either because their coded base powers are 10 (100 for max flare), it's just the damage calculation that acts as 0

DaWoblefet:
This was already done in https://github.com/smogon/pokemon-showdown/commit/d9e9864dbec4c3c977a3417705ef2e213c580f02; as said, the display is accurate to cart. Changing these to 1 BP doesn't really change anything because they aren't 1 BP either. You also can't just change them to 0 because 0 BP is interpreted as a no-damage move, when it's just 0 BP before entering the base damage formula.

PR #8620, opened 2022-01-09
Wi-Fi: Convert to TSX

KrisXV:
but actually this time


Zarel (approved):
Assuming this hasn't changed since last time, it was fine last time.

AnnikaCodes:
I don't know a ton about Preact but this seems fine and Zarel has also approved it, so I'm going to merge. Thanks!

PR #8624, opened 2022-01-12
Describe how non-dynamax max-moves work

William-Edward:
This could save hackmons players some confusion. Wasn't sure if it's 'non-dynamax' or 'without a base move' specifically. Both seem ambiguous in their own way, Also not even sure if 0 power is accurate.

### DIFF
--- a/data/text/moves.ts
+++ b/data/text/moves.ts
@@ -3484,27 +3484,27 @@ export const MovesText: {[k: string]: MoveText} = {
 	},
 	maxairstream: {
 		name: "Max Airstream",
-		desc: "Power is equal to the base move's Max Move power. If Max Airstream is used without Dynamaxing, its actual power is 0, but is displayed as 10. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute.",
+		desc: "Power is equal to the base move's Max Move power. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute. This effect does not happen if the user is not Dynamaxed. If this move is used as a base move, it deals damage with a power of 0.",
 		shortDesc: "Base move affects power. Allies: +1 Speed.",
 	},
 	maxdarkness: {
 		name: "Max Darkness",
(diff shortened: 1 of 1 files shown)


DaWoblefet:
They don't apply their Max Move effects either, so that deserves a mention (e.g. Max Flare won't set sun). They also aren't _technically_ 0 power, because if they were, they'd get bumped up to 1 base power by the minimum base power check. Also, a Dynamaxed hacked Max Move is still treated as 0 BP going into the damage formula, though that _will_ set Max Move effects (e.g. using Dynamaxed hacked Max Flare will set sun). Not sure what the most concise way to convey all that information is, though.

William-Edward:
The best I could come up with is for example:

"Power is equal to the base move's Max Move power. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute. If this move is used without a base move and without Dynamaxing, its actual power is 1, but is displayed as 10, and the move will have no additional effects. If there is no base move but the user is Dynamaxed, additional effects occur as normal, but the base power is still 1"

Is it worth saying the base power is 0 but will be calculated as 1?

data/text/moves.ts
```diff
@@ -3484,27 +3484,27 @@ export const MovesText: {[k: string]: MoveText} = {
 	},
 	maxairstream: {
 		name: "Max Airstream",
-		desc: "Power is equal to the base move's Max Move power. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute.",
+		desc: "Power is equal to the base move's Max Move power. If Max Airstream is used without Dynamaxing, its actual power is 0, but is displayed as 10. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute.",
```
Marty-D, line 3487:
```suggestion
		desc: "Power is equal to the base move's Max Move power. If this move is successful, the Speed of each Pokemon on the user's side is raised by 1 stage, even if they have a substitute. This effect does not happen if the user is not Dynamaxed. If this move is used as a base move, it deals damage with a power of 0.",
```
Same for all of them.

William-Edward (commented):
Worked i hope? I dont use github very often :p

Marty-D (approved):
Perfect, thanks!

PR #8626, opened 2022-01-12
[Gen 5] Random Battle: Fix (PU) Pokemons level

Tarrembeau:


AnnikaCodes:
Pokémon that are below PU by usage should probably have a high level than Pokémon deemed so strong they were _banned_ from PU?

Tarrembeau:
> Pokémon that are below PU by usage should probably have a high level than Pokémon deemed so strong they were _banned_ from PU?

Yes. Also we realized that PU mons don't have any level given and are still lvl 80 as well, so we need to fix it as well

PR #8627, opened 2022-01-12
BD/SP Random Battle: Update movepools

Tarrembeau:
This update is mainly made to tone down the impact of stealth rocks in bdsp rands, most notably by adding defog to all the pokémon who can and want to use it. We're also fixing Registeel in the process and changing Dustox and Tropius completely.


Tarrembeau:
Oops, where did I go wrong? :/

AnnikaCodes:
Looks good, thanks!

PR #8628, opened 2022-01-14
Allow bots to receive private mod actions

FeluciaPS:
This seems like a leftover from when bots were ranked above moderators, and nobody ever noticed that that piece of functionality changed. If this is intended then disregard my PR :)

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -328,7 +328,7 @@ export abstract class BasicRoom {
 		if (this.userCount) Sockets.roomBroadcast(this.roomid, message);
 	}
 	sendMods(data: string) {
-		this.sendRankedUsers(data, '%');
+		this.sendRankedUsers(data, '*');
 	}
 	sendRankedUsers(data: string, minRank: GroupSymbol = '+') {
 		if (this.settings.staffRoom) {


DaWoblefet:
Bots have access to modlog, so that makes sense to me.

server/rooms.ts
```diff
@@ -328,7 +328,7 @@ export abstract class BasicRoom {
 		if (this.userCount) Sockets.roomBroadcast(this.roomid, message);
 	}
 	sendMods(data: string) {
-		this.sendRankedUsers(data, '%');
+		this.sendRankedUsers(data, '*');
```
AnnikaCodes, line 331: This will also send private mod actions to Section Leaders. Is that okay?

KrisXV: They should already be getting them in theory, since anyone who has access to the staff room gets notified by this currently (which is a different bug, but still fine imo).

mia-pi-git: >  (which is a different bug, but still fine imo)

fwiw, not a bug - staff room access grants `user.isStaff`, which intentionally gives access to all private mod actions.

AnnikaCodes:
Looks good, thanks!

PR #8629, opened 2022-01-15
Implement secondaries stat chance overflow

DaWoblefet:
For stat increases, and plausibly stat drops, secondaries will overflow when both Serene Grace and Pledge Rainbow are simultaneously active. This is because the variable that stores the value is an unsigned 8-bit integer. The only move this affects is Charge Beam. See the [battle mechanics research thread](https://www.smogon.com/forums/posts/9076566/) for more details on its discovery.

In battle-actions, in addition to the ``secondaries`` function, there's another function called ``selfDrops`` that also checks for secondary chances as far as I can tell, but it seems to be unused throughout the codebase. I have no idea what it does but don't think it needs to be touched here.

### DIFF
--- a/test/sim/abilities/serenegrace.js
+++ b/test/sim/abilities/serenegrace.js
@@ -10,7 +10,7 @@ describe(`Serene Grace`, function () {
 		battle.destroy();
 	});
 
-	it(`should not stack Pledge Rainbow for flinches`, function () {
+	it(`should not stack with Pledge Rainbow for flinches`, function () {
 		// hardcoded RNG seed to not flinch if it was 60% odds
 		battle = common.createBattle({gameType: 'doubles', seed: [1, 2, 3, 5]}, [[
 			{species: 'wynaut', ability: 'serenegrace', moves: ['bite', 'waterpledge']},
--- a/test/sim/abilities/serenegrace.js
+++ b/test/sim/abilities/serenegrace.js
@@ -38,8 +38,13 @@ describe(`Serene Grace`, function () {
 
 		// Modding secondary chances so it will either always apply or never apply
 		// (64 * 2 * 2) % 256 === 0
-		Dex.moves.get('poweruppunch').secondaries[0].chance = 64;
-		Dex.moves.get('acidspray').secondaries[0].chance = 64;
+		battle.onEvent('ModifyMove', battle.format, 1, function (move) {
+			if (move.secondaries) {
+				for (const secondary of move.secondaries) {
+					secondary.chance = 64;
+				}
+			}
+		});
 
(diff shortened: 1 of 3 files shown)


test/sim/abilities/serenegrace.js
```diff
@@ -0,0 +1,75 @@
...
+const assert = require('./../../assert');
+const common = require('./../../common');
+
+let battle;
+
+describe(`Serene Grace`, function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
```
Slayer95, line 12: This needs an ``after()`` hook to restore the secondary chances when the test is done. Also it would be the best practice to accordingly mod them in a ``before()`` hook.

Slayer95:
``selfDrops`` seems to be for Close Combat et. al., and, relevantly, Diamond Storm?

Also, ``56ba7829661`` removed a debug instruction from battle.random()
``this.debug('randBW(' + (m ? (n ? m + ', ' + n : m) : '') + ') = ' + result)``

In hindsight, adding that back would allow for a test easily checking for the actual overflow in Charge Beam, and would in general improve many prng related tests, wdyt?

test/sim/abilities/serenegrace.js
DaWoblefet: Very good catch! I resolved this by editing the secondary mid-event rather than modifying the actual move in Dex.

DaWoblefet:
> `selfDrops` seems to be for Close Combat et. al., and, relevantly, Diamond Storm?

It's not for Diamond Storm, because Diamond Storm is a secondary (and so ``isSecondary`` is true and RNG isn't called). Close Combat would be example. I can't think of a percentage-based effect offhand like that has boosts but isn't a secondary, though.

> Also, [56ba782](https://github.com/smogon/pokemon-showdown/commit/56ba78296614b185d3485229c5885ad6442e4a42) removed a debug instruction from `battle.random()` `this.debug('randBW(' + (m ? (n ? m + ', ' + n : m) : '') + ') = ' + result)`
> 
> In hindsight, adding that back would allow for a test easily checking for the actual overflow in Charge Beam, and would in general improve many prng related tests, wdyt?

I don't understand what it was like previously so I can't say. If I can help it, I prefer not involving RNG in tests at all, but if you have to, I don't mind just fooling with RNG seeds until we get one that works. Fwiw you can't actually check for debug

Slayer95:
That's not what I was suggesting. Anyway, I realized my actual idea was impracticable, so please disregard that comment.

This now looks good ~

test/sim/abilities/serenegrace.js
Slayer95: Right, that's a better approach.

PR #8631, opened 2022-01-20
Avatars: Add a command to move all of a user's avatars at once

mia-pi-git:


server/chat-commands/avatars.tsx
```diff
@@ -759,6 +759,34 @@ export const commands: Chat.ChatCommands = {
 		</>);
 	},
 
+	async moveavatars(target, room, user) {
+		this.checkCan('bypassall');
+		const [from, to] = target.split(',').map(toID);
+		if (!from || !to) {
+			return this.parse(`/help moveavatars`);
+		}
+		if (!customAvatars[from] || !customAvatars[from].allowed.length) {
```
AnnikaCodes, line 768:
```suggestion
		if (!customAvatars[from]?.allowed.length) {
```

Zarel:
Looks good, thanks!

PR #8632, opened 2022-01-22
Gen7 Random Battle: Add Rhydon

Tarrembeau:
Remove stealth rock from rhyperior as well to differentiate them better.


AnnikaCodes:
Thanks!

PR #8637, opened 2022-01-24
Shared Power: Fix sharing weather suppression

HisuianZoroark:
Shared Power currently does not share Cloud Nine or Air Lock's weather suppression, so this should properly fix it.


KrisXV:
ty

PR #8638, opened 2022-01-24
Here I changed some max level and minimum level

Vaibhav04Goyal:
You can do confirm as 

Max 50 Level is changed to 100
and Min is changed from 10 to 5

Thanks

### DIFF
--- a/config/CUSTOM-RULES.md
+++ b/config/CUSTOM-RULES.md
@@ -228,9 +228,9 @@ Value rules
 
 Certain rules can specify a value.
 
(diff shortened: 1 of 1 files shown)


mia-pi-git:
Is there a need for this? This documentation is just meant to demonstrate the usage of the rule.

DaWoblefet:
I don't think this improves the documentation. Banning Pokemon above level 100 isn't relevant to the vast majority of formats, so it's not a relevant example. I don't see any significant difference in level 5 versus level 10 either; it seems to me the concept isn't expressed any more clearly by picking a different number.

PR #8639, opened 2022-01-25
Fix Bird-type crashes

DaWoblefet:
Bird-type has caused two recent independent crashes - one with Revelationmons, and one with some unholy combination of Shared Power + Super Staff Bros 4. Both crashes are fixed by not allowing the Bird-type to be set in those scenarios. I also fixed a typo in Revelationmons message at the start of the battle.


KrisXV:
ty

PR #8641, opened 2022-01-25
Usersearch: Convert to TSX

KrisXV:


AnnikaCodes:
This seems okay to me at a glance, but I haven't learned Preact yet so I'm not sure I feel comfortable merging this unilaterally.

server/chat-plugins/usersearch.tsx
```diff
@@ -7,76 +7,101 @@ export const nameList = new Set<string>(JSON.parse(
...
-			results.online.push(`${!page ? ONLINE_SYMBOL : ''} ${curUser.name}`);
-		} else {
-			results.offline.push(`${!page ? OFFLINE_SYMBOL : ''} ${curUser.name}`);
+class PunishmentHTML extends Chat.JSX.Component<{userid: ID, target: string}> {
+	render() {
+		const {userid, target} = {...this.props};
+		const buf = [];
+		for (const cmdName of ['Forcerename', 'Namelock', 'Weeknamelock']) {
+			// We have to use dangerouslySetInnerHTML here because otherwise the `value`
+			// property of the button tag is auto escaped, making &#10; into &amp;#10;
```
Zarel, line 16: Why not just unescape it? Use `\n` instead of `&#10;`? Would that work?

KrisXV: I tried that, it doesn't work.

AnnikaCodes:
Thanks!

PR #8643, opened 2022-01-27
Abuse-monitor: Sort recommendations by user

mia-pi-git:
Convenience for staff members, so they can see who got which recommendation. Mostly a PR since this involves a migration.

### DIFF
--- a/server/chat-plugins/abuse-monitor.ts
+++ b/server/chat-plugins/abuse-monitor.ts
@@ -42,6 +42,8 @@ const REPORT_NAMECOLORS: {[k: string]: string} = {
 	other: '', // black - empty since handled by dark mode
 };
 
+export let migrated = global.Chat?.oldPlugins['abuse-monitor']?.migrated || false;
+
 export const cache: {
 	[roomid: string]: {
 		users: Record<string, number>,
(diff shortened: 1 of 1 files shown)


server/chat-plugins/abuse-monitor.ts
```diff
@@ -42,16 +42,31 @@ const REPORT_NAMECOLORS: {[k: string]: string} = {
...
+} = (() => {
+	const plugin = global.Chat?.oldPlugins['abuse-monitor'];
+	if (!plugin?.cache) return {};
+	if (plugin.migrated) return plugin.cache;
+	for (const k in plugin.cache) {
+		const cur = plugin.cache[k];
+		if (typeof cur.recommended?.type === 'string') { // would be object if it was the new entry
+			// we cannot feasibly determine who it was (but it __is__ logged in <<abuselog>>, so staff can)
+			delete cur.recommended;
+		}
```
AnnikaCodes, line 65: Should we really discard data about recommendations just because we don't know which user it was for?

mia-pi-git: Not like there's a better way to do it. Plus, it's logged to abuselog anyway, so staff can still see it.

AnnikaCodes: You could save it under something like `an unknown user`, but it's probably fine as-is.

AnnikaCodes:
Thanks!

PR #8644, opened 2022-01-27
Add Legends: Arceus Pokemon

KrisXV:


Karthik99999:
Wouldn't it be better if the pla mons used their own `isNonstandard` value instead of using `Future`? It would make it easier to specifically unban pla mons for a format

KrisXV:
> Wouldn't it be better if the pla mons used their own `isNonstandard` value instead of using `Future`? It would make it easier to specifically unban pla mons for a format

Potentially, but they're the only pokemon using "Future" so it's effectively the same regardless. We can add a PLA isNonstandard value in the future if they end up getting some weird lgpe treatment next gen

sim/dex-species.ts
```diff
@@ -481,7 +481,8 @@ export class DexSpecies {
 					species.tier = species.doublesTier = 'Illegal';
 				}
 			}
-			species.nfe = !!(species.evos.length && this.get(species.evos[0]).gen <= this.dex.gen);
+			const legalEvos = species.evos.filter(mon => !this.get(mon).isNonstandard);
+			species.nfe = !!(legalEvos.length && this.get(legalEvos[0]).gen <= this.dex.gen);
```
urkerab, line 485: Would `some` be better than messing around with `filter`, something like this?
```suggestion
			species.nfe = species.evos.some(mon => this.get(mon).gen <= this.dex.gen && !this.get(mon).isNonstandard);
```

KrisXV: These effectively do the same thing

data/pokedex.ts
```diff
@@ -1428,6 +1445,25 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
...
+		formeOrder: ["Arcanine", "Arcanine-Hisui"],
+	},
+	arcaninehisui: {
+		num: 59,
+		name: "Arcanine-Hisui",
+		baseSpecies: "Arcanine",
+		forme: "Hisui",
+		types: ["Fire", "Rock"],
+		genderRatio: {M: 0.75, F: 0.25},
+		baseStats: {hp: 90, atk: 110, def: 80, spa: 100, spd: 80, spe: 95},
```
Marty-D, line 1458:
```suggestion
		baseStats: {hp: 95, atk: 115, def: 80, spa: 95, spd: 80, spe: 90},
```

data/pokedex.ts
```diff
@@ -12077,6 +12358,23 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
...
 		eggGroups: ["Dragon"],
+		otherFormes: ["Goodra-Hisui"],
+		formeOrder: ["Goodra", "Goodra-Hisui"],
+	},
+	goodrahisui: {
+		num: 706,
+		name: "Goodra-Hisui",
+		baseSpecies: "Goodra",
+		forme: "Hisui",
+		types: ["Dragon"],
```
Marty-D, line 12369:
```suggestion
		types: ["Steel", "Dragon"],
```

KrisXV: I got a bit copypaste happy towards the end… oops

Marty-D (approved):
Thanks! 👍

Zarel:
Probably a value of `Spinoff` makes sense (replacing `LGPE`) so we don't have to keep doing this.

sim/dex-species.ts
urkerab: I wouldn't have suggested code if I thought that it didn't calculate `nfe` correctly... the point was just to have nicer-looking code.

PR #8645, opened 2022-01-30
Prevent illegal evolutions from activating Eviolite

HisuianZoroark:
It's currently possible to successfully run Ursaring with an Eviolite when it supposedly shouldn't work, this should hopefully fix that.


KrisXV:
This is just a visual bug, on the client fwiw, so this code doesnt do anything

HisuianZoroark:
yeah seems like it is, thank you.

PR #8661, opened 2022-02-08
Pokebilities: Streamline ability bans

HisuianZoroark:
Allows one to ban abilities without having to go through every single pokemon who has said ability. This also eases complex banning in pokebilities which makes tournament codes exponentially more condensed.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1643,6 +1643,21 @@ export const Formats: FormatList = [
 			// Shadow Tag/Arena Trap
 			'Diglett-Base', 'Dugtrio-Base', 'Gothita', 'Gothitelle', 'Gothorita', 'Trapinch', 'Wobbuffet', 'Wynaut',
 		],
+		onValidateSet(set) {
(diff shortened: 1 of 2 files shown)


KrisXV (approved):
I think this looks fine, but I'm going to wait until @Zarel or @Marty-D takes a look since it deals with team validator

Marty-D (approved):
Sure, why not. Thanks! 👍

config/formats.ts
```diff
@@ -1643,6 +1643,21 @@ export const Formats: FormatList = [
 			// Shadow Tag/Arena Trap
 			'Diglett-Base', 'Dugtrio-Base', 'Gothita', 'Gothitelle', 'Gothorita', 'Trapinch', 'Wobbuffet', 'Wynaut',
 		],
+		onValidateSet(set) {
```
urkerab, line 1646: The third parameter is actually the `setHas` variable, so it seems as if you could actually move the code from the team validator block to this function. (But I haven't actually tried it to see whether it works.)

PR #8662, opened 2022-02-09
Support classifying message toxicity locally

mia-pi-git:
This PR both adds a local message classifier, and moves the code for the remote classifier using Perspective to a directory in server/. The local classifier is in server/artemis/local.ts, the remote is in server/artemis/remote.ts. The local classifier uses a python dependency to classify messages - the TS file uses `child_process.spawn` to wrap the script. It is kept in a StreamProcessManager, where each stream spawns an instance of the python process. This is done so the spawning isn't done in the main process, since that causes noticeable lag. The remote code was just copy/pasted from the abuse-monitor chat plugin.

### DIFF
--- a/server/artemis/install
+++ b/server/artemis/install
@@ -1,4 +1,6 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
+set -o errexit
+set -o errtrace
 cd server/artemis/
 python3 -m pip install -r requirements.txt > /dev/null
 echo "Dependencies installed. Starting model installation..."
--- a/server/artemis/install
+++ b/server/artemis/install
@@ -1,0 +1,6 @@
+#!/usr/bin/env sh
+cd server/artemis/
+python3 -m pip install -r requirements.txt > /dev/null
+echo "Dependencies installed. Starting model installation..."
+python3 -c "import detoxify; detoxify.Detoxify('unbiased')" > /dev/null
+echo "Model installed. Local classifier is ready to use."
--- a/server/artemis/local.ts
+++ b/server/artemis/local.ts
@@ -1,0 +1,152 @@
+/**
+ * Typescript wrapper around the Python Artemis model.
+ * By Mia.
+ * @author mia-pi-git
+ */
+
+import * as child_process from 'child_process';
+import {ProcessManager, Streams, Utils, Repl} from '../../lib';
+import {Config} from '../config-loader';
+import {toID} from '../../sim/dex-data';
+
+class ArtemisStream extends Streams.ObjectReadWriteStream<string> {
+	tasks = new Set<string>();
+	private process: child_process.ChildProcessWithoutNullStreams;
+	constructor() {
+		super();
+		this.process = child_process.spawn('python3', [
+			'-u', __dirname + '/model.py', Config.debugartemisprocesses ? "debug" : "",
+		].filter(Boolean));
+		this.listen();
+	}
+	listen() {
+		this.process.stdout.setEncoding('utf8');
+		this.process.stderr.setEncoding('utf8');
+		this.process.stdout.on('data', (data) => {
+			// so many bugs were created by \nready\n
+			data = data.trim();
+			const [taskId, dataStr] = data.split("|");
+			if (this.tasks.has(taskId)) {
+				this.tasks.delete(taskId);
+				return this.push(`${taskId}\n${dataStr}`);
+			}
+			if (taskId === 'error') { // there was a major crash and the script is no longer running
+				const info = JSON.parse(dataStr);
+				Monitor.crashlog(new Error(info.error), "An Artemis script", info);
+				try {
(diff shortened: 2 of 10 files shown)


server/artemis/install
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+cd server/artemis/
```
AnnikaCodes, line 2:
```suggestion
#!/usr/bin/env bash
set -o errexit
set -o errtrace
cd server/artemis/
```

server/artemis/model.py
```diff
@@ -0,0 +1,53 @@
...
+def now():
+	return int(time.time() * 1000)
+
+logfile = open("logs/artemis.log", "a")
+def log(message):
+	if not debug:
+		return None
+
+	logfile.write(str(now()) + ":" + str.rstrip(message) + "\n")
+	logfile.flush()
```
AnnikaCodes, line 31:
```suggestion
	if debug:
		logfile.write(str(now()) + ":" + str.rstrip(message) + "\n")
		logfile.flush()
```

server/artemis/model.py
```diff
@@ -0,0 +1,53 @@
...
+	for line in sys.stdin:
+		log("in:" + line)
+		parts = line.split("|")
+		out = parts.pop(0) + "|"
+		try:
+			res = model.predict("|".join(parts))
+			for key in res: res[key] = str(res[key]) # json.dumps doesn't like floats
+			out += json.dumps(res)
+		except BaseException as e:
+			out += json.dumps({'error': f"{e}"})
```
AnnikaCodes, line 45: If you're using f-strings here (thus preventing the script from running on Python version that don't support f-strings), why not use them for the rest of the string interpolation you do?

server/artemis/remote.ts
```diff
@@ -0,0 +1,176 @@
...
+		return PM.respawn();
+	}
+	spawn(number: number) {
+		PM.spawn(number);
+	}
+	getActiveProcesses() {
+		return PM.processes.length;
+	}
+}
+
```
AnnikaCodes, line 176: I assume this file is just copy/pasted from the preexisting Artemis code? Are there changes I should be aware of?

server/artemis/local.ts
```diff
@@ -0,0 +1,158 @@
...
+			// floats have to be made into strings because python's json.dumps
+			// doesn't like float32s
+			data[k] = parseFloat(data[k]);
+		}
+		return data as Record<string, number>;
+	}
+}
+
+// main module check necessary since this gets required in other non-parent processes sometimes
+// when that happens we do not want to take over or set up or anything
```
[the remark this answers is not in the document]
mia-pi-git: No. It's not one or the other, it's both for different applications.

server/artemis/remote.ts
mia-pi-git: No. It's all the same.

server/artemis/model.py
mia-pi-git: Yeah, probably should. Done in https://github.com/smogon/pokemon-showdown/pull/8662/commits/ae3f15d7577854f769423d08b4c5220a64cf4b74.

server/artemis/local.ts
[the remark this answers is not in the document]
AnnikaCodes: Oh, okay.

server/artemis/local.ts
```diff
@@ -0,0 +1,165 @@
...
+				const info = JSON.parse(dataStr);
+				Monitor.crashlog(new Error(info.error), "An Artemis script", info);
+				try {
+					this.pushEnd(); // push end first so the stream always closes
+					this.process.disconnect();
+				} catch {}
+			}
+		});
+		this.process.stderr.on('data', data => {
+			data += "";
```
AnnikaCodes, line 43: ?

mia-pi-git: data is a Buffer by default. This just makes it into a string.

AnnikaCodes: Why choose `+= ""` instead of `.toString()`?

mia-pi-git: Looks nicer, in my opinion.

AnnikaCodes (approved):
This looks fine to me, assuming it's been tested.

server/artemis/local.ts
AnnikaCodes: I think it makes the code more difficult to read for people who aren't familiar with JavaScript's type coercion.

mia-pi-git: We can junk it anyway since setEncoding('utf-8') means the data is a string. This was just a prior relic of that.

mia-pi-git:
> This looks fine to me, assuming it's been tested.

Yes, I've done a bunch of testing.

AnnikaCodes:
Thanks!

PR #8667, opened 2022-02-14
Add a timeout for userprefixes

KrisXV:


server/chat-plugins/username-prefixes.ts
```diff
@@ -6,58 +6,79 @@
...
 		if (configJustLoaded) {
 			// if we just loaded the config file, ensure that all prefixes are IDs
-			if (Config.forcedprefixes.privacy) Config.forcedprefixes.privacy = Config.forcedprefixes.privacy.map(toID);
-			if (Config.forcedprefixes.modchat) Config.forcedprefixes.modchat = Config.forcedprefixes.modchat.map(toID);
+			if (!Array.isArray(Config.forcedprefixes)) {
+				const convertedPrefixes = [];
+				for (const type in Config.forcedprefixes) {
+					for (const prefix of Config.forcedprefixes[type]) {
+						convertedPrefixes.push({type, prefix, expireAt: Date.now() + WEEK});
+						this.timeouts.set(prefix, setTimeout(() => {
```
mia-pi-git, line 34: I think it would be better to keep one master timeout that iterates over all the prefixes and just clears out ones that are dated - would scale better and would be easier to destroy.

KrisXV: So... a weekly interval?

KrisXV: That's easy I just want to make sure that's what you meant

KrisXV: Also, I'm thinking that's not very friendly to the strict nature of prefixes lasting a week; If the server restarts midway through a ladder tour, then the prefix lasts for an additional week and can be used to manipulate game publicity

mia-pi-git: So run it once a day?

server/chat-plugins/username-prefixes.ts
```diff
@@ -6,58 +6,103 @@
 import {FS} from '../../lib';
 
 const PREFIXES_FILE = 'config/chat-plugins/username-prefixes.json';
+const WEEK = 24 * 7 * 60 * 60 * 1000;
```
AnnikaCodes, line 9:
```suggestion
const PREFIX_DURATION = 10 * 24 * 60 * 60 * 1000;
```

AnnikaCodes (changes requested):
I spoke to policy US about this and they want prefixes to last 10 days.

server/chat-plugins/username-prefixes.ts
mia-pi-git: I would also run the purge on startup on possible. But definitely run the timer at least once a day.

AnnikaCodes: One master timeout _is_ better at scale, but there shouldn't be more than 5-10 prefixes at once (and probably more like 2), so I'm not sure that scalability matters much here.

mia-pi-git: I don't think it hurts much to prep for scale. A single timer also is less memory (less to keep track of) and it's easier to manage (don't have to calculate remaining time for each prefix on every startup, etc).

server/chat-plugins/username-prefixes.ts
```diff
@@ -109,11 +154,13 @@ export const commands: Chat.ChatCommands = {
 
 			const types = target ? [prefixManager.validateType(toID(target))] : ['privacy', 'modchat'];
 
+			const entries = Config.forcedprefixes.filter((x: any) => types.includes(x.type)).map((x: any) => x.prefix);
+
 			return this.sendReplyBox(types.map(type => {
-				const info = Config.forcedprefixes[type].length ?
-					`<code>${Config.forcedprefixes[type].join('</code>, <code>')}</code>` : `none`;
+				const info = entries.filter((x: any) => x.type === type).length ?
+					`<code>${entries.filter((x: any) => x.type === type).join('</code>, <code>')}</code>` : `none`;
```
AnnikaCodes, line 161: `entries` should be an array of strings, right, since you've mapped `x => x.prefix`?

KrisXV: oh yeah it is, ill fix in a bit

server/chat-plugins/username-prefixes.ts
KrisXV: makes sense, did

server/chat-plugins/username-prefixes.ts
```diff
@@ -109,11 +153,14 @@ export const commands: Chat.ChatCommands = {
...
+
 			return this.sendReplyBox(types.map(type => {
-				const info = Config.forcedprefixes[type].length ?
-					`<code>${Config.forcedprefixes[type].join('</code>, <code>')}</code>` : `none`;
+				const prefixes = entries.filter((x: any) => x.type === type).map((x: any) => x.prefix);
+				const info = prefixes.length ?
+					`<code>${prefixes.join('</code>, <code>')}</code>` : `none`;
 				return `Username prefixes that disable <strong>${type}</strong>: ${info}.`;
-			}).join(`<br />`));
+			}).join('<br />'));
```
AnnikaCodes, line 163: Why change this? PS' string guidelines stipulate that `` ` `` should be used for HTML. IMO it doesn't matter enough to change code that already misuses it, but why go out of your way to change code that already adheres to our quote conventions?

KrisXV: It wasn't intentional... I accidentally converted it because i deleted the code then rewrote it from memory

AnnikaCodes (approved):
Okay, I think this is good to go! Thanks!

PR #8671, opened 2022-02-19
Fix Heal Bell/Aromatherapy in Multi Battles

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9127908

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -502,8 +502,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {snatch: 1, distance: 1},
 		onHit(pokemon, source, move) {
 			this.add('-activate', source, 'move: Aromatherapy');
+			const side = pokemon.side;
+			let allies = side.pokemon;
+			if (side.allySide) allies = allies.concat(side.allySide.pokemon);
 			let success = false;
-			for (const ally of pokemon.side.pokemon) {
+			for (const ally of allies) {
 				if (ally !== source && ((ally.hasAbility('sapsipper')) ||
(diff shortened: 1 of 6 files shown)


DaWoblefet:
There was an addition to the skipped test in https://github.com/smogon/pokemon-showdown/pull/8571 that I would like in here; can you add that and make sure it passes too?

data/moves.ts
```diff
@@ -502,8 +502,11 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {snatch: 1, distance: 1},
 		onHit(pokemon, source, move) {
 			this.add('-activate', source, 'move: Aromatherapy');
+			const side = pokemon.side;
+			let allies = side.pokemon;
+			if (side.allySide) allies = allies.concat(side.allySide.pokemon);
 			let success = false;
-			for (const ally of pokemon.side.pokemon) {
+			for (const ally of allies) {
```
urkerab, line 509:
```suggestion
			for (const ally of pokemon.side.allies()) {
```

Karthik99999: yeah the issue with #Pokemon.allies is that it only gets active allies. Maybe we could make a new function that gets inactive ones as well?

urkerab: Sorry, I forgot that the `allyTeam` target doesn't actually target the team, it's just for show, and Aromatherapy and Heal Bell actually have to manually iterate the team. (I don't know whether it would be possible to fix that; you would need to activate the move in a separate event if you did.)

Marty-D (approved):
Nice, thanks! 👍

PR #8672, opened 2022-02-20
Tournaments: Add /recenttours

KrisXV:
Requested by tournaments staff and is already used in other rooms with bots. I also fixed the settings permissions because they were never supposed to be accessible to non-ROs

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -15,6 +15,8 @@ export interface TournamentRoomSettings {
 	forceTimer?: boolean;
 	playerCap?: number;
 	showSampleTeams?: boolean;
+	recentToursLength?: number;
+	recentTours?: {name: string, time: number}[];
 }
 
 type Generator = RoundRobin | Elimination;
@@ -1185,6 +1187,19 @@ export class Tournament extends Rooms.RoomGame {
 			bracketData: this.getBracketData(),
 		};
 		this.room.add(`|tournament|end|${JSON.stringify(update)}`);
+		const settings = this.room.settings.tournaments;
+		if (settings?.recentToursLength) {
+			if (!settings.recentTours) settings.recentTours = [];
+			const name = Dex.formats.get(this.name).exists ? Dex.formats.get(this.name).name :
+				`${this.name} (${Dex.formats.get(this.baseFormat).name})`;
+			settings.recentTours.unshift({name, time: Date.now()});
+			// Use a while loop here in case the threshold gets lowered with /tour settings recenttours
+			// to trim down multiple at once
+			while (settings.recentTours.length > settings.recentToursLength) {
+				settings.recentTours.splice(settings.recentTours.length - 1, 1);
+			}
+			this.room.saveSettings();
+		}
 		this.remove();
 	}
 }
@@ -1263,6 +1278,29 @@ function createTournament(
 }
 
 const commands: Chat.ChatCommands = {
+	recenttours(target, room, user) {
+		this.runBroadcast();
+		room = this.requireRoom();
+		if (!room.settings.tournaments?.recentToursLength) {
+			throw new Chat.ErrorMessage(`Recent tournaments aren't documented in this room.`);
+		}
+		if (!room.settings.tournaments?.recentTours?.length) {
+			throw new Chat.ErrorMessage(`There haven't been any documented tournaments in this room recently.`);
+		}
+		// Shorten array if the recentToursLength gets adjusted
+		const array = room.settings.tournaments.recentTours.filter((x, i) => (
+			i <= room!.settings.tournaments!.recentToursLength! - 1
+		));
+		const {name, time} = array[0];
+		let buf = `The last tournament ended ${Chat.toDurationString(Date.now() - time)} ago - ${name}`;
+		if (array.length > 1) {
+			buf += `<hr /><strong>Previous tournments:</strong> `;
+			buf += array.filter((x, i) => i !== 0).map(x => x.name).join(', ');
+		}
+		this.sendReplyBox(buf);
+	},
+	recenttourshelp: [`/recenttours - Displays the n most recent tour(s).`],
+
 	tour: 'tournament',
 	tours: 'tournament',
 	tournaments: 'tournament',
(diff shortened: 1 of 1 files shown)


server/tournaments/index.ts
```diff
@@ -1185,6 +1188,19 @@ export class Tournament extends Rooms.RoomGame {
...
+		const settings = this.room.settings.tournaments;
+		if (settings?.recentToursLength) {
+			if (!settings.recentTours) settings.recentTours = [];
+			const name = Dex.formats.get(this.name).exists ? Dex.formats.get(this.name).name :
+				`${this.name} (${Dex.formats.get(this.baseFormat).name})`;
+			settings.recentTours.unshift({name, baseFormat: this.baseFormat, time: Date.now()});
+			// Use a while loop here in case the threshold gets lowered with /tour settings recenttours
+			// to trim down multiple at once
+			while (settings.recentTours.length > settings.recentToursLength) {
+				settings.recentTours.splice(settings.recentTours.length - 1, 1);
```
AnnikaCodes, line 1200: Would it be more readable to use `Array#pop` here?

KrisXV: ...I forgot Array#pop existed. oops

server/tournaments/index.ts
```diff
@@ -1263,6 +1287,30 @@ function createTournament(
...
+		if (!room.settings.tournaments?.recentToursLength) {
+			throw new Chat.ErrorMessage(`Recent tournaments aren't documented in this room.`);
+		}
+		if (!room.settings.tournaments?.recentTours?.length) {
+			throw new Chat.ErrorMessage(`There haven't been any documented tournaments in this room recently.`);
+		}
+		// Shorten array if the recentToursLength gets adjusted
+		const array = room.settings.tournaments.recentTours.filter((x, i) => (
+			i <= room!.settings.tournaments!.recentToursLength! - 1
+		));
```
AnnikaCodes, line 1303: If `recentTourLength` has changed, why not either remove the elements from `recentTours` (instead of making a new array), or just trim `recentTours` when you change `recentTourLength`?

server/tournaments/index.ts
```diff
@@ -1263,6 +1287,30 @@ function createTournament(
...
+		const {name, time} = array[0];
+		let buf = `The last tournament ended ${Chat.toDurationString(Date.now() - time)} ago - ${name}`;
+		if (array.length > 1) {
+			buf += `<hr /><strong>Previous tournments:</strong> `;
+			buf += array.filter((x, i) => i !== 0).map(x => x.name).join(', ');
+		}
+		this.sendReplyBox(buf);
+	},
+	recenttourshelp: [`/recenttours - Displays the n most recent tour(s).`],
```
AnnikaCodes, line 1312: n?

server/tournaments/index.ts
KrisXV: I figured people might want to preserve it while coding but looking back I don't know why I thought that since it seems pretty minute, so I'll fix it to do the latter

server/tournaments/index.ts
KrisXV: variable n as in n, n-1, n-2, i couldnt figure out a better way to phrase this while remaining compact

server/tournaments/index.ts
```diff
@@ -2172,6 +2220,73 @@ const commands: Chat.ChatCommands = {
...
+					return this.parse(`/help tour settings`);
+				}
+				if (!room.settings.tournaments) room.settings.tournaments = {};
+				if (!isNaN(num) && num <= 15 && num >= 1) {
+					if (room.settings.tournaments.recentToursLength === num) {
+						throw new Chat.ErrorMessage(`Recent tournament threshold is already set to ${num}.`);
+					}
+					room.settings.tournaments.recentToursLength = num;
+					room.saveSettings();
+					this.privateModAction(`Recent tournament threshold was set to ${num} by ${user.name}.`);
```
AnnikaCodes, line 2239: This wording is a little clunky; maybe "number of recent tournaments to record"?

server/tournaments/index.ts
```diff
@@ -2172,6 +2220,73 @@ const commands: Chat.ChatCommands = {
...
+						if (toID(target) === 'forcedelete') {
+							delete room.settings.tournaments.recentTours;
+							this.privateModAction(`Recent tournaments list was deleted by ${user.name}.`);
+							this.modlog('TOUR SETTINGS', null, `recent tours: delete`);
+						}
+						room.saveSettings();
+						this.privateModAction(`Recent tournament threshold was turned off by ${user.name}.`);
+						this.modlog('TOUR SETTINGS', null, `recent threshold: off`);
+					} else {
+						if (toID(target) !== 'forcedelete') {
```
AnnikaCodes, line 2253: Why not use a variable for `toID(target) === 'forcedelete'`?

server/tournaments/index.ts
```diff
@@ -2172,6 +2220,73 @@ const commands: Chat.ChatCommands = {
...
+							throw new Chat.ErrorMessage(`Recent tournament threshold is already disabled.`);
+						}
+					}
+				} else {
+					this.sendReply(`Usage: ${this.cmdToken}${this.fullCmd} <number|off|forcedelete>`);
```
AnnikaCodes, line 2258: Is there a reason that this isn't just in a normal help command? It has no aliases and I think most people can figure out the `/` vs. `!` distinction themselves.

server/tournaments/index.ts
AnnikaCodes: You never mention `n` anywhere else in the help, though, so it's unclear what it refers to. It seems like it just displays however many recent tournaments have been recorded (which will be at minimum 0 and at most `settings.tournaments.recentTourLength`).

server/tournaments/index.ts
KrisXV: It is in the normal help command fwiw, it's just consistency with the other tours commands, I will likely refactor it later on

KrisXV:
@AnnikaCodes done

server/tournaments/index.ts
AnnikaCodes: If it's for consistency that's okay.

AnnikaCodes:
Thanks!

PR #8673, opened 2022-02-22
Gen 2: Fix Return/Frustration min damage

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9116008

Return and Frustration should be able to deal 0 damage (see bug report above for in game footage)


AnnikaCodes:
Thanks!

PR #8674, opened 2022-02-23
Fix /modjoin in battles

MathyFurret:
This fixes a bug where if Player 1 sets modjoin and turns it off, Player 2 still can't turn it back on.

This doesn't fix everything wrong with modjoin in battles; for example, if Player 1 sets it to +, Player 2 can't set it higher. This is quite minor and requires more reworking than is probably worth for a kinda-deprecated feature.


AnnikaCodes:
Thanks!

PR #8675, opened 2022-02-24
Metronome Battle: Retain Fur Coat Ban

VineST:
The retest has concluded, and Fur Coat remains banned while Fluffy goes free.
https://www.smogon.com/forums/threads/metronome-battle-fluffy-fur-coat-retest-voting.3698121/


AnnikaCodes (approved):
Thanks!

PR #8677, opened 2022-02-25
Rotation Battles

MathyFurret:
An [approved suggestion](https://www.smogon.com/forums/threads/rotation-battles.3666026/) on Smogon.

This updates the simulator to handle Rotation Battles. They're really just Single Battles with some extra steps, so not a lot needed to be done.

Currently this is a draft to review the protocol and API changes. Once these are approved, I will begin working on the client updates and playability on the main server should be decided. When a format is finalized I will add unit tests but I have tested the core features in the command line and there are no obvious bugs.

## Protocol changes

Input: The `rotateleft` and `rotateright` choices have been added. Both must be followed by a `move` choice.

Output: The `|rotateleft|SIDE` and `|rotateright|SIDE` majors have been added.

No backwards-compatibility issues should arise.

### DIFF
--- a/sim/SIM-PROTOCOL.md
+++ b/sim/SIM-PROTOCOL.md
@@ -295,6 +295,11 @@ message).
 > Moves already active `POKEMON` to active field `POSITION` where the
 > leftmost position is 0 and each position to the right counts up by 1.
 
+`|rotateleft|SIDE` or `|rotateright|SIDE`
+
+> In a Rotation Battle, `SIDE`'s Pokémon rotated in the given direction,
+> and their 2nd or 3rd Pokémon is now active, respectively
+
 `|cant|POKEMON|REASON` or `|cant|POKEMON|REASON|MOVE`
 
 > The Pokémon `POKEMON` could not perform a move because of the indicated
@@ -686,6 +691,9 @@ To be exact, `CHOICE` is one of:
 - `move MOVESPEC zmove`, to use a z-move version of a move
 
 - `move MOVESPEC max`, to Dynamax/Gigantamax and make a move
+
+- `rotateright move MOVESPEC ...` or `rotateleft move MOVESPEC ...`, 
+in Rotation Battles to rotate to your second or third Pokémon, respectively, and use its move
 
 - `switch SWITCHSPEC`, to make a switch
 
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -183,6 +183,25 @@ export class BattleActions {
 		pokemon.draggedIn = null;
 		return true;
 	}
+	rotateIn(pokemon: Pokemon) {
+		const side = pokemon.side;
+		if (pokemon === side.pokemon[1]) {
+			side.pokemon.unshift(...side.pokemon.splice(1, 2));
+			side.active[0].isActive = false;
+			side.active[0] = pokemon;
+			pokemon.isActive = true;
+			this.battle.add('rotateright', side);
+		} else if (pokemon === side.pokemon[2]) {
+			side.pokemon.unshift(...side.pokemon.splice(2, 1));
+			side.active[0].isActive = false;
+			side.active[0] = pokemon;
+			pokemon.isActive = true;
+			this.battle.add('rotateleft', side);
+		}
+		for (let i = 0; i < side.pokemon.length; i++) {
+			side.pokemon[i].position = i;
+		}
+	}
 
 	// #endregion
 
--- a/sim/battle-queue.ts
+++ b/sim/battle-queue.ts
@@ -49,8 +49,8 @@ export interface MoveAction {
 /** A switch action */
 export interface SwitchAction {
 	/** action type */
-	choice: 'switch' | 'instaswitch';
-	order: 3 | 103;
+	choice: 'switch' | 'instaswitch' | 'rotate';
+	order: 3 | 103 | 104;
 	/** priority of the action (lower first) */
 	priority: number;
 	/** speed of pokemon switching (higher first if priority tie) */
@@ -177,8 +177,9 @@ export class BattleQueue {
 				runSwitch: 101,
 				runPrimal: 102,
 				switch: 103,
-				megaEvo: 104,
-				runDynamax: 105,
+				rotate: 104,
+				megaEvo: 105,
+				runDynamax: 106,
 
 				shift: 200,
 				// default is 200 (for moves)
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -1219,7 +1219,9 @@ export class Battle {
 			for (let i = 0; i < this.sides.length; i++) {
 				const side = this.sides[i];
 				if (!side.pokemonLeft) continue;
-				const activeData = side.active.map(pokemon => pokemon?.getMoveRequestData());
+				let active = side.active;
+				if (this.gameType === 'rotation') active = active.concat(side.pokemon.slice(1, 3));
+				const activeData = active.map(pokemon => pokemon?.getMoveRequestData());
 				requests[i] = {active: activeData, side: side.getRequestData()};
 				if (side.allySide) {
 					requests[i].ally = side.allySide.getRequestData(true);
@@ -1352,7 +1354,8 @@ export class Battle {
 		if (!side.pokemonLeft) return [];
 
 		const canSwitchIn = [];
-		for (let i = side.active.length; i < side.pokemon.length; i++) {
+		const startPos = this.gameType === 'rotation' ? 3 : side.active.length;
+		for (let i = startPos; i < side.pokemon.length; i++) {
 			const pokemon = side.pokemon[i];
 			if (!pokemon.fainted) {
 				canSwitchIn.push(pokemon);
@@ -2379,6 +2382,16 @@ export class Battle {
 						this.actions.switchIn(side.pokemon[i], i);
 					}
 				}
+				if (this.gameType === 'rotation') {
+					for (let i = 1; i < 3; i++) {
+						// temporarily flag as active for Illusion, etc
+						// TODO is this actually needed?
+						side.pokemon[i].isActive = true;
+						this.runEvent('BeforeSwitchIn', side.pokemon[i]);
+						this.add('switch', side.pokemon[i], side.pokemon[i].getDetails);
+						side.pokemon[i].isActive = false;
+					}
+				}
 			}
 			for (const pokemon of this.getAllPokemon()) {
 				this.singleEvent('Start', this.dex.conditions.getByID(pokemon.species.id), pokemon.speciesState, pokemon);
@@ -2445,6 +2458,9 @@ export class Battle {
 					break;
 				}
 			}
+			break;
+		case 'rotate':
+			this.actions.rotateIn(action.target);
 			break;
 		case 'runUnnerve':
 			this.singleEvent('PreStart', action.pokemon.getAbility(), action.pokemon.abilityState, action.pokemon);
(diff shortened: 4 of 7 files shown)


MathyFurret:
More info:

In a Rotation Battle, each player starts by sending out 3 Pokemon at once, but only 1 is truly active at a time. Players can rotate to one of their "semi-active" Pokemon and have it use a move in the same turn. Semi-active Pokemon keep volatiles when rotating out but don't tick residuals.

The `switch` majors at the start of the battle include the semi-active Pokemon, as do the `active` arrays in request objects. Otherwise nothing else is sent about semi-active Pokemon that wouldn't be sent for inactive Pokemon, because they don't actually do anything when they aren't active.

sim/SIM-PROTOCOL.md
```diff
@@ -295,6 +295,11 @@ message).
 > Moves already active `POKEMON` to active field `POSITION` where the
 > leftmost position is 0 and each position to the right counts up by 1.
 
+`|rotateleft|SIDE` or `|rotateright|SIDE`
```
mia-pi-git, line 298: Nit: `rotate|direction|SIDE` seems cleaner / more intuitive.

sim/SIM-PROTOCOL.md
```diff
@@ -687,6 +692,9 @@ To be exact, `CHOICE` is one of:
 
 - `move MOVESPEC max`, to Dynamax/Gigantamax and make a move
 
+- `rotateright move MOVESPEC ...` or `rotateleft move MOVESPEC ...`,
```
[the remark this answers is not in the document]
MathyFurret: Not yet, but it's still up for debate; personally i prefer having them as 1 word here but I could go either way

sim/battle-queue.ts
```diff
@@ -177,8 +177,9 @@ export class BattleQueue {
 				runSwitch: 101,
 				runPrimal: 102,
 				switch: 103,
-				megaEvo: 104,
-				runDynamax: 105,
+				rotate: 104,
+				megaEvo: 105,
+				runDynamax: 106,
```
urkerab, line 182: Is this definitely the order that things happen in-game?

sim/battle.ts
```diff
@@ -1219,7 +1219,9 @@ export class Battle {
 			for (let i = 0; i < this.sides.length; i++) {
 				const side = this.sides[i];
 				if (!side.pokemonLeft) continue;
-				const activeData = side.active.map(pokemon => pokemon?.getMoveRequestData());
+				let active = side.active;
+				if (this.gameType === 'rotation') active = active.concat(side.pokemon.slice(1, 3));
```
urkerab, line 1223: Given the number of times that you use it, I think possibly there should be a property or method of `side` that returns three Pokémon when playing a rotation battle, so you can use that where you're currently having to special-case rotation battles. (Here you would just use it directly, but in other places you would use the length instead of special casing 3 or `this.active.length`.)

sim/pokemon.ts
```diff
@@ -671,6 +671,7 @@ export class Pokemon {
 	getAtLoc(targetLoc: number) {
 		let side = this.battle.sides[targetLoc < 0 ? this.side.n % 2 : (this.side.n + 1) % 2];
 		targetLoc = Math.abs(targetLoc);
+		if (this.battle.gameType === 'rotation') return side.pokemon[targetLoc - 1];
```
urkerab, line 674: Why is this needed, in the sense that why would anyone be trying to target position 2 or 3 during a rotation battle?

sim/battle.ts
```diff
@@ -2557,6 +2583,8 @@ export class Battle {
 			}
 		}
 
+		if (rotateFirst) return false;
```
urkerab, line 2586: What should happen if both Pokémon are knocked out, but one side can switch while the other needs to rotate?

sim/side.ts
```diff
@@ -918,6 +939,12 @@ export class Side {
 				if (data) return this.emitChoiceError(`Unrecognized data after "shift": ${data}`);
 				if (!this.chooseShift()) return false;
 				break;
+			case 'rotateright':
+			case 'rotateleft':
+				if (this.battle.gameType !== 'rotation') return this.emitChoiceError("Can't rotate: Not a rotation battle");
+				if (!data.startsWith('move ')) return this.emitChoiceError("Rotations must be followed by a move choice");
+				this.choice.willRotate = choiceType === 'rotateleft' ? 'left' : 'right';
+				return this.choose(data, true);
```
urkerab, line 947: This is one of those edge cases where my preference would be to fall through to the `move` choice (after adjusting the data appropriately) although my protocol option of having the rotation be a move flag bypasses this problem entirely.

sim/battle.ts
```diff
@@ -2379,6 +2382,16 @@ export class Battle {
 						this.actions.switchIn(side.pokemon[i], i);
 					}
 				}
+				if (this.gameType === 'rotation') {
+					for (let i = 1; i < 3; i++) {
+						// temporarily flag as active for Illusion, etc
+						// TODO is this actually needed?
+						side.pokemon[i].isActive = true;
+						this.runEvent('BeforeSwitchIn', side.pokemon[i]);
+						this.add('switch', side.pokemon[i], side.pokemon[i].getDetails);
```
urkerab, line 2391: Should this be using `this.actions.switchIn`?

urkerab (commented):
Looking good but I'd just like some clarification on some of the code if you don't mind.

sim/battle-queue.ts
MathyFurret: Woble said he'd do some reasearch, but rotations definitely happen after hard-switches.

sim/pokemon.ts
MathyFurret: When rotating and using a self-targetting move targetLoc is set as -2 or -3, which is beyond the bounds of side.active. This causes the `if` below to assume it's a multi battle and index a side that doesn't exist, which crashes.

sim/battle.ts
MathyFurret: I'm 90% sure that the auto-rotation should happen first, which is the purpose of this check.

sim/battle.ts
MathyFurret: I avoid `switchIn` because effects such as Intimidate shouldn't activate for the 2nd and 3rd Pokemon.

sim/battle-queue.ts
urkerab: Ah, you're waiting on research? Fair enough.

sim/battle.ts
urkerab: Ah, I was confused by `Ilusion, etc` - is this here only to set up Illusion? (When does the ability of the 2nd or 3rd Pokémon actually start?)

sim/battle.ts
urkerab: Actually I had the wrong use case. What happens if you rotate in and use Pursuit on the Pokémon that the other side is switching? If that uses this code path then the Pursuit executes before the rotation, since it's the `BeforeSwitchOut` event that triggers it.

sim/pokemon.ts
urkerab: Ah, so could my suggestion of a side property/method for "active except first 3 if rotation battle" take care of this case too?

sim/battle.ts
MathyFurret: They don't. This is the case for both gen 5 and 6.

Slayer95:
Could you add tests to spec the different behavior between right and left?

That is, whether "rotate right" means "rotate active pokemon to the right", or if that meaning is conveyed by "rotate left" ("rotate left pokemon into active position").

sim/battle-actions.ts
```diff
@@ -183,6 +183,27 @@ export class BattleActions {
...
+			side.pokemon.unshift(...side.pokemon.splice(2, 1));
+			this.battle.add('rotate', 'left', side);
+			break;
+		}
+		for (let i = 0; i < side.pokemon.length; i++) {
+			side.pokemon[i].position = i;
+		}
+		// TODO: Primal Reversion, Slow Start, anything else that should trigger
+		// for a Pokemon that started the battle semi-active
+	}
```
Slayer95, line 206: You seem to be breaking the match between the indices of ``active`` and ``pokemon``.

Is that a bug? Or could you please explain why that is needed?

sim/side.ts
```diff
@@ -262,6 +265,12 @@ export class Side {
 	hasAlly(pokemon: Pokemon) {
 		return pokemon.side === this || pokemon.side === this.allySide;
 	}
+	get subActives() {
+		return this.battle.gameType === 'rotation' ? this.pokemon.slice(1, 3) : [];
+	}
+	get activeAndSubActives() {
+		return this.active.concat(this.subActives);
+	}
```
Slayer95, line 273: Getters are frowned upon by @Zarel

sim/battle-actions.ts
MathyFurret: `active` is only length 1 in Rotation Battles. (I changed the stub that had Rotation Battles set `active` to have a length of 3.)

MathyFurret:
> Could you add tests to spec the different behavior between right and left? (The current tests have Pokémon with repeat movesets, so it isn't distinguishable).

The first new test in decisions.js does this, by checking species. (p1 goes right, p2 goes left)

I could change the movesets too, just to further ensure that the choice doesn't throw an error.

sim/side.ts
```diff
@@ -592,7 +610,7 @@ export class Side {
 			return this.emitChoiceError(`Can't move: You can only ultra burst once per battle`);
 		}
 		let dynamax = (megaDynaOrZ === 'dynamax');
-		const canDynamax = this.activeRequest?.active[this.active.indexOf(pokemon)].canDynamax;
+		const canDynamax = this.activeRequest?.active[this.pokemon.indexOf(pokemon)].canDynamax;
```
Slayer95, line 613: > You seem to be breaking the match between the indices of active and pokemon.

Oh, I understand now. Subactive Pokemon are retrofitted into ``activeRequest.active``, which is why this line had to be updated, and not because of some mismatch. You are making ``side.active`` an array of a single Pokemon, right?

sim/side.ts
MathyFurret: Yeah, now that I've given it some thought, there isn't much of a point to make these `get`

test/sim/decisions.js
```diff
@@ -484,6 +484,49 @@ describe('Choices', function () {
 		});
 	});
 
+	describe("Rotation requests", function () {
+		it("should allow rotations in rotation battles", function () {
+			battle = common.gen(5).createBattle({gameType: 'rotation'}, TRIPLES_TEAMS.full);
+
+			battle.makeChoices('rotate right move tackle', 'rotate left move tackle');
+			assert.species(battle.p1.active[0], "Ivysaur");
+			assert.species(battle.p2.active[0], "Blastoise");
```
Slayer95, line 493: Since you are using ``TRIPLES_TEAMS``, could you unhardcode the species and refer to them by their array indices?

```js
TRIPLES_TEAMS.full[0][2]
TRIPLES_TEAMS.full[1][1]
```

Slayer95:
Looks good so far. IMO you should go ahead with work on the client to provide a full picture of the proposal.

Slayer95:
> this is draft to review the protocol and API change

Protocol has already been discussed, but IIRC a draft cannot be marked as Approved?

MathyFurret:
Hey, researchers, I would like as many of these cases tested as possible before I wrap up the server code (may add stuff as me/others think of them):

### Both gens ###

- Is the order of rotations decided by the speed of the Pokemon that start in front, the Pokemon rotating in, or is it always random?
- Confirm order of rotation, hard switches, Slow Start activation
- What happens if P1 hard switches while P2 is rotating to a Pokemon using Pursuit?
- Imprison
- Mirror Move (and similar moves?)
- Are onAlly... abilities active for sub-active Pokemon? (Victory Star was introduced the same gen so very well could be)

### Gen 6 ###
- Confirm order of Primal Reversion alongside the above features

MathyFurret:
I think I want to look at look at https://github.com/smogon/pokemon-showdown/projects/3#card-69824553 because it probably goes along with the stuff I haven't implemented yet. That's in turn blocked until #8354 merges IMO, because it requires a refactor of how the Update event works.

sim/battle.ts
MathyFurret: i should have just looked at the old research threads lols https://www.smogon.com/forums/threads/b-w-research-thread.79340/post-3675294 pursuit should not interact if the user has to rotate in first. the code _should_ already be working fine because events are not triggered for sub-active Pokemon, but i'll add a test case for it

PR #8678, opened 2022-03-03
Refactor room games to use generic typing

MathyFurret:
Redux of #7716 apparently.

`RoomGame` is now an abstract class with a type parameter for the player class. Because the `makePlayer` method must call the player constructor method directly, it is now an abstract method. Games without their own player class now extend `SimpleRoomGame` which has a default `makePlayer` implementation.

This makes RoomGame subclasses ever so slightly neater: for example, they no longer need to override `players` and `playerTable`.

### DIFF
--- a/server/chat-plugins/blackjack.ts
+++ b/server/chat-plugins/blackjack.ts
@@ -17,7 +17,7 @@ type Deck =
 type Symbols = '♥' | '♦' | '♣' | '♠';
 type SymbolName = 'Hearts' | 'Diamonds' | 'Clubs' | 'Spades';
 
-export class Blackjack extends Rooms.RoomGame {
+export class Blackjack extends Rooms.RoomGame<BlackjackPlayer> {
 	room: Room;
 
 	roomID: RoomID;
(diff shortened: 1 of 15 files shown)


server/room-game.ts
```diff
@@ -16,9 +16,9 @@
  */
 
 // globally Rooms.RoomGamePlayer
-export class RoomGamePlayer {
+export class RoomGamePlayer<GameClass extends RoomGame = SimpleRoomGame> {
```
AnnikaCodes, line 19: I think it's a little simpler to name these `Game` and `Player` instead of `GameClass` and `PlayerClass`, but either way is fine.

AnnikaCodes (approved):
I really like this!

server/chat-plugins/blackjack.ts
```diff
@@ -485,7 +483,7 @@ export class Blackjack extends Rooms.RoomGame {
 		this.turnLog += turnLine;
 		if (player.cards.length > 2) this.display(turnLine, false, player.name);
 
-		if (player === this.dealer) {
+		if (player instanceof BlackjackDealer) {
```
KrisXV, line 486: is instanceof hotpatchable?

AnnikaCodes: Given that the entire plugin would be hotpatched I think it'd be ok.

AnnikaCodes:
Thanks!

PR #8679, opened 2022-03-04
Hackmons Cup: Fix custom un/bans with formes

MathyFurret:
This fixes two bugs when using a custom ruleset for Hackmons Cup:
- Pokemon with multiple formes weren't being added to the pool when whitelisted except for formes that *did* have a different `isNonstandard` value (e.g. a ruleset of `-allpokemon, +mewtwo` would *only* add the megas in Gen 8)
- For Pokemon with multiple formes, only the first forme was being added to the pool (e.g. a ruleset of `-allpokemon, +zapdos` would only allow Zapdos instead of having an equal chance to pick Zapdos or Zapdos-Galar)

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -547,7 +547,7 @@ export class RandomTeams {
 
 					banReason = ruleTable.check('basepokemon:' + toID(species.baseSpecies));
 					if (banReason) continue;
-					if (banReason !== '' || this.dex.species.get(species.baseSpecies).isNonstandard === species.isNonstandard) {
+					if (banReason !== '' || this.dex.species.get(species.baseSpecies).isNonstandard !== species.isNonstandard) {
 						const nonexistentCheck = Tags.nonexistent.genericFilter!(species) && nonexistentBanReason;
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
> For Pokemon with multiple formes, only the first forme was being added to the pool (e.g. a ruleset of -allpokemon, +zapdos would only allow Zapdos instead of having an equal chance to pick Zapdos or Zapdos-Galar)

Is this a bug? Should the formes be unbanned separately (i.e. `-allpokemon, +zapdos` for only Zapdos-Kanto and `-allpokemon, +zapdos, +zapdos-galar` for both) instead? Otherwise the PR is fine.

MathyFurret:
`+zapdos` resolves to `+basepokemon:zapdos` which unbans both. However even if you manually unban both, only the 1st would be able to show up -- the bug was that only 1 pokemon with a given dex no. would be added to the pool

WeWuzNidokangz:
Thank-you for noticing and fixing these problems!

AnnikaCodes:
Thanks!

PR #8680, opened 2022-03-04
Check for Sky Drop before resolving mega action

Karthik99999:


DaWoblefet:
This passes the current Sky Drop Mega Evolution test and looks a lot cleaner. Thanks Karthik!

PR #8684, opened 2022-03-07
Spotlights: Fix "Refresh" button opening another tab

MathyFurret:
This fixes a bug where the "Refresh" button in the default `/viewspotlights` page would open another tab instead of refreshing the current tab. (e.g. clicking it on `view-spotlights-lobby` would open another tab `view-spotlights-lobby-`)


AnnikaCodes:
Good catch, thanks!

PR #8685, opened 2022-03-07
Stop sending -formechange for Gmax

Karthik99999:
Client Side: smogon/pokemon-showdown-client#1946

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -708,8 +708,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 			if (['cramorantgulping', 'cramorantgorging'].includes(pokemon.species.id) && !pokemon.transformed) {
 				pokemon.formeChange('cramorant');
 			}
-			this.add('-start', pokemon, 'Dynamax');
-			if (pokemon.gigantamax) this.add('-formechange', pokemon, pokemon.species.name + '-Gmax');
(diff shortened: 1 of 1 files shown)


Zarel (approved):
Looks good; should be merged one restart after the client change.

data/conditions.ts
```diff
@@ -708,8 +708,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 			if (['cramorantgulping', 'cramorantgorging'].includes(pokemon.species.id) && !pokemon.transformed) {
 				pokemon.formeChange('cramorant');
 			}
-			this.add('-start', pokemon, 'Dynamax');
-			if (pokemon.gigantamax) this.add('-formechange', pokemon, pokemon.species.name + '-Gmax');
+			this.add('-start', pokemon, 'Dynamax', pokemon.gigantamax ? 'Gmax' : '');
```
Karthik99999, line 711: Even though this got approved already, would it be a good idea to only send "G" rather than "Gmax" (like how packed teams work)? Since the client will only care about the argument not being an empty string.

Zarel: I don't think this is necessary. The protocol doesn't pack quite this tightly anywhere else.

DaWoblefet:
Thanks Karthik!

PR #8688, opened 2022-03-15
Fix battle not ending after faint from future move

MathyFurret:
Fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9152590


MathyFurret:
xref https://github.com/smogon/pokemon-showdown/projects/3#card-69824150

Marty-D (approved):
Thanks!

PR #8691, opened 2022-03-15
Scavengers: Spam Filter twist

sparkychildcharlie:


server/chat-plugins/scavenger-games.ts
```diff
@@ -32,6 +32,13 @@ interface GameMode {
 	[k: string]: any;
 }
 
+function toSeconds(time: string) {
+	// hhmmss => ss
+	const parts = time.split(':').reverse();
+	const seconds = parts.map((value: string, index: number) => parseInt(value) * Math.pow(60, index));
+	return seconds.reduce((a, b) => a + b);
```
mia-pi-git, line 39:
```suggestion
	return parts
		.map((value: string, index: number) => parseInt(value) * Math.pow(60, index));
		.reduce((a, b) => a + b);
```

mia-pi-git: Also, are the typedefs in the map() callback necessary? Should be implicit.

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+			player.incorrect.push(id);
+		},
+
+		onComplete(player, time, blitz) {
+			const seconds = toSeconds(time);
+			if (!player.incorrect) return {name: player.name, total: seconds, blitz, time, original_time: time};
+
+			const deduction = 30 * player.incorrect.length;
+			const total = seconds + deduction;
+			const final_time = Chat.toDurationString(total * 1000, {hhmmss: true});
```
mia-pi-git, line 243:
```suggestion
			const finalTime = Chat.toDurationString(total * 1000, {hhmmss: true});
```

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+			if (total > 60) blitz = false;
+
+			return {name: player.name, total, blitz, time: final_time, original_time: time};
+		},
+
+		onConfirmCompletion(player, time, blitz, place, result) {
+			blitz = result.blitz;
+			time = result.time;
+			const deduction_message = player.incorrect?.length ?
```
mia-pi-git, line 252:
```suggestion
			const deductionMessage = player.incorrect?.length ?
```

server/chat-plugins/scavengers.ts
```diff
@@ -702,8 +702,12 @@ export class ScavengerHunt extends Rooms.RoomGame<ScavengerHuntPlayer> {
 		this.completed.push(result);
 		const place = formatOrder(this.completed.length);
 
-		this.runEvent('ConfirmCompletion', player, time, blitz);
-		this.announce(Utils.html`<em>${result.name}</em> has finished the hunt in ${place} place! (${time}${(blitz ? " - BLITZ" : "")})`);
+		const completion_message = this.runEvent('ConfirmCompletion', player, time, blitz, place, result);
```
mia-pi-git, line 705: Broken record, but camelCase over snake_case.

aQrator:
Confirming RO approval

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+				`${player.incorrect.length} incorrect ${Chat.plural(player.incorrect.length, "guesses", "guess")}` :
+				"Perfect!";
+			return `<em>${Utils.escapeHTML(player.name)}</em> has finished the hunt! (Final Time: ${time} - ${deduction_message}${(blitz ? " - BLITZ" : "")})`;
+		},
+
+		onEnd() {
+			this.completed.sort((a, b) => a.total - b.total);
```
mia-pi-git, line 259:
```suggestion
			Utils.sortBy(this.completed, entry => -entry.total)
```

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+			if (total > 60) blitz = false;
+
+			return {name: player.name, total, blitz, time: final_time, original_time: time};
+		},
+
+		onConfirmCompletion(player, time, blitz, place, result) {
+			blitz = result.blitz;
+			time = result.time;
+			const deduction_message = player.incorrect?.length ?
+				`${player.incorrect.length} incorrect ${Chat.plural(player.incorrect.length, "guesses", "guess")}` :
```
mia-pi-git, line 253:
```suggestion
				`${Chat.count(player.incorrect, 'incorrect guesses', 'incorrect guess'}` :
```

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+
+			player.incorrect.push(id);
+		},
+
+		onComplete(player, time, blitz) {
+			const seconds = toSeconds(time);
+			if (!player.incorrect) return {name: player.name, total: seconds, blitz, time, original_time: time};
+
+			const deduction = 30 * player.incorrect.length;
+			const total = seconds + deduction;
```
mia-pi-git, line 242:
```suggestion
			const total = seconds + (30 * player.incorrect.length);
```
Nitpick. Declaring variables you only use once is messier, imo.

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +220,46 @@ const TWISTS: {[k: string]: Twist} = {
...
+	spamfilter: {
+		name: 'Spam Filter',
+		id: 'spamfilter',
+
+		desc: 'Every wrong answer adds 30 seconds to your final time!',
+
+		onIncorrectAnswer(player: ScavengerHuntPlayer, value: string) {
+			if (!player.incorrect) player.incorrect = [];
+			const id = `${player.currentQuestion}-${value}`;
+			if (player.incorrect.includes(id)) return;
```
mia-pi-git, line 232: This wouldn't be every wrong answer, no? This would be every unique wrong answer.

mia-pi-git: Also, why not use a `Set`?

server/chat-plugins/scavenger-games.ts
sparkychildcharlie: probably not - i probably was just trying to hard convert a string to a number before and be lazy with types

server/chat-plugins/scavenger-games.ts
sparkychildcharlie: yeah that's correct, no need to beat a dead horse.  Let's not punish people for not using their brains (too much)

I don't code much so I never really looked into sets.  I assume it's kinda the same since Sets just have unique ids? I feel like it's fine the way it is now zzz

server/chat-plugins/scavenger-games.ts
sparkychildcharlie: I don't agree with this one.  It seems like a lot of extra steps looking at utils file. a.total is always type number, so there's no reason to use Utils which has a lot more type definition in it's comparative process.

mia-pi-git: The Utils differences don't matter much. This is just more legible / consistent (since sortBy is preferred)

server/chat-plugins/scavenger-games.ts
AnnikaCodes: The advantage to using a `Set` is that you wouldn't have to worry about forgetting the `if (player.incorrect.includes(id)) return;` line. Refactoring to use a `Set` can go in a separate PR, though.

server/chat-plugins/scavenger-games.ts
```diff
@@ -213,6 +219,45 @@ const TWISTS: {[k: string]: Twist} = {
...
+
+			player.incorrect.push(id);
+		},
+
+		onComplete(player, time, blitz) {
+			const seconds = toSeconds(time);
+			if (!player.incorrect) return {name: player.name, total: seconds, blitz, time, original_time: time};
```
AnnikaCodes, line 238: What if `player.incorrect` is an empty array?

server/chat-plugins/scavengers.ts
```diff
@@ -702,8 +702,12 @@ export class ScavengerHunt extends Rooms.RoomGame<ScavengerHuntPlayer> {
...
 		const place = formatOrder(this.completed.length);
 
-		this.runEvent('ConfirmCompletion', player, time, blitz);
-		this.announce(Utils.html`<em>${result.name}</em> has finished the hunt in ${place} place! (${time}${(blitz ? " - BLITZ" : "")})`);
+		const completionMessage = this.runEvent('ConfirmCompletion', player, time, blitz, place, result);
+		this.announce(
+			completionMessage ?
+				completionMessage :
+				Utils.html`<em>${result.name}</em> has finished the hunt in ${place} place! (${time}${(blitz ? " - BLITZ" : "")})`
+		);
```
AnnikaCodes, line 710:
```suggestion
		this.announce(
			completionMessage ||
			Utils.html`<em>${result.name}</em> has finished the hunt in ${place} place! (${time}${(blitz ? " - BLITZ" : "")})`
		);
```

sparkychildcharlie:
is it fine now @AnnikaCodes

server/chat-plugins/scavenger-games.ts
AnnikaCodes: Oh, hmm, it seems like this wouldn't affect anything aside from potentially letting `blitz` remain true even when the seconds was greater than 60.

AnnikaCodes:
Thanks!

PR #8693, opened 2022-03-18
Add a Dockerfile for running the server

brandon-layton:
Having a docker image to run the server with makes it simple for a lot of people to host there own server that they might want to use. Having to set this up natively or on a vm is not appealing for many people and may be a stopping point for some when they want to run there own server. A container will make running a server more accessible for people to do. 

I have looked at previous efforts and there seem to be old images and Dockerfiles out there but none that are kept up to date with the latest version. The best effort I saw was with this pull request #6184 and there were some good points to why it was not accepted. The biggest I saw was about the data in certain directories not being persistent. In my version of the Dockerbuild I added volumes for the directories that were stated to be a concern. Adding more as they are needed is also not a hard thing to do. 

I also have my Dockerfile get the latest release automatically from a git clone into the container. This also make the Dockerfile not have to be updated everytime a new release comes out.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,38 @@
+# Pokémon Showdown Docker
+
+This Dockerfile will build an image that can be used to run a docker container with. This application needs runs on port 8000, so you need to bind that port to your host. If you want to use a custom config you will also need to supply your own `config.js`, an example of this can be found at `config/config-example.js` of this repository.
+
+
+# Building & Running
+
+To build this image copy `Dockerfile` onto your computer to run with docker run or docker compose. You will need docker and/or docker-compose installed depending on what method you use. This Dockerfile will run the latest released version of Pokémon Showdown at the time  of building.
+
+##  Using docker run
+
+We will build the image first and then run it as a container. 
+While in the same directory this command will build the image and tag it as `pokemon-showdown`.
+```
+docker build . -t pokmon-showdown
+```
+
+After that we will want to run our image, publish port 8000, and add our custom config we have to the container. 
+```
+docker run -p 8000 -v /path/to/config.js:/config.js pokemon-showdown
+```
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -0,0 +1,38 @@
+# Pokémon Showdown Docker
+
+This Dockerfile will build an image that can be used to run a docker container with. This application needs runs on port 8000, so you need to bind that port to your host. If you want to use a custom config you will also need to supply your own `config.js`, an example of this can be found at `config/config-example.js` of this repository.
```
AnnikaCodes, line 3:
```suggestion
This Dockerfile will build an image that can be used to run a docker container with. This application needs to run on port 8000, so you need to bind that port on your host. If you want to use a custom config, you will also need to supply your own `config.js`; an example of this can be found at `config/config-example.js` in this repository.
```

docker/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+# Building & Running
+
+To build this image copy `Dockerfile` onto your computer to run with docker run or docker compose. You will need docker and/or docker-compose installed depending on what method you use. This Dockerfile will run the latest released version of Pokémon Showdown at the time  of building.
```
AnnikaCodes, line 8:
```suggestion
To build this image, copy `Dockerfile` onto your computer to run with `docker run` or `docker compose`. You will need `docker` and/or `docker-compose` installed, depending on which method you use. This Dockerfile will run the most recent version of Pokémon Showdown available on GitHub at the time of building.
```

docker/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+##  Using docker run
```
AnnikaCodes, line 10:
```suggestion
## Using `docker run`
```

AnnikaCodes (commented):
This seems fine, though I don't know how to write Dockerfiles and thus can't effectively review it.

mia-pi-git:

PR #8695, opened 2022-03-18
Fix some ability interactions with OMs

urkerab:
For instance, if you have a Drizzle team in Shared Power, team members sharing Drizzle will show `[object Object]` instead of Drizzle as the ability. Regular formats are unaffected.


KrisXV (approved):

Marty-D (approved):
Thanks! 👍

PR #8696, opened 2022-03-19
Gen 6: Fix Nature Power

Karthik99999:
Nature Power shouldn't be affected by Psychic Terrain in Gen 6

### DIFF
--- a/data/mods/gen6/moves.ts
+++ b/data/mods/gen6/moves.ts
@@ -152,6 +152,21 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		basePower: 65,
 	},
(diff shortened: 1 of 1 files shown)


KrisXV:
I don't see why this is an important thing to fix, obviously Psychic Terrain didn't exist in gen 6 but we have a ton of interactions implemented that dont exist until future generations that work in past gens for custom game etc. For example, should Dazzling just not do anything in gen 6?

Karthik99999:
You know what, I wasn't even aware that psychic terrain was not in Gen 6. I don't think this needs to be fixed lol I should've checked that

KrisXV:
Lol no worries, it happens

PR #8700, opened 2022-03-21
Mafia: Fix transitions between deathstates

ayiadvance:
No longer decrements number of players when someone transitions between death states, and no longer allows a certain type of kill to be used on a player already in that death state.


AnnikaCodes:
I'm not too knowledgeable about Mafia but this seems fine.

PR #8704, opened 2022-03-25
Scavengers: fix sorting direction

sparkychildcharlie:
@Zarel can you patch this after it goes in? since it's used daily by scavs

I think mia got the sort direction wrong when she made the suggestion


mia-pi-git:
Thanks!

PR #8705, opened 2022-03-26
/analysis fixes

Karthik99999:
- Uses the right format id for anything goes (https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9161077)

- Fixes link for pokemon with a forme in its name (https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9161082)


DaWoblefet:
Thanks Karthik!

PR #8708, opened 2022-04-01
Add Crazyhouse Rule

HisuianZoroark:


data/rulesets.ts
```diff
@@ -1604,6 +1604,79 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+		},
+		onFaint(target, source, effect) {
+			if (!target.m.numSwaps) {
+				target.m.numSwaps = 0;
+			}
+			target.m.numSwaps++;
+			if (effect && effect.effectType === 'Move' && source.side.pokemon.length < 24 &&
+                source.side !== target.side && target.m.numSwaps < 4) {
+				const Pokemon: typeof import('../sim/pokemon').Pokemon =
+                    require('../sim/pokemon').Pokemon;
```
AnnikaCodes, line 1637: Can't this be a top-level `import`?

data/rulesets.ts
```diff
@@ -1604,6 +1604,79 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+				const hpCost = this.clampIntRange(Math.floor((target.baseMaxhp * target.m.numSwaps) / 4), 1);
+				// Just in case(tm) and for Shedinja
+				if (hpCost === target.baseMaxhp) {
+					target.m.outofplay = true;
+					return;
+				}
+				source.side.pokemonLeft++;
+				source.side.pokemon.length++;
+
+				const newPoke = new Pokemon(target.set, source.side);
```
KrisXV, line 1647: Do you really need to use the Pokemon class here? can't you just deepClone the Pokemon?

HisuianZoroark: Calling deepClone causes a crash. This was the next best thing.

data/rulesets.ts
HisuianZoroark: It can if it needs to be, but as far as I know only crazyhouse uses this.

data/rulesets.ts
HisuianZoroark: Oh also, it does help immensely because it treats the new mon as as owned by the capturer, making adding it to the opponent's team easier to do without much hiccups.

urkerab: The old Gen 7 Super Staff Bros had code to swap Pokémon between sides without cloning them, but I don't know whether it's applicable to what you're trying to achieve here.

HisuianZoroark: > The old Gen 7 Super Staff Bros had code to swap Pokémon between sides without cloning them

can you point me to it? It could work but I couldn't find it after searching

urkerab: https://github.com/smogon/pokemon-showdown/blob/8768b2d7fb71840521f30e659d7fe717c3cd47f1/config/formats.js#L1941

HisuianZoroark: oh thanks! I can probably use that. The only problem might be pointing issues, but I am willing to give it a shot.

HisuianZoroark: > The old Gen 7 Super Staff Bros had code to swap Pokémon between sides without cloning them, but I don't know whether it's applicable to what you're trying to achieve here.
Trying to adapt the code into what I am trying to do causes an invalid switch error, which is why I have a clone to help prevent this sort of crash. It might be possible with more tinkering, but i can't think of any way off the top of my head.

HoeenCoder: Looks like deepClone crashes because Pokemon objects reference a Side object, which are have a reference loop: https://github.com/smogon/pokemon-showdown/blob/master/sim/side.ts#L66

Unsure where to go from here for fixing this, or if we even should fix infinite recursion in deepClone.

KrisXV:
I'm waiting to merge this until it gets a comment from zarel or marty since it modifies the battle.ts file; however, it is likely going to win OMotM so I will ping @Marty-D to take a quick look at the battle.ts file aspect to make sure it doesn't do anything bad that I might've missed.

HisuianZoroark:
> I'm waiting to merge this until it gets a comment from zarel or marty since it modifies the battle.ts file; however, it is likely going to win OMotM so I will ping @Marty-D to take a quick look at the battle.ts file aspect to make sure it doesn't do anything bad that I might've missed.

that's fine. The code in battle.ts is only temporary anyways.

KrisXV:
been 3 days and nothing so i'll merge

data/rulesets.ts
Zarel: Deep cloning isn't meant to be used for class instances in the first place.

data/rulesets.ts
```diff
@@ -1604,6 +1605,72 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+				if (hpCost === target.baseMaxhp) {
+					target.m.outofplay = true;
+					return;
+				}
+				source.side.pokemonLeft++;
+				source.side.pokemon.length++;
+
+				// A new Pokemon is created and stuff gets aside akin to a deep clone.
+				// This is because deepClone crashes when side is called recursively.
+				// Until a refactor is made to prevent it, this is the best option to prevent crashes.
```
Zarel, line 1648: This comment is false; this is a shallow clone, not a deep clone.

data/rulesets.ts
Zarel: We could get rid of the `Pokemon` dependency if we had am `addPokemon` function for it in `Side`.

Zarel: Why are you cloning in the first place? What do you want to carry over that isn't in the set?

data/rulesets.ts
```diff
@@ -1604,6 +1605,72 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+			if (effect && effect.effectType === 'Move' && source.side.pokemon.length < 24 &&
+                source.side !== target.side && target.m.numSwaps < 4) {
+				const hpCost = this.clampIntRange(Math.floor((target.baseMaxhp * target.m.numSwaps) / 4), 1);
+				// Just in case(tm) and for Shedinja
+				if (hpCost === target.baseMaxhp) {
+					target.m.outofplay = true;
+					return;
+				}
+				source.side.pokemonLeft++;
+				source.side.pokemon.length++;
```
Zarel, line 1644: Oh... oh no...

While this is technically valid JS, generally it's best not to mutate an array's length directly. PS uses `array.push` to add things to arrays, which is the usual way to do it.

PR #8713, opened 2022-04-04
Combine `lint` and `full-lint` NPM scripts

AnnikaCodes:
Now that [`lint` and `full-lint` use the same configuration](https://github.com/smogon/pokemon-showdown/commit/c57fcf25c69454bc8c9a21e183c75cfe75ed68a9), the only difference between them is the `--max-warnings` flag.

This separation causes some problems:
- In small terminal windows, warnings can be obscured by terminal scroll due to the large volume of unit tests. These warnings in turn cause CI failures since GitHub Actions runs `full-lint`.
- When changes to the linting script need to be made, **three** scripts need to be updated. This can lead to oversights such as `full-lint` not being updated for `.tsx` files when JSX support was added.

I've made this a PR in case anyone else has opinions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,13 +34,13 @@
     "build": "node build",
     "tsc": "tsc",
     "fast-lint": "eslint . --config .eslintrc-no-types.json --cache --cache-location .eslintcache-no-types --ext .js,.ts,.tsx",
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
Oh I didn't realise checks were failing here

PR #8714, opened 2022-04-05
Don't show /offertie hint in tour battles

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9175609


DaWoblefet:
Thanks Karthik!

PR #8715, opened 2022-04-06
Change wss URL in PROTOCOL.md

Nixola:
The previously shown URL (https://sim.smogon.com/showdown/websocket), along with sim3.smogon.com or sim.psim.us, doesn't work, since it serves a certificate that's only valid for sim3.psim.us, which is also the default domain currently used by the client.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -9,7 +9,7 @@ Showdown directly using WebSocket:
 
     ws://sim.smogon.com:8000/showdown/websocket
       or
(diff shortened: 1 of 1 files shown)


AnnikaCodes:
I think the purpose of the `sim.smogon.com` URL is to make sure the URL doesn't change when Showdown switches servers (there has been a `sim2` in the past). This looks fine as a short-term fix, but going forward it might be worth adding a certificate for `sim.smogon.com`?

PR #8719, opened 2022-04-10
End Dynamax before start of 4th turn

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-52058688


DaWoblefet (approved):
Went through and reviewed this in-depth with Karthik. While I think there's a decent chance Game Freak does the Speed sorting for Dynamax differently (they probably do it more than once), I think this implementation works well for our purposes.

PR #8722, opened 2022-04-17
Implement held Poke Ball legality

DaWoblefet:
Various Poke Balls have only been available at one point or another as held items:

- Cherish Ball and Park Ball have never been available
- Dream Ball and Safari Ball are only available as of Gen 8
- Beast Ball, Dream Ball, and Safari Ball aren't available in BDSP
- The Apricorn balls (Fast, Heavy, Level, Love, Lure, and Moon) are only available in Gens 2, 4, and 7-onwards
- Sport Ball is only available in Gens 2, 4, and 8

For Gems, Fairy Gem should be marked as ``Unobtainable``, while everything but Normal Gem should presumably be marked as ``Past``. Fairy Gem was added in Gen 6 but never released, whereas all the other Gems used to be in Gen 5. I know there's a test saying otherwise but I'm pretty sure they should be semantically marked as ``Past`` and the test itself is wrong (checking [Discord activity around the time](https://discord.com/channels/630837856075513856/630845310033330206/699491404690227232), I think pre just edited it to make it pass, back when we were using ``isUnreleased`` notation).

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -752,6 +752,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		num: 16,
 		gen: 4,
 		isPokeball: true,
+		isNonstandard: "Unobtainable",
 	},
 	chestoberry: {
 		name: "Chesto Berry",
@@ -3773,6 +3774,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		num: 500,
 		gen: 4,
 		isPokeball: true,
+		isNonstandard: "Unobtainable",
 	},
 	passhoberry: {
 		name: "Passho Berry",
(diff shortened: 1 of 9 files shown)


DaWoblefet:
> I think if it functions properly when hacked in it should be marked “Unobtainable” and otherwise be marked “Past”, regardless of past obtainability. Even with a different distinction, I don’t think any of them could apply to Fairy Gem in a way that makes it different from Rock Gem in gens 6+. For instance, Floette-Eternal is currently marked as “Past” despite never being released, so at the very least if the other gems are “Past” then Fairy Gem should be too.

I think this is the way we're supposed to use it, so I've corrected Gems and Poke Balls to use "Past" and "Unobtainable" appropriately. Gems other than Normal Gem do not function in Gen 8, but they will function just fine in Gens 6-7.

PR #8726, opened 2022-04-21
Fix Red Card's interaction with Life Orb and Shell Bell

### LINKED ISSUE #8504: Shell Bell vs Red Card interaction
https://replay.pokemonshowdown.com/gen8doublesou-1441038151-w0989yl4fnw1p0xgts3uq9qdn92zlz1pw
https://replay.pokemonshowdown.com/gen8ou-1441038726-4f78s8oai7gj21cn35o2ejwm1ckvg97pw

As you can see in these two replays, Shell Bell activates before switching out due to Red Card. Per the Bulbapedia wiki, "The Red Card switches out the attacker before its own held Life Orb or Shell Bell has a chance to activate." (so the way Red Card interacts with Life Orb is also probably affected!)

https://bulbapedia.bulbagarden.net/wiki/Red_Card

mdebackr:
Manually tested with odd interactions with Red Card and Life Orb/Shell Bell, including the use of ingrain or if the attacking pokemon is the last remaining pokemon on its side. Removed import from npm test in my previous pull request.

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -2905,7 +2905,7 @@ export const Items: {[itemid: string]: ItemData} = {
 			return this.chainModify([5324, 4096]);
 		},
 		onAfterMoveSecondarySelf(source, target, move) {
-			if (source && source !== target && move && move.category !== 'Status') {
+			if (source && source !== target && move && move.category !== 'Status' && !source.forceSwitchFlag) {
 				this.damage(source.baseMaxhp / 10, source, source, this.dex.items.get('lifeorb'));
 			}
 		},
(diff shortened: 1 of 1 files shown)


MathyFurret:
Next time instead of closing and opening your PR, just ask for help on discord or PS, we can help you neaten up your commits if needed.

DaWoblefet:
I believe this would technically work, but I'll let Marty give his comments of doing it this way instead of modifying the event. Post-move check success needs a thorough review anyway, so I think this would an OK solution.

PR #8727, opened 2022-04-21
Fix extraneous failure messages

MathyFurret:
https://github.com/smogon/pokemon-showdown/projects/3#card-69826019

for Defense Curl and Rest+Uproar

Test:
image


DaWoblefet:
I believe Minimize does the same thing too if you want to lump that in. Those are the only 3 I know offhand.

DaWoblefet:
Thanks Mathy!

PR #8729, opened 2022-04-22
Reveal hacked Zacian/Zamazenta forme in Team Preview

MathyFurret:
https://github.com/smogon/pokemon-showdown/projects/3#card-69824841

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -444,8 +444,12 @@ export const Rulesets: {[k: string]: FormatData} = {
 		onTeamPreview() {
 			this.add('clearpoke');
 			for (const pokemon of this.getAllPokemon()) {
-				const details = pokemon.details.replace(', shiny', '')
+				let details = pokemon.details.replace(', shiny', '')
 					.replace(/(Arceus|Gourgeist|Pumpkaboo|Xerneas|Silvally|Urshifu)(-[a-zA-Z?-]+)?/g, '$1-*');
+				if (this.ruleTable.has('obtainableformes')) {
+					// Crowned formes are revealed in Hackmons, but not in regular play.
+					details = details.replace(/(Zacian|Zamazenta)(-[a-zA-Z?-]+)?/g, '$1-*');
+				}
 				this.add('poke', pokemon.side.id, details, '');
 			}
 			this.makeRequest('teampreview');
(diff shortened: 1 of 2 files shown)


data/rulesets.ts
```diff
@@ -445,7 +445,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 			this.add('clearpoke');
 			for (const pokemon of this.getAllPokemon()) {
 				const details = pokemon.details.replace(', shiny', '')
-					.replace(/(Arceus|Gourgeist|Pumpkaboo|Xerneas|Silvally|Zacian|Zamazenta|Urshifu)(-[a-zA-Z?-]+)?/g, '$1-*');
```
Karthik99999, line 448: You might not be done with the pr yet, but removing Zacian/Zamazenta from this list would reveal their formes in regular formats like VGC.

MathyFurret:
Alright, there was more to this than I thought. I haven't played SwSh yet so I'm a bit unfamiliar with how the item works.

Looking into it more, it seems that the auto-conversion to Crowned forme is done in `TeamValidator#validateForme` method, which is done iff `ruleTable.has('obtainableformes')`. Would the most sensible solution just be to repeat this check to decide whether to hide the forme in team preview?

DaWoblefet:
This is almost correct, just one edge case missing - a regular Zacian/Zamazenta, holding Rusted Sword/Shield in hackmons should also be regular Zacian/Zamazenta at Team Preview. Once it hits the battle, it will in fact turn into its Crowned form, but at team preview it is its Hero form. See the comments on https://github.com/smogon/pokemon-showdown/pull/8261 for more info.

MathyFurret:
I'm debating how I want to go about doing this. Particularly with people who input a Crowned holding Rusted Thing in Hackmons when they didn't know they could hack the Crowned form in. 

All the other cases are pretty clear-cut: it would show Hero in Team Preview (and be denoted by the asterisk) unless it's a hacked Crowned

DaWoblefet:
I was stuck on this too on Karthik's PR. We already send that the Pokemon is Crowned by the time of Team Preview, but we don't know if it's a legitimate Crowned or a hacked Crowned. You can't even check if the Pokemon is holding the correct Rusted item, because you could always just hack in a Zamazenta-Crowned holding Rusted Shield.

Like I said in that PR, you probably couldn't do it accurately without getting rid of the convenience feature. If you can't think of anything else, I'll just merge as-is, since your change makes it more correct than it is right now.

MathyFurret:
The correct fix probably involves handling zac/zama later than team validation

DaWoblefet:
Superseded by https://github.com/smogon/pokemon-showdown/pull/8753; thank you for the ground work and tests!

PR #8730, opened 2022-04-22
Proc hazards in the order they were set

MathyFurret:
https://github.com/smogon/pokemon-showdown/projects/3#card-23557574

Entry hazards now use a new `EntryHazard` event, which is run immediately after `SwitchIn` and uses a stable sort for its resolution. (String keys in objects are guaranteed to remain in insertion order, and `Array#sort` is guaranteed stable.)

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -451,12 +451,6 @@ export const Conditions: {[k: string]: ConditionData} = {
 			return this.chainModify([5325, 4096]);
 		},
 	},
-	entryhazard: {
-		// this is a side condition
-		onSwitchIn(pokemon) {
-			this.runEvent('EntryHazard', pokemon);
-		},
-	},
 
 	// weather is implemented here since it's so important to the game
 
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -451,6 +451,12 @@ export const Conditions: {[k: string]: ConditionData} = {
 			return this.chainModify([5325, 4096]);
 		},
 	},
+	entryhazard: {
+		// this is a side condition
+		onSwitchIn(pokemon) {
+			this.runEvent('EntryHazard', pokemon);
+		},
+	},
 
 	// weather is implemented here since it's so important to the game
 
--- a/data/items.ts
+++ b/data/items.ts
@@ -2349,7 +2349,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		},
 		num: 1120,
 		gen: 8,
-		// Hazard Immunity implemented in moves.js
+		// Hazard Immunity implemented in BattleActions#runSwitch
 	},
 	helixfossil: {
 		name: "Helix Fossil",
--- a/data/items.ts
+++ b/data/items.ts
@@ -2349,7 +2349,7 @@ export const Items: {[itemid: string]: ItemData} = {
 		},
 		num: 1120,
 		gen: 8,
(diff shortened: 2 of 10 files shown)


Karthik99999:
The extra condition seems a bit unnecessary to me. Why not just call the `onEntryHazard` event everywhere the `onSwitchIn` event is called?

DaWoblefet:
> Actually this raises an interesting question: does Court Change preserve the order of hazards across sides?
Pretty sure yes re: Court Change, but I'll double check to make sure.

You could probably move Heavy-Duty Boots code to the generic ``entryhazard`` condition.

All the tests relevant to the following pass, but I figure it's relevant to the implementation so I'll post it here anyway: entry hazards are a part of Game Freak's equivalent of onSwitchIn, along with the following effects (this list is as of Gen 7):
<details>
<summary>List of effects in Game Freak's onSwitchIn equivalent</summary>
Slot effects:
Lunar Dance
Healing Wish
Z-Memento / Z-Parting Shot

Side effects:
Stealth Rock
Spikes
Toxic Spikes
Sticky Web

Abilities:
Intimidate
Slow Start
Limber
Insomnia
Magma Armor
Immunity
Water Veil
Drizzle
Drought
Sand Stream
Snow Warning
Primordial Sea
Desolate Land
Delta Stream
Air Lock / Cloud Nine
Trace
Download
Forewarn
Anticipation
Frisk
Mold Breaker
Pressure
Unnerve (note: typically Unnerve will activate in the event before Game Freak's onSwitchIn)
Imposter
Dark Aura
Fairy Aura
Aura Break
Water Bubble
Electric Surge
Psychic Surge
Misty Surge
Grassy Surge
Comatose

Items:
Cheri
Chesto
Rawst
Aspear
Persim
Pecha
Lum
Leppa
Oran
Berry Juice
Sitrus
Figy
Wiki
Mago
Aguav
Iapapa
Liechi
Ganlon
Salac
Petaya
Apicot
Lansat
Starf
Micle
Mental Herb
Amulet Coin
Air Balloon
</details>

Then, the [order of modifier application within a given event](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-8685465) (which is broadly not implemented generally on the sim yet) would cause these effects to activate for each Pokemon in order of their Speed. In this case, it's slot effects > side effects (hazards) > Abilities > items.

So things like this will happen as a result: if Choice Scarf Landorus-T and Tapu Koko simulatenously switch in, with Stealth Rock active, first Stealth Rock damages Landorus-T, then Intimidate activates, then Stealth Rock damages Tapu Koko, then Electric Surge activates. Another simpler example is that Healing Wish activates before hazard damage.

MathyFurret:
actually, if all hazards are grouped between slot effects and abilities, that sounds like a reason we _should_ do what karthik said

DaWoblefet:
> Actually this raises an interesting question: does Court Change preserve the order of hazards across sides?

Confirming it does preserve the order as suspected.

MathyFurret:
> You could probably move Heavy-Duty Boots code
does this mean Poison type with HDB doesn't remove tspikes? because as it was coded before, they would

DaWoblefet:
> does this mean Poison type with HDB doesn't remove tspikes? because as it was coded before, they would

Nope they do; I forgot about that, so it is best to leave the HDB implementation alone.

DaWoblefet:
Thanks a bunch Mathy!

PR #8731, opened 2022-04-22
Fix damage from future attacks

wimeng:
[https://github.com/smogon/pokemon-showdown/projects/3#card-70965392](https://github.com/smogon/pokemon-showdown/projects/3#card-70965392)

This adds a new attribute ```recentForme``` to ```ActiveMove```, letting it store the forme of the Pokemon which used the move. This variable is updated with the latest form if the future attack's user is still on the field. The code now passes the previously skipped unit test in futuresight.js.

### DIFF
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -1604,6 +1604,7 @@ export class BattleActions {
 			defBoosts = 0;
 		}
 
+		if (move.isFutureMove && !source.isActive && move.recentForme) attacker.setSpecies(move.recentForme);
 		let attack = attacker.calculateStat(attackStat, atkBoosts);
 		let defense = defender.calculateStat(defenseStat, defBoosts);
 
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -3477,6 +3477,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 					effectType: 'Move',
 					isFutureMove: true,
 					type: 'Steel',
+					recentForme: source.species,
 				},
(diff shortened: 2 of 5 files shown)


DaWoblefet (approved):
I like this! I'll let Marty give his thoughts.

sim/battle-actions.ts
```diff
@@ -1604,6 +1604,7 @@ export class BattleActions {
 			defBoosts = 0;
 		}
 
+		if (move.isFutureMove && !source.isActive && move.recentForme) attacker.setSpecies(move.recentForme);
```
Slayer95, line 1607: Doesn't this mean that a Ditto that transformed into Jirachi to use Doom Desire and then was switched out, after hitting the Ditto will go back to the field as a Jirachi?

Slayer95:
According to the card, stat stages should also be stored, right?

wimeng:
I believe that both stat stages and abilities should not be used in damage calculation when the user is not on the field. There are corresponding unit tests in futuresight.js which test for these.

sim/battle-actions.ts
wimeng: It appears that this isn't the case when I run a server using this code locally. I'm not entirely certain why this is, I would guess that there are certain mechanisms for changing a Pokemon back to its base forme upon switching in.

AnnikaCodes:
Thanks! I've merged this for now; if there is an issue regarding stat stage storage we can fix that in a follow-up PR.

PR #8732, opened 2022-04-23
Gen 1: Fix Transform bugs

Karthik99999:
- Fix copying stat boosts (https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8843431)
- Fix crit rate not using base species for speed stat (https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/post-6283457)

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1174,7 +1174,15 @@ export class Pokemon {
 		}
 		let boostName: BoostID;
 		for (boostName in pokemon.boosts) {
-			this.boosts[boostName] = pokemon.boosts[boostName]!;
+			this.boosts[boostName] = pokemon.boosts[boostName];
+			if (this.battle.gen <= 1) {
+				if (boostName === 'evasion' || boostName === 'accuracy') continue;
+				if (this.boosts[boostName] >= 0) {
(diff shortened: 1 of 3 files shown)


sim/pokemon.ts
```diff
@@ -1174,7 +1174,15 @@ export class Pokemon {
 		}
 		let boostName: BoostID;
 		for (boostName in pokemon.boosts) {
-			this.boosts[boostName] = pokemon.boosts[boostName]!;
+			this.boosts[boostName] = pokemon.boosts[boostName];
+			if (this.battle.gen <= 1) {
```
DaWoblefet, line 1178: I would just do equals gen 1, there's nothing prior to that.

Karthik99999: there was other code in this file that does `<= 1` which is why I did that (`=== 1` and `<= 1` are used pretty interchangeably throughout the entire codebase though so 🤷‍♂️)

DaWoblefet: Idk why we'd do it that way, but I suppose that can be a thing for another day.

DaWoblefet:
Thanks as always, Karthik!

PR #8733, opened 2022-04-24
Fix Focus Punch/Beak Blast/Shell Trap charge timing

Karthik99999:
https://github.com/smogon/pokemon-showdown/projects/3#card-61383963

### DIFF
--- a/sim/battle-queue.ts
+++ b/sim/battle-queue.ts
@@ -18,7 +18,7 @@ import type {Battle} from './battle';
 /** A move action */
 export interface MoveAction {
 	/** action type */
-	choice: 'move' | 'beforeTurnMove';
+	choice: 'move' | 'beforeTurnMove' | 'priorityChargeMove';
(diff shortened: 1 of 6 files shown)


sim/battle-queue.ts
```diff
@@ -18,7 +18,7 @@ import type {Battle} from './battle';
 /** A move action */
 export interface MoveAction {
 	/** action type */
-	choice: 'move' | 'beforeTurnMove';
+	choice: 'move' | 'beforeTurnMove' | 'priorityChargeMove';
 	order: 3 | 5 | 200 | 201 | 199;
```
urkerab, line 22: This list should probably have your new order here.

Personally I'm not a fan of the code duplication that Gen 4 requires. Is it possible keep the `beforeTurnMove` action but just change its choice and/or order in Gen 4?

Marty-D (approved):
Nice, thanks! 👍

PR #8735, opened 2022-04-24
Fix normal move target resolution in Triple Battles

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8973895

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2230,7 +2230,10 @@ export class Battle {
 				// even if a move can target an ally, auto-resolution will never make it target an ally
 				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your ally
 				const adjacentFoes = pokemon.adjacentFoes();
-				return adjacentFoes.length ? this.sample(adjacentFoes) : null;
+				if (adjacentFoes.length) {
+					return this.sample(adjacentFoes);
+				}
+				return pokemon.side.foe.active[pokemon.side.foe.active.length - 1 - pokemon.position];
 			}
 		}
 		return pokemon.side.randomFoe() || pokemon.side.foe.active[0];
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2228,11 +2228,9 @@ export class Battle {
 		if (this.activePerHalf > 2) {
 			if (move.target === 'adjacentFoe' || move.target === 'normal' || move.target === 'randomNormal') {
 				// even if a move can target an ally, auto-resolution will never make it target an ally
-				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your all
+				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your ally
 				const adjacentFoes = pokemon.adjacentFoes();
-				if (adjacentFoes.length) return this.sample(adjacentFoes);
-				// no valid target at all, return a possibly-fainted foe for any possible redirection
-				return pokemon.side.foe.active[0];
+				return adjacentFoes.length ? this.sample(adjacentFoes) : null;
 			}
 		}
 		return pokemon.side.randomFoe() || pokemon.side.foe.active[0];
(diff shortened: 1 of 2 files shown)


sim/battle.ts
```diff
@@ -2228,11 +2228,9 @@ export class Battle {
 		if (this.activePerHalf > 2) {
 			if (move.target === 'adjacentFoe' || move.target === 'normal' || move.target === 'randomNormal') {
 				// even if a move can target an ally, auto-resolution will never make it target an ally
-				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your all
+				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your ally
 				const adjacentFoes = pokemon.adjacentFoes();
-				if (adjacentFoes.length) return this.sample(adjacentFoes);
-				// no valid target at all, return a possibly-fainted foe for any possible redirection
-				return pokemon.side.foe.active[0];
+				return adjacentFoes.length ? this.sample(adjacentFoes) : null;
```
urkerab, line 2233: Is it safe to return `null` here? The old code (i.e. before #7929 broke it) returned the slot opposite the attacker.
https://github.com/smogon/pokemon-showdown/blob/39f5717c255cef59ca32c9351b5919032729e064/sim/battle.ts#L2056-L2062

Karthik99999: Given that all the tests pass I would like to say yes. This code only runs for triple battles and something like `any` will skip to the last return statement where `Side#randomFoe()` is called.

Karthik99999: So would the storm drain user be in range or out of range? and should the storm drain successfully retarget the move?

Karthik99999: Ok I did that test locally and yeah Storm Drain isn't redirecting when it should. I'm not really sure what you would return here then.

Slayer95: Return the Pokémon in front of the attacker. The old code needs to be properly updated to account for more than 2 sides though.

sim/battle.ts
```diff
@@ -2228,11 +2228,12 @@ export class Battle {
 		if (this.activePerHalf > 2) {
 			if (move.target === 'adjacentFoe' || move.target === 'normal' || move.target === 'randomNormal') {
 				// even if a move can target an ally, auto-resolution will never make it target an ally
-				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your all
+				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your ally
 				const adjacentFoes = pokemon.adjacentFoes();
-				if (adjacentFoes.length) return this.sample(adjacentFoes);
-				// no valid target at all, return a possibly-fainted foe for any possible redirection
```
Slayer95, line 2234: I don't understand why you'd want to remove this comment. On the contrary, it should be expanded upon to prevent this issue from happening again.

sim/battle.ts
```diff
@@ -2228,11 +2228,11 @@ export class Battle {
 		if (this.activePerHalf > 2) {
 			if (move.target === 'adjacentFoe' || move.target === 'normal' || move.target === 'randomNormal') {
 				// even if a move can target an ally, auto-resolution will never make it target an ally
-				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your all
+				// i.e. if both your opponents faint before you use Flamethrower, it will fail instead of targeting your ally
 				const adjacentFoes = pokemon.adjacentFoes();
 				if (adjacentFoes.length) return this.sample(adjacentFoes);
-				// no valid target at all, return a possibly-fainted foe for any possible redirection
-				return pokemon.side.foe.active[0];
+				// no valid target at all, return slot directly across for any possible redirection
```
Slayer95, line 2234:
```js
			// copies across in doubles/triples
			// (also copies across in multibattle and diagonally in free-for-all,
			// but side.foe already takes care of those)
const target = pokemon.side.foe.active[pokemon.side.foe.active.length - 1 - pokemon.position];
```

https://github.com/smogon/pokemon-showdown/blob/4291ccb51bcee3f33c4b18b136f92dc439b65143/data/abilities.ts

The comment is inaccurate, but the implementation looks good.

Marty-D (approved):
Nice, thanks! 👍

PR #8738, opened 2022-04-26
Fix Pursuit activating on non-adjacent foes in Triples

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9117672


Marty-D (commented):
Neat, thanks! 👍

PR #8739, opened 2022-04-27
Fix Slow Start not halving Sp. Atk for special Z-Moves

Karthik99999:
> If the user holds a typed Z-Crystal and uses a Z-Move based on a special move, Slow Start will also reduce the Special Attack stat for the duration of that move only.

This comes from Bulbapedia and it sounds like I should also be checking `isZOrMaxPowered` but I wasn't sure so I didn't add that check.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3428,11 +3428,6 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			onModifyAtk(atk, pokemon) {
 				return this.chainModify(0.5);
 			},
-			onModifySpA(spa, pokemon, target, move) {
-				if (move.isZ) {
-					return this.chainModify(0.5);
-				}
-			},
 			onModifySpe(spe, pokemon) {
 				return this.chainModify(0.5);
 			},
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -3428,6 +3428,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			onModifyAtk(atk, pokemon) {
 				return this.chainModify(0.5);
 			},
+			onModifySpA(spa, pokemon, target, move) {
+				if (move.isZ) {
+					return this.chainModify(0.5);
+				}
+			},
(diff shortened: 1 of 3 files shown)


DaWoblefet:
Yep, here's [cart footage too](https://youtu.be/4s3AQmC6BOQ?t=1620) since I remember finding this one really funny.

> This comes from Bulbapedia and it sounds like I should also be checking `isZOrMaxPowered` but I wasn't sure so I didn't add that check.

Nope, Slow Start doesn't halve Special max moves.

Karthik99999:
> Nope, Slow Start doesn't halve Special max moves.

I was thinking more along the lines of hacked z-moves, but I'm pretty sure now that slow start should still decrease sp. atk for those.

DaWoblefet:
> Bulbapedia specifies that this happens with typed Z crystals off of special moves. Does it not happen with, say, Stoked Sparksurfer or other special Z moves that don’t have a physical counterpart?

This is correct; if the move is not marked as physical in its data, it should not be halved. So Stoked Sparksurfer, Light that Burns the Sky, etc. should never be halved by Slow Start.

In Gen 7, the way the game does the check for Slow Start halving the attacking stat is by seeing if the category in move data is special, rather than if the move is in fact right now physical/special (like Huge Power or Muscle Band would). So Photon Geyser and LTBTS will always ignore Slow Start. In Gen 7, Sunsteel Strike and friends ignored abilities, including your own ability. But if you use a called Sunsteel Strike and friends, e.g. via Sleep Talk, then you won't have the Mold Breaker properties. So a called physical Slow Start Photon Geyser _should_ halve the user's Attack during damage calculation but it in fact does not.

In Gen 8, Sunsteel Strike and friends no longer ignore your own abilities, and according to Zaggyo Slow Start can weaken both physical Photon Geyser and physical Shell Side Arm now. So if Z-moves were in Gen 8 I'd assume they wouldn't have this behavior there either.

DaWoblefet (approved):
Looks good to me, thanks for adding the extra test!

PR #8740, opened 2022-04-28
Friends: Hide advanced commands in help broadcast

MathyFurret:
99% of people will just use the GUI to do the things under the cut, and `!help friends` is something that's broadcasted quite frequently to introduce people to the feature, so this will be far more user-friendly.


mia-pi-git:
LGTM.

mia-pi-git:
Thanks!

PR #8741, opened 2022-04-28
Fix choice lock after dynamax ends

MathyFurret:
Dynamax ending check was moved to happen before checking disabled moves.


DaWoblefet:
Thanks for the fix and unit test Mathy!

PR #8753, opened 2022-05-06
Properly handle Zacian/Zamazenta Crowned formes

Karthik99999:
- Crowned formes will now use `battleOnly` instead of `changesFrom`, which means the validator will handle them similarly to other battle-only formes.
- Before the first switch-ins, base forme Zacian/Zamazenta holding Rusted Sword/Shield will have their species changed to their Crowned formes, abilities changed to Intrepid Sword/Dauntless Shield, and Iron Head changed to Behemoth Blade/Bash.
- This also fixes issue that Zacian/Zamazenta had with Team Preview in Hackmons. Now only the base forme Zacian/Zamazenta will have `-*` appended to their name at Team Preview, while Crowned forme will be shown normally, as you can no longer have Crowned forme at Team Preview unless `Obtainable Formes` is not in the ruleset.
- Most importantly, the convenience feature of being able to select Crowned formes and Behemoth moves in the teambuilder will still work

Drafting for now in case I missed something in my species-changing code.

This also removes the need for #8729

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -681,25 +681,6 @@ export const Formats: FormatList = [
 			'Water Bubble', 'Wonder Guard', 'Comatose + Sleep Talk', 'Rusted Sword', 'Court Change', 'Bolt Beak', 'Double Iron Bash',
 			'Octolock', 'Shell Smash',
 		],
-		onChangeSet(set) {
-			const item = this.dex.toID(set.item);
-			if (set.species === 'Zacian' && item === 'rustedsword') {
-				set.species = 'Zacian-Crowned';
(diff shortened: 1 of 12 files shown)


MathyFurret:
feel free to adopt my unit tests from 8729

Karthik99999:
Ok I think my code should be correct, only question I have now is do I remove `forcedForme` from Rusted Sword/Rusted Shield? I feel like I should since the other items that used `forcedForme` change the Pokemon's forme outside of the battle but I'm not 100% sure.

DaWoblefet:
Tested all the edge cases I know and it looks good. Thanks Karthik!

PR #8756, opened 2022-05-06
Fix Burning Jealousy burning Pokemon that just switched in

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9208445


DaWoblefet:
Thanks Karthik!

PR #8757, opened 2022-05-07
fix Spikes missing against flying/digging semi-invulnerable opponent (gen 2)

Rezzo64:
Just fixing the bug I reported as below:

(https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-74#post-9187155)

Double commit to remove the compiler warning on line length for line 255


Marty-D:
Good catch, thanks! Just backporting the target handling from future gens fixes this issue for Spikes and any other relevant moves people feel like playing with in a Gen 2 environment, without adding them to the whitelists.

PR #8761, opened 2022-05-08
Validate tour rules in generator before setting it

tmagicturtle:
Hi, I'm sure the PR name doesn't really make sense. This fixes the bug I posted here:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9210558

Basically, prior to this fix, this.fullFormat would get set, and THEN it would check if it would work with formats with generated teams. What happens is, if it doesn't work with those formats, then it throws an error, but this.fullFormat remains, and all players get a "team generator (random) failed using these rules and seed (6503,27336,63804,42074)" error.

This fix holds off on setting this.fullFormat until after generated team formats are checked.


DaWoblefet:
Thanks fart!

PR #8763, opened 2022-05-09
Pokebilities AAA: Fix Traced Primal weather not ending

WeWuzNidokangz:
Bug reported in several places:-
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-76#post-9208581
https://replay.pokemonshowdown.com/gen8pokebilitiesaaa-1570487910


WeWuzNidokangz:
Faint time is also a problem. removeVolatile can't be used here because it doesn't pass the hp check or the status.id check.

PR #8764, opened 2022-05-09
Pokebilities AAA: Fix Sheer Force interaction with abilities that add secondaries

WeWuzNidokangz:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-76#post-9207294

Shoutouts to mathfreak for the detailed analysis.

~~This change makes Sheer Force take priority deterministically. Although we can't know for certain how SF 'should' interact with abilities that add secondary effects, it seems logical that affected moves would function like regular SF-compatible moves.~~

I changed my mind after looking into this further:-

https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/page-62#post-8905424

Based on this, it seems that we _can_ know that Stench would activate after Sheer Force on cart if you could somehow have both. Poison Touch still seems to be undetermined behavior but given the above, I think it is more coherent and should be more intuitive to users if it works the same way, with SF always losing in priority when it interacts with abilities that add secondary effects.

### DIFF
--- a/data/mods/pokebilities/abilities.ts
+++ b/data/mods/pokebilities/abilities.ts
@@ -220,6 +220,11 @@ export const Abilities: {[k: string]: ModdedAbilityData} = {
 			}
 		},
 	},
+	sheerforce: {
+		inherit: true,
+		// Activate after abilities that can add a secondary effect
+		onModifyMovePriority: -3,
+	},
 	trace: {
 		inherit: true,
 		onUpdate(pokemon) {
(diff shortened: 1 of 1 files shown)


WeWuzNidokangz:
Hisuan Zoroark mentioned King's Rock as a precedent (activates after SF) which at the least would create a new bug if SF is moved to this priority, so drafting for now to get more opinions on this.

PR #8769, opened 2022-05-12
Metronome fixes

Karthik99999:
- Add Dragon Hammer to the list of uncallable moves in SwSh (https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-9213663)
- Allow `Unobtainable` moves to be called (fixes Psycho Boost not being called by Metronome in BDSP)
- Prevent Metronome from calling moves the user already knows in gens 2 and 4


Marty-D (approved):
Thanks! 👍

PR #8774, opened 2022-05-14
Pokebilities: Fix Intimidate counters

WeWuzNidokangz:
Bug report: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-76#post-9205832
Replays: https://replay.pokemonshowdown.com/gen8pokebilities-1566499155-msrutjsdz77jinmo32cqlmy1swgaajspw
https://replay.pokemonshowdown.com/gen8pokebilitiesaaa-1573974793-fg5n05q18nzvsxk8ky8mfdibsww7558pw

Intimidate can activate from one side's onSwitchIn before the other side has received its counter-ability volatile from its own onSwitchIn, so it seems like at least some abilities need to be applied before this. However, just moving all addVolatiles to onBeforeSwitchIn would make the timing of abilities that are announced on switch-in incorrect.

It would be good if there is a less janky way to address this, especially since it's possible that there might be an ability that needs to passively exist during the opponent's switch-in that also announces on the user's switch in, although I couldn't think of an extant example of one.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1723,7 +1723,7 @@ export const Formats: FormatList = [
 			// handle switch-in ability-to-ability interactions, e.g. Intimidate counters
 			const neededBeforeSwitchInIDs = [
 				'clearbody', 'competitive', 'contrary', 'defiant', 'fullmetalbody', 'hypercutter', 'innerfocus',
-				'mirrorarmor', 'oblivious', 'owntempo', 'rattled', 'scrappy', 'simple', 'whitesmoke'
+				'mirrorarmor', 'oblivious', 'owntempo', 'rattled', 'scrappy', 'simple', 'whitesmoke',
 			];
 			if (pokemon.m.innates) {
 				for (const innate of pokemon.m.innates) {
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1718,10 +1718,26 @@ export const Formats: FormatList = [
 					.filter(ability => ability !== pokemon.ability);
 			}
 		},
+		onBeforeSwitchIn(pokemon) {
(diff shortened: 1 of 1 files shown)


WeWuzNidokangz:
Bump?

KrisXV:
Sorry, i missed this, ill take a glance now

KrisXV:
@WeWuzNidokangz is there way to make the onSwitchInPriority higher without interfering with something to guarantee everything has its abilities before stuff activates?

WeWuzNidokangz:
> @WeWuzNidokangz is there way to make the onSwitchInPriority higher without interfering with something to guarantee everything has its abilities before stuff activates?

Just increasing priority here globally doesn't help because the sides are racing each other's version of the same callback. I think if the onSwitchInPriority was changed from being a fixed number to a method that returns a different priority depending on which specific abilities the mons have, it would be similar but worse than what is happening with onBeforeSwitchIn in this change because I don't think it could correctly handle situations where both sides have both Intimidate and an Intimidate counter-ability. 

Looking at it more and comparing to how the real abilities work, I think the deeper problem is that volatiles run the Start event as soon as they are added, whereas real abilities passively exist and can react to other abilities before they have been Started themselves in runSwitch, meaning that they don't need to be created in a staggered way like this to interact correctly. So I think the 'proper' way to handle it would be for OMs like this to not use volatiles, or for volatiles to have more general functionality so that that they could behave more similarly to abilities(?), either of which seems like a really big change.

WeWuzNidokangz:
@KrisXV

AnnikaCodes:
@KrisXV are you still interested in reviewing this?

PR #8776, opened 2022-05-14
Gen 1: Fix (High) Jump Kick interaction with Ghost-types

KrisXV:
Jump Kick/High Jump Kick should deal 1 HP recoil upon fail regardless of the target


DaWoblefet:
Thanks Kris!

PR #8781, opened 2022-05-18
Fix multiple Gen 1 Stadium bugs

Karthik99999:
- Shows confusion message after Thrash ends
- Makes Toxic actually work like Toxic
- Makes Leech seed stack with Toxic
- Fixes client messages sent by Haze


DaWoblefet:
Thanks Karthik!

PR #8785, opened 2022-05-20
Kleavor and Ursaluna evo fields

Glazelf:
- Change `evoType` from `"other"` to `"useItem"` for Kleavor and Ursaluna.
- Move item names to `evoItem`.
- Set `evoCondition` for Ursaluna to `"full moon"`.

This is better since Kleavor requires no extra conditions aside from using an item.

Not sure about Ursaluna, but felt like this approach is still more consistent with other Pokémon than what is currently used.

### DIFF
--- a/data/pokedex.ts
+++ b/data/pokedex.ts
@@ -16335,8 +16335,8 @@ export const Pokedex: {[speciesid: string]: SpeciesData} = {
 		weightkg: 89,
 		color: "Brown",
 		prevo: "Scyther",
-		evoType: "other",
-		evoCondition: "Black Augurite",
+		evoType: "useItem",
+		evoItem: "Black Augurite",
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I think tests are failing because we don't actually have those items in our codebase.

Glazelf:
> I think tests are failing because we don't actually have those items in our codebase.

I didn't consider that as a necessity lol. That's my bad I suppose.
What would be better, adding PLA evo items or reverting to the uglier formatting of PLA evos?

DaWoblefet:
I personally think it's fine as-is - our PLA support is intentionally limited, and I don't think it's worthwhile to add a bunch of unusable items to slightly improve evolution condition descriptions.

PR #8789, opened 2022-05-24
Add ???-type Arceus forme

Karthik99999:
https://www.smogon.com/forums/threads/add-type-arceus-to-showdown.3655284/

I named the forme Arceus-Unknown because that's the name it has in the gen 4 sprites on PS

I also made some gen 5 style sprites for this forme since there are only gen 4 sprites on PS, not sure on the best way to send those over (or if you would even accept them).


data/learnsets.ts
```diff
@@ -48358,6 +48358,9 @@ export const Learnsets: {[speciesid: string]: LearnsetData} = {
 	arceuswater: {
 		eventOnly: true,
 	},
+	arceusunknown: {
+		eventOnly: true,
+	},
```
Karthik99999, line 48363: I am also not sure if I should actually make it `eventOnly`

AnnikaCodes:
> I also made some gen 5 style sprites for this forme since there are only gen 4 sprites on PS, not sure on the best way to send those over (or if you would even accept them).

I think @Marty-D can upload those for you.

AnnikaCodes (changes requested):
Can you please push another commit or rebase this so that GitHub reruns the CI with the latest fixes for the TypeScript issues?

PR #8794, opened 2022-05-26
Fix UNO crash and add various UNO QoL fixes

DaWoblefet:
- Fixes UNO crash introduced by https://github.com/smogon/pokemon-showdown/pull/8614; this previous PR accidentally broke a number of things (if you DQ another person whose turn is not active, it would still advance the player count; it would also DQ a player, remove the ID from the array, then attempt to find that ID in the player array which is what caused the crash).
- Adds autodq alias to timer to be consistent with tournaments syntax
- Adds proper error message if you try to start an UNO game with less than 2 users
- Adds logic to let you leave the UNO game during the game in addition to signups


DaWoblefet:
Build passes locally - this is because of the typescript bump.

AnnikaCodes:
These changes look OK to me, but I don't know if any current contributors are interested in maintaining UNO. Kris has a lot of entries in `git blame`, so I've requested review from them, but that may just have been from typescripting.

AnnikaCodes (changes requested):
Can you please push another commit or rebase this so that GitHub reruns the CI with the latest fixes?

DaWoblefet:
Anyone have opinions for how to fix the TS error? I need to both return false and throw an error message for starting with less than 2 players. I can't use errorReply because there's no ``Chat.CommandContext`` available. If there's an easy solution I'll do that, otherwise I don't really feel like refactoring UNO _that_ much :P

DaWoblefet:
Mia gave good advice!

PR #8796, opened 2022-05-29
Check parent roomauth for modjoin

MathyFurret:
Context: https://www.smogon.com/forums/threads/allow-room-owners-to-join-invite-only-subrooms.3702238/

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -558,7 +558,7 @@ export abstract class BasicRoom {
 		if (!Users.Auth.isAuthLevel(modjoinSetting)) {
 			Monitor.error(`Invalid modjoin setting in ${this.roomid}: ${modjoinSetting}`);
 		}
-		return Users.globalAuth.atLeast(user, modjoinSetting);
+		return this.auth.atLeast(user, modjoinSetting);
 	}
 	mute(user: User, setTime?: number) {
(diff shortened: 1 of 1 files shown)


server/rooms.ts
```diff
@@ -558,7 +558,7 @@ export abstract class BasicRoom {
 		if (!Users.Auth.isAuthLevel(modjoinSetting)) {
 			Monitor.error(`Invalid modjoin setting in ${this.roomid}: ${modjoinSetting}`);
 		}
-		return Users.globalAuth.atLeast(user, modjoinSetting);
+		return this.auth.atLeast(user, modjoinSetting);
```
AnnikaCodes, line 561: Wouldn't this cause issues if a subroom were set to modjoin `&`, thus allowing subroom auth to join but not the parent room's owner?

AnnikaCodes: Never mind - it looks like groupchats restrict modjoin to +/% so this fix is OK.

AnnikaCodes:
Thanks!

mia-pi-git:
This seems to have made it so admins can no longer join private rooms. See https://github.com/smogon/pokemon-showdown/commit/4a24d3ee1effa26f300837772576c2f2fc0a254b.

PR #8803, opened 2022-06-03
Add/modify error messages in dexsearch, movesearch

glacialcascade:
Some minor changes:
- Add periods to the end of "The parameter '...' cannot have alternative parameters" error messages for consistency with other error messages
- Add error message for using alternative parameters for `monotype` and `natdex` in dexsearch, and `all` and `natdex` in movesearch (as orGroup.skip is set to true for those, similar to the other cases)
- Add "A search cannot include and exclude 'monotype'." error message as it works similarly to the other parameters with that message (just setting a flag rather than adding another orGroup; currently, `/nds mono, !mono` is the same as `/nds !mono` but `/nds !mono, mono` is the same as `/nds mono`)

It might also be helpful to add an error message if the parameter `natdex` or `all` is negated (as e.g. `!natdex` is functionally identical to `natdex`, which may be surprising behavior).

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -632,12 +632,16 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 			}
 
 			if (['mono', 'monotype'].includes(toID(target))) {
+				if (singleTypeSearch === isNotSearch) return {error: "A search cannot include and exclude 'monotype'."};
+				if (parameters.length > 1) return {error: "The parameter 'monotype' cannot have alternative parameters."};
 				singleTypeSearch = !isNotSearch;
 				orGroup.skip = true;
 				continue;
 			}
 
 			if (target === 'natdex') {
+				if (isNotSearch) return {error: "A search cannot exclude 'natdex'."};
+				if (parameters.length > 1) return {error: "The parameter 'natdex' cannot have alternative parameters."};
 				nationalSearch = true;
 				orGroup.skip = true;
 				continue;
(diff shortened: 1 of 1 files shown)


server/chat-plugins/datasearch.ts
```diff
@@ -679,7 +682,7 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
 			}
 
 			if (target.endsWith(' asc') || target.endsWith(' desc')) {
-				if (parameters.length > 1) return {error: `The parameter '${target.split(' ')[1]}' cannot have alternative parameters`};
+				if (parameters.length > 1) return {error: `The parameter '${target.split(' ')[1]}' cannot have alternative parameters.`};
```
AnnikaCodes, line 685:
```suggestion
				if (parameters.length > 1) {
					return {error: `The parameter '${target.split(' ')[1]}' cannot have alternative parameters.`};
				}
```

server/chat-plugins/datasearch.ts
```diff
@@ -632,12 +632,15 @@ function runDexsearch(target: string, cmd: string, canAll: boolean, message: str
...
 			if (['mono', 'monotype'].includes(toID(target))) {
+				if (singleTypeSearch === isNotSearch) return {error: "A search cannot include and exclude 'monotype'."};
+				if (parameters.length > 1) return {error: "The parameter 'monotype' cannot have alternative parameters."};
 				singleTypeSearch = !isNotSearch;
 				orGroup.skip = true;
 				continue;
 			}
 
 			if (target === 'natdex') {
+				if (parameters.length > 1) return {error: "The parameter 'natdex' cannot have alternative parameters."};
```
AnnikaCodes, line 643: Is this the case? `/ds natdex, water, earthquake, bst>600` and similar searches work fine for me.

glacialcascade: This is meant to prevent stuff like `/ds natdex|blue` where the natdex parameter is part of a group of alternatives

AnnikaCodes: Ah, okay.

AnnikaCodes:
Thanks!

PR #8805, opened 2022-06-06
Fix !pivot, !recovery, !zrecovery, !priority in dexsearch

glacialcascade:
Currently, running any of `/nds !pivot`, `/nds !recovery`, `/nds !zrecovery`, and `/nds !priority` results in "No Pokémon found.". This is because they are adding an empty DexOrGroup
```js
  {
    abilities: {},
    tiers: {},
    doublesTiers: {},
    colors: {},
    'egg groups': {},
    formes: {},
    gens: {},
    moves: {},
    types: {},
    resists: {},
    weak: {},
    stats: {},
    skip: false
  }
```
to `searches`, which naturally no Pokémon can satisfy and therefore eliminates all possibilities. This change sets `skip` to `true` for these empty DexOrGroups so that they will be ignored when finding possibilities.


AnnikaCodes:
Thanks!

PR #8807, opened 2022-06-09
Gen 1: Fix Haze and Disable

gigalh128:
I wrote many more test cases for the functionality of both haze and disable from generation 1. Please take a look at the way I implemented the fix for pokemon not being able to move on the turn they are healed by haze/thawed because I'm not 100% certain that this is the correct way to do this.

### DIFF
--- a/data/mods/gen1/README.md
+++ b/data/mods/gen1/README.md
@@ -41,7 +41,7 @@ Status
 ------
 Freeze never thaws unless hit by a Fire-type attack or by Haze.
 Sleep lasts 1-7 turns and you wake up at the end of the turn.
-If a pokemon has its sleep or freeze status healed by haze, it will not be able to move on that turn.
+If a Pokemon has its sleep or freeze status healed by Haze, it will not be able to move on that turn.
 
 1/256 miss
 ----------
--- a/data/mods/gen1/README.md
+++ b/data/mods/gen1/README.md
@@ -55,4 +55,4 @@ TODO. Glitches not implemented
 There are a couple of divisions by zero in the original game in the cartridge. Those have not been implemented.
 
 If a Pokemon has its sleep or freeze status healed by Haze, it will not be able to move on that turn (skip turn glitch).
-This also applies to a frozen pokemon that has just been thawed by a fire type move.
+This also applies to a frozen Pokemon that has just been thawed by a fire type move.
--- a/data/mods/gen1/README.md
+++ b/data/mods/gen1/README.md
@@ -41,6 +41,7 @@ Status
 ------
 Freeze never thaws unless hit by a Fire-type attack or by Haze.
 Sleep lasts 1-7 turns and you wake up at the end of the turn.
+If a pokemon has its sleep or freeze status healed by haze, it will not be able to move on that turn.
 
 1/256 miss
 ----------
--- a/data/mods/gen1/README.md
+++ b/data/mods/gen1/README.md
@@ -41,7 +41,6 @@ Status
 ------
 Freeze never thaws unless hit by a Fire-type attack or by Haze.
 Sleep lasts 1-7 turns and you wake up at the end of the turn.
-If a Pokemon has its sleep or freeze status healed by Haze, it will not be able to move on that turn.
 
 1/256 miss
 ----------
@@ -54,3 +53,6 @@ Partial trapping moves let either Pokémon switch but target will be unable to m
 TODO. Glitches not implemented
 ------------------------------
 There are a couple of divisions by zero in the original game in the cartridge. Those have not been implemented.
+
+If a Pokemon has its sleep or freeze status healed by Haze, it will not be able to move on that turn (skip turn glitch).
+This also applies to a frozen pokemon that has just been thawed by a fire type move.
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -268,38 +268,39 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		},
 	},
 	disable: {
-		inherit: true,
+		num: 50,
+		accuracy: 55,
+		basePower: 0,
+		category: "Status",
+		name: "Disable",
+		pp: 20,
+		priority: 0,
+		flags: {protect: 1, mirror: 1, bypasssub: 1},
+		volatileStatus: 'disable',
 		onTryHit(target) {
-			const ppSum = target.moveSlots.reduce((acc, val) => acc + val.pp, 0);
-			if (!ppSum) {
-				return false; // all moves had 0 pp
-			}
+			return target.moveSlots.some(ms => ms.pp > 0) && undefined;
 		},
 		condition: {
-			duration: 4,
-			durationCallback(target, source, effect) {
-				// duration is 0 to 7 turns in gen 1
-				const duration = this.random(0, 7);
-				return duration;
-			},
 			onStart(pokemon) {
-				if (!this.queue.willMove(pokemon)) {
-					this.effectState.duration++;
-				}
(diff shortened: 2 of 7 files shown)


data/mods/gen1/moves.ts
```diff
@@ -403,20 +412,41 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+					this.log[this.log.length - 1] += silentHack;
+				}
+				if ('lightscreen' in pokemon.volatiles) {
+					pokemon.removeVolatile('lightscreen');
+					this.log[this.log.length - 1] += silentHack;
+				}
+				if ('reflect' in pokemon.volatiles) {
+					pokemon.removeVolatile('reflect');
+					this.log[this.log.length - 1] += silentHack;
 				}
```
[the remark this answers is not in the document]
gigalh128: Hey there, I didn't follow this exactly because I actually messed up here, and the first three need the silent hack whereas the last 4 are treated differently. Doing this based off of https://gist.github.com/scheibo/ddb13426a10335b7a471b92c5b258b49#file-haze-patch=

I did fix all of the silent hack ones to be done in a loop though.

gigalh128:
Changes in string quotes are based off of https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#strings

I got it wrong before and also fixed the formatting for the rest of the test cases.

data/mods/gen1/README.md
```diff
@@ -41,6 +41,7 @@ Status
 ------
 Freeze never thaws unless hit by a Fire-type attack or by Haze.
 Sleep lasts 1-7 turns and you wake up at the end of the turn.
+If a pokemon has its sleep or freeze status healed by haze, it will not be able to move on that turn.
```
AnnikaCodes, line 44:
```suggestion
If a Pokemon has its sleep or freeze status healed by Haze, it will not be able to move on that turn.
```

sim/battle.ts
```diff
@@ -834,6 +834,14 @@ export class Battle {
 			// 	('0000' + (this.event.modifier * 4096).toString(16)).slice(-4).toUpperCase());
 			relayVar = this.modify(relayVar, this.event.modifier);
 		}
+
+		// handle haze/thawing in gen 1 here since it needs to be part of the BeforeMove event
+		if (eventid === 'BeforeMove' && target instanceof Pokemon &&
+		(target.lastStatusCuredThisTurn === 'slp' || target.lastStatusCuredThisTurn === 'frz') &&
+		this.gen === 1) {
```
AnnikaCodes, line 841:
```suggestion
		if (
			eventid === 'BeforeMove' &&
			target instanceof Pokemon &&
			(target.lastStatusCuredThisTurn === 'slp' || target.lastStatusCuredThisTurn === 'frz') &&
			this.gen === 1
		) {
```

sim/battle.ts
```diff
@@ -834,6 +834,14 @@ export class Battle {
 			// 	('0000' + (this.event.modifier * 4096).toString(16)).slice(-4).toUpperCase());
 			relayVar = this.modify(relayVar, this.event.modifier);
 		}
+
+		// handle haze/thawing in gen 1 here since it needs to be part of the BeforeMove event
```
AnnikaCodes, line 838: This doesn't feel like the right place for this, but I'm not familiar enough with the simulation engine to say where it should go instead, so I guess it's fine.

data/mods/gen1/moves.ts
```diff
@@ -268,30 +268,39 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		},
 	},
 	disable: {
-		inherit: true,
```
DaWoblefet, line 271: You should keep inherit: true and just override functions / other properties as necessary. You have a lot of extra duplicate code!

data/mods/gen1/moves.ts
```diff
@@ -268,30 +268,39 @@ export const Moves: {[k: string]: ModdedMoveData} = {
...
+		accuracy: 55,
+		basePower: 0,
+		category: "Status",
+		name: "Disable",
+		pp: 20,
+		priority: 0,
+		flags: {protect: 1, mirror: 1, bypasssub: 1},
+		volatileStatus: 'disable',
+		onTryHit(target) {
+			return target.moveSlots.some(ms => ms.pp > 0) && undefined;
```
DaWoblefet, line 281: This shouldn't need && undefined.

DaWoblefet: You're missing a check to see if one of the opponent's moves are already Disabled as well (right now it will just silently proceed).

test/sim/moves/disable.js
```diff
@@ -12,36 +12,71 @@ describe('Disable', function () {
...
-			{species: 'Spearow', item: 'assaultvest', moves: ['growl']},
+			{species: "Spearow", item: 'assaultvest', moves: ['growl']},
 		]]);
 
 		battle.makeChoices();
 		assert(battle.log.indexOf('|-fail|p1a: Wynaut') > 0, `Disable should have failed vs Struggle`);
 	});
+
+	it(`should work on the first turn in generation 1 
+		if opponent has move with pp`, function () {
```
DaWoblefet, line 49: Accidental newline

test/sim/moves/haze.js
```diff
@@ -80,4 +80,97 @@ describe('Haze - RBY', function () {
...
+
+		battle.makeChoices('move disable', 'move tackle');
+		assert('disable' in p2volatiles);
+
+		battle.makeChoices('move haze', 'move tackle');
+		assert(!('disable' in p2volatiles));
+	});
+
+	it(`should still make previously disabled pokemon (on the same turn) 
+		with 1 move use struggle`, function () {
```
DaWoblefet, line 116: Ditto extra newline

DaWoblefet (changes requested):
I don't think the new ``lastStatusCuredThisTurn`` property is necessary. I would try investigate if ``onEnd()`` for slp and frz can be leveraged.

gigalh128:
Hey there, thanks for the review. I have a few notes on the commit I just made.

> You should keep inherit: true and just override functions / other properties as necessary. You have a lot of extra duplicate code!

The backwards inheritance from gen 8 was causing a lot of issues for disable, namely with its validity conditions and its duration mechanic, so I decided to stop using inheritance for disable to be safe.

> This shouldn't need && undefined.

This function isn't supposed to return if the two conditions (finding a move with more than 0 pp and not being disabled already) are true, so I added an `&& undefined` to make sure it did not return true. The reasoning for this comes from https://github.com/smogon/pokemon-showdown/blob/master/sim/pokemon.ts#L211. Disable hasn't finished executing at this point, so it is imperative that we DO NOT return true.

> I don't think the new `lastStatusCuredThisTurn` property is necessary. I would try investigate if `onEnd()` for slp and frz can be leveraged.

Unfortunately, `onEnd()` for slp and frz doesn't actually get triggered when the status is cured, so it's useless in this scenario (in contrast to something like confusion being healed by haze, where it does get triggered). Since I figured this would be a point of further contention and would like to have my pull request merged in a reasonable amount of time, I have decided to kick this problem down the road by rolling back its changes. Please let me know if you would prefer I don't do that.

AnnikaCodes (approved):
This seems fine to me, although I'm not too familiar with Gen 1 mechanics.

AnnikaCodes:
> The backwards inheritance from gen 8 was causing a lot of issues for disable, namely with its validity conditions and its duration mechanic, so I decided to stop using inheritance for disable to be safe.

This should probably be addressed in the future, but inheritance bugs aren't really in the scope of this PR.

DaWoblefet:
I appreciate your patience here @gigalh128; thank you much for the fix!

PR #8808, opened 2022-06-09
Fix /exportinputlog

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9240009

- Allows anyone to use the command without consent of all players in random battles.
- Now actually sends the button for the other users to consent to the battle log being shared. Previously `battle.players` was being used to loop through all the players, which is empty after the battle has ended. Now `battle.playerTable` is the object that gets iterated over and `Users.get` is called directly instead of `player.getUser`, as `player.getUser` will return null after the battle has ended.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -649,7 +649,7 @@ export const commands: Chat.ChatCommands = {
 			return;
 		}
 		this.checkCan('exportinputlog', null, room);
-		if (user.can('forcewin')) {
+		if (user.can('forcewin') || Dex.formats.get(battle.format).team) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I'm curious when this broke - it definitely used to work, because I was using it regularly for bug reports before I became an admin. My hunch is the RoomGame API refactor? https://github.com/smogon/pokemon-showdown/commit/2093539e06181b850bc5f4cea5f8339a3ce5292d

DaWoblefet:
Either way, looks good. Thanks Karthik!

PR #8813, opened 2022-06-16
Fix "inverse" argument in /weakness

MathyFurret:
"inverse" is a format so it was getting removed by splitFormat. Now /weak will work even while Inverse is OMotM.


AnnikaCodes:
Thanks!

PR #8814, opened 2022-06-16
Gen 1: Add incompatibilities to NC97 Move Legality

PlagueVonKarma:
Fixing what I reported here: https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/post-9235903
Marty said to just copy the banlist from gen2 and add to it, so here I am!

Details on what's up;
- Before Yellow came out, Nidoran-M learned Double Kick at L42, so Nidoking can never get it alongside Thrash.
- Metapod doesn't learn Harden if evolving from Caterpie prior to Yellow, you have to catch a wild Metapod to have it. Ergo, Tackle and String Shot can't be on the same set.

Extremely minor, but that's most of my pull requests. Thank you for your time!


HoeenCoder:
Re-opened upon request

Marty-D (approved):
Neat, thanks! 👍

PR #8815, opened 2022-06-16
Fix White Herb interactions with Intimidate and KO stat-boosting Abilities

SirenXD:
Fix for the White Herb Intimidate double-battle interaction as described here:
https://github.com/smogon/pokemon-showdown/pull/5185

And for abilities that boost stats on KO, as described here:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-32#post-8664442

For Intimidate:

I found that the originally proposed solution (or at least the idea behind it) appears to work fine. I modified the current test for intimidate to include a single-switch + stat reducing attack turn, which correctly leaves Wynaut at -1 ATK, as well as it passes every other test. I didn't see any comments in the PR that explains why it wont work as a fix.

For Abilities:

I'm new to this code-base so this may not be the best solution. I came across this solution exclusively with Grim Neigh before just adding the other abilities. Assuming they all work the same way, they should work too. I was not able to see how Abilities that activate mid-turn are handled. From what I could tell, it looks like they're inserted into the battle queue mid-turn where action.choice is "move", but I could be wrong about that.

Current solution is to not use the item if there's a Pokemon in the Faint Queue and the Pokemon's ability is within the list. I tried to make sure I got all the abilities that boost on KO, but I may have missed one or some.

grimneigh, beastboost, soulheart, chillingneigh, moxie are all covered in the list. Added tests to cover 4 of these, as I wasn't sure how Beast Boost worked to guarantee I could write a test case to isolate it's changed stat every time. They all passed.

I could see some potential issues with the ability fix that are due to my lack of knowledge of the things I'm using for the fix (specifically in doubles when Wynaut does not get the kill but partner Pokemon does, is the first thing that comes to mind).

If there's any issues with either of these fixes, I can revisit this and take another look. I will likely look into the ability fix in double battles later today or tomorrow.

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -6677,6 +6677,16 @@ export const Items: {[itemid: string]: ItemData} = {
 			},
 		},
 		onUpdate(pokemon) {
+
+
+			let itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), ('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
+
+			for (const action of this.queue) {
+				if (action.choice === 'runSwitch' || (itemUpdateIgnoredAbilities.includes(pokemon.ability) && this.faintQueue.length > 0)) {
+					return;
+				}
+			}
+
 			let activate = false;
 			const boosts: SparseBoostsTable = {};
 			let i: BoostID;
--- a/data/items.ts
+++ b/data/items.ts
@@ -6677,16 +6677,12 @@ export const Items: {[itemid: string]: ItemData} = {
 			},
 		},
 		onUpdate(pokemon) {
-
-
-			let itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), ('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
-
+			const itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), ('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
 			for (const action of this.queue) {
 				if (action.choice === 'runSwitch' || (itemUpdateIgnoredAbilities.includes(pokemon.ability) && this.faintQueue.length > 0)) {
 					return;
 				}
 			}
-
 			let activate = false;
 			const boosts: SparseBoostsTable = {};
 			let i: BoostID;
--- a/data/items.ts
+++ b/data/items.ts
@@ -6677,9 +6677,13 @@ export const Items: {[itemid: string]: ItemData} = {
 			},
 		},
 		onUpdate(pokemon) {
-			const itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), ('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
+			const itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), 
+				('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
 			for (const action of this.queue) {
-				if (action.choice === 'runSwitch' || (itemUpdateIgnoredAbilities.includes(pokemon.ability) && this.faintQueue.length > 0)) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
From the start, I'll just mention that I'm pretty sure the "correct" fix is refactoring White Herb to not use onUpdate(), and instead trigger at specific event timings. I'm confident we don't even have all the requisite events to do that on the sim currently, though.

test/sim/items/whiteherb.js
```diff
@@ -24,27 +24,71 @@ describe("White Herb", function () {
 		assert.statStage(wynaut, 'spa', 0);
 	});
 
-	it.skip('should activate after Abilities that boost stats on KOs', function () {
+	it('should activate after Abilities that boost SPA on KOs', function () {
```
DaWoblefet, line 27: I don't see much reason to include a ton of each Ability in these tests. Soul-Heart and one of the Moxie copies should be good enough.

data/items.ts
```diff
@@ -6677,6 +6677,12 @@ export const Items: {[itemid: string]: ItemData} = {
 			},
 		},
 		onUpdate(pokemon) {
+			const itemUpdateIgnoredAbilities = [('grimneigh' as ID), ('beastboost' as ID), ('soulheart' as ID), ('chillingneigh' as ID), ('moxie' as ID)];
```
DaWoblefet, line 6680: I worry that this will cause other legitimate interactions, e.g. a Beast Boost Pokemon attacking into Gooey and KOing it, from resolving the Speed drop.

SirenXD: It probably will, yeah. I'll have to look into beast boost on it's own to provide a better solution, and I'll have to look out for edge cases like this.

Just to confirm, the resolve order should be Attack into Gooey -> Kill -> Move Stat Drop (if applicable, i.e. Superpower) -> Speed Drop -> Ability Boost -> White Herb?

test/sim/items/whiteherb.js
SirenXD: There isn't if they all function exactly the same (within their respective group). I can prune this to just Moxie and Soul-Heart, and will likely have to make Beast Boost it's own test.

data/items.ts
SirenXD: Tested Beast Boost into Gooey and KO, all stat drops are resolved.

Current Order is:

Kill -> Move Stat Drop -> Gooey Speed Drop -> Ability Boost -> White Herb

If this order is not intended, let me know and I'll see what I can do to have this order swapped around as necessary.

SirenXD:
> (specifically in doubles when Wynaut does not get the kill but partner Pokemon does, is the first thing that comes to mind).

Tested this, although I'm not entirely sure what I was thinking when writing it. The following situations appear to be resolving as intended:

Move causes stat drop but doesn't get kill -> White Herb resolves stat drop(s).
Move causes stat drop and gets kill -> Ability increases stat -> White Herb resolves stat drop(s)
Move causes stat drop and gets kill -> Gooey activates and reduces speed -> Ability increases stat -> White Herb resolves stat drop(s)

Tested in both single and double battle.

Beast Boost appears to be working correctly as well, tried with having an untouched stat get changed (spd) and having one of the move-reduced stats (i.e. spa + draco meteor) get changed, and it worked as expected in both situations.

> I'll just mention that I'm pretty sure the "correct" fix is refactoring White Herb to not use onUpdate(), and instead trigger at specific event timings. I'm confident we don't even have all the requisite events to do that on the sim currently, though.

That's fair. I think the current ability fix is not ideal since the list is hard-coded, but I was not able to discern a method to tell that an ability was going to modify stats.

I can see that the ability's stat boost are defined as a "move", but could not see that there was a move with the abilities ID in the BattleQueue. I would guess this is either because the moves in queue aren't linked to the Ability's ID in the way I was checking, or it has to do with the doUpdate event call frequency and the timing that the "move" is injected into the BattleQueue (if it even is). I could attempt to refactor it to use `onBoost` but I would still need to be able/know how to check the Battle Queue for further incoming stat changes in the same turn to know if I should wait on using the White Herb or not.

PR #8817, opened 2022-06-19
Gen 1 UU: Ban Agility + Partial Trapping

PlagueVonKarma:
https://www.smogon.com/forums/threads/rby-uu-agility-partial-trapping-apt-voting.3703325/page-2#post-9249504

This is going to affect a tournament coming up within the next few days!

I added a rule to rulesets.ts under the assumption this ban may be repeated in other lower tiers. I've seen some groans about it in NU. Technically, only Wrap and Fire Spin are possible alongside Agility; even with Tradebacks, the only new user of this combination is Arcanine with Fire Spin. However, in the spirit of the ban and potential use in something like STABmons room tours, I slapped Bind and Clamp in there as well. Inefficient, but we may as well future-proof.

If the clause name is too long, "APT Clause" is possible, which I think would be understandable at its core. Not sure if I did that right...neither with the position or var name. Regardless, here we are!

Thank you for your time!

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1585,7 +1585,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Prevents Pok\u00e9mon from having moves that would only be obtainable in Pok\u00e9mon Crystal.",
 		// Implemented in mods/gen2/rulesets.ts
 	},
-	aptclause: {
+	agilitypartialtrappingclause: {
 		effectType: 'ValidatorRule',
 		name: 'Agility + Partial Trapping Clause',
 		desc: "Bans the combination of Agility and partial trapping moves like Wrap.",
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1585,6 +1585,12 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Prevents Pok\u00e9mon from having moves that would only be obtainable in Pok\u00e9mon Crystal.",
 		// Implemented in mods/gen2/rulesets.ts
 	},
+	aptclause: {
+		effectType: 'ValidatorRule',
+		name: 'Agility + Partial Trapping Clause',
+		desc: "Bans the combination of Agility and partial trapping moves like Wrap.",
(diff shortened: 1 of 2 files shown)


data/rulesets.ts
```diff
@@ -1585,6 +1585,12 @@ export const Rulesets: {[k: string]: FormatData} = {
 		desc: "Prevents Pok\u00e9mon from having moves that would only be obtainable in Pok\u00e9mon Crystal.",
 		// Implemented in mods/gen2/rulesets.ts
 	},
+	agilitypartialtrappingclause: {
+		effectType: 'ValidatorRule',
+		name: 'Agility + Partial Trapping Clause',
```
KrisXV, line 1590: I don't like the + being in the name here, can it be renamed to AgiliTrap or something?

PlagueVonKarma: That's not really a "used" term. If you want something else, APT would be the appropriate name, which was used in the suspect test's terminology. 

https://www.smogon.com/forums/threads/rby-uu-agility-partial-trapping-apt-suspect-test-apt-banned.3700597/

KrisXV: I don't really care what the name is, as long as there isn't a + in the rule name. It doesn't functionally break anything, but the ambiguity with how unbans work in challenges/how complex bans work in challenges makes me prefer that this was just manually added to the UU ruleset directly. There also doesn't need to be a clause banning these if only one format bans them in the first place, so I'm unsure if this is even a necessity.

PlagueVonKarma: Aye, right. We can go with APT, but lemme just verify this last bit before I change stuff, just to be sure.

> There also doesn't need to be a clause banning these if only one format bans them in the first place, so I'm unsure if this is even a necessity.

I'm mainly going in with the rationale that it's highly likely other RBY tiers will ban the combination, NU especially. If it's not a concern (we _could_ "cross that bridge when we come to it", I try to make things easier later) then yeah we can just do the ban that way. Let me know what you think.

KrisXV: I do not like Chris "Cross That Bridge" Christie so lets just do APT

PR #8822, opened 2022-06-26
Gen 1: Add Stadium Zero Mod

PlagueVonKarma:
This satisfies the approved suggestion by myself [here](https://www.smogon.com/forums/threads/pokemon-stadium-jp-mechanics-mod.3668277/).

This mod is complete with the exceptions of;
- Explosion and Self-Destruct should KO the user when hitting a Substitute
- Stadium's accuracy should have a 1/65535 uncertainty, as it re-rolls the accuracy check if it gets a 255. This is a bug in the standard Stadium sim, and I'm not looking to fix that as I don't know how!
- I believe EBC auto-ties will need looking at again; this mod has the same issues as RBY does, basically.
Otherwise, I'm 99.9% sure it's done. I'm setting it as a draft as I'm not sure whether it's ready or not, and it's probably best it goes through whatever rituals you have.

I also include an "Expansion Disk" rule that "un-dexits" the Pokemon while having the standard ruleset. I think this could be beneficial for those who want to use certain Pokemon in these mechanics; most notably Persian and Rhydon, which are very important contenders. It was intended in the original game, but didn't get released due to the N64 DD's failure...it's a nice touch, right? Or is it best just to have people do "+Unobtainable"?

This Pull Request also adds two formats a la how you did Stadium 2: Stadium Zero OU and Nintendo Cup '98. There is an NC97 version but it's pretty minor so I don't think it's worth looking at. I guess Nintendo Cup '99 could be added down the line for Stadium 1?

### DIFF
--- a/data/mods/gen1stadium0/conditions.ts
+++ b/data/mods/gen1stadium0/conditions.ts
@@ -1,0 +1,99 @@
+// Stadium Zero's statuses are a mix of RBY and Stadium's. It uses Stadium sleep, but RBY burn/tox/par. It's hard to understand.
+export const Conditions: {[k: string]: ModdedConditionData} = {
+	brn: {
+		name: 'brn',
+		effectType: 'Status',
+		onStart(target) {
+			this.add('-status', target, 'brn');
+			target.addVolatile('brnattackdrop'); // TESTME: This should be ignored after a boost.
+		},
+		onAfterMoveSelfPriority: 2,
+		onAfterMoveSelf(pokemon) {
+			const toxicCounter = pokemon.volatiles['residualdmg'] ? pokemon.volatiles['residualdmg'].counter : 1;
+			this.damage(this.clampIntRange(Math.floor(pokemon.maxhp / 16), 1) * toxicCounter, pokemon);
+			if (pokemon.volatiles['residualdmg']) {
+				this.hint("In Gen 1, Toxic's counter is retained after Rest and applies to PSN/BRN.", true);
+			}
+		},
+		onSwitchIn(pokemon) {
+			pokemon.addVolatile('brnattackdrop');
+		},
+		onAfterSwitchInSelf(pokemon) {
+			this.damage(this.clampIntRange(Math.floor(pokemon.maxhp / 16), 1));
+		},
+	},
+	par: {
+		name: 'par',
+		effectType: 'Status',
+		onStart(target) {
+			this.add('-status', target, 'par');
+			target.addVolatile('parspeeddrop');  // TESTME: This should be ignored after a boost.
+		},
+		onBeforeMovePriority: 2,
+		onBeforeMove(pokemon) {
+			if (this.randomChance(63, 256)) {
+				this.add('cant', pokemon, 'par');
+				pokemon.removeVolatile('bide');
+				pokemon.removeVolatile('twoturnmove');
+				pokemon.removeVolatile('fly');
+				pokemon.removeVolatile('dig');
+				pokemon.removeVolatile('solarbeam');
+				pokemon.removeVolatile('skullbash');
+				pokemon.removeVolatile('partialtrappinglock');
+				return false;
+			}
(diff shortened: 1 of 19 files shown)


PlagueVonKarma:
Oh yo, it actually passed without script issues or anything!

If you think this is all gravy, I'm fine with it being merged as-is. It's really those two bugs up top I take issue with and uhhh don't know how to sort out because I am a cavewoman! I guess if you want you could rename the NC97 legality rule thing for here specifically?

data/mods/gen1stadium0/conditions.ts
```diff
@@ -0,0 +1,99 @@
+// Stadium Zero's statuses are a mix of RBY and Stadium's. It uses Stadium sleep, but RBY burn/tox/par. It's hard to understand.
+export const Conditions: {[k: string]: ModdedConditionData} = {
+	brn: {
+		name: 'brn',
+		effectType: 'Status',
+		onStart(target) {
+			this.add('-status', target, 'brn');
+			target.addVolatile('brnattackdrop');// TESTME: This should be ignored after a boost.
```
AnnikaCodes, line 8: Do you want help writing tests for these?

PlagueVonKarma: That could be handy, yeah!

PlagueVonKarma:
yeah idk what's up with that test 

never done these things before and it makes my head spin

data/mods/gen1stadium0/rulesets.ts
```diff
@@ -0,0 +1,94 @@
+export const Rulesets: {[k: string]: ModdedFormatData} = {
+	standard: {
+		effectType: 'ValidatorRule',
+		name: 'Standard',
+		ruleset: ['Obtainable', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Exact HP Mod', 'Cancel Mod'],
+	},
+	expansiondisk: { // Stadium Zero was intended to have an Expansion Disk that legalised all the "dexited" Pokemon.
+		effectType: 'ValidatorRule',
+		name: 'Expansion Disk',
+		desc: 'Legalises Pok\u00e9mon unobtainable in Pok\u00e9mon Stadium Zero.',
```
KrisXV, line 10:
```suggestion
		desc: 'Legalizes Pok\u00e9mon unobtainable in Pok\u00e9mon Stadium Zero.',
```

sim/team-validator.ts
```diff
@@ -1674,7 +1674,7 @@ export class TeamValidator {
 			problems.push(`This format is in gen ${dex.gen} and ${name} is from gen ${eventData.generation}${etc}.`);
 		}
 
-		if (eventData.japan && dex.currentMod !== 'gen1jpn') {
+		if (eventData.japan && dex.currentMod !== 'gen1jpn' || eventData.japan && dex.currentMod !== 'gen1stadium0') {
```
KrisXV, line 1677:
```suggestion
		if (eventData.japan && !['gen1jpn', 'gen1stadium0'].includes(dex.currentMod)) {
```

test/sim/moves/agility-swords-dance.js
```diff
@@ -0,0 +1,51 @@
...
+const common = require('./../../common');
+
+let battle;
+
+describe('Agility + Swords Dance', function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it('[Gen 1] Agility should negate the paralysis speed drop.', function () {
```
AnnikaCodes, line 13: It looks like this test is failing - can you investigate?

PlagueVonKarma: I think it's cus of the format choice? Not sure. I asked Kris for help with fixing it, because I really, really, really don't know how this stuff works. I know I've said this a lot of times but I'm really not a programmer...

PR #8826, opened 2022-07-04
Update README.md

BUCKFAE:
Using `n latest` did not work for me. I had to explicitly use Node 14.


monsanto:
I'd rather do the current LTS if we deemed this necessary. Node 14 (and this PR) is pretty old, so I don't we'll be merging this.

PR #8832, opened 2022-07-11
Add Jump! Magikarp format

DaWoblefet:
https://www.smogon.com/forums/threads/jump-magikarp.3704588/

This isn't complete, because we don't really have a way of enforcing "you must bring Magikarp" - or any "you must bring [Pokemon]" logic at all. I think we should just use standard PS timer and ignore the complications of implementing the VGC timeout logic at Team Preview with this, but that still doesn't solve the issue with forcing a player to bring Magikarp. I'm not even sure where to put the hardcode if we want to go that route.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -460,6 +460,7 @@ export const Formats: FormatList = [
 		],
 
 		mod: 'gen8',
+		gameType: 'doubles',
 		ruleset: ['Flat Rules', '!! Picked Team Size = 2', '!! Adjust Level = 50', 'Min Source Gen = 8'],
 		banlist: ['Sub-Legendary'],
 		onValidateTeam(team) {
@@ -475,6 +476,7 @@ export const Formats: FormatList = [
 				return [`Your team must contain Magikarp.`];
 			}
 		},
+		// Bringing 1 Magikarp during Team Preview hardcoded in sim/side.ts#chooseTeam
 	},
 	{
 		name: "[Gen 8] Doubles Custom Game",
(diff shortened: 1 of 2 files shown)


config/formats.ts
```diff
@@ -461,6 +452,29 @@ export const Formats: FormatList = [
 			}
 		},
 	},
+	{
+		name: "[Gen 8] Jump! Magikarp",
+		threads: [
```
KrisXV, line 457:
```suggestion
		desc: `Every team must contain Magikarp.`,
		threads: [
```

KrisXV:
I think this validation is fine for now, we don't need a force x logic to be made ASAP for easy future proofing since it doesn't do anything except delay this. We can look into more proper routes in the future.

DaWoblefet:
> I think this validation is fine for now, we don't need a force x logic to be made ASAP for easy future proofing since it doesn't do anything except delay this. We can look into more proper routes in the future.

There's no enforcement of actually bringing Magikarp to the game right now as it is. It only enforces a Magikarp existing on the team.

KrisXV:
> > I think this validation is fine for now, we don't need a force x logic to be made ASAP for easy future proofing since it doesn't do anything except delay this. We can look into more proper routes in the future.
> 
> There's no enforcement of actually bringing Magikarp to the game (selecting it during Team Preview) right now as it is. It only enforces a Magikarp existing on the team.

Oh, I misinterpreted what you were saying. mb

DaWoblefet:
I talked with HoeenHero and the correct place to add the logic would be in sim/side.ts#chooseTeam. I'll work on that later this evening.

DaWoblefet:
Not the prettiest on the client, but it functions!
image

Marty-D (approved):

PR #8834, opened 2022-07-13
Gen 1: Add Nintendo Cup 1999

PlagueVonKarma:
Annika said I could do this so uh here I am! NC99 is basically the first UU tier to ever exist and has some incredibly interesting history, it's pretty neat. 

This is being done with the presumption that #8822 will go through with Nintendo Cup 1998, essentially completing the "quadrilogy" of the original BSS formats. Given the Stadium mod is way, way, way more accurate than it used to be, with some ancient (and some egregious) bugs being fixed, it makes sense from this perspective as well.

It's a draft because I have zero clue why my validator code isn't working and am praying that someone knows why. Also, it'll create a merge conflict with the NC98 mod because I cannot future-proof to save my life :3

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -3877,7 +3877,7 @@ export const Formats: FormatList = [
         mod: 'gen1stadium', 
 		searchShow: false,
         ruleset: ['Adjust Level Down = 50', 'Picked Team Size = 3', 
-				'Obtainable', 'Team Preview', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'Exact HP Mod', 'Cancel Mod', 'Nintendo Cup 1999 Move Legality'],
+				'Obtainable', 'Team Preview', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'Exact HP Mod', 'Cancel Mod'],
         banlist: ['Venusaur', 'Dugtrio', 'Alakazam', 'Golem', 'Magneton', 'Gengar', 'Hypno', 'Electrode', 'Exeggutor', 'Chansey', 'Kangaskhan', 'Starmie', 'Jynx', 'Tauros', 'Gyarados', 'Lapras', 'Ditto', 'Vaporeon', 'Jolteon', 'Snorlax', 'Articuno', 'Zapdos', 'Dragonite', 'Mewtwo', 'Mew', 'Flareon + Focus Energy + Ember', 'Nidoking + Fury Attack + Thrash'],
     },
 	{
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -3877,7 +3877,7 @@ export const Formats: FormatList = [
         mod: 'gen1stadium', 
 		searchShow: false,
         ruleset: ['Adjust Level Down = 50', 'Picked Team Size = 3', 
-				'Obtainable', 'Team Preview', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'Exact HP Mod', 'Cancel Mod'],
+				'Obtainable', 'Team Preview', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'Exact HP Mod', 'Cancel Mod', 'Nintendo Cup 1999 Move Legality'],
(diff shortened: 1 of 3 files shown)


KrisXV:
I'm going to try to not sound pessimistic but apologies in advance if I still come off as such. What's the need for all of these Gen 1 formats? We normally don't add formats unless it's convenience for a team tour or circuit (old gen 1v1s for 1v1PL/circuit for example) or has big community outcry (BW ZU). I haven't really seen much of that for this or Stadium Zero, though the latter makes it possible to simulate Japanese stadium mechanics, which is a different story. I just don't really see a point to have this much inflation in gen 1 formats past the bare minimum to make simulation possible.

KrisXV:
Feel free to PM me on discord and I'll talk more since it's easier to discuss there, but I wanted to share my opinion on this before reviewing so we don't waste time on something until we're sure it'll be added.

AnnikaCodes:
Here's my 2 cents:

Personally I don't see what the issue is with adding more challenge-only formats as long as there is some interest from players. The challenge format selection menu is already so large that it doesn't fit on my screen without scrolling, so I don't think the clutter argument is very important (which is the main reason we limit ladder formats, AFAIK). It also seems arbitrary to support Nintendo Cup 1997 and not the other Nintendo Cups; although they are all fairly obscure, so are decade-old VGC formats and OMs with only one Smogon post in their thread in the past year (we make both of these formats available for challenges).

config/formats.ts
```diff
@@ -3866,6 +3866,22 @@ export const Formats: FormatList = [
...
+			`&bullet; <a href="https://www.smogon.com/forums/threads/stadium-cup-format-megathread.3695448/post-9066167">Stadium Cup Format Megathread</a>`,
+		],
+		
+		mod: 'gen1stadium', 
+		searchShow: false,
+		ruleset: ['Adjust Level Down = 50', 'Picked Team Size = 3', 
+				'Obtainable', 'Team Preview', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'Exact HP Mod', 'Cancel Mod'],
+		banlist: ['Venusaur', 'Dugtrio', 'Alakazam', 'Golem', 'Magneton', 'Gengar', 'Hypno', 'Electrode', 'Exeggutor', 'Chansey', 'Kangaskhan', 'Starmie', 
+		'Jynx', 'Tauros', 'Gyarados', 'Lapras', 'Ditto', 'Vaporeon', 'Jolteon', 'Snorlax', 'Articuno', 'Zapdos', 'Dragonite', 'Mewtwo', 'Mew', 
+		'Flareon + Focus Energy + Ember', 'Nidoking + Fury Attack + Thrash'],
```
AnnikaCodes, line 3883: Running `npm run fix` should fix the formatting here.

PlagueVonKarma: oh i got an error when i tried that so i just kinda left it--

KrisXV:
I mentioned this in PvK's PMs on discord, but my ideal method for adding Nintendo Cup 1999 is via adding a ValidatorRule called "Allow Japanese Events" similar to that of "Allow Tradebacks" so players can just add the rule to Gen 1 Stadium OU in challenges/tournaments and manually ban/unban the other Pokemon, or even going a step further and adding a "Nintendo Cup 1999 Legality" rule to make it easy to add the banlist. I don't like the argument of "the formats list is already cluttered so what hurt will another format do," especially when I've been aiming to minimize the list to what's necessary/popular.

PlagueVonKarma:
> I mentioned this in PvK's PMs on discord, but my ideal method for adding Nintendo Cup 1999 is via adding a ValidatorRule called "Allow Japanese Events" similar to that of "Allow Tradebacks" so players can just add the rule to Gen 1 Stadium OU in challenges/tournaments and manually ban/unban the other Pokemon, or even going a step further and adding a "Nintendo Cup 1999 Legality" rule to make it easy to add the banlist. I don't like the argument of "the formats list is already cluttered so what hurt will another format do," especially when I've been aiming to minimize the list to what's necessary/popular.

Fwiw this method would also benefit GSC as it also has a Japan-exclusive event or two, and is compatible with the Gen 1 events as well. Extremely minor overall, but it hurts no one.

PR #8835, opened 2022-07-14
Fix Jumpluff EVs

sexy90gxebattlefactoryplayer:
Jumpluff has 252 spdef EVs that should be speed. https://www.smogon.com/dex/sm/pokemon/jumpluff/

- Wigglytuff


DaWoblefet:
Good find, thanks!

PR #8836, opened 2022-07-15
Gen 1: Properly fix Mist

KrisXV:
This properly addresses the double failure messages that #8818 attempted to fix and adjusts protocol back to what it should be. The double failure was popping up because `gen1/scripts.ts` wasn't respecting `SILENT_FAIL`.

### DIFF
--- a/data/mods/gen1/scripts.ts
+++ b/data/mods/gen1/scripts.ts
@@ -467,9 +467,9 @@ export const Scripts: ModdedBattleScriptsData = {
 					return false;
 				}
 				if (moveData.boosts && target.hp) {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Thanks Kris!

TheVanadium:
Is there supposed to be a comma in the "But, it failed" message?

DaWoblefet:
> Is there supposed to be a comma in the "But, it failed" message?

Yes. That's how it's written for Gen 1 failure messages: https://github.com/pret/pokeyellow/blob/501a21fa5dc6f01c0563316c9cd209cccb1eef45/data/text/text_5.asm

PR #8840, opened 2022-07-18
Add a GUI for using custom rules

mia-pi-git:
This allows people to select a format and toggle clauses/rules/bans/value rules easily, rather than just having to _know_ the syntax offhand. I couldn't simplify bans any further, so I put an explanation of the syntax on there. For the rest, I added checkboxes for clauses and checkboxes/input tags for value rules.
This was suggested [here](https://www.smogon.com/forums/threads/ability-to-create-custom-formats-via-a-format-builder-tool.3704660/).

Some screenshots:
Base page where one selects a format: https://user-images.githubusercontent.com/49593536/179430289-20b58323-99e3-4aee-b74c-e0efaff7cab8.png
Main page: https://user-images.githubusercontent.com/49593536/179430297-72228ac7-4fea-4478-9afd-d662abed4582.png
A snippet of the clause editor: https://user-images.githubusercontent.com/49593536/179430312-d995d9d7-5712-4b3d-a72c-d50b4efeb1ac.png
Value rule editor: https://user-images.githubusercontent.com/49593536/179430325-2389a3ce-25a7-43e1-a9cb-4357ea3aa832.png

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2954,13 +2954,13 @@ export const pages: Chat.PageTable = {
 		buf += `Bans/Unbans: <input name="bans" /> <small>(separated by commas)</small><br /><br />`;
 		buf += `<details class="readmore"><summary><u><strong>Clauses</strong></u></summary>`;
 		buf += `<p>The following rules can be added to challenges/tournaments to modify the style of play. `;
-		buf += `Alternatively, already present rules can be removed from formats by preceding the rule name with <code>!</code></p>`;
+		buf += `Alternatively, already present rules can be removed from formats by preceding the rule name with <code>!</code>.</p>`;
 		buf += `<p>However, some rules, like <code>Obtainable</code>, are made of subrules, that can be individually turned on and off.</p>`;
 		buf += `<p>Note that if you do not explicitly check a mod to include it, it will not be included in the final output.</p>`;
 		buf += `<div class="ladder"><table><tr><th>Rule Name</th><th>Description</th><th>Toggle</th></tr>`;
 		for (const rule of rules) {
 			if (rule.hasValue) continue;
-			const desc = rule.desc ? rule.desc : "No description.";
+			const desc = rule.desc || "No description.";
 			const id = toID(rule.name);
 			buf += `<tr>`;
 			buf += `<td>${rule.name}</td><td>${desc}</td>`;
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2974,7 +2974,7 @@ export const pages: Chat.PageTable = {
 		buf += `<div class="ladder"><table><tr><th>Rule Name</th><th>Description</th><th>Toggle</th></tr>`;
 		for (const rule of rules) {
 			if (!rule.hasValue) continue;
-			const desc = rule.desc ? rule.desc : "No description.";
+			const desc = rule.desc || "No description.";
 			buf += `<tr>`;
 			const id = toID(rule.name);
 			const existing = ruleTable.valueRules.get(id);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2746,6 +2746,109 @@ export const commands: Chat.ChatCommands = {
 		`!code [code] - Broadcasts code to a room. Accepts multi-line arguments. Requires: + % @ & #`,
 		`/code [code] - Shows you code. Accepts multi-line arguments.`,
 	],
+
+	buildformat(target, room, user) {
+		target = toID(target);
+		if (target && this.connection.openPages?.has('buildformat')) {
+			this.closePage('buildformat');
+		}
+		return this.parse(`/j view-buildformat${target ? `-${target}` : ""}`);
+	},
+
+	makecustomchallenge(target, room, user) {
+		target = target.trim();
+		if (!target.length) {
+			return this.popupReply(`No parameters given.`);
+		}
+		const args = Chat.parseArguments(target, ' | ', {
+			allowEmpty: true, useIDs: false,
+		});
+		const format = Dex.formats.get(toID(args.format[0]));
+		if (!format.exists) {
+			return this.popupReply(`The format '${format}' does not exist.`);
+		}
+		delete args.format;
+		const targetUserID = toID(args.user[0]);
+		if (targetUserID) {
+			this.checkChat();
+			if (!Users.get(targetUserID)) {
+				return this.popupReply(`User '${targetUserID}' not found.`);
+			}
+		}
+		delete args.user;
+		const challengeBuf = [];
+		if (args.bans?.[0]) {
+			const bans = args.bans[0].split(',').map(f => f.trim());
+			challengeBuf.push(...bans);
+		}
+		delete args.bans;
+
+		for (const k in args) {
+			if (k.endsWith('-enabled')) continue;
+			let name = k;
+			if (name.endsWith('-val')) {
+				name = name.slice(0, -4);
+			}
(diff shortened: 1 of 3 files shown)


server/chat-commands/info.ts
```diff
@@ -2746,6 +2746,110 @@ export const commands: Chat.ChatCommands = {
...
+				name = name.slice(0, -4);
+			}
+			const rule = Dex.data.Rulesets[name];
+			if (!rule || rule.effectType === 'Format') {
+				return this.popupReply(`Invalid rule or modifier: ${name}`);
+			}
+			const id = rule.id || toID(rule.name);
+			let val = args[k][0];
+			// BUG: when there are numbers in the form elem/param name,
+			// the `{name}` specifier doesn't get removed from the string
```
AnnikaCodes, line 2800: should this be fixed before the PR goes live?

server/chat-commands/info.ts
```diff
@@ -2746,6 +2746,110 @@ export const commands: Chat.ChatCommands = {
...
+		}
+		if (targetUserID) {
+			this.closePage(`buildformat`);
+			return this.parse(`/challenge ${targetUserID},${fullFormat}`);
+		} else {
+			this.connection.send(
+				`>view-buildformat-${format.id}\n|selectorhtml|#output|` +
+				`Here's the string for your desired rules!<br /><code>${fullFormat}</code><br />` +
+				`Use <code>/challenge [user],${fullFormat}</code> to challenge someone with it!`
+			);
```
AnnikaCodes, line 2850: should it also prompt the user visually to enter in a username if they didn't specify one earlier?

mia-pi-git: Nah, it's designed for it to be optional. If a user isn't given, then it just spits out the string so that they can use it on their own (plus it gives directions on how to do so).

server/chat-commands/info.ts
```diff
@@ -2746,6 +2746,110 @@ export const commands: Chat.ChatCommands = {
...
+		}
+		delete args.user;
+		const challengeBuf = [];
+		if (args.bans?.[0]) {
+			const bans = args.bans[0].split(',').map(f => f.trim());
+			challengeBuf.push(...bans);
+		}
+		delete args.bans;
+
+		((global as any).args ||= {})[user.id] = args;
```
AnnikaCodes, line 2786: Why use a global variable here?

mia-pi-git: Whoops. That was for debugging.

server/chat-commands/info.ts
mia-pi-git: Handled in dms.

server/chat-commands/info.ts
```diff
@@ -2804,6 +2907,91 @@ export const pages: Chat.PageTable = {
...
+		buf += `<li><code>- Baton Pass</code>: Ban a move/item/ability/etc</li></ul>`;
+		buf += `<h3>Group bans</h3>`;
+		buf += `<ul><li><code>- OU</code> or <code>- DUU</code>: Ban a tier</li>`;
+		buf += `<li><code>- Mega</code> or <code>- CAP</code>: Ban a Pok&eacute;mon category</li></ul>`;
+		buf += `<h3>Complex bans</h3>`;
+		buf += `<ul><li><code>- Blaziken + Speed Boost</code>: Ban a combination of things in a single Pokemon (you can have a Blaziken, and you can have Speed Boost on the same team, but the Blaziken can't have Speed Boost)</li>`;
+		buf += `<li><code>- Drizzle ++ Swift Swim</code>: Ban a combination of things in a team (if any Pokémon on your team have Drizzle, no Pokémon can have Swift Swim)</li></ul>`;
+		buf += `<h2><u>Unbans</u></h2>`;
+		buf += `<p>Using a <code>+</code> instead of a <code>-</code> unbans that category.</p>`;
+		buf += `<ul><li><code>+ Blaziken</code>: Unban/unrestrict a Pok&eacute;mon.</li></ul></small></details><br />`;
```
AnnikaCodes, line 2952: As a sidenote, don't we have the ability to use TSX in plugins now? This would be a perfect use for it; TSX would make this much easier to read since we could indent the HTML tags.

server/chat-commands/info.ts
```diff
@@ -2804,6 +2907,91 @@ export const pages: Chat.PageTable = {
...
+		cmd.push(`bans={bans}`);
+		buf += `Bans/Unbans: <input name="bans" /> <small>(separated by commas)</small><br /><br />`;
+		buf += `<details class="readmore"><summary><u><strong>Clauses</strong></u></summary>`;
+		buf += `<p>The following rules can be added to challenges/tournaments to modify the style of play. `;
+		buf += `Alternatively, already present rules can be removed from formats by preceding the rule name with <code>!</code></p>`;
```
AnnikaCodes, line 2957:
```suggestion
		buf += `Alternatively, already present rules can be removed from formats by preceding the rule name with <code>!</code>.</p>`;
```

server/chat-commands/info.ts
```diff
@@ -2804,6 +2907,91 @@ export const pages: Chat.PageTable = {
...
+		buf += `<p>However, some rules, like <code>Obtainable</code>, are made of subrules, that can be individually turned on and off.</p>`;
+		buf += `<p>Note that if you do not explicitly check a mod to include it, it will not be included in the final output.</p>`;
+		buf += `<div class="ladder"><table><tr><th>Rule Name</th><th>Description</th><th>Toggle</th></tr>`;
+		for (const rule of rules) {
+			if (rule.hasValue) continue;
+			const desc = rule.desc ? rule.desc : "No description.";
```
AnnikaCodes, line 2963:
```suggestion
			const desc = rule.desc || "No description.";
```

server/chat-commands/info.ts
```diff
@@ -2804,6 +2907,91 @@ export const pages: Chat.PageTable = {
...
+			buf += `</tr>`;
+			cmd.push(`${id}={${id}}`);
+		}
+		buf += `</table></div></details><br />`;
+		buf += `<details class="readmore"><summary><u><strong>Value Rules</strong></u></summary>`;
+		buf += `Click the "enabled" checkbox and fill in a value to enable a rule - uncheck the box to remove the rule.`;
+		buf += `<div class="ladder"><table><tr><th>Rule Name</th><th>Description</th><th>Toggle</th></tr>`;
+		for (const rule of rules) {
+			if (!rule.hasValue) continue;
+			const desc = rule.desc ? rule.desc : "No description.";
```
AnnikaCodes, line 2977:
```suggestion
			const desc = rule.desc || "No description.";
```

server/chat-commands/info.ts
mia-pi-git: I can do this in another PR.

AnnikaCodes:
Thanks!

server/chat-commands/info.ts
```diff
@@ -2746,6 +2746,109 @@ export const commands: Chat.ChatCommands = {
...
+			if (k.endsWith('-enabled')) continue;
+			let name = k;
+			if (name.endsWith('-val')) {
+				name = name.slice(0, -4);
+			}
+			const rule = Dex.data.Rulesets[name];
+			if (!rule || rule.effectType === 'Format') {
+				return this.popupReply(`Invalid rule or modifier: ${name}`);
+			}
+			const id = rule.id || toID(rule.name);
```
urkerab, line 2796: What was the intent of this line? `Dex.data.Rulesets` can only be indexed by an ID, and the raw data itself never contains the `id` property (this was overlooked because the type declaration allows it).

PR #8841, opened 2022-07-18
Gen 1: Move Slowbro to OU

PlagueVonKarma:
https://www.smogon.com/forums/threads/slowbro-and-the-ou-uu-cutoff.3704191/post-9278948
The "situation" was resolved and Slowbro is being maintained as OU. What a bother...


DaWoblefet:
Thanks Plague Von Karma!

PR #8843, opened 2022-07-21
Replace BDSP PH with Sw/Sh PH

TomOfTomKat:
After a recent definition change, BDSP PH is being replaced with Sw/Sh PH. This commit is to remove BDSP PH from the challengeable list and replace it with Sw/Sh PH. Announcement can be found here: https://www.smogon.com/forums/threads/pure-hackmons.3656851/post-9279060


DaWoblefet:
Thanks!

PR #8849, opened 2022-07-25
Random Battle updates

ACakeWearingAHat:
- Basculin has Superpower over Head Smash
- Silvally-Flying has Flame Charge over Fire Fang
- Druddigon gains Gunk Shot and cannot obtain Sheer Force by having Gunk Shot in singles
- Rotom formes will be less common
- Palkia will get Lustrous Orb instead of Leftovers when it does not qualify for a Choice item


AnnikaCodes:
Thanks!

PR #8850, opened 2022-07-27
Gen 7 Battle Factory: Remove Quagsire from UU

sexy90gxebattlefactoryplayer:
Quagsire remains banned after an initial ban and subsequent retest (https://www.smogon.com/forums/threads/sm-uu-quagsire-revote.3702984/page-2). It should stay banned for the foreseeable future so I'm removing it from the UU tier of BF.

\- Wigglytuff


sexy90gxebattlefactoryplayer:
I understand there's a script somewhere to check if the Pokemon's tier is higher than the tier being generated ingame, but because of Quag's complicated history in SM UU, I think it'd be best for clarity's sake to remove it here.

AnnikaCodes:
Thanks!

PR #8851, opened 2022-07-27
Rooms: Ensure battles save all data correctly

mia-pi-git:
This ensures battle rating is saved, properly persists timer, and ensures people join correctly. It also adds a notification telling people their battles will be restored shortly (since battle reloading can take a bit), which is removed when they have a battle restored.

### DIFF
--- a/server/rooms.ts
+++ b/server/rooms.ts
@@ -1279,6 +1279,28 @@ export class GlobalRoomState {
 		let count = 0;
 		if (!Config.usepostgres) return 0;
 		const logDatabase = new PostgresDatabase();
+		let [{value}] = await logDatabase.query(`SELECT value FROM battle_db_info WHERE key = 'version'`).catch(
+			() => [{value: "0"}]
+		);
+		value = Number(value);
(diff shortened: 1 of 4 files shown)


server/rooms.ts
```diff
@@ -1279,6 +1279,28 @@ export class GlobalRoomState {
 		let count = 0;
 		if (!Config.usepostgres) return 0;
 		const logDatabase = new PostgresDatabase();
+		let [{value}] = await logDatabase.query(`SELECT value FROM battle_db_info WHERE key = 'version'`).catch(
+			() => [{value: "0"}]
+		);
+		value = Number(value);
+		const files = FS(`databases/migrations/storedbattles`)
+			.readdirSync()
+			.map(f => Number(f.slice(1).split('.')[0]));
```
AnnikaCodes, line 1288:
```suggestion
			.filter(f => f.endsWith('.sql'))
			.map(f => Number(f.slice(1).split('.')[0]));
```
Stray files like `.DS_Store` or an editor swap file could cause problems.

PR #8854, opened 2022-08-01
Category Swap: Fix special cases and rulify

WeWuzNidokangz:
The special casing is needed to make certain moves follow the OM rule correctly. Rulification is just to support mashups better and to future-proof mechanics support in case the format is removed after LCotM.

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1996,78 +1996,78 @@ export const Rulesets: {[k: string]: FormatData} = {
 			}
 
 			switch (move.id) {
-				case 'doomdesire': {
-					move.onTry = function (source, target) {
-						if (!target.side.addSlotCondition(target, 'futuremove')) return false;
-						Object.assign(target.side.slotConditions[target.position]['futuremove'], {
-							move: 'doomdesire',
-							source: source,
-							moveData: {
-								id: 'doomdesire',
-								name: "Doom Desire",
-								accuracy: 100,
-								basePower: 140,
-								category: "Physical",
-								priority: 0,
-								flags: {},
-								effectType: 'Move',
-								isFutureMove: true,
-								type: 'Steel',
-							},
-						});
-						this.add('-start', source, 'Doom Desire');
-						return this.NOT_FAIL;
-					};
+			case 'doomdesire': {
+				move.onTry = function (source, subtarget) {
+					if (!subtarget.side.addSlotCondition(subtarget, 'futuremove')) return false;
+					Object.assign(subtarget.side.slotConditions[subtarget.position]['futuremove'], {
+						move: 'doomdesire',
+						source: source,
+						moveData: {
+							id: 'doomdesire',
+							name: "Doom Desire",
+							accuracy: 100,
+							basePower: 140,
+							category: "Physical",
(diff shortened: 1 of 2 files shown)


KrisXV:
These special cases for photon geyser/clones are not mentioned anywhere in the op of the thread and they were not mentioned while the meta was in the submissions stage. I also didn’t make this a rule on purpose.

KrisXV:
> These special cases for photon geyser/clones are not mentioned anywhere in the op of the thread and they were not mentioned while the meta was in the submissions stage. I also didn’t make this a rule on purpose.

Btw with this I’m saying make sure to clarify how they work in the op before I merge this

data/rulesets.ts
```diff
@@ -1979,4 +1979,96 @@ export const Rulesets: {[k: string]: FormatData} = {
 			}
 		},
 	},
+	categoryswaprule: {
```
KrisXV, line 1982:
```suggestion
	categoryswapmod: {
```

data/rulesets.ts
```diff
@@ -1979,4 +1979,96 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+		effectType: 'Rule',
+		name: 'Category Swap Rule',
```
KrisXV, line 1984:
```suggestion
		name: 'Category Swap Mod',
```

data/rulesets.ts
```diff
@@ -1979,4 +1979,96 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+		desc: `All physical moves become special, and all special moves become physical.`,
+		onBegin() {
+			this.add('rule', 'Category Swap Rule: All physical moves become special, and all special moves become physical');
```
KrisXV, line 1987:
```suggestion
			this.add('rule', 'Category Swap Mod: All physical moves become special, and vice versa');
```

WeWuzNidokangz:
> These special cases for photon geyser/clones are not mentioned anywhere in the op of the thread and they were not mentioned while the meta was in the submissions stage. I also didn’t make this a rule on purpose

Okay, I'll ask for that to be checked. It was reported as a bug to me before without special-casing, but I don't remember if it was somebody on council who reported/confirmed it.

https://www.smogon.com/forums/threads/category-swap.3702709/page-3#post-9293421

Is there any benefit to _not_ making this a rule? It seems likely that it will become necessary to maintain the functionality at some point, while in the meantime having the mechanics entangled with the format just makes things like Category Swap AG, Ubers, etc unnecessarily convoluted to express and maintain.

data/rulesets.ts
```diff
@@ -1979,4 +1979,96 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+							effectType: 'Move',
+							isFutureMove: true,
+							type: 'Steel',
+						},
+					});
+					this.add('-start', source, 'Doom Desire');
+					return this.NOT_FAIL;
+				};
+			}
+			break;
```
KrisXV, line 2022: This needs to be in the brackets

KrisXV:
I discuss making these rules with the om mod team at the end of each month, so we haven’t had a chance to discuss it yet

PR #8857, opened 2022-08-06
Fix /analysis for Zygarde-10%

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9297815


AnnikaCodes:
Good catch!

PR #8862, opened 2022-08-13
`/roomauth`: Fix duplicate "names in bold" message

MathyFurret:
This fixes a bug where subrooms would cause the "Names in **bold** are online" message to appear twice (or potentially more for nested subrooms). Additionally, I moved the message to a separate line for clarity.


AnnikaCodes:
Thanks!

PR #8863, opened 2022-08-13
Fix adding players in 2-player battles

MathyFurret:
The message "You cannot join this battle, as you were not originally playing in it" should display if and only if the battle was restored from a reset and hasn't already resumed. Adding players for reasons such as /leavegame substitution and /importinputlog will work again.


AnnikaCodes:
Thanks!

PR #8867, opened 2022-08-15
Add Partners in Crime

HisuianZoroark:


KrisXV:
I can review this after work

sim/pokemon.ts
```diff
@@ -19,6 +19,7 @@ interface MoveSlot {
 	disabledSource?: string;
 	used: boolean;
 	virtual?: boolean;
+	originalPoke?: number; // Partners in Crime
```
KrisXV, line 22: im unsure how i feel about adding stuff used just for oms onto these values. @Marty-D thoughts?

data/mods/partnersincrime/scripts.ts
```diff
@@ -0,0 +1,341 @@
...
+			}
+			// Entrainment
+			if (this.m.innate && this.m.innate.endsWith(ability.id)) {
+				this.removeVolatile(this.m.innate);
+				delete this.m.innate;
+			}
+			this.abilityOrder = this.battle.abilityOrder++;
+			return oldAbility;
+		},
+		hasAbility(ability: string | string[]) {
```
KrisXV, line 221:
```suggestion
		hasAbility(ability) {
```

data/mods/partnersincrime/scripts.ts
```diff
@@ -0,0 +1,341 @@
...
+			if (!Array.isArray(ability)) {
+				if (ownAbility === this.battle.toID(ability) || allyAbility === this.battle.toID(ability)) return true;
+			} else {
+				 if (ability.map(this.battle.toID).includes(ownAbility) || ability.map(this.battle.toID).includes(allyAbility)) {
+					 return true;
+				 }
+			}
+			return false;
+		},
+		transformInto(pokemon: Pokemon, effect?: Effect | null) {
```
KrisXV, line 235:
```suggestion
		transformInto(pokemon, effect) {
```

data/mods/partnersincrime/scripts.ts
```diff
@@ -0,0 +1,341 @@
...
+
+		// Partners in Crime moveSlot updating
+		// Must be highest priority so imprison doesn't lag behind.
+		for (const side of this.sides) {
+			for (const pokemon of side.active) {
+				pokemon.moveSlots = pokemon.moveSlots.filter(move => move.originalPoke === pokemon.m.value);
+				const ally = side.active.find(mon => mon && mon !== pokemon && !mon.fainted);
+				let allyMoves = ally ? this.dex.deepClone(ally.moveSlots) : [];
+				if (ally) {
+					// @ts-ignore
```
KrisXV, line 16: i don't like using ts-ignore here, it's better to typecast as `any` typically. What was the error originally?

HisuianZoroark: this can honestly be merged with the above line but I want to find a solution to transform that doesn't bleed into pokemon.ts first before i handle this.

KrisXV: you can keep it, just change the typecast

config/formats.ts
```diff
@@ -2093,6 +2093,118 @@ export const Formats: FormatList = [
...
+				for (const moveSlot of pokemon.baseMoveSlots) {
+					if (!moveSlot.originalPoke) moveSlot.originalPoke = pokemon.m.value;
+				}
+				x++;
+			}
+		},
+		onBeforeSwitchIn(pokemon) {
+			let ngas = false;
+			for (const poke of this.getAllActive()) {
+				if (['neutralizinggas'].includes(this.toID(poke.ability))) {
```
KrisXV, line 2133:
```suggestion
				if (this.toID(poke.ability) === ('neutralizinggas' as ID)) {
```

config/formats.ts
```diff
@@ -2093,6 +2093,118 @@ export const Formats: FormatList = [
...
+					delete ally.volatiles[ally.m.innate];
+					if (!ngas || pokemon.getAbility().isPermanent) ally.addVolatile(ally.m.innate);
+				}
+			}
+		},
+		onSwitchInPriority: 2,
+		onSwitchIn(pokemon) {
+			let ngas = false;
+			for (const poke of this.getAllActive()) {
+				if (['neutralizinggas'].includes(this.toID(poke.ability))) {
```
KrisXV, line 2165: do the same thing here

sim/pokemon.ts
Marty-D: Not a huge fan, what makes this necessary now when it wasn't necessary to implement this OM back in October 2019?

KrisXV: I believe the reason why PiC was removed was because a ton of changes to the sim were made, so I don't think a property like this was needed before. I wonder if using `pokemon.m` can be used for this though?

HisuianZoroark: > I wonder if using pokemon.m can be used for this though?
It's possible. Just pushed the an alternative solution to Transform/mimic support that doesn't interfere with the core sim files.

data/mods/partnersincrime/scripts.ts
KrisXV: ig keep this for now, try to figure out an alternative to it in the meantime

KrisXV:
swag

PR #8869, opened 2022-08-18
Gen 1 JPN: Japanese event move incompatibilities

PlagueVonKarma:
So I realised that #8821 missed out on actually making the combination of Tackle and Dragon Rage illegal on Magikarp, which was kind of the point of what I was doing, so that was pretty dumb. Anyway, I also looked into the [Fan Club Chairman stuff](https://bulbapedia.bulbagarden.net/wiki/List_of_Japanese_event_Pok%C3%A9mon_distributions_(Generation_I)#Pok.C3.A9mon_Stamp_Pok.C3.A9mon) and realised they have similar issues, so let's sort this out.

- Rapidash is given at L40 with the moveset of Ember, Fire Spin, Stomp, and Pay Day. This means Growl and Tail Whip are impossible.
- Fearow is given at L25 with the moveset of Growl, Leer, Fury Attack, and Pay Day. This means Mirror Move(!) and Peck are impossible.
- Magikarp is given at L15 with Splash and Dragon Rage, with Tackle purposefully removed and thus made impossible.

This is your textbook move incompatibility problem in RBY: caught at a higher level, no move reminder available, so earlier moves cannot be obtained. We know the movesets of each event - check gen2/learnsets.ts for that - so this implementation should be fairly straightforward. Like other incompatibilities brought upon by these issues, I've hardcoded it, only this time into standard.

### DIFF
--- a/data/mods/gen1jpn/rulesets.ts
+++ b/data/mods/gen1jpn/rulesets.ts
@@ -3,7 +3,7 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 		effectType: 'ValidatorRule',
 		name: 'Standard',
 		ruleset: ['Obtainable', 'Desync Clause Mod', 'Sleep Clause Mod', 'Freeze Clause Mod', 'Species Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Endless Battle Clause', 'HP Percentage Mod', 'Cancel Mod'],
-		banlist: ['Dig', 'Fly', 'Rapidash + Pay Day + Growl', 'Rapidash + Pay Day + Tail Whip', 'Fearow + Pay Day + Peck', 'Fearow + Pay Day + Mirror Move', 'Magikarp + Dragon Rage + Tackle'],
+		banlist: ['Dig', 'Fly'],
 	},
 	nintendocup1997movelegality: {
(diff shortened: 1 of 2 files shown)


Marty-D:
The `standard` banlist isn't meant for legality; since this applies to every format it should be done like the rest of the rulesets files, overriding the `obtainablemoves` rule. Here's the one from Gen 7's mod as an example.
```js
	obtainablemoves: {
		inherit: true,
		banlist: [
			// Leaf Blade: Gen 6+ Nuzleaf level-up
			// Sucker Punch: Gen 4 Shiftry tutor
			'Shiftry + Leaf Blade + Sucker Punch',

			// Aura Break Zygarde can't be changed to 10% forme in gen 7
			// making moves only obtainable from gen 6 illegal
			'Zygarde-10% + Aura Break + Rock Smash',
			'Zygarde-10% + Aura Break + Secret Power',
			'Zygarde-10% + Aura Break + Strength',
		],
	},
```

In your case, you'll want to copy this rule from the Gen 2 mod (which is actually just Gen 1 legality, not sure who put it in Gen 2), and add your new bans to it.

PlagueVonKarma:
@Marty-D this should fix the problem! I put this pull request as a commented source.

PR #8870, opened 2022-08-19
Media approvals: Notify user on denial

MathyFurret:
https://www.smogon.com/forums/threads/deny-notification-on-requestshow.3705738/


AnnikaCodes:
Thanks!

PR #8872, opened 2022-08-22
Fix abilitysearch for special char separators

sparksblade:
Steam Engine's ability having 'Fire/Water' meant '/as water' was not returning the ability.


AnnikaCodes:
Thanks!

PR #8874, opened 2022-08-23
BDSP: Reintroduce Battle Festival Doubles as a challenge format

Karthik99999:
Many people want to play this format in the downtime between now and Scarlet & Violet VGC


KrisXV:
ok fine

PR #8875, opened 2022-08-25
Use opposing pokemon by default for `/weakness` & `/coverage` chat commands

chaorace:
This PR enhances the existing `/weakness` and `/coverage` commands to have useful default values when used in a battle room's chat. More specifically, here's how the new defaults work:

* `/weakness` now becomes `/weakness [currently active opposing pokemon's species]`
* `/coverage` now becomes `/coverage [currenly active opposing pokemon's type 1], [currenly active opposing pokemon's type 2]`

When there are multiple active pokemon on the opposing side who are eligible to be used in the default, these default commands will automatically multiplex to run once per eligible target.

Default resolution for a transformed Pokemon always bases information on the transformation.
Default resolution for a Pokemon actively under the effects of the Illusion ability will base imformation on the illusion.

Type resolution in all other cases is driven by `Pokemon.apparentType`, which is the same variable that drives the in-game type display.
- Prior to Gen-7, this display only updates when the species changes, so type resolution will only reflect what is known about the species.
- Post Gen-7, the type display is more dynamic and so too is the default parameter resolution. Note that the simulated type display is still not necessarily a perfect reflection of the actual typing used for damage calculation, so don't expect perfection here either.

Finally, please note that any hidden type information that could possibly be leaked by this implementation is almost certainly already being leaked by the simulator client's type display. This is because both elements are directly driven by the same underlying simulator variable. If any such hidden information leak is discovered, the remediation needs to be targetted at the implementation of `Pokemon.apparentType`, since otherwise the information will just continue leaking out via the client type display.


chaorace:
Rebased changes onto latest master

AnnikaCodes:
This seems fine at a glance, but I don't know enough about the battle simulator to be comfortable merging it.

singiamtel:
It currently doesn't show anything when the opponent is out of Pokémon, maybe it should give a warning of some kind instead of failing quietly

chaorace:
> It currently doesn't show anything when ~the opponent is out of Pokémon~ the battle is over, maybe it should give a warning of some kind instead of failing quietly

Which scenarios are you finding that it fails quietly? I tried to make it so that you would always see the `/help {command}` message whenever a valid target could not be found. If that's not sufficient, I'm happy to add some extra messages for various conditions. Did you have anything specific in mind?

singiamtel:
> > It currently doesn't show anything when ~the opponent is out of Pokémon~ the battle is over, maybe it should give a warning of some kind instead of failing quietly
> 
> Which scenarios are you finding that it fails quietly? I tried to make it so that you would always see the `/help {command}` message whenever a valid target could not be found. If that's not sufficient, I'm happy to add some extra messages for various conditions. Did you have anything specific in mind?

It should show some error when you use it on battles that have already finished, in my opinion. Now it sends nothing back

chaorace:
@singiamtel Currently, the convention with these two commands is to display their help text when they are incorrectly used. _I_ like that convention and think the message is plenty sufficient to explain why it didn't do anything against an empty field.

Of course, reasonable people can disagree on that point and it's certainly not something I plan to make a fuss over. All I ask is that you look at this example of the updated help text and tell me that it is not sufficient in lieu of an error:
```
/coverage - Provides the best effectiveness match-up defending against STAB moves from the active Pokemon on the opposing side.
/coverage [move 1], [move 2] ... - Provides the best effectiveness match-up against all defending types for given moves or attacking types
!coverage [move 1], [move 2] ... - Shows this information to everyone.
Adding the parameter 'all' or 'table' will display the information with a table of all type combinations.
```

If that's the case and we do want a genuine error message, would something like this be acceptable?
`A target Pokemon cannot be automatically selected for this command when used against an empty field`

PR #8877, opened 2022-08-26
Sketch Gen 8 BDSP Update

EnPassant4264:
The whole point of the Sketch Gen 8 Moves rule is to _avoid_ hardcoding the team validator based on format, so this updates the BDSP formats to match.

Not sure where the first four commits came from since they're already on main.


urkerab:
> Not sure where the first four commits came from since they're already on main.

Try rebasing instead of merging?

EnPassant4264:
Okay, I rebased it, and that's where the 41 new commits here are.

DaWoblefet:
With all the extra stuff in this PR, it's really hard to review. You can reach out to any of the devs on Discord (myself included) to help with fixing this up so it can be accepted.

KrisXV:
The hardcode is fine since it applies to the entire BDSP mod

PR #8879, opened 2022-08-27
Gen 2: Remove unnecessary changes for Block 

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9315445

### DIFF
--- a/data/mods/gen2/moves.ts
+++ b/data/mods/gen2/moves.ts
@@ -111,11 +111,6 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			},
 		},
 	},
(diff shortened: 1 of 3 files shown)


Marty-D:
This isn't true; in XY the flag was only removed from Block and Mean Look, and they either forgot Spider Web or intentionally left it. I'll accept deleting Block from Gen 2 though if you want to just keep that change.

Karthik99999:
Bulbapedia lied to me but I guess that's not a surprise.

Marty-D:
Thanks! 👍

PR #8882, opened 2022-08-30
Gen 1: Fix Dig & Fly

PlagueVonKarma:
This is a patch that will add the semi-invulnerability glitch to Dig & Fly. It was previously not fixed because the moves were banned, but with alternate formats being explored and Nintendo Cup '97 being appreciated by a small group of Japanese players, this issue is becoming more and more egregious. Arguably one of the biggest problems with the simulation, honestly?

The main reason this happened was that the para/confusion statuses removed the Dig/Fly volatiles upon being triggered, which was just...outright wrong. `twoturnmove` also removed the volatiles, which I changed to only happen for Solar Beam, Razor Wind, and Skull Bash, as in those cases, it's working correctly. With these, the glitch starts functioning correctly! 

Thanks to Karthik and pyuk for helping extensively with this!

--

Considerations have been made for Stadium. I simply put the old Dig/Fly weirdness into `moves.ts` and `conditions.ts`, as whatever that code is doing, it's working correctly in Stadium's case. I'll be ensuring this all works for my Stadium Zero patch as well.


PlagueVonKarma:
Alright, so I managed to get the move locking to work properly in the most flagrant display of me overthinking things ever. Thanks to pyuk for the pointers there.

--

> To fix the other issue I think you need to override `onTryMove` (inherited from gen 7) and remove the code that aborts the charge up if the user already has the move’s volatile.

@pyuk-bot I'm not actually sure what I'm supposed to remove here. I tried different areas but it always went wrong. It would end up causing an infinite loop of the Pokemon flying or digging. I guess that's the point, but even when the move is supposed to work, it wouldn't. 

Also, self-targeting moves [do work](https://cdn.discordapp.com/attachments/785695381635596298/1014514910564192276/fly_glitchsd.mp4) (thanks to Frrf for getting me the footage and listening to my coding pain) on a cartridge, but they miss the afflicted Pokemon in my current code. What would I need to change to make this work properly?

Sorry for being such a pain with all this...

PlagueVonKarma:
So I added that and it works well, but there is a kink in this.

So Charizard is (I believe) the only Pokemon to get Dig and Fly at once, so in the case this happens, the locked move will need to swap. Right now, it can't...

PlagueVonKarma:
Ok, I managed to fix the self-targeting bit on my own, I tested with Recover and it worked just fine.

With that, I believe this is ready for review!

data/mods/gen1/conditions.ts
```diff
@@ -269,4 +265,13 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 			this.effectState.duration = 2;
 		},
 	},
+	twoturnmove: {
+		// Skull Bash, SolarBeam, etc...
+		inherit: true,
+		onEnd(target) { // this stops dig and fly from being removed
+			target.removeVolatile('solarbeam');
+			target.removeVolatile('skullbash');
+			target.removeVolatile('razorwind');
```
Karthik99999, line 274:
```suggestion
		onEnd(target) {
			if (this.effectState.move !== 'dig' || this.effectState.move !== 'fly') {
				target.removeVolatile(this.effectState.move);
			}
```

HoeenCoder (approved):
Looks good to me, can you update and test this on the latest commit?

PlagueVonKarma:
> Looks good to me, can you update and test this on the latest commit?
@HoeenCoder 
So I'm trying it on the latest commit now, and something is somehow making the glitch not happen again. I'm not sure what's causing it, it was only barely functioning before, could you have a look?

HoeenCoder:
> > Looks good to me, can you update and test this on the latest commit?
> 
> @HoeenCoder So I'm trying it on the latest commit now, and something is somehow making the glitch not happen again, unless I am simply not running the localhost correctly. I'm not sure what's causing it, it was only barely functioning before, could you have a look?

What exactly is not working? I'm having troubles with my local test server in general currently so I can't test things now.

PlagueVonKarma:
> > > Looks good to me, can you update and test this on the latest commit?
> > 
> > @HoeenCoder So I'm trying it on the latest commit now, and something is somehow making the glitch not happen again, unless I am simply not running the localhost correctly. I'm not sure what's causing it, it was only barely functioning before, could you have a look?
> 
> What exactly is not working? I'm having troubles with my local test server in general currently so I can't test things now.

Basically all of it. It runs but the glitch doesn't happen, but I have a feeling I'm running something wrong LOL

HoeenCoder:
> > > > Looks good to me, can you update and test this on the latest commit?
> > > 
> > > 
> > > @HoeenCoder So I'm trying it on the latest commit now, and something is somehow making the glitch not happen again, unless I am simply not running the localhost correctly. I'm not sure what's causing it, it was only barely functioning before, could you have a look?
> > 
> > 
> > What exactly is not working? I'm having troubles with my local test server in general currently so I can't test things now.
> 
> Basically all of it. It runs but the glitch doesn't happen, but I have a feeling I'm running something wrong LOL

Contact me on discord, I'll see what I can do when I'm free in the evenings

PR #8884, opened 2022-09-01
Update formats.ts

TomOfTomKat:
Currently Chimera 1v1 is incorrectly validating illegal pokemon (including Pokestar mons), illegal movesets, and illegal abilities. Reverting to the ruleset that was used last time it had a ladder.


KrisXV:
In the future just update your open pull requests, otherwise the pr area gets flooded

PR #8885, opened 2022-09-01
Add umbrella item alias

singiamtel:
Added alias 'umbrella' -> 'Utility Umbrella' to make it easier to find via /dt


KrisXV:
sounds good to me! thank you!

PR #8891, opened 2022-09-11
Chat: Don't hyperlink parts of words with common TLDs

MathyFurret:
Typing something like "spiderman.jpg" is very often used to reference a meme or other image without actually linking or embedding it. However the link regex treats these like .jp links which is never intended.

### DIFF
--- a/server/chat-formatter.ts
+++ b/server/chat-formatter.ts
@@ -22,7 +22,7 @@ REGEXFREE SOURCE FOR LINKREGEX
 			\b [a-z0-9-]+ ( \. [a-z0-9-]+ )* \.
 			(
 				# followed either a common TLD...
-				com? | org | net | edu | info | us | jp(?!e?g)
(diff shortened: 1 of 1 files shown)


DaWoblefet:
This regex change doesn't seem to do anything when I test it:
image

MathyFurret:
Did you build the client with this change?

DaWoblefet:
> Did you full-build the client with this change?

I did not! After doing so, it appears to work correctly.

server/chat-formatter.ts
```diff
@@ -22,7 +22,7 @@ REGEXFREE SOURCE FOR LINKREGEX
 			\b [a-z0-9-]+ ( \. [a-z0-9-]+ )* \.
```
urkerab, line 22: I see there's a word boundary `\b` at the beginning of the domain, so perhaps what this really needs is one at the end?

AnnikaCodes:
Thanks!

PR #8893, opened 2022-09-13
Hangman: Make guesses clearable and loggable

MathyFurret:
https://www.smogon.com/forums/threads/ability-to-clear-guesses-in-hangman.3707793/ for context


mia-pi-git:
LGTM

PR #8895, opened 2022-09-13
Correct color of some Pokémon

richi3f:
Correct color of Morpeko (Hangry Mode), Hisuian Qwilfish, Hisuian Sneasel, Hisuian Braviary, Enamorus, and Sneasler.


DaWoblefet:
Thank you!

PR #8913, opened 2022-09-24
Fix Rollout/Ice Ball description

Karthik99999:
The rollout storage glitch should only be mentioned in the dlc1 mod for gen 8.


DaWoblefet:
While we're at it, might as well include Ice Face in the Gen 8 DLC1 mod.

DaWoblefet:
Thanks Karthik!

PR #8914, opened 2022-09-24
Let's Go: Mega Evolve Y & Itemless Megas

EvezaGit:
In Gen 7 Let's Go formats, these changes allow Pokemon to Mega Evolve without a held item, and allow Pokemon with multiple Mega Evolutions to choose between their Mega Evolutions during battle. This is accurate to cartridge. Have tested fairly extensively on my side server, so hopefully it all works! 

This pull request has been made simultaneously with one for smogon/pokemon-showdown-client.

### DIFF
--- a/data/mods/gen7letsgo/scripts.ts
+++ b/data/mods/gen7letsgo/scripts.ts
@@ -48,8 +48,8 @@ export const Scripts: ModdedBattleScriptsData = {
 		// mega evolution in LGPE does not require anything to be held
 		canMegaEvo(pokemon: Pokemon) {
 			const species = pokemon.baseSpecies;
-			for (let id in species.otherFormes) {
-				let forme = species.otherFormes[id];
+			for (const id in species.otherFormes) {
+				const forme = species.otherFormes[id];
 				if (forme.includes("-Mega") && forme !== pokemon.name) {
 					return forme;
 				}
--- a/data/mods/gen7letsgo/scripts.ts
+++ b/data/mods/gen7letsgo/scripts.ts
@@ -60,8 +60,8 @@ export const Scripts: ModdedBattleScriptsData = {
 		// lets player select either Charizard or Mewtwo mega evolution during battle
 		canMegaEvoY(pokemon: Pokemon) {
 			const species = pokemon.baseSpecies;
-			for (let id in species.otherFormes) {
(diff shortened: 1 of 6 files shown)


DaWoblefet:
Good start! Is there any particular reason to split up the protocol here into a mega and megay, as opposed to say, supporting an additional optional parameter on ``mega`` to a specific X or Y?

EssCode:
> Good start! Is there any particular reason to split up the protocol here into a mega and megay, as opposed to say, supporting an additional optional parameter on `mega` to a specific X or Y?

Have tried that approach in the past and it generally felt less convenient to implement- much easier to add a new simple parameter than to refactor every use of the mega value. Could try and implement it that way in the future as it's technically cleaner but this felt better than nothing for the time being

DaWoblefet:
It sounds to me like just adding a new parameter for the X/Y forms, and then you've got the logic for Mega Evolution without items already. Since all you're adding is an additional parameter, it's not a huge refactor of the protocol. I also think it's weird to assume X is the default Mega when you pick between two Megas.

KrisXV:
I'm pretty sure X is assumed to be default purely because it's before Y alphabetically. Either way, this is already a lot of work, so changing it entirely again would be a lot of work. This is a good stepping stone IMO until a refactor happens later down the road, because a refactor of that level would require a lot of bugfixing from my experience.

DaWoblefet:
My concern is that if we add this new megay protocol, then refactor it, you have to keep the old megay protocol around on the client to support older replays with it. We also have a lot of code reduplication that shouldn't be needed. Like granted, LGPE isn't exactly our highest-activity format or anything, but I think we can do it and it won't be _super_ bad.

EssCode:
That's a reasonable concern. I can look into the parameter alternative tomorrow if it's nobody else's priority, should help that I'm more familiar with the way the client and server interface now.

EssCode:
Nope, that kind of implementation is still beyond me. Before even considering the issue of turning every instance of canMegaEvo into an array, adding a parameter to 'mega' in the choose command is finicky enough, as is carrying the value of that parameter all the way to runMegaEvo. If anyone else wants to give it a shot in the near future that would be great, otherwise this is somewhat awkward as there's some minor interest in LGPE VGC and this mechanic is quite important there.

PR #8916, opened 2022-09-25
Add Hindi to the youtube groupwatch whitelist

KennedyLFC:
at the request of Arsenal, they have an event planned to watch a Hindi movie using it


DaWoblefet:
Is there a reason why we have this hardcoded list if we're just going to keep adding more and more rooms to it? Let's make it something configurable through something like an admin command in roomsettings or something (to be clear you don't have to do this Kennedy).

mia-pi-git:
TBH, I could probably just make it open to all public rooms. I doubt it'll get used enough to have too much of an issue with quota.

AnnikaCodes:
This is Main-specific anyway, so it should be configurable by admins or at least moved to config.js, if not removed altogether.

KennedyLFC:
could we merge this anyways just as a short-term thing? Hindi needs it for an event they have planned

PR #8917, opened 2022-09-25
/movesearch: Search past gen learnsets

dot-Comfey:
The code for /movesearch in past gens would add a move for a Pokemon without regards to if it could learn the move in a given generation; this change will check a move's source generations and add moves only if the move can be learned in a current or past generation.

This fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8608692

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -1618,13 +1618,28 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			usedSpecies = Utils.deepClone(mod.species.get(usedSpecies.baseSpecies));
 			usedSpeciesLearnset = Utils.deepClone(mod.species.getLearnset(usedSpecies.id) || {});
 		}
-		const lsetData = new Set(Object.keys(usedSpeciesLearnset));
+		const lsetData = new Set();
+		for (const move in usedSpeciesLearnset) {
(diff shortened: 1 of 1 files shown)


server/chat-plugins/datasearch.ts
```diff
@@ -1618,13 +1618,28 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 			usedSpecies = Utils.deepClone(mod.species.get(usedSpecies.baseSpecies));
 			usedSpeciesLearnset = Utils.deepClone(mod.species.getLearnset(usedSpecies.id) || {});
 		}
-		const lsetData = new Set(Object.keys(usedSpeciesLearnset));
+		const lsetData = new Set();
+		for (const move in usedSpeciesLearnset) {
+			let learnset = mod.species.getLearnset(usedSpecies.id);
```
AnnikaCodes, line 1623:
```suggestion
			const learnset = mod.species.getLearnset(usedSpecies.id);
```

AnnikaCodes: You should use `const` for any variable you don't modify.

dot-Comfey:
I realized I should be creating new branches to create pull requests instead of using the master branch, so I'll close this pull request and open a new one.

PR #8918, opened 2022-09-26
Movesearch: Search past gen learnsets

dot-Comfey:
Re-creation of https://github.com/smogon/pokemon-showdown/pull/8917

The code for /movesearch in past gens would add a move for a Pokemon without regards to if it could learn the move in a given generation; this change will check a move's source generations and add moves only if the move can be learned in a current or past generation.

This fixes https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8608692


DaWoblefet:
This looks great to me! Not sure if @AnnikaCodes had any pending thoughts but this looks good to go to master.

AnnikaCodes:
no thoughts head empty

DaWoblefet:
Thanks dot Comfey!

PR #8919, opened 2022-09-29
Improved typing for `utils.ts`, and for iterating over base stats

GregLyons:
## Improved typing for `utils.ts`

- `compare`
  - Taking in arguments of same type now enforced by generic parameter.
  - Type assertions could be removed to make the code safer (e.g. by adding a `typeof b === 'string'` check in the case where `a` is `string`), but the check at the beginning of the function should suffice.
- `deepClone`
  - Takes in object of a given type, and returns object of the same type.
  - Added type assertions in the body to fix type errors.

## Arrays for iterating over stats

Added `Stats` namespace to `global-types.ts`. Now one can iterate over them directly in the code with `for (let statName of Stats.statIDs)` and get Intellisense (i.e. autocomplete on stat names) and alleviate type errors (when doing `for (let statName in species.baseStats)`, `statName` is only cast as `string`).

## Added missing tiers to `TierTypes`

Upon strengthening the `deepClone` type from `any`, I had to fix some type errors. Some of them involved trying to update the tier of the cloned object to, e.g. `"MnM"`, which isn't actually a tier. Since the clone is no longer `any`, this type error came up. I'm assuming such tiers should be added to `TierTypes.Others`, but if this is not the case I'm happy to add them where they should be added.

## Other tricky changes

- `othermetas.ts`
  - Removed `delta.types = null`, which was throwing a type error.
  - This change would be relevant to line 657, where `getDataPokemonHTML` is called on `deltas`. The `if (species.types)` will now go through, but then the `for` loop will do nothing because the array is empty. Thus, this change shouldn't have any effect beyond fixing the type error.

### DIFF
--- a/data/mods/gen7mixandmega/scripts.ts
+++ b/data/mods/gen7mixandmega/scripts.ts
@@ -103,7 +103,7 @@ export const Scripts: ModdedBattleScriptsData = {
 				baseStats[statName] = this.battle.clampIntRange(baseStats[statName] + deltas.baseStats[statName], 1, 255);
 			}
 			species.weighthg = Math.max(1, species.weighthg + deltas.weighthg);
-			// @ts-ignore
+			// @ts-ignore originalMega is not a defined property of the Species class, but we need to add it here
 			species.originalMega = deltas.originalMega;
 			species.requiredItem = deltas.requiredItem;
 			if (deltas.isMega) species.isMega = true;
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -162,8 +162,6 @@ export function visualize(value: any, depth = 0): string {
  * In other words: `[num, str]` will be sorted A to Z, `[num, {reverse: str}]` will be sorted Z to A.
  */
 export function compare<T extends Comparable>(a: T, b: T): number {
-	if (typeof a !== typeof b) throw new Error(`Passed values ${a} and ${b} must be of same type`);
-
 	if (typeof a === 'number') {
 		return a - (b as number);
 	}
@@ -181,7 +179,7 @@ export function compare<T extends Comparable>(a: T, b: T): number {
 		return 0;
 	}
 	if ('reverse' in a) {
-		return compare((b as { reverse: Comparable }).reverse, a.reverse);
+		return compare((b as {reverse: Comparable}).reverse, a.reverse);
 	}
 	throw new Error(`Passed value ${a} is not comparable`);
 }


server/chat-plugins/othermetas.ts
```diff
@@ -591,7 +590,7 @@ export const commands: Chat.ChatCommands = {
 		} else if (mixedSpecies.weighthg >= 100) {
 			weighthit = 40;
 		}
-		const details: {[k: string]: string} = {
```
GregLyons, line 594: Because `num` and `gen` are numbers.

server/chat-plugins/othermetas.ts
```diff
@@ -591,7 +590,7 @@ export const commands: Chat.ChatCommands = {
...
+		const details: {[k: string]: string | number} = {
```
GregLyons, line 593: Because `gen` and `num` are `number` type

server/chat-plugins/othermetas.ts
```diff
@@ -462,15 +461,15 @@ export const commands: Chat.ChatCommands = {
...
+			for (const stat of Stats.statIDs) {
 				species.baseStats[stat] = flippedStats[stat];
 			}
 			this.sendReply(`|raw|${Chat.getDataPokemonHTML(species, dex.gen)}`);
 			return;
 		}
 		const stats = Object.values(species.baseStats).reverse();
 		for (const [i, statName] of Object.keys(species.baseStats).entries()) {
-			species.baseStats[statName] = stats[i];
+			species.baseStats[statName as keyof typeof species.baseStats] = stats[i];
```
GregLyons, line 472: Basically, `statName` is of type `StatID`

sim/global-types.ts
```diff
@@ -24,8 +24,14 @@ type PokemonSlot = '' | string & {__isSlot: true};
...
 type GenderName = 'M' | 'F' | 'N' | '';
-type StatIDExceptHP = 'atk' | 'def' | 'spa' | 'spd' | 'spe';
-type StatID = 'hp' | StatIDExceptHP;
+const _statIDsExceptHP = ['atk', 'def', 'spa', 'spd', 'spe'] as const;
+const _statIDs = [..._statIDsExceptHP, 'hp'] as const;
+namespace Stats {
+	export const statIDsExceptHP = _statIDsExceptHP;
+	export const statIDs = _statIDs;
+}
+type StatIDExceptHP = typeof _statIDsExceptHP[number];
```
GregLyons, line 33: Re-defining in terms of array so only need to change the stat list in one place. `as const` changes the array into a tuple, so that `StatIDExceptHP` is the type union of the elements of `_statIDsExceptHP`, i.e. `'atk' | 'def' | 'spa' | 'spd' | 'spe'`, just as before.

DaWoblefet:
Looks very good from a cursory glance to me.

GregLyons:
🤦‍♂️ I ran `npm test` after making the change to `deepClone`, which is how I found that lint error about `startsWith`. I guess I must not have let it run all the way or something. Now it looks like it passes `npm run test` (both mocha and tsc).

I wasn't able to import `Stats` into `data/mods/mixandmega/scripts.ts` for some reason (I'm not experienced with `namespace`), so I had to use an alternate method.

config/formats.ts
```diff
@@ -1278,7 +1278,7 @@ export const Formats: FormatList = [
 
 			const mixedSpecies = this.dex.deepClone(species);
 			mixedSpecies.weightkg =
-				Math.max(0.1, +(species.weightkg + crossSpecies.weightkg - crossPrevoSpecies.weightkg)).toFixed(1);
+				parseInt(Math.max(0.1, +(species.weightkg + crossSpecies.weightkg - crossPrevoSpecies.weightkg)).toFixed(1));
```
AnnikaCodes, line 1281: Does Math.max not return a `number`?

data/mods/gen7mixandmega/scripts.ts
```diff
@@ -99,10 +99,11 @@ export const Scripts: ModdedBattleScriptsData = {
 				species.types = [species.types[0], deltas.type];
 			}
 			const baseStats = species.baseStats;
-			for (const statName in baseStats) {
+			for (const statName of Stats.statIDs) {
 				baseStats[statName] = this.battle.clampIntRange(baseStats[statName] + deltas.baseStats[statName], 1, 255);
 			}
 			species.weighthg = Math.max(1, species.weighthg + deltas.weighthg);
+			// @ts-ignore
```
AnnikaCodes, line 106: ts-ignores should explain why they are necessary

lib/utils.ts
```diff
@@ -162,7 +161,9 @@ export function visualize(value: any, depth = 0): string {
  *
  * In other words: `[num, str]` will be sorted A to Z, `[num, {reverse: str}]` will be sorted Z to A.
  */
-export function compare(a: Comparable, b: Comparable): number {
+export function compare<T extends Comparable>(a: T, b: T): number {
+	if (typeof a !== typeof b) throw new Error(`Passed values ${a} and ${b} must be of same type`);
```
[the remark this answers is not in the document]
AnnikaCodes: I don't know that this should be a runtime error.

lib/utils.ts
```diff
@@ -180,7 +181,7 @@ export function compare(a: Comparable, b: Comparable): number {
 		return 0;
 	}
 	if ('reverse' in a) {
-		return compare((b as {reverse: string}).reverse, a.reverse);
+		return compare((b as { reverse: Comparable }).reverse, a.reverse);
```
AnnikaCodes, line 184:
```suggestion
		return compare((b as {reverse: Comparable}).reverse, a.reverse);
```
style nit

sim/global-types.ts
AnnikaCodes: Do these need to be defined as `const`s and then added to the namespace? Can't you just say 
```
namespace Stats {
	export const statIDsExceptHP = ['atk', 'def', 'spa', 'spd', 'spe'] as const;
...
```

AnnikaCodes:
This looks good to me!

sim/global-types.ts
GregLyons: If you define them in the namespace, you can't use them outside of it for `StatIDExceptHP`, etc. I did it this way so the arrays in the namespace and the types outside it could be in sync. We _could_ move `StatID` into the namespace, but then we'd need to refactor a lot of imports, which I wanted to avoid.

config/formats.ts
GregLyons: The `toFixed(1)` changes it to a `string`, but it wasn't being picked up b/c `mixedSpecies` was `any`, and I guess JS was silently converting `string` to `number` in calculations.

lib/utils.ts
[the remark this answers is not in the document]
GregLyons: Yeah, actually I think this check is unnecessary since now TS compiler should prevent it. I removed it.

data/mods/gen7mixandmega/scripts.ts
GregLyons: Added an explanation. Lmk if it's preferable to do `(species as any).originalMega` instead, in this case, which would be more consistent with how it's done in the rest of the codebase when adding a property to a class.

On a side note, I wonder if it would be helpful to add something like a `MixedSpecies` type, which would extend `Species` with the `originalMega` property. We could probably have a generic `AugmentedSpecies<T>` class, where `MixedSpecies = AugmentedSpecies<{ originalMega: ___ }>`. Then `doGetMixedSpecies` would return that `AugmentedSpecies` class instead, so callers of this function know that it returns something with the `originalMega` property.

lib/utils.ts
GregLyons: Applied

AnnikaCodes (approved):

AnnikaCodes:
I think there's probably still more incremental improvements here, but this is a good start :)

AnnikaCodes:
We need to add tests for creating battles in mods before remerging this

PR #8920, opened 2022-09-29
Fix Defiant/Competitive announcing at +6 atk/sp. atk

Karthik99999:


urkerab:
What's the current and desired behaviour at +6 atk/sp. atk?

urkerab:
(in particular, with regards to the "can't go any higher" message at +6)

Karthik99999:
current behavior is the both the ability activation message and "can't go any higher" message is shown, desired behavior is for only the "can't go any higher" message to be shown

DaWoblefet:
> What's the current and desired behaviour at +6 atk/sp. atk?

See this for an example: https://twitter.com/DaWoblefet/status/1575293196887887872

urkerab:
Well, given that it's my fault that the sim does this, I think I should at least say how I would have chosen to code the fix.

Currently, `Battle#boost` for an Ability will by default announce the ability only if the boost is successful and will be silent if the boost failed (due to maxing out). This is suitable for e.g. Speed Boost.
Meanwhile Gooey manually announces itself and passes `true` for `isSecondary` which prevents a duplicate ability announcement while also noisily maxing out, unless the boost was inhibited (e.g. Clear Body).
So my approach would be that Defiant should pass `true` for `isSelf` instead (no Ability currently does this). This currently has the default effect, but the code to noisily max out for `isSecondary` could trivially be modified to affect `isSelf` as well. (This would then remove the need for Defiant to announce itself manually.)

DaWoblefet:
Thank you both!

PR #8921, opened 2022-10-04
feat(nvmrc): adding nvmrc file

girordo:
Adding nvmrc file for node version check


DaWoblefet:
Can you explain what this is intended for? We don't require people be on the latest version of Node.js to run Showdown, nor do we require an older version to run Showdown.

girordo:
Usually dotfiles like .nvmrc are used to config for a better developer experience and make the repo works fine.
I understand that I should better improve my first comment, I sorry about that.
If you don't need this version, it's good to specify what version is better to run.

DaWoblefet:
I'm not a huge fan of this personally, because we basically just recommend the latest stable version. But with this file, we'd have to continually update it with each new node release. Based on their documentation, maybe we want to use ``lts/*``? I'm still not convinced it's needed, though; it looks like you have to install additional tools to even use this.

PR #8922, opened 2022-10-05
Add FAQs for custom avatars and the privacy policy

KennedyLFC:
these are very frequently asked questions across the sim and having global FAQs for them rather than relying on room FAQs is just simplifying and uniformising things.


server/chat-commands/info.ts
```diff
@@ -2072,6 +2072,12 @@ export const commands: Chat.ChatCommands = {
 		if (showAll || ['vpn', 'proxy'].includes(target)) {
 			buffer.push(`<a href="https://pokemonshowdown.com/${this.tr`pages/proxyhelp`}">${this.tr`Proxy lock help`}</a>`);
 		}
+		if (showAll || ['ca', 'customavatar', 'customavatars'].includes('target') {
```
DaWoblefet, line 2075:
```suggestion
		if (showAll || ['ca', 'customavatar', 'customavatars'].includes('target')) {
```
Sorry, I'm bad lol

DaWoblefet:
Thanks Kennedy!

PR #8923, opened 2022-10-06
Correct Node version documentation

DaWoblefet:
I think this is everything; updates any old references we have to clarify we require at least 14.


AnnikaCodes:
Thanks!

PR #8924, opened 2022-10-06
Fix minor friends bugs

DaWoblefet:
Fixed an issue where if you are blocking incoming friend requests, you cannot cancel sent friend requests:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9323392

Clarified the response of an error message where sending friend requests would be rejected if both incoming+outgoing requests are greater than max requests:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9334011


AnnikaCodes:
Thanks!

PR #8926, opened 2022-10-09
add missed portuguese translations

hpbonfim:
I noticed that some sentences in Portuguese were missing and I added them!


AnnikaCodes:
I don't speak Portuguese but I can understand a bit of it, and this seems to be correct. If Portuguese staff want to change it then that's fine but this seems like an improvement over English, thanks!

PR #8928, opened 2022-10-10
Validator: Push problems for all failed events

DaWoblefet:
Currently, when validating events that fail because of "either/or" reasons, the validator just picks the first event it can find and claims incompatibility with that event. For example, consider a Shiny Eternatus with 0 Atk IVs. Compare before and after:

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

So I think it's much clearer to list all possible events and their incompatibility with each. This does have the downside of Pokemon with many events making the problems it gives back a bit longer, but I think it's a fine trade-off. This doesn't happen with _every_ event Pokemon incompatibility edge case though; Pokemon that can't be Shiny are a common example of ones affected:

<details>
<summary>Example with Shiny Victini</summary>

</details>

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -684,18 +684,15 @@ export class TeamValidator {
 				} else {
 					problems.push(`${species.name} is only obtainable from events - it needs to match one of its events, such as:`);
 				}
-				let eventInfo = eventData[0];
-				let eventNum = 1;
 				for (const [i, event] of eventData.entries()) {
 					if (event.generation <= dex.gen && event.generation >= this.minSourceGen) {
-						eventInfo = event;
-						eventNum = i + 1;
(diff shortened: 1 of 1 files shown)


sim/team-validator.ts
```diff
@@ -684,18 +684,15 @@ export class TeamValidator {
 				} else {
 					problems.push(`${species.name} is only obtainable from events - it needs to match one of its events, such as:`);
```
urkerab, line 685:
```suggestion
					problems.push(`${species.name} is only obtainable from events - it needs to match one of its events:`);
```

"such as" no longer makes sense if you're listing all of the events.

DaWoblefet:
Build failure is from a typescript bump, not me :P

PR #8930, opened 2022-10-11
Fix Neutralizing Gas Primal weather interaction

singiamtel:
Neutralizing gas now negates strong weather abilities and restores them on exit.

As suggested on [this Mechanics Bugs card](https://github.com/smogon/pokemon-showdown/projects/3#card-85843136)

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -2348,6 +2348,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 			if (pokemon.transformed) return;
 			this.add('-ability', pokemon, 'Neutralizing Gas');
 			pokemon.abilityState.ending = false;
+			const strongWeathers = ['desolateland', 'primordialsea', 'deltastream'];
 			for (const target of this.getAllActive()) {
 				if (target.illusion) {
 					this.singleEvent('End', this.dex.abilities.get('Illusion'), target.abilityState, target, pokemon, 'neutralizinggas');
@@ -2355,6 +2356,9 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 				if (target.volatiles['slowstart']) {
 					delete target.volatiles['slowstart'];
 					this.add('-end', target, 'Slow Start', '[silent]');
+				}
+				if (strongWeathers.includes(target.getAbility().id)) {
+					this.singleEvent('End', this.dex.abilities.get(target.getAbility().id), target.abilityState, target, pokemon, 'neutralizinggas');
 				}
 			}
 		},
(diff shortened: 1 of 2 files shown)


DaWoblefet (changes requested):
Just tried this solution out, and while it does pass the test, it doesn't actually fix the bug. If you want to go the hardcode route, I would recommend the same style that we do for Illusion currently (see Neutralizing Gas's ``onPreStart``)

image

DaWoblefet:
The "correct" fix would be to make it so Neutralizing Gas sends an 'End' event for _all_ Abilities. However, we already aren't doing that (see current hardcodes for Illusion and Slow Start), so it's just as well we add these three to the list as well.

singiamtel:
Sorry, I should have tested this better. It was only working when Neutralizing Gas was on the battlefield before Desolate Land. I added a check to `onPreStart` and it should be working properly now

sim/field.ts
```diff
@@ -103,11 +103,16 @@ export class Field {
 	}
 
 	suppressingWeather() {
+		const strongWeathers = ['desolateland', 'primordialsea', 'deltastream'];
```
DaWoblefet, line 106: Changes in sim/field.ts should no longer be necessary with us ending the weather properly.

data/abilities.ts
```diff
@@ -2357,6 +2357,10 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 					this.add('-end', target, 'Slow Start', '[silent]');
 				}
 			}
+			const strongWeathers = ['desolateland', 'primordialsea', 'deltastream'];
+			if (strongWeathers.includes(this.field.getWeather().id)) {
+				this.field.clearWeather();
```
DaWoblefet, line 2362: This works; what happens if you send an 'End' event associated with the Pokemon causing the Primal weather event? I would prefer to do it that way rather than just clearing weather directly.

singiamtel: Yea that's cleaner. It works too, just changed it

data/abilities.ts
```diff
@@ -2356,6 +2357,9 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 					delete target.volatiles['slowstart'];
 					this.add('-end', target, 'Slow Start', '[silent]');
 				}
+				if (strongWeathers.includes(target.getAbility().id)) {
+					this.add('-end', target, target.getAbility().id, '[silent]');
```
DaWoblefet, line 2361:
```suggestion
					this.singleEvent('End', this.dex.abilities.get(target.getAbility().id), target.abilityState, target, pokemon, 'neutralizinggas');
```
Should make it so you can drop changes in sim/field.ts

singiamtel:
Brilliant! Removed the changes in ``field.ts`` and updated the event as you suggested

DaWoblefet (approved):

DaWoblefet:
Thank you much!

PR #8932, opened 2022-10-11
Utils: Avoid `any`

foolmoron:
- Improved types in `utils` to avoid `any`
  - https://github.com/smogon/pokemon-showdown/commit/139470c576fa8eb7f79eeac40335c77738910124
- Fixed a bunch of usage code that is now type-aware
  - There's some slight behavioral changes here so please check this since I don't have full context!
  - https://github.com/smogon/pokemon-showdown/commit/83345e180c4c74c923faafbb317c5b0a19fe9bd0

### DIFF
--- a/data/mods/gen1/rulesets.ts
+++ b/data/mods/gen1/rulesets.ts
@@ -17,7 +17,8 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 			const bst = newSpecies.bst;
 			if (bst <= 350) {
 				newSpecies.bst = 0;
-				for (const stat in newSpecies.baseStats) {
+				let stat: StatID;
+				for (stat in newSpecies.baseStats) {
 					if (stat === 'spd') continue;
 					newSpecies.baseStats[stat] = this.clampIntRange(newSpecies.baseStats[stat] * 2, 1, 255);
 					newSpecies.bst += newSpecies.baseStats[stat];
@@ -44,7 +45,7 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 				spe: newSpecies.baseStats.hp,
 			};
 			for (const i in newSpecies.baseStats) {
-				newSpecies.baseStats[i] = stats[i];
+				newSpecies.baseStats[i as StatID] = stats[i as StatID];
 			}
 			return newSpecies;
 		},
@@ -61,7 +62,8 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 			const pst: number = newSpecies.bst - newSpecies.baseStats['hp'];
 			const scale = 500 - newSpecies.baseStats['hp'];
 			newSpecies.bst = newSpecies.baseStats['hp'];
-			for (const stat in newSpecies.baseStats) {
+			let stat: StatID;
+			for (stat in newSpecies.baseStats) {
 				if (stat === 'hp' || stat === 'spd') continue;
 				newSpecies.baseStats[stat] = this.clampIntRange(newSpecies.baseStats[stat] * scale / pst, 1, 255);
 				newSpecies.bst += newSpecies.baseStats[stat];
--- a/data/mods/gen1/rulesets.ts
+++ b/data/mods/gen1/rulesets.ts
@@ -44,8 +44,9 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 				spd: newSpecies.baseStats.atk,
 				spe: newSpecies.baseStats.hp,
 			};
-			for (const i in newSpecies.baseStats) {
-				newSpecies.baseStats[i as StatID] = stats[i as StatID];
+			let stat: StatID;
+			for (stat in newSpecies.baseStats) {
+				newSpecies.baseStats[stat] = stats[stat];
 			}
 			return newSpecies;
 		},
--- a/data/mods/gen7mixandmega/scripts.ts
+++ b/data/mods/gen7mixandmega/scripts.ts
@@ -1,3 +1,7 @@
+interface SpeciesMixAndMega extends Mutable<Species> {
+	originalMega?: string | Species;
+}
+
 export const Scripts: ModdedBattleScriptsData = {
 	inherit: 'gen7',
 	init() {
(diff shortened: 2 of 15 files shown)


sim/global-types.ts
```diff
@@ -64,7 +64,7 @@ namespace TierTypes {
 	export type Singles = "AG" | "Uber" | "(Uber)" | "OU" | "(OU)" | "UUBL" | "UU" | "RUBL" | "RU" | "NUBL" | "NU" |
 	"(NU)" | "PUBL" | "PU" | "(PU)" | "NFE" | "LC";
 	export type Doubles = "DUber" | "(DUber)" | "DOU" | "(DOU)" | "DBL" | "DUU" | "(DUU)" | "NFE" | "LC";
-	export type Other = "Unreleased" | "Illegal" | "CAP" | "CAP NFE" | "CAP LC";
+	export type Other = "Unreleased" | "Illegal" | "CAP" | "CAP NFE" | "CAP LC" | "NS" | "CE" | "MnM";
```
foolmoron, line 67: @KrisXV Continuing from discord: Is there another place we can put these special tier values? It seems dangerous to have values that are outside of the enum types. Maybe another enum called `Special` or something?

KrisXV: These are never actually used for tiering and the reason for declaring types for `tier` etc is to ensure there's no typos when editing data in `formats-data` files. We could do an enum specific to the OM chat file or something, but these shouldn't be here.

data/mods/mixandmega/scripts.ts
```diff
@@ -95,10 +95,12 @@ export const Scripts: ModdedBattleScriptsData = {
 				species.types = [species.types[0], deltas.type];
 			}
 			const baseStats = species.baseStats;
-			for (const statName in baseStats) {
+			let statName: StatID;
+			for (statName in baseStats) {
 				baseStats[statName] = this.battle.clampIntRange(baseStats[statName] + deltas.baseStats[statName], 1, 255);
 			}
 			species.weighthg = Math.max(1, species.weighthg + deltas.weighthg);
+			// @ts-ignore
```
KrisXV, line 103: I don't like using `ts-ignore` in places where it was previously used and then removed, I would rather typecast `any` or `unknown` here.

data/mods/gen7mixandmega/scripts.ts
```diff
@@ -99,10 +99,12 @@ export const Scripts: ModdedBattleScriptsData = {
 				species.types = [species.types[0], deltas.type];
 			}
 			const baseStats = species.baseStats;
-			for (const statName in baseStats) {
+			let statName: StatID;
+			for (statName in baseStats) {
 				baseStats[statName] = this.battle.clampIntRange(baseStats[statName] + deltas.baseStats[statName], 1, 255);
 			}
 			species.weighthg = Math.max(1, species.weighthg + deltas.weighthg);
+			// @ts-ignore
```
KrisXV, line 107: ditto

KrisXV:
I'm renaming this PR because the title causes ambiguity with our `utils` file and property

data/mods/gen1/rulesets.ts
```diff
@@ -44,7 +45,7 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 				spe: newSpecies.baseStats.hp,
 			};
 			for (const i in newSpecies.baseStats) {
-				newSpecies.baseStats[i] = stats[i];
+				newSpecies.baseStats[i as StatID] = stats[i as StatID];
```
KrisXV, line 48: I think I prefer the type declaring above the `for` loop compared to typecasting.

server/chat-plugins/datasearch.ts
```diff
@@ -1613,7 +1613,7 @@ function runMovesearch(target: string, cmd: string, canAll: boolean, message: st
 
 	const getFullLearnsetOfPokemon = (species: Species) => {
 		let usedSpecies: Species = Utils.deepClone(species);
-		let usedSpeciesLearnset: LearnsetData = Utils.deepClone(mod.species.getLearnset(usedSpecies.id));
+		let usedSpeciesLearnset: LearnsetData | undefined = Utils.deepClone(mod.species.getLearnset(usedSpecies.id));
```
KrisXV, line 1616: Realistically couldn't the typecast here just be removed?

foolmoron: Yes, but it resulted in a somewhat messy type:
```
Mutable<{
    [moveid: string]: string[];
} | undefined>
```

But I def prefer leaving the type off, so I'll update it

sim/global-types.ts
foolmoron: I see. I have an idea, will give it a shot

data/mods/mixandmega/scripts.ts
foolmoron: I was following the pattern used in other `originalMega` references like this:
https://github.com/smogon/pokemon-showdown/blob/master/config/formats.ts#L781-L782

I feel like the right solution is to add this prop to `Species`

data/mods/gen1/rulesets.ts
foolmoron: in latest rev

sim/global-types.ts
foolmoron: How's the latest rev? Keeps the values in that file

data/mods/mixandmega/scripts.ts
KrisXV: No, there’s no need to edit the species class for an OM

sim/global-types.ts
KrisXV: That's fine

data/mods/gen7mixandmega/scripts.ts
AnnikaCodes: This at the least needs an explanation if it's not going to be converted to a cast.

data/mods/mixandmega/scripts.ts
foolmoron: Casted in latest!

data/mods/gen7mixandmega/scripts.ts
```diff
@@ -25,15 +29,17 @@ export const Scripts: ModdedBattleScriptsData = {
 
 			const isUltraBurst = !pokemon.canMegaEvo;
 			// @ts-ignore
-			const species: Species = this.getMixedSpecies(pokemon.m.originalSpecies, pokemon.canMegaEvo || pokemon.canUltraBurst);
```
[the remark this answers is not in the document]
foolmoron: Yes because that is for the `this.getMixedSpecies` reference, which is a custom func that does not exist on the `this` type

mia-pi-git:
@foolmoron, are you still interested in working on this?

foolmoron:
> @foolmoron, are you still interested in working on this?

Prob not sorry

mia-pi-git:
Okay, thanks!

PR #8935, opened 2022-10-14
Gen2 Random Battles: HP grass instead of Giga Drain on celebi and vileplume

adrivrie:
Just a small sets change.

https://discord.com/channels/294651453279174656/809413306854932510/1030479956762963999


AnnikaCodes:
I don't really see the benefit but people seem approving.

PR #8936, opened 2022-10-15
Gen2 Random Battles: cull explosion and selfdestruct on sleeptalk

adrivrie:
This change right now should only affect Steelix, but the general rule is valid.

https://discord.com/channels/294651453279174656/809413306854932510/1030858745858756658


AnnikaCodes:
Thanks!

Issue #8868, opened 2022-08-16
Policy Suggestion: Mechanics Fixing PRs Must Have Tests


gigalh128:
Showdown has issues with buggy move mechanics being found in production. Ideally, we would have tests for moves to catch these before this happens. I know we can't feasibly write comprehensive tests to prevent this, as there are just too many moves. However, we can mitigate this, and one simple idea is to require anyone who makes a PR fixing a certain mechanic to also make unit tests that demonstrate the mechanic has been fixed correctly. This is what I have done in my [gen 1 fix for haze and disable](https://github.com/smogon/pokemon-showdown/pull/8807). By implementing this policy, we can build a better culture of writing unit tests so that bugs are found **before** reaching production, and doing so will also get people more familiar with the Showdown codebase.

Let me know your thoughts on this policy suggestion.

AnnikaCodes:
I support this. Unit tests are generally helpful and mechanics are relatively easy to test. I do think that for new contributors, we should make it clear that maintainers will help them to write unit tests (testing frameworks can be hard to wrap your head around as a new contributor).

DaWoblefet:
I would prefer this as an encouraged suggestion, not mandatory. Certain small mechanics fixes don't warrant tests, and I wouldn't want someone to say, open a PR with "Prevent Dynamax Cannon from being called by Instruct" with a test where it's not really needed. I am in agreement with Annika that seasoned maintainers helping with tests is a great middle ground if it's a new developer.

joebangara43903:
suck my nigger dick

Zarel:
The reason I've never required this is because it's kind of hostile to new contributors. "Thanks for helping us for free, now here's some more work for you."

I've also been on the other end of this ( https://github.com/mochajs/mocha/pull/2746 ), where I submit a bugfix, and they demand specific kinds of tests that aren't possible, and the bug sits unfixed because that's the rules.

I agree with Wob; encouraging more unit tests is good, but I think it would be bad to make it a requirement.

Slayer95:
Consensus doesn't favor this approach.

The barrier of entry for contributors, and especially mechanics contributors is already high enough.

PR #8938, opened 2022-10-17
Revert "Improved typing for `utils.ts`, and for iterating over base stats"

AnnikaCodes:
Reverts smogon/pokemon-showdown#8919. This broke certain mods and needs further testing.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1193,7 +1193,7 @@ export const Formats: FormatList = [
 
 			const mixedSpecies = this.dex.deepClone(species);
 			mixedSpecies.weightkg =
-				parseInt(Math.max(0.1, +(species.weightkg + crossSpecies.weightkg - crossPrevoSpecies.weightkg)).toFixed(1));
+				Math.max(0.1, +(species.weightkg + crossSpecies.weightkg - crossPrevoSpecies.weightkg)).toFixed(1);
 			mixedSpecies.nfe = false;
 			mixedSpecies.evos = [];
 			mixedSpecies.eggGroups = crossSpecies.eggGroups;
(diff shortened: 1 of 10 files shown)


GregLyons:
1. Was it a compile-time error or a run-time error? And which mods were breaking? Aside from the `parseInt` and `startsWith` changes, the only changes which should change the run-time behavior are the ones that change `stat in newSpecies.baseStats` to `stat of Stats.statIDs`. I was assuming that the object `newSpecies.baseStats` always has `[hp, atk, def, spa, spd, spe]` as keys, but if at some point that object has a key added or removed, then my change would've broken that. If instead it's compile-time errors, I'm happy to fix those and re-open the PR.
2. What unit-testing framework do y'all use? (#8868) I'm most familiar with Jest, but I'm willing to pick up whatever PS uses. It seems like the relevant code (mods that tweak base stats) would be relatively easy to write some extra tests for.

AnnikaCodes:
Runtime error, the problem was that `Stats` wasn't defined in that file. We use Mocha for testing.

GregLyons:
I see. Which mods were breaking? Was it all the mods in which `Stats.statIDs` was used? Perhaps I'm misunderstanding how `namespace`s work in TS, but I'm not seeing any errors on my end, even when I run `npm run test` (I did before, but ran it again after pulling in most recent changes). Is there a way I could reproduce this error for further testing?

GregLyons:
🤦 [namespaces don't actually do anything in JS](https://stackoverflow.com/questions/60376241/typescript-namespace-not-resolved-during-runtime). Sorry for not realizing that. I'll open another PR soon with different fixes.

PR #8940, opened 2022-10-19
Improved typing in `utils.ts`

GregLyons:
This addresses #8919 and the subsequent  #8938. The issue was that I was trying to use a `namespace` to export arrays to be iterated over, but this won't work at runtime (the TS compiler won't catch this, see e.g. [this thread](https://stackoverflow.com/questions/60376241/typescript-namespace-not-resolved-during-runtime)).


data/mods/gen1/rulesets.ts
```diff
@@ -17,7 +17,8 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 			const bst = newSpecies.bst;
 			if (bst <= 350) {
 				newSpecies.bst = 0;
-				for (const stat in newSpecies.baseStats) {
+				let stat: StatID;
+				for (stat in newSpecies.baseStats) {
```
GregLyons, line 21: This tells `TS` that `stat` is a `StatID`, not merely `string`

GregLyons:
I don't see any tests for specific mods. If I wanted to write some tests for the stat mods that this code affects, where would be a good place to put them? In `test/sim/misc`?

If desired, I can try to add these tests onto this PR as well.

lib/utils.ts
```diff
@@ -318,12 +317,12 @@ export function clearRequireCache(options: {exclude?: string[]} = {}) {
 	}
 }
 
-export function deepClone(obj: any): any {
```
GregLyons, line 321: Perhaps `mutableDeepClone` would be a more precise name. I used the `Mutable` type since in virtually all of the cases I saw, this is used to make a copy of an object and then add/mutate a property.

Since the object being cloned is often marked as `readonly`, merely `T` wouldn't suffice, as then the `readonly` would be carried over to the clone, defeating the purpose of the clone.

DaWoblefet:
> I don't see any tests for specific mods. If I wanted to write some tests for the stat mods that this code affects, where would be a good place to put them? In `test/sim/misc`?
> 
> If desired, I can try to add these tests onto this PR as well.

Mods aren't good about having tests alongside their code (some are seasonal and get rotated in/out of the codebase which would cause test issues if the formats don't exist). I'd prefer them appropriately sequestered in tests as well, maybe `test/sim/mods`.

GregLyons:
Converting to draft for the time being while I write unit tests. I think I need some guidance on writing them.

1. Is adding the mod to `common.js` all I need to do so that I'm capable of testing the mod?
2. Is there a way for me to access the Pokemon's base stat post-mod directly? If not, is there an easy way for me to calculate what the stat after level/IV/EV AND mod should be?

test/sim/mods/350cup.js
```diff
@@ -0,0 +1,44 @@
...
+
+		const bulbasaur = battle.p1.active[0];
+
+		const atk = bulbasaur.getStat('atk');
+		const def = bulbasaur.getStat('def');
+		const spa = bulbasaur.getStat('spa');
+		const spd = bulbasaur.getStat('spd');
+		const spe = bulbasaur.getStat('spe');
+
+		assert.equal(atk, 98);
```
GregLyons, line 26: Not sure how to get a Pokemon's base stats post-mod

PR #8946, opened 2022-10-22
Random Doubles updates

ACakeWearingAHat:
-So many movesets have changed. In general, movesets have been trimmed to remove unnecessary coverage, unnecessary instances of Protect, and most instances of Ally Switch.
-The total bulk minimum for Assault Vest in Doubles has been lowered to 275, rather than 280. This allows Incineroar to have an Assault Vest, primarily.
-Pheromosa and frail Regenerator Pokemon will no longer get a Sitrus Berry in Doubles; in cases where they used to get a Sitrus Berry, they will now get a Life Orb.
-In Doubles, Protect will no longer be generated alongside reliable recovery moves.
-Muddy Water will not be on sets with Liquidation.
-Galarian Mr. Mime will no longer appear in Random Doubles.


AnnikaCodes:
thanks!

PR #8951, opened 2022-10-26
Fix events having access to past-gen exclusive moves

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9373659

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -1778,7 +1778,7 @@ export class TeamValidator {
 		if (ruleTable.has('obtainablemoves')) {
 			if (eventData.generation > setSources.maxSourceGen() && !(dex.gen <= 2 && eventSpecies.gen === 1)) {
 				if (fastReturn) return true;
-				problems.push(`${name} must not have moves only learnable in gen ${setSource.maxSourceGen()}${etc}.`);
+				problems.push(`${name} must not have moves only learnable in gen ${setSources.maxSourceGen()}${etc}.`);
 			}
 		}
 		if (ruleTable.has('obtainableabilities')) {
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -689,7 +689,9 @@ export class TeamValidator {
 						const eventInfo = event;
 						const eventNum = i + 1;
 						const eventName = eventData.length > 1 ? ` #${eventNum}` : ``;
-						const eventProblems = this.validateEvent(set, eventInfo, eventSpecies, setSources, ` to be`, `from its event${eventName}`);
+						const eventProblems = this.validateEvent(
+							set, eventInfo, eventSpecies, setSources, ` to be`, `from its event${eventName}`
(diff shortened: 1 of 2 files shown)


sim/team-validator.ts
```diff
@@ -1767,6 +1775,12 @@ export class TeamValidator {
 		}
 		// Event-related ability restrictions only matter if we care about illegal abilities
 		const ruleTable = this.ruleTable;
+		if (ruleTable.has('obtainablemoves')) {
+			if (eventData.generation > setSources.maxSourceGen() && !(dex.gen <= 2 && eventSpecies.gen === 1)) {
+				if (fastReturn) return true;
+				problems.push(`${name} must not have moves only learnable in gen ${setSource.maxSourceGen()}${etc}.`);
```
Karthik99999, line 1781: Would appreciate suggestions for this message

singiamtel: Maybe just ``${name} does not learn ${move.name} in gen ${dex.gen}${etc}.``

sim/team-validator.ts
```diff
@@ -1767,6 +1775,13 @@ export class TeamValidator {
 		}
 		// Event-related ability restrictions only matter if we care about illegal abilities
```
urkerab, line 1776: This comment is a bit confusing now that you've added a check for illegal moves.

sim/team-validator.ts
```diff
@@ -1767,6 +1775,13 @@ export class TeamValidator {
...
 		const ruleTable = this.ruleTable;
+		if (ruleTable.has('obtainablemoves')) {
+			const tradebackEligible = ruleTable.has('allowtradeback') || (dex.gen === 2 && species.gen === 1);
```
urkerab, line 1779: The other uses of `ruleTable.has('allowtradeback')` don't do this so I assume that there is a reason this is done differently here.

Karthik99999: My logic was that if tradebacks were allowed then this issue of not being able to use moves from a past gen wouldn't be a problem. Though I guess that's under the assumption that `allowtradeback` was also meant for you to be able to bring the mon back up to the current gen, which I'm not sure is the case.

sim/team-validator.ts
Karthik99999: is this comment even necessary?

Marty-D (approved):

sim/team-validator.ts
DaWoblefet: Yeah, you can take it out.

sim/team-validator.ts
DaWoblefet: I'm in agreement here with urkerab that it should be removed; that's more of a policy thing with allowtradeback in custom rules for modern gens.

DaWoblefet:
Thanks Karthik!

PR #8956, opened 2022-10-29
Fix Mimic accuracy in gens 1 and 2

livid-washed:
Mimic accuracy fixed, changed from can't miss to 100% for gens 1 and 2. 100% is the correct accuracy.

Reported as a bug (for gen 1): 
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-14#post-8558474

Mimic's accuracy: https://bulbapedia.bulbagarden.net/wiki/Mimic_(move)

### DIFF
--- a/data/mods/gen2/moves.ts
+++ b/data/mods/gen2/moves.ts
@@ -415,7 +415,6 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	},
 	mimic: {
 		inherit: true,
-		accuracy: 100,
 		ignoreAccuracy: undefined,
 		ignoreEvasion: undefined,
 		noSketch: true,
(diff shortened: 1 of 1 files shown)


data/mods/gen2/moves.ts
```diff
@@ -415,7 +415,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	},
 	mimic: {
 		inherit: true,
-		accuracy: true,
+		accuracy: 100,
```
DaWoblefet, line 418: You can just remove this altogether; it'll inherit the 100 from Gen 3.

livid-washed: Yep, thanks! I have removed it

DaWoblefet:
Thank you much!

PR #8957, opened 2022-10-29
Gen 8: Fix failure message not showing for Purify

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9376335

Specified gen 8 since the fail message already shows in gen 7 due to Purifiy's `onHit` event in that gen returning `false` instead of `NOT_FAIL`.


DaWoblefet:
Thanks Karthik!

PR #8960, opened 2022-10-30
Gen 1: Fix Thunder Wave not affecting Ground-types with Inverse Mod

livid-washed:
This fixes Ground-types not being affected by Thunder Wave in Gen 1 with Inverse Mod.

https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-15#post-8801689

Ground-types are still immune to Thunder Wave in regular battles, since Thunder Wave does not ignore type immunities.


DaWoblefet:
Thanks, looks good!

PR #8961, opened 2022-10-30
Gen 1: Fix Twineedle poison chance

livid-washed:
Twineedle should only be able to poison on its second hit in gen 1. Currently it can poison on both hits.

https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-15#post-8757544

The patch is done in an identical way to gen 2.


AnnikaCodes:
Thanks!

PR #8963, opened 2022-11-02
Gen 1: Fix Hyper Beam unbroken Substitute interaction

livid-washed:
This patch fixes the bug "Hyper Beam should only skip recharge turns when KOing a Substitute, not every time", listed here:

Previously, Hyper Beam would skip recharge turns when hitting a Substitute, regardless of whether the Substitute was KOd.

### DIFF
--- a/test/sim/moves/hyperbeam.js
+++ b/test/sim/moves/hyperbeam.js
@@ -42,13 +42,13 @@ describe(`Hyper Beam`, function () {
 		assert.false.cantMove(() => battle.choose('p1', 'move tackle'));
 	});
 
-	it.skip(`[Gen 1] should force a recharge turn after damaging, but not breaking a Substitute`, function () {
+	it(`[Gen 1] should force a recharge turn after damaging, but not breaking a Substitute`, function () {
 		battle = common.gen(1).createBattle([[
-			{species: 'chansey', moves: ['hyperbeam', 'tackle']},
+			{species: 'slowbro', moves: ['hyperbeam', 'tackle']},
 		], [
 			{species: 'rhydon', moves: ['substitute']},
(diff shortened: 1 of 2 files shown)


test/sim/moves/hyperbeam.js
```diff
@@ -42,13 +42,13 @@ describe(`Hyper Beam`, function () {
 		assert.false.cantMove(() => battle.choose('p1', 'move tackle'));
 	});
 
-	it.skip(`[Gen 1] should force a recharge turn after damaging, but not breaking a Substitute`, function () {
+	it(`[Gen 1] should force a recharge turn after damaging, but not breaking a Substitute`, function () {
 		battle = common.gen(1).createBattle([[
-			{species: 'chansey', moves: ['hyperbeam', 'tackle']},
+			{species: 'slowbro', moves: ['hyperbeam', 'tackle']},
```
AnnikaCodes, line 47: Why is this changed? Are we sure that Slowbro's Hyper Beam isn't just breaking the substitute?

livid-washed: Chansey is faster than Rhydon so the previous test was inappropriate. We want to test hitting the Substitute, not the Rhydon itself. Slowbro is slower, and also never breaks Rhydon's sub even with a max roll crit.

AnnikaCodes: Gotcha.

AnnikaCodes:
Thanks!

AnnikaCodes:
If you're looking to fix more Gen 1 Hyper Beam inaccuracies, there's also a bug where Hyper Beam never skips the recharge turn when it's called by another move.

PR #8966, opened 2022-11-04
Gen 1: Fix Mirror Move Hyper Beam interaction

livid-washed:
This patch fixes the interaction between Mirror Move and Hyper Beam. When Mirror Move calls Hyper Beam and KO's the enemy, the Mirror Move user should not need to recharge. 

https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-16#post-9004208

Partial trapping and charging moves called by Mirror Move are still messed up, but I think that's also an issue with the implementation of those moves themselves.

### DIFF
--- a/data/mods/gen1/scripts.ts
+++ b/data/mods/gen1/scripts.ts
@@ -67,8 +67,7 @@ export const Scripts: ModdedBattleScriptsData = {
 	},
 	actions: {
 		// This function is the main one when running a move.
-		// It deals with the beforeMove and AfterMoveSelf events.
-		// This leads with partial trapping moves shennanigans after the move has been used.
+		// It deals with the beforeMove event
 		// It also deals with how PP reduction works on gen 1.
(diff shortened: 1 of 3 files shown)


livid-washed:
Some explanation for the changes, feel free to ask any questions or point out any errors.

Previously: 

- runMove does "before move" events, calls useMove, then does "after move" events
- useMove uses the move

Now:

- runMove does "before move" events and calls useMove
- useMove calls useInnerMove, then does "after move" events, but only if it is the 'final' move
- useInnerMove uses the move (identical copy of previous useMove)

When Mirror Move (or Metronome) is successfully used, it calls useMove for the new selected move.

Previously, when using Mirror Move, the "after move" events are done with Mirror Move as the move. Now, the "after move" events are done with the new selected move. For example, if the chosen move is Hyper Beam and the opponent is KO'd, then the Mirror Move user will now have the recharge status removed.

AnnikaCodes:
Thanks!

PR #8968, opened 2022-11-04
Validator: Correctly enforce per-game evolution restrictions

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8811572
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-8833356

### DIFF
--- a/sim/dex-species.ts
+++ b/sim/dex-species.ts
@@ -136,6 +136,8 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 	declare readonly evoItem?: string;
 	/** Evolution move. falsy if doesn't evolve. */
 	readonly evoMove?: string;
+	/** Evolution region. falsy if doesn't evolve. */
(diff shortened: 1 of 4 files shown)


sim/dex-species.ts
```diff
@@ -136,6 +136,8 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 	declare readonly evoItem?: string;
 	/** Evolution move. falsy if doesn't evolve. */
 	readonly evoMove?: string;
+	/** Evolution region. falsy if doesn't evolve. */
```
DaWoblefet, line 139: I would make this clear that this is _exclusive_ evo region, i.e. it cannot evolve anywhere else, even if it might be obtainable elsewhere.

DaWoblefet:
Thanks Karthik!

PR #8969, opened 2022-11-05
Gen 1: Fix Counter interaction with Substitute

livid-washed:
This patch fixes Counter so that it can counter attacks that hit a substitute. Previously, Counter could not counter attacks that hit a substitute, even if they are 'counterable' in Gen 1. The damage that is countered is the 'uncapped' damage (for non-Stadium), which can be important if the attack breaks the substitute.

https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-16#post-8973042
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-30#post-8645108
Also fixes the same Counter/Substitute issue in Gen 1 Stadium, listed here:
https://www.smogon.com/forums/threads/stadium-bug-report-thread.3526616/page-3#post-8644615

(Does NOT fix the same issue with Bide: probably not working on that move for a while)

### DIFF
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -835,6 +835,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 				uncappedDamage = this.runEvent('SubDamage', target, source, move, uncappedDamage);
 				if (!uncappedDamage) return uncappedDamage;
 				source.lastDamage = uncappedDamage;
+				this.lastDamage = uncappedDamage;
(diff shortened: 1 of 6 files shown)


AnnikaCodes (approved):
This looks good to me!

Also, what is your username on Showdown?

livid-washed:
Thanks!

My username is **livid washed**.

AnnikaCodes:
Thanks!

PR #8970, opened 2022-11-06
Battle Factory: Raise max generation attempts

MathyFurret:
Discussed with @ACakeWearingAHat 

This raises the number of attempts taken by the Battle Factory algorithm to generate a team with the desired parameters, such as having exactly 1 Stealth Rocker and having exactly one hazard remover.

This should be considered only a temporary migitation; a better solution for the future would be to rewrite the generator around guaranteeing teammates with certain features

This won't badly tank performance; time and space complexity are both O(n) with the number of attempts, which is distributed geometrically

### DIFF
--- a/data/mods/gen6/random-teams.ts
+++ b/data/mods/gen6/random-teams.ts
@@ -1202,7 +1202,7 @@ export class RandomGen6Teams extends RandomGen7Teams {
 	randomFactoryTeam(side: PlayerOptions, depth = 0): RandomTeamsTypes.RandomFactorySet[] {
 		this.enforceNoDirectCustomBanlistChanges();
 
-		const forceResult = (depth >= 4);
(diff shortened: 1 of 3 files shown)


data/mods/gen6/random-teams.ts
```diff
@@ -1202,7 +1202,7 @@ export class RandomGen6Teams extends RandomGen7Teams {
 	randomFactoryTeam(side: PlayerOptions, depth = 0): RandomTeamsTypes.RandomFactorySet[] {
 		this.enforceNoDirectCustomBanlistChanges();
 
-		const forceResult = (depth >= 4);
+		const forceResult = (depth >= 12);
```
AnnikaCodes, line 1205: Maybe in the future this should be a constant somewhere?

AnnikaCodes:
Thanks!

PR #8974, opened 2022-11-07
Gen 1: Fix simultaneous Counters so that they both fail

livid-washed:
This patch fixes Counter so that it will always fail when the opponent's last selected move is Counter, including when both Counters are selected on the same turn.

Mentioned here and confirmed in many other places:


AnnikaCodes:
Thanks!

PR #8976, opened 2022-11-08
Fix: Trigger emergency exit after Mind Blown

singiamtel:
As suggested here: [Various Emergency Exit bugs](https://github.com/smogon/pokemon-showdown/projects/3#card-85841440)

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -11162,6 +11162,12 @@ export const Moves: {[moveid: string]: MoveData} = {
 				this.damage(Math.round(pokemon.maxhp / 2), pokemon, pokemon, this.dex.conditions.get('Mind Blown'), true);
 			}
 		},
+		onAfterMoveSecondarySelf(pokemon, target, move) {
+			const maxhp = pokemon.getUndynamaxedHP(pokemon.maxhp);
+			if (pokemon.hp && pokemon.ability === "emergencyexit" && pokemon.getUndynamaxedHP() <= maxhp / 2) {
+				this.runEvent('EmergencyExit', pokemon);
+			}
+		},
 		secondary: null,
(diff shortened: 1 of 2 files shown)


data/moves.ts
```diff
@@ -11162,6 +11162,12 @@ export const Moves: {[moveid: string]: MoveData} = {
 				this.damage(Math.round(pokemon.maxhp / 2), pokemon, pokemon, this.dex.conditions.get('Mind Blown'), true);
```
singiamtel, line 11162: Could also move this logic right after the recoil damage in here, not sure if this is the proper use for ``onAfterMoveSecondarySelf``

Karthik99999: Probably better to do that. Also while in practice a pokemon with under half hp would just get KO'd by the recoil, I still think you should check the targets hp before dealing the damage, then checking to see if the user went from above half to below half to keep it consistent with the rest of the emergency exit checks. Here's an example for recoil moves done in `battle-actions.ts`

```js
if (move.recoil && move.totalDamage) {
	const hpBeforeRecoil = pokemon.hp;
	this.battle.damage(this.calcRecoilDamage(move.totalDamage, move), pokemon, pokemon, 'recoil');
	if (pokemon.hp <= pokemon.maxhp / 2 && hpBeforeRecoil > pokemon.maxhp / 2) {
		this.battle.runEvent('EmergencyExit', pokemon, pokemon);
	}
}
```

singiamtel:
How should I proceed here? I'd like to try to fix some other Emergency Exit bugs, is it fine if we edit this PR to allow them?

DaWoblefet:
Thanks, singiamtel!

PR #8977, opened 2022-11-10
Add more debug messages for moves with custom base power

DaWoblefet:
Adds base power debugging messages (see [Discord discussion](https://discord.com/channels/630837856075513856/630845310033330206/1031028616458407976)). All the moves with custom base power should be covered now (save ones that are immediately obvious, e.g. Pledge moves or Magnitude).

In addition, this fixes Flail / Reversal's ratio calculation in Gen 5+. @Marty-D can you verify if they floor or not in Gen 4?

I've also removed an unnecessary piece of inheritance for Hidden Power in Gen III.

### DIFF
--- a/data/mods/gen2/moves.ts
+++ b/data/mods/gen2/moves.ts
@@ -302,9 +302,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	frustration: {
 		inherit: true,
 		basePowerCallback(pokemon) {
-			const bp = Math.floor(((255 - pokemon.happiness) * 10) / 25) || null;
-			this.debug('BP: ' + bp);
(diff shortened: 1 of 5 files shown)


DaWoblefet:
Weird, these build failures are because tests that have Frustration and Return are bombing out because of the new debug line.

DaWoblefet:
> moveName += ' ' + basePowerCallback.call(this.battle, this);

Tried this and it still crashed. I'm just going to not include debug prints for Return/Frustration; they don't vary during the middle of battle anyway.

DaWoblefet:
The old UPC stuff says Flail/Reversal is also floored, so I'm proceeding forward.

PR #8978, opened 2022-11-10
Add Ability Clause

HisuianZoroark:
Requested by UT

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -591,6 +591,48 @@ export const Rulesets: {[k: string]: FormatData} = {
 			}
 		},
 	},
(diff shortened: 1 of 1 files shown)


KrisXV:
Rename to ability clause pls

DaWoblefet:
Most of this code is shared with ``2abilityclause``. I wonder if there's a way to make this more generic in case we ever want ``3abilityclause`` etc.

KrisXV:
Yes, we can change it into a value rule easily

KrisXV:
I'll work on turning into a value rule in a bit, its better to get this in before gen 9 starts

PR #8981, opened 2022-11-12
Mafia: Add /mafia resetgame command and some other minor stuff.

motogp-bit:
/mafia resetgame resets the game to night 0,removes all action data from the html tab,redistributes the roles and clears all deadlines.Players who died or were kicked are not added,players who subbed in or were added are not removed and no other setting set by the host is changed. (Copied some of the night commands because the other option included a lot of spaghetti).

Return to day X button removed when x = 0 (it couldn't possibly work in any way and would always throw an error message).

Nightstart is now default,/mafia daystart can be used instead to start at day.Actions from night 0 are not tracked.


Alexander489:
staff approved

Issue #8955, opened 2022-10-28
Swallow healing incorrect


davidstone:
The following battle log

```
<< >battle-gen3ou-1697426101
|player|p2|david stone|169|
|teamsize|p1|1
|teamsize|p2|3
|gen|3
|tier|[Gen 3] OU
|rule|Sleep Clause Mod: Limit one foe put to sleep
|rule|Switch Priority Clause Mod: Faster Pokémon switch first
|rule|Species Clause: Limit one of each Pokémon
|rule|OHKO Clause: OHKO moves are banned
|rule|Evasion Items Clause: Evasion items are banned
|rule|Evasion Moves Clause: Evasion moves are banned
|rule|Endless Battle Clause: Forcing endless battles is banned
|rule|HP Percentage Mod: HP is shown in percentages
|rule|One Boost Passer Clause: Limit one Baton Passer that has a way to boost its stats
|rule|Freeze Clause Mod: Limit one foe frozen
|
|t:|1666993134
|start
|switch|p1a: Smeargle|Smeargle, M|251/251
|switch|p2a: Smeargle|Smeargle, L40, F|100/100
|turn|1
|request|{"active":[{"moves":[{"move":"Stockpile","id":"stockpile","pp":16,"maxpp":16,"target":"self","disabled":false},{"move":"Swallow","id":"swallow","pp":16,"maxpp":16,"target":"self","disabled":false},{"move":"Seismic Toss","id":"seismictoss","pp":31,"maxpp":32,"target":"normal","disabled":false}]}],"side":{"name":"obi","id":"p1","pokemon":[{"ident":"p1: Smeargle","details":"Smeargle, M","condition":"1/251","active":true,"stats":{"atk":45,"def":106,"spa":76,"spd":126,"spe":186},"moves":["stockpile","swallow","seismictoss"],"baseAbility":"owntempo","item":"","pokeball":"pokeball"}]},"rqid":4}
|
|t:|1666993142
|move|p1a: Smeargle|Seismic Toss|p2a: Smeargle
|-damage|p2a: Smeargle|1/100
|move|p2a: Smeargle|Endeavor|p1a: Smeargle
|-damage|p1a: Smeargle|1/251
|
|upkeep
|turn|2
[...]

PR #8983, opened 2022-11-13
Fix: round down swallow healing

cooshaug:
Fixing bug described in #8955. Swallow healing should round down, not up. 

Before, the rounding relied on [battle.modify](https://github.com/smogon/pokemon-showdown/blob/master/sim/battle.ts#L2067), which would sometimes round up. For this reason, I'm don't think it's suited for healing ([here](https://github.com/smogon/pokemon-showdown/blob/master/data/moves.ts#L11553), for example). Otherwise, the method is used for stat changes, like Attack.

### DIFF
--- a/test/sim/moves/swallow.js
+++ b/test/sim/moves/swallow.js
@@ -1,0 +1,60 @@
+"use strict";
+
+const assert = require("./../../assert");
+const common = require("./../../common");
+
+let battle;
+
+describe("Swallow", function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it("should heal 1/4 health (rounded down) after 1 stockpile", function () {
+		battle = common.createBattle([
+			[
+				{
+					species: "Seviper",
+					ability: "shedskin",
+					moves: ["stockpile", "swallow"],
+				},
+			],
+			[{species: "Zangoose", ability: "immunity", moves: ["sleeptalk", "falseswipe"]}],
+		]);
+
+		battle.makeChoices("move stockpile", "move falseswipe");
+		const hpBeforeHeal = battle.p1.active[0].hp;
+		battle.makeChoices("move swallow", "move sleeptalk");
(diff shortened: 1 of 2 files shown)


test/sim/moves/swallow.js
```diff
@@ -0,0 +1,60 @@
...
+const common = require("./../../common");
+
+let battle;
+
+describe("Swallow", function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it("should heal 1/4 health (rounded down) after 1 stockpile", function () {
```
DaWoblefet, line 13: The formatting of tests here needs reworked; try to model it after tests from around that area (Sparkling Aria is a good example).

test/sim/moves/swallow.js
```diff
@@ -0,0 +1,60 @@
...
+				{
+					species: "Seviper",
+					ability: "shedskin",
+					moves: ["stockpile", "swallow"],
+				},
+			],
+			[{species: "Zangoose", ability: "immunity", moves: ["sleeptalk", "falseswipe"]}],
+		]);
+
+		battle.makeChoices("move stockpile", "move falseswipe");
```
DaWoblefet, line 25: You can use something like Seismic Toss instead so HP is fixed throughout the whole process.

DaWoblefet:
Actually, never mind all that - the user posting the issue was mistaken. ``this.modify`` is the correct function to call here. In the modern games at least, Swallow uses half down rounding, i.e. 0.5 or smaller rounds down, anything else rounds up. It's the same style that Mega Launcher Heal Pulse, Morning Sun / Synthesis / Moonlight, Shore Up, and Floral Healing in Grassy Terrain use. So in the user's case, they were simply mistaken. A Pokemon with 251 HP should in fact recover 63 HP from a 1 Stockpile Swallow, not 62.

PR #8985, opened 2022-11-14
/learn: Add hidden ability checks

dot-Comfey:
It's like 5 days until the turn of the next generation, and this is such a niche feature of a command, but whatever, I decided to work on this anyway. This changes the behavior of /learn when checking the validity of a Pokemon with a combination of a given move and a hidden ability. The following changes were made:

- Method validateEventHiddenAbility was added as a split from code originally in validateEvent, and is called by its original method and the method for /learn. It returns null if the ability being hidden or not is compatible with a given event, otherwise it returns the reason for incompatibility.
- If a Pokemon is marked as having its hidden ability, its sources should not be restricted if the Ability Patch can be used. This shouldn't change any Pokemon legality, but should prevent /learn from returning no sources for a move that is legitimately learnable.
- If a Pokemon does not have a set sourcesBefore, /learn will go through all of its possible event sources and check if it is compatible with a hidden ability. Events that do not have a hidden ability will be removed from sources, and if this results in a Pokemon having no sources, the command returns that the move can't be learned.
- If a Pokemon is marked as a species that can only be male with a hidden ability, Gen 5 egg sources will be removed (this is already detected by /learn in its current state, with a visual indicator of no hidden ability). If this results in a Pokemon having no sources, the command returns that the move can't be learned.
- A purely aesthetic change I added is that /learn will fail if used with a hidden ability in a Gen before Gen 5. I made this change because it will always return that the move can't be learned, and hidden ability slots, which can have significance in Gens with hidden abilities, are of no relevance in other Gens.

These changes deal with the bug reported in this post: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9361247

These changes only attempt to check if a combination of a hidden ability and a move or moves are compatible; it does not check if a hidden ability alone is legal. Also when making these changes, I realized that there might be a possible legality error with Gen 5 male only hidden abilities: in Gen 6, a Tyrogue shouldn't be able to have its hidden ability Vital Spirit and know Bullet Punch and Work Up. This is because Bullet Punch is a move Tyrogue learns only through breeding in Gens 4-6, and because it has its hidden ability, it can't come from Gen 4, which was before hidden abilities, and it can't come from Gen 5 because Tyrogue is a male only species, so it is also a male only hidden ability species and shouldn't have egg moves from Gen 5. So this Tyrogue must come from Gen 6, but it knows Work Up, a move which it can only learn in Gen 5. If this is indeed not supposed to be legal, I would have caught this in my changes, but for now, I would prefer for my changes to be consistent with the team validator.

### DIFF
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -2469,8 +2469,9 @@ function runLearn(target: string, cmd: string, canAll: boolean, formatid: string
 				const source = setSources.sources[sourceIndex];
 				if (source.charAt(1) === 'S' && learnset.eventData) {
 					const eventData = learnset.eventData[parseInt(source.substr(2))];
-					const hiddenAbilityProblem = validator.validateEventHiddenAbility(species.name, true, eventData, 
-						` because it has a move only available from an event`);
+					const hiddenAbilityProblem = validator.validateEventHiddenAbility(
+						species.name, true, eventData, ` because it has a move only available from an event`
+					);
 					if (hiddenAbilityProblem) {
 						setSources.sources.splice(sourceIndex, 1);
 						sourceIndex--;
--- a/server/chat-plugins/datasearch.ts
+++ b/server/chat-plugins/datasearch.ts
@@ -8,2635 +8,2634 @@
  * @license MIT
  */
 
- import {ProcessManager, Utils} from '../../lib';
- import {TeamValidator} from '../../sim/team-validator';
- import {Chat} from '../chat';
- 
- interface DexOrGroup {
-	 abilities: {[k: string]: boolean};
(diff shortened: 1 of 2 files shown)


KrisXV:
hey @dot-Comfey sorry for the long delay on this, could you fix the merge conflict? ill take a look at this once that's done.

dot-Comfey:
I don't like this code (it was written 8 months ago and I know how to write better code) but whatever.

KrisXV:
If you'd like to rewrite it then I will leave it open for you do do that

dot-Comfey:
There are still some inconsistencies between the validator and /learn in regards to move legality for other specific sets, but I feel my code now satisfies what I was going for in this PR.

PR #8986, opened 2022-11-14
Gen8 Random Battles: Limit the amount of pokemon weak to any type to three

adrivrie:
Keeps a counter with weaknesses and rejects mons when their typing would put one of them above 3 (or above half the team's size for larger teams).

This was attempted earlier and abandoned. I found the bug in the original implementation and fixed it.

https://discord.com/channels/294651453279174656/809406358654681168/1041802698112122951

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -2405,6 +2405,7 @@ export class RandomTeams {
 		const tierCount: {[k: string]: number} = {};
 		const typeCount: {[k: string]: number} = {};
 		const typeComboCount: {[k: string]: number} = {};
+		const typeWeaknesses: {[k: string]: number} = {};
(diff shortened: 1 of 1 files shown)


adrivrie:
> What does this change from the previous attempt?

Should probably have mentioned this immediately, sorry:

The previous attempt [lines 2465-2474 here](https://github.com/smogon/pokemon-showdown/commit/52bf7847042335d1a6eb71fae2e6a2ffbb38cdfe) already incremented the weakness counters before it was certain a species would not be rejected. Thus, the weakness counter increased even for some rejected species leading to cascading failures resulting finally in team generation (for certain seeds) to fail entirely. 

This commit increments the counters only when a mon is added to the team.

AnnikaCodes:
Thanks!

PR #8987, opened 2022-11-15
Add Gen 9 Partners in Crime

HisuianZoroark:
Does what is outlined [here](https://www.smogon.com/forums/threads/partners-in-crime.3706080/page-6#post-9392994)

Also adds some support for items that were forgotten about in the original implementation, and adds some support for moves/Leppa Berry with the new PP mechanic in place.


HisuianZoroark:
After some discussion, this will now house the Gen 9 Partners in Crime format on top of the mechanics already introduced in this Pull Request.

PR #8988, opened 2022-11-15
Adding Rollout storage glitch

cooshaug:
This is ticketed [here](https://github.com/smogon/pokemon-showdown/projects/3#card-61615000)

Want to practise some web-dev. Hence, why I'm picking up a slightly obscure ticket. 

Some notes:
* I'm not sure I've correctly used (or abused) some of the event hooks - for example, `onAfterMove` for checking if we need to store the multiplier for the next non-rollout move. 
* The ticket also specifies that this bug was present pre-patch in Gen 8, though I'm not sure how to check for this (some `gen` attributes need to be ints, so "8 DLC1" wouldn't work).

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -765,7 +765,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 				effect && effect.effectType === 'Move' &&
 				['mimikyu', 'mimikyutotem'].includes(target.species.id) && !target.transformed
 			) {
-				this.add('-activate', target, 'ability: Disguise');
+				if (this.gen === 7 && ["rollout", "iceball"].includes(effect.id)) {
+					source.volatiles[effect.id].contactHitCount--;
+				}
+
+				this.add("-activate", target, "ability: Disguise");
 				this.effectState.busted = true;
 				return 0;
 			}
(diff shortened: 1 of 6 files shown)


data/abilities.ts
```diff
@@ -765,7 +765,11 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 				effect && effect.effectType === 'Move' &&
 				['mimikyu', 'mimikyutotem'].includes(target.species.id) && !target.transformed
 			) {
-				this.add('-activate', target, 'ability: Disguise');
+				if (this.gen === 7 && ["rollout", "iceball"].includes(effect.id)) {
```
DaWoblefet, line 768: This behavior should be inside the mod/ directory for ``gen8dlc1`` and ``gen7``.

test/sim/moves/rollout.js
```diff
@@ -130,20 +130,25 @@ for (const move of moves) {
...
 
 				for (let i = 0; i < 5; i++) { battle.makeChoices(); }
 				battle.makeChoices('move watergun', 'switch 2');
-				const snorlax = battle.p2.active[0];
-				const damage = snorlax.maxhp - snorlax.hp;
+				let snorlax = battle.p2.active[0];
+				let damage = snorlax.maxhp - snorlax.hp;
 				assert.bounded(damage, [147, 174]); // 40 * 2^4 BP; would be 40 BP otherwise, range 10-12
+				battle.makeChoices("move watergun", "switch 3");
+				snorlax = battle.p2.active[0];
```
DaWoblefet, line 149: I would recommend switching to a different Pokemon, just so it's very clear it's a different Snorlax.

DaWoblefet:
I really like this implementation! Thanks a bunch.

PR #8989, opened 2022-11-15
Calc damage for Metronome(item) properly when using Copycat/Metronome(move)

singiamtel:


MathyFurret:
While this technically fixes this niche bug it introduces another niche bug: Failed Copycat won't reset the Metronome counter. Also AFAIK this behavior goes back further than Gen 8, meaning there are more move-calling moves to account for.

singiamtel:
> While this technically fixes this niche bug it introduces another niche bug: Failed Copycat won't reset the Metronome counter.


That is true. Do you have any suggestion, or should I just add a check on Copycat/Metronome?

> Also AFAIK this behavior goes back further than Gen 8, meaning there are more move-calling moves to account for.


What other moves should I account for?

MathyFurret:
https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_call_other_moves but I don't think a hardcoded list is the right way to go about this. At this point in the check we don't yet know if these moves will have successfully called another move.

I looked at this one day but couldn't think of an elegant solution. It probably involves delaying resetting the counter/last move until later (onAfterMove?)

singiamtel:
>I looked at this one day but couldn't think of an elegant solution. It probably involves delaying resetting the counter/last move until later (onAfterMove?)

Yeah, you're probably right. I'm not sure how to do it without powering up the next different move too, I'll try to implement that in the evening. I'll mark it as draft until it's fixed.

PR #8991, opened 2022-11-17
Gen 1: Fix Rage not being triggered by missed Disable/Explosion

livid-washed:
This patch fixes the Rage condition so that the rage boost occurs when targeted by Disable or Selfdestruct/Explosion, even if those moves miss or fail. Adds some corresponding tests as well.

Disable bug reported here: https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-14#post-8503476

Explosion/Selfdestruct: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9399712


AnnikaCodes:
Thanks for all these mechanics fixes!

PR #8997, opened 2022-11-18
Add Gen 9

KrisXV:
Thanks to @Marty-D @HoeenCoder @DaWoblefet @EvezaGit @Karthik99999 for helping out


Zarel:
Hehe:
image

(There aren't any other traces of Quagsire-Paldea left, but apparently someone forgot its entry in `data/text` which I found while doing translations.)

PR #9004, opened 2022-11-18
Only change category of Tera Blast if user is Terastallized

Karthik99999:


KrisXV:
Can u edit the description too pls

PR #9008, opened 2022-11-18
Fix Charcadet evoConditions

Glazelf:
Current evoConditions imply that the Malicious and Auspicious Armors need to be used within Zapapico.
This is not the case in my own experience and I can't find this info reflected anywhere else either.


HoeenCoder:
I'm going to ask @Marty-D to take a look at one.

PR #9009, opened 2022-11-18
Unit Tests: Transition to Gen 9

MathyFurret:
* Default gen is now Gen 9
* Tests specifically for Dynamax and other legacy mechanics were kept Gen 8
* FFA/Multi will continue to target Gen 8
* A couple tests I wasn't sure how to transition so I left a TODO notice

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -23,7 +23,7 @@ class TestTools {
 		this.currentMod = mod;
 		this.dex = Dex.mod(mod);
 
-		this.modPrefix = this.dex.isBase ? `[gen8] ` : `[${mod}] `;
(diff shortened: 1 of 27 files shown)


AnnikaCodes:
Thanks!

OpacousPuppet:
When you say that FFA will continue to target Gen 8, could you be more specific? Do you mean that it will continue to target Gen 8 in the meantime until developers can code Gen 9 FFA?

MathyFurret:
> When you say that FFA will continue to target Gen 8, could you be more specific? Do you mean that it will continue to target Gen 8 in the meantime until developers can code Gen 9 FFA?

Yes the tests should be updated when gen 9 ffa is added

PR #9013, opened 2022-11-19
Fix various forme and move legality

KrisXV:
- Kanto Tauros should be unreleased
- Johto Wooper/Quagsire should be unreleased
- All non-Fancy Vivillon are unreleased
- Pokeball Vivillon is dexited
- Spewpa evolves into Fancy Vivillon (It's Fancy Spewpa but thats another issue)
- Cramorant is dexited
- Genesect formes should be tiered together in Gen 8
- Proton Geyser and Mind Blown and Take Heart are dexited


AnnikaCodes:
Thanks!

PR #9015, opened 2022-11-19
Gen 1: Fix Wrap when hitting a substitute

livid-washed:
This patch fixes the behavior of Wrap when it hits a substitute. Previously, a faster target with substitute up would be able to attack every second turn when being wrapped, and there were some other inaccurate behaviors as well.

https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-11#post-6893249
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-14#post-8442413
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-15#post-8622471
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-16#post-8999638

scripts.ts: By resetting the duration of the target's partial trapping every time Wrap hits the target's substitute, we ensure that the target will remain partially trapped until Wrap ends. Previously, a faster target would be able to attack every second turn as the duration decreased from 2 to 0.

moves.ts: Wrap hitting a Ghost type's substitute will be treated as a hit, resulting in the same partial trapping as with hitting a non-Ghost type.

### DIFF
--- a/data/mods/gen1/moves.ts
+++ b/data/mods/gen1/moves.ts
@@ -827,9 +827,9 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 				// Substitute, here we deliberately use the uncapped damage when tracking lastDamage etc.
 				// Also, multi-hit moves must always deal the same damage as the first hit for any subsequent hits
 				let uncappedDamage = move.hit > 1 ? source.lastDamage : this.actions.getDamage(source, target, move);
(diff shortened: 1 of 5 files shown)


AnnikaCodes:
Thanks

PR #9018, opened 2022-11-20
Clean Toxic Debris desc

Glazelf:
Specify physical attacks, special attacks don't trigger Toxic Debris. Moves do not need to make contact.  
It functions correctly but the description is vague and might confuse people.


KrisXV:
Thank you!

PR #9019, opened 2022-11-20
Gen 1: Sonic Boom can hit Ghost-types

livid-washed:
In Gen 1, Sonic Boom can hit Ghost-types, which is not currently implemented. Mentioned here and in many other places, and confirmed with in-game testing:

https://www.smogon.com/forums/threads/mechanics-differences-between-generations.3668073/post-8553244


AnnikaCodes:
Thanks!

PR #9022, opened 2022-11-20
Datasearch: Don't hardcode max gen numbers

MathyFurret:
This fixes gen 9 not working as a param in some commands, and prevents the same from happening in the future


KrisXV:
ty!

PR #9032, opened 2022-11-22
Gen 1 Stadium: fix battle crashing when Wrap KOs an enemy

livid-washed:
This patch fixes an omission with a previous Wrap-related fix that could cause Stadium battles to crash when a wrapping move KOs an enemy.

Previous change: https://github.com/smogon/pokemon-showdown/pull/9007

Crash: https://replay.pokemonshowdown.com/gen1stadiumou-1717693225-ea0vym1qnuplwtvfsf9780bo9pcg4uipw


AnnikaCodes:
Thanks!

PR #8929, opened 2022-10-11
Hangman: Prevent impossible guesses and fix HTML escape

DaWoblefet:
Fixes 3 issues with hangman:

1. When a letter is already successfully guessed, other impossible guesses can be submitted that don't take into account the current letters revealed.
<details>
<summary>Before</summary>

</details>
<details>
<summary>After</summary>

</details>

2. Words with HTML will break formatting, typically hiding or otherwise obscuring the message. Here's an example with `</div>`.
<details>
<summary>Before</summary>

</details>
<details>
<summary>After</summary>

</details>

3. Hint length just said "too long" and didn't clarify the number of characters. I refactored all of the length requirements to be constants at the top of the file.

PR #9033, opened 2022-11-22
Hangman: Prevent impossible guesses, escape HTML

MathyFurret:
Fixed version of #8929. This escapes the HTML in the secret word *before* the styling for unrevealed letters is added


AnnikaCodes (approved):
Looks good! Glad to see there are tests.

PR #9034, opened 2022-11-22
Gen 1: Fix Explosion not fainting user when target is semi-invulnerable

livid-washed:
This patch fixes Selfdestruct/Explosion when used against semi-invulnerable (Dig/Fly) enemies. Using Explosion should faint the user, but currently on Showdown it doesn't.

Also, adding a hint for the Disable/Explosion-Rage interaction.


AnnikaCodes:
Thanks!

PR #9035, opened 2022-11-22
Implement Revival Blessing

HisuianZoroark:
Client code does not exist yet, just wanted to get the server side stuff out at the moment.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2614,11 +2614,16 @@ export class Battle {
 		);
 
 		for (let i = 0; i < this.sides.length; i++) {
-			if (switches[i] && !this.canSwitch(this.sides[i])) {
+      let reviveSwitch = false; // Used to ignore the fake switch for Revival Blessing
+			if (switches[i] && !this.canSwitch(this.sides[i]) && !reviveSwitch) {
 				for (const pokemon of this.sides[i].active) {
-					pokemon.switchFlag = false;
+          if (this.sides[i].slotConditions[pokemon.position]['revivalblessing']) {
+            reviveSwitch = true;
+            continue;
+          }
+					if (!reviveSwitch) pokemon.switchFlag = false;
 				}
-				switches[i] = false;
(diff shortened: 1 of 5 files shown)


HisuianZoroark:
I think I can safely say that this and the accompanying client pull request are finished? I'll send in a demo video to showcase the move when I have more spare time.

sim/battle.ts
```diff
@@ -2614,11 +2614,16 @@ export class Battle {
 		);
 
 		for (let i = 0; i < this.sides.length; i++) {
-			if (switches[i] && !this.canSwitch(this.sides[i])) {
+			let reviveSwitch = false; // Used to ignore the fake switch for Revival Blessing
+			if (switches[i] && !this.canSwitch(this.sides[i]) && !reviveSwitch) {
```
HoeenCoder, line 2618: `&& !reviveSwitch` is redudant because you define that as `false` on the previous line.

HoeenCoder (commented):
Excellent work!

sim/battle.ts
HisuianZoroark: Yeah that's valid

Karthik99999:
Can you add
```js
heal: "  [POKEMON] was revived and is ready to fight again!",
```
to revival blessing in `data/text/moves.ts`? No extra client changes are needed

data/text/moves.ts
```diff
@@ -5080,6 +5080,7 @@ export const MovesText: {[k: string]: MoveText} = {
 	revivalblessing: {
 		name: "Revival Blessing",
 		shortDesc: "Revives a fainted Pokemon to 50% HP.",
+		heal: "  [POKEMON] was revived and is ready to fight again!",
```
Karthik99999, line 5083:
```suggestion

		heal: "  [POKEMON] was revived and is ready to fight again!",
```
very minor nitpick of adding an extra line to match formatting of everything else in the text files

PR #9037, opened 2022-11-22
Fix Rage Fist bugs

cooshaug:
[Bug Report](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9407560)

Rage Fist:
*  Status moves now do not increase Rage Fist BP
*  Capping BP at 350
* Should now work with transform (`timesAttacked` should be inherited)

### DIFF
--- a/test/sim/moves/ragefist.js
+++ b/test/sim/moves/ragefist.js
@@ -33,4 +33,23 @@ describe('Rage Fist', function () {
 		rageFistDamage = battle.p2.active[0].maxhp - rageFistDamage - battle.p2.active[0].hp;
 		assert.bounded(rageFistDamage, [17, 21]);
 	});
+
+	it('should increase BP after each hit of multi-hit moves', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'vitalspirit', moves: ['sleeptalk', 'ragefist']}]});
+		battle.setPlayer('p2', {team: [{species: 'Umbreon', ability: 'shellarmor', moves: ['bulletseed', 'sleeptalk']}]});
+		battle.makeChoices();
+		const rageFistMutliplier = 1 * battle.log.filter(s => s.includes('hitcount'))[0].slice(-1) + 1;
+		battle.makeChoices('move ragefist', 'move sleeptalk');
+		const rageFistDamage = battle.p2.active[0].maxhp - battle.p2.active[0].hp;
+		assert.bounded(rageFistDamage, [17 * rageFistMutliplier, 21 * rageFistMutliplier]); // damage range here isn't exact, but it's enough for this test
+	});
+
+	it.skip('should, when copied with Copycat, inherit the number of times hit', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'vitalspirit', moves: ['ragefist']}]});
+		battle.setPlayer('p2', {team: [{species: 'Umbreon', ability: 'shellarmor', moves: ['copycat']}]});
+		battle.makeChoices();
+		assert.bounded(battle.p1.active[0].maxhp - battle.p1.active[0].hp, [39, 46]);
+	});
 });
--- a/test/sim/moves/ragefist.js
+++ b/test/sim/moves/ragefist.js
@@ -45,11 +45,11 @@ describe('Rage Fist', function () {
 		assert.bounded(rageFistDamage, [17 * rageFistMutliplier, 21 * rageFistMutliplier]); // damage range here isn't exact, but it's enough for this test
 	});
 
-	it.skip('should, when copied with Copycat, inherit the number of times hit', function () {
+	it('should, when copied with Copycat, inherit the number of times hit', function () {
 		battle = common.createBattle();
-		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'vitalspirit', moves: ['ragefist']}]});
-		battle.setPlayer('p2', {team: [{species: 'Umbreon', ability: 'shellarmor', moves: ['copycat']}]});
+		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'shellarmor', moves: ['ragefist']}]});
+		battle.setPlayer('p2', {team: [{species: 'Umbreon', moves: ['copycat']}]});
 		battle.makeChoices();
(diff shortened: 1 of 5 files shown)


DaWoblefet:
> Hitting yourself in confusion increases Rage Fist BP

Can you clarify the cart footage / research where this is documented?

cooshaug:
> > Hitting yourself in confusion increases Rage Fist BP
> 
> Can you clarify the cart footage / research where this is documented? Even if it does work that way, I don't think a separate`timesHitItselfInConfusion` variable adds much value.

* Rage Fist's [wiki page](https://bulbapedia.bulbagarden.net/wiki/Rage_Fist_(move)) says that hitting self in confusion counts as a "hit".
* `timesHitItselfInConfusion` may not add much value on its own, but including self-confusion hits in `timesAttacked` would be very confusing for future uses. If we want to combine them, I suggest changing the name to `timesHit` - wdyt?

HoeenCoder:
> * `timesHitItselfInConfusion` may not add much value on its own, but including self-confusion hits in `timesAttacked` would be very confusing for future uses. If we want to combine them, I suggest changing the name to `timesHit` - wdyt?

Sounds good to me!

DaWoblefet:
Bulbapedia was mistaken. Self-hit confusion doesn't count at all ([see research](https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/post-9409510)).

If you could in this PR, fixing Rage Fist to cap at 350 BP should be fairly straightforward.

cooshaug:
For the copycat case, I'm struggling to find an elegant to implement it. I think a `sourcePokemon` attr to the battle's `lastMove` attr would really help, e.g.:

```	
ragefist: {
		basePowerCallback(pokemon, target, move) {
			let multiplier = pokemon.timesAttacked;
			if (move.sourceEffect === 'copycat') {
				multiplier = this.lastMove.sourcePokemon.timesAttacked;
			}
			return Math.min(350, 50 + 50 * multiplier);
		}
```

DaWoblefet:
> For the copycat case, I'm struggling to find an elegant way to implement it. I think a `sourcePokemon` attr to the battle's `lastMove` attr would really help, e.g.:
> 
> ```
> ragefist: {
> 		basePowerCallback(pokemon, target, move) {
> 			let multiplier = pokemon.timesAttacked;
> 			if (move.sourceEffect === 'copycat') {
> 				multiplier = this.lastMove.sourcePokemon.timesAttacked;
> 			}
> 			return Math.min(350, 50 + 50 * multiplier);
> 		}
> ```
> 
> Another workaround might be to backtrack through the log.
> 
> If ye can think of a way to fetch this info, let me know! Otherwise, if we proceed with this solution, it's best done in a separate PR.

Oh, to be clear, it's not using some other Pokemon's  timesHit - it's using the Pokemon's own timesHit counter. Something like the move Metronome would behave the same way, as an example. I don't think you need to change anything you did for that (so just remove the final test and you're good to go!)

cooshaug:
> > For the copycat case, I'm struggling to find an elegant way to implement it. I think a `sourcePokemon` attr to the battle's `lastMove` attr would really help, e.g.:
> > ```
> > ragefist: {
> > 		basePowerCallback(pokemon, target, move) {
> > 			let multiplier = pokemon.timesAttacked;
> > 			if (move.sourceEffect === 'copycat') {
> > 				multiplier = this.lastMove.sourcePokemon.timesAttacked;
> > 			}
> > 			return Math.min(350, 50 + 50 * multiplier);
> > 		}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Another workaround might be to backtrack through the log.
> > If ye can think of a way to fetch this info, let me know! Otherwise, if we proceed with this solution, it's best done in a separate PR.
> 
> Oh, to be clear, it's not using some other Pokemon's timesAttacked - it's using the Pokemon's own timesAttacked counter. Something like the move Metronome would behave the same way, as an example. I don't think you need to change anything you did for that (so just remove the final test and you're good to go!)

Oops - my mistake, I misread.

test/sim/moves/ragefist.js
```diff
@@ -0,0 +1,55 @@
...
+		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'vitalspirit', moves: ['sleeptalk', 'ragefist']}]});
+		battle.setPlayer('p2', {team: [{species: 'Umbreon', ability: 'shellarmor', moves: ['bulletseed', 'sleeptalk']}]});
+		battle.makeChoices();
+		const rageFistMutliplier = 1 * battle.log.filter(s => s.includes('hitcount'))[0].slice(-1) + 1;
+		battle.makeChoices('move ragefist', 'move sleeptalk');
+		const rageFistDamage = battle.p2.active[0].maxhp - battle.p2.active[0].hp;
+		assert.bounded(rageFistDamage, [17 * rageFistMutliplier, 21 * rageFistMutliplier]); // damage range here isn't exact, but it's enough for this test
+	});
+
+	it('should, when copied with Copycat, inherit the number of times hit', function () {
```
DaWoblefet, line 48: I would change this to something like 'it should use the user's own number of times hit even if called by another move', since it's not inheriting anything.

test/sim/moves/ragefist.js
```diff
@@ -0,0 +1,55 @@
...
+		assert.bounded(rageFistDamage, [17, 21]);
+		battle.makeChoices();
+		rageFistDamage = battle.p2.active[0].maxhp - rageFistDamage - battle.p2.active[0].hp;
+		assert.bounded(rageFistDamage, [17, 21]);
+	});
+
+	it('should increase BP after each hit of multi-hit moves', function () {
+		battle = common.createBattle();
+		battle.setPlayer('p1', {team: [{species: 'Primeape', ability: 'vitalspirit', moves: ['sleeptalk', 'ragefist']}]});
+		battle.setPlayer('p2', {team: [{species: 'Umbreon', ability: 'shellarmor', moves: ['bulletseed', 'sleeptalk']}]});
```
DaWoblefet, line 40: Use Skill Link here to guarantee a constant number of hits.

test/sim/moves/ragefist.js
cooshaug: Yep, that's indeed better

test/sim/moves/ragefist.js
cooshaug: Sounds good

DaWoblefet:
Thanks for your work on this, cooshaug!

PR #9042, opened 2022-11-23
Mark several forms as past gen only

dot-Comfey:
This change prevents these forms from being used in formats that ban nonexistent Pokemon but don't validate obtainable forms, such as Gen 9 Hackmons.

Reported here: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9403024


KrisXV:
thank you!

PR #9044, opened 2022-11-23
Gen 1: Fix Confusion/Psybeam confusion chance and substitute interaction

livid-washed:
In Gen 1, Confusion/Psybeam's secondary effect should be able to confuse enemies behind a substitute, as long as the substitute is unbroken. Also, the confusion chance should be 25/256, not 26/256. This is in contrast to other moves with '10%' chance of a secondary effect, like Flamethrower, which have a 26/256 chance.

https://github.com/pkmn/engine/blob/main/src/lib/gen1/README.md
https://github.com/pret/pokered/blob/master/engine/battle/effects.asm (lines 1112-1116)


AnnikaCodes:
Thanks!

PR #9045, opened 2022-11-23
Correct Ting-Lu's stats

EvezaGit:
Ting-Lu's day 1 patch nerf was only partially implemented: -10 HP, **-5 Def, +5 SpA**.


KrisXV:
ty

PR #9047, opened 2022-11-23
Fix Transform copying Terastallization

MathyFurret:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9409227


DaWoblefet:
Thanks Mathy!

PR #9049, opened 2022-11-24
Fix Flying Tera Type Pokemon using Roost

MathyFurret:
https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/post-9403448

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1968,7 +1968,7 @@ export class Pokemon {
 		const item = (this.ignoringItem() ? '' : this.item);
 		if (item === 'ironball') return true;
 		// If a Fire/Flying type uses Burn Up and Roost, it becomes ???/Flying-type, but it's still grounded.
-		if (!negateImmunity && this.hasType('Flying') && !('roost' in this.volatiles)) return false;
+		if (!negateImmunity && this.hasType('Flying') && !(this.hasType('???') && 'roost' in this.volatiles)) return false;
 		if (this.hasAbility('levitate') && !this.battle.suppressingAbility()) return null;
 		if ('magnetrise' in this.volatiles) return false;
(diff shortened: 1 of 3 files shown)


test/sim/moves/roost.js
```diff
@@ -78,6 +78,16 @@ describe('Roost', function () {
...
 		assert.equal(battle.p1.active[0].hp, battle.p1.active[0].maxhp); // Immune to Astonish
 	});
+
+	it('should not remove Flying type during Terastallization', function () {
+		battle = common.createBattle([[
+			{species: "Dudunsparce", ability: "runaway", moves: ['roost'], teraType: "Flying"},
+		], [
+			{species: "Bonsly", ability: "sturdy", moves: ['earthquake']},
+		]]);
+		battle.makeChoices('move roost terastallize', 'auto');
```
DaWoblefet, line 88: Roost won't remove the user's Flying-type at 100% HP either, so this test needs modified.

sim/pokemon.ts
```diff
@@ -1968,7 +1968,7 @@ export class Pokemon {
 		const item = (this.ignoringItem() ? '' : this.item);
 		if (item === 'ironball') return true;
 		// If a Fire/Flying type uses Burn Up and Roost, it becomes ???/Flying-type, but it's still grounded.
```
MathyFurret, line 1970: I couldn't find a unit test for this interaction but I don't trust myself to write one, it's beyond my comprehension.

MathyFurret: Actually, has Ground immunity for Tera Flying Roost actually been tested?

MathyFurret: https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/post-9410489 It has now

PR #9053, opened 2022-11-24
Gen 1: remove code referring to screens as side conditions

livid-washed:
Gen 1 treats Reflect/Light Screen as volatiles, not side conditions, so there is no reason to have these lines of code. This won't actually change any mechanics, just clean up some unnecessary code.


DaWoblefet:
Thanks livid-washed!

PR #9054, opened 2022-11-24
Gen 1: Fix inaccuracies in damage calculation

livid-washed:
This patches fixes a couple of inaccuracies in damage calculation:

- The damage should be clamped between 0 and 997 (i.e. only clamped from above), not 1 and 997.
- Type effectiveness should be applied individually against each of the target's types. For example, Ice attacks against Articuno are not very effective against the first type, and super effective against the second, and this affects the damage calculation.

https://www.smogon.com/forums/threads/past-gens-research-thread.3506992/#post-5878612


AnnikaCodes:
Thanks!

PR #9055, opened 2022-11-24
Gen 2: Fix traded-back events for Gen 2 evos

MathyFurret:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9410059


DaWoblefet:
Thanks Mathy!

PR #9056, opened 2022-11-24
Fix Protean activating on Terastallized Pokemon

TheImmortal:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9409142


DaWoblefet:
This will also apply to Libero.

TheImmortal:
> This was already fixed by [8a84d0b](https://github.com/smogon/pokemon-showdown/commit/8a84d0b2f0fb35c4426773b25e1ff4fc140991a4). Protean returns early without doing anything if `setType` returns false, which it does now for terastallized Pokemon.

Alright.

PR #9060, opened 2022-11-24
/statcalc: Check for Pokemon before IVs

dot-Comfey:
Reported here: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9408443


AnnikaCodes:
Thanks!

PR #9062, opened 2022-11-25
Gen 1: Remove unnecessary RNG rolls for secondaries

livid-washed:
This patch removes useless RNG rolls for Gen 1 secondary effects (e.g. Ice Beam's chance to freeze) when the move hits a substitute or the target has already fainted. The code will now first check if the target is available before rolling RNG to see if the secondary effect will be applied.


AnnikaCodes:
Thanks!

PR #9068, opened 2022-11-26
Add test for Opportunist crash

MathyFurret:


MathyFurret:
Marty just pushed a fix but I will keep the test

AnnikaCodes:
Thanks!

PR #9069, opened 2022-11-26
Unit tests: Refactor validator tests

MathyFurret:
Team validator unit tests are now split into several files under a `team-validator` directory:
* `basic.js` - Basic functionality
* `breeding.js` - Related to breeding and egg moves
* `custom-rules.js` - All the tests from the "custom rules" section
* `events.js` - Tests related to event Pokemon
* `formes.js` - Tests related to Pokemon with multiple formes
* `misc.js` - Tests too specific to be considered "basic" but that didn't fit anywhere else

Some stupidly long tests with extremely broad assertions were broken down.

Let me know if there's any tests you would move to a different file.


test/sim/team-validator/basic.js
```diff
@@ -0,0 +1,302 @@
...
+			{species: 'xerneas', ability: 'fairyaura', moves: ['snore'], ivs: {hp: 0, atk: 0, def: 0, spa: 0}, evs: {hp: 1}},
+		];
+		let illegal = TeamValidator.get('gen8anythinggoes').validateTeam(team);
+		assert(illegal);
+
+		illegal = TeamValidator.get('gen8purehackmons').validateTeam(team);
+		assert.equal(illegal, null);
+	});
+
+	it('should validate the Diancie released with zero perfect IVs', function () {
```
DaWoblefet, line 104: events.js

DaWoblefet:
Everything else looks good!

One of the things that's always bothered me is how we use ``assert(illegal)`` or ``assert.equals(illegal, null)``, which I think isn't very intuitive and has tripped me up more than once (e.g. typoing ``assert(illegal, null)``). What do you think about also refactoring to use a function like ``assert.legalTeam`` and explicitly the result of ``ValidateTeam`` along with ``true`` or ``false``? This might be a prospect for down the road and not in this PR.

MathyFurret:
Loved the idea so much that I implemented it right away. Wasn't too hard with a regex replace. I also moved that 1 test as you requested.

I think some other tests need to be refactored to use `assert.false.method` as well

DaWoblefet:
Really good work, thanks Mathy!

PR #9071, opened 2022-11-26
Gen 1: Fix stat modification, implement stat overflow glitch

livid-washed:
This patches implements the stat down modifier overflow glitch (division by zero not implemented for obvious reasons):

https://www.youtube.com/watch?v=y2AOm7r39Jg

- Stats are no longer automatically clamped from above by 999. This clamping is applied only when stats are being raised by boosting.
- When a stat is boosted at 999 or unboosted at 1, the boost/unboost is followed by an unboost/boost of -/+1 to the stat, respectively. The stat remains at 999 or 1. For example, if a Mewtwo is at +4 Special with 999 Special and uses Amnesia, the final stat stage is +5, and the Special remains at 999. A hint message will display for the first time this occurs in a battle.
- When a stat is unboosted, the stat is no longer clamped at 999, which can cause the stat down modifier overflow glitch if the stat is 1024 or higher. We still avoid division by zero glitches by ensuring that the final defense used in the damage calculation is at least 1, even though it can be 0 on cart.
-  The parspeeddrop and brnattackdrop volatiles are removed since they don't actually do anything, but can be misleading for people reading the code.

### DIFF
--- a/data/mods/gen1/conditions.ts
+++ b/data/mods/gen1/conditions.ts
@@ -14,7 +14,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		effectType: 'Status',
 		onStart(target) {
 			this.add('-status', target, 'brn');
-			target.addVolatile('brnattackdrop');
 		},
 		onAfterMoveSelfPriority: 2,
 		onAfterMoveSelf(pokemon) {
@@ -23,9 +22,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 			if (pokemon.volatiles['residualdmg']) {
 				this.hint("In Gen 1, Toxic's counter is retained after Rest and applies to PSN/BRN.", true);
 			}
-		},
-		onSwitchIn(pokemon) {
-			pokemon.addVolatile('brnattackdrop');
 		},
 		onAfterSwitchInSelf(pokemon) {
 			this.damage(this.clampIntRange(Math.floor(pokemon.maxhp / 16), 1));
@@ -36,7 +32,6 @@ export const Conditions: {[id: string]: ModdedConditionData} = {
 		effectType: 'Status',
 		onStart(target) {
 			this.add('-status', target, 'par');
-			target.addVolatile('parspeeddrop');
 		},
 		onBeforeMovePriority: 2,
 		onBeforeMove(pokemon) {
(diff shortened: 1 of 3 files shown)


PlagueVonKarma:
> We still avoid division by zero glitches by ensuring that the final defense used in the damage calculation is at least 1, even though it can be 0 on cart.

So RBY players tend to get _reaaaaally_ grabby about cart accuracy, moreso than any community I have ever been in, so I thought I'd bring up a few things just to make sure.

Some players have said this behaviour should fall under Desync Clause Mod, given the spirit of it was targeted at avoiding "literal game breaking" while not being inherently "wrong". The idea was to force a 1/256 miss if a situation _would_ cause a Link Battle to softlock, as technically, any softlock is avoidable. 

I remember @Enigami saying a suitable method would be banning Pokemon with stats that can result in a divide-by-zero error. I don't know if this is what the SWSH 600-and-something error ended up doing, or if that went the route of not being simulated at all. At least in RBY's case, PS has historically not simulated divide-by-zero errors in general - see Reflect / Light Screen's implementation - so I guess your way would be the consistent way to go about it.

It's probably nothing, but I thought I'd open some small talk just to make sure everything is alright.

livid-washed:
> Some players have said this behaviour should fall under Desync Clause Mod, given the spirit of it was targeted at avoiding "literal game breaking" while not being inherently "wrong". The idea was to force a 1/256 miss if a situation _would_ cause a Link Battle to softlock, as technically, any softlock is avoidable.
> 
> I remember @Enigami saying a suitable method would be banning Pokemon with stats that can result in a divide-by-zero error, which would not require complex simulation like the "force 1/256 if causes error" thing, but probably elaborate math instead. Paging him for that.

I think that is a good idea, one that someone else has discussed with me as well. I assume that no Pokemon have problematic stats when naturally maxed out? That would be my only concern, since preventing Pokemon from being their natural best doesn't feel right. Though I guess a difference of -1 stat isn't that big of a deal.

I think any such rule change (forbidden certain stats) would need to be agreed by RBY 'authorities' (council?), so I don't think it's my place to implement such a change. And of course, implementing move failure/miss for the / 0 case without restrictions on allowed stats seems abusable. As you mentioned, the divide-by-zero issue already exists currently thanks to Reflect / Light Screen, and I am for now just sticking with the current method that is used for dealing with it. 

Overall, I suppose there are two separate things here: (a) implementing the stat overflow glitch, and (b) how to deal with divide-by-zero. Here, I am doing (a) and leaving (b) as the current status quo, to be decided by others if they wish to change it.

AnnikaCodes:
Thanks!

PR #9074, opened 2022-11-27
Polls: Allow escaping separators

jmclemo6:
This commit allows users to escape separators in polls. Approved [on the forums](https://www.smogon.com/forums/threads/allow-users-to-escape-separators-in-polls.3709730/#post-9412630)


AnnikaCodes (approved):
seems fine

AnnikaCodes:
Thanks!

PR #9078, opened 2022-11-27
Fix Zangoose learnsets and give Pichu family Volt Tackle

DaWoblefet:
What should probably be Zangoose's egg moves are actually listed in its reminder moves: https://pastebin.com/N9Zgp9Qj

Not 100% sure on the source yet for Pichu.


DaWoblefet:
Pichu is Egg move-only, best as I can tell. However, since it's one of those weird cases, I'm not sure if the correct thing we do is mark as R or E.

DaWoblefet:
Talked with Anubis, and R seems to be the appropriate category.

PR #9080, opened 2022-11-27
Info: Support accessing a log of alternate accounts

mia-pi-git:
Approved by US.
This PR infrastructure to record namechanges server-side. It's meant to be a native integration of Officer Jenny's `~alts` command that's used frequently among staff, but one that has a better data source than just watching join/leave messages. The renames will be stored in a fairly simple database table that's primary keyed on `oldID, newID` to cut down on data use. Each pair will also have the latest IP stored, updated whenever that rename happens.
Unlike `~alts`, this is gated to global staff for privacy reasons (and related IP data is restricted to moderators and up).

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2866,9 +2866,93 @@ export const commands: Chat.ChatCommands = {
 			);
 		}
 	},
+	altlog: 'altslog',
+	altslog(target, room, user) {
+		this.checkCan('lock');
+		target = toID(target);
+		if (!target) {
+			return this.parse(`/help altslog`);
+		}
+		return this.parse(`/join view-altslog-${target}`);
+	},
+	altsloghelp: [
+		`/altslog [userid] - View the alternate account history for the given [userid]. Requires: % @ &`,
+	],
+};
+
+export const handlers: Chat.Handlers = {
+	onRename(user, oldID, newID) {
+		if (oldID === newID || !Config.usesqlite || [oldID, newID].some(f => f.startsWith('guest'))) return;
+		void Chat.database.run(
+			`REPLACE INTO alts_log (to_id, from_id, ip) VALUES (?, ?, ?)`,
+			[newID, oldID, user.latestIp]
+		);
+	},
 };
 
 export const pages: Chat.PageTable = {
+	async altslog(query, user) {
+		this.checkCan('lock');
+		this.title = '[Alts Log]';
+		const target = toID(query.shift());
+		if (!target) {
+			return this.errorReply(`Please specify a user to find alternate accounts for.`);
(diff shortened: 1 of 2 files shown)


server/chat-commands/info.ts
```diff
@@ -2866,9 +2866,93 @@ export const commands: Chat.ChatCommands = {
...
+	async altslog(query, user) {
+		this.checkCan('lock');
+		this.title = '[Alts Log]';
+		const target = toID(query.shift());
+		if (!target) {
+			return this.errorReply(`Please specify a user to find alternate accounts for.`);
+		}
+		this.title += ` ${target}`;
+		if (!Config.usesqlite) {
+			return this.errorReply(`The alternate account log is current disabled.`);
```
AnnikaCodes, line 2903:
```suggestion
			return this.errorReply(`The alternate account log is currently disabled.`);
```

server/chat-commands/info.ts
```diff
@@ -2866,9 +2866,93 @@ export const commands: Chat.ChatCommands = {
...
+		}
+		const rawLimit = query.shift() || "100";
+		const num = parseInt(rawLimit);
+		if (num > 3000) {
+			return this.errorReply(`3000 is the maximum number of results from the alternate account log.`);
+		}
+		if (isNaN(num)) {
+			return this.errorReply(`The max results must be a real number (received "${rawLimit}")`);
+		}
```
AnnikaCodes, line 2912: Should you check for negative numbers?

mia-pi-git: ...Maybe? This is only realistically gonna be used in a button and SQLite _does_ technically support negative limits. There's no real harm if we don't check - dunno what to tell anyone going out of their way to handicap themselves.
There's also no real reason not to check, so it pretty much comes down to preference. I have no opinion either way.

server/chat-commands/info.ts
mia-pi-git: Whoops.

server/chat-commands/info.ts
AnnikaCodes: A negative LIMIT will make SQLite return all results, which defeats the max of 3000.

mia-pi-git: Does it? Fascinating. Will swap, then.

AnnikaCodes: [SQLite docs](https://www.sqlite.org/lang_select.html#limitoffset):
> If the LIMIT expression evaluates to a negative value, then there is no upper bound on the number of rows returned.

AnnikaCodes (approved):

AnnikaCodes:
Thanks!

PR #9083, opened 2022-11-28
Use `[Gen 9] Anything Goes` as default format for /vtm

Karthik99999:


server/chat-commands/core.ts
```diff
@@ -1395,11 +1395,9 @@ export const commands: Chat.ChatCommands = {
 		}
 		if (!target) return this.errorReply(this.tr`Provide a valid format.`);
 		const originalFormat = Dex.formats.get(target);
-		// Note: The default here of [Gen 8] Anything Goes isn't normally hit; since the web client will send a default format
-		const format = originalFormat.effectType === 'Format' ? originalFormat : Dex.formats.get(
-			'[Gen 8] Anything Goes'
-		);
-		if (format.effectType !== this.tr`Format`) return this.popupReply(this.tr`Please provide a valid format.`);
+		// Note: The default here of [Gen 9] Anything Goes isn't normally hit; since the web client will send a default format
+		const format = originalFormat.effectType === 'Format' ? originalFormat : Dex.formats.get('[Gen 9] Anything Goes');
```
urkerab, line 1399: Would it be too cheaty to use `'ag'` as that's always an alias for the latest gen Anything Goes?

Karthik99999: Probably not. I would use `'anythinggoes'` to make it more clear though

DaWoblefet:
Thanks Karthik!

PR #9085, opened 2022-11-28
Gen 1: Fix Toxic-Haze interaction

livid-washed:
This patches fixes the interaction between Toxic and Haze. After Haze is used, on cart, the previous Toxic counter is never referred to again. I simulate this behavior by making Haze remove the Toxic counter. While this is different to cart in implementation, it should be identical in result.


AnnikaCodes:
Thanks!

PR #9088, opened 2022-11-29
Linked: Fix generation consistency

WeWuzNidokangz:
Maybe this format is intended to be removed soon, but currently it doesn't work properly, and if teams get past the validator it seems like there is a risk of causing in-battle crashes by mixing Gen 8 and 9 mechanics.


WeWuzNidokangz:
It seems like a similar problem is also occurring with Pokebilities. If there's no problem with this solution, I'll do the same with that OM.

KrisXV:
Go ahead and do that for now

WeWuzNidokangz:
OK. I looked over other mods just in case, but it looks like these were the only two not inheriting from the proper gen.

PR #9089, opened 2022-11-29
Ability Shield: Unskip tests that pass

im-tofa:
Now that the updated implementation by pyuk has been merged into master, the previously failing tests can be unskipped


DaWoblefet:
Thanks im-tofa!

PR #9093, opened 2022-11-30
Add unobtainable Gen 9 Egg moves

DaWoblefet:
See https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9416548

### DIFF
--- a/data/mods/gen1stadium/rulesets.ts
+++ b/data/mods/gen1stadium/rulesets.ts
@@ -3,5 +3,16 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 		effectType: 'ValidatorRule',
 		name: 'Standard',
 		ruleset: ['Obtainable', 'Stadium Sleep Clause', 'Freeze Clause Mod', 'Species Clause', 'OHKO Clause', 'Evasion Moves Clause', 'Exact HP Mod', 'Cancel Mod'],
+		banlist: [
+			// Only obtainable as Egg move, no valid parent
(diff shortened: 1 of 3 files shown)


DaWoblefet:
Actually this doesn't work either, because obtainablemoves is gonna backport with inheritance. I tried to mirror what we did for BDSP, but this isn't really a good solution either.

Karthik99999:
Can't you just make the banlist an empty array in Gen 8?

KrisXV:
> Marking them as unobtainable would work, but then STABmons people would get mad bc unobtainable moves are banned from that format (even though other hisuian moves aren’t banned bc they aren’t marked as unobtainable for some reason)

Why are you trying to bring OM philosophy into this? Regardless, I don’t like the unobtainable approach since that’s not something we’ve done in the past either.

DaWoblefet:
I think the correct solution would be, for Gen 8+, check if any obtainable Pokemon learn the move from a non-Egg source. I'll see if that's feasible or not to do.

DaWoblefet:
> I think the correct solution would be, for Gen 8+, check if any obtainable Pokemon learn the move from a non-Egg source. I'll see if that's feasible or not to do.

It probably is, but beyond the scope of what I want to do. Even if we did get this working, we'd have to hardcode in events or something to handle the cases like Shroodle getting Cross Poison or the Mareep line getting Electroweb.

DaWoblefet:
Superseded by https://github.com/smogon/pokemon-showdown/pull/9096

PR #9097, opened 2022-11-30
Ceaseless Edge & Stone Axe: Update implementation based on research thread

im-tofa:
Added an implementation and some tests based on what I could derive from the research thread (great thread!). The Magic Bounce test is based on the fact that one of the researchers stated it has similar mechanics as Toxic Debris, which was demonstrated to not be bounced earlier in the thread!

I have not thought too much about interactions with other abilities/moves aside from Magic Bounce, Protect, Substitute.

No released Pokémon has this move at the moment :)


data/moves.ts
```diff
@@ -2203,11 +2203,17 @@ export const Moves: {[moveid: string]: MoveData} = {
...
+			for (const side of source.side.foeSidesWithConditions()) {
+				side.addSideCondition('spikes');
+			}
+		},
+		onAfterSubDamage(damage, target, source, effect) {
+			for (const side of source.side.foeSidesWithConditions()) {
+				side.addSideCondition('spikes');
+			}
 		},
+		// critRatio: 2,
```
Karthik99999, line 2216:
```suggestion
		// critRatio: 2,
		self: {
			onHit(source) {
				for (const side of source.side.foeSidesWithConditions()) {
					side.addSideCondition('spikes');
				}
			},
		},
```

KrisXV:
This implementation causes the move to no longer get boosted by Sheer Force when it's supposed to.

Karthik99999:
> This implementation causes the move to no longer get boosted by Sheer Force when it's supposed to.

I think we would have to properly implement the sheer force flag for this but that's out of scope of this pr

data/moves.ts
im-tofa: Oh this triggers onHit even for sub hits, nice! I'll fix this, ty :)

im-tofa:
> I'd bet Stone Axe works the same way.
Very good point! I'll add it to this PR tomorrow, have to sleep soon

> This implementation causes the move to no longer get boosted by Sheer Force when it's supposed to.
Is it confirmed to be Sheer Force-boostable (& therefore also not put up spikes when the user has Sheer Force)? If so I'll add tests for that as well

Karthik99999:
> Is it confirmed to be Sheer Force-boostable (& therefore also not put up spikes when the user has Sheer Force)? If so I'll add tests for that as well

It has the Sheer Force flag (in game, we currently don't support it on showdown) so yes it is affected by Sheer Force

im-tofa:
Ok, I'll take another look at it later this week then :)

data/moves.ts
```diff
@@ -2204,9 +2204,13 @@ export const Moves: {[moveid: string]: MoveData} = {
 		priority: 0,
 		flags: {contact: 1, protect: 1, mirror: 1, slicing: 1},
 		// critRatio: 2,
-		secondary: {
-			chance: 100,
-			sideCondition: 'spikes',
+		self: {
+			// TODO fix sheer force interaction
```
urkerab, line 2208: Diamond Storm just puts in a dummy `secondary: {},`.

im-tofa: Worked like a charm, thanks! :)

im-tofa:
Thanks for the feedback everyone! Added an analogous Stone Axe implementation & fixed the Sheer Force interaction using urkerab's suggestion :)

PR #9099, opened 2022-12-01
Gen 1: Minor corrections for stat down overflow

livid-washed:
Fixes a couple little issues with the stat down overflow patch:
- Adds a space to the hint regarding div by 0
- Removes an unremoved console.log from the test
- Remove clamping for Reflect / Light Screen


AnnikaCodes:
Thanks!

PR #9101, opened 2022-12-01
Small code cleanup in HC team generation

adrivrie:
EVs generation in Hackmons Cup had an exception for gen 6 for seemingly no reason. This exception was removed, which has no effect as gen6hackmonscup does not exist anymore.

Also, changes EV generation to the same format as IV generation for clarity.


adrivrie:
I had no idea. My bad!

PR #9102, opened 2022-12-01
Update commandline -h generate-team

adrivrie:
The -h for the command line tool (as well as the COMMANDLINE.md file) had as the default for the generate-team option "gen7randombattles". This should be gen9randombattles now.


DaWoblefet:
There is no Gen 9 Random Battle yet, so this can't be the case.

adrivrie:
It is, and running it without argument throws an error (because indeed there's no gen9 rands yet). I figured hard-defaulting it to gen8 now makes little sense as this hasn't been updated in years, and it will only keep failing for a limited time.

DaWoblefet:
This will be merged once Gen 9 randbats goes live.

AnnikaCodes:
Why not just write "the latest Random Battles format"?

MathyFurret:
Beep :3

monsanto:
Annika's suggestion seems reasonable to me

adrivrie:
Changed it to the "latest Random Battles format".

DaWoblefet:
Thanks adrivrie!

PR #9106, opened 2022-12-02
Salt Cure's effect should be a Secondary effect

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9418556
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9417534

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -15863,7 +15863,6 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 15,
 		priority: 0,
 		flags: {protect: 1, mirror: 1},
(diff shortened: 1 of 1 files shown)


pacmanboss256:
does it get affected by sheer force or is it like sparkling aria? cetitan can learn entrainment with sheer force to potentially test this

Karthik99999:
It's blocked by Covert Cloak (and presumably Shield Dust) too so it's not like Sparkling Aria

pacmanboss256:
ah alright then

PR #9114, opened 2022-12-03
Costar should copy crit volatiles

Karthik99999:


urkerab:
If this is now a clone of Psych Up, is there a way of sharing the code somehow? (Always assuming that it would be worthwhile.)

PR #9117, opened 2022-12-04
Gen 1: Fix mist fail message appearing twice

TheVanadium:
Issue reported on
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/page-16#post-9252475


AnnikaCodes:
This was fixed #8836 as far as I can tell; has it been reproduced since then?

TheVanadium:
It has not been replicated. I assumed it hadn't been fixed, as it is in the following Smogon thread as unfixed:
image

I will revert these changes.

PR #9127, opened 2022-12-05
Add [Gen 9] Doubles Hackmons Cup

adrivrie:
Adds gen 9 doubles hackmons cup as a challenge-only format. Also moves CC2v2 from singles to randomized formats.

Discussed in the rands auth discord here: https://discord.com/channels/294651453279174656/1041098998624288990/1049303050042482748


config/formats.ts
```diff
@@ -2089,6 +2081,24 @@ export const Formats: FormatList = [
...
+
+		mod: 'gen9',
+		gameType: 'doubles',
+		team: 'randomHC',
+		searchShow: false,
+		ruleset: ['HP Percentage Mod', 'Cancel Mod'],
+		banlist: ['Nonexistent'],
+	},
+	{
+		name: "[Gen 9] Challenge Cup 2v2",
```
AnnikaCodes, line 2095: Why move CC2v2 to Randomized Formats but not CC1v1?

adrivrie: I figured the reason for having them in the "top slot" was to make a couple rands formats more visible while awaiting gen9, but it was a little grating to have a doubles format in the singles category, and we figured it'd be better in rands than doubles. I don't really mind either way however, just prefer cc2v2 not to be in singles.

Also, feel free to just close this as you already added DHC :)

MathyFurret:
This is already done #9184

adrivrie:
Indeed it is, thought it was already closed. Closing now!

PR #9129, opened 2022-12-05
Mafia: Fix a bug with resetgame

motogp-bit:
Resetgame can possibly call distributeRoles() when game is set to idea,causing 'Cannot read property 'safeName' of undefined' because role is undefined after idearoll.

This is probably a temporary fix because it actually doesn't allow resetgame to be used on idea whatsoever,even if the game is no longer idea.


AnnikaCodes:
This doesn't seem like it will make anything /worse/.

PR #9130, opened 2022-12-06
Add missing Metronome moves + update Metronome description to Gen 9

kittenchilly:
Metronome can't select Order Up, and can select Dragon Hammer.
Also updates the description with all the new moves it can't select.
Old description is still in Gen 8.


DaWoblefet:
Although not strictly necessary, while you're working on it, would you mind adding the correct noMetronome list to the Gen 8 mod? It'd be nice to make sure the lists line up.

It also looks like the new description is missing Order Up. Tidy Up also isn't in the correct spot alphabetically. There's also the [Dragon Hammer discrepancy from last gen](https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/post-9213663). I'll ask Anubis if she knows about its behavior in Gen 9.

kittenchilly:
Weird, I got the list directly from the data the actual Metronome in the sim uses. This means Metronome can incorrectly use Order Up right now, if that's true.

PR #9131, opened 2022-12-06
Punishments: Fix muting users on shared IP

MathyFurret:
If the user is apparently banned but got in the room anyways they should be mutable. I will add some more context for this in private


mia-pi-git:
Discussed this in #staff, determined this wasn't the ideal way to fix it (bug in punishments code, not in /mute, which should be handled in punishments). Closing this.

MathyFurret:
Should be fixed by https://github.com/smogon/pokemon-showdown/commit/b90d453572ccf5a743588af227e3aeb7e5408189

PR #9134, opened 2022-12-07
Account for moves with <60 Base Power when terastallized

KrisXV:


sim/battle-actions.ts
```diff
@@ -1609,6 +1609,17 @@ export class BattleActions {
 			basePower = 0;
 		}
 
+		if (
+			basePower < 60 && source.getTypes(true).includes(move.type) && source.terastallized && move.priority <= 0 &&
```
urkerab, line 1613: `source.getTypes(true)` will always return `source.terastallized` if it's set.

KrisXV: I'm unsure what you're saying here

urkerab: I don't know what the correct condition is, but `source.getTypes(true).includes(move.type) && source.terastallized` is the same as `move.type === source.terastallized`.

KrisXV: Yeah, but the current implementation is used in other places in the sim currently so I wanna keep consistent w/ that

MathyFurret: I used the new form in #9138 i like it better

urkerab: I couldn't find any other place using `source.getTypes(true)`. I did find one using `source.getTypes(false, true)` but obviously that's completely different.

DaWoblefet:
Thanks Kris!

PR #9135, opened 2022-12-07
Implement fainted forme regression

KrisXV:
mainly only relevant for natdex, there's a thing with revival blessing not un-greying the icons on the sidebar that i cant quite figure out but functionally this works as intended (barring the zero to hero interaction, we dont rly know how that should actually work with revival yet)

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -131,7 +131,10 @@ export const Formats: FormatList = [
 			'Picked Team Size = 1', 'Max Team Size = 3',
 			'Standard', 'Terastal Clause', 'Sleep Moves Clause', 'Accuracy Moves Clause', '!Sleep Clause Mod',
 		],
-		banlist: ['Chi-Yu', 'Cinderace', 'Dragonite', 'Koraidon', 'Mimikyu', 'Miraidon', 'Scream Tail', 'Moody', 'Focus Band', 'Focus Sash', 'King\'s Rock', 'Quick Claw', 'Acupressure', 'Perish Song'],
+		banlist: [
+			'Chi-Yu', 'Cinderace', 'Dragonite', 'Flutter Mane', 'Gholdengo', 'Koraidon', 'Mimikyu', 'Miraidon', 'Scream Tail',
+			'Moody', 'Focus Band', 'Focus Sash', 'King\'s Rock', 'Quick Claw', 'Acupressure', 'Perish Song',
+		],
 	},
 	{
 		name: "[Gen 9] Anything Goes",


sim/battle.ts
```diff
@@ -2316,6 +2316,14 @@ export class Battle {
 				pokemon.isActive = false;
 				pokemon.isStarted = false;
 				delete pokemon.terastallized;
+				if (pokemon.preFormeChangeInfo) {
+					pokemon.setSpecies(pokemon.preFormeChangeInfo.species);
+					pokemon.baseSpecies = pokemon.preFormeChangeInfo.species;
+					pokemon.baseAbility = pokemon.ability = pokemon.preFormeChangeInfo.ability;
+					pokemon.details = pokemon.species.name + (pokemon.level === 100 ? '' : ', L' + pokemon.level) +
+						(pokemon.gender === '' ? '' : ', ' + pokemon.gender) + (pokemon.set.shiny ? ', shiny' : '');
+					this.add('detailschange', pokemon, (pokemon.illusion || pokemon).details);
```
urkerab, line 2325: `illusion` got cleared ten lines above here.

sim/pokemon.ts
```diff
@@ -1356,6 +1360,10 @@ export class Pokemon {
 				this.ability = ''; // Don't allow Illusion to wear off
 			}
 			this.setAbility(species.abilities['0'], null, true);
+			this.preFormeChangeInfo = {
+				species: preFormeChangeSpecies,
+				ability: this.baseAbility,
```
urkerab, line 1365: How does these differ from `this.set.species` and `this.set.ability`?

urkerab (commented):
Don't you need an exemption for Shaymin-Sky (which remains in land forme if it faints after being frozen)? (Might be as simple as checking whether the new forme is battle-only.)

sim/battle.ts
```diff
@@ -2316,6 +2316,14 @@ export class Battle {
 				pokemon.isActive = false;
 				pokemon.isStarted = false;
 				delete pokemon.terastallized;
+				if (pokemon.preFormeChangeInfo) {
+					pokemon.setSpecies(pokemon.preFormeChangeInfo.species);
+					pokemon.baseSpecies = pokemon.preFormeChangeInfo.species;
```
Karthik99999, line 2321: I don't think we can do this for every species. For example, Palafin doesn't go back to Zero forme when it faints.

KrisXV: I was trying to avoid adding a property in pokedex or something, but it seems like the solution is to either do that or hardcode a list here similar to the disguise/iceface hardcode

sim/pokemon.ts
KrisXV: the Species here is a Species class, not much difference but I think I prefer it so it doesn't need to be grabbed with species.get again, there's no real difference for ability but again I like using the existing properties in the class if/when possible

sim/battle.ts
KrisXV: O

KrisXV:
i didnt mean to close this but ill get back to this eventually

PR #9138, opened 2022-12-08
Tera + Adaptability STAB is x2.25

MathyFurret:
https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/post-9426618


urkerab:
Although apparently you lose Adaptability on your original STAB.

https://www.smogon.com/forums/posts/9426572

test/sim/misc/terastal.js
```diff
@@ -137,4 +137,26 @@ describe("Terastallization", function () {
...
+
+	it("should combine with Adaptability for an overall STAB of x2.25", () => {
+		battle = common.createBattle([[
+			{species: "Dragalge", ability: 'adaptability', moves: ['acid'], teraType: "Poison"},
+		], [
+			{species: "Mareep", ability: 'static', moves: ['sleeptalk']},
+		]]);
+		battle.makeChoices('move acid terastallize', 'auto');
+		const damage = battle.p2.active[0].maxhp - battle.p2.active[0].hp;
+		assert.bounded(damage, [119, 141], "Actual damage: " + damage);
```
Marty-D, line 149: Is this accounting for the fact that Acid will be 60 power in this case? I would use some move with more than 60 power to start with if not, so it's still correct after the other PR is merged.

sim/battle-actions.ts
```diff
@@ -1698,21 +1698,21 @@ export class BattleActions {
 		baseDamage = this.battle.randomizer(baseDamage);
 
 		// STAB
-		if (move.forceSTAB || (type !== '???' && pokemon.hasType(type))) {
+		if (move.forceSTAB || (type !== '???' &&
+			(pokemon.hasType(type) || (pokemon.terastallized && pokemon.getTypes(false, true).includes(type))))) {
```
AnnikaCodes, line 1702: This is an aside, but `getTypes` would be a great place to use an object to pass options — if I'm not already familiar with the function, the arguments in `pokemon.getTypes(false, true)` don't make any sense.

AnnikaCodes:
CI failed because of `npm install`, which is odd; I'm rerunning it.

PR #9140, opened 2022-12-08
Update Pokedex number order

Glazelf:
According to Joe Merrick [[1](https://twitter.com/JoeMerrick/status/1600547621520932864)][[2](https://twitter.com/JoeMerrick/status/1600813442188746752)] there is an alternative internal order for the natdex that's backed by the TCG releases.  
I haven't seen anyone else really talk about this yet, but I don't use the Smogon forums. :P  

This PR simply reorders the Pokédex file entries and changes the natdex numbers to match that order.  
If anything else needs to be altered to accomodate for this lmk, but I couldn't find anything by searching for the natdex numbers that would break because of this.


Glazelf:
Forgot to update from upstream beforehand but should be good now.  🤦

Marty-D:
Thanks, I would merge this after reviewing but changing the dex numbers will break the sprite sheet, so I have to make a new sprite sheet first. I'll mark this as WIP until that's done.

PR #9143, opened 2022-12-08
Gens 2-3: Fix accuracy of moves

livid-washed:
This patch fixes the accuracy of some moves for Gen 2-3:

It is important to list moves that don't check for accuracy as having true accuracy, even if they are listed in-game as having 100% accuracy. Otherwise other effects that can cause moves to miss (like BrightPowder) can come into play when they shouldn't. https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-84#post-9373204

Gen 2:
- Lock-On and Pain Split should have 100% accuracy (not true)
- Nightmare should have true accuracy (not 100%)
https://www.smogon.com/forums/threads/gen-2-ps-development-post-bugs-here.3524926/post-8557865

Gen 3:
- Block, Mean Look, Mimic, Pain Split, Role Play, Skill Swap, Spider Web, Vital Throw, and Yawn bypass accuracy checks, even though they are listed in-game as 100% accuracy. They should have true accuracy.
- Foresight, Odor Sleuth, and Mind Reader should have 100% accuracy (not true)
- Memento and Nightmare should have true accuracy (not 100%)
These were listed on Bulbapedia and confirmed via assembly and some in-game testing (obtaining misses for Foresight, Odor Sleuth, and Mind Reader after accuracy is lowered).

### DIFF
--- a/data/mods/gen2/moves.ts
+++ b/data/mods/gen2/moves.ts
@@ -514,6 +514,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	},
 	nightmare: {
 		inherit: true,
+		accuracy: 100,
 		condition: {
 			noCopy: true,
 			onStart(pokemon) {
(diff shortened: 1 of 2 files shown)


AnnikaCodes:
Thanks!

PR #9149, opened 2022-12-09
Gen 4: Ensure Brick Break breaks screens of immune targets

vivapuffs:
https://github.com/smogon/pokemon-showdown/projects/3#card-61615516
This is my first time contributing to an open source project, please let me know if I've made any mistakes

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -178,7 +178,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		recoil: [1, 3],
 	},
-		brickbreak: {
+	brickbreak: {
 		inherit: true,
 		ignoreImmunity: true,
 		onTryHit(pokemon, target){
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -181,14 +181,14 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 	brickbreak: {
 		inherit: true,
 		ignoreImmunity: true,
-		onTryHit(pokemon, target){
+		onTryHit(pokemon, target) {
 			pokemon.side.removeSideCondition('reflect');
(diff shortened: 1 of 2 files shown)


data/mods/gen4/moves.ts
```diff
@@ -178,6 +178,18 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		recoil: [1, 3],
 	},
+	brickbreak: {
+		inherit: true,
+		ignoreImmunity: true,
+		onTryHit(pokemon, target){
+			pokemon.side.removeSideCondition('reflect');
+			pokemon.side.removeSideCondition('lightscreen');
+			if (pokemon.hasType('Ghost')){
```
KrisXV, line 187:
```suggestion
			if (pokemon.hasType('Ghost')) {
```

MathyFurret:
Hi! There's a unit test written for this here: https://github.com/smogon/pokemon-showdown/blob/master/test/sim/moves/brickbreak.js#L41 Could you delete the ".skip" so it runs?

vivapuffs:
> Hi! There's a unit test written for this here: https://github.com/smogon/pokemon-showdown/blob/master/test/sim/moves/brickbreak.js#L41 Could you delete the ".skip" so it runs?

oh whoops should be done now

urkerab:
What if the user's ability is Scrappy?

vivapuffs:
good catch, is the fix i put in my latest commit appropriate?

urkerab:
You should use `hasAbility` to check for an ability, as that takes Gastro Acid into account. You should also use `runImmunity` to check for an immunity, as that allows Foresight to negate the immunity.

vivapuffs:
it appears to work correctly with just runImmunity (see latest commit,) tested against ghost mon, ghost mon under foresight, and scrappy mon vs. ghost mon

PR #9155, opened 2022-12-11
Fix copycat dancer interaction

aviettran:
Previously, when copycat selected a dance move, a dancer would then target the pokemon that used copycat instead of copycat's inner move's target.

Correct interaction verified in game: https://twitter.com/WeedleTwineedle/status/1601925296898592770

### DIFF
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -328,7 +328,9 @@ export class BattleActions {
 				if (this.battle.faintMessages()) break;
 				if (dancer.fainted) continue;
 				this.battle.add('-activate', dancer, 'ability: Dancer');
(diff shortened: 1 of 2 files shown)


DaWoblefet:
This looks good to me, but this sort of thing is right up @pyuk-bot's alley so I'd like pyuk to review.

Karthik99999:
> You know, tbh I’m not even sure how dancer is able to copy called dance moves at all. Doesn’t copycat call `useMove` directly, which should skip the dancer check entirely?

`move` is set to `Battle#activeMove` once we get to the dancer part, which should be the dance move called by copycat

DaWoblefet:
Thanks aviettran!

PR #9156, opened 2022-12-11
Gen 1: Actually fix Metronome calling Counter

livid-washed:
A bit of an oopsie with the last Counter patch, now Metronome calling Counter should fail for real.

https://github.com/smogon/pokemon-showdown/pull/9136 was the previous patch


AnnikaCodes:
thanks!

PR #9157, opened 2022-12-11
Fix various flag inconsistencies in moves.

kittenchilly:
Updates Copycat, Encore, Instruct, Me First, Mimic, Sleep Talk, and various smaller things that I noticed when doing a passthrough of comparing all the flags in SV to the ones we define.

Also implements metronome for gen 8 mod


data/moves.ts
```diff
@@ -17222,14 +17222,14 @@ export const Moves: {[moveid: string]: MoveData} = {
...
 			const noSleepTalk = [
-				'assist', 'beakblast', 'belch', 'bide', 'celebrate', 'chatter', 'copycat', 'dynamaxcannon', 'focuspunch', 'mefirst', 'metronome', 'mimic', 'mirrormove', 'naturepower', 'shelltrap', 'sketch', 'sleeptalk', 'uproar',
+				'assist', 'beakblast', 'belch', 'bide', 'blazingtorque', 'bounce', 'celebrate', 'chatter', 'combattorque', 'copycat', 'dig', 'dive', 'dynamaxcannon', 'fly', 'focuspunch', 'freezeshock', 'geomancy', 'mefirst', 'metronome', 'mimic', 'mirrormove', 'naturepower', 'noxioustorque', 'phantomforce', 'razorwind', 'shadowforce', 'shelltrap', 'sketch', 'skullbash', 'skyattack', 'skydrop', 'sleeptalk', 'solarbeam', 'solarblade', 'struggle', 'uproar', 'wickedtorque'
 			];
 			const moves = [];
 			for (const moveSlot of pokemon.moveSlots) {
 				const moveid = moveSlot.id;
 				if (!moveid) continue;
 				const move = this.dex.moves.get(moveid);
-				if (noSleepTalk.includes(moveid) || move.flags['charge'] || (move.isZ && move.basePower !== 1) || move.isMax) {
```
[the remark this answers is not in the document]
kittenchilly: Meteor Beam, interestingly enough.

data/moves.ts
```diff
@@ -9782,12 +9782,12 @@ export const Moves: {[moveid: string]: MoveData} = {
 			const lastMove = target.lastMove;
 			const moveIndex = target.moves.indexOf(lastMove.id);
 			const noInstruct = [
-				'assist', 'beakblast', 'belch', 'bide', 'celebrate', 'copycat', 'dynamaxcannon', 'focuspunch', 'iceball', 'instruct', 'kingsshield', 'mefirst', 'metronome', 'mimic', 'mirrormove', 'naturepower', 'obstruct', 'outrage', 'petaldance', 'rollout', 'shelltrap', 'sketch', 'sleeptalk', 'struggle', 'thrash', 'transform', 'uproar',
+				'assist', 'beakblast', 'belch', 'bide', 'blazingtorque', 'bounce', 'celebrate', 'chatter', 'combattorque', 'copycat', 'dig', 'dive', 'dynamaxcannon', 'fly', 'focuspunch', 'freezeshock', 'geomancy', 'holdhands', 'iceball', 'iceburn', 'instruct', 'kingsshield', 'magicaltorque', 'mefirst', 'meteorassault', 'metronome', 'mimic', 'mirrormove', 'naturepower', 'noxioustorque', 'obstruct', 'outrage', 'petaldance', 'phantomforce', 'razorwind', 'rollout', 'shadowforce', 'shelltrap', 'sketch', 'skullbash', 'skyattack', 'skydrop', 'sleeptalk', 'solarbeam', 'solarblade', 'struggle', 'thrash', 'transform', 'uproar', 'wickedtorque',
```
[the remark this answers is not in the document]
kittenchilly: I saw a few charge and recharge moves that didn't have the NoInstruct flag, but since Hyper Beam apparently can't be used by instruct but it doesn't have the NoInstruct flag, it seems the game has a separate check aside from the flag. There seem to be some moves that would fail if run through both though, for some reason.

data/moves.ts
```diff
@@ -2851,7 +2851,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		flags: {},
 		onHit(pokemon) {
 			const noCopycat = [
-				'assist', 'banefulbunker', 'beakblast', 'behemothbash', 'behemothblade', 'belch', 'bestow', 'celebrate', 'chatter', 'circlethrow', 'copycat', 'counter', 'covet', 'craftyshield', 'destinybond', 'detect', 'dragontail', 'dynamaxcannon', 'endure', 'feint', 'focuspunch', 'followme', 'helpinghand', 'holdhands', 'kingsshield', 'matblock', 'mefirst', 'metronome', 'mimic', 'mirrorcoat', 'mirrormove', 'naturepower', 'obstruct', 'protect', 'ragepowder', 'roar', 'shelltrap', 'sketch', 'sleeptalk', 'snatch', 'spikyshield', 'spotlight', 'struggle', 'switcheroo', 'thief', 'transform', 'trick', 'whirlwind',
```
Marty-D, line 2854: If you're going to remove things from these lists, you need to override them in the Gen 8 mod with the old lists, which were actually manually tested because they weren't explicitly flagged before Gen 9.

kittenchilly: Done.

data/moves.ts
```diff
@@ -2849,18 +2849,18 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 20,
 		priority: 0,
 		flags: {},
-		onHit(pokemon) {
-			const noCopycat = [
-				'assist', 'banefulbunker', 'beakblast', 'behemothbash', 'behemothblade', 'belch', 'bestow', 'celebrate', 'chatter', 'circlethrow', 'copycat', 'counter', 'covet', 'craftyshield', 'destinybond', 'detect', 'dragontail', 'dynamaxcannon', 'endure', 'feint', 'focuspunch', 'followme', 'helpinghand', 'holdhands', 'kingsshield', 'matblock', 'mefirst', 'metronome', 'mimic', 'mirrorcoat', 'mirrormove', 'naturepower', 'obstruct', 'protect', 'ragepowder', 'roar', 'shelltrap', 'sketch', 'sleeptalk', 'snatch', 'spikyshield', 'spotlight', 'struggle', 'switcheroo', 'thief', 'transform', 'trick', 'whirlwind',
-			];
+		noCopycat: [
```
Marty-D, line 2852: This isn't what I meant, I meant override `onHit` like everything else; this is a half-refactor because Gen 5 overrides `onHit` again and this just becomes unused extra data. One day these should all be refactored into move flags, but that shouldn't be the goal of this PR.

You also have to override these things for more than just Copycat - any move list where you've added a move that existed in the gen before it, or removed any move has to be accounted for.

PR #9161, opened 2022-12-12
Add rating for every ability in gen 8

adrivrie:
Adds a rating to every ability in gen8, taken from the final commit before the gen9 implementation.

With gen9 random battles in development, ability scores are being changed resulting in changes in past-gen set generation. This is especially prevalent because the set generation system will behave differently in gen9. 

The cognitive load of having to take into account the effects on all previous gens when changing an ability rating is now too large, so things are starting to fall through the cracks. This is compounded by the non-programmers who do the bulk of randbats upkeep not being familiar with inheritance as a concept. As a specific example: the rating of guts was raised by 0.5 in preparation for gen9, leading to Machamp in gens 4 and 6 (and others?) rolling Dynamic Punch without No guard.

By adding all abilities' ratings to gen8, we limit the complexity to what it is now, (partially) freezing sets at a point where we know they work.

https://discord.com/channels/294651453279174656/809413252517462026/1051852914928386158

### DIFF
--- a/data/mods/gen8/abilities.ts
+++ b/data/mods/gen8/abilities.ts
@@ -1,4 +1,106 @@
+/*
+
+Ratings and how they work:
+
+-1: Detrimental
+	  An ability that severely harms the user.
+	ex. Defeatist, Slow Start
+
+ 0: Useless
+	  An ability with no overall benefit in a singles battle.
+	ex. Color Change, Plus
+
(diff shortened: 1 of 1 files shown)


ACakeWearingAHat:
I approve of this change for Random Battles so long as @KrisXV does

PR #9163, opened 2022-12-12
Transforming into Pokemon with the same ability (Gen 5+)

benstensen:
This pull request is in response to [this card](https://github.com/smogon/pokemon-showdown/projects/3#card-85843016). 

> In Gen 5+, Transforming with an entrance Ability (e.g. Intimidate) into a Pokemon with the same entrance Ability will not reactivate the Ability.

This behavior has been implemented through an additional check in setAbility. The corresponding test in `test/sim/moves/transform.js` is now passing.


DaWoblefet:
Thanks benstensen!

PR #9169, opened 2022-12-13
Implement Gen 9 Random Battle

ACakeWearingAHat:
Credits: 
Framework: @livid-washed 
Coding: me and @livid-washed 
Sets: me and the random battles room staff
/randbats command (will be appended onto this PR): @MathyFurret 
There are a lot of commits, we know. It's a big project. 

Gen 9 Random Battles uses a different generation method and set structure than previous generations of random battles. These differences can be summarized in three points.
1. **Pokemon are classified into roles**. Each role determines the moves prioritized and the items generated, as well as other miscellaneous effects. There are ten roles in total. Each pokemon may be assigned multiple roles; each role will have its own movepool and Tera type pool. This is a method of splitting sets that does not require use of cosmetically identical formes.

2. **There is a new data file specifically for Random Battles**. This data file removes clutter from the formats-data file. It contains all the fields needed: species, level, roles (with tera types and movepools under the role field). This data file will be only for [Gen 9] Random Battle; other identical data files can be made for other formats. This opens the way for individualization of free-for-all sets in the future, a major shortcoming of keeping it all in one file.

3. **Move generation is frontloaded, rather than backloaded**. In past generations, the moveset is first filled out, and then moves are rejected from it until a desirable moveset is settled upon. This system had many problems, among which are the code being complex and esoteric, certain bugs being nigh-impossible to fix or identify, and certain desirable sets being impossible to reliably achieve. This new system instead tries to settle on a correct moveset on the first try by selecting particularly desirable options directly from the pool, and then removing undesirable options from the unselected movepool, before generating random moves. This system of enforcement > incompatibilities > random move generation has a significantly lower chance of failure than the old system and a higher chance to get sets that the pokemon particularly appreciates. Most importantly, this new system drastically decreases the amount of hardcoding (and coding in general) required to keep the format maintenanced; the move incompatibility section is nearly 10x smaller than Gen 8's shouldCullMove section.

Still left to do, albeit maybe not in this PR:
-Update client-side such that [Gen 9] Random Battle is the default ladder and challengeable format. 
-There's a bit in /server/ladders.ts that seems to adjust Gen 8 OU and Gen 8 Random Battle's ladders in some way. Might be important to do this for their Gen 9 variants. I don't know the specifics.
-the /setodds command needs to be updated to accommodate for the new data file, I believe.


data/random-teams.ts
```diff
@@ -350,1463 +306,805 @@ export class RandomTeams {
...
+		if (species.requiredMove) {
+			const move = this.dex.moves.get(species.requiredMove).id;
+			counter = this.addMove(move, moves, types, abilities, teamDetails, species, isLead, isDoubles,
+				movePool, teraType, role);
+		}
+
+		// Add other moves you really want to have, e.g. STAB, recovery, setup, depending on role.
+
+		// Enforce STAB
+		types.forEach((type, index) => {
```
mia-pi-git, line 603: `for-of` over forEach, please.

livid-washed: > `for-of` over forEach, please.

Thanks! Forgot to change that when we stopped using the index.

Karthik99999: For future reference, you can get the index with `for-of` by using `Array.entries()`

data/random-teams.ts
```diff
@@ -1830,165 +1128,116 @@ export class RandomTeams {
...
+			PU: 88, "(PU)": 88, NFE: 88,
+		};
+		return tierScale[tier] || 80;
 	}
 
 	randomSet(
 		species: string | Species,
 		teamDetails: RandomTeamsTypes.TeamDetails = {},
 		isLead = false,
 		isDoubles = false,
```
KrisXV, line 1237:
```suggestion
		isDoubles = false
```

data/random-teams.ts
```diff
@@ -1997,16 +1246,28 @@ export class RandomTeams {
...
 		}
-		if (species.name.endsWith('-Gmax')) {
-			forme = species.name.slice(0, -5);
-			gmax = true;
+		let sets;
+		if (isDoubles) {
+			sets = this.randomDoublesSets[species.id]["sets"];
+		} else {
+			sets = this.randomSets[species.id]["sets"];
+		}
```
KrisXV, line 1254:
```suggestion
		const sets = (this as any)[`random${isDoubles ? 'Doubles' : ''}Sets`][species.id]["sets"];
```

data/random-teams.ts
```diff
@@ -1997,16 +1246,28 @@ export class RandomTeams {
...
+		const possible_sets = [];
+		for (const set of sets) {
+			if (teamDetails.teraBlast && set.role === "Tera Blast user") {
+				continue;
+			}
+			possible_sets.push(set);
```
KrisXV, line 1260:
```suggestion
		const possibleSets = [];
		for (const set of sets) {
			if (teamDetails.teraBlast && set.role === "Tera Blast user") {
				continue;
			}
			possibleSets.push(set);
```

data/random-teams.ts
```diff
@@ -2020,394 +1281,132 @@ export class RandomTeams {
...
+				pokemonPool.push(pokemon);
+				if (!baseSpeciesPool.includes(species.baseSpecies)) baseSpeciesPool.push(species.baseSpecies);
 			}
-			pokemonPool.push(species.id);
 		}
-		return pokemonPool;
+		return [pokemonPool, baseSpeciesPool];
 	}
 
+	randomSets: AnyObject = require('./random-sets.json');
```
KrisXV, line 1407:
```suggestion

	// TODO: Make types for this
	randomSets: AnyObject = require('./random-sets.json');
```

data/random-teams.ts
```diff
@@ -2582,16 +1554,21 @@ export class RandomTeams {
...
 			if (set.moves.includes('stickyweb')) teamDetails.stickyWeb = 1;
 			if (set.moves.includes('toxicspikes')) teamDetails.toxicSpikes = 1;
 			if (set.moves.includes('defog')) teamDetails.defog = 1;
 			if (set.moves.includes('rapidspin')) teamDetails.rapidSpin = 1;
+			if (set.moves.includes('mortalspin')) teamDetails.rapidSpin = 1;
+			if (set.moves.includes('tidyup')) teamDetails.rapidSpin = 1;
 			if (set.moves.includes('auroraveil') || (set.moves.includes('reflect') && set.moves.includes('lightscreen'))) {
 				teamDetails.screens = 1;
 			}
+			if (set.role === ('Tera Blast user')) teamDetails.teraBlast = 1;
```
KrisXV, line 1571:
```suggestion
			if (set.role === 'Tera Blast user') teamDetails.teraBlast = 1;
```

KrisXV (changes requested):
Can you move the battle factory files/code into the gen8 mod?

livid-washed:
> Can you move the battle factory files/code into the gen8 mod?

Done, thanks for the review!

KrisXV (approved):

tools/build-utils.js
```diff
@@ -236,9 +236,10 @@ exports.transpile = (doForce, decl) => {
 	}
 
 	// sucrase doesn't copy JSON over, so we'll have to do it ourselves
-	copyOverDataJSON('bss-factory-sets.json');
+	copyOverDataJSON('random-sets.json');
 	copyOverDataJSON('cap-1v1-sets.json');
```
KrisXV, line 240:
```suggestion
	copyOverDataJSON('mods/gen8/cap-1v1-sets.json');
```
can you also move that file into the gen 8 mod?

KrisXV:
Swag

PR #9170, opened 2022-12-13
Make Maushold-Four the base forme

kittenchilly:
Maushold Family of Four is the base form in the games, while Family of Three is the rare form. Family of Three was erroneously made the base form on Showdown.
This change is made to be consistent with Dudunsparce, Sinistea, and Polteageist.


KrisXV:
There was nothing erroneous about this. Maushold Family of Three is the forme indexed first and listed as “Maushold-0” in every datamine so far. You can tell the difference because the family of three forme weighs less than the family of four.

PR #9171, opened 2022-12-13
Revert "Fix Copycat Dancer interaction (#9155)"

aviettran:
This seems to be causing problems worse than what it was trying to fix. I recommend reverting while I come up with a better fix.

https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-98

This reverts commit 95d1b924274b1c619de582447edb9847670359f1.


aviettran:
@pyuk-bot sure thing. Probably a good idea to leave my new test in place (although with a refactor, it's likely to work just fine).

PR #9173, opened 2022-12-13
Moderation: Correct /force notice for weeklocks

MathyFurret:


server/chat-commands/moderation.ts
```diff
@@ -869,7 +869,7 @@ export const commands: Chat.ChatCommands = {
 
 		if (!targetUser && !Punishments.search(userid).length && !force) {
 			return this.errorReply(
-				`User '${targetUsername}' not found. Use \`\`/forcelock\`\` if you need to to lock them anyway.`
+				`User '${targetUsername}' not found. Use \`\`/force${month ? 'month' : week ? 'week' : ''}lock\`\` if you need to to lock them anyway.`
```
AnnikaCodes, line 872:
```suggestion
				`User '${targetUsername}' not found. Use \`\`/force${month ? 'month' : (week ? 'week' : '')}lock\`\` if you need to to lock them anyway.`
```
This makes it more readable.

MathyFurret: I always prefered chaining ternaries without the parentheses, but I suppose that's fine

PR #9177, opened 2022-12-14
Fix Cloud Nine message

MathyFurret:


DaWoblefet:
Thanks Mathy! (for anyone reading, Air Lock was already correct)

PR #9180, opened 2022-12-14
Fix Order Up failure and Sheer Force mechanics

DaWoblefet:
Fixes Sheer Force-boosted Order Up behavior (should increase BP and not actually negate the stat boost lol) and fixes the behavior with failures (failing at roughly the same timing as Steel Beam would set its 50% HP flag or later will still give Dondozo the boost).

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -13194,11 +13194,10 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1, pulse: 1, mirror: 1},
-		onAfterMoveSecondarySelf(pokemon, target, move) {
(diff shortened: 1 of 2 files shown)


data/moves.ts
```diff
@@ -13211,7 +13210,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 				break;
 			}
 		},
-		secondary: null,
+		hasSheerForce: true,
```
Marty-D, line 13213:
```suggestion
		secondary: null,
		hasSheerForce: true,
```
Every move has `secondary` defined even if it's null, might as well leave it. Sheer Force still handles these weird cases appropriately.

DaWoblefet:
I'm dumb, it was master not linting, not this PR lol

PR #9182, opened 2022-12-14
Add tests for Gen 9 Random Battle

livid-washed:
This adds two tests for Gen 9 Random Battle:

1. Checks that every Pokemon gets a moveset with 4 moves (except Ditto). 
2. Checks that every move on every set is obtainable. Some moves are only obtainable with a non-default teamDetails (for example, Cryogonal only gets Haze if the team has a hazard remover), so we consider all possible relevant teamDetails. This looks like it would take a long time, but the move is usually obtained after only a few moveset generations, and as soon as we obtain the move, we can stop and move on to the next move. On my local machine this test took about 500ms, though I've marked it as slow to future-proof.

### DIFF
--- a/test/random-battles/gen9.js
+++ b/test/random-battles/gen9.js
@@ -43,10 +43,10 @@ describe('[Gen 9] Random Battle', () => {
 						const stealthRock = Math.floor(i / 2) % 2;
 						const stickyWeb = Math.floor(i / 4) % 2;
 						teamDetails = {'defog': defog, 'stealthRock': stealthRock, 'stickyWeb': stickyWeb};
-						for (let i = 0; i < rounds; i++) {
+						for (let j = 0; j < rounds; j++) {
 							// randomMoveset() deletes moves from the movePool, so create a copy
 							const movePoolCopy = [...movePool];
-							const teraType = teraTypes[i % teraTypes.length];
+							const teraType = teraTypes[j % teraTypes.length];
 							const moveSet = generator.randomMoveset(types, abilities, teamDetails, species, false, false, movePoolCopy, teraType, role);
 							if (moveSet.has(move)) {
 								moveObtained = true;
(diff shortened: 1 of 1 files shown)


test/random-battles/gen9.js
```diff
@@ -0,0 +1,63 @@
...
+			const sets = setsJSON[pokemon]["sets"];
+			const types = species.types;
+			const abilities = new Set(Object.values(species.abilities));
+			if (species.unreleasedHidden) abilities.delete(species.abilities.H);
+			for (const set of sets) {
+				const role = set.role;
+				const movePool = set.movepool.map(m => dex.moves.get(m).id);
+				const teraTypes = set.teraTypes;
+				// Every move in the movePool should be obtainable
+				for (const move of movePool) {
```
MathyFurret, line 37: I wonder if it would be more efficient to create a `Set` out of the movepool, remove the moves when they are found, and then break once the set is empty.

livid-washed: > I wonder if it would be more efficient to create a `Set` out of the movepool, remove the moves when they are found, and then break once the set is empty.

Good idea, I will try that

livid-washed: Seems to work :)

test/random-battles/gen9.js
```diff
@@ -0,0 +1,57 @@
...
+				const role = set.role;
+				const moves = new Set(set.movepool.map(m => dex.moves.get(m).id));
+				const teraTypes = set.teraTypes;
+				let teamDetails = {};
+				// Go through all possible teamDetails combinations, if necessary
+				for (let i = 0; i < 8; i++) {
+					const defog = i % 2;
+					const stealthRock = Math.floor(i / 2) % 2;
+					const stickyWeb = Math.floor(i / 4) % 2;
+					teamDetails = {'defog': defog, 'stealthRock': stealthRock, 'stickyWeb': stickyWeb};
```
mia-pi-git, line 42: Nit: ```suggestion
					teamDetails = {defog, stealthRock, stickyWeb};
```
You can just pass in variables to an object when initializing and the key names will be the same as the variable names. Since the names of the keys are the same as the variables, it's a bit cleaner to do it here.

MathyFurret: do we not have the eslint rule for this enabled?

livid-washed: I guess not. Thanks for that!

PR #9184, opened 2022-12-14
Update randomized formats

MathyFurret:
WIP -- basically, adding more gen 9 stuff, moving filler to the randomized section, removing/hiding old gen stuff


MathyFurret:
That should be everything @ACakeWearingAHat suggested, but I think it was brought up that if we are going to completely hide formats like this, then could we also add all old gen HC/CC?

config/formats.ts
```diff
@@ -2063,6 +2037,86 @@ export const Formats: FormatList = [
...
+		team: 'random',
+		ruleset: ['Obtainable', 'Same Type Clause', 'HP Percentage Mod', 'Cancel Mod', 'Sleep Clause Mod'],
+	},
+	{
+		name: "[Gen 9] Random Battle Mayhem",
+		desc: `[Gen 9] Random Battle with Team Preview and elements of Camomons, Inverse, Scalemons, and Shared Power.`,
+
+		mod: 'sharedpower',
+		team: 'random',
+		ruleset: ['[Gen 8] Random Battle', 'Team Preview', 'Camomons Mod', 'Inverse Mod', 'Scalemons Mod'],
```
Karthik99999, line 2053:
```suggestion
		ruleset: ['[Gen 9] Random Battle', 'Team Preview', 'Camomons Mod', 'Inverse Mod', 'Scalemons Mod'],
```

PR #9190, opened 2022-12-15
Datasearch: Show formes if sorted

MathyFurret:
https://www.smogon.com/forums/threads/dexsearch-should-show-extra-forms-if-a-sort-is-used.3709971/


AnnikaCodes:
LGTM

PR #9191, opened 2022-12-15
Computer-Generated Teams: Fix weakness counter incrementing too soon

adrivrie:
The weakness counter to prevent having 4+ mons weak to the same type is currently incrementing while still checking. This can cause types to be already incremented before the species is rejected because of another type, thus incorrectly increasing the counter. In the worst (rare!) case this leads to teams being generated with too many weaknesses as there is no more room to reject any.


AnnikaCodes:
Good catch!

PR #9194, opened 2022-12-15
Fix Shed Tail rounding down

adrivrie:
The amount of damage the user of Shed Tail takes is 50% of its max hp rounded up.

https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9436537

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -16458,13 +16458,13 @@ export const Moves: {[moveid: string]: MoveData} = {
 				this.add('-fail', source, 'move: Shed Tail');
 				return this.NOT_FAIL;
 			}
-			if (source.hp <= source.maxhp / 2 || source.maxhp === 1) { // Shedinja clause
+			if (source.hp <= (source.maxhp + 1) / 2) { // rounds up
 				this.add('-fail', source, 'move: Shed Tail', '[weak]');
 				return this.NOT_FAIL;
 			}
 		},
 		onHit(target) {
-			this.directDamage(target.maxhp / 2);
+			this.directDamage((target.maxhp + 1) / 2); // rounds up
 		},
 		self: {
 			onHit(source) {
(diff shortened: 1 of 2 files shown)


data/moves.ts
```diff
@@ -16458,13 +16458,13 @@ export const Moves: {[moveid: string]: MoveData} = {
 				this.add('-fail', source, 'move: Shed Tail');
 				return this.NOT_FAIL;
 			}
-			if (source.hp <= source.maxhp / 2 || source.maxhp === 1) { // Shedinja clause
+			if (source.hp <= (source.maxhp + 1) / 2) { // rounds up
```
Karthik99999, line 16461:
```suggestion
			if (source.hp <= Math.ceil(source.maxhp / 2) || source.maxhp === 1) { // Shedinja clause
```

data/moves.ts
```diff
@@ -16458,13 +16458,13 @@ export const Moves: {[moveid: string]: MoveData} = {
...
 				this.add('-fail', source, 'move: Shed Tail', '[weak]');
 				return this.NOT_FAIL;
 			}
 		},
 		onHit(target) {
-			this.directDamage(target.maxhp / 2);
+			this.directDamage((target.maxhp + 1) / 2); // rounds up
```
Karthik99999, line 16467:
```suggestion
			this.directDamage(Math.ceil(target.maxhp / 2));
```

data/moves.ts
adrivrie: Turned the +1 stuff (I was thinking of integer division) to Math.ceil, but I think the shedinja clause should not return.

Karthik99999:
The description should be updated to say "rounded half up"

DaWoblefet:
Thanks adrivrie!

PR #9197, opened 2022-12-16
Computer-Generated Teams: Prevent duplicate moves

adrivrie:
It's possible currently to get the same move twice on a mon. For example: https://replay.pokemonshowdown.com/gen9computergeneratedteams-1744282911

By deduplicating the movepool before moveset generation we can prevent this. If some but not all moves are duplicates, this also prevents duplicated moves from having artificially doubles weights.


adrivrie:
I don't think adding moves from different sources and then deduplicating the overlap is a very weird control flow. Would you prefer removing duplicates after/during every "source"?

PR #9199, opened 2022-12-17
Update DUbers and DUU threads and ladders

DaWoblefet:
DOU tier leaders approved removing the ladders for Gen 8 DUbers/DUU. I assume whether or not they should remain as challengeable formats will be determined whenever we do a format sweep. I also updated DUbers/DUU with the new Gen 9 threads and did some aliases.


KrisXV:
I was gonna leave DUU until gen 9 was out of alpha but considering they never do tiering anyway I guess this is fine

PR #9201, opened 2022-12-17
Gen 1: Fix Bide's implementation

livid-washed:
This patch fixes many issues with the current implementation of Bide:

- Bide's damage accumulation is now done correctly: the stored damage is incremented (by battle.lastDamage) every time the Bide user 'moves', not when the Bide user gets hit. This includes just before unleashing Bide, and does not include the turn that Bide is used. This means the stored damage can increase if the opponent switches or uses a move that doesn't reset lastDamage (e.g. Teleport).
- Bide's duration is paused if the user is asleep, frozen, partially trapped or flinching (and Bide's damage does not accumulate in that turn).
- Bide can now hit an enemy Substitute.
- Bide's stored damage will be reset to 0 if an enemy faints. This situation can result in a desync on cart, so we invoke Desync Clause Mod.
- Bide can now be disabled when active. If Bide is disabled while active, Bide's duration will be paused until the disable is over (similar to sleep).

Sources: https://www.smogon.com/rb/articles/rby_mechanics_guide, assembly, and in-game testing.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1670,6 +1670,10 @@ export const Formats: FormatList = [
 					// in gen 1, fainting skips the rest of the turn
 					// residuals don't exist in gen 1
 					this.queue.clear();
+					// Fainting clears accumulated Bide damage
+					for (const pokemon of this.getAllActive()) {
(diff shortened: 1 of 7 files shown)


AnnikaCodes:
Thanks!

PR #9203, opened 2022-12-18
Migrate to esbuild for compilation

mia-pi-git:
Discussed in #chat-dev. `esbuild` is much faster, needing only 4 seconds to compile, and should use significantly less memory than compiling on the fly. 
The structure I took was from chaos's suggestion - compiling the `.js` files right next to their `.ts` origin, which ensures we don't need any `.*-dist` hackery, only that we need to reference `.js` instead of `.ts` in some direct calls (like in SQL extension files). Changing imports isn't necessary since require prioritizes `.js` files.
Would be nice to get this out soon, given that hotpatches the last few days haven't been super stable.


pokemon-showdown
```diff
@@ -170,14 +137,14 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 	case 'simulate-battle':
 		{
 			sucraseBuild();
-			var BattleTextStream = require('./.sim-dist/battle-stream').BattleTextStream;
-			var Streams = require('./.lib-dist/streams');
+			const BattleTextStream = require('./sim/battle-stream.js').BattleTextStream;
+			var Streams = require('./lib/streams');
 			var stdin = Streams.stdin();
-			var stdout = Streams.stdout();
+			const stdout = Streams.stdout();
```
monsanto, line 143: In the future, please do these kinds of changes in a separate commit, it makes it harder for me to review the change.

pokemon-showdown
```diff
@@ -222,13 +189,13 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
 	case 'json-team':
 		{
 			sucraseBuild();
-			var Teams = require('./.sim-dist/teams').Teams;
-			var Streams = require('./.lib-dist/streams');
+			var Teams = require('./.sim/teams').Teams;
+			var Streams = require('./.lib/streams');
```
monsanto, line 193: This . error seems common, are you doing this by hand? Next time `perl -pi -e "s/.lib-dist/lib/g" **/*.ts` etc. I'm stopping the review here, but I may be able to look at it again after you double check your work.

pokemon-showdown
mia-pi-git: Oh, I think this was just eslint --fix. I'll revert that in a sec.

pokemon-showdown
mia-pi-git: Nope, did it with VSCode find and replace. Probably just missed a . in the replace. One sec.

mia-pi-git: Reverted and ran the proper find/replace this time.

pokemon-showdown
mia-pi-git: Reverted.

pokemon-showdown
```diff
@@ -1,11 +1,16 @@
...
+// failing before
+/* eslint-disable block-scoped-var, no-var, no-redeclare*/
 
 // Before running, ensure that the Typescript files have been compiled.
 //
 // Simply doing `require('../build')` here doesn't work because the `replace`
 // step of the build is asynchronous and we can't wait for it to finish easily
 // unless we shell out and use `execSync`. However, the heuristic of simply
-// checking for the presence of  './.sim-dist/dex' to determine whether a build
+// checking for the presence of './.sim/dex' to determine whether a build
```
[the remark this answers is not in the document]
mia-pi-git: See below - addressed / fixed.

mia-pi-git:
On further review, I changed the structure to compile to a single `dist` directory instead of alongside. Makes things less messy with similar lack of complication (just a few changes to `__dirname` stuff).

monsanto:
> On further review, I changed the structure to compile to a single `dist` directory instead of alongside. Makes things less messy with similar lack of complication (just a few changes to `__dirname` stuff).

You can't do this and use `ts-node` for development, as `__dirname` will be different. I don't see why it makes things messier.

DaWoblefet:
Reviewed this with Mia for Windows. The server builds and runs which is excellent. There's an issue where tests do not work if the path contains spaces (everything else has worked with spaces in the past). There's also a small TypeError when running any test:
``TypeError: Cannot read properties of undefined (reading 'name')
    at Object.loadTranslations (C:\Users\dawob\VS-Code-Projects\Pokemon-Showdown\dist\server\chat.js:1428:21)``

We should probably tell people when migrating to npm install before starting the server again. It will attempt to install dependencies and _sort of_ does it, but it seems a lot more consistent to run it before starting the server. We can also tell people they can delete any leftover .*-dist directories.

sim/dex-species.ts
```diff
@@ -237,6 +237,13 @@ export class Species extends BasicEffect implements Readonly<BasicEffect & Speci
 
 	constructor(data: AnyObject) {
 		super(data);
+		// FIXME: These shouldn't be necessary, Object.assign in the super should serve this purpose but doesn't
```
sirDonovan, line 240: There are two options to fix this described [here](https://github.com/evanw/esbuild/issues/2284#issuecomment-1141399397)

mia-pi-git: I never would have found this. Thanks so much!

mia-pi-git:
> There's also a small TypeError when running any test:
> `TypeError: Cannot read properties of undefined (reading 'name') at Object.loadTranslations (C:\Users\dawob\VS-Code-Projects\Pokemon-Showdown\dist\server\chat.js:1428:21)`

This was a bug - I fixed it.

mia-pi-git:
> We should probably tell people when migrating to `npm install` before starting the server again. It will attempt to install dependencies and _sort of_ does it, but it seems a lot more consistent to run it before starting the server.

This seems to be because it only installs some deps, instead of the full dev set. If you're running node build, you probably need dev dependencies, so I'm going to remove that part. 

> We can also tell people they can delete any leftover .*-dist directories.

I'll add a note if any of the dist dirs exist.

.eslintrc-no-types.json
```diff
@@ -11,6 +11,7 @@
     "logs/",
     "node_modules/",
     ".*-dist/",
```
monsanto, line 13: Can be removed?

lib/fs.ts
```diff
@@ -25,7 +25,9 @@ import * as fs from 'fs';
 import * as pathModule from 'path';
 import {ReadStream, WriteStream} from './streams';
 
-const ROOT_PATH = pathModule.resolve(__dirname, '..');
+// account for pwd/dist/lib
+const DIST = `${pathModule.sep}dist${pathModule.sep}`;
```
monsanto, line 29: Is this `sep` stuff necessary? Windows accepts `/` and I think `/` is used elsewhere in the code anyway

test/common.js
```diff
@@ -3,7 +3,7 @@
 const path = require('path');
 const fs = require('fs');
 const assert = require('./assert');
-const Sim = require('./../sim');
+const Sim = require('./../dist/sim');
```
monsanto, line 6: I thought this kind of thing wasn't going to be necessary?

mia-pi-git: Yeah, I thought it would be easier to do this than to copy over the entire test directory.

lib/fs.ts
mia-pi-git: In my tests with Wob, it was an issue. Not sure why it's not necessary other places - I think Node resolves / to path.sep?

.eslintrc-no-types.json
mia-pi-git: Missed this one - thanks!

lib/fs.ts
monsanto: Ok that's fine, we don't have to dwell on it given that we need to ship this but put a comment explaining the situation I guess

test/common.js
monsanto: Feels half-baked to me, but I don't want to hold up fixing the performance issues. A follow-up PR doing something nice with this would be appreciated, I should be able to `rg dist` and see nothing

mia-pi-git: OK.

monsanto:
It looks OK to me but I don't have time to test it myself. It sounds like you all have actually run it and it works fine, so if that is the case consider this approval from me.

.gitignore
```diff
@@ -23,6 +23,11 @@ tools/modlog/converter.js
 *.tsbuildinfo
 **/sourceMaps
 
+/**/*.js
+!test/**/*.js
```
AnnikaCodes, line 27: should this also exclude tools/**/*.js? there's a bunch of JS files in there

build
```diff
@@ -22,19 +22,15 @@ function shell(cmd) {
...
+		throw new Error("esbuild version too old");
 	}
 } catch (e) {
 	console.log('Installing dependencies...');
-	shell('npm install' + (decl ? '' : ' --production'));
-	force = true;
+	shell('npm install');
 }
-// for some reason, sucrase won't be requirable until a tick has passed
+// for some reason, esbuild won't be requirable until a tick has passed
```
AnnikaCodes, line 33: does this bug actually happen with esbuild?

mia-pi-git: This bug is with all node modules, not just sucrase.

.gitignore
mia-pi-git: Aren't there some compiled files in there? I manually excluded the non-compiled ones.

package.json
```diff
@@ -2,14 +2,13 @@
   "name": "pokemon-showdown",
   "description": "The server for the Pokémon Showdown battle simulator",
   "version": "0.11.7",
-  "main": ".sim-dist/index.js",
+  "main": "dist/sim/index.js",
```
AnnikaCodes, line 5: this is off-topic, but why is `sim/index` the entry point for what's described as the *server*?

mia-pi-git: It's for the node package, which is meant to export simulation tools.

pokemon-showdown
```diff
@@ -128,27 +95,27 @@ if (!process.argv[2] || /^[0-9]+$/.test(process.argv[2])) {
...
 		{
-			sucraseBuild();
-			var Teams = require('./.sim-dist/teams').Teams;
-			var TeamValidator = require('./.sim-dist/team-validator').TeamValidator;
+			ensureBuilt();
+			var Teams = require('./sim/teams.js').Teams;
+			var TeamValidator = require('./sim/team-validator.js').TeamValidator;
 			var validator = TeamValidator.get(process.argv[3]);
-			var Streams = require('./.lib-dist/streams');
+			var Streams = require('./lib/streams.js');
```
AnnikaCodes, line 118: Do you need to specify `.js` here?

server/chat-commands/admin.ts
```diff
@@ -570,6 +569,7 @@ export const commands: Chat.ChatCommands = {
 			return this.errorReply("Wait for /updateserver to finish before hotpatching.");
 		}
 
+		await this.parse(`/rebuild`);
```
AnnikaCodes, line 572: In the past, we've rebuilt after `/updateserver` as well because the server was configured to not rebuild on startup in order to save time during restarts. Is `esbuild` fast enough that it won't add much downtime to restarts, or should we continue building in `/updateserver` like we did when we used sucrase?

pokemon-showdown
mia-pi-git: Nope. This should be pointing to ./dist/lib, though.

server/chat-commands/admin.ts
mia-pi-git: The current build script takes 4 seconds to run. Up to you.

pokemon-showdown
mia-pi-git: Fixed, thanks.

server/chat-commands/admin.ts
```diff
@@ -885,8 +885,9 @@ export const commands: Chat.ChatCommands = {
 		const processes = new Map<string, ProcessData>();
 		const ramUnits = ["KiB", "MiB", "GiB", "TiB"];
 
+		const cwd = FS('.').path;
```
AnnikaCodes, line 888: Why do you sometimes use `FS('.').path` and sometimes `FS.ROOT_PATH`? They should both be the directory PS is in, right?

mia-pi-git: Every FS('.').path reference is from before I added FS.ROOT_PATH. I'll swap the references.

sim/dex-moves.ts
```diff
@@ -326,6 +326,8 @@ export interface ActiveMove extends MutableMove, RuinableMove {
 	typeChangerBoosted?: Effect;
 	willChangeForme?: boolean;
 	infiltrates?: boolean;
+	/** Should Order Up try its after-move boost effect?*/
+	orderUpBoost?: boolean;
```
AnnikaCodes, line 330: why is this here

test/common.js
AnnikaCodes: if we are copying over the test directory anyway, we might consider typescripting new test files...

mia-pi-git: We aren't copying over the test directory - that's why there are dist references. We could do that, I suppose, but that seems like something more for the followup.

tools/build-utils.js
```diff
@@ -1,248 +1,48 @@
...
+	);
+
+	const gitignored = ['./server/chat-plugins/private'];
+	for (const path of gitignored) {
+		if (fs.existsSync(path)) {
+			shell([
+				`find ${path}`, `grep "\\.[tj]s\\b"`, `grep -vF ".d.ts"`, `grep -v "node_modules"`, `grep -v global`,
+				`xargs node_modules/.bin/esbuild --log-level=error --outbase=${path} --outdir=./dist/${path} ` +
+				`--format=cjs --tsconfig=./tsconfig.json`,
+			].join(' | '), true);
```
AnnikaCodes, line 41: Has this been tested on Windows?

mia-pi-git: Yes, see above - Wob tested all of this on Windows with me.

AnnikaCodes (approved):
This looks good to me, but it should be tested before we deploy it (especially things like Trivia's SQL transactions and room-battle / battlesearch child processes).

sim/dex-moves.ts
mia-pi-git: No idea where this is, as dex-moves doesn't show on the current list of changed files. Resolving for now.

server/chat-commands/admin.ts
AnnikaCodes: Let's leave it like this for now. 4 seconds on your computer is probably 1-3 seconds on sim3, which may not be noticeable. If it is bothering people we should change it back.

package.json
AnnikaCodes: then the description is wrong

.gitignore
AnnikaCodes: Did you? tools/simulate/index.js and tools/team-generation-benchmark.js seem to still be gitignored?

server/chat-commands/admin.ts
mia-pi-git: OK.

Zarel:
I'm pretty unhappy with this, since it means we've lost line numbers for crashlogs.

Looking at benchmarks, swc isn't any slower than esbuild. We could have stayed on swc and just dropped ts-node, if that was the problem.

monsanto:
I'll also add that the transition here away from ts-node was a huge pain in the ass so it would be nice to not do it again. I imagine just changing esbuild -> SWC and keeping the directory structure the same would be less painful, but still...

mia-pi-git:
> I'll also add that the transition here away from ts-node was a huge pain in the ass so it would be nice to not do it again. I imagine just changing esbuild -> SWC and keeping the directory structure the same would be less painful, but still...

We shouldn't need to move to SWC - we just need to set up source map support properly (see #server-dev?)

Zarel:
We've been talking about the source map thing; Mia said it was working like over a week ago and we still keep getting crashlogs in JS files. Nothing we've tried has worked. Sooner or later we should just give up and do SWC.

Zarel:
@mia-pi-git you can have as much time as you need to get it working, but if you're stuck with no ideas for a long time, I think that's a good reason to switch back to swc.

monsanto:
I don't know what you think the difference would be. It's another transpiler that has to generate source map files, unless you're talking about restoring ts-node, which you definitely shouldn't do. If mia is having to tinker with runtime support for loading the source map files, then you're going to run into the exact same problems with SWC

Zarel:
We've fixed source map support so this is no longer an issue, but for the record, the nice thing about swc is that it supports preserving line numbers while transpiling, so you don't need source maps to know what line number a crash happened at.

PR #9204, opened 2022-12-18
Refactor Revival Blessing to be an action + fixes

MathyFurret:
* Pokemon revived in an active slot in Doubles should run a switch immediately
* Pokemon should not be able to make a move the same turn they fainted and were revived
* Fixes Revival Blessing choices not exporting to input log correctly

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2556,8 +2556,11 @@ export class Battle {
 		case 'revivalblessing':
 			action.pokemon.side.pokemonLeft++;
 			if (action.target.position < action.pokemon.side.active.length) {
-				action.target.isActive = true;
-				this.queue.cancelMove(action.target);
+				this.queue.addChoice({
+					choice: 'instaswitch',
+					pokemon: action.target,
+					target: action.target,
+				});
 			}
 			action.target.fainted = false;
 			action.target.faintQueued = false;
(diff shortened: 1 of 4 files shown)


sim/battle.ts
```diff
@@ -2553,6 +2553,18 @@ export class Battle {
 				}
 			}
 			break;
+		case 'revivalblessing':
+			action.pokemon.side.pokemonLeft++;
+			if (action.target.position < action.pokemon.side.active.length) action.target.isActive = true;
```
Karthik99999, line 2558:
```suggestion
			if (action.target.position < action.pokemon.side.active.length) {
						action.target.isActive = true;
						this.queue.cancelMove(action.target);
			}
```
Probably good to fix this bug while we're at it.

Karthik99999: Actually found another bug with this, in that the Pokemon should actually be switching in rather than being set to active. I tested this with an intimidate pokemon, and reviving it while it was in active slot made intimidate reactivate.

Karthik99999: It also happens after the revive message is shown so the placement of this needs to be moved.

MathyFurret: Oh I know what you're talking about. The same thing happens with Revives in previous gens too (going back to at least gen 5 where it was possible to use one via Wonder Launcher). I will implement, thanks

MathyFurret:
Works great under the hood, but a client change may be needed to fix the switch animation.

PR #9205, opened 2022-12-18
Gen IX: Fix Curse targeting behavior

DaWoblefet:
In Gen 9, Ghost-type Curse can now be targeted at either opponent or an ally. However, if Curse is targeted at an ally, it instead selects either opponent randomly instead. I used the ``ModifyTarget`` handler to accomplish this, but I'm not sure if I've done everything quite right with it, so reviews are appreciated.

### DIFF
--- a/test/sim/moves/curse.js
+++ b/test/sim/moves/curse.js
@@ -30,17 +30,6 @@ describe('Curse', function () {
 		assert.equal(battle.p1.active[0].getMoveRequestData().moves[0].target, 'normal');
 	});
 
-	it(`should request the Ghost target after the user becomes Ghost`, function () {
-		battle = common.createBattle([[
-			{species: 'Rapidash', moves: ['curse']},
-		], [
-			{species: 'Trevenant', moves: ['trickortreat']},
-		]]);
-		assert.equal(battle.p1.active[0].getMoveRequestData().moves[0].target, 'self');
-		battle.makeChoices();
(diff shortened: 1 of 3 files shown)


test/sim/moves/curse.js
```diff
@@ -10,82 +10,135 @@ describe('Curse', function () {
...
+		assert.fullHP(wynaut);
+		assert(caterpie.maxhp !== caterpie.hp || metapod.maxhp !== metapod.hp, `Either Caterpie or Metapod should have lost HP from Curse`);
+	});
 
-		battle.makeChoices('move spite', 'auto');
-		// Check residual damage
-		assert.equal(hps[0], battle.p1.active[0].hp); // Curse user unaffected
-		assert.notEqual(hps[1], battle.p2.active[0].hp); // Curse residual damage
+	it(`[Gen 7] should target the ally if the target is an ally`, function () {
+		battle = common.gen(7).createBattle({gameType: 'doubles'}, [[
```
urkerab, line 131: If you're reverting the behaviour in the Gen 8 mod then surely that's the one you should use for testing?

urkerab: (Or should the behaviour actually have been reverted in the Gen 7 mod?)

DaWoblefet: Gen 8 Curse can't target an ally in the first place. I took off the inheritance in the Gen 8 mod because I think it makes more sense to drop it there rather than 7 in case there's some jank with LightningRod Electrify Curse targeting or something that I haven't tested in 8.

PR #9207, opened 2022-12-18
Random Battles updates

ACakeWearingAHat:
-Several minor updates to movesets and tera types
-Stop enforcing Leaf Storm on Grass-types, instead forcing a random grass move if the user could possibly get Leaf Storm and does not already have a grass move (It's a lot simpler than it sounds)
-Prevent Air Balloon Rotom-Heat
-whoops arceus-poison was level 80 haha
-Prevent Armor Tail
-Prevent Choice Scarf Zoroark-Hisui
-Add a Swords Dance set to Lokix
-Rework Sunflora to a Choice Specs user so it can maybe be less trash

This update to sets was discussed by and approved by the Random Battles roomstaff.

### DIFF
--- a/data/random-sets.json
+++ b/data/random-sets.json
@@ -585,8 +585,8 @@
         "sets": [
             {
                 "role": "Wallbreaker",
-                "movepool": ["Dazzling Gleam", "Earth Power", "Solar Beam", "Sunny Day"],
-                "teraTypes": ["Grass", "Ground"]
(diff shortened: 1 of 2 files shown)


livid-washed:
Looks good to me :)

MathyFurret (approved):

mia-pi-git:
Thanks!

PR #9208, opened 2022-12-19
Random Battles updates

ACakeWearingAHat:
- Pikachu will now always get Surf
- Annihilape will no longer get Life Orb
- Clear Smog and Toxic cannot be on the same set; this is important for Amoonguss
- Normal-types will always get a Normal-type STAB attack if they can roll Future Sight; this is important for Farigiraf
- Iron Bundle is no longer capable of using Tera Blast
- Farigiraf now has a Future Sight + WishTect set and an AV set.
- Ting-Lu now has a separate set that can run Choice Band or Assault Vest.
- Gholdengo will now always run Make It Rain
- Several other misc. set changes

These changes were approved by the Random Battles room staff.

### DIFF
--- a/data/random-sets.json
+++ b/data/random-sets.json
@@ -19,8 +19,8 @@
         "sets": [
             {
                 "role": "Fast Attacker",
-                "movepool": ["Nuzzle", "Play Rough", "Surf", "Volt Switch", "Volt Tackle"],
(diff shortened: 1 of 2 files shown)


MathyFurret (approved):

mia-pi-git:
Thanks!

PR #9216, opened 2022-12-21
Random Battles updates

ACakeWearingAHat:
-Corrosion will now generate specifically when Glimmora has Toxic and lacks coverage for Steel-types.
-Pelipper now has a Choice Specs set.
-Espathra's Tera Blast set is now Life Orb Roost + Tera Fire.
-Grafaiai's offensive sets were combined into one movepool, Super Fang cannot generate with Switcheroo, and an Encore + Parting Shot set was added.
-Several other minor updates to sets and tera types.
-Cud Chew code was made more simple and futureproofed.

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -881,7 +881,7 @@ export class RandomTeams {
 		case 'Intimidate':
 			if (abilities.has('Hustle')) return true;
 			if (abilities.has('Sheer Force') && !!counter.get('sheerforce')) return true;
-			return (abilities.has('Stakeout') || moves.has('substitute'));
+			return (abilities.has('Stakeout'));
 		case 'Iron Fist':
 			return !counter.ironFist;
 		case 'Justified':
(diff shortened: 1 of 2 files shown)


data/random-teams.ts
```diff
@@ -500,8 +500,8 @@ export class RandomTeams {
 		// This space reserved for assorted hardcodes that otherwise make little sense out of context
 		// Landorus
 		this.incompatibleMoves(moves, movePool, 'nastyplot', 'rockslide');
-		// Persian and Seviper
-		this.incompatibleMoves(moves, movePool, 'switcheroo', ['fakeout', 'suckerpunch']);
+		// Persian and Grafaiai
+		this.incompatibleMoves(moves, movePool, 'switcheroo', ['fakeout', 'superfang']);
```
adrivrie, line 504: I would make this `['trick', 'switcheroo']` to make the rule more general, despite it not actually occurring right now

adrivrie (approved):

data/random-teams.ts
ACakeWearingAHat: Once we have reason to remove it from the hardcodes section, we'll do that. Thank you, though!

livid-washed:
I approve of these changes

PR #9217, opened 2022-12-22
Gen 3: Fix Brick Break ignoring Ghost immunity

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9443365


livid-washed:
You could add `ignoreImmunity: true,` to gen3/moves.ts and remove `return false;` from onTryHit(). Gen 3 runs the TryHit event immediately before checking type immunity.

Marty-D:
Thanks, but the original was completely wrong so I've reverted it.

PR #9218, opened 2022-12-22
Gen IX: Really fix Curse targeting behavior

DaWoblefet:
I am like, 90% sure the only reason this was crashing was because of the restart (restarts load saved battles with new mechanics), but in the meantime Karthik pointed out it's better to put the target change in ``onModifyMove`` anyway.


DaWoblefet:
Neither do I, but minimally I can say all the battles that crashed imported just fine with no crashes on this PR.

PR #9223, opened 2022-12-23
Add new testing API extending /evalbattle scope

Slayer95:
- Support for ``p1``, ``p2``, ``p1a``, ``p1b``...
- Support for fast access to specific Pokémon via their species ids.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -14,6 +14,14 @@ function capitalize(word) {
 }
 
 /**
+ * @typedef TestContext
+ * @prop {Sim.Battle} battle - Simulated battle
+ * @prop {Sim.Side} [key: SideID] - Sides: p1, p2...
+ * @prop {Sim.Pokemon} [key: PokemonSlot] - Pokemon: p1a, p1b, p2a...
+ * @prop {Sim.Pokemon} [key: string]
+ */
+
+/**
  * The default random number generator seed used if one is not given.
  */
 const DEFAULT_SEED = [0x09917, 0x06924, 0x0e1c8, 0x06af0];
@@ -110,6 +118,34 @@ class TestTools {
 		}
 
 		return new Sim.Battle(battleOptions);
+	}
+
+	/**
+	 * Creates a new Battle and returns it.
+	 *
+	 * @param {Object} [options]
+	 * @param {Team[]} [teams]
+	 * @returns {TestContext} Test context.
+	 */
+	testCtx(options, teams) {
+		const battle = this.createBattle(options, teams);
+		const ctx = {battle};
+		const bySpecies = new Map();
+		for (const side of battle.sides) {
+			ctx[side.id] = side;
+			for (const [i, pokemon] of side.pokemon.entries()) {
+				ctx[`${side.id}${String.fromCharCode(97 + i)}`] = pokemon;
+				if (bySpecies.has(pokemon.species.id)) {
+					bySpecies.set(pokemon.species.id, null);
+				} else {
+					bySpecies.set(pokemon.species.id, pokemon);
+				}
+			}
+		}
+		for (const [speciesId, pokemon] of bySpecies) {
+			if (pokemon) ctx[speciesId] = pokemon;
+		}
+		return ctx;
 	}
 
 	/**
--- a/test/sim/abilities/berserk.js
+++ b/test/sim/abilities/berserk.js
@@ -3,41 +3,39 @@
 const assert = require('./../../assert');
 const common = require('./../../common');
 
-let battle;
+let ctx;
 
 describe('Berserk', function () {
 	afterEach(function () {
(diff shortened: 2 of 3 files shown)


MathyFurret:
This looks nice but one suggestion: the Pokemon identifiers shouldn't have the forme names

test/common.js
```diff
@@ -112,6 +120,34 @@ class TestTools {
...
 	}
 
+	/**
+	 * Creates a new Battle and returns it.
+	 *
+	 * @param {Object} [options]
+	 * @param {Team[]} [teams]
+	 * @returns {TestContext} Test context.
+	 */
+	testCtx(options, teams) {
```
mia-pi-git, line 130:
```suggestion
	testCtx(options, teams) {
```
Nit: `testContext` instead of `testCtx` for the function name is clearer and more consistent with the rest of the codebase. I'd lean towards `getTestContext` given it's a function, though.

Slayer95:
Updated per review

DaWoblefet:
I have mixed feelings.

On readability:
I like the nice shorthand with p2a and mon names. It makes it easier to follow along during the logic of the test. I do think it makes the initial declaration a bit worse - the Dancer examples you converted are pretty long now. If you're checking multiple Pokemon in a non-singles format with a hardcoded RNG seed you may well have to scroll to read everything.

On writability:
This is more a logistical thing, but I'm not a huge fan of adding another "new way we write tests" in our already split tests codebase. It's of course not your job to convert thousands of tests.

In the transition period, it introduces a real inconvenience when working with a tests file that already exists. Normally, when folks write tests, they simply copy another test in the correct format and just edit that. But then you can't have both this new style and the old style of tests running in the same file, because battle means something different. This would require a developer submitting a fix for one mechanic to update the formatting of potentially dozens more in the same file to use the new context-based format (minimally, you can't just drop a ``let ctx;`` and ``ctx.battle.destroy()`` and expect it to behave).

On performance:
Is there any significant difference? Just benchmarking what you've currently done before and after is fine. A full mocha already takes a decent bit to run and I'd rather us steer away from performance loss unless it's a significant benefit.

Slayer95:
Thanks for your thorough review, @DaWoblefet 

> I like the nice shorthand with p2a and mon names [...] I do think it makes the initial declaration a bit worse

I noticed the recurrent pattern of declaring ``const [species] = battle.pX.active[?]`` which many contributors follow, so the destructuring makes it more concise.

> the Dancer examples you converted are pretty long now.

Good point. Then it's probably better to avoid renaming the destructured Pokémon positions. Then the reader will not actually need to read the first line and _just know_ what the variables mean, which is the main objective.

> But then you can't have both this new style and the old style of tests running in the same file, because battle means something different.

In fact, it's the same type of ``battle``, just with different scopes. Therefore, there wouldn't be any problem keeping both styles in a single file.

> It's of course not your job to convert thousands of tests.

Though I may just migrate them all automatically, so there won't be any transition. For free, really. Of course, that's assuming you all like this API. I would've already done had I not been afraid it would make the diff huge.

> On performance:
Is there any significant difference? Just benchmarking what you've currently done before and after is fin

I'll have a look.

DaWoblefet:
Just to clarify, what is actually needed then to run context-based tests simultaneously with the current format? I kept getting issues related to battles being unable to be destroyed. For my testing, I just copy-pasted one of the new Dancer tests into the Room Service tests file.

If you just copy-paste the test straight-up:
<details>
  1) Room Service
       should only copy dance moves used by other Pokemon:
     ReferenceError: ctx is not defined
      at Context.<anonymous> (test\sim\items\roomservice.js:48:34)
      at process.processImmediate (node:internal/timers:471:21)

  2) Room Service
       "after each" hook for "should only copy dance moves used by other Pokemon":
     TypeError: Cannot read properties of null (reading 'destroy')
      at Battle.destroy (dist\sim\battle.js:2546:16)
      at Context.<anonymous> (test\sim\items\roomservice.js:10:10)
      at process.processImmediate (node:internal/timers:471:21)
</details>
If you add ``let ctx;`` and ``ctx.battle.destroy();``
<details>
  1) Room Service
       "after each" hook for "should activate when Trick Room is set":
     TypeError: Cannot read properties of undefined (reading 'battle')
      at Context.<anonymous> (test\sim\items\roomservice.js:12:7)
      at process.processImmediate (node:internal/timers:471:21)
</details>

Whatever I am doing wrong is likely what the novice developer would be doing wrong.

Slayer95:
Oh I see! Yea, sorry. It turns out that the ``afterEach`` hooks would then need to be updated with a conditional clause:``if (battle) battle.destroy()`` and ``if (ctx) ctx.battle.destroy()``

It would be for the best if an agreement could be reached to migrate the tests. I'll post back when I do the benchmarks.

PR #9226, opened 2022-12-23
Fix Hyper Drill breaking protect instead of ignoring it

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9439330


DaWoblefet:
Thanks Karthik!

PR #9227, opened 2022-12-24
Implement proper Origin forme mechanics

Karthik99999:
- Added the Adamant Crystal, Lustrous Globe, and Griseous Core, which are all required to be held by their respective Origin formes
- The Griseous Orb is now Identical to the other Orbs, providing a power boost to base Giratina and is able to be knocked off

This definitely would affect metagames like randbats, which currently has the origin formes holding items like Choice Specs and Choice Scarf, so this would be a pretty big nerf to them. ~~This is also a direct nerf to Giratina Origin, as it can no longer get the power boost from Griseous Orb while still being locked to one item.~~ I don't think there's much of an argument to not implement this since we usually try to follow cart mechanics as much as possible, but I just thought that some might not want this change with it affecting certain metagames too much despite being mechanics that could (with a very low chance) be changed with an update or DLC.

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -64,9 +64,6 @@ export const Items: {[itemid: string]: ItemData} = {
 	adamantcrystal: {
 		name: "Adamant Crystal",
 		spritenum: 0, // TODO
-		fling: {
-			basePower: 60,
-		},
 		onTakeItem(item, pokemon, source) {
 			if ((source && source.baseSpecies.num === 483) || pokemon.baseSpecies.num === 483) {
 				return false;
@@ -2333,9 +2330,6 @@ export const Items: {[itemid: string]: ItemData} = {
 	griseouscore: {
 		name: "Griseous Core",
 		spritenum: 0, // TODO
-		fling: {
-			basePower: 60,
-		},
 		onTakeItem(item, pokemon, source) {
 			if ((source && source.baseSpecies.num === 487) || pokemon.baseSpecies.num === 487) {
 				return false;
@@ -3229,9 +3223,6 @@ export const Items: {[itemid: string]: ItemData} = {
 	lustrousglobe: {
 		name: "Lustrous Globe",
 		spritenum: 0, // TODO
-		fling: {
-			basePower: 60,
-		},
 		onTakeItem(item, pokemon, source) {
 			if ((source && source.baseSpecies.num === 484) || pokemon.baseSpecies.num === 484) {
 				return false;
--- a/data/items.ts
+++ b/data/items.ts
@@ -61,6 +61,24 @@ export const Items: {[itemid: string]: ItemData} = {
 		num: 545,
 		gen: 5,
 	},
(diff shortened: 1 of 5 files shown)


data/items.ts
```diff
@@ -61,6 +61,21 @@ export const Items: {[itemid: string]: ItemData} = {
 		num: 545,
 		gen: 5,
 	},
+	adamantcrystal: {
+		name: "Adamant Crystal",
+		spritenum: 0, // TODO
+		onTakeItem(item, pokemon, source) {
+			if ((source && source.baseSpecies.num === 483) || pokemon.baseSpecies.num === 483) {
```
KrisXV, line 68:
```suggestion
			if (source?.baseSpecies.num === 483 || pokemon.baseSpecies.num === 483) {
```

data/items.ts
```diff
@@ -2312,6 +2327,21 @@ export const Items: {[itemid: string]: ItemData} = {
 		num: 286,
 		gen: 4,
 	},
+	griseouscore: {
+		name: "Griseous Core",
+		spritenum: 0, // TODO
```
KrisXV, line 2332: It's not accurate but maybe we could use the existing orb item sprites as placeholders?

data/items.ts
```diff
@@ -2312,6 +2327,21 @@ export const Items: {[itemid: string]: ItemData} = {
...
+		onTakeItem(item, pokemon, source) {
+			if ((source && source.baseSpecies.num === 487) || pokemon.baseSpecies.num === 487) {
```
KrisXV, line 2334:
```suggestion
			if (source?.baseSpecies.num === 487 || pokemon.baseSpecies.num === 487) {
```

data/items.ts
```diff
@@ -3197,6 +3220,21 @@ export const Items: {[itemid: string]: ItemData} = {
 		gen: 2,
 		isPokeball: true,
 	},
+	lustrousglobe: {
+		name: "Lustrous Globe",
+		spritenum: 0, // TODO
+		onTakeItem(item, pokemon, source) {
+			if ((source && source.baseSpecies.num === 484) || pokemon.baseSpecies.num === 484) {
```
KrisXV, line 3227:
```suggestion
			if (source?.baseSpecies.num === 484 || pokemon.baseSpecies.num === 484) {
```

KrisXV (approved):

Karthik99999:
> I haven’t seen any confirmation one way or the other that the transformation items do or don’t have the power boost.

Just tested this in game, and I actually did get damage rolls that showed that the origin items do have the damage boost that the orbs get. Will go ahead and implement that, thank you for bringing that up.

data/items.ts
Karthik99999: Yeah that seems fine. I'll leave the TODO comment so we can remember to change it.

DaWoblefet:
Thanks Karthik!

PR #9228, opened 2022-12-25
Build: Generate sourcemaps

MathyFurret:
This is needed for debuggers to work

LMK if this should be made optional via environment variable


AnnikaCodes:
Thanks!

PR #9229, opened 2022-12-25
Show Terastallization in switch details

MathyFurret:
Requires client update


MathyFurret:
It may be redundant to include the type, but I will probably keep it just for convenience with Illusion shenanigans

PR #9232, opened 2022-12-26
Tournaments: Shuffle loser bracket seeds

MathyFurret:
Currently, the losers' bracket is seeded in the same order as the winners' bracket. This too often results in players being matched up with the same opponent they had in the winners' bracket. Shuffling the seeds at each level will keep this from happening as often.

### DIFF
--- a/server/tournaments/generator-elimination.ts
+++ b/server/tournaments/generator-elimination.ts
@@ -245,6 +245,7 @@ export class Elimination {
 			for (const depth in matchesByDepth) {
 				if (depth === '0') continue;
 				const matchesThisDepth = matchesByDepth[depth];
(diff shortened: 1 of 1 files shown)


MathyFurret:
Currently deciding if i should figure out how to test this in a unit test environment or just invite people to my side server to test. I can only open so many incognito windows before i go insane

AnnikaCodes:
@sirDonovan might know about testing tournaments? I can deploy this on staging.psim.us if you want.

sirDonovan:
In the past I used a private server to test changes with multiple players when needed so that would be helpful @AnnikaCodes

MathyFurret:
Tried to recruit people from Main's Lobby to test but this doesn't seem conclusive
image

PR #9239, opened 2022-12-27
Don't reveal Dudunsparce forme on Team Preview

Karthik99999:
~~I know this literally does not matter~~
The Dudunsparce formes share a menu sprite, so you shouldn't be able to tell which one the opponent brought.


KrisXV:
ty

PR #9240, opened 2022-12-27
Gen 5: Fix Umbreon learnset

MathyFurret:


Marty-D:
Just to note, I think this applies to every Pokemon; presumably moves already learned in BW were not updated for B2W2.

PR #9242, opened 2022-12-27
Fix mocha for paths with whitespace

adrivrie:
Adding quotes around the path in a command line command make it work with whitespace in the path.

I thought "User Files" was a default directory in Windows, but now I'm thinking I may have put the whitespace there myself which would make me very silly. Regardless, this fix should help some people (including me).


AnnikaCodes:
Thanks!

PR #9243, opened 2022-12-27
Gen 1: Implement Dig/Fly invulnerability glitch, partially fix charging moves and Mirror Move

livid-washed:
This patch does a few things:
- pokemon.lastMove is reset to null whenever a switch is made. This ensures that Mirror Move fails when used by a newly switched in Pokemon, against an enemy that has not attacked since the switch.
- The charging turn of a two-turn move does not update pokemon.lastMove. In particular, Mirror Move will use the previous move made by the enemy (or fail if there was no previous move). If the enemy's Metronome calls a two-turn move, Mirror Move will call Metronome.
- Two-turn moves are paused if the Pokemon using the two-turn move is put to sleep, frozen, flinched, or partially trapped. The Pokemon cannot switch.
- Two-turn moves end if they fail because they are disabled.
- Two-turn moves only use PP in the attacking turn, not the charging turn.
- If a two-turn move is called by Metronome or Mirror Move, the calling move uses PP only in the attacking turn.
- Introduce a new 'invulnerability' (semi-invulnerability) volatile for Gen 1, associated with Dig and Fly.
- Dig/Fly invulnerability glitch: if a Pokemon gets fully paralyzed while semi-invulnerable, it will remain semi-invulnerable until it switches out or fully executes a Dig/Fly.

This should close the other PR relating to Dig/Fly glitch: https://github.com/smogon/pokemon-showdown/pull/8882/files

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,5 +1,5 @@
 {
-  "spec": ["test/main.js", "test/lib/**/*.js", "test/server/**/*.js", "test/sim/**/*.js", "test/tools/**/*.js", "test/random-battles/**/*.js"],
+  "spec": ["test/main.js", "test/sim/moves/mirrormove.js", "test/sim/misc/twoturnmoves.js"],
   "grep": "^((?!\\(slow\\)).)*$",
   "reporter": "dot",
   "ui": "bdd",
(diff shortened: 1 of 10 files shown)


AnnikaCodes:
Thanks!

PR #9245, opened 2022-12-28
Fix Glaive Rush again

Karthik99999:
Hopefully this is the last time we fix this


DaWoblefet:
Thanks Karthik (and Marty!)

PR #9250, opened 2022-12-29
Fix Room Service boost source

Karthik99999:
Probably would be better to refactor the pseudo weather event to use `eachEvent` but I'm lazy


DaWoblefet:
Thanks Karthik!

PR #9256, opened 2022-12-30
Gen 6-8 Random Battles: Fix unused moves

MathyFurret:
discussed w/ randbats team, will get final review here


ACakeWearingAHat (approved):
We discussed this pretty in depth in the Randscord. I approve of the result after giving it a good look-over.

PR #9261, opened 2022-12-30
Old Gens: Clean up non-occurring events and conditions

livid-washed:
This patch removes sections of code that are irrelevant to gens 1-2.

Removed the following:
All gens: move.negateSecondary and move.forceStatus (completely unused)
Gens 1-3: Pseudo Weather (Trick Room, etc)
Gen 1-2:
- TryHitField event (Mud Sport/Water Sport), TryHitSide event (Magic Coat and abilities)
- TrySecondaryHit event (unused)
- DragOut event (ingrain and abilities)
- CriticalHit event (lucky chant and abilities)
- BasePower event (future gens where base power gets adjusted by items/abilities/Charge etc)
- move.overrideOffensivePokemon and move.overrideOffensiveStat (moves like Foul Play, Psyshock, Body Press)
- AfterEachBoost and AfterBoost events (items like Mirror Herb and abilities)
- move.willCrit is either false or undefined for all moves
- more.ignoreAccuracy, move.ignoreEvasion

Gen 1:
- move.beforeMoveCallback (Bide which is now implemented correctly, focus punch)
- ModifyMove event (many things but not relevant to gen 1, e.g. Weather Ball)
- move.heal (healing is coded manually in moves.ts)
- move.sideCondition (Safeguard, etc)
- move.selfSwitch (Baton Pass, U-turn, etc)
- move.basePowerCallback (Rollout, etc)


Marty-D:
I'm not one to care about supporting past gen functionality of things that only exist in future gens, but certainly other people do. Are these things already non-functional in the respective Custom Game or does this render them non-functional?

livid-washed:
> I'm not one to care about supporting past gen functionality of things that only exist in future gens, but certainly other people do. Are these things already non-functional in the respective Custom Game or does this render them non-functional?

Hi Marty,

I am mainly doing this to make it much easier to fix actual Gen 1/2 bugs that can occur in actual battles. Having the non-occurring clutter around makes it significantly harder to work out where the important code belongs. 

To answer your question, there are a few future-gen things that 'work' in Gen 1/2 Custom Game that would be made non-functional by this PR, though it is kind of a mixed bag. Whether future-gen effects work or not is pretty arbitrary, and even the effects that exist may not fully work. For example, Gravity, which this PR removes, exists in Gen 1 Custom Game at the moment, but it doesn't increase the accuracy of moves, though it does let moves like Earthquake hit Flying-types. On the other hand, rain (or other weather) does not work at all in Gen 1 Custom Game. In general, there isn't really any reason why effect A should exist and effect B shouldn't. It seems more consistent to have them both not work.

Even if we care about using future-gen moves in Gen 1, I think it would make more sense to only include effects that would also occur in Gen 1. For example, future-gen recharging moves like Frenzy Plant work the same way as Hyper Beam and won't change with this patch.

Marty-D:
If we go in this direction I would rather just prevent future-gen stuff from being used in past gens at all, which guards against any possible crashes and other stuff no one should have to think about when working on current gen things.

Either way, I hesitate to merge large changes like this around SPL time, as we have been burned several times in other years with problems in tour matches after certain past gen updates.

livid-washed:
Might come back to this in the future, but gonna close for now

PR #9266, opened 2022-12-31
Implement opt-in Open Team Sheets

DaWoblefet:
WIP with major help from Karthik.

The announcement of open team sheets in VGC play was met with mixed reception. For various reasons, some players prefer closed team sheets, and some prefer open team sheets. It is undesirable to make two ladders, however, as that is likely to fracture the playerbase and probably be lopsided in favor of closed team sheets. The current solution is to make opting into OTS very easy, while also not being forced to reveal the player's OTS to spectators or in replays. Players can then choose to agree or not to agree to OTS.

The main issue with current implementation is that player-specific information (e.g. specific player damage, or in this case, uhtml with team sheet information) is not retained if a player happens to refresh during a game because of e.g. a bad connection. We may opt to refactor so the acceptance button prompts are also visible to spectators and just don't do anything when clicked, then change the uhtml "View Open Team Sheets" when both players have accepted.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -235,7 +235,7 @@ export const Formats: FormatList = [
 
 		mod: 'gen9',
 		gameType: 'doubles',
-		ruleset: ['Flat Rules', '!! Adjust Level = 50', 'Paldea Pokedex', 'Min Source Gen = 9', 'VGC Timer'],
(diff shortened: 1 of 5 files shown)


Karthik99999:
I overrode `onJoin` in `RoomGame` due to `|splits|` not getting properly resent after DCs. The "right" thing to do would be to fix that issue but we didn't want to indefinitely block this feature because of that.

PR #9268, opened 2022-12-31
Gen 8 Battle Factory: Remove banned Pokemon

adrivrie:
Magby has been banned from LC, so removing it from BF.


adrivrie:
It turns out Scrafty and Indeedee-f were also removed from their BF tier, so removing those sets as well.

KrisXV:
ty

PR #9269, opened 2022-12-31
Fix private logs not getting sent when connecting to GameRoom

Karthik99999:


DaWoblefet:
This doesn't fix private information not showing up on refreshing (because at the time the log is sent, the user is a guest account), but it does load private information if the whole tab is closed, Showdown's loaded up again, the player logs in, then re-joins the game.

DaWoblefet:
Thanks Karthik!

PR #9270, opened 2022-12-31
Improvements to room FAQs and repeats

dot-Comfey:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9451198 I tested this on a local server running the most recent version of Pokemon Showdown, so while I never caused a crash on the main server, the bug still exists there.

- Removed /rfaq alias check when setting a topic to repeat. If a /rfaq alias is set to repeat, the text of the topic the alias is set for is displayed, rather than the topic's ID.
- If a /rfaq is deleted, repeats linked to it or one of its aliases will also be deleted, preventing a crash caused by trying to display a nonexistent topic. Originally supported for just topics, but support for aliases was added, which the lack of support for is what caused crashes.
- A separate problem, but the ability to set a topic as an alias of itself has been removed. Because this is possible, you can have a topic which is registered in the server, but does not appear at all if you type /rfaq.

While it seems like it wasn't intended for /rfaq aliases to be used for repeats, since there's a way to get around it, might as well just work with it instead of trying to prevent it.

### DIFF
--- a/server/chat-plugins/room-faqs.ts
+++ b/server/chat-plugins/room-faqs.ts
@@ -111,16 +111,24 @@ export const commands: Chat.ChatCommands = {
 		if (!(roomFaqs[room.roomid] && roomFaqs[room.roomid][topic])) return this.errorReply("Invalid topic.");
 		if (
 			room.settings.repeats?.length &&
-			room.settings.repeats.filter(x => x.faq && x.id === (getAlias(room!.roomid, topic) || topic)).length
+			room.settings.repeats.filter(x => x.faq && x.id === topic).length
 		) {
-			this.parse(`/msgroom ${room.roomid},/removerepeat ${getAlias(room.roomid, topic) || topic}`);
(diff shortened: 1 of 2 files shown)


server/chat-plugins/room-faqs.ts
```diff
@@ -111,16 +111,25 @@ export const commands: Chat.ChatCommands = {
...
+				delete roomFaqs[room!.roomid][val];
+				if (
+					room?.settings.repeats?.length &&
+					room.settings.repeats.filter(x => x.faq && x.id === val).length
+				) {
+					this.parse(`/msgroom ${room.roomid},/removerepeat ${val}`);
+				}
+				return true;
+			}
 		);
```
MathyFurret, line 131: This should now be using a for-of loop instead of map, that's what the linter error was meant to encourage in the first place.

mia-pi-git:
SGTM. Thanks!

PR #9274, opened 2023-01-02
Gen 8 Random Battle: Fix lead bugs

MathyFurret:
This pull request improves unit testing by including the Pokemon in the lead slot sometimes when isLead is not explicitly set. This revealed 2 bugs (heh), that are fixed as follows:

* Shuckle was not rolling Sticky Web in the lead slot as it was getting rejected in favor of Stealth Rock.
  * Now, Shuckle *always* runs *both* hazards.
* Scyther was jntended to run Heavy-Duty Boots if it had 3 attacks. However, in the lead position it was still being given Eviolite.
  * This was in fact intended; the test was assuming Scyther would never be the lead. However if it has U-Turn it should run Boots even if it's the lead, so the code and test were updated to reflect that.

### DIFF
--- a/data/mods/gen8/random-teams.ts
+++ b/data/mods/gen8/random-teams.ts
@@ -1723,11 +1723,7 @@ export class RandomGen8Teams {
 		if (moves.has('bellydrum') && moves.has('substitute')) return 'Salac Berry';
 
 		// Misc item generation logic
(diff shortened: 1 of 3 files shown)


ACakeWearingAHat (approved):
Discussed in Random Battles discord

PR #9277, opened 2023-01-02
Change Delta Stream's effectiveness activation to use `-fieldactivate` instead of `-activate`

MrSableye:
Apologies if this is too opinionated of a change, but I'm *fairly* certain this is the only usage of `-activate`  that doesn't provide the target of the effect being activated. The target isn't actively used in rendering on the client, but this should improve the sim protocol's consistency.

~~Also if it matters, Power Construct is an example of an `-activate` message that provides the target without using it to render anything on the client. (see: [Power Construct's protocol message](https://github.com/smogon/pokemon-showdown/blob/b37e8b2499d6940200e5026dc06c1f760feeb4e1/data/abilities.ts#L3049) and [its text not making use of the target](https://github.com/smogon/pokemon-showdown/blob/b57c75bf5431ef8032ba851125b116af83a9074b/data/text/abilities.ts#L1228)).~~ Actually the target is still used to display "[Zygarde's Power Construct]" so it's less precedent, thanks @urkerab 

Either way, it improves consistency and also makes it easier for alternate clients to parse `-activate` messages

### DIFF
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -715,7 +715,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 		onEffectivenessPriority: -1,
 		onEffectiveness(typeMod, target, type, move) {
 			if (move && move.effectType === 'Move' && move.category !== 'Status' && type === 'Flying' && typeMod > 0) {
-				this.add('-activate', target, 'deltastream');
+				this.add('-fieldactivate', 'Delta Stream');
 				return 0;
 			}
 		},
--- a/data/conditions.ts
+++ b/data/conditions.ts
@@ -715,7 +715,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 		onEffectivenessPriority: -1,
 		onEffectiveness(typeMod, target, type, move) {
 			if (move && move.effectType === 'Move' && move.category !== 'Status' && type === 'Flying' && typeMod > 0) {
-				this.add('-activate', '', 'deltastream');
+				this.add('-activate', target, 'deltastream');
 				return 0;
 			}
 		},


urkerab:
Note that the client needs to know whose Ability it is anyway so it can display "[Zygarde's Power Construct]" (and also update the tooltip, although the subsequent transformation would do that anyway), so that's not necessarily a good example.

MrSableye:
Fair enough! I think the consistency point still stands and it shouldn't change anything on the client

DaWoblefet:
The reason it doesn't currently send a target is because it would be an information leak to send a target. In doubles, for example, if you use a move reduced by Delta Stream that hits both opponents, when the "The mysterious strong winds weakened the attack!" string is played, you don't know which one Delta Stream is affecting in that moment, which matters for niche situations where you don't actually know the typing of the target (Illusion, hacked Arceus, things like that). The client shouldn't know which target caused the message to be played because it's not revealed in the string.

data/conditions.ts
```diff
@@ -715,7 +715,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 		onEffectivenessPriority: -1,
 		onEffectiveness(typeMod, target, type, move) {
 			if (move && move.effectType === 'Move' && move.category !== 'Status' && type === 'Flying' && typeMod > 0) {
-				this.add('-activate', '', 'deltastream');
+				this.add('-activate', target, 'deltastream');
```
Karthik99999, line 718:
```suggestion
				this.add('-fieldactivate', 'Delta Stream');
```
I think `-fieldactivate` is the most correct here since we're dealing with a weather and we don't need to send an empty string to not have a target.

MrSableye: Nice! Presumably this would make it work similarly to Pay Day's " Coins were scattered everywhere!"?

Karthik99999: yep (even though Pay Day should probably use `-activate` since it's coming from a specific mon that everyone knows used the move)

DaWoblefet:
Looks good, thanks!

PR #9283, opened 2023-01-04
Tournaments: Support team locking

MathyFurret:
Needs support from client, will write that later.

This adds a new command `/tour teamlock` that allows a tournament to force players to use the same team throughout the tournament. It can only be set before the tournament starts.

Potential issue: This can break if a ruleset is hotpatched during the tournament that makes a player's team invalid. Maybe I should add a way for prepBattle to bypass validation?

### DIFF
--- a/server/tournaments/index.ts
+++ b/server/tournaments/index.ts
@@ -60,6 +60,10 @@ export class TournamentPlayer extends Rooms.RoomGamePlayer<Tournament> {
 	isEliminated: boolean;
 	autoDisqualifyWarned: boolean;
 	lastActionTime: number;
+	/**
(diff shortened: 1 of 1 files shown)


mia-pi-git:
> Potential issue: This can break if a ruleset is hotpatched during the tournament that makes a player's team invalid. Maybe I should add a way for prepBattle to bypass validation?

You could probably just validate once at the beginning and then assume it's fine the rest of the time.

MathyFurret:
Unless it becomes a serious issue I really don't want to rewrite battleready for this, there's not an easy fix because it validates legality and applies filters at the same time

PR #9291, opened 2023-01-07
Random Battles Updates

ACakeWearingAHat:
We have been advised by Annika that objective winrates are unlikely to be procured before Random Battles Team Tournament starts. The reason for this is that attempting to collect winrates results in either Showdown's function being reduced or extremely inaccurate winrates. Annika does not know why this is the case. Subjective level balancing is being implemented as an emergency measure.

Aside from that: 
Tera Ground will no longer appear on Gastrodon
Calm Mind Umbreon will no longer exist
Oricorio will get Defog more often
Hisuian Decidueye will now get Defog
Delibird's suicide move will be changed, again, due to bugged sets.

### DIFF
--- a/data/random-sets.json
+++ b/data/random-sets.json
@@ -1939,7 +1939,7 @@
         ]
     },
     "amoonguss": {
-        "level": 81,
+        "level": 82,
         "sets": [
             {
                 "role": "Bulky Support",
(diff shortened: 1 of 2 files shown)


ACakeWearingAHat (commented):
These changes are approved by Random Battles roomstaff.

ACakeWearingAHat:
I've made a hard stop to any further updates in this patch.

ACakeWearingAHat:
This update still has staff approval. I will keep adding things that have staff approval until it is merged.

PR #9293, opened 2023-01-08
Fix ruleset interactions with Rayquaza-Mega

MathyFurret:
* Rayquaza's ability to Mega Evolve post-gen 7 now checks for "+Past" in the ruleset instead of "Standard NatDex". There are probably other cases of this hardcode that should be changed, but leave that for another PR.
  * NatDex Draft is now able to Mega Evolve Rayquaza
* Banning Rayquaza-Mega no longer implicitly bans Rayquaza + Dragon Ascent; the new behavior is to add an implicit Mega Rayquaza Clause
  * While this change is significant to many formats, tiering folk apparently prefer this if I understand correctly. If it becomes problematic in the future I am more than willing to make a Policy Review thread
  * This fixes Rayquaza validity in National Dex Ubers

### DIFF
--- a/sim/dex-formats.ts
+++ b/sim/dex-formats.ts
@@ -813,6 +813,31 @@ export class DexFormats {
 
 		ruleTable.resolveNumbers(format, this.dex);
 
+		const canMegaEvo = this.dex.gen <= 7 || ruleTable.has('+pokemontag:past');
+		if (ruleTable.has('obtainableformes') && canMegaEvo) {
+			// Banning Rayquaza-Mega implicitly adds Mega Rayquaza Clause
+			const mray = this.dex.species.get('rayquazamega');
+			let mrayBanned = false;
+			const banReason = ruleTable.check('pokemon:' + mray.id);
+			if (banReason) {
+				mrayBanned = true;
+			} else if (banReason !== '') {
+				for (const ruleid of ruleTable.tagRules) {
+					if (ruleid.startsWith('*')) continue;
+					const tag = Tags[ruleid.slice(12)];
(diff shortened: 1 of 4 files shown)


sim/dex-formats.ts
```diff
@@ -813,6 +813,31 @@ export class DexFormats {
...
+					if (ruleid.startsWith('*')) continue;
+					const tag = Tags[ruleid.slice(12)];
+					if (tag.speciesFilter?.(mray)) {
+						if (ruleid.startsWith('+')) break;
+						mrayBanned = true;
+						break;
+					}
+				}
+			}
+			if (mrayBanned) {
```
urkerab, line 835: Unless I'm missing something...
```suggestion
			if (ruleTable.isBannedSpecies(mray)) {
```
Although at this point it might make more sense for the Mega Rayquaza Clause to ban Rayquaza-Mega and have `canMegaEvo` check whether it's banned.

MathyFurret: isBannedSpecies checks the base forme as well, which isn't desirable here. However, I suppose `isBannedSpecies('rayquazamega') && !isBannedSpecies('rayquaza')` would work, and better, too.

urkerab:
(As it turns out, making the Mega Rayquaza Clause ban Mega Rayquaza and then having `canMegaEvo` check for the ban would also have fixed PR #9299 at the same time.)

MathyFurret:
This behavior already betrays the expectation that bans only affect teambuilding; I don't want to make it worse by allowing it to happen without a message to players at the start of the battle

PR #9294, opened 2023-01-08
Gen 1: Fix battle log for charging moves, add hint for stat overflow

livid-washed:
This fixes the battle log so that the second turn of a two-turn move will include "[from]<move>", e.g. "[from]Skull Bash". This was unintentionally removed in https://github.com/smogon/pokemon-showdown/pull/9243.

Edit: also adds a hint for stat overflow in Gen 1


AnnikaCodes:
Thanks!

PR #9295, opened 2023-01-08
Update regex for extracting privileged messages

MrSableye:
# Overview
Fixes an issue where privileged information via `split` is leaked to spectators. Only observed occurneces happen in gen 3 due to Pressure's mechanics being the only message that doesn't have shared information and the data-stripping regex not handling it well.

### Team (for both players)
```
Moltres @ Leftovers  
Ability: Pressure  
EVs: 48 HP  
IVs: 0 Atk  
- Agility  

Claydol @ Choice Band  
Ability: Levitate  
EVs: 56 HP / 252 Spe  
- Protect  
- Facade
```

### Actions
You need to separate clients. Two players and one spectator. When all clients are connected, one player should switch to Moltres and one should switch to Claydol.

## "Raw" Update
Un-processed message before `ServerStream#extractChannel` is applied:
```
|
|t:|1673213422
|split|p2
|switch|p2a: Moltres|Moltres|333/333
|switch|p2a: Moltres|Moltres|100/100
|split|p2
|-ability|p2a: Moltres|Pressure|[silent]

|split|p1
|switch|p1a: Claydol|Claydol|275/275
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

### P1 perspective (player switching in Claydol) [CORRECT]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|100/100
|switch|p1a: Claydol|Claydol|275/275
|
|upkeep
|turn|2
```

### P2 perspective (player switching in Moltres) [CORRECT]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|333/333
|-ability|p2a: Moltres|Pressure|[silent]
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

### Spectator perspective [LEAK]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|100/100
|split|p1
|switch|p1a: Claydol|Claydol|275/275
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

### P1 perspective [CORRECT]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|100/100
|switch|p1a: Claydol|Claydol|275/275
|
|upkeep
|turn|2
```

### P2 perspective [CORRECT]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|333/333
|-ability|p2a: Moltres|Pressure|[silent]
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

### Spectator perspective [CORRECT]
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|100/100
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

## Additional debugging/verification
The issues lie in the last regex being applied, so I've provided some basic samples to play around with regex either in-server or through something like https://regex101.com/ (using the substitution function is useful). Applying the old replace `replace(/\n\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '\n')` doesn't capture all splits properly. Applying the new replace `replace(/(?<=\n)\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '')` does.
### P1 perspective before enemy secrets are removed
```
|
|t:|1673213422
|switch|p2a: Moltres|Moltres|333/333
|-ability|p2a: Moltres|Pressure|[silent]
|split|p1
|switch|p1a: Claydol|Claydol|275/275
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```
### P2 perspective before enemy secrets are removed
```
|
|t:|1673213422
|split|p2
|switch|p2a: Moltres|Moltres|333/333
|switch|p2a: Moltres|Moltres|100/100
|split|p2
|-ability|p2a: Moltres|Pressure|[silent]

|switch|p1a: Claydol|Claydol|275/275
|
|upkeep
|turn|2
```
### Spectator perspective before player secrets are removed
```
|
|t:|1673213422
|split|p2
|switch|p2a: Moltres|Moltres|333/333
|switch|p2a: Moltres|Moltres|100/100
|split|p2
|-ability|p2a: Moltres|Pressure|[silent]

|split|p1
|switch|p1a: Claydol|Claydol|275/275
|switch|p1a: Claydol|Claydol|100/100
|
|upkeep
|turn|2
```

Apologies for the large dump here, this was a weird edge case that was difficult to find by virtue of it being basically 100% Gen 3-specific from my observations. But it would likely occur again if any future gen introduced similar completely secret information.

### DIFF
--- a/server/sockets.ts
+++ b/server/sockets.ts
@@ -257,7 +257,7 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 			const roomChannel = this.roomChannels.get(roomid);
 			for (const [curSocketid, curSocket] of room) {
 				const channelid = roomChannel?.get(curSocketid) || 0;
-				if (!messages[channelid]) messages[channelid] = this.extractChannel(message, channelid);
+				if (!messages[channelid]) messages[channelid] = ServerStream.extractChannel(message, channelid);
 				curSocket.write(messages[channelid]!);
 			}
 		},
@@ -426,23 +426,23 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 		console.log(`Test your server at http://${config.bindaddress === '0.0.0.0' ? 'localhost' : config.bindaddress}:${config.port}`);
 	}
 
-	extractChannel(message: string, channelid: -1 | ChannelID) {
+	static extractChannel(message: string, channelid: -1 | ChannelID) {
 		if (channelid === -1) {
 			// Grab all privileged messages
-			return message.replace(/\n\|split\|p[1234]\n([^\n]*)\n(?:[^\n]*)/g, '\n$1');
+			return message.replace(/(?:(?<=\n)|^)\|split\|p[1234]\n([^\n]*)\n(?:[^\n]*)/g, '$1').replace(/\n$/g, '');
 		}
 
 		// Grab privileged messages channel has access to
 		switch (channelid) {
-		case 1: message = message.replace(/\n\|split\|p1\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
-		case 2: message = message.replace(/\n\|split\|p2\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
-		case 3: message = message.replace(/\n\|split\|p3\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
-		case 4: message = message.replace(/\n\|split\|p4\n([^\n]*)\n(?:[^\n]*)/g, '\n$1'); break;
+		case 1: message = message.replace(/(?:(?<=\n)|^)\|split\|p1\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 2: message = message.replace(/(?:(?<=\n)|^)\|split\|p2\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 3: message = message.replace(/(?:(?<=\n)|^)\|split\|p3\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 4: message = message.replace(/(?:(?<=\n)|^)\|split\|p4\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
 		}
 
 		// Discard remaining privileged messages
 		// Note: the last \n? is for privileged messages that are empty when non-privileged
-		return message.replace(/(?<=\n)\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '');
+		return message.replace(/(?:(?<=\n)|^)\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?(\n$)?/g, '').replace(/\n$/g, '');
 	}
 
 	/**
--- a/server/sockets.ts
+++ b/server/sockets.ts
@@ -427,31 +427,34 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 	}
 
 	static extractChannel(message: string, channelid: -1 | ChannelID) {
-		const messages = message.split('\n');
+		const lines: (string | undefined)[] = message.split('\n');
 
-		for (let i = 0; i < messages.length - 2; i++) {
-			const line = messages[i];
+		for (let i = 0; i < lines.length - 2; i++) {
+			const line = lines[i];
+			if (!line) continue;
+
 			const splitMatch = /^\|split\|p([1234])$/.exec(line);
-
 			if (!splitMatch) continue;
 
 			const [, playerMatch] = splitMatch;
 
 			const player = parseInt(playerMatch);
-			const secretMessage = messages[i + 1];
-			const sharedMessage = messages[i + 2];
+			const secretMessage = lines[i + 1]!; // Assume anything ahead of us is defined if the line is defined
+			const sharedMessage = lines[i + 2]!;
 			const hasPrivilege = (player === channelid) || (channelid === -1);
 
 			let channelMessage = sharedMessage;
 			if (hasPrivilege) channelMessage = secretMessage; // Expose secrets to matching players
-			const isEmptyChannelMessage = channelMessage.length === 0; // If the line is empty, prepare to remove it
+			const isEmptyChannelMessage = channelMessage.length === 0;
 
-			messages[i] = channelMessage;
-			messages.splice(i + 1 - (isEmptyChannelMessage ? 1 : 0), 2 + (isEmptyChannelMessage ? 1 : 0));
-			if (isEmptyChannelMessage) i--;
+			lines[i] = isEmptyChannelMessage ? undefined : channelMessage;
+			lines[i + 1] = undefined;
+			lines[i + 2] = undefined;
 		}
 
-		return messages.join('\n');
+		return lines
+			.filter((line) => line !== undefined)
+			.join('\n');
 	}
 
 	/**
--- a/server/sockets.ts
+++ b/server/sockets.ts
@@ -21,6 +21,39 @@ import {IPTools} from './ip-tools';
 type StreamWorker = ProcessManager.StreamWorker;
 type ChannelID = 0 | 1 | 2 | 3 | 4;
 
+export function extractChannelMessages(message: string) {
+	const channelMessages = new Map<ChannelID | -1, string[]>(
+		[-1, 0, 1, 2, 3, 4].map<[ChannelID | -1, string[]]>((channelid) => [channelid as ChannelID | -1, []]),
+	);
+	const messages: string[] = message.split('\n');
+
+	for (let i = 0; i < messages.length; i++) {
+		const line = messages[i];
+		if (!line) continue;
+
+		const splitMatch = /^\|split\|p([1234])$/.exec(line);
+		if (!splitMatch) {
+			channelMessages.forEach((log) => log.push(line));
+			continue;
+		}
+
+		const [, playerMatch] = splitMatch;
+		const player = parseInt(playerMatch);
+		const secretMessage = messages[i + 1];
+		const sharedMessage = messages[i + 2];
+
+		channelMessages.forEach((log, channelid) => {
+			const hasPrivilege = (player === channelid) || (channelid === -1);
+			const channelMessage = hasPrivilege ? secretMessage : sharedMessage;
+			if (channelMessage) log.push(channelMessage);
+		});
+
+		i += 2;
+	}
+
+	return channelMessages;
+}
+
 export const Sockets = new class {
 	async onSpawn(worker: StreamWorker) {
 		const id = worker.workerid;
@@ -254,10 +287,11 @@ export class ServerStream extends Streams.ObjectReadWriteStream<string> {
 				null, null, null, null, null,
 			];
 			const message = data.substr(nlLoc + 1);
+			const channelMessages = extractChannelMessages(message);
 			const roomChannel = this.roomChannels.get(roomid);
 			for (const [curSocketid, curSocket] of room) {
 				const channelid = roomChannel?.get(curSocketid) || 0;
-				if (!messages[channelid]) messages[channelid] = ServerStream.extractChannel(message, channelid);
+				if (!messages[channelid]) messages[channelid] = channelMessages.get(channelid)!.join('\n');
 				curSocket.write(messages[channelid]!);
 			}
 		},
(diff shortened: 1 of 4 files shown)


sim/battle.ts
```diff
@@ -2934,7 +2934,7 @@ export class Battle {
 
 		// Discard remaining secret data
 		// Note: the last \n? is for secret data that are empty when shared
-		return data.replace(/\n\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '\n');
+		return data.replace(/(?<=\n)\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '');
```
urkerab, line 2937: Do you know why this is the only regexp in this function that's affected? (They are so similar that it seems odd to me that only one would be affected.)
I was also wondering whether using `^` with the `m` flag would be more readable?

MrSableye: Only the final regexes are affected because the others don't have the conditional replacement of the last `\n`, which seems to be problematic. I still invite people to mess around with the regexes on their own to verify this a bit more, since regexes tend to be funky.

As for using `m`, did you have something like `^\|split\|(?:[^\n]*)$\n^(?:[^\n]*)$\n(^$\n)?/gm` in mind?

monsanto:
Nice catch. A test would be appreciated, the fix is fiddly and refactoring any of this code would be a daunting task in general

MrSableye:
Sure, let me take a shot at writing some tests here because yeah, I'd prefer not to introduce another bug considering how "low level" this is

monsanto:
I'm struggling a bit with how to handle this PR. I understand you are doing the best you can with existing code, but that existing code isn't good, so it bleeds into my assessment of the fix. Here are my thoughts:

- The regular expression used to match is hard to understand at a glance (bad property for something security sensitive to have!). My eyes are glazing over at the number of \ns.
- The "fix" is added to every regex, but it sounds like it only matters for one of them. That means that someone later would expend a lot of effort trying to understand that prefix you added, when it doesn't do anything, but there would be resistance to changing it because of fear that it will expose secrets, so it will never be removed. Then someone else will waste a lot of time trying to understand it and the cycle will repeat.
- There are seemingly two different methods for doing the same thing, with more or less the same implementation. 
- One of those methods is not tested.

I think the way forward depends on your motivation. What I'd like, as a maintainer, is for this code to be obviously correct and not leak any information. But maybe you don't want to go that far. Let me know what you think of my assessment and what you think a good way forward would be. If I have some time later I can try to play with it too, and maybe others can as well.

MrSableye:
>The regular expression used to match is hard to understand at a glance, which is a bad property for something security sensitive to have. My eyes are glazing over at the number of \ns.
Yeah these regexes are painful to look at.

>The "fix" is added to every regex, but it sounds like it only matters for one of them. That means that someone later would expend a lot of effort trying to understand that prefix you added, when it doesn't do anything, but there would be resistance to changing it because of fear that it will expose secrets, so it will never be removed. Then someone else will waste a lot of time trying to understand it and the cycle will repeat.
My fix actually fixes another issue which I don't think ever occurs, but I wanted to make sure was covered which was updates which the first `split` is the first line, which is why I ended up adding a similar fix to each regex.

>There are seemingly two different methods for doing the same thing, with more or less the same implementation.
Agreed, these could probably share the same code. It'd take a bit of refactoring, but likely worth it.

>One of those methods is not tested.
Something that should be solvable if refactored above.

As for the way forward, I'd be glad to work on a more clean/readable approach to this and try to make it sharable if that's desired. I've already been looking at this for long enough to know what exactly needs to be done and I don't mind doing more work; the only reason I stuck with using regex is mostly to avoid making too many waves in an area of code that pretty much every message sent to the client sees.

tl;dr: I'm glad to refactor this area with a blessing from the maintainers.

sim/battle.ts
```diff
@@ -2921,20 +2921,20 @@ export class Battle {
...
+		case 'p1': data = data.replace(/(?:(?<=\n)|^)\|split\|p1\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 'p2': data = data.replace(/(?:(?<=\n)|^)\|split\|p2\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 'p3': data = data.replace(/(?:(?<=\n)|^)\|split\|p3\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
+		case 'p4': data = data.replace(/(?:(?<=\n)|^)\|split\|p4\n([^\n]*)\n(?:[^\n]*)/g, '$1'); break;
 		}
 
 		// Discard remaining secret data
 		// Note: the last \n? is for secret data that are empty when shared
-		return data.replace(/\n\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?/g, '\n');
+		return data.replace(/(?:(?<=\n)|^)\|split\|(?:[^\n]*)\n(?:[^\n]*)\n\n?(\n$)?/g, '').replace(/\n$/g, '');
```
urkerab, line 2937: I think `(?:(?<=\n)|^)` can be `(?<=\n|^)`, which would be slightly less unreadable. But maybe I've caused you enough trouble already; it's just that all the regexps in this function are very similar, the basic difference being which player they match, and I actually thought that it would get more confusing if they diverged.

MrSableye:
@monsanto How do you feel about an approach like the current version? Should be more readable this way

monsanto:
I did take a look at the implementation, but I probably won't be able to give feedback until tomorrow, when I have some time to actually check out the code and play with it. The test case was useful for understanding the problem, so thanks for that.

monsanto:
- Your tests enforce that the log does not end with \n. Is the sim actually this picky?
- Split logic appears to also be in battle-stream.ts. It is unclear if it suffers from the same problem. Perhaps it is not possible at this time to consolidate parsing.

MrSableye:
>Your tests enforce that the log does not end with \n. Is the sim actually this picky?
No, it isn't. It's not a requirement of the system, but it seems like a fair thing to test as part of the system and makes it a little more consistent. Glad to remove it or not test on it if desired

>Split logic appears to also be in battle-stream.ts. It is unclear if it suffers from the same problem. Perhaps it is not possible at this time to consolidate parsing.
It doesn't suffer from the same problem, but iirc it might suffer from an issue of exposing the presence of secret information by sending non-privileged consumers an empty line. If there's a good place to locate this code, there shouldn't be an issue with applying it here. There's also some split logic in `roomlog.ts` as well, but it handles it entirely differently (somewhat similarly to my implementation here) and shouldn't be affected. 

Also as a minor update, I've observed that this bug now affects the open team sheets requests for VGC intermitently (https://github.com/smogon/pokemon-showdown/blob/1db21ae74ca3785d207f21fb2b7b7168a7d5040a/data/rulesets.ts#L1624)

monsanto:
So I appreciate what you're trying to do by avoiding regex but I think it isn't quite the right refactor. Operating on data structures rather than doing text munging is the right approach, but it is introducing a deserialization/serialization at an arbitrary point in the program, when really this should only be happening at the edges of the program. After looking at the rest of the architecture I imagine doing this right would require quite an involved refactor, so I think we should fix this bug and maybe come back to that later if there is interest.

The logic modulo delimiters is possible to express cleanly with a regular expression, and I would do so as follows

```typescript
const splitRegex = /^\|split\|(.*)\n(.*)\n(.*)/gm;
message = message.replace(splitRegex, (_, player, secret, shared) => {
    if (channelid === -1 || player === "p" + channelid) {
        return secret;
    } else {
        return shared;
    }
});
return message;
```

The changes here are that
- There is only one regex, and one pass over the string with it
- Superfluous `\n`s have been removed by writing it more idiomatically with `.` (which already doesn't match newlines) and `^` + `m` flag. There is a slight logic difference in that the fields do not match `\r`, if this is actually used in sim code then it needs to be covered in a test.
- The fields are all named; even just seeing you call things `shared` and `secret` in the tests aided my understanding as I was not familiar with this construct before this PR.
- The full structure of the construct is apparent: in the previous regex code, the last use was trying to be clever by not matching the last field, and this may have contributed to the bug.

This brings us to the concern of delimiters. The problem is that with `\n` as a separator (between messages), and not a terminator (always at the end of a message) is that it is hard to do a structure preserving replace. This can be normalized with the following regular expression

```typescript
message = message.replace(/(?<=\n)\n+|^\n+|\n+$/g, "");
```

but it feels inefficient. It is possible to do this in one pass over the string with a similar approach by incrementally advancing the regex with the `lastIndex` property. But will it yield a performance benefit? It is already the case that this code is unnecessarily slow, as it is possible to extract all channels in one pass, and the existing code has a loop that calls `extractChannel`. And again, this normalization afaict would not be required when always using `\n` as a line terminator but maybe the rest of the code is not setup to make this assumption and it would cause a leak. Dunno. Thoughts are welcome and I would still be interested in investigating some code sharing.

MrSableye:
How about my current approach? Refactored it out to a single function to be shared and it generates all channels in one pass. Its final location is TBD, I don't particularly like it sitting in `sockets.ts`, but I'm struggling to find an appropriate spot. This is also pretty similar to:
https://github.com/smogon/pokemon-showdown/blob/ac3647cba01702e1385d39d33a913f45f723fc05/server/roomlogs.ts#L86-L106

It's a necessary evil (from my perspective/understanding) that this logic will need to be explicitly called in different places as it has two entry points: websocket connections on the server for room battles (`socket.ts` logic) and local streams for simulated battles (`battle-stream.ts`).

Also, for reference, I don't believe the data being sent to any of these functions treats `\n` as a message/line/whatever terminator as most of them are doing a simple `join('\n')` on a list of messages to be sent. So the last line won't necessarily have a `\n`. Although, it really wouldn't matter if it did since the client strips trailing `\n`s.

server/sockets.ts
```diff
@@ -21,6 +21,39 @@ import {IPTools} from './ip-tools';
 type StreamWorker = ProcessManager.StreamWorker;
 type ChannelID = 0 | 1 | 2 | 3 | 4;
 
+export function extractChannelMessages(message: string) {
+	const channelMessages = new Map<ChannelID | -1, string[]>(
```
MrSableye, line 25: I'm not personally sold on returning a map. This function could:
- accept a list of ChannelIDs and then return the logs in the order they were provided
- accept nothing but return a list of logs in ascending order (-1, 0, 1, 2, 3, 4)

Not sure of the best approach here.

server/sockets.ts
```diff
@@ -21,6 +21,46 @@ import {IPTools} from './ip-tools';
...
+			Object.values(channelMessages).forEach((log) => log.push(line));
+			continue;
+		}
+
+		const [, playerMatch] = splitMatch;
+		const player = playerMatch;
+		const secretMessage = messages[i + 1];
+		const sharedMessage = messages[i + 2];
+
+		Object.entries(channelMessages).forEach(([channelId, log]) => {
```
monsanto, line 52: Prefer for-of

monsanto (commented):
Apologies for the delay. Feel free to ping me if I am taking too long.

When I mentioned doing everything in one pass it was more of a justification for doing things clearly at the expense of speed. But I think it's worthwhile to make it faster too if it is in our power to do so without sacrificing clarity.

My understanding is that this is doing extra work in some cases. Maybe it would be better to take a list of channel ids to process.

I see you did not like my idea of processing via regular expression. I'm still not crazy about deserializing the message (`split("\n")`) and reserializing it (`join("\n")`) at a suboptimal point in the program, but maybe it is OK.

Part of what is bothering me about me about your approach is that messages aren't actually delimited by `"\n"`. `|split|` is a weird meta-construct that takes 3 lines, and your code has to account for that looking ahead in `messages` and adjusting `i` outside of the `for` header. It's also the case that you are doing unnecessary work by splitting chunks in between `|split|` constructs by line, and then rejoining them later. Maybe the speed difference doesn't matter, but the combination of the two just feels funny to me, it feels like the wrong approach. But it's also the case that the `\n` fiddling of my approach is not really great either. I bet that there is a synthesis of the two approaches that doesn't have any drawbacks.

MrSableye:
> Maybe it would be better to take a list of channel ids to process.

I'll just do this. I can even do some fancy TypeScript magic to make it is only indexable by the IDs you provide :)

> I see you did not like my idea of processing via regular expression. I'm still not crazy about deserializing the message (split("\n")) and reserializing it (join("\n")) at a suboptimal point in the program, but maybe it is OK. [...] |split| is a weird meta-construct that takes 3 lines

I played around with your regex solution a bit and I'm having trouble making it handle and remove completely empty lines (as is, it leaves empty `\n`s which is an information leak). It seems hard to work around this due to regex replacement order (sometimes you're left with a superfluous `\n` if you have two consequetive `|split|`s) which circles us back to whether `\n` is a delimiter or a line terminator (or both?) and how/if/when we can handle the replacement of these extra lines. I don't have the expertise or confidence to say that Showdown is consistent in how `\n` is treated which is part of the reason I thought going with the splitting approach made sense since it breaks things down clearly into lines with a concrete way to look at future lines and replace them without significantly altering the underlying structure like a regex replace might.

But take what I said with a grain of salt, I could be missing something really silly that'd simplify things with the regex, I just haven't found it

MrSableye:
> I'm still not crazy about deserializing the message (split("\n")) and reserializing it (join("\n"))

If it helps, I could return just the joined messages instead of the array of messages to make it easier on callers, if desired.

monsanto:
The channel IDs stuff looks good, thanks. I'm going to try my hand at this one more time and if I don't come up with something w/o an obvious downside we'll just merge what you have.

monsanto:
You can cut the logic by more than half with an approach like this:

```ts
const RE = /^\|split\|p([1234])\n(.*)\n(.*)|.+/gm;
for (const [lineMatch, playerMatch, secretMessage, sharedMessage] of message.matchAll(RE)) {
	const player = playerMatch ? parseInt(playerMatch) : 0;
	for (const channelId of channelIdSet) {
		let line = lineMatch;
		if (player) {
			line = channelId === -1 || player === channelId ? secretMessage : sharedMessage;
			if (!line) continue;
		}
		channelMessages[channelId].push(line);
	}
}
```

This works similarly to what you have, but:
- Matches split constructs as a whole, instead of reconstructing them from their constituent lines
- Otherwise, matches non-empty lines (so you don't have to filter them out)

Going to see if I can pare this down further.

MrSableye:
@monsanto Implemented your approach, super clean.

monsanto:
Great, thanks. I'll merge this tomorrow. I might make some slight tweaks as I don't think, for example, that it's necessary to allocate a `Set` for each call. But nothing big or that you would need to be on the hook for (unless you want to be, and make changes before I get to it)

monsanto:
I took another pass over this and I think beyond the `Set` issue I would say that it would be more clear if channel `0` was named `spectator` and channel `-1` was named `omniscient`, since it doesn't really matter if they're integers. Also it would potentially be better if `extractChannelMessages` returned a record of strings instead of requiring the caller to join. But these can probably be addressed post-merge and I don't want to hold up this PR any more than it is.

Thanks again.

PR #9299, opened 2023-01-09
National Dex Ubers: Fix Dragon Ascent preventing Rayquaza from using Terastal

Zrp200:
Currently Rayquaza cannot Terastal while knowing Dragon Ascent in National Dex Ubers, even though it cannot Mega Evolve thanks to Mega Rayquaza Clause. 

Because Terastal is handled through a field rather than a method, MRC's failure to update it caused inconsistent behavior. This resolves the inconsistency with gen 8 implementation behavior (Rayquaza was able to dynamax under MRC); as such, Rayquaza should be able to use Terastal when using Dragon Ascent under MRC.

This resolves https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9441141

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -1213,7 +1213,11 @@ export const Rulesets: {[k: string]: FormatData} = {
 		onBegin() {
 			this.add('rule', 'Mega Rayquaza Clause: You cannot mega evolve Rayquaza');
 			for (const pokemon of this.getAllPokemon()) {
-				if (pokemon.species.id === 'rayquaza') pokemon.canMegaEvo = null;
+				if (pokemon.species.id === 'rayquaza') {
+					pokemon.canMegaEvo = null;
+					// ability to terastal was determined before the clause activated, causing incorrect behavior
+					pokemon.canTerastallize = this.actions.canTerastallize(pokemon);
+				}
 			}
 		},
 	},


urkerab:
@KrisXV What do you think about my other approach that I mentioned in #9293 whereby `canMegaEvo` checks for the MRC directly, meaning that the Terastallisation check is correct first time?

KrisXV:
@urkerab im indifferent for the most part, but ill leave this open for more discussion

Zrp200:
I'm aware this is a somewhat awkward interaction, but it is very desirable to have NatDex mechanics mirror in-game mechanics as much as possible. 
If Rayquaza was added into the game right now it would undoubtedly be able to Terastal while knowing Dragon Ascent, and the fact this isn't true for NatDex Ubers, where Rayquaza cannot Mega Evolve, is a constant source of confusion within the community.

Zrp200:
> @Zrp200 have you tested this in previous gens? I assume it doesnt impact them at all?
@KrisXV 

https://github.com/smogon/pokemon-showdown/blob/21d0ffaa7bff93665660d8c42d505d911a754330/sim/battle-actions.ts#L1855-L1863
This is the current implementation of that method. One of the checks specifically checks for if the gen is 9, which handles the previous gen cases. 

If we are worried about any checks that would be done beforehand, we could always move the gen check to be checked before anything else.

KrisXV:
No, I was just ensuring that you tested this before merging.

Zrp200:
I did test it. As expected, nothing was amiss even in prior gens.

PR #9301, opened 2023-01-09
Gen 1: Fix Bide selections

livid-washed:
This patch fixes Bide so that it restricts the choices of the Pokemon using it, by disabling the Pokemon's other choices. If a Bide user is disabled, its only choice will be Struggle.


AnnikaCodes:
Thanks!

Karthik99999:
Why not use onLockMove for this?

PR #9304, opened 2023-01-10
VGC/BSS: Add Series 2

DaWoblefet:
Also refactors Paradox Pokemon to not count as Sub-Legendary, since we have the tag for Paradox Pokemon now and it was just an assumption they were classified as Sub-Legendary.


DaWoblefet:
Merging with Kris's blessing.

PR #9307, opened 2023-01-10
Battle Factory updates

adrivrie:
Many sets in Battle Factory (mostly gen 6) had less than maximum EVs without any reason. This fixes that, as well as some uncontroversial set changes in gens 6 and 8 BF, and in gen 8 BSSF.


adrivrie:
Because of a sudden influx of energy into the format by rands staff this became a larger patch, but it's still all uncontroversial set changes.

ACakeWearingAHat (approved):

PR #9309, opened 2023-01-10
Fix "Opponent withdrew" message to ignore nicknames when requested.

anandoza:
- Changed "withdrew" message to use FULLNAME instead of NICKNAME

When ignoring nicknames, the "Opponent withdrew [pokemon]" sentence still used the nickname, which is confusing.


anandoza:
For an example, see this forum post/screenshot: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-63#post-8988978

DaWoblefet:
This isn't the correct place to fix the issue. The bug is that while ignoring nicknames, the switch out message still occurs. But in fact on cart if you switch out a Pokemon with a nickname, it would display the nickname. The correct fix here is client-side.

PR #9315, opened 2023-01-12
Gen 1: Fix Thrash and Rage mechanics

livid-washed:
This patch fixes thrashing moves' mechanics in Gen 1:

- Thrash will now lock the player in no matter what. Previously it would not lock the player if the move hits and breaks a substitute.
- Thrash is now paused when the user is asleep, frozen, flinching, or partially trapped.
- Thrash is now paused when the user has Thrash disabled.

It also implements the Thrash and Rage accuracy bug: https://www.youtube.com/watch?v=NC5gbJeExbs


AnnikaCodes:
Thanks!

PR #9317, opened 2023-01-13
Fix/override cant protocol text for Disable

pauliwag:
Fix latter part of [this](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-85#post-9380156) bug.


DaWoblefet:
Thanks pauliwag!

PR #9320, opened 2023-01-13
Terastal information is no longer sent to the client for Pokemon that cannot terastallize.

Zrp200:
This includes: 

* when terastal has already been used 
* when terastal is banned 
* when terastal is unusuable due to other mechanics (natdex) 

This should also resolve smogon/pokemon-showdown-client#2086

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1085,7 +1085,7 @@ export class Pokemon {
 			entry.commanding = !!this.volatiles['commanding'] && !this.fainted;
 			entry.reviving = this.isActive && !!this.side.slotConditions[this.position]['revivalblessing'];
 		}
-		if (this.battle.gen === 9) {
+		if (this.battle.gen === 9 && this.canTerastallize) {
 			entry.teraType = this.teraType;
 			entry.terastallized = this.terastallized || '';
 		}


DaWoblefet:
What happens when a Pokemon is already Terastallized?

Zrp200:
> What happens when a Pokemon is already Terastallized?

It appears to me that it still works, but I can put a check in there for it anyway just in case.

sim/pokemon.ts
```diff
@@ -1085,7 +1085,7 @@ export class Pokemon {
 			entry.commanding = !!this.volatiles['commanding'] && !this.fainted;
 			entry.reviving = this.isActive && !!this.side.slotConditions[this.position]['revivalblessing'];
 		}
-		if (this.battle.gen === 9) {
+		if (this.terastallized || this.battle.gen === 9 && this.canTerastallize) {
```
Zrp200, line 1088: Whether or not this check is needed is debatable, though it does safeguard from future code's assumptions of both being defined at once.

Currently, the client does assume that `teraType` is always defined when `terastallized` is defined.

[battle-tooltips.ts](https://github.com/smogon/pokemon-showdown-client/blob/07074574258956117f01bdf57e329da7603156e8/src/battle-tooltips.ts#L798)
```typescript
			let types = serverPokemon?.terastallized ? [serverPokemon.teraType] : this.getPokemonTypes(pokemon);
```
In truth, though, this check is completely unneeded since `getPokemonTypes()` actually checks `terastallized` itself. If the line of code that assumes `teraType` is defined is refactored to ```typescript
let types = this.getPokemonTypes(pokemon)
```
it might be desirable to just always send `terastallized` to the client, as it is by default falsey regardless, and then only gate sending of `teraType`.

PR #9321, opened 2023-01-13
Randbats: Add a chat plugin to track winrates

mia-pi-git:
Requested by randbats staff. 
Since the current winrates script is broken, and rands maintainers depend on winrates to balance the format, this is a more long-term solution (since it can be accessed at any time without a sim3 administrator). 

Would be nice if this could be merged within a few days - rands maintainers need it soon.


MathyFurret:
I personally liked the Z-scores that came with the old stats, but I don't know what the rest of the team thinks so I asked them.

The formula would be: `2 * Math.sqrt(timesGenerated) * (numWins / timesGenerated - 0.5)`

adrivrie:
Is there a downside to including gens 1 and 8?

mia-pi-git:
> Is there a downside to including gens 1 and 8?

I discussed this with randbats staff. They weren’t interested in tracking 1 or 8, so I left them out since it means less server load and less space used.

PR #9323, opened 2023-01-14
Rooms: Support changing case/spaces of room title

MathyFurret:
Allow changing the title of a room thru /renameroom if it would not change the room ID (e.g. testroom -> Test Room)

Also fixes a bug where "g" gets added to the room ID with /renamegroupchat


AnnikaCodes:
This _looks_ like it would work, but have you tested it?

AnnikaCodes:
Thanks

PR #9326, opened 2023-01-14
Gen 4: Fix Copycat

MathyFurret:
It should not copy called moves.

### DIFF
--- a/sim/battle-actions.ts
+++ b/sim/battle-actions.ts
@@ -301,7 +301,7 @@ export class BattleActions {
 			pokemon.side.zMoveUsed = true;
 		}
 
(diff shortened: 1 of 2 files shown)


sim/battle-actions.ts
```diff
@@ -301,6 +301,8 @@ export class BattleActions {
 			pokemon.side.zMoveUsed = true;
 		}
 
+		const oldActiveMove = this.battle.activeMove;
```
urkerab, line 304: (This seems to be the same as `move`, since that's what we set it from on line 247; at least, I can't see how it could have changed yet.)

DaWoblefet:
Thanks Mathy!

PR #9328, opened 2023-01-15
Fix interaction of Ruin Abilities with themselves

HisuianZoroark:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-104#post-9468014

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -357,9 +357,9 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		},
 		onAnyModifySpD(spd, target, source, move) {
 			const abilityHolder = this.effectState.target;
-			if (abilityHolder === target) return;
+			if (target.hasAbility('Beads of Ruin')) return;
 			if (!move.ruinedSpD?.hasAbility('Beads of Ruin')) move.ruinedSpD = abilityHolder;
-			if (move.ruinedSpD !== abilityHolder && move.ruinedSpD !== target) return;
+			if (move.ruinedSpD !== abilityHolder) return;
 			this.debug('Beads of Ruin SpD drop');
 			return this.chainModify(0.75);
 		},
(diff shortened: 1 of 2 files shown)


data/abilities.ts
```diff
@@ -357,7 +357,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		},
 		onAnyModifySpD(spd, target, source, move) {
 			const abilityHolder = this.effectState.target;
-			if (abilityHolder === target) return;
+			if (abilityHolder === target || target.hasAbility('Beads of Ruin')) return;
```
Karthik99999, line 360: we can probably just shorten these to just the hasAbility check right? since that must return true on the ability holder

HisuianZoroark: That feels correct

HisuianZoroark:
Implemented all suggestions, thanks!

DaWoblefet:
Thank you!

data/abilities.ts
```diff
@@ -357,9 +357,9 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		},
 		onAnyModifySpD(spd, target, source, move) {
 			const abilityHolder = this.effectState.target;
-			if (abilityHolder === target) return;
+			if (target.hasAbility('Beads of Ruin')) return;
 			if (!move.ruinedSpD?.hasAbility('Beads of Ruin')) move.ruinedSpD = abilityHolder;
-			if (move.ruinedSpD !== abilityHolder && move.ruinedSpD !== target) return;
+			if (move.ruinedSpD !== abilityHolder) return;
```
urkerab, line 362: (Although with the new mechanics couldn't `ruinedSpD` et. al. just be boolean?)

urkerab: Oh, Def and SpD are harder because of spread moves. Never mind then.

PR #9333, opened 2023-01-16
Fix Disable and Cursed Body interactions with Metronome- and Copycat-actuated two-turn moves

pauliwag:
Fix [this](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-85#post-9380156) bug.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -3689,8 +3689,15 @@ export const Moves: {[moveid: string]: MoveData} = {
 					this.debug(`Pokemon hasn't moved yet`);
 					return false;
 				}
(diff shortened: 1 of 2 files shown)


DaWoblefet:
This implementation won't work because there's more moves than just Copycat/Metronome that can call other moves. Hardcoding all of the moves that call other moves isn't the right strategy here.

pauliwag:
Thank you for your feedback, and sorry my formatter broke your CI 😅  I will close this PR and perhaps revisit in the future.

PR #9335, opened 2023-01-19
Add Gen 9 Trademarked

HisuianZoroark:
What's new is a once per turn limit on each individual Pokemon trying to use a trademark. This fixes most, if not all, of the issues that cause trademarked to get taken down last gen. Tested this rigorously so it does not crash the server.


KrisXV:
ty!

PR #9344, opened 2023-01-22
Random Battles updates

ACakeWearingAHat:
-Dudunsparce will not always get Tera STAB, because it doesn't *always* want shadow ball
-Magearna now has a Shift Gear Weakness Policy Tera Blast user set
-Vespiquen now gets Pressure; Spiritomb now does not
-Forretress, Carbink, and Donphan will no longer get useless Tera types
-Several other minor set changes
-Ice-types will now always get Freeze-Dry.

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -165,7 +165,7 @@ export class RandomTeams {
 				)
 			),
 			Ground: (movePool, moves, abilities, types, counter) => !counter.get('Ground'),
-			Ice: (movePool, moves, abilities, types, counter) => !counter.get('Ice'),
+			Ice: (movePool, moves, abilities, types, counter) => (movePool.includes('freezedry') || !counter.get('Ice')),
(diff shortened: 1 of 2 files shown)


data/random-teams.ts
```diff
@@ -676,7 +676,7 @@ export class RandomTeams {
 				let moveType = move.type;
 				if (['judgment', 'revelationdance'].includes(moveid)) moveType = types[0];
 				if (!this.noStab.includes(moveid) && (move.basePower > 30 || move.multihit || move.basePowerCallback)) {
-					if (teraType === moveType) {
+					if (teraType === moveType && species.baseSpecies !== 'Dudunsparce') {
```
livid-washed, line 679:
```suggestion
					if (teraType === moveType) {
```

Dudunsparce check would be better placed in line 672

livid-washed (changes requested):

MathyFurret (approved):
LGTM.

PR #9345, opened 2023-01-22
Gen 1 Thrash: Confusion from thrashing should be silently added

livid-washed:
Confusion due to fatigue (after thrashing) was incorrectly and unintentionally made non-silent in the previous PR linked below. This fixes it, so that the confusion is once again silent.

https://github.com/smogon/pokemon-showdown/pull/9315


AnnikaCodes:
tyty

PR #9357, opened 2023-01-25
Gen 2: Fix Encore

livid-washed:
This fixes Encore in Gen 2:

We introduce a new property ``pokemon.lastMoveEncore`` that only exists for Gen 2. It is identical to ``pokemon.lastMove``, except that it is not reset when the other side switches. This property is used for Encore, and ensures that enemies can still be Encored even if their last move was before the Encore user switched in.


KrisXV:
thank u

PR #9360, opened 2023-01-27
Trademarked: Restrict nonstatus moves

HisuianZoroark:
Removed too much of the absolute trademark bans that I forgot that damaging moves shouldn't be trademarks.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -1506,7 +1506,7 @@ export const Formats: FormatList = [
 			}
 			// Absolute trademark bans
 			if (ability.category !== 'Status') {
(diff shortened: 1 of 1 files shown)


config/formats.ts
```diff
@@ -1504,6 +1504,10 @@ export const Formats: FormatList = [
 			if (!ability.exists) { // Not even a real move
 				return this.validateSet(set, teamHas);
 			}
+			// Absolute trademark bans
+			if (ability.category !== 'Status') {
+				return [`${ability.name} is not a status move, and cannot be used as a trademark.`];
```
KrisXV, line 1509:
```suggestion
				return [`${ability.name} is not a status move and cannot be used as a trademark.`];
```

PR #9362, opened 2023-01-27
Fix Illusioned Tera Type not disappearing

MathyFurret:
I'm not 100% satisfied with this solution because it causes a type change to appear on the real Pokemon that's the same as its base type, so it will still say "type changed." But I may as well submit it anyways while I try to figure it out, to allow other people to chime in.

### DIFF
--- a/data/abilities.ts
+++ b/data/abilities.ts
@@ -1860,6 +1860,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		},
 		onEnd(pokemon) {
 			if (pokemon.illusion) {
+				pokemon.apparentType = pokemon.illusion.apparentType
 				this.debug('illusion cleared');
 				pokemon.illusion = null;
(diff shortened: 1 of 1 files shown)


data/abilities.ts
```diff
@@ -1860,6 +1860,7 @@ export const Abilities: {[abilityid: string]: AbilityData} = {
 		},
 		onEnd(pokemon) {
 			if (pokemon.illusion) {
+				pokemon.apparentType = pokemon.illusion.apparentType;
```
MathyFurret, line 1863: An alternative is this:
```suggestion
				if (!pokemon.terastallized) {
					this.add('-end', pokemon, 'typechange', '[silent]');
					pokemon.apparentType = pokemon.baseSpecies.types.join('/');
				}
```
This fixes the issue in the OP but causes actual type changes (Soak) to disappear until the end of the turn. Maybe that's better, I can't decide.

MathyFurret:
Thought about it and this^ is less preferable because sending a "typechange ended" message when only one typechange ended is a confusing implementation

PR #9363, opened 2023-01-28
Gen 9 Random Battles update

livid-washed:
- Some minor tera type and movepool updates, including giving all Pokemon with Last Respects a fixed Choice Scarf set
- Salt Cure now counts as STAB, and is an enforced move
- Created a list of Pokemon called PriorityPokemon, which have STAB priority enforced, and the STAB priority counts as regular (and tera, if appropriate) STAB even if it is in the noSTAB list. (For example, Breloom has Mach Punch enforced, and this counts as Fighting STAB, so Close Combat is not required).

https://discord.com/channels/294651453279174656/1041098998624288990/1052446471615684689

### DIFF
--- a/data/random-sets.json
+++ b/data/random-sets.json
@@ -481,7 +481,7 @@
             {
                 "role": "Fast Attacker",
                 "movepool": ["Aura Sphere", "Fire Blast", "Nasty Plot", "Psystrike", "Recover", "Shadow Ball"],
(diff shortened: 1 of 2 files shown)


ACakeWearingAHat (approved):
This is the entire contents of the "To-Do" pin in the Random Battles auth discord. Thank you.

ACakeWearingAHat:
(still good)

PR #9368, opened 2023-01-28
Mafia: Prevent incorrect alignment.

Lucas-Meijer:
A new theme includes a role that can change someone's alignment to Town when discarded.


HoeenCoder:
Looks good, let me know when you fix the null assertion and we can merge.

PR #9369, opened 2023-01-28
Cure burns when Thermal Exchange is obtained

HisuianZoroark:
Confirmed by PS description and research thread


DaWoblefet:
Thanks Hisuian Zoroark!

PR #9374, opened 2023-02-01
Random Battles Monthly Spotlight: POTD

livid-washed:
Implements the new Random Battles monthly spotlight: POTD

The POTD will always generate in slot 2, unless maxTeamSize is 1, in which case it generates in slot 1.

https://www.smogon.com/forums/threads/random-battles-monthly-spotlight-cycle-5-voting.3715058/#post-9491679


KrisXV:
ty!

PR #9376, opened 2023-02-01
POTD fixes

livid-washed:
This patch makes some minor fixes to the POTD format:
- Adds a broadcast when each battle starts, stating the Pokemon of the Day
- Exempts POTD from type combo check. Previously, the POTD would be rejected if the lead Pokemon generated had the same typing as the POTD.

The former was requested by rands staff, including here: https://discord.com/channels/294651453279174656/1012049648472559656/1070451381649027142


livid-washed:
> Rather than exempting it entirely, you could add its type combo to `typeComboCount` before the first Pokémon is chosen since we already know for sure it will be on the team.

Yes, that's a great suggestion! Much more elegant (and accurate, for generating the lead). Thanks!

This now adds the POTD's type and weakness counters at the beginning, and skips updating those counters for the POTD when it gets generated.

PR #9380, opened 2023-02-03
NatDex: Fix Revival Blessing + Pursuit interaction

MathyFurret:
Old behavior: Pursuit would proc on a mon using Revival Blessing, if that mon had any unfainted allies

New behavior: Pursuit will not proc on Revival Blessing

I believe this is sensible, as Revival Blessing is not actually a switch.

### DIFF
--- a/sim/battle.ts
+++ b/sim/battle.ts
@@ -2695,12 +2695,12 @@ export class Battle {
 						reviveSwitch = true;
 						continue;
 					}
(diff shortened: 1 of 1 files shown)


DaWoblefet:
Has there been approval from the National Dex council for this? It almost certainly isn't supposed to do this but they should be OKing all NatDex interactions like this.

DaWoblefet:
Thanks Mathy!

PR #9386, opened 2023-02-06
Gen 3-7 Random Battle: Prepare for level balancing

adrivrie:
This makes all levels in gens 3-7 Random Battles explicit the way they are in gen8. That is, every set should have a level attached to it in formats-data.ts.

This will make it much easier to interact with for level balancing scripts we will likely soon start using. Although no final decision has been made, it seems likely that winrate balancing will soon commence for gens 3-7.

Currently, levels are decided by tier-scaling, modified bst-scaling and hard-coded exceptions.

To be clear: this PR does not change functionality.

### DIFF
--- a/data/mods/gen3/formats-data.ts
+++ b/data/mods/gen3/formats-data.ts
@@ -7,6 +7,7 @@ export const FormatsData: {[k: string]: ModdedSpeciesFormatsData} = {
 	},
 	venusaur: {
 		randomBattleMoves: ["curse", "earthquake", "hiddenpowerrock", "leechseed", "sleeppowder", "sludgebomb", "swordsdance", "synthesis"],
+		randomBattleLevel: 82,
 		tier: "UUBL",
 	},
 	charmander: {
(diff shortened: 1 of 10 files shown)


adrivrie:
Bumping this because I think it would be nice to have, even before changing any functionality. It would confirm that this format is acceptable, so I can continue working on balance tooling without fear that the i/o needs changing.

Also, I would not want the actual level changes to be in the same commit, to keep it as clean as possible in case something needs to be reverted.

If we don't want this merged before it's actually used by something, I'd still appreciate a review so I know it's sufficient.

monsanto:
I'm operating as a reviewer of last resort here. Conceptually this seems fine to me, I like consistency between generations. 

The only potential concern I have is that if there is no followup on the winrate balancing script--or if for some reason it is decided not to use this approach--then changes to old gen tiers require the `randomBattleLevel` to also change. This is pretty rare but it sometimes happens. I don't know if it is an important concern or not.

If whoever is in charge of old gen randbats has approved winrate balancing then I don't have an issue with guaranteeing that `randomBattleLevel` is present in the data files.

adrivrie:
As for now it really seems like winrate based level changes will become the standard for gens 3+, though it will likely be postponed until the end of March for preparation reasons, and because there's a big tournament going on now, and we don't want to shake things up too dramatically. Of course people could change their minds, but that's always a risk, and I think we should be able to revert this commit cleanly in that case.

monsanto:
Per Discord convo this is approved by me, merge can wait until the tooling is written

adrivrie:
After more discussion this is now on hold to see if we want to do a more thorough revamp, separating rands data from formats/tiers data.

adrivrie:
Closing as superseded by #9413

PR #9387, opened 2023-02-10
Prevent Commanded Dondozo from switching out through Baton Pass

HisuianZoroark:
One of the things from DarkFE's recent Commander mechanics post in the research thread. I am not sure how many other moves (Teleport, Chilly Reception, etc.) this applies to.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -1118,8 +1118,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 40,
 		priority: 0,
 		flags: {},
-		onHit(target) {
-			if (!this.canSwitch(target.side)) {
+		onTryHit(target) {
+			if (!this.canSwitch(target.side) || target.volatiles['commanded']) {
 				this.attrLastMove('[still]');
 				this.add('-fail', target);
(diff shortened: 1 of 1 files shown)


data/moves.ts
```diff
@@ -1118,8 +1118,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 40,
 		priority: 0,
 		flags: {},
-		onHit(target) {
-			if (!this.canSwitch(target.side)) {
+		onTryHit(target) {
```
DaWoblefet, line 1121: Did this change as well? Baton Pass failures weren't this early in previous generations.

HisuianZoroark: Ah sorry, the onHit failure caused a strange issue where the commanded dondozo could still switch out even though it triggered a failure. I can refactor the commander failure to be somewhere else if needed.

DaWoblefet: Apparently we're using the same thing for Healing Wish and Lunar Dance to get them to work, so I'll just merge now since it's only a slight mechanics error and fixes a more important one.

PR #9393, opened 2023-02-11
Random Battles updates

ACakeWearingAHat:
Do not hotpatch until Sunday

-Revamped most arceus sets
-Removed Tera Blast User Tatsugiri and Sylveon; in their place, Ghost Tera Blast Falinks (which can switch out!) and Ground Tera Blast Florges will now exist. 
-New framework exists for the existence of bulkier Tera Blast users. Tera Blast users with a recovery move and 2 attacks will get Leftovers instead of the usual Life Orb.
-Wet Tauros can now roll Raging Bull, but its other water moves are incompatible with it.
-Special Toxicroak was removed in an attempt to increase its win rate.
-Moves are shuffled now instead of always generating STABs in slots 1 and 2 both for aesthetic purposes and to improve Camomons Mod.
-Many other set changes.

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -131,7 +131,10 @@ export const Formats: FormatList = [
 			'Picked Team Size = 1', 'Max Team Size = 3',
 			'Standard', 'Terastal Clause', 'Sleep Moves Clause', 'Accuracy Moves Clause', '!Sleep Clause Mod',
 		],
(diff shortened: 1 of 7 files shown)


livid-washed (approved):

AnnikaCodes:
thanks!

PR #9395, opened 2023-02-13
Item Pool for /randbats command

ThreeEeveesInACoat:
I'm not super experienced with this specific language, but I've always thought it'd be nice to know what items a pokemon can have in the /randbats command, instead of using the /rcalc. I started doing this (all items up to line 1262 are there, as well as a few others scattered around, but there may be some things I've overlooked or gotten wrong, this is the first time I've done something like this), and I spend around an hour or two on what I have thus far. Some help with making sure that what I'm trying to do is even possible or if I'm just royally screwing up everything would be nice. If this does work, I can finish the rest soon. Once again, not very good with Github or Typescript, so anyone with a bit more experience who can help out would be nice, I just want to make sure I don't waste any more time on something that isn't gonna work in the first place.

### DIFF
--- a/data/random-sets.json
+++ b/data/random-sets.json
@@ -5,12 +5,15 @@
             {
                 "role": "Fast Attacker",
                 "movepool": ["Earthquake", "Flamethrower", "Focus Blast", "Hurricane", "Will-O-Wisp"],
-                "teraTypes": ["Fire", "Ground", "Water"]
+                "teraTypes": ["Fire", "Ground", "Water"],
+                "itempool": ["Heavy-Duty Boots"]
             },
             {
                 "role": "Setup Sweeper",
                 "movepool": ["Dragon Dance", "Earthquake", "Flare Blitz", "Swords Dance", "Thunder Punch"],
-                "teraTypes": ["Fire", "Ground"]
+                "teraTypes": ["Fire", "Ground"],
+                "itempool": ["Heavy-Duty Boots"]
+
             }
         ]
     },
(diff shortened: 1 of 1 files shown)


adrivrie:
Hello! As you will have noticed, the items that each pokemon can get are not currently in the code. This is because items, unlike moves, are a sort of emergent property. Item generation works with general-ish rules: If a pokemon has four attacks they often get a choice item, with three attacks it's often life orb, and many more rules and exceptions govern the items. You can see this [here](https://github.com/smogon/pokemon-showdown/blob/master/data/random-teams.ts#L1014). These rules can, and do occasionally, change. Furthermore, movepool changes can have quirky/rare interactions with item generation that we don't foresee.

This system makes it so that there's no simple way of displaying the items in the `/randbats` command, especially given the new "role" system.

What you're doing is taking the item list from the calculator. The way that the calculator gets this data is by taking it from [this repository](https://pkmn.github.io/randbats/), specifically [here](https://github.com/pkmn/randbats/blob/main/data/gen9randombattle.json). The way this thing knows which items are possible, is that every hour it checks whether the code has changed, and if it has, it generates 100 000 teams and notes the results. One problem is that it may not capture rare items on rare pokemon, which you may not see in that many teams. A more fundamental issue is that adding this data makes it somewhat circular. That is, this codebase informs the data repo, which in turn changes the code here.

What I'm trying to say is that this is something that we don't really want to have. Every solution (such as ourselves generating a bunch of teams every patch) creates more problems, and in total it's simply not currently deemed worth it.

If you wanted to work on this specifically then I'm afraid I'll have to disappoint you. If you wanted to help out the project in general however, I recommend stopping by the Discord or finding some approved [suggestions](https://www.smogon.com/forums/forums/suggestions.517/) or confirmed [bugs](https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/) to work on.

Apologies for the wall of text!

DaWoblefet:
Closing for the reasons listed above, but thanks for giving it a whirl!

PR #9396, opened 2023-02-15
Refactor some move lists to move flags

DaWoblefet:
This refactors Encore, Me First, Future moves, always Pressure, Pledge Combo moves, Sleep Talk, Assist, Copycat, Instruct, Mimic, and Parental Bond to be in flags, consistent with how Gen IX is handling them now.
- New flag names (open to discussion): ``failencore, failmefirst, futuremove, mustpressure, pledgecombo, nosleeptalk, noassist, failcopycat, failinstruct, failmimic, noparentalbond.``
- Opted not to refactor these to be flags:
  - "always Sheer Force" flag (our secondaries system is different enough from cart that this isn't worth it when it's already working)
  - "don't display effectiveness in move selection" flag (no PS support for that cart feature yet)
  - "can't use twice" flag (literally only Gigaton Hammer)
  - Metronome flag (because I did this all manually and am NOT doing Metronome without a script)
  - 5 mystery flags that are still a mystery

It also corrects dozens of move flags. The general process was first to compare the effect's description to the cartridge list, then compare PS's move data to the cartridge list. They should all be synchronized. I was only able to verify directly from the cartridge lists in Gens 6-9, so for Gens 5 and prior I assumed descriptions were accurate. I also used BDSP's cartridge list, so it could _hypothetically_ be missing SwSh DLC moves, but in fact I'm pretty sure there are none missing because they don't show up in Gen 9's list.
- Tera Blast now interacts correctly with Pressure (deducts 2 PP if any opponent has Pressure, even if the target wasn't a Pressure target).
- Parental Bond now no longer deals damage twice on future moves.
- Parental Bond now interactions correctly with Misty Explosion (deal damage once but play the "Hit 1 time!" string).
- Future Sight has the allyanim flag (Gen 9 only).
- Mirror Move now can call any of the LGPE moves and Grassy Glide (note: I did not have an LGPE move data dump; this was using the Gen 8/9 dumps).
- The 5 Torque moves got a bunch of new flags correctly added.
- Crafty Shield and Obstruct are now properly flagged as callable by Copycat. Mirror Coat can also be called (Gen 9 only).
- Instruct properly flags Chatter and Hold Hands as not callable.
- Belch and Comeuppance are now properly flagged as not callable by Me First. Chatter is now flagged as callable (Gen 9 only). In Gen 4, Me First cannot call Me First or real Struggle; it can call hacked Struggle. I also tested Metal Burst and it can call that successfully.
- Mimic added a bunch of new unMimicable moves in Gen 9: Assist, Belch, Celebrate, Copycat, Hold Hands, Me First, Metronome, Mirror Move, Nature Power, Sleep Talk 
- Likely a bunch of other small discrepancies that I didn't explicitly take notes about. Again, if description didn't match the list before, it does now.

I've included all these new flags in movesearch and also cleaned up the formatting of the help information.


sim/dex-moves.ts
```diff
@@ -25,25 +25,36 @@ export type MoveTarget =
...
 	gravity?: 1; // Prevented from being executed or selected during Gravity's effect.
 	heal?: 1; // Prevented from being executed or selected during Heal Block's effect.
+	instruct?: 1; // Cannot be repeated by Instruct.
+	mefirst?: 1; // Cannot be selected by Me First.
+	mimic?: 1; // Cannot be copied by Mimic.
 	mirror?: 1; // Can be copied by Mirror Move.
-	allyanim?: 1; // The move has an animation when used on an ally.
+	mustpressure?: 1; // Additional PP is deducted due to Pressure when it ordinarily would not.
 	nonsky?: 1; // Prevented from being executed or selected in a Sky Battle.
+	parentalbond?: 1; // Cannot be made to hit twice via Parental Bond.
```
KrisXV, line 49: considering the purpose of this flag is to prevent it from triggering on parental bond, wouldn't something like `unbound` make more sense? similar to sleep talk etc flags

DaWoblefet: I don't like 'unbound' but 'non[thing]' or 'no[thing]' or 'cant[thing]' are all options I think are reasonable and am very open to any name suggestions.

data/mods/gen6/moves.ts
```diff
@@ -31,10 +39,9 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		condition: {
 			duration: 3,
 			onStart(target) {
-				const noEncore = ['encore', 'mimic', 'mirrormove', 'sketch', 'struggle', 'transform'];
```
urkerab, line 34: (I think what happened here is that the `noEncore` list diverged for Gen 7, and then the code was updated for the mechanics in later generations, but since those mechanics won't exist in Gen 6, the Gen 9 code would actually work with the move flag-based approach. I notice you already did the assist cases, although those are easier because it's not a condition.)

data/mods/gen8/moves.ts
```diff
@@ -211,6 +216,10 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		isNonstandard: null,
 	},
+	futuresight: {
+		inherit: true,
+		flags: {futuremove: 1},
+	},
```
urkerab, line 222: Is this a deliberate behaviour change? What about Doom Desire?

data/mods/gen6/moves.ts
```diff
@@ -3,6 +3,14 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		priority: 1,
 	},
+	assist: {
+		inherit: true,
+		flags: {},
+	},
```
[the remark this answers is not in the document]
DaWoblefet: There's quite a bit of oldgens inheritance I still have to go through and clean up.

data/mods/gen8/moves.ts
DaWoblefet: Future Sight has the allyanim flag in 9 but not in 8.

data/mods/gen6/moves.ts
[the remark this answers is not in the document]
DaWoblefet: Oldgen cleanup done.

DaWoblefet:
Updated the PR description with a more robust list of changes. At this point, I think the only thing left is debating the flag names.

DaWoblefet:
After a brief talk with Marty, I changed some flags as ``fail[thing]`` and ``no[thing]``, consistent with how the flags are labeled in the move data dumps. I do think this provides additional clarity, at a slight cost of needing to flags for some moves to multiple lines because they're past the 120 character limit.

kittenchilly:
I was thinking that it might be easier to add the metronome flag as a nometronome flag, since there are fewer moves that don't have the metronome flag than those that do. It's not cart accurate but it should work the same, and is similar to how we already do it.

kittenchilly:
Also shouldn't we implement the "can't use twice" flag anyways? Since it's a flag, I assume GF wants to give it to future moves. Also allows custom moves to easily use it if they wanted.

DaWoblefet:
> I was thinking that it might be easier to add the metronome flag as a nometronome flag, since there are fewer moves that don't have the metronome flag than those that do. It's not cart accurate but it should work the same, and is similar to how we already do it.

I would prefer us use the actual game flags, but in either case I don't plan to move Metronome over in this PR because I would need to script it and I'd rather just say "everything but Metronome right now".

> Also shouldn't we implement the "can't use twice" flag anyways? Since it's a flag, I assume GF wants to give it to future moves. Also allows custom moves to easily use it if they wanted.

Abstracting the behavior is fine, but it feels kind of silly to make it a flag for a single move. Would absolutely be fine with someone doing it in a follow-up PR.

Zarel (approved):
This is great! I'd love to merge this ASAP so we don't have to worry about merge conflicts.

DaWoblefet:
This has Marty's blessing as well now, so I'll merge.

PR #9398, opened 2023-02-17
Gen3 sleeptalk cant call rest

frank113:
1. Added `'rest'` to `NoSleepTalk` array to ensure that sleep talk will never call rest in generation 3
2. Added unit test for generation 3 to ensure that sleep talk never calls rest even with missing HP.

### DIFF
--- a/data/mods/gen3/moves.ts
+++ b/data/mods/gen3/moves.ts
@@ -537,7 +537,7 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 			for (const moveSlot of pokemon.moveSlots) {
 				const move = moveSlot.id;
 				const pp = moveSlot.pp;
-				const NoSleepTalk = ['assist', 'bide', 'focuspunch', 'metronome', 'mirrormove', 'sleeptalk', 'uproar'];
+				const NoSleepTalk = ['assist', 'bide', 'focuspunch', 'metronome', 'mirrormove', 'rest', 'sleeptalk', 'uproar'];
 				if (move && !(NoSleepTalk.includes(move) || this.dex.moves.get(move).flags['charge'])) {
 					moves.push({move: move, pp: pp});
 				}
--- a/test/sim/moves/sleeptalk.js
+++ b/test/sim/moves/sleeptalk.js
@@ -43,4 +43,27 @@ describe('Sleep Talk', function () {
 		assert.equal(breloom.hp, hp);
 		assert.equal(move.pp, move.maxpp - 2);
 	});
+
+	it('should not be able to choose rest in Gen 3 battles', function() {
+		battle = common.gen(3).createBattle([[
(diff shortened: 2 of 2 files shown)


DaWoblefet:
This behavior isn't cartridge-accurate: Sleep Talk can in fact call Rest in Gen III: https://www.youtube.com/watch?v=tRsS8FJETwM

If you're trying to address https://github.com/smogon/pokemon-showdown/projects/3#card-85838170, the issue is that Sleep Talk shouldn't be able to call Rest and then Rest be successful.

frank113:
> This behavior isn't cartridge-accurate: Sleep Talk can in fact call Rest in Gen III: https://www.youtube.com/watch?v=tRsS8FJETwM
> 
> If you're trying to address https://github.com/smogon/pokemon-showdown/projects/3#card-85838170, the issue is that Sleep Talk shouldn't be able to call Rest and then Rest be successful.

Understood, thanks for clarifying! I updated the code to allow for sleep talk to call rest yet ensured that if rest is called it will fail. The corresponding test in `sleeptalk` was updated as well.

data/mods/gen3/moves.ts
```diff
@@ -550,6 +550,11 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 				this.add('cant', pokemon, 'nopp', randomMove.move);
 				return;
 			}
+			if (randomMove.move === 'rest') {
```
DaWoblefet, line 553: This isn't the right place either. The failure condition should be in Rest. Rest _itself_ should be the thing failing. Sleep Talk should successfully be able to call Rest.

DaWoblefet: I do notice that Sleep Talk itself doesn't have a proper message sent on failure either in 3 (https://replay.pokemonshowdown.com/gen3customgame-1802139448), so there may be something in both Sleep Talk and Rest to be fixed? Or maybe it has to do with the move call. Either way, hardcoding Rest somewhere isn't the correct fix.

frank113:
I reverted sleep talk to its initial state by removing the hard coding. You had the right idea concerning the need to modify `rest` rather than `sleep talk` -- thanks for that. 

I decided to modify `rest` to check for sleep. Prior to this PR rest does not actually fail in generation 3 while asleep because the move never gets called -- `can't` is logged instead. Placing the sleep check in `onTryMove` sidestepped the default `onTry` and ensures that the failure condition for rest is only met via `sleep talk`.

What do you think about the use of `onTryMove` vs. another handler? I worry about interactions in which `rest` can be called while asleep, for example `sleep talk -> metronome -> rest`.

DaWoblefet:
Apologies for the long time delay here again, but this does not work either. In fact, the current implementation overrides behavior to ensure Rest doesn't work at full HP or if the user has Insomnia/Vital Spirit.

See my comments here about the correct fix: https://github.com/smogon/pokemon-showdown/pull/9556#issuecomment-1605791334

PR #9408, opened 2023-02-21
Fix current generation reference on /ms

JagSamurott:
A fix based on this bug: https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9493133 

Galar is changed to Paldea appropriately


DaWoblefet:
In https://github.com/smogon/pokemon-showdown/pull/9396 I rewrote this to be

``<code>/ms</code> will search all non-dexited moves (clickable in that game); you can include dexited moves by using <code>/nms</code> or by adding <code>natdex</code> as a parameter.``

I think using that kind of language is preferable, since technically ms8 wouldn't search Paldea moves.

JagSamurott:
Ah, I missed that, and it makes more sense. Was just looking for some bugs to fix and will close this

PR #9409, opened 2023-02-21
Scavengers: Make /scav queue broadcastable

PartMan7:
Did this use a function.toString().includes call before?


MathyFurret:
> Did this use a function.toString().includes call before?

It [still does](https://github.com/smogon/pokemon-showdown/blob/master/server/chat.ts#L1963)

PartMan7:
I- wha-

If reading is a skill, I have a skill issue.

(This still fixes the intended bug with /scav queue being unbroadcastable at the moment)

monsanto:
Looks OK to me, thanks

PR #9410, opened 2023-02-21
Codeowners: Fix alphabetical order in chat-plugins

PartMan7:
That _one_ line being out of alphabetical order irked me


CODEOWNERS
```diff
@@ -20,7 +21,6 @@ server/chat-plugins/rock-paper-scissors.ts @mia-pi-git
 server/chat-plugins/sample-teams.ts @KrisXV
 server/chat-plugins/scavenger*.ts @xfix @sparkychildcharlie
```
AustinXII, line 22: should we get rid of sparkychild from this?

KrisXV: this is something that can be discussed for another pr, so im merging this as-is for now

PR #9412, opened 2023-02-23
Gen 9 Random Battle: Fix STAB enforcement

MathyFurret:
This allows more than one STAB move to be added if necessary. Most notably this makes Freeze Dry correctly forced on mons that have it and another Ice move.

### DIFF
--- a/test/random-battles/gen9.js
+++ b/test/random-battles/gen9.js
@@ -3,7 +3,7 @@
  */
 'use strict';
 
-const {testTeam} = require('./tools');
+const {testTeam, testAlwaysHasMove} = require('./tools');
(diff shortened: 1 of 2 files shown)


test/random-battles/gen9.js
```diff
@@ -60,6 +60,10 @@ describe('[Gen 9] Random Battle', () => {
 			}
 		}
 	});
+
+	it("should always give Iron Bundle Freeze Dry", () => {
```
KrisXV, line 64:
```suggestion
	it("should always give Iron Bundle Freeze-Dry", () => {
```

ACakeWearingAHat (approved):

PR #9416, opened 2023-02-26
Hackmons Cup: Support Force Select

WeWuzNidokangz:
Feature requested by OM Mashups staff. Makes the Force Select rule function (properly) in Hackmons Cup formats by requiring the team generator to include the specified Pokemon. Improves tour code validation of Force Select.

* While testing this, I realized that currently, rather than doing nothing, adding Force Select to a generated team tour/battle can create broken battles where one or both players may be unable to make a valid move because they are required to select a Pokemon that doesn't actually exist on their team. This change prohibits this, but even if this feature is rejected, some defense against this should be added to Showdown. Example:-

```
/tour new [Gen 9] Random Battle, Elimination
/tour rules Force Select=Pikachu, Team Preview
/tour autostart 5
/tour autodq 2
```

* It struck me that data/mods/gen8/random-teams.ts currently has a very large amount of seemingly redundant code with the main random-teams.ts, though I didn't look at this deeply so maybe there is some good reason for this that I don't understand.


WeWuzNidokangz:
[b145196](https://github.com/smogon/pokemon-showdown/pull/9416/commits/b145196125293eb8110cc30a4009efcfd8ab28ce)

I don't know why this seemingly regular line return triggered "Trailing spaces not allowed" here. I thought it was a bit more readable to have a gap but it's all good I guess...

WeWuzNidokangz:
[c79c93a](https://github.com/smogon/pokemon-showdown/pull/9416/commits/c79c93afdc698ac3ef97186595c793d459ab6a5f)

Ironically, it seems like there is a good reason to override <Gen 8 behavior now (maybe just Gen 1). This line from here (https://github.com/smogon/pokemon-showdown/pull/9427) can't be added to the <Gen 8 because it makes Gen 1 Hackmons Cup unlaunchable as it doesn't support bans yet the format data bans Nonexistent. I probably should try adding ban support for Gen 1 HC but it's beyond the scope of this PR.

KrisXV:
@livid-washed @MathyFurret @ACakeWearingAHat thoughts?

data/mods/gen8/random-teams.ts
```diff
@@ -304,9 +304,12 @@ export class RandomGen8Teams {
 
 	/**
 	 * Check if user has directly tried to ban/unban/restrict things in a custom battle.
-	 * Doesn't count bans nested inside other formats/rules.
+	 * Doesn't count bans nested inside other formats/rules except Force Select.
 	 */
 	private hasDirectCustomBanlistChanges() {
+		// Force Select is functionally a ban on not including a specific Pokemon.
```
Slayer95, line 310: This premise is actually wrong. That should only happen in Species Clause formats. Which, granted, Hackmons Cup effectively is, but it's still the case that this is the wrong level to enforce Species Clause.

WeWuzNidokangz: Currently Force Select seems to have no relationship to Species Clause. For example, when creating battles like this...

```
/tour new [Gen 9] Anything Goes, Elimination
/tour rules Force Select=Pikachu
/tour autostart 5
/tour autodq 2
```

...the tour and battles can start. Each player has to include at least one Pikachu on their team, but since there is no Species Clause, they may have more than one each.

Is this behavior a bug? If needed, Force Select could require Species Clause in its onValidateRule to prevent this. But if the current behavior doesn't cause any other bugs, it feels like this would just be cutting out functionality that somebody might want to use.

KrisXV:
@livid-washed @ACakeWearingAHat bump

PR #9422, opened 2023-02-27
Add descriptions for new moves

Glazelf:
Tried to keep them in line with the general structure and phrasing of other move descriptions.  
Other weather-related moves' long description also specify how the move is affected by Primal weathers. 
I don't know how Hydro Steam is affected by Desolate Land, but I assume it gets nullified still. Regardless I chose to avoid mentioning this weather.


Marty-D:
Thanks, I would have made my changes directly in here and merged, but GitHub doesn't think this file exists on your branch and wouldn't let me.

Glazelf:
Weird. I have no idea what would have caused that haha. Thanks for the heads up, though!

PR #9423, opened 2023-02-28
Fix Fly Smack Down Stomping Tantrum interaction

aden243:


data/conditions.ts
```diff
@@ -305,6 +305,7 @@ export const Conditions: {[k: string]: ConditionData} = {
 		},
 		onEnd(target) {
 			target.removeVolatile(this.effectState.move);
+			pokemon.moveThisTurnResult = null;
```
MathyFurret, line 308: `pokemon` is not defined here. please test your code!

AnnikaCodes:
Is there possibly a more substantive title for this PR?

Zarel:
While looking at the code, I had a suspicion that it might work if I fixed the typo, and it turns out it does seem to. What does everyone think, given that?

Zarel:
> This won't work if the two-turn move fails normally (e.g. if you Dive into a Storm Drain, or Fly misses). Stomping Tantrum should get 150 BP there, but this change would always make the result null after removing the particular twoturn volatile.

Sounds like we need a test about that.

Zarel:
Okay, I'm pretty sure this is unmaintained, so I'm closing this now. Feel free to reopen if you want to try again at fixing this, @aden243.

PR #9425, opened 2023-02-28
Tera Donation: Fixing Tera Blast

Seerd:
Tera Blast was staying Normal, changing the type at the same type it changes the category solves this.


KrisXV:
This isn’t the proper fix, it’d be easier to just fix the typo that’s causing this in the first place

PR #9427, opened 2023-02-28
Add Randomized Format Spotlight for March

MathyFurret:
if there's a more Proper Order for these lmk and i will make a script to sort them


KrisXV:
Alphabetical

livid-washed:
I'm pretty certain you need to add the bans/unbans as `customRules`, rather than ban/unban list. `randomHCTeam()` only looks at the custom rules, not the banlist or unbanlist. (Currently, Broken Cup is the same as Hackmons Cup).

MathyFurret:
> I'm pretty certain you need to add the bans/unbans as `customRules`, rather than ban/unban list. `randomHCTeam()` only looks at the custom rules, not the banlist or unbanlist. (Currently, Broken Cup is the same as Hackmons Cup).

well thats dumb and counterintuitive, so i will allow it to check for the other properties too

urkerab:
I always thought the order was Pokémon, then Abilities, then Items, then Moves, sorted alphabetically within each group.

MathyFurret:
> I always thought the order was Pokémon, then Abilities, then Items, then Moves, sorted alphabetically within each group.

ah thanks, i wasnt sure where items went in this. i can fix it

WeWuzNidokangz:
> > I'm pretty certain you need to add the bans/unbans as `customRules`, rather than ban/unban list. `randomHCTeam()` only looks at the custom rules, not the banlist or unbanlist. (Currently, Broken Cup is the same as Hackmons Cup).
> 
> well thats dumb and counterintuitive, so i will allow it to check for the other properties too

I admit this is unintuitive since it's different to how non-random formats work, but the reason I did it like that was because if it checked ALL bans, it would include bans effected by format stacking OMs that don't have their own rules with the intention of only adding their mechanics, which would annoying to deal with at best (having to re-unban all of the vanilla OM Pokemon bans etc individually) and at worst could remove the functionality of adding the OM mechanic at all from many formats by throwing the custom banlist unsupported error.

I think adding a reference to bans only on the base format like this PR does is a good change, though this may be the first time a built-in randoms format has a meaningful banlist(?)

PR #9429, opened 2023-03-01
Implement new Order Up mechanics

Karthik99999:
https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/post-9525866

While this is not a perfect implementation as this makes the boost not occur if the user has Sheer Force even though in game the boost happens no matter what, it should be good enough for now as it's more important to not have the boost happen when missing/hitting into a protect.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -13230,11 +13230,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1},
-		onUseMoveMessage(source, target, move) {
-			move.orderUpBoost = true;
-		},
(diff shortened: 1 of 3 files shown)


data/moves.ts
```diff
@@ -13230,11 +13230,8 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 10,
 		priority: 0,
 		flags: {protect: 1},
-		onUseMoveMessage(source, target, move) {
```
urkerab, line 13233: It might be possible to just use a different event here; `TryHit` might work (the `singleEvent` version runs shortly before a move hits) or there's always `AfterDamage` and `AfterSubDamage`.

PR #9435, opened 2023-03-01
VGC/BSS: Add Regulation C

sabertazimi:


DaWoblefet:
We typically start BSS off as non-ladderable (just add ``searchShow: false``). I also want to verify these are going to be the English names of the format.

sabertazimi:
> 

Fixed for BSS.

DaWoblefet:
It's almost certainly going to be the same in English (in the [English publication for "Season 4"](https://sv-news.pokemon.co.jp/en/page/47.html), they started to call Series 2 "Regulation B") but I would prefer to hold out until we see something official in English just to make sure.

DaWoblefet:
Apparently the official name is 'Regulation Set C', but they also use 'regulation set Regulation B' as seen previously, so we'll go with the shorter of the two. https://sv-news.pokemon.co.jp/en/page/75.html. Thanks sabertazimi!

PR #9440, opened 2023-03-03
Hackmons Cup: Stop banning all natures when there are custom bans

adrivrie:
Currently when generating a team for hackmonscup with any sort of custom ban, all natures get banned. This is because of a weird typo(?) that has persisted for a long time and affects all gens. The code checked the same condition twice in a row: `if (!hasCustomBans)` and the `else` was misplaced.

With the introduction of brokencup in  #9427 the check for custom bans changed so that the default gen9HC now has custom bans (specifically, it bans `'Nonexistent'`), so natures were accidentally removed from default HC. The bug always existed but was never noticed before.

### DIFF
--- a/data/mods/gen8/random-teams.ts
+++ b/data/mods/gen8/random-teams.ts
@@ -766,7 +766,6 @@ export class RandomGen8Teams {
 				}
 				// There is no 'nature:nonature' rule so do not constrain pool size
 			}
-
 		}
 
 		const randomN = this.randomNPokemon(this.maxTeamSize, this.forceMonotype, undefined,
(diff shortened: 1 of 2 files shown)


ACakeWearingAHat (approved):

livid-washed (approved):

WeWuzNidokangz:
> Currently when generating a team for hackmonscup with any sort of custom ban, all natures get banned. This is because of a weird typo(?) that has persisted for a long time and affects all gens. The code checked the same condition twice in a row: if (!hasCustomBans) and the else was misplaced.

Yeah, I think it was probably a mistake of pasting something twice, sorry. Thanks for finding and fixing this.

KrisXV:
nice catch

PR #9447, opened 2023-03-06
Validator: Identify moves learned only in a later gen as sketched in previous gens

HisuianZoroark:
Previous logic caused Necturna to learn Shell Smash in previous gens without restriction because sources had a value, and thus any moves Necturna learned in future generations also passed the validator as "naturally learned" in previous generations. This seems to fix it but unsure if there is a better way.


Marty-D:
Ah I had this on my to-do list and I can check it off, thanks! Looks like a fine solution to me 👍

PR #9459, opened 2023-03-12
Almost Any Ability: Ban Good as Gold

Thinkerino:
Post: https://www.smogon.com/forums/threads/almost-any-ability-several-bans-at-post-485-good-as-gold-ban-518.3710568/post-9539286


AnnikaCodes:
Thanks!

PR #9466, opened 2023-03-14
Abilities: Update Sheer Force's description to include Gen9

PartMan7:
...which is just Anger Shell


PartMan7:
(No one saw that typo)

PR #9467, opened 2023-03-14
FIx crash with Force Monotype

KrisXV:


KrisXV:
Umm... did not mean to make this a pullreq. oops!

PR #9468, opened 2023-03-15
Mirror Armor: Fix crash

HisuianZoroark:


KrisXV:
Merging this now because our tests still pass + it fixes a crash, if someone finds a better solution down the line feel free to make another pull request

PR #9470, opened 2023-03-18
Add CI action to automatically publish new releases

### LINKED ISSUE #9428: Update the npm package
Hi there! I was trying to install the package locally on my machine, however I keep running into typescript compilation errors.

I asked around on the Typescript discord, and after investagion we found that the issue was due to this export in this file
```typescript
// node_modules/pokemon-showdown/.sim-dist/index.d.ts
export * from '../lib';
```
However this issue appears to have since been fixed in this repository, and just requires being published to npm

Wazbat:
Apologies for the delay in getting started with this, but here is what I have so far regarding automated creation of releases. This should be ready to go if it looks satisfactory to you, once the repo secret is created

I wanted to check if this is satisfactory for you @monsanto as you mentioned keeping this convenient, easy and automated.

With this CI, in order to update the NPM package, all that is necessary is a new github release to be created, which anyone with permissions should be able to do in ~15 seconds with 4 clicks and typing a new version number (github will auto generate release notes too with a button)

As you can see by the workflow, it'll run the automated tests, and if they pass it'll do the following:
~~1. Install dependencies~~
~~2. Run `npm run build`~~
~~3. Bump the version in the package.json~~
~~4. Run `npm publish`~~
~~5. If all the above succedes, it'll then create an automated PR to update the package.json version on master~~
1. Compare the package.json version numbers between the current commit and the previous push
2. If diferent, build the code, then publish the npm package

Resolves #9428


Zarel:
Hm. Is this even the hard part? The main reason _I_ haven't published a new release is that we transitioned to `tsnode-swc` and stopped maintaining the old compiler, and we never found time to make sure everything still worked.

monsanto:
@Zarel Did you read the linked thread and my rationale for doing this? If you're willing to take responsibility for manual releases then whatever but I think we'd be in a stronger position if they were automated. It's another thing for you to remember to do and if you happen to take a break the npm packages will likely once again be in disrepair, and it will be unpleasant for me (or whoever else) to deal with.

Wazbat:
@monsanto I could update the CI to make this even more automated, however I think just using github releases is a good middleground

You don't have to keep the releases regularly up to date with every commit or anything. But just even creating a release once every few months would be more than enough. It's only 4 clicks and takes a few seconds, and I think anyone with the apropriate permissions can do it. Everything is fully automated from then on

You honestly don't even need to keep doing them. If the package ever has issues again like the current one, someone can just open an issue here in the repo asking for a new release and someone can do it, as again it takes just a few seconds.

I'd be happy to do them myself too if necessary

Zarel:
@monsanto your posts in the linked thread don't really get into why we don't currently have regular releases.

This PR automates a part that I was not aware was a bottleneck – publishing to NPM. The reason we stopped publishing to NPM is because we switched from sucrase to tsnode-swc, and no one had time to keep maintaining the NPM package. The PR to switch said something along the lines of "no I don't have time to add support for building an NPM package".

The reason I have not published NPM packages since then is because I was not aware that our NPM package build system was still working.

Zarel:
I do apologize for putting this here, though; you're right that this PR does add useful automation even if it isn't the bottleneck.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          title: Bump package.json version to ${{ github.ref_name }}
+          body: |
+            Bump package.json version to ${{ github.ref_name }}
+          branch: npm-version-bump
+          base: master
+          labels: |
+            dependencies
+            version-bump
```
Zarel, line 36: It feels weird to create a PR for this. Especially since we should probably update `package.json` before tagging a new version, rather than after.

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
+          cache: 'npm'
+      - run: npm ci
+      - run: npm run build
```
Zarel, line 19: Apparently `npm ci` already does this. I suppose doing it again can't hurt, so the question is more "is it more important to futureproof against changes where `npm ci` no longer builds, or more important to have the release workflow accurately reflect what steps do what."

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+      # Bump the version in package.json
```
Zarel, line 20: What's this comment for?

.github/workflows/release.yml
Zarel: Thinking more about this, this seems outright wrong to me. The published NPM package's `package.json` needs to match the version number passed to `npm version`, which implies that `package.json` _must_ be updated before tagging.

.github/workflows/release.yml
Wazbat: True. I was looking inside the build script and did see that it also installed dependencies. My main reason for doing it here was just in case that script ever changed, as _in theory_ all a build script should do is build, not install dependencies

I don't think there's any harm in running the dependency install twice. `npm ci` is the same as `npm i` if I recall correctly, but it just fails if the package.json and package-lock.json don't match, as an extra layer of protection

.github/workflows/release.yml
Wazbat: Just to explain what exactly the step after does, in case someone viewing the file is unfamiliar with the npm version command

.github/workflows/release.yml
Wazbat: This flow here is designed with the original goal of making the whole process as simple and automated as possible, so with that in mind I set it up like this to keep everything simple.

Instead of having to bump the version in the package.json yourself (which I wasn't sure if it was being maintained or not), you simply create a GitHub release.
The CI then checks out the repo locally, and uses `npm version` to locally bump the version in the package.json to the same number as the tag that was created using the github release.
Then, in the next step, when npm publish runs, locally the version number in the package.json is bumped, so it'll publish that new version

However, this commit is only local in the CI machine, so to keep the repo in sync, it then opens a PR with the change to the version number in the package.json

Wazbat: This can be changed of course, but the alternative to this is a user manually bumping the version on their local machine, opening a PR, merging it, then creating a release with that exact same version

It's a lot more manual steps with room for error for the same outcome

Wazbat: Oh!!! I think I understand what you mean

While the published npm version will have the correct version number, the git tag/release's package.json will still technically be on the previous version

I can see why that would be undesirable, if not outright wrong

If you want I can change the whole flow so that instead of being triggered on release, it triggers on commit

If a commit to the master branch bumps the package.json version, that will then be published to npm via `npm publish`

I can also add a separate super simple workflow that is triggered manually via the UI, where from a drop-down you simply select the version to bump (major, minor, patch) with a single click and it'll automatically create a PR to update the version

Wazbat:
@Zarel I've reworked the CI since your last review

Now, we only publish to npm on push to master, if the version field is changed. This can be done either manually by a person, or using a new workflow I added, which can be triggered directly from the actions tab on github. 

I don't have an example of this exact workflow, but clicking this button in the UI will now trigger the action, which will open a PR to update the version field in the package.json and package-lock.json files
image

monsanto:
> @monsanto your posts in the linked thread don't really get into why we don't currently have regular releases.
> 
> This PR automates a part that I was not aware was a bottleneck – publishing to NPM. The reason we stopped publishing to NPM is because we switched from sucrase to tsnode-swc, and no one had time to keep maintaining the NPM package. The PR to switch said something along the lines of "no I don't have time to add support for building an NPM package".
> 
> The reason I have not published NPM packages since then is because I was not aware that our NPM package build system was still working.

I feel like we are talking past each other. What I asked for in the PS repo, and in the calc repo, is an automated build and release system that has zero overhead for our organization. If the build script has decayed, then that is evidence for my case--I don't have time to support it. The interface to building needs to be stable and shared between what we use in production (which will actually be tested on a regular basis) and what is used for NPM (which won't). If you aren't around to publish NPM packages, then that is evidence for my case--I don't want to learn how to do it, or bother with it. Whether or not this PR actually achieves this goal is besides the point, however.

.github/workflows/release.yml
Zarel: We seem to have a postinstall script that builds.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
+name: Publish to npm
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'package*.json'
```
Zarel, line 7: Is there any reason this isn't just `package.json`? If only `package-lock` is changed, that's probably bad and should be ignored, right?

```suggestion
      - 'package.json'
```

.github/workflows/release.yml
Wazbat: Oh huh, I didn't see that script... I suppose it does. I'm a little confused what the best practice here is then. Installing dependencies builds the app, but also building the app installs dependencies

We could keep things simple and just only run `npm ci`, but if that postinstall script is ever removed it'd break things

In theory there's no harm in running both the install and build steps in CI, so perhaps for clarity we should leave them both in?

Zarel (approved):
Looks great, other than my suggested change! I'd prefer not to ever use the PR workflow but there's nothing wrong with having it in there.

.github/workflows/publish.yml
Wazbat: Yeah I suppose you're right there. I added the wildcard thinking "oh both the files can change so we should watch both" but thinking about it yeah, we only need to watch the package.json

.github/workflows/release.yml
Zarel: I don't have a strong opinion here. It seems you prefer to leave both in, which is fine by me.

Wazbat:
> I'd prefer not to ever use the PR workflow but there's nothing wrong with having it in there.

That's fair! Yeah I just added it for convenience. Might be fun to test if it works later, but there's no need to use it

Wazbat:
Oh yeah, the last thing to check before or after merging. Someone with permission needs to create the appropriate npm secret. Not sure if that's already been done. The name has to march the secret name in the CI

Wazbat:
Just to make sure this doesn't get stale @Zarel @monsanto , who is in charge of creating the NPM secret? That's the only thing needed for this workflow to work

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+      - test
+      - get-version
+    runs-on: ubuntu-latest
+    # We only want to publish if the package.json version field has changed
+    if: needs.get-version.outputs.current_version != needs.get-version.outputs.last_version
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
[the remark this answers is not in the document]
AnnikaCodes: Does the package lock in the repository not work under Node 14?
Anyway, our `tsconfig.json` should tell the build system to target es2020 (which is compatible with Node 14) regardless of the current Node version: https://github.com/smogon/pokemon-showdown/blob/70759c5c13ff1d4ea56ee63d44a5202e67f20eee/tsconfig.json#L2-L3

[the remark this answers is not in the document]
Wazbat: > Does the package lock in the repository not work under Node 14?

I don't think so. I tried on my local machine via nvm and it fails, but 19 is fine. If it's ok to change that I'll update the CI to use a modern version of node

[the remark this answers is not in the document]
Dirain1700: npm v6 or lower cant parse the lockVersion v2 or higher, should install by `npm i`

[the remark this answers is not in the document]
Wazbat: I've updated to a modern version of node which fixes this

.github/workflows/update_version.yml
```diff
@@ -0,0 +1,42 @@
...
+          cache: 'npm'
+      - id: bump_version
+        run: |
+          echo "old_version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
+          npm version ${{ github.event.inputs.version }} --git-tag-version false
+          echo "new_version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
+
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          title: Bump package.json version to v${{ steps.bump_version.outputs.version }}}
```
Dirain1700, line 36:
```suggestion
          title: Bump package.json version to v${{ steps.bump_version.outputs.new_version }}}
```
https://github.com/Dirain1700/workflows-test/pull/26

Dirain1700:
@Zarel @monsanto @AnnikaCodes
at first sorry for mentioning

When running `npm version` and on npm v6 or lower, npm couldn't change the package version on package-lock.json, which is resolved on vpm v8 (Node v16) or higher.

CC0758D3-1675-4AF6-AB79-64B4348708B4

Can you choose from
1: Ignoring package-lock.json
2: Update Node.js version to v16

Thanks.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,47 @@
...
+      - get-version
+    runs-on: ubuntu-latest
+    # We only want to publish if the package.json version field has changed
+    if: needs.get-version.outputs.current_version != needs.get-version.outputs.last_version
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 19
+          cache: 'npm'
```
Dirain1700, line 42:
```suggestion
          cache: 'npm'
          always-auth: true
```

The CI fails unless adding this line: https://github.com/Dirain1700/workflows-test/actions/runs/4520306972



Have you tested these CIs? Follow my example CI and change it

[Actions link](https://github.com/Dirain1700/workflows-test/actions/runs/4520384963)

[Auto-commit link](https://github.com/Dirain1700/workflows-test/commit/73b7a77a6fea7727fda1f7c675cf5ff183fd1733)

monsanto:
We should upgrade node 14 in production at this point, it's really old.

.github/workflows/update_version.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 19
+          cache: 'npm'
+      - id: bump_version
+        run: |
+          echo "old_version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
+          npm version ${{ github.event.inputs.version }} --git-tag-version false
```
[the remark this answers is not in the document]
Wazbat: Are you sure that's supported? I was looking in the npm docs and don't see that mentioned
https://docs.npmjs.com/cli/v8/commands/npm-version#git-tag-version

It might be supported, but deprecated

.github/workflows/update_version.yml
Wazbat: Oh my bad!

Also thanks for testing these!

.github/workflows/publish.yml
Wazbat: I think that's failing due to the NODE_AUTH_TOKEN not being set though. Do you have a NPM_TOKEN set in that repository you're testing it in? Otherwise publish will never work
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages

You are correct though that that always-auth is necessary, but as per the github actions docs, adding a `registry-url` should be all that's necessary

This workflow is pretty much identical to the docs there

Dirain1700: Of course, I had a secret named NPM_TOKEN on that repo (if what you said is true, then my CI must have failed)

anyaways, there is nothing wrong with having authority, i will recommend this.

Wazbat: Could you clarify what you mean by having authority? Do you mean the argument to setup node?

.github/workflows/update_version.yml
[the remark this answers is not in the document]
Dirain1700: > If run in a git repo, it will also create a version commit and tag. This behavior is controlled by git-tag-version (see below), and can be disabled on the command line by running npm --no-git-tag-version version. It will fail if the working directory is not clean, unless the -f or --force flag is set.

from [v9 docs](https://docs.npmjs.com/cli/v9/commands/npm-version)

.github/workflows/publish.yml
Dirain1700: In fact, your prediction is off. You should specify explicitly
2D46E0A3-1BFB-48AC-A9F5-DCE7815726F5

Wazbat: Oh yeah, you're 100% correct. If ommited it does indeed set that flag to false
image

I just want to check though, are you sure that tag is necessary? Every single online sample GA npm publish workflow I've found does not state that option.

Why is it required? Do we know _why_ it's failing without it?

Dirain1700: that image is your script and failing, i specified that on my script

Wazbat: Aye, however I've since added the `registry-url` which was missing from mine, could you edit yours to include that option? From what I've been searching, the always-auth option shouldn't be necessary, just the `registry-url` which I forgot to add

.github/workflows/update_version.yml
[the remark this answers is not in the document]
Wazbat: Ahhh, yep I was looking at the wrong docs, thanks!

.github/workflows/publish.yml
Dirain1700: Yeah, `always-auth` is not needed, thats my bad.

Wazbat: No problem! Yeah I forgot the registry in the first place so that probably would have fixed it instead

Zarel:
Hi! Sorry this fell off my radar, I've been working on other stuff. Yeah, we can bump the required Node version to 16.

Wazbat:
That sounds like a good idea, though that shouldn't be necesarry for this PR I think

.github/workflows/publish.yml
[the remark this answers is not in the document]
Dirain1700: anyway v19 will be deprecated soon, use v16 instead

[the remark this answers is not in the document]
Wazbat: Are you sure? v16 is under maintenance. I do agree v19 might not be the best idea after checking it, but isn't v18 the best one to use for LTS?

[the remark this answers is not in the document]
Wazbat: image
image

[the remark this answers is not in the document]
Dirain1700: uh i thought it ends in 2024
but Zarel didn't say using v18, please wait his response, in any case it is not recommended using odd number version.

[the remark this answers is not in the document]
Wazbat: Gotcha, yeah I agree we probably shouldn't use v19. I recently upgraded some other workflows of mine to node 19 based on a dependabot recomendation, however I didn't actually check when supported ended. I assumed it would last a bit longer than this.

In theory we can use any version for the CI, as long as the build works. Developers are free to use any version on their local machines

Just going to @Zarel to confirm what version we should use for the CI

AnnikaCodes:
I think we should use Node 16 for CI so that we maintain compatibility with that version — supporting all current and security-supported LTS releases seems doable to me. (We could also consider running CI on v16, v18, and v20 via a matrix if we really want to.)

Wazbat:
Sorry for the ping @Zarel , but we're just waiting for confirmation about what node version to use in the CI runners, I don't want this PR to get stale as I think we're pretty much sorted

Zarel:
We had a build issue so we're on Node 16 for CI now. It looks like this is set to build with Node 19, which seems fine to me?

Zarel:
As far as I'm aware, the main thing this is blocked on is getting some secrets put in the right places, which I'm working on now.

.github/workflows/publish.yml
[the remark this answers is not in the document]
Zarel: Okay, yes, I'd prefre 18.

Wazbat:
> We had a build issue so we're on Node 16 for CI now. It looks like this is set to build with Node 19, which seems fine to me?

19 works fine, the only issue is it's going to be deprecated soon as it's an odd numbered version. 18 is the one with longest support at the moment

Dirain1700:
(Sorry from side) When I tested this CI, I used Read/Write Granular Access Tokens (Click `Access Tokens` on top page)
Organization is not required, if the token is only for pokemon-showdown, i think

IMG_6762

Wazbat:
A lot of the guides also mention specifically using Automation tokens too, something to keep in mind

image

Zarel:
All right, looks like everything is ready. Are you on Discord? I'd be nice to have you on call if after merging, we try it and something breaks.

Wazbat:
Oh sure can do, I'll join. I'm Wazbat#0387

PR #9475, opened 2023-03-24
Gen 1: Fix stat dropping at 1

livid-washed:
If a Pokemon has a stat of 1 and it is the target of a stat drop, the stat should not be recalculated.


AnnikaCodes:
Thanks!

PR #9476, opened 2023-03-25
Random Battles updates

ACakeWearingAHat:
-Substitute users will always get Soundproof when possible (Hisuian Electrode)
-Garchomp can get Spikes now.
-Farigiraf has gained Tera Normal on its Future Sight WishTect set.
-Cloyster is no longer a Tera Blast user, and instead uses Tera Ground Drill Run.
-Hyper Voice will now always appear on Normal-types that can run it.
-Arboliva now gets Ground Tera as an option to compensate for forced Hyper Voice decreasing the rate of Earth Power.
-Hisuian Zoroark can now also use Tera Fighting, and will get Focus Blast more often.
-Sandy Shocks can now run a Choice item Tera Blast set. It is Tera Flying or Ice.

### DIFF
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -173,7 +173,6 @@ export class RandomTeams {
 			Ice: (movePool, moves, abilities, types, counter) => (movePool.includes('freezedry') || !counter.get('Ice')),
 			Normal: (movePool, moves, abilities, types, counter) => {
 				if (movePool.includes('boomburst') || movePool.includes('hypervoice')) return true;
-				return (!counter.get('Normal') && movePool.includes('futuresight'));
 			},
 			Poison: (movePool, moves, abilities, types, counter) => {
 				if (types.includes('Ground')) return false;
--- a/data/random-teams.ts
+++ b/data/random-teams.ts
@@ -171,9 +171,7 @@ export class RandomTeams {
 			),
 			Ground: (movePool, moves, abilities, types, counter) => !counter.get('Ground'),
 			Ice: (movePool, moves, abilities, types, counter) => (movePool.includes('freezedry') || !counter.get('Ice')),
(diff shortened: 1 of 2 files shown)


data/random-teams.ts
```diff
@@ -172,7 +172,7 @@ export class RandomTeams {
 			Ground: (movePool, moves, abilities, types, counter) => !counter.get('Ground'),
 			Ice: (movePool, moves, abilities, types, counter) => (movePool.includes('freezedry') || !counter.get('Ice')),
 			Normal: (movePool, moves, abilities, types, counter) => {
-				if (movePool.includes('boomburst')) return true;
+				if (movePool.includes('boomburst') || movePool.includes('hypervoice')) return true;
 				return (!counter.get('Normal') && movePool.includes('futuresight'));
```
livid-washed, line 176:
```suggestion
```

data/random-teams.ts
```diff
@@ -172,7 +172,7 @@ export class RandomTeams {
 			Ground: (movePool, moves, abilities, types, counter) => !counter.get('Ground'),
 			Ice: (movePool, moves, abilities, types, counter) => (movePool.includes('freezedry') || !counter.get('Ice')),
 			Normal: (movePool, moves, abilities, types, counter) => {
-				if (movePool.includes('boomburst')) return true;
+				if (movePool.includes('boomburst') || movePool.includes('hypervoice')) return true;
```
livid-washed, line 175:
```suggestion
				return (movePool.includes('boomburst') || movePool.includes('hypervoice'));
```

livid-washed: Forcing Normal STAB on Farigiraf's sets with Future Sight is no longer needed, since Hyper Voice is forced.

livid-washed (approved):

PR #9477, opened 2023-03-26
Mix and Mega: Fix error message

HisuianZoroark:
The crash was because when an ability such as Protosynthesis, Quark Drive, or Slow Start ends, the pokemon tries to delete a volatile, showing a client error in the process (probably because the pokemon wasn't out?)


KrisXV:
thx

PR #9478, opened 2023-03-26
Update Gen 9 format URLs

KingNeodude:
Added Metagame Discussion to Gen 9 OU
Added Sample Teams to Gen 9 OU
Added Viability Rankings to Gen 9 OU, UU, RU, NU


AnnikaCodes:
Thanks!

PR #9483, opened 2023-03-30
Scavengers: Update the Minesweeper twist description

PartMan7:
Non-urgent; updating the description with the how-to for the future. Have tested the twist itself on main; everything's working swimmingly!


PartMan7:
...forgot to actually pull first, whoops

PR #9484, opened 2023-03-30
Room-settings: Reject setting modjoin past % in groupchats

Dirain1700:
I could set modjoin to @ by using `/modjoin sync`, maybe this should be blocked?


Dirain1700:
To be honest, I don't think it need to ban setting modjoin to trusted/whitelist, do you have any reasons to keep it this?

Dirain1700:
image

AnnikaCodes:
Global staff should be able to join groupchats yes. What happens if you set modchat to @ and *then* turn on `/modjoin sync`? Should there be a check in the modjoin command also?

Dirain1700:
Yes, currently it does so should check in modchat command.
IMG_7246

PR #9487, opened 2023-03-31
Old Gens Random Battles updates

livid-washed:
Gen 1: 
- limit teams to 2 Fire weaknesses
- Pokemon with no Physical attacks (and no Mimic or Transform) will get minimal Attack EVs/DVs

Gen 3: Electrode wants Static rather than Soundproof

Gen 4:
- some set updates, mostly for mons with low winrate, see the pin in the gen 4 channel
- set Spinda to level 100

Gen 5: Replace Wrap with Knock Off on Shuckle

Gen 6-7:
- Add Mega-Rayquaza, at level 67 in both gens
- A couple minor set changes, including removing Dragon Ascent from regular Rayquaza
- Delphox wants Blaze rather than Magician

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -2101,7 +2101,7 @@ export const Formats: FormatList = [
 
 		mod: 'gen7',
 		team: 'random',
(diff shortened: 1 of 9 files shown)


AnnikaCodes:
thanks

PR #9492, opened 2023-04-03
Scavengers: Fix minesweeper crashing if someone leaves

PartMan7:
(I'll strongly type Scavs at a later date)


mia-pi-git:
Thanks!

PR #9497, opened 2023-04-05
Add a plugin to manage permanent punishments in-sim

mia-pi-git:
Requested by US. 
Detailed specs are in the US forum, but the rough tl;dr of this is: staff makes nom via chat page with relevant info, server collects modlog/alts/etc and sends it to US in a method similar to helptickets. US can then decide via chat page on what to do (the page provides buttons to change standings for the primary userid, individual alts, and everyone on an IP) then they can resolve it with a reason. On resolving, a post is made in the permalocks thread, detailing username, reason, and IPs (with geolocation data).

I'm not super sold on the API for geolocation data, but it's free and provides all the data we need.


DaWoblefet:
Mia asked me for a review. I was only able to investigate a little bit directly because I don't have a local modlog, but everything here looks safe to merge as a soft launch (since everything is gated behind higher permissions).

mia-pi-git:
Bump. Would like to get this out soon-ish.

AnnikaCodes:
This seems fine to me; has it been thoroughly tested locally? If not I can check it out myself; I have a local modlog setup.

mia-pi-git:
> This seems fine to me; has it been thoroughly tested locally? If not I can check it out myself; I have a local modlog setup.

It has been, yes.

PR #9505, opened 2023-04-07
Gen 1 and Gen 2: Nickname Clause

andrebastosdias:
I don't know if there is a reason for the Nickname Clause not being active in these generations. Although it is not compulsory, its implementation simplifies development by enabling a straightforward selection of a Pokémon by its nickname.

### DIFF
--- a/data/mods/gen1/rulesets.ts
+++ b/data/mods/gen1/rulesets.ts
@@ -2,7 +2,7 @@ export const Rulesets: {[k: string]: ModdedFormatData} = {
 	standard: {
 		effectType: 'ValidatorRule',
 		name: 'Standard',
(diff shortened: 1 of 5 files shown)


DaWoblefet:
Has a tier leader for either Gen 1 or 2 signed off on this change? I also agree it makes intuitive sense to have the clause, but we'd want to see something first from a TL before merging.

Zarel:
We didn't ask before adding Nickname Clause to newer gens, either; I did it purely as a QoL thing, and omitting it from past gens is purely an oversight. I would rather merge, and revert if pastgen councils take issue with this, than leave this open forever because no one can get ahold of them.

Zarel:
Thanks!

Zarel:
Thanks!

PR #9506, opened 2023-04-07
Github Actions: Perform selective eslint & testing for PRs

im-tofa:
A few potential optimizations to the Github Actions job that could be worth trying out, if desired

1. Use [actions/cache@v3](https://github.com/actions/cache) to create a Github-hosted, immutable cache entry that stores a specified directory (in this case, `node_modules`). A new entry is created whenever `package-lock.json` is updated or a new `node-version` is specified in the action itself. Old entries are auto-evicted from the cache after a few days of inactivity or due to lack of space. Similar to what is done [here](https://github.com/prettier/eslint-config-prettier/blob/main/.github/workflows/check.yml), saves about a minute in the `npm install` step once cached
2. Use [tj-actions/changed-files@v35](https://github.com/tj-actions/changed-files) to fetch all changed/modified js, ts & tsx files in a PR, and use eslint only on those. If there's a push to master (force push or PR merge), a full lint is run instead. Saves 1-2 minutes for smaller PRs

The remaining large chunk of time is in `tsc` + `mocha` (~6-8m), mainly `mocha`. The runners on Github Actions also seem to vary quite a lot in performance in general between runs, but idk how much we can do about that

Examples:
1. [push-to-master](https://github.com/im-tofa/pokemon-showdown/actions/runs/4638042638/jobs/8207490831)
2. [pull request](https://github.com/im-tofa/pokemon-showdown/actions/runs/4638087164/jobs/8207571092?pr=2)

NOTE: I bumped the node version in Github Actions to `16.x` since I couldn't figure out why `14.x` won't pass `tsc` (happened in `master` as well). I'll change it back to `14.x` after someone who knows more is able to fix it in master

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,8 @@ jobs:
           
     - run: npm ci
     
-    - name: Get all changed *.js, *.ts & *.tsx file(s)
+    - name: Determine which files to lint (if pull request)
+      if: ${{ github.event_name == 'pull_request' }}
       id: changed-files
       uses: tj-actions/changed-files@v35
       with:
@@ -45,11 +46,8 @@ jobs:
           **/*.ts
           **/*.tsx
 
-    - name: What files where changed?
-      run: |
-        echo ${{ steps.changed-files.outputs.all_changed_and_modified_files }}
-
-    - name: Detect changed directories
+    - name: Determine whether test/sim or test/random-battles need to run (if pull request)
+      if: ${{ github.event_name == 'pull_request' }}
       id: changed-directories
       uses: tj-actions/changed-files@v35
       with:
@@ -59,14 +57,14 @@ jobs:
           sim/**
 
     - name: Run selective lint & neccessary tests (if pull request)
-      run: npm run full-test-ci
       if: ${{ github.event_name == 'pull_request' }}
+      run: npm run full-test-ci
       env:
         CI: true
         FILES: ${{ steps.changed-files.outputs.all_changed_and_modified_files }}
         SKIPSIMTESTS: ${{ steps.changed-directories.outputs.all_changed_and_modified_files == '' }}
 
-    - name: Run full lint & test (if pushing to master)
+    - name: Run full lint & test (if push to master)
       run: npm run full-test
       if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
       env:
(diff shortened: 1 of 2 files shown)


Zarel:
While I don't have a _problem_ with this, I don't think either of these will have much impact on build times, the vast majority of which are spent in Mocha.

Plausibly the most impactful thing we can do is only run sim tests when the sim is changed, which would have a pretty huge impact.

im-tofa:
Thanks for the feedback :)

> Plausibly the most impactful thing we can do is only run sim tests when the sim is changed, which would have a pretty huge impact.

Good point! I could take a look at adding that when I'm able to, should be possible to deduce with the `changed-files`-action. Would it be enough to say "if the `sim` or `data` directories are unchanged, there is no need to run `sim` tests"? I could add this for only PRs or for both PRs & merges/push-to-master

Zarel:
> Would it be enough to say "if the `sim` or `data` directories are unchanged, there is no need to run `sim` tests"? I could add this for only PRs or for both PRs & merges/push-to-master

Yes, `sim` and `random-battles`.

Zarel:
Although `config/formats` does also mean needing to re-run tests.

im-tofa:
@Zarel I added your suggestion now; the workflow will skip tests in `test/sim/**` and `test/random-battles/**` if none of the changes are in `data/**`, `sim/**` or `config/formats.ts`. Non-sim PRs (e.g this PR) run much faster now :)

I also removed the `node_modules` caching, it only worked with `npm install` and felt pretty sketchy anyway. 

All of the optimizations are for PRs only, push-to-master still runs full lint & test just in case 👍

singiamtel:
I'm not an expert in GitHub CI, but aren't you able to do this for changed files instead?

```yaml
on:
  push:
    branches: [ master ]
    paths:
      - "config/formats.ts"
      - "data/**"
      - "sim/**"
  pull_request:
    branches: [ master ]
    paths:
      - "config/formats.ts"
      - "data/**"
      - "sim/**"
```

From [Github docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore)

im-tofa:
> I'm not an expert in GitHub CI, but aren't you able to do this for changed files instead?
> 
> ```yaml
> on:
>   push:
>     branches: [ master ]
>     paths:
>       - "config/formats.ts"
>       - "data/**"
>       - "sim/**"
>   pull_request:
>     branches: [ master ]
>     paths:
>       - "config/formats.ts"
>       - "data/**"
>       - "sim/**"
> ```
> 
> From [Github docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore)

Didn't know this was a thing, thanks for sharing! However while this PR focuses on sim-tests, linting, building & other testing is always done. It doesn't look like this type of conditional execution is possible with the quoted suggestion without having separate workflows

Zarel:
This is exactly what I wanted! Thank you!

PR #9507, opened 2023-04-07
/setodds: disable setodds command for spinoff formats

adrivrie:
Currently /setodds on computergeneratedteams causes a crash, and on challenge cup and hackmons cup gives the wrong result by using the set generator of "normal" rands. The code already assumed it was dealing with regular random battles, this just makes it fail gracefully if that's not the case.


adrivrie:
Something went very wrong here, sorry

PR #9509, opened 2023-04-08
Move Iron Leaves to UUBL

shrianshChari:


KrisXV:
I ended up doing this in https://github.com/smogon/pokemon-showdown/commit/9c0b6c59a7a471e71885d983b3f5005eb71f7768 because these things should be somewhat high priority and the commit history on your branch was a bit botched. Thanks for contributing though, don't let this discourage you!

PR #9513, opened 2023-04-10
Add a chat plugin for storing teams remotely

mia-pi-git:
tl;dr pokepaste, but on PS. 
There are some limits to ensure people don't do what they do with pokepaste - IE, people have to upload valid teams, and there is a limit to start with (though it's possible we could remove that, it's mainly there since i'm being cautious.)

I also included basic searching functionality, and the ability to search publicly shared teams in the same vein as replays.


server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,692 @@
...
+
+export const TeamsHandler = new class {
+	database = new PostgresDatabase();
+	readyPromise: Promise<void> | null = new Promise<void>((resolve) => {
+		if (!Config.usepostgres) {
+			this.readyPromise = null;
+			return resolve();
+		}
+		void this.database.query('SELECT * FROM teams LIMIT 1').catch(async () => {
+			await this.database.query(FS(`databases/schemas/teams.sql`).readSync());
```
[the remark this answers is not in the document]
mia-pi-git: Yeah, fair. Swapped it to an async arrow function.

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,692 @@
...
+		`<button class="button" name="send" value="/j ${context.pageid}" style="float: right">` +
+		` <i class="fa fa-refresh"></i> ${context.tr('Refresh')}</button>`
+	);
+}
+
+export const TeamsHandler = new class {
+	database = new PostgresDatabase();
+	readyPromise: Promise<void> | null = new Promise<void>((resolve) => {
+		if (!Config.usepostgres) {
+			this.readyPromise = null;
```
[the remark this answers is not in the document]
mia-pi-git: (swapped it to arrow function, no longer applies)

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,686 @@
...
+		}
+		if (!includePrivate) where.push('private != true');
+
+		const result = await this.query<StoredTeam>(
+			`SELECT * FROM teams${where.length ? ` WHERE ${where.join(' AND ')}` : ''} LIMIT ${count}`,
+			args,
+		);
+		return result.filter(row => {
+			const team = Teams.unpack(row.team)!;
+			if (row.private && !(row.ownerid === user.id || user.can('rangeban'))) {
```
thejetou, line 76: Why should admins be able to view a user's private teams?

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,686 @@
...
+		if (isUpdate) {
+			existing = await this.get(isUpdate);
+			if (!existing) {
+				connection.popup("You're trying to edit a team that doesn't exist.");
+				return null;
+			}
+		}
+
+		const validatorResult = await TeamValidatorAsync.get(`gen${Dex.gen}customgame`).validateTeam(rawTeam, {
+			removeNicknames: true, // less headache to moderate this way
```
thejetou, line 135: Why? Aren't nicknames filtered anyway?

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,686 @@
...
+			throw new Chat.ErrorMessage(message);
+		};
+
+		if (!Config.usepostgres || !Config.usepostgresteams) {
+			err(`The teams database is currently disabled.`);
+		}
+		if (!Users.globalAuth.atLeast(user, Config.usepostgresteams)) {
+			err("You cannot currently use the teams database.");
+		}
+		if (user.locked || user.semilocked) err("You cannot use the teams database while locked.");
```
thejetou, line 277: Why not?

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,686 @@
...
+			if (teamData.ownerid !== user.id && !user.can('rangeban')) {
+				return this.errorReply("You cannot delete teams you do not own.");
+			}
+			await TeamsHandler.delete(teamid);
+			this.popupReply(`Team ${teamid} deleted.`);
+			for (const page of connection.openPages || new Set()) {
+				if (page.startsWith('teams-')) this.refreshPage(page);
+			}
+		},
+		async setprivacy(target, room, user, connection) {
```
thejetou, line 381: There should be a third level of privacy: unlisted

Currently, your only two options are to either have it completely open or only viewable by yourself. This doesn't support a use case where you built a team for a teammate but obviously don't want it public but want your teammate. Unlisted would basically be a team link but with an extra `password` field that's randomly generated for a team a la private PS replays.

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,686 @@
...
+			buf += `<strong>What's the name of the team?</strong><br />`;
+			buf += `<input name="name" /><br />`;
+
+			buf += `<strong>What's the team's format?</strong><br />`;
+			buf += `<input name="format" /><br />`;
+
+			buf += `<strong>Should the team be private? (yes/no)</strong><br />`;
+			buf += `<select name="privacy" /><option value="1">Yes</option><option value="0">No</option></select><br />`;
+
+			buf += `<strong>Provide the team:</strong><br />`;
```
thejetou, line 522: There should be a way to paste a folder worth of teams which is useful for [team dumps](https://www.smogon.com/forums/threads/smogon-premier-league-xiv-sv-ou-discussion.3714163/page-2#post-9575543).

thejetou (commented):
This is a general design review; I didn't look too hard into the nitty gritty of the code.

server/chat-plugins/teams.ts
mia-pi-git: Not exactly. Right now, we have 'searchable' or 'unlisted' is probably a more accurate way of putting it. Anyone can view a team if they have the number, but it'll only show up in searches if it's marked as public.

mia-pi-git: Now that I say this to myself, team IDs should probably be hashes so they aren't guessed easily.

server/chat-plugins/teams.ts
mia-pi-git: Admins can view private replays, I just applied the same principle here.

server/chat-plugins/teams.ts
mia-pi-git: I don't really think it's necessary to have nicknames in uploaded teams (how is that crucial to the set?) but i'm open to changing my mind on that one.

server/chat-plugins/teams.ts
thejetou: There's a few reasons why admins should be able to view private replays like if a user is boosting on the ladder but I don't see a reason why admins should be able to view private teams since they are pretty sensitive information.

server/chat-plugins/teams.ts
thejetou: It's not crucial information to a set but it provides flavor since some players have standard nicknames they give to specific Pokemon and whatnot. I just don't see the harm in allowing nicknames personally since they're already chat filtered.

server/chat-plugins/teams.ts
mia-pi-git: Fair enough. Done.

server/chat-plugins/teams.ts
mia-pi-git: My thing is like, if we allow nicknames, then staff have to moderate this. And I don't think anyone wants to do that.

thejetou:
A button where the Pokepaste button currently is to upload a team would be nice and ease discoverability.

server/chat-plugins/teams.ts
```diff
@@ -0,0 +1,691 @@
...
+const MAX_SEARCH = 3000;
+
+export interface StoredTeam {
+	teamid: string;
+	team: string;
+	ownerid: ID;
+	format: ID;
+	title: string | null;
+	date: Date;
+	private: boolean;
```
AnnikaCodes, line 22: Does the database interface handle conversion from integers to booleans? I'm pretty sure SQLite's `BOOLEAN` type is just an int.

mia-pi-git: We're not using SQLite for this - we're using Postgres.

AnnikaCodes:
This seems fine at first glance; has it been tested?

mia-pi-git:
> This seems fine at first glance; has it been tested?

Yes.

mia-pi-git:
Thanks all!

PR #9523, opened 2023-04-18
Update modchat help

singiamtel:
Two different small bugs, unrelated

### DIFF
--- a/server/chat-commands/room-settings.ts
+++ b/server/chat-commands/room-settings.ts
@@ -155,7 +155,7 @@ export const commands: Chat.ChatCommands = {
 		room.saveSettings();
 	},
 	modchathelp: [
-		`/modchat [off/autoconfirmed/trusted/+/%/@/*/player/#/&] - Set the level of moderated chat. Requires: % \u2606 for off/autoconfirmed/+ options, * @ # & for all the options`,
+		`/modchat [off/autoconfirmed/trusted/+/%/@/*/player/#/&] - Set the level of moderated chat. Requires: % \u2606 for off/autoconfirmed/+/player options, * @ # & for all the options`,
 	],
 
 	automodchat(target, room, user) {


server/chat-commands/room-settings.ts
```diff
@@ -155,7 +155,7 @@ export const commands: Chat.ChatCommands = {
 		room.saveSettings();
 	},
 	modchathelp: [
-		`/modchat [off/autoconfirmed/trusted/+/%/@/*/player/#/&] - Set the level of moderated chat. Requires: % \u2606 for off/autoconfirmed/+ options, * @ # & for all the options`,
+		`/modchat [off/autoconfirmed/trusted/+/%/@/*/player/#/&] - Set the level of moderated chat. Requires: % \u2606 for off/autoconfirmed/+/player options, * @ # & for all the options`,
```
singiamtel, line 158: Tested it and both players and drivers are able to ``/modchat player``, so I'm updating it here

server/chat-plugins/chatlog.ts
```diff
@@ -600,7 +600,7 @@ export abstract class Searcher {
 	) {
 		let buf = Utils.html`<div class="pad"><h2>Linecounts on `;
 		buf += `${roomid}${user ? ` for the user ${user}` : ` (top ${MAX_TOPUSERS})`}</h2>`;
-		buf += `<strong>Total lines: {total}</strong><br />`;
+		if (!user) buf += `<strong>Total lines: {total}</strong><br />`;
```
singiamtel, line 603: ``/linecount id=`` seems to be broken, shows {total} rather than the number. They are displayed below so my suggestion would be to hide this one when id= is provided, and leave it as is when there's no id

Current:
image

Change:
image

mia-pi-git:
In the future: Please make separate PRs for separate fixes. 
Can you remove the chatlog change from this PR? It's not the correct fix (total lines should be showing if a user is specified), so I'll merge this one once it's out.

singiamtel:
> In the future: Please make separate PRs for separate fixes. Can you remove the chatlog change from this PR? It's not the correct fix (total lines should be showing if a user is specified), so I'll merge this one once it's out. Thanks for the /help fix!

Fair enough, change reverted. Should I open a new PR for the ``/linecount``, to discuss what the fix should be?

AnnikaCodes:
Thanks!

PR #9528, opened 2023-04-19
Fix Mycelium Might and Quick Claw interaction

singiamtel:
Quick claw can no longer boost status moves that had lower priority because of mycelium might.

https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-108#post-9509334

Added some tests too with a fixed seed

### DIFF
--- a/data/items.ts
+++ b/data/items.ts
@@ -4518,7 +4518,8 @@ export const Items: {[itemid: string]: ItemData} = {
 	},
 	quickclaw: {
 		onFractionalPriorityPriority: -2,
-		onFractionalPriority(priority, pokemon) {
+		onFractionalPriority(priority, pokemon, target, move) {
+			if (move.status && pokemon.hasAbility("myceliummight")) return;
 			if (priority <= 0 && this.randomChance(1, 5)) {
 				this.add('-activate', pokemon, 'item: Quick Claw');
 				return 0.1;
--- a/data/items.ts
+++ b/data/items.ts
@@ -4519,7 +4519,7 @@ export const Items: {[itemid: string]: ItemData} = {
 	quickclaw: {
 		onFractionalPriorityPriority: -2,
 		onFractionalPriority(priority, pokemon, target, move) {
-			if (move.status && pokemon.hasAbility("myceliummight")) return;
+			if (move.category === "Status" && pokemon.hasAbility("myceliummight")) return;
 			if (priority <= 0 && this.randomChance(1, 5)) {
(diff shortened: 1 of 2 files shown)


data/items.ts
```diff
@@ -4518,8 +4518,9 @@ export const Items: {[itemid: string]: ItemData} = {
 	},
 	quickclaw: {
 		onFractionalPriorityPriority: -2,
-		onFractionalPriority(priority, pokemon) {
+		onFractionalPriority(priority, pokemon, target, move) {
 			if (priority <= 0 && this.randomChance(1, 5)) {
+				if (move.status && pokemon.hasAbility("myceliummight")) return;
```
Karthik99999, line 4523:
```suggestion
			if (move.status && pokemon.hasAbility('Mycelium Might')) return;
			if (priority <= 0 && this.randomChance(1, 5)) {
```
just so that `randomChance` doesn't run which affects rng

singiamtel: Implemented. 
Unrelated question, do you guys prefer if I rebase changes to keep one commit, or do you do that on merge?

Karthik99999: PRs are always squashed, no need force push everything into one commit.

data/items.ts
```diff
@@ -4518,7 +4518,8 @@ export const Items: {[itemid: string]: ItemData} = {
 	},
 	quickclaw: {
 		onFractionalPriorityPriority: -2,
-		onFractionalPriority(priority, pokemon) {
+		onFractionalPriority(priority, pokemon, target, move) {
+			if (move.status && pokemon.hasAbility("myceliummight")) return;
```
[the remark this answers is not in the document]
singiamtel: I guess I was testing it with spore... Thanks for feedback, implemented

test/sim/abilities/myceliummight.js
```diff
@@ -19,4 +19,22 @@ describe("Mycelium Might", function () {
 		battle.makeChoices();
 		assert.false.fullHP(battle.p2.active[0]);
 	});
+	it("should never trigger your own quick claw if using a status move", function () {
+		battle = common.createBattle({seed: [3, 1, 1, 1]}, [[
```
DaWoblefet, line 23: I think you can use {forceRandomChance: true} here to avoid a hardcoded seed, same for the second test.

singiamtel: Didn't know that was a thing, cool. Implemented

DaWoblefet:
Thanks singiamtel!

Issue #9531, opened 2023-04-20
Gen 1: Ditto base attack stat changes from 194 to 166


andrebastosdias:
While simulating Gen 1 Random Battles, I've noticed a recurring issue with one particular behavior that keeps appearing in my assertions. Although I am unable to reproduce it as it seems to happen randomly, there are instances where, upon switching out a Ditto, its base attack stat on the request object becomes 166 instead of the expected 194.

I don't know if this is an error or a mechanic.

Here's an example (search for "details":"Ditto"):

```
[...]

PR #9532, opened 2023-04-20
Gens 1-4: fix bug where Transform users' stats get permanently altered

livid-washed:
This fixes a bug in gens 1-4. If Pokemon 1 Transformed into another Pokemon with different IVs, switches out, and switches back in, it would incorrectly calculate its stats based on the second Pokemon's IVs, rather than its own IVs.

The intention of copying IVs is for Hidden Power typing and power, but those are copied separately already.

This fixes the issue raised here: https://github.com/smogon/pokemon-showdown/issues/9531


andrebastosdias:
I'm not sure if this is the same issue that I posted in [#9531](https://github.com/smogon/pokemon-showdown/issues/9531). From what I checked (by running "infinite" random battles), the issue occured only with the Attack stat and it always changed from 194 to 166

PR #9534, opened 2023-04-21
Chat: Implement /requestcode

singiamtel:
Approved here: https://www.smogon.com/forums/threads/add-requestcode.3709337/

I mostly tried to copy how /requestshow is implemented. 
It's my first time dealing with chat commands, would appreciate any feedback/suggestion.

### DIFF
--- a/server/chat-commands/core.ts
+++ b/server/chat-commands/core.ts
@@ -1638,7 +1638,7 @@ export const commands: Chat.ChatCommands = {
 process.nextTick(() => {
 	// We might want to migrate most of this to a JSON schema of command attributes.
 	Chat.multiLinePattern.register(
-		'>>>? ', '/(?:room|staff)intro ', '/(?:staff)?topic ', '/(?:add|widen)datacenters ', '/bash ', '!code ', '/code ', '/modnote ', '/mn ',
+		'>>>? ', '/(?:room|staff)intro ', '/(?:staff)?topic ', '/(?:add|widen)datacenters ', '/bash ', '!code ', '/code ', '/requestcode', '/modnote ', '/mn ',
 		'/eval', '!eval', '/evalbattle', '/evalsql', '>>sql',
 		'/importinputlog '
 	);
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2619,6 +2619,102 @@ export const commands: Chat.ChatCommands = {
 	},
 	denyshowhelp: [`/denyshow [user] - Denies the media display request of [user]. Requires: % @ # &`],
 
+	requestcode(target, room, user) {
+		room = this.requireRoom();
+		this.checkChat();
+		target = this.message.substr(this.cmdToken.length + this.cmd.length + +this.message.includes(' ')).trimRight();
+		if (!room.settings.requestCodeEnabled) {
+			return this.errorReply(`Code approvals are disabled in this room.`);
+		}
+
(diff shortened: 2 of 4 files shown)


mia-pi-git:
Personally, I still fail to see why this is necessary when someone can just post a link to a code-sharing website (pastebin, pastie, etc). It doesn’t seem like anyone actually addressed that in the thread, and this is a niche enough request it feels important to consider.

singiamtel:
The same could be said about /requestshow and Imgur. It's more convenient to be able to do it directly from the room, and doesn't require staff to click random links to different pastebin clones.

PartMan7:
Agreed with the cute weeb here, linking to a paste is far less likely to be read than a !code, and it's significantly more effort than a requestcode.

mia-pi-git:
> The same could be said about /requestshow and Imgur

Those are commonly used, though. This seems to only be desired by a single room.

HuntressRubyRose:
Tier rooms voices and above frequently use code to share teams without leaving the site.

singiamtel:
How should we proceed here? I thought this was approved already, but if you're not convinced that it's needed, should we go back to discussing it in the Smogon thread?

AnnikaCodes:
Like I said on Smogon, I'm not sure why showing code needs to be blocked from regular users in the first place (is the spam really so much worse than just normal chat? it's not like images where you can post shock content). If some rooms want to keep it blocked, would it make sense to make it a roomsetting that Tech & Code can toggle to let regular users post code blocks?

mia-pi-git:
> Like I said on Smogon, I'm not sure why showing code needs to be blocked from regular users in the first place (is the spam really so much worse than just normal chat? it's not like images where you can post shock content). If some rooms want to keep it blocked, would it make sense to make it a roomsetting that Tech & Code can toggle to let regular users post code blocks?

Said as much in #staff but might as well repost: against this for feature bloat reasons, which a roomsetting would only aggravate. Plus, /code is restricted for a reason. If people want to post long blocks of text, making them go through an external host seems more prudent (easier to remove from the screen that way).

AnnikaCodes:
Upper Staff have discussed this and the general consensus is that a roomsettings toggle to let anyone use `!code` that defaults to not letting them use it is fine.

mia-pi-git:
> Upper Staff have discussed this and the general consensus is that a roomsettings toggle to let anyone use `!code` that defaults to not letting them use it is fine.

It may be prudent to, instead, have /permissions support allowing a given rank to broadcast specific commands. This is something people have asked about before, and it would be more general-use than a toggle for one command.

AnnikaCodes:
> > Upper Staff have discussed this and the general consensus is that a roomsettings toggle to let anyone use `!code` that defaults to not letting them use it is fine.
> 
> It may be prudent to, instead, have /permissions support allowing a given rank to broadcast specific commands. This is something people have asked about before, and it would be more general-use than a toggle for one command.

Currently, we check if the user of a command has 'jurisdiction' over unranked users when testing permissions to broadcast (or use other commands without a target). Since you don't generally have jurisdiction over people of the same rank, I propose that we add an exception so that if `/permissions` has allowed a command to be used by unranked users, this jurisdiction check is skipped. Would this cause any problems I'm not seeing?

AnnikaCodes:
Sorry @singiamtel; unfortunately US wanted a different solution to this.

singiamtel:
Understood, thanks for the follow-up

PR #9539, opened 2023-04-25
Tier Shift: New values

Seerd:


DaWoblefet:
Seems to be accurate as of https://www.smogon.com/forums/threads/tier-shift.3720396/. Thanks!

PR #9541, opened 2023-04-27
`/requestshow`: Fix comments

Dirain1700:
### Fix permissions

At first, `user.can('showmedia', null, room, '/show')` return the user has a rank which voice or higher. if `!show` disallowed for some rank and the rank was voice and higher, they could not use  both of `!show` and `/requestshow`.
By using `Chat.checkCan('show', null, room)`, we could solve this problem. if the user could not use `!show`, they could use `/requestshow`.

### Fix comments
If the comment starts/ends with spaces, the comment was caught by `Chat.checkChat()`. Solved this by triming spaces.

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2517,7 +2517,7 @@ export const commands: Chat.ChatCommands = {
 		if (!room.settings.requestShowEnabled) {
 			return this.errorReply(`Media approvals are disabled in this room.`);
 		}
-		if (this.checkCan('show', null, room)) return this.errorReply(`Use !show instead.`);
+		if (user.can('showmedia', null, room, '/show')) return this.errorReply(`Use !show instead.`);
 		if (room.pendingApprovals?.has(user.id)) return this.errorReply('You have a request pending already.');
 		if (!toID(target)) return this.parse(`/help requestshow`);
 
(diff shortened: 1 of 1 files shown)


mia-pi-git:
From what I understand of what you're describing, this is intentional behavior - people who can use /show cannot use /requestshow.

Dirain1700:
In room settings, Room Owners can restrict permissions to disallow to use !show. disallowed rank users can't use /requestshow

For example, In Anime and Manga room, voices cannot use !show, and /requestshow, too.
It should use same permission checker for !show and /requestshow

mia-pi-git:
It should not be. Room owners may want to disable both.

mia-pi-git:
What you've described so far is intentional behavior, not a bug. If you remove the permissions change and just leave in the comment fix, I'll merge this.

Dirain1700:
Well, these are matters I want to tell:

- ROs can disable requestshow in another settings
- In some rooms voices cannot use !show
- In the room (!show disallowed for voices), regular user can use /requestshow, but voices can't

server/chat-commands/info.ts
```diff
@@ -2517,7 +2517,7 @@ export const commands: Chat.ChatCommands = {
 		if (!room.settings.requestShowEnabled) {
 			return this.errorReply(`Media approvals are disabled in this room.`);
 		}
-		if (user.can('showmedia', null, room, '/show')) return this.errorReply(`Use !show instead.`);
+		if (this.checkCan('show', null, room)) return this.errorReply(`Use !show instead.`);
```
mia-pi-git, line 2520:
```suggestion
		if (user.can('showmedia', null, room, '/show')) return this.errorReply(`Use !show instead.`);
```

mia-pi-git:
Thanks!

PR #9542, opened 2023-04-28
/requestshow: Allow using comma in comments

Dirain1700:
`[link, comment]` ignores latter comments splited by comma if the comment includes. Use `[link, ...comments]` to allow it.
Also fix #9541, comment should be trimed and stored to variable, not just checking

### DIFF
--- a/server/chat-commands/info.ts
+++ b/server/chat-commands/info.ts
@@ -2521,7 +2521,8 @@ export const commands: Chat.ChatCommands = {
 		if (room.pendingApprovals?.has(user.id)) return this.errorReply('You have a request pending already.');
 		if (!toID(target)) return this.parse(`/help requestshow`);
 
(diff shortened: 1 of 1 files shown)


server/chat-commands/info.ts
```diff
@@ -2521,7 +2521,8 @@ export const commands: Chat.ChatCommands = {
 		if (room.pendingApprovals?.has(user.id)) return this.errorReply('You have a request pending already.');
 		if (!toID(target)) return this.parse(`/help requestshow`);
 
-		let [link, comment] = target.split(',');
+		let [link, ...comments] = target.split(',');
+		const comment = comments.join(',').trim();
```
mia-pi-git, line 2525:
```suggestion
		const [link, comment] = this.splitOne(target);
```

mia-pi-git:
Thanks!

PR #9544, opened 2023-04-30
Random Battles: 2023-04 End-of-month changes

adrivrie:
April monthly randbats balance. Ideally is merged as soon as possible so the May stats are based on the new levels.

### DIFF
--- a/data/mods/gen3/random-teams.ts
+++ b/data/mods/gen3/random-teams.ts
@@ -246,8 +246,7 @@ export class RandomGen3Teams extends RandomGen4Teams {
 		// First, the high-priority items
 		if (species.name === 'Ditto') return this.sample(['Metal Powder', 'Quick Claw']);
 		if (species.name === 'Farfetch\u2019d') return 'Stick';
-		if (species.name === 'Latias') return 'Soul Dew';
-		if (species.name === 'Latios') return 'Soul Dew';
+		if (species.name === 'Latias' || species.name === 'Latios') return 'Soul Dew';
 		if (species.name === 'Marowak') return 'Thick Club';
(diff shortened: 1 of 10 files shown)


livid-washed (approved):

data/mods/gen3/random-teams.ts
```diff
@@ -246,6 +246,8 @@ export class RandomGen3Teams extends RandomGen4Teams {
 		// First, the high-priority items
 		if (species.name === 'Ditto') return this.sample(['Metal Powder', 'Quick Claw']);
 		if (species.name === 'Farfetch\u2019d') return 'Stick';
+		if (species.name === 'Latias') return 'Soul Dew';
+		if (species.name === 'Latios') return 'Soul Dew';
```
ACakeWearingAHat, line 250: these can be combined with a ||

ACakeWearingAHat: same for every gen

ACakeWearingAHat (changes requested):

adrivrie:
This PR now also includes a more general Random Battles patch: adding soul dew to (non-mega) Lati@s and revamping their sets in gens 3-6, and removing Toxtricity Low-Key from gen9 randbats.

This was combined into the same PR to save some merging effort as these changes would likely have caused a merge conflict otherwise. If this is inconvenient or (too) bad practice, please let us know the preferred course of action

data/mods/gen3/random-teams.ts
adrivrie: Eh, sure

ACakeWearingAHat (approved):

PR #9547, opened 2023-05-01
WIP: Tournaments: Add support for swiss-style tournaments

CheeseMuffin:
Add swiss-style tournaments as an option (instead of round robin or elimination). This was asked for a long time ago (https://github.com/smogon/pokemon-showdown/issues/2205) and is approved in the forum suggestions (https://www.smogon.com/forums/threads/add-swiss-as-a-valid-tour-format.3658511/#post-8328188). Am also planning on adding support for bo3 for all types of tours but that will happen after this. It's still a bit in-progress, but wanted to make sure the general idea was good before I do the more complicated stuff.

Algorithm: 
Does roughly what https://www.npmjs.com/package/swiss-pairing does (Edmonds Blossom algorithm). In practice, with 1000 entrants, takes a few seconds to pair (image). Assuming this algorithm is okay, I plan on reimplementing Edmonds Blossom on our side (so the dependency this PR currently adds will not be there in the final version), but wanted to ensure it was okay first as it is a bit complicated. This should functionally do the same thing as normal Swiss pairings (pair within the same score group and avoid rematches), though the weighting procedure might need some adjustments.

Screenshots (all of these lists are scrollable):
Signups:
image
In-Progress:
image
Finished:
image

Very minimal client-side changes at the moment since I'm not great at those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "0.11.9",
   "main": "dist/sim/index.js",
   "dependencies": {
+    "edmonds-blossom": "^1.0.0",
     "esbuild": "^0.16.10",
     "preact": "^10.5.15",
     "preact-render-to-string": "^5.1.19",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
+        "edmonds-blossom": "^1.0.0",
         "esbuild": "^0.16.10",
         "preact": "^10.5.15",
         "preact-render-to-string": "^5.1.19",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "version": "0.11.9",
   "main": "dist/sim/index.js",
   "dependencies": {
+    "edmonds-blossom": "^1.0.0",
```
AnnikaCodes, line 7: I assume @Zarel will prefer to wait to merge this until you reimplement this dependency in-house.

CheeseMuffin: In the chat-dev channel on discord, Zarel said the dependency was fine.

CheeseMuffin:
I'll get around to fixing all the TODOs I left in and then i'll request a proper review, though idk who the right person for that is since I don't see anybody in the codeowners file

PR #9549, opened 2023-05-01
Support hotpatching formats without reloading processes

mia-pi-git:
Should make it a lot easier to hotpatch formats consecutively, without worrying about memory from extra processes lying around. 

This will require a restart.

### DIFF
--- a/lib/process-manager.ts
+++ b/lib/process-manager.ts
@@ -605,16 +605,19 @@ export class StreamProcessManager extends ProcessManager<StreamProcessWrapper> {
 	activeStreams: Map<string, Streams.ObjectReadWriteStream<string>>;
 	_createStream: () => Streams.ObjectReadWriteStream<string>;
 	messageCallback?: (message: string) => any;
+	childMessageCallback?: (message: string) => any;
 
 	constructor(
 		module: NodeJS.Module,
 		createStream: () => Streams.ObjectReadWriteStream<string>,
-		messageCallback?: (message: string) => any
+		messageCallback?: (message: string) => any,
+		childMessageCallback?: (message: string) => any,
 	) {
 		super(module);
 		this.activeStreams = new Map();
 		this._createStream = createStream;
 		this.messageCallback = messageCallback;
+		this.childMessageCallback = childMessageCallback;
 
 		processManagers.push(this);
 	}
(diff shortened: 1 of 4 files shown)


Slayer95:
This probably helps with memory spikes in such scenarios. However, hotpatching doesn't internally remove the old code from memory (``require.cache = {}``) doesn't help. So that would mean going back to the time when PS needed to be restarted every week / few days.

Hence, If this goes through, there would still need to be another way to respawn the simulator processes.

Slayer95:
See also https://github.com/nodejs/node/issues/8443

mia-pi-git:
> This probably helps with memory spikes in such scenarios. However, hotpatching doesn't internally remove the old code from memory (`require.cache = {}` doesn't help.) So that would mean going back to the time when PS needed to be restarted every week / few days.
> 
> Hence, If this goes through, there would still need to be another way to respawn the simulator processes.

Does deleting the cache remove the reference? It seems to work fine in chat hotpatches?

Slayer95:
Unfortunately, it's unclear how thorough is enough. There is also the V8 code cache, that won't get cleared even if the Node.js modules are.

Slayer95:
What I'd suggest is tracking the active battles when ``/hotpatch`` is ran, and schedule a non-stackable task to run ``Rooms.PM.respawn()`` when they are all over. If battles are hotpatched again, then the task for respawn is overriden.

AnnikaCodes:
Do old battle simulation processes still cause memory problems? Our server consistently has 20+ GB free and I don't think child processes' memory usage counts against the main process's Node memory cap.

As a test, I just hotpatched formats on Main. Before the hotpatch:
> [Main process] RSS: 3.03 GiB, Heap: 1.61 GiB / 2.57 GiB
(and ~3.5 GB used by all battle simulator processes with 36G available)

After: 
> [Main process] RSS: 3.07 GiB, Heap: 1.71 GiB / 2.60 GiB
(and ~3.8 GB used by all battle simulator processes with 36G available, which later went up to 4.1G with 35G available)

So I'm not sure how important saving memory here is. If we can figure out how to flush the cache properly and reliably it would be an improvement, but I don't think having a larger permanent memory usage increase after hotpatching (necessitating more restarts) is a good tradeoff.

mia-pi-git:
Agreed, probably not worth it.

PR #9551, opened 2023-05-01
Make Natdex Tier Shift work

placuszek15:
It works on some example challenges on a local repository, i got spooked once by node telling me its a slow battle so .has _maybe_ could be made faster

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -2069,8 +2069,8 @@ export const Rulesets: {[k: string]: FormatData} = {
 				nfe: 30,
 				lc: 30,
 			};
-			let isNatdex: boolean = this.ruleTable.has("standardnatdex");
-			let tier: string = isNatdex ? this.toID(species.natDexTier) : this.toID(species.tier);
+			let isNatDex: boolean = this.ruleTable.has("standardnatdex");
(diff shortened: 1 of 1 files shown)


placuszek15:
Second commit hopefully makes the non-pokemon bans work, previous code was using === instead of toID which meant it didn't work

data/rulesets.ts
```diff
@@ -2069,12 +2069,12 @@ export const Rulesets: {[k: string]: FormatData} = {
 				nfe: 30,
 				lc: 30,
 			};
-			let tier: string = this.toID(species.tier);
+			let isNatdex: boolean = this.ruleTable.has("standardnatdex");
```
KrisXV, line 2072:
```suggestion
			let isNatDex: boolean = this.ruleTable.has("standardnatdex");
```

data/rulesets.ts
```diff
@@ -2069,12 +2069,12 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+			let tier: string = isNatdex ? this.toID(species.natDexTier) : this.toID(species.tier);
```
KrisXV, line 2073:
```suggestion
			let tier: string = this.toID(isNatDex ? species.natDexTier : species.tier);
```

PR #9556, opened 2023-05-09
Fix Gen3 Rest Sleep Talk interaction

MYNAMEISRADU:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9106563

Added fail message to the OnTry function of the move Rest if user is currently asleep. This is probably an issue due to the tryMoveHit function being modified in the gen3 scripts file

### DIFF
--- a/data/mods/gen3/moves.ts
+++ b/data/mods/gen3/moves.ts
@@ -527,6 +527,25 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		pp: 20,
 	},
+	rest: {
+		 inherit: true,
+		 OnTry (source) {
+			 if (source.status === 'slp') {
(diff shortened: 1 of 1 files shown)


DaWoblefet:
I apologize for the long delay on this review, but I don't think it's the correct fix. Currently _all_ effects that return false in onTry() events in Gen 3 won't play the appropriate failure message. So for example, Sleep Talk while the user is awake, or Fake Out if the user's already been on the field for a turn, or Follow Me in singles should all play a failure message but don't. So the fix would be looking into why "But it failed!" generic messages aren't being handled correctly.

_Technically_ we would need to modify Rest in one of the generations to account for the different string used. However, in general, we support very, very few oldgens move strings. I'd rather not override the function just to play the string, especially since it's not an information leak or anything.

PR #9557, opened 2023-05-09
Fix Volbeat Dizzy Punch Lunge legality issue

MYNAMEISRADU:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/page-68#post-9063784

added two different sets of level 1 movesets with each excluding one of the illegal moves (I presume this is how it's meant to be done)

### DIFF
--- a/data/learnsets.ts
+++ b/data/learnsets.ts
@@ -37666,6 +37666,10 @@ export const Learnsets: {[speciesid: string]: LearnsetData} = {
 			waterpulse: ["7T", "6T", "4M", "3M"],
 			zenheadbutt: ["7T", "7L33", "6T", "6L33", "5T", "5L37", "4T", "4L37"],
 		},
+		eventData: [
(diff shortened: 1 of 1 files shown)


DaWoblefet:
There is a skipped test for this in test/sim/team-validator/breeding.js, which I would recommend unskipping to test your work. In this case, adding additional event data will not fix the issue, because the validator already sees a Volbeat with Lunge and Dizzy Punch as legal. So adding additional event data would never make it illegal.

I think hardcoding this edge case is OK since it only affects the Volbeat line. It would need to be done somewhere in sim/team-validator.ts, likely in the findEggMoveFathers() function, where similar edge cases like Nidoran or Shedinja are handled.

MYNAMEISRADU:
I tried working with the findEggMoveFathers() function by bypassing the gen6 check for only volbeat specifically in gen7 but that still didn't fix the legality issue, and since I'm very new to the code I couldn't really figure out how to progress further.

PR #9564, opened 2023-05-17
Formats: Fix various style inconsistencies

adrivrie:
Happened to find some inconsistent use of `'` versus `"`, so I went and tried to find more style inconsistensies.


config/formats.ts
```diff
@@ -1887,118 +1887,118 @@ export const Formats: FormatList = [
...
-			"Deoxys-Attack", "Deoxys-Defense", "Deoxys-Speed", "Desolate Land", "Dialga", "Dialga-Origin", "Diamond Storm",
-			"Diancie", "Diancie-Mega", "Dire Claw", "Disable", "Discharge", "Dondozo", "Doom Desire", "Double Iron Bash",
-			"Download", "Draco Meteor", "Draco Plate", "Dragapult", "Dragon Ascent", "Dragon Dance", "Dragon Darts",
-			"Dragon Energy", "Dragon Hammer", "Dragon Pulse", "Dragon Tail", "Dragonite", "Dragonium Z", "Drain Punch",
-			"Dread Plate", "Drill Peck", "Drizzle", "Drought", "Drum Beating", "Dry Skin", "Duraludon", "Dusknoir",
-			"Dynamax Cannon", "Earth Eater", "Earth Plate", "Earth Power", "Earthquake", "Eerie Spell", "Effect Spore",
-			"Eject Pack", "Electivire", "Electric Surge", "Electrium Z", "Electro Drift", "Emboar", "Empoleon", "Enamorus",
-			"Enamorus-Therian", "Encore", "Energy Ball", "Entei", "Eruption", "Espeon", "Esper Wing", "Eternatus",
-			"Eternatus-Eternamax", "Exeggutor", "Exeggutor-Alola", "Expanding Force", "Expert Belt", "Explosion",
-			"Extreme Evoboost", "Extreme Speed", "Fairium Z", "Fake Out", "Feraligatr", "Fiery Wrath", "Fightinium Z",
```
AnnikaCodes, line 1925: Aren't banlists displayed directly to the user by the `/tier` command?

adrivrie: They are indeed, but all other (un)banlists and rulesets all use single quotes. Perhaps the idea is that the "main use" of rulesets is in-code, and it being shown to users is just a side effect of the implementation of `/tier`?

If you prefer I can revert this part. Getting everything into double quotes is not something I feel like undertaking manually though, and I can't think of a good way to automate it.

AnnikaCodes (approved):
This seems fine to me; @KrisXV do you have an opinion? You work on formats a lot more than I do.

AnnikaCodes:
Thankns!

PR #9568, opened 2023-05-26
Fix local ladders inputing bad entries on load

Karthik99999:
Caused by #4516 where the behavior of skipping the first line of the tsv ladder file was accidentally removed.


AnnikaCodes (approved):
For the record, the data is saved with a header row here: https://github.com/smogon/pokemon-showdown/blob/d74822bbc74c095aa312053f7034baee3068d99e/server/ladders-local.ts#L97-L98

AnnikaCodes:
Thanks!

PR #9570, opened 2023-05-28
Add missing general translations

DieterReinert:


AnnikaCodes (approved):
Looks good, thanks!

AnnikaCodes:
> theres a much easier way to do this that doesn't involve all of this support, using a just one library. is there a reason why it's set up this way

Do you have a particular library you recommend? Generally, Showdown tries to avoid unnecessary dependencies, but we can definitely consider managing translations another way.

mynameisthunder:
> > theres a much easier way to do this that doesn't involve all of this support, using a just one library. is there a reason why it's set up this way
> 
> Do you have a particular library you recommend? Generally, Showdown tries to avoid unnecessary dependencies, but we can definitely consider managing translations another way.

absolutely, at a few companies ive worked at we used 

[https://react.i18next.com/latest/usetranslation-hook](https://react.i18next.com/latest/usetranslation-hook)

PR #9580, opened 2023-06-01
BSS/VGC: Add Regulation D

Karthik99999:
Pending verification on if transfer mechanics are correct here

### DIFF
--- a/config/formats.ts
+++ b/config/formats.ts
@@ -230,18 +230,18 @@ export const Formats: FormatList = [
 		],
 	},
 	{
-		name: "[Gen 9] Battle Stadium Singles Series 2",
-
-		mod: 'gen9',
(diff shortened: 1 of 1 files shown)


DaWoblefet:
My initial worry was about whether or not Regulation D was going to enforce SV-origin Pokemon, which would exclude quite a lot of Home-exclusive stuff (e.g. Regieleki or Alolan Raichu) but still allow SV-origin Teddiursa that was evolved to Ursaluna in PLA and then traded back. Everything I'm reading implies that this rule will just change when Reg D rolls around. There's also a bunch of move legality stuff to investigate thanks to research from Anubis, but that applies for every SV format, not just our Nintendo formats.

Thanks Karthik!

PR #9581, opened 2023-06-01
Add /am edithistory to /help am

Lionyx:
I keep forgetting about this command's name and having it there is going to be very intuitive


AnnikaCodes (approved):

AnnikaCodes:
Thanks!

PR #9589, opened 2023-06-05
Revelationmons: Update bans

Seerd:


AnnikaCodes:
Is there a source for these bans?

PR #9590, opened 2023-06-05
MnM: Unrestrict Zapdos-Galar

TomOfTomKat:


AnnikaCodes:
Is there a Smogon post for this rule change?

PR #9594, opened 2023-06-09
Add National Dex Ubers

Amurakaede:


monsanto:
This also removes National Dex AG, is that intentional? This format exists in a previous generation.

PR #9601, opened 2023-06-13
Implement Gen 9 Random Doubles Battle

livid-washed:
Credits:
Coding: me and @ACakeWearingAHat
Sets: @ACakeWearingAHat, Arcticblast, Celever, Eeveon7, longhiep341, PlatinumCheesecake
/randdubs command (will be appended onto this PR): @MathyFurret

Gen 9 Random Doubles Battle follows the same team/set generation as singles, described here: https://github.com/smogon/pokemon-showdown/pull/9169. It has different role names, aside from Tera Blast user which is in common with singles.

This PR also adds Gen 9 Random Doubles, FFA, and multi to the formats and tracks winrates for randdubs which will be used for level balancing.


adrivrie (approved):
Looks good to me! There's a couple non-dubs changes that I would have personally preferred to be in a separate commit, but I don't think that's enforced in this repo.

MathyFurret (approved):

DaWoblefet:
Good work, randbats team!

PR #9604, opened 2023-06-15
Add Freedom Cup

DaWoblefet:
To go live June 15th 11:59:59 pm. See https://www.smogon.com/forums/threads/1000-ladder-tour-freedom-cup-playing-competitive-pokemon-how-the-pokemon-company-intended.3722982/

The only thing left for the format is debating where we want it to live. Currently, it's in Pet Mods, but doesn't mod any actual game mechanics. SV Singles or Other Metagames would be the other candidates.


KrisXV:
After thinking through more, I'd move this for SV Singles since it has more viewability there

PR #9605, opened 2023-06-16
Send revealed set information in 'poke' messages

qwertystrargs:
Related to smogon/pokemon-showdown-client#2127

Sends information for open team sheets via a new method of `Battle`, 
including data for the client's Pokemon properties as kwArgs in `|poke|` messages.

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -2,7 +2,6 @@
 
 import {Utils} from "../lib";
 import {Pokemon} from "../sim/pokemon";
-import {Teams} from "../sim/teams";
 
 // The list of formats is stored in config/formats.js
 export const Rulesets: {[k: string]: FormatData} = {
@@ -535,12 +534,18 @@ export const Rulesets: {[k: string]: FormatData} = {
 			}
 		},
 		onTeamPreview() {
-			this.add('clearpoke');
-			for (const pokemon of this.getAllPokemon()) {
-				const details = pokemon.details.replace(', shiny', '')
-					.replace(/(Greninja|Gourgeist|Pumpkaboo|Xerneas|Silvally|Urshifu|Dudunsparce)(-[a-zA-Z?-]+)?/g, '$1-*')
-					.replace(/(Zacian|Zamazenta)(?!-Crowned)/g, '$1-*'); // Hacked-in Crowned formes will be revealed
-				this.add('poke', pokemon.side.id, details, '');
+			if (this.ruleTable.has(`forceopenteamsheets`)) {
+				this.showOpenTeamSheets(true);
+			} else {
+				this.add('clearpoke');
+				for (const pokemon of this.getAllPokemon()) {
+					const teraTypeNote = this.ruleTable.has(`teratypepreview`) && pokemon.canTerastallize ?
+						`[teraType] ${pokemon.teraType}` : ``;
+					const details = pokemon.details.replace(', shiny', '')
+						.replace(/(Greninja|Gourgeist|Pumpkaboo|Xerneas|Silvally|Urshifu|Dudunsparce)(-[a-zA-Z?-]+)?/g, '$1-*')
+						.replace(/(Zacian|Zamazenta)(?!-Crowned)/g, '$1-*'); // Hacked-in Crowned formes will be revealed
+					this.add('poke', pokemon.side.id, details, teraTypeNote);
+				}
 			}
 			this.makeRequest('teampreview');
 			if (this.ruleTable.has(`teratypepreview`)) {
@@ -1660,7 +1665,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 	openteamsheets: {
 		effectType: 'Rule',
 		name: 'Open Team Sheets',
-		desc: "Allows each player to see the Pok&eacute;mon and all non-stat information about them, before they choose their lead Pok&eacute;mon",
+		desc: "If all players agree during Team Preview, all non-stat Pok&eacute;mon set information is revealed to them",
 		onTeamPreview() {
 			const msg = 'uhtml|otsrequest|<button name="send" value="/acceptopenteamsheets" class="button" style="margin-right: 10px;"><strong>Accept Open Team Sheets</strong></button><button name="send" value="/rejectopenteamsheets" class="button" style="margin-top: 10px"><strong>Deny Open Team Sheets</strong></button>';
 			for (const side of this.sides) {
@@ -1676,14 +1681,8 @@ export const Rulesets: {[k: string]: FormatData} = {
 	forceopenteamsheets: {
 		effectType: 'Rule',
 		name: 'Force Open Team Sheets',
-		desc: "Allows each player to see the Pok&eacute;mon and all non-stat information about them, before they choose their lead Pok&eacute;mon",
-		onTeamPreview() {
-			let buf = 'raw|';
-			for (const side of this.sides) {
-				buf += Utils.html`<div class="infobox" style="margin-top:5px"><details><summary>Open Team Sheet for ${side.name}</summary>${Teams.export(side.team, {hideStats: true})}</details></div>`;
-			}
-			this.add(buf);
-		},
+		desc: "Allows players to see all non-stat Pok&eacute;mon set information during Team Preview",
+		// hardcoded in the teampreview rule
 	},
 	aaarestrictedabilities: {
 		effectType: 'ValidatorRule',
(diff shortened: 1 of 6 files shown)


server/chat-commands/core.ts
```diff
@@ -804,15 +804,7 @@ export const commands: Chat.ChatCommands = {
...
-			let buf = '|uhtml|ots|';
-			for (const curPlayer of battle.players) {
-				const team = await battle.getTeam(curPlayer.id);
-				if (!team) continue;
-				buf += Utils.html`<div class="infobox" style="margin-top:5px"><details><summary>Open Team Sheet for ${curPlayer.name}</summary>${Teams.export(team, {hideStats: true})}</details></div>`;
-			}
-			for (const curPlayer of battle.players) {
-				curPlayer.sendRoom(buf);
-			}
+			battle.forceOTS();
```
Slayer95, line 807: This, however, does change the implementation from "players only" to "everyone".

data/rulesets.ts
```diff
@@ -1676,14 +1695,12 @@ export const Rulesets: {[k: string]: FormatData} = {
...
 		effectType: 'Rule',
 		name: 'Force Open Team Sheets',
-		desc: "Allows each player to see the Pok&eacute;mon and all non-stat information about them, before they choose their lead Pok&eacute;mon",
-		onTeamPreview() {
-			let buf = 'raw|';
-			for (const side of this.sides) {
-				buf += Utils.html`<div class="infobox" style="margin-top:5px"><details><summary>Open Team Sheet for ${side.name}</summary>${Teams.export(side.team, {hideStats: true})}</details></div>`;
-			}
-			this.add(buf);
+		desc: "Allows each player to see every Pokémon and all non-stat information about them, before they choose their lead Pokémon",
```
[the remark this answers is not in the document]
Slayer95: I think some tournaments might want the "everyone" variant, while casual/ladder play would prefer the "each player" variant.

data/rulesets.ts
```diff
@@ -530,58 +530,26 @@ export const Rulesets: {[k: string]: FormatData} = {
...
-					setNotes += `|[moveList] ${moveList.join(';')}`;
-					if (this.ruleTable.has('broadcastopenteamsheets')) {
-						this.add('poke', pokemon.side.id, details, setNotes);
-					} else {
-						const message = ['poke', pokemon.side.id, details, setNotes].join('|');
-						for (const side of this.sides) this.addSplit(side.id, [message]);
+					let buf = ``;
+					for (const pokemon of side.pokemon) {
+						buf += buf ? ` / ` : `raw|${side.name}'s Tera Types:<br />`;
+						buf += `<psicon pokemon="${pokemon.species.id}" /><psicon type="${pokemon.teraType}" />`;
```
Slayer95: This is nice!

data/rulesets.ts
```diff
@@ -530,58 +530,26 @@ export const Rulesets: {[k: string]: FormatData} = {
...
-					}
-					setNotes += `|[moveList] ${moveList.join(';')}`;
-					if (this.ruleTable.has('broadcastopenteamsheets')) {
-						this.add('poke', pokemon.side.id, details, setNotes);
-					} else {
-						const message = ['poke', pokemon.side.id, details, setNotes].join('|');
-						for (const side of this.sides) this.addSplit(side.id, [message]);
+					let buf = ``;
+					for (const pokemon of side.pokemon) {
+						buf += buf ? ` / ` : `raw|${side.name}'s Tera Types:<br />`;
```
Slayer95: And this needs a escape.

data/rulesets.ts
qwertystrargs: For the record, this is not from my code (it was introduced last January).

Slayer95: Oh, you should probably rebase in that case.

qwertystrargs: From what I can tell, based on the result ("Files Changed" tab), the subject `teraType` -gallery code is  *not* considered part of my changes.
Also, I was under the impression that modifying this PR (name/title is inaccurate) such as by rebasing, renaming, etc. would cause issues
(... and I have mishandled these PRs enough already.)

(Sorry if this is terribly erroneous, I am still not especially familiar with Git /GitHub.)

Slayer95: > Also, I was under the impression that modifying this PR (name/title is inaccurate) such as by rebasing, renaming, etc. would cause issues

You would normally be playing very safe but following such guidelines, but you have already messed with the PR by doing [Merge 'origin/master' into openTeamSheets](https://github.com/smogon/pokemon-showdown/pull/9605/commits/e58bec3f4f333cea18f094839605d5b04d12747c)

That's why you see below the error: <code>"This branch cannot be rebased due to conflicts"</code>. Since you are new with Git, resolving conflicts might be challenging, so you could also do a backup and work from a fresh PR from an up-to-date repository. You may also ask for help in #development and/or #techcode.

data/rulesets.ts
```diff
@@ -1697,7 +1697,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 		onTeamPreview() {
 			let buf = 'raw|';
 			for (const side of this.sides) {
-				buf += Utils.html`<div class="infobox" style="margin-top:5px"><details><summary>Open Team Sheet for ${side.name}</summary>${Teams.export(side.team, {hideStats: true})}</details></div>`;
+				buf += Utils.html`<div class="infobox" style="margin-top:5px"><details><summary>Open Team Sheet for ${Utils.escapeHTML(side.name)}</summary>${Teams.export(side.team, {hideStats: true})}</details></div>`;
```
Slayer95, line 1700: Utils.html already takes care of escaping.

data/rulesets.ts
```diff
@@ -2033,7 +2032,7 @@ export const Rulesets: {[k: string]: FormatData} = {
 				newPoke.position = newPos;
 				source.side.pokemon[newPos] = newPoke;
 				this.add('poke', source.side.pokemon[newPos].side.id, source.side.pokemon[newPos].details, '');
-				this.add('-message', `${target.name} was captured by ${newPoke.side.name}!`);
+				this.add('-message', `${target.name} was captured by ${Utils.escapeHTML(newPoke.side.name)}!`);
```
Slayer95, line 2035: Only raw messages need to be escaped by the server. This is not, so it's automatically escaped client-side.

Slayer95:
@qwertystrargs , it looks like this PR might be compatible with the current client.

So that we needn't wait for the client PR to go live before. That would be for the best, since client PRs are often daunting for reviewers. Can you confirm whether that's the case?

qwertystrargs:
I believe both of these PRs are compatible independent of one another.
Currently,
- With just the server changes, open team sheets should only (visibly) update pokemon details (shinyness, forme) during Team Preview and reset it once all players agree (which unfortunately(?) adds another team-species reminder (not changed for client)).
-  With just the client changes, the only effects should be the `terastallized` property being retained until revival and the addition of a `teraType` property (which would be limited to a "this was previously terastallized" reminder, for after revival).

qwertystrargs:
Before I forget, would be reasonable to increment a [Zoroark]'s `previouslySwitchedIn` when its illusion ends instead of when it switches in, as means of tracking what the client has seen (for formats without Team Preview)? Perhaps alternatively, would it be preferable to simply remove the changes related to `updatepoke` and reintroduce them at a later point, after (/if) the other changes are accepted?

Slayer95:
Not sure if I am understanding correctly the scenario, but to start with, I don't think OTS without Team Preview makes much sense. Opt-in OTS as implemented in PS inherently leaks information. (Player A sees a Pokémon that may run any of 5+ sets, * asks for team sheets*) The later the request to disclose team sheets, and the more asymmetrical the information known by the players, the worse the result. If there is a Zoroark or multiple Pokémon of the same species in-game, all the worst.

Therefore, trying to track what the client sees in order to extend support for non-fair opt-in team sheets looks like walking in the wrong direction. 

Nevertheless, the current extent of the proposed |updatepoke| changes seem minimal enough. However, no comment on its correctness.

qwertystrargs:
I will remove the changes related to `|updatepoke|` later, but if you noticed any other design failures I have overlooked, I should like to know (for when/if I revisit `|updatepoke|`).
Or perhaps you could explain what you meant by "**non-fair** opt-in team sheets"?

Related to my last comment, the condition for using `|poke|` over `|updatepoke|` should be changed to catch unrevealed pokemon, instead of having the client assume the pokemon exists after its search fails (plus the added condition fails when "extra" illusions exist).
Checking for a teampreview rule and `.previouslySwitchedIn === 0` seems to work except for unrevealed Zoroarks which have switched in as illusions before. (Although, adding a separate property which could also be set in teampreviews would probably be better.)

The only other _major_ problem I am aware of is the need to skip the kwArgs-sending loop when without Species Clause, which can be resolved with `onValidateRule()` (or by filtering out possible duplicates, like `getAllActive()`). That leaves some minor tooltip issues with Hidden Power and Illusion (worst case the wrong "Possible Illusion" gets the data?).

PR #9606, opened 2023-06-17
Refactor Ash-Greninja

KrisXV:
This refactor fixes an issue with Battle Bond Greninja being unusable in gen 9, as the current implementation is reliant on Ash-Greninja being marked as usable. This wasn't an issue for previous gens because both forms were either dexited or not; however, Ash-Greninja does not exist in Gen 9, so marking it as usable in order to allow this ability wouldn't work.


DaWoblefet (approved):
Build passes locally; we'll need to eventually figure out why it's not behaving on PRs that modify learnsets. I worked with Kris and all relevant mechanics and legality interactions seem good to go. Downstream sources will need to note that we are now using ``Greninja-Bond`` as the form that exclusively has Battle Bond.

Marty-D:
> Build passes locally

`npm run full-test` doesn't pass locally either; ExhaustiveRunner hangs forever on gen9 and gen7 custom games, which strongly implies a Greninja issue. Skipping the gen9 and gen7 runs allows the build to pass.

PR #9607, opened 2023-06-17
Implemented KunkFuFighting Convergence validation, needs to be tested…

FernySG:
… but should account for banned abilities (pokemon that naturally acquire them are allowed), Mega abilities/moves, and more edge cases.


FernySG:
Sorry! Didn't mean to send this.

PR #9612, opened 2023-06-20
Update /help searchlogs

singiamtel:
/searchlogs is now only accesible by admins, so I'm updating the docs.


DaWoblefet:
This is good since it doesn't appear searchlogs is high on the priority list to fix. Thanks!

PR #9613, opened 2023-06-22
Fix Greninja-Bond with exhaustive runner

DaWoblefet:
Currently, exhaustive runner looks at the species as defined by details information provided in pokemon.ts#getSwitchRequestData. However, Greninja-Bond (and any later edits to Zygarde or Rockruff) can't send its form in details, because it's visually indistinguishable from regular Greninja. There are a few solutions to this:

- Hardcode exceptions to skip over Greninja-Bond, etc.
- Add species name to pokemon.ts#getSwitchRequestData

This approach goes with the latter. I assume you would always know your own species at all times. Maybe there's an argument that you wouldn't always know your ally's species in a Multi Battle, but 1) we don't support multi battle hackmons where you could even theoretically bring a regular Greninja with Battle Bond, 2) I think it's a reasonable assumption that partner players in a Multi Battle should always know species names of their allies. (Hopefully I am not mistaken on what this function in fact does).

In testing, I was not able to remove the Pikachu / Pichu hardcodes when stress-testing by looping through 30 different RNG seeds. Battles would crash saying the player couldn't pass but needed to switch in a Pokemon.

### DIFF
--- a/sim/pokemon.ts
+++ b/sim/pokemon.ts
@@ -1061,7 +1061,6 @@ export class Pokemon {
 	getSwitchRequestData(forAlly?: boolean) {
 		const entry: AnyObject = {
 			ident: this.fullname,
-			species: this.species.name,
 			details: this.details,
 			condition: this.getHealth().secret,
 			active: (this.position < this.side.active.length),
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Talked it over with Marty, and we think it's best to just hardcode Greninja for now. Instead of fooling with main sim functions to make exhaustive-runner work, we'll fool with exhaustive-runner.

PR #9620, opened 2023-06-28
Fix Gen 4 Brick Break Ghost Type Interaction

jzfraser:
In Gen 4 Brick Break should break reflect, even on a ghost type.

Fixes https://github.com/smogon/pokemon-showdown/projects/3#card-61615516

First time trying to contribute here so criticism very welcome!

### DIFF
--- a/data/mods/gen4/moves.ts
+++ b/data/mods/gen4/moves.ts
@@ -182,6 +182,16 @@ export const Moves: import('../../../sim/dex-moves').ModdedMoveDataTable = {
 		inherit: true,
 		recoil: [1, 3],
 	},
+	brickbreak: {
+		inherit: true,
+		onTryHit() {},
+		onTryMove(source, target) {
+			if (!target.isProtected()) {
+				target.side.removeSideCondition('reflect');
+				target.side.removeSideCondition('light screen');
+			}
+		},
+	},
 	bulletseed: {
 		inherit: true,
 		basePower: 10,
--- a/data/mods/gen3/moves.ts
+++ b/data/mods/gen3/moves.ts
@@ -130,6 +130,7 @@ export const Moves: import('../../../sim/dex-moves').ModdedMoveDataTable = {
 			foe.removeSideCondition('reflect');
 			foe.removeSideCondition('lightscreen');
 		},
+		onTryMove() {},
 	},
 	charge: {
 		inherit: true,
--- a/test/sim/moves/brickbreak.js
+++ b/test/sim/moves/brickbreak.js
@@ -24,6 +24,20 @@ describe('Brick Break', () => {
 		assert.false(battle.p2.sideConditions['reflect']);
 	});
 
+	it('should not break Reflect when enemy uses Protect', () => {
+		battle = common.gen(4).createBattle([[
+			{ species: "mew", moves: ['brickbreak', 'splash'] },
+		], [
+			{ species: "ninjask", moves: ['reflect', 'splash', 'protect'] },
+		]]);
+
+		battle.makeChoices('move splash', 'move reflect');
+		assert(battle.p2.sideConditions['reflect']);
+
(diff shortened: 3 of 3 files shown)


data/mods/gen4/moves.ts
```diff
@@ -155,6 +155,12 @@ export const Moves: {[k: string]: ModdedMoveData} = {
 		inherit: true,
 		recoil: [1, 3],
 	},
+	brickbreak: {
+		inherit: true,
+		onTryMove(_attacker, defender, _move) {
+			defender.side.removeSideCondition('reflect');
+		},
```
AnnikaCodes, line 162:
```suggestion
		onTryMove(attacker, defender) {
			defender.side.removeSideCondition('reflect');
		},
```
Usually, PS either omits the names of unused arguments (if they're after the ones that are used) or keep them as-is with no underscore.

jzfraser: Fixed, thank you!

DaWoblefet (changes requested):
Sorry for the delay on reviewing this. You'll need to do a few extra things:

- Remove Light Screen as well. Brick Break into a Ghost-type will also break Light Screen.
- Override the event appropriately in each generation. Since we're using a different event, we want to make sure it's not being inherited accidentally. Here in the Gen 4 mod, you'll want to add:

``onTryHit() {}``

And in the Gen 3 mod, you'll want to add:
``onTryMove() {}``

However, you might need to find a different event other than onTryMove; that seems to be too early, because Brick Break will _not_ remove screens on Protect in Gen 4, but this change would let Brick Break do that. So it'd need to be something in Gen 4's checks done before move success _after_ Protect, but _before_ type immunities. If it'd help, I can push up another test for you to validate against, or you can copy the current Gen 4 test and just replace one of the Splash with Protect and change the last assert().

jzfraser:
No worries at all, thanks for the review and the explanation of what still needs to be done!

I can add the test to make sure it doesn't go past protect myself, and I'll take a look at the event options to keep the screen removal from happening too early.

jzfraser:
Alright, I was unable to find a different event that would still work the right way, so instead I just added a check to prevent this from happening if the target is protected. Let me know if this works or if you still think I need to look for a different event!

DaWoblefet:
> Also, the test says that the screens should be removed in Gen 4 _and earlier_, but the card just says this should happen in Gen 4. Do you know which is the case?

I tested in Gen 3 and it's the same behavior as in Gen 4 - using Brick Break into a Ghost-type will still break Reflect, and using it into Protect will not break Reflect.

As for the implementation, I'm still not a fan of using onTryMove(). Because of how events are arranged in Gen 4, lots of stuff can happen in the meantime, including, say, move accuracy. As it is, a missed Brick Break would break screens on this implementation when it should not. onTryHit seems to be the right event. I think it should probably be a sort of ``tryHitMovePriority`` and we assign it an earlier priority before Protect, like how Magic Bounce is handled. Looks like we can't currently assign priority directly to moves like that, though. I guess we could add that. Do you have any other ideas on implementation @Marty-D?

jzfraser:
I've been working on figuring out a way to get `onTryHit` to work against ghost types, but it seems like this happens after whichever event does the type checking, causing `onTryHit` to not be run at all.

Totally possible that I'm just not understanding how these events work, though.

jzfraser:
Started looking into this again but it looks like that mechanics bugs project no longer exists. Happy to give this a better shot but I'm not sure where it stands at this point. If I see a response on here I'll respond accordingly, otherwise I'll close this PR after some time.

jzfraser:
Also, the main problem I am finding in implementing this is that type immunity is checked a few steps before `onTryHit`. Because of that, another solution will be needed, either other than or additional to `onTryHit` and `onTryHitPriority`, as the `onTryHit` event is not run when the pokemon is immune, as far as I can tell.

PR #9621, opened 2023-06-29
Gen 1: Fix Inverse Mod

livid-washed:
My previous patch https://github.com/smogon/pokemon-showdown/pull/9054 inadvertently caused Inverse Mod to no longer work in Generation 1. This PR will fix that.

### DIFF
--- a/data/mods/gen1/scripts.ts
+++ b/data/mods/gen1/scripts.ts
@@ -903,7 +903,8 @@ export const Scripts: ModdedBattleScriptsData = {
 			// Type effectiveness.
 			// In Gen 1, type effectiveness is applied against each of the target's types.
 			for (const targetType of target.types) {
-				const typeMod = this.battle.dex.getEffectiveness(type, targetType);
+				let typeMod = this.battle.dex.getEffectiveness(type, targetType);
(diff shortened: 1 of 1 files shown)


data/mods/gen1/scripts.ts
```diff
@@ -903,7 +903,8 @@ export const Scripts: ModdedBattleScriptsData = {
 			// Type effectiveness.
 			// In Gen 1, type effectiveness is applied against each of the target's types.
 			for (const targetType of target.types) {
-				const typeMod = this.battle.dex.getEffectiveness(type, targetType);
+				let typeMod = this.battle.dex.getEffectiveness(type, targetType);
+				typeMod = this.battle.runEvent('Effectiveness', this.battle, targetType, move, typeMod);
```
AnnikaCodes, line 907: Does `getEffectiveness` have a side effect? If not, you're just overwriting the value with `runEvent`'s return value.

livid-washed: `getEffectiveness` outputs the (regular) type effectiveness; this is used as an input to `runEvent` which flips this value if Inverse Mod is active

PR #9640, opened 2023-07-07
Fix Illusion leak when disguised as Greninja-Bond

Karthik99999:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9694181


DaWoblefet:
This problem will also exist whenever we transition Zygarde and Rockruff over, but this seems to be the best way to do it for now. Thanks Karthik!

PR #9646, opened 2023-07-10
Add Pokemon GO support

dot-Comfey:


sora10pls:
Nice work on this! There are a lot of small legality quirks that are noted here not a lot of people knowing about (e.g. Lv. 1 Egg hatches post-XP buff).

I'd like to chime in with my own knowledge on GO legality, sourced from extensive research done by myself and Anubis over the course of a few months when GO Transporter was first released (and many years after).

* You mention that Giovanni's Shadow Pokémon are obtainable at a minimum of Lv. 11, which does not really matter. Shadow Pokémon are always encountered at Lv. 8, and do not scale according to your Trainer Level, like Eggs do (up to a maximum of 20). In addition to this, from our own testing, trading Purified Pokémon to lower-level accounts makes them eligible to down-level to the Level they were encountered at (Lv. 8).
* There are a number of edge cases where Pokémon normally obtainable with one method are very briefly made available via another method. Your current implementation does not account for these. Some examples include:
   * Alolan Raichu in wild encounters (Elite Raid Boss bonus encounters, Season 11: Hidden Gems in the northern hemisphere)
   * Hisuian Growlithe in research encounters (Season 11: Hidden Gems field research), and starting with GO Fest 2023: Global, in the wild
   * Latias and Latios in wild encounters (GO Snapshot photobombs, GO Tour: Hoenn rare wild encounters)
   * Deoxys obtained via Special Trade (for a brief period of time, Deoxys was allowed to be traded, making it eligible for sub-10 IVs)
   * Pawniard in research encounters (Level 43 Challenge)
   * Pumpkaboo in Eggs (for a brief period of time, Small Size Pumpkaboo was in the Halloween event Egg pool, before being removed)
   * Rockruff in wild encounters (Daily Adventure Incense)

I have not gone through your entire list for ommissions, but these are some notable ones I found and wanted to share. There are others that I found, and there could be more that I simply overlooked.

A project I actively maintain, [PoGoEncTool](https://github.com/projectpokemon/PoGoEncTool/), contains a database of all legal Pokémon encounters featured in GO, that can be transferred to HOME, including many obscure quirks and mistakes Niantic has made along the way. Its only current use is with PKHeX, as saving changes with the program will generate binaries containing all the possible encounters in a format PKHeX can use to parse for legality checking and encounter generation.

Since it is designed specifically for HOME connectivity, this data should suit you well for your own purposes. If you have any questions, feel free to reply and I'll share some insight when I can.

data/pokemongo.ts
```diff
@@ -0,0 +1,961 @@
+/**
```
AnnikaCodes, line 1: Normally we put encounter data in a `learnsets.ts` file — is there a reason you chose to split off Pokémon GO encounters into their own file?

KrisXV: I prefer the data being split from learnsets, the learnsets file is already a massive file on its own and is a pain to edit, so id prefer if we didnt have even more encounter legality data in there

DaWoblefet:
I don't have much to say about the implementation here (validator code is hard), but I think it would be good to have some tests written for Go legality, like level 1 Endeavor Donphan, etc. You can include those in ``test/sim/team-validator/misc.js``, or if you think a lot of tests would be good to handle multiple edge cases, refactoring to another file is good too.

sora10pls:
Noticed there are various future GO Fest 2023 encounters that are included in your database, while others are omitted. The event has not started yet, but will in a few weeks, so you should remain consistent with what you do and don't include. All of the relevant species are as follows...

First time available in the wild:
* Hisuian Growlithe
* Lucario
* Timburr
* Pawniard

First time available in Special Research:
* Diancie

New Shiny debuts:
* Unown M
* Unown !
* Shellos (all forms)
* Petilil
* Joltik
* Golett
* Skrelp
* Goomy
* Pumpkaboo (all forms)*
* Dewpider
* Oranguru

\*Shiny Pumpkaboo debuted during the 2022 Halloween event, but could only be encountered in a costume, and is not a standard wild encounter outside of Halloween, so it could not be transferred to HOME.

Everything listed above, sans Diancie, will be available in the wild (least restrictive encounter type).

Zygarde is also included in your database, but is currently disabled from being transferred to HOME. This restriction, as well as its trade restriction, will probably be removed once all of its story related content has been released, but should note it regardless.

DaWoblefet:
Thanks for your work here, dot Comfey!

PR #9648, opened 2023-07-11
Properly implement Chloroblast recoil

Karthik99999:
And by properly I mean the dumb way GF implemented this.

https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9699207
https://twitter.com/zakocky/status/1663886466525827072

### DIFF
--- a/test/sim/moves/chloroblast.js
+++ b/test/sim/moves/chloroblast.js
@@ -12,7 +12,7 @@ describe('Chloroblast', function () {
 
 	it('should deal recoil damage to the user equal to half its max HP, rounded up', function () {
 		battle = common.createBattle([[
-			{species: "Electrode-Hisui", ability: 'noguard', moves: ['chloroblast']},
+			{species: "Electrode-Hisui", item: 'widelens', moves: ['chloroblast']},
 		], [
 			{species: "Blissey", moves: ['sleeptalk']},
 		]]);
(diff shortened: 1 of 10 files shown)


DaWoblefet:
Do we have tests distinguishing this from Mind Blown, etc? If not I think those would be good to add.

test/sim/moves/chloroblast.js
```diff
@@ -0,0 +1,54 @@
...
+let battle;
+
+describe('Chloroblast', function () {
+	afterEach(function () {
+		battle.destroy();
+	});
+
+	it('should deal recoil damage to the user equal to half its max HP, rounded up', function () {
+		battle = common.createBattle([[
+			{species: "Electrode-Hisui", moves: ['chloroblast']},
```
DaWoblefet, line 15: Use more No Guard where possible to avoid misses in case RNG cases.

test/sim/moves/chloroblast.js
```diff
@@ -0,0 +1,54 @@
...
+		]]);
+		battle.makeChoices();
+		assert.fullHP(battle.p1.active[0]);
+	});
+
+	it('should not have its base power boosted by Reckless', function () {
+		battle = common.createBattle([[
+			{species: "Electrode-Hisui", ability: 'reckless', moves: ['chloroblast']},
+		], [
+			{species: "Blissey", moves: ['sleeptalk']},
```
DaWoblefet, line 48: Shell Armor + Wide Lens works fine for this one.

KrisXV:
Would this be better implemented in the calcRecoilDamage function itself?

Karthik99999:
> Would this be better implemented in the calcRecoilDamage function itself?

I originally didn't do this because it would add an extra parameter to calcRecoilDamage, but now that I realize substitute has to do recoil damage itself it's probably better to implement it in calcRecoilDamage.

Karthik99999:
> Did anyone ever get around to testing how this works with Parental Bond?

I just tested it and the recoil was still 50%.

DaWoblefet:
Thanks Karthik!

PR #9649, opened 2023-07-11
Gen4: Brick break now breaks reflect when hitting ghosts

singiamtel:
https://github.com/smogon/pokemon-showdown/projects/3#card-61615516

This fix is a bit awkward, but I couldn't find any weird side effects. It's the same way Order Up worked before #9429


DaWoblefet:
This is already in-progress: https://github.com/smogon/pokemon-showdown/pull/9620

The implementation there is closer to what we'd be looking for, so I'll just leave that one open for now.

singiamtel:
Oh, I didn't see the PR, my bad

PR #9652, opened 2023-07-12
fixed mega snorlax typing

socoolmudkip:


socoolmudkip:
my bad, meant to not merge on main lol

PR #9655, opened 2023-07-15
Fix Gen 1 Stadium Overflow

Tree69420:
On showdown in gen 1 stadium OU, when a pokemon's stat is over 1024, it overflows to a small number, this is not cart-accurate as stadium does not have overflows. This code caps stats at 999 to be cart-accurate.
https://www.smogon.com/forums/threads/stadium-bug-report-thread.3526616/post-9700865


Tree69420:
oh uh i didnt realize pull requests worked like this but I also fixed the interaction between sleep/freeze and counter. 
https://www.smogon.com/forums/threads/rby-tradebacks-bug-report-thread.3524844/post-9565241

DaWoblefet:
Thanks!

PR #9656, opened 2023-07-15
Monotype Random Battles: adjust type combination limit

livid-washed:
In Gens 4-5, this fully removes the type combination limit for Monotype Random Battles, since some types have very few possible type combinations. This should fix the infrequent crash that can occur while generating teams.

In Gens 6-9, the type combination limit has been increased from 2 to 3. The previous limit of 2 was too restrictive and could lead to problems with team generation.


ACakeWearingAHat (approved):

DaWoblefet:
I love it when crashes are fixed :) Thank you!

PR #9657, opened 2023-07-16
Fix , all in /as and /is

dot-Comfey:


AnnikaCodes:
I'm not sure how this helps, but it also seems good to use split instead of substr here.

PR #9595, opened 2023-06-10
Tournaments: Hide toolbox for players who has ended all battle

Dirain1700:
This PR has 2 changes:
・ Hide toolbox when eliminated players reconnected
・ Hide toolbox if the player have ended all battles in Round Robin

PR #9662, opened 2023-07-18
Tournaments: Show toolbox for reconnected players if the tournament was not started

Dirain1700:
Testing for my previous PR was not enough, sorry
Fix #9595


DaWoblefet:
Thanks Dirain1700!

PR #9679, opened 2023-07-24
Prevent Commanded Dondozo from switching out through pivot moves

HisuianZoroark:
The sequel to #9387 

As seen here: https://www.smogon.com/forums/threads/scarlet-violet-battle-mechanics-research.3709545/page-31#post-9538928

This also fixes the slight mechanics error regarding Baton Pass used to implement the previous Baton Pass with Commander failure.

### DIFF
--- a/data/moves.ts
+++ b/data/moves.ts
@@ -1084,7 +1084,7 @@ export const Moves: {[moveid: string]: MoveData} = {
 		pp: 40,
 		priority: 0,
 		flags: {},
-		onTryHit(target) {
+		onHit(target) {
 			if (!this.canSwitch(target.side) || target.volatiles['commanded']) {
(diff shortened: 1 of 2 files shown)


DaWoblefet:
Needing to have commanded and commander volatiles hardcoded everywhere kinda sucks. Is there a way we can move this to a common function associated with switching out?

HisuianZoroark:
> Needing to have commanded and commander volatiles hardcoded everywhere kinda sucks. Is there a way we can move this to a common function associated with switching out?

canSwitchOut() only takes side as an argument, maybe i can change that for an optional source parameter.

DaWoblefet:
For now, I will merge this, and it can be left as something for later to review.

PR #9682, opened 2023-07-26
Fix Ice Spinner removing terrain when fainting of recoil

adrivrie:
As per https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9716181.


DaWoblefet:
This doesn't fully fix the interactions with Ice Spinner in checks done after move success, but in general, we have pretty bad support for those right now so it's good to make the interaction slightly more correct.

Thanks adrivrie!

~~also one of these days I'll do a rebase correctly first time~~

PR #9683, opened 2023-07-26
Set Importer: Fix CAP Pokemon gens

adrivrie:
All CAP mons have a gen property in pokedex,ts, so we should use that.

The ranges were inaccurate, certainly because of a typo (-8 instead of -18 on line 157) but also otherwise. I think the reason is that the numbering for CAP changed at some point, though I can find no evidence in the history.

Anyway, hopefully this resolves it.


monsanto:
I'll try this patch next time I update the sets. Thanks

monsanto:
If there are any problems with this we'll have to revert.

PR #9685, opened 2023-07-29
Random Battles: July 2023 level balancing

adrivrie:
Monthly level balancing. For the first time includes Gen 8. Doubles is also generally included but will be in another merge request together with a general patch to avoid merge conflicts.

If possible, please merge and patch soonish to the month rollover.

Thank you :)


ACakeWearingAHat (approved):
this sure is a semiautomatic adjustment of countless integers so there's not much TO review

thanks as always, bud.

PR #9688, opened 2023-07-31
Movesearch: Remove empty spaces from 'sheer force' before validating

singiamtel:
This allows to run ``/ms sheer force``, same as how magic bounce works


DaWoblefet:
Looks good, thanks singiamtel!

PR #9689, opened 2023-08-01
Fix Gen 9 Hackmons legality

KrisXV:
https://www.smogon.com/forums/posts/9713378
https://www.smogon.com/forums/posts/9716092/

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -537,9 +537,12 @@ export const Rulesets: {[k: string]: FormatData} = {
 		onTeamPreview() {
 			this.add('clearpoke');
 			for (const pokemon of this.getAllPokemon()) {
-				const details = pokemon.details.replace(', shiny', '')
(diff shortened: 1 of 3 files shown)


data/rulesets.ts
```diff
@@ -2456,4 +2459,51 @@ export const Rulesets: {[k: string]: FormatData} = {
...
+				}
+			}
+			return problems;
+		},
+	},
+	speciesrevealclause: {
+		effectType: 'Rule',
+		name: 'Species Reveal Clause',
+		desc: "Reveals a Pok&eacute;mon's true species in hackmons-based metagames.",
+		// Hardcoded into effect, cannot be disabled.
```
DaWoblefet, line 2504: I would add a comment that it ties into Team Preview as well.

DaWoblefet:
Currently, BH has requested the current implementation of Hackmons be kept for the purposes of seasonal play. I would eventually like to see most of this logic moved to the validator once that no longer needs support. In the meantime, this implementation all functions well and should let us avoid any surprises with other OMs.

Thanks Kris!

PR #9692, opened 2023-08-02
Revamp [Gen 7] Random Battle

ACakeWearingAHat:
Framework by @livid-washed 
random-teams filled out by me
sets created by the Random Battles room staff, with special thanks to Bobomania

This is finished and ready for review.

This pull request completely overhauls Gen 7 Random Battles to utilize the more modern and streamlined Gen 9 Random Battles setgen system, allowing us to have split sets, frontloaded move generation, and role-based move and item enforcement logic.

This will massively change how the format plays.


data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +30,666 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+		moves: Set<string>,
+		types: string[],
+		abilities: Set<string>,
+		teamDetails: RandomTeamsTypes.TeamDetails,
+		species: Species,
+		isLead: boolean,
+		isDoubles: boolean,
+		movePool: string[],
+		preferredType: string,
+		role: string,
```
AnnikaCodes, line 419: Why not add a proper type for roles? That would catch typos made in the names of roles.

data/mods/gen7/random-teams.ts
```diff
@@ -602,121 +701,105 @@ export class RandomGen7Teams extends RandomGen8Teams {
...
+			// Slam part is for Tauros
+			return (moves.has('bodyslam') || species.id === 'staraptor');
+		case 'Iron Fist':
+			// Dynamic Punch bit is for Golurk
+			return (!counter.get(toID(ability)) || moves.has('dynamicpunch'));
 		case 'Lightning Rod':
 			return (
-				species.types.includes('Ground') ||
-				(!!teamDetails.rain || moves.has('raindance')) && abilities.has('Swift Swim')
+				types.has('Ground') || species.id === 'marowakalola' ||
```
AnnikaCodes, line 745: In this method, there's an odd mixture of hardcoded species IDs and broader conditions that are noted in comments as being specific to one species.

data/mods/gen7/random-teams.ts
```diff
@@ -742,31 +827,47 @@ export class RandomGen7Teams extends RandomGen8Teams {
...
+		if (species.id === 'druddigon' && role === 'Bulky Support') return 'Rough Skin';
+		if (abilities.has('Scrappy') && moves.has('boomburst')) return 'Scrappy';
+		if (abilities.has('Shed Skin') && moves.has('rest') && !moves.has('sleeptalk')) return 'Shed Skin';
+		if (abilities.has('Sniper') && moves.has('focusenergy')) return 'Sniper';
+		if (species.name === 'Kommo-o' && role === 'Z-Move user') return 'Soundproof';
+		if (species.id === 'stunfisk') return 'Static';
+		if (species.id === 'breloom') return 'Technician';
+		if (abilities.has('Toxic Boost')) return 'Toxic Boost';
+		if (species.id === 'porygon2') return 'Trace';
+		if (types.has('Water') && counter.get('setup') && abilities.has('Weak Armor')) return 'Weak Armor';
```
AnnikaCodes, line 864: This also seems like something that could be broken out into top-level constants to make this easier to read.

data/mods/gen7/random-teams.ts
```diff
@@ -807,319 +908,180 @@ export class RandomGen7Teams extends RandomGen8Teams {
...
+		if (role === 'Z-Move user') {
+			// Specific Z-Crystals
+			if (species.baseSpecies === 'Arceus' && species.requiredItems) return species.requiredItems[1];
+			if (species.name === 'Raichu-Alola') return 'Aloraichium Z';
+			if (species.name === 'Decidueye') return 'Decidium Z';
+			if (species.name === 'Kommo-o') return 'Kommonium Z';
+			if (species.name === 'Lunala') return 'Lunalium Z';
+			if (species.baseSpecies === 'Lycanroc') return 'Lycanium Z';
+			if (species.name === 'Marshadow') return 'Marshadium Z';
+			if (species.name === 'Mew') return 'Mewnium Z';
```
AnnikaCodes, line 932: Possibly this could be simplified by iterating over all Z-crystals and checking `itemUser`, but this way is probably more performant (albeit potentially trickier to maintain).

test/random-battles/all-gens.js
```diff
@@ -17,10 +17,13 @@ describe('value rule support (slow)', () => {
 
 		testTeam({format: 'gen8multirandombattle', rounds: 100}, team => assert.equal(team.length, 3));
 		testTeam({format: 'gen8cap1v1', rounds: 100}, team => assert.equal(team.length, 3));
+
+		testTeam({format: 'gen7randombattle', rounds: 100}, team => assert.equal(team.length, 6));
 	});
 
 	// TODO: Support gen 9 set format
 	for (let gen = 1; gen <= 8; gen++) {
+		if (gen === 7) continue;
```
AnnikaCodes, line 26: Why are we exempting Gen 7 from these tests? Does this revamp not support Max Move Count?

test/random-battles/gen7.js
```diff
@@ -4,29 +4,94 @@
...
+					if (set.preferredTypes) {
+						for (const type of set.preferredTypes) {
+							const dexType = Dex.types.get(type);
+							assert(dexType.exists, `${species.name} has invalid Preferred Type: ${type}`);
+							assert.equal(type, dexType.name, `${species.name} has misformatted Preferred Type: ${type}`);
+						}
+						for (let i = 0; i < set.preferredTypes.length - 1; i++) {
+							assert(set.preferredTypes[i + 1] > set.preferredTypes[i], `${species} preferredTypes should be sorted alphabetically`);
+						}
+					}
```
AnnikaCodes, line 44: Good work! I'm glad we have tests to validate the team data.

data/mods/gen6/random-teams.ts
```diff
@@ -1136,14 +1142,10 @@ export class RandomGen6Teams extends RandomGen7Teams {
 			moves.add('thunder');
 		}
 
-		if (species.battleOnly && !species.requiredAbility) {
```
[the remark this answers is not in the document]
livid-washed: Yes, it correctly generates abilities based on the base species. In a previous PR (https://github.com/smogon/pokemon-showdown/pull/9642) we refactored ability selection to an individual method `getAbility()`, which takes care of megas and their base species and made this line no longer necessary. I just forgot to remove it.

data/mods/gen2/random-teams.ts
```diff
@@ -115,9 +115,10 @@ export class RandomGen2Teams extends RandomGen3Teams {
 
 	getItem(
 		ability: string,
-		types: Set<string>,
+		types: string[],
```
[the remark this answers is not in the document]
livid-washed: Mainly because I wanted `getItem()` to be consistent with gen 9, no real reason otherwise. While it doesn't matter for item generation, type order can matter for other things like determining the type of Revelation Dance (which in turn affects move enforcement, etc)

test/random-battles/all-gens.js
livid-washed: Due to the frontloading of moveset generation, it doesn't support Max Move Count if it is set lower than 4, as some sets have more than one 'enforced' move.

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +30,666 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	randomSets: AnyObject = require('./random-sets.json');
 
 	constructor(format: Format | string, prng: PRNG | PRNGSeed | null) {
 		super(format, prng);
 
-		this.noStab = [...this.noStab, 'voltswitch'];
+		if (this.format.gameType === 'singles') this.noStab = NoStab;
 
-		this.moveEnforcementCheckers = {
+		this.moveEnforcementCheckers = (this.format.gameType !== 'singles') ? this.moveEnforcementCheckers : {
```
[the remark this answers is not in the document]
livid-washed: I think the team generator for both singles and doubles ultimately needs to be exported in a single class in `random-teams.ts`, and having gen 7 singles inherit from gen 7 doubles, which in turn inherited from gen 8 was the best way I could think of to achieve this. I'm open to alternative solutions.

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +30,666 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+			if (moves.size + movePool.length <= this.maxMoveCount) return;
+		}
+		if (teamDetails.spikes && teamDetails.spikes >= 2) {
+			if (movePool.includes('spikes')) this.fastPop(movePool, movePool.indexOf('spikes'));
+			if (moves.size + movePool.length <= this.maxMoveCount) return;
+		}
+
+		// Develop additional move lists
+		const pivotingMoves = ['partingshot', 'uturn', 'voltswitch'];
+		const badWithSetup = ['defog', 'dragontail', 'haze', 'healbell', 'nuzzle', 'pursuit', 'rapidspin', 'toxic'];
```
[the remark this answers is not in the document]
livid-washed: `badWithSetup` is an array that only makes sense in the context of move incompatibilites, it is not fundamental enough to put as a top-level constant. I've moved `pivotingMoves` to the top, here and in gen 9.

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +30,666 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+		this.incompatibleMoves(moves, movePool, 'hornleech', 'woodhammer');
+		this.incompatibleMoves(moves, movePool, ['gigadrain', 'leafstorm'], ['leafstorm', 'petaldance', 'powerwhip']);
+		this.incompatibleMoves(moves, movePool, 'wildcharge', 'thunderbolt');
+		this.incompatibleMoves(moves, movePool, 'gunkshot', 'poisonjab');
+		this.incompatibleMoves(moves, movePool, ['drainpunch', 'focusblast'], ['closecombat', 'highjumpkick', 'superpower']);
+		this.incompatibleMoves(moves, movePool, 'stoneedge', 'headsmash');
+		this.incompatibleMoves(moves, movePool, 'dracometeor', 'dragonpulse');
+		this.incompatibleMoves(moves, movePool, 'dragonclaw', 'outrage');
+		this.incompatibleMoves(moves, movePool, 'knockoff', ['darkestlariat', 'darkpulse', 'foulplay']);
+		if (!types.includes('Dark') && preferredType !== 'Dark') {
```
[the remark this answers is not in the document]
livid-washed: I think these are more suited to be in the body of the method, especially since many of the incompatibilities are for specific Pokemon rather than all Pokemon in general. If anything, I think it would be more confusing to have even more long objects/arrays at the top of the file, I'd prefer for top-level constants to be only for the most general-purpose things. From experience in gen 9, having them in the body does not make them any harder to modify.

data/mods/gen7/random-teams.ts
livid-washed: We have used hardcoded `species.id` checks when necessary - however, if there was no need or benefit to have those `species.id` checks, we did not include them, and instead just made a note of it. For example, we want to reject No Guard on Golurk sets with Shadow Punch (but not its other set with Dynamic Punch), so we rejected No Guard with Shadow Punch. There is no need to have a `species.id` check since Golurk is the only Pokemon that can have both.

data/mods/gen2/random-teams.ts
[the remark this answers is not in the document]
AnnikaCodes: I suppose that's fine.

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
AnnikaCodes: It's kind of clunky (and feels like it defeats some of the point of separating them) but I suppose that's fine; I forgot that they are expected to be together. An alternative would be to modify the code that calls the team generator to get singles and doubles teams from different classes for Gen 7, so the classes could be properly separate. This is similar to what computer generated teams did.

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
AnnikaCodes: What do you mean by "for specific Pokemon"? The ones I grabbed for the review comment don't have any indication that they're targeted at certain species. Anyway, the benefit is you wouldn't have repeat `this.incompatibleMoves(moves, movePool,` all the time, since you could just iterate over the incompatibility array.

data/mods/gen7/random-teams.ts
AnnikaCodes: PS contributing guidelines [generally prefer](https://github.com/smogon/pokemon-showdown/blob/master/CONTRIBUTING.md#c2-document-in-names-if-possible) that we document things within the code rather than through comments, if possible.

I think what you describe is acceptable as long as the move/ability/etc checks actually matter within the species and aren't just a stand-in for checking the species (and wouldn't be something you care about on other species).

test/random-battles/all-gens.js
AnnikaCodes: Then why not add a generation-specific check to skip testing only those `Max Move Count`s lower than 4 in Gen 7, similar to what Gen 1 has? That seems better than bypassing the test altogether.

livid-washed: I've adjusted the Max Move Count tests to include gen 7 and gen 9

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
livid-washed: That makes sense, I went ahead and put the general incompatibilities in its own array. Since it's only relevant in this single context, I think it still belongs in the method and not at the top of the file.

data/mods/gen7/random-teams.ts
livid-washed: In this case, I disagree. There's too much variety in ability enforcement conditions and I think going through it line by line is better than trying to use multiple constants to capture them. I think this is also easier for maintenance if we ever need to make small changes to enforcement conditions.

data/mods/gen7/random-teams.ts
livid-washed: Agreed, I think the list is short enough that this is more performant, especially since some Z-Move users with specific Z-crystals opt not to use them, like Tapu Koko.

data/mods/gen7/random-teams.ts
livid-washed: I've created a type for roles

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
livid-washed: I've implemented your suggestion and made the required changes to `sim/teams.ts`

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
AnnikaCodes: OK; having the data in the method seems a bit counterintuitive to me but if it's beneficial to randbats coders I suppose it's fine.

I still don't understand why the incompatibilies ostensibly for particular species aren't in the array; it might be faster to only run those if you're generating for a particular species, but you run `incompatibleMoves` on them regardless of species! Is 
```ts
// Lunatone
this.incompatibleMoves(moves, movePool, 'moonlight', 'rockpolish');
```
somehow better than a commented entry in the array:
```ts
const incompatiblePairs = [
    ...,
    // For Lunatone
    ['moonlight', 'rockpolish'],
    ...,
];
```

data/mods/gen7/random-teams.ts
AnnikaCodes: OK. I do think this would more readable if it were grouped by species-based checks, role-based checks, and ability-based checks. Right now they're all interspersed. Also, is there a reason the Z-move check for Kommo-o uses the species's English name instead of its ID?

data/mods/gen7/random-teams.ts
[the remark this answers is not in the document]
livid-washed: Sure, I've moved the applicable hardcodes into the array

data/mods/gen7/random-teams.ts
livid-washed: I've separated the ability enforcements into species-based and non species-based checks. Some of the species-based checks also check other things, but I think this is quite readable. While the non species-based checks aren't too long (only 8 lines), so I don't think grouping it further would be beneficial.

For the second question, no reason. I've changed Kommo-o and other species names checks in this section like Ambipom to species ID checks.

data/mods/gen6/random-teams.ts
```diff
@@ -4,12 +4,19 @@ import {PRNG, PRNGSeed} from '../../../sim/prng';
 import {Utils} from '../../../lib';
 import {toID} from '../../../sim/dex';
 
+// Moves that shouldn't be the only STAB moves:
+const NoStab = [
```
mia-pi-git, line 8: `camelCase` :(

data/mods/gen8/random-teams.ts
```diff
@@ -1691,6 +1693,8 @@ export class RandomGen8Teams {
 		teamDetails: RandomTeamsTypes.TeamDetails,
 		species: Species,
 		isDoubles: boolean,
+		preferredType = '',
```
mia-pi-git, line 1696: Why have these default to `''` when the type after them is a required type? You're always going to be specifying them anyways, so it seems like it'd be better to just make the type explicit instead?

data/random-teams.ts
```diff
@@ -102,6 +102,10 @@ const Hazards = [
 const ProtectMove = [
 	'banefulbunker', 'protect', 'spikyshield',
 ];
+// Moves that switch the user out
+const PivotingMoves = [
```
mia-pi-git, line 106: `camelCase`

server/chat-plugins/randombattles/index.ts
```diff
@@ -173,6 +174,21 @@ function getData(species: string | Species, format: string | Format): any | null
 	return data;
 }
 
+/**
+ * Gets the random battles data for a Pokemon for doubles where doubles is its own file
+ * but it still uses the old system. Currently only gen7dubs.
+ */
+function getStandaloneDoublesData(species: string | Species, format: string | Format): any | null {
```
mia-pi-git, line 181: Name's a bit confusing - maybe roll this functionality into one function for all of doubles?

server/chat-plugins/randombattles/index.ts
```diff
@@ -785,21 +801,39 @@ export const commands: Chat.ChatCommands = {
 			setCount = 1;
 		} else {
 			const setsToCheck = [species];
-			if (dex.gen > 7 && !isNoDMax) setsToCheck.push(dex.species.get(`${args[0]}gmax`));
+			if (dex.gen >= 8 && !isNoDMax) setsToCheck.push(dex.species.get(`${args[0]}gmax`));
 			if (species.otherFormes) setsToCheck.push(...species.otherFormes.map(pkmn => dex.species.get(pkmn)));
-			if (dex.gen >= 9) {
+			if (dex.gen >= 9 || dex.gen === 7 && !isDoubles) {
```
mia-pi-git, line 806:
```suggestion
			if (dex.gen >= 9 || (dex.gen === 7 && !isDoubles)) {
```

server/chat-plugins/randombattles/index.ts
```diff
@@ -965,13 +999,15 @@ export const commands: Chat.ChatCommands = {
...
+		if (dex.gen >= 9 || dex.gen === 7 && format.gameType !== 'doubles') {
+			setExists = !!getSets(species, format);
+		} else if (dex.gen === 7 && format.gameType === 'doubles') {
+			setExists = !!getStandaloneDoublesData(species, format);
 		} else {
 			const data = getData(species, format);
 			if (!data) {
 				setExists = false;
-			} else if (format.gameType === 'doubles' || format.gameType === 'freeforall') {
+			} else if (format.gameType === 'doubles' && dex.gen !== 7 || format.gameType === 'freeforall') {
```
mia-pi-git, line 1010:
```suggestion
			} else if ((format.gameType === 'doubles' && dex.gen !== 7) || format.gameType === 'freeforall') {
```

data/random-teams.ts
AnnikaCodes: I think these top-level constants could just as well be `SHOUTING_SNAKE_CASE`?

data/mods/gen8/random-teams.ts
livid-washed: Agreed, fixed

test/random-battles/gen7.js
```diff
@@ -4,29 +4,94 @@
...
 
-	it('All moves on all sets should be obtainable', () => {
-		const rounds = 500;
-		for (const pokemon of Object.keys(dataJSON)) {
+	describe("New set format", () => {
+		const filename = '../../data/mods/gen7/random-sets.json';
+		it(`${filename} should have valid set data`, () => {
+			const setsJSON = require(filename);
+			const validRoles = ["Fast Attacker", "Setup Sweeper", "Wallbreaker", "Z-Move user",
+				"Bulky Attacker", "Bulky Setup", "Staller", "Bulky Support", "Fast Support", "AV Pivot"];
```
KrisXV, line 19:
```suggestion
			const validRoles = [
				"Fast Attacker", "Setup Sweeper", "Wallbreaker", "Z-Move user", "Bulky Attacker",
				"Bulky Setup", "Staller", "Bulky Support", "Fast Support", "AV Pivot",
			];
```

data/mods/gen6/random-teams.ts
```diff
@@ -4,12 +4,19 @@ import {PRNG, PRNGSeed} from '../../../sim/prng';
 import {Utils} from '../../../lib';
 import {toID} from '../../../sim/dex';
 
+// Moves that shouldn't be the only STAB moves:
+const noStab = [
```
KrisXV, line 8:
```suggestion
const NO_STAB = [
```

data/mods/gen6/random-teams.ts
```diff
@@ -4,12 +4,19 @@ import {PRNG, PRNGSeed} from '../../../sim/prng';
...
+	'skydrop', 'snarl', 'suckerpunch', 'uturn', 'watershuriken', 'vacuumwave', 'voltswitch', 'waterspout',
+];
+
 export class RandomGen6Teams extends RandomGen7Teams {
 	randomData: {[species: string]: OldRandomBattleSpecies} = require('./random-data.json');
 
 	constructor(format: Format | string, prng: PRNG | PRNGSeed | null) {
 		super(format, prng);
-		this.noStab = [...this.noStab, 'aquajet', 'fakeout', 'iceshard', 'machpunch', 'quickattack', 'vacuumwave'];
+		this.noStab = noStab;
```
KrisXV, line 19:
```suggestion
		this.noStab = NO_STAB;
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
 	rock: {def: 30, spd: 30, spe: 30},
 };
 
+// Moves that restore HP:
+const RecoveryMove = [
```
KrisXV, line 33:
```suggestion
const RECOVERY_MOVES = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'healorder', 'milkdrink', 'moonlight', 'morningsun', 'recover', 'recycle', 'roost', 'shoreup', 'slackoff', 'softboiled', 'strengthsap', 'synthesis',
+];
+// Moves that drop stats:
+const ContraryMoves = [
```
KrisXV, line 37:
```suggestion
const CONTRARY_MOVES = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'closecombat', 'leafstorm', 'overheat', 'superpower', 'vcreate',
+];
+// Moves that boost Attack:
+const PhysicalSetup = [
```
KrisXV, line 41:
```suggestion
const PHYSICAL_SETUP = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'bellydrum', 'bulkup', 'coil', 'curse', 'dragondance', 'honeclaws', 'howl', 'meditate', 'poweruppunch', 'screech', 'swordsdance',
+];
+// Moves which boost Special Attack:
+const SpecialSetup = [
```
KrisXV, line 45:
```suggestion
const SPECIAL_SETUP = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'calmmind', 'chargebeam', 'geomancy', 'nastyplot', 'quiverdance', 'tailglow',
+];
+// Moves that boost Attack AND Special Attack:
+const MixedSetup = [
```
KrisXV, line 49:
```suggestion
const MIXED_SETUP = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'celebrate', 'growth', 'happyhour', 'holdhands', 'shellsmash', 'workup',
+];
+// Some moves that only boost Speed:
+const SpeedSetup = [
```
KrisXV, line 53:
```suggestion
const SPEED_SETUP = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'agility', 'autotomize', 'flamecharge', 'rockpolish',
+];
+// Conglomerate for ease of access
+const Setup = [
```
KrisXV, line 57:
```suggestion
const SETUP = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'acidarmor', 'agility', 'autotomize', 'bellydrum', 'bulkup', 'calmmind', 'celebrate', 'coil', 'conversion', 'curse', 'dragondance',
+	'electricterrain', 'flamecharge', 'focusenergy', 'geomancy', 'growth', 'happyhour', 'holdhands', 'honeclaws', 'howl', 'irondefense', 'meditate',
+	'nastyplot', 'poweruppunch', 'quiverdance', 'raindance', 'rockpolish', 'shellsmash', 'shiftgear', 'swordsdance', 'tailglow', 'workup',
+];
+// Moves that shouldn't be the only STAB moves:
+const NoStab = [
```
KrisXV, line 63:
```suggestion
const NO_STAB = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'accelerock', 'aquajet', 'bulletpunch', 'clearsmog', 'dragontail', 'eruption', 'explosion',
+	'fakeout', 'firstimpression', 'flamecharge', 'futuresight', 'iceshard', 'icywind', 'incinerate', 'machpunch', 'nuzzle',
+	'pluck', 'poweruppunch', 'pursuit', 'quickattack', 'rapidspin', 'reversal', 'selfdestruct', 'shadowsneak', 'skyattack',
+	'skydrop', 'snarl', 'suckerpunch', 'uturn', 'watershuriken', 'vacuumwave', 'voltswitch', 'waterspout',
+];
+// Hazard-setting moves
+const Hazards = [
```
KrisXV, line 70:
```suggestion
const HAZARDS = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'spikes', 'stealthrock', 'stickyweb', 'toxicspikes',
+];
+// Protect and its variants
+const ProtectMove = [
```
KrisXV, line 74:
```suggestion
const PROTECT_MOVES = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'banefulbunker', 'kingsshield', 'protect', 'spikyshield',
+];
+// Moves that switch the user out
+const PivotingMoves = [
```
KrisXV, line 78:
```suggestion
const PIVOT_MOVES = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+	'partingshot', 'uturn', 'voltswitch',
+];
+
+// Moves that should be paired together when possible
+const MovePairs = [
```
KrisXV, line 83:
```suggestion
const MOVE_PAIRS = [
```

data/mods/gen7/random-teams.ts
```diff
@@ -29,568 +29,678 @@ const ZeroAttackHPIVs: {[k: string]: SparseStatsTable} = {
...
+const priorityPokemon = [
+	'aegislashblade', 'banette', 'breloom', 'cacturne', 'doublade', 'dusknoir', 'golisopod', 'honchkrow', 'mimikyu', 'scizor', 'scizormega', 'shedinja',
+];
+function sereneGraceBenefits(move: Move) {
+	return move.secondary?.chance && move.secondary.chance >= 20 && move.secondary.chance < 100;
+}
+
 export class RandomGen7Teams extends RandomGen8Teams {
-	randomData: {[species: string]: OldRandomBattleSpecies} = require('./random-data.json');
+	randomSets: AnyObject = require('./random-sets.json');
```
KrisXV, line 102: I would prefer a type declaration for this

data/random-teams.ts
KrisXV: i agree with annika

KrisXV (changes requested):
ill take another look once these get addressed

data/mods/gen7/random-teams.ts
ACakeWearingAHat: this'll all require some cleanup but it shouldn't be too hard to fix it

data/random-teams.ts
ACakeWearingAHat: we'll make these adjustments in the gen 9 file for gen 9's next current-gen set update, alongside the arrayification of the move incompatibility list and some set updates. This update will occur this weekend.

ACakeWearingAHat: resolving because i have confirmed that annika does not want this to happen in this pr

server/chat-plugins/randombattles/index.ts
livid-washed: Since gen 7 randdubs will continue using the old system, I believe that folding it into `getData()` would be the simplest way to go about it.

KrisXV (approved):

ACakeWearingAHat:
We would like this onsite as soon as is feasible.

PR #9698, opened 2023-08-05
Fix Go legality in formats without obtainablemoves

dot-Comfey:
https://www.smogon.com/forums/threads/bug-reports-v4-read-original-post-before-posting.3663703/post-9727878 This was happening as a result of move sources being generated for all formats rather than only ones that check for move legality.

In addition, I noticed that my implementation allowed Gen 6+ legendaries to get away with illegal moves if their only Pokemon GO identifier was having less than three perfect IVs; now the code should be performing legality checks only after all identifiers are checked.

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -823,11 +823,11 @@ export class TeamValidator {
 		}
 
 		// Attempt move validation again after verifying Pokemon GO origin
-		if (setSources.isFromPokemonGo) {
+		if (ruleTable.has('obtainablemoves') && setSources.isFromPokemonGo) {
 			setSources.restrictiveMoves = [];
 			setSources.sources = ['8V'];
 			setSources.sourcesBefore = 0;
-			if (moveProblems && !moveProblems.length && ruleTable.has('obtainablemoves')) {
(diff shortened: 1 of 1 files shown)


sim/team-validator.ts
```diff
@@ -832,6 +822,17 @@ export class TeamValidator {
 			problems.push(...this.validateStats(set, species, setSources, pokemonGoProblems));
 		}
 
+		// Attempt move validation again after verifying Pokemon GO origin
+		if (setSources.isFromPokemonGo) {
```
DaWoblefet, line 826: I think you can move the obtainablemoves check up here rather than later in the if statement - it makes it clear that this won't run if it's Custom Game etc.

DaWoblefet: Or maybe validateMoves() should just always return early if obtainablemoves isn't in the ruleTable? That would catch this problem from occurring in the future as well.

KrisXV:
this looks good to me, @DaWoblefet feel free to merge if you agree

DaWoblefet:
Thanks dot-Comfey!

PR #9699, opened 2023-08-05
Frantic Fusions: Fix validation errors

HisuianZoroark:
Automatically including !obtainable abilities seems to be a source of the errors, but also Nickname Clause doesn't turn "off" when removed in a part of the hardcode, allowing nickname clause to be removed fixes that issue about nicknaming pokemon in ruleset mods

### DIFF
--- a/sim/team-validator.ts
+++ b/sim/team-validator.ts
@@ -896,7 +896,8 @@ export class TeamValidator {
 			// nickname is the name of a species
 			if (nameSpecies.baseSpecies === species.baseSpecies) {
 				set.name = species.baseSpecies;
-			} else if (nameSpecies.name !== species.name && nameSpecies.name !== species.baseSpecies && ruleTable.has('nicknameclause')) {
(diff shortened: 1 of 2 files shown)


sim/team-validator.ts
```diff
@@ -896,7 +896,8 @@ export class TeamValidator {
 			// nickname is the name of a species
 			if (nameSpecies.baseSpecies === species.baseSpecies) {
 				set.name = species.baseSpecies;
-			} else if (nameSpecies.name !== species.name && nameSpecies.name !== species.baseSpecies) {
+			} else if (nameSpecies.name !== species.name && nameSpecies.name !== species.baseSpecies &&
+          ruleTable.has('nicknameclause')) {
```
KrisXV, line 900:
```suggestion
			} else if (nameSpecies.name !== species.name &&
				nameSpecies.name !== species.baseSpecies && ruleTable.has('nicknameclause')) {
```

PR #9700, opened 2023-08-05
`/fuse`: Allow providing one argument

Seerd:
Let you check the stats/abilities each Pokémon gives.

### DIFF
--- a/server/chat-plugins/othermetas.ts
+++ b/server/chat-plugins/othermetas.ts
@@ -574,7 +574,7 @@ export const commands: Chat.ChatCommands = {
 		this.sendReply(`|raw|${buf}`);
 	},
 	franticfusionshelp: [
-		`/fuse <pokemon>, <fusion>[, generation] - Shows the stats and abilities that <pokemon> would get when fused with <fusion>.`,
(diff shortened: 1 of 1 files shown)


server/chat-plugins/othermetas.ts
```diff
@@ -559,7 +574,7 @@ export const commands: Chat.ChatCommands = {
 		this.sendReply(`|raw|${buf}`);
 	},
 	franticfusionshelp: [
-		`/fuse <pokemon>, <fusion>[, generation] - Shows the stats and abilities that <pokemon> would get when fused with <fusion>.`,
+		`/fuse <pokemon>, <fusion>[, generation] - Shows the stats and abilities that <pokemon> would get when fused with <fusion>, if only one Pok\u00e9mon is ingressed it shows the values it would add.`,
```
KrisXV, line 577:
```suggestion
		`/fuse <pokemon>, <fusion>[, generation] - Shows the stats and abilities that <pokemon> would get when fused with <fusion>.`,
		`/fuse <pokemon>[, generation] - Shows the stats and abilities that <pokemon> donates.`,
```

PR #9701, opened 2023-08-06
Add Tera Switch Mod

Seerd:
For the forgotten OM: https://www.smogon.com/forums/threads/tera-switch.3715743/
Needs to fix visual bug that prevents Tera icon/type from disappearing when going back to original types, plz help.

### DIFF
--- a/data/rulesets.ts
+++ b/data/rulesets.ts
@@ -2596,4 +2596,26 @@ export const Rulesets: {[k: string]: FormatData} = {
 			return newSpecies;
 		},
 	},
+	teraswitchmod: {
+		effectType: 'Rule',
+		name: "Tera Switch Mod",
+		desc: `The Pokémon automatically switches in and out of its base typing and its designated Tera typing at the end of each turn .`,
+		onBegin() {
+			this.add('rule', 'Tera Switch Mod: The Pokémon automatically switches in and out of its base typing and its designated Tera typing at the end of each turn ');
+		},
+		onResidualOrder: 29,
+		onResidual(pokemon) {
+			if (!pokemon.terastallized) {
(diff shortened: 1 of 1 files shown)


MathyFurret:
That's because there isn't a message defined to remove Tera yet. Necessecity's the mother of invention, after all.

KrisXV:
necessecity

data/rulesets.ts
```diff
@@ -2596,4 +2596,25 @@ export const Rulesets: {[k: string]: FormatData} = {
 			return newSpecies;
 		},
 	},
+	teraswitchmod: {
+		effectType: 'Rule',
+		name: "Tera Switch Mod",
+		desc: `The Pokémon automatically switches in and out of its base typing and its designated Tera typing at the end of each turn .`,
```
AnnikaCodes, line 2602:
```suggestion
		desc: `The Pokémon automatically switches in and out of its base typing and its designated Tera typing at the end of each turn.`,
```

Seerd:
Have been trying some stuff without luck, can't figure out how to fix that visual bug.

MathyFurret:
ok so what you would *actually* need to do is make a new message type that tells the client to clear a tera. something like `-unterastallize|POKEMON`

there are 3 parts to this:
* document the message in SIM-PROTOCOL.md (apparently optional because `-terastallize` hasn't even been added yet loool)
* on the client, around [here](https://github.com/smogon/pokemon-showdown-client/blob/master/src/battle.ts#L2509), add a case for your new message that undoes all the stuff that tera does. there's the log text aspect too
* send the message whenever you un-tera (with `this.add`)

Seerd:
Thanks, now it does update the types every turn.

